PR #433, opened 2020-07-20
Show full payload in source debugger

prabrishac:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/services/source-debugger/eventUploader.go
+++ b/services/source-debugger/eventUploader.go
@@ -22,25 +22,11 @@ type GatewayEventBatchT struct {
 	eventBatch string
 }
 
-//EventUploadT is a structure to hold actual event data
-type EventUploadT struct {
-	Event             string      `json:"event,omitempty"`
-	Integrations      interface{} `json:"integrations,omitempty"`
-	Properties        interface{} `json:"properties,omitempty"`
-	OriginalTimestamp string      `json:"originalTimestamp,omitempty"`
-	AnonymousID       string      `json:"anonymousId,omitempty"`
-	UserID            string      `json:"userId,omitempty"`
-	MessageID         string      `json:"messageId,omitempty"`
-	Name              string      `json:"name,omitempty"`
-	SentAt            string      `json:"sentAt,omitempty"`
-	Type              string      `json:"type,omitempty"`
-}
-
 //EventUploadBatchT is a structure to hold batch of events
 type EventUploadBatchT struct {
 	WriteKey   string
 	ReceivedAt string
-	Batch      []EventUploadT
+	Batch      []map[string]interface{}
 }
 
 var uploadEnabledWriteKeys []string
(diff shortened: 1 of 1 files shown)


services/source-debugger/eventUploader.go
```diff
@@ -22,25 +22,11 @@ type GatewayEventBatchT struct {
...
-	SentAt            string      `json:"sentAt,omitempty"`
-	Type              string      `json:"type,omitempty"`
-}
-
 //EventUploadBatchT is a structure to hold batch of events
 type EventUploadBatchT struct {
 	WriteKey   string
 	ReceivedAt string
-	Batch      []EventUploadT
+	Batch      []map[string]interface{}
```
SumanthPuram, line 29: Define a type for EventUpload.
`type EventUploadT map[string]interface{}`

SumanthPuram: Makes all further references more readable.

prabrishac: done.

services/source-debugger/eventUploader.go
```diff
@@ -137,17 +126,20 @@ func uploadEvents(eventBuffer []*GatewayEventBatchT) {
 
 			//updating originalTimestamp in the event using the formula
 			//receivedAt - (sentAt - originalTimeStamp)
-			orgTS, err := time.Parse(time.RFC3339, ev.OriginalTimestamp)
+			originalTimestamp := ev["originalTimestamp"].(string)
```
psrikanth88, line 129: check if event has originalTimestamp property. Only events sent with SDK's have originalTimestamp. And maybe older SDK's do not have it also.
Also handle casting to string error?

services/source-debugger/eventUploader.go
```diff
@@ -137,17 +126,20 @@ func uploadEvents(eventBuffer []*GatewayEventBatchT) {
...
+			sentAt := ev["sentAt"].(string)
```
psrikanth88, line 130: same err handling as above

psrikanth88: also maybe add sentAt in gateway if not present in payload
something like we do for receivedAt https://github.com/rudderlabs/rudder-server/blob/master/gateway/gateway.go#L395 but with additional if present check

services/source-debugger/eventUploader.go
prabrishac: done. reusing the same logic from processor.go

services/source-debugger/eventUploader.go
prabrishac: done. reusing the same logic from processor.go. As discussed, did not add it in gateway for now.

prabrishac:
Corresponding config BE change: https://github.com/rudderlabs/rudder-config-backend/pull/217

PR #436, opened 2020-07-22
Error handling

sivashanmukh:
**Fixes** # (*issue*)

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
## Checklist:


rruntime/goroutine-factory.go
```diff
@@ -22,21 +26,30 @@ import (
...
 					"GoRoutines": {
 						"Number": runtime.NumGoroutine(),
 					}})
 
 				misc.RecordAppError(fmt.Errorf("%v", r))
 				logger.Fatal(r)
-				panic(r)
+				err, ok := r.(string)
+				_ = err
+				_ = ok
```
snarkychef, line 45: What are these 3 lines of code doing?

rruntime/goroutine-factory.go
```diff
@@ -22,21 +26,30 @@ import (
...
 //      	rt.workerProcess(worker)
 //    })
-func Go(function func()) {
+//When the go routine panics with an error, and if a custom errorHandler is passed
+func GoHandleError(function func(), errorHandler func(err error)) {
 	go func() {
 		ctx := bugsnag.StartSession(context.Background())
 		defer func() {
 			if r := recover(); r != nil {
+
```
snarkychef, line 35: Also, are we planning to give support for suppressing panic with custom panic handler? If so, we should call custom panic handler first and also decide if we want to further here to notify bugsnag, reportapperror etc...

router/router.go
```diff
@@ -201,7 +201,7 @@ func (rt *HandleT) workerProcess(worker *workerT) {
 			if rt.customDestinationManager != nil {
 				respStatusCode, respStatus, respBody = rt.customDestinationManager.SendData(job.EventPayload, paramaters.SourceID, paramaters.DestinationID)
 			} else {
-				respStatusCode, respStatus, respBody = rt.netHandle.sendPost(job.EventPayload)
+				respStatusCode, respStatus, respBody, err = rt.netHandle.sendPost(job.EventPayload)
```
snarkychef, line 204: `err` handling pending?

README.md
```diff
@@ -1,6 +1,8 @@
 ![Build Status][build status]
 [![Release]][release]
 
+[Build](https://codebuild.us-east-1.amazonaws.com/badges?uuid=eyJlbmNyeXB0ZWREYXRhIjoiOW0vREcyTDh3QXcyTkRyd0FPOUV1eTdvMXViTGFuQzAwZ0NHbllZdTJKelVVb0srN3kyVTczSHJ0Y3BzeW1BT3RNaCtITjZoc0Q3bW9kZEZiVlV1L0djPSIsIml2UGFyYW1ldGVyU3BlYyI6Im0vL01Rc09wNXJmMDI1Z2QiLCJtYXRlcmlhbFNldFNlcmlhbCI6MX0%3D&branch=master)
```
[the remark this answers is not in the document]
sivashanmukh: Yea Makes sense.

processor/integrations/integrations.go
```diff
@@ -19,16 +20,31 @@ var (
 	destTransformURL, userTransformURL string
 	customDestination                  []string
 	whSchemaVersion                    string
+	postParamentersTFields               []string
```
SumanthPuram, line 23: Typo? postParamentersTFields

processor/integrations/integrations.go
```diff
@@ -38,111 +54,29 @@ const (
...
-		panic(fmt.Errorf("typecast of parsedJSON.Get(\"files\") to interface{} failed"))
-	}
-	return postInfo
+// PostParametersT is a struct for holding all the values from transformerResponse and use them to publish an event to a destination 
+type PostParametersT struct {
+	Type          string      `json:"type"`
+	URL           string      `json:"endpoint"`
+	RequestMethod string      `json:"method"`
+	UserID        string      `json:"userId"`
+	Headers       interface{} `json:"headers"`
```
SumanthPuram, line 63: Make all these `map[string]interface{}` except files?

router/network.go
```diff
@@ -129,8 +35,14 @@ func (network *NetHandleT) processNewResponseType(jsonData []byte) (int, string,
 	// so, code addition should be done here instead of version bumping of response.
 	if isRest && !isMultipart {
 		requestMethod := postInfo.RequestMethod
-		requestBody := postInfo.Body.(map[string]interface{})
-		requestQueryParams := postInfo.QueryParams.(map[string]interface{})
+		requestBody, ok := postInfo.Body.(map[string]interface{})
```
SumanthPuram, line 38: We can remove the typecasting here.

processor/integrations/integrations.go
```diff
@@ -19,16 +20,31 @@ var (
...
 }
 
+// This is called in init and it should be a one time call. Having reflect calls happening during runtime is not a great idea.
+// We unmarshal json response from transformer into PostParametersT struct.
+// Since unmarshal doesn't check if the fields are present in the json or not and instead just initialze to zero value, we have to manually do this check on all fields before unmarshaling
+// This function gets a list of fields tagged as json from the struct and populates in postParamentersTFields
+
+func populatePostParamenterFields() {
+	v := reflect.TypeOf(PostParametersT{})
+	postParamentersTFields = make([]string, v.NumField(), v.NumField())
```
SumanthPuram, line 42: https://www.calhoun.io/how-to-use-slice-capacity-and-length-in-go/

This is a good article to read. We need to clean up our overall code. But I think we should follow one of these standards.

```
	postParamentersTFields = make([]string, v.NumField())
    for i := 0; i < v.NumField(); i++ {
		postParamentersTFields[i] = strings.Split(v.Field(i).Tag.Get("json"), ",")[0]
    }
``` 

Or 

```
	postParamentersTFields = make([]string, 0, v.NumField())
    for i := 0; i < v.NumField(); i++ {
		postParamentersTFields = append(postParamentersTFields, strings.Split(v.Field(i).Tag.Get("json"), ",")[0])
    }
```

config/backend-config/backend-config_test.go
```diff
@@ -3,7 +3,8 @@ package backendconfig
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
+
```
SumanthPuram, line 6: Why do we need to add errors package in every file?

processor/integrations/integrations.go
sivashanmukh: I wasn't entirely sure if we can do it because this typecast is happening in it's caller and that too only conditionally.

config/backend-config/backend-config_test.go
sivashanmukh: It's not added. It's replaced.

processor/integrations/integrations.go
sivashanmukh: I used it for the lack of better name.

processor/integrations/integrations.go
sivashanmukh: Actually giving length and capacity both same value is redundant and instead just giving length should suffice. 

Alternatively since I know the length prematurely, I could have just created an array instead of a slice.

processor/integrations/integrations.go
SumanthPuram: I meant PostParamenters* has an extra "n"
It should PostParameters*

processor/integrations/integrations.go
SumanthPuram: We can do this. It makes things more readable. Please change it.

processor/integrations/integrations.go
```diff
@@ -38,113 +55,31 @@ const (
...
-	if !ok {
-		panic(fmt.Errorf("typecast of parsedJSON.Get(\"params\") to interface{} failed"))
-	}
-	postInfo.Files, ok = parsedJSON.Get("files").Value().(interface{})
-	if !ok {
-		panic(fmt.Errorf("typecast of parsedJSON.Get(\"files\") to interface{} failed"))
+	for _, v := range postParametersTFields {
+		if !parsedJSON.Get(v).Exists() {
+			errMessage := fmt.Sprintf("missing expected field : %s in transformer response : %v", v, parsedJSON)
+			err = errors.New(errMessage)
```
SumanthPuram, line 76: You should return error here?

processor/integrations/integrations.go
snarkychef: I prefer the second approach using append. I feel it reads better as it will be clear that we are making up the list (appending) as we go.

processor/integrations/integrations.go
```diff
@@ -38,113 +54,38 @@ const (
...
-	}
-	return version
-}
-
-// GetPostInfoNew parses the transformer response
-func GetPostInfoNew(transformRaw json.RawMessage) PostParameterNewT {
-	var postInfo PostParameterNewT
-	var ok bool
-	parsedJSON := gjson.ParseBytes(transformRaw)
-	if parsedJSON.Get("output").Exists() {
```
snarkychef, line 84: this block of code is not required anymore?

router/network.go
```diff
@@ -19,108 +19,16 @@ type NetHandleT struct {
 	httpClient *http.Client
 }
 
-func (network *NetHandleT) processOldResponseType(jsonData []byte) (int, string, string) {
+func (network *NetHandleT) sendPost(jsonData []byte) (int, string, string) {
```
snarkychef, line 22: Please add a comment describing what this function returns, because it returns 3 things and 2 of the same type. Maybe even named returns. But I am not a big fan of bare return statements. Even though you are using named returns, I would still prefer return statement with parameters included.

processor/integrations/integrations.go
sivashanmukh: Yea. Not required afaik. @SumanthPuram can validate

SumanthPuram: @sivashanmukh Can you test this with a user-transformation?

There are 4 cases we need to validate.

User transformation - success/failure
Dest transformation - success/faulure

SumanthPuram: A couple more:
Source transformation - success/failure

SumanthPuram: Test with the latest transformer. You can the see the output that the Transformer is returning in versionedRouter.js.

sivashanmukh: I have tested it throughly today, and this is not required.

`processor.go: func (proc *HandleT) getDestTransformerEvents(....
`
The output in the transformer response is handled here and only its contents are stored in the job which is picked by router. So this is not required.

router/network.go
```diff
@@ -19,108 +19,18 @@ type NetHandleT struct {
...
-	if resp != nil && resp.Body != nil {
-		respBody, _ = ioutil.ReadAll(resp.Body)
-		defer resp.Body.Close()
-	}
-
+	postInfo, err := integrations.GetPostInfo(jsonData)
 	if err != nil {
-		logger.Error("Errored when sending request to the server", err)
-		return http.StatusGatewayTimeout, "", string(respBody)
+		panic(err)
```
psrikanth88, line 29: Should we return an error with custom message, instead of a panic ? We can use backed up aborted rt_jobs and replay them if needed.

Although each individual transformer builds response upon this https://github.com/rudderlabs/rudder-transformer/blob/master/v0/destinations/util.js#L79-L94, it might be flaky to build on guarantee that the response from transfomer will always have queryParams, files etc.

psrikanth88: Maybe we can add  the check to verify all the postInfo are in the response to transformer .
Check in versioedRouter.js before sending response and throw an error in transformer if not

processor/integrations/integrations.go
SumanthPuram: Perfect

router/network.go
SumanthPuram: @psrikanth88 Till we enable backing up aborted jobs by default and the restream functionality might be safe to panic. Another option is we can enable backing up aborted jobs by default and then return an error instead of panic.

SumanthPuram: @psrikanth88 @sivashanmukh WDYT?

SumanthPuram:
```
Another option is we can enable backing up aborted jobs by default and then return an error instead of panic.
```

SumanthPuram: Return an error in VersionedRouter if the format doesn't match.

PR #464, opened 2020-07-31
Support for overriding config with env variables

SumanthPuram:
**Fixes** # (*issue*)

## Type of change
## Checklist:


SumanthPuram:
Some of our customers don't want to provide the API keys to our control plane. This will add support for them to specify a specific value in the control plane which looks like `env.AMPLITUDE_API_KEY`, then looks for `AMPLITUDE_API_KEY` in the data plane when running the server.

config/backend-config/workspace-config.go
```diff
@@ -44,6 +49,35 @@ func (workspaceConfig *WorkspaceConfig) GetRegulations() (RegulationsT, bool) {
...
+	flattenedConfig, err := flatten.Flatten(configMap, "", flatten.DotStyle)
+
+	for configKey, v := range flattenedConfig {
+		reflectType := reflect.TypeOf(v)
+		if reflectType != nil && reflectType.String() == "string" {
+			valString := v.(string)
+			shouldReplace := strings.HasPrefix(strings.TrimSpace(valString), configEnvReplacer)
+			if shouldReplace {
+				envVariable := valString[len(configEnvReplacer):]
+				response, err = sjson.SetBytes(response, configKey, config.GetEnv(envVariable, ""))
```
psrikanth88, line 70: maybe a log here if env is not found, can be helpful in debugging

config/backend-config/workspace-config.go
```diff
@@ -67,8 +101,9 @@ func (workspaceConfig *WorkspaceConfig) getFromAPI() (SourcesT, bool) {
 		return SourcesT{}, false
 	}
 
+	updatedResponseBody := replaceConfigWithEnvVariables(respBody)
 	var sourcesJSON SourcesT
-	err = json.Unmarshal(respBody, &sourcesJSON)
+	err = json.Unmarshal(updatedResponseBody, &sourcesJSON)
```
psrikanth88, line 106: why not replace after unmarshal

config/backend-config/workspace-config.go
lokey: Maybe even panic here and ask them to either unset the `env.` value in webapp or add it in env here? Otherwise it is easy to miss.

config/backend-config/workspace-config.go
lokey, line 106: Can we rename the function from getFromApi or move this logic to some other function?

PR #477, opened 2020-08-06
Timestamp fix

dhawal1248:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -565,7 +565,9 @@ func enhanceWithTimeFields(event *transformer.TransformerEventT, singularEventMa
 	event.Message["receivedAt"] = receivedAt.Format(misc.RFC3339Milli)
 	event.Message["originalTimestamp"] = originalTimestamp.Format(misc.RFC3339Milli)
 	event.Message["sentAt"] = sentAt.Format(misc.RFC3339Milli)
-	event.Message["timestamp"] = misc.GetChronologicalTimeStamp(receivedAt, sentAt, originalTimestamp).Format(misc.RFC3339Milli)
+	if _, ok := event.Message["timestamp"]; !ok {
+		event.Message["timestamp"] = misc.GetChronologicalTimeStamp(receivedAt, sentAt, originalTimestamp).Format(misc.RFC3339Milli)
+	}
 }
 
 // add metadata to each singularEvent which will be returned by transformer in response
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -10,7 +10,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/araddon/dateparse"
 	"github.com/rudderlabs/rudder-server/config"
 	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
 	"github.com/rudderlabs/rudder-server/gateway"
@@ -541,19 +540,8 @@ func getBackendEnabledDestinationTypes(writeKey string) map[string]backendconfig
 
 func getTimestampFromEvent(event types.SingularEventT, field string) (timestamp time.Time) {
 	var ok bool
-	if timestamp, ok = getParsedTimestamp(event[field]); !ok {
+	if timestamp, ok = misc.GetParsedTimestamp(event[field]); !ok {
 		timestamp = time.Now()
-	}
-	return
-}
-
-func getParsedTimestamp(input interface{}) (parsedTimestamp time.Time, valid bool) {
-	if timestampStr, typecasted := input.(string); typecasted {
-		var err error
-		parsedTimestamp, err = dateparse.ParseAny(timestampStr)
-		if err == nil {
-			valid = true
-		}
 	}
 	return
 }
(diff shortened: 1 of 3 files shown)


processor/processor.go
```diff
@@ -565,7 +565,9 @@ func enhanceWithTimeFields(event *transformer.TransformerEventT, singularEventMa
 	event.Message["receivedAt"] = receivedAt.Format(misc.RFC3339Milli)
 	event.Message["originalTimestamp"] = originalTimestamp.Format(misc.RFC3339Milli)
 	event.Message["sentAt"] = sentAt.Format(misc.RFC3339Milli)
-	event.Message["timestamp"] = misc.GetChronologicalTimeStamp(receivedAt, sentAt, originalTimestamp).Format(misc.RFC3339Milli)
+	if _, ok := event.Message["timestamp"]; !ok {
```
SumanthPuram, line 568: We should format this as well to misc.RFC3339Milli

services/source-debugger/eventUploader.go
```diff
@@ -148,7 +148,9 @@ func uploadEvents(eventBuffer []*GatewayEventBatchT) {
 			sentAt := getTimestampFromEvent(ev, "sentAt")
 
 			ev["originalTimestamp"] = originalTimestamp.Format(time.RFC3339)
-			ev["timestamp"] = misc.GetChronologicalTimeStamp(receivedAtTS, sentAt, originalTimestamp).Format(time.RFC3339)
+			if _, ok := ev["timestamp"]; !ok {
```
SumanthPuram, line 151: Source debugger shows the events received at the gateway. The `timestamp` is computed later in the processor.
I don't think we should even send our computed timestamp here. 

In fact, we should not add any new fields here in eventUploader. Just send the event as it is received.
We can remove lines from 143 to 153. Let's just append ev to array.

Check with @prabrisha-rudder once, since she made the changes last time. I might have missed something.

processor/processor.go
psrikanth88, line 568: Should we check the data type / format of the string sent by user in timestamp field. Can you verify the segment behavior once like what happens if user sends int, bool, invalid timestamp format, different timestamp formats etc.
(also check if it's not empty string, whitespaces etc.)

processor/processor.go
```diff
@@ -556,16 +553,37 @@ func getTimestampFromEvent(event types.SingularEventT, field string) time.Time {
...
+
+	if evTimestamp, ok := event.Message["timestamp"]; ok {
+		var err bool
+		timestamp, err = getParsedTimestamp(evTimestamp)
+		if err {
+			timestamp = misc.GetChronologicalTimeStamp(receivedAt, sentAt, originalTimestamp)
+		}
+	} else {
+		timestamp = misc.GetChronologicalTimeStamp(receivedAt, sentAt, originalTimestamp)
+	}
```
psrikanth88, line 580:
```
var timestamp time.Time
var ok bool
if timestamp, ok = getParsedTimestamp(event.Message["timestamp"]); !ok {
	timestamp = misc.GetChronologicalTimeStamp(receivedAt, sentAt, originalTimestamp)
}
```

psrikanth88: and

```
func getParsedTimestamp(input interface{}) (parsedTimestamp time.Time, valid bool) {
	if  timestampStr, typecasted := input.(string); typecasted {
		var err error
		parsedTimestamp, err = dateparse.ParseAny(timestampStr)
		if err == nil {
			valid = true
		}
	}
	return
}
```

processor/processor.go
```diff
@@ -541,13 +541,10 @@ func getBackendEnabledDestinationTypes(writeKey string) map[string]backendconfig
...
-	if _, ok := event[field]; ok {
-		timestampStr, typecasted := event[field].(string)
-		if typecasted {
-			timestamp, err = dateparse.ParseAny(timestampStr)
-		}
-		if !typecasted || err != nil {
+	var err bool
+	if evTimestamp, ok := event[field]; ok {
+		timestamp, err = getParsedTimestamp(evTimestamp)
+		if err {
```
psrikanth88, line 547: and similar here as below

processor/processor.go
```diff
@@ -539,33 +539,44 @@ func getBackendEnabledDestinationTypes(writeKey string) map[string]backendconfig
...
+
+func getParsedTimestamp(input interface{}) (parsedTimestamp time.Time, valid bool) {
+	if timestampStr, typecasted := input.(string); typecasted {
+		var err error
+		parsedTimestamp, err = dateparse.ParseAny(timestampStr)
+		if err == nil {
+			valid = true
+		}
+	}
+	return
```
SumanthPuram, line 558: Let's be explicit about returning values.
`return parsedTimestamp`

processor/processor.go
```diff
@@ -539,33 +539,44 @@ func getBackendEnabledDestinationTypes(writeKey string) map[string]backendconfig
...
+func getTimestampFromEvent(event types.SingularEventT, field string) (timestamp time.Time) {
+	var ok bool
+	if timestamp, ok = getParsedTimestamp(event[field]); !ok {
 		timestamp = time.Now()
 	}
-	return timestamp
+	return
+}
+
+func getParsedTimestamp(input interface{}) (parsedTimestamp time.Time, valid bool) {
```
SumanthPuram, line 550: Let's move this function to `misc.go`?

processor/processor.go
dhawal1248: done

processor/processor.go
dhawal1248: moved

PR #499, opened 2020-08-19
Table dumps to cloud object storage logic optimisation and data race fixes.

chandumlg:
**Fixes** # (*issue*)

## Type of change
## Checklist:


jobsdb/jobsdb.go
```diff
@@ -2058,22 +2058,15 @@ func (jd *HandleT) backupTable(backupDSRange dataSetRangeT, isJobStatusTable boo
...
-			}
-			contentSlice[idx] = rowBytes
+		if len(rawJSONRows) < 2 {
+			break
 		}
-		// append new line character at end, before next backupRowsBatch is appended to same file
-		content := append(bytes.Join(contentSlice[:], []byte("\n")), []byte("\n")...)
-		gzWriter.Write(content)
+
+		rawJSONRows = bytes.Replace(rawJSONRows, []byte(", \n "), []byte("\n"), -1) //replacing ", \n " with "\n"
```
snarkychef, line 2065: This feels little risky to me, as we might end up with malformed json. Let's add an assert for number of rows equals number of replacements? I have checked current schemas, and as we are using jsonb types for json, any newlines in the data must be stripped off already. But in case ", \n" happens in some data, it might cause issue.

jobsdb/jobsdb.go
```diff
@@ -2058,22 +2058,15 @@ func (jd *HandleT) backupTable(backupDSRange dataSetRangeT, isJobStatusTable boo
...
+		rawJSONRows = rawJSONRows[1 : len(rawJSONRows)-1]                           //stripping starting '[' and ending ']'
```
snarkychef, line 2066: Let's add an assert to ensure the first and last characters are `[` and `]`?

jobsdb/jobsdb.go
```diff
@@ -2058,22 +2058,15 @@ func (jd *HandleT) backupTable(backupDSRange dataSetRangeT, isJobStatusTable boo
...
-			panic(err)
-		}
-		contentSlice := make([][]byte, len(rows))
-		for idx, row := range rows {
-			rowBytes, err := json.Marshal(row)
-			if err != nil {
-				panic(err)
-			}
-			contentSlice[idx] = rowBytes
+		if len(rawJSONRows) < 2 {
```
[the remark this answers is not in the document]
chandumlg: This is just a safe check so that subslice condition below doesn't panic.
Ideally this should never happen. We have a good number of checks to make sure that the number of rows returned are non zero.

row.Scan for the query `select json_agg (select *...) ....` throws some other error (not ErrNoRows)

jobsdb/jobsdb.go
chandumlg: json_agg returns a JSON Array. https://www.postgresql.org/docs/9.5/functions-aggregate.html.

If we want an assert, we can add.

jobsdb/jobsdb.go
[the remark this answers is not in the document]
chandumlg: removed this check.

jobsdb/jobsdb.go
chandumlg: added assert.

jobsdb/jobsdb.go
chandumlg: assert added.

snarkychef (approved):

SumanthPuram (approved):

PR #520, opened 2020-08-28
added google pub/sub as destination

ruchiramoitra:
**Fixes** # (*issue*)

## Type of change
## Checklist:


services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,164 @@
...
+		if topicIdString == "" {
+			respStatus = "Failure"
+			responseMessage = "[GooglePubSub] error :: empty delivery stream"
+			return 400, respStatus, responseMessage
+		}
+		topic := pbs.Topic(topicIdString)
+		result := topic.Publish(ctx, &pubsub.Message{Data: []byte(value)})
+
+		serverID, err := result.Get(ctx)
+		if err != nil {
```
prabrishac, line 103: write the logic in a separate function

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,164 @@
...
+				statusCode = 500
+				break
+			case codes.Unavailable:
+				statusCode = 503
+				break
+			case codes.Unauthenticated:
+				statusCode = 401
+				break
+			default:
+				statusCode = 400
```
prabrishac, line 142: make the default one 500

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,164 @@
...
+			logger.Error(responseMessage)
+			statusCode := 400
+			return statusCode, respStatus, responseMessage
+		}
+		if topicIdString == "" {
+			respStatus = "Failure"
+			responseMessage = "[GooglePubSub] error :: empty topic id string"
+			return 400, respStatus, responseMessage
+		}
+		topic := pbs.Topic(topicIdString)
```
snarkychef, line 99: // If a Topic's Publish method is called, it has background goroutines
// associated with it. Clean them up by calling Topic.Stop.
//
// Avoid creating many Topic instances if you use them to publish.
func (c *Client) Topic(id string) *Topic {

Referring to the method's documentation, there are following issues:
1. We are not calling topic.Stop(), which will leave behind goroutines. Verify this by looking at dataplane health endpoint, which contains goroutine count. 
2. I think we should reuse topic objects instead of creating them again and again. And stop topics whenever producer is closed.

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,164 @@
...
+		result := topic.Publish(ctx, &pubsub.Message{Data: []byte(value)})
```
snarkychef, line 100: // Publish publishes msg to the topic asynchronously. Messages are batched and
// sent according to the topic's PublishSettings. Publish never blocks.
//
// Publish returns a non-nil PublishResult which will be ready when the
// message has been sent (or has failed to be sent) to the server.
//
// Publish creates goroutines for batching and sending messages. These goroutines
// need to be stopped by calling t.Stop(). Once stopped, future calls to Publish
// will immediately return a PublishResult with an error.
func (t *Topic) Publish(ctx context.Context, msg *Message) *PublishResult {

Referring to the method's documentation, `Messages are batched and sent according to the topic's PublishSettings.` message will be sent to server once 10ms delayThreshold is reached according to the DefaultPublishSettings. This will limit our throughput. I think we should get rid of the delay threshold to immediately send our event, as we are not batching. 

Right after creating topic, we should set `topic.PublishSettings.DelayThreshold = 0`

snarkychef (changes requested):

services/streammanager/googlepubsub/googlepubsubmanager.go
ruchiramoitra: done

services/streammanager/googlepubsub/googlepubsubmanager.go
ruchiramoitra: done

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,164 @@
...
+	"google.golang.org/grpc/status"
+)
+
+type Config struct {
+	Credentials string `json:"credentials"`
+	ProjectId   string `json:"projectId"`
+}
+
+// NewProducer creates a producer based on destination config
+func NewProducer(destinationConfig interface{}) (pubsub.Client, error) {
```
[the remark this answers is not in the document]
ruchiramoitra: changed

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,164 @@
...
+	err = json.Unmarshal(jsonConfig, &config)
+	if err != nil {
+		return pubsub.Client{}, fmt.Errorf("[GooglePubSub] error  :: error in GooglePubSub while unmarshelling destination config:: %w", err)
+	}
+	var client *pubsub.Client
+	if config.Credentials != "" && config.ProjectId != "" {
+		client, err = pubsub.NewClient(ctx, config.ProjectId, option.WithCredentialsJSON([]byte(config.Credentials)))
+	}
+	if err != nil {
+		return pubsub.Client{}, err
```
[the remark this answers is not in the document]
ruchiramoitra: noted

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return *client, nil
+}
+func Produce(jsonData json.RawMessage, producer interface{}, destConfig interface{}) (statusCode int, respStatus string, responseMessage string) {
+	parsedJSON := gjson.ParseBytes(jsonData)
+	pbs, ok := producer.(pubsub.Client)
```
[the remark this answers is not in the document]
ruchiramoitra: now changing the logic we are sending nil when error occurs in new producer. In this scenario check will not pass?

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,164 @@
...
+		result := topic.Publish(ctx, &pubsub.Message{Data: []byte(value)})
+
+		serverID, err := result.Get(ctx)
+		if err != nil {
+			statusCode = getError(err)
+			responseMessage = "[GooglePubSub] error :: Failed to publish:" + err.Error()
+			respStatus = "Failure"
+			return statusCode, respStatus, responseMessage
+		} else {
+			responseMessage = "Message publish with id %v" + serverID
```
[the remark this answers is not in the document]
ruchiramoitra: changed

services/streammanager/googlepubsub/googlepubsubmanager.go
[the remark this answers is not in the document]
snarkychef: Yeah

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,164 @@
...
+	var config Config
+	jsonConfig, err := json.Marshal(destConfig)
+	if err != nil {
+		respStatus = "Failure"
+		responseMessage = "[GooglePubSub] error :: " + err.Error()
+		logger.Errorf("[GooglePubSub] error  :: %w", err)
+		statusCode := 400
+		return statusCode, respStatus, responseMessage
+	}
+	err = json.Unmarshal(jsonConfig, &config)
```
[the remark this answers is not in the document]
ruchiramoitra: yes it is not being used in producer. Redundant so removed

services/streammanager/googlepubsub/googlepubsubmanager.go
ruchiramoitra: added this.

services/streammanager/googlepubsub/googlepubsubmanager.go
ruchiramoitra: done

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,198 @@
...
+
+type Config struct {
+	Credentials     string              `json:"credentials"`
+	ProjectId       string              `json:"projectId"`
+	EventToTopicMap []map[string]string `json:"eventToTopicMap"`
+}
+
+type pubsubClient struct {
+	Pbs          *pubsub.Client
+	EventToTopic []map[string]interface{}
```
snarkychef, line 25: Looking at the code, I think this should be 
```
Topics map[string]*pubsub.Topic     # map from topic name to topic object
``` 
instead of 
```
EventToTopic []map[string]interface{}
```
Because as far as I can tell, we are accessing topic object using topic name.

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -0,0 +1,198 @@
...
+				break
+			}
+		}
+		if topic == nil {
+			statusCode = 400
+			responseMessage = "[GooglePubSub] error :: Topic not found in project"
+			respStatus = "Failure"
+			return statusCode, respStatus, responseMessage
+		}
+		topic.PublishSettings.DelayThreshold = 0
```
snarkychef, line 115: Setting this once should be enough, while topic creation.

snarkychef: Are we doing load tests? Did we observe any difference because of this change?

snarkychef (changes requested):

services/streammanager/googlepubsub/googlepubsubmanager.go
ruchiramoitra: changed

services/streammanager/googlepubsub/googlepubsubmanager.go
ruchiramoitra: done.
Will do load test.

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -99,9 +94,8 @@ func Produce(jsonData json.RawMessage, producer interface{}, destConfig interfac
 			return 400, respStatus, responseMessage
 		}
 		var topic *pubsub.Topic
-		for _, s := range pbs.TopicMap {
-			splitString := strings.Split(s.String(), "/")
-			if splitString[3] == topicIdString {
+		for i, s := range pbs.TopicMap {
```
snarkychef, line 97: You can access the topic as 
```
topic := pbs.TopicMap[topicIdString]
```
and get rid of this for loop.

snarkychef: Also, make sure unnecessary things are not logged. like the logger.Info(s.String()) below.

snarkychef (changes requested):

services/streammanager/googlepubsub/googlepubsubmanager.go
ruchiramoitra: done. and removed.

snarkychef (approved):

arnab-p (approved):

PR #560, opened 2020-09-24
force reset db migration version on change to older versions

psrikanth88:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/services/sql-migrator/migrator.go
+++ b/services/sql-migrator/migrator.go
@@ -27,6 +27,10 @@ type Migrator struct {
 
 	// Handle is the sql.DB handle used to execute migration statements
 	Handle *sql.DB
+
+	// Indicates if migration version should be force reset to latest on file in case of revert to lower version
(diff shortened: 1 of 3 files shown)


psrikanth88:
This is to handle cases where we want to revert back to version where database is in a older version.

Eg. Current wh_schemas_migrations is v2. On new feature deploy (which has sql migrations), db is set to v3. On reverting back the server to older commits, database version should be set back to v3 to avoid any migration crashes. This assumes that database changes in v3 are compatible with older versions (old server version can be safely run with database changes made in v3)

fxenik (commented):
Look's good!

fxenik (approved):

sivashanmukh:
@psrikanth88 

Lets say we had v3 and we went back to v2, this PR will handle that assuming the db changes are backward compatible.
But what happens when -
- putting v3 again after this revert?
- putting a v3new after this revert?

sivashanmukh (approved):

PR #589, opened 2020-10-08
Query warehouse through rudder-cli

psrikanth88:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/warehouse/client/client.go
+++ b/warehouse/client/client.go
@@ -1,0 +1,107 @@
+package client
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"cloud.google.com/go/bigquery"
+	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+	"google.golang.org/api/iterator"
+)
+
+const (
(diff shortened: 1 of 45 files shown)


SumanthPuram (commented):
Iterating over rows and columns is the same in all the warehouse files (except BQ).

Instead of that, can we move the logic one level up (instead of redshift/bigquery etc). We will do in something like warehouse/querier.go? We can create custom functions in the interface for special BQ handling.

warehouse/client/client.go
```diff
@@ -0,0 +1,107 @@
...
+type Client struct {
+	SQL  *sql.DB
+	BQ   *bigquery.Client
+	Type string
+}
+
+func (cl *Client) sqlQuery(statement string) (result warehouseutils.QueryResult, err error) {
+	rows, err := cl.SQL.Query(statement)
+	if err != nil && err != sql.ErrNoRows {
+		return
```
SumanthPuram, line 27: `return result, err` consistently at all places.

SumanthPuram (approved):

PR #602, opened 2020-10-16
Batching at router [WIP]

chandumlg:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/config/config.toml
+++ b/config/config.toml
@@ -154,10 +154,12 @@ kafkaDialTimeoutInSec = 10
 retryTimeWindowInMins = 180
 minRetryBackoffInS = 10
 maxRetryBackoffInS = 300
+workerRequestBatchTimeoutInS = 5
     [Router.BRAZE]
     forceHTTP1=true
     httpTimeoutInS=120
     httpMaxIdleConnsPerHost=32
+    workerVersion=1
     # settings similar to below one can be used to throttle router
     # sending events to downstream destinations
     # [Router.throttler.AM]
--- a/config/config.toml
+++ b/config/config.toml
@@ -159,7 +159,7 @@ workerRequestBatchTimeoutInS = 5
     forceHTTP1=true
     httpTimeoutInS=120
     httpMaxIdleConnsPerHost=32
-    workerVersion=1
+    workerVersion=solo
     # settings similar to below one can be used to throttle router
     # sending events to downstream destinations
     # [Router.throttler.AM]
--- a/router/router.go
+++ b/router/router.go
@@ -1,14 +1,11 @@
 package router
 
 import (
-	"encoding/json"
 	"fmt"
 	"math"
 	"math/rand"
 	"sort"
-	"strconv"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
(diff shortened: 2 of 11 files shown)


config/config.toml
```diff
@@ -154,10 +154,12 @@ kafkaDialTimeoutInSec = 10
 retryTimeWindowInMins = 180
 minRetryBackoffInS = 10
 maxRetryBackoffInS = 300
+workerRequestBatchTimeoutInS = 5
```
snarkychef, line 157: We need one more configuration for MaxBatchSize. And a boolean flag for enabling batching.

Also, I think we can remove the worker prefix.

router/router.go
```diff
@@ -1,40 +1,47 @@
...
-	"github.com/rudderlabs/rudder-server/processor/integrations"
 	"github.com/rudderlabs/rudder-server/rruntime"
 	"github.com/rudderlabs/rudder-server/services/customdestinationmanager"
-	destinationdebugger "github.com/rudderlabs/rudder-server/services/destination-debugger"
 	"github.com/rudderlabs/rudder-server/services/stats"
 	"github.com/rudderlabs/rudder-server/utils/logger"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 )
 
+type WorkerFactory interface {
```
snarkychef, line 24: This interface looks more like worker manager to me than a factory.

router/router.go
```diff
@@ -1,40 +1,47 @@
...
+	CreateWorkers(rt *HandleT)
+	GetWorker(idx int) WorkerInterface
+	ResetSleep()
+}
+
+type WorkerInterface interface {
+	PostJobOnWorker(job *jobsdb.JobT)
+	WorkerProcess(rt *HandleT)
+	GetFailedJobIDMutex() *sync.RWMutex
```
snarkychef, line 33: I was under the impression that failed job handling will only happen in final worker in pipeline and we shouldn't need this in batch worker.

router/router.go
```diff
@@ -118,6 +115,7 @@ func loadConfig() {
 	retryTimeWindow = config.GetDuration("Router.retryTimeWindowInMins", time.Duration(180)) * time.Minute
 	minRetryBackoff = config.GetDuration("Router.minRetryBackoffInS", time.Duration(10)) * time.Second
 	maxRetryBackoff = config.GetDuration("Router.maxRetryBackoffInS", time.Duration(300)) * time.Second
+	workerRequestBatchTimeout = config.GetDuration("Router.workerRequestBatchTimeoutInS", 5) * time.Second
```
snarkychef, line 118: Let's start supporting multiple units of duration going forward? We can have fixed suffix list like 
"20s" for 20 seconds
"20ms" for 20 milli seconds etc.

Because we faced difficulty recently when we wanted tune a timeout configuration to milli seconds but it only supported seconds.

router/router.go
```diff
@@ -1,40 +1,47 @@
...
 	"github.com/rudderlabs/rudder-server/rruntime"
 	"github.com/rudderlabs/rudder-server/services/customdestinationmanager"
-	destinationdebugger "github.com/rudderlabs/rudder-server/services/destination-debugger"
 	"github.com/rudderlabs/rudder-server/services/stats"
 	"github.com/rudderlabs/rudder-server/utils/logger"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 )
 
+type WorkerFactory interface {
+	CreateWorkers(rt *HandleT)
```
snarkychef, line 25: I would probably call this LaunchWorkers rather than CreateWorkers

router/router.go
```diff
@@ -1,40 +1,47 @@
...
+	GetWorker(idx int) WorkerInterface
+	ResetSleep()
+}
+
+type WorkerInterface interface {
+	PostJobOnWorker(job *jobsdb.JobT)
+	WorkerProcess(rt *HandleT)
```
snarkychef, line 32: Passing rt object here adds complexity I think. We don't why/how it is accessing it. I think it's better to pass actual parameters that's required for that workers. Maybe you can pass some kind of Options/Config object based on the worker in addition to any fixed parameters.

router/router.go
snarkychef: Shouldn't this also take worker count? Because you are accessing workers using idx in next function. You need to know how many workers are present right?

PR #653, opened 2020-11-12
Adding stats for aborted job count in batch router

sivashanmukh:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/router/batchrouter/batchrouter.go
+++ b/router/batchrouter/batchrouter.go
@@ -147,7 +147,7 @@ type ErrorResponseT struct {
 	Error string
 }
 
-func updateDestStatusStats(id string, jobStateCount map[string]int, destType string, isWarehouse bool, parameterFilters []jobsdb.ParameterFilterT) {
+func sendDestStatusStats(id string, jobStateCount map[string]int, destType string, isWarehouse bool, parameterFilters []jobsdb.ParameterFilterT) {
 	tags := map[string]string{
 		"module":      "batch_router",
 		"destType":    destType,
@@ -160,7 +160,7 @@ func updateDestStatusStats(id string, jobStateCount map[string]int, destType str
 	for jobState, count := range jobStateCount {
 		tags["job_state"] = jobState
 		if count > 0 {
-			stats.NewTaggedStat("batch_job_states", stats.CountType, tags).Count(count)
+			stats.NewTaggedStat("event_status", stats.CountType, tags).Count(count)
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


chandumlg (approved):

gane5hvarma (approved):

SumanthPuram (commented):
Instead of doing it here, might be better to do it in recordUploadStatus function? That's where we are tracking event deliveries too.

sivashanmukh:
@SumanthPuram There is another setJobStatus [here](https://github.com/rudderlabs/rudder-server/blob/fae990229522e6eda04b7b780ded81ddff998bde/router/batchrouter/batchrouter.go#L606). We won't count those if we move it inside recordUploadStat. @psrikanth88 should we panic on [this error](https://github.com/rudderlabs/rudder-server/blob/fae990229522e6eda04b7b780ded81ddff998bde/router/batchrouter/batchrouter.go#L605)?

SumanthPuram:
I think the handling for jobsBySource needs to be improved.

[Here](https://github.com/rudderlabs/rudder-server/blob/fae990229522e6eda04b7b780ded81ddff998bde/router/batchrouter/batchrouter.go#L573) we mark `executing` irrespective of whether src/destination config is present. 

[Here](https://github.com/rudderlabs/rudder-server/blob/fae990229522e6eda04b7b780ded81ddff998bde/router/batchrouter/batchrouter.go#L606) we mark `failed` even if the config is not present.

We should search for the source in the config in both places. If we don't find this in the first, we should not even mark it `executing`

sivashanmukh:
I have cleaned it up a little more. This time I moved out the aborted check, consolidated success and fail counts into the same single metric with different tags to differentiate. Regarding updating the job statuses in a better way. Let's take it as a different task.

gane5hvarma (approved):

SumanthPuram (approved):

router/batchrouter/batchrouter.go
```diff
@@ -147,14 +147,22 @@ type ErrorResponseT struct {
...
-	} else {
-		destStatsD = stats.NewBatchDestStat("batch_router.dest_failed_attempts", stats.CountType, id)
+func sendDestStatusStats(id string, jobStateCount map[string]int, destType string, isWarehouse bool, parameterFilters []jobsdb.ParameterFilterT) {
+	tags := map[string]string{
+		"module":      "batch_router",
+		"destType":    destType,
+		"isWarehouse": fmt.Sprintf("%t", isWarehouse),
+	}
+	for _, paramFilter := range parameterFilters {
+		tags[fmt.Sprintf("param_%s", paramFilter.Name)] = paramFilter.Value
```
SumanthPuram, line 157: The parameter filter also has received_at as a key and the value of received_at timestamp. Timestamps will blow up influx.

Eg
```
{ 
  "source_id": "s1",
  "destination_id": "d1",
  "received_at": "t1"
}
```

Instead of sending all parameter filters, let's send only source and destinations? 
We should use this function for representing any `source` or `destination` in tags.

`misc.GetTagName(destination.Destination.ID, destination.Destination.Name)`

Also the tag names should `source` and `destination` to maintain parity across all metrics.

SumanthPuram (commented):
Need to fix the tags.

router/batchrouter/batchrouter.go
sivashanmukh: In this context, parameter filters doesn't have received_at. It has only one parameter and that is destId

```
	parameterFilters := []jobsdb.ParameterFilterT{
		{
			Name:     "destination_id",
			Value:    batchJobs.BatchDestination.Destination.ID,
			Optional: false,
		},
	}

```
In any case, I have removed the paramFilter piece and am adding source and destination ids like you suggested. Take a look. Should we add source/dest names here for better readability?

PR #661, opened 2020-11-18
Drop extra requests in processor loop read

psrikanth88:
**Fixes** # (*issue*)

## Type of change
## Checklist:


SumanthPuram (commented):
Needs changes

processor/processor.go
```diff
@@ -997,12 +985,22 @@ func (proc *HandleT) handlePendingGatewayJobs() bool {
...
 	proc.logger.Debugf("Processor DB Read size: %v", toQuery)
 	//Should not have any failure while processing (in v0) so
 	//retryList should be empty. Remove the assert
-	retryList := proc.gatewayDB.GetToRetry([]string{gateway.CustomVal}, toQuery, nil)
-	toQuery -= len(retryList)
-	unprocessedList := proc.gatewayDB.GetUnprocessed([]string{gateway.CustomVal}, toQuery, nil)
+	unTruncatedRetryList := proc.gatewayDB.GetToRetry([]string{gateway.CustomVal}, toQuery, nil)
+	retryList, totalRetryEvents := getTruncatedEventList(unTruncatedRetryList, toProcessEvents)
+
+	toQuery -= totalRetryEvents
```
SumanthPuram, line 995: Should this be 
toQuery -= len(retryList)
?

SumanthPuram (commented):
.

processor/processor.go
psrikanth88: dbReadBatchSize - 1000, maxLoopProcessEvents - 1000

assuming we got 500 events (200 requests) from querying retry jobs, it should be enough thatt we query for 500 unprocessed jobs rather than 800. (assuming 1 jobs has atleast one event)

SumanthPuram: Got it. Looks good to me.

Given that we can assume there is at least one event in a job, if someone configures these variables wrong i.e., `dbReadSize > maxNumEvents in loop`

Should we make the

 `dbReadSize = min(dbReadSize, maxNumEvents)`

SumanthPuram (approved):

processor/processor.go
chandumlg: Deciding `dbReadBatchSize` based on `avgEventsInRequest` is risky when we change `avgEventsInRequest` and restart. We may end up not processing events in order.

processor/processor.go
```diff
@@ -997,12 +992,28 @@ func (proc *HandleT) handlePendingGatewayJobs() bool {
...
+
+	unTruncatedRetryList := proc.gatewayDB.GetToRetry([]string{gateway.CustomVal}, toQuery, nil)
+	retryList, totalRetryEvents, hasMaxEvents = getTruncatedEventList(unTruncatedRetryList, toProcessEvents)
+
+	if len(unTruncatedRetryList) >= dbReadBatchSize || hasMaxEvents {
+		// skip querying for unprocessed jobs if either retreived dbReadBatchSize or retreived maxEventToProcess
+	} else {
+		eventsLeftToProcess := maxEventsToProcess - totalRetryEvents
+		toQuery = misc.MinInt(eventsLeftToProcess, dbReadBatchSize)
+		if toProcessEvents > 0 {
```
chandumlg, line 1012: Do we need this check?

You are never decrementing toProcessEvents.

chandumlg (approved):

processor/processor.go
```diff
@@ -989,6 +964,26 @@ func (proc *HandleT) processJobsForDest(jobList []*jobsdb.JobT, parsedEventList
 	proc.pStatsDBW.Print()
 }
 
+func getTruncatedEventList(jobList []*jobsdb.JobT, maxEvents int) (truncatedList []*jobsdb.JobT, totalEvents int, gotMaxEvents bool) {
```
SumanthPuram, line 967: 1. Do we need gotMaxEvents 
2. Can we do with a single return? Makes it more readable in this case.
```
	lastIndex := len(jobList)

  for idx, job := range jobList {
		eventsInJob := len(gjson.GetBytes(job.EventPayload, "batch").Array())
		totalEvents += eventsInJob
		if totalEvents >= maxEvents {
			// return atleast one job to prevent processing first job with events > maxEvents
			// eg. first job has 1500 events and maxEvents is only 1000
			if idx == 0 {
				lastIndex = 1
			} else if totalEvents == maxEvents {
				// if both are equal, include current index of slice also
				lastIndex = idx + 1
			} else {
                lastIndex = idx
                totalEvents = totalEvents - eventsInJob
            }
            break
		 }
	}

  return jobList[:lastIndex], totalEvents
```

processor/processor.go
```diff
@@ -1000,9 +995,22 @@ func (proc *HandleT) handlePendingGatewayJobs() bool {
...
-	retryList := proc.gatewayDB.GetToRetry([]string{gateway.CustomVal}, toQuery, nil)
-	toQuery -= len(retryList)
-	unprocessedList := proc.gatewayDB.GetUnprocessed([]string{gateway.CustomVal}, toQuery, nil)
+
+	var retryList, unprocessedList []*jobsdb.JobT
+	var totalRetryEvents, totalUnprocessedEvents int
+	var gotMaxEvents bool
+
+	unTruncatedRetryList := proc.gatewayDB.GetToRetry([]string{gateway.CustomVal}, toQuery, nil)
+	retryList, totalRetryEvents, gotMaxEvents = getTruncatedEventList(unTruncatedRetryList, maxEventsToProcess)
```
SumanthPuram, line 1004:
```
retryList, totalRetryEvents = getTruncatedEventList(unTruncatedRetryList, maxEventsToProcess)
eventsLeftToProcess := maxEventsToProcess - totalRetryEvents

if len(unTruncatedRetryList) >= dbReadBatchSize || eventsLeftToProcess <= 0 {
```

SumanthPuram (commented):
Minor comments.

processor/processor.go
psrikanth88: @SumanthPuram  suppose maxEvents is 1000 and totalRetryEvents is 950 although there are more retryJobs left (eg. next retry job has 100 events, including it in retryJobs will make totalRetryEvents 1050 ( > 1000). This is the reason why gotMaxEvents is also returned

PR #709, opened 2020-12-08
migrating dedup

bipin-rudder:
**Fixes** # (*issue*)

## Type of change
## Checklist:


processor/processor.go
```diff
@@ -1033,6 +1041,10 @@ func (proc *HandleT) handlePendingGatewayJobs() bool {
 
 	proc.statDBR.End()
 
+	if enableDedup {
+		unprocessedList = proc.dedupHandler.DedupJobs(unprocessedList)
```
SumanthPuram, line 1045: We are writing to BadgerDB here. But, there is a chance that the application crashes before the processor marked the gateway jobs successful.

That will cause events to be dropped by the processor in the next loop. 

We should write to badger only after writing success to gateway jobsdb.

services/dedup/dedup.go
```diff
@@ -0,0 +1,233 @@
...
+			for msgId := range allMessageIdsSet {
+				messageIdsArr = append(messageIdsArr, msgId)
+			}
+			rruntime.Go(func() {
+				d.writeToBadger(messageIdsArr)
+			})
+		}
+		//Mark all as executing so next query doesn't pick it up
+		var statusList []*jobsdb.JobStatusT
+		for _, batchEvent := range duplicateList {
```
SumanthPuram, line 216: Dedup should not be directly writing to jobsDB. Instead, it should only work on top of badger.

SumanthPuram: It would be ideal if we could restrict jobsdb transactions only to the processor.

SumanthPuram (changes requested):

PR #746, opened 2020-12-28
disable nullable support for clickhouse destination

gane5hvarma:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-server/746)


warehouse/clickhouse/clickhouse.go
```diff
@@ -187,30 +190,35 @@ func (ch *HandleT) getConnectionCredentials() credentialsT {
...
 func getClickHouseColumnTypeForSpecificTable(tableName string, columnType string, notNullableKey bool) string {
 	if notNullableKey {
 		return columnType
 	}
+	// Nullable is not disabled for users and identity table
 	if tableName == warehouseutils.UsersTable {
 		return fmt.Sprintf(`SimpleAggregateFunction(anyLast, Nullable(%s))`, columnType)
 	}
+	if tableName != warehouseutils.IdentifiesTable && disableNullable {
+		return fmt.Sprintf(`%s`, columnType)
```
sivashanmukh, line 220: Remove fmt.Sprintf() and directly return

warehouse/clickhouse/clickhouse.go
```diff
@@ -270,28 +278,41 @@ func generateArgumentString(arg string, length int) string {
 
 // typecastDataFromType typeCasts string data to the mentioned data type
 func typecastDataFromType(data string, dataType string) interface{} {
+
 	switch dataType {
 	case "int":
 		i, err := strconv.Atoi(data)
 		if err != nil {
+			if disableNullable {
```
sivashanmukh, line 286: Maintain a map[string]interface{} where key is dataType and value is the default value for that type.

Let the error be caught after switch and set the value corresponding to that dataType on error.

sivashanmukh (approved):
LGTM

warehouse/clickhouse/clickhouse.go
```diff
@@ -187,30 +197,35 @@ func (ch *HandleT) getConnectionCredentials() credentialsT {
...
 
-func getClickHouseCodecForColumnType(columnType string) string {
+func getClickHouseCodecForColumnType(columnType string, tableName string) string {
 	switch columnType {
 	case "datetime":
-		return "Codec(DoubleDelta, LZ4)"
-	default:
-		return ""
+		if disableNullable && !(tableName == warehouseutils.IdentifiesTable || tableName == warehouseutils.UsersTable) {
+			return "Codec(DoubleDelta, LZ4)"
```
psrikanth88, line 211: @gane5hvarma are we adding compression only for datetime fields?

psrikanth88 (approved):

PR #747, opened 2020-12-28
Rohith feature unbatch server sdk user events

cisse21:
**Fixes** # (*issue*)

## Type of change
## Checklist:


gateway/gateway.go
```diff
@@ -586,14 +583,42 @@ func (gateway *HandleT) beaconBatchHandler(w http.ResponseWriter, r *http.Reques
...
+	var errorMessage = ""
+	writeKey, _, ok := r.BasicAuth()
+	if !ok || writeKey == "" {
+		errorMessage = response.GetStatus(response.NoWriteKeyInBasicAuth)
+		misc.IncrementMapByKey(sourceFailStats, "noWriteKey", 1)
+		gateway.updateSourceStats(sourceFailStats, "gateway.write_key_failed_requests")
+		return errorMessage
+	}
+	payload, err := gateway.getPayloadFromRequest(r)
+	if err == nil {
```
sivashanmukh, line 597: Lets indent the successpath to the left by checking for err != nil and returning

gateway/gateway.go
```diff
@@ -542,6 +526,19 @@ func (gateway *HandleT) eventSchemaWebHandler(wrappedFunc func(http.ResponseWrit
...
+	if r.Body != nil {
+		payload, err := ioutil.ReadAll(r.Body)
+		r.Body.Close()
+		return payload, err
+	}
+	return []byte{}, errors.New("RequestBodyNil")
+}
+
+func (gateway *HandleT) webImportHandler(w http.ResponseWriter, r *http.Request) {
+	gateway.webHandler(w, r, "import")
```
sivashanmukh, line 539: Why forward this to webhandler and then do a check on reqType == 'import'. Instead just directly send to a different well defined function?

gateway/gateway.go
```diff
@@ -586,14 +583,42 @@ func (gateway *HandleT) beaconBatchHandler(w http.ResponseWriter, r *http.Reques
...
+	writeKey, _, ok := r.BasicAuth()
+	if !ok || writeKey == "" {
+		errorMessage = response.GetStatus(response.NoWriteKeyInBasicAuth)
+		misc.IncrementMapByKey(sourceFailStats, "noWriteKey", 1)
+		gateway.updateSourceStats(sourceFailStats, "gateway.write_key_failed_requests")
+		return errorMessage
+	}
+	payload, err := gateway.getPayloadFromRequest(r)
+	if err == nil {
+		if reqType == "import" {
```
sivashanmukh, line 598: As mentioned in my other comment, this check seems out of place here. I think we should directly handle it by sending the request to its own handler instead of webhandler

gateway/gateway.go
```diff
@@ -586,14 +583,42 @@ func (gateway *HandleT) beaconBatchHandler(w http.ResponseWriter, r *http.Reques
...
+			errorMessage = gateway.userWebImportHandler(w, r, "batch", payload, writeKey)
+		} else {
+			done := make(chan string, 1)
+			gateway.AddToWebRequestQ(r, &w, done, reqType, payload, writeKey)
+			//Wait for batcher process to be done
+			errorMessage = <-done
+		}
+	} else if err.Error() == "RequestBodyNil" {
```
sivashanmukh, line 606: Don't we want to collect any stats in this case?

gateway/gateway.go
```diff
@@ -49,10 +49,11 @@ import (
  */
 
 type webRequestT struct {
```
sivashanmukh, line 51: Since the contents of struct changed drastically and don't even have request in it anymore, should we consider renaming it appropriately? @chandumlg

gateway/webhook/webhook.go
```diff
@@ -243,15 +243,22 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
...
 	done := make(chan string)
-	webhook.gwHandle.AddToWebRequestQ(req.request, req.writer, done, "batch")
+	var errorMessage = ""
+	payload, err := ioutil.ReadAll(req.request.Body)
+	req.request.Body.Close()
+	if err == nil {
+		webhook.gwHandle.AddToWebRequestQ(req.request, req.writer, done, "batch", payload, req.writeKey)
+		errorMessage = <-done
+	} else {
+		errorMessage = "Invalid JSON"
```
sivashanmukh, line 258: How do we know its Invalid JSON before we even try to UnMarshal it?

sivashanmukh (changes requested):
I haven't reviewed the tests yet. Will review them in the final pass.

gateway/gateway.go
cisse21: Add Source Fail Stats to update source stats

chandumlg (approved):

gateway/gateway.go
```diff
@@ -586,23 +583,121 @@ func (gateway *HandleT) beaconBatchHandler(w http.ResponseWriter, r *http.Reques
...
+}
+
+func (gateway *HandleT) userWebImportHandler(w *http.ResponseWriter, r *http.Request, reqType string, payload []byte, writeKey string) string {
+	errorMessage := ""
+	usersPayload, payloadError := gateway.getUsersPayload(payload)
+	if payloadError != nil {
+		return payloadError.Error()
+	}
+	count := len(usersPayload)
 	done := make(chan string, 1)
```
sivashanmukh, line 654: I think we should buffer this channel to the size of `len(payload)`. Otherwise, in case the empty buffer of `userWebRequestWorker.webRequestQ` used inside `AddToWebRequestQ` is smaller than the `len(payload)` the subsequent flow gets stuck in the process of putting response in the done channel we pass which is only of size 1 because we don't yet start listening on done until we are done putting messages.

gateway/gateway.go
```diff
@@ -586,23 +583,121 @@ func (gateway *HandleT) beaconBatchHandler(w http.ResponseWriter, r *http.Reques
...
+		if err != nil {
+			gateway.logger.Debug(err.Error())
+			http.Error(w, response.GetStatus(err.Error()), 400)
+		}
+	}()
+	payload, writeKey, err := gateway.getPayloadAndWriteKey(w, r, reqType)
+	if err != nil {
+		return
+	}
+	err = gateway.checkAndAddToWebRequestQ(&w, r, reqType, payload, writeKey)
```
sivashanmukh, line 636: Lets rename this function as well. We are not just adding it in the queue. We are processing the request and getting it back as well. Lets call it checkAndProcessWebRequest?

sivashanmukh (approved):
LGTM

PR #766, opened 2021-01-08
WH: Remove slave dependency on warehouseDB

psrikanth88:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/warehouse/slave.go
+++ b/warehouse/slave.go
@@ -143,16 +143,17 @@ type loadFileUploadJob struct {
 }
 
 type loadFileUploadOutputT struct {
-	tableName string
-	location  string
-	totalRows int
+	TableName     string
+	Location      string
+	TotalRows     int
+	StagingFileID int64
 }
 
-func (jobRun *JobRunT) uploadLoadFilesToObjectStorage() ([]int64, error) {
+func (jobRun *JobRunT) uploadLoadFilesToObjectStorage() ([]loadFileUploadOutputT, error) {
 	job := jobRun.job
 	uploader, err := job.getFileManager()
 	if err != nil {
-		return []int64{}, err
+		return []loadFileUploadOutputT{}, err
 	}
 	// var loadFileIDs []int64
 	var loadFileUploadOutputs []loadFileUploadOutputT
@@ -179,9 +180,10 @@ func (jobRun *JobRunT) uploadLoadFilesToObjectStorage() ([]int64, error) {
 						return
 					}
 					loadFileOutputChan <- loadFileUploadOutputT{
-						tableName: tableName,
-						location:  uploadOutput.Location,
-						totalRows: jobRun.tableEventCountMap[tableName],
+						TableName:     tableName,
+						Location:      uploadOutput.Location,
+						TotalRows:     jobRun.tableEventCountMap[tableName],
+						StagingFileID: jobRun.job.StagingFileID,
 					}
 				}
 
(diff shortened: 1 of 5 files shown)


warehouse/slave.go
```diff
@@ -559,13 +560,12 @@ func processClaimedJob(claimedJob pgnotifier.ClaimT) {
 		return
 	}
 	job.BatchID = claimedJob.BatchID
-	ids, err := processStagingFile(job)
-	job.setStagingFileStatus(err)
```
sivashanmukh, line 563: Why is this removed? This was added in a previous PR, right?

sivashanmukh, line 563: If this function is not used anywhere else, we should remove the function altogether.

psrikanth88: removed

warehouse/upload.go
```diff
@@ -1153,6 +1184,58 @@ func (job *UploadJobT) createLoadFiles(generateAll bool) (startLoadFileID int64,
 	return startLoadFileID, endLoadFileID, nil
 }
 
+func (job *UploadJobT) setStagingFileSuccess(stagingFileIDs []int64) {
+	sqlStatement := fmt.Sprintf(`UPDATE %s SET status=$1, updated_at=$2 WHERE id=ANY($3)`, warehouseutils.WarehouseStagingFilesTable)
```
SumanthPuram, line 1188: Minor comment, ignore if not needed. Should we use `where id IN $3` instead of `where id=ANY($3)`

Both should work, but `IN` seems to be more intuitive and we use that across other modules.

warehouse/upload.go
```diff
@@ -1131,19 +1133,48 @@ func (job *UploadJobT) createLoadFiles(generateAll bool) (startLoadFileID int64,
...
+				if len(output) == 0 {
+					pkgLogger.Errorf("[WH]: No LoadFiles returned by wh worker")
+					continue
+				}
+				loadFiles = append(loadFiles, output...)
+				successfulStagingFileIDs = append(successfulStagingFileIDs, resp.JobID)
+			}
+			err = job.bulkInstertLoadFileRecords(loadFiles)
+			if err != nil {
+				saveLoadFileErrs = append(saveLoadFileErrs, err)
```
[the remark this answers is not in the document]
sivashanmukh: If we have a continue, wg.Done() should be done before that. What's the purpose of continue? Do you intend that we skip job.setStagingFileSuccess?

warehouse/upload.go
psrikanth88: ANY seems to be more verstaile https://stackoverflow.com/a/34627688
Also pq.Array can be used instead of us preparing a comma separated list of values

warehouse/upload.go
```diff
@@ -1131,19 +1133,48 @@ func (job *UploadJobT) createLoadFiles(generateAll bool) (startLoadFileID int64,
...
 			}
+			job.setStagingFileSuccess(successfulStagingFileIDs)
 			wg.Done()
 		})
 	}
 
 	wg.Wait()
 
+	if len(saveLoadFileErrs) > 0 {
```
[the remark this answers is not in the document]
psrikanth88: will take this up in different PR

warehouse/upload.go
sivashanmukh: I would suggest that we start using ANY here and switch everywhere else with IN. It seems to be a better way to construct queries from code.

sivashanmukh (approved):
LGTM

warehouse/upload.go
SumanthPuram: https://stackoverflow.com/a/34627688 ->

This says IN uses indexes and ANY doesn't.

SumanthPuram (approved):

warehouse/upload.go
sivashanmukh: I too was confused when I saw that and had done an EXPLAIN on both variants in our query like so
`EXPLAIN ANALYZE UPDATE wh_staging_files SET status='success', updated_at='2021-01-01' WHERE id=ANY('{3,2,1}');
`
`EXPLAIN ANALYZE UPDATE wh_staging_files SET status='success', updated_at='2021-01-01' WHERE id IN (3,2,1);`

Both of them had given the same query plan. I had then read the reply again and what you mentioned along with the other link below it are special cases where the index wasn't used and we don't really use IN or ANY in that way afaik.

The one that you saw, points to this link -> https://dba.stackexchange.com/questions/125413/index-not-used-with-any-but-used-with-in/125500#125500 which is the special case the answer discusses.

SumanthPuram: Just checked this. 

For composite indexes, ANY is not using the index but IN is using it. Since this is a primary key, we will go ahead with ANY.

SumanthPuram (approved):

PR #773, opened 2021-01-15
Change logic to pick which uploads to process

psrikanth88:
**Fixes** # (*issue*)

## Type of change
## Checklist:


warehouse/warehouse.go
```diff
@@ -453,13 +462,18 @@ func (wh *HandleT) getPendingUploads(warehouse warehouseutils.WarehouseT) ([]Upl
 	return uploads, nil
 }
 
+func namespaceIdentifier(warehouse warehouseutils.WarehouseT) string {
```
sivashanmukh, line 465: Why not use workerIdentifier itself instead?

psrikanth88: @sivashanmukh missed that the func exists, used that

warehouse/warehouse.go
```diff
@@ -578,25 +589,6 @@ func (wh *HandleT) processJobs(warehouse warehouseutils.WarehouseT) (err error)
 		delete(inRecoveryMap, warehouse.Destination.ID)
 	}
 
-	// Step 2: Handle any Pending uploads
-	// An upload job is pending if it is neither exported nor aborted
-
-	pendingUploads, err := wh.getPendingUploads(warehouse)
```
sivashanmukh, line 584: This functions definition can be removed since its not used anywhere else

warehouse/warehouse.go
```diff
@@ -578,25 +589,6 @@ func (wh *HandleT) processJobs(warehouse warehouseutils.WarehouseT) (err error)
...
-	if err != nil {
-		pkgLogger.Errorf("[WH]: Failed to get pending uploads: %s with error %w", warehouse.Identifier, err)
-		return err
-	}
-
-	uploadJob, err := wh.getUploadJobsForPendingUploads(warehouse, whManager, pendingUploads)
```
sivashanmukh, line 590: This functions definition can be removed since its not used anywhere else

warehouse/warehouse.go
psrikanth88: removed

warehouse/warehouse.go
psrikanth88: removed

sivashanmukh (approved):
LGTM

warehouse/warehouse.go
```diff
@@ -732,15 +641,139 @@ func (wh *HandleT) mainLoop() {
 	}
 }
 
+func (wh *HandleT) getUploadsToProcess(availableWorkers int, skipIdentifiers []string) ([]*UploadJobT, error) {
+
+	var skipIdentifiersSQL string
+	if len(skipIdentifiers) > 0 {
+		skipIdentifiersSQL = fmt.Sprintf(`and concat(t.namespace, ':', t.destination_id) != ALL($1)`)
```
sivashanmukh, line 648: This concat should be with _ instead of column, because that's what we get from workerIdentifier func

SumanthPuram (approved):

PR #790, opened 2021-01-21
WH: Fix scheduling using lastEventAt

psrikanth88:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/warehouse/warehouse.go
+++ b/warehouse/warehouse.go
@@ -472,12 +472,17 @@ func isDestInProgress(warehouse warehouseutils.WarehouseT) bool {
 	return false
 }
 
-func uploadFrequencyExceeded(warehouse warehouseutils.WarehouseT, syncFrequency string) bool {
+func getUploadFreqInS(syncFrequency string) int64 {
 	freqInS := uploadFreqInS
 	if syncFrequency != "" {
 		freqInMin, _ := strconv.ParseInt(syncFrequency, 10, 64)
 		freqInS = freqInMin * 60
 	}
+	return freqInS
+}
+
+func uploadFrequencyExceeded(warehouse warehouseutils.WarehouseT, syncFrequency string) bool {
+	freqInS := getUploadFreqInS(syncFrequency)
 	lastEventTimeMapLock.Lock()
 	defer lastEventTimeMapLock.Unlock()
 	if lastExecTime, ok := lastEventTimeMap[warehouse.Identifier]; ok && timeutil.Now().Unix()-lastExecTime < freqInS {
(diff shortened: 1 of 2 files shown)


warehouse/warehouse.go
```diff
@@ -486,10 +491,33 @@ func uploadFrequencyExceeded(warehouse warehouseutils.WarehouseT, syncFrequency
...
+	unixTime := lastEventAt.Unix()
+
+	freqInS := getUploadFreqInS(warehouseutils.GetConfigValue(warehouseutils.SyncFrequency, warehouse))
+
+	lastStagingFileCreatedAt := getLastStagingFileCreatedAt(warehouse)
+	// set lastEvent to currentTime in cases where
+	// 1. if there is lag in creating staging files - last staging file is more than freqInS old
+	// 2. and if there is less than freqInS worth of data in pending staging files
+	if time.Now().Add(time.Duration(-freqInS)*time.Second).Sub(lastStagingFileCreatedAt) > 0 && lastStagingFileCreatedAt.Sub(lastEventAt) < time.Duration(freqInS)*time.Second {
+		unixTime = time.Now().Unix()
```
sivashanmukh, line 515: This fixes our issue, but the naming is kind of messed up. Basically, in the same map, we are storing last_exec_at when there is lag and last_event_at otherwise.

Instead lets keep setting last_exec_at like we used to do before because it only has in-memory existence. And then fix the check accounting for the lag in `uploadFrequencyExceeded`?

sivashanmukh, line 515: This fixes our issue, but, in the same map, we are storing last_exec_at when there is lag and last_event_at otherwise.

Instead lets keep setting `last_exec_at` like we used to do before because it only has in-memory existence. And then fix the check accounting for the lag in `uploadFrequencyExceeded`?

sivashanmukh, line 515: This fixes our issue, but, in the `lastEventTimeMap` map, we are storing `last_exec_at` when there is lag and `last_event_at` otherwise.

Instead lets `lastEventTimeMap` to `lastExecTimeMap` like before and keep setting `last_exec_at` like we used to do before because it only has in-memory existence. And then fix the check accounting for the lag in `uploadFrequencyExceeded`?

sivashanmukh (changes requested):

sivashanmukh:
@psrikanth88 Notification test

warehouse/warehouse.go
psrikanth88: changed the variable for better readability

SumanthPuram (approved):

sivashanmukh (approved):
LGTM

PR #795, opened 2021-01-26
Table skip load bug fix

sivashanmukh:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/warehouse/upload.go
+++ b/warehouse/upload.go
@@ -188,19 +188,6 @@ func (job *UploadJobT) initTableUploads() error {
 	return createTableUploads(job.upload.ID, tables)
 }
 
-func (job *UploadJobT) shouldTableBeLoaded(tableName string) (bool, error) {
-	tableUpload := NewTableUpload(job.upload.ID, tableName)
-	loaded, err := tableUpload.hasBeenLoaded()
-	if err != nil {
-		return false, err
-	}
-	hasLoadfiles, err := job.hasLoadFiles(tableName)
-	if err != nil {
-		return false, err
-	}
-	return !loaded && hasLoadfiles, nil
-}
-
 func (job *UploadJobT) syncRemoteSchema() (hasSchemaChanged bool, err error) {
 	schemaHandle := SchemaHandleT{
 		warehouse:    job.warehouse,
@@ -690,13 +677,12 @@ func (job *UploadJobT) loadTable(tName string) (alteredSchema bool, err error) {
 }
 
 func (job *UploadJobT) loadUserTables() ([]error, error) {
-	var loadTables bool
+	var hasLoadFiles bool
 	userTables := []string{job.identifiesTableName(), job.usersTableName()}
 
 	var err error
 	previouslyFailedTables, currentJobSucceededTables := job.getTablesToSkip()
 	for _, tName := range userTables {
-		loadTables, err = job.shouldTableBeLoaded(tName)
 		if _, ok := currentJobSucceededTables[tName]; ok {
 			continue
 		}
@@ -704,10 +690,11 @@ func (job *UploadJobT) loadUserTables() ([]error, error) {
 			err = fmt.Errorf("Skipping %s tables because it previously failed to load in an earlier job: %d", tName, prevJobID)
 			break
 		}
+		hasLoadFiles, err = job.hasLoadFiles(tName)
 		if err != nil {
 			break
 		}
-		if loadTables {
+		if hasLoadFiles {
 			// There is at least one table to load
 			break
 		}
@@ -717,7 +704,7 @@ func (job *UploadJobT) loadUserTables() ([]error, error) {
 		return []error{err}, nil
 	}
 
-	if !loadTables {
+	if !hasLoadFiles {
 		return []error{}, nil
 	}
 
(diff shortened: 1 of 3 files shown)


SumanthPuram (approved):

warehouse/upload.go
```diff
@@ -614,11 +585,20 @@ func (job *UploadJobT) loadAllTablesExcept(skipPrevLoadedTableNames []string) []
...
+			wg.Done()
+			continue
+		}
+		if prevJobID, ok := previouslyFailedTables[tableName]; ok {
+			loadErrors = append(loadErrors, fmt.Errorf("Skipping %s tables because it previously failed to load in an earlier job: %d", tableName, prevJobID))
+			wg.Done()
+		}
 		hasLoadFiles, err := job.hasLoadFiles(tableName)
 		if err != nil {
 			loadErrors = append(loadErrors, err)
```
psrikanth88, line 604: add missing wg.Done() here

warehouse/upload.go
```diff
@@ -716,7 +704,7 @@ func (job *UploadJobT) loadUserTables() (loadErrors []error, tableUploadErr erro
 		return []error{err}, nil
 	}
 
-	if !loadTables {
+	if !hasLoadFiles {
```
psrikanth88, line 707: In the case where prev upload has {identifies: S, users: F}, should we load identifies in the current upload? `hasLoadFiles` will always be true for current upload right because we do not skip load file generation for selective tables ?

warehouse/upload.go
```diff
@@ -764,46 +752,50 @@ func (job *UploadJobT) loadIdentityTables(populateHistoricIdentities bool) (load
 	identityTables := []string{job.identityMergeRulesTableName(), job.identityMappingsTableName()}
 
 	var alteredSchema bool
```
psrikanth88, line 754: In above line `if generated, _ := job.areIdentityTablesLoadFilesGenerated(); !generated`
should we even do the `resolveIdentities()` part? Maybe we can skip at the start of the func only if something in prev uploads has failed?

warehouse/upload.go
```diff
@@ -520,31 +500,22 @@ func getTableUploadStatusMap(tableUploadStatuses []*TableUploadStatusT) map[int6
...
 	tableUploadStatus := getTableUploadStatusMap(tableUploadStatuses)
-	previouslyFailedTableMap := make(map[string]bool)
+	previouslyFailedTableMap := make(map[string]int64)
 	currentlySucceededTableMap := make(map[string]bool)
 	for uploadID, tableStatusMap := range tableUploadStatus {
 		for tableName, status := range tableStatusMap {
-			if uploadID < job.upload.ID && status == TableUploadExportingFailed { //Previous upload and table upload failed
-				previouslyFailedTableMap[tableName] = true
+			if uploadID < job.upload.ID && (status == TableUploadExportingFailed || status == UserTableUploadExportingFailed) { //Previous upload and table upload failed
+				previouslyFailedTableMap[tableName] = uploadID
```
psrikanth88, line 511: include `exporting_identities_failed` also?

warehouse/warehouse.go
```diff
@@ -522,7 +522,7 @@ func (wh *HandleT) getUploadJobsForPendingUploads(warehouse warehouseutils.Wareh
 
 		if !wh.canStartPendingUpload(pendingUpload, warehouse) {
 			pkgLogger.Debugf("[WH]: Skipping pending upload for %s since current time less than next retry time", warehouse.Identifier)
-			if pendingUpload.Status != TableUploadExportingFailed {
+			if pendingUpload.Status != TableUploadExportingFailed && pendingUpload.Status != UserTableUploadExportingFailed {
```
psrikanth88, line 525: include exporting_identities_failed also?

warehouse/upload.go
```diff
@@ -520,31 +500,22 @@ func getTableUploadStatusMap(tableUploadStatuses []*TableUploadStatusT) map[int6
 	return tableUploadStatus
 }
 
-func (job *UploadJobT) getTablesToSkip() ([]string, []string) {
+func (job *UploadJobT) getTablesToSkip() (map[string]int64, map[string]bool) {
```
psrikanth88, line 503: this table does not include logic specific to any step(ExportingData or ExportingUserData). Can we add these on the `previouslyFailedTableMap, currentlySucceededTableMap` on the upload and check if they are already populated to skip evaluating this func multiple times?

psrikanth88 (changes requested):

warehouse/upload.go
sivashanmukh: In the case you mentioned, we would be breaking with an error with this check -> `if prevJobID, ok := previouslyFailedTables[tName]; ok {`

So even before this `if !hasLoadFilesCheck {` is done, we would be returning from the function with non-nil error.

I had left the hasLoadFilesCheck just because it was already being done. We can remove it if we are confident that there is no way we arrive here without loadFiles. Infact we can remove such checks for other table exporting steps as well?

warehouse/upload.go
sivashanmukh: Done

warehouse/upload.go
sivashanmukh: Done

warehouse/upload.go
sivashanmukh: I didn't bother with the multiple calls to this function, because once we merge the three steps to run concurrently, I thought we could just make a single function call to this guy and pass on the two maps to all three concurrent steps in one go.

psrikanth88 (approved):

PR #810, opened 2021-02-01
warehouse api's

gane5hvarma:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-server/810)


gane5hvarma:
TODO: Need to add index on `created_at` column

PR #834, opened 2021-02-11
load users table in one tx with identify table

gane5hvarma:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-server/834)

### DIFF
--- a/warehouse/postgres/postgres.go
+++ b/warehouse/postgres/postgres.go
@@ -320,6 +320,7 @@ func (pg *HandleT) loadTable(tableName string, tableSchemaInUpload warehouseutil
 	}
 	tableResp.name = stagingTableName
 	if skipTempTableDelete {
+		txn.Exec(`SAVEPOINT IDENTIFY_LOADED`)
 		tableResp.tx = txn
 		return
 	}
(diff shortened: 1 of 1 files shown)


warehouse/postgres/postgres.go
```diff
@@ -326,7 +335,7 @@ func (pg *HandleT) loadTable(tableName string, tableSchemaInUpload warehouseutil
 func (pg *HandleT) loadUserTables() (errorMap map[string]error) {
 	errorMap = map[string]error{warehouseutils.IdentifiesTable: nil}
 	pkgLogger.Infof("PG: Starting load for identifies and users tables\n")
-	identifyStagingTable, err := pg.loadTable(warehouseutils.IdentifiesTable, pg.Uploader.GetTableSchemaInUpload(warehouseutils.IdentifiesTable), true)
+	identifyStagingTableResp, err := pg.loadTable(warehouseutils.IdentifiesTable, pg.Uploader.GetTableSchemaInUpload(warehouseutils.IdentifiesTable), true)
```
sivashanmukh, line 338: Instead of creating the transaction inside loadTable and passing it out as a response, lets create it before this line, pass it to the loadTable as an argument and just do a tx.Exec inside and throw the txErr out. Let this scope decide if you have to rollback or commit.

PR #913, opened 2021-04-05
Google sheets destination

utsabc:
## Google Sheets Destination
## Checklist:


ruchiramoitra (commented):
line 62  check as discussed
line 65 try sending cox

ruchiramoitra (approved):

services/streammanager/googlesheets/schemas/googlesheetsschema.go
```diff
@@ -0,0 +1,85 @@
+package schemas
+
+import (
+	"github.com/tidwall/gjson"
+)
+
+func GetSheetsHeaderSchema() []interface{} {
+	hschema := []interface{}{
+		"id",
```
SumanthPuram, line 9: Can we reuse these fields instead of listing them again in the next function?

services/streammanager/googlesheets/schemas/googlesheetsschema.go
```diff
@@ -0,0 +1,85 @@
...
+		parsedJSON.Get("context_traits").String(),
+		parsedJSON.Get("context_userAgent").String(),
+		parsedJSON.Get("event_text").String(),
+		parsedJSON.Get("event").String(),
+		parsedJSON.Get("event_name").String(),
+		parsedJSON.Get("sent_at").String(),
+		parsedJSON.Get("received_at").String(),
+		parsedJSON.Get("timestamp").String(),
+		parsedJSON.Get("original_timestamp").String(),
+		parsedJSON.Get("properties").String()}
```
SumanthPuram, line 82: Instead of writing properties as a string, should we write flattened properties?

Also why should we hard code these things? I think we should write everything similar to what we do with warehouses.

SumanthPuram (changes requested):

services/streammanager/googlesheets/schemas/googlesheetsschema.go
utsabc: We have addressed this in the new change

services/streammanager/googlesheets/schemas/googlesheetsschema.go
utsabc: HI @SumanthPuram ,
For Point 1. We want to keep some message fields like `properties`, `traits`, and `context.traits` free-flowing object in Google Sheets as these are user-customizable
Screenshot 2021-04-15 at 1 27 06 PM

For Point 2, and 3. We did some implementation changes on how we are doing the transformation so that we can remove the schema dependency for the server altogether.  We have added the changes with inline explanation of our approach.

ruchiramoitra (approved):

ruchiramoitra (approved):

services/streammanager/googlesheets/googlesheetsmanager.go
```diff
@@ -0,0 +1,236 @@
...
+// and we are storing the data into designated position in array based on transformer
+// mappings.
+// Example payload we have from transformer:
+// { message:
+// 	 { 	'0': { key: 'anonymous_id', value: '' },
+// 	 	'1': { key: 'user_id', value: 'userTest004' },
+// 	 	'2': { key: 'event', value: 'Page Call' },
+// 	 	..}
+// }
+func ParseTransformedData(source gjson.Result) ([]string, []string, error) {
```
SumanthPuram, line 189: Starting with Capital letter makes a function public. There is no need for this function to be exposed outside.

SumanthPuram: Same with all functions. Don't make the functions public unless needed.

services/streammanager/googlesheets/googlesheetsmanager.go
```diff
@@ -0,0 +1,236 @@
...
+		responseMessage = "[GoogleSheets] error :: Failed to parse transformed data ::" + parseErr.Error()
+		return 400, respStatus, responseMessage
+
+	}
+	// Converting the string array of keys to an array of interface compatibale with sheets-api (Keys relates to sheet header)
+	headerRow := GetSheetsData(keys)
+	// Creating value range for inserting row into sheet
+	var vr sheets.ValueRange
+	vr.MajorDimension = "ROWS"
+	vr.Range = spreadSheetTab + "!A1"
```
SumanthPuram, line 135: What is this for?

services/streammanager/googlesheets/googlesheetsmanager.go
```diff
@@ -0,0 +1,236 @@
...
+	// With our global - sheet client now updated we are ready to proceed for further
+	// api calls.
+
+	// Here we extract the keys, values of the transformed json into two string arrays
+	// because google sheets-api works with array of data if any error occurs here we
+	// log it as parse error for that event
+	keys, values, parseErr := ParseTransformedData(parsedJSON)
+
+	if parseErr != nil {
+		respStatus = "Faliure"
```
SumanthPuram, line 125: Typo: Failure

services/streammanager/googlesheets/googlesheetsmanager.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// Here we recieve the transformed json with all the required mappings
+	parsedJSON := gjson.ParseBytes(jsonData)
+	// Storing sheet_id and sheet_tab for the specific message
+	spreadSheetId := parsedJSON.Get("spreadSheetId").String()
+	spreadSheetTab := parsedJSON.Get("spreadSheetTab").String()
+
+	var pServiceMessage string
+	// First we are pinging to google-sheets using existing client
+	pkgLogger.Debug("[Google Sheets] Pingging to Sheets using existing client", googleAPIService)
```
SumanthPuram, line 92: Do we need to ping first? We will end up doing 2 calls for every event. 

Instead, can we try to post and if it fails , then based on the error code - refresh the token.

services/streammanager/googlesheets/googlesheetsmanager.go
```diff
@@ -0,0 +1,236 @@
...
+	data := make([]interface{}, len(typedata))
+	for key, value := range typedata {
+		data[key] = value
+	}
+	return data
+}
+
+// This method produces a google-sheets client from a jwt.Config client by
+// retrieveing access token
+func GenerateServiceWithrefreshToken(jwtconfig jwt.Config) (*sheets.Service, error) {
```
SumanthPuram, line 221: `WithRefresh`

services/streammanager/googlesheets/googlesheetsmanager.go
utsabc: Fixed it

services/streammanager/googlesheets/googlesheetsmanager.go
utsabc: This is a specific `struct` exposed by google-sheets-api for marking which way we want to add the data. ROW or COLUMN along with from which Starting point in Particular Sheet. This is required for using the `sheets-service` APIs

services/streammanager/googlesheets/googlesheetsmanager.go
utsabc: Fixed all occurrences

services/streammanager/googlesheets/googlesheetsmanager.go
utsabc: Did a revamp of the existing code to support this. Added a `func` to handle insertion of data using sheets-service, which internally handles if there is a token issue in the error by creating a new client and calls itself recursively in order to prevent the current event-loss.

Add inline comments explaining functionality

services/streammanager/googlesheets/googlesheetsmanager.go
utsabc: Fixed

ruchiramoitra (approved):

psrikanth88 (approved):

arnab-p (approved):

PR #927, opened 2021-04-14
success upload exists

Jayachand:
**Fixes** # (*issue*)

## Type of change
## Checklist:


warehouse/warehouse.go
```diff
@@ -725,6 +729,59 @@ func (wh *HandleT) runUploadJobAllocator() {
 	}
 }
 
+func (wh *HandleT) uploadStatusTrack() {
+	for {
+		time.Sleep(uploadStatusTrackFrequency)
```
psrikanth88, line 734: put this sleep at the end of function instead? just to get atleast one stat when pod restarts

warehouse/warehouse.go
```diff
@@ -725,6 +729,59 @@ func (wh *HandleT) runUploadJobAllocator() {
...
+			}
+
+			sqlStatement := fmt.Sprintf(`
+				select 
+					cast( case when count(*) > 0 then 1 else 0 end as bit ), 
+						(select count(*) from %[7]s where source_id='%[2]s' and destination_id='%[3]s' and created_at > now() - interval '%[6]d MIN') as staging_files_count 
+				from %[1]s where source_id='%[2]s' and destination_id='%[3]s' and (status='%[4]s' or status='%[5]s') and updated_at > now() - interval '%[6]d MIN'`,
+				warehouseutils.WarehouseUploadsTable, source.ID, destination.ID, ExportedData, Aborted, timeWindow, warehouseutils.WarehouseStagingFilesTable)
+
+			rows, err := wh.dbHandle.Query(sqlStatement)
```
psrikanth88, line 763: make it `QueryRow` since only 1 row is returned from query always

warehouse/warehouse.go
```diff
@@ -725,6 +729,59 @@ func (wh *HandleT) runUploadJobAllocator() {
...
+
+			timeWindow := uploadBufferTimeInMin
+			if value, err := strconv.Atoi(syncFrequency); err == nil {
+				timeWindow += value
+			}
+
+			sqlStatement := fmt.Sprintf(`
+				select 
+					cast( case when count(*) > 0 then 1 else 0 end as bit ), 
+						(select count(*) from %[7]s where source_id='%[2]s' and destination_id='%[3]s' and created_at > now() - interval '%[6]d MIN') as staging_files_count
```
psrikanth88, line 759: let's do the staging file count query separately first and then do the uploads query if there are staging files? Also, for staging files let us do something like `select created_at from wh_staging_files order by created_at desc limit 1` and compare it with time window. This might be faster than filtering on time window in query

psrikanth88 (changes requested):

warehouse/warehouse.go
Jayachand: done

warehouse/warehouse.go
Jayachand: done

warehouse/warehouse.go
Jayachand: done

psrikanth88 (approved):

sivashanmukh (approved):
LGTM

Issue #569, opened 2020-09-30
docker images tags do not correspond with releases


jalberto:
will be very useful if docker image tags use the same version system as the release in Github

jalberto:
Lates tag is from 7mo ago

![image](https://user-images.githubusercontent.com/99927/96117910-12c5fa00-0eeb-11eb-8eeb-25d058707a21.png)

while lates stable 0.1.9 release is from 20 Jun

yanivyalda:
agreed it's confusing. We don't know which version to use as well.

marcossv9:
+1 to this. Another good option could be to add a **stable** docker image also. Every one-two days there is a new docker-image.. It is not good practice..

jalberto:
@marcossv9 a fixed tag will bring the same issues, the ideal solution is to have a new tag for each release, but that tag needs to be in sync with the repo.

My perception is rudder lads are not too motivated to make the self-hosted option easier to use.

SumanthPuram:
@jalberto @marcossv9 @yanivyalda 
Apologies for the delayed response. We did chat internally and wanted to get your feedback. The following PR is our plan towards managing the release tags in docker and deprecating Github releases. 

Please let us know your feedback on this.


https://github.com/rudderlabs/rudder-server/pull/999

chandumlg:
We now follow semver and docker tags are same as the releases.

PR #999, opened 2021-05-19
Publish open source release policy

SumanthPuram:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,0 +1,8 @@
+### Open source Releases
+
+GitHub releases are deprecated. All releases will be made on [Docker Hub](https://hub.docker.com/repository/docker/rudderlabs/rudder-server).
+
+There will be a monthly `rudder-server` release in Dockerhub. The tag will be the date and time when the release is created (Eg. `18052021.042954`).
+
+- If you want to auto-update to the latest tag, please use `1-alpine` tag. This will be automatically updated after every release.
+- If you want to stay on a release, use the time based tag instead (Eg. `18052021.042954`).


jalberto:
My 2 cents coming from #569 

Software versioning is not just a number or counter, but a sensible way to answer these questions:

- Which version I am using?
  - This is solved with the current proposal, as each docker image will have its own version
  - as the proposed versioning is date-based, and created once per month if there is a hotfix on the very same day, will be confusing. Which one is the hotfix? `18052021.042954` or `18052021.042854`
- Which codebase/commit/tag am I running?
  - This is not solved with the proposal, as there is not a pointer to the code being used in the docker image
  - ideally, this pointer will be a tag in the git repo
  - proposed version format will  work as a tag in git
- Which deps should I use with this version?
  - as rudder-server needs to coordinate with other rudder services like transformers 
  - how I know if `server-18052021.042954` will work with `transformer-10102021.0000`?
  - here is where date-based versioning has issues. is easier to understand SemVer for this, so `server-1.2.9` will work with `transformers-1.2.x`
- it is safe to upgrade or there are breaking changes?
  - here again, date-based has issues, as by date you don't know even if it is a stable version or not
  - SemVer main reason is to solve this, and mostly deps managers will understand it out of the box
- where I check for changes for this version compared with the previous version?
  - so each month I will need to de-compile the docker-image, to know which are the breaking changes?
  - the changleog can use date-based or SemVer as far is clear which changes are in which version

SumanthPuram:
Great, thanks for these details.

To summarise
1. rudder-server and rudder-transformer releases should use sem-ver maintaining compatibility
2. Create tags in GitHub corresponding to releases
3. Add change log as part of Github release

Anything else that you want to suggest?

jalberto:
Great summary @SumanthPuram I tend to over-explain my thoughts :)

I think that will cover all. To clarify I am not suggesting to move from based-date to SemVer, as both can work, but I think is harder with date-based versioning.

Thanks!

SumanthPuram:
Thanks! We will work on this, will keep you posted.

jalberto:
Any progress on this?

mderazon:
@SumanthPuram any way to track breaking changes between releases of Rudderstack ?
For example, changes in transformers like https://github.com/rudderlabs/rudder-transformer/pull/1091 break analytics but I can't find a way to know about these things when upgrading to a new Docker version

mderazon:
https://github.com/rudderlabs/rudder-server/blob/6996d7eeae8f0c85c5c89b51264671e3d6ba68d7/releases.md?plain=1#L3

Is this updated ? I don't see any semver tag in Dockerhub

jalberto:
I really think there is no interest in having a good enough versioning system that allows open-source users to have a more reliable installation

PR #1000, opened 2021-05-19
Pixel API Response: pixel output issue

shrouti1507:


gateway/gateway.go
```diff
@@ -785,6 +794,48 @@ func (gateway *HandleT) webRequestHandler(rh RequestHandler, w http.ResponseWrit
...
+	var errorMessage string
+	defer func() {
+		if errorMessage != "" {
+			gateway.logger.Debug(errorMessage)
+			//http.Error(w, response.GetStatus(errorMessage), 400)
+		}
+	}()
+	payload, writeKey, err := gateway.getPayloadAndWriteKey(w, r)
+	if err != nil {
+		errorMessage = err.Error()
```
ruchiramoitra, line 810: if error is there directly return from here 200 and image.

arnab-p (approved):

arnab-p (approved):

gateway/gateway.go
```diff
@@ -785,6 +790,29 @@ func (gateway *HandleT) webRequestHandler(rh RequestHandler, w http.ResponseWrit
...
+	payload, writeKey, err := gateway.getPayloadAndWriteKey(w, r)
+	if err != nil {
+		return
+	}
+	errorMessage = rh.ProcessRequest(gateway, &w, r, reqType, payload, writeKey)
+
+	atomic.AddUint64(&gateway.ackCount, 1)
+	gateway.trackRequestMetrics(errorMessage)
+
+	gateway.logger.Debug(response.GetStatus(""))
```
chandumlg, line 813: Remove this log

gateway/gateway.go
```diff
@@ -603,6 +603,7 @@ func (gateway *HandleT) webGroupHandler(w http.ResponseWriter, r *http.Request)
 }
 
 func (gateway *HandleT) pixelPageHandler(w http.ResponseWriter, r *http.Request) {
+
```
chandumlg, line 606: remove this empty line

gateway/gateway.go
```diff
@@ -917,34 +945,24 @@ func (gateway *HandleT) setWebPayload(r *http.Request, qp url.Values, reqType st
...
-			// make a new request
-			req, _ := http.NewRequest(http.MethodPost, "", nil)
 
-			// set basic auth header
-			req.SetBasicAuth(writeKey[0], "")
-			delete(queryParams, "writeKey")
+	queryParams := r.URL.Query()
+	writeKey := queryParams["writeKey"]
+	// make a new request
+	req, _ := http.NewRequest(http.MethodPost, "", nil)
```
chandumlg, line 952: handle error here.

gateway/gateway.go
```diff
@@ -917,34 +945,24 @@ func (gateway *HandleT) setWebPayload(r *http.Request, qp url.Values, reqType st
...
-			err := gateway.setWebPayload(req, queryParams, reqType)
-			if err != nil {
-				http.Error(w, err.Error(), http.StatusBadRequest)
-				return
-			}
+	// set X-Forwarded-For header
+	req.Header.Add("X-Forwarded-For", r.Header.Get("X-Forwarded-For"))
+
+	// convert the pixel request(r) to a web request(req)
+	_ = gateway.setWebPayload(req, queryParams, reqType)
```
chandumlg, line 962: handle error here.

gateway/gateway.go
shrouti1507: error is ignored because we are ideally trying to send 200 with pixel value irrespective of any situation

gateway/gateway.go
shrouti1507: error is ignored because we are ideally trying to send 200 with pixel value irrespective of any situation

gateway/gateway.go
```diff
@@ -785,6 +790,29 @@ func (gateway *HandleT) webRequestHandler(rh RequestHandler, w http.ResponseWrit
...
+	atomic.AddUint64(&gateway.recvCount, 1)
+	var errorMessage string
+	defer func() {
+		if errorMessage != "" {
+			gateway.logger.Debug(errorMessage)
+		}
+	}()
+	payload, writeKey, err := gateway.getPayloadAndWriteKey(w, r)
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
shrouti1507: errorMessage is ignored because we are ideally trying to send 200 with pixel value irrespective of any situation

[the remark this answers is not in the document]
chandumlg: We log the errorMessage in the defer func() above.

gateway/gateway.go
chandumlg: Server will crash if above statement fails with error. Handle the error and send 200.

chandumlg (approved):

ruchiramoitra (approved):
LGTM

gateway/gateway.go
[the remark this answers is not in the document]
SumanthPuram: Why are we returning 200 for failures? Is that expected?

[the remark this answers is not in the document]
SumanthPuram: @shrouti1507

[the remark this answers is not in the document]
chandumlg: From @arnab-p 
"The idea was to send success always irrespective of success/failure. That way we can ensure that a broken image won’t be shown. They’re OK even if we silently drop the event"

arnab-p (approved):

PR #1100, opened 2021-06-28
wh: support pending-events and trigger-upload APIs

dhawal1248:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/warehouse/utils/utils.go
+++ b/warehouse/utils/utils.go
@@ -168,6 +168,20 @@ type QueryResult struct {
 	Values  [][]string
 }
 
+type PendingEventsRequest struct {
+	SourceID  string `json:"source_id"`
+	TaskRunID string `json:"task_run_id"`
+}
+
+type PendingEventsResponse struct {
+	PendingEvents bool `json:"pending_events"`
+}
+
+type TriggerUploadRequest struct {
+	SourceID      string `json:"source_id"`
+	DestinationID string `json:"destination_id"`
+}
+
 func TimingFromJSONString(str sql.NullString) (status string, recordedTime time.Time) {
 	timingsMap := gjson.Parse(str.String).Map()
 	for s, t := range timingsMap {
--- a/warehouse/utils/utils.go
+++ b/warehouse/utils/utils.go
@@ -168,16 +168,16 @@ type QueryResult struct {
 	Values  [][]string
 }
 
-type PendingEventsRequest struct {
+type PendingEventsRequestT struct {
 	SourceID  string `json:"source_id"`
 	TaskRunID string `json:"task_run_id"`
 }
 
-type PendingEventsResponse struct {
+type PendingEventsResponseT struct {
 	PendingEvents bool `json:"pending_events"`
 }
 
-type TriggerUploadRequest struct {
+type TriggerUploadRequestT struct {
 	SourceID      string `json:"source_id"`
 	DestinationID string `json:"destination_id"`
 }
--- a/warehouse/utils/utils.go
+++ b/warehouse/utils/utils.go
@@ -174,7 +174,9 @@ type PendingEventsRequestT struct {
 }
 
 type PendingEventsResponseT struct {
-	PendingEvents bool `json:"pending_events"`
+	PendingEvents            bool `json:"pending_events"`
+	PendingStagingFilesCount int64  `json:"pending_staging_files"`
+	PendingUploadCount       int64  `json:"pending_uploads"`
 }
 
 type TriggerUploadRequestT struct {
--- a/warehouse/warehouse.go
+++ b/warehouse/warehouse.go
@@ -256,14 +256,7 @@ func (wh *HandleT) backendConfigSubscriber() {
 				wh.workerChannelMapLock.Unlock()
 
 				destinationsMapLock.Lock()
-				// TODO: ask if the `warehouse` var defined above can be used here
-				destinationsMap[destination.ID] = warehouseutils.WarehouseT{
-					Source:      source,
-					Destination: destination,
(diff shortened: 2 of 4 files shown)


warehouse/warehouse.go
```diff
@@ -1106,6 +1341,8 @@ func startWebHandler() {
 	if isMaster() {
 		backendconfig.WaitForConfig()
 		http.HandleFunc("/v1/process", processHandler)
+		http.HandleFunc("/v1/pending-events", pendingEventsHandler)
```
psrikanth88, line 1344: let's make it `/v1/warehouse/pending-events`. Should have done it ideally for /process too but we should start doing it for new routes

warehouse/utils/utils.go
```diff
@@ -168,6 +168,20 @@ type QueryResult struct {
 	Values  [][]string
 }
 
+type PendingEventsRequest struct {
```
psrikanth88, line 171: lets add a T at the end `PendingEventsRequestT` ? just check once if that's what we are doing for other structs too

warehouse/warehouse.go
```diff
@@ -1066,6 +1079,228 @@ func processHandler(w http.ResponseWriter, r *http.Request) {
...
+	pendingStagingFileCount, err := getPendingStagingFileCount(sourceID)
+	if err != nil {
+		err := fmt.Errorf("Error getting pending staging file count : %v", err)
+		pkgLogger.Errorf("[WH]: %v", err)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	// get pending uploads
+	pendingUploadCount, err := getPendingUploadCount(sourceID)
```
psrikanth88, line 1126: do this only if pendingStagingFileCount == 0 ?

warehouse/warehouse.go
```diff
@@ -1066,6 +1079,228 @@ func processHandler(w http.ResponseWriter, r *http.Request) {
...
+								WHERE %[1]s.id > %[2]v AND %[1]s.source_id='%[3]s'`,
+		warehouseutils.WarehouseStagingFilesTable, lastStagingFileID, sourceID)
+
+	err = dbHandle.QueryRow(sqlStatement).Scan(&fileCount)
+	if err != nil && err != sql.ErrNoRows {
+		err = fmt.Errorf("Query: %s failed with Error : %w", sqlStatement, err)
+		return
+	}
+
+	return
```
psrikanth88, line 1202: explicitly return nil error ? since sql.ErrNoRows can also be considered not an error for our use case

warehouse/warehouse.go
```diff
@@ -1066,6 +1079,228 @@ func processHandler(w http.ResponseWriter, r *http.Request) {
...
+								WHERE %[1]s.status NOT IN ('%[2]s', '%[3]s') AND %[1]s.source_id='%[4]s'
+	`, warehouseutils.WarehouseUploadsTable, ExportedData, Aborted, sourceID)
+
+	err = dbHandle.QueryRow(sqlStatement).Scan(&uploadCount)
+	if err != nil && err != sql.ErrNoRows {
+		err = fmt.Errorf("Query: %s failed with Error : %w", sqlStatement, err)
+		return
+	}
+
+	return
```
psrikanth88, line 1217: return nil error same as above

psrikanth88 (approved):

psrikanth88 (approved):

warehouse/warehouse.go
```diff
@@ -1066,6 +1079,232 @@ func processHandler(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	wh := make([]warehouseutils.WarehouseT, 0)
+
+	if sourceID != "" && destID == "" {
+		// get all wh destinations for given source id
+		destinationsMapLock.Lock()
+		for _, dst := range destinationsMap {
```
psrikanth88, line 1262: this will only give last source of a given destination. some sources might be missing if we use this map.

psrikanth88 (changes requested):

warehouse/warehouse.go
```diff
@@ -1066,6 +1079,232 @@ func processHandler(w http.ResponseWriter, r *http.Request) {
 	recordStagedRowsStat(stagingFile.TotalEvents, stagingFile.BatchDestination.Destination.DestinationDefinition.Name, stagingFile.BatchDestination.Destination.ID, stagingFile.BatchDestination.Source.Name, stagingFile.BatchDestination.Destination.Name)
 }
 
+func pendingEventsHandler(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
dhawal1248: Added file counts to response.Didn't find any other places where we could add logs.

warehouse/warehouse.go
```diff
@@ -1066,6 +1093,239 @@ func processHandler(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+
+	// if there are any pending staging files or uploads, set pending events as true
+	if (pendingStagingFileCount + pendingUploadCount) > int64(0) {
+		pendingEvents = true
+	}
+
+	// trigger upload if there are pending events
+	if pendingEvents {
```
psrikanth88, line 1158: let us take a qquery param along with the endpoint `/v1/warehouse/pending-events?triggerPending=true` and only trigger if true so that we can also use the endpoint to just get pending events without triggering. This can be used later for features like disabling Sync Now button in case there is a pending upload going on etc.

psrikanth88 (approved):

psrikanth88 (approved):

PR #1163, opened 2021-08-10
wh dest sync now

bipin-rudder:
**Fixes** # (*issue*)

## Type of change
## Checklist:


warehouse/warehouse.go
```diff
@@ -1349,6 +1361,58 @@ func triggerUploadHandler(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 }
 
+func TriggerUploadHandler(srcId string, destId string) (error){
```
psrikanth88, line 1364: should we DRY this block since we have it in warehouse.go too

proto/warehouse/warehouse.proto
```diff
@@ -11,6 +11,8 @@ service Warehouse {
   rpc GetWHUploads (WHUploadsRequest) returns (WHUploadsResponse);
   rpc GetWHUpload (WHUploadRequest) returns (WHUploadResponse);
   rpc TriggerWHUpload (WHUploadRequest) returns (google.protobuf.Empty);
+  rpc TriggerWHUploads (WHUploadsRequest) returns (google.protobuf.Empty);
```
psrikanth88, line 14: so, we would not respond with something like there are no pending files that need to be uploaded? can we use the getpendingstagingfiles func to quickly check that before triggering or something

psrikanth88 (approved):

psrikanth88 (approved):

PR #1167, opened 2021-08-12
filter out unsupported message types

arajguha:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1190,7 +1190,7 @@ func (proc *HandleT) processJobsForDest(jobList []*jobsdb.JobT, parsedEventList 
 		// OR
 		//router and transformer supports router transform, then no destination transformation happens.
 		if transformAt == "none" || (transformAt == "router" && transformAtFromFeaturesFile != "") {
-			response = ConvertToTransformerResponse(eventsToTransform)
+			response = ConvertToFilteredTransformerResponse(eventsToTransform, transformAt != "none")
 		} else {
 			destTransformationStat.transformTime.Start()
 			response = proc.transformer.Transform(eventsToTransform, url, transformBatchSize)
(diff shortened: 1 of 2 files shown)


processor/processor.go
```diff
@@ -1167,6 +1167,31 @@ func (proc *HandleT) processJobsForDest(jobList []*jobsdb.JobT, parsedEventList
...
 
+		// filter out events with unsupported message types
+		filteredEvents := []transformer.TransformerEventT{}
+		for _, event := range eventsToTransform {
+			destinationDef := event.Destination.DestinationDefinition
+			supportedTypes, ok := destinationDef.Config["supportedMessageTypes"]
+			if ok {
+				messageType := event.Message["type"].(string)
+				found := false
+				for _, typ := range supportedTypes.([]interface{}) {
```
cisse21, line 1178: Can we use misc.Contains to check whether the given type exists or not ?

processor/processor.go
```diff
@@ -1167,6 +1167,31 @@ func (proc *HandleT) processJobsForDest(jobList []*jobsdb.JobT, parsedEventList
...
+					if messageType == typ.(string) {
+						found = true
+						filteredEvents = append(filteredEvents, event)
+					}
+				}
+				if !found {
+					pkgLogger.Info("Message Type " + messageType + " not supported for Destination " + destinationDef.Name)
```
cisse21, line 1185: Would it be better to have it as a Debug Log rather than an Info log as it might be very high in number

cisse21 (changes requested):

cisse21 (approved):

processor/processor.go
```diff
@@ -1398,14 +1398,43 @@ func (proc *HandleT) processJobsForDest(jobList []*jobsdb.JobT, parsedEventList
...
+			messageType, typOk := event.Message["type"].(string)
+			if !typOk {
+				// add to FailedEvents
+				errMessage = "Invalid message type. Type assertion failed"
+				resp = transformer.TransformerResponseT{Output: event.Message, StatusCode: 400, Metadata: event.Metadata, Error: errMessage}
+				failedEvents = append(failedEvents, resp)
+				continue
+			}
+
+			if misc.Contains(supportedTypes, messageType) {
```
psrikanth88, line 1421: trim the messageType and do case insensitive match just in case?

processor/processor.go
```diff
@@ -1398,14 +1398,43 @@ func (proc *HandleT) processJobsForDest(jobList []*jobsdb.JobT, parsedEventList
...
+				resp = transformer.TransformerResponseT{Output: event.Message, StatusCode: 200, Metadata: event.Metadata}
+				responses = append(responses, resp)
+			} else {
+				// add to FailedEvents
+				errMessage = "Message type " + messageType + " not supported"
+				resp = transformer.TransformerResponseT{Output: event.Message, StatusCode: 400, Metadata: event.Metadata, Error: errMessage}
```
psrikanth88, line 1427: Should we set a different field in response struct, so that it is saved so in the proc_err tables and can be also used for use cases like filter out when sending events to live events (if we do so), skip in replay etc

processor/processor.go
```diff
@@ -1398,14 +1398,43 @@ func (proc *HandleT) processJobsForDest(jobList []*jobsdb.JobT, parsedEventList
 	proc.pStatsDBW.Print()
 }
 
-func ConvertToTransformerResponse(events []transformer.TransformerEventT) transformer.ResponseT {
+func ConvertToTransformerResponse(events []transformer.TransformerEventT, filterUnsupportedMessageTypes bool) transformer.ResponseT {
```
psrikanth88, line 1401: modify name of func to also reflect that we are filtering by message type ?

psrikanth88 (changes requested):

processor/processor.go
chandumlg: done

processor/processor.go
chandumlg: save or not to proc_error will be taken up in another PR.

chandumlg (approved):

arnab-p (approved):

PR #1192, opened 2021-08-23
Introducing ctx for graceful shutdown

lvrach:
**Fixes** # (*issue*)

## Type of change

## Dependencies

[Ctx graceful shutdown](https://github.com/rudderlabs/rudder-server-enterprise/pull/34) for migration feature.

### Motivation 

While trying to automate the integration testing of the service, I was able to gracefully bring the service to a stop. Moreover, because the testing suite was cleaning up the external dependenes after the test finished, the service was fataling , causing the test to always return with an error.

### Goal 
The primary goal of this PR, is to make sure that no goroutine that access the database is running when the main returns.

The secondary goal, which is important as well, is to enable graceful termination so not issues are caused in production during a shutdown event. 

Please refer to [internal doc](https://www.notion.so/rudderstacks/Graceful-termination-of-rudder-server-cb7ce31a9e954361acf0f0ceadb0c00f) for further information on the topic.

### Issue

Currently when a TERM signal is received, the processing in all go routines (including http handling) will be interrupted abruptly for an  [`os.Exit(1)`](https://github.com/rudderlabs/rudder-server/blob/master/main.go#L169).

### Approach

1. Propagate the TERM signal in every running process via context.Context. 
2. Modify processes to be cancelable. 
3. Wait for all process / go-routines to finish before exiting the main go routine

### Introducing context

`ctx context.Context` is introduced as the first argument in many methods and functions

### http.Server shutdown
From:
```go
gateway.logger.Fatal(srv.ListenAndServe())
```

To:
```go
	g.Go(func() error {
		<-ctx.Done()
		return srv.Shutdown(ctx)
	})
	g.Go(func() error {
		return srv.ListenAndServe()
		// 	gateway.logger.Fatal() -> should we log or propagate the error ?
	})

	return g.Wait()
```

### for loops

From:
```go
    for {
        doStuff()
        time.Sleep(...)
      }
```

To: 

```go

	for {
		select {
		case <-ctx.Done():
			return nil
		case <-time.After(time.Second * 5):
			doStuff()
		}
	}
```

### `go fn()` replaced by errgroup

```go
		go gateway.StartAdminHandler()
```

```go
	g.Go((func() error {
		return doStuff()
	})
	
	....
	
	g.Wait()
```

### rruntime replaced by errgroup

From:
```go
	rruntime.Go(func() {
		doStuff()
	})
```

To:
```go
	g.Go(misc.WithBugsnag(func() error {
		return doStuff()
	}))
	
	....
	
	g.Wait()
```

### Use pingCtx()

When searching for instances of `rruntime` I have stabled upon this pattern

```go
	pingResultChannel := make(chan error, 1)
	defer as.Db.Close()
	rruntime.Go(func() {
		pingResultChannel <- as.Db.Ping()
	})
	var timeOut time.Duration = 5
	select {
	case err = <-pingResultChannel:
	case <-time.After(timeOut * time.Second):
		err = fmt.Errorf("connection testing timed out after %d sec", timeOut)
	}
```

Which can be done, with:  https://pkg.go.dev/database/sql#Conn.PingContext

```go
	ctx, cancel := context.WithTimeout(context.TODO(), timeOut)
	defer cancel()

	err = as.Db.PingContext(ctx)
	if err == context.DeadlineExceeded {
		return fmt.Errorf("connection testing timed out after %d sec", timeOut/time.Second)
	}
```

`context.TODO()` -> Eventually context should be passed on those methods as well, for example if we want to shutdown without waiting for ping to timeout.

### .Shutdown() method

`.Shutdown() ` When shutdown is called all the goroutines associated with the service will stop and the channels will close. The method will block, until everything is properly terminated. No other method should be called after or during shutdown.

If  `.Shutdown()` is used, the method that spawn up the workers (usually `.Setup()`` is non-blocking and doesn't require context. For instance .Setup will run the go routines on the background, a context and an errgroup will be used internally to handle their termination.

To achieve this, the following properties where added to the struct:

```go
	backgroundWait   func() error
	backgroundCancel context.CancelFunc
```

In simple case, Shutdown is doing:
```go
func (srv ..) Shutdown() {
    srv.backgroundCancel()
    srv.backgroundWait()
} 
```

`Shutdown()` used for the following reasons:
1. Shutdown is complex and termination of internal components should happen in a specific order.
2. We want to make sure no other service is using it, before it terminates. With Shutdown we can easily facillitate this, usually by doing:
```go
service.Setup()
defer service.Shutdown()
```
3. When the `.Setup()` method should run and return successfully,  before any other method could be called. The setup with the execution of workers could be separated in a diffrent method, but the refactoring was not straight forward and in most cases reasons (1.) and (2.) where also at play.

### draining/closing channels

When it comes to channel the naive approach:

```go
	for {
		select {
		case <-ctx.Done():
			return nil
		case <-ch:
			doStuff()
		}
	}
```

Will not work:
1. If the channel is buffered, messages in the channel will be lost.
2. The producer/publisher `ch <- {}` might block forever, either the channel is unbuffered or full and the consumer has already exited.

Thus, it is important to:
 1. Stop go routines/workers, only when their channels is closed. Either doing a `data, hasMore := <-ch` or `data := range ch`.
 2. The go routines/ worker that publish messages to a channel should be responsible for closing their channel after they are done.
 
 
Sometimes the service has no direct control of the publishers of a channel. A very common occurrence was exposed methods publishing to a channel. In this case, the `.Shutdown()` method is responsible for closing the channel. In this case, it is import for no other struct method to be called during or after  `.Shutdown()`. Reminding here, that publishing to a closed channel will lead to panic. `backgroundCtx.Err()` could be used to determine if the service is Shutting down, and return an error before attempting to publish to the channel  

If multiple workers are publish to the channel, the close should happen after all the workers have finished. An additional `errgroup` is used to facilitate this.
   

## Risks

Potential risk from the introduced changes:

### sync calls

A lot of calls have become synchronous/blocking. If they are not handled correctly, they could cause the execution in the block in an unexpected place without initializing everything. 
 
They can even cause dead-locks, for instance:
```go
	moduleLoadLock.Lock()
	defer moduleLoadLock.Unlock()

```

later in the method was a non blocking call, spawning a go routine
```go
    processorInstance.Start()
```

Once `processorInstance.Start(ctx)`  was converted to a blocking call, `moduleLoadLock.Unlock()` won't be call, thus preventing other methods to initialise.

## How to test

run `go run main.go --cleardb --normal-mode`
then `Ctrl+C`

### Inspect  

By applying the following patch: 

```
 
+       _ "net/http/pprof"
+
        // This is necessary for compatibility with enterprise features
        _ "github.com/rudderlabs/rudder-server/imports"
 )
@@ -106,6 +108,10 @@ func main() {
                cancel()
        }()
 
+       go func() {
+               fmt.Println(http.ListenAndServe("localhost:6060", nil))
+       }()
+
        options := app.LoadOptions()
 ```

Will expose:
http://localhost:6060/debug/pprof/goroutine?debug=1

You can use this endpoint to understand which goroutines are preventing the shutdown.

Moreover, to verify that goroutines have closed gracefully, you can add:
```
@@ -216,6 +222,9 @@ func main() {
                logger.Log.Sync()
        }
        stats.StopRuntimeStats()
+
+       chh := make(chan os.Signal, 1)
+       <-chh
 }
```

After the following message:
``
`2021-09-06T17:13:38.913+0300    INFO    rudder-server/main.go:215       Graceful terminal after 4.988371842s, with 72 go-routines
```
You can use http://localhost:6060/debug/pprof/goroutine?debug=1 for remaining go-routines.

Expected go-routines patterns for this PR:
```
backendConfigSubscriber
debugger.
enterprise/reporting
enterprise/suppress
```
And other non-rudder-server related

After applying this, to kill the service:
```sh
sudo lsof -i tcp:6060 | grep LISTEN  
kill -9 <pid>
```

## Checklist:


gateway/gateway.go
```diff
@@ -198,26 +202,65 @@ func (gateway *HandleT) initUserWebRequestWorkers() {
...
+			return nil
 		})
 	}
+	g.Wait()
+
+	close(gateway.userWorkerBatchRequestQ)
+}
+
+func (gateway *HandleT) Shutdown() {
+	fmt.Println("gateway shutdown start")
```
lvrach, line 232:
```suggestion
```

gateway/gateway.go
```diff
@@ -198,26 +202,65 @@ func (gateway *HandleT) initUserWebRequestWorkers() {
...
+
+	if gateway.httpWebServer != nil {
+		gateway.httpWebServer.Shutdown(context.Background())
+		fmt.Println("gateway httpWebServer shutdown")
+	}
+	gateway.backgroundCancel()
+	gateway.webhookHandler.Shutdown()
+	fmt.Println("gateway webhookHandler shutdown")
```
lvrach, line 240:
```suggestion
```

gateway/gateway.go
```diff
@@ -198,26 +202,65 @@ func (gateway *HandleT) initUserWebRequestWorkers() {
...
+
+	// UserWebRequestWorkers
+	for _, worker := range gateway.userWebRequestWorkers {
+		close(worker.webRequestQ)
+	}
+	fmt.Println("close worker")
```
lvrach, line 246:
```suggestion
```

jobsdb/jobsdb.go
```diff
@@ -617,52 +646,47 @@ func (jd *HandleT) workersAndAuxSetup(ownerType OwnerType, tablePrefix string, r
...
 	config.RegisterBoolConfigVariable(true, &jd.enableReaderQueue, true, enableReaderQueueKeys...)
-	jd.storeChannel = make(chan writeJob)
-	jd.storeWithRetryChannel = make(chan writeJob)
-	jd.updateJobStatusChannel = make(chan writeJob)
-	jd.deleteExecutingChannel = make(chan writeJob)
+	// jd.storeChannel = make(chan writeJob)
+	// jd.storeWithRetryChannel = make(chan writeJob)
+	// jd.updateJobStatusChannel = make(chan writeJob)
+	// jd.deleteExecutingChannel = make(chan writeJob)
+	jd.writeChannel = make(chan writeJob)
```
lvrach, line 653: Do we have a use for these channels ?

jobsdb/jobsdb.go
```diff
@@ -617,52 +643,52 @@ func (jd *HandleT) workersAndAuxSetup(ownerType OwnerType, tablePrefix string, r
...
-	})
+func (jd *HandleT) startMigrateDSLoop(ctx context.Context) {
+	jd.backgroundGroup.Go(misc.WithBugsnag(func() error {
+		jd.migrateDSLoop(jd.backgroundCtx)
+		return nil
+	}))
 }
 
-func (jd *HandleT) readerSetup() {
+func (jd *HandleT) readerSetup(ctx context.Context) {
```
cisse21, line 691: here ctx and jd.backgroundCtx is the same .. Can we eliminate backgroundCtx if it isn't required

gateway/gateway.go
```diff
@@ -1350,7 +1428,7 @@ Setup initializes this module:
 
 This function will block until backend config is initialy received.
 */
-func (gateway *HandleT) Setup(application app.Interface, backendConfig backendconfig.BackendConfig, jobsDB jobsdb.JobsDB, rateLimiter ratelimiter.RateLimiter, versionHandler func(w http.ResponseWriter, r *http.Request)) {
+func (gateway *HandleT) Setup(ctx context.Context, application app.Interface, backendConfig backendconfig.BackendConfig, jobsDB jobsdb.JobsDB, rateLimiter ratelimiter.RateLimiter, versionHandler func(w http.ResponseWriter, r *http.Request)) {
```
cisse21, line 1431: Do we need to pass the ctx to this function since we don't seem to be using this anywhere

router/batchrouter/batchrouter.go
```diff
@@ -1349,9 +1355,13 @@ func (worker *workerT) workerProcess() {
...
+			_, hasMore := <-worker.resumeChannel
+			if hasMore {
+				pkgLogger.Infof("Batch Router worker %d is resumed. Dest type: %s", worker.workerID, worker.brt.destType)
+			} else {
+				// TODO: SHOULD WE CLEAN UP ?
+				return
+			}
 		case batchDestData, hasMore := <-brt.processQ:
 			if !hasMore {
 				// TODO: SHOULD WE CLEAN UP ?
```
lvrach, line 1367: @chandumlg is anything we should do here, in order to cleanup before shutdown ?

cisse21 (approved):

config/config.yaml
```diff
@@ -3,7 +3,7 @@ gwDBRetention: 0h
 routerDBRetention: 0h
 enableProcessor: true
 enableRouter: true
-enableStats: true
+enableStats: false
```
chandumlg, line 6: Revert

chandumlg (approved):

PR #1200, opened 2021-08-24
Config Polling Improvements

Sidddddarth:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/config/backend-config/backend-config.go
+++ b/config/backend-config/backend-config.go
@@ -45,6 +45,7 @@ var (
 	maxRegulationsPerRequest              int
 	configEnvReplacementEnabled           bool
 	successfulQueryTimeStamp              time.Time
+	useCacheForHostedConfig               bool
 
 	//DefaultBackendConfig will be initialized be Setup to either a WorkspaceConfig or MultiWorkspaceConfig.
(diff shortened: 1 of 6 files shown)


Sidddddarth:
Tests pending, will correct them if we can go ahead with this.

lvrach:
What is the motivation behind this PR? Can you provide a description of the problems we are trying to solve and an overview of the solution ?

Sidddddarth:
> What is the motivation behind this PR? Can you provide a description of the problems we are trying to solve and an overview of the solution ?

motivation: to improve customer experience(predominantly in hosted workspaces), where fetching config would take too long(order of 10s of seconds -> I checked just now and it took 25-35 seconds)

problems to be solved: the entire hosted config is very large(35MB) and takes too long to query, so the problem to solve is to reduce the latency and the size of the payload that's polled for every 5(say) seconds.

overview of the proposed solution: Fetching a diff in the config every 5  seconds instead of the entire config always. Only at startup does the server poll for a full config. The control-plane exposed a new endpoint to do so, rudder-server fetches the diff in config from that endpoint and applies it on top of the config that it already has.

PR #1176, opened 2021-08-20
Integration service tests using docker for dependencies - `init()` refactor version

lvrach:
**Fixes** # (*issue*)

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added unit tests for the code
- [ ] I have made corresponding changes to the documentation

### How to run:

```sh
go test -timeout 0 ./ -v
```

How does it look like:
```sh
=== RUN   TestWebhook
webhookurl http://127.0.0.1:58402
workspace config path: /var/folders/7j/5bfjw_rs303d_bl5tylsxwfm0000gn/T/workspaceConfig.293613532.json
serviceHealthEndpoint http://localhost:58400/health

....

 ... service logs ... 

....

{"appType": "EMBEDDED", "server":"UP", "db":"UP","acceptingEvents":"TRUE","routingEvents":"TRUE","mode":"NORMAL","goroutines":"643", "backendConfigMode": "JSON", "lastSync":"2021-09-06T12:28:21+03:00", "lastRegulationSync":""}
2021-09-06T12:28:44.692+0300    INFO    backend-config/workspace-config.go:104  Reading workspace config from JSON file
Sending Track Event
OK
Event Sent Successfully
--- PASS: TestWebhook (0.66s)
PASS
Test on hold, before cleanup
Press Ctrl+C to exit
```

Once you see:
```
Event Sent Successfully
--- PASS: TestWebhook (0.66s)
PASS
Test on hold, before cleanup
Press Ctrl+C to exit
```
You can `Ctrl+C` to exit.

PR #1221, opened 2021-09-06
Destination OAuth Error Handling

sanpj2292:
Implementation of Destination Oauth in Router for which some changes have also been made in Processor. 
This PR has been raised to transformer-proxy as we would need to do some testing using the network layer & also the diff between dest-oauth branch and master has many changes that are included in transformer-proxy. Hence to check the changes for destination OAuth clearly, the PR has been raised to transformer-proxy branch.

Kindly do not merge this PR

** The fmt.Print or fmt.Println or fmt.Printf 's  found anywhere in the PR will be removed before final merge**

**Fixes** # (*issue*)

## Type of change
## Checklist:


router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,179 @@
...
+		expiredToken: accessToken,
+	}
+	res, err := json.Marshal(refTokenBody)
+	if err != nil {
+		// TODO: Is this way ok ? - Check with team
+		return http.StatusBadRequest, err.Error()
+	}
+	resp, refErr := authErrHandler.client.Post(refreshUrl, "application/json; charset=utf-8", bytes.NewBuffer(res))
+	if refErr != nil {
+		panic(refErr)
```
utsabc, line 122: Maybe we should throw some 5XX error from here as we are refreshing at this poin

config/config.yaml
```diff
@@ -110,6 +110,8 @@ Router:
     forceHTTP1: true
     httpTimeout: 120s
     httpMaxIdleConnsPerHost: 32
+  BQSTREAM:
```
cisse21, line 113: need to add this config in the operator too if we are planning to only enable transformerProxy for this destination

processor/processor.go
```diff
@@ -1219,6 +1221,19 @@ func (proc *HandleT) processJobsForDest(jobList []*jobsdb.JobT, parsedEventList
 		connectionDetailsMap := make(map[string]*types.ConnectionDetails)
 		statusDetailsMap := make(map[string]*types.StatusDetail)
 		successCountMap := make(map[string]int64)
+
+		// This mapping is used to map destination with it's respective account for refresh token capability
+		destToAccountIdMapping := make(map[string]string)
+		for _, eventToTransform := range eventsToTransform {
+			// default value or if accountId doesn't exist or not convertible to string
+			destToAccountIdMapping[eventToTransform.Destination.ID] = ""
+			if x, found := eventToTransform.Destination.Config["accountId"]; found {
```
cisse21, line 1230: rename x to accountIdInterface?

config/config.yaml
sanpj2292: Then how will we enable the transformerProxy for this destination alone ?

cisse21: This file wont be present when deployed to production. To have this variable enabled for a particular destination we need to change it in rudderstack-operator repo

processor/processor.go
```diff
@@ -1265,6 +1282,8 @@ func (proc *HandleT) processJobsForDest(jobList []*jobsdb.JobT, parsedEventList
 				SourceTaskRunID: sourceTaskRunId,
 				SourceJobID:     sourceJobId,
 				SourceJobRunID:  sourceJobRunId,
+				WorkspaceId:     workspaceId, // Changed for Auth Error Implementation
```
cisse21, line 1285: We can remove the comment for this I guess

router/router.go
```diff
@@ -1849,3 +1860,66 @@ func (rt *HandleT) Resume() {
 
 	rt.paused = false
 }
+
+// Currently the retry logic has been implemented for only OAuth
+func (rt *HandleT) SendToTransformerProxyWithRetry(val integrations.PostParametersT,
```
cisse21, line 1865: why do we need to pass accessToken as an explicit parameter when it is already part of PostParametersT

cisse21, line 1865: why do we need to pass accessToken as an explicit parameter when it is already part of PostParametersT

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,178 @@
...
+	statusCode, resp := authErrHandler.processResponse(res, doErr)
+	authErrHandler.oauthErrHandlerReqTimerStat.End()
+	authErrHandler.logger.Debugf("[%s request] :: Disable Response received : %s", loggerNm)
+	return statusCode, resp
+}
+
+func (authErrHandler *OAuthErrResHandler) processResponse(resp *http.Response, err error) (statusCode int, respBody string) {
+	var respData []byte
+	if resp != nil && resp.Body != nil {
+		respData, _ = ioutil.ReadAll(resp.Body)
```
cisse21, line 155: handle the error during ReadAll. probably ignore this attempt and retry

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,178 @@
...
+		//Detecting content type of the respBody
+		contentTypeHeader = http.DetectContentType(respData)
+	}
+	//If content type is not of type "*text*", overriding it with empty string
+	if !(strings.Contains(strings.ToLower(contentTypeHeader), "text") ||
+		strings.Contains(strings.ToLower(contentTypeHeader), "application/json") ||
+		strings.Contains(strings.ToLower(contentTypeHeader), "application/xml")) {
+		respData = []byte("")
+	}
+	if err != nil {
```
cisse21, line 172: If error is not nil then we need not do the all the prior computations .Can we move this check to the top and return immediately?

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,178 @@
...
+		respData, _ = ioutil.ReadAll(resp.Body)
+		defer resp.Body.Close()
+	}
+	var contentTypeHeader string
+	if resp != nil && resp.Header != nil {
+		contentTypeHeader = resp.Header.Get("Content-Type")
+	}
+	if contentTypeHeader == "" {
+		//Detecting content type of the respBody
+		contentTypeHeader = http.DetectContentType(respData)
```
cisse21, line 164: Can't we ignore the part where we fetch the contentType from header and just reply on this?

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,178 @@
...
+	statusCode, response := authErrHandler.processResponse(refreshResponse, refreshErr)
+	authErrHandler.oauthErrHandlerReqTimerStat.End()
+	authErrHandler.logger.Debugf("[%s request] :: Refresh token response received : %s", loggerNm, response)
+	return statusCode, response
+}
+
+func (authErrHandler *OAuthErrResHandler) DisableDestination(destination backendconfig.DestinationT, workspaceId string) (statusCode int, respBody string) {
+	authErrHandler.oauthErrHandlerReqTimerStat.Start()
+	destinationId := destination.ID
+	disableURL := fmt.Sprintf("%s/workspaces/%s/destinations/%s/disable", configBEURL, workspaceId, destinationId)
```
cisse21, line 132: Does this endpoint disable the destination on ConfigBackend ?

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,178 @@
...
+	authErrHandler.tr = &http.Transport{}
+	//This timeout is kind of modifiable & it seemed like 10 mins for this is too much!
+	authErrHandler.client = &http.Client{Transport: authErrHandler.tr, Timeout: 5 * time.Minute}
+	authErrHandler.oauthErrHandlerReqTimerStat = stats.NewStat("router.processor.oauthErrorHandler_request_time", stats.TimerType)
+	authErrHandler.oauthErrHandlerNetReqTimerStat = stats.NewStat("router.oauthErrorHandler_network_request_time", stats.TimerType)
+}
+
+func (authErrHandler *OAuthErrResHandler) RefreshToken(workspaceId string, accountId string, accessToken string) (statusCode int, respBody string) {
+	authErrHandler.oauthErrHandlerReqTimerStat.Start()
+	if len(strings.TrimSpace(accessToken)) == 0 {
```
cisse21, line 105: use the helper function here in router/utils?

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,178 @@
...
+		hasExpired:   true,
+		expiredToken: accessToken,
+	}
+	res, err := json.Marshal(refTokenBody)
+	if err != nil {
+		panic(err)
+	}
+	refreshResponse, refreshErr := authErrHandler.client.Post(refreshUrl, "application/json; charset=utf-8", bytes.NewBuffer(res))
+	if refreshErr != nil {
+		return http.StatusBadRequest, err.Error()
```
cisse21, line 121: return refreshErr.Error()

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	authErrHandler.logger.Debugf("[%s request] :: Disable Request sent : %s", loggerNm)
+	authErrHandler.oauthErrHandlerNetReqTimerStat.Start()
+	res, doErr := authErrHandler.client.Do(req)
+	authErrHandler.oauthErrHandlerNetReqTimerStat.End()
+	if doErr != nil {
+		// Abort on receiving an error
+		return http.StatusBadRequest, err.Error()
+	}
+	statusCode, resp := authErrHandler.processResponse(res, doErr)
```
cisse21, line 146: why do we need to pass on error when we are returning it if not nil above

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	res, err := json.Marshal(refTokenBody)
+	if err != nil {
+		panic(err)
+	}
+	refreshResponse, refreshErr := authErrHandler.client.Post(refreshUrl, "application/json; charset=utf-8", bytes.NewBuffer(res))
+	if refreshErr != nil {
+		return http.StatusBadRequest, err.Error()
+	}
+	statusCode, response := authErrHandler.processResponse(refreshResponse, refreshErr)
```
cisse21, line 123: why do we need to pass on error when we are returning it if not nil above

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,178 @@
...
+		HasExpired:   refTokenBody.hasExpired,
+		ExpiredToken: refTokenBody.expiredToken,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return j, nil
+}
+
+func init() {
```
cisse21, line 88: https://github.com/rudderlabs/rudder-server/pull/1176/files

We have depricated the use of init(). Please refer the above PR

router/router.go
```diff
@@ -1849,3 +1860,66 @@ func (rt *HandleT) Resume() {
...
+	destinationJob types.DestinationJobT, accessToken string, retryCount int) (statusCode int, response string) {
+	var respStatusCode int
+	var respBodyTemp string
+
+	respStatusCode, respBodyTemp = rt.transformer.Send(val, rt.destName, accessToken)
+	if retryCount >= 2 {
```
cisse21, line 1871: use maxFailedCountForJob for retrycount check rather than hardcoding to 2

router/router.go
```diff
@@ -1849,3 +1860,66 @@ func (rt *HandleT) Resume() {
...
+		// TODO: Should we send any other error here ?
+		return http.StatusInternalServerError, err.Error()
+	}
+	workspaceId := destinationJob.JobMetadataArray[0].WorkspaceId
+	var stCd int = 0
+	var res string = ""
+	// Check the category
+	// Trigger the refresh endpoint/disable endpoint
+	if errOutput.Output.AuthErrorCategory == oauth.DISABLE_DEST {
+		stCd, res = rt.oauth.DisableDestination(destinationJob.Destination, workspaceId)
```
cisse21, line 1889: change stcd to StatusCode and res to response?

router/router.go
```diff
@@ -1849,3 +1860,66 @@ func (rt *HandleT) Resume() {
...
+			val.WorkspaceId = workspaceId
+			val.AccountId = accountId
+			// This will be "true" only when a refresh executed successfully
+			var accountSecret oauth.AccountSecret
+
+			if router_utils.IsNotEmptyString(res) {
+				if err := json.Unmarshal([]byte(res), &accountSecret); err != nil {
+					// Some problem with AccountSecret unmarshalling
+					return http.StatusInternalServerError, err.Error()
+				} else if !router_utils.IsNotEmptyString(accountSecret.AccessToken) {
```
cisse21, line 1909: need not check for expirationDate? What would be the case where we would be having an empty accessToken. Is it being retried on transformer side for such cases?

router/router.go
```diff
@@ -1849,3 +1860,66 @@ func (rt *HandleT) Resume() {
...
+					// Status is 200, but no accesstoken or expirationDate is sent
+					return http.StatusInternalServerError, `Empty Token cannot be processed further`
+				}
+			}
+			// Retry with Refreshed Token(variable "res" - contains refreshed access token & expirationDate)
+			return rt.SendToTransformerProxyWithRetry(val, destinationJob, res, retryCount)
```
cisse21, line 1915: res is the response from the API call , wouldn't it be a JSON ? or will we be returned just an access token as response

router/router.go
```diff
@@ -1849,3 +1860,66 @@ func (rt *HandleT) Resume() {
...
+	fmt.Printf("SendToProxy Response Code: %d\n", respStatusCode)
+	fmt.Printf("SendToProxy Response: %s\n", respBodyTemp)
+	var errOutput oauth.ErrorOutput
+	if err := json.Unmarshal([]byte(respBodyTemp), &errOutput); err != nil {
+		// If respBodyTemp comes out with a string, then this will occur
+		// TODO: Should we send any other error here ?
+		return http.StatusInternalServerError, err.Error()
+	}
+	workspaceId := destinationJob.JobMetadataArray[0].WorkspaceId
+	var stCd int = 0
```
cisse21, line 1884: need not initialise with values, they would be defaulted to 0 and ""

router/router.go
```diff
@@ -550,6 +557,7 @@ func (worker *workerT) handleWorkerDestinationJobs() {
 					respStatusCode, respBody = worker.rt.customDestinationManager.SendData(destinationJob.Message, sourceID, destinationID)
 				} else {
 					result := getIterableStruct(destinationJob.Message, transformAt)
+					fmt.Println(destinationJob)
```
cisse21, line 560: remove fmt.Println

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: Yes

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: Sure, makes sense

router/router.go
sanpj2292: accessToken parameter(method arg) is like a refreshed access token(which will be obtained after one refresh token retrial)
The one part of PostParamtersT will be set using this paramter(accessToken). Let me know if this makes sense

router/router.go
sanpj2292: There can be valid cases where a destination doesn't expire the access token at all in which case expiration date will not be available. Since this is a valid case, I've removed. that check. Let me know if it doesn't make sense

router/router.go
sanpj2292: It would be a stringified json

router/router.go
sanpj2292: Sure

router/router.go
sanpj2292: 👍🏼

router/router.go
cisse21: We can get on a call to discuss this once on Monday

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: Yes We don't need to pass error. Will change it

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: Will change this

cisse21 (approved):

router/network.go
```diff
@@ -161,10 +161,12 @@ func (network *NetHandleT) SendPost(ctx context.Context, structData integrations
 			contentTypeHeader = http.DetectContentType(respBody)
 		}
 
-		//If content type is not of type "*text*", overriding it with empty string
+		// If content type is not of type "*text*", overriding it with empty string
 		if !(strings.Contains(strings.ToLower(contentTypeHeader), "text") ||
 			strings.Contains(strings.ToLower(contentTypeHeader), "application/json") ||
 			strings.Contains(strings.ToLower(contentTypeHeader), "application/xml")) {
+			// REVERT: This is only for debugging purposes
+			fmt.Printf(`Original Response: %v`, respBody)
```
cisse21, line 169: Remove this print statement

sanpj2292: Sure

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,546 @@
...
+	}
+
+	authErrHandlerTimeStart := time.Now()
+	res, doErr := authErrHandler.client.Do(req)
+	stats.NewTaggedStat("cp_request_latency", stats.TimerType, stats.Tags{
+		"url":         cpReq.Url,
+		"destination": cpReq.destName,
+	}).SendTiming(time.Since(authErrHandlerTimeStart))
+	authErrHandler.logger.Debugf("[%s request] :: destination request sent\n", loggerNm)
+	if doErr != nil {
```
cisse21, line 513: body.Close here?

sanpj2292: Done

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,546 @@
...
+	authErrHandler.logger = pkgLogger
+	authErrHandler.tr = &http.Transport{}
+	//This timeout is kind of modifiable & it seemed like 10 mins for this is too much!
+	authErrHandler.client = &http.Client{}
+	authErrHandler.destLockMap = destLockMap
+	authErrHandler.accountLockMap = accountLockMap
+	authErrHandler.lockMapWMutex = lockMapMutex
+	authErrHandler.destAuthInfoMap = destAuthInfoMap
+	authErrHandler.refreshActiveMap = refreshActiveMap
+	authErrHandler.disableDestActiveMap = disableDestActiveMap
```
lvrach, line 156: What is the reason for initializing those maps and mutexes as global scope variables and then passing them in the handler?

Can just define them inside here and avoid globals?

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,546 @@
...
+	}
+
+	refTokenBody := RefreshTokenBodyParams{}
+	if router_utils.IsNotEmptyString(refTokenParams.AccessToken) {
+		refTokenBody = RefreshTokenBodyParams{
+			HasExpired:   true,
+			ExpiredToken: refTokenParams.AccessToken,
+		}
+	}
+	authErrHandler.accountLockMap[refTokenParams.AccountId].RLock()
```
lvrach, line 208: This map read is not thread-safe. You need to obtain a read lock:
```
	resHandler.lockMapWMutex.Lock()
```

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,546 @@
...
+	refVal, ok := authErrHandler.destAuthInfoMap[refTokenParams.AccountId]
+	if ok {
+		isInvalidAccessTokenForRefresh := (router_utils.IsNotEmptyString(refVal.Account.AccessToken) &&
+			refVal.Account.AccessToken != refTokenParams.AccessToken)
+		if isInvalidAccessTokenForRefresh {
+			authErrHandler.accountLockMap[refTokenParams.AccountId].RUnlock()
+			// authStats.statName = fmt.Sprintf("%s_success", refTokenParams.EventNamePrefix)
+			// authStats.errorMessage = ""
+			// authStats.SendCountStat()
```
lvrach, line 217: Should this code remain commented out ?

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,546 @@
...
+	if doErr != nil {
+		// Abort on receiving an error
+		authErrHandler.logger.Errorf("[%s request] :: destination request failed: %+v\n", loggerNm, doErr)
+		return http.StatusBadRequest, doErr.Error()
+	}
+	statusCode, resp := processResponse(res)
+	return statusCode, resp
+}
+
+func (resHandler *OAuthErrResHandler) NewMutex(id string, errCategory string) error {
```
lvrach, line 525: A few concerns regarding this method:
1. It seems to me that this method should not be exported, thus 
```suggestion
func (resHandler *OAuthErrResHandler) newMutex(id string, errCategory string) error {
``` (lower case)
2. Is this method is control-coupled by the `errCategory` function

A suggestion here would be to do something like this:
```suggestion
 setMapMutex(map[string]*sync.RWMutex, id string)
 ```
 
 This way you can avoid:
 ```
 switch errCategory {
	case DISABLE_DEST:
		mutexMap = resHandler.destLockMap
	case REFRESH_TOKEN:
		mutexMap = resHandler.accountLockMap
	default:
		// when errorCategory i empty
		resHandler.logger.Debugf("[%s request] :: Case missing for mutex for %s\n", loggerNm, id)
		return fmt.Errorf(`except %v, %v error category is not supported`, DISABLE_DEST, REFRESH_TOKEN)
	}
```

Also makes the function, generic enough to be used in other places as well.

3. If you care about performance, consider using a Read lock for checking if the id already exists and then switch to a write `Lock` .

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,546 @@
...
+		}
+	}
+	//Detecting content type of the respData
+	contentTypeHeader := http.DetectContentType(respData)
+	//If content type is not of type "*text*", overriding it with empty string
+	if !(strings.Contains(strings.ToLower(contentTypeHeader), "text") ||
+		strings.Contains(strings.ToLower(contentTypeHeader), "application/json") ||
+		strings.Contains(strings.ToLower(contentTypeHeader), "application/xml")) {
+		respData = []byte("")
+	}
```
lvrach, line 478: [minor]
```suggestion
	contentTypeHeader := strings.ToLower(http.DetectContentType(respData))
	//If content type is not of type "*text*", overriding it with empty string
	if !(strings.Contains(contentTypeHeader, "text") ||
		strings.Contains(contentTypeHeader, "application/json") ||
		strings.Contains(contentTypeHeader, "application/xml")) {
		respData = []byte("")
	}
```

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -0,0 +1,546 @@
...
+		statName:        "",
+		isCallToCpApi:   false,
+		authErrCategory: DISABLE_DEST,
+		errorMessage:    "",
+		destDefName:     destination.DestinationDefinition.Name,
+	}
+	defer func() {
+		disableDestStats.statName = "disable_destination_total_req_latency"
+		disableDestStats.isCallToCpApi = false
+		disableDestStats.SendTimerStats(authErrHandlerTimeStart)
```
lvrach, line 400: [minor] This part of code here concerns me a bit, in terms of readability and accidents. Having to change the properties of a struct just before you call a method of it, can cause confusion and issues. 

Is there a way to avoid doing this? For example, what if:
```suggestion
		disableDestStats.SendTimerStats("disable_destination_total_req_latency", authErrHandlerTimeStart, false)
```

Or even better create one method for `isCallToCpApi=true` and one for `isCallToCpApi=false`

lvrach (commented):
Added a few comments, feel free to ignore minor. I will keep reviewing

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: I think I referred this from somewhere but I don't remember now. I will change it.

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: Makes sense. Done!

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: Done

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: If I am understanding correctly, Is this what you're trying to say ?
```golang
authErrHandler.lockMapWMutex.RLock()
authErrHandler.accountLockMap[refTokenParams.AccountId].RLock()
// After all the reads are complete
authErrHandler.lockMapWMutex.RUnlock()
```

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292:
```golang
setMapMutex(map[string]*sync.RWMutex, id string)
```
Sorry couldn't completely understand your perspective about this. Can you explain it a bit more ?

lvrach (approved):

lvrach (approved):
Looks good from security perspective

router/router.go
```diff
@@ -604,12 +638,33 @@ func (worker *workerT) handleWorkerDestinationJobs(ctx context.Context) {
...
+										workerId:       worker.workerID,
+										destRespStCd:   destRespStCd,
+										destResBody:    destRespBody,
+										trRespStCd:     respStatusCode,
+										trRespBody:     respBodyTemp,
+									})
+								}
 							}
 							// response transform end
 							if isSuccessStatus(respStatusCode) {
```
utsabc, line 670: We have the option of not saving the destination response in case of success like this, did we remove that?
`	if isSuccessStatus(respStatusCode) && !worker.rt.saveDestinationResponseOverride {
					if saveDestinationResponse {
						if !getRouterConfigBool("saveDestinationResponse", worker.rt.destName, true) {
							respBody = ""
						}
					} else {
						respBody = ""
					}
				}`

sanpj2292: I am not sure about removing, `router.go` I've used the source code implemented by you in major parts like this.

sanpj2292: Also if you see Line numbers 694 - 702 you'll find what you were asking for. Can you confirm ?

utsabc: yeah found it was looking at wrong place no issues

sanpj2292:
Please check `Load Test Results` in this document

**Load Test Results**:
https://www.notion.so/rudderstacks/BQ-Stream-Test-Report-0a921410b87a4afda565a98b089f2fa3

PR #1252, opened 2021-09-20
Bump to go version: 1.16

lvrach:
# What

Change version 1.13 to 1.16, by search and replace. `go mod tidy`  and `go mod vendor` made some changes, committing those as well.

# Why 
Reasons to upgrade go version:
* new features, like embedding 
* tooling improvements
* new API methods, stay up to date with the latest recommend ones and avoid deprecated ones, like `ioutil`
* perfromance improvements 
* keep up to date by incremental updates,  to  avoid big version bumps that could be expensive.

Going from 1.13 to 1.16 we have the following changes:
* https://golang.org/doc/go1.14
* https://golang.org/doc/go1.15
* https://golang.org/doc/go1.16

# Risks
I am running the rudder-server code base locally using 1.16 for more than a month. I haven't experienced a problem with code and tooling changes.

Also I have checked the change log for changes that could potential affects us.

The following are not breaking, but would be good to be aware of:

### Embedded Files
_One of the features that is a reason by it self to upgrade_

The new embed package provides access to files embedded in the program during compilation using the new //go:embed directive.

### File Systems
The new io/fs package defines the fs.FS interface, an abstraction for read-only trees of files. The standard library packages have been adapted to make use of the interface as appropriate.
On the producer side of the interface, the new embed.FS type implements fs.FS, as does zip.Reader. The new os.DirFS function provides an implementation of fs.FS backed by a tree of operating system files.
On the consumer side, the new http.FS function converts an fs.FS to an http.FileSystem. Also, the html/template and text/template packages’ ParseFS functions and methods read templates from an fs.FS.
For testing code that implements fs.FS, the new testing/fstest package provides a TestFS function that checks for and reports common mistakes. It also provides a simple in-memory file system implementation, MapFS, which can be useful for testing code that accepts fs.FS implementations.

### Deprecation of io/ioutil

The io/ioutil package has turned out to be a poorly defined and hard to understand collection of things. All functionality provided by the package has been moved to other packages. The io/ioutil package remains and will continue to work as before, but we encourage new code to use the new definitions in the io and os packages. Here is a list of the new locations of the names exported by io/ioutil:
* Discard => io.Discard
* NopCloser => io.NopCloser
* ReadAll => io.ReadAll
* ReadDir => os.ReadDir (note: returns a slice of os.DirEntry rather than a slice of fs.FileInfo)
* ReadFile => os.ReadFile
* TempDir => os.MkdirTemp
* TempFile => os.CreateTemp
* WriteFile => os.WriteFile

### Vendoring (1.14)
When the main module contains a top-level vendor directory and its go.mod file specifies go 1.14 or higher, the go command now defaults to -mod=vendor for operations that accept that flag. A new value for that flag, -mod=mod, causes the go command to instead load modules from the module cache (as when no vendor directory is present).
When -mod=vendor is set (explicitly or by default), the go command now verifies that the main module's vendor/modules.txt file is consistent with its go.mod file.
go list -m no longer silently omits transitive dependencies that do not provide packages in the vendor directory. It now fails explicitly if -mod=vendor is set and information is requested for a module not mentioned in vendor/modules.txt.

### Modules (1.16)
Module-aware mode is enabled by default, regardless of whether a go.mod file is present in the current working directory or a parent directory. More precisely, the GO111MODULE environment variable now defaults to on. To switch to the previous behavior, set GO111MODULE to auto.
Build commands like go build and go test no longer modify go.mod and go.sum by default. Instead, they report an error if a module requirement or checksum needs to be added or updated (as if the -mod=readonly flag were used). Module requirements and sums may be adjusted with go mod tidy or go get.


chandumlg (approved):

cisse21 (approved):

vendor/modules.txt
```diff
@@ -37,25 +47,33 @@ github.com/ClickHouse/clickhouse-go/lib/protocol
...
 # github.com/DATA-DOG/go-sqlmock v1.5.0
+## explicit
 github.com/DATA-DOG/go-sqlmock
 # github.com/DataDog/zstd v1.4.1
 github.com/DataDog/zstd
 # github.com/EagleChen/mapmutex v0.0.0-20180418073615-e1a5ae258d8d
+## explicit
 github.com/EagleChen/mapmutex
 # github.com/EagleChen/restrictor v0.0.0-20180420073700-9b81bbf8df1d
+## explicit
```
Sidddddarth, line 58: @lvrach What are these `explicit` lines about?

Sidddddarth (approved):

vendor/modules.txt
lvrach: > As of Go 1.14, the go command will also verify that your project’s vendor/modules.txt file is consistent with its go.mod file. If not, you may get an error, such as the one shown below, when building your project:

So for every entry in the the go.mod you have a `## explicit` comment annotation.

PR #1267, opened 2021-09-23
Cleanup Empty Folder structures while removing files.

achettyiitr:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/utils/misc/misc.go
+++ b/utils/misc/misc.go
@@ -284,12 +284,12 @@ func RemoveFilePaths(filePaths ...string) {
 		if err != nil {
 			pkgLogger.Error(err)
 		}
+		RemoveEmptyFolderStructureForFilePath(fp)
 	}
-	removeEmptyFolderStructure(filePaths[0])
 }
 
-// removeEmptyFolderStructure recursively cleans up everything till it reaches the stage where the folders are not empty or parent.
-func removeEmptyFolderStructure(fp string) {
+// RemoveEmptyFolderStructureForFilePath recursively cleans up everything till it reaches the stage where the folders are not empty or parent.
+func RemoveEmptyFolderStructureForFilePath(fp string) {
 	if fp == "" {
 		return
 	}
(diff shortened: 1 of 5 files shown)


utils/misc/misc.go
```diff
@@ -276,13 +277,30 @@ func UnZipSingleFile(outputfile string, filename string) {
 	rc.Close()
 }
 
-func RemoveFilePaths(filepaths ...string) {
-	for _, filepath := range filepaths {
+// RemoveFilePaths removes filePaths as well as cleans up the empty folder structure.
+func RemoveFilePaths(filePaths ...string) {
```
sivashanmukh, line 281: Why do we need filepaths to be a `...string`. It looks like all the callers of this send only one path.

utils/misc/misc.go
```diff
@@ -276,13 +277,30 @@ func UnZipSingleFile(outputfile string, filename string) {
...
 		err := os.Remove(filepath)
 		if err != nil {
 			pkgLogger.Error(err)
 		}
 	}
+	removeEmptyFolderStructure(filePaths...)
+}
+
+// removeEmptyFolderStructure recursively cleans up everything till it reaches the stage where the folders are not empty or parent.
+func removeEmptyFolderStructure(filepaths ...string) {
```
sivashanmukh, line 292: Why do we need filepaths to be `...string`. It looks like we are only using `filepaths[0]` anyways.

sivashanmukh (approved):
LGTM

sivashanmukh (changes requested):
Based on previous comments

utils/misc/misc.go
achettyiitr: Removed ...

utils/misc/misc.go
achettyiitr: RemoveFilePaths is an older implementation. I just updated the comment at the top and added cleanup for folders here.
Since RemoveFilePaths is being used at other places as well like batchrouter.go, azure-synapse, clickhouse.go, identify.go where we are using ...

psrikanth88 (approved):

PR #1288, opened 2021-09-29
Add logs and metrics to investigate body read issues

lvrach:
We need more observability around request body reading, to investigate customer issue.

1. `gateway.requestSizeStat.SendTiming(time.Duration(len(body)) * time.Millisecond)` -> hack to get the proper body size, we currently have 0 due to flooring
2.  Measure the time it takes to read a body, so we can compare it with total request time.
3. Log Content-Length and partial body, if we get an error for ReadAll

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -149,6 +149,7 @@ type HandleT struct {
 	requestSizeStat                                            stats.RudderStats
 	dbWritesStat                                               stats.RudderStats
 	dbWorkersBufferFullStat, dbWorkersTimeOutStat              stats.RudderStats
+	bodyReadTimeStat                                           stats.RudderStats
 	trackSuccessCount                                          int
 	trackFailureCount                                          int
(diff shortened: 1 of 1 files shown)


gateway/gateway.go
```diff
@@ -1434,6 +1448,7 @@ func (gateway *HandleT) Setup(application app.Interface, backendConfig backendco
 	gateway.dbWritesStat = gateway.stats.NewStat("gateway.db_writes", stats.CountType)
 	gateway.dbWorkersBufferFullStat = gateway.stats.NewStat("gateway.db_workers_buffer_full", stats.CountType)
 	gateway.dbWorkersTimeOutStat = gateway.stats.NewStat("gateway.db_workers_time_out", stats.CountType)
+	gateway.bodyReadTimeStat = gateway.stats.NewStat("gateway.batch_size", stats.TimerType)
```
SumanthPuram, line 1451: Should we fix the stat name? `bodyReadTime vs batchSize`

PR #1332, opened 2021-10-12
Track Event delivery time for fresh uploads

Jayachand:
**Fixes** # (*issue*)
- Send event delivery time stat for only fresh uploads
- Retried uploads will have parameter retried set in upload metadata

## Type of change
## Checklist:


warehouse/stats.go
```diff
@@ -161,7 +162,22 @@ func (job *UploadJobT) recordTableLoad(tableName string, numEvents int64) {
...
+
+	config := job.warehouse.Destination.Config
+	syncFrequency := "1440"
+	if config[warehouseutils.SyncFrequency] != nil {
+		syncFrequency, _ = config[warehouseutils.SyncFrequency].(string)
+	}
+
+	var syncValue int
+	if val, err := strconv.Atoi(syncFrequency); err == nil {
+		syncValue = 2 * val
```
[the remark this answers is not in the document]
Jayachand: added

warehouse/upload.go
```diff
@@ -1292,6 +1292,7 @@ func (job *UploadJobT) triggerUploadNow() (err error) {
 		metadata = make(map[string]string)
 	}
 	metadata["nextRetryTime"] = time.Now().Add(-time.Hour * 1).Format(time.RFC3339)
+	metadata["retried"] = "true"
```
dhawal1248, line 1295: set and read actual boolean true and false here??

warehouse/stats.go
```diff
@@ -162,7 +164,17 @@ func (job *UploadJobT) recordTableLoad(tableName string, numEvents int64) {
 		pkgLogger.Errorf("[WH]: Failed to generate delay metrics: %s, Err: %v", job.warehouse.Identifier, err)
 		return
 	}
-	job.timerStat("event_delivery_time", tag{name: "tableName", value: strings.ToLower(tableName)}).SendTiming(time.Since(firstEventAt))
+
+	config := job.warehouse.Destination.Config
```
achettyiitr, line 168: Should we move this inside the if block of retried?

warehouse/upload.go
```diff
@@ -1292,6 +1292,7 @@ func (job *UploadJobT) triggerUploadNow() (err error) {
 		metadata = make(map[string]string)
 	}
 	metadata["nextRetryTime"] = time.Now().Add(-time.Hour * 1).Format(time.RFC3339)
```
dhawal1248, line 1294: also set a custom priority - something like 50 so that triggered uploads are picked up faster
We can even take this as an input if required

warehouse/stats.go
achettyiitr: Then in that case everyone don't need to do the syncFrequency retrieval from config.

dhawal1248 (commented):
Added minor comments

warehouse/upload.go
achettyiitr: We are already handling this when are are doing initUpload.

warehouse/stats.go
Jayachand: Moved inside. compared to actual uploads, retried will be minor

sivashanmukh (approved):
LGTM

warehouse/upload.go
```diff
@@ -1287,12 +1287,14 @@ func (job *UploadJobT) triggerUploadNow() (err error) {
...
-	var metadata map[string]string
+	var metadata map[string]interface{}
 	unmarshallErr := json.Unmarshal(upload.Metadata, &metadata)
 	if unmarshallErr != nil {
-		metadata = make(map[string]string)
+		metadata = make(map[string]interface{})
 	}
 	metadata["nextRetryTime"] = time.Now().Add(-time.Hour * 1).Format(time.RFC3339)
+	metadata["retried"] = true
+	metadata["priority"] = 50
```
dhawal1248, line 1297: also set nextRetryTime so that we can be sure that it will be picked up asap ?

dhawal1248 (commented):
Added a minor comment

dhawal1248 (approved):

PR #1338, opened 2021-10-14
Enable concurrency on processor on pipeline processing 

lvrach:
## Change summary

First step was to refactor the code to put the pipeline ( = source and destination pair) processing, you can see the change in this PR https://github.com/rudderlabs/rudder-server/pull/1338/commits/0f4d607ecb1d76a7d98a13cccb4c6425e0c3b89c. To do this I used goland refactoring feature, with careful adjustments afterwards.

Having processing in a separate function made it easy to run it concurrently, by spinning one go routine per pipeline and waiting for them to be done. 
```go
	chOut := make(chan processPipelineOutput, 1)
	wg := sync.WaitGroup{}
	wg.Add(len(groupedEvents))
	for srcAndDestKey, eventList := range groupedEvents {
		srcAndDestKey, eventList := srcAndDestKey, eventList
		go func() {
			defer wg.Done()
			chOut <- proc.processPipeline( srcAndDestKey, eventList,. ..)
		}()
	}
	go func() {
		wg.Wait()
		close(chOut)
	}()
```

While the parent go-routine aggregates the results from individual processing:

```go
	for o := range chOut {
		batchDestJobs = append(batchDestJobs, o.batchDestJobs...)
		destJobs = append(destJobs, o.destJobs...)
		reportMetrics = append(reportMetrics, o.reportMetrics...)
		for k, v := range o.errorsPerDestID {
			procErrorJobsByDestID[k] = append(procErrorJobsByDestID[k], v...)
		}
	}
```

To my surprise after this change, no performance improvement has noticeable between master and this branch. 

1 source @ 300 rps -> 4 destinations. 

_concurrent-proc (left) vs master (right)_ 

Screenshot 2021-10-15 at 10 34 27 AM

### Transport refactor 

Transport code was not concurrent safe, it was meant to be called only sequentially and it had an access lock mutex to enforce it.

One option was to go with multiple transporter instances per pipeline, which I did in order to quickly test the perfromance. However, the way transporter worked, with a fixed number of worker per instance, having arbitrary instances running depending on the number of pipelines could lead to unpredictable perfromance.

I would like to have a single transporter, with all the request sharing the same http connection pool. If there is reason to limit the number of concurrent request to transformers it should happen in a single place and apply the same independent of the customers configuration.

Moreover, the worker pattern implementation was adding complexity and it was even harder to make concurrent safe. Thus I use a simple fan-out/fan-in pattern.

### Stats .Start/.End -> .SendTime

The affected code contained a lot of non concurrent-safe usage of stats. I've modified them to use .SendTime.

## Impact

After those changes the impact is clear, the `concurrent-proc` is consistently better from `master` for setups with more than one pipeline. While I haven't seen any perfromance degradation yet.

You can find benchmark results down in the comments

## TODOs 

* Add tests for the transformer
* Determine if a harder limit for concurrent requests to transformers needs to be imposed.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -163,8 +163,9 @@ Processor:
   avgEventsInRequest: 1 # DEPRECATED: don't use avgEventsInRequest, rudder-server will automatically adapt
   transformBatchSize: 100
   userTransformBatchSize: 200
-  maxChanSize: 2048
-  numTransformWorker: 8
+  maxConcurrency: 200
+  maxHTTPConnections: 100
+  maxHTTPIdleConnections: 50
   maxRetry: 30
   retrySleep: 100ms
   errReadLoopSleep: 30s


lvrach:
1 source @ 200 rps -> 6 destinations, 5 minutes run

master:
Screenshot 2021-10-19 at 5 07 33 PM

concurrent-proc:
Screenshot 2021-10-19 at 5 09 10 PM

PR #1341, opened 2021-10-16
Response transform

utsabc:
## Response Transformer Support

Here the main objective is to handle incompatible destination responses for rudder-server using a transformer. 
The idea is once we deliver the event to the destination we want to parse the response returned by the destination by making a response-transformation request to the transformer. 

The transformer takes the responsibility of sending appropriate responses and code for the server (namely 200, 429, 500, and 400) so that we can abort, and retry destination jobs and avoid silent failures.

Test Report for Response Transform
https://docs.google.com/document/d/1M19vgBC0XVBln8Wu-suQx9mzjkoDf94tVk0jr4Znbe8/edit?usp=sharing

Load test Report 
without Response Transform (GA, Braze and AM)
https://snapshot.raintank.io/dashboard/snapshot/WY9tKHAqIvXCiQYIC5QfVlwp44s3zAxq

with Response Transform Enabled Destination (GA and Braze)
https://snapshot.raintank.io/dashboard/snapshot/3gXdgSuZF3876y3zPfl6Vfh6NO3NkzsX

with Response Transform Enabled Destination (GA, Braze, and AM with lower load)
https://snapshot.raintank.io/dashboard/snapshot/EDy31wzPV1CAfwLlRyLX7kBD07PrJIjX?orgId=2


router/transformer/transformer.go
```diff
@@ -150,11 +156,52 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
+		//Detecting content type of the respBody
+		contentTypeHeader = http.DetectContentType(respData)
+	}
+	//If content type is not of type "*text*", overriding it with empty string
+	if !(strings.Contains(strings.ToLower(contentTypeHeader), "text") ||
+		strings.Contains(strings.ToLower(contentTypeHeader), "application/json") ||
+		strings.Contains(strings.ToLower(contentTypeHeader), "application/xml")) {
+		respData = []byte("")
+	}
+	if err != nil {
```
sanpj2292, line 188: Move this block to right after `Line: 168`

router/transformer/transformer.go
```diff
@@ -150,11 +156,52 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
+	resp.Body.Close()
+	return resp.StatusCode, string(respData)
+}
+
+//is it ok to use same client for network and transformer calls? need to understand timeout setup in router
 func (trans *HandleT) Setup() {
 	trans.logger = pkgLogger
 	trans.tr = &http.Transport{}
-	trans.client = &http.Client{Transport: trans.tr}
+	trans.client = &http.Client{Transport: trans.tr, Timeout: 10 * time.Minute}
```
sanpj2292, line 201: I don't think we would need a specific timeout here. Let's remove it

utsabc: This is the timeout for requests to transformers. When we were doing proxying we kept it 10min for safety as we there could be long requests to destinations. Now we can remove this/ reduce it to much smaller durationion

router/transformer/transformer.go
utsabc: Yep

router/transformer/transformer.go
```diff
@@ -150,11 +156,52 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
+		respData, _ = io.ReadAll(resp.Body)
+	}
+	integrations.CollectIntgErrorStats([]byte(gjson.GetBytes(respData, "output").Raw), false)
+	var contentTypeHeader string
+	if resp != nil && resp.Header != nil {
+		contentTypeHeader = resp.Header.Get("Content-Type")
+	}
+	if contentTypeHeader == "" {
+		//Detecting content type of the respBody
+		contentTypeHeader = http.DetectContentType(respData)
```
cisse21, line 185: why do we need lines 180-184 when we can get the content type using the detectcontenttype function

router/transformer/transformer.go
cisse21: We should ideally restrict any http timeouts to not more than 30s. Probably can have it configured for destinations which by default take longer

Can we add it as a configurable variable if it makes sense

router/router.go
```diff
@@ -616,7 +633,7 @@ func (worker *workerT) handleWorkerDestinationJobs(ctx context.Context) {
 				deliveryLatencyStat.End()
 				// END: request to destination endpoint
 
-				if isSuccessStatus(respStatusCode) {
+				if isSuccessStatus(respStatusCode) && !worker.rt.saveDestinationResponseOverride {
```
cisse21, line 636: Do we still need the variable saveDestinationResponse ? If not we can probably remove it and use saveDestinationResponseOverride

router/transformer/transformer.go
```diff
@@ -150,11 +156,52 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
 }
 
+func (trans *HandleT) ResponseTransform(responseData integrations.DeliveryResponseT, destName string) (statusCode int, respBody string) {
+	rawJSON, err := json.Marshal(responseData)
+	if err != nil {
+		panic(err)
+	}
+	var resp *http.Response
+	var respData []byte
+	url := getResponseTransformURL(destName)
```
cisse21, line 166: What if the URL is not implemented on transformer. To ensure we have backwards compatibility if the HTTP request returns 404 , can we use the original statusCode and body?

router/router.go
```diff
@@ -1778,10 +1797,14 @@ func (rt *HandleT) Setup(backendConfig backendconfig.BackendConfig, jobsDB, erro
...
 	retryTimeWindowKeys := []string{"Router." + rt.destName + "." + "retryTimeWindow", "Router." + rt.destName + "." + "retryTimeWindowInMins", "Router." + "retryTimeWindow", "Router." + "retryTimeWindowInMins"}
 	savePayloadOnErrorKeys := []string{"Router." + rt.destName + "." + "savePayloadOnError", "Router." + "savePayloadOnError"}
+	responseTransformKeys := []string{"Router." + rt.destName + "." + "responseTransform", "Router." + "responseTransform"}
+	saveDestinationResponseOverrideKeys := []string{"Router." + rt.destName + "." + "saveDestinationResponseOverride", "Router." + "saveDestinationResponseOverride"}
 	config.RegisterIntConfigVariable(3, &rt.maxFailedCountForJob, true, 1, maxFailedCountKeys...)
 	config.RegisterDurationConfigVariable(180, &rt.retryTimeWindow, true, time.Minute, retryTimeWindowKeys...)
 	config.RegisterBoolConfigVariable(false, &rt.enableBatching, false, "Router."+rt.destName+"."+"enableBatching")
 	config.RegisterBoolConfigVariable(false, &rt.savePayloadOnError, true, savePayloadOnErrorKeys...)
+	config.RegisterBoolConfigVariable(false, &rt.responseTransform, true, responseTransformKeys...)
```
cisse21, line 1806: Can you change the third argument to false for both these variables. These are not hot reloadable and would require a restart to get applied

cisse21 (changes requested):

router/transformer/transformer.go
cisse21: Also in cases of 5xx from transformer we should be using the original sendPost response code to ensure we don't deliver the same data twice to the destination

router/transformer/transformer.go
utsabc: As discussed with @sanpj2292  we will for now use default timeout of go-HTTP-client and and not configure explicitly if required we can do it as part of enhancement

router/router.go
utsabc: Let's keep both, for now, saveDestinationResponse is a config from the control plane. We wanted to have some control from the server hence kept it

router/router.go
utsabc: You mean `responseTransform` and `savePayloadOnError` ?

cisse21: Yes

router/transformer/transformer.go
utsabc: Yes

router/transformer/transformer.go
cisse21: Go HTTP client does not have an explicit timeout I guess and requests can get stuck for a while

cisse21 (changes requested):

router/transformer/transformer.go
utsabc: Yeah we can remove that logic

cisse21 (approved):

router/transformer/transformer.go
```diff
@@ -159,35 +159,94 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
+		}
+
+		if requestFailed {
+			trans.logger.Errorf("Failed request succeeded after %v retries, URL: %v", retryCount, url)
+		}
+
+		if resp != nil {
+			//handling for 404
+			if resp.StatusCode == 404 {
+				trans.logger.Errorf("[Response transform doesnot exist for URL: %v]", url)
```
utsabc, line 209: TODO: Collect detailed metric in this scenario and get ourselves alerted

router/transformer/transformer.go
```diff
@@ -159,35 +159,94 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
+				panic(fmt.Errorf("Transformer HTTP connection error: URL: %v Error: %+v", url, err))
+			}
+			retryCount++
+			time.Sleep(retrySleep)
+			//Refresh the connection
+			continue
+		}
+
+		if resp != nil && resp.Body != nil {
+			tempRespData, _ = io.ReadAll(resp.Body)
```
cisse21, line 190: panic for this error

cisse21 (changes requested):

router/transformer/transformer.go
```diff
@@ -150,11 +157,106 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
+	}
+	var resp *http.Response
+	var respData []byte
+	var respCode int
+	var tempRespData []byte
+	url := getResponseTransformURL(destName)
+	for {
+		trans.transformerResponseTransformRequestTime.Start()
+		resp, err = trans.client.Post(url, "application/json; charset=utf-8", bytes.NewBuffer(rawJSON))
+		trans.transformerResponseTransformRequestTime.End()
```
lvrach, line 175: Using timer start/end is not thread safe. We will get incorrect reading if it runs concurrently.

```suggestion
		s := time.Now()
		resp, err = trans.client.Post(url, "application/json; charset=utf-8", bytes.NewBuffer(rawJSON))
		
		trans.transformerResponseTransformRequestTime.SendTiming(time.Since(s))
```

cisse21 (approved):

router/transformer/transformer.go
utsabc: This is addressed

router/transformer/transformer.go
utsabc: This is addressed

processor/integrations/integrations.go
```diff
@@ -57,6 +58,51 @@ type PostParametersT struct {
...
+	Metadata                 map[string]interface{} `json:"metadata"`
+	ResponseTransformFailure bool                   `json:"responseTransformFailure"`
+	FailureAt                string                 `json:"failureAt"`
+}
+
+type TransErrorT struct {
+	ErrorDetailed TransErrorSpecT `json:"errorDetailed"`
+}
+
+func CollectIntgErrorStats(input []byte, transformationError bool) {
```
lvrach, line 82: Deepsource reported this as a control-coupled function: https://deepsource.io/gh/rudderlabs/rudder-server/run/9bce0483-bfc7-4f18-b43a-458da7d9d9cb/go/RVV-A0005

I agree with it, having:
```
func CollectIntgTransformErrorStats(input []byte) {
```

and a 
```
func CollectIntgDestErrorStats(input []byte) {
```

Would be more readable.

router/transformer/transformer.go
```diff
@@ -161,11 +168,103 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
+		panic(err)
+	}
+	var resp *http.Response
+	var respData []byte
+	var respCode int
+	var tempRespData []byte
+	url := getResponseTransformURL(destName)
+	for {
+		s := time.Now()
+		resp, err = trans.client.Post(url, "application/json; charset=utf-8", bytes.NewBuffer(rawJSON))
```
lvrach, line 185: Ideally we should use the ctx in our post request https://pkg.go.dev/net/http#NewRequestWithContext

processor/integrations/integrations.go
utsabc: Got it, addressing this

router/transformer/transformer.go
```diff
@@ -161,11 +168,103 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
+	}
+	return respCode, string(respData)
+}
+
+//is it ok to use same client for network and transformer calls? need to understand timeout setup in router
 func (trans *HandleT) Setup() {
 	trans.logger = pkgLogger
 	trans.tr = &http.Transport{}
-	trans.client = &http.Client{Transport: trans.tr}
+	trans.client = &http.Client{Transport: trans.tr, Timeout: 30 * time.Second}
```
lvrach, line 264: Should this be configurable?

utsabc: This is addressed

router/transformer/transformer.go
utsabc: This is addressed

lvrach (approved):
Looks good from security perspective

PR #1395, opened 2021-11-01
Additional metrics

lvrach:
While performing some stress-test I found the need for some new metrics:

## Introducing histogram type

For more accurately observing the values of a metric with avg, 95, 99, min, max. This is similar to timer. The counter type is problematic in some use cases because it sums all the values captured during a sampling period.

If for example, you want to find how many events/bytes are picked in every loop (multiple loops happen inside a sampling period). The counter can only give you a total summary. While with histogram they are many options.

## Introducing .Since

To avoid repetition, I have added `.Since()` for timer stats.   

### runtime 
Adding runtime metric `mem.gc.cpu_percent`: 

```
	// GCCPUFraction is the fraction of this program's available
	// CPU time used by the GC since the program started.
	//
	// GCCPUFraction is expressed as a number between 0 and 1,
	// where 0 means GC has consumed none of this program's CPU. A
	// program's available CPU time is defined as the integral of
	// GOMAXPROCS since the program started. That is, if
	// GOMAXPROCS is 2 and a program has been running for 10
	// seconds, its "available CPU" is 20 seconds. GCCPUFraction
	// does not include CPU time used for write barrier activity.
	//
	// This is the same as the fraction of CPU reported by
	// GODEBUG=gctrace=1.
```
https://pkg.go.dev/runtime

### processor 

`processor.db_read_requests` number of requests read from db in every loop. Same us `processor.num_requests`, but with histogram type.

`processor.db_read_events` number of events read from db in every loop. Same us  `processor.num_events`, , but with histogram type.

`processor.db_read_payload_bytes` measuring the total size in bytes of jobs event_payload  in every loop.

`processor.db_write_status_time` time it takes to write job status table

`processor.db_write_jobs_time` time it takes to write to all jobs tables

`processor.db_write_payload_bytes` jobs event_payload total size in bytes, written to router or batch_router dbs,  (by module)
`processor.db_write_events` jobs/events written to router or batch_router dbs (by module)

### transformer

`processor.transformer_request_batch_count` how many batches are created for a single transform call.
`processor.transformer_request_time` pure HTTP request time (doesn't include time waiting for scheduling)

with tags:
```
		"dest_name":    event.Destination.DestinationDefinition.Name,
		"dest_id":   event.Destination.DestinationDefinition.ID,
		"src_id": event.Metadata.SourceID,
```

## Extra 

`avgEventsInRequest` ->  removed `# DEPRECATED`


processor/transformer/transformer.go
```diff
@@ -256,6 +269,19 @@ func (trans *HandleT) Validate(clientEvents []TransformerEventT,
...
 }
 
+func (trans *HandleT) requestTime(s stats.Tags, d time.Duration) {
+	stats.NewTaggedStat("processor.transformer_request_time", stats.TimerType, s).SendTiming(d)
+}
+
+func statsTags(event TransformerEventT) stats.Tags {
+	return stats.Tags{
+		"dest_type": event.Destination.DestinationDefinition.Name,
+		"dest_name": event.Destination.Name,
```
cisse21, line 279: Do we really need a destination name tag when we have both type and ID

lvrach: I am thinking it could make debugging easier.

lvrach: It creates more problems than it solves, I have remove the `Destination.Name`

cisse21: Generally Influx would create more Indexes based on the tags so was just thinking of removing redundancy

cisse21 (approved):

chandumlg (approved):

PR #1404, opened 2021-11-05
Reporting Service TCs

rspriyamdixit:
**Fixes** # (*issue*)

## Type of change
## Checklist:


docker_test.go
```diff
@@ -417,8 +418,71 @@ func run(m *testing.M) (int, error) {
...
+	}
+	defer func() {
+		if err := pool.Purge(timescaleRes); err != nil {
+			log.Printf("Could not purge resource: %s \n", err)
+		}
+	}()
+	timescaleDB_DSN := fmt.Sprintf("postgresql://postgres:password@host.docker.internal:%s/%s?sslmode=disable", timescaleRes.GetPort("5432/tcp"), database)
+	fmt.Println("timesacaleDB_DSN",timescaleDB_DSN)
+	timescaleDB_DSN_1 := fmt.Sprintf("postgresql://postgres:password@localhost:%s/%s?sslmode=disable", timescaleRes.GetPort("5432/tcp"), database)
+	if err := pool.Retry(func() error {
```
lvrach, line 446: Suggestion, for better readability:
```suggestion
	timescaleDB_DSN_Internal := fmt.Sprintf("postgresql://postgres:password@host.docker.internal:%s/%s?sslmode=disable", timescaleRes.GetPort("5432/tcp"), database)
	fmt.Println("timesacaleDB_DSN",timescaleDB_DSN)
	timescaleDB_DSN_viaHost := fmt.Sprintf("postgresql://postgres:password@localhost:%s/%s?sslmode=disable", timescaleRes.GetPort("5432/tcp"), database)
	if err := pool.Retry(func() error {
```

lvrach: Also, consider making `timescaleDB_DSN_viaHost` a local if it is used only for readiness

docker_test.go
```diff
@@ -697,6 +762,16 @@ func TestPostgres(t *testing.T) {
 	require.Equal(t, myEvent.count, "1")
 }
 
+// Verify Event in POSTGRES
+func TestReportingService(t *testing.T) {
```
lvrach, line 766: This test assumes it will run after something else. For data to be there. In this case, it makes sense to move this check in the same test that produces data.

docker_test.go
```diff
@@ -697,6 +762,16 @@ func TestPostgres(t *testing.T) {
...
+	var myEvent Event
+	require.Eventually(t, func() bool {
+		eventSql := "select count (*) from metrics"
+		rs_db.QueryRow(eventSql).Scan(&myEvent.count)
+		return myEvent.count == "11"
```
lvrach, line 771: * The `myEvent.count == "11"` seems overly specific and easy to break. 

* We don't check for the content of those metrics here. Thus, if we simply insert empty or incorrect rows the test won't detect it.

* Is it possible to query the reporting service for the expected data without looking directly into the database. APIs are generally more stable than underlying db schema.

PR #1405, opened 2021-11-05
Introduce pipelining

lvrach:
## Overview 

This perfromance is inspired by [cpu instruction pipelining](https://en.wikipedia.org/wiki/Instruction_pipelining).

The aim of pipelining is to improve throughput for processor by keeping all the components busy, while the processing of events will remain sequential and order won't be affected.

Screenshot 2021-10-26 at 12 43 18 PM

More details in [notion](https://www.notion.so/rudderstacks/Processor-Pipelining-e1e1b29ff77d4318af57bb64c9da1ef8
)

## Implementation

Understanding that processor is a critical and a complex component of our system, I tried to introduce pipelining with as little changes as possible.

### breaking `processJobsForDest`

The first was to break processJobsForDest into 3 components: marshaling, transformation and storage. 
To do this, I used goland refactoring mechanism to extract methods from the code. 
Then I convert all the arguments of this function to a struct. 
Finally, I made sure the intermediate components accept as input a single struct and return a struct that can be used as input in the next component.

### mainPipeline
`mainPipeline(ctx)` is the method that setups and coordinates pipelining. It setups a go routine for each component. Setups channels that act as buffer between components.

### markExecuting

Because we can read again from the database, before the latest retrieved jobs have mark as completed. We need to mark the as executing, before we can read again. Similar to what router is doing.

### `Processor.enablePipelining` flag

This flag can be used to enable or disable pipelining. If pipeline flag is disabled, the `mainLoop(ctx)` will run instead trying to mimic the existing behaviour. This is useful, if some unexpected bug or perfromance degradation is introduced with this feature.

### `processPipeline` -> `processSrcDst`

I have used the word pipeline to refer to a product language, that describes the connection between a source and a destination. However, with the introduction of this feature that name caused confusion.

### Cleanup `AfterJobID`

After the patch that disabled read cursor https://github.com/rudderlabs/rudder-server/pull/1428, a proper cleanup of this feature was done.

## Impact

[25% - 50%] throughput improvement. In theory 100% is expected when db_write = transform time.

### Sustainable throughput 

master: (34k per min):
Screenshot 2021-11-06 at 10 19 39 PM

proc-pipe (51k per min):
Screenshot 2021-11-06 at 10 16 01 PM

### Under maximum pressure

This comparison is done with 1400 RPS as input. With (left part) and without (right part) pipelining.
 
Screenshot 2021-11-11 at 9 23 19 PM

### Under normal pressure

Handling 600 events from single source to four destinations.

Testing pipelining enabled (left) and disabled (right).

When pipelining is disabled the e2e delivery latency gets much higher, even if gw table count is under 3 in both cases

Screenshot 2021-11-16 at 4 17 48 PM


processor/processor.go
```diff
@@ -374,7 +381,12 @@ func (proc *HandleT) Start(ctx context.Context) {
 		if err := proc.backendConfig.WaitForConfig(ctx); err != nil {
 			return err
 		}
-		proc.mainLoop(ctx)
+		if enablePipelining {
+			fmt.Print("pipeline")
```
chandumlg, line 385: Remove the log

processor/processor.go
```diff
@@ -1875,6 +2008,125 @@ func (proc *HandleT) mainLoop(ctx context.Context) {
...
+
+// mainPipeline: new way of handling jobs
+//
+// [getJobs] -chProc-> [processJobsForDest] -chTrans-> [transformations] -chStore-> [Store]
+func (proc *HandleT) mainPipeline(ctx context.Context) {
+	//waiting for reporting client setup
+	if proc.reporting != nil && proc.reportingEnabled {
+		proc.reporting.WaitForSetup(ctx, types.CORE_REPORTING_CLIENT)
+	}
+	fmt.Println("mainPipeline")
```
chandumlg, line 2044: Remove the log

chandumlg (approved):

processor/processor.go
```diff
@@ -420,8 +434,13 @@ var (
 )
 
 func loadConfig() {
+	config.RegisterBoolConfigVariable(true, &enablePipelining, false, "Processor.enablePipelining")
+	config.RegisterIntConfigVariable(1, &pipelineBufferedItems, true, 1, "Processor.pipelineBufferedItems")
```
snarkychef, line 438: I see this is defined as hot reloadable. It doesn't look straight forward hot reloadable in code below. The only way for the new value to reflect is by pausing and resuming the processor? Just for clarification.

snarkychef: If that is indeed the case, we should add a comment?

processor/processor.go
```diff
@@ -420,8 +434,13 @@ var (
...
 	config.RegisterDurationConfigVariable(time.Duration(5000), &maxLoopSleep, true, time.Millisecond, []string{"Processor.maxLoopSleep", "Processor.maxLoopSleepInMS"}...)
+	config.RegisterDurationConfigVariable(time.Duration(200), &readLoopSleep, true, time.Millisecond, "Processor.readLoopSleep")
+	//DEPRECATED: used only on the old mainLoop:
 	config.RegisterDurationConfigVariable(time.Duration(10), &loopSleep, true, time.Millisecond, []string{"Processor.loopSleep", "Processor.loopSleepInMS"}...)
+	//DEPRECATED: used only on the old mainLoop:
```
snarkychef, line 443: We use fixedLoopSleep in our multi-tenant environments, to optimise for Postgres(shared) resource usage. We need to add readLoopSleep to match the existing fixedLoopSleep in that environment. This needs to be taken care of at the time or before deploying this change to production (multi-tenant customers)

processor/processor.go
```diff
@@ -1875,6 +2007,124 @@ func (proc *HandleT) mainLoop(ctx context.Context) {
...
+		}
+	}()
+
+	chTrans := make(chan transformationMessage, bufferSize)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		defer close(chTrans)
+		for jobs := range chProc {
+			chTrans <- proc.processJobsForDest(jobs, nil)
```
snarkychef, line 2101: We are always calling this function with second argument as nil?

processor/processor.go
```diff
@@ -1259,26 +1279,67 @@ func (proc *HandleT) processJobsForDest(jobList []*jobsdb.JobT, parsedEventList
...
+func (proc *HandleT) transformations(in transformationMessage) storeMessage {
 	//Now do the actual transformation. We call it in batches, once
 	//for each destination ID
 
+	var procErrorJobsByDestID = make(map[string][]*jobsdb.JobT)
+	var batchDestJobs []*jobsdb.JobT
+	var destJobs []*jobsdb.JobT
+
 	destProcStart := time.Now()
 	proc.logger.Debug("[Processor: processJobsForDest] calling transformations")
```
snarkychef, line 1327: What is this debug log for? Do we need to update the log to reflect new function name? Do we need the log at all?

processor/processor.go
lvrach: Good catch it should not be hot reloadable

processor/processor.go
lvrach: Yea that correct. My intention was to remove it in the next PR actually.

cisse21 (approved):

cisse21:
Any comments on this PR would be addressed in a follow up PR .

cc : @lvrach  @snarkychef

PR #1417, opened 2021-11-09
Response transform, fix for empty payload on response transform retry

utsabc:
**Fixes** # (*issue*)

Fix for payload value being set to nil upon retry of response transform, as the length of the reader (*strings.reader) type was not set to 0 during retry. Hence used Seek from `i.o.Seeker` interface to set the offset of the payload on retry

## Type of change
- [X] Bug fix (non-breaking change which fixes an issue)
## Checklist:
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -180,10 +180,10 @@ func (trans *HandleT) ResponseTransform(ctx context.Context, responseData integr
 	var respData []byte
 	var respCode int
 	var tempRespData []byte
-	var payload io.Reader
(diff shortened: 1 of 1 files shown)


cisse21 (approved):

sanpj2292 (approved):
Basically the reader after being read, was changing the index in the same object. Hence for subsequent request we couldn't use this reader. We added this fix to reset the read index in subsequent requests

PR #1424, opened 2021-11-11
MT: In Memory Stats

cisse21:
**Fixes** # (*issue*)
Adding support for required in memory-stats at processor and router level.

Stats Required : 
Processor Write Stats per Customer per DestType/DestID 
- This stat aims at maintaining the total number of non terminal jobs present in jobsDB. 

Router In Rate Stats per Customer per DestType/DestID
- This stat tries to maintain the moving average of count of jobs processor writes into Router/BatchRouter Tables

Latency Stats per Customer per DestId/DestType
- This stat maintains the latencies of delivery times for Router

## Type of change
## Checklist:

### DIFF
--- a/utils/misc/misc.go
+++ b/utils/misc/misc.go
@@ -67,6 +67,17 @@ type RudderError struct {
 	Code              int
 }
 
+type Pair struct {
+	Key   string
+	Value int
+}
+
+type PairList []Pair
+
+func (p PairList) Len() int           { return len(p) }
+func (p PairList) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+func (p PairList) Less(i, j int) bool { return p[i].Value < p[j].Value }
+
 var pkgLogger logger.LoggerI
 
 func Init() {
@@ -1272,4 +1283,22 @@ func GetJsonSchemaDTFromGoDT(goType string) string {
 		return "boolean"
 	}
 	return "object"
+}
+
+func sortMap(inputMap map[string]int) map[string]int {
+	p := make(PairList, len(inputMap))
+
+	i := 0
+	for k, v := range inputMap {
+		p[i] = Pair{k, v}
+		i++
+	}
+
+	sort.Sort(p)
+	//p is sorted
+	outMap := make(map[string]int)
+	for _, k := range p {
+		outMap[k.Key] = k.Value
+	}
+	return outMap
 }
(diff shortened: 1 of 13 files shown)


utils/misc/misc.go
```diff
@@ -45,6 +45,10 @@ import (
 
 var AppStartTime int64
 var errorStorePath string
+var jobQueryBatchSize int
+var RouterInMemoryJobCounts map[string]map[string]map[string]int
+var routerJobCountMutex sync.RWMutex
+var ProcessorJobsMovingAverages map[string]map[string]map[string]MovingAverage
```
lvrach, line 51: I would suggest avoid introducing more global variables and instead create a struct will contain the variables and logic necessary for stats to work.

Also, I don't think `utils/misc` is the correct package to host such a rich functionality, instead a new `service/tenantstats` could be used.

utils/misc/misc.go
```diff
@@ -80,6 +91,96 @@ func LoadDestinations() ([]string, []string) {
...
 }
 
+func AddToInMemoryCount(customerID string, destinationType string, count int, tableType string) {
+	_, ok := RouterInMemoryJobCounts[tableType][customerID]
+	if !ok {
+		routerJobCountMutex.Lock()
+		RouterInMemoryJobCounts[tableType][customerID] = make(map[string]int)
+		routerJobCountMutex.Unlock()
+	}
+	RouterInMemoryJobCounts[tableType][customerID][destinationType] += count
```
lvrach, line 101: This code doesn't protects from reads happening while writing to the map. It only prevent concurrent writes, which is not enough.

You need a read lock every-time you try to read from the map. 

Similarly for the methods bellow.
```suggestion
	routerJobCountMutex.RLock()
	_, ok := RouterInMemoryJobCounts[tableType][customerID]
	if !ok {
		routerJobCountMutex.RUnLock()
		routerJobCountMutex.Lock()
		RouterInMemoryJobCounts[tableType][customerID] = make(map[string]int)
		routerJobCountMutex.Unlock()
		routerJobCountMutex.RLock()
	}
	RouterInMemoryJobCounts[tableType][customerID][destinationType] += count
	routerJobCountMutex.RUnlock()
```

lvrach (commented):
Apart from my suggestion of moving the stats collector implementation into a separate service and having a struct instead of global variables.  I think tests are also important and easy to do here. In addition to simple test, [benchmarks](https://pkg.go.dev/testing#hdr-Benchmarks) can be used to test both for perfromance and concurrency issues (with `-race`).

PR #1425, opened 2021-11-11
deletion from S3 with pagination

saurav-malani:
**Fixes** # (*issue*)

## Type of change
## Checklist:


regulation-worker/internal/delete/batch/batch.go
```diff
@@ -0,0 +1,379 @@
...
+		return fmt.Errorf("error while writing to statusTrackerFile: %w", err)
+	}
+	if _, err := f.Write([]byte("\n")); err != nil {
+		return fmt.Errorf("error while writing to statusTrackerFile: %w", err)
+	}
+	if err := f.Close(); err != nil {
+		return fmt.Errorf("error while closing statusTrackerFile: %w", err)
+	}
+	return nil
+}
```
lvrach, line 138: file close is not called in case of a write error
check https://www.joeshaw.org/dont-defer-close-on-writable-files/
```suggestion
func updateStatusTrackerFile(fileName string) (err error) {
	f, err := os.OpenFile(statusTrackerFile, os.O_APPEND|os.O_WRONLY, 0644)
	if err != nil {
		return fmt.Errorf("error while opening statusTrackerFile: %w", err)
	}
        defer func() {
           cerr := f.Close()
            if err == nil {
               err = cerr
            }
        }()
        
	if _, err := io.WriteString(f, fileName + "\n")); err != nil {
		err = fmt.Errorf("error while writing to statusTrackerFile: %w", err)
	        return err
	}
```


```

**Review by lvrach — COMMENTED** [review]
partial review

File: regulation-worker/internal/delete/batch/s3.go [hunk 2]
```diff
@@ -0,0 +1,58 @@
...
+		}
+
+		if users.Phone != nil {
+			searchObject = append(searchObject, "/"...)
+			searchObject = append(searchObject, []byte(*users.Phone)...)
+			searchObject = append(searchObject, "/d;"...)
+		}
+	}
+
+	out, err := exec.Command("sed", "-e", string(searchObject), uncompressedFileName).Output()
```
lvrach, line 50: As we discussed there is a limit on the number of bytes you can provide as a single argument. In most cases is 4k bytes. I believe this is too restrictive for our use case. 

It better to write all the rules into a temporary file and provide this file instead:

you can use a [bufio writer](https://pkg.go.dev/bufio) to efficiently write to the file instead of appending all the rules in memory.

```suggestion
       out, err := exec.Command("sed", "-f", sedCommandsFile uncompressedFileName).Output()
```

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -0,0 +1,379 @@
...
+	err = g.Wait()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (b *Batch) cleanup(prefix string) {
+
+	err := b.FM.DeleteObjects([]string{prefix + "/" + statusTrackerFile})
```
lvrach, line 375: better use [filepath.Join](https://pkg.go.dev/path/filepath#Join) to make it safe and portable.
```suggestion
	err := b.FM.DeleteObjects(filepath.Join(prefix, statusTrackerFile))
```

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -0,0 +1,379 @@
...
+			fileNamePrefix := strings.Split(files[_i].Key, "/")
+			err := batch.delete(gCtx, job.UserAttributes, fileNamePrefix[len(fileNamePrefix)-1])
+			if err != nil {
+				return fmt.Errorf("error while deleting object, %w", err)
+			}
+			err = withExpBackoff(batch.upload, gCtx, files[_i].Key)
+			if err != nil {
+				return fmt.Errorf("error while uploading cleaned file, %w", err)
+			}
+			<-goRoutineCount
```
lvrach, line 362: This could create problems, and potentially block the process under some edge conditions. Because `<-goRoutineCount` doesn't called in case of an error

to be safe:
```
defer func() {
<-goRoutineCount
}
```
in the begging of goroutine.

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -0,0 +1,379 @@
...
+	defer close(goRoutineCount)
+	for i := 0; i < len(files); i++ {
+		_i := i
+		goRoutineCount <- true
+		g.Go(func() error {
+			err = withExpBackoff(batch.download, gCtx, files[_i].Key)
+			if err != nil {
+				return err
+			}
+			fileNamePrefix := strings.Split(files[_i].Key, "/")
```
lvrach, line 353: Safer to use [filepath.Split](https://pkg.go.dev/path/filepath#Split)

Also what `fileNamePrefix` is here and why we need it  ?

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -0,0 +1,379 @@
...
+	//if statusTracker.txt exists then read it & remove all those files name from above gzFilesObjects,
+	//since those files are already cleaned.
+	if exist {
+
+		err = batch.download(ctx, statusTrackerFile)
+		if err != nil {
+			return fmt.Errorf("error while downloading statusTrackerFile:%w", err)
+		}
+
+		data, err := os.ReadFile(statusTrackerFile)
```
lvrach, line 326: How big can this file get, is it ok to read everything in memory ?

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+	if err := f.Close(); err != nil {
+		return fmt.Errorf("error while closing statusTrackerFile: %w", err)
+	}
+	return nil
+}
+
+//downloads `fileName` locally
+//Note: download happens concurrently in 5 go routine by default
+func (b *Batch) download(ctx context.Context, fileName string) error {
```
lvrach, line 142: I believe it would be better if download, was using a specific directory to store all the files, but default a random folder insider tmp.

Moreover use random file names for the downloaded files to avoid naming conflicts.

the `Batch` struct will have an optional string to store the directory and download will return the full path of the random file name.

Is important for those files to get deleted in case of a crash for privacy and regulatory reasons. In most cases tmp directory gets cleared after restart so it is safer there.

lvrach (changes requested):

regulation-worker/cmd/main_test.go
```diff
@@ -42,24 +42,26 @@ func handler() http.Handler {
...
-		main.Run(svcCtx)
-		<-c
+
+		os.Setenv("CONFIG_BACKEND_URL", "https://api.dev.rudderlabs.com")
+		os.Setenv("WORKSPACE_TOKEN", "216Co97d9So9TkqphM0cxBzRxc3")
+		os.Setenv("CONFIG_PATH", "./test_config.yaml")
+		config.Load()
+		logger.Init()
+		backendconfig.Init()
+		code <- m.Run()
```
lvrach, line 56:
```suggestion
	 main.Run()
	 doce <- struct{}
```

regulation-worker/cmd/main_test.go
```diff
@@ -42,24 +42,26 @@ func handler() http.Handler {
 func run(m *testing.M) int {
 	svr := httptest.NewServer(handler())
 	defer svr.Close()
-	workspaceID := "1zzAn8ZshcdkLN5TvP86VqLMT90"
+	workspaceID := "216Co97d9So9TkqphM0cxBzRxc3"
 	svcCtx, svcCancel := context.WithCancel(context.Background())
-
+	code := make(chan int)
```
lvrach, line 47:
```suggestion
	done := make(chan struct{})
```

regulation-worker/cmd/main_test.go
```diff
@@ -42,24 +42,26 @@ func handler() http.Handler {
...
-	config.Load()
-	backendconfig.Init()
-	code := m.Run()
-
-	return code
+	_ = os.Setenv("workspaceID", workspaceID)
+	_ = os.Setenv("urlPrefix", svr.URL)
+	main.Run(svcCtx)
+	statusCode := <-code
+	return statusCode
```
lvrach, line 64:
```suggestion
```suggestion
	code := main.Run(svcCtx)
	<-done
	return code
``````

File: services/filemanager/s3manager.go [hunk 11]
```diff
@@ -68,19 +69,13 @@ func (manager *S3Manager) Upload(file *os.File, prefixes ...string) (UploadOutpu
 		uploadInput.ServerSideEncryption = aws.String("AES256")
 	}
 	output, err := s3manager.Upload(uploadInput)
-	if err != nil {
```
lvrach, line 71: Why we have removed this?

regulation-worker/internal/service/looper.go
```diff
@@ -17,9 +17,21 @@ func (l *Looper) Loop(ctx context.Context) error {
...
+			if isContextCanceled {
+				return nil
+			}
 		} else if err != nil {
 			return err
 		}
 	}
 }
+
+func sleepContext(ctx context.Context, delay time.Duration) bool {
```
lvrach, line 30: cool idea 🚀  Maybe we can added in a common library.

regulation-worker/internal/service/looper.go
```diff
@@ -17,9 +17,21 @@ func (l *Looper) Loop(ctx context.Context) error {
 	for {
 		err := l.Svc.JobSvc(ctx)
 		if err == model.ErrNoRunnableJob {
-			time.Sleep(10 * time.Minute)
+			isContextCanceled := sleepContext(ctx, 10*time.Minute)
+			if isContextCanceled {
+				return nil
+			}
```
lvrach, line 23: [minor]
```suggestion
			if ctxCanceled := sleepCtx(ctx, 10*time.Minute);  ctxCanceled {
				return nil
			}
```

regulation-worker/internal/model/model.go
```diff
@@ -6,19 +6,23 @@ import (
...
-	JobStatusComplete  JobStatus = "complete"
-	JobStatusFailed    JobStatus = "failed"
+	JobStatusUndefined         JobStatus = ""
+	JobStatusPending           JobStatus = "pending"
+	JobStatusRunning           JobStatus = "running"
+	JobStatusComplete          JobStatus = "complete"
+	JobStatusFailed            JobStatus = "failed"
+	JobStatusNotSupported      JobStatus = "not supported"
+	JobStatusInvalidFormat     JobStatus = "invalid format"
+	JobStatusInvalidCredential JobStatus = "invalid credential"
```
lvrach, line 25: What if we use `JobStatusFailed` in place of invalid format/ credential, but also include an error code/message that explains the reason.

regulation-worker/internal/delete/delete_test.go
```diff
@@ -0,0 +1,180 @@
...
+// 					"accessKey":   "xyz",
+// 					"enableSSE":   false,
+// 				},
+// 				DestinationID: "1234",
+// 				Type:          "random",
+// 				Name:          "mock_random",
+// 			},
+// 		},
+// 	}
+// 	for _, tt := range tests {
```
lvrach, line 166: what if we uncomment but skip them
```suggestion
 	for _, tt := range tests {
        t.Skip()
```

regulation-worker/internal/delete/delete.go
```diff
@@ -1,33 +1,46 @@
...
+	Delete(ctx context.Context, job model.Job, destConfig map[string]interface{}, destName string) model.JobStatus
+}
+
+type batchManager interface {
+	Delete(ctx context.Context, job model.Job, destConfig map[string]interface{}, destName string) model.JobStatus
+}
+
+type customManager interface {
+	Delete(ctx context.Context, job model.Job, destConfig map[string]interface{}, destName string) model.JobStatus
 }
```
lvrach, line 24: Those interface are similar, why not  using a single one
```suggestion
type deleter interface {
	Delete(ctx context.Context, job model.Job, destConfig map[string]interface{}, destName string) model.JobStatus
}
```

regulation-worker/internal/delete/delete.go
```diff
@@ -1,33 +1,46 @@
...
-type Deleter struct {
-	API   deleter
-	Batch deleter
+
+type DeleteFacade struct {
+	AM apiManager
+	BM batchManager
+	CM customManager
```
lvrach, line 29:
```suggestion
	AM deleter
	BM deleter
	CM deleter
```

regulation-worker/internal/service/looper.go
saurav-malani: sure

services/filemanager/s3manager.go
saurav-malani: sorry, did it by mistake

regulation-worker/internal/model/model.go
saurav-malani: I was thinking of updating the current status shown in regulation-manager DB with all of these to have more transparency. 
I am planning to log all the errors. 
Given above two points, I am not sure, to whom passing the error would be helpful.

regulation-worker/internal/service/looper.go
saurav-malani: added to utils/misc/misc.go

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -0,0 +1,560 @@
...
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/cenkalti/backoff"
+	"github.com/rudderlabs/rudder-server/config"
+	"github.com/rudderlabs/rudder-server/regulation-worker/internal/model"
+	"github.com/rudderlabs/rudder-server/services/filemanager"
+	_ "go.uber.org/automaxprocs"
```
chandumlg, line 26: remove this

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -0,0 +1,560 @@
...
+}
+
+func (b *Batch) cleanup(prefix string) {
+	err := b.FM.DeleteObjects([]string{filepath.Join(prefix, statusTrackerFileName)})
+	if err != nil {
+		fmt.Println("error during cleanup: %w", err)
+	}
+	err = os.RemoveAll(b.TmpDirPath)
+	if err != nil {
+		fmt.Println("error while deleting temporary directory: %w", err)
```
chandumlg, line 558: User logger instead of fmt.Println (Not urgent, may be next PR)

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -0,0 +1,560 @@
...
+	"github.com/rudderlabs/rudder-server/config"
+	"github.com/rudderlabs/rudder-server/regulation-worker/internal/model"
+	"github.com/rudderlabs/rudder-server/services/filemanager"
+	_ "go.uber.org/automaxprocs"
+	"golang.org/x/sync/errgroup"
+)
+
+var (
+	regexRequiredSuffix   = regexp.MustCompile(".json.gz$")
+	statusTrackerFileName = "statusTracker.txt"
```
chandumlg, line 32: Let's have rudder in the name?

lvrach (approved):

chandumlg (approved):

PR #1427, opened 2021-11-11
unionQuery

Sidddddarth:
**Fixes** # (*issue*)

## Type of change
## Checklist:

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -122,6 +122,11 @@ type JobsDB interface {
 	GetJournalEntries(opType string) (entries []JournalEntryT)
 	JournalDeleteEntry(opID int64)
 	JournalMarkStart(opType string, opPayload json.RawMessage) int64
+
+	//union
+	GetUnprocessedUnion(map[string]int, GetQueryParamsT) []*JobT
+	GetProcessedUnion(map[string]int, GetQueryParamsT) []*JobT
+	GetCustomerCounts(int) map[string]int
 }
 
 /*
--- a/jobsdb/unionQuery.go
+++ b/jobsdb/unionQuery.go
@@ -1,0 +1,284 @@
+package jobsdb
+
+import (
+	"database/sql"
+	"fmt"
+	"strings"
+)
+
+const (
+	selectQuery = `SELECT jobs.job_id, jobs.uuid, jobs.user_id, jobs.parameters, jobs.custom_val, jobs.event_payload, jobs.event_count, jobs.created_at, jobs.expire_at, jobs.customer, sum(jobs.event_count) over (order by jobs.job_id asc) as running_event_counts `
+)
+
+//used to get pickup-counts during server start-up
+func (jd *HandleT) GetCustomerCounts(defaultBatchSize int) map[string]int {
+	unionMap := make(map[string]int)
+	//just using first DS here
+	//get all jobs all over DSs and then calculate
+	//use DSListLock also
+
+	// jd.dsMigrationLock.RLock()
+	jd.dsListLock.RLock()
+	// defer jd.dsMigrationLock.RUnlock()
+	defer jd.dsListLock.RUnlock()
+	rows, err := jd.dbHandle.Query(fmt.Sprintf(`select customer, count(job_id) from %s group by customer;`, jd.getDSList(false)[0].JobTable))
+	jd.assertError(err)
+
+	for rows.Next() {
+		var customer string
+		var count int
+		err := rows.Scan(&customer, &count)
+		jd.assertError(err)
+		unionMap[customer] = count
+	}
+
+	return unionMap
+}
+
+//Unprocessed
+
+func (jd *HandleT) getUnprocessedUnionQuerystring(unionMap map[string]int, ds dataSetT, params GetQueryParamsT) string {
+	var queries []string
+
+	for customer, count := range unionMap {
+		//do cache stuff here
+		queries = append(queries, jd.getSingleCustomerUnprocessedQueryString(customer, count, ds, params, true))
+	}
+
+	return `(` + strings.Join(queries, `) UNION (`) + `)`
+}
+
+func (jd *HandleT) getSingleCustomerUnprocessedQueryString(customer string, count int, ds dataSetT, params GetQueryParamsT, order bool) string {
+	customValFilters := params.CustomValFilters
+	parameterFilters := params.ParameterFilters
+	var sqlStatement string
+
+	if useJoinForUnprocessed {
+		// event_count default 1, number of items in payload
+		sqlStatement = fmt.Sprintf(
+			selectQuery+
+				`FROM %[1]s AS jobs `+
+				`LEFT JOIN %[2]s AS job_status ON jobs.job_id=job_status.job_id `+
+				`WHERE job_status.job_id is NULL AND customer='%[3]s'`,
+			ds.JobTable, ds.JobStatusTable, customer)
+	} else {
+		sqlStatement = fmt.Sprintf(
+			selectQuery+
+				` FROM %[1]s AS jobs `+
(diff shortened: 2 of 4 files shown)


jobsdb/unionQuery.go
```diff
@@ -0,0 +1,284 @@
...
+	jd.assertError(err)
+
+	for rows.Next() {
+		var customer string
+		var count int
+		err := rows.Scan(&customer, &count)
+		jd.assertError(err)
+		unionMap[customer] = count
+	}
+
```
lvrach, line 34: https://golang.org/doc/database/querying#multiple_rows


They are two reasons  `.Next()` will return false and stop iteration:
1. All rows have been consumed 
2. An error took place while iterating

https://pkg.go.dev/database/sql#Rows.Next

For the second reason, once the iteration is over we need to check `.Err`

```suggestion
	}
    if err = rows.Err(); err != nil {
        jd.assertError(err)
    }
```

Maybe we need to introduce this check if it is missing.

jobsdb/unionQuery.go
```diff
@@ -0,0 +1,284 @@
...
+	return unionMap
+}
+
+//Unprocessed
+
+func (jd *HandleT) getUnprocessedUnionQuerystring(unionMap map[string]int, ds dataSetT, params GetQueryParamsT) string {
```
lvrach, line 40: `unionMap` is not very descriptive, a suggestion:

```suggestion
func (jd *HandleT) getUnprocessedUnionQuerystring(customerCount map[string]int, ds dataSetT, params GetQueryParamsT) string {
```

jobsdb/unionQuery.go
```diff
@@ -0,0 +1,284 @@
...
+	defer jd.dsListLock.RUnlock()
+
+	dsList := jd.getDSList(false)
+	outJobs := make([]*JobT, 0)
+
+	//removed count assert, because that params.count is not used..?
+
+	for _, ds := range dsList {
+		jobs := jd.getUnprocessedUnionDS(ds, unionMap, params)
+		outJobs = append(outJobs, jobs...)
```
lvrach, line 123: What if all customer counts are zero? Also is it possible to get to negative counts? I think we need to double-check for that.

Early stop if nothing more to read?

```suggestion
		jobs := jd.getUnprocessedUnionDS(ds, unionMap, params)
		outJobs = append(outJobs, jobs...)
		if len(unionMap) == 0 {
		    break 
		}
```

jobsdb/unionQuery.go
```diff
@@ -0,0 +1,284 @@
...
+	var jobList []*JobT
+	for rows.Next() {
+		var job JobT
+		var _null int
+		err := rows.Scan(&job.JobID, &job.UUID, &job.UserID, &job.Parameters, &job.CustomVal,
+			&job.EventPayload, &job.EventCount, &job.CreatedAt, &job.ExpireAt, &job.Customer, &_null)
+		jd.assertError(err)
+		jobList = append(jobList, &job)
+
+		unionMap[job.Customer] -= 1
```
lvrach, line 153: Shouldn't it be possible, but should we check for `job.Customer` key missing from the `unionMap`?

jobsdb/unionQuery.go
```diff
@@ -0,0 +1,284 @@
...
+
+	//removed count assert, because that params.count is not used..?
+
+	for _, ds := range dsList {
+		jobs := jd.getUnprocessedUnionDS(ds, unionMap, params)
+		outJobs = append(outJobs, jobs...)
+		// count -= len(jobs) //this is not necessary anymore right?
+
+		//we could further limit jobs by eventCount - ignoring that for now
+		// just focusing on getting so many rows of jobs for now
```
lvrach, line 127: We never store multiple events per job in the `router` / `batchrouter` tables. We want to have one jobs per event so we can handle failures and retries independently.

Thus, event count is not necessary here. Plus it is going to introduce a lot of complexity. My opinion is to not implement this.

jobsdb/unionQuery.go
```diff
@@ -0,0 +1,284 @@
...
+	jd.dsListLock.RLock()
+	defer jd.dsMigrationLock.RUnlock()
+	defer jd.dsListLock.RUnlock()
+
+	dsList := jd.getDSList(false)
+	outJobs := make([]*JobT, 0)
+
+	for _, ds := range dsList {
+		jobs := jd.getProcessedUnionDS(ds, unionMap, params)
+		outJobs = append(outJobs, jobs...)
```
lvrach, line 181: Similar to `GetUnprocessedUnion`, we should check and stop early if all counters are zero. Maybe also assert for negative counts.

jobsdb/jobsdb.go
```diff
@@ -122,6 +122,11 @@ type JobsDB interface {
 	GetJournalEntries(opType string) (entries []JournalEntryT)
 	JournalDeleteEntry(opID int64)
 	JournalMarkStart(opType string, opPayload json.RawMessage) int64
+
+	//union
+	GetUnprocessedUnion(map[string]int, GetQueryParamsT) []*JobT
+	GetProcessedUnion(map[string]int, GetQueryParamsT) []*JobT
+	GetCustomerCounts(int) map[string]int
 }
```
lvrach, line 130: We should create a separate interface, that includes all the methods the router code needs.

jobsdb/unionQuery.go
```diff
@@ -0,0 +1,284 @@
...
+				`LEFT JOIN %[2]s AS job_status ON jobs.job_id=job_status.job_id `+
+				`WHERE job_status.job_id is NULL AND customer='%[3]s'`,
+			ds.JobTable, ds.JobStatusTable, customer)
+	} else {
+		sqlStatement = fmt.Sprintf(
+			selectQuery+
+				` FROM %[1]s AS jobs `+
+				`WHERE customer='%[3]s' AND jobs.job_id NOT IN (SELECT DISTINCT(job_status.job_id) FROM %[2]s AS job_status)`,
+			ds.JobTable, ds.JobStatusTable, customer)
+	}
```
lvrach, line 70: I think is a good time to decide to keep one implementation or the other. It will simplify our code and reduce the paths we need to test.

jobsdb/unionQuery.go
Sidddddarth: oh yes, that missed my mind. Won't go ahead with that anymore.

jobsdb/unionQuery.go
Sidddddarth: Yes, it defaults to true. Never seen otherwise, removing the else branch.

jobsdb/unionQuery.go
Sidddddarth: I was under the impression that we would always pick a minimum number of jobs per customer(the `delta` value from fairness equation). 
And the cache, which is next in line would help to not query all customers:
>  if the cache says there's no jobs, we won't include that customer in the union query.

And yes, I'll add the check for negative counts shortly.

jobsdb/unionQuery.go
Sidddddarth: the unionMap would be passed to jobsdb by the router, the keys are not updated(except deleting the key if we get all the jobs), only the values are decreased by the number of jobs we find in each DS. `job.Customer` is the same value scanned that we send in the query, so it's bound to exist in the Map. Also we're deleting the key from the map when we have retrieved the number of jobs we wanted, but the `limit` in the query helps with that situation.
Can you elaborate on more situations where we might come across this..?

jobsdb/unionQuery.go
Sidddddarth: I just realized what you're saying. This is required, yes.

jobsdb/unionQuery.go
Sidddddarth: true, it was just a placeholder.
This looks good

jobsdb/unionQuery.go
Sidddddarth: Please elaborate more about the negative counts, maybe with an example showing how we might come across it.

PR #1434, opened 2021-11-13
Bugfix: response transformer added wrapper response from trasnformer,…

utsabc:
… Response Transform Bugfixes

**Fixes** # (*issue*)
 - Wrapper of status code from transformer
 - handling of failure cases
 
## Type of change
## Checklist:


router/transformer/transformer.go
```diff
@@ -209,62 +221,45 @@ func (trans *HandleT) ResponseTransform(ctx context.Context, responseData integr
...
-			respCode = resp.StatusCode
+			respCode = int(transformerResponse.Status)
 			break
+
+		}
+
+		// fallback if both resp and err are nil
+		if resp == nil && err == nil {
+			panic(fmt.Errorf(`[Response Transform] Client returned nil response and nil error`))
 		}
```
lvrach, line 261:
```suggestion
```
This condition could never happen,  either resp or err will be nil but never both.

> An error is returned if caused by client policy (such as CheckRedirect), or failure to speak HTTP (such as a network connectivity problem). A non-2xx status code doesn't cause an error.
> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

router/transformer/transformer.go
```diff
@@ -186,19 +186,31 @@ func (trans *HandleT) ResponseTransform(ctx context.Context, responseData integr
...
+		//for successful response
+		if err != nil || (resp != nil && resp.StatusCode != http.StatusOK) {
+			var tStatus string
+			if resp != nil {
+				tStatus = resp.Status
+				if resp.StatusCode == http.StatusNotFound {
+					panic(fmt.Errorf("[Response transform doesnot exist for URL: URL: %v", url))
+				}
+
+			}
```
lvrach, line 208: This part or code is a bit hard to read/follow. Also, it is not very idiomatic, usually, we check for err != nil before doing anything else, but checks here are a bit convoluted.

May I suggest a more idiomatic way in combination with if-else or early returnn

```go

if err != nil {
    // ...
	retryCount++
	time.Sleep(retrySleep)
    continue
}

// if err == nil, resp is always set

if resp.StatusCode == http.StatusNotFound {
      panic(fmt.Errorf("[Response transform doesnot exist for URL: URL: %v", url))
}

if resp.StatusCode != http.StatusOK {
	// check of other conditions here	   
	retryCount++
	time.Sleep(retrySleep)
	continue
}

if resp.Body == nil {
  // error handling if body is missing 
}

tempRespData, err = io.ReadAll(resp.Body)
if err != nil {

}




```

router/transformer/transformer.go
```diff
@@ -209,62 +221,45 @@ func (trans *HandleT) ResponseTransform(ctx context.Context, responseData integr
 		if resp != nil && resp.Body != nil {
 			tempRespData, _ = io.ReadAll(resp.Body)
```
lvrach, line 222: Not introduced in this PR, but the error is ignored here.

lvrach (commented):
I have added some suggestions regarding a more idiomatic error handling approach.

Also:
1. Is it possible to write a test for the bug this PR is trying to fix ?
2. Since we are changing this, you could try to introduce https://pkg.go.dev/github.com/cenkalti/backoff/v4, the same way it is used in other parts of our system. It will simplify our code and backoff is better than simple simple.

router/transformer/transformer.go
utsabc: Yeah makes sense, will address this

router/transformer/transformer.go
utsabc: Got it,will address this

router/transformer/transformer.go
sanpj2292: The motivation behind writing the code the way it's written

We retry when there is a problem with transformer service. Hence we've kept the transformer service error handling at a single place.
Also this would avoid code duplication
```golang

 if err != nil || (resp.StatusCode != 200) {
   // There is prob with transformer service
 } 
```
Idiomatic perspective I understand. Want to understand why do you think it's not readable. Can you shed more light on this ?

router/transformer/transformer.go
```diff
@@ -169,103 +170,71 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
 		panic(err)
 	}
-	var resp *http.Response
+
 	var respData []byte
 	var respCode int
-	var tempRespData []byte
+
 	url := getResponseTransformURL(destName)
 	payload := strings.NewReader(string(rawJSON))
```
lvrach, line 183:
```suggestion
	payload := []byte(rawJSON)
```

router/transformer/transformer.go
```diff
@@ -279,6 +248,50 @@ func (trans *HandleT) Setup() {
 	trans.transformerResponseTransformRequestTime = stats.NewStat("router.transformer_response_transform_time", stats.TimerType)
 }
 
+func (trans *HandleT) makeHTTPRequest(ctx context.Context, url string, payload *strings.Reader) ([]byte, int, error) {
```
lvrach, line 251:
```suggestion
func (trans *HandleT) makeHTTPRequest(ctx context.Context, url string, payload []byte) ([]byte, int, error) {
```

router/transformer/transformer.go
```diff
@@ -279,6 +248,50 @@ func (trans *HandleT) Setup() {
...
+	var respData []byte
+	var respCode int
+	payload.Seek(0, io.SeekStart)
+
+	req, err := http.NewRequestWithContext(ctx, "POST", url, payload)
```
lvrach, line 256:
```suggestion
	req, err := http.NewRequestWithContext(ctx, "POST", url, io.NewReader(payload))
```

router/transformer/transformer.go
```diff
@@ -279,6 +248,50 @@ func (trans *HandleT) Setup() {
 	trans.transformerResponseTransformRequestTime = stats.NewStat("router.transformer_response_transform_time", stats.TimerType)
 }
 
+func (trans *HandleT) makeHTTPRequest(ctx context.Context, url string, payload *strings.Reader) ([]byte, int, error) {
+	var respData []byte
+	var respCode int
+	payload.Seek(0, io.SeekStart)
```
sanpj2292, line 254: @utsabc You can remove `.Seek` here & use as @lvrach suggested. This is more from a readability point of view. Kindly make these changes

router/transformer/transformer.go
utsabc: I am not able to get the `NewReader` method from `io` package.  I am able to use `bytes.NewReader()` which is working in expected way

```
req, err := http.NewRequestWithContext(ctx, "POST", url, bytes.NewReader(payload))
```

router/transformer/transformer.go
utsabc: Addressed this suggestion

router/transformer/transformer.go
utsabc: Addressed this suggestion

router/transformer/transformer.go
utsabc: Yes addressed this

cisse21 (approved):

sanpj2292 (approved):

PR #1435, opened 2021-11-15
Temp folder check.

achettyiitr:
**Fixes** # (*issue*)

## Type of change
## Checklist:


utils/misc/misc.go
```diff
@@ -296,12 +298,35 @@ func RemoveFilePaths(filePaths ...string) {
...
+	return
+}
+
 // RemoveEmptyFolderStructureForFilePath recursively cleans up everything till it reaches the stage where the folders are not empty or parent.
 func RemoveEmptyFolderStructureForFilePath(fp string) {
 	if fp == "" {
 		return
 	}
-	for currDir := filepath.Dir(fp); currDir != "/" && currDir != "."; {
+	for currDir := filepath.Dir(fp); currDir != "/" && currDir != "." && !Contains(tempFolders, currDir); {
```
psrikanth88, line 329: @achettyiitr just for my understanding, what was the issue with current code. `rmDir` would only delete empty folders, right?

utils/misc/misc.go
```diff
@@ -296,12 +298,35 @@ func RemoveFilePaths(filePaths ...string) {
...
+// GetTempFolders returns all temporary folders.
+func GetTempFolders() (tempFolders []string) {
+	tmpDirPath, err := CreateTMPDIR()
+	if err != nil {
+		return
+	}
+
+	tempFolders = []string{
+		tmpDirPath,
+		fmt.Sprintf("%s/%s", tmpDirPath, "rudder-async-destination-logs"),
```
psrikanth88, line 310: maybe we maintain a map of these tmp paths and make sure we use them as variables in all places we use them. So that we would not miss adding any new path introduced in code

achettyiitr: Sure.

utils/misc/misc.go
achettyiitr: Actually the issue was, When we were in the deletion flow we were stopping at the root.
Let's say that the time window start at t1 and completes at t2. Before t1 if somebody has created the folders, so we would have deleted it during this time interval. Now after t2, if we are trying to access the files we will get error.

utils/misc/misc.go
achettyiitr: Done.

utils/misc/misc.go
```diff
@@ -296,12 +311,34 @@ func RemoveFilePaths(filePaths ...string) {
...
+	return
+}
+
 // RemoveEmptyFolderStructureForFilePath recursively cleans up everything till it reaches the stage where the folders are not empty or parent.
 func RemoveEmptyFolderStructureForFilePath(fp string) {
 	if fp == "" {
 		return
 	}
-	for currDir := filepath.Dir(fp); currDir != "/" && currDir != "."; {
+	for currDir := filepath.Dir(fp); currDir != "/" && currDir != "." && !Contains(tempFolders, currDir); {
```
achettyiitr, line 341: Use contains for map.

achettyiitr: Done.

sivashanmukh (approved):
LGTM

utils/misc/misc.go
```diff
@@ -296,12 +311,39 @@ func RemoveFilePaths(filePaths ...string) {
...
+
 // RemoveEmptyFolderStructureForFilePath recursively cleans up everything till it reaches the stage where the folders are not empty or parent.
 func RemoveEmptyFolderStructureForFilePath(fp string) {
 	if fp == "" {
 		return
 	}
 	for currDir := filepath.Dir(fp); currDir != "/" && currDir != "."; {
+		// Checking if the currDir is present in the temporary folders or not
+		// If present we should stop at that point.
+		if _, ok := tempFolders[currDir]; ok {
```
dhawal1248, line 344: Can you rename it to something like reservedFolders/internalFolders?

dhawal1248 (approved):
Small commment, otherwise LGTM

utils/misc/misc.go
achettyiitr: Done.

PR #1450, opened 2021-11-23
Feat/live event cache for destination

kuldeep0020:
The scope of this PR is as follows:
- Implementation of an in-memory cache
- Use of the implemented cache to store the source, destination and transformation events if the events are not being viewed live by the user on UI. We store some events and once user wants to view live events then we sends these stored events as well. 

## Type of change
## Checklist:
- [x] I have added unit tests for the code


services/debugger/destination/eventDeliveryStatusUploader.go
```diff
@@ -35,23 +35,43 @@ var uploader debugger.UploaderI
 var (
 	configBackendURL                  string
 	disableEventDeliveryStatusUploads bool
+	eventsDeliveryCacheMap            map[string][]*DeliveryStatusT
+	eventsCacheMapLock                sync.RWMutex
```
lvrach, line 39: Would it be possible to avoid using global variables and instead have them as struct properties. Maybe `EventDeliveryStatusUploader` struct would be suitable for this.

services/debugger/destination/eventDeliveryStatusUploader.go
```diff
@@ -135,3 +161,18 @@ func backendConfigSubscriber(backendConfig backendconfig.BackendConfig) {
...
 }
+
+func recordHistoricEventsDelivery(destinationIDs []string) {
+	for _, destinationID := range destinationIDs {
+		var historicEventsDelivery []*DeliveryStatusT
+		eventsCacheMapLock.Lock()
+		if deliveryStatus, ok := eventsDeliveryCacheMap[destinationID]; ok {
+			historicEventsDelivery = deliveryStatus
+			delete(eventsDeliveryCacheMap, destinationID)
+		}
```
lvrach, line 172: I am not aware of the product requirements here, but should we delete them every time we send them ?

services/debugger/destination/eventDeliveryStatusUploader.go
```diff
@@ -35,23 +35,43 @@ var uploader debugger.UploaderI
...
+
+	if _, ok := cache[destinationID]; !ok {
+		cache[destinationID] = make([]*DeliveryStatusT, 0, maxEventsCacheSize)
+	}
+	eventsCache := cache[destinationID]
+	eventsCache = append(eventsCache, deliveryStatus)
+	if len(eventsCache) > maxEventsCacheSize {
+		eventsCache = eventsCache[len(eventsCache)-maxEventsCacheSize:]
+	}
+	cache[destinationID] = eventsCache
```
lvrach, line 72: An alternate way to implement this would be a ring buffer:
https://github.com/smallnest/ringbuffer
https://pkg.go.dev/container/ring

It gives performance benefits (no memory allocation) and you could avoid Locks at some extend.

However, it would be more complex and I am not sure if our use case here justifies this complexity.

services/debugger/transformation/transformationStatusUploader.go
```diff
@@ -295,3 +224,122 @@ func getEventsAfterTransform(singularEvent types.SingularEventT, receivedAt time
...
+	reportedMessageIDs := make(map[string]struct{})
+	eventBeforeMap := make(map[string]*EventBeforeTransform)
+	eventAfterMap := make(map[string]*EventsAfterTransform)
+	for _, userTransformerEvent := range tStatus.UserTransformedEvents {
+		if userTransformerEvent.Metadata.MessageID != "" {
+			reportedMessageIDs[userTransformerEvent.Metadata.MessageID] = struct{}{}
+			singularEventWithReceivedAt := tStatus.EventsByMessageID[userTransformerEvent.Metadata.MessageID]
+			if _, ok := eventBeforeMap[userTransformerEvent.Metadata.MessageID]; !ok {
+				eventBeforeMap[userTransformerEvent.Metadata.MessageID] = getEventBeforeTransform(singularEventWithReceivedAt.SingularEvent, singularEventWithReceivedAt.ReceivedAt)
+			}
```
lvrach, line 269: This section of code, is rather hard to read. What if:
```suggestion
                metadata := userTransformerEvent.Metadata
		if metadata.MessageID != "" {
			reportedMessageIDs[metadata.MessageID] = struct{}{}
			singularEventWithReceivedAt := tStatus.EventsByMessageID[metadata.MessageID]
			if _, ok := eventBeforeMap[metadata.MessageID]; !ok {
				eventBeforeMap[metadata.MessageID] = getEventBeforeTransform(singularEventWithReceivedAt.SingularEvent, singularEventWithReceivedAt.ReceivedAt)
			}
```

lvrach (commented):
I left a few non-blocking comments. 

Also:
* Could we provide tests for the changes here? 
* Should we keep cached items (potentially) for ever in memory, should they eventually be expired (for GDPR) reasons. This related to case that a few events for a destination and they never send again, we keep them in memory while a customer has ask for data deletion. If 30 days pass with out a server restart it could lead to a violation.

cisse21:
Can we add another go-routine which periodically cleans up the jobs which are present in cache so that we don't bloat up our memory?

lvrach:
> Can we add another go-routine which periodically cleans up the jobs which are present in cache so that we don't bloat up our memory?

The number of events we keep per destination is bounded, so memory bloat up isn't concerning me, but keeping events for longer than 30 days is an issue.

services/debugger/destination/eventDeliveryStatusUploader.go
kuldeep0020: Sounds good.

services/debugger/transformation/transformationStatusUploader.go
kuldeep0020: Introduce variables for the object parameters which are being referenced multiple time

services/debugger/destination/eventDeliveryStatusUploader.go
kuldeep0020: See if we can avoid allocation of new memory every time we add element in the cache.

services/debugger/destination/eventDeliveryStatusUploader.go
kuldeep0020: Keep this in cache, don't delete it, if the control plane supports the polling for the events then we can send the events in case of let say a browser refresh.

services/debugger/destination/cache.go
```diff
@@ -0,0 +1,108 @@
...
+	lock sync.RWMutex
+	Size   int
+	KeyTTL time.Duration
+	m      map[string]*cacheItem
+	once sync.Once
+}
+
+func (c *Cache) init() {
+	c.once.Do(func() {
+		c.Size = CacheMaxSize
```
cisse21, line 35: Why do we need to store CacheMaxSize in Cache Struct. If we store it this way , configuring this variable wont be hot reloadable

services/debugger/source/cache.go
```diff
@@ -0,0 +1,109 @@
+package sourcedebugger
+
```
cisse21, line 2: Can we move all the cache layers onto the top(debugger level) if we are repeating the same code at multiple places.Or is there a specific reason we did this?

services/debugger/destination/cache.go
```diff
@@ -0,0 +1,108 @@
...
+	CacheMaxSize int
+)
+
+func loadCacheConfig() {
+	config.RegisterIntConfigVariable(2, &CacheMaxSize, true, 1, "DestinationDebugger.maxEventsCacheSize")
+	config.RegisterDurationConfigVariable(720 * time.Hour, &CacheTTL, true, 1, "DestinationDebugger.maxEventsCacheTTL")
+}
+
+type cacheItem struct {
+	objs       []*DeliveryStatusT
```
cisse21, line 21: can we rename objs to deliveryEvents?

services/debugger/destination/cache.go
```diff
@@ -0,0 +1,108 @@
...
+
+type Cache struct {
+	lock sync.RWMutex
+	Size   int
+	KeyTTL time.Duration
+	m      map[string]*cacheItem
+	once sync.Once
+}
+
+func (c *Cache) init() {
```
cisse21, line 33: can we change c to cache and Cache to CacheT to maintain consistency with rest of the code base

services/debugger/destination/cache.go
```diff
@@ -0,0 +1,108 @@
...
+	config.RegisterDurationConfigVariable(720 * time.Hour, &CacheTTL, true, 1, "DestinationDebugger.maxEventsCacheTTL")
+}
+
+type cacheItem struct {
+	objs       []*DeliveryStatusT
+	lastAccess time.Time
+}
+
+type Cache struct {
+	lock sync.RWMutex
```
cisse21, line 26: change this to cacheLock

services/debugger/destination/cache.go
```diff
@@ -0,0 +1,108 @@
...
+	c.lock.Lock()
+	if deliveryStatus, ok := c.m[key]; ok {
+		historicEventsDeliveryStatus = deliveryStatus.objs
+		delete(c.m, key)
+	}
+	c.lock.Unlock()
+	return historicEventsDeliveryStatus
+}
+
+func (c *Cache) readData(key string) []*DeliveryStatusT {
```
cisse21, line 83: Looks like we are no longer using this function.Can this be removed

kuldeep0020: Done

services/debugger/destination/cache.go
kuldeep0020: done

services/debugger/destination/cache.go
kuldeep0020: changed to data, keeping it more generic

services/debugger/source/cache.go
kuldeep0020: done

services/debugger/destination/cache.go
kuldeep0020: fixed

services/debugger/destination/cache.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (c *Cache) init() {
+	c.once.Do(func() {
+		c.Size = CacheMaxSize
+		c.m = make(map[string]*cacheItem, c.Size)
+		c.KeyTTL = CacheTTL
+
+		go func() {
+			for now := range time.Tick(time.Second) {
```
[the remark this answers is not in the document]
kuldeep0020: Done

services/debugger/destination/cache.go
```diff
@@ -0,0 +1,108 @@
...
+	KeyTTL time.Duration
+	m      map[string]*cacheItem
+	once sync.Once
+}
+
+func (c *Cache) init() {
+	c.once.Do(func() {
+		c.Size = CacheMaxSize
+		c.m = make(map[string]*cacheItem, c.Size)
+		c.KeyTTL = CacheTTL
```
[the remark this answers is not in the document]
kuldeep0020: Fixed

services/debugger/destination/cache.go
```diff
@@ -0,0 +1,108 @@
...
+type cacheItem struct {
+	objs       []*DeliveryStatusT
+	lastAccess time.Time
+}
+
+type Cache struct {
+	lock sync.RWMutex
+	Size   int
+	KeyTTL time.Duration
+	m      map[string]*cacheItem
```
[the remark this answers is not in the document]
kuldeep0020: done

services/debugger/destination/cache.go
```diff
@@ -0,0 +1,108 @@
...
+
+type cacheItem struct {
+	objs       []*DeliveryStatusT
+	lastAccess time.Time
+}
+
+type Cache struct {
+	lock sync.RWMutex
+	Size   int
+	KeyTTL time.Duration
```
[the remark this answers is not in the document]
kuldeep0020: Done

services/debugger/destination/eventDeliveryStatusUploader.go
```diff
@@ -86,6 +90,10 @@ func Setup(backendConfig backendconfig.BackendConfig) {
 	rruntime.Go(func() {
 		backendConfigSubscriber(backendConfig)
 	})
+
+	rruntime.Go(func() {
```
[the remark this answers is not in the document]
kuldeep0020: Removed

services/debugger/destination/cache.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+var (
+	CacheTTL     time.Duration
+	CacheMaxSize int
+)
+
+func loadCacheConfig() {
+	config.RegisterIntConfigVariable(2, &CacheMaxSize, true, 1, "DestinationDebugger.maxEventsCacheSize")
+	config.RegisterDurationConfigVariable(720 * time.Hour, &CacheTTL, true, 1, "DestinationDebugger.maxEventsCacheTTL")
```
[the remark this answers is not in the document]
kuldeep0020: Done

config/config.yaml
```diff
@@ -47,10 +47,16 @@ Debugger:
   retrySleep: 100ms
 SourceDebugger:
   disableEventUploads: false
+  maxEventsCacheSize: 3
```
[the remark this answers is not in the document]
kuldeep0020: Done

cisse21 (approved):

services/debugger/destination/cache.go
lvrach: I should reply on this one, because I had ask for this change.

It is idiomatic in to avoid name repetition, in this case the struct is already name Cache so naming the lock `cacheLock` is considered redundant. Given this a private field I think `cacheLock` doesn't help more than `lock`.

I understand that naming of things is hard and controversial, so I am not very strong on this one.

services/debugger/cache_test.go
```diff
@@ -0,0 +1,69 @@
...
+			Expect(len(c.cacheMap)).To(Equal(1))
+			Expect(len(c.cacheMap[testKey].data)).To(Equal(1))
+			Expect(c.cacheMap[testKey].data[0]).To(Equal(testValue))
+		})
+
+		It("Cache timeout", func() {
+			var c CacheT
+			c.init()
+			c.keyTTL = time.Second
+			c.cleanupFreq = time.Second
```
lvrach, line 35: c.init() should not be called from outside the methods of the structs
It also should not be called before variables are specified

time.Second is to much for a test,
```suggestion
		    c := CacheT{
		        keyTTL:  10 * time.Millisecond, 
     			cleanupFreq: 10 * time.Millisecond,
		    }
```

services/debugger/cache_test.go
```diff
@@ -0,0 +1,69 @@
...
+			Expect(c.keyTTL).To(Equal(time.Second))
+			c.Update(testKey, testValue)
+			Expect(len(c.cacheMap)).To(Equal(1))
+			Expect(len(c.cacheMap[testKey].data)).To(Equal(1))
+			Expect(c.cacheMap[testKey].data[0]).To(Equal(testValue))
+			time.Sleep(time.Second * 2)
+			Expect(len(c.cacheMap)).To(Equal(0))
```
lvrach, line 42: Avoid sleeps in tests, eventually checks for a condition in a loop and is faster and more robust.
```suggestion
			Eventually(func() int {
			return len(c.cacheMap) }).To(Equal(0))
```

services/debugger/cache.go
```diff
@@ -0,0 +1,100 @@
...
+type cacheItem struct {
+	data       [][]byte
+	lastAccess time.Time
+}
+
+/*
+	CacheT is an in-memory cache. Each key-value pair stored in this cache have a TTL and one goroutine removes the
+key-value pair form the cache which is older than TTL time.
+*/
+type CacheT struct {
```
lvrach, line 30: Not sure why the `T` prefix is added here, I am aware there is an old convention in the rudder-server but I haven't understood why we are doing it.

```suggestion
type Cache struct {
```

services/debugger/cache_test.go
```diff
@@ -0,0 +1,69 @@
...
+			var c CacheT
+			c.Update(testKey, testValue)
+			v := c.ReadAndPopData(testKey)
+			Expect(v).To(Equal([][]byte{testValue}))
+			Expect(len(c.cacheMap)).To(Equal(0))
+		})
+
+		It("Cache data store limit", func() {
+			var c CacheT
+			c.init()
```
lvrach, line 55:
```suggestion
```

services/debugger/cache.go
```diff
@@ -0,0 +1,100 @@
...
+	lock        sync.RWMutex
+	keyTTL      time.Duration // Time after which the data will be expired and removed from the cache
+	cleanupFreq time.Duration // This is the time at which a cleaner goroutines  checks whether the data is expired in cache
+	size        int           // This is the size upto which this cache can store a value corresponding to any key
+	cacheMap    map[string]*cacheItem
+	once        sync.Once
```
lvrach, line 36:
```suggestion
	KeyTTL      time.Duration // Time after which the data will be expired and removed from the cache
	CleanupFreq time.Duration // This is the time at which a cleaner goroutines  checks whether the data is expired in cache
	Size        int           // This is the size upto which this cache can store a value corresponding to any key
	cacheMap    map[string]*cacheItem
	lock        sync.RWMutex
	once        sync.Once
```

configurable properties should be exportable fields

services/debugger/cache.go
```diff
@@ -0,0 +1,100 @@
...
+	"time"
+)
+
+/*
+	loadCacheConfig sets the properties of the cache after reading it from the config file.
+This gives a feature of hot readability as well.
+ */
+func (cache *CacheT) loadCacheConfig() {
+	config.RegisterIntConfigVariable(3, &cache.size, true, 1, "LiveEvent.Cache.size")
+	config.RegisterDurationConfigVariable(time.Duration(720), &cache.keyTTL, true, time.Hour,
```
lvrach, line 15: I am not sure if it a good idea to register these variables multiple times, cc @chandumlg

lvrach (changes requested):

services/debugger/cache.go
cisse21: This was changed just to adhere to the convention

services/debugger/cache.go
```diff
@@ -0,0 +1,100 @@
...
+cleanupFreq, size, empty cacheMap
+ */
+func (cache *CacheT) init() {
+	cache.once.Do(func() {
+		cache.loadCacheConfig()
+		cache.cacheMap = make(map[string]*cacheItem, cache.size)
+
+		go func() {
+			for {
+				time.Sleep(cache.cleanupFreq)
```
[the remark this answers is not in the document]
cisse21: using time.Tick initialised the for Loop and caused the variable to be non hot reloadable, so instead we moved to this approach

services/debugger/destination/cache.go
cisse21: It was previously used as c.lock inside the code so I suggested this change

config/config.yaml
```diff
@@ -45,6 +45,11 @@ Debugger:
   maxRetry: 3
   batchTimeout: 2s
   retrySleep: 100ms
+LiveEvent:
+  cache:
+    size: 3
+    ttl: 30d
```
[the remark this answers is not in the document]
kuldeep0020: done

services/debugger/cache.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+/*
+	loadCacheConfig sets the properties of the cache after reading it from the config file.
+This gives a feature of hot readability as well.
+ */
+func (cache *CacheT) loadCacheConfig() {
+	config.RegisterIntConfigVariable(3, &cache.size, true, 1, "LiveEvent.Cache.size")
+	config.RegisterDurationConfigVariable(time.Duration(720), &cache.keyTTL, true, time.Hour,
+		"LiveEvent.cache.ttl")            // default keyTTL is 30 days
```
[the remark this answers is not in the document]
kuldeep0020: done

services/debugger/cache.go
kuldeep0020: done

services/debugger/cache_test.go
kuldeep0020: done

services/debugger/cache_test.go
kuldeep0020: done

lvrach (approved):

PR #1478, opened 2021-12-06
Throttle for no event order

cisse21:
**Fixes** # (*issue*)
When we have no eventOrdering enabled ,we are currently not enforcing throttling on the jobs.This PR attempts to fix this issue . Also since we are not marking jobs as throttled in any part of the code we have removed that section of the code

## Type of change
## Checklist:


router/router.go
```diff
@@ -1148,6 +1145,19 @@ func (rt *HandleT) findWorker(job *jobsdb.JobT, throttledAtTime time.Time) (toSe
 		return nil
 	}
 
+	err := json.Unmarshal(job.Parameters, &parameters)
+
+	if err == nil && rt.canThrottle(parameters.DestinationID, userID, throttledAtTime) {
```
lvrach, line 1150: 1. I prefer if error checking is seperated from logic, and ideally imditly after method invocation. Requires less cognitive effort to read it.

2.  not introduce in this PR but, `shouldThrottle` seems more suitable than `canThrottle`.
```suggestion
	err := json.Unmarshal(job.Parameters, &parameters)
    if err != nil {
        // error log
        return nil
    }
	if rt.shouldThrottle(parameters.DestinationID, userID, throttledAtTime) {
```

router/router_test.go
```diff
@@ -185,8 +185,7 @@ var _ = Describe("Router", func() {
 			}
 
 			callRetry := c.mockRouterJobsDB.EXPECT().GetToRetry(jobsdb.GetQueryParamsT{CustomValFilters: []string{CustomVal["GA"]}, JobCount: c.dbReadBatchSize}).Return(toRetryJobsList).Times(1)
```
lvrach, line 187: Could we devise a test for throttling, would it be possible?

lvrach (approved):
LGTM, with minor comment. A throttling test would be nice if possible

chandumlg (approved):

PR #1485, opened 2021-12-08
Uses the latest updated_at time; instead of id to load the latest schema

saivarunr:
**Fixes** # (*issue*)
Uses the latest updated_at time; instead of id to load the latest schema

## Type of change
- [ x] Bug fix (non-breaking change which fixes an issue)
## Checklist:
- [ x] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code

What is the issue?
The code fetches the schema changed on the warehouse destination and stores it locally, but it queries to fetch by the latest-id which need not be the latest, as we're [upsert-ing the schema here](https://github.com/rudderlabs/rudder-server/blob/ebd99531d6f9f6f2d6a21d59408c0baf7142d195/warehouse/schema.go#L91-L96)

What is the fix?
The change proposed will get the latest schema known which is by the latest time updated_at and uses it to write to destination.

### DIFF
--- a/warehouse/schema.go
+++ b/warehouse/schema.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+
 	"github.com/rudderlabs/rudder-server/utils/misc"
 	"github.com/rudderlabs/rudder-server/utils/timeutil"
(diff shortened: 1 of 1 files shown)


sivashanmukh:
Closed in favor of 3cee13a0

PR #1537, opened 2021-12-30
Adding query params in payload

utsabc:
**Description* # 
- added support for putting list of string values in config.yaml
- added logic for parsing query_params in webhook only for configured sources

## Type of change
## Checklist:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -570,7 +570,6 @@ func RegisterStringSliceConfigVariable(defaultValue []string, ptr *[]string, isH
 
 	if !isSet {
 		*ptr = defaultValue
-		fmt.Println("vallue from config", *ptr)
 	}
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -536,6 +536,43 @@ func RegisterStringConfigVariable(defaultValue string, ptr *string, isHotReloada
 	}
 }
 
+func RegisterStringSliceConfigVariable(defaultValue []string, ptr *[]string, isHotReloadable bool, keys ...string) {
+	configVarLock.Lock()
+	defer configVarLock.Unlock()
+	var isSet bool
+	configVar := ConfigVar{
+		value:           ptr,
+		isHotReloadable: isHotReloadable,
+		defaultValue:    defaultValue,
+		keys:            keys,
+	}
+	if isHotReloadable {
+		hotReloadableConfig[keys[0]] = &configVar
+	} else {
+		nonHotReloadableConfig[keys[0]] = &configVar
+	}
+	for _, key := range keys {
+		if IsTransformedEnvSet(key) {
+			isSet = true
+			*ptr = GetStringSlice(key, defaultValue)
+			break
+		}
+	}
+	if !isSet {
+		for _, key := range keys {
+			if viper.IsSet(key) {
+				isSet = true
+				*ptr = GetStringSlice(key, defaultValue)
+				break
+			}
+		}
+	}
+
+	if !isSet {
+		*ptr = defaultValue
+	}
+}
+
 // GetInt64 is wrapper for viper's GetInt
 func GetInt64(key string, defaultValue int64) (value int64) {
 
@@ -576,6 +613,26 @@ func GetString(key string, defaultValue string) (value string) {
 		return defaultValue
 	}
 	return viper.GetString(key)
+}
+
+// GetStringSlice is wrapper for viper's GetStringSlice
+func GetStringSlice(key string, defaultValue []string) (value []string) {
+	fmt.Println("TransformKey(key)", TransformKey(key))
+	envVal := GetEnv(TransformKey(key), "")
+
+	// parsing comma separated string from env and populating a slice
+	if envVal != "" {
+		envValList := strings.Split(envVal, ",")
+		for i, elem := range envValList {
+			envValList[i] = strings.ToLower(strings.Trim(elem, " "))
+		}
+		return envValList
+	}
+
+	if !viper.IsSet(key) {
+		return defaultValue
+	}
+	return viper.GetStringSlice(key)
 }
 
 // GetDuration is wrapper for viper's GetDuration
--- a/config/config.go
+++ b/config/config.go
@@ -570,6 +570,7 @@ func RegisterStringSliceConfigVariable(defaultValue []string, ptr *[]string, isH
 
 	if !isSet {
 		*ptr = defaultValue
+		fmt.Println("vallue from config", *ptr)
 	}
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -617,7 +617,6 @@ func GetString(key string, defaultValue string) (value string) {
 
 // GetStringSlice is wrapper for viper's GetStringSlice
 func GetStringSlice(key string, defaultValue []string) (value []string) {
-	fmt.Println("TransformKey(key)", TransformKey(key))
 	envVal := GetEnv(TransformKey(key), "")
 
 	// parsing comma separated string from env and populating a slice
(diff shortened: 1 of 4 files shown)


gateway/webhook/webhook.go
```diff
@@ -247,6 +247,23 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
...
+				req.done <- webhookErrorRespT{err: response.GetStatus(response.RequestBodyReadFailed)}
+				continue
+			}
+
+			tempBody["query_parameters"] = queryParams
+			body, err = json.Marshal(tempBody)
+
+			if err != nil {
+				req.done <- webhookErrorRespT{err: response.GetStatus(response.RequestBodyReadFailed)}
+				continue
```
saikumarrs, line 264: Error types need to be changed

arajguha: can you suggest what the error types should be changed to?

saikumarrs: Not sure if we have them already but something more appropriate,
- RequestBodyUnmarshalFailed
- RequestBodyMarshalFailed
or something more generic for this kind of errors.

gateway/webhook/webhook.go
```diff
@@ -247,6 +248,27 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
 				continue
 			}
 
+			// begin
+			if parseAllQueryParams {
```
saikumarrs, line 252: I don't know if we have source information at this level, but if we have, I would drive it through a configuration that is set only to "Shopify" by default as we don't need this for all sources.

gateway/webhook/webhook.go
arajguha: got it. I could find only one more marshalling error. So will update it

gateway/webhook/webhook.go
arajguha: I think the information we have at this point is just that it's a webhook type source.

arajguha: we could have some information by parsing the payload being sent the webhook URL maybe. But again for that, we are parsing anyway. So not sure on this one.

saikumarrs: @cisse21 any thoughts?

gateway/webhook/webhook.go
```diff
@@ -247,6 +248,27 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
...
+				var tempBody map[string]interface{}
+				err = json.Unmarshal(body, &tempBody)
+
+				if err != nil {
+					req.done <- webhookErrorRespT{err: response.GetStatus(response.RequestBodyReadFailed)}
+					continue
+				}
+
+				tempBody["query_parameters"] = queryParams
+				body, err = json.Marshal(tempBody)
```
saikumarrs, line 263: This additional marshaling logic here seems to be having some impact on the throughput definitely not negligible. One option we are thinking of is to send the query parameters as additional data alongside event payload to the transformer instead of incorporating them within the event body.

@lvrach @cisse21 @chandumlg need help around this. Any other suggestions?

gateway/webhook/webhook.go
arajguha: @saikumarrs  `breq.sourceType` . Here, we have the information re sourceType (here it is Shopify). So do we want to add a check and only parse query params for `breq.sourceType == "Shopify"` ? Will that work?

gateway/webhook/webhook.go
lvrach: > marshaling logic here seems to be having some impact on the throughput definitely not negligible
It would be expectable, but do we know by how much ?

> One option we are thinking of is to send the query parameters as additional data alongside event payload
That is a valid option, but can cause a lot of changes down the line. A similar kind of solution would be to change the payload schema (in a backwards compatible way) to have `body` and `query` at the top level.


A hacky solution for perfromance, is instead of Unmarshal to manipulate json in place. sudo-code:
```go
paramsRawJSON, err = json.Marshal(queryParams)

jsonLastCurl  := body.FindLastIndex("}")
body[:jsonLastCurl] + `"query_parameters":` + paramsRawJSON   + body[body[jsonLastCurl:len(body)] 

```

saikumarrs: Thanks, @lvrach.

> A similar kind of solution would be to change the payload schema (in a backwards compatible way) to have body and query at the top level.
If this is something we can explore, can you help us?

@arajguha please share the old and new (after incorporating the hacky solution) latency reports (probably via a new private channel).

arajguha: on it

saikumarrs (approved):

gateway/webhook/webhook.go
```diff
@@ -247,6 +248,22 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
...
+				if err != nil {
+					req.done <- webhookErrorRespT{err: response.GetStatus(response.ErrorInMarshal)}
+					continue
+				}
+
+				closingBraceIdx := bytes.LastIndexByte(body, '}')
+				appendData := []byte(`, "query_parameters": ` + string(paramsBytes))
+
+				body = append(body[:closingBraceIdx], appendData...)
+				body = append(body, '}')
```
lvrach, line 264: The implementation looks solid, it would be even better if it was accompanied by a test.

The easiest way to test the inject logic, is to extracted in a separate function, for example:

```go
           func injectJSON(rawJSON []byte, key string, value interface{}) error {
				paramsBytes, err := json.Marshal(value)
				if err != nil {
				    return fmt.Errorf("marshal injected value: %w", err)
				}
				closingBraceIdx := bytes.LastIndexByte(rawJSON, '}')
				appendData := []byte(`, "` + key + `": ` + string(paramsBytes))

				rawJSON = append(rawJSON[:closingBraceIdx], appendData...)
				rawJSON = append(rawJSON, '}')
		   }
			```

**Review by lvrach — APPROVED** [review]
LGTM, approving with non-blocking comment.

File: gateway/webhook/webhook.go (hunk 4 above)
Comment (saikumarrs) [reply gateway/webhook/webhook.go:264]: @arajguha / @utsabc can you address this?

File: config/config.go [hunk 5]
```diff
@@ -578,6 +615,26 @@ func GetString(key string, defaultValue string) (value string) {
 	return viper.GetString(key)
 }
 
+// GetStringSlice is wrapper for viper's GetStringSlice
+func GetStringSlice(key string, defaultValue []string) (value []string) {
+	fmt.Println("TransformKey(key)", TransformKey(key))
```
cisse21, line 620: Do we need this log here ?

gateway/webhook/webhook.go
```diff
@@ -247,6 +248,22 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
 				continue
 			}
 
+			if misc.ContainsString(sourceListForParsingParams, strings.ToLower(breq.sourceType)) {
```
cisse21, line 251: does it make sense to convert both of them to lower case for comparison check?

gateway/webhook/webhook.go
```diff
@@ -247,6 +248,22 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
...
+				queryParams := req.request.URL.Query()
+				paramsBytes, err := json.Marshal(queryParams)
+
+				if err != nil {
+					req.done <- webhookErrorRespT{err: response.GetStatus(response.ErrorInMarshal)}
+					continue
+				}
+
+				closingBraceIdx := bytes.LastIndexByte(body, '}')
+				appendData := []byte(`, "query_parameters": ` + string(paramsBytes))
```
cisse21, line 261: Can we append paramsBytes to `[]byte(`, "query_parameters":` , Probably skip the conversion to string and back to byte array

cisse21 (changes requested):

config/config.go
arajguha: will remove this. kept this for debugging

gateway/webhook/webhook.go
arajguha: can do that

cisse21 (approved):

PR #1555, opened 2022-01-07
FileManager tests using emulator to upload, list, get download key, download & delete object.

saurav-malani:
Added tests for following file manager:-
1. S3
2. minio
3. Digital Ocean Space
4. Azure blob storage
5. GCS 

Testing each of these file manager involves testing all of it's exposed methods by:-
1. running an emulator to mock actual server. 
2. uploading test data stored in goldenFile to emulator.
3. listing it.
4. getting download key from file location.
5. getting object name from file location.
6. downloading one of the file from the emulator & verifying if it matches with the one present in golden file.
7. deleting it from emulator.


services/filemanager/s3manager.go
```diff
@@ -102,6 +82,35 @@ func (manager *S3Manager) Download(output *os.File, key string) error {
...
+	https://bucket-name.s3.amazonaws.com/key - >> key
+*/
+
+func (manager *S3Manager) GetObjectNameFromLocation(location string) (string, error) {
+	parsedUrl, err := url.Parse(location)
+	if err != nil {
+		return "", err
+	}
+	trimedUrl := strings.TrimLeft(parsedUrl.Path, "/")
+	if manager.Config.S3ForcePathStyle != nil && *manager.Config.S3ForcePathStyle {
```
saurav-malani, line 96: added check for `S3ForcePathStyle` flag. Since, if  `S3ForcePathStyle` is set to `true` then location url will of type: `http://localhost:55190/filemanager-test-1/some-prefix/original.json.gz`. And, if not set, then it will of type: `http://filemanager-test-1.localhost:55191/some-prefix/original.json.gz`. I.e. in 1st case bucketName if part of path, whereas in 2nd case, it is a part of host.

services/filemanager/s3manager.go
```diff
@@ -102,6 +82,35 @@ func (manager *S3Manager) Download(output *os.File, key string) error {
 	return nil
 }
 
+/*
+GetObjectNameFromLocation gets the object name/key name from the object location url
+	https://bucket-name.s3.amazonaws.com/key - >> key
+*/
+
+func (manager *S3Manager) GetObjectNameFromLocation(location string) (string, error) {
```
saurav-malani, line 90: according to me `GetObjectNameFromLocation` & `GetDownloadKeyFromFileLocation` are serving the same purpose. returning the objectName, which is actually used as key.

services/filemanager/s3manager.go
```diff
@@ -153,53 +162,42 @@ func (manager *S3Manager) getSession() (*session.Session, error) {
...
-	uri, err := url.Parse(location)
-	if err != nil {
-		return "", err
-	}
-	host := uri.Host
-	path := uri.Path[1:]
-	if strings.Contains(host, manager.Config.Bucket) {
-		return path, nil
-	}
-	return strings.TrimPrefix(path, fmt.Sprintf(`%s/`, manager.Config.Bucket)), nil
```
lvrach, line 200: this was probably code for the forcePath option. Kinda dangerous I would say, using `strings.Contains(host, manager.Config.Bucket`  what if the name/key contained the key ?

services/filemanager/s3manager.go
lvrach: Not sure if this happens only when `S3ForcePathStyle` is set. The `Force` makes me think that is if this is not provided we can not guarantee that will not be `S3PathStyle` ...

lvrach: Checkout the following: https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html

lvrach: I believe that the best way to determine if is path style is to check the domain name for the bucket.

If `S3ForcePathStyle` is set OR domain doesn't contain bucket name -> pathStyle

saurav-malani: changes done for this.

services/filemanager/azureBlobStoragemanager.go
```diff
@@ -39,21 +56,25 @@ func (manager *AzureBlobStorageManager) getContainerURL() (azblob.ContainerURL,
...
+
 	p := azblob.NewPipeline(credential, azblob.PipelineOptions{})
 
 	// From the Azure portal, get your storage account blob service URL endpoint.
-	URL, _ := url.Parse(
-		fmt.Sprintf("https://%s.blob.core.windows.net/%s", accountName, manager.Config.Container))
+	URL := manager.getBaseURL()
+	url, err := url.Parse(URL)
+	if err != nil {
+		return azblob.ContainerURL{}, err
```
lvrach, line 74: Few things here:
1. shadowing the import package `url` is not a good idea.
2. `manager.getBaseURL()` should construct and return a url.URL, it safer this way and you don't need to do Parse and error handle.

services/filemanager/azureBlobStoragemanager.go
```diff
@@ -131,6 +135,8 @@ func (manager *AzureBlobStorageManager) ListFilesWithPrefix(prefix string, maxIt
 	}
 
 	// List the blobs in the container
+	ctx := context.Background()
+	marker := string("")
```
lvrach, line 139: Generally more common to do:

```suggestion
    var marker string
```
When you can provide a useful value for something.

services/filemanager/azureBlobStoragemanager.go
```diff
@@ -131,6 +135,8 @@ func (manager *AzureBlobStorageManager) ListFilesWithPrefix(prefix string, maxIt
 	}
 
 	// List the blobs in the container
+	ctx := context.Background()
```
lvrach, line 138: Let's keep this on the top, a reminder that this method should accept a context at some point

services/filemanager/digitalOceanSpacesManager.go
```diff
@@ -133,8 +140,47 @@ func (manager *DOSpacesManager) ListFilesWithPrefix(prefix string, maxItems int6
 	return
 }
 
-func (manager *DOSpacesManager) DeleteObjects(locations []string) (err error) {
-	return
+func (manager *DOSpacesManager) DeleteObjects(keys []string) (err error) {
+	sess := manager.getSession()
+	if err != nil {
+		return fmt.Errorf(`error starting S3 session: %v`, err)
```
lvrach, line 146: mention s3 here might be confusing, also error messages should not contain the word error:
```suggestion
		return fmt.Errorf(`get session: %v`, err)
```

```go
err := fmt.Errorf("establish connection error")

err2 := fmt.Errorf(`error starting S3 session: %v`, err)

if err2 != nil {
    log.Error(err2)
}
```

Would print: `ERROR: error starting S3 session: establish connection error`
if avoid adding error in fmt.Errorf:
`ERROR: starting S3 session: establish connection`

services/filemanager/digitalOceanSpacesManager.go
```diff
@@ -133,8 +140,47 @@ func (manager *DOSpacesManager) ListFilesWithPrefix(prefix string, maxItems int6
...
+			Delete: &s3.Delete{
+				Objects: objects[i:j],
+			},
+		}
+		_, err := svc.DeleteObjects(input)
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				default:
+					pkgLogger.Errorf(`Error while deleting S3 objects: %v, error code: %v`, aerr.Error(), aerr.Code())
```
lvrach, line 173: A lot of coded is added here, is it necessary for testing/refactoring ?

services/filemanager/digitalOceanSpacesManager.go
```diff
@@ -158,15 +206,39 @@ func GetDOSpacesConfig(config map[string]interface{}) *DOSpacesConfig {
 	if config["accessKey"] != nil {
 		accessKey = config["accessKey"].(string)
 	}
-	return &DOSpacesConfig{Bucket: bucketName, EndPoint: endPoint, Prefix: prefix, AccessKeyID: accessKeyID, AccessKey: accessKey}
+	if config["region"] != nil {
+		tmp := config["region"].(string)
+		region = &tmp
+	}
+	if config["forcePathStyle"] != nil {
+		tmp := config["forcePathStyle"].(bool)
```
lvrach, line 214: Given incorrect type: `config["forcePathStyle"] = "foo"`

 this code will panic. Are we ok with this ?

services/filemanager/miniomanager.go
```diff
@@ -25,6 +27,7 @@ func (manager *MinioManager) Upload(file *os.File, prefixes ...string) (UploadOu
 		return UploadOutput{}, err
 	}
 	if err = minioClient.MakeBucket(manager.Config.Bucket, "us-east-1"); err != nil {
+		fmt.Println("error while creating bucket: ", err)
```
lvrach, line 30:
```suggestion
```
debug log ?

services/filemanager/miniomanager.go
```diff
@@ -79,7 +82,12 @@ func (manager *MinioManager) GetObjectNameFromLocation(location string) (string,
 
 //TODO complete this
 func (manager *MinioManager) GetDownloadKeyFromFileLocation(location string) string {
-	return location
+	parsedUrl, err := url.Parse(location)
+	if err != nil {
+		fmt.Println("error while parsing location url: ", err)
+	}
+	trimedUrl := strings.TrimLeft(parsedUrl.Path, "/")
+	return strings.TrimPrefix(trimedUrl, fmt.Sprintf(`%s/`, manager.Config.Bucket))
```
lvrach, line 90: What if s3 path style is not used ? Is it always used for minio ? Was it working before ?

services/filemanager/miniomanager.go
```diff
@@ -116,8 +124,19 @@ func GetMinioConfig(config map[string]interface{}) *MinioConfig {
 	}
 }
 
-func (manager *MinioManager) DeleteObjects(locations []string) (err error) {
-	return
+func (manager *MinioManager) DeleteObjects(keys []string) (err error) {
+	minioClient, err := minio.New(manager.Config.EndPoint, manager.Config.AccessKeyID, manager.Config.SecretAccessKey, manager.Config.UseSSL)
```
lvrach, line 128: Should a minioClient being created every time ?
Could it be part of the struct ?

services/filemanager/azureBlobStoragemanager.go
```diff
@@ -31,6 +31,23 @@ func supressMinorErrors(err error) error {
...
+	endpoint := "blob.core.windows.net"
+	if manager.Config.EndPoint != nil {
+		endpoint = *manager.Config.EndPoint
+	}
+
+	baseURL := fmt.Sprintf("%s://%s.%s/%s/", protocol, manager.Config.AccountName, endpoint, manager.Config.Container)
+	if manager.Config.ForcePathStyle != nil && *manager.Config.ForcePathStyle {
+		baseURL = fmt.Sprintf("%s://%s/%s/%s/", protocol, *manager.Config.EndPoint, manager.Config.AccountName, manager.Config.Container)
+	}
+	return baseURL
```
lvrach, line 49: construct and return url.URL

services/filemanager/fileManager_test.go
```diff
@@ -0,0 +1,339 @@
...
+//run minio server & store data in it.
+func run(m *testing.M) int {
+
+	flag.BoolVar(&hold, "hold", false, "hold environment clean-up after test execution until Ctrl+C is provided")
+	flag.Parse()
+
+	//docker pool setup
+	pool, err := dockertest.NewPool("")
+	if err != nil {
+		log.Fatalf("Could not connect to docker: %s", err)
```
lvrach, line 55: Lets avoid Fatal here:
```suggestion
		panic(fmt.Errorf("Could not connect to docker: %s", err))
```

services/filemanager/digitalOceanSpacesManager.go
```diff
@@ -81,37 +87,38 @@ func (manager *DOSpacesManager) Download(output *os.File, key string) error {
...
+	return trimedUrl
 }
 
 /*
 GetObjectNameFromLocation gets the object name/key name from the object location url
 	https://rudder.sgp1.digitaloceanspaces.com/key - >> key
 */
 func (manager *DOSpacesManager) GetObjectNameFromLocation(location string) (string, error) {
-	uri, err := url.Parse(location)
+	parsedUrl, err := url.Parse(location)
```
lvrach, line 107: minor
```suggestion
	parsedURL, err := url.Parse(location)
```

services/filemanager/fileManager_test.go
```diff
@@ -0,0 +1,387 @@
...
+	for _, tt := range tests {
+
+		t.Run(tt.name, func(t *testing.T) {
+			fmFactory := filemanager.FileManagerFactoryT{}
+			fm, err := fmFactory.New(&filemanager.SettingsT{
+				Provider: tt.destName,
+				Config:   tt.config,
+			})
+			if err != nil {
+				panic(err)
```
lvrach, line 271:
```suggestion
				t.Fatal(err)
```

services/filemanager/fileManager_test.go
```diff
@@ -0,0 +1,387 @@
...
+			}
+
+			//upload all files
+			fmt.Println(" uploading files")
```
lvrach, line 275:
```suggestion
			t.Log(" uploading files")
```

services/filemanager/fileManager_test.go
```diff
@@ -0,0 +1,387 @@
...
+			uploadOutputs := make([]filemanager.UploadOutput, 0)
+			for _, file := range fileList {
+				filePtr, err := os.Open(file)
+				if err != nil {
+					fmt.Println("error while opening testData file to upload: ", err)
```
lvrach, line 280:
```suggestion
					t.Fatalf("error while opening testData file to upload: ", err)
```

services/filemanager/fileManager_test.go
```diff
@@ -0,0 +1,387 @@
...
+				}
+				uploadOutput, err := fm.Upload(filePtr)
+				if err != nil {
+					panic(err)
+				}
+				uploadOutputs = append(uploadOutputs, uploadOutput)
+				filePtr.Close()
+			}
+			fmt.Println("file uploading successful... trying to list uploaded files")
```
lvrach, line 289:
```suggestion
```

services/filemanager/fileManager_test.go
```diff
@@ -0,0 +1,387 @@
...
+			//list files using ListFilesWithPrefix
+			originalFileObject, err := fm.ListFilesWithPrefix("", 1000)
+			if err != nil {
+				fmt.Println("error while getting file object: ", err)
+			}
```
lvrach, line 294:
```suggestion
			require.NoError(t, err, "while getting file object")
```

services/filemanager/digitalOceanSpacesManager.go
saurav-malani: this code is not added just for testing... for all the file managers, whichever function wasn't implemented, I implemented them. So, this is to delete from bucket.

saurav-malani: also, since DeleteObjects takes a max of 1000 objects at a time. So, we have the `for` loop to delete in batch of 1000 objects at a time.

services/filemanager/miniomanager.go
saurav-malani: minio by always uses pathStyle url, as per my knowledge. and previously we didn't had a working code for it.

services/filemanager/miniomanager.go
saurav-malani: good idea

lvrach (approved):

chandumlg (approved):

PR #1596, opened 2022-01-25
Batch integration test

saurav-malani:
These tests test e2e regulation worker to delete from batch type: S3. 
Batch deleter test does the following:
1. Creates minio server to mock s3. 
2. Inserts mock data in it.
2. mocks regulation-manager to get job & update job status
3. mocks config-backend to get config of the destination to delete from.
4. run regulation-worker.
5. regulation worker picks job from mock regulation-manager & delete from batch destination.
6. verification: verifying if after deletion mock s3 file is updated and is same as the golden file.


regulation-worker/cmd/main_test.go
```diff
@@ -287,6 +411,69 @@ func getWorkspaceConfig(w http.ResponseWriter, r *http.Request) {
 	_, _ = w.Write(body)
 }
 
+func verifyBatchDelete() error {
```
lvrach, line 414: Is better for testing helper functions to use `t` for assertions instead of returning an error
```suggestion
func verifyBatchDelete(t *testing.T) {
    t.Helper()
```

services/filemanager/miniomanager.go
```diff
@@ -25,6 +27,7 @@ func (manager *MinioManager) Upload(file *os.File, prefixes ...string) (UploadOu
 		return UploadOutput{}, err
 	}
 	if err = minioClient.MakeBucket(manager.Config.Bucket, "us-east-1"); err != nil {
+		fmt.Println("error while creating bucket: ", err)
```
lvrach, line 30:
```suggestion
```
Debug log ?

lvrach (approved):

chandumlg (approved):

PR #1599, opened 2022-01-26
Integration test code refactor

rspriyamdixit:
**Fixes** # (*issue*)
Temporary PR for initial Review :
1. Code restructure 
2. .env file for integration test
3. Warehouse Test restructure
## Type of change
## Checklist:


lvrach (changes requested):
Please avoid the usage of global variables, like:
```
var (
	pool             *dockertest.Pool
	err              error
	network          *dc.Network
	brokerPort       string
	localhostPort    string
	localhostPortInt int
)
```

in favor of returning concreate structs.

lvrach (changes requested):
The usage of global variables is not resolved. I've left an example on how can be done for Kafka. It should be hard to apply this in other destinations as well.

testhelper/destination/rudderserver.go
```diff
@@ -0,0 +1,136 @@
...
+type ServerTest struct {
+	db *sql.DB
+}
+
+type MINIO struct {
+	minioEndpoint   string
+	minioBucketName string
+}
+
+func SetJobsDB(Test *Test) (*sql.DB, *dockertest.Resource) {
```
lvrach, line 31: Even if we currently only use sql.DB, is better to return the whole `type ServerTest struct`. Making the `db` exportable `DB`. 

This way if we need to return something extra like:
* database name
* username
* password
* port

We can easily add them.

I would suggest even now to return DB_DSN in form of `url.URL`

testhelper/destination/kafka.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+		Env: []string{"ZOOKEEPER_CLIENT_PORT=2181"},
+	})
+	if err != nil {
+		fmt.Println(err)
+	}
+	return Test, z
+}
+
+func SetKafka(Test *Test, z *dockertest.Resource) *dockertest.Resource {
```
lvrach, line 57: this can remain unexported
```suggestion
func setKafka(Test *Test, z *dockertest.Resource) *dockertest.Resource {
```

lvrach (approved):

chandumlg (approved):

PR #1606, opened 2022-01-27
Processor cpu optimisations and tracing

lvrach:
Requires https://github.com/rudderlabs/rudder-server-enterprise/pull/62

# Impact 

It is hard to gauge the exact impact of the changes overall, given that the transformation path has a lot of dependencies, thus creating benchmarks or isolating performance requires a lot of effort.

I run rudder-server under the same load and configuration with and without the changes for a long duration (more than an hour) and measure the average CPU and dest processing times: 

## CPU

**Without** the changes: 
Screenshot 2022-01-30 at 10 01 16 PM

**With** the changes: 
Screenshot 2022-01-30 at 7 15 22 PM

Around ~15% less CPU usage is expected.

## Processing

**With** the changes: 
Screenshot 2022-01-30 at 7 11 59 PM

**Without** the changes: 
Screenshot 2022-01-30 at 9 58 00 PM

# How 

## Observability

`/profile` and `/trace` endpoints was used extensively to find critical paths needing optimization. In addition to that custom tracing annotation was added, to improve observability over transformation code.

## Optimise JSON (Un) Marshaling

Start using  [json-iterator](https://github.com/json-iterator/go) for (un)marshaling operations. 

## Optimise UUID generation 

by replacing `gofrs/uuid` with  `google/uuid`.

As you can observe from the benchmarks below `google/uuid` is much faster than  `gofrs/uuid`. The main reason is google one is using a random pool so it doesn't have to do a syscall every time.

However, replacing gofrs completely wasn't something I wanted to risk doing just before the new release. Since it would require significant testing. So I've created `misc.FastUUID()`, which generates uuid with google's implementation and coverts it via binary to gofrs struct.

### benchmarks

`"github.com/google/uuid"` vs `"github.com/gofrs/uuid"`
```
goos: darwin
goarch: arm64
pkg: github.com/rudderlabs/rudder-server/processor
Benchmark_GOOGLE_UUID-10              	24369831	        50.32 ns/op	       0 B/op	       0 allocs/op
Benchmark_GOOGLE_UUID_STR_GOFRS-10    	 4280739	       272.5 ns/op	      96 B/op	       2 allocs/op
Benchmark_GOOGLE_UUID_BIN_GOFRS-10    	21232407	        58.20 ns/op	       0 B/op	       0 allocs/op
Benchmark_FAST_UUID-10                	18141252	        59.94 ns/op	       0 B/op	       0 allocs/op
Benchmark_GOFRS_UUID-10               	 3281251	       346.7 ns/op	      16 B/op	       1 allocs/op
```


processor/processor.go
```diff
@@ -856,7 +865,7 @@ func (proc *HandleT) updateMetricMaps(countMetadataMap map[string]MetricMetadata
 		var eventType string
 		eventName = event.Metadata.EventName
 		eventType = event.Metadata.EventType
-		countKey := fmt.Sprint(event.Metadata.SourceID, METRICKEYDELIMITER, event.Metadata.DestinationID, METRICKEYDELIMITER, event.Metadata.SourceBatchID, METRICKEYDELIMITER, eventName, METRICKEYDELIMITER, eventType)
+		countKey := strings.Join([]string{event.Metadata.SourceID, event.Metadata.DestinationID, event.Metadata.SourceBatchID, eventName, eventType}, METRICKEYDELIMITER)
```
lvrach, line 868: need further investigation / optimisation

processor/transformer/transformer.go
```diff
@@ -354,7 +371,11 @@ func (trans *HandleT) request(url string, data []TransformerEventT) []Transforme
 	var transformerResponses []TransformerResponseT
 	if resp.StatusCode == http.StatusOK {
 		integrations.CollectIntgTransformErrorStats(respData)
-		err = json.Unmarshal(respData, &transformerResponses)
+
+		trace.Logf(ctx, "Unmarshal", "response raw size: %d", len(respData))
```
cisse21, line 375: Do we need these trace logs in Production?

utils/misc/misc.go
```diff
@@ -992,6 +1000,11 @@ func IsValidUUID(uuid string) bool {
 	return r.MatchString(uuid)
 }
 
+func FastUUID() uuid.UUID {
+	b, _ := gluuid.New().MarshalBinary()
```
chandumlg, line 1004: handle error?

chandumlg (approved):

.enterprise-commit
```diff
@@ -1 +1 @@
-d654771f7f6d6457eec7622d28637c5a23c1edbb
+a4b3a4a5e407cb976e67769102f5b0707566f27f
```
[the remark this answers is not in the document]
chandumlg: enterprise PR commit updated.

processor/transformer/transformer.go
chandumlg: These will help analysing traces.

cisse21 (approved):

PR #1609, opened 2022-01-28
Multitenant: Fair Pickup Algorithm

cisse21:
# Fair Pickup Algorithm 

The major addition of this PR was the implementation of a fair pickup algorithm for the router. 
https://www.notion.so/rudderstacks/Rudder-Server-Router-Fairness-d607203b08024e2aa0415381f72f30ad
## Tenant Stats

[services/multitenant/tenantstats.go](https://github.com/rudderlabs/rudder-server/pull/1609/files#diff-33993c76966b72dcca4a64a9e4fb0bd71565741152a64321f758d01460451f75) is a new component in rudder-server.

It is responsible for collecting tenant and destination-specific stats:
* Router jobsdb **pile up** (unprocessed jobs) per destType, tenant (workspaceID). This is done by incrementing on processor db writes and decrementing on router pickup. It also queries the db to init in the beginning.
* **Failure rate** moving average per destType/tenant, computed on router
* Delivery latency 

Having the above stats in-memory  `Tenant Stats`, runs the actual pickup algorithm on `GetRouterPickupJobs`. Along with some constraints (how many jobs we can pick up this time), it returns how many jobs should be picked by for every customer.

## UnionQuery

[jobsdb/unionQuery.go](https://github.com/rudderlabs/rudder-server/pull/1609/files#diff-751b0a50aa5b5c56377a09785722fca6fd2d8437e1dd197b61ae267051daa822) extends jobsdb functionality with the needs of multitenant.

Using [composition](https://www.geeksforgeeks.org/composition-in-golang/), all of the `jobsdb.HandleT` fields and methods are available in the new component as well.

```
type MultiTenantHandleT struct {
	*HandleT
	Cache CacheOperator
}
```

We have also abstracted the Cache, for testing and [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle).

### GetAllJobs

The most significant method is `GetAllJobs`, a new DB query that can fetch jobs with **multiple states from multiple users** 

In the existing implementation, we were only picking jobs in order independently of workspaceID. This however can lead to suboptimal system throughput and starvation of some tenants (workspaceIDs).  In tandem with tenant stats, this new query ensure we are getting the optimal amount of jobs per workspaceID.

In the previous implementation, we were using 3 separate queries to get all the jobs (you can see that in legacy implementation) of different states. Now we only need a single query.

## Delivery time limit

resultSetT TODO

## Switching on/off

The [null object pattern](https://sourcemaking.com/design_patterns/null_object) has been used, to enable convenient switching of new and existing functionality.

Instead of using `if` statements to select between the two behaviors a common interface is used, with two implementations. We [DI](https://en.wikipedia.org/wiki/Dependency_injection) the implementation we want early in the program. 

### tenantstats

A `noop` implementation for tenantstats has been added. Most of the methods do nothing, except of [GetRouterPickupJobs](https://github.com/rudderlabs/rudder-server/pull/1609/files#diff-4c47c8895388c7e9232046929d59b2a09ea220cbfc82fd2e6b242e4425de1d4cR17) which returns a single workspace with `{"0": ...}` all the available jobs.

### legacy union query 

[jobsdb/unionQueryLegacy.go 
](jobsdb/unionQueryLegacy.go) Implement all of the interface methods, but used the existing implementation and doesn't support quering from multiple users. It expects the map (`{"0": ...}`) that the noop tenantstats implementation returns.

### Reporting

Reporting functionality and feature flag already existed, but it was realling heavily on if statements. Our existing implementation took advantage of some computation done for reporting. Thus, we didn't want to have the computation inside a reporting flag.

Thus the same null object pattern has been followed by reporting as well.

 
### Factories

Router factories have been introduced because the number of params we were passing inside `app/apphandlers/*` was getting out of hand.


services/db/nullRecovery.go
```diff
@@ -21,7 +21,8 @@ func HandleNullRecovery(forceNormal bool, forceDegraded bool, forceStandBy bool,
 		pkgLogger.Info("Setting server mode to Migration. If this is not intended remove environment variables related to Migration.")
 		forceMode = migrationMode
 	} else {
-		forceMode = getForceRecoveryMode(false, false, forceStandBy)
+		//Check for its implications ... Changing it to enable degraded for now
+		forceMode = getForceRecoveryMode(forceNormal, forceDegraded, forceStandBy)
```
cisse21, line 25: Something which needs to be looked at and tested a bit more

gateway/gateway.go
```diff
@@ -1099,8 +1113,13 @@ func (gateway *HandleT) webRequestHandler(rh RequestHandler, w http.ResponseWrit
 	var errorMessage string
 	defer func() {
 		if errorMessage != "" {
-			gateway.logger.Info(fmt.Sprintf("IP: %s -- %s -- Response: 400, %s", misc.GetIPFromReq(r), r.URL.Path, response.GetStatus(errorMessage)))
-			http.Error(w, response.GetStatus(errorMessage), 400)
+			if strings.Contains(errorMessage, response.GetStatus(response.TooManyRequests)) {
+				gateway.logger.Info(fmt.Sprintf("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, http.StatusTooManyRequests, errorMessage))
```
lvrach, line 1117:
```suggestion
				gateway.logger.Infof("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, http.StatusTooManyRequests, errorMessage)
```

utils/misc/misc.go
```diff
@@ -85,6 +85,17 @@ type RudderError struct {
...
+type Pair struct {
+	Key   string
+	Value float64
+}
+
+type PairList []Pair
+
+func (p PairList) Len() int           { return len(p) }
+func (p PairList) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+func (p PairList) Less(i, j int) bool { return p[i].Value < p[j].Value }
```
lvrach, line 97: Not sure if `misc` is the correct place for this. Could be moved to router package or sub-package since it is router specific at the moment.

cisse21: Updated this

jobsdb/jobsdb.go
```diff
@@ -2011,18 +2084,18 @@ func (jd *HandleT) getProcessedJobsDS(ds dataSetT, getAll bool, limitCount int,
...
+	// }
 
 	tags := StatTagsT{CustomValFilters: params.CustomValFilters, StateFilters: params.StateFilters, ParameterFilters: params.ParameterFilters}
 	queryStat := jd.getTimerStat("processed_ds_time", tags)
 	queryStat.Start()
 	defer queryStat.End()
 
 	// We don't reset this in case of error for now, as any error in this function causes panic
-	jd.markClearEmptyResult(ds, stateFilters, customValFilters, parameterFilters, willTryToSet, nil)
+	// jd.markClearEmptyResult(ds, stateFilters, customValFilters, parameterFilters, willTryToSet, nil)
```
lvrach, line 2098: What are we going to do regarding this commented-out code?

jobsdb/jobsdb.go
```diff
@@ -2011,18 +2084,18 @@ func (jd *HandleT) getProcessedJobsDS(ds dataSetT, getAll bool, limitCount int,
...
 	checkValidJobState(jd, stateFilters)
 
-	if jd.isEmptyResult(ds, stateFilters, customValFilters, parameterFilters) {
-		jd.logger.Debugf("[getProcessedJobsDS] Empty cache hit for ds: %v, stateFilters: %v, customValFilters: %v, parameterFilters: %v", ds, stateFilters, customValFilters, parameterFilters)
-		return []*JobT{}
-	}
+	// if jd.isEmptyResult(ds, stateFilters, customValFilters, parameterFilters) {
+	// 	jd.logger.Debugf("[getProcessedJobsDS] Empty cache hit for ds: %v, stateFilters: %v, customValFilters: %v, parameterFilters: %v", ds, stateFilters, customValFilters, parameterFilters)
+	// 	return []*JobT{}
+	// }
```
lvrach, line 2090: What are we going to do regarding this commented-out code?

.github/workflows/tests.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-go@v2
       with:
-        go-version: '^1.16.4'
+        go-version: '~1.16.4'
```
lvrach, line 37: FYI: `'^1.16.4'` was matching with the latest go version 1.17.x - `^` can use any minor or patch version bigger than the specified one. We changed it to `~` to only match with any bigger patch version.

utils/misc/ewma.go
```diff
@@ -0,0 +1,138 @@
+package misc
+
+import "sync"
+
+const (
+	// By default, we average over a one-minute period, which means the average
+	// age of the metrics in the period is 30 seconds.
+	AVG_METRIC_AGE float64 = 30.0
```
cisse21, line 8: should we be making these variables configurable ? or go ahead with the defaults

jobsdb/jobsdb.go
```diff
@@ -1406,8 +1419,8 @@ func (jd *HandleT) setSequenceNumber(newDSIdx string) dataSetT {
 	//Now set the min JobID for the new DS just added to be 1 more than previous max
 	if len(dRangeList) > 0 {
 		newDSMin := dRangeList[len(dRangeList)-1].maxJobID
-		jd.assert(newDSMin > 0, fmt.Sprintf("newDSMin:%d <= 0", newDSMin))
-		sqlStatement := fmt.Sprintf(`SELECT setval(pg_get_serial_sequence('%s_jobs_%s', 'job_id'), %d)`,
+		// jd.assert(newDSMin > 0, fmt.Sprintf("newDSMin:%d <= 0", newDSMin))
```
lvrach, line 1422:
```suggestion
		// jd.assert(newDSMin > 0, fmt.Sprintf("newDSMin:%d <= 0", newDSMin))
```

chandumlg (approved):

Sidddddarth (approved):

PR #1622, opened 2022-02-02
chore(tests): Remove sleep(), use wait logic

lvrach:
* Fixes overall testing time. Average execution time for the affected tests is now 6x faster (down to 2 from 12 seconds).
* All `• [SLOW TEST:5.000 seconds]` tags are now removed from this part.

## Type of change
## Checklist:

### DIFF
--- a/mocks/services/debugger/uploader.go
+++ b/mocks/services/debugger/uploader.go
@@ -72,11 +72,9 @@ func (m *MockUploaderI) EXPECT() *MockUploaderIMockRecorder {
 }
 
 // RecordEvent mocks base method.
-func (m *MockUploaderI) RecordEvent(arg0 interface{}) bool {
+func (m *MockUploaderI) RecordEvent(arg0 interface{}) {
 	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "RecordEvent", arg0)
-	ret0, _ := ret[0].(bool)
-	return ret0
+	m.ctrl.Call(m, "RecordEvent", arg0)
 }
 
 // RecordEvent indicates an expected call of RecordEvent.
@@ -95,4 +93,16 @@ func (m *MockUploaderI) Start() {
 func (mr *MockUploaderIMockRecorder) Start() *gomock.Call {
 	mr.mock.ctrl.T.Helper()
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Start", reflect.TypeOf((*MockUploaderI)(nil).Start))
+}
+
+// Stop mocks base method.
+func (m *MockUploaderI) Stop() {
+	m.ctrl.T.Helper()
+	m.ctrl.Call(m, "Stop")
+}
+
+// Stop indicates an expected call of Stop.
+func (mr *MockUploaderIMockRecorder) Stop() *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Stop", reflect.TypeOf((*MockUploaderI)(nil).Stop))
 }
--- a/services/debugger/destination/eventDeliveryStatusUploader_test.go
+++ b/services/debugger/destination/eventDeliveryStatusUploader_test.go
@@ -6,6 +6,7 @@ import (
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
 	"github.com/rudderlabs/rudder-server/config"
 	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
 	mocksBackendConfig "github.com/rudderlabs/rudder-server/mocks/config/backend-config"
@@ -204,6 +205,7 @@ var _ = Describe("eventDeliveryStatusUploader", func() {
 			EventName:     `some_event_name`,
 			EventType:     `some_event_type`,
 		}
+		disableEventDeliveryStatusUploads = false
 	})
 
 	AfterEach(func() {
@@ -211,7 +213,7 @@ var _ = Describe("eventDeliveryStatusUploader", func() {
 	})
 
 	Context("RecordEventDeliveryStatus", func() {
-		It("returns false if disableEventDeliveryStatusUploads is false", func() {
+		It("returns false if disableEventDeliveryStatusUploads is true", func() {
 			mockCall := c.mockBackendConfig.EXPECT().Subscribe(gomock.Any(), backendconfig.TopicBackendConfig).
 				Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) {
 					// on Subscribe, emulate a backend configuration event
@@ -222,12 +224,13 @@ var _ = Describe("eventDeliveryStatusUploader", func() {
 					}()
 				})
 			tFunc := c.asyncHelper.ExpectAndNotifyCallback()
-			mockCall.Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) { tFunc() }).Return().Times(1)
+			mockCall.Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) {
+				tFunc()
+			}).Return().Times(1)
 
-			time.Sleep(1 * time.Second)
+			c.asyncHelper.WaitWithTimeout(5 * time.Second)
 			disableEventDeliveryStatusUploads = true
 			Expect(RecordEventDeliveryStatus(DestinationIDEnabledA, &deliveryStatus)).To(BeFalse())
-			disableEventDeliveryStatusUploads = false
 		})
 
 		It("returns false if destination_id is not in uploadEnabledDestinationIDs", func() {
@@ -241,9 +244,11 @@ var _ = Describe("eventDeliveryStatusUploader", func() {
 					}()
 				})
 			tFunc := c.asyncHelper.ExpectAndNotifyCallback()
-			mockCall.Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) { tFunc() }).Return().Times(1)
+			mockCall.Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) {
+				tFunc()
+			}).Return().Times(1)
 
-			time.Sleep(1 * time.Second)
+			c.asyncHelper.WaitWithTimeout(5 * time.Second)
 			Expect(RecordEventDeliveryStatus(DestinationIDEnabledB, &deliveryStatus)).To(BeFalse())
 		})
 
@@ -258,10 +263,13 @@ var _ = Describe("eventDeliveryStatusUploader", func() {
 					}()
 				})
 			tFunc := c.asyncHelper.ExpectAndNotifyCallback()
-			mockCall.Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) { tFunc() }).Return().Times(1)
+			mockCall.Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) {
+				tFunc()
+			}).Return().Times(1)
 
-			time.Sleep(1 * time.Second)
-			Expect(RecordEventDeliveryStatus(DestinationIDEnabledA, &deliveryStatus)).To(BeTrue())
+			c.asyncHelper.WaitWithTimeout(5 * time.Second)
+			eventuallyFunc := func() bool { return RecordEventDeliveryStatus(DestinationIDEnabledA, &deliveryStatus) }
+			Eventually(eventuallyFunc).Should(BeTrue())
 		})
 
 		It("transforms payload properly", func() {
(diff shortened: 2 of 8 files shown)


chandumlg (approved):

cisse21 (approved):

services/debugger/uploader.go
```diff
@@ -137,9 +152,13 @@ func (uploader *Uploader) handleEvents() {
 	}
 }
 
-func (uploader *Uploader) flushEvents() {
+func (uploader *Uploader) flushEvents(ctx context.Context) {
 	for {
 		time.Sleep(uploader.batchTimeout)
+		select {
+		case <-ctx.Done():
+		case <-time.After(uploader.batchTimeout):
```
lvrach, line 160: time.After inside this loop can create memory leaks, maybe we should use a ticker instead. 🤔

services/debugger/uploader.go
```diff
@@ -137,9 +152,13 @@ func (uploader *Uploader) handleEvents() {
 	}
 }
 
-func (uploader *Uploader) flushEvents() {
+func (uploader *Uploader) flushEvents(ctx context.Context) {
 	for {
 		time.Sleep(uploader.batchTimeout)
```
lvrach, line 157:
```suggestion
```
should we remove this ?

services/debugger/uploader.go
lvrach: My bad actually, in this case, we should be good.

services/debugger/destination/eventDeliveryStatusUploader_test.go
```diff
@@ -275,9 +301,12 @@ var _ = Describe("eventDeliveryStatusUploader", func() {
...
 				})
 			tFunc := c.asyncHelper.ExpectAndNotifyCallback()
-			mockCall.Do(func(channel chan utils.DataEvent, topic backendconfig.Topic) { tFunc() }).Return().Times(1)
+			mockCall.Do(func(channel chan utils.DataEvent, topic backendconfig.Topic) {
+				tFunc()
+				wg.Done()
+			}).Return().Times(1)
 
-			time.Sleep(1 * time.Second)
+			syncutils.WaitTimeout(&wg, waitTimeout)
```
lvrach, line 309: Is this safe, will leave things running in the background?

mocks/services/debugger/uploader.go
```diff
@@ -96,3 +96,15 @@ func (mr *MockUploaderIMockRecorder) Start() *gomock.Call {
...
+// Stop mocks base method.
+func (m *MockUploaderI) Stop() {
+	m.ctrl.T.Helper()
+	m.ctrl.Call(m, "Stop")
+}
+
+// Stop indicates an expected call of Stop.
+func (mr *MockUploaderIMockRecorder) Stop() *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Stop", reflect.TypeOf((*MockUploaderI)(nil).Stop))
```
mcchran, line 109: long line

mcchran:
Why rudder transformer is included in the change file list?

services/debugger/destination/eventDeliveryStatusUploader_test.go
```diff
@@ -212,6 +216,9 @@ var _ = Describe("eventDeliveryStatusUploader", func() {
 
 	Context("RecordEventDeliveryStatus", func() {
 		It("returns false if disableEventDeliveryStatusUploads is false", func() {
+			wg := sync.WaitGroup{}
+			wg.Add(1)
```
fracasula, line 220: A WaitGroup seems a bit overkilling for this, it looks like it could be done with a simple channel and a `<-time.After()`. No big deal though given that it's a test.

services/debugger/uploader.go
```diff
@@ -57,20 +61,31 @@ func New(url string, transformer Transformer) UploaderI {
...
+	uploader.bgWaitGroup.Add(1)
 	rruntime.Go(func() {
-		uploader.flushEvents()
+		uploader.flushEvents(ctx)
+		uploader.bgWaitGroup.Done()
 	})
 }
 
+func (upload *Uploader) Stop() {
+	close(upload.eventBatchChannel)
```
fracasula, line 85: This is going to panic if we call `Stop` more than once.

services/debugger/uploader.go
fracasula: I don't understand why nothing is done though. The context is done and we don't care and keep going. Same with batchTimeout which in this case looks more like a backoff interval rather than a timeout.

utils/sync/wait_with_timeout.go
```diff
@@ -0,0 +1,22 @@
...
+	go func() {
+		defer close(c)
+		wg.Wait()
+	}()
+	select {
+	case <-c:
+		return false // completed normally
+	case <-time.After(timeout):
+		return true // timed out
+	}
```
fracasula, line 21: There might be cases where the routine never returns, why not just using normal channels and/or contexts?
Also considering the package where this lives people might be keen to use it in production code as well, not just tests, and this should be avoided imho.

fracasula: This morning, after some consideration followed by reviewing this PR yesterday, I wrote this implementation: https://github.com/rudderlabs/rudder-server/pull/1845

It's an implementation that shouldn't leak, let me know your opinion. If you like the package I can remove it from draft and have it reviewed.

fracasula: I closed the PR and I vote for rewriting the whole `uploader` given that it is a small package it looks very doable :+1:

services/debugger/source/eventUploader_test.go
```diff
@@ -99,13 +102,18 @@ var _ = Describe("eventUploader", func() {
...
 
-			time.Sleep(3 * time.Second)
+			c.asyncHelper.WaitWithTimeout(5 * time.Second)
 			Expect(RecordEvent(sample_writeKey, recordingEvent)).To(BeFalse())
 		})
 
 		It("transforms payload properly", func() {
+			//wg := sync.WaitGroup{}
+			//wg.Add(1)
+
```
fracasula, line 116:
```suggestion
```

services/debugger/source/eventUploader_test.go
```diff
@@ -115,9 +123,12 @@ var _ = Describe("eventUploader", func() {
 					}()
 				})
 			tFunc := c.asyncHelper.ExpectAndNotifyCallback()
-			mockCall.Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) { tFunc() }).Return().Times(1)
+			mockCall.Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) {
+				tFunc()
+				//wg.Done()
```
fracasula, line 128:
```suggestion
```

services/debugger/destination/eventDeliveryStatusUploader_test.go
```diff
@@ -222,12 +224,13 @@ var _ = Describe("eventDeliveryStatusUploader", func() {
 					}()
 				})
 			tFunc := c.asyncHelper.ExpectAndNotifyCallback()
-			mockCall.Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) { tFunc() }).Return().Times(1)
+			mockCall.Do(func(channel chan pubsub.DataEvent, topic backendconfig.Topic) {
+				tFunc()
+			}).Return().Times(1)
 
-			time.Sleep(1 * time.Second)
+			c.asyncHelper.WaitWithTimeout(5 * time.Second)
```
fracasula, line 231: 5s for a unit tests seems like a very long timeout, can you give a bit more context as to why we would need 5s?

chrikar:
> Why rudder transformer is included in the change file list?

Nice catch @mcchran ! I'll amend that

services/debugger/uploader.go
fracasula: Why was this resolved? Shouldn't we return on `ctx.Done()` instead of flushing? WDYT?
Second thing, shouldn't we rename `batchTimeout` to `flushInterval` or something? It is not being used as a timeout, right?

services/debugger/uploader.go
```diff
@@ -57,25 +61,35 @@ func New(url string, transformer Transformer) UploaderI {
...
 	client := &http.Client{}
 
-	uploader := &Uploader{url: url, transformer: transformer, eventBatchChannel: eventBatchChannel, eventBuffer: eventBuffer, Client: client}
+	uploader := &Uploader{url: url, transformer: transformer, eventBatchChannel: eventBatchChannel, eventBuffer: eventBuffer, Client: client, bgWaitGroup: sync.WaitGroup{}}
 	uploader.Setup()
 	return uploader
 }
 
 func (uploader *Uploader) Start() {
+	ctx, cancel := context.WithCancel(context.Background())
```
fracasula, line 70: Can we potentially have `Start()` accept a `context.Context` as a parameter? I think it could play a very important role in the scope of this PR. WDYT?

chrikar: So that we can use the ctx to match the Start() with the Stop()?

services/debugger/destination/eventDeliveryStatusUploader_test.go
chrikar: All the tests in the service/debugger directory run in 2s locally for me. So this is mostly like a theoretical maximum. We could lower to 1 second if that helps.

services/debugger/source/eventUploader_test.go
chrikar: Oops :)

services/debugger/uploader.go
fracasula: We wouldn't even need `Stop()`, right? Plus it should make the whole thing easier to test.

services/debugger/uploader.go
```diff
@@ -57,25 +61,35 @@ func New(url string, transformer Transformer) UploaderI {
...
+
 	rruntime.Go(func() {
 		uploader.handleEvents()
+		cancel()
 	})
+
+	uploader.bgWaitGroup.Add(1)
 	rruntime.Go(func() {
```
fracasula, line 78: Why are we using `rruntime.Go` instead of simply doing `go`?

chrikar: Afaik the plan is to remove this in the near future. Also, this discussion is outside the scope of this PR, and should be discussed independently, so I'll resolve this for now

fracasula: Sure, I was just asking why it's being done this way, not asking for changing it here.
Although I must say that given that the plan is to remove them you could consider dropping this one now :slightly_smiling_face: 

Up to you :+1:

services/debugger/uploader.go
chrikar: 1. In case of ctx.Done() we fist want the events stored to be flashed and then return to exit the loop.
2. I like this suggestion. I will leave the configuration parameter we use to input the value for this as is for now though `Debugger.batchTimeoutInS`.

services/debugger/uploader.go
chrikar: The chain of events as it is now is: 
Stop() -> stop(channel) -> handleEvents() loop exit -> ctx cancel -> flushEvents trigger -> flush events (and return).
So, we need stop to stop the channel and trigger this.
Unless you mean that we can remove stop() altogether and use the cancel() in the test scope?

mocks/services/debugger/uploader.go
chrikar: Do we have a convention setup for this? I don't see anything in the guidelines.
Personally, I'm in favour of 125 characters a a limit for more modern display sizes and resolutions.

fracasula: In my editor (GoLand) I think it is set to 120 but yeah I agree that that doesn't seem like a long line.

services/debugger/uploader.go
fracasula: I meant that we can just pass `ctx` to `Start()` and use that to stop everything. If the context gets canceled the uploader stops everything. No need to `Stop()`, correct?

services/debugger/uploader.go
fracasula: > 1. In case of ctx.Done() we fist want the events stored to be flashed and then return to exit the loop.

Can you explain why please? Are we relying on the uploader in a way that if we don't upload the data is lost? If not and the context is canceled then we must honour it and return. WDYT?

lvrach: @fracasula Why is important for you to exit without flushing existing data?

flushing the data is not important, but it helps with testing and in ensures more consistent behaviour.

fracasula: @lvrach I think that is the expectation in every Go codebase. You pass a context, the context is canceled, all operations are terminated as soon as possible. Plus, this is an OpenSource project which makes conventions/standards even more important. To make an example the opposite would be as bad as having a method called `SendSync` which sends asynchronously instead. It says one thing but does another.

In any case, why would you cancel a context and keep operations running instead? Do you see an advantage on doing so?

services/debugger/uploader_test.go
```diff
@@ -173,22 +177,21 @@ var _ = Describe("Uploader", func() {
 
 					rawJSON, err := json.Marshal(data)
 					Expect(err).To(BeNil())
+					//wg.Done()
```
lvrach, line 180:
```suggestion
```

services/debugger/uploader.go
chrikar: As discussed we're going to follow this guideline https://www.rudderstack.com/blog/implementing-graceful-shutdown-in-go/

PR #1662, opened 2022-02-15
Add Timeout PostToWarehouse

cisse21:
**Fixes** # (*issue*)
https://rudderlabs.slack.com/archives/C01HTT66UMB/p1644907137267929
## Type of change
## Checklist:

### DIFF
--- a/router/batchrouter/batchrouter.go
+++ b/router/batchrouter/batchrouter.go
@@ -60,6 +60,7 @@ var (
 	readPerDestination                 bool
 	disableEgress                      bool
 	toAbortDestinationIDs              string
+	netClientTimeout                   time.Duration
 	transformerURL                     string
 	datePrefixOverride                 string
 	dateFormatLayouts                  map[string]string // string -> string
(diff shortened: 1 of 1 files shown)


achettyiitr (approved):
LGTM

router/batchrouter/batchrouter.go
```diff
@@ -60,6 +60,7 @@ var (
 	readPerDestination                 bool
 	disableEgress                      bool
 	toAbortDestinationIDs              string
+	netClientTimeout                   time.Duration
```
lvrach, line 63: Maybe we can name it httpTimeout to be consistent.

lvrach (approved):

router/batchrouter/batchrouter.go
```diff
@@ -2015,6 +2016,7 @@ func loadConfig() {
 	config.RegisterBoolConfigVariable(false, &disableEgress, false, "disableEgress")
 	config.RegisterBoolConfigVariable(true, &readPerDestination, false, "BatchRouter.readPerDestination")
 	config.RegisterStringConfigVariable("", &toAbortDestinationIDs, true, "BatchRouter.toAbortDestinationIDs")
+	config.RegisterDurationConfigVariable(10, &netClientTimeout, false, time.Second, "BatchRouter.httpTimeout")
```
lvrach, line 2019: Do we know the 95p time this requests needs ? Maybe 30 sec is safer.

cisse21: We don't have any stats around here I guess... Would it make sense to add one here? Also since this is a call within the same namespace, I thought we could have it at 10 seconds

router/batchrouter/batchrouter.go
cisse21: This is the same name we are using in router for time out, Do we change it at both the places?

Sidddddarth (approved):

PR #1670, opened 2022-02-17
Support for multiple hosts

cisse21:
**Fixes** # (*issue*)
https://www.notion.so/rudderstacks/Kafka-Config-Improvements-e6bf6890614549989a3be174027428c1
## Type of change
## Checklist:

### DIFF
--- a/services/streammanager/kafka/kafkamanager.go
+++ b/services/streammanager/kafka/kafkamanager.go
@@ -149,10 +149,12 @@ func NewProducer(destinationConfig interface{}, o Opts) (sarama.SyncProducer, er
 
 	if isSslEnabled {
 		caCertificate := destConfig.CACertificate
-		tlsConfig := NewTLSConfig(caCertificate)
-		if tlsConfig != nil {
-			config.Net.TLS.Config = tlsConfig
-			config.Net.TLS.Enable = true
+		config.Net.TLS.Enable = true
+		if caCertificate != "" {
+			tlsConfig := NewTLSConfig(caCertificate)
+			if tlsConfig != nil {
+				config.Net.TLS.Config = tlsConfig
+			}
 		}
 		if destConfig.UseSASL {
 			// SASL is enabled only with SSL
--- a/services/streammanager/kafka/kafkamanager.go
+++ b/services/streammanager/kafka/kafkamanager.go
@@ -130,6 +130,7 @@ func NewProducer(destinationConfig interface{}, o Opts) (sarama.SyncProducer, er
 
 	var destConfig = Config{}
 	jsonConfig, err := json.Marshal(destinationConfig)
+	hosts := make([]string, 0)
 	if err != nil {
 		return nil, fmt.Errorf("[Kafka] Error while marshaling destination Config %+v, with Error : %w", destinationConfig, err)
 	}
@@ -137,10 +138,11 @@ func NewProducer(destinationConfig interface{}, o Opts) (sarama.SyncProducer, er
 	if err != nil {
 		return nil, fmt.Errorf("[Kafka] Error while unmarshalling dest config :: %w", err)
 	}
-	hostName := destConfig.HostName + ":" + destConfig.Port
+	hostNames := strings.Split(destConfig.HostName, ",")
+	for _, hostName := range hostNames {
+		hosts = append(hosts, hostName+":"+destConfig.Port)
+	}
 	isSslEnabled := destConfig.SslEnabled
-
-	hosts := []string{hostName}
 
 	config := getDefaultConfiguration()
 	config.Producer.Timeout = o.Timeout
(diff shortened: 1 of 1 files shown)


services/streammanager/kafka/kafkamanager.go
```diff
@@ -130,27 +130,31 @@ func NewProducer(destinationConfig interface{}, o Opts) (sarama.SyncProducer, er
 
 	var destConfig = Config{}
 	jsonConfig, err := json.Marshal(destinationConfig)
+	hosts := make([]string, 0)
```
lvrach, line 133: [minor]
It would be preferable not to interrupt error checking with other code. 
```suggestion
```

services/streammanager/kafka/kafkamanager.go
```diff
@@ -130,27 +130,31 @@ func NewProducer(destinationConfig interface{}, o Opts) (sarama.SyncProducer, er
...
 	if err != nil {
 		return nil, fmt.Errorf("[Kafka] Error while marshaling destination Config %+v, with Error : %w", destinationConfig, err)
 	}
 	err = json.Unmarshal(jsonConfig, &destConfig)
 	if err != nil {
 		return nil, fmt.Errorf("[Kafka] Error while unmarshalling dest config :: %w", err)
 	}
-	hostName := destConfig.HostName + ":" + destConfig.Port
+	hostNames := strings.Split(destConfig.HostName, ",")
```
lvrach, line 141:
```suggestion
    hosts := make([]string, 0)
	hostNames := strings.Split(destConfig.HostName, ",")
```

services/streammanager/kafka/kafkamanager.go
```diff
@@ -130,27 +130,31 @@ func NewProducer(destinationConfig interface{}, o Opts) (sarama.SyncProducer, er
...
+	for _, hostName := range hostNames {
+		hosts = append(hosts, hostName+":"+destConfig.Port)
```
lvrach, line 143: What if someone wants to provide multiple hosts with different ports? We should consider supporting a configuration similar to [boostrap.servers](https://jaceklaskowski.gitbooks.io/apache-kafka/content/kafka-properties-bootstrap-servers.html)

services/streammanager/kafka/kafkamanager.go
```diff
@@ -307,7 +311,7 @@ func NewTLSConfig(caCertFile string) *tls.Config {
 	tlsConfig.RootCAs = caCertPool
 
 	//tlsConfig.BuildNameToCertificate()
-	tlsConfig.InsecureSkipVerify = true
+	tlsConfig.InsecureSkipVerify = false
```
lvrach, line 314: I remember having issues with some existing clients with expired certificates.
I believe this should be configurable even if the option is hidden from the UI. So that we have the option of applying a quick fix for clients with expired certificates and similar SSL issues. 

Especially since we don't have a way of notifying them, that their keys are expired.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -130,27 +130,31 @@ func NewProducer(destinationConfig interface{}, o Opts) (sarama.SyncProducer, er
...
 	config.Producer.Timeout = o.Timeout
 
 	if isSslEnabled {
 		caCertificate := destConfig.CACertificate
-		tlsConfig := NewTLSConfig(caCertificate)
-		if tlsConfig != nil {
-			config.Net.TLS.Config = tlsConfig
-			config.Net.TLS.Enable = true
+		config.Net.TLS.Enable = true
+		if caCertificate != "" {
```
lvrach, line 153: Wouldn't it be safer to have an explicit `destConfig.CustomCertificate` bool option instead of checking for empty string ?

services/streammanager/kafka/kafkamanager.go
cisse21: I spoke with @snarkychef  on this and he  suggested it would be rare that we would have a requirement of supporting a combination of host and portNames

services/streammanager/kafka/kafkamanager.go
cisse21: For quick fix , We could disable SSL which would mean it would not even use this function.This flag essentially skips verifying the certificate so I guess turning off the flag in the mean time would do no harm

services/streammanager/kafka/kafkamanager.go
cisse21: So you mean to say we should have another toggle on UI to use customer certificate?

services/streammanager/kafka/kafkamanager.go
lvrach: Not sure if it would be possible to connect to an SSL endpoint with SSL disabled on the client-side.

cisse21: Yeah makes sense

chandumlg (approved):

PR #1677, opened 2022-02-18
Fix Azure blob issue

lvrach:
# Changes

## Fix azure 400 issue
https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=1333fec0e54647cea5de5e851f8f3262

Following the official example, I used:
```
	serviceURL := azblob.NewServiceURL(*baseURL, p)
``` 
To build the URL. We are not sure regarding the root cause of the problem, but constructing the URL this way solved the `400` issue.

## Return code

Ensure we are returning the `m.Run()` code. Until now test suite was passing, even if some tests were failing inside the `filemanager` package.

 ## Skip GCS tests

Once the test suite was failing properly, we discovered that GCS test is failing. We've time-boxed some time to fix it, but unfortunately, we could. We tested with the actual destination and it works, so we decide to skip this test for now.

```
    skip:     "storage emulator is not stable",
```


saurav-malani (approved):
these changes seems to be fixing the azure blob issue. But, Azure emulator tests are failing. I will try to fix the test and push the PR.

cisse21 (approved):

PR #1703, opened 2022-02-25
isSuccessStatus->isJobTerminated

Sidddddarth:
Lets the router worker mark  a job as waiting only if they follow a `non-terminal` job for a user.
Previous behaviour was to mark any job following a `non-succeeding` job of a user as waiting.

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -871,6 +871,34 @@ func (worker *workerT) handleWorkerDestinationJobs(ctx context.Context) {
 		}
 	}
 
+	//if batching/routerTransform is enabled, we need to make sure that all the routerJobs status are written to DB.
+	//if in any case transformer doesn't send all the job ids back, setting their statuses as failed
+	for _, routerJob := range worker.routerJobs {
+		if _, ok := handledJobMetadatas[routerJob.JobMetadata.JobID]; !ok {
+			routerJobResponses = append(routerJobResponses, &RouterJobResponse{
+				jobID: routerJob.JobMetadata.JobID,
+				destinationJob: &types.DestinationJobT{
+					Destination: routerJob.Destination,
+					Message:     routerJob.Message,
+					JobMetadataArray: []types.JobMetadataT{
+						{
+							SourceID: routerJob.JobMetadata.SourceID,
+						},
+					},
+				},
+				destinationJobMetadata: &types.JobMetadataT{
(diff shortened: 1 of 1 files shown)


cisse21 (approved):

atzoum:
@Sidddddarth  do we have a state diagram describing all possible job states and their transitions?

Sidddddarth:
> @Sidddddarth do we have a state diagram describing all possible job states and their transitions?

There's no state diagram per se
but the basic idea is that 
- all events of a user that follow a failed event are marked `waiting`
- non-terminal(failed and waiting) jobs are picked up in the next `generatorLoop` and if a worker is found(which is where waiting jobs are ignored) for an event -> the event is passed to the worker for further routing
- and so and attempt is made to send the original failed job per user, the `waiting` jobs keep waiting till the failed job has reached a terminal state
- after which the waiting jobs are then sent to workers.

atzoum:
> > @Sidddddarth do we have a state diagram describing all possible job states and their transitions?
> 
> There's no state diagram per se but the basic idea is that
> 
> * all events of a user that follow a failed event are marked `waiting`
> * non-terminal(failed and waiting) jobs are picked up in the next `generatorLoop` and if a worker is found(which is where waiting jobs are ignored) for an event -> the event is passed to the worker for further routing
> * and so and attempt is made to send the original failed job per user, the `waiting` jobs keep waiting till the failed job has reached a terminal state
> * after which the waiting jobs are then sent to workers.

So all possible states are? `failed`, `waiting`...

atzoum:
[This ](https://github.com/rudderlabs/rudder-server/blob/master/jobsdb%2Fjobsdb.go#L470) is what I was looking for :)

Sidddddarth:
> So all possible states are? `failed`, `waiting`...

No, those are just some of the states we focused on while making this changes

> [This ](https://github.com/rudderlabs/rudder-server/blob/master/jobsdb%2Fjobsdb.go#L470)is what I was looking for :)

Understood
Those are all the states, yes
but the transition will need more

> Is it possible to showcase the correct (expected) behavior by adding a relevant scenario in router_test.go?
handleWorkerDestinationJobs is too complex (at least at first sight) to be able to reason about the change by merely reading it :)

You're right
I'll try and add some test cases around it

chandumlg (approved):

PR #1717, opened 2022-03-02
OOM kill fix processor

saurav-malani:
## Description of the change

>Till now, all the the jobs read from the DB where processed at a time. So, when we had enough jobs in gw table we were reading 10k jobs at a time. And, we are doing processing and transformation for all the jobs at a time, because of which memory usage was quite high. 
>But, here we are dividing read jobs (say 10k) into sub-batch of 5, and then processing each sub-batch at a time, instead of processing the complete batch at a time.
>Till now, since all the 10k jobs where processed at a time, there are a lot memory fluctuation. As, with slight increase in event size or number of transformations, all 10k jobs in memory where affected. But now, since we are processing only 2k jobs at a time. So, memory fluctuation has reduced significantly.

Screenshot 2022-03-10 at 11 34 24 PM

## Performance & Throughput Changes

RAM usage has come down to 33% (i.e. average usage before:8.17GB, now: 2.75GB) of what was being used before, without compromise in throughput.

>Resource Usage before:
image

>Resource Usage after changes:
image

>Grafana Resource usage Snapshot:
https://snapshots.raintank.io/dashboard/snapshot/vDDuNRQ28LqAlkZ8D4yZWHTKY8oUpkVl

>Throughput Before:
image

>Throughput After:
image

>Grafana Throughput Snapshot
https://snapshots.raintank.io/dashboard/snapshot/a2mM09f1dpMwlqBymOSJ017F1KifuXnx

## Notion Link (with details of all the experiments & environment)

> [Notion Link](https://www.notion.so/OOM-kill-Investigation-2bf973a1386c42f5b949543a715633b4#db99deb89d674c2f9f1ad1ee88d28bbe)


processor/processor.go
```diff
@@ -432,6 +432,7 @@ func (proc *HandleT) Shutdown() {
 var (
 	enablePipelining          bool
 	pipelineBufferedItems     int
+	subJobCount               int
```
atzoum, line 435: Why not splitting the 10K jobs into 10K subjobs of 1 job, instead of 5 subjobs of 2K jobs?

atzoum: To make the most efficient use of memory using asynchronous step processing through channels, each of the three channels should receive a singular item (message) instead of the full batch. Anything in between is going to be a compromise. Do you have any metrics to share with regards to how much the throughput decreases using singular items? Even better, would it be too hard for me to setup a test bench locally?
The structs that are being passed around through the channels appear to be complex indeed, holding stuff other than job messages, so I will need some time to study and try to explain why throughput decreases as you mentioned, however I believe it might be worth giving some effort towards this direction.

processor/processor.go
```diff
@@ -460,7 +461,8 @@ var (
 
 func loadConfig() {
 	config.RegisterBoolConfigVariable(true, &enablePipelining, false, "Processor.enablePipelining")
-	config.RegisterIntConfigVariable(1, &pipelineBufferedItems, false, 1, "Processor.pipelineBufferedItems")
+	config.RegisterIntConfigVariable(0, &pipelineBufferedItems, false, 1, "Processor.pipelineBufferedItems")
+	config.RegisterIntConfigVariable(5, &subJobCount, false, 1, "Processor.subJobCount")
```
atzoum, line 465: @saurav-malani  when using a very low `maxEventsToProcess`, e.g. 100, this would split events in 5 subJobs of 20 items each. Similarly, when using a very high  `maxEventsToProcess`, e.g. 100000, this would split events in 5 subJobs of 20000 items each. Based on the above and given the fact that you have found 2000 items per batch to be the most efficient number, should we use a configurable `subJobSize` (default 2000) and derive the `subJobCount` from it, instead of the opposite?

processor/processor.go
atzoum: It appears that transformer HTTP calls are being batched for same src+dest key messages during

1. the `process` step for performing trackingPlan validation 
2. the `transform` step for custom & destination-specific transformations

This can explain why throughput decreases drastically when splitting the batch in very small sub-batches.

processor/processor.go
```diff
@@ -2160,38 +2173,50 @@ func (proc *HandleT) pipelineWithPause(ctx context.Context, fn func(ctx context.
 	}
 }
 
+//Since mainpipeline func can process jobs in a single batch (when number of jobs are less)
+// and in multiple batches (when number of jobs are more). So, to identify if the received
+//job is a sub-part of a job is a single job. So, `isSplit` variable is needed to be passed
+//accross go routines.
+type subJobT struct {
+	subJobs []*jobsdb.JobT
+	isSplit bool
```
atzoum, line 2182: In case you follow the configurable `subJobSize` strategy, then you should replace `isSplit` with `subJobCount` here

processor/processor.go
```diff
@@ -2230,22 +2279,92 @@ func (proc *HandleT) mainPipeline(ctx context.Context) {
...
 		defer close(chStore)
-
+		defer close(triggerStore)
 		for msg := range chTrans {
+
 			chStore <- proc.transformations(msg)
+			//since we don't want to call DB write for each sub-jobs. That's why we are waiting until all the sub-jobs are processed.
+			//Once all the sub-jobs are processed, we trigger the DB write by `triggerStore` channel.
+			if len(chStore) == subJobCount || !msg.isSplit {
+				triggerStore <- 1
```
atzoum, line 2296: In case you follow the configurable `subJobSize` strategy, then you could push here the calculated `subJobCount`, instead of the literal value `1`

processor/processor.go
saurav-malani: makes sense. I assumed that `maxEventsToProcess` is 10k always. But, since it's configurable, this is not a correct assumption. Thanks for pointing it out.

atzoum: here is some sample code for splitting

```
type subJob struct {
	jobs  []*jobsdb.JobT
	index int
	size  int
}

func splitJobs(jobs []*jobsdb.JobT, batchSize int) []subJob {
	res := []subJob{}
	jobsSize := len(jobs)
	if batchSize >= jobsSize {
		res = append(res, subJob{jobs, 0, 1})
		return res
	}
	subJobCount := jobsSize / batchSize
	if jobsSize%batchSize > 0 {
		subJobCount++
	}
	for i := 0; i < subJobCount; i++ {
		startIdx := i * batchSize
		endIdx := (i + 1) * batchSize
		if endIdx > jobsSize {
			endIdx = jobsSize
		}
		res = append(res, subJob{jobs[startIdx:endIdx], 0, 1})
	}
	return res
}
```

and the relevant test for possible scenarios

```
type subJob struct {
	jobs  []*jobsdb.JobT
	index int
	size  int
}

func splitJobs(jobs []*jobsdb.JobT, batchSize int) []subJob {
	res := []subJob{}
	jobsSize := len(jobs)
	if batchSize >= jobsSize {
		res = append(res, subJob{jobs, 0, 1})
		return res
	}
	subJobCount := jobsSize / batchSize
	if jobsSize%batchSize > 0 {
		subJobCount++
	}
	for i := 0; i < subJobCount; i++ {
		startIdx := i * batchSize
		endIdx := (i + 1) * batchSize
		if endIdx > jobsSize {
			endIdx = jobsSize
		}
		res = append(res, subJob{jobs[startIdx:endIdx], 0, 1})
	}
	return res
}
```

processor/processor.go
```diff
@@ -2230,22 +2281,91 @@ func (proc *HandleT) mainPipeline(ctx context.Context) {
 		}
 	}()
 
-	chStore := make(chan storeMessage, bufferSize)
+	triggerStore := make(chan int, 1)
+	chStore := make(chan storeMessage, subJobCount)
```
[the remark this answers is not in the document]
saurav-malani: done

processor/processor.go
saurav-malani: we are capturing that in the `msg.subJobCount`. I have changed `<-1` to `<-true`.

saurav-malani: the reason I have included `subJobCount` in the struct, is to maintain consistency across all the the go routine in the mainpipeline

processor/processor.go
saurav-malani: done

processor/processor.go
saurav-malani: yup

processor/processor.go
```diff
@@ -2230,28 +2279,67 @@ func (proc *HandleT) mainPipeline(ctx context.Context) {
 		}
 	}()
 
-	chStore := make(chan storeMessage, bufferSize)
+	chStore := make(chan storeMessage, maxEventsToProcess/subJobSize)
```
lvrach, line 2282: Do we still need this?

processor/processor.go
saurav-malani: I have created a `splitter` func to convert jobs into subJobs.

processor/processor.go
saurav-malani: yup, not needed anymore.

saurav-malani: converting it back to `bufferSize`

processor/processor.go
```diff
@@ -2160,38 +2172,75 @@ func (proc *HandleT) pipelineWithPause(ctx context.Context, fn func(ctx context.
...
 			case <-time.After(nextSleepTime):
+
 				if !isUnLocked {
 					nextSleepTime = proc.maxLoopSleep
 					continue
 				}
-
 				jobs := proc.getJobs()
 				if len(jobs) == 0 {
-					// no jobs found, double sleep time until maxLoopSleep
```
lvrach, line 2194: Why did you remove this comment?

processor/processor.go
```diff
@@ -2160,35 +2172,73 @@ func (proc *HandleT) pipelineWithPause(ctx context.Context, fn func(ctx context.
...
+//Since mainpipeline func can process jobs in a single batch (when number of jobs are less)
+// and in multiple batches (when number of jobs are more). So, to identify if the received
+//job is a sub-part of a job is a single job. So, `isSplit` variable is needed to be passed
+//accross go routines.
+type subJobT struct {
+	subJobs  []*jobsdb.JobT
+	haveMore bool
+}
+
+func splitJob(jobs []*jobsdb.JobT) []subJobT {
```
atzoum, line 2184: @saurav-malani splitJobs function may not respect the `subJobSize` for the last page if the `size` of the jobs is not a multiple of the `subJobSize`.

E.g. if `jobSize` is 3999 and `subJobSize` is 2000, `splitJob` will produce an array with a single element of size 3999, instead of two elements, one with 2000 and another with 1999.

Please consider consulting my example here and also make sure to add the relevant tests that I have included:
https://github.com/rudderlabs/rudder-server/pull/1717#discussion_r823629433

processor/processor.go
```diff
@@ -2200,22 +2250,22 @@ func (proc *HandleT) mainPipeline(ctx context.Context) {
...
 				if err != nil {
 					pkgLogger.Error(err)
 					panic(err)
 				}
-
 				events := 0
 				for i := range jobs {
 					events += jobs[i].EventCount
 				}
-				// nextSleepTime is dependent on the number of events read in this loop
```
atzoum, line 2214: please revive this deleted comment

processor/processor.go
```diff
@@ -2232,26 +2282,65 @@ func (proc *HandleT) mainPipeline(ctx context.Context) {
...
 	wg.Add(1)
 	go func() {
+		var mergedJob storeMessage
+		firstSubJob := true
 		defer wg.Done()
-		for msg := range chStore {
-			proc.Store(msg)
+		for subJob := range chStore {
+
+			if firstSubJob {
```
atzoum, line 2301: to skip creating duplicate structs in case of unsplitted messages, you could add the following fragment above:

```
if firstSubJob && !subJob.haveMore {
	proc.Store(subJob)
	continue
}
```

atzoum (approved):
LGTM, good job Saurav!

processor/processor.go
```diff
@@ -2160,23 +2190,64 @@ func (proc *HandleT) pipelineWithPause(ctx context.Context, fn func(ctx context.
 	}
 }
 
+//`jobSplitter` func Splits the read Jobs into sub-batches after reading from DB to process.
+//`subJobMerger` func merges the split jobs into a single batch before writing to DB.
+//So, to keep track of sub-batch we have `hasMore` variable.
+//each sub-batch has `hasMore`. If, a sub-batch is the last one from the batch it's marked as `false`, else `true`.
+type subJobT struct {
```
lvrach, line 2197: lets avoid the `T` suffix

```suggestion
type subJob struct {
```

processor/processor.go
```diff
@@ -2160,23 +2190,64 @@ func (proc *HandleT) pipelineWithPause(ctx context.Context, fn func(ctx context.
...
+}
+
 // mainPipeline: new way of handling jobs
 //
 // [getJobs] -chProc-> [processJobsForDest] -chTrans-> [transformations] -chStore-> [Store]
 func (proc *HandleT) mainPipeline(ctx context.Context) {
 	//waiting for reporting client setup
+	proc.logger.Info("Processor mainPipeline started")
+	proc.logger.Info("Processor subJobSize= ", subJobSize)
+	proc.logger.Info("Processor pipelineBufferedItems= ", pipelineBufferedItems)
```
lvrach, line 2237: [minor] For more concise logs:

```suggestion
	proc.logger.Info("Processor mainPipeline started")
	proc.logger.Info("Processor subJobSize= ", subJobSize)
	proc.logger.Info("Processor pipelineBufferedItems= ", pipelineBufferedItems)
```

lvrach (approved):
LGTM! minor comments only. 

Thanks for addressing the comments mentioned before.

processor/processor.go
```diff
@@ -365,6 +369,11 @@ func (proc *HandleT) Setup(backendConfig backendconfig.BackendConfig, gatewayDB
 	proc.statBatchDestNumOutputEvents = proc.stats.NewTaggedStat("processor.num_output_events", stats.CountType, stats.Tags{
 		"module": "batch_router",
 	})
+	proc.DBReadThroughput = proc.stats.NewStat("processor.db_read_throughput", stats.CountType)
```
atzoum, line 372: Shouldn't all throughputs be calculated as `Gauge`s instead of `Counter`s?
What's the point of having a counter increased by a value (throughput) every time we calculate this value?

atzoum: any comment @saurav-malani ?

chandumlg (approved):

PR #1736, opened 2022-03-10
Capture pending-events metrics and publish to StatsD/InfluxDB

Sidddddarth:
## Description of the change

- Introduced `metric` package for capturing metrics in rudder-server independently from the `stats` package. 3 different metric types are currently supported by `metric.Registry`: `Counter`, `Gauge` & `MovingAverage`. Using prometheus' implementation for `Gauge` & `Counter`, kept the existing ewma implementation for `MovingAverage`.
- `jobsdb_rt_pending_events_count` & `jobsdb_batch_rt_pending_events_count` metrics are being captured in `metric.Registry` by `multitenant` and picked periodically by `stats` to be published in StatsD.
- `JobsDB`s of `rt` & `batch_rt` tables expose their `pending-events` through rudder-cli.

## Further work
The following metrics are now calculated (mainly) by the `multitenant` package:
1. `Gauge` of number of unprocessed events. Dimensions: `dbPrefix={rt,batch_rt}`, `destType`, `workspaceId` 
1. `Gauge` of the last drain occurrence. Dimensions: `dbPrefix=rt`, `destType`, `workspaceId` 
2. `MovingAverage` of events' input rate Dimensions: `dbPrefix={rt,batch_rt}`, `destType`, `workspaceId` 
4. `MovingAverage` of events' (success) output rate. Dimensions: `dbPrefix=rt`, `destType`, `workspaceId` 
5. `MovingAverage` of events' failure rate. Dimensions: `dbPrefix=rt`, `destType`, `workspaceId`

It is proposed to decouple the above metrics' calculation from the `multitenant` package and transfer the responsibility of performing such calculations to `jobsdb` package.
The following `JobsDB` methods should be responsible for updating the aforementioned metrics, whenever called:

```
Store(jobList []*JobT) error
storeWithRetryEach(jobList []*JobT) map[uuid.UUID]string
UpdateJobStatus(statusList []*JobStatusT, customValFilters []string, parameterFilters []ParameterFilterT) error
UpdateJobStatusInTxn(txHandler *sql.Tx, statusList []*JobStatusT, customValFilters []string, parameterFilters []ParameterFilterT) error
```

**After this change:**
1. `MultiTenant` interface could be simplified to a single method:
```
type MultiTenantI interface {
	GetRouterPickupJobs(destType string, noOfWorkers int, routerTimeOut time.Duration, jobQueryBatchSize int, timeGained float64) (map[string]int, map[string]float64)
}
```
2. The `processor` would no longer need to depend on `multitenant`.
3. All jobsDBs will have their metrics available and published to InfluxDB, not only `pending_events` metric for `rt` & `batch_rt`.

## Notion Link

> https://www.notion.so/rudderstacks/Enable-collection-of-multi-tenant-in-memory-stats-all-the-time-19668e935ae4406cb082e02c89989a8a
> https://www.notion.so/rudderstacks/add-stats-for-batch_rt-similar-to-what-we-have-for-rt-in-case-of-multi-tenant-0578c916013d47b6a5f45674b36cb060
> https://www.notion.so/rudderstacks/Report-collected-in-memory-stats-to-influxDB-8fb6645415944e95b773965385765425

## Type of change

## Checklist:

## Security


services/multitenant/tenantstats.go
```diff
@@ -58,19 +68,59 @@ func NewStats(routerDB jobsdb.MultiTenantJobsDB) *MultitenantStatsT {
...
+								"workspace":   workspace,
+								"destType":    destType,
+								"tablePrefix": tableType,
+							})
+						pendingEventCountStat.Gauge(destTypeCount)
+					}
+				}
+			}
+		}()
+	}
```
lvrach, line 114: Could we decouple stats keeping functionality from sending metrics to a different component as we discussed?

services/multitenant/tenantstats.go
```diff
@@ -58,19 +68,59 @@ func NewStats(routerDB jobsdb.MultiTenantJobsDB) *MultitenantStatsT {
...
 	multitenantStat.lastDrainedTimestamps = make(map[string]map[string]time.Time)
 	multitenantStat.failureRate = make(map[string]map[string]misc.MovingAverage)
-	multitenantStat.routerTenantLatencyStat = make(map[string]map[string]misc.MovingAverage)
-	multitenantStat.processorStageTime = time.Now()
 	pileUpStatMap := make(map[string]map[string]int)
-	routerDB.GetPileUpCounts(pileUpStatMap)
-	for workspace := range pileUpStatMap {
-		for destType := range pileUpStatMap[workspace] {
-			multitenantStat.AddToInMemoryCount(workspace, destType, pileUpStatMap[workspace][destType], "router")
+	for routerType := range routerDBs {
```
cisse21, line 72: Should probably have destType as a key even in batch router stats

cisse21 (changes requested):

services/multitenant/tenantstats.go
Sidddddarth: put it in processor

processor/processor.go
```diff
@@ -117,6 +117,7 @@ type HandleT struct {
 	multitenantI                   multitenant.MultiTenantI
 	backgroundWait                 func() error
 	backgroundCancel               context.CancelFunc
+	atomicAdd                      uint64
```
atzoum, line 120: I understand that this is here only for debugging purposes. If so, please make sure to cleanup afterwards

processor/processor.go
```diff
@@ -284,6 +285,8 @@ func (proc *HandleT) Status() interface{} {
 		proc.dedupHandler.PrintHistogram()
 	}
 
+	statusRes["pending-events-count"] = append(statusRes["pending-events-count"], proc.multitenantI.Status())
```
atzoum, line 288: Should the processor report in its own status the number of pending events in router & batch router queues?

services/multitenant/tenantstats.go
```diff
@@ -36,6 +73,7 @@ type MultiTenantI interface {
 	RemoveFromInMemoryCount(workspaceID string, destinationType string, count int, tableType string)
 	ReportProcLoopAddStats(stats map[string]map[string]int, tableType string)
 	UpdateWorkspaceLatencyMap(destType string, workspaceID string, val float64)
+	Status() map[string]map[string]map[string]int
```
atzoum, line 40: Looks like this map is calling for a composite key, since we are not really interested in actually accessing intermediate levels of the nested maps.
E.g.
```
type MultiTenantStatusKey struct {
	Dataset         string
	Workspace       string
	DestinationType string
}
...
Status() map[Key]int {

```

processor/processor.go
Sidddddarth: yeah
this is a placeholder for now(put it here because the metric call is made from the processor as well)
I'll move it to the respective queues.

services/multitenant/channelCounter.go
```diff
@@ -0,0 +1,52 @@
...
+	counter := &Counter{make(chan func(), 256), make(map[string]map[string]int)}
+	go func(counter *Counter) {
+		for f := range counter.ch {
+			f()
+		}
+	}(counter)
+	return counter
+}
+
+func (c *Counter) Add(workspaceID, dest string, num int) {
```
atzoum, line 18: minor: `Increment`/`Decrement` appear to be more appropriate words for counters

services/multitenant/channelCounter.go
```diff
@@ -0,0 +1,52 @@
+package multitenant
+
+type Counter struct {
+	ch                    chan func()
```
atzoum, line 4: is the purpose of this channel to help with concurrent map access? If so, should we use a `sync.RWMutex` instead as proposed [here](https://go.dev/blog/maps#concurrency)?

Sidddddarth: it is for concurrent map access, as well as to maintain the order of that access. I read a bit about "mutex fairness" and it didn't seem like ordering was maintained.
Whereas with this, I'm fairly certain the order of access is maintained.

atzoum: what is the concern?

Sidddddarth: one concern is that too many goroutines racing for the mutex around (de)incrementing the count. This is slowing down the processes on those goroutines(processor storing to rt, brt queues).

I suspect this is somehow leading to subtract(terminal counts) before we even add them to pending count.
Hence an attempt to use channels. I'll give the atomic operation(prometheus) you suggested and look for improvements.

atzoum: Incrementing & decrementing an in-memory counter shouldn't be considered to be a cause of delays regardless the number of competing writers. I believe that it is more important here to have readable code and consistent behaviour, i.e. ensure that the counter is safe for concurrent access.

Sidddddarth: The counter is safe for concurrent access right?
while the +/- operation itself isn't delaying, wouldn't taking locks be a cause for that..?

atzoum: > The counter is safe for concurrent access right?

which counter?

Sidddddarth: the map of counts(`workspaceDestCountMap`) in `Counter` struct.

atzoum: yes, looks to be safe. However it seems to be a complex pattern to reuse.

services/multitenant/tenantstats.go
```diff
@@ -113,95 +116,50 @@ func (multitenantStat *MultitenantStatsT) CalculateSuccessFailureCounts(workspac
...
+					multitenantStat.routerInputRates[dbPrefix][workspaceKey][destType].Add(0)
+				}
 			}
 		}
 	}
 	multitenantStat.processorStageTime = time.Now()
 }
 
-func (multitenantStat *MultitenantStatsT) GetRouterPickupJobs(destType string, noOfWorkers int, routerTimeOut time.Duration, jobQueryBatchSize int, timeGained float64) (map[string]int, map[string]float64) {
+func (multitenantStat *MultitenantStatsT) GetRouterPickupJobs(dbPrefix string, noOfWorkers int, routerTimeOut time.Duration, jobQueryBatchSize int, timeGained float64) (map[string]int, map[string]float64) {
```
Sidddddarth, line 156: I think this remains destType
dbPrefix is "rt" for this function anyway - because this is called for router only.
May need to have dbPrefix and destType both if both prefixes call `GetRouterPickupJobs`

services/metric/manager.go
```diff
@@ -0,0 +1,76 @@
...
+func (r *manager) createRegistry(key string) {
+	r.registries[key] = NewRegistry()
+}
+
+func (r *manager) GetRegistry(key string) Registry {
+	return r.registries[key]
+}
+
+func GetPendingEventsMeasurement(tablePrefix string, workspace string, destType string) Gauge {
+	return GetManager().GetRegistry(PENDING_EVENTS_REGISTRY).MustGetGauge(&pendingEventsMeasurement{tablePrefix, workspace, destType})
```
Sidddddarth, line 55: a more general doubt: would using pointers as a key in the map yield expected results..?
`&pendingEventsMeasurement{tablePrefix, workspace, destType}`

services/multitenant/tenantstats.go
atzoum: Correct, I messed this one up. Thanks for spotting it! I will revert and make sure this didn't 
cascade anywhere else.

services/metric/manager.go
atzoum: the Registry accepts an interface as an argument, thus the pointer. It is true that we would be in trouble if using pointers as map keys, thus we use the value returned from the `GetKey` method as a key, which is the actual struct. 
I admit that this is not the simplest/cleanest solution right now, we can further discuss to find a better design

Sidddddarth: oh right. I misread the return in `GetKey`.
This looks alright

services/metric/registry.go
```diff
@@ -0,0 +1,232 @@
...
+	r.store.Range(f)
+}
+
+func (r *registry) GetMetricsByName(name string) []TagsWithValue {
+	metricsSet, ok := r.nameIndex.Load(name)
+	if !ok {
+		return []TagsWithValue{}
+	}
+	values := []TagsWithValue{}
+	for _, value := range metricsSet.(map[Measurement]TagsWithValue) {
```
cisse21, line 217: Shouldn't we be checking the success of typecast? Like we do for the above functions

cisse21 (changes requested):

cisse21:
@atzoum  Can we also probably migrate all the tenantStats to the new modules? Also I think we should be doing some performance/benchmark testing to ensure that the performance is not degraded with these newer modules before we go ahead and make any more changes

services/metric/registry.go
atzoum: > Shouldn't we be checking the success of typecast? Like we do for the above functions

Not really since the `Registry` has full ownership of what gets inserted in the `nameIndex` map. Only values generated by the `indexGenerator` can be added to the `nameIndex` which are of the appropriate type.
In the future, as soon as generics make their way into the `sync` package we will be able to have `sync.Map[K, V]` and this and other similar casts will become redundant :)

atzoum:
> @atzoum Can we also probably migrate all the tenantStats to the new modules? Also I think we should be doing some performance/benchmark testing to ensure that the performance is not degraded with these newer modules before we go ahead and make any more changes

Yes that's the idea. I was also thinking something more radical than that, i.e. to fully decouple `tenantStats` from metrics collection, since these metrics are valuable outside the context of multitenant too. `tenantStats` should only be responsible for implementing the pickup algorithm, still relying on these metrics. But JobsDB appears to be a more appropriate component for capturing these metrics since all write operations happen through it after all.
The middle-ground solution would be to keep things as is and merely register all `tenantStats` metrics in `metric.Registry`. The only win with this approach would be that extra metrics will become available in InfluxDB. 

Regarding performance, I don't expect any surprises there since:
1. We are using the `Counter` & `Gauge` implementations of prometheus which are highly optimized already.
2. We are using a `sync.Map` to address all other goroutine coordination concerns, which is something highly optimized for our use case, where the entry for a given key is only ever written once but read many times.

Sidddddarth:
> > @atzoum Can we also probably migrate all the tenantStats to the new modules? Also I think we should be doing some performance/benchmark testing to ensure that the performance is not degraded with these newer modules before we go ahead and make any more changes
> 
> Yes that's the idea. I was also thinking something more radical than that, i.e. to fully decouple `tenantStats` from metrics collection, since these metrics are valuable outside the context of multitenant too. `tenantStats` should only be responsible for implementing the pickup algorithm, still relying on these metrics. JobsDB appears to be a more appropriate component for capturing these metrics since all write operations happen through it after all. The middle-ground solution would be to keep things as is and merely register all `tenantStats` metrics in `metric.Registry`. The only win with this approach would be that extra metrics will become available in InfluxDB.
> 
> Regarding performance, I don't expect any surprises there since:
> 
>     1. We are using the `Counter` & `Gauge` implementations of prometheus which are highly optimized already.
> 
>     2. We are using a `sync.Map` to address all other goroutine coordination concerns, which is something highly optimized for our use case, where the entry for a given key is only ever written once but read many times.

On a tangent: might as well link the jobsdb cache here(counter=0 means `no jobs` otherwise `has jobs`)

cisse21:
`https://snapshots-origin.grafana.net/dashboard/snapshot/RHYerFDQrMESkHnYZFC0gB99hiJqNdVK : Master
`
`https://snapshots-origin.grafana.net/dashboard/snapshot/tJgQpTYOQd8KrBRS6XMBBLsigNNe5L9N?orgId=2&refresh=10s : Metrics PR`

Have done a preliminary analysis with this PR and the current Master, The performance seems to be a tad bit downgraded after the changes to the metrics. However this test setup has one source and one destination(`WEBHOOK`) for 200 workspaces , need to verify if this performance would remain consistent even in the case where we have multiple routers running.

atzoum:
Here are benchmark results comparing:
- Accessing a `metric.Registry` and incrementing a `metric.Counter`
- Accessing a `map[string]int` using a mutex lock and incrementing the counter

The results indicate that as concurrency levels and CPUs increase, the metric registry starts performing better. It performs worse though when there is low concurrency & CPUs.

```
BenchmarkRegistryGetCounterAndInc1              12117880                86.83 ns/op
BenchmarkRegistryGetCounterAndInc1-2            13755355                85.89 ns/op
BenchmarkRegistryGetCounterAndInc1-4            14091225                84.94 ns/op
BenchmarkRegistryGetCounterAndInc1-6            13658461                88.03 ns/op
BenchmarkRegistryGetCounterAndInc1-8            13574800                88.67 ns/op
BenchmarkRegistryGetCounterAndInc10             13763809                87.33 ns/op
BenchmarkRegistryGetCounterAndInc10-2           18780444                63.12 ns/op
BenchmarkRegistryGetCounterAndInc10-4           24814129                43.47 ns/op
BenchmarkRegistryGetCounterAndInc10-6           16493910                70.92 ns/op
BenchmarkRegistryGetCounterAndInc10-8           18177583                65.92 ns/op
BenchmarkRegistryGetCounterAndInc100            13609389                87.34 ns/op
BenchmarkRegistryGetCounterAndInc100-2          18413791                63.96 ns/op
BenchmarkRegistryGetCounterAndInc100-4          25806867                43.89 ns/op
BenchmarkRegistryGetCounterAndInc100-6          16953501                71.90 ns/op
BenchmarkRegistryGetCounterAndInc100-8          17984386                66.68 ns/op
BenchmarkMutexMapGetIntAndInc1                  58581950                18.45 ns/op
BenchmarkMutexMapGetIntAndInc1-2                64421314                18.46 ns/op
BenchmarkMutexMapGetIntAndInc1-4                64513960                18.37 ns/op
BenchmarkMutexMapGetIntAndInc1-6                64521477                18.38 ns/op
BenchmarkMutexMapGetIntAndInc1-8                64537814                18.43 ns/op
BenchmarkMutexMapGetIntAndInc10                 65235710                29.72 ns/op
BenchmarkMutexMapGetIntAndInc10-2                7396996               176.7 ns/op
BenchmarkMutexMapGetIntAndInc10-4               10374482               124.8 ns/op
BenchmarkMutexMapGetIntAndInc10-6                9382900               126.9 ns/op
BenchmarkMutexMapGetIntAndInc10-8                9347695               127.3 ns/op
BenchmarkMutexMapGetIntAndInc100                57121887                18.51 ns/op
BenchmarkMutexMapGetIntAndInc100-2               6345979               194.6 ns/op
BenchmarkMutexMapGetIntAndInc100-4               8331866               155.5 ns/op
BenchmarkMutexMapGetIntAndInc100-6               8490924               141.2 ns/op
BenchmarkMutexMapGetIntAndInc100-8               8469376               139.5 ns/op
```

atzoum:
Just performed some tests using my dev environment with 1 workspace & 1 destination, sending 1M events in 30 minutes and couldn't spot any difference in performance between `master` and this pull request:
Screenshot 2022-03-29 at 12 09 51 PM

atzoum (approved):

jobsdb/jobsdb.go
```diff
@@ -460,6 +461,21 @@ func (jd *HandleT) Status() interface{} {
 	if db.IsValidMigrationMode(jd.migrationState.migrationMode) {
 		statusObj["migration-state"] = jd.migrationState
 	}
+	pendingEventMetrics := metric.GetManager().GetRegistry(metric.PUBLISHED_METRICS).GetMetricsByName(fmt.Sprintf(metric.JOBSDB_PENDING_EVENTS_COUNT, jd.tablePrefix))
```
lvrach, line 464: [minor] avoiding long lines:

```suggestion
	pendingEventMetrics := metric.GetManager()
	 .GetRegistry(metric.PUBLISHED_METRICS)
	 .GetMetricsByName(fmt.Sprintf(metric.JOBSDB_PENDING_EVENTS_COUNT, jd.tablePrefix))
```

jobsdb/jobsdb.go
```diff
@@ -460,6 +461,21 @@ func (jd *HandleT) Status() interface{} {
...
+				pendingEvents = append(pendingEvents, map[string]interface{}{
+					"tags":  pendingEvent.Tags,
+					"count": count,
+				})
+			}
+		}
+		statusObj["pending-events"] = pendingEvents
+	}
+
 	return statusObj
```
lvrach, line 479: [minor]
```suggestion
	if len(pendingEventMetrics) == 0 {
		return statusObj
	}
	pendingEvents := []map[string]interface{}{}
	for _, pendingEvent := range pendingEventMetrics {
		count := pendingEvent.Value.(metric.Gauge).IntValue()
		if count > 0 {
			pendingEvents = append(pendingEvents, map[string]interface{}{
				"tags":  pendingEvent.Tags,
				"count": count,
			})
		}
	}
	statusObj["pending-events"] = pendingEvents
	

	return statusObj
```

router/router.go
```diff
@@ -1541,31 +1534,31 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
...
+		case jobsdb.Succeeded.State:
+			routerWorkspaceJobStatusCount[workspaceID] += 1
+			sd.Count++
+			rt.MultitenantI.CalculateSuccessFailureCounts(workspaceID, rt.destName, true, false)
+		case jobsdb.Aborted.State:
+			routerWorkspaceJobStatusCount[workspaceID] += 1
+			sd.Count++
+			rt.MultitenantI.CalculateSuccessFailureCounts(workspaceID, rt.destName, false, true)
+			routerAbortedJobs = append(routerAbortedJobs, resp.JobT)
+			PrepareJobRunIdAbortedEventsMap(resp.JobT.Parameters, jobRunIDAbortedEventsMap)
```
lvrach, line 1555: 👏

router/router.go
```diff
@@ -1604,11 +1597,11 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
 	}
 	//REPORTING - ROUTER - END
 
-	for workspace := range routerWorkspaceJobStatusCount {
-		for destType := range routerWorkspaceJobStatusCount[workspace] {
-			rt.MultitenantI.RemoveFromInMemoryCount(workspace, destType, routerWorkspaceJobStatusCount[workspace][destType], "router")
+	defer func() {
+		for workspace := range routerWorkspaceJobStatusCount {
+			metric.GetPendingEventsMeasurement("rt", workspace, rt.destName).Sub(float64(routerWorkspaceJobStatusCount[workspace]))
```
lvrach, line 1602: [minor] 
```suggestion
			metric.GetPendingEventsMeasurement("rt", workspace, rt.destName)
			    .Sub(float64(routerWorkspaceJobStatusCount[workspace]))
```

lvrach:
> performing such calculations to jobsdb package.
> The following JobsDB methods should be responsible for updating the aforementioned metrics
I am not sure about directly injecting things to JobsDB methods, I would prefer a wrapper/decorator pattern.

lvrach:
Although we should eventually do it, I am a bit concern having `workspaceId` as a dimension. We should be careful when deploying this to hosted, the number of `timeseries` will explode.

chandumlg:
https://snapshots-origin.grafana.net/dashboard/snapshot/UHOK423o2e7SSi8qh2Sf2XHd4Wh1MX6W?orgId=2

Performance on dev environment look okay.

cisse21 (approved):

lvrach:
@chandumlg did the performance tests included multiple workspace IDs?

cisse21:
@lvrach  It involved 200 Workspace ID's with one destination each. The difference with my previous experiment was that while in my experiment there were 500's too, here it is always 200

jobsdb/jobsdb.go
```diff
@@ -464,19 +464,21 @@ func (jd *HandleT) Status() interface{} {
...
-		for _, pendingEvent := range pendingEventMetrics {
-			count := pendingEvent.Value.(metric.Gauge).IntValue()
-			if count > 0 {
-				pendingEvents = append(pendingEvents, map[string]interface{}{
-					"tags":  pendingEvent.Tags,
-					"count": count,
-				})
-			}
+	if len(pendingEventMetrics) == 0 {
+		return statusObj
```
atzoum, line 469: @chandumlg the initial idea was to return an empty array if no pending events where present

lvrach (approved):

router/router.go
atzoum: Kudos to @Sidddddarth

PR #1744, opened 2022-03-14
handling transformer response unmarshal error

chandumlg:
## Description of the change

Handle incorrect format of transformer response instead of panicking.

## Notion Link

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=212368cf19d844baa75552685b8f4ca8

## Type of change

## Checklist:

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -633,7 +633,7 @@ func getIterableStruct(payload []byte, transformAt string) ([]integrations.PostP
 		}
 	}
 
-	return responseArray, fmt.Errorf("transformer response unmarshal error: %w", err)
+	return responseArray, err
 }
 
 func (worker *workerT) handleWorkerDestinationJobs(ctx context.Context) {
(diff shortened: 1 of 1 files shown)


cisse21 (approved):

router/router.go
```diff
@@ -752,52 +749,56 @@ func (worker *workerT) handleWorkerDestinationJobs(ctx context.Context) {
...
-								resp := worker.rt.netHandle.SendPost(sendCtx, val)
-								respStatusCode, respBodyTemp, respContentType = resp.StatusCode, string(resp.ResponseBody), resp.ResponseContentType
-								// stat end
-								worker.routerDeliveryLatencyStat.SendTiming(time.Since(rdl_time))
-							}
-							// transformer proxy end
-							if isSuccessStatus(respStatusCode) {
+					result, err := getIterableStruct(destinationJob.Message, transformAt)
+					if err != nil {
+						respStatusCode, respBody = 400, fmt.Errorf("transformer response unmarshal error: %w", err).Error()
```
atzoum, line 754: Failure to process the server's response is actually a client processing error, however we are masking it as a server error `400` response. This appears to be a pattern that we are already using, so it is acceptable.
Nevertheless, would it be valuable to discriminate client processing errors by using some special status code?

The following link provides all assigned (& unassigned) status codes:
https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
Ranges 419-420, 432-450 & 452-499 are unassigned. I would pick 499 which might be the last status code we should expect to become assigned to some special meaning.

atzoum (approved):
I did one comment for the future , otherwise LGTM

PR #1749, opened 2022-03-15
Fix: Save responses from the destinations with odd content-type header

atzoum:
## Description of the change

Introduce filtering of a destination's body based on regex matching on Content-Type.
This fix uses the standard `mime.ParseMediaType` function to parse the Content-Type
header, which removes any parameters and converts the resulting media type to lowercase.

The regex pattern allows for a broader range of media types to be allowed compared
to the previous filtering algorithm. More specifically, response media types such as
`application/problem+json` will no longer have their body filtered out.

## Notion Link
https://www.notion.so/rudderstacks/Save-responses-from-the-destinations-with-odd-content-type-header-ff7397e0ebac46a2af5a28a1047ddbc5

## Type of change

## Checklist:

- [x] I have added unit tests for the code

## Security

### DIFF
--- a/router/network.go
+++ b/router/network.go
@@ -8,8 +8,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
@@ -19,6 +21,12 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/misc"
 	"github.com/rudderlabs/rudder-server/utils/sysUtils"
 )
+
+var contentTypeRegex *regexp.Regexp
+
+func init() {
+	contentTypeRegex = regexp.MustCompile(`^(text/[a-z0-9.-]+)|(application/([a-z0-9.-]+\+)?(json|xml))$`)
+}
 
 //NetHandleT is the wrapper holding private variables
 type NetHandleT struct {
@@ -182,12 +190,13 @@ func (network *NetHandleT) SendPost(ctx context.Context, structData integrations
 			//Detecting content type of the respBody
 			contentTypeHeader = http.DetectContentType(respBody)
 		}
+		mediaType, _, _ := mime.ParseMediaType(contentTypeHeader)
 
-		// If content type is not of type "*text*", overriding it with empty string
-		if !(strings.Contains(strings.ToLower(contentTypeHeader), "text") ||
-			strings.Contains(strings.ToLower(contentTypeHeader), "application/json") ||
-			strings.Contains(strings.ToLower(contentTypeHeader), "application/xml")) {
-			respBody = []byte("")
+		// If media type is not in some human readable format (text,json,xml), override the response with an empty string
+		// https://www.iana.org/assignments/media-types/media-types.xhtml
+		isHumanReadable := contentTypeRegex.MatchString(mediaType)
+		if !isHumanReadable {
+			respBody = []byte("redacted due to unsupported content-type")
 		}
 
 		if err != nil {
--- a/router/network_test.go
+++ b/router/network_test.go
@@ -8,7 +8,8 @@ import (
 
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
-	"github.com/onsi/gomega"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
 	mocksSysUtils "github.com/rudderlabs/rudder-server/mocks/utils/sysUtils"
 	"github.com/rudderlabs/rudder-server/processor/integrations"
 	"github.com/rudderlabs/rudder-server/utils/logger"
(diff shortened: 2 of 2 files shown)


router/network.go
```diff
@@ -182,11 +184,12 @@ func (network *NetHandleT) SendPost(ctx context.Context, structData integrations
...
+		mediaType, _, _ := mime.ParseMediaType(contentTypeHeader)
 
-		// If content type is not of type "*text*", overriding it with empty string
-		if !(strings.Contains(strings.ToLower(contentTypeHeader), "text") ||
-			strings.Contains(strings.ToLower(contentTypeHeader), "application/json") ||
-			strings.Contains(strings.ToLower(contentTypeHeader), "application/xml")) {
+		// If media type is not in some human readable format (text,json,xml), override the response with an empty string
+		// https://www.iana.org/assignments/media-types/media-types.xhtml
+		isHumanReadable, _ := regexp.MatchString(`^(text/[a-z0-9.-]+)|(application/([a-z0-9.-]+\+)?(json|xml))$`, mediaType)
+		if !isHumanReadable {
```
atzoum, line 192: @chandumlg this is my interpretation of why we are dropping the response body for some content types, please feel free to comment.

router/network.go
```diff
@@ -182,11 +184,12 @@ func (network *NetHandleT) SendPost(ctx context.Context, structData integrations
...
 		}
+		mediaType, _, _ := mime.ParseMediaType(contentTypeHeader)
 
-		// If content type is not of type "*text*", overriding it with empty string
-		if !(strings.Contains(strings.ToLower(contentTypeHeader), "text") ||
-			strings.Contains(strings.ToLower(contentTypeHeader), "application/json") ||
-			strings.Contains(strings.ToLower(contentTypeHeader), "application/xml")) {
+		// If media type is not in some human readable format (text,json,xml), override the response with an empty string
+		// https://www.iana.org/assignments/media-types/media-types.xhtml
+		isHumanReadable, _ := regexp.MatchString(`^(text/[a-z0-9.-]+)|(application/([a-z0-9.-]+\+)?(json|xml))$`, mediaType)
```
lvrach, line 191: For performance reasons, I would suggest that you [compile](https://pkg.go.dev/regexp#MustCompile) once and then use it here.

You can do that by using init() and compiling as a global variable

lvrach: Also, I have to say I am not very fun of using regex. They are hard to read,  to ensure correctness, and can impact performance.

In this case, It seems to be better than doing multiple Contains, and the json / xml suffix would make the existing approach even more complex.

It would be nice, to add some tests for this. Maybe Isolate this to a separate function. Some test cases would help document this function.

lvrach (approved):
LGTM, with one comment for performance.

router/network.go
atzoum: > For performance reasons, I would suggest that you [compile](https://pkg.go.dev/regexp#MustCompile) once and then use it here.
> 
> You can do that by using init() and compiling as a global variable

would this
```
var contentTypeRegex *regexp.Regexp

func init() {
	contentTypeRegex = regexp.MustCompile(`^(text/[a-z0-9.-]+)|(application/([a-z0-9.-]+\+)?(json|xml))$`)
}
```
be better than that?
```
var contentTypeRegex = regexp.MustCompile(`^(text/[a-z0-9.-]+)|(application/([a-z0-9.-]+\+)?(json|xml))$`)
```

why?

router/network.go
cisse21: The reason why we drop few metrics is because for some cases the response type might be in image format or something which is too heavy and might cause the status table to bloat up

atzoum: > The reason why we drop few metrics is because for some cases the response type might be in image format or something which is too heavy and might cause the status table to bloat up

Thanks @cisse21, do you believe the current comment is sufficient or should I change it?

router/network.go
```diff
@@ -182,11 +190,12 @@ func (network *NetHandleT) SendPost(ctx context.Context, structData integrations
...
 
-		// If content type is not of type "*text*", overriding it with empty string
-		if !(strings.Contains(strings.ToLower(contentTypeHeader), "text") ||
-			strings.Contains(strings.ToLower(contentTypeHeader), "application/json") ||
-			strings.Contains(strings.ToLower(contentTypeHeader), "application/xml")) {
+		// If media type is not in some human readable format (text,json,xml), override the response with an empty string
+		// https://www.iana.org/assignments/media-types/media-types.xhtml
+		isHumanReadable := contentTypeRegex.MatchString(mediaType)
+		if !isHumanReadable {
 			respBody = []byte("")
```
lvrach, line 199:
```suggestion
			respBody = []byte(fmt.Sprintf("Response contains unsupported content type: %q",  mediaType))
```

router/network.go
```diff
@@ -182,11 +190,12 @@ func (network *NetHandleT) SendPost(ctx context.Context, structData integrations
 			//Detecting content type of the respBody
 			contentTypeHeader = http.DetectContentType(respBody)
 		}
+		mediaType, _, _ := mime.ParseMediaType(contentTypeHeader)
```
lvrach, line 193: we are ignoring error here

lvrach (approved):

PR #1754, opened 2022-03-16
retention support per destination id

chandumlg:
## Description of the change

Support to retire jobs of a particular destination through config

## Notion Link

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=aa64b05b1986448ba4f97bb9417849d0

## Type of change

## Checklist:

## Security

### DIFF
--- a/router/utils/utils.go
+++ b/router/utils/utils.go
@@ -56,7 +56,7 @@ func ToBeDrained(job *jobsdb.JobT, destID, toAbortDestinationIDs string, destina
 	if jobReceivedAt.Exists() {
 		jobReceivedAtTime, err := time.Parse(misc.RFC3339Milli, jobReceivedAt.String())
 		if err == nil {
-			if time.Now().UTC().Sub(jobReceivedAtTime.UTC()) > getRetentionTimeForDestination(destID) {
+			if time.Since(jobReceivedAtTime) > getRetentionTimeForDestination(destID) {
 				return true, "job expired"
 			}
 		}
--- a/router/utils/utils.go
+++ b/router/utils/utils.go
@@ -41,13 +41,22 @@ func loadConfig() {
 	config.RegisterDurationConfigVariable(time.Duration(720), &JobRetention, true, time.Hour, "Router.jobRetention")
 }
 
+func getRetentionTimeForDestination(destID string) time.Duration {
+	destJobRetentionFound := config.IsSet("Router." + destID + ".jobRetention")
+	if destJobRetentionFound {
+		return config.GetDuration("Router"+"."+destID+"jobRetention", JobRetention, time.Hour)
+	}
+
+	return JobRetention
+}
+
 func ToBeDrained(job *jobsdb.JobT, destID, toAbortDestinationIDs string, destinationsMap map[string]*BatchDestinationT) (bool, string) {
 	//drain if job is older than a day
 	jobReceivedAt := gjson.GetBytes(job.Parameters, "received_at")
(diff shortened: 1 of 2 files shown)


router/utils/utils.go
```diff
@@ -41,13 +41,22 @@ func loadConfig() {
...
+}
+
 func ToBeDrained(job *jobsdb.JobT, destID, toAbortDestinationIDs string, destinationsMap map[string]*BatchDestinationT) (bool, string) {
 	//drain if job is older than a day
 	jobReceivedAt := gjson.GetBytes(job.Parameters, "received_at")
 	if jobReceivedAt.Exists() {
 		jobReceivedAtTime, err := time.Parse(misc.RFC3339Milli, jobReceivedAt.String())
 		if err == nil {
-			if time.Now().UTC().Sub(jobReceivedAtTime.UTC()) > JobRetention {
+			if time.Now().UTC().Sub(jobReceivedAtTime.UTC()) > getRetentionTimeForDestination(destID) {
```
lvrach, line 59:
```suggestion
			if time.Since(jobReceivedAtTime) > getRetentionTimeForDestination(destID) {
```

I believe `UTC()` doesn't change the underlying time, but have to check.

router/utils/utils.go
```diff
@@ -41,13 +41,22 @@ func loadConfig() {
 	config.RegisterDurationConfigVariable(time.Duration(720), &JobRetention, true, time.Hour, "Router.jobRetention")
 }
 
+func getRetentionTimeForDestination(destID string) time.Duration {
+	destJobRetentionFound := config.IsSet("Router." + destID + ".jobRetention")
```
lvrach, line 45: make `destID` value safe for usage on yaml or env variables.

router/utils/utils.go
chandumlg: Yes. Checked it. No need to check in UTCs

router/utils/utils.go
chandumlg: We already do this in a bunch of places. If needed, clean up all at once later?

atzoum: For converting a config property to an environment variable what we do now is:

1. Convert from camel case to snake case
2. Replace `.` with `_`
3. prepend `RSERVER_`


Here is an excerpt from the documentation of a [configuration management spec](https://download.eclipse.org/microprofile/microprofile-config-2.0-M1/microprofile-config-spec.html#default_configsources.env.mapping) used in java, which I find insightful. We may follow the same (or similar) approach:

> Some operating systems allow only alphabetic characters or an underscore, _, in environment variables. Other characters such as ., /, etc may be disallowed. In order to set a value for a config property that has a name containing such disallowed characters from an environment variable, the following rules are used.
> 
> The ConfigSource for the environment variables searches three environment variables for a given property name (e.g. com.ACME.size):
>
> Exact match (i.e. com.ACME.size)
>
> Replace each character that is neither alphanumeric nor _ with _ (i.e. com_ACME_size)
> 
> Replace each character that is neither alphanumeric nor _ with _; then convert the name to upper case (i.e. COM_ACME_SIZE)

>

atzoum (approved):
LGTM if we only plan to use `yaml` as a configuration source for such parameters. Otherwise needs more work to properly support such parameters as environment variables.

router/utils/utils.go
chandumlg: Good point. We will clean up later.

PR #1763, opened 2022-03-21
Dynamic cluster: Lifecycle control for server mode changes

lvrach:
_Paired programed this PR with @kuldeep0020_ 

## Dynamic cluster 
Screenshot 2022-03-24 at 7 54 50 AM

Dynamic cluster is a component of the system responsible for dynamically - without requiring server restart - changing the state for rudder-server, based on cluster needs.

Two aspects can change:
* server-mode("NORMAL", "DEGRADED")
* workspaceIDs (not implemented in this PR)

The dynamic cluster relies on a `stateProvider`  to get notified for latest desired state.

A stateProvider can be two things:
* ETCD ( dynamic config)
* env ( static - sends the state once)

It is also important for the desired server-mode to be Acknowledged. For this a new struct has been introduced `ModeACK`. 

## Test and Risks

This code is tested both with unit and integration tests, but at this point is not used. So it is safe to merge it without creating any risk.

## Notion Link

> [Notion Link
](https://www.notion.so/rudderstacks/Dynamic-Cluster-Manager-479c878d94f24a3791e3fcfd3be272f8)
## Type of change

## Checklist:

## Security


utils/types/servermode/servermode.go
```diff
@@ -0,0 +1,29 @@
+package servermode
+
+type Mode string
+
+const (
+	UndefinedMode Mode = ""
+	NormalMode    Mode = "NORMAL"
+	DegradedMode  Mode = "DEGRADED"
```
lvrach, line 8:
```suggestion
	DegradedMode  Mode = "DEGRADED"
	StandByMode    Mode = "STANDBY"
```

app/cluster/dynamic.go
```diff
@@ -0,0 +1,104 @@
...
+	BatchRouterDB Lifecycle
+	ErrorDB       Lifecycle
+
+	Processor Lifecycle
+	Router    Lifecycle
+
+	currentMode servermode.Mode
+}
+
+func (d *Dynamic) Run(ctx context.Context) error {
```
lvrach, line 33: Start with degraded mode instead of undefined.

processor/manager.go
```diff
@@ -32,25 +34,29 @@ func (proc *LifecycleManager) Run(ctx context.Context) error {
 // StartNew starts a processor, this is not a blocking call.
```
atzoum, line 34: documentation needs to be updated, at least the first word :)

app/cluster/dyanmic_test.go
```diff
@@ -0,0 +1,206 @@
+package cluster_test
```
atzoum, line 1: The initial intention must have been to name this file `dynamic_test.go`

router/manager/manager.go
```diff
@@ -15,9 +15,9 @@ import (
 )
 
 var (
-	objectStorageDestinations []string
-	asyncDestinations         []string
-	warehouseDestinations     []string
+	objectStorageDestinations = []string{"S3", "GCS", "AZURE_BLOB", "MINIO", "DIGITAL_OCEAN_SPACES"}
```
atzoum, line 18: I cannot find how these slices were initialized before

app/cluster/dynamic.go
```diff
@@ -0,0 +1,104 @@
...
+	ServerMode() <-chan servermode.ModeAck
+}
+
+type Lifecycle interface {
+	Start()
+	Stop()
+}
+
+type Dynamic struct {
+	Provider modeProvider
```
atzoum, line 20: This is interesting... a public field of private type! Wonder how this would behave in practice :)

utils/types/servermode/servermode.go
```diff
@@ -0,0 +1,29 @@
+package servermode
+
+type Mode string
+
+const (
+	UndefinedMode Mode = ""
```
lvrach, line 6:
```suggestion
```

atzoum, line 6: One way to get away with `UndefinedMode` which doesn't look that much different from `DegradedMode` so far, would be to use int enum types, e.g.

```go
type Mode int

const (
	Degraded Mode = iota
        Normal
)

func (m Mode) String() string {
	return [2]string{"NORMAL", "DEGRADED"}[m]
}

func ValueOf(m string) (Mode, error) {

	switch m {
	case "NORMAL":
		return Normal, nil
	case "DEGRADED":
		return Degraded, nil
	}
	return Degraded, fmt.Errorf("unknown mode %s", m)

}
```

app/cluster/dynamic.go
```diff
@@ -0,0 +1,104 @@
...
+			err := d.handleModeChange(newMode.Mode())
+			if err != nil {
+				return err
+			}
+			newMode.Ack()
+		}
+	}
+}
+
+func (d *Dynamic) start() {
```
atzoum, line 51: Is there any specific reasoning behind the current start/stop order?
Normally, one would expect the stop order to be the inverse of the start order.

A start strategy that just came into mind is to start based on the inverse order of the event processing steps, e.g.

1. RouterDB
2. Router
4. BatchRouterDB
5. BatchRouter
6. ErrorDB
7. GatewayDB
8. Processor

In the above list I would also expect to see a Gateway too. What would happen if we had a running Gateway operating against a stopped GatewayDB?

utils/types/servermode/servermode.go
```diff
@@ -0,0 +1,29 @@
...
+	NormalMode    Mode = "NORMAL"
+	DegradedMode  Mode = "DEGRADED"
+)
+
+type ModeAck struct {
```
atzoum, line 11: does `servermode.Ack` look better than `servermode.ModeAck`?

router/manager/manager.go
lvrach: cc @kuldeep0020

utils/types/servermode/servermode.go
lvrach: We have removed the undefined mode actually. 
We need a parse/validator method, not sure if we need iotas seems a bit overkill.

router/manager/manager.go
kuldeep0020: These were hardcoded before as well, we are not reading these from any config or storage as of now.

app/cluster/dynamic.go
lvrach: > Normally, one would expect the stop order to be the inverse of the start order.

That is the idea, well in our case as long as dbs start before the processor/router we are set. We are thinking of start/stop them concurrently to speed up things.

Also, router includes batchrouter.

lvrach: > In the above list I would also expect to see a Gateway too. What would happen if we had a running Gateway operating against a stopped GatewayDB?

This part is a bit more complicated, the short answer is: For now gateway is going to run in another process. Dynamic server mode in embedded setup, needs more effort and we are going to tackle it later.

utils/types/servermode/servermode.go
lvrach: actually does, yes.

app/cluster/dynamic.go
lvrach: We can export it ok, but no one should actually use this interface. 
Since will create coupling with this component dependency.

app/cluster/dynamic.go
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+
+	"github.com/rudderlabs/rudder-server/utils/types/servermode"
+)
+
+type modeProvider interface {
+	ServerMode() <-chan servermode.ModeAck
+}
+
+type Lifecycle interface {
```
lvrach, line 27:
```suggestion
type lifecycle interface {
```

utils/types/servermode/servermode.go
lvrach: Thinking more about it, I like the `iota` approach, I can see the benefits.

atzoum: I was thinking about writing a generator for sparing us with the boilerplate code, but looks like someone else caught me already!
https://github.com/abice/go-enum

As a bonus, this generator handles json marshalling/unmarshalling concerns too, which allows us to se enum types in structs used for json binding

app/cluster/dynamic.go
kuldeep0020: done

utils/types/servermode/servermode.go
kuldeep0020: actually, I agree.

app/cluster/dynamic.go
kuldeep0020: won't affect because processor or router will only be started after DBs

router/manager/manager.go
kuldeep0020: This is hardcoded as of now in router.go

processor/manager.go
kuldeep0020: done

app/cluster/dynamic.go
kuldeep0020: done

utils/types/servermode/servermode.go
kuldeep0020: done

app/cluster/dynamic.go
```diff
@@ -0,0 +1,155 @@
...
+	//	switch newMode {
+	//	case servermode.NormalMode:
+	//		d.logger.Info("Transiting the server from UndefinedMode to NormalMode")
+	//		d.start()
+	//	case servermode.DegradedMode:
+	//		d.logger.Info("Server is running in UndefinedMode, can not transit to DegradedMode.")
+	//	default:
+	//		d.logger.Errorf("Unsupported transition from UndefinedMode to %s \n", newMode)
+	//		return fmt.Errorf("unsupported transition from UndefinedMode to %s", newMode)
+	//	}
```
lvrach, line 132:
```suggestion
```

processor/manager.go
```diff
@@ -32,25 +34,29 @@ func (proc *LifecycleManager) Run(ctx context.Context) error {
...
 //If the processor is not completely started and the data started coming then also it will not be problematic as we
 //are assuming that the DBs will be up.
-func (proc *LifecycleManager) StartNew() {
-	proc.HandleT.Setup(proc.backendConfig, proc.gatewayDB, proc.routerDB, proc.batchRouterDB,
-		proc.errDB, proc.clearDB, proc.reporting, proc.multitenantStats)
+func (proc *LifecycleManager) Start() {
+	if proc.Transformer != nil {
```
atzoum, line 38: this appears to be a late stage for performing configuration changes. Why is it necessary and why can't we do this earlier, when the `LifecycleManager` is being created?

app/cluster/dyanmic_test.go
```diff
@@ -0,0 +1,220 @@
...
+	//	require.True(t, routerDB.callOrder < processor.callOrder)
+	//	require.True(t, batchRouterDB.callOrder < processor.callOrder)
+	//	require.True(t, errorDB.callOrder < processor.callOrder)
+	//
+	//	t.Log("dbs should be started before router")
+	//	require.True(t, gatewayDB.callOrder < router.callOrder)
+	//	require.True(t, routerDB.callOrder < router.callOrder)
+	//	require.True(t, batchRouterDB.callOrder < router.callOrder)
+	//	require.True(t, errorDB.callOrder < router.callOrder)
+	//})
```
lvrach, line 127:
```suggestion
```

processor/manager.go
lvrach: This pattern of lazy inti is quite common in Go. There is no `New` function for `LifecycleManager`

chandumlg (approved):

kuldeep0020 (approved):

processor/manager.go
atzoum: Instead of having an exported struct field `Transformer` which ultimately acts as an **indirect setter** for the private field `HandleT.transformer` wouldn't it be simpler and more straightforward to omit this field from the struct, remove any special handling from the `Start` method and add the missing setter method?

```go
func (proc *LifecycleManager) SetTransformer(t transformer.Transformer) {
    proc.HandleT.transformer = proc.Transformer
}
```

PR #1766, opened 2022-03-22
context timeout pending events queries

chandumlg:
## Description of the change

Timing out pending events queries and returning 1 in case of errors.

## Notion Link

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=8f4502d905fb47a3a5a0fe5d41fe841c

## Type of change

## Checklist:

## Security

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -922,22 +922,25 @@ func (gateway *HandleT) pendingEventsHandler(w http.ResponseWriter, r *http.Requ
 		}
 	}
 
+	ctx, cancel := context.WithTimeout(context.TODO(), config.GetDuration("Gateway.pendingEventsQueryTimeout", time.Duration(10), time.Second))
+	defer cancel()
+
 	var pending bool
 	if !excludeGateway {
-		pending = gateway.readonlyGatewayDB.HavePendingJobs([]string{CustomVal}, -1, gwParameterFilters)
+		pending = gateway.readonlyGatewayDB.HavePendingJobs(ctx, []string{CustomVal}, -1, gwParameterFilters)
 		if pending {
 			w.Write([]byte(fmt.Sprintf("{ \"pending_events\": %d }", getIntResponseFromBool(pending))))
 			return
 		}
 	}
 
-	pending = gateway.readonlyRouterDB.HavePendingJobs(nil, -1, rtParameterFilters)
+	pending = gateway.readonlyRouterDB.HavePendingJobs(ctx, nil, -1, rtParameterFilters)
 	if pending {
 		w.Write([]byte(fmt.Sprintf("{ \"pending_events\": %d }", getIntResponseFromBool(pending))))
 		return
 	}
 
-	pending = gateway.readonlyBatchRouterDB.HavePendingJobs(nil, -1, rtParameterFilters)
+	pending = gateway.readonlyBatchRouterDB.HavePendingJobs(ctx, nil, -1, rtParameterFilters)
 	if pending {
 		w.Write([]byte(fmt.Sprintf("{ \"pending_events\": %d }", getIntResponseFromBool(pending))))
 		return
--- a/jobsdb/readonly_jobsdb.go
+++ b/jobsdb/readonly_jobsdb.go
@@ -1,12 +1,14 @@
 package jobsdb
 
 import (
+	"context"
 	"database/sql"
 	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
 
+	"github.com/rudderlabs/rudder-server/config"
 	"github.com/rudderlabs/rudder-server/utils/logger"
 
 	"time"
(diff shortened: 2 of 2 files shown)


gateway/gateway.go
```diff
@@ -922,22 +922,25 @@ func (gateway *HandleT) pendingEventsHandler(w http.ResponseWriter, r *http.Requ
 		}
 	}
 
+	ctx, cancel := context.WithTimeout(context.TODO(), config.GetDuration("Gateway.pendingEventsQueryTimeout", time.Duration(10), time.Second))
```
lvrach, line 925: We could get the context for the [request](https://pkg.go.dev/net/http#Request.Context):

```suggestion
	ctx, cancel := context.WithTimeout(r.Context(), config.GetDuration("Gateway.pendingEventsQueryTimeout", time.Duration(10), time.Second))
```

The benefit is HTTP request timeouts or request getting canceled will also affect the context.

jobsdb/readonly_jobsdb.go
```diff
@@ -22,7 +24,7 @@ import (
 ReadonlyJobsDB interface contains public methods to access JobsDB data
 */
 type ReadonlyJobsDB interface {
-	HavePendingJobs(customValFilters []string, count int, parameterFilters []ParameterFilterT) bool
+	HavePendingJobs(ctx context.Context, customValFilters []string, count int, parameterFilters []ParameterFilterT) bool
```
atzoum, line 27: The current solution encapsulates error handling behaviour as implementation details of the`ReadonlyJobsDB#HavePendingJobs` method. The method treats an `error` as a `true` result, instead of letting the caller handle the error appropriately.
If we have any need in the future to use `ReadonlyJobsDB#HavePendingJobs` for some other purpose and want to be able to interpret an error in a different way, then we should simplify it already.

Based on the above, I propose to change the method signatures of all methods inside `readonly_jobsdb.go` (both private and public) which now accept a `ctx` to return the error instead of handling it on their own, so that the error can bubble up to the caller.

1. Change the following method signatures:
```
func (jd *ReadonlyHandleT) HavePendingJobs(ctx context.Context, customValFilters []string, count int, parameterFilters []ParameterFilterT) bool, error
func (jd *ReadonlyHandleT) haveUnprocessedJobs(ctx context.Context, customValFilters []string, parameterFilters []ParameterFilterT) bool, error
func (jd *ReadonlyHandleT) getUnprocessedJobsDSCount(ctx context.Context, ds dataSetT, customValFilters []string, parameterFilters []ParameterFilterT) int64, error
func (jd *ReadonlyHandleT) haveNonSucceededJobs(ctx context.Context, customValFilters []string, parameterFilters []ParameterFilterT) bool, error
func (jd *ReadonlyHandleT) haveProcessedJobs(ctx context.Context, stateFilter []string, customValFilters []string, parameterFilters []ParameterFilterT) bool, error
func (jd *ReadonlyHandleT) haveProcessedJobs(ctx context.Context, stateFilter []string, customValFilters []string, parameterFilters []ParameterFilterT) bool, error
```
2. Modify method implementations so that they return errors when they happen, instead of handling them internally.
3. Catch the error at the `gateway` and treat it as `pending=true`

It should really be the responsibility of the caller, i.e. the `gateway` to decide how to interpret an error and in this specific case, the `gateway` can decide that whenever an error occurs it should assume that pending events do exist.

chandumlg: Changed HavePendingJobs. Since that is the only function gateway calls. We can change other functions later.

gateway/gateway.go
```diff
@@ -922,24 +922,27 @@ func (gateway *HandleT) pendingEventsHandler(w http.ResponseWriter, r *http.Requ
...
 	}
 
-	pending = gateway.readonlyBatchRouterDB.HavePendingJobs(nil, -1, rtParameterFilters)
-	if pending {
-		w.Write([]byte(fmt.Sprintf("{ \"pending_events\": %d }", getIntResponseFromBool(pending))))
+	pending, err = gateway.readonlyBatchRouterDB.HavePendingJobs(ctx, nil, -1, rtParameterFilters)
+	if err != nil || pending {
+		w.Write([]byte(fmt.Sprintf("{ \"pending_events\": %d }", getIntResponseFromBool(true))))
 		return
 	}
```
atzoum, line 948: Looking at the implementation of `getWarehousePending`, we return `false` in case of an `error` which appears to be in contrast with what we are doing elsewhere. Is this intentional? Any idea why?

gateway/gateway.go
```diff
@@ -922,24 +922,27 @@ func (gateway *HandleT) pendingEventsHandler(w http.ResponseWriter, r *http.Requ
...
 
-	pending = gateway.readonlyBatchRouterDB.HavePendingJobs(nil, -1, rtParameterFilters)
-	if pending {
-		w.Write([]byte(fmt.Sprintf("{ \"pending_events\": %d }", getIntResponseFromBool(pending))))
+	pending, err = gateway.readonlyBatchRouterDB.HavePendingJobs(ctx, nil, -1, rtParameterFilters)
+	if err != nil || pending {
+		w.Write([]byte(fmt.Sprintf("{ \"pending_events\": %d }", getIntResponseFromBool(true))))
```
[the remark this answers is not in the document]
lvrach:
```go
w.Write([]byte(`{ "pending_events": 1 }`))
```

gateway/gateway.go
chandumlg: Not sure why. But I understand there is a contrast.

One thing is getting warehouse status is not db intensive. So, I suggest for now leave it as is. We can have warehouse team address this later.

lvrach (approved):
LGTM, thanks for addressing the comments.

Issue #1672, opened 2022-02-17
Metrics missing


azhelev:
**Describe the bug**
Some metrics are not available. By not available i mean they are not emitted at all. Most importantly `recovery.mode_normal` is not there. If i restart the server several times or crash it the metric may appear.
Several other metrics are also behaving strangely. `jobsdb.gw_tables_count`, `jobsdb.tables_count`, `jobsdb.batch_rt_tables_count` show a value initially but then they disappear.

![image](https://user-images.githubusercontent.com/12055675/154448384-ad38ecde-8a35-4a79-b762-f121007134ae.png)

**Steps to reproduce the bug**
1. Start rudder-server

**Expected behavior**
All metrics are emitted and available

**Any additional context**
Running self-hosted rudderstack on kubernetes, installed with the helm chart. Metrics are obtained by the telegraf sidecar container.

chandumlg:
@azhelev Table count metrics are emitted whenever a new dataset is created. If there is not activity (meaning no flow of events and no new datasets are created), metrics are emitted. Run some load and check.

`recovery.mode_normal` is a gauge type metric, which is fired every 10s. It works for me. Can you double check?

Thank you for your interest.

azhelev:
@chandumlg thanks for the clarification the table count metrics, that explains a lot.
I'm sure about the `recovery.mode_normal` metric, see the following screenshot.

![image](https://user-images.githubusercontent.com/12055675/154480653-606035e6-1886-43d6-8868-ce08f9c6f1cc.png)

Here i intentionally crashed the rudder-server process (exec into the pod and kill the pid). After the container came back up the metric showed up and is emitted regularly as it should along with some other metrics that were not present.
Notice how gateway requests and number of events metrics are present all the time.
![image](https://user-images.githubusercontent.com/12055675/154481024-7fb31a63-9dc4-430b-88ff-b28d90a7da58.png)

Regards
Atanas

chandumlg:
@saurav-malani Can you take a look at this pls.

saurav-malani:
@azhelev we managed to replicate the issue. Also, we have identified, the reason behind why we are not getting the metrics. Essentially, it is related to the ordering of `telegraf_sidecar` & `backend` containers. We are working on fixing it. 
Thanks for bringing this to our notice.

saurav-malani:
@azhelev we have fixed the issue. Thanks for pointing it out.

PR #1770, opened 2022-03-23
Metrics issue

saurav-malani:
Issue link: https://github.com/rudderlabs/rudder-server/issues/1672


services/stats/stats.go
```diff
@@ -93,26 +97,59 @@ type RudderStatsT struct {
...
+		//statsd.New returns client even when err!=nil, although that client won't work.
+		//But, since we don't want getting new client to be a blocking call. So, setting it even with the non-working client.
+		//And, will be updated once we have `err==nil`, until then stats sent to telegraf will be dropped.
+		client, err = statsd.New(opts...)
+		if err != nil {
+			isMockClient = true
+			pkgLogger.Errorf("error while setting statsd client: %v", err)
+		}
+		return err
+	}, bo); err != nil {
```
chrikar, line 116: Minor: For code readability, I would extract the function as a separate declaration before calling `backk.Retry(func, bo)`

services/stats/stats.go
```diff
@@ -93,26 +97,59 @@ type RudderStatsT struct {
...
+		if bo.NextBackOff() == backoff.Stop {
+			//if setup didn't succeed in bo.MaxElapsedTime, then we panic. Since, if `RSERVER_ENABLE_STATS` is `true`
+			// & we failed to complete statsd setup in maxElapsedTime then something is wrong. And, we don't want to loose metrics by ignoring the error.
+			panic(err)
```
chrikar, line 120: I'm not very fond of panic-ing. I would suggest we switch to a context based cancel process, so that the server can shut down gracefully. Of course, that might require further changes in the Setup() process beyond the scope of this PR, but it something we need to keep in mind for the future and how we're handling services startup.

services/stats/stats.go
```diff
@@ -93,26 +97,59 @@ type RudderStatsT struct {
...
 	}
-
-	var err error
 	conn = statsd.Address(statsdServerURL)
-	//TODO: Add tags by calling a function...
+	// since, we don't want setup to be a blocking call, creating a separate `go routine`` for retry to get statsd client.
+	var err error
+	//NOTE: this is to get atleast a dummy client, even in case of failure. So, that nil pointer error is not received when client is called.
 	client, err = statsd.New(conn, statsd.TagsFormat(getTagsFormat()), defaultTags())
 	if err != nil {
```
chrikar, line 137: I think we've got a logical error here. If the new statsd connection succeeds with the first attempt, then we're never reaching the call to `collectRuntimeStats(client)`.

services/stats/stats.go
saurav-malani: make sense.

services/stats/stats.go
```diff
@@ -93,26 +97,59 @@ type RudderStatsT struct {
...
-		// If nothing is listening on the target port, an error is returned and
-		// the returned client does nothing but is still usable. So we can
-		// just log the error and go on.
-		pkgLogger.Error(err)
-	}
-	if client != nil {
 		rruntime.Go(func() {
+			//this is a blocking call
```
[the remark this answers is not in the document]
saurav-malani: removed

services/stats/stats.go
saurav-malani: modified

services/stats/stats.go
saurav-malani: thanks for pointing this out, called `collectRuntimeStats` inside `if`  condition by mistake. Fixed.

saurav-malani: `statsd.New` returns a real client when `err==nil` & dummy client when `err!=nil`. That is the reason we are calling `statsd.New` once without retry mechanism. Since, retry mechanism is non-blocking. So, to avoid calling any of the `RudderStats` methods before getting a client (real/dummy), we are making a single blocking call request. So, that we don't get `nil` pointer error.

services/stats/stats.go
chrikar: I see that other services in main.go are already using contexts, so if it's not too much trouble let's try to add this and avoid the panic(). Otherwise, let's open a ticket to handle that separately asap.

services/stats/stats.go
```diff
@@ -93,29 +97,66 @@ type RudderStatsT struct {
...
 		return
 	}
-
-	var err error
 	conn = statsd.Address(statsdServerURL)
-	//TODO: Add tags by calling a function...
+	// since, we don't want setup to be a blocking call, creating a separate `go routine`` for retry to get statsd client.
+	var err error
+	//NOTE: this is to get atleast a dummy client, even in case of failure. So, that nil pointer error is not received when client is called.
 	client, err = statsd.New(conn, statsd.TagsFormat(getTagsFormat()), defaultTags())
```
chrikar, line 139: We don't need this line, as this is called in `getNewStatsdClientWithExpoBackoff()` anyway

mcchran (dismissed):
LGTM.

services/stats/stats.go
```diff
@@ -93,29 +97,66 @@ type RudderStatsT struct {
...
-	}
-	if client != nil {
-		rruntime.Go(func() {
-			collectRuntimeStats(client)
-		})
-	}
+	rruntime.Go(func() {
+		if err != nil {
+
+			getNewStatsdClientWithExpoBackoff(conn, statsd.TagsFormat(getTagsFormat()), defaultTags())
```
chrikar, line 143: Let's rename this to avoid the deepsource error:
```
function 'getNewStatsdClientWithExpoBackoff' seems to be a getter but it does not return any result
```

services/stats/stats.go
chrikar: I've opened a ticket for this: https://www.notion.so/rudderstacks/Use-context-cancellation-instead-of-panic-to-handle-failures-during-server-startup-7fc2948a6fdb40beb46f9e0f059b098b

chrikar (approved):

services/stats/stats.go
atzoum: IMO there is no point in panicking or shutting down gracefully. If we fail to establish a statsd connection after 15 minutes (which is the maximum elapsed time of the backoff library), then it should be about time to accept it and stop trying! Just make sure that we can handle this scenario as well, .e.g. return some error from `getNewStatsdClientWithExpoBackoff` so that the calling routine can decide how to handle the result.

services/stats/stats.go
```diff
@@ -35,6 +36,9 @@ var enableGCStats bool
 var rc runtimeStatsCollector
 var pkgLogger logger.LoggerI
 var statsSamplingRate float32
+var isMockClient bool
```
atzoum, line 39: I propose to group all tag-related variables together. Also some renaming might be appropriate. What I have in mind is:
```
// place all variables in a group
taggedClientsMapLock // locks should be above the resources they are protecting
taggedClientsMap
tagValsList // rename to taggedClientKeys
tagStrList // rename to taggedClientTags
isMockClient // rename to connEstablished and adapt logic with the reversal of this boolean
```

services/stats/stats.go
```diff
@@ -93,29 +97,66 @@ type RudderStatsT struct {
 	dontProcess bool
 }
 
+func getNewStatsdClientWithExpoBackoff(opts ...statsd.Option) {
```
atzoum, line 100: as mentioned in another comment, this method should return an `error` if the maximum wait time is reached.

services/stats/stats.go
```diff
@@ -93,29 +97,66 @@ type RudderStatsT struct {
...
-		rruntime.Go(func() {
-			collectRuntimeStats(client)
-		})
-	}
+	rruntime.Go(func() {
+		if err != nil {
+
+			getNewStatsdClientWithExpoBackoff(conn, statsd.TagsFormat(getTagsFormat()), defaultTags())
+
+			taggedClientsMapLock.Lock()
```
atzoum, line 145: if we receive an error from the previous method, maybe we should disable stats (`statsEnabled = false`) and return.

atzoum (commented):
@saurav-malani after studying `statsd.Client` implementation I believe that you have followed the best approach and there is not much we can do better. I added a few comments regarding naming of variables & the dreadful panicking! Otherwise LGTM!

services/stats/stats.go
lvrach: We can also place everything inside a `var` blog:

```go
var (

// place all variables in a group
taggedClientsMapLock // locks should be above the resources they are protecting
taggedClientsMap
tagValsList // rename to taggedClientPendingKeys
tagStrList  // rename to taggedClientPendingTags
isMockClient // rename to connEstablished and adapt logic with the reversal of this boolean
)
```

services/stats/stats.go
```diff
@@ -93,29 +97,66 @@ type RudderStatsT struct {
...
+			for i, tagValue := range tagValsList {
+				taggedClient := client.Clone(conn, statsd.TagsFormat(getTagsFormat()), defaultTags(), statsd.Tags(tagValue...), statsd.SampleRate(statsSamplingRate))
+				taggedClientsMap[tagStrList[i]] = taggedClient
+			}
+			taggedClientsMapLock.Unlock()
+
+			pkgLogger.Info("statsd client setup succeeded.")
+			isMockClient = false
```
atzoum, line 153: this might need to happen inside the lock

services/stats/stats.go
lvrach: Will it ever reach maximum, or retry constantly at a lesser pace?

services/stats/stats.go
lvrach: > If we fail to establish a statsd connection after 15 minutes
I believe we should keep retrying to connect forever every 15 minutes

atzoum: As an alternative, setting the `MaxElapsedTime` of the backoff to `0` would mean that we will never get a stop signal from the backoff!

services/stats/stats.go
lvrach, line 39: This is beyond the scope of the PR, but we should avoid using global for those values, they could be part of `HandleT` struct (well `HandleT` should be renamed as well). 

And instead of a global `Setup()`, sync.once could be used, to do lazy init

mcchran (changes requested):

services/stats/stats.go
atzoum: not needed if we retry forever

services/stats/stats.go
atzoum: otherwise, if we keep retrying forever, no adaptation is needed!

services/stats/stats.go
chrikar: Indeed, not sending metrics is not critical to the function of the service, but it is something we should be aware of, which right now we get from the error logs printed every minute.

saurav-malani: there are actually many Setup() functions in main.go that requires us to pass context. Also, we are using `panic` to frequently, and most of them could be handled more gracefully, I suppose. So, I believe creating a separate ticket for this would be better.

services/stats/stats.go
saurav-malani: we actually need this to get dummy client & avoid nil pointer in client. Since, `exponentialBackoff` is non blocking.

services/stats/stats.go
saurav-malani: since we have global variable `client` that is being set by calling this function, converting `getNewStatsdClientWithExpoBackoff` to `setNewStatsdClientWithExpoBackoff`

services/stats/stats.go
saurav-malani: done

services/stats/stats.go
saurav-malani: made change to return error from this function, but simply logging it in `Setup` func. Currently, since none of the `Setup's` called from main.go returns `error`. I feel it might be better to take it up together for all of them, and also handle other panics. @atzoum  WDYT?

services/stats/stats.go
```diff
@@ -93,29 +97,66 @@ type RudderStatsT struct {
 	dontProcess bool
 }
 
+func getNewStatsdClientWithExpoBackoff(opts ...statsd.Option) {
+	maxWait := time.Minute * 10
+	bo := backoff.NewExponentialBackOff()
+	bo.MaxInterval = time.Minute
+	bo.MaxElapsedTime = maxWait
```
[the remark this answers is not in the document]
saurav-malani: @lvrach but do we really want to retry it forever? Or, retry it for sometime & then log the error.
CC: @chandumlg

services/stats/stats.go
saurav-malani: `statsEnabled` seems to be an environment variable, which is meant to be set by us. So, do you think it would a good idea to modify it inside code? As I feel, it might create confusion in some cases.

services/stats/stats.go
saurav-malani: makes sense, modified

services/stats/stats.go
```diff
@@ -163,22 +204,21 @@ func newTaggedStat(Name string, StatType string, tags Tags, samplingRate float32
...
-		} else {
-			taggedClient = client.Clone(conn, statsd.TagsFormat(getTagsFormat()), defaultTags(), statsd.Tags(tagVals...), statsd.SampleRate(samplingRate))
+
+		if isMockClient {
+			tagStrList = append(tagStrList, tagStr)
+			tagValsList = append(tagValsList, tagVals)
 		}
+		taggedClient = client.Clone(conn, statsd.TagsFormat(getTagsFormat()), defaultTags(), statsd.Tags(tagVals...), statsd.SampleRate(samplingRate))
+
+		taggedClientsMapLock.Lock()
```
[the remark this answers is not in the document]
saurav-malani: makes sense, modified.

services/stats/stats.go
atzoum: > made changes to return error from this function, but simply logging it in `Setup` func. Currently, since none of the `Setup's` called from main.go returns `error`. I feel it might be better to take it up together for all of them, and also handle other panics. @atzoum WDYT?

What you describe sounds good, however I cannot see it materialized in this PR

services/stats/stats.go
atzoum: > `statsEnabled` seems to be an environment variable, which is meant to be set by us. So, do you think it would a good idea to modify it inside code? As I feel, it might create confusion in some cases.

The idea behind this proposal is that there is no point in keep filling the tag arrays if we know that we are not going to use them. Practically, the moment that we stop trying to enable stats is a good time to admit that stats are not enabled after all!

services/stats/stats.go
saurav-malani: can you check now?

services/stats/stats.go
saurav-malani: makes sense.

services/stats/stats.go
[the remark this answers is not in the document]
chandumlg: Retrying with some interval should be okay. No big deal.

services/stats/stats.go
```diff
@@ -195,7 +241,7 @@ func NewTaggedStat(Name string, StatType string, tags Tags) (rStats RudderStats)
 	return DefaultStats.NewTaggedStat(Name, StatType, tags)
 }
 
-// Count increases the stat by n. Only applies to CountType stats
+//
```
lvrach, line 244: Do we need the empty comment? Did you intend for this change to happen?

services/stats/stats.go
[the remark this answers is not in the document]
lvrach: > Retrying with some interval should be okay. No big deal.
@chandumlg are you ok with retrying forever?

mcchran (approved):

PR #1776, opened 2022-03-24
Mitigation for stuck processor: Panic on processor store timeout

lvrach:
## Issue: stuck processor

The processor can get stuck under some conditions. We know that in this case writes to batchrouter (or router) at wait for a very long time or forever.

We still have found the root cause of this issue, but given the urgency and impact of this problem, we would like to provide quick mitigation.

**Note**: we could have also used metrics, to automatically restart the server and avoid this hack. However, automatic actions based on metrics are only available for Prometheus alerts, not Kapacitor.

## Mitigation

When the store method is called, we create a context with a timeout. We check in a goroutine if the context is Done due to Deadline, if yes then we panic.

It might seem crude way but is better to panic than block forever. Unfortunately, there wasn't any easy way to exit gracefully.

## Timeout

I've set the timeout to 5 minutes, I wanted to make sure that we want `panic` in case of a slow insert, but we will detect a stuck processor fast enough.

## How to test
apply the following change

```diff
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -173,7 +173,7 @@ Processor:
   loopSleep: 10ms
   maxLoopSleep: 5000ms
   fixedLoopSleep: 0ms
-  storeTimeout: 5m
+  storeTimeout: 1ms
   maxLoopProcessEvents: 10000
   transformBatchSize: 100
   userTransformBatchSize: 200
```

Run the server and generate an event. You should see:

```
panic: processor .Store() timed out after 1ms

goroutine 1525 [running]:
github.com/rudderlabs/rudder-server/processor.(*HandleT).Store.func1({0x102f03c38, 0x14001c6aa20}, 0x14000466c00)
        /Users/leonidas/src/rudder/rudder-server/processor/processor.go:1487 +0x118
created by github.com/rudderlabs/rudder-server/processor.(*HandleT).Store
        /Users/leonidas/src/rudder/rudder-server/processor/processor.go:1484 +0x98
exit status 2
```

## Notion Link

[Notion Link
](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=c6a8cbadea854c60bb705e757bc7ee94)
## Type of change

## Checklist:

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -114,6 +114,7 @@ type HandleT struct {
 	transformerFeatures            json.RawMessage
 	readLoopSleep                  time.Duration
 	maxLoopSleep                   time.Duration
+	storeTimeout                   time.Duration
 	multitenantI                   multitenant.MultiTenantI
 	backgroundWait                 func() error
 	backgroundCancel               context.CancelFunc
@@ -304,6 +305,7 @@ func (proc *HandleT) Setup(backendConfig backendconfig.BackendConfig, gatewayDB 
 
 	proc.readLoopSleep = readLoopSleep
 	proc.maxLoopSleep = maxLoopSleep
+	proc.storeTimeout = storeTimeout
 
 	proc.multitenantI = multitenantStat
 	proc.gatewayDB = gatewayDB
@@ -444,6 +446,7 @@ var (
 	subJobSize                int
 	readLoopSleep             time.Duration
 	maxLoopSleep              time.Duration
+	storeTimeout              time.Duration
 	loopSleep                 time.Duration // DEPRECATED: used only on the old mainLoop
 	fixedLoopSleep            time.Duration // DEPRECATED: used only on the old mainLoop
 	maxEventsToProcess        int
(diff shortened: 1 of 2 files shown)


processor/processor.go
```diff
@@ -473,6 +476,8 @@ func loadConfig() {
 	config.RegisterIntConfigVariable(0, &pipelineBufferedItems, false, 1, "Processor.pipelineBufferedItems")
 	config.RegisterIntConfigVariable(2000, &subJobSize, false, 1, "Processor.subJobSize")
 	config.RegisterDurationConfigVariable(time.Duration(5000), &maxLoopSleep, true, time.Millisecond, []string{"Processor.maxLoopSleep", "Processor.maxLoopSleepInMS"}...)
+	config.RegisterDurationConfigVariable(5*time.Minute, &storeTimeout, true, time.Millisecond, "Processor.storeTimeout")
```
chandumlg, line 479: Default value passed needs correction.

chandumlg: 5 min = time.Duration(5*time.Minute/time.Millisecond)

processor/processor.go
```diff
@@ -1473,6 +1478,16 @@ type storeMessage struct {
 }
 
 func (proc *HandleT) Store(in storeMessage) {
+	ctx, cancel := context.WithTimeout(context.TODO(), proc.storeTimeout)
```
atzoum, line 1481: please add a TODO to remove this once proper timeouts and error handling is implemented for database operations

processor/processor.go
snarkychef, line 479: I feel it's better to use time.Minute as unit here, to reduce impact of misconfiguration.

processor/processor.go
```diff
@@ -1473,6 +1478,16 @@ type storeMessage struct {
...
+	defer cancel()
+
+	go func() {
+		<-ctx.Done()
+		if ctx.Err() == context.DeadlineExceeded {
+			panic(fmt.Sprintf("processor .Store() timed out after %s", proc.storeTimeout))
```
snarkychef, line 1487: As we store data into multiple tables (multiple transactions?), we might be writing some duplicates, depending on which query caused the panic.

processor/processor.go
lvrach: Oh got it, I will fix it. 

I find the base unit confusing (4th param). I am expecting to get what I provide as a default.

processor/processor.go
lvrach: That is true, but I believe this is expected, to happen in abnormal situations. Otherwise, we need to introduce transactions.

Since we have seen this happen on batchrouter as far as I know, we can switch the order to avoid duplicates in this case.

chandumlg (approved):

processor/processor.go
lvrach: I have switched the order of inserts, this would make double writes less likely

snarkychef (approved):

PR #1785, opened 2022-03-30
Fix: Jobsdb migration bug

atzoum:
## Description of the change

If server crashes (panics)  while the `migrateDSLoop` is running and has managed to create a new `dataSet` but has not managed to mark the start of copy operation in the journal yet, then the dataSet's tables are not cleaned up during recovery.

This fix reverses the order of actions so that the copy operation is recorded in the journal before any new dataSet gets created. To achieve this, `HandleT#addNewDS` method had to be refactored to accept the `dataSetT` to be created as a parameter instead of calculating it within the method's body  based on the previous method parameters. Due to this change, all invocations of this method had to be adapted as well.

_**Hint:**_
The jobDB's journal table appears to be a poor man's method for performing transaction demarcation. To understand this a bit better, please consider first that long-running transactions are not preferred by `rudder-server` due to the involved lock contention. But, there are occasions where we need to perform multiple database operations as atomic operations (either succeed completely or fail completely). An operation entry in the journal table marks the start of such a transaction. If the operation fails to complete successfully, the server will crash (panic). Upon restart, the recovery process will kick in, it will scan the journal for incomplete operations and perform all necessary compensation actions, a.k.a. rollback. After recovery finishes, the server's services are safe to start again against a consistent state.

## Notion Link

https://www.notion.so/rudderstacks/Jobsdb-migration-bug-f71cff7202734ce0814b9f807777d008

## Type of change

## Checklist:

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -866,7 +866,7 @@ func (jd *HandleT) writerSetup(ctx context.Context) {
 
 	//If no DS present, add one
 	if len(jd.datasetList) == 0 {
-		jd.addDS(jd.newDataSetStruct(jd.computeNewIdxForAppend()), true)
+		jd.addNewDS(newDataSet(jd.tablePrefix, jd.computeNewIdxForAppend()))
 	}
 
 	jd.backgroundGroup.Go(misc.WithBugsnag(func() error {
@@ -1233,9 +1233,9 @@ func mapDSToLevel(ds dataSetT) (levelInt int, levelVals []int, err error) {
 	return len(levelVals), levelVals, nil
 }
 
-func (jd *HandleT) newDataSetStruct(dsIdx string) dataSetT {
-	jobTable := fmt.Sprintf("%s_jobs_%s", jd.tablePrefix, dsIdx)
-	jobStatusTable := fmt.Sprintf("%s_job_status_%s", jd.tablePrefix, dsIdx)
+func newDataSet(tablePrefix string, dsIdx string) dataSetT {
+	jobTable := fmt.Sprintf("%s_jobs_%s", tablePrefix, dsIdx)
+	jobStatusTable := fmt.Sprintf("%s_job_status_%s", tablePrefix, dsIdx)
 	return dataSetT{
 		JobTable:       jobTable,
 		JobStatusTable: jobStatusTable,
@@ -1243,26 +1243,26 @@ func (jd *HandleT) newDataSetStruct(dsIdx string) dataSetT {
 	}
 }
 
-func (jd *HandleT) addDS(ds dataSetT, isNew bool) {
+func (jd *HandleT) addNewDS(ds dataSetT) {
 	jd.logger.Infof("Creating new DS %+v", ds)
-
 	queryStat := stats.NewTaggedStat("add_new_ds", stats.TimerType, stats.Tags{"customVal": jd.tablePrefix})
-
 	queryStat.Start()
 	defer queryStat.End()
+	jd.createDS(ds, true)
+	// Tracking time interval between new ds creations. Hence calling end before start
+	if jd.isStatNewDSPeriodInitialized {
+		jd.statNewDSPeriod.End()
+	}
+	jd.statNewDSPeriod.Start()
+	jd.isStatNewDSPeriodInitialized = true
+}
 
-	defer func() {
-		if isNew {
-			// Tracking time interval between new ds creations. Hence calling end before start
-			if jd.isStatNewDSPeriodInitialized {
-				jd.statNewDSPeriod.End()
-			}
-			jd.statNewDSPeriod.Start()
-			jd.isStatNewDSPeriodInitialized = true
-		}
-	}()
-
-	jd.createDS(ds, isNew)
+func (jd *HandleT) addDS(ds dataSetT) {
+	jd.logger.Infof("Creating DS %+v", ds)
+	queryStat := stats.NewTaggedStat("add_new_ds", stats.TimerType, stats.Tags{"customVal": jd.tablePrefix})
+	queryStat.Start()
+	defer queryStat.End()
+	jd.createDS(ds, false)
 }
 
 func (jd *HandleT) computeNewIdxForAppend() string {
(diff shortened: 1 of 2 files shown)


chandumlg (approved):

jobsdb/jobsdb.go
```diff
@@ -1240,37 +1233,26 @@ func mapDSToLevel(ds dataSetT) (levelInt int, levelVals []int, err error) {
...
+		JobStatusTable: jobStatusTable,
+		Index:          dsIdx,
+	}
 }
 
-func (jd *HandleT) addNewDS(newDSType string, insertBeforeDS dataSetT) dataSetT {
-	jd.logger.Infof("Creating new DS of type %s before ds %s for %s jobsdb", newDSType, insertBeforeDS.Index, jd.tablePrefix)
-	var newDSIdx string
-	appendLast := newDSType == appendToDsList
+func (jd *HandleT) addDS(ds dataSetT, isNew bool) {
```
lvrach, line 1246: I don't like the fact that we are passing the isNew bool, it makes things less readable and harder to follow.

Could we have a better approach, for example to different methods:
* `addDS(ds dataSetT)`
* `addNewDs(ds dataSetT)`

jobsdb/jobsdb.go
```diff
@@ -1525,16 +1504,12 @@ func (jd *HandleT) createDS(appendLast bool, newDSIdx string) dataSetT {
...
-	if appendLast {
-		newDSWithSeqNumber := jd.setSequenceNumber(newDSIdx)
-		jd.JournalMarkDone(opID)
-		return newDSWithSeqNumber
-	}
-
-	//This is the migration case. We don't yet update the in-memory list till
+	//In case of a migration, we don't yet update the in-memory list till
 	//we finish the migration
+	if refreshList {
```
lvrach, line 1509: I can not easily follow this logic, why `refreshList` leads to `jd.setSequenceNumber(newDS.Index)` being called ?

lvrach (commented):
Would be realistic to add tests for this bug/change? I was thinking an integration test where journal table is partially created

jobsdb/jobsdb.go
```diff
@@ -873,7 +866,7 @@ func (jd *HandleT) writerSetup(ctx context.Context) {
 
 	//If no DS present, add one
 	if len(jd.datasetList) == 0 {
-		jd.addNewDS(appendToDsList, dataSetT{})
+		jd.addDS(jd.newDataSetStruct(jd.computeNewIdxForAppend()), true)
```
lvrach, line 869: Not in the scope of this PR, but I find `newDataSetStruct` method name a unusual:

```suggestion
		jd.addDS(jd.newDataSet(jd.computeNewIdxForAppend()), true)
```
or 

```suggestion
		jd.addDS(NewDataSet(jd.tablePrefix, jd.computeNewIdxForAppend()), true)
```

jobsdb/jobsdb.go
```diff
@@ -1240,37 +1233,26 @@ func mapDSToLevel(ds dataSetT) (levelInt int, levelVals []int, err error) {
...
+		Index:          dsIdx,
+	}
 }
 
-func (jd *HandleT) addNewDS(newDSType string, insertBeforeDS dataSetT) dataSetT {
-	jd.logger.Infof("Creating new DS of type %s before ds %s for %s jobsdb", newDSType, insertBeforeDS.Index, jd.tablePrefix)
-	var newDSIdx string
-	appendLast := newDSType == appendToDsList
+func (jd *HandleT) addDS(ds dataSetT, isNew bool) {
+	jd.logger.Infof("Creating new DS %+v", ds)
```
lvrach, line 1247: [minor] This `new` might be misleading if ds already created

jobsdb/jobsdb.go
```diff
@@ -1240,37 +1233,26 @@ func mapDSToLevel(ds dataSetT) (levelInt int, levelVals []int, err error) {
...
-		newDSIdx = jd.computeNewIdxForIntraNodeMigration(insertBeforeDS)
-	case insertForImport:
-		newDSIdx = jd.computeNewIdxForInterNodeMigration(insertBeforeDS)
-	default:
-		panic("Unknown usage for newDSType : " + newDSType)
-
-	}
-	jd.assert(newDSIdx != "", "newDSIdx is empty")
-
 	defer func() {
```
lvrach, line 1254: defer probably is not needed anymore, since `jd.createDS` is not returning. Except if we need this to run for panic.

jobsdb/jobsdb.go
```diff
@@ -1240,37 +1233,26 @@ func mapDSToLevel(ds dataSetT) (levelInt int, levelVals []int, err error) {
...
 
-	switch newDSType {
-	case appendToDsList:
-		newDSIdx = jd.computeNewIdxForAppend()
-	case insertForMigration:
-		newDSIdx = jd.computeNewIdxForIntraNodeMigration(insertBeforeDS)
-	case insertForImport:
-		newDSIdx = jd.computeNewIdxForInterNodeMigration(insertBeforeDS)
-	default:
-		panic("Unknown usage for newDSType : " + newDSType)
```
lvrach, line 1267: Just making sure:
We no longer need this, since we directly pass the index when calling `addDs`. Correct?

lvrach (approved):
The change to resolve the bug looks good to me.

I believe we can take the chance to improve the readability of this part further. As well as adding writing a test, for the recovery behavior.

However, if the fix needs to be urgently released I am cool with merging this and addressing the comments in another PR.

jobsdb/jobsdb.go
atzoum: > I can not easily follow this logic, why `refreshList` leads to `jd.setSequenceNumber(newDS.Index)` being called ?

Didn't understand this either. The refactoring's purpose was to leave business logic I couldn't comprehend intact!

jobsdb/jobsdb.go
atzoum: correct, index calculation is performed earlier

jobsdb/jobsdb.go
atzoum: will remove it, if there is a panic the less we should care about is updating a statistic about how often we create new datasets :)

PR #1799, opened 2022-04-05
fix: `pre_drop` table pileup issue

saurav-malani:
## Description of the issue

> Currently, when `BackupEnabled` is `true`, we try backing-up `pre_drop` tables to the configured bucket. And, drop the table only after the upload is successful. But, there is no upper limit on how many times (or for how long) we are going to retry in case of failure. Because of this, we end up retrying infinitely & never really drop the table if config is incorrect or if access key is has expired.

## Description of the change

>We have introduced an `exponentialBackoff` retry mechanism with maximum retry time of 10 minute, i.e. if upload didn't succeed in 10 min, retry will be stopped & `pre_drop` tables will be dropped.

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)

## Notion Link
https://www.notion.so/rudderstacks/Handle-pre_drop-table-upload-errors-d7c049e11ad84a779bee843e0d4c4c94

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -2989,10 +2989,7 @@ func (jd *HandleT) backupTable(ctx context.Context, backupDSRange *dataSetRangeT
 
 	jd.logger.Infof("[JobsDB] :: Uploading backup table to object storage: %v", tableName)
 	var output filemanager.UploadOutput
-	// get a file uploader
-	fileUploader, errored := jd.getFileUploader()
-	jd.assertError(errored)
-	output, err = jd.backupUploadWithExponentialBackoff(ctx, fileUploader.Upload, file, pathPrefixes...)
+	output, err = jd.backupUploadWithExponentialBackoff(ctx, file, pathPrefixes...)
 	if err != nil {
 		storageProvider := config.GetEnv("JOBS_BACKUP_STORAGE_PROVIDER", "S3")
 		jd.logger.Errorf("[JobsDB] :: Failed to upload table %v dump to %s. Error: %s", tableName, storageProvider, err.Error())
@@ -3006,7 +3003,11 @@ func (jd *HandleT) backupTable(ctx context.Context, backupDSRange *dataSetRangeT
 	return true, nil
 }
 
-func (jd *HandleT) backupUploadWithExponentialBackoff(ctx context.Context, fu func(context.Context, *os.File, ...string) (filemanager.UploadOutput, error), file *os.File, pathPrefixes ...string) (filemanager.UploadOutput, error) {
+func (jd *HandleT) backupUploadWithExponentialBackoff(ctx context.Context, file *os.File, pathPrefixes ...string) (filemanager.UploadOutput, error) {
+	// get a file uploader
+	fileUploader, errored := jd.getFileUploader()
+	jd.assertError(errored)
+
 	bo := backoff.NewExponentialBackOff()
 	bo.MaxInterval = time.Minute
 	bo.MaxElapsedTime = jd.maxBackupRetryTime
@@ -3015,19 +3016,12 @@ func (jd *HandleT) backupUploadWithExponentialBackoff(ctx context.Context, fu fu
 	var err error
 	var output filemanager.UploadOutput
 	backup := func() error {
-		output, err = fu(ctx, file, pathPrefixes...)
-		if err != nil {
-			return err
-		}
-		return nil
+		output, err = fileUploader.Upload(ctx, file, pathPrefixes...)
+		return err
 	}
 
-	if err = backoff.Retry(backup, boCtx); err != nil {
-		if bo.NextBackOff() == backoff.Stop {
-			return output, err
-		}
-	}
-	return output, nil
+	err = backoff.Retry(backup, boCtx)
+	return output, err
 }
 
 func (jd *HandleT) getBackupDSRange() *dataSetRangeT {
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -2772,6 +2772,10 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
 
 		// drop dataset after successfully uploading both jobs and jobs_status to s3
 		opID = jd.JournalMarkStart(backupDropDSOperation, opPayload)
+		//Currently, we retry uploading a table for sometime & if it fails. We only drop that table & not all `pre_drop` tables.
+		// So, in situation when new table creation rate is more than drop. We will still have pipe up issue.
+		// An easy way to fix this is, if at any point of time exponential retry fails then instead of just dropping that particular
+		// table drop all subsequent `pre_drop` table. As, most likely the upload of rest of the table will also fail with the same error.
 		jd.dropDS(backupDS, false)
 		jd.JournalMarkDone(opID)
 	}
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -407,7 +407,7 @@ func (jd *HandleT) getBackUpSettings() *BackupSettingsT {
 	config.RegisterBoolConfigVariable(false, &instanceBackupEnabled, false, fmt.Sprintf("JobsDB.backup.%v.enabled", jd.tablePrefix))
 	config.RegisterBoolConfigVariable(false, &instanceBackupFailedAndAborted, false, fmt.Sprintf("JobsDB.backup.%v.failedOnly", jd.tablePrefix))
 	config.RegisterStringConfigVariable(jd.tablePrefix, &pathPrefix, false, fmt.Sprintf("JobsDB.backup.%v.pathPrefix", jd.tablePrefix))
-	config.RegisterDurationConfigVariable(time.Minute*10, &jd.maxBackupRetryTime, false, time.Minute, "JobsDB.backup.maxRetry")
+	config.RegisterDurationConfigVariable(10, &jd.maxBackupRetryTime, false, time.Minute, "JobsDB.backup.maxRetry")
 	backupSettings := BackupSettingsT{BackupEnabled: masterBackupEnabled && instanceBackupEnabled,
 		FailedOnly: instanceBackupFailedAndAborted, PathPrefix: strings.TrimSpace(pathPrefix)}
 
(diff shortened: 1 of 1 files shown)


jobsdb/jobsdb.go
```diff
@@ -3024,7 +3046,7 @@ func (jd *HandleT) getBackupDSRange() dataSetRangeT {
 		}
 	}
 	if len(dnumList) == 0 {
-		return backupDSRange
+		return &backupDSRange
```
cisse21, line 3049: What was the need to change the return type to a pointer when we are not modifying this variable?

jobsdb/jobsdb.go
```diff
@@ -3007,7 +3005,31 @@ func (jd *HandleT) backupTable(backupDSRange dataSetRangeT, isJobStatusTable boo
 	return true, nil
 }
 
-func (jd *HandleT) getBackupDSRange() dataSetRangeT {
+func backupUploadWithExponentialBackoff(fu func(context.Context, *os.File, ...string) (filemanager.UploadOutput, error), ctx context.Context, file *os.File, pathPrefixes ...string) (filemanager.UploadOutput, error) {
+	bo := backoff.NewExponentialBackOff()
+	bo.MaxInterval = time.Minute
```
cisse21, line 3010: Let's have these variables configurable

jobsdb/jobsdb.go
saurav-malani: got deep source error, that passing large variable as value & should be passed as pointer.

cisse21: Not sure if this particular struct is a large variable but probably @lvrach and @atzoum can help out here

```
type dataSetRangeT struct {
	minJobID  int64
	maxJobID  int64
	startTime int64
	endTime   int64
	ds        dataSetT
}
```

and 

```
type dataSetT struct {
	JobTable       string `json:"job"`
	JobStatusTable string `json:"status"`
	Index          string `json:"index"`
}
```

jobsdb/jobsdb.go
saurav-malani: done

jobsdb/jobsdb.go
```diff
@@ -139,6 +140,7 @@ const (
 var globalDBHandle *sql.DB
 var masterBackupEnabled, instanceBackupEnabled, instanceBackupFailedAndAborted bool
 var pathPrefix string
+var maxBackupRetryTime time.Duration
```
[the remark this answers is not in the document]
saurav-malani: done

jobsdb/jobsdb.go
```diff
@@ -3007,7 +3006,31 @@ func (jd *HandleT) backupTable(backupDSRange dataSetRangeT, isJobStatusTable boo
 	return true, nil
 }
 
-func (jd *HandleT) getBackupDSRange() dataSetRangeT {
+func (jd *HandleT) backupUploadWithExponentialBackoff(ctx context.Context, fu func(context.Context, *os.File, ...string) (filemanager.UploadOutput, error), file *os.File, pathPrefixes ...string) (filemanager.UploadOutput, error) {
```
atzoum, line 3009: The method's signature is a bit awkward. I don't think we gain anything by creating the fileUploader outside and passing the upload function to this method. I also propose to further simplify this method (both signature and body), e.g.

```
func (jd *HandleT) backupUploadWithExponentialBackoff(ctx context.Context, file *os.File, pathPrefixes ...string) (filemanager.UploadOutput, error) {
	// get a file uploader
	fileUploader, errored := jd.getFileUploader()
	jd.assertError(errored)

	bo := backoff.NewExponentialBackOff()
	bo.MaxInterval = time.Minute
	bo.MaxElapsedTime = jd.maxBackupRetryTime
	boCtx := backoff.WithContext(bo, ctx)

	var err error
	var output filemanager.UploadOutput
	backupOperation := func() error {
		output, err = fileUploader.Upload(ctx, file, pathPrefixes...)
		return err
	}

	err = backoff.Retry(backupOperation, boCtx)
	return output, err
}
```

jobsdb/jobsdb.go
```diff
@@ -2992,8 +2992,7 @@ func (jd *HandleT) backupTable(backupDSRange dataSetRangeT, isJobStatusTable boo
 	// get a file uploader
 	fileUploader, errored := jd.getFileUploader()
```
atzoum, line 2993: we can skip creating the fileUploader here

jobsdb/jobsdb.go
atzoum: I don't have a strong opinion on this matter. I understand both the performance benefit of passing references instead of copying values around & the benefit of immutable structures. But anyway, a structure is shallow-copied when passed by value, correct? So immutability can only be achieved under conditions.

In the specific scenario, I tend to agree with @cisse21, but I am OK with keeping the linter happy too!

jobsdb/jobsdb.go
saurav-malani: done

jobsdb/jobsdb.go
```diff
@@ -3007,7 +3003,28 @@ func (jd *HandleT) backupTable(backupDSRange dataSetRangeT, isJobStatusTable boo
...
+	jd.assertError(errored)
+
+	bo := backoff.NewExponentialBackOff()
+	bo.MaxInterval = time.Minute
+	bo.MaxElapsedTime = jd.maxBackupRetryTime
+	boCtx := backoff.WithContext(bo, ctx)
+
+	var err error
+	var output filemanager.UploadOutput
+	backup := func() error {
```
atzoum, line 3018: minor: for simplicity you can consider using an anonymous function, e.g. 
```go
var err error
var output filemanager.UploadOutput
err = backoff.Retry(func() error {
    output, err = fileUploader.Upload(ctx, file, pathPrefixes...)
    return err
}, boCtx)
return output, err
```

atzoum (approved):
LGTM, only one last (minor) comment

jobsdb/jobsdb.go
saurav-malani: @chrikar pointed out last time that, this decreases the code readability. So, did it like this here. 
ref: https://github.com/rudderlabs/rudder-server/pull/1770#discussion_r837354087

atzoum: Yeap, that's why I mentioned this is minor, .i.e. subjective. What is readable to me might not be for someone else!

atzoum:
Please update the pull request's title according to our [new policy](https://www.notion.so/rudderstacks/Pull-Requests-40a4c6bd7a5e4387ba9029bab297c9e3)

jobsdb/jobsdb.go
```diff
@@ -3007,7 +3003,28 @@ func (jd *HandleT) backupTable(backupDSRange dataSetRangeT, isJobStatusTable boo
 	return true, nil
 }
 
-func (jd *HandleT) getBackupDSRange() dataSetRangeT {
+func (jd *HandleT) backupUploadWithExponentialBackoff(ctx context.Context, file *os.File, pathPrefixes ...string) (filemanager.UploadOutput, error) {
+	// get a file uploader
+	fileUploader, errored := jd.getFileUploader()
+	jd.assertError(errored)
```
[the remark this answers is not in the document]
lvrach: If we return an error instead of panic, the operation will be done with just an error. Maybe we don't want this.

```go
			err := jd.backupDS(ctx, backupDSRange)
			if err != nil {
				jd.logger.Errorf("[JobsDB] :: Failed to backup jobs table %v. Err: %v", backupDSRange.ds.JobStatusTable, err)
			}
			jd.JournalMarkDone(opID)
```

lvrach (approved):

jobsdb/jobsdb.go
```diff
@@ -405,7 +407,7 @@ func (jd *HandleT) getBackUpSettings() *BackupSettingsT {
 	config.RegisterBoolConfigVariable(false, &instanceBackupEnabled, false, fmt.Sprintf("JobsDB.backup.%v.enabled", jd.tablePrefix))
 	config.RegisterBoolConfigVariable(false, &instanceBackupFailedAndAborted, false, fmt.Sprintf("JobsDB.backup.%v.failedOnly", jd.tablePrefix))
 	config.RegisterStringConfigVariable(jd.tablePrefix, &pathPrefix, false, fmt.Sprintf("JobsDB.backup.%v.pathPrefix", jd.tablePrefix))
-
+	config.RegisterDurationConfigVariable(time.Minute*10, &jd.maxBackupRetryTime, false, time.Minute, "JobsDB.backup.maxRetry")
```
chandumlg, line 410: time.Minute*10 -> 10

jobsdb/jobsdb.go
```diff
@@ -2761,11 +2763,9 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
...
 		if isBackupConfigured() {
 			opID = jd.JournalMarkStart(backupDSOperation, opPayload)
-			success := jd.backupDS(backupDSRange)
-			if !success {
-				jd.removeTableJSONDumps()
-				jd.JournalMarkDone(opID)
-				continue
+			err := jd.backupDS(ctx, backupDSRange)
+			if err != nil {
+				jd.logger.Errorf("[JobsDB] :: Failed to backup jobs table %v. Err: %v", backupDSRange.ds.JobStatusTable, err)
```
chandumlg, line 2768: Don't we need this? 

jd.removeTableJSONDumps()

jobsdb/jobsdb.go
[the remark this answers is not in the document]
chandumlg: That should be okay right @lvrach? Instead of panicking.

chandumlg (changes requested):
Left a few comments.

jobsdb/jobsdb.go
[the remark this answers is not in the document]
lvrach: If we are not able to create the file manager, we will conclude the operation and print an error. Sounds good to me.

jobsdb/jobsdb.go
saurav-malani: since, in `backupTable` function we are already doing `os.Remove()`. So, I don't think we need to call `jd.removeTableJSONDumps()`.

jobsdb/jobsdb.go
```diff
@@ -2778,34 +2778,34 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
...
 	// return after backing up aboprted jobs if the flag is turned on
 	// backupDS is only called when BackupSettings.BackupEnabled is true
 	if jd.BackupSettings.FailedOnly {
 		jd.logger.Info("[JobsDB] ::  backupDS: starting backing up aborted")
-		_, err := jd.backupTable(backupDSRange, false)
+		_, err := jd.backupTable(ctx, backupDSRange, false)
 		if err != nil {
 			jd.logger.Errorf("[JobsDB] :: Failed to backup aborted jobs table %v. Err: %v", backupDSRange.ds.JobStatusTable, err)
-			return false
+			return err
```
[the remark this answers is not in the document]
saurav-malani: Good idea, but actually caller function is printing the log with table name details itself. So, to avoid redundant contents in error log, I am returning just `err`.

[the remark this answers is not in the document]
saurav-malani: removed `jd.logger.Errorf` from `backupDS` func. Since, caller function is printing the logs anyways.

jobsdb/jobsdb.go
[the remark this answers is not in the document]
saurav-malani: I just copied this piece of code into exponentialBackoff function without giving much thought to it. But, yes it makes sense to not assert. Removed it.

cisse21 (approved):

jobsdb/jobsdb.go
chandumlg: Got it.

chandumlg (approved):

PR #1800, opened 2022-04-05
fix(jobsdb): execute ANALYZE after completing batch jobsdb operations

atzoum:
## Description of the change

During a migration or import, new tables get created and populated in bulk with a large amount of rows.
It is possible that the AUTOVACUUM process does not manage to analyze those tables before we start querying them
which will result in the database's query planner making wrong decisions that may lead to never-ending queries.

This fix makes sure that whenever data is inserted in bulk in JobsDB, affected tables are explicitly ANALYZEd
before we release them into the wild

**Refactoring Notes**

1. `updateJobStatusDS` was renamed to `copyJobStatusDS`.

2. Both `storeJobsDS` and `storeJobsDSInTxn` methods where using a `bool` method parameter as a control flag. This control parameter was removed from both methods and methods were split to (1) `storeJobsDS` & `copyJobsDS` and (2) `storeJobsDSInTxn` & `copyJobsDSInTxn`.

3. Introduced `doInTransaction` as a common abstraction for demarcating transaction boundaries within other methods. _Please advise with regards to where we could move this method in order to be reused across the codebase._

4. `migrateJobs` was changed to perform copy operations to both tables (jobs & job_status) in a single transaction using `doInTransaction`.

## Notion Link

https://www.notion.so/rudderstacks/Batch-router-query-investigation-5cb1986fd3d547cc85964b97ade2e97a

## Type of change

## Checklist:

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -1780,30 +1780,30 @@ func (jd *HandleT) migrateJobs(srcDS dataSetT, destDS dataSetT) (noJobsMigrated 
 		0, GetQueryParamsT{StateFilters: getValidNonTerminalStates()})
 	jobsToMigrate := append(unprocessedList, retryList...)
 	noJobsMigrated = len(jobsToMigrate)
-	//Copy the jobs over. Second parameter (true) makes sure job_id is copied over
-	//instead of getting auto-assigned
-	err = jd.storeJobsDS(destDS, true, jobsToMigrate) //TODO: switch to transaction
-	jd.assertError(err)
 
-	//Now copy over the latest status of the unfinished jobs
-	var statusList []*JobStatusT
-	for _, job := range retryList {
-		newStatus := JobStatusT{
-			JobID:         job.JobID,
-			JobState:      job.LastJobStatus.JobState,
-			AttemptNum:    job.LastJobStatus.AttemptNum,
-			ExecTime:      job.LastJobStatus.ExecTime,
-			RetryTime:     job.LastJobStatus.RetryTime,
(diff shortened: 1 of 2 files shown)


jobsdb/jobsdb.go
```diff
@@ -1831,61 +1831,77 @@ func (jd *HandleT) postMigrateHandleDS(migrateFrom []dataSetT) error {
...
-	defer func() {
-		customValParamMap := make(map[string]map[string]map[string]struct{})
-		var workspaces []string //for bursting old cache
-		for _, job := range jobList {
-			if !misc.ContainsString(workspaces, job.WorkspaceId) {
-				workspaces = append(workspaces, job.WorkspaceId)
-			}
-			jd.populateCustomValParamMap(customValParamMap, job.CustomVal, job.Parameters, job.WorkspaceId)
-		}
+	defer jd.clearCache(ds, jobList)
```
atzoum, line 1841: @lvrach to consider for the future, since caching appears to be an aspect (horizontal concern). Instead of injecting caching logic in each and every method of `jobsdb.HandleT` we could use the decorator pattern (embedded struct) and handle all caching concerns on a different layer.
[This example](https://fabianlindfors.se/blog/decorators-in-go-using-embedded-structs/) also deals with caching 🤣

lvrach: Totally agree with that, I have proposed a jobsdb decorator a while back. Be prepared caused it might need some effort to introduce.

chandumlg (approved):
LGTM

cisse21 (approved):

PR #1803, opened 2022-04-05
Use legacy pickup algorithm on non multi-tenant environments. 

lvrach:
## Context

`multitenantStats` are responsible for two things:
1. Collect input/output jobs from processor/router
2. Determine the number of pickup jobs based on collected stats

Until [pending-event metrics PR](https://github.com/rudderlabs/rudder-server/pull/1736), there was a flag to enable/disable `multitenantStats` via `EnableMultitenancy`.

After pending-event metrics, the `multitenantStats` are always enabled.

## Refactor

This PR enables/disables only the `pickup jobs` functionality depending on the `EnableMultitenancy` flag.

## Notion Link

> [Notion Link
](https://www.notion.so/rudderstacks/Use-legacy-pickup-algorithm-on-non-multi-tenant-environments-7474d1ff5dac489a902ec64e7d8ad75b)

## Security

### DIFF
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -63,7 +63,7 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 	var batchRouterDB jobsdb.HandleT
 	var procErrorDB jobsdb.HandleT
 
-	var tenantRouterDB jobsdb.MultiTenantJobsDB = &jobsdb.MultiTenantLegacy{HandleT: &routerDB}
+	var tenantRouterDB jobsdb.MultiTenantJobsDB
 	var multitenantStats multitenant.MultiTenantI = multitenant.NOOP
 
 	pkgLogger.Info("Clearing DB ", options.ClearDB)
@@ -90,11 +90,17 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 
 		if config.GetBool("EnableMultitenancy", false) {
 			tenantRouterDB = &jobsdb.MultiTenantHandleT{HandleT: &routerDB}
+			multitenantStats = multitenant.NewStats(map[string]jobsdb.MultiTenantJobsDB{
+				"rt":       tenantRouterDB,
+				"batch_rt": &jobsdb.MultiTenantLegacy{HandleT: &batchRouterDB},
+			})
+		} else {
+			tenantRouterDB = &jobsdb.MultiTenantLegacy{HandleT: &routerDB}
+			multitenantStats = multitenant.WithLegacyPickupJobs(multitenant.NewStats(map[string]jobsdb.MultiTenantJobsDB{
+				"rt":       tenantRouterDB,
+				"batch_rt": &jobsdb.MultiTenantLegacy{HandleT: &batchRouterDB},
+			}))
 		}
-		multitenantStats = multitenant.NewStats(map[string]jobsdb.MultiTenantJobsDB{
-			"rt":       tenantRouterDB,
-			"batch_rt": &jobsdb.MultiTenantLegacy{HandleT: &batchRouterDB},
-		})
 	}
 
 	enableGateway := true
--- a/services/multitenant/legacy.go
+++ b/services/multitenant/legacy.go
@@ -4,15 +4,15 @@ import (
 	"time"
 )
 
-type legacy struct {
+type Legacy struct {
 	*MultitenantStatsT
 }
 
-func WithLegacyPickupJobs(stats *MultitenantStatsT) *legacy {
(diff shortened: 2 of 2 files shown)


services/multitenant/legacy.go
```diff
@@ -0,0 +1,19 @@
+package multitenant
+
+import (
+	"time"
+)
+
+type Legacy struct {
```
atzoum, line 7: It hadn't crossed my mind that there is an easy way to implement the decorator pattern in go using embedded structs, nice! 👍 

If I had to grumble about something, this might be naming, but I won't do it, since I haven't even reached to a consensus with myself about naming stuff in go yet!

This however appears to be a good opportunity to express my conviction, again, about `MultitenantI` interface, which is that only `GetRouterPickupJobs` method should be part of it, everything else appears to be part of something else.

services/multitenant/legacy.go
```diff
@@ -0,0 +1,19 @@
...
+	*MultitenantStatsT
+}
+
+func WithLegacyPickupJobs(stats *MultitenantStatsT) *Legacy {
+	return &Legacy{stats}
+}
+
+func (*Legacy) GetRouterPickupJobs(_ string, _ int, _ time.Duration, jobQueryBatchSize int, _ float64) (map[string]int, map[string]float64) {
```
atzoum, line 15: These two return types look like good examples why using custom types instead of generic types (maps, strings, ints, etc.) might be useful sometimes for us uninitiated reviewers!

1. A reviewer who hasn't come across this piece of code before, cannot possibly know what a `map[string]int` and a `map[string]float64` stand for.
2. There is no documentation on this or the interface's method about what this method is supposed to return.
3. Even if there was some documentation on the interface, it would be hard from this method to jump to the interface and read the documentation without trying too hard (there might be some magical shortcut in VSCode which can help you find interface methods implemented by this struct that I am unaware of).
4. If instead of these 2 maps, 2 specific types were created with meaningful names, the names of the types alone might be enough to give some valuable context. If these types were documented as well, it would be like living in GitHub reviewer's paradise!

Please consider all the above as my excuse for not being able to understand whether what you return here is valid or not!

atzoum (approved):

app/apphandlers/embeddedAppHandler.go
```diff
@@ -90,11 +90,17 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 
 		if config.GetBool("EnableMultitenancy", false) {
 			tenantRouterDB = &jobsdb.MultiTenantHandleT{HandleT: &routerDB}
+			multitenantStats = multitenant.NewStats(map[string]jobsdb.MultiTenantJobsDB{
+				"rt":       tenantRouterDB,
+				"batch_rt": &jobsdb.MultiTenantLegacy{HandleT: &batchRouterDB},
+			})
+		} else {
+			tenantRouterDB = &jobsdb.MultiTenantLegacy{HandleT: &routerDB}
+			multitenantStats = multitenant.WithLegacyPickupJobs(multitenant.NewStats(map[string]jobsdb.MultiTenantJobsDB{
```
atzoum, line 99: if `EnableMultitenancy` is `true`, why are we using `WithLegacyPickupJobs`?

atzoum, line 99: Sorry for missing this one during the previous sprint

cisse21 (approved):

services/multitenant/legacy.go
cisse21: @atzoum  Do you think named return types( example :` jobPerWorkspace map[string]int , pickedCustomerLatencyMap map[string]int`) would make sense or generic types ?

chandumlg (approved):

services/multitenant/legacy.go
atzoum: @cisse21 any kind of custom typing can be useful, depending on the context.
At least for well established business identifiers of a domain, e.g in our case `workspaceId`, `sourceId`, `destinationId`, `destinationType`, `jobsDbPrefix` etc. it is worthwhile using custom types. It is common for such identifiers to be used as method parameters, or in maps and other parametrized types, so the gains become apparent immediately, e.g.
It is more straightforward to have a 
```
map[workspaceId]map[destinationType]int
```
rather than a 
```
map[string]map[string]int
```

At a higher level, it might also make sense to create a custom type which encapsulates another parametrized type, e.g. 
```
type pendingEventsCounters map[workspaceId]map[destinationType]int
```
The above type discloses even more information about its intention. Only by looking at it you can understand that it is supposed to hold counters grouped by workspace id & destination type. So you don't really care about what is the actual name of the variable holding this type (if this is a return type it might not be named), you can rely only on the type itself to infer its purpose.

@lvrach what is your opinion?

PR #1809, opened 2022-04-06
Cleanup docker kafka network and skip failing TestDynamicClusterManager

chrikar:
## Description of the change

Our current tests on`master` are commonly failing for two reasons:
1. TestDynamicClusterManager is failing
2. Clickhouse network is unable to create a network with the desired subnet. This appears to be side-effect of not deleting each kafka network that we create during docker testing.
Note: The subnet used by the kafka network is now printed to facilitate troubleshooting in potential conflicts in the future.

## Notion Link

[Notion Link](https://www.notion.so/rudderstacks/Fix-failing-CI-due-to-TestDynamicClusterManager-and-docker-network-leftovers-20efc40b287b4dbeb2cbc99de5c53390)

## Type of change

## Checklist:

## Security

### DIFF
--- a/testhelper/destination/kafka.go
+++ b/testhelper/destination/kafka.go
@@ -27,9 +27,19 @@ func SetupKafka(pool *dockertest.Pool, d deferer) (*KafkaResource, error) {
 		return nil, fmt.Errorf("Could not create docker network: %w", err)
 	}
 	d.Defer(func() error {
-		// TODO delete network
+		if err := pool.Client.RemoveNetwork(network.ID); err != nil {
+			return fmt.Errorf("Could not remove kafka network: %w \n", err)
+		}
 		return nil
 	})
+
+	networkInfo, err := pool.Client.NetworkInfo(network.ID)
+	if err != nil {
+		return nil, fmt.Errorf("Could not get docker network info: %w", err)
+	}
+	for _, ipamConfig := range networkInfo.IPAM.Config {
+		log.Println("kafka network subnet:", ipamConfig.Subnet)
+	}
 
 	zookeeperPortInt, err := freeport.GetFreePort()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


testhelper/destination/kafka.go
```diff
@@ -27,10 +27,20 @@ func SetupKafka(pool *dockertest.Pool, d deferer) (*KafkaResource, error) {
...
 		return nil
 	})
 
+	networkInfo, err := pool.Client.NetworkInfo(network.ID)
+	if err != nil {
+		return nil, fmt.Errorf("Could not get docker network info: %w", err)
+	}
+	for _, ipamConfig := range networkInfo.IPAM.Config {
+		log.Println("kafka network subnet:", ipamConfig.Subnet)
+	}
```
lvrach, line 42: Do we need the log here?

lvrach (approved):
It was long due to deleting the network, nice!

testhelper/destination/kafka.go
chrikar: Need it, no. But there's still a chance we might get a subnet conflict with the clickhouse network, and this can help us understand if that is the case or not.

chandumlg (approved):

testhelper/destination/kafka.go
chrikar: @atzoum we're creating two docker networks now:
1. kafka_network: no particular settings, so will get an available subnet from docker service
2. clickhouse-network: requests the subnet `172.23.0.0/24` and fails if that subnet is not available, which is our current problem.

Ideally, we'd like to have no fixed subnets unless we really need them to. So, we want to discuss a bit with the warehouse devs about removing that if possible. Until that happens, this PR will mitigate our problem with failing CI tests.

PR #1813, opened 2022-04-06
WIP: Using leaky bucket approach

fracasula:
Just playing around.

### DIFF
--- a/utils/pubsub.go
+++ b/utils/pubsub.go
@@ -25,14 +25,14 @@ type EventBus struct {
 	// lastEvent holds the last event for each topic so that we can send it to new subscribers
 	lastEvent map[string]*DataEvent
 
-	subscribersMutex sync.RWMutex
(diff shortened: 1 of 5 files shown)


utils/pubsub.go
```diff
@@ -81,45 +81,39 @@ type subPublisher struct {
...
 	r.lastValue = data
-	r.lastValueLock.Unlock()
 
-	r.startedLock.Lock()
-	// start publish loop if not started
-	if !r.started {
-		go r.startLoop()
-		r.started = true
+	select {
+	case r.ping <- struct{}{}: // signals the startLoop that it has to read the value
```
atzoum, line 103: please explain the behaviour here. We wouldn't want publish method to block, the consumer's channel might take a long time to process an event

atzoum: actually this is pretty smart!

fracasula:
Closing since @atzoum has adopted this in his PR :+1:

PR #1819, opened 2022-04-07
Fix.recover panic

fracasula:
## Description of the change

Fixing panics on recover

## Notion Link

[> Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=2c36ad782c444537b8ae2880ed5c986c)

## Type of change

## Checklist:

## Security

### DIFF
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -59,7 +59,7 @@ type PackageStatusHandler interface {
 // RegisterAdminHandler is used by other packages to
 // expose admin functions over the unix socket based rpc interface
 func RegisterAdminHandler(name string, handler interface{}) {
-	instance.rpcServer.RegisterName(name, handler)
+	_ = instance.rpcServer.RegisterName(name, handler) // @TODO fix ignored error
 }
 
 // RegisterStatusHandler expects object implementing PackageStatusHandler interface
@@ -80,12 +80,12 @@ func Init() {
 		statushandlers: make(map[string]PackageStatusHandler),
 		rpcServer:      rpc.NewServer(),
 	}
-	instance.rpcServer.Register(instance)
+	_ = instance.rpcServer.Register(instance) // @TODO fix ignored error
 	pkgLogger = logger.NewLogger().Child("admin")
 }
 
 // Status reports overall server status by fetching status of all registered admin handlers
-func (a Admin) Status(noArgs struct{}, reply *string) (err error) {
+func (a Admin) Status(_ struct{}, reply *string) (err error) {
 	defer func() {
 		if r := recover(); r != nil {
 			pkgLogger.Error(r)
@@ -104,7 +104,7 @@ func (a Admin) Status(noArgs struct{}, reply *string) (err error) {
 }
 
 // PrintStack fetches stack traces of all running goroutines
-func (a Admin) PrintStack(noArgs struct{}, reply *string) (err error) {
+func (a Admin) PrintStack(_ struct{}, reply *string) (err error) {
 	defer func() {
 		if r := recover(); r != nil {
 			pkgLogger.Error(r)
@@ -129,8 +129,8 @@ func (a Admin) HeapDump(path *string, reply *string) (err error) {
 	if err != nil {
 		return err
 	}
-	defer f.Close()
-	pprof.Lookup("heap").WriteTo(f, 1)
+	defer func() { _ = f.Close() }()
+	_ = pprof.Lookup("heap").WriteTo(f, 1)
 	*reply = "Heap profile written to " + *path
 	return nil
 }
@@ -159,7 +159,7 @@ func (a Admin) StartCpuProfile(path *string, reply *string) (err error) {
 }
 
 // StopCpuProfile stops writing already cpu profile
-func (a Admin) StopCpuProfile(noArgs struct{}, reply *string) (err error) {
+func (a Admin) StopCpuProfile(_ struct{}, reply *string) (err error) {
 	defer func() {
 		if r := recover(); r != nil {
 			pkgLogger.Error(r)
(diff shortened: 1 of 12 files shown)


router/destinationReponseHandler.go
```diff
@@ -147,14 +147,7 @@ func (handler *JSONResponseHandler) IsSuccessStatus(respCode int, respBody strin
 //TXTResponseHandler -- start
 
 //IsSuccessStatus - returns the status code based on the response code and body
-func (handler *TXTResponseHandler) IsSuccessStatus(respCode int, respBody string) (returnCode int) {
```
atzoum, line 150: @chandumlg what was the initial intention here?

router/admin.go
```diff
@@ -127,7 +128,7 @@ func (r *RouterRpcHandler) GetDSStats(dsName string, result *string) (err error)
 	if err != nil {
 		return err
 	}
-	defer dbHandle.Close()
+	defer func() { _ = dbHandle.Close() }()
```
atzoum, line 131: What was the problem with the previous approach?

config/config.go
```diff
@@ -75,7 +75,7 @@ func UpdateConfig() {
 func watchForConfigChange() {
 	defer func() {
 		if r := recover(); r != nil {
-			err := fmt.Errorf("failed while trying to update Config Variables with Error: %w", r.(error))
+			err := fmt.Errorf("cannot to update Config Variables: %v", r)
```
atzoum, line 78: new text doesn't appear to be syntactically correct
```suggestion
			err := fmt.Errorf("cannot update Config Variables: %v", r)
```

warehouse/warehouse.go
```diff
@@ -1679,7 +1679,7 @@ func Start(ctx context.Context, app app.Interface) error {
 
 	runningMode := config.GetEnv("RSERVER_WAREHOUSE_RUNNING_MODE", "")
 	if runningMode == DegradedMode {
-		pkgLogger.Infof("WH: Running warehouse service in degared mode...")
+		pkgLogger.Infof("WH: Running warehouse service in degraded mode...")
```
atzoum, line 1682: @chandumlg is it safe to change existing log statements? I don't see why it shouldn't be unless we are performing any kind of log processing, e.g. for alerting or reporting purposes that relies on the existing log formats.

admin/admin.go
```diff
@@ -43,33 +42,33 @@ import (
...
+// output of Status() is expected to be json encodable by default
 type PackageStatusHandler interface {
 	Status() interface{}
 }
 
 // RegisterAdminHandler is used by other packages to
 // expose admin functions over the unix socket based rpc interface
 func RegisterAdminHandler(name string, handler interface{}) {
-	instance.rpcServer.RegisterName(name, handler)
+	_ = instance.rpcServer.RegisterName(name, handler) // @TODO fix ignored error
```
atzoum, line 62: Is the `@` prefix needed? On a side note, this is going to be one endless todo list!

atzoum (changes requested):

router/admin.go
fracasula: Ignoring errors bothers linters, in this case we could issue a warning or simply ignore. By making the assignment explicit we're making clear that we're OK with ignoring a DB connection closure error.

admin/admin.go
fracasula: Should work either way in most editors I guess. It's just a comment anyway.

router/destinationReponseHandler.go
fracasula: I'd :heart: to know too. It just proxies the code at the moment.

admin/admin.go
fracasula: In regards to the endless list, once I'm more confident with the codebase I'll try something. Ignoring a failure while registering a handler seems like it could be serious.

admin/admin.go
```diff
@@ -230,15 +229,15 @@ func (a Admin) GetFormattedEnv(env string, reply *string) (err error) {
...
+// StartServer starts an HTTP server listening on unix socket and serving rpc communication
 func StartServer(ctx context.Context) error {
 	tmpDirPath, err := misc.CreateTMPDIR()
 	if err != nil {
 		panic(err)
 	}
 	sockAddr := filepath.Join(tmpDirPath, "rudder-server.sock")
 	if err := os.RemoveAll(sockAddr); err != nil {
-		pkgLogger.Fatal(err)
+		pkgLogger.Fatal(err) // @TODO return?
```
lvrach, line 240: We can extend the scope of this one:  https://www.notion.so/rudderstacks/rudder-server-use-context-cancellation-instead-of-panic-to-handle-failures-during-server-startup-cf5d05ce758643f68eebe204eede37ce

cc @chrikar

lvrach (approved):

router/admin.go
kuldeep0020: If we are doing this, then why not put error log also. All the places?

router/destinationReponseHandler.go
chandumlg: Copy paste issue may be. Thanks for the fix.

router/admin.go
fracasula: > If we are doing this, then why not put error log also. All the places?

One bit at a time I'm sure we'll make it consistent :+1: 

In regards to the error log I'd say that usually nobody really does anything when they can't close a connection so it would just mean polluting the logs. If we try to close it more than once for example we could get `sql: duplicate driverConn close` but it's not a big deal. If you feel like having those in the logs can benefit someone then perhaps I can raise a `Warn` instead of an `Error`. Thoughts?

atzoum (approved):

router/admin.go
chandumlg: Agree with log pollution.

PR #1774, opened 2022-03-24
Integrate cluster manager

lvrach:
## Description of the change

> Introduced a new way to start and manage the mode of the server. Now JobsDB is being created with the help of new JobsDB creators and managed with the help of jobsDB interface. Processor and Routers are also created by the respective creators and managed by the help of their respective lifecycle manager interfaces. 
JobsDBs along with Processor and Router are passed to Cluster Manager in order to manage (start/stop) individual components of the server based on the expected state of the server or to help the server reach to the desired state (Normal/Degraded Mode).
The cluster manager needs one mode provide also based on which it gets the signal of mode changes. 
server_LCM

Configs: 
LegacyAppHandler   EnableMT
True                          True       Older App handlers With New Pickup Algo (In-mem stats being populated)
False                         False      New AppHandler with Old Pickup Algo (In-mem stats being populated)
False                         True        New AppHandler with New Pickup algo (In-mem stats being populated) –Hosted/end-goal
True                          False       Older AppHandler with older pickup Algo (In-mem stats being populated) – Current Default

## Notion Link

> [Notion Link](https://www.notion.so/rudderstacks/Integration-of-Cluster-Manager-451b5da33c2944919996532b8d8d1a0e)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added unit tests for the code
- [ ] I have made corresponding changes to the documentation

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #1839, opened 2022-04-12
feat(multi-tenant): etcd provider and config backend updates

lvrach:
## Overview
This PR contains two major changes:

* Introducing **etcd provider**
* Change backend config to support dynamic **workspaceID updates**

Screenshot 2022-04-13 at 9 15 21 PM

## Context

**Dynamic cluster** is responsible for changing:
1. **the mode** serve is in between NORMAL and DEGRADED (only accept requests, but no further processing)
2. **workspaceIDs** the server is responsible for handling

**without the server needing to be restarted** like before, by listening for changes using the state providers.

more details in the [PR it was introduced](https://github.com/rudderlabs/rudder-server/pull/1774)

In this PR **the etcd provider** is introduced, giving the ability to change mode or workspacesID from the values stored etcd server.

**Static provider** is also there, as a way to provide static behaviour in this dynamic system.

We are making this changes in order to enable multi-tenant support in rudder-server, you can get a better overall idea from the [design document](https://www.notion.so/rudderstacks/Multi-tenant-Overview-315a8a85027141f5a5119d5833ed0360).

## ETCD

_The diagram below explains the communication between components_
Screenshot 2022-04-13 at 9 00 11 PM

More specifically and following the defined protocol, it will watch the following keys:
```go
const (
	modeRequestKeyPattern        = `/%s/server/%s/mode`       // /<releaseName>/server/<serverIndex>/mode
	workspacesRequestsKeyPattern = `/%s/server/%s/workspaces` // /<releaseName>/server/<serverIndex>/workspaces
)
```

**NOTE:** `releaseName` and `serverIndex` are provided as env variables to rudder-server.

The etcd values are JSON and require acknowledgement.

## Notion Link

> [Notion Link](https://www.notion.so/rudderstacks/ETCD-integration-with-rudder-server-28d8c8d8989e4061b8042ecd8d00ab13)

## Type of change

## Checklist:

## Security


app/cluster/dynamic.go
```diff
@@ -125,6 +163,15 @@ func (d *Dynamic) stop() {
 	d.serverStopCountStat.Increment()
 }
 
+func (d *Dynamic) handleWorkspaceChange(workspaces string) error {
+	if d.currentWorkspaceIDs == workspaces {
+		return nil
+	}
+	d.BackendConfig.StopPolling()
+	d.BackendConfig.StartPolling(workspaces)
+	return nil
```
cisse21, line 172: We should also call waitForConfig at this point to ensure the config is completely received by the server

app/cluster/dynamic.go
```diff
@@ -62,33 +73,60 @@ func (d *Dynamic) init() {
...
+	serverModeChan := d.Provider.ServerMode(ctx)
+	workspaceIDsChan := d.Provider.WorkspaceIDs(ctx)
+
 	for {
 		select {
 		case <-ctx.Done():
 			if d.currentMode == servermode.NormalMode {
 				d.stop()
 			}
 			return nil
```
fracasula, line 97: Wouldn't it make sense to return `ctx.Err()` instead?

app/cluster/dynamic.go
```diff
@@ -125,6 +163,15 @@ func (d *Dynamic) stop() {
 	d.serverStopCountStat.Increment()
 }
 
+func (d *Dynamic) handleWorkspaceChange(workspaces string) error {
+	if d.currentWorkspaceIDs == workspaces {
+		return nil
+	}
+	d.BackendConfig.StopPolling()
```
fracasula, line 170: qq: is `StopPolling()` idempotent? What if we stop but polling was never started?

app/cluster/dynamic_test.go
```diff
@@ -167,6 +186,22 @@ func TestDynamicCluster(t *testing.T) {
 		require.True(t, errorDB.callOrder > router.callOrder)
 	})
 
+	t.Run("Update workspaceIDs", func(t *testing.T) {
+		chACK := make(chan bool)
```
fracasula, line 190:
```suggestion
		chACK := make(chan struct{})
```

By doing `bool` you're implicitly telling the reader that the value matters whereas you are just using the channel to signal here.

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+)
+
+const (
+	modeRequestKeyPattern        = `/%s/server/%s/mode`       // /<releaseName>/server/<serverIndex>/mode
+	workspacesRequestsKeyPattern = `/%s/server/%s/workspaces` // /<releaseName>/server/<serverIndex>/workspaces
+)
+
+var _ cluster.ModeProvider = &ETCDManager{}
+
+type ETCDConfig struct {
```
fracasula, line 41: Do we need to export this struct?

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+			manager.initErr = err
+			return
+		}
+		manager.Client = cli
+	})
+
+	return manager.initErr
+}
+
+func (manager *ETCDManager) Ping() error {
```
fracasula, line 123: This will ping only once so the name is misleading. One would expect something over the wire each time this method is called but that's not the case. Perhaps consider renaming (e.g. `HadConnected`) or consider making it ping for real.

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+		close(resultChan)
+	}()
+
+	return resultChan
+}
+
+func (manager *ETCDManager) Close() {
+	if manager.Client != nil {
+		manager.Client.Close()
```
fracasula, line 277: Could do `_ = manager.Client.Close()`

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func (manager *ETCDManager) prepareMode(raw []byte) servermode.ModeRequest {
+	var req modeRequestValue
+	err := json.Unmarshal(raw, &req)
+	if err != nil {
+		return servermode.ModeError(fmt.Errorf("unmarshal mode request: %w", err))
+	}
+
+	mode := servermode.Mode(req.Mode)
```
fracasula, line 137: This casting is redundant, you can do `mode := req.Mode`.

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	return servermode.NewModeRequest(
+		mode,
+		func() error {
+			ackValue, err := json.MarshalToString(modeAckValue{
+				Status: mode,
+			})
+			if err != nil {
+				return err
```
fracasula, line 149: Some context here would be appreciated, same on the error below (e.g. `return fmt.Errorf("could not marshal mode ack value: %w", err)`).

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	if len(resp.Kvs) == 0 {
+		return errChModeRequest(errors.New("no workspace found"))
+	}
+
+	resultChan <- manager.prepareMode(resp.Kvs[0].Value)
+
+	etcdWatchChan := manager.Client.Watch(ctx, modeRequestKey, clientv3.WithRev(resp.Header.Revision+1))
+	go func() {
```
fracasula, line 183: Shouldn't you propagate the context in these two for loops? The same goes for the `WorkspaceIDs` method.

fracasula (changes requested):
Just a preliminary review, I haven't gone through everything and I still have to fully understand about the general change/scope, but I'll try to finish the review today. Thanks for making the changes :+1: 

Sidenote: DeepSource is failing, I had rid us of all warnings a few days ago, please fix the build: https://deepsource.io/gh/rudderlabs/rudder-server/run/443433e0-887a-4ed9-bba6-61c71d2ed135/go/

app/cluster/dynamic.go
lvrach: Not sure about that, `ctx` cancellation is the expected way to terminate this component I don't see any benefit in returning an error.

app/cluster/dynamic_test.go
lvrach: Minor, but it makes sense.

app/cluster/state/etcd.go
lvrach: Yes, https://github.com/rudderlabs/rudder-server/blob/42bc59080fd22751d0b6ce399ed25bc0a39adcce/app/cluster/state/etcd_test.go#L98

`ETCDConfig` is the struct to be used if you want to configure the component.

app/cluster/state/etcd.go
lvrach: Good, point. I wanted to do a health check. Fixed.

app/cluster/dynamic.go
```diff
@@ -17,15 +21,20 @@ var (
...
+	ServerMode(ctx context.Context) <-chan servermode.ModeRequest
+	WorkspaceIDs(ctx context.Context) <-chan workspace.WorkspacesRequest
 }
 
 type lifecycle interface {
 	Start()
 	Stop()
 }
 
+type startStopPooling interface {
```
kuldeep0020, line 33: Maybe rename it and change the methods name as well

app/cluster/dynamic.go
```diff
@@ -125,6 +163,15 @@ func (d *Dynamic) stop() {
 	d.serverStopCountStat.Increment()
 }
 
+func (d *Dynamic) handleWorkspaceChange(workspaces string) error {
+	if d.currentWorkspaceIDs == workspaces {
```
kuldeep0020, line 167: should we have a validator for workspaces?

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+	"sync"
+	"time"
+
+	jsoniter "github.com/json-iterator/go"
+	"github.com/rudderlabs/rudder-server/app/cluster"
+	"github.com/rudderlabs/rudder-server/config"
+	"github.com/rudderlabs/rudder-server/utils/types/servermode"
+	"github.com/rudderlabs/rudder-server/utils/types/workspace"
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
```
kuldeep0020, line 17:
```suggestion
	etcdClientv3 "go.etcd.io/etcd/client/v3"
```

app/cluster/dynamic.go
lvrach: Not sure what kind of validator we can do for workspaceIDs, we are pretty ok to accept any value as far as understand. Maybe repeated values could be a problem ...

cc @cisse21

app/cluster/state/etcd.go
lvrach: I am not sure I understand what you mean by that.

Context cancellation will cause `.Watch` to close the channel, thus the loops will stop and the go routine will return.

app/cluster/state/etcd.go
lvrach: using `snakeCase` is not recommended for package names, I could use `etcdclientv3`, but it would be ugly/hard to read. 

How about just `etcd` ?

app/cluster/dynamic.go
lvrach: `StopPolling()` is not idempotent. The assumption is that BackendConfig will always be running before dynamicCluster is started. 

I am concerned about this as well, I have some plans to refactor backend config to improve on this. This PR is already too big, I would like to do it in a separate one.

app/cluster/dynamic.go
fracasula: > I don't see any benefit in returning an error.

That's just because nothing inside actually honours the context (e.g. `req.Ack()` is not `req.Ack(ctx)`).
Let's say that then we add one. They would propagate the error like most libraries do when a context is canceled so that then on the other side you can do stuff like `errors.Is(err, context.DeadlineExceeded)` and act accordingly (if you want to).

If you don't return it I think you might just be complicating things so that in some cases it is propagated and in some others not. Returning it is the most obvious (and expected) thing to do in my opinion although the problem might not present itself now, this code might be changed by someone else in the future and not consider this.

It is up to you but keep in mind that sometimes not doing the obvious thing is a cost that you might have to pay later.

app/cluster/state/etcd.go
fracasula: Right, but I saw no usage from outside this package and thought that one would rely on `func EnvETCDConfig() *ETCDConfig` to initialize it. It's minor ofc so if needed then ciao :man_shrugging:

app/cluster/state/etcd.go
fracasula: Indeed.

> When renaming an imported package, the local name should follow the same guidelines as package names (lower case, no under_scores or mixedCaps).

Reference: https://go.dev/blog/package-names?utm_source=dlvr.it&utm_medium=twitter

`etcd` or `etcd3` both look good to me.

app/cluster/state/etcd.go
fracasula: Did you push the fix? :thinking:

app/cluster/state/etcd.go
fracasula: Right, so it is coupled to `.Watch` which also happens to be a library we don't have control over since it's a 3rd party.

I meant that instead of:

```go
for watchResp := range etcdWatchChan {
```

One should do:

```go
for {
    select {
    case <-ctx.Done():
        return ctx.Err()
    case watchResp := <-etcdWatchCan:
        // do your magic
```

app/cluster/state/etcd.go
cisse21: it is used as clientV3 even on kubernetes repo so I would say we can just go ahead with what we have right now.

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+
+const (
+	modeRequestKeyPattern        = `/%s/server/%s/mode`       // /<releaseName>/server/<serverIndex>/mode
+	workspacesRequestsKeyPattern = `/%s/server/%s/workspaces` // /<releaseName>/server/<serverIndex>/workspaces
+)
+
+var _ cluster.ModeProvider = &ETCDManager{}
+
+type ETCDConfig struct {
+	ReleaseName          string
```
cisse21, line 42: we can safely change ReleaseName to Namespace as more people might be relevant with this name

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+			Endpoints:            manager.Config.Endpoints,
+			DialTimeout:          manager.Config.dialTimeout,
+			DialKeepAliveTime:    manager.Config.dialKeepAliveTime,
+			DialKeepAliveTimeout: manager.Config.dialKeepAliveTimeout,
+			DialOptions: []grpc.DialOption{
+				grpc.WithBlock(), // block until the underlying connection is up
+			},
+		})
+		if err != nil {
+			manager.initErr = err
```
cisse21, line 114: Why are we propogating an error which we encountered while trying to connect to etcd? Shouldn't we just exit the process and panic?

app/cluster/state/etcd.go
lvrach: I disagree, the producer is responsible for closing the channel. In this case, the producer is etcd client.
They are very clear in their documentation of how this should be used:
> 	// Watch watches on a key or prefix. The watched events will be returned
	// through the returned channel. If revisions waiting to be sent over the
	// watch are compacted, then the watch will be canceled by the server, the
	// client will post a compacted error watch response, and the channel will close.
	// If the requested revision is 0 or unspecified, the returned channel will
	// return watch events that happen after the server receives the watch request.
	// If the context "ctx" is canceled or timed out, returned "WatchChan" is closed,
	// and "WatchResponse" from this closed channel has zero events and nil "Err()".
	// The context "ctx" MUST be canceled, as soon as watcher is no longer being used,
	// to release the associated resources.
	//
	// If the context is "context.Background/TODO", returned "WatchChan" will
	// not be closed and block until event is triggered, except when server
	// returns a non-recoverable error (e.g. ErrCompacted).
	// For example, when context passed with "WithRequireLeader" and the
	// connected server has no leader (e.g. due to network partition),
	// error "etcdserver: no leader" (ErrNoLeader) will be returned,
	// and then "WatchChan" is closed with non-nil "Err()".
	// In order to prevent a watch stream being stuck in a partitioned node,
	// make sure to wrap context with "WithRequireLeader".

app/cluster/dynamic.go
cisse21: repeated values too should not be a problem but not sure what validations can be done if required as all of the data resides on control plane and we would have to request control plane to validate them(which is the source for scheduler anyways)

app/cluster/dynamic.go
```diff
@@ -17,15 +21,20 @@ var (
 )
 
 type ModeProvider interface {
```
atzoum, line 23: Now it is not only about `servermode`. What about
```
type EventChannelProvider interface {
	GetServerModeEventChannel(ctx context.Context) <-chan servermode.ChangeEvent
	GetWorkspaceIdsEventChannel(ctx context.Context) <-chan workspace.ChangeEvent
}
```

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+	Status string `json:"status"`
+}
+
+func EnvETCDConfig() *ETCDConfig {
+	endpoints := strings.Split(config.GetEnv("ETCD_HOSTS", "127.0.0.1:2379"), `,`)
+	releaseName := config.GetEnv("RELEASE_NAME", `multitenantv1`)
+	serverIndex := config.GetInstanceID()
+
+	envConfigOnce.Do(func() {
+		config.RegisterDurationConfigVariable(time.Duration(15), &etcdGetTimeout, true, time.Second, "etcd.getTimeout")
```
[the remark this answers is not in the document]
lvrach: Yeap hot reloading doesn't make sense.

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func errChModeRequest(err error) <-chan servermode.ModeRequest {
+	ch := make(chan servermode.ModeRequest, 1)
+	ch <- servermode.ModeError(err)
+	close(ch)
+	return ch
+}
+
+func (manager *ETCDManager) ServerMode(ctx context.Context) <-chan servermode.ModeRequest {
```
atzoum, line 163: Is this ever going to be used by more than one components/goroutines?
If the answer is no, return an error or panic the 2nd time, as a counter-measure for anybody trying to misuse it.
If the answer is yes, it wouldn't be efficient to create a new etcd watch every time a new consumer arrives. We can figure ways around this (typical Rx issue, we could use our existing `pubsub`), but I think the answer to the original question must be no, correct?

app/cluster/state/etcd.go
lvrach: I would reserve panic for something much worse / unexpected. A communication err is bad, and we are going to exit gracefully, by propagating the error all the way to the top.

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func (manager *ETCDManager) Ping() error {
+	if err := manager.init(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (manager *ETCDManager) prepareMode(raw []byte) servermode.ModeRequest {
```
atzoum, line 130: `unmarshalModeChangeEvent`

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func errChWorkspacesRequest(err error) <-chan workspace.WorkspacesRequest {
+	ch := make(chan workspace.WorkspacesRequest, 1)
+	ch <- workspace.WorkspacesError(err)
+	close(ch)
+	return ch
+}
+
+func (manager *ETCDManager) prepareWorkspace(raw []byte) workspace.WorkspacesRequest {
```
atzoum, line 212: `unmarshalWorskpaceChangeEvent`

app/cluster/state/etcd.go
fracasula: That's good enough :+1:

app/cluster/state/static.go
```diff
@@ -1,29 +1,44 @@
...
+	ch := make(chan servermode.ModeRequest, 1)
+	ch <- servermode.NewModeRequest(servermode.Mode(s.Mode), func() error {
+		return nil
 	})
-	s.ch <- servermode.WithACK(servermode.Mode(s.Mode), func() {})
-	return s.ch, nil
+
+	go func() {
+		<-ctx.Done()
+		close(ch)
```
fracasula, line 27: Correct me if I'm wrong but nobody else can push on `ch` (given the return type) and we are pushing a single message only in it, correct?

If that stands then I got two observations:
1. there is no need for the routine to close the channel given that it's buffered, you could just push into it, close it and return it. there is no use for a context as well so it could be renamed `_ context.Context` because pushing a single message there won't be a blocking operation anyway.
2. unless the channel with one element is needed for some reason, couldn't we just return the element without the channel at all?

utils/types/workspace/workspace.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func NewWorkspacesRequest(workspaceIDs []string, ack func() error) WorkspacesRequest {
+	return WorkspacesRequest{
+		workspaceIDs: workspaceIDs,
+		ack:          ack,
+	}
+}
+
+func WorkspacesError(err error) WorkspacesRequest {
```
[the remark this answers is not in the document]
lvrach: `ChangeEventError` ?

gateway/webhook/webhook.go
```diff
@@ -194,7 +194,7 @@ func (webhook *HandleT) RequestHandler(w http.ResponseWriter, r *http.Request) {
 			code = resp.statusCode
 		}
 		pkgLogger.Infof("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, code, resp.err)
-		http.Error(w, resp.err, code)
+    http.Error(w, resp.err, code)
```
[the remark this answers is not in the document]
lvrach: damn that was accidental

utils/types/workspace/workspace.go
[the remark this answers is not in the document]
atzoum: nice

config/backend-config/multitenant-workspace-config.go
```diff
@@ -63,13 +63,10 @@ func (workspaceConfig *MultiTenantWorkspaceConfig) Get(workspaces string) (Confi
...
 // getFromApi gets the workspace config from api
 func (workspaceConfig *MultiTenantWorkspaceConfig) getFromAPI(workspaceArr string) (ConfigT, bool) {
-	url := fmt.Sprintf("%s/hostedWorkspaceConfig?fetchAll=true", configBackendURL)
-	//TODO : Uncomment it and add tests once we have cluster managers ready
-	//To support existing multitenant behaviour
-	// url := fmt.Sprintf("%s/multitenantWorkspaceConfig?ids=[%s]", configBackendURL, workspaceArr)
-	// workspacesString := ""
-	// url = url + workspacesString
-	// url = url + "&fetchAll=true"
+	url := fmt.Sprintf("%s/multitenantWorkspaceConfig?ids=[%s]", configBackendURL, workspaceArr)
```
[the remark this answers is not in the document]
lvrach: I agree with this suggestion.

Regarding `multi-valued query parameters`:
1. I am not sure if the control plane will accept something like this.
2. Given the number of workspaceIDs we expect, it might have a big overhead. 

cc @cisse21

config/backend-config/multitenant-workspace-config.go
```diff
@@ -81,7 +78,7 @@ func (workspaceConfig *MultiTenantWorkspaceConfig) getFromAPI(workspaceArr strin
 
 	backoffWithMaxRetry := backoff.WithMaxRetries(backoff.NewExponentialBackOff(), 3)
 	err := backoff.RetryNotify(operation, backoffWithMaxRetry, func(err error, t time.Duration) {
-		pkgLogger.Errorf("[[ Workspace-config ]] Failed to fetch config from API with error: %v, retrying after %v", err, t)
+		pkgLogger.Errorf("[[MultiTenant-Workspace-config ]] Failed to fetch config from API with error: %v, retrying after %v", err, t)
```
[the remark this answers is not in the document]
lvrach: >  it because we are missing the logger's name in our log format?
no, we don't. Logger name is there

I don't know why we are doing it like this, cc @cisse21

app/cluster/dynamic.go
```diff
@@ -39,12 +48,14 @@ type Dynamic struct {
...
 
-	currentMode servermode.Mode
+	currentMode         servermode.Mode
+	currentWorkspaceIDs string
 
 	serverStartTimeStat  stats.RudderStats
 	serverStopTimeStat   stats.RudderStats
 	serverStartCountStat stats.RudderStats
 	serverStopCountStat  stats.RudderStats
+	BackendConfig        startStopPooling
```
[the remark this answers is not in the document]
lvrach: You mean why using this interface:
```go
	Stop()
	StartWithIDs(workspaces string)
	WaitForConfig(ctx context.Context) error
```

Instead of:
```
type BackendConfig interface {
	SetUp()
	Get(string) (ConfigT, bool)
	WaitForConfig(ctx context.Context) error
	Subscribe(channel chan pubsub.DataEvent, topic Topic)
	Stop()
	StartWithIDs(workspaces string)
}
```

app/cluster/dynamic.go
```diff
@@ -125,6 +163,15 @@ func (d *Dynamic) stop() {
 	d.serverStopCountStat.Increment()
 }
 
+func (d *Dynamic) handleWorkspaceChange(workspaces string) error {
```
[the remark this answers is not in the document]
lvrach: I agree I wanted to change that, but this change would propagate to a lot of work. I wanted to cut this PR short. I am planning a config specific refactor after this one is merged.

app/cluster/dynamic.go
lvrach: Is not very idiomatic for Go to use the `Get` / `Set` prefix. The same goes for return type as a suffix `EventChannel`.

In general shorter names are preferred in Go.

app/cluster/state/etcd.go
lvrach: , is `no`, but I find both proposals to be kind of an overkill. What we are trying to protect against, what is the benefit of adding this complexity?

app/cluster/state/etcd.go
lvrach: I am ok with `unmarshalMode`

app/cluster/state/static.go
lvrach: The expected behaviour for the provider is to keep the channel open until context is cancelled.

app/cluster/state/etcd.go
lvrach: I think there is some miss understanding. Context cancellation will cause `.Watch` to close the channel it returns:
> If the context "ctx" is canceled or timed out, returned "WatchChan" is closed

In our `.ServerMode`  / `.WorkspaceIDs` implementation when `etcdWatchChan` is closed, the for loops are exited and we close the `resultChan`. 

> Pushing to that channel while the context has been canceled can still happen

Yes, for me this is not necessarily bad. I understand that it can delay the shutdown, for me, this is like an inflight request (the scheduler request something, so we are going to do it and awk and then shutdown). If we see that this creates problems then we can ignore it.

app/cluster/dynamic.go
lvrach: So I am going without a validator here.

config/backend-config/multitenant-workspace-config.go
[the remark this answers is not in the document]
cisse21: The current way is the contract we have with control-plane ..... would require changes from CP side if we need to incorporate such changes. In production we were expecting not to have more than 200-300 workspaces on a single rudder-server so might not be a big overhead I guess

config/backend-config/multitenant-workspace-config.go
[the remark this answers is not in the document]
cisse21: It has been part of legacy code... we can get rid of them

config/backend-config/multitenant-workspace-config.go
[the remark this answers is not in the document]
cisse21: https://rudderlabs.slack.com/archives/C02A3UPEWMU/p1644392187564439

app/cluster/dynamic.go
atzoum: ignore get/set, only focus on the interface name :)

app/cluster/state/etcd.go
atzoum: misuse, due to missing documentation

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,289 @@
...
+		for watchResp := range etcdWatchChan {
+			if watchResp.Err() != nil {
+				resultChan <- servermode.ChangeEventError(watchResp.Err())
+				continue
+			}
+
+			for _, event := range watchResp.Events {
+				switch event.Type {
+				case mvccpb.PUT:
+					resultChan <- manager.prepareMode(event.Kv.Value)
```
cisse21, line 203: We should be appending the `/<instance-id>` to the ack_key which we get from ETCD.Where do you think we should be handling that?

lvrach: I would expect that the `ack_key` would contain the exact key and would not require any additions or modifications from rudder-server.

Since the rudder-server already watches an instance specific key (assuming <instance-id> =  <serverIndex>) `/<releaseName>/server/<serverIndex>/mode`, I believe scheduler should provide an ack_key that contains ` <instance-id>` already.

What do you think ?

cisse21: Yeah we should be able to incorporate this in scheduler I guess pretty easily from scheduler side. We can let it remain as id

kuldeep0020 (approved):
LGTM (This is not going to be used for now)

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,289 @@
...
+
+func (manager *ETCDManager) prepareWorkspace(raw []byte) workspace.ChangeEvent {
+	var req workspacesRequestsValue
+	err := json.Unmarshal(raw, &req)
+	if err != nil {
+		return workspace.ChangeEventError(err)
+	}
+
+	return workspace.NewWorkspacesRequest(
+		strings.Split(req.Workspaces, ","),
```
cisse21, line 230: Is the splitting just to let the reader know that we are serving multiple workspaces?

lvrach: Yes, I think is better to move this as an array for two reasons:
1. Readability: _I expect many ids, thus an array/slice_
2. Decouple transport layer with business logic: the fact that we are using commas to separate the workspace ids is specific to implementation, we might as well use a JSON array, or `;` - separated fields. To prevent the core logic of the system to care about this is better to do the parsing here.

app/cluster/dynamic.go
lvrach: what do you think of `StateChangeProvider`  ?

config/backend-config/multitenant-workspace-config.go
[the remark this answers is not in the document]
lvrach: removed

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,289 @@
...
+	return manager.initErr
+}
+
+// Ping ensures the connection to etcd is alive
+func (manager *ETCDManager) Ping() error {
+	if err := manager.init(); err != nil {
+		return err
+	}
+
+	_, err := manager.Client.Cluster.MemberList(context.Background())
```
cisse21, line 129: Why did we choose to use the MemberList function to determine the health of the cluster? I have seen few repos using a get call on any random key for the same functionality

config/backend-config/multitenant-workspace-config.go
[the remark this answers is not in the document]
lvrach: Only implemented the first suggestion for now.

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,289 @@
...
+	modeRequestKey := fmt.Sprintf(modeRequestKeyPattern, manager.Config.Namespace, manager.Config.ServerIndex)
+
+	resultChan := make(chan servermode.ChangeEvent, 1)
+	resp, err := manager.Client.Get(ctx, modeRequestKey)
+	if err != nil {
+		return errChModeRequest(err)
+	}
+
+	if len(resp.Kvs) == 0 {
+		return errChModeRequest(errors.New("no workspace found"))
```
cisse21, line 187: This might be a plausible case when we are using manual mode, We would be first scaling up the server with empty workspaces and then would move a workspace on to that empty server

cisse21: Should we be returning an error ? I feel we can just ignore and move ahead

config/backend-config/multitenant-workspace-config.go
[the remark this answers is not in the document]
lvrach: I have the following issue:
```
  Got: https://api.rudderlabs.com/multitenantWorkspaceConfig?fetchAll=true&ids=%5BtestToken%5D (string)
  Want: is equal to https://api.rudderlabs.com/multitenantWorkspaceConfig?ids=[testToken]&fetchAll=true (string)
  ```

File: app/cluster/state/etcd.go [hunk 32]
```diff
@@ -0,0 +1,289 @@
...
+				resultChan <- servermode.ChangeEventError(watchResp.Err())
+				continue
+			}
+
+			for _, event := range watchResp.Events {
+				switch event.Type {
+				case mvccpb.PUT:
+					resultChan <- manager.prepareMode(event.Kv.Value)
+				default:
+					resultChan <- servermode.ChangeEventError(fmt.Errorf("unknown event type %q", event.Type))
```
cisse21, line 205: Do we plan to handle different errors from etcd layer differently on the dynamic cluster manager?

cisse21: Some errors like this should be pretty harmless where as the watch error might need some handling

app/cluster/state/etcd.go
cisse21: This seems to be an incorrect error at server Mode... If we don't get any mode from ETCD, we should start by default in Degraded

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,289 @@
...
+	modeRequestKey := fmt.Sprintf(workspacesRequestsKeyPattern, manager.Config.Namespace, manager.Config.ServerIndex)
+
+	resultChan := make(chan workspace.ChangeEvent, 1)
+	resp, err := manager.Client.Get(ctx, modeRequestKey)
+	if err != nil {
+		return errChWorkspacesRequest(err)
+	}
+
+	if len(resp.Kvs) == 0 {
+		return errChWorkspacesRequest(errors.New("no workspace found"))
```
cisse21, line 257: This might be a plausible case when we are using manual mode, We would be first scaling up the server with empty workspaces and then would move a workspace on to that empty server

cisse21: Should we be returning an error ? I feel we can just ignore and move ahead

app/cluster/dynamic.go
```diff
@@ -62,33 +74,60 @@ func (d *Dynamic) init() {
...
 			if d.currentMode == servermode.NormalMode {
 				d.stop()
 			}
 			return nil
-		case newMode := <-modeChan:
-			d.logger.Infof("Got trigger to change the mode, new mode: %s, old mode: %s", newMode.Mode(), d.currentMode)
-			err = d.handleModeChange(newMode.Mode())
+		case req := <-serverModeChan:
+			if req.Err() != nil {
+				return req.Err()
```
cisse21, line 101: Might be a dumb question but where exactly do we handle this error? I see this being propogated till main but never acted upon or logged? Any reason why

app/cluster/dynamic.go
cisse21: One case which we might need to consider here is that the workspace string need not match but might have the same workspace ID's where this check would fail

atzoum (approved):
@lvrach I don't have anything major, you may treat all my comments as suggestions

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,289 @@
...
+	resultChan <- manager.unmarshalMode(resp.Kvs[0].Value)
+
+	etcdWatchChan := manager.Client.Watch(ctx, modeRequestKey, clientv3.WithRev(resp.Header.Revision+1))
+	go func() {
+		for watchResp := range etcdWatchChan {
+			if watchResp.Err() != nil {
+				resultChan <- servermode.ChangeEventError(watchResp.Err())
+				continue
+			}
+
```
cisse21, line 199: Do you think it makes sense to have an in-memory value for Revision which keeps getting and we only process those watch events which are higher than this particular Revision value? Just so to avoid processing stale keys

lvrach: Given how etcd client I don't think this is necessary. Get will fetch the latest revision and watch will start from that revision and keep fetching monotonically.  

Thus I don't think is possible to receive an older revision from what we already have.

app/cluster/dynamic.go
lvrach: My understanding is that an error like this should cause this application to stop. By propagating to main we ensure it will stop in a graceful way.

app/cluster/state/etcd.go
lvrach: Yeah, I was thinking about this as well. 

Some errors like the parsing could probably be handled. 

We need to ensure that scheduler and other components will be ok with server ignoring some values.

For simplicity I kept it like this, until we revisit our error handling in a higher level.

app/cluster/state/etcd.go
lvrach: If this is possible to happen we need to revisit this. In my view is quite unsafe to start a watch without getting the key (revision specifically) first.
 It might be possible between get and watch, for a put to happen. This is going to be lost if I remember correctly.

Let's rethink about it

app/cluster/state/etcd.go
lvrach: No reason in particular. I look inside the etcd implementation and looked that this is called quite frequently for coordination. 

No strong point on this, I can change it if you want.

app/cluster/dynamic.go
lvrach: I was thinking to introduce [sets](https://github.com/deckarep/golang-set) for storing and comparing workspaceIDs.

app/cluster/state/etcd.go
lvrach: What if we wait until the key is available ? This is the latest change I made. Should we do the same for workspaceIDs ?

app/cluster/state/etcd.go
cisse21: Contract should ensure that Scheduler does not return any other values apart from PUT value.So should be safe to ignore I guess

app/cluster/state/etcd.go
cisse21: We can start from the initial revision right? for this particular key. If i remember it correctly there is also something called mod revision at a key level

app/cluster/state/etcd.go
cisse21: Just wanted to know the idea behind it. Anything that helps establish that there is a connection should be good enough

app/cluster/state/etcd.go
cisse21: Yeah workspaces and mode behave in a similar fashion so it should be same for both I guess

lvrach: will return an empty workspace

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,289 @@
...
+			for _, event := range watchResp.Events {
+				switch event.Type {
+				case mvccpb.PUT:
+					resultChan <- manager.unmarshalMode(event.Kv.Value)
+				default:
```
lvrach, line 204: Add metric & log, do not return an error.

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,289 @@
...
+		config.RegisterDurationConfigVariable(time.Duration(30), &keepaliveTime, false, time.Second, "etcd.keepaliveTime")
+		config.RegisterDurationConfigVariable(time.Duration(10), &keepaliveTimeout, false, time.Second, "etcd.keepaliveTimeout")
+		config.RegisterDurationConfigVariable(time.Duration(20), &dialTimeout, false, time.Second, "etcd.dialTimeout")
+	})
+
+	return &ETCDConfig{
+		Endpoints:            endpoints,
+		Namespace:            namespace,
+		ServerIndex:          serverIndex,
+		etcdWatchTimeout:     etcdWatchTimeout,
```
lvrach, line 87:
```suggestion
		etcdGetTimeout:     etcdGetTimeout,
		etcdPutTimeout:     etcdPutTimeout,
		etcdWatchTimeout:     etcdWatchTimeout,
```

app/cluster/state/etcd.go
```diff
@@ -0,0 +1,289 @@
...
+	return servermode.NewChangeEvent(
+		mode,
+		func() error {
+			ackValue, err := json.MarshalToString(modeAckValue{
+				Status: mode,
+			})
+			if err != nil {
+				return fmt.Errorf("marshal ack value: %w", err)
+			}
+			_, err = manager.Client.Put(context.Background(), req.AckKey, ackValue)
```
fracasula, line 158: [minor] Perhaps we can convert some of these `context.Background()` to `context.TODO()` :+1:

app/cluster/dynamic.go
```diff
@@ -62,33 +74,60 @@ func (d *Dynamic) init() {
...
+	defer cancel()
+
+	serverModeChan := d.Provider.ServerMode(ctx)
+	workspaceIDsChan := d.Provider.WorkspaceIDs(ctx)
+
 	for {
 		select {
 		case <-ctx.Done():
 			if d.currentMode == servermode.NormalMode {
 				d.stop()
```
fracasula, line 96: [question] First of all I'm sorry if this was already explained :sweat_smile: 
Can you explain why on context cancelation if the current mode is `NORMAL` we "stop" and if not (meaning it is `DEGRADED`, right?) then we don't stop but we still return? Does it mean that the processor and router will still be running?

app/cluster/dynamic.go
fracasula: That would be better than a string comparison and could deal with both duplicates and IDs stored in a different order (e.g. `1,2` != `2,1`). Before going down that road I'd do a bunch of benchmarks to compare at least 2/3 different approaches for our use case.

app/cluster/dynamic.go
cisse21: If the mode is `DEGRADED` by default it would mean that both processor and router as stopped

app/cluster/state/etcd.go
lvrach: I have passed a proper context here

cisse21 (approved):

fracasula (approved):

PR #1847, opened 2022-04-13
WIP: feat: Gateway back pressure mechanism implementation

saurav-malani:
## Description of the change

>To avoid OOM kill of gateway in scenarios where we have a lot of sudden spike in gateway traffic or when DB is under pressure. We have introduced a back pressure mechanism that restrict on maximum number of active request in memory, by using counting Semaphore.

## Notion Link

> [Notion Link](https://www.notion.so/rudderstacks/Rate-limiting-in-Gateway-5a00130da4a545f58d147c81867afd4f)

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

## Security

### DIFF
--- a/gateway/configuration.go
+++ b/gateway/configuration.go
@@ -43,6 +43,10 @@ func loadConfig() {
 	config.RegisterDurationConfigVariable(time.Duration(10), &WriteTimeout, false, time.Second, []string{"WriteTimeout", "WriteTimeOutInSec"}...)
 	config.RegisterDurationConfigVariable(time.Duration(720), &IdleTimeout, false, time.Second, []string{"IdleTimeout", "IdleTimeoutInSec"}...)
 	config.RegisterIntConfigVariable(524288, &MaxHeaderBytes, false, 1, "MaxHeaderBytes")
+	//default is set with assumption that 4GB of RAM is present, with avg. event size=5KB & avg. batch size=10.
+	config.RegisterIntConfigVariable(80000, &maxClients, false, 1, "Gateway.maxClients")
+	config.RegisterBoolConfigVariable(true, &enableClientLimit, false, "Gateway.enableClientLimit")
+
 }
 
 // MaxReqSize is the maximum request body size, in bytes, accepted by gateway web handlers
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -76,7 +76,7 @@ type batchWebRequestT struct {
 
 var (
 	webPort, maxUserWebRequestWorkerProcess, maxDBWriterProcess, adminWebPort int
-	maxUserWebRequestBatchSize, maxDBBatchSize, MaxHeaderBytes                int
+	maxUserWebRequestBatchSize, maxDBBatchSize, MaxHeaderBytes, maxClients    int
 	userWebRequestBatchTimeout, dbBatchWriteTimeout                           time.Duration
 	enabledWriteKeysSourceMap                                                 map[string]backendconfig.SourceT
 	enabledWriteKeyWebhookMap                                                 map[string]string
@@ -84,7 +84,7 @@ var (
 	sourceIDToNameMap                                                         map[string]string
 	configSubscriberLock                                                      sync.RWMutex
 	maxReqSize                                                                int
-	enableRateLimit                                                           bool
+	enableRateLimit, enableClientLimit                                        bool
 	enableSuppressUserFeature                                                 bool
 	enableEventSchemasFeature                                                 bool
 	diagnosisTickerTime                                                       time.Duration
@@ -162,6 +162,7 @@ type HandleT struct {
 	addToBatchRequestQWaitTime                                 stats.RudderStats
 	trackSuccessCount                                          int
 	trackFailureCount                                          int
+	activeClientCount                                          int
 	requestMetricLock                                          sync.RWMutex
 	diagnosisTicker                                            *time.Ticker
 	webRequestBatchCount                                       uint64
@@ -177,9 +178,9 @@ type HandleT struct {
 	netHandle                                                  *http.Client
 	httpTimeout                                                time.Duration
 	httpWebServer                                              *http.Server
-
-	backgroundCancel context.CancelFunc
-	backgroundWait   func() error
+	client                                                     chan struct{}
+	backgroundCancel                                           context.CancelFunc
+	backgroundWait                                             func() error
 }
 
 func (gateway *HandleT) updateSourceStats(sourceStats map[string]int, bucket string, sourceTagMap map[string]string) {
@@ -656,7 +657,20 @@ func (gateway *HandleT) printStats(ctx context.Context) {
 	}
 }
 
-func (gateway *HandleT) stat(wrappedFunc func(http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
+func (gateway *HandleT) withMiddleware(fu func(http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
+	//this is to make sure that we don't have more than `maxClient` in-memory at any point of time. As, having more http client than `maxClient`
+	// may lead to gateway OOM kill.
+	if enableClientLimit {
+		gateway.client <- struct{}{}
+		defer func() { <-gateway.client }()
+	}
+
+	gateway.activeClientCount++
+	defer func() { gateway.activeClientCount-- }()
+
+	return gateway.withStat(fu)
+}
+func (gateway *HandleT) withStat(wrappedFunc func(http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
 		latencyStat := gateway.stats.NewSampledTaggedStat("gateway.response_time", stats.TimerType, map[string]string{"reqType": r.URL.Path})
 		latencyStat.Start()
@@ -1381,23 +1395,23 @@ func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
 	gateway.logger.Infof("Starting in %d", webPort)
 	srvMux := mux.NewRouter()
 	srvMux.Use(headerMiddleware)
-	srvMux.HandleFunc("/v1/batch", gateway.stat(gateway.webBatchHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/identify", gateway.stat(gateway.webIdentifyHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/track", gateway.stat(gateway.webTrackHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/page", gateway.stat(gateway.webPageHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/screen", gateway.stat(gateway.webScreenHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/alias", gateway.stat(gateway.webAliasHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/merge", gateway.stat(gateway.webMergeHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/group", gateway.stat(gateway.webGroupHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/batch", gateway.withMiddleware(gateway.webBatchHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/identify", gateway.withMiddleware(gateway.webIdentifyHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/track", gateway.withMiddleware(gateway.webTrackHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/page", gateway.withMiddleware(gateway.webPageHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/screen", gateway.withMiddleware(gateway.webScreenHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/alias", gateway.withMiddleware(gateway.webAliasHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/merge", gateway.withMiddleware(gateway.webMergeHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/group", gateway.withMiddleware(gateway.webGroupHandler)).Methods("POST")
 	srvMux.HandleFunc("/health", gateway.healthHandler).Methods("GET")
-	srvMux.HandleFunc("/v1/import", gateway.stat(gateway.webImportHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/audiencelist", gateway.stat(gateway.webAudienceListHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/import", gateway.withMiddleware(gateway.webImportHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/audiencelist", gateway.withMiddleware(gateway.webAudienceListHandler)).Methods("POST")
 	srvMux.HandleFunc("/", gateway.healthHandler).Methods("GET")
-	srvMux.HandleFunc("/pixel/v1/track", gateway.stat(gateway.pixelTrackHandler)).Methods("GET")
-	srvMux.HandleFunc("/pixel/v1/page", gateway.stat(gateway.pixelPageHandler)).Methods("GET")
+	srvMux.HandleFunc("/pixel/v1/track", gateway.withMiddleware(gateway.pixelTrackHandler)).Methods("GET")
+	srvMux.HandleFunc("/pixel/v1/page", gateway.withMiddleware(gateway.pixelPageHandler)).Methods("GET")
 	srvMux.HandleFunc("/version", gateway.versionHandler).Methods("GET")
-	srvMux.HandleFunc("/v1/webhook", gateway.stat(gateway.webhookHandler.RequestHandler)).Methods("POST", "GET")
-	srvMux.HandleFunc("/beacon/v1/batch", gateway.stat(gateway.beaconBatchHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/webhook", gateway.withMiddleware(gateway.webhookHandler.RequestHandler)).Methods("POST", "GET")
+	srvMux.HandleFunc("/beacon/v1/batch", gateway.withMiddleware(gateway.beaconBatchHandler)).Methods("POST")
 
 	if enableEventSchemasFeature {
 		srvMux.HandleFunc("/schemas/event-models", gateway.eventSchemaWebHandler(gateway.eventSchemaHandler.GetEventModels)).Methods("GET")
@@ -1410,10 +1424,10 @@ func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
 	}
 
 	//todo: remove in next release
-	srvMux.HandleFunc("/v1/pending-events", gateway.stat(gateway.pendingEventsHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/clear", gateway.stat(gateway.ClearHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/failed-events", gateway.stat(gateway.fetchFailedEventsHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/clear-failed-events", gateway.stat(gateway.clearFailedEventsHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/pending-events", gateway.withMiddleware(gateway.pendingEventsHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/clear", gateway.withMiddleware(gateway.ClearHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/failed-events", gateway.withMiddleware(gateway.fetchFailedEventsHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/clear-failed-events", gateway.withMiddleware(gateway.clearFailedEventsHandler)).Methods("POST")
 
 	c := cors.New(cors.Options{
 		AllowOriginFunc:  reflectOrigin,
@@ -1458,9 +1472,9 @@ func (gateway *HandleT) StartAdminHandler(ctx context.Context) error {
 	gateway.logger.Infof("Starting AdminHandler in %d", adminWebPort)
 	srvMux := mux.NewRouter()
 	srvMux.Use(headerMiddleware)
-	srvMux.HandleFunc("/v1/clear", gateway.stat(gateway.ClearHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/clear", gateway.stat(gateway.OperationStatusHandler)).Methods("GET")
-	srvMux.HandleFunc("/v1/pending-events", gateway.stat(gateway.pendingEventsHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/clear", gateway.withMiddleware(gateway.ClearHandler)).Methods("POST")
+	srvMux.HandleFunc("/v1/clear", gateway.withMiddleware(gateway.OperationStatusHandler)).Methods("GET")
+	srvMux.HandleFunc("/v1/pending-events", gateway.withMiddleware(gateway.pendingEventsHandler)).Methods("POST")
 
 	srv := &http.Server{
 		Addr:    ":" + strconv.Itoa(adminWebPort),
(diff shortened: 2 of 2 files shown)


gateway/gateway.go
```diff
@@ -1381,23 +1402,23 @@ func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
 	gateway.logger.Infof("Starting in %d", webPort)
 	srvMux := mux.NewRouter()
 	srvMux.Use(headerMiddleware)
```
atzoum, line 1404: IMO we should have 3 middlewares, with the following order:

1. `statsMiddleware`
3. `activeClientsMiddleware`
2. `headerMiddleware`

This also implies that we shall also clean our handler functions below!

gateway/gateway.go
```diff
@@ -162,6 +162,7 @@ type HandleT struct {
 	addToBatchRequestQWaitTime                                 stats.RudderStats
 	trackSuccessCount                                          int
 	trackFailureCount                                          int
+	activeClientCount                                          int
```
atzoum, line 165: use `int32`

gateway/gateway.go
```diff
@@ -656,7 +657,27 @@ func (gateway *HandleT) printStats(ctx context.Context) {
...
+	//NOTE: even after we hit the limit of `maxNoCLient`. New request will still pileup until we hit the backlog limit set by `listen` command.
+	// By default it is around 128.If a connection request arrives when the queue is full, the client may receive an error with an indication of ECONNREFUSED.
+	if enableClientLimit {
+		gateway.client <- struct{}{}
+		defer func() {
+			<-gateway.client
+		}()
+	}
+
+	gateway.activeClientCount++
```
atzoum, line 672: `atomic.AddInt32(&gateway.activeClientCount, 1)`

gateway/gateway.go
```diff
@@ -656,7 +657,27 @@ func (gateway *HandleT) printStats(ctx context.Context) {
...
+	defer func() {
+		gateway.activeClientCount--
```
atzoum, line 674: `atomic.AddInt32(&gateway.activeClientCount, -1)`

gateway/gateway.go
```diff
@@ -656,7 +657,27 @@ func (gateway *HandleT) printStats(ctx context.Context) {
...
 }
 
-func (gateway *HandleT) stat(wrappedFunc func(http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
+func (gateway *HandleT) withMiddleware(fu func(http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
+	//this is to make sure that we don't have more than `maxClient` in-memory at any point of time. As, having more http client than `maxClient`
+	// may lead to gateway OOM kill.
+	//NOTE: even after we hit the limit of `maxNoCLient`. New request will still pileup until we hit the backlog limit set by `listen` command.
+	// By default it is around 128.If a connection request arrives when the queue is full, the client may receive an error with an indication of ECONNREFUSED.
+	if enableClientLimit {
+		gateway.client <- struct{}{}
```
atzoum, line 666: Why is it better to let requests wait here, instead of responding immediately with `HTTP 503`? At this point, we already have `maxActiveClients` goroutines active and competing with each other, we don't need any more than that!

gateway/gateway.go
```diff
@@ -656,7 +657,27 @@ func (gateway *HandleT) printStats(ctx context.Context) {
 	}
 }
 
-func (gateway *HandleT) stat(wrappedFunc func(http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
+func (gateway *HandleT) withMiddleware(fu func(http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
+	//this is to make sure that we don't have more than `maxClient` in-memory at any point of time. As, having more http client than `maxClient`
+	// may lead to gateway OOM kill.
+	//NOTE: even after we hit the limit of `maxNoCLient`. New request will still pileup until we hit the backlog limit set by `listen` command.
```
atzoum, line 663: > backlog limit set by `listen` command

please explain

atzoum (changes requested):

gateway/configuration.go
```diff
@@ -43,6 +43,10 @@ func loadConfig() {
 	config.RegisterDurationConfigVariable(time.Duration(10), &WriteTimeout, false, time.Second, []string{"WriteTimeout", "WriteTimeOutInSec"}...)
 	config.RegisterDurationConfigVariable(time.Duration(720), &IdleTimeout, false, time.Second, []string{"IdleTimeout", "IdleTimeoutInSec"}...)
 	config.RegisterIntConfigVariable(524288, &MaxHeaderBytes, false, 1, "MaxHeaderBytes")
+	//default is set with assumption that 4GB of RAM is present, with avg. event size=5KB & avg. batch size=10.
+	config.RegisterIntConfigVariable(80000, &maxClients, false, 1, "Gateway.maxClients")
+	config.RegisterBoolConfigVariable(true, &enableClientLimit, false, "Gateway.enableClientLimit")
```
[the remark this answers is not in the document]
saurav-malani: I actually wanted to have a default values, so that anyone how is enabling doesn't have to think & calculate, what the appropriate number should be. And, can use the default one. But, since we don't have much clarity on what that number should be, I will remove two variable and combine them into one. As you mentioned.

gateway/gateway.go
saurav-malani: made changes with 3 middleware in chain.

gateway/gateway.go
saurav-malani: making it int32, so that we can do atomic addition & subtraction.

gateway/gateway.go
saurav-malani: yup, I had this in mind. But, forgot to make change. Anyways done.

gateway/gateway.go
saurav-malani: added `case` to handle this.

gateway/gateway.go
saurav-malani: changed the implementation. So, this isn't relevant anymore.

gateway/gateway.go
```diff
@@ -1380,24 +1372,24 @@ func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
...
-	srvMux.Use(headerMiddleware)
-	srvMux.HandleFunc("/v1/batch", gateway.stat(gateway.webBatchHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/identify", gateway.stat(gateway.webIdentifyHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/track", gateway.stat(gateway.webTrackHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/page", gateway.stat(gateway.webPageHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/screen", gateway.stat(gateway.webScreenHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/alias", gateway.stat(gateway.webAliasHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/merge", gateway.stat(gateway.webMergeHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/group", gateway.stat(gateway.webGroupHandler)).Methods("POST")
+	srvMux.Use(gateway.statMiddleware, gateway.maxActiveClientMiddleware, headerMiddleware)
```
atzoum, line 1375: @chandumlg with this change we are enabling stats collection for `/schemas` endpoints as well, I don't see any problem with that, on the contrary I believe it is desirable to collect metrics for all possible paths. Can you confirm?

gateway/gateway.go
```diff
@@ -1479,6 +1471,39 @@ func (gateway *HandleT) StartAdminHandler(ctx context.Context) error {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if maxActiveClients != 0 {
+			select {
+			case gateway.activeClient <- struct{}{}:
+				defer func() {
+					<-gateway.activeClient
+				}()
+				next.ServeHTTP(w, r)
+			default:
+				http.Error(w, "server is under pressure", response.GetStatusCode(response.TooManyRequestsInMemory))
```
atzoum, line 1491:
```suggestion
				http.Error(w, http.StatusText(http.StatusServiceUnavailable), http.StatusServiceUnavailable)
```

do we have any specific reason not to prefer using the standard `net/http` status and messages?

gateway/gateway.go
```diff
@@ -1479,6 +1471,39 @@ func (gateway *HandleT) StartAdminHandler(ctx context.Context) error {
...
+				return
+			}
+		}
+	})
+}
+
+func (gateway *HandleT) statMiddleware(next http.Handler) http.Handler {
```
atzoum, line 1498: @chandumlg @lvrach  Is there a place where we collect response status code metrics per path? IMO this is the place to capture such metrics so that we don't (unintentionally) miss anything.

So to sum up, I propose that:
1. timer stat should be enhanced with the `code` tag
2. introduce a new counter stat `gateway.response_count` using the same set of tags (`reqType`, & `code`)

_Note:_
With regards to naming, I would normally expect to see the following names being used here:
```
gateway.request_time{path="/v1/identify", code="200" }
gateway.request_count{path="/v1/identify", code="200" }
```

atzoum (changes requested):

gateway/gateway.go
chandumlg: Yeah no problem. Let's enable.

gateway/gateway.go
chandumlg: Right now, we don't have response status code in the metrics. 

Can we do this in another pr, where we have a task to clean up gateway response codes? CC @fracasula

gateway/gateway.go
```diff
@@ -1479,6 +1471,39 @@ func (gateway *HandleT) StartAdminHandler(ctx context.Context) error {
 	return g.Wait()
 }
 
+func (gateway *HandleT) maxActiveClientMiddleware(next http.Handler) http.Handler {
```
atzoum, line 1474:
```suggestion
func (gateway *HandleT) maxActiveClientMiddleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		atomic.AddInt32(&gateway.activeClientCount, 1)
		defer func() {
			atomic.AddInt32(&gateway.activeClientCount, -1)
		}()
		if maxActiveClients != 0 {
			//this is to make sure that we don't have more than `maxClient` in-memory at any point of time. As, having more http client than `maxClient`
			// may lead to gateway OOM kill.
			select {
			case gateway.activeClient <- struct{}{}:
				defer func() {
					<-gateway.activeClient
				}()
			default:
				http.Error(w, "server is under pressure", response.GetStatusCode(response.TooManyRequestsInMemory))
				return
			}
		}
		next.ServeHTTP(w, r)
	})
}
```

gateway/gateway.go
saurav-malani: actually, for rest of the place in gateway code I saw that we have defined a response package & we were using it to respond. So, I also did the same way. But, just noticed that we only calling `response.GetStatus` in gateway and not `response.GetStatusCode`. So, modified it.

gateway/gateway.go
saurav-malani: great point, I have modified my implementation. Thanks!

gateway/gateway.go
```diff
@@ -1479,6 +1471,39 @@ func (gateway *HandleT) StartAdminHandler(ctx context.Context) error {
...
+		atomic.AddInt32(&gateway.activeClientCount, 1)
+		defer func() {
+			atomic.AddInt32(&gateway.activeClientCount, -1)
+		}()
+		if maxActiveClients != 0 {
+			select {
+			case gateway.activeClient <- struct{}{}:
+				defer func() {
+					<-gateway.activeClient
+				}()
```
lvrach, line 1488: I would separate counting and limit into a separate middleware.
I would also make them self contain and independent from the gateway component, not for reusability but to avoid accidental and unnecessary coupling.

sudo code:
```go
package middleware
func  LimitConcurrentRequests(maxRequest int) func(http.Handler) http.Handler {
    activeClient := make(chan struct{}, maxRequest)
    return func(next.http.Handler) http.Handler {
          ....
          next.ServeHTTP(w, r)
     }


     }
}
````

File: gateway/response/response.go [hunk 13]
```diff
@@ -22,6 +22,9 @@ const (
 	RequestBodyReadFailed = "Failed to read body from request"
 	//RequestBodyTooLarge - Request size exceeds max limit
 	RequestBodyTooLarge = "Request size exceeds max limit"
+	//TooManyRequestsInMemory - This can be because of sudden spike in requests or maybe DB is under pressure, because of which DB write
+	//is taking more time than usual, because of which currently we have too many jobs in-memory.
+	TooManyRequestsInMemory = "Too many request in-memory"
```
atzoum, line 27: we shouldn't disclose this kind of information to our clients

gateway/gateway.go
lvrach: This will make it easy to test as well, please include tests

gateway/gateway.go
```diff
@@ -1479,6 +1471,39 @@ func (gateway *HandleT) StartAdminHandler(ctx context.Context) error {
...
+				http.Error(w, response.GetStatus(response.TooManyRequestsInMemory), response.GetStatusCode(response.TooManyRequestsInMemory))
+				return
+			}
+		}
+		next.ServeHTTP(w, r)
+	})
+}
+
+func (gateway *HandleT) statMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
lvrach, line 1499: Similar to the above we can avoid coupling with gateway, the only dependency is the stats package. Number of active requests can be counted here.

gateway/response/response.go
```diff
@@ -76,6 +79,8 @@ func loadStatusMap() {
 	statusMap[RequestBodyTooLarge] = ResponseStatus{message: RequestBodyTooLarge, code: http.StatusRequestEntityTooLarge}
 	statusMap[InvalidWriteKey] = ResponseStatus{message: InvalidWriteKey, code: http.StatusUnauthorized}
 	statusMap[InvalidJSON] = ResponseStatus{message: InvalidJSON, code: http.StatusBadRequest}
+	statusMap[TooManyRequestsInMemory] = ResponseStatus{message: TooManyRequestsInMemory, code: http.StatusServiceUnavailable}
```
atzoum, line 82:
```suggestion
	statusMap[TooManyRequestsInMemory] = ResponseStatus{message: http.StatusText(http.StatusServiceUnavailable), code: http.StatusServiceUnavailable}
```

atzoum (changes requested):

lvrach (changes requested):

gateway/response/response.go
saurav-malani: I didn't know we have `http.StatusText()`, this makes use of `statusMap[]` redundant here. So, removed it.

lvrach:
removed the Ready for Review, until changes are addressed/pushed

gateway/gateway.go
saurav-malani: modified as per your feedback.

middleware/middleware.go
```diff
@@ -0,0 +1,75 @@
...
+	"context"
+	"net/http"
+	"strings"
+	"sync/atomic"
+	"time"
+
+	"github.com/rudderlabs/rudder-server/services/stats"
+)
+
+func MaxConcurrentRequestsMiddleware(maxConcurrentRequests int) func(http.Handler) http.Handler {
```
lvrach, line 13: [minor] More compact name, no need to include `Middleware` in the function name since it is on the package.

Think of usage `middleware.MaxConcurrentRequestsMiddleware( )`, it can be a bit excessive. 

```suggestion
func LimitConcurrentRequests(maxRequests int) func(http.Handler) http.Handler {
```

middleware/middleware.go
```diff
@@ -0,0 +1,75 @@
...
+			defer atomic.AddInt32(&concurrentRequests, -1)
+
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+//Currently sets the content-type only for eventSchemas, health responses.
+//Note : responses via http.Error aren't affected. They default to text/plain
+func HeaderMiddleware() func(http.Handler) http.Handler {
```
lvrach, line 66: [minor] 
1. Include the function name in the comment.
2. Suggest a more appropriate title. `ContentType`

```suggestion
// ContentType Currently sets the content-type only for eventSchemas, health responses.
// Note : responses via http.Error aren't affected. They default to text/plain
func ContentType() func(http.Handler) http.Handler {
```

middleware/middleware_test.go
```diff
@@ -0,0 +1,77 @@
...
+		{
+			name:                  "max concurrent request: 1",
+			maxConcurrentRequests: 1,
+			totalReq:              10,
+		},
+		{
+			name:                  "max concurrent request: 10",
+			maxConcurrentRequests: 10,
+			totalReq:              10,
+		},
```
lvrach, line 41: [minor] and subjective.  I would prefer to see expected responses here, instead of computing them in:
```go
require.Equal(t, tt.totalReq-tt.maxConcurrentRequests, resp503, "actual 503 resp different than expected.")
```

lvrach (approved):

atzoum (approved):
LGTM

middleware/middleware.go
```diff
@@ -0,0 +1,75 @@
...
+		// may lead to gateway OOM kill.
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if maxConcurrentRequests != 0 {
+				select {
+				case requests <- struct{}{}:
+					defer func() {
+						<-requests
+					}()
+				default:
+					http.Error(w, http.StatusText(http.StatusServiceUnavailable), http.StatusServiceUnavailable)
```
fracasula, line 27: What does it happen if we keep returning 503 for too long? Would this affect some load balancing mechanism in the middle? Also, is this monitored somehow?

fracasula (changes requested):
Looks fairly good, I've added very few comments, the blocking one is the one about doing `atomic.LoadInt32(&concurrentRequests)`. Let me know if it doesn't make sense of course.

middleware/middleware.go
```diff
@@ -0,0 +1,75 @@
...
+func StatMiddleware(ctx context.Context) func(http.Handler) http.Handler {
+	var concurrentRequests int32
+	activeClientCount := stats.NewStat("gateway.concurrent_requests_count", stats.GaugeType)
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(10 * time.Second):
+				activeClientCount.Gauge(concurrentRequests)
```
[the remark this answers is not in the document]
lvrach: Good point!

middleware/middleware.go
saurav-malani: done

middleware/middleware.go
saurav-malani: done

middleware/middleware_test.go
```diff
@@ -0,0 +1,77 @@
...
+					respRecorder := httptest.NewRecorder()
+					req := httptest.NewRequest("GET", "http://testing", nil)
+
+					handlerToTest.ServeHTTP(respRecorder, req)
+					if respRecorder.Code == 200 {
+						resp200++
+					} else if respRecorder.Code == 503 {
+						resp503++
+					} else {
+						randomResp++
```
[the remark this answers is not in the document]
saurav-malani: oh, complete missed it.

middleware/middleware_test.go
saurav-malani: the reason I have written it like this is, to make it more intuitive that anything over allowed concurrentRequests would be 503. ` tt.totalReq-tt.maxConcurrentRequests` mathematically indicates that. @lvrach WDYT?

gateway/gateway.go
```diff
@@ -1380,24 +1371,24 @@ func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
...
-	srvMux.Use(headerMiddleware)
-	srvMux.HandleFunc("/v1/batch", gateway.stat(gateway.webBatchHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/identify", gateway.stat(gateway.webIdentifyHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/track", gateway.stat(gateway.webTrackHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/page", gateway.stat(gateway.webPageHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/screen", gateway.stat(gateway.webScreenHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/alias", gateway.stat(gateway.webAliasHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/merge", gateway.stat(gateway.webMergeHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/group", gateway.stat(gateway.webGroupHandler)).Methods("POST")
+	srvMux.Use(middleware.StatMiddleware(ctx), middleware.MaxConcurrentRequestsMiddleware(maxConcurrentRequests), middleware.HeaderMiddleware())
```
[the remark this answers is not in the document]
saurav-malani: makes sense, modified.

gateway/gateway.go
```diff
@@ -1457,10 +1448,10 @@ func (gateway *HandleT) StartAdminHandler(ctx context.Context) error {
 
 	gateway.logger.Infof("Starting AdminHandler in %d", adminWebPort)
 	srvMux := mux.NewRouter()
-	srvMux.Use(headerMiddleware)
-	srvMux.HandleFunc("/v1/clear", gateway.stat(gateway.ClearHandler)).Methods("POST")
-	srvMux.HandleFunc("/v1/clear", gateway.stat(gateway.OperationStatusHandler)).Methods("GET")
-	srvMux.HandleFunc("/v1/pending-events", gateway.stat(gateway.pendingEventsHandler)).Methods("POST")
+	srvMux.Use(middleware.StatMiddleware(ctx), middleware.MaxConcurrentRequestsMiddleware(maxConcurrentRequests), middleware.HeaderMiddleware())
```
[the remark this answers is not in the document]
saurav-malani: modified

middleware/middleware.go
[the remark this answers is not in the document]
saurav-malani: complete makes sense, didn't knew something like `LoadInt32` exists. 😄

middleware/middleware_test.go
```diff
@@ -57,9 +58,9 @@ func TestMaxConcurrentRequestsMiddleware(t *testing.T) {
 
 					handlerToTest.ServeHTTP(respRecorder, req)
 					if respRecorder.Code == 200 {
-						resp200++
+						atomic.AddInt32(&resp200, 1)
 					} else if respRecorder.Code == 503 {
-						resp503++
+						atomic.AddInt32(&resp503, 1)
 					} else {
 						randomResp++
```
fracasula, line 65: Perhpas you have forgotten the atomic part here :slightly_smiling_face:

middleware/middleware.go
saurav-malani: so essentially, with this PR we want to make sure that gateway doesn't get OOM killed, if DB write is slow for some reason in comparison to ingress. So, ideally we shouldn't return too many 503's for too long. Only for sometime, when DB is under pressure, or if someone sends too many requests in a short period of time, & I believe rejecting in this case is fine. 
But, we are planning to keep a count of 503's returned & maybe add an alert for the same, if too many 503 are returned in a separate PR.
Also, I don't think this will affect load balancing mechanism in the middle. Though, not sure here. 

@atzoum @lvrach @chandumlg any thoughts?

middleware/middleware_test.go
saurav-malani: Oops, fixed.

middleware/middleware.go
atzoum: capturing statistics of response codes per request path (information typically available in an access log) can be part of another pull request.
Alerts on top of such metrics can follow at a latter phase

fracasula: Given the nature of this PR (rejecting requests) and given that it is a new behaviour we are introducing I'd suggest adding the metrics now. Also considering how easy you added a gauge a few lines below with `stats.DefaultStats`. Whatever the outcome whenever something like this is introduced I think it's beneficial to be able to see immediately the effects of such change. It is also not clear to @saurav-malani whether this might affect load balancing of something else in the middle so if we end up with an incident, having someone respond and see a big bunch of 503s could help them with the incident resolution.

fracasula: Unless of course we are already collecting response codes statistics in some other layer (at Netdata we were). @lvrach ?

atzoum: @fracasula we will selectively enable the limit on specific customer(s), by default it is disabled. The existing metric, number of active connections should be enough for monitoring & troubleshooting at this stage.

fracasula (approved):

chandumlg (approved):

PR #417, opened 2020-07-07
Close all warhouse connections

psrikanth88:
**Fixes** # (*issue*)

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added unit tests for the code
- [ ] I have made corresponding changes to the documentation

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-server/417)

PR #1869, opened 2022-04-20
feat(core): add support for transient sources

atzoum:
# Description

A **Source** can now be configured as `transient`. The system treats events of transient sources in the following way:

1. Events will not be backed up to the object storage.
2. Failed processor events will not be backed up in `rudder-processor-errors` either, this process is a.k.a `stash` internally.
3. Sample events will not be sent to the reporting service.

## Design notes

- Introduced `transientsource.Service` which is responsible for providing all required services pertaining to transient sources. There are 3 ways you can create a `transientsource.Service`
   1. `transientsource.NewService` - creates a service which will listen for configuration changes and update its list of transient source ids automatically.
   2. `transientsource.NewStaticService` creates a service which uses a static list of transient source ids.
   3. `transientsource.NewEmptyService` creates a static service operating against an empty list of transient source ids.
- Introduced `prebackup.Handler`. Each `jobsdb` (reader or readWriter) instance, after creating a pre-drop table and before backing it up to the object storage, will call its list of pre-backup handlers, if any. Currently only a single type of pre-backup handler exists and is configured, which is responsible for deleting events whose `source_id` is transient.

## Notion Ticket

https://www.notion.so/rudderstacks/Disable-gateway-dumps-for-certain-sources-e32acf723d3e45b48af7bd707afb5262

## Security

### DIFF
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -13,6 +13,7 @@ import (
 	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
 	"github.com/rudderlabs/rudder-server/gateway"
 	"github.com/rudderlabs/rudder-server/jobsdb"
+	"github.com/rudderlabs/rudder-server/jobsdb/prebackup"
 	"github.com/rudderlabs/rudder-server/processor"
 	ratelimiter "github.com/rudderlabs/rudder-server/rate-limiter"
 	"github.com/rudderlabs/rudder-server/router"
@@ -23,6 +24,7 @@ import (
 	sourcedebugger "github.com/rudderlabs/rudder-server/services/debugger/source"
 	transformationdebugger "github.com/rudderlabs/rudder-server/services/debugger/transformation"
 	"github.com/rudderlabs/rudder-server/services/multitenant"
+	"github.com/rudderlabs/rudder-server/services/transientsource"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 	"github.com/rudderlabs/rudder-server/utils/types"
 	"github.com/rudderlabs/rudder-server/utils/types/servermode"
@@ -69,6 +71,10 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 
 	migrationMode := embedded.App.Options().MigrationMode
 	reportingI := embedded.App.Features().Reporting.GetReportingInstance()
+	transientSources := transientsource.NewService(ctx, backendconfig.DefaultBackendConfig)
+	prebackupHandlers := []prebackup.Handler{
+		prebackup.DropSourceIds(transientSources.SourceIdsSupplier()),
+	}
 
 	//IMP NOTE: All the jobsdb setups must happen before migrator setup.
 	// This gwDBForProcessor should only be used by processor as this is supposed to be stopped and started with the
@@ -80,6 +86,7 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 		jobsdb.WithMigrationMode(migrationMode),
 		jobsdb.WithStatusHandler(),
 		jobsdb.WithQueryFilterKeys(jobsdb.QueryFiltersT{}),
+		jobsdb.WithPreBackupHandlers(prebackupHandlers),
 	)
 	defer gwDBForProcessor.Close()
 	routerDB := jobsdb.NewForReadWrite(
@@ -89,6 +96,7 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 		jobsdb.WithMigrationMode(migrationMode),
 		jobsdb.WithStatusHandler(),
 		jobsdb.WithQueryFilterKeys(router.QueryFilters),
+		jobsdb.WithPreBackupHandlers(prebackupHandlers),
 	)
 	defer routerDB.Close()
 	batchRouterDB := jobsdb.NewForReadWrite(
@@ -98,6 +106,7 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 		jobsdb.WithMigrationMode(migrationMode),
 		jobsdb.WithStatusHandler(),
 		jobsdb.WithQueryFilterKeys(batchrouter.QueryFilters),
+		jobsdb.WithPreBackupHandlers(prebackupHandlers),
 	)
 	defer batchRouterDB.Close()
 	errDB := jobsdb.NewForReadWrite(
@@ -107,6 +116,7 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 		jobsdb.WithMigrationMode(migrationMode),
 		jobsdb.WithStatusHandler(),
 		jobsdb.WithQueryFilterKeys(jobsdb.QueryFiltersT{}),
+		jobsdb.WithPreBackupHandlers(prebackupHandlers),
 	)
 
 	var tenantRouterDB jobsdb.MultiTenantJobsDB
@@ -134,7 +144,7 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 					g.Go(misc.WithBugsnag(func() error {
 						StartProcessor(
 							ctx, &clearDB, gwDBForProcessor, routerDB, batchRouterDB, errDB,
-							reportingI, multitenant.NOOP,
+							reportingI, multitenant.NOOP, transientSources,
 						)
 						return nil
 					}))
@@ -143,7 +153,7 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 			startRouterFunc := func() {
 				if enableRouter {
 					g.Go(misc.WithBugsnag(func() error {
-						StartRouter(ctx, tenantRouterDB, batchRouterDB, errDB, reportingI, multitenant.NOOP)
+						StartRouter(ctx, tenantRouterDB, batchRouterDB, errDB, reportingI, multitenant.NOOP, transientSources)
 						return nil
 					}))
 				}
@@ -170,20 +180,22 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 		modeProvider = state.NewStaticProvider(servermode.DegradedMode)
 	}
 
-	proc := processor.New(ctx, &options.ClearDB, gwDBForProcessor, routerDB, batchRouterDB, errDB, multitenantStats, reportingI)
+	proc := processor.New(ctx, &options.ClearDB, gwDBForProcessor, routerDB, batchRouterDB, errDB, multitenantStats, reportingI, transientSources)
 	rtFactory := &router.Factory{
-		Reporting:     reportingI,
-		Multitenant:   multitenantStats,
-		BackendConfig: backendconfig.DefaultBackendConfig,
-		RouterDB:      tenantRouterDB,
-		ProcErrorDB:   errDB,
+		Reporting:        reportingI,
+		Multitenant:      multitenantStats,
+		BackendConfig:    backendconfig.DefaultBackendConfig,
+		RouterDB:         tenantRouterDB,
+		ProcErrorDB:      errDB,
+		TransientSources: transientSources,
 	}
 	brtFactory := &batchrouter.Factory{
-		Reporting:     reportingI,
-		Multitenant:   multitenantStats,
-		BackendConfig: backendconfig.DefaultBackendConfig,
-		RouterDB:      batchRouterDB,
-		ProcErrorDB:   errDB,
+		Reporting:        reportingI,
+		Multitenant:      multitenantStats,
+		BackendConfig:    backendconfig.DefaultBackendConfig,
+		RouterDB:         batchRouterDB,
+		ProcErrorDB:      errDB,
+		TransientSources: transientSources,
 	}
 	rt := routerManager.New(rtFactory, brtFactory, backendconfig.DefaultBackendConfig)
 
@@ -200,7 +212,7 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 
 	if enableReplay && embedded.App.Features().Replay != nil {
 		var replayDB jobsdb.HandleT
-		replayDB.Setup(jobsdb.ReadWrite, options.ClearDB, "replay", routerDBRetention, migrationMode, true, jobsdb.QueryFiltersT{})
+		replayDB.Setup(jobsdb.ReadWrite, options.ClearDB, "replay", routerDBRetention, migrationMode, true, jobsdb.QueryFiltersT{}, prebackupHandlers)
 		defer replayDB.TearDown()
 		embedded.App.Features().Replay.Setup(&replayDB, gwDBForProcessor, routerDB, batchRouterDB)
 	}
--- a/app/apphandlers/processorAppHandler.go
+++ b/app/apphandlers/processorAppHandler.go
@@ -20,6 +20,7 @@ import (
 	"github.com/rudderlabs/rudder-server/config"
 	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
 	"github.com/rudderlabs/rudder-server/jobsdb"
+	"github.com/rudderlabs/rudder-server/jobsdb/prebackup"
 	proc "github.com/rudderlabs/rudder-server/processor"
 	"github.com/rudderlabs/rudder-server/router"
 	"github.com/rudderlabs/rudder-server/router/batchrouter"
@@ -28,6 +29,7 @@ import (
 	destinationdebugger "github.com/rudderlabs/rudder-server/services/debugger/destination"
 	transformationdebugger "github.com/rudderlabs/rudder-server/services/debugger/transformation"
 	"github.com/rudderlabs/rudder-server/services/multitenant"
+	"github.com/rudderlabs/rudder-server/services/transientsource"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 	"github.com/rudderlabs/rudder-server/utils/types"
 
@@ -94,6 +96,10 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 
 	migrationMode := processor.App.Options().MigrationMode
 	reportingI := processor.App.Features().Reporting.GetReportingInstance()
+	transientSources := transientsource.NewService(ctx, backendconfig.DefaultBackendConfig)
+	prebackupHandlers := []prebackup.Handler{
+		prebackup.DropSourceIds(transientSources.SourceIdsSupplier()),
+	}
 
 	//IMP NOTE: All the jobsdb setups must happen before migrator setup.
 	gwDBForProcessor := jobsdb.NewForRead(
@@ -103,6 +109,7 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 		jobsdb.WithMigrationMode(migrationMode),
 		jobsdb.WithStatusHandler(),
 		jobsdb.WithQueryFilterKeys(jobsdb.QueryFiltersT{}),
+		jobsdb.WithPreBackupHandlers(prebackupHandlers),
 	)
 	defer gwDBForProcessor.Close()
 	gatewayDB = *gwDBForProcessor
@@ -113,6 +120,7 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 		jobsdb.WithMigrationMode(migrationMode),
 		jobsdb.WithStatusHandler(),
 		jobsdb.WithQueryFilterKeys(router.QueryFilters),
+		jobsdb.WithPreBackupHandlers(prebackupHandlers),
 	)
 	defer routerDB.Close()
 	batchRouterDB := jobsdb.NewForReadWrite(
@@ -122,6 +130,7 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 		jobsdb.WithMigrationMode(migrationMode),
 		jobsdb.WithStatusHandler(),
 		jobsdb.WithQueryFilterKeys(batchrouter.QueryFilters),
+		jobsdb.WithPreBackupHandlers(prebackupHandlers),
 	)
 	defer batchRouterDB.Close()
 	errDB := jobsdb.NewForReadWrite(
@@ -131,6 +140,7 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 		jobsdb.WithMigrationMode(migrationMode),
 		jobsdb.WithStatusHandler(),
 		jobsdb.WithQueryFilterKeys(jobsdb.QueryFiltersT{}),
+		jobsdb.WithPreBackupHandlers(prebackupHandlers),
 	)
 	var tenantRouterDB jobsdb.MultiTenantJobsDB
 	var multitenantStats multitenant.MultiTenantI
@@ -156,7 +166,7 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 					if enableProcessor {
 						StartProcessor(
 							ctx, &clearDB, gwDBForProcessor, routerDB, batchRouterDB, errDB,
-							reportingI, multitenant.NOOP,
+							reportingI, multitenant.NOOP, transientSources,
 						)
 					}
 					return nil
@@ -165,7 +175,7 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 			startRouterFunc := func() {
 				if enableRouter {
 					g.Go(func() error {
-						StartRouter(ctx, tenantRouterDB, batchRouterDB, errDB, reportingI, multitenant.NOOP)
+						StartRouter(ctx, tenantRouterDB, batchRouterDB, errDB, reportingI, multitenant.NOOP, transientSources)
 						return nil
 					})
 				}
@@ -204,21 +214,23 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 		return fmt.Errorf("unsupported deployment type: %q", deploymentType)
 	}
 
-	p := proc.New(ctx, &options.ClearDB, gwDBForProcessor, routerDB, batchRouterDB, errDB, multitenantStats, reportingI)
+	p := proc.New(ctx, &options.ClearDB, gwDBForProcessor, routerDB, batchRouterDB, errDB, multitenantStats, reportingI, transientSources)
 
 	rtFactory := &router.Factory{
-		Reporting:     reportingI,
-		Multitenant:   multitenantStats,
-		BackendConfig: backendconfig.DefaultBackendConfig,
-		RouterDB:      tenantRouterDB,
-		ProcErrorDB:   errDB,
+		Reporting:        reportingI,
+		Multitenant:      multitenantStats,
+		BackendConfig:    backendconfig.DefaultBackendConfig,
+		RouterDB:         tenantRouterDB,
+		ProcErrorDB:      errDB,
+		TransientSources: transientSources,
 	}
 	brtFactory := &batchrouter.Factory{
-		Reporting:     reportingI,
-		Multitenant:   multitenantStats,
-		BackendConfig: backendconfig.DefaultBackendConfig,
-		RouterDB:      batchRouterDB,
-		ProcErrorDB:   errDB,
+		Reporting:        reportingI,
+		Multitenant:      multitenantStats,
+		BackendConfig:    backendconfig.DefaultBackendConfig,
+		RouterDB:         batchRouterDB,
+		ProcErrorDB:      errDB,
+		TransientSources: transientSources,
 	}
 	rt := routerManager.New(rtFactory, brtFactory, backendconfig.DefaultBackendConfig)
 
@@ -235,7 +247,7 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 
 	if enableReplay && processor.App.Features().Replay != nil {
 		var replayDB jobsdb.HandleT
-		replayDB.Setup(jobsdb.ReadWrite, options.ClearDB, "replay", routerDBRetention, migrationMode, true, jobsdb.QueryFiltersT{})
+		replayDB.Setup(jobsdb.ReadWrite, options.ClearDB, "replay", routerDBRetention, migrationMode, true, jobsdb.QueryFiltersT{}, prebackupHandlers)
 		defer replayDB.TearDown()
 		processor.App.Features().Replay.Setup(&replayDB, gwDBForProcessor, routerDB, batchRouterDB)
 	}
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -143,27 +143,30 @@ type userWebRequestWorkerT struct {
 
 //HandleT is the struct returned by the Setup call
 type HandleT struct {
-	application                                                app.Interface
-	userWorkerBatchRequestQ                                    chan *userWorkerBatchRequestT
-	batchUserWorkerBatchRequestQ                               chan *batchUserWorkerBatchRequestT
-	jobsDB                                                     jobsdb.JobsDB
-	ackCount                                                   uint64
-	recvCount                                                  uint64
-	backendConfig                                              backendconfig.BackendConfig
-	rateLimiter                                                ratelimiter.RateLimiter
-	stats                                                      stats.Stats
-	batchSizeStat                                              stats.RudderStats
-	requestSizeStat                                            stats.RudderStats
-	dbWritesStat                                               stats.RudderStats
-	dbWorkersBufferFullStat, dbWorkersTimeOutStat              stats.RudderStats
-	bodyReadTimeStat                                           stats.RudderStats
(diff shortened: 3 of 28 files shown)


gateway/gateway.go
```diff
@@ -1253,7 +1256,7 @@ func (gateway *HandleT) collectMetrics(ctx context.Context) {
 			case <-ctx.Done():
 				return
 			case <-gateway.diagnosisTicker.C:
-				gateway.requestMetricLock.RLock()
+				gateway.requestMetricLock.Lock()
```
atzoum, line 1259: we are updating the metric so a lock is required instead of a read lock

app/apphandlers/embeddedAppHandler.go
```diff
@@ -172,7 +182,7 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 		modeProvider = state.NewStaticProvider(servermode.DegradedMode)
 	}
 
-	proc := processor.New(ctx, &options.ClearDB, gwDBForProcessor, routerDB, batchRouterDB, errDB, multitenantStats, reportingI)
+	proc := processor.New(ctx, &options.ClearDB, gwDBForProcessor, routerDB, batchRouterDB, errDB, multitenantStats, reportingI, transientSources)
```
atzoum, line 185: `processor` is the one that initialises `stash`, therefore we need to pass `transientSources` to it.

jobsdb/jobsdb.go
```diff
@@ -1620,21 +1630,36 @@ func (jd *HandleT) renameDS(ds dataSetT, allowMissing bool) {
 	var sqlStatement string
 	var renamedJobStatusTable = fmt.Sprintf(`pre_drop_%s`, ds.JobStatusTable)
 	var renamedJobTable = fmt.Sprintf(`pre_drop_%s`, ds.JobTable)
+	err := jd.doInTransaction(func(tx *sql.Tx) error {
```
atzoum, line 1633: renaming the table and executing the jobsdb's pre-backup handlers is performed in a single transaction.

atzoum:
Do we really need `build/wait-for-go/wait-for-go` version controlled?

app/apphandlers/embeddedAppHandler.go
```diff
@@ -172,20 +182,22 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
...
```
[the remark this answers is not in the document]
fracasula: I used DI for many years and I'm not a big fan to be honest, I'll try to explain why since now it's been a few years without and I'm not looking back.

I think it's yet another thing that we put on top and that one has to learn before being able to contribute. It bloates the code, adds lines to debug and lines to the stack and it makes the binary heavier. Also, and of course it is subjective, I personally prefer to initialize things manually and they are usually not that complex or boring to setup if you follow the SOLID principles (emphasis on the `S` of SOLID, if you're unit really has a single responsibility I'm pretty sure it'll have only a small set of things that you need to inject). By doing that we'd only have very few structs like the processor that would need lots of deps and DI wouldn't bring enough to the table to justify a new framework in the project.

app/apphandlers/processorAppHandler.go
```diff
@@ -344,15 +361,15 @@ func (processor *ProcessorApp) LegacyStart(ctx context.Context, options *app.Opt
 				clearDB := false
 				if enableProcessor {
 					g.Go(func() error {
-						StartProcessor(ctx, &clearDB, &gatewayDB, &routerDB, &batchRouterDB, &procErrorDB, reportingI, multitenantStats)
+						StartProcessor(ctx, &clearDB, &gatewayDB, &routerDB, &batchRouterDB, &procErrorDB, reportingI, multitenantStats, transientSources)
```
fracasula, line 364: Not really related to this PR, just thinking out loud here so feel free to ignore me. I wonder why we even use an `errgroup` with these routines given that they don't return any error at all.

gateway/gateway.go
fracasula: might as well convert it to a `Mutex` then, `RLock()` doesn't seem to be used anywhere else.

jobsdb/integration_test.go
```diff
@@ -148,7 +149,7 @@ func TestJobsDB(t *testing.T) {
 		CustomVal: true,
 	}
 
-	jobDB.Setup(jobsdb.ReadWrite, false, "batch_rt", dbRetention, migrationMode, true, queryFilters)
+	jobDB.Setup(jobsdb.ReadWrite, false, "batch_rt", dbRetention, migrationMode, true, queryFilters, []prebackup.Handler{})
```
fracasula, line 152: I wonder if we could just pass `nil` instead of an empty slice here. It shouldn't break your code.

jobsdb/jobsdb.go
```diff
@@ -1615,26 +1625,55 @@ func (jd *HandleT) invalidateCache(ds dataSetT) {
...
-	if allowMissing {
-		sqlStatement = fmt.Sprintf(`ALTER TABLE IF EXISTS "%s" RENAME TO "%s"`, ds.JobStatusTable, renamedJobStatusTable)
-	} else {
+	err := jd.doInTransaction(func(tx *sql.Tx) error {
 		sqlStatement = fmt.Sprintf(`ALTER TABLE "%s" RENAME TO "%s"`, ds.JobStatusTable, renamedJobStatusTable)
-	}
-	_, err := jd.dbHandle.Exec(sqlStatement)
+		_, err := jd.dbHandle.Exec(sqlStatement)
+		if err != nil {
+			return err
```
fracasula, line 1637: [minor] might be useful to return an error with a bit more context here like `return fmt.Errorf("could not rename status table: %w", err)`.

jobsdb/jobsdb.go
```diff
@@ -1615,26 +1625,55 @@ func (jd *HandleT) invalidateCache(ds dataSetT) {
...
+		}
+
+		sqlStatement = fmt.Sprintf(`ALTER TABLE "%s" RENAME TO "%s"`, ds.JobTable, renamedJobTable)
+		_, err = jd.dbHandle.Exec(sqlStatement)
+		if err != nil {
+			return err
+		}
+		if len(jd.preBackupHandlers) > 0 {
```
fracasula, line 1645: [minor] Do we really need to check for the length of this slice? :thinking: If empty the `for` won't do a thing anyway and you'll return `nil` afterwards, right?

jobsdb/jobsdb.go
```diff
@@ -1615,26 +1625,55 @@ func (jd *HandleT) invalidateCache(ds dataSetT) {
...
+				}
+			}
+		}
+		return nil
+	})
 	jd.assertError(err)
+}
+
+// renameDS renames a dataset if it exists
+func (jd *HandleT) renameDS(ds dataSetT) {
```
fracasula, line 1659: [question] what does `ds` mean here?

jobsdb/jobsdb.go
```diff
@@ -1615,26 +1625,55 @@ func (jd *HandleT) invalidateCache(ds dataSetT) {
...
+	var sqlStatement string
+	var renamedJobStatusTable = fmt.Sprintf(`pre_drop_%s`, ds.JobStatusTable)
+	var renamedJobTable = fmt.Sprintf(`pre_drop_%s`, ds.JobTable)
+	err := jd.doInTransaction(func(tx *sql.Tx) error {
+		sqlStatement = fmt.Sprintf(`ALTER TABLE IF EXISTS "%s" RENAME TO "%s"`, ds.JobStatusTable, renamedJobStatusTable)
```
fracasula, line 1664: Idr if it would work because we're talking about a table name and not a regular argument but perhaps you know. Do you need `fmt.Sprintf` here or we can use prepared statements instead? As in doing `ALTER TABLE $1` and then `db.dbHandle.Exec(sqlStatement, "TABLE_NAME")`.

jobsdb/jobsdb.go
```diff
@@ -1615,26 +1625,55 @@ func (jd *HandleT) invalidateCache(ds dataSetT) {
...
 	var renamedJobTable = fmt.Sprintf(`pre_drop_%s`, ds.JobTable)
-
-	if allowMissing {
-		sqlStatement = fmt.Sprintf(`ALTER TABLE IF EXISTS "%s" RENAME TO "%s"`, ds.JobStatusTable, renamedJobStatusTable)
-	} else {
+	err := jd.doInTransaction(func(tx *sql.Tx) error {
 		sqlStatement = fmt.Sprintf(`ALTER TABLE "%s" RENAME TO "%s"`, ds.JobStatusTable, renamedJobStatusTable)
-	}
-	_, err := jd.dbHandle.Exec(sqlStatement)
+		_, err := jd.dbHandle.Exec(sqlStatement)
```
fracasula, line 1635: Shouldn't this be `tx.Exec(sqlStatement)` instead? :thinking: 
Might be useful to also have a test to check that these ops are treated atomically.

jobsdb/jobsdb.go
fracasula: https://go.dev/play/p/q-0toDi7oF1

jobsdb/prebackup/prebackup.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/lib/pq"
+)
+
+// Handler alters a jobsdb table before backup happens
+type Handler interface {
```
fracasula, line 12: [mino] Can't this be defined in `jobsdb.go` instead and remove the coupling between the two packages? `DropSourceIds` can still return an unexported `struct` (i.e. `*dropSourceIds`). Thoughts?

jobsdb/jobsdb.go
fracasula: Yeah it's not possible, it looks like it could come in the future though: https://github.com/golang/go/issues/18478

However, something we could do is use [QuoteIdentifier](https://pkg.go.dev/github.com/lib/pq#QuoteIdentifier) instead. Potentially we could also avoid `fmt.Sprintf` which uses reflection (slower), not really needed though.

jobsdb/prebackup/prebackup.go
```diff
@@ -0,0 +1,59 @@
...
+	_, err = jsStmt.ExecContext(ctx, sourceIdsParam)
+	if err != nil {
+		return err
+	}
+
+	// Last cleanup events from the jobs table
+	jSql := fmt.Sprintf(`DELETE FROM "%[1]s" WHERE parameters->>'source_id' = ANY($1)`, jobsTable)
+	jStmt, err := txn.Prepare(jSql)
+	if err != nil {
+		return err
```
fracasula, line 52: [minor] err context might be nice here as well, up to you.

jobsdb/prebackup/prebackup_test.go
```diff
@@ -0,0 +1,194 @@
...
+	handler := DropSourceIds(func() []string { return []string{} })
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	var txn *sql.Tx
+
+	// when I execute the dropSourceIds handler for an empty source ids list, a canceled context and nil transaction
+	err := handler.Handle(ctx, txn, "jobs", "job_status")
+
+	// then handler doesn't do anything and no error is returned
+	if err != nil {
```
fracasula, line 71: [nitpick] Could use `require.NoError(t, err)` to make it shorter, ofc totally up to you.

jobsdb/prebackup/prebackup_test.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	fmt.Println("Test on hold, before cleanup")
+	fmt.Println("Press Ctrl+C to exit")
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	<-c
+	return res
+}
+
+func createTables(t *testing.T, txn *sql.Tx, jobsTable, jobStatusTable string) {
```
fracasula, line 120: Do we have migrations that would create these tables in production for us? If yes could we potentially use those in the test instead?

app/apphandlers/embeddedAppHandler.go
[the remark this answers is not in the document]
atzoum: Yes, you always have to give something in order to take something else. I have used DI for many years as well, although on a different language, and it something that I am missing here, where we keep on solving the dependency injection problem in all sorts of other different ways :)

I am not expecting us to use DI, it is hard to retrofit such a design choice anyway. Just thought that this is a good example to argue in favour of DI in general :)

gateway/gateway.go
atzoum: yes indeed

jobsdb/integration_test.go
atzoum: Because it is easier to read, an empty slice of `prebackup.Handler` is more explicit than a nil which can be anything. Performance here is not an issue.

jobsdb/jobsdb.go
atzoum: Yes, this was one incomplete refactoring laying around... hopefully you were there to spot it, thanks!

jobsdb/jobsdb.go
atzoum: yes, will do

jobsdb/jobsdb.go
atzoum: no we don't need to

jobsdb/jobsdb.go
atzoum: Yes & no.  Double-quotes & zero-bytes is not something that we are really at risk of, given the conventions that we are currently using for table names in rudder-server. I don't have any objection to doing what you propose though, as long as we do this on a separate task and for all occurrences in the codebase, otherwise it doesn't offer any advantage. Still I would treat it with low priority.

jobsdb/prebackup/prebackup.go
atzoum: I don't understand the value of an unexported interface, please elaborate a bit on this proposal.

jobsdb/prebackup/prebackup.go
atzoum: yes

jobsdb/prebackup/prebackup_test.go
atzoum: ok

jobsdb/prebackup/prebackup_test.go
atzoum: I cannot tell. Why do we need production (redacted?) data for this though? It is a simple, straightforward, unit test.

fracasula: Hold on, I thought you were bringing up a real DB so I thought it was an integration test? If that's the case I think it makes sense to be as close to production as possible. If you can (and want) it'd be great to have the same schema as it is applied in production here to avoid human mistakes and possible drifts (as in one change applied in production but then we forgot to apply it here in the test schema). Not blocking comment btw :+1:

jobsdb/jobsdb.go
fracasula: Yeah sure, no problem with `fmt.Sprintf()` as well, it's a super minor optimization, I mostly asked because I genuinely didn't remember whether it was possible to "prepare" a table name :sweat_smile:

jobsdb/integration_test.go
fracasula: What do you mean it can be anything? You can't pass anything there, it can either be `nil` or a slice of that type or it won't compile, right? In any case, minor optimization but I don't see the reason to allocate. Plus it is a test so completely up to you man :+1:

jobsdb/prebackup/prebackup.go
fracasula: It's just a minor software design (not blocking) question but I'd like to know your opinion about this.

To me it looks like the `Handler` interface is not needed in the `prebackup` package, you're just defining it upfront (kinda Java style).
The only thing using it there is `func DropSourceIds` but there is no reason to return an interface there, it could just return a struct instead.

It seems more "idiomatic" in Golang to have the package that uses the abstraction to
define it and to me that package is the `jobsdb` (i.e. in `WithPreBackupHandlers`).

Long story short, I'd move the interface in the `jobsdb` package and call it something like `PreBackupper` because that's what it does. `Handler` is a very generic name and it doesn't tell us what that abstraction actually means to the `jobsdb` package.

Once we do that we realize that the `prebackup` package doesn't even have to be called `prebackup`. What it does is that it gives you an utility function to drop jobs by `source_id`, we don't even care if those are transient either (not at that stage at least I think?).

So the `prebackup` could be `sourcedropper` and we could remove all references to
handlers and transient things from there and have a properly named interface in `jobsdb` instead.

These are my two cents, your code does what is supposed to do and there is nothing wrong with it, but as a reader of your code this is the impression I got :slightly_smiling_face:

jobsdb/prebackup/prebackup_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+func checkCount(t *testing.T, rows *sql.Rows) (count int) {
+	for rows.Next() {
+		err := rows.Scan(&count)
+		if err != nil {
+			t.Fatalf("failed to count rows: %s", err)
+		}
+	}
+	return count
+}
```
fracasula, line 193: [optional]

```suggestion
func requireRowsCount(t *testing.T, txn *sql.Tx, tableName string, expectedCount int) {
	t.Helper()
	rows := txn.QueryRow(fmt.Sprintf("SELECT COUNT(*) as count FROM  %s", tableName))

	var count int
	require.NoError(t, rows.Scan(&count))
	require.EqualValues(t, expectedCount, count)
}
```

Given that you're passing `t` already we could have it as a `t.Helper()` and get rid of `checkCount` thanks to `txn.QueryRow`.

processor/processor.go
```diff
@@ -969,7 +973,7 @@ func (proc *HandleT) getFailedEventJobs(response transformer.ResponseT, commonMe
 
 		for _, message := range messages {
 			sampleEvent, err := jsonfast.Marshal(message)
-			if err != nil {
+			if err != nil || proc.transientSources.Apply(commonMetaData.SourceID) {
 				proc.logger.Errorf(`[Processor: getFailedEventJobs] Failed to unmarshal first element in failed events: %v`, err)
```
fracasula, line 977: Is the log message still relevant is we get in here due to `proc.transientSources.Apply(commonMetaData.SourceID)` instead of because of `err != nil`?

services/transientsource/transientsource.go
```diff
@@ -0,0 +1,140 @@
...
+	"sync"
+
+	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
+	"github.com/rudderlabs/rudder-server/jobsdb"
+	"github.com/rudderlabs/rudder-server/utils/pubsub"
+	"github.com/tidwall/gjson"
+)
+
+// Service provides services related to transient source ids
+type Service interface {
```
fracasula, line 15: Another software design dilemma for you :sweat_smile: 

Again, your code does what is supposed to do but I'd like to point out that declaring interfaces named after "things" instead of "actions" is kind of a bad practice.

That's why we usually see interfaces in the Go standard library end with "er" (e.g. `Server`, `Reader`, `Writer`, `Closer`, etcetera) and not `Service`.

The main reason is that if you name interfaces after things they'll notoriously get big (or at least bigger than their "action" couterpart and less understable) and give birth to fat objects with multiple responsibilities (see SOLID principles).

In fact we see packages using this `Service` that don't even need the whole set of methods (i.e. the `batchrouter` which is just using `Apply(sourceID)`) so now we're under the impression that the `batchrouter` needs the whole service and we don't even know what a service is/does unless we look at the whole thing.

FYI, interfaces can also be embedded like `Reader` + `Closer` = [ReadCloser](https://pkg.go.dev/io#ReadCloser), so if some object is in need of more "doers" in a single abstraction it can still be done like in the standard library.

Again, not a blocking issue but please do keep in mind that these kind of interfaces are usually a code smell.

services/transientsource/transientsource.go
```diff
@@ -0,0 +1,140 @@
...
+
+// NewStaticService creates a new service that operates against a predefined list of transient source ids.
+// Useful for tests.
+func NewStaticService(sourceIds []string) Service {
+	s := &service{
+		init:         make(chan struct{}),
+		sourceIds:    sourceIds,
+		sourceIdsMap: asMap(sourceIds),
+	}
+	close(s.init)
```
fracasula, line 57: [minor] Do we really need to make the channel and then close it? Can't we just not make it?

services/transientsource/transientsource.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (r *service) ApplyJob(job *jobsdb.JobT) bool {
+	return r.ApplyParams(job.Parameters)
+}
+
+// updateLoop uses backend config to retrieve & keep up-to-date the list of transient source ids
+func (r *service) updateLoop(ctx context.Context, config backendconfig.BackendConfig) {
+
+	var onceInit sync.Once
+	go func() {
```
fracasula, line 93: Minor but still worth mentioning. This kind of APIs where a function seems synchronous but is not should be avoided.

So instead of:

```go
func someFunctionThatSeemsSynchronous() {
    go func() {
        // let's do stuff asynchronously instead
    }()
}
```

We should consider doing:

```go
func someFunctionThatIsActuallySynchronous() {
    // let's do stuff synchronouly!!!
}

func someCaller() {
    // if I need to call it synchronously then I do
    someFunctionThatIsActuallySynchronous()

    // but if I need to go asynchronous then I do
    go someFunctionThatIsActuallySynchronous()
}
```

The advantages are:
1. the name of the function is not misleading anymore
2. it's up to the caller whether it wants to go sync or async
3. the caller might want to call multiple functions asynchronouly so having the way I'm proposing would simplify the go routines coordination, which should be on the caller side

So I suggest you do ` go updateLoop()` and remove the go routine from inside it.

As a sidenote `onceInit` could potentially be part of the `struct` and do `r.onceInit.Do(...)`.

fracasula (approved):
Logic wise looks good to me but I've added a few comments because I don't think your code is always idiomatic and sometimes I find it hard to follow.

The only blocking comment would be this one about transactions: https://github.com/rudderlabs/rudder-server/pull/1869#discussion_r862680137

jobsdb/integration_test.go
atzoum: if you were to only read this line, ignoring the actual method signature, you can now understand that the caller passes an empty slice of `prebackup.Handler`. If this changes to `nil`, you won't understand much until you go and look at the method's actual signature.

jobsdb/prebackup/prebackup.go
atzoum: Yes, what you propose is valid, still the interface needs to be exported :)

fracasula: > Yes, what you propose is valid, still the interface needs to be exported :)

Not here where I added the comment, it would be in a different package where it would be more clear, both the name and its purpose.

> I don't consider it wrong to explicitly set the return type to the interface in question, instead of using an unexported struct type

That depends, we should do things when it makes sense, I don't think it does in this case.

You're returning an interface that is simply not needed here and IMO that led to:
* interface name that could be more specific like I'm suggesting instead of `Handler`
* interface in a package where it is not needed, leads to confusion
* more coupling between packages
* misleading name for the `prebackup` package

I haven't heard an argument that explains why it makes sense to return an interface here so my suggestion so far, along with the above, is: let's not return interfaces just because we are used to do so. I know that is very common the first years of Go, especially when coming from a different background so it's cool. That said, if you believe your approach is better then go ahead :+1: 

FYI: Here's some stuff worth [reading](https://blog.chewxy.com/2018/03/18/golang-interfaces/).

services/transientsource/transientsource.go
atzoum: Yes, I have read [the bible](https://go.dev/doc/effective_go) and I know how I should name single-method interfaces, I also know how we should combine and name 2 single method interfaces in one. But should 2 be the limit for the number of an interface's methods? And should everything be named using -er? No, I don't believe that naming an interface after a [thing](https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L46) is a bad practice.

Why is it better to have a single struct implementing 5 single-method interfaces, instead of a struct implementing a single 5-method interface (can use embedding there as well) when these methods are related to each other?

Is [this design](https://github.com/rudderlabs/rudder-server/blob/69857216e8b736d2924a761a8eef578edce8d40e/app/cluster/dynamic.go#L42) better? In this scenario, we are declaring the `lifecycle` interface (a.k.a. `StartStopper`) as we only need to use 2 methods inside `Dynamic`. But in the `Dynamic` struct, not all `lifecycle` fields are made equal. Therefore, we are forced to use the field's name as a means to indicate the actual type of the field that we expect, e.g. in `GatewayDB`  we need a `jobsdb.HandleT`, in `Processor` we expect  `processor.HandleT` etc.
With the above design I would argue that by using the small `lifecycle` interface for everything we essentially give away many of the benefits provided by strong typing.

jobsdb/prebackup/prebackup.go
atzoum: Unfortunately, I cannot relate (yet) to [this](https://blog.chewxy.com/2018/03/18/golang-interfaces/) since it doesn't explain, or otherwise I fail to understand, why we shouldn't write types to fulfil interfaces. Until I do, my mind will continue working like [this](https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L193) 🤣

services/transientsource/transientsource.go
atzoum: Thanks, you are correct!

processor/processor.go
atzoum: no it is not relevant, thanks for spotting once again!

jobsdb/prebackup/prebackup_test.go
atzoum: thanks!

app/apphandlers/processorAppHandler.go
atzoum: it waits for all go routines to finish in [#417](https://github.com/rudderlabs/rudder-server/pull/1869/files#diff-a56cac86f1d08aa5ed9a9f58c33b5f65aa8b34574a018ed4b12d14cc4b190179R417) so even if this specific task doesn't appear to through an error, we still need to wait for it to complete, upon context cancellation.

fracasula: Yeah, looks like a wait group could do the trick. Might be clearer since `errgroup` clearly has `err` in its name it implies that the error is important.

router/utils/utils.go
```diff
@@ -45,7 +45,7 @@ func loadConfig() {
 func getRetentionTimeForDestination(destID string) time.Duration {
 	destJobRetentionFound := config.IsSet("Router." + destID + ".jobRetention")
 	if destJobRetentionFound {
-		return config.GetDuration("Router"+"."+destID+"jobRetention", 720, time.Hour)
+		return config.GetDuration("Router."+destID+".jobRetention", 720, time.Hour)
```
atzoum, line 48: @chandumlg are we really using this configuration option in production? How can this be?

cisse21: We use this for open source people so that if for some reason their server goes into degraded mode then upon resumption of normal mode we don’t loose their jobs. We set it to 24hr in our production setup through operator

atzoum: Are we setting it properly (with the dot), improperly (without the dot) or both ways?

cisse21: I’m sorry I didn’t understand what you are trying to say. Can you elaborate a bit?

cisse21: https://github.com/rudderlabs/rudderstack-operator/blob/master/operator-helm/valuefiles/values.yaml#L121

This is what we set in production environment

atzoum: @cisse21  I am referring to the per-destinationId configuration override.
The current code in `master` as you can see here is testing the existence of `"Router." + destID + ".jobRetention"` and then reads the value from `"Router"+"."+destID+"jobRetention"`!

cisse21: Got It…. We set it at global level which is not without any destination Id flag. We can override the default behaviour for destination Id by setting appropriate flag per customer

chandumlg: @atzoum This is a bug. Thanks for fixing this. 👍

jobsdb/jobsdb.go
```diff
@@ -699,14 +710,15 @@ multiple users of JobsDB
 dsRetentionPeriod = A DS is not deleted if it has some activity
 in the retention time
 */
-func (jd *HandleT) Setup(ownerType OwnerType, clearAll bool, tablePrefix string, retentionPeriod time.Duration, migrationMode string, registerStatusHandler bool, queryFilterKeys QueryFiltersT) {
+func (jd *HandleT) Setup(ownerType OwnerType, clearAll bool, tablePrefix string, retentionPeriod time.Duration, migrationMode string, registerStatusHandler bool, queryFilterKeys QueryFiltersT, preBackupHandlers []prebackup.Handler) {
```
lvrach, line 713:
```suggestion
func (jd *HandleT) Setup(ownerType OwnerType, clearAll bool, tablePrefix string, retentionPeriod time.Duration, migrationMode string, registerStatusHandler bool, queryFilterKeys QueryFiltersT) {
```

We don't need to pass `preBackupHandlers` here. `.Setup` is already having too many parameters. 

1. Use New...(), WithPreBackupHandle pattern
2. expose `preBackupHandlers` and inject the handler directly.

app/apphandlers/embeddedAppHandler.go
```diff
@@ -82,6 +88,7 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 		jobsdb.WithMigrationMode(migrationMode),
 		jobsdb.WithStatusHandler(),
 		jobsdb.WithQueryFilterKeys(jobsdb.QueryFiltersT{}),
+		jobsdb.WithPreBackupHandlers(prebackupHandlers),
```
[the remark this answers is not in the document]
lvrach: I believe it is a step into the right direction, I would go with it a step further.

Instead of a prebackup handler, I would go with a predelete handler. There you can have multiple implementations:
* Do nothing (no backup at all)
* Backup everything
* Filter backup

It will allow us to test and extend the full backup functionality. 

I understand that this might exceed to scope of this PR, consider the comment non-blocking.

jobsdb/jobsdb.go
```diff
@@ -1615,26 +1627,67 @@ func (jd *HandleT) invalidateCache(ds dataSetT) {
 	}
 }
 
-//Rename a dataset
-func (jd *HandleT) renameDS(ds dataSetT, allowMissing bool) {
+//mustRenameDS renames a dataset
+func (jd *HandleT) mustRenameDS(ds dataSetT) {
```
lvrach, line 1631:
```suggestion
func (jd *HandleT) mustRenameDS(ds dataSetT) error {
```

I would suggest starting returning errors in newly introduce methods

processor/processor.go
```diff
@@ -284,7 +286,7 @@ func (proc *HandleT) Status() interface{} {
 func (proc *HandleT) Setup(
 	backendConfig backendconfig.BackendConfig, gatewayDB jobsdb.JobsDB, routerDB jobsdb.JobsDB,
 	batchRouterDB jobsdb.JobsDB, errorDB jobsdb.JobsDB, clearDB *bool, reporting types.ReportingI,
-	multiTenantStat multitenant.MultiTenantI,
+	multiTenantStat multitenant.MultiTenantI, transientSources transientsource.Service,
```
lvrach, line 289: Having to propagate transientSources inside proc, just to pass it to stash does not look very good. We can just inject it to errorDB, before we pass it here right?

If we do that, all the modifications in stash and proc won't be necessery.

router/router.go
```diff
@@ -1540,7 +1543,11 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
 			if err != nil {
 				errorCode = 200 //TODO handle properly
 			}
-			sd = utilTypes.CreateStatusDetail(resp.status.JobState, 0, errorCode, string(resp.status.ErrorResponse), resp.JobT.EventPayload, eventName, eventType)
+			sampleEvent := resp.JobT.EventPayload
+			if rt.transientSources.Apply(parameters.SourceID) {
```
lvrach, line 1547: [question] Not sure I understand the motivation behind this change, can you provide more context?

atzoum: yes it is the side-effect #3 of transient sources
> 3. Sample events will not be sent to the reporting service.

processor/processor.go
atzoum: `proccessor` needs `transientSources` in order to perform sampleEvent filtering.
`stash` needs `transientSources` in order to perform event filtering as well.
`jobsdb` currently doesn't need  `transientSources`, it depends on `prebackup.Handlers`

jobsdb/jobsdb.go
atzoum: ok, but still the caller will panic on error, at least for the time being

jobsdb/jobsdb.go
atzoum: Why exposing it for initialisation purposes is better than including it in the existing initialiser function?

Currently `jobsdb.HandleT` has none of its dependency fields exported, which is not bad.
Generally speaking, refactoring jobsdb's initialization is indeed needed, but shall we deal with it now?

app/apphandlers/embeddedAppHandler.go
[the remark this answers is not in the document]
atzoum: I like it, but needs rewriting, thus should treat it as a separate task. I will include one in the backlog

chandumlg (approved):

PR #1874, opened 2022-04-21
fix: disable stash backup if masterBackup is disabled.

cisse21:
# Description

We get various requests from CS team to disable backups for a few customers for compliance reasons. We should be able to disable the backups with a single flag and this change helps us in achieving that

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=9759e895d487439cbe75813164bda3b0)

## Security


chandumlg (approved):

processor/stash/stash.go
```diff
@@ -93,7 +93,7 @@ func (st *HandleT) Start(ctx context.Context) {
 }
 
 func (st *HandleT) setupFileUploader() {
-	if errorStashEnabled {
+	if errorStashEnabled && jobsdb.IsMasterBackupEnabled() {
```
atzoum, line 96: What is the difference between the stash backup and the jobsdb's own backup?
Each jobsdb also has its own `BackupSettings#IsBackupEnabled` method which tests both `masterBackupEnabled` and `instanceBackupEnabled` flags. Is the second flag relevant here?

cisse21: You are right .... each jobsdb has its own backup enabled but in the case of proc_error, since there is no jobsdb reader(the one responsible for backing up) we run backups of the error table in a separate process and enable them with `errorStashEnabled` flag

Sidddddarth (approved):

processor/stash/stash.go
atzoum: @cisse21 thanks, I see that the `proc_error` db is setup using `jobsdb.ReadWrite`, which means that both migration and backup loops are running for it (the archiver as well which I don't really understand what this does exactly but maybe not important for this discussion).

So, we have `stash` moving entries to object storage and updating the job statuses in `proc_error`. The migration loop will take care of old tables and since backup is not enabled for `proc_error` the entries will not end up two times in S3.

What about enabling backup for `proc_error` as well and instead of having `stash` sending entries to the backup table, use it only for marking jobs as `Succeeded`. Then the migration and backup loops will do the rest. Having `stash` performing the job of jobsdb's internal loops might be problematic in the future. In fact, now that I need to introduce a filtering mechanism which will skip backing up specific `source_id`s I need to adapt `stash` as well. It would be better if I only had to introduce the change in a single place instead.

cisse21: @atzoum  Ideally we should be having all of them at the same place and that makes sense too but we would need to provide backwards compatibility for a few months since we have a lot of open source users and on-prem users whom we would not wanna break the existing flow

psrikanth88: @atzoum @cisse21  The main reason `stash` was introduced instead of relying on `jobsdb` backup was to back up the errors to object storage at a deterministic rate. In a lot of customer use cases, they might not even have 100K errors in say a month, so the lag in delivering these backups is very high. This would require setting jobsdb size to a very small one and that also needs tuning and its hard to tune as errors might not be coming at a constant rate. One other reason for faster and deterministic backups via stash is to allow for developing features that directly query object storage to surface errors to user rather than going through a central storage powered by RudderStack

atzoum: Thank you @psrikanth88 for this. So for `proc_error`, `stash` is it's router. I am currently working for [this task](https://www.notion.so/rudderstacks/Disable-gateway-dumps-for-certain-sources-e32acf723d3e45b48af7bd707afb5262) and it is not clear whether filtering based on `source_id` shall be performed by `stash` or not. Now, I tend to believe that it doesn't. @chandumlg?

PR #1875, opened 2022-04-22
feat: Kafka client

fracasula:
# Description

Due to the fact that we keep getting issues while publishing with Sarama, we decided to move away from it.

Here I'm adding a very simple abstraction so that we can use our own library to talk with Kafka. If the need arises we'll be able to switch to another underlying library without having to change the way we configure our client and do things with it.

## Notion Ticket

[< Notion Link >](https://www.notion.so/rudderstacks/Kafka-timeout-not-working-properly-a2ffbc0c620442a3a7afb9f49c8c09fa)

## Security

### DIFF
--- a/utils/kafkaclient/config.go
+++ b/utils/kafkaclient/config.go
@@ -1,0 +1,89 @@
+package kafkaclient
+
+import "time"
+
+type ScramHashGenerator uint8
+
+const (
+	ScramPlainText ScramHashGenerator = iota
+	ScramSHA256
+	ScramSHA512
+)
+
+const (
+	defaultDialTimeout = 10 * time.Second
+	defaultOpTimeout   = 2 * time.Second
+)
+
+// Option is an abstraction used to allow the configuration of a client
+type Option interface {
+	apply(*config)
+}
+
+type withOption struct{ setup func(*config) }
+
+func (w withOption) apply(c *config) { w.setup(c) }
+
+type config struct {
+	dialTimeout time.Duration
+	opTimeout   time.Duration
+	tlsConfig   *tlsConfig
+	saslConfig  *saslConfig
+}
+
+func (c *config) defaults() {
+	if c.dialTimeout < 1 {
+		c.dialTimeout = defaultDialTimeout
+	}
+	if c.opTimeout < 1 {
+		c.opTimeout = defaultOpTimeout
+	}
+}
+
+type tlsConfig struct {
+	cert, key,
+	caCertificate []byte
+	insecureSkipVerify bool
+}
+
+type saslConfig struct {
+	scramHashGen       ScramHashGenerator
+	username, password string
+}
+
+// WithDialTimeout sets the maximum amount of time a dial will wait for a connect to complete
+func WithDialTimeout(t time.Duration) Option {
+	return withOption{setup: func(c *config) {
+		c.dialTimeout = t
+	}}
+}
+
+// WithOpTimeout sets the maximum amount of time for operations to complete (e.g. read, write, wait for batches)
+func WithOpTimeout(t time.Duration) Option {
+	return withOption{setup: func(c *config) {
+		c.opTimeout = t
+	}}
+}
+
+// WithTLS adds TLS support
+func WithTLS(cert, key, caCertificate []byte, insecureSkipVerify bool) Option {
+	return withOption{setup: func(c *config) {
+		c.tlsConfig = &tlsConfig{
+			cert:               cert,
+			key:                key,
+			caCertificate:      caCertificate,
+			insecureSkipVerify: insecureSkipVerify,
+		}
+	}}
+}
+
+// WithSASL adds SASL support
+func WithSASL(hg ScramHashGenerator, username, password string) Option {
+	return withOption{setup: func(c *config) {
+		c.saslConfig = &saslConfig{
+			scramHashGen: hg,
+			username:     username,
+			password:     password,
+		}
+	}}
+}
--- a/utils/kafkaclient/producer.go
+++ b/utils/kafkaclient/producer.go
@@ -1,0 +1,78 @@
+package kafkaclient
+
+import (
+	"context"
+	"net"
+
+	"github.com/segmentio/kafka-go"
+)
+
+// Producer provides a high-level API for producing messages to Kafka
+type Producer struct {
+	writer *kafka.Writer
+}
+
+// NewProducer instantiates a new producer. To use it asynchronously just do "go p.Publish(ctx, msgs)".
+func (c *client) NewProducer(topic string) *Producer {
+	return &Producer{
+		writer: &kafka.Writer{
+			Topic:                  topic,
+			Balancer:               &kafka.Hash{},
+			BatchTimeout:           c.config.opTimeout,
+			ReadTimeout:            c.config.opTimeout,
+			WriteTimeout:           c.config.opTimeout,
+			RequiredAcks:           kafka.RequireAll,
+			AllowAutoTopicCreation: true,
+			Async:                  false,
+			Compression:            0,
+			Transport: &kafka.Transport{
+				DialTimeout: c.config.dialTimeout,
+				Dial: func(ctx context.Context, network string, address string) (net.Conn, error) {
+					return c.dialer.DialContext(ctx, network, address)
+				},
+			},
+		},
+	}
+}
+
+// Close tries to close the producer, but it will return sooner if the context is canceled.
+// A routine in background will still try to close the producer since the underlying library does not support
+// contexts on Close().
+func (p *Producer) Close(ctx context.Context) error {
+	done := make(chan error, 1)
+	go func() {
+		done <- p.writer.Close()
+	}()
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case err := <-done:
+		return err
+	}
+}
+
+// Publish allows the production of one or more message to Kafka.
+// To use it asynchronously just do "go p.Publish(ctx, msgs)".
+func (p *Producer) Publish(ctx context.Context, msgs ...Message) error {
+	messages := make([]kafka.Message, len(msgs))
+	for i := range msgs {
+		var headers []kafka.Header
+		if l := len(msgs[i].Headers); l > 0 {
+			headers = make([]kafka.Header, l)
+			for k := range msgs[i].Headers {
+				headers[k] = kafka.Header{
+					Key:   msgs[i].Headers[k].Key,
+					Value: msgs[i].Headers[k].Value,
+				}
+			}
+		}
+		messages[i] = kafka.Message{
+			Key:     msgs[i].Key,
+			Value:   msgs[i].Value,
+			Time:    msgs[i].Timestamp,
+			Headers: headers,
+		}
+	}
+	return p.writer.WriteMessages(ctx, messages...)
+}
--- a/go.mod
+++ b/go.mod
@@ -45,6 +45,7 @@ require (
 	github.com/phayes/freeport v0.0.0-20180830031419-95f893ade6f2
 	github.com/rs/cors v1.7.0
 	github.com/rudderlabs/analytics-go v3.3.1+incompatible
+	github.com/segmentio/kafka-go v0.4.31
 	github.com/shurcooL/vfsgen v0.0.0-20200824052919-0d455de96546
 	github.com/snowflakedb/gosnowflake v1.6.3
 	github.com/spaolacci/murmur3 v1.1.0
@@ -126,7 +127,7 @@ require (
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
 	github.com/jstemmer/go-junit-report v0.9.1 // indirect
 	github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0 // indirect
-	github.com/klauspost/compress v1.13.6 // indirect
+	github.com/klauspost/compress v1.14.2 // indirect
 	github.com/klauspost/cpuid v1.2.3 // indirect
 	github.com/magiconair/properties v1.8.5 // indirect
 	github.com/mattn/go-ieproxy v0.0.1 // indirect
@@ -143,7 +144,7 @@ require (
 	github.com/patrickmn/go-cache v2.1.0+incompatible // indirect
 	github.com/pelletier/go-toml v1.9.3 // indirect
 	github.com/pierrec/lz4 v2.6.1+incompatible // indirect
-	github.com/pierrec/lz4/v4 v4.1.8 // indirect
+	github.com/pierrec/lz4/v4 v4.1.14 // indirect
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -205,7 +205,6 @@ github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869/go.mod h1:Ekp36dR
 github.com/boombuler/barcode v1.0.0/go.mod h1:paBWMcWSl3LHKBqUq+rly7CNSldXjb2rDl3JlRe0mD8=
 github.com/bshuster-repo/logrus-logstash-hook v0.4.1/go.mod h1:zsTqEiSzDgAa/8GZR7E1qaXrhYNDKBYy5/dWPTIflbk=
 github.com/buger/jsonparser v0.0.0-20180808090653-f4dd9f5a6b44/go.mod h1:bbYlZJ7hK1yFx9hf58LP0zeX7UjIGs20ufpu3evjr+s=
-github.com/bugsnag/bugsnag-go v0.0.0-20141110184014-b1d153021fcd h1:rFt+Y/IK1aEZkEHchZRSq9OQbsSzIT/OrI8YFFmRIng=
 github.com/bugsnag/bugsnag-go v0.0.0-20141110184014-b1d153021fcd/go.mod h1:2oa8nejYd4cQ/b0hMIopN0lCRxU0bueqREvZLWFrtK8=
 github.com/bugsnag/bugsnag-go/v2 v2.1.2 h1:R5rgYn5w5LhVIere+n+Ah49pa4E1b3OP2bSwHtURmv8=
 github.com/bugsnag/bugsnag-go/v2 v2.1.2/go.mod h1:mJCnw33SPVYPFTsAeSR/kpwuyvjTXrPK5w/XZfTUwMU=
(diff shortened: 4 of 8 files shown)


fracasula:
Oh I forgot to add a `TestProducerTimeout` :man_facepalming: 
I tested it before writing the client but then forgot to add a real test in the end. Will do it soon, the PR is still reviewable :+1:

atzoum (commented):
@fracasula you have done some thorough and elaborate work! I don't have any comment on the quality of this work, however I am failing to acknowledge the merits of such a meta-client abstraction, apart from the fact that it is thoroughly tested.

My main arguments:
- We currently don't really use the kafka client in countless places across our codebase, instead it is limited to `kafkamanager.go` alone. Thus we are not threatened with a great refactoring impact if we ever want to switch the client implementation or upgrade the existing one, something that this abstraction could safeguard us from.
- As a matter of fact, `kafkamanager.go` is our current kafka abstraction, although not the most elegant one! Furthermore, we do have some kafka configuration abstraction(s) already in the form of structs which correspond to control plane's configuration options and these are really hard to change at this late stage.
- I cannot see what is preventing us from setting up test scenarios with various configuration combinations in terms of the current `kafkamanager.go` implementation. Isn't this required anyway, even with the meta-client?
- It isn't a goal to write our own kafka client but to change the one that we are using. The tests of this meta-client verify that the abstraction is written well, but we are still missing the actual tests, the ones that will verify the integration, i.e. that we can (1) use segment's kafka client instead of sarama, (2) with or without the use of a meta-client.

A secondary comment regarding the actual design of the meta-client itself is that I would expect to see the use of interfaces for kafka producer and consumer for two reasons:
   1. Facilitating testing and mocking.
   2. Facilitating any future switch to a different client library. As long as you keep the same interface you can provide a new implementation of this meta-client and even use a feature flag for controlling your rollout strategy from the old to the new & easily revert a failed rollout if necessary.

Open to discuss, as always :)

fracasula:
@atzoum thanks for your comment, I'll address it in separate answers below :+1:

> * We currently don't really use the kafka client in countless places across our codebase, instead it is limited to `kafkamanager.go` alone. Thus we are not threatened with a great refactoring impact if we ever want to switch the client implementation or upgrade the existing one, something that this abstraction could safeguard us from.

This doesn't mean we shouldn't wrap it. To me it's a matter of segregation of responsibilities. Here I just want to make sure that whatever client we use (sarama, confluent, segment, etc...) we can do the things we want to do (connect plainly, with auth, with scram, with tls, etcetera).

If you got a regression you'll immediately know whether it is a client issue or a stream manager issue. Plus this abstraction layer is fairly simple. What took me long were the tests.

> * I cannot see what is preventing us from setting up test scenarios with various configuration combinations in terms of the current `kafkamanager.go` implementation. Isn't this required anyway, even with the meta-client?

I think you're mixing things. In the `kafkamanager` we don't care what client we use, we just read from a source and write to a destination. Most tests in the `kafkamanager` can be done with a mocked client. In fact that's what I'm planning to do in the second PR.

With a mocked client we can test not only against the expected outcomes but also against the unexpected ones. What does the `kafkamanager` do when the consumer returns a message with weird data (sorta like chaos testing)?

Then as I see fit of course I might be adding additional integration tests as well according to the stream manager business logic.

Additional concerns: it is not easy to mock a 3rd party client and if we do then we'll have to change both the code and the tests. With my approach you have isolation meaning:
1. in the client package you just test that the client does what is supposed to do and it has no regression after updates or after switching to other clients
2. in the stream manager you can have both unit and integration tests to test both good and unexpected behaviours. do you want to change client? fine, the stream manager will not change, nor will its unit/integration tests because we depend on our own client now. in fact, not even the `client_test.go` will change.

> * It isn't a goal to write our own kafka client but to change the one that we are using. The tests of this meta-client verify that the abstraction is written well, but we are still missing the actual tests, the ones that will verify the integration, i.e. that we can (1) use segment's kafka client instead of sarama, (2) with or without the use of a meta-client.

The goal will be achieved with more than one PR, I thought we said to keep 'em small.

So, I guess you missed the PR description where I explained that the integration will happen in a second PR (I'm working on that already :slightly_smiling_face:).

Also, see [this comment](https://github.com/rudderlabs/rudder-server/pull/1875#issuecomment-1113484135), there will soon be a `TestProducerTimeout` as well.

> A secondary comment regarding the actual design of the meta-client itself is that I would expect to see the use of interfaces for kafka producer and consumer for two reasons:
> 
> 1. Facilitating testing and mocking.
> 2. Facilitating any future switch to a different client library. As long as you keep the same interface you can provide a new implementation of this meta-client and even use a feature flag for controlling your rollout strategy from the old to the new & easily revert a failed rollout if necessary.

It is perfectly fine to depend on your own client and you don't even need interfaces in this stage. Once I'll need to mock the client in the stream manager I'll just declare an interface there. That interface will quite likely look like this:

```go
type producer interface {
	Publish(ctx context.Context, msgs ...kafkaclient.Message) error
}
```

And then I can make it return whatever I need it to.

This is why I believe that with my approach we'll have better tests, better coverage, better isolation, less coupling and clearer components responsibilities.

The same would apply to other parts of the system. For example I don't believe that we should be using the etcd client directly.

utils/kafkaclient/config.go
```diff
@@ -0,0 +1,156 @@
+package kafkaclient
```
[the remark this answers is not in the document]
fracasula: There is a greater plan, please have a look at my big fat comment below :laughing: I'm going to resolve this one since it is a duplicate.

atzoum:
> I think you're mixing things. In the `kafkamanager` we don't care what client we use, we just read from a source and write to a destination. Most tests in the `kafkamanager` can be done with a mocked client. In fact that's what I'm planning to do in the second PR.

I mentioned `kafkamanager.go` which appears to be the only place that we care about the kafka client that we use so far. Were you referring to something else?

atzoum:
@fracasula you may proceed as you see appropriate, I just want to make sure that you are confident this is really an abstraction layer that will pay dividends in the future and is worth the extra mile, instead of simply switching the client library and writing the same set of tests against the existing methods:

- kafka.NewProducer
- kafka.NewProducerForAzureEventHub
- kafka.NewProducerForConfluentCloud
- kafka.Produce

fracasula:
> > I think you're mixing things. In the `kafkamanager` we don't care what client we use, we just read from a source and write to a destination. Most tests in the `kafkamanager` can be done with a mocked client. In fact that's what I'm planning to do in the second PR.
> 
> I mentioned `kafkamanager.go` which appears to be the only place that we care about the kafka client that we use so far. Were you referring to something else?

I'm sorry, I'm not following :disappointed: perhaps we can have a quick call about this to clarify things.

> > The goal will be achieved with more than one PR, I thought we said to keep 'em small.
> 
> It would be useful to share the plan at this early stage

I thought I did, both in the daily scrum call and in the PR description. I apologize if this was not clear. I will try to make things more clear in the future. Thanks for pointing it out @atzoum :+1: 

> @fracasula you may proceed as you see appropriate, I just want to make sure that you are confident this is really an abstraction layer that will pay dividends in the future and is worth the extra mile, instead of simply switching the client library and writing the same set of tests against the existing methods:
> 
> * kafka.NewProducer
> * kafka.NewProducerForAzureEventHub
> * kafka.NewProducerForConfluentCloud
> * kafka.Produce

I do believe that it's worth it already. I'm very thourough with the tests and I hope we won't have to mix them up and have all in the stream manager. I'd like to keep the client tests separate and then have dedicated stream manager tests. I hope it'll be clearer once the 2nd PR comes in unless there's a strong feeling from the team that this approach shouldn't go through at all.

In regards to the `Azure` and `Confluent` cloud tests I was thinking that they can potentially live in the `client_test.go` here (but in another PR). I'll let you know once I progress more with the other PR.

fracasula:
@atzoum I pushed a `TestConfluentAzureCloud`. for now it doesn't run on CI but I'm working on it. I can run it locally but I had to create my own Confluent Cloud account to get an API Key/Secret pair.

Also I'm thinking that I can remove `ContainerOption`, `ProducerOption` and all the `With*` functions and just have the config struct exported. That is going to remove some code and make the meta-client way smaller.

utils/kafkaclient/producer.go
```diff
@@ -0,0 +1,182 @@
...
+// sends requests.
+func WithProducerClientID(clientID string) ProducerOption {
+	return withProducerOption{setup: func(c *producerConfig) {
+		c.clientID = clientID
+	}}
+}
+
+// WithProducerBatchTimeout sets the time limit on how often incomplete message batches will be flushed to
+//	// kafka.
+func WithProducerBatchTimeout(t time.Duration) ProducerOption {
```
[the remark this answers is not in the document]
fracasula: Good point, thanks! :+1:  I removed the configuration option. We can put it back if and when needed.

utils/kafkaclient/config.go
```diff
@@ -0,0 +1,124 @@
...
+		return "sha-512"
+	default:
+		panic(fmt.Errorf("scram hash generator out of the known domain %d", s))
+	}
+}
+
+type Config struct {
+	ClientID    string
+	DialTimeout time.Duration
+	TLS         *TLS
```
atzoum, line 38: The type of configuration that we support appears to be very specific:
https://www.rudderstack.com/docs/destinations/streaming-platforms/kafka/#apache-kafka

e.g. we don't appear to be supporting client certificates, only a CA certificate for SSL and SASL when SSL is enabled.

lvrach:
Haven't looked at the code in detail, but first of all great job on the testing effort.

@fracasula I will agree with @atzoum, regarding the necessity of an additional abstraction (kafkaclient).

I understand the benefit of a thin abstraction layer for Kafka that is decoupled for the more business-oriented `kafkamanager`. 

However, the `kafkaclient` includes code for functionality that we don't need (produce, list topics, etc) and an abstraction always has an extra cost. In this case, we have already paid the initial implementation cost the maintenance and readability is the one I worry about.

I would prefer if `kafkaclient` was tight to the needs of the stream manager, support only the need functionality and has a `kafkamanager` sub-package.

fracasula:
> @fracasula I will agree with @atzoum, regarding the necessity of an additional abstraction (kafkaclient).
> 
> I understand the benefit of a thin abstraction layer for Kafka that is decoupled for the more business-oriented `kafkamanager`.
> 
> However, the `kafkaclient` implements and tests functionality that we don't need (produce, list topics, etc) and abstractions always have an extra cost. In this case, we have already paid the initial implementation cost the maintenance and readability is the one I worry about. I am still not sure if the cost is justified in this case.

@lvrach half the code is used in tests, see `createTopic`, `listTopics`, `consumer.go` etcetera:
* `createTopic` is needed to automate the creation of a topic to be used in tests
* `listTopics` is used to make sure the topics needed are indeed there
* `consumer.go` is used to make sure that the producer actually sent the messages to Kafka and that Kafka got them
  * I'd like to make `docker_test.go` use it as well if it makes sense

If you don't consider the ones I listed above, what is actually left of the client is not much and it gives me the ability to properly test the `kafkamanager` by mocking the client when needed. It's basically just moving the logic away from there in another package, I don't think you'll save much code the other way around.

As I explained in previous comments this gives us a few benefits along with tests isolation. In the `kafkamanager` I'm planning to mostly put under tests:
* bootstrapping
* json marshalling
* status codes / errors returned by Produce

Everything else is under test here already.

> I would prefer if `kafkaclient` was tight to the needs of the stream manager, support only the needed functionality and live a `kafkamanager` sub-package.

@lvrach Sure, I can move it in the `kafkamanager` as a subpackage but then we won't be able to use it in `docker_test.go` for example. If that what you guys want please confirm and I can proceed with the relocation of the package.

fracasula:
@lvrach @atzoum as agreed during our call I will:
* move the client into a `kafkamanager` subpackage
* create test helpers for the createTopic and listTopics functions
* later on make changes to docker_test.go to make it use the client

utils/kafkaclient/config.go
fracasula: It will be the responsibility of the `kafkamanager` to use only the proper TLS configuration :+1: This will live there anyway as subpackage as agreed.

fracasula:
@lvrach @atzoum can you please have another look?

services/streammanager/kafka/client/producer.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"context"
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/segmentio/kafka-go"
+)
+
+type ProducerConfig struct {
```
atzoum, line 12: I believe we need another configuration field for the partitioner as well. Now, we are using `sarama.NewReferenceHashPartitioner` now, I assume this is the same with segment's `kafka.Murmur2Balancer`

services/streammanager/kafka/client/producer.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	if c.config.SASL != nil {
+		transport.SASL, err = c.config.SASL.build()
+		if err != nil {
+			return nil, fmt.Errorf("could not build SASL configuration: %w", err)
+		}
+	}
+
+	p = &Producer{
+		writer: &kafka.Writer{
```
atzoum, line 64: Judging by [this](https://github.com/segmentio/kafka-go/blob/main/writer.go#L796), I believe we should explicitly set `BatchTimeout` to something very small, e.g. `1*time.NanoSecond`.

atzoum (changes requested):

services/streammanager/kafka/client/config.go
```diff
@@ -0,0 +1,124 @@
...
+	Cert, Key,
+	CACertificate []byte
+	WithSystemCertPool,
+	InsecureSkipVerify bool
+}
+
+func (c *TLS) build() (*tls.Config, error) {
+	loadCerts := len(c.Cert) > 0 && len(c.Key) > 0
+
+	if !loadCerts && !c.InsecureSkipVerify && !c.WithSystemCertPool {
```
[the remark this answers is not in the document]
fracasula: Yes, good catch! :+1: Fix pushed.

services/streammanager/kafka/client/producer.go
fracasula: I think we should use `*kafka.Hash` to replace the `sarama.NewReferenceHashPartitioner` unless I'm misunderstanding something. That's why I hard-coded it in the producer [here](https://github.com/rudderlabs/rudder-server/blob/99fe96e42ccc97e9400000a585584d99dec999e7/services/streammanager/kafka/client/producer.go#L67). Why do you say we should use the murmur2 one, can you share the relevant docs?

services/streammanager/kafka/client/producer.go
fracasula: I thought it had to be greater than one to create issues. The default is `1` though and while testing I didn't encounter any delays, it seems to send messages synchronously.

> // When sending synchronously and the writer's batch size is configured to be
> // greater than 1, this method blocks until either a full batch can be assembled
> // or the batch timeout is reached.

But if we don't provide it then it defaults to 1s:

```go
func (w *Writer) batchTimeout() time.Duration {
	if w.BatchTimeout > 0 {
		return w.BatchTimeout
	}
	return 1 * time.Second
}
```

* https://github.com/segmentio/kafka-go/blob/main/writer.go#L796
* https://github.com/segmentio/kafka-go/blob/6691dda207ff50d6a50baed99748c620d569aaa9/writer.go#L577

It would be very weird that in order to make it sync people would have to come up with the 1µs setting btw, so it seems OK to me :thinking: 

WDYT?

services/streammanager/kafka/client/producer.go
atzoum: [here](https://github.com/Shopify/sarama/blob/main/partitioner.go#L170)
> NewReferenceHashPartitioner is like NewHashPartitioner except that it handles absolute values in the same way as the reference Java implementation. NewHashPartitioner was supposed to do that but it had a mistake and now there are people depending on both behaviors. This will all go away on the next major version bump.

[here](https://github.com/segmentio/kafka-go#java)
> Use the kafka.Murmur2Balancer balancer to get the same behaviour as the canonical Java client's default partitioner.

services/streammanager/kafka/client/producer.go
atzoum: The default BatchSize is [100](https://github.com/segmentio/kafka-go/blob/6691dda207ff50d6a50baed99748c620d569aaa9/writer.go#L782), so according to the documentation it will wait. Setting the `BatchSize` to 0 might make more sense, as long as this doesn't have as a side effect to send a varArg of messages in separate requests, which I don't think so.

internal/buildtag/cloud_on.go
```diff
@@ -0,0 +1,5 @@
+//go:build cloud
+
+package buildtag
+
+const Cloud = true
```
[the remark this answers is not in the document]
fracasula: It's a build tag used to flag tests that have a dependency on a cloud provider. They are a must on CI but I'm running them locally too. I guess locally they make sense if you're making changes on a piece of code that is tested by those.

What name do you propose?

docker_test.go
```diff
@@ -1632,3 +1632,7 @@ func initWHClickHouseClusterModeSetup(t *testing.T) {
 		}
 	}
 }
+
+type logger struct{}
+
+func (*logger) Log(args ...interface{}) { log.Println(args...) }
```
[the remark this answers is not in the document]
fracasula: Right, I replaced it there because we were using `log.Println` anyway. I'll look into the RS logger then thanks :+1:

services/streammanager/kafka/client/producer.go
fracasula: What the hell I confused size with timeout :laughing: Thanks man, I'll look into it then :+1:

services/streammanager/kafka/client/producer.go
fracasula: @atzoum https://github.com/segmentio/kafka-go/pull/906

fracasula: I'm using my own fork for now till we hear from them.

fracasula:
@lvrach @atzoum can you have another look please?

atzoum (approved):

internal/buildtag/cloud_on.go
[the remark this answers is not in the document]
lvrach: I was thinking we could keep them cloud provider-specific.

Like we have done here https://github.com/rudderlabs/rudder-server/blob/master/services/streammanager/bqstream/bqstreammanager_test.go#L24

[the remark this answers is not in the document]
fracasula: @lvrach there you're not using a tag though. I could remove it and rely solely on the env vars like in the test you linked. would that be OK?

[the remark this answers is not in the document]
fracasula: @lvrach done [here](https://github.com/rudderlabs/rudder-server/pull/1875/commits/07e040a45c791529c13724f38199062ccf1eb7b6).

lvrach (approved):

PR #1884, opened 2022-04-25
feat(jobsdb): support limiting jobs' pickup based on total payload size

Sidddddarth:
# Description
In addition to the events limit, a limit can now be placed on the total payload size of jobs. This new limit uses the same zero-value semantics as the events limit does, i.e. using 0 as `PayloadSizeLimit` disables the limit.

Additionally the following changes have been performed:

1. Refactored `GetQueryParamsT`, by renaming its limit-related fields & dropping some unused filters.
2. Refactored `JobsDB.DeleteExecuting`, dropped `GetQueryParamsT` from method parameter in order to fix the zero-value usage inconsistency of `JobsLimit`. Zero and less than zero in this case was disabling the limit whereas in all other cases it acted as a valid limit, i.e. no results were returned.
3. Fixed overflow behaviour of `EventsLimit` & `PayloadSizeLimit`. Now, we overflow the imposed limit only in the case where a single result is to be returned so as not to bring processing to a halt.
4. Fixed a bug where attemptNum, last job state etc. were omitted while scanning rows after union-querying.

## Configuration

The configuration option is reloadable per routerType, batchrouterType, processor.

Default limits have been calculated using a placeholder of 10kb/event:

- processor: 10KB * 10k = 100MB
- routers: 10KB * 10k = 100MB
- batchrouter: 10KB * 100k = 1GB

`Note:` In practice a json payload of 1K will require more than 1K (~ 20-30% more) to be stored as a JSONB column in Postgresql.

## Notion Ticket

- [Pick events from gateway jobsdb based on event size](https://www.notion.so/rudderstacks/Pick-events-from-gateway-jobsdb-based-on-event-size-7214ed5a0fe4468da3e01b9eb662f026)
- [Pick events from router jobsdb based on event size
](https://www.notion.so/rudderstacks/Pick-events-from-router-jobsdb-based-on-event-size-c993c8901e4b403aa5d534d9598f22f5)
- [Pick events from batch router jobsdb based on event size
](https://www.notion.so/rudderstacks/Pick-events-from-batch-router-jobsdb-based-on-event-size-c95a646ffb4b4d96992d1192a4e6c2ab)

## Security


jobsdb/jobsdb.go
```diff
@@ -84,6 +84,7 @@ type GetQueryParamsT struct {
 	StateFilters                  []string
 	JobCount                      int
 	EventCount                    int
+	PayloadSize                   *int64
```
atzoum, line 87: why `*int64` instead of `int`?

atzoum, line 87: minor: `MaxPayloadSize` or `PayloadLimit` might be a better word choice than `PayloadSize`

jobsdb/jobsdb.go
```diff
@@ -2341,13 +2347,24 @@ func (jd *HandleT) getProcessedJobsDS(ds dataSetT, getAll bool, limitCount int,
 			ds.JobTable, ds.JobStatusTable, stateQuery, customValQuery, sourceQuery, limitQuery)
 
 		args := []interface{}{getTimeNowFunc()}
+
+		var wrapQuery []string
 		if params.EventCount > 0 {
-			sqlStatement = fmt.Sprintf(`SELECT * FROM (`+sqlStatement+`) t WHERE running_event_counts - t.event_count + 1 <= $%d;`, len(args)+1)
 			// EXPLAIN `running_event_counts - t.event_count + 1`: If the event count limit "splits" a job we want this jobs to be returned.
 			//				`+1` prevents a job with event count of 1 to be returned.
```
atzoum, line 2354: I fail to understand the purpose of `+1`, let alone if this is indeed required in the `payload_size` query

jobsdb/jobsdb.go
```diff
@@ -3615,6 +3657,17 @@ func (jd *HandleT) getUnprocessed(params GetQueryParamsT) []*JobT {
 				break
 			}
 		}
+		if limitByPayloadSize {
+			sumPayloadSize := int64(0)
+			for _, j := range jobs {
+				sumPayloadSize += j.PayloadSize
+			}
+			*(params.PayloadSize) -= sumPayloadSize
+			// received event count could exceed the requested event count, by the spillover of the last selected job
```
atzoum, line 3666: please adapt the comment

jobsdb/unionQuery.go
```diff
@@ -39,7 +39,7 @@ type MultiTenantJobsDB interface {
 	GetPileUpCounts(map[string]map[string]int)
 }
 
-func (mj *MultiTenantHandleT) getSingleWorkspaceQueryString(workspace string, count int, ds dataSetT, params GetQueryParamsT, order bool) string {
+func (mj *MultiTenantHandleT) getSingleWorkspaceQueryString(workspace string, count int, payloadLimit int64, ds dataSetT, params GetQueryParamsT, order bool) string {
```
atzoum, line 42: why do we need the `payloadLimit` parameter and how is this different from `params.PayloadSize`?

jobsdb/unionQuery.go
```diff
@@ -83,17 +85,28 @@ func (mj *MultiTenantHandleT) GetAllJobs(workspaceCount map[string]int, params G
...
 
+	if params.PayloadSize != nil {
+		if *(params.PayloadSize) <= 0 {
+			return outJobs
+		}
+	}
+
+	workspacePayloadLimitMap := make(map[string]int64)
+	for workspace, count := range workspaceCount {
+		workspacePayloadLimitMap[workspace] = int64(float64(int64(count)*(*(params.PayloadSize))) / float64(params.JobCount))
```
atzoum, line 96: What is the logic behind the payload limit that we are trying to impose? What is the meaning of `params.PayloadSize`? The total maximum payload size of all messages or the maximum payload size of a single message? Why would we want to limit the maximum payload size of a single message?

Assuming that we have 10 workspaces and each workspace has 2 jobs (10%). If my total payload limit is 10MB, shouldn't I roughly set a maximum payload size of 1MB (10MB*10%) per workspace?

atzoum (changes requested):

atzoum:
@Sidddddarth what is the performance penalty of this extra query condition against a fully loaded dataset?

`<time of query with payload limit> - <time of query without payload limit>`

jobsdb/jobsdb.go
lvrach: I am responsible for this ambiguity, so let me try to explain:

The reason behind this: `running_event_counts - t.event_count + 1` is to deal with situations where a single item that the pickup size. For example, the max size is 10_000 bytes and we have an item with 11_000 bytes.

We have two options: pick or not pick the overflowing item. Because we wanted to play it safe, we prefer to pick an oversized item rather than pick nothing at all. Not being able to pick an item, due to an incorrect config or unexpected job size could lead to the processor being stuck.

We prefer availability, this way under any conditions, we ensure to pick at least one item. Rather than correctness: always return 

This is what `running_event_counts - t.event_count` is doing.

Now for the `+1`, if we used just `running_event_counts - t.event_count` we would always pick one more item even if the item size is 1 (which is not possible to overflow). 

The ideal way to handle this is to only allow overflow when it will otherwise lead to not picking a large item and getting stuck.
`running_event_counts - t.event_count + 1` was an easy way to compromise, but it hinders readability.

lvrach: For payload_size that `1` is not common/possible the `+1` can be removed.

jobsdb/unionQuery.go
Sidddddarth: >What is the meaning of params.PayloadSize?

it's a limit on the total payloads read from the db.

> shouldn't I roughly set a maximum payload size of 1MB (10MB*10%) per workspace?

That's the plan. each workspace gets a limit on payload size in proportion to the number of jobs we want to pick for that workspace.

jobsdb/jobsdb.go
Sidddddarth: Agreed. In hindsight, it shouldn't have made it in.

jobsdb/unionQuery.go
Sidddddarth: `params.PayloadSize` as it is right now is the total payload limit. For all workspaces.
`payloadLimit` is just for that particular `workspace`

jobsdb/jobsdb.go
Sidddddarth: In case of single-tenant server, the same params is passed to `GetToRetry`, `GetWaiting` and `GetUnprocessed`. I didn't want to have to go over the jobs again and update the payloadLimit, instead I chose to just update it while scanning the rows into jobs. A pointer made that convenient.
Ideally a pointer to the entire params can be passed, which would solve majority of such problems.( looking at having to do `toQuery -= len(retryList)` between `GetToRetry` and `GetWaiting` and `GetUnprocessed`

atzoum: Then this is something that needs to be made clear in the doc, at least.
My personal opinion is that it is unintuitive and maybe an anti-pattern to update a requested query parameter while scanning the results. If we need to communicate this information back to the caller, we could do this by including it in the return value(s)

jobsdb/unionQuery.go
atzoum: The only reason we are passing `GetQueryParamsT` appears to be for including `params.StateFilters[0]` in a log message.  Can we exchange `params GetQueryParamsT` with `state string`?

jobsdb/unionQuery.go
atzoum: I must have been confused then by the too many casts, isn't there anything we could do to simplify? e.g.

```
percentage := float64(count)/float64(params.JobCount)
payloadLimit := percentage*float64(*params.PayloadSize)
workspacePayloadLimitMap[workspace] = int64(payloadLimit)
```

jobsdb/jobsdb.go
Sidddddarth: correction: it's not being updated while scanning, but after the scan is complete(refer line 3663 below).
Also, please elaborate why it's an anti-pattern.

atzoum: Function parameters are usually best to be considered as immutable and a function communicates information to the caller using its output(s). Pure functions is something that we should aim for & pure functions don't mutate state or parameters.
A function mutating variables outside of its scope can cause unexpected results, thus is considered to be dangerous. Definitely, any function mutating its parameters should at least explicitly document this extensively.

As an everyday example, imagine that you go to a pizza shop and you write down your (peculiar) order: "I want up to 3 pizzas, but they shouldn't be any more than 1000kcal all together". What you expect from the shop is to get the receipt, your pizzas and maybe the number of their calories. You don't expect that they will get your order, strikethrough the number 1000kcal and replaced it with the remaining calories :) Wouldn't that be awkward?

jobsdb/unionQuery.go
Sidddddarth: Good catch.
In fact I think we can leave `params` out entirely considering it's getting all jobs and including one state in the log doesn't add much.
We are setting it ourselves anyway(`params.StateFilters = []string{NotProcessed.State, Waiting.State, Failed.State}`)

jobsdb/jobsdb.go
Sidddddarth: > running_event_counts <= $%d OR ROW_NUMBER = 1

say ds-1 has jobs with event_count 1,2,3,10 and ds-2 has jobs with event-count 4
in this case first three jobs from ds-1 and the first job from ds-2 will be picked, leaving 4th job in ds-1 behind.
this would lead to an event ordering issue..?

jobsdb/jobsdb.go
Sidddddarth: indeed
using `PayloadLimit`

jobsdb/jobsdb.go
atzoum: You are right, but still this is something that can be solved if getProcessedJobsDS returns an additional value indicating whether the limit has been reached or not, so that the caller knows when to stop and not rely on the result's size

Sidddddarth:
> @Sidddddarth what is the performance penalty of this extra query condition against a fully loaded dataset?
> 
> `<time of query with payload limit> - <time of query without payload limit>`

image


It's taking more or less the same time.
NO REAL DECREASE IN PERFORMANCE.

atzoum (approved):

router/batchrouter/batchrouter.go
```diff
@@ -1526,9 +1550,24 @@ func (worker *workerT) workerProcess() {
...
 					brt.logger.Debugf("BRT: %s: DB about to read for parameter Filters: %v ", brt.destType, parameterFilters)
-
-					retryList := brt.jobsDB.GetToRetry(jobsdb.GetQueryParamsT{CustomValFilters: []string{brt.destType}, JobCount: toQuery, ParameterFilters: parameterFilters, IgnoreCustomValFiltersInQuery: true})
-					unprocessedList := brt.jobsDB.GetUnprocessed(jobsdb.GetQueryParamsT{CustomValFilters: []string{brt.destType}, JobCount: toQuery, ParameterFilters: parameterFilters, IgnoreCustomValFiltersInQuery: true})
+					retryList := brt.jobsDB.GetToRetry(
+						jobsdb.GetQueryParamsT{
+							CustomValFilters:              []string{brt.destType},
+							JobsLimit:                     toQuery,
+							ParameterFilters:              parameterFilters,
+							IgnoreCustomValFiltersInQuery: true,
```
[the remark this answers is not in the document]
lvrach: Not sure about 1 and 2, but params should take care of part 3 if `readPerDestination` is set

router/batchrouter/batchrouter.go
```diff
@@ -1526,9 +1550,24 @@ func (worker *workerT) workerProcess() {
...
+							PayloadSizeLimit:              brt.payloadLimit,
+						},
+					)
+					unprocessedList := brt.jobsDB.GetUnprocessed(
```
lvrach, line 1562: We could probably use `.GetAll` methid here

lvrach (approved):

processor/stash/stash.go
```diff
@@ -234,9 +234,9 @@ func (st *HandleT) readErrJobsLoop(ctx context.Context) {
 			}
 			//NOTE: sending custom val filters array of size 1 to take advantage of cache in jobsdb.
 			toQuery := errDBReadBatchSize
-			retryList := st.errorDB.GetToRetry(jobsdb.GetQueryParamsT{CustomValFilters: []string{""}, JobCount: toQuery, IgnoreCustomValFiltersInQuery: true})
+			retryList := st.errorDB.GetToRetry(jobsdb.GetQueryParamsT{CustomValFilters: []string{""}, JobsLimit: toQuery, IgnoreCustomValFiltersInQuery: true})
 			toQuery -= len(retryList)
-			unprocessedList := st.errorDB.GetUnprocessed(jobsdb.GetQueryParamsT{CustomValFilters: []string{""}, JobCount: toQuery, IgnoreCustomValFiltersInQuery: true})
+			unprocessedList := st.errorDB.GetUnprocessed(jobsdb.GetQueryParamsT{CustomValFilters: []string{""}, JobsLimit: toQuery, IgnoreCustomValFiltersInQuery: true})
```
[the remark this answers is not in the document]
atzoum: it is easy to do it now

router/batchrouter/batchrouter.go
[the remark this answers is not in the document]
atzoum: If this is a query optimisation, e.g. don't include the `customVal` filter if a `destinationId` parameter filter is provided, then we could implement this in a better way, automatically so that it applies in all queries and we can drop this flag from the params

processor/processor.go
```diff
@@ -2441,7 +2446,7 @@ func throughputPerSecond(processedJob int, timeTaken time.Duration) int {
 }
 
 func (proc *HandleT) crashRecover() {
-	proc.gatewayDB.DeleteExecuting(jobsdb.GetQueryParamsT{CustomValFilters: []string{GWCustomVal}, JobCount: -1})
+	proc.gatewayDB.DeleteExecuting(GWCustomVal)
```
cisse21, line 2449: Can't we have the custom Val empty here just like we do it everywhere else?

jobsdb/unionQueryLegacy.go
```diff
@@ -5,20 +5,62 @@ type MultiTenantLegacy struct {
...
+	}
+}
+
+func limitsReached(eventsLimitEnabled, payloadLimitEnabled bool, params GetQueryParamsT) bool {
+	return params.JobsLimit <= 0 ||
+		(eventsLimitEnabled && params.EventsLimit <= 0) ||
+		(payloadLimitEnabled && params.PayloadSizeLimit <= 0)
+}
+
+func getTotalEvents(jobs []*JobT) int {
```
cisse21, line 52: Do we need this check since we call GetAllJobs only in router and every job is an event

jobsdb/unionQuery.go
```diff
@@ -83,11 +77,24 @@ func (mj *MultiTenantHandleT) GetAllJobs(workspaceCount map[string]int, params G
 	dsList := mj.getDSList(false)
 	outJobs := make([]*JobT, 0)
 
+	workspacePayloadLimitMap := make(map[string]int64)
+	for workspace, count := range workspaceCount {
+		percentage := math.Max(float64(count), 0) / float64(params.JobsLimit)
+		payloadLimit := percentage * float64(params.PayloadSizeLimit)
+		workspacePayloadLimitMap[workspace] = int64(payloadLimit)
```
cisse21, line 84: What would happen if the allotted payload size of a customer is always lesser than the size he has for a single job. Does it make sense to add a safety buffer here of say 1 job to be picked always?

jobsdb/unionQuery.go
```diff
@@ -150,42 +156,53 @@ func (mj *MultiTenantHandleT) getUnionDS(ds dataSetT, workspaceCount map[string]
 	for rows.Next() {
 		var job JobT
 		var _null int
-		columns, err := rows.Columns()
```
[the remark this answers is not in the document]
cisse21: me and @Sidddddarth  actually discussed this. There is no column named `LastJobStatus` in the joined table, so this block of the code never got executed and every job was being assumed as an unprocessed job and was retried till eternity(3hrs on hosted)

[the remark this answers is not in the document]
cisse21: The current approach does not differentiate between a processed and unprocessed job and populates the data based on the query results

jobsdb/unionQuery.go
Sidddddarth: this is enforced..
guaranteed minimum of 1 job, maximum up to the limit.

processor/processor.go
atzoum: Yes, if you believe this is safe, it would simplify even further the `DeleteExecuting` method!

cisse21: we can have the parameters in `DeleteExecuting` probably for any future use-cases but we don't need to do it at GW level since we have only one customVal

jobsdb/unionQuery.go
atzoum: To be exact, in `unionQuery`, whenever you are using a `PayloadLimit` we always overflow, i.e. we always return 1 more job than this limit allows

processor/stash/stash.go
```diff
@@ -234,9 +237,9 @@ func (st *HandleT) readErrJobsLoop(ctx context.Context) {
 			}
 			//NOTE: sending custom val filters array of size 1 to take advantage of cache in jobsdb.
 			toQuery := errDBReadBatchSize
-			retryList := st.errorDB.GetToRetry(jobsdb.GetQueryParamsT{CustomValFilters: []string{""}, JobCount: toQuery, IgnoreCustomValFiltersInQuery: true})
+			retryList := st.errorDB.GetToRetry(jobsdb.GetQueryParamsT{CustomValFilters: []string{""}, IgnoreCustomValFiltersInQuery: true, JobsLimit: toQuery, PayloadSizeLimit: payloadLimit})
 			toQuery -= len(retryList)
-			unprocessedList := st.errorDB.GetUnprocessed(jobsdb.GetQueryParamsT{CustomValFilters: []string{""}, JobCount: toQuery, IgnoreCustomValFiltersInQuery: true})
+			unprocessedList := st.errorDB.GetUnprocessed(jobsdb.GetQueryParamsT{CustomValFilters: []string{""}, IgnoreCustomValFiltersInQuery: true, JobsLimit: toQuery, PayloadSizeLimit: payloadLimit})
```
chandumlg, line 242: Don't we need to check if we have hit the limits before fetching unprocessed?

router/batchrouter/batchrouter.go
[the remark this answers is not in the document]
chandumlg: 1 is done for jobsdb cache to work properly I believe.

processor/stash/stash.go
atzoum: Yes, to be efficient though it needs further refactoring so that instead of a slice of jobs we return some more information, e.g. 

```
type JobsResult struct {
  Jobs          []*JobT
  EventsCount   int
  PayloadCount  int64
  LimitsReached bool
}
```
For the time being this should be enough though, but will open another task to implement it in the next sprint.

router/batchrouter/batchrouter.go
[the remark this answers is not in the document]
atzoum: Inside the cache we appear to be using `CustomValFilter` as a key and ignore the existence of `IgnoreCustomValFiltersInQuery`. 

Isn't the following scenario problematic?

1. Query using {CustomValFilters: "val"} returns 0 jobs and is cached.
2. Next query using {CustomValFilters: "val", IgnoreCustomValFiltersInQuery: true} would return some jobs if it went through the database but due to the cache we reply with 0 jobs.

PR #1885, opened 2022-04-25
fix: disable creation of views when dedup is enabled in BQ

Jayachand:
# Description

- Views not required in bqDedup method
- If switched to partition method from dedup
    - required users_view will be created while loading users table
    - views for new track events will be created
    - skipped views will be left alone. customer can create view himself or make a request to us

## Notion Ticket

https://www.notion.so/rudderstacks/skip-views-generation-when-dedup-is-enabled-in-BQ-2cbec577f93a4bc3abaea7b89c076c3e

## Security

### DIFF
--- a/warehouse/bigquery/bigquery.go
+++ b/warehouse/bigquery/bigquery.go
@@ -115,7 +115,9 @@ func (bq *HandleT) CreateTable(tableName string, columnMap map[string]string) (e
 		return
 	}
 
-	err = bq.createTableView(tableName, columnMap)
+	if !isDedupEnabled {
+		err = bq.createTableView(tableName, columnMap)
+	}
 	return
 }
 


warehouse/bigquery/bigquery.go
```diff
@@ -115,7 +115,9 @@ func (bq *HandleT) CreateTable(tableName string, columnMap map[string]string) (e
 		return
 	}
 
-	err = bq.createTableView(tableName, columnMap)
+	if !isDedupEnabled {
+		err = bq.createTableView(tableName, columnMap)
+	}
```
psrikanth88, line 120: for my understanding - excluding users table, customer needs to create views himself if the table was created when dedup is enabled and now switched to append method, right?

Jayachand: That's right

psrikanth88 (approved):

achettyiitr (approved):
LGTM

PR #1911, opened 2022-05-03
refactor(jobsdb): support executing commands in the context of an existing transaction

atzoum:
# Description

Refactoring jobsdb in order to support executing commands (append jobs & update job statuses) in the context of an existing transaction.

In the context of this refactoring a number of tasks have been performed, most importantly:
- introduce `WithTx`, `WithStoreSafeTx`, `WithUpdateSafeTx` methods in JobsDB
- rename, reorganise and document JobsDB methods (see JobsDB interface)
- drop `globalDBHandle`
- drop unnecessary utility methods from JobsDB
- all `InTxn` commands now respect the writerQueue configuration
- use `SAVEPOINT` where appropriate
- setup connection pool limits based on jobsdb configuration
- unexport private-at-heart types
- update enterprise-commit
- use semaphore channels instead of goroutine workers for implementing reader/writer queues

## Review Guide

1. Please start by reading the new `jobsdb.JobsDB` interface definition
2. Read the rest of the pull request changes, top-to-bottom. As you read these changes you will find comments in key sections which there to guide you and explain critical changes. 

## Notion Ticket

[Refactor jobsdb operations to execute in the context of an existing transaction](https://www.notion.so/rudderstacks/Refactor-jobsdb-operations-to-execute-in-the-context-of-an-existing-transaction-227d67e9a8fd4e6db2ca26f71f6817f3)

[Refactor jobsdb reader/writer queue](https://www.notion.so/rudderstacks/Refactor-jobsdb-reader-writer-queue-3063893b909345f29b3165eb659a0b6f)

## Security


atzoum:
A performance benchmark (see `BenchmarkJobsdb` in `jobsdb/integration_test.go`) comparing the previous implementation with the new one indicates no regression.

```markdown
# Before
BenchmarkJobsdb/store_and_consume_10000_jobs_using_1_stream(s)-10         	       1	17571726041 ns/op	576593440 B/op	 6211230 allocs/op
BenchmarkJobsdb/store_and_consume_10000_jobs_using_10_stream(s)-10        	       1	6995922292 ns/op	579987200 B/op	 6259936 allocs/op
BenchmarkJobsdb/store_and_consume_10000_jobs_using_200_stream(s)-10       	       1	6704573667 ns/op	586918048 B/op	 6332132 allocs/op
BenchmarkJobsdb/store_and_consume_10000_jobs_using_500_stream(s)-10       	       1	7406804292 ns/op	593137704 B/op	 6406599 allocs/op

# After
BenchmarkJobsdb/store_and_consume_10000_jobs_using_1_stream(s)-10         	       1	17258266875 ns/op	576617376 B/op	 6213160 allocs/op
BenchmarkJobsdb/store_and_consume_10000_jobs_using_10_stream(s)-10        	       1	5147047875 ns/op	580197320 B/op	 6264283 allocs/op
BenchmarkJobsdb/store_and_consume_10000_jobs_using_200_stream(s)-10       	       1	5060879625 ns/op	583470256 B/op	 6294680 allocs/op
BenchmarkJobsdb/store_and_consume_10000_jobs_using_500_stream(s)-10       	       1	6563832125 ns/op	589489216 B/op	 6378574 allocs/op


```

jobsdb/jobsdb.go
```diff
@@ -100,43 +100,149 @@ var getTimeNowFunc = func() time.Time {
...
+	// GetPileUpCounts returns statistics (counters) of incomplete jobs
+	// grouped by workspaceId and destination type
+	GetPileUpCounts(statMap map[string]map[string]int)
+
+	/* Admin */
 
 	Status() interface{}
-	GetIdentifier() string
+	IsHealthy() bool
 	DeleteExecuting(params GetQueryParamsT)
```
cisse21, line 233: Shouldn't this also come under Queries section?

atzoum: indeed `IsHealthy` performs a `SELECT` to check on the database's health but I don't believe that it qualifies as a valuable query, thus I grouped it under the `Admin` group, along with `Status` & `DeleteExecuting`, which are methods used for administrative purposes and not for actual business.

jobsdb/jobsdb.go
```diff
@@ -1532,64 +1554,97 @@ func (jd *HandleT) GetMaxDSIndex() (maxDSIndex int64) {
...
-	err = txn.Commit()
+	sqlStatement = fmt.Sprintf(`DROP TABLE IF EXISTS "%s"`, ds.JobStatusTable)
+	jd.prepareAndExecStmtInTx(tx, sqlStatement)
+	sqlStatement = fmt.Sprintf(`DROP TABLE IF EXISTS "%s"`, ds.JobTable)
+	jd.prepareAndExecStmtInTx(tx, sqlStatement)
+	err = tx.Commit()
 	jd.assertError(err)
 
+	// TODO: should we really do all these post-drop stuff during recovery
+	jd.postDropDs(ds)
```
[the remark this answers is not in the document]
chandumlg: During recovery not required.

jobsdb/jobsdb.go
```diff
@@ -118,43 +118,149 @@ var getTimeNowFunc = func() time.Time {
...
 
+// AppendSafeTx sealed interface
+type AppendSafeTx interface {
+	Tx() *sql.Tx
+	appendSafeTxIdentifier() string
+}
+
+type appendSafeTx struct {
+	tx       *sql.Tx
+	identity string
```
mcchran, line 129: is this an identifier or an identity?

jobsdb/jobsdb.go
```diff
@@ -118,43 +118,149 @@ var getTimeNowFunc = func() time.Time {
...
+	return r.identity
+}
+
+func (r *appendSafeTx) Tx() *sql.Tx {
+	return r.tx
+}
+
+// EmptyAppendSafeTx returns an empty interface usable only for tests
+func EmptyAppendSafeTx() AppendSafeTx {
+	return &appendSafeTx{}
```
mcchran, line 142: why we need that? I thought we can just mock the AppendSafeTx public interface in the tests, cant' we?

app/app.go
```diff
@@ -138,7 +138,7 @@ func HealthHandler(w http.ResponseWriter, r *http.Request, jobsDB jobsdb.JobsDB)
 	var dbService string = "UP"
 	var enabledRouter string = "TRUE"
 	var backendConfigMode string = "API"
-	if !jobsDB.CheckPGHealth() {
+	if !jobsDB.IsHealthy() {
```
lvrach, line 141: Since we change the name of the method, I will suggest naming it `.Ping` and return an error instead of bool

```go
Ping() error
```

config/config.yaml
```diff
@@ -90,6 +90,7 @@ JobsDB:
       failedOnly: false
   gw:
     enableWriterQueue: false
+    maxOpenConnections: 64
```
lvrach, line 93: I would suggest having gw specific defaults in our code, this way is not necessary for this or devops config to be there for the system to work with reasonable configuration.

gateway/gateway.go
```diff
@@ -304,9 +304,9 @@ func (gateway *HandleT) dbWriterWorkerProcess(process int) {
 		}
 
 		if gwAllowPartialWriteWithErrors {
-			errorMessagesMap = gateway.jobsDB.StoreWithRetryEach(jobList)
+			errorMessagesMap = gateway.jobsDB.AppendWithRetryEach(jobList)
 		} else {
-			err := gateway.jobsDB.Store(jobList)
+			err := gateway.jobsDB.Append(jobList)
```
lvrach, line 309: Some alternative names
`Append` -> `Push`, `Publish`, `Enqueue`

jobsdb/integration_test.go
```diff
@@ -789,100 +789,172 @@ func TestJobsDB_IncompatiblePayload(t *testing.T) {
...
+		totalJobs = 10000
+		// maxDsSize is the maximum ds size, controls how often a new DS will be created
+		maxDsSize = 500
+		// pageSize is the batch size for appending and retrieving jobs within each worker
+		pageSize = 10
+	)
+
+	concurrencies := []int{16, 64, 256, 512}
+
+	os.Setenv("RSERVER_JOBS_DB_MAX_DSSIZE", fmt.Sprintf("%d", maxDsSize))
```
lvrach, line 807: Prefer https://pkg.go.dev/testing#B.Setenv 
> Setenv calls os.Setenv(key, value) and uses Cleanup to restore the environment variable to its original value after the test.


```suggestion
	b.Setenv("RSERVER_JOBS_DB_MAX_DSSIZE", fmt.Sprintf("%d", maxDsSize))
```

jobsdb/integration_test.go
```diff
@@ -789,100 +789,172 @@ func TestJobsDB_IncompatiblePayload(t *testing.T) {
...
 	queryFilters := jobsdb.QueryFiltersT{
 		CustomVal: true,
 	}
 
-	jobDB.Setup(jobsdb.ReadWrite, false, "batch_rt", dbRetention, migrationMode, true, queryFilters)
-	defer jobDB.TearDown()
+	for _, concurrency := range concurrencies {
+		jobsDb1 := jobsdb.HandleT{}
+		os.Setenv("RSERVER_JOBS_DB_ENABLE_WRITER_QUEUE", "true")
+		os.Setenv("RSERVER_JOBS_DB_ENABLE_READER_QUEUE", "true")
```
lvrach, line 822:
```suggestion
		b.Setenv("RSERVER_JOBS_DB_ENABLE_WRITER_QUEUE", "true")
		b.Setenv("RSERVER_JOBS_DB_ENABLE_READER_QUEUE", "true")
```

jobsdb/integration_test.go
```diff
@@ -789,100 +789,172 @@ func TestJobsDB_IncompatiblePayload(t *testing.T) {
...
-				UserID:       "a-292e-4e79-9880-f8009e0ae4a3",
-				UUID:         uuid.Must(uuid.NewV4()),
-				CustomVal:    customVal,
-			}
-		}
+		jobsDb2 := jobsdb.HandleT{}
+		os.Setenv("RSERVER_JOBS_DB_ENABLE_WRITER_QUEUE", "false")
+		os.Setenv("RSERVER_JOBS_DB_ENABLE_READER_QUEUE", "false")
+		os.Setenv("RSERVER_JOBS_DB_GW_MAX_OPEN_CONNECTIONS", "64")
```
lvrach, line 833:
```suggestion
		b.Setenv("RSERVER_JOBS_DB_ENABLE_WRITER_QUEUE", "false")
		b.Setenv("RSERVER_JOBS_DB_ENABLE_READER_QUEUE", "false")
		b.Setenv("RSERVER_JOBS_DB_GW_MAX_OPEN_CONNECTIONS", "64")
```

services/metric/registry_test.go
```diff
@@ -158,6 +158,7 @@ func benchmarkRegistryGetCounterAndInc(b *testing.B, concurrency int) {
 	end.Add(concurrency)
 	for i := 0; i < concurrency; i++ {
 		go func() {
+			start.Wait()
```
lvrach, line 161: [question] why we are doing this  ? 🤔

router/batchrouter/batchrouter.go
```diff
@@ -422,15 +422,18 @@ func (brt *HandleT) pollAsyncStatus(ctx context.Context) {
 											statusList = append(statusList, status)
 										}
 										brt.failedJobCount.Count(len(statusList))
-										txn := brt.jobsDB.BeginGlobalTransaction()
-										brt.jobsDB.AcquireUpdateJobStatusLocks()
-										err = brt.jobsDB.UpdateJobStatusInTxn(txn, statusList, []string{brt.destType}, parameterFilters)
+										err := brt.jobsDB.WithUpdateSafeTx(func(tx jobsdb.UpdateSafeTx) error {
```
mcchran, line 425: Don't get the structure of the interface. `WithUpdateSafeTx` should afford the following signature:
`jobsDB.WithUpdateSafeTx(brt, func(tx jobsdb.UpdateSafeTx) error) error`

router/batchrouter/batchrouter.go
```diff
@@ -422,15 +422,18 @@ func (brt *HandleT) pollAsyncStatus(ctx context.Context) {
...
+											err = brt.jobsDB.UpdateJobStatusInTx(tx, statusList, []string{brt.destType}, parameterFilters)
```
mcchran, line 426: `UpdateJobStatusInTx` do we need to perform that unsafely? If not it makes sense to have the UpdateSafeTx as part of the `UpdateJobStatusInTx` of we should decorate the later to and make the decorated version public.

gateway/gateway_test.go
```diff
@@ -226,7 +226,7 @@ var _ = Describe("Gateway Enterprise", func() {
 		It("should accept events from normal users", func() {
 			allowedUserEventData := fmt.Sprintf("{\"batch\":[{\"userId\": \"%s\"}]}", NormalUserID)
 
-			mockCall := c.mockJobsDB.EXPECT().StoreWithRetryEach(gomock.Any()).DoAndReturn(jobsToEmptyErrors).Times(1)
+			mockCall := c.mockJobsDB.EXPECT().AppendWithRetryEach(gomock.Any()).DoAndReturn(jobsToEmptyErrors).Times(1)
```
mcchran, line 229: be aware here we do not test the arguments of AppendWithRetryEach, I assume we don't have to right?

jobsdb/integration_test.go
mcchran: FYI this cannot be used in prallel tests.

jobsdb/jobsdb.go
atzoum: Unfortunately we cannot mock this since it is a sealed interface

jobsdb/jobsdb.go
atzoum: what do you prefer? :)

gateway/gateway.go
atzoum: **Poll**

👍  Append
👎  Push
😄  Publish
🎉  Enqueue
❤️  Store

services/metric/registry_test.go
atzoum: so that nobody starts before the benchmark's timer starts, but not 100% sure it is needed

router/batchrouter/batchrouter.go
atzoum: which interface, `jobsdb.UpdateSafeTx` or `jobsdb.JobsDB`?

router/batchrouter/batchrouter.go
atzoum: can you provide an example?

jobsdb/jobsdb.go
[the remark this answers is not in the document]
atzoum: will delete

config/config.yaml
atzoum: I have started working on refactoring jobsdb initialisation & configuration, but as a separate task tracked [here](https://www.notion.so/rudderstacks/Refactor-jobsdb-intialisation-configuration-58e13e67e6194d0893fb8346ec38a4fa).

app/app.go
mcchran: It is not just a ping we perform here... but I like the ping error approach.

gateway/gateway.go
gane5hvarma: .Go - Tells jobslist to go in jobsdb 😁

jobsdb/integration_test.go
atzoum: yes, it is a benchmark so shouldn't be an issue

gateway/gateway.go
fracasula: I voted to keep `Store` although it is more generic as a name just to avoid confusion because inside it is processing db requests with a specific name and that name has not changed. It is still invoked as `store` [here](https://github.com/rudderlabs/rudder-server/blob/b5a90cfe67405e47db73557b6b997ec01c91d03c/jobsdb/jobsdb.go#L3742) and that is used for metrics.

I prefer to have a closer match between code and metrics, I would find it less confusing. If that is not the case please let me know.

atzoum: Will retain the old name since renaming would require changes in the metrics, grafana dashboards and it is not the main focus of this pull request

lvrach (approved):
Approved, thanks for taking care of all suggestions.

please remember to update `.enterprise-commit` before mergin

PR #1913, opened 2022-05-04
feat: upgrade ginkgo to V2

pChondros:
# Description

Ginkgo needs to be updated to V2 so that we can move forward with the upgrade of Golang to v1.18

## Notion Ticket

[rudder-server: Update ginkgo to v2 before upgrading to golang v1.18](https://www.notion.so/rudderstacks/rudder-server-Update-ginkgo-to-v2-before-upgrading-to-golang-v1-18-dca19a01ffd04edcb9b60dd2aa2c986f) 

## Security


.github/workflows/tests.yaml
```diff
@@ -116,7 +117,8 @@ jobs:
           ${{ runner.os }}-go-
     - run: go version
     - run: go mod download # Not required, used to segregate module download vs test times
-    - run: (cd /tmp && go get -u github.com/onsi/ginkgo/ginkgo@v1.16.5)
+    - run: (cd /tmp && go get -u github.com/onsi/ginkgo/v2 && go install -mod=mod github.com/onsi/ginkgo/v2/ginkgo@latest)
```
lvrach, line 120: We no longer need to keep backwards compatibility at this point, we can just use install
```suggestion
    - run: (cd /tmp && go install -mod=mod github.com/onsi/ginkgo/v2/ginkgo@latest)
```

atzoum (approved):

PR #1920, opened 2022-05-06
feat: kafkamanager with new client

fracasula:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security


go.mod
```diff
@@ -70,7 +69,7 @@ require (
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 )
 
-replace github.com/segmentio/kafka-go => github.com/fracasula/kafka-go v0.4.32-0.20220504142613-8272bc65c694
+replace github.com/segmentio/kafka-go => github.com/fracasula/kafka-go v0.4.32-0.20220509075905-7ccc623033ba
```
fracasula, line 72: Needed till this is merged: https://github.com/segmentio/kafka-go/pull/906

services/streammanager/kafka/client/producer.go
```diff
@@ -117,5 +131,20 @@ func (p *Producer) Publish(ctx context.Context, msgs ...Message) error {
 			Headers: headers,
 		}
 	}
-	return p.writer.WriteMessages(ctx, messages...)
+
+	var err error
+	for i := 0; i < p.config.MaxRetries; i++ {
+		err = p.writer.WriteMessages(ctx, messages...)
+		if errors.Is(err, kafka.LeaderNotAvailable) {
```
lvrach, line 138: We only sleep if the error is `kafka.LeaderNotAvailable`, why not sleeping in other cases as well ?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -82,269 +139,227 @@ func loadConfig() {
...
+	if err != nil {
+		return nil, fmt.Errorf("[Azure Event Hubs] Cannot create client: %w", err)
+	}
+	if err = c.Ping(context.TODO()); err != nil {
+		return nil, fmt.Errorf("[Azure Event Hubs] Cannot connect: %w", err)
+	}
 
-	return sarama.NewSyncProducer(hosts, conf)
+	return c.NewProducer(destConfig.Topic, client.ProducerConfig{
+		WriteTimeout:  o.Timeout,
```
atzoum, line 257: shouldn't we use `kafkaWriteTimeout` here?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -31,49 +33,104 @@ type Config struct {
...
+type logger interface {
+	Error(args ...interface{})
+	Errorf(format string, args ...interface{})
+}
 
-const (
-	azureEventHubUser = "$ConnectionString"
+var (
+	clientCert, clientKey                []byte
+	kafkaDialTimeout                     time.Duration
```
atzoum, line 109: shouldn't we use this value as the config's `DialTimeout`?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -354,78 +369,72 @@ func Produce(jsonData json.RawMessage, producer interface{}, destConfig interfac
...
-	batchedMessage, err := prepareBatchedMessage(topic, batch, timestamp)
+	timestamp := time.Now()
+	batchOfMessages, err := prepareBatchOfMessages(topic, batch, timestamp)
 	if err != nil {
-		return 400, "Failure", "Error while preparing batched message :: " + err.Error()
+		return 400, "Failure", "Error while preparing batched message: " + err.Error()
 	}
 
-	err = kafkaProducer.SendMessages(batchedMessage)
+	err = p.Publish(context.TODO(), batchOfMessages...)
```
atzoum, line 396: shouldn't we use here a timeout context using as timeout the `Opt.Timeout` passed in the producer function(s)?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -354,78 +369,72 @@ func Produce(jsonData json.RawMessage, producer interface{}, destConfig interfac
...
 	}
-	userID, _ := parsedJSON.Get("userId").Value().(string)
 
+	timestamp := time.Now()
+	userID, _ := parsedJSON.Get("userId").Value().(string)
 	message := prepareMessage(topic, userID, value, timestamp)
-
-	partition, offset, err := kafkaProducer.SendMessage(message)
-	if err != nil {
+	if err = p.Publish(context.TODO(), message); err != nil {
```
atzoum, line 421: shouldn't we use here a timeout context using as timeout the Opt.Timeout passed in the producer function(s)?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -82,269 +139,227 @@ func loadConfig() {
...
 	}
 
-	conf.Producer.Timeout = o.Timeout
+	c, err := client.NewAzureEventHubs(
+		destConfig.BootstrapServer, destConfig.EventHubsConnectionString, client.Config{},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("[Azure Event Hubs] Cannot create client: %w", err)
+	}
+	if err = c.Ping(context.TODO()); err != nil {
```
atzoum, line 252: no timeout here? shouldn't we use `kafkaDialTimeout` instead?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -354,78 +369,72 @@ func Produce(jsonData json.RawMessage, producer interface{}, destConfig interfac
...
+	return getStatusCodeFromError(err), returnMessage, err.Error()
 }
 
 // getStatusCodeFromError parses the error and returns the status so that event gets retried or failed.
 func getStatusCodeFromError(err error) int {
 	statusCode := 500
-
 	errorString := err.Error()
 
 	for _, s := range abortableErrors {
```
atzoum, line 440: shouldn't we adapt `abortableErrors` as well?
They should be different, e.g. here is what sarama has:
https://github.com/Shopify/sarama/blob/f07b7b87556741de750337199ccd02aa2a996fe2/errors.go
And here are segment's:
https://github.com/segmentio/kafka-go/blob/main/error.go

Shouldn't we abstract this as well on our generic client?

services/streammanager/kafka/client/producer.go
```diff
@@ -117,5 +131,20 @@ func (p *Producer) Publish(ctx context.Context, msgs ...Message) error {
 			Headers: headers,
 		}
 	}
-	return p.writer.WriteMessages(ctx, messages...)
+
+	var err error
+	for i := 0; i < p.config.MaxRetries; i++ {
```
atzoum, line 136: why aren't we using the `Writer.MaxAttempts` which uses [backoff](https://github.com/segmentio/kafka-go/blob/main/writer.go#L1089)? The default `Writer.MaxAttempts` is 10, so we are actually using it, so why do we need another retry mechanism?

docker_test.go
```diff
@@ -331,9 +321,10 @@ func run(m *testing.M) (int, error) {
 	cleanup := &testhelper.Cleanup{}
 	defer cleanup.Run()
 
-	KafkaContainer, err = destination.SetupKafka(pool, cleanup, destination.WithLogger(&testLogger{
-		logger.NewLogger().Child("kafka"),
-	}))
+	KafkaContainer, err = destination.SetupKafka(pool, cleanup,
```
atzoum, line 324: Some of us are using Apple with a M1 chip (arm64) and both `bitnami/kafka` & `bitnami/zookeeper` don't provide a docker image for `arm64`. As a result these tests delay dramatically and ultimately fail when running on a M1 chip which is bad development experience. If we cannot find a solution to that, then we should disable these tests when `runtime.GOARCH == arm64`
@pChondros @mcchran

services/streammanager/kafka/client/producer.go
fracasula: @lvrach I simply got this code from their docs (they retry on this error only or on `context.DeadlineExceeded`) but you are right, it can make sense to retry on other errors too.

From what I can see, errors that can make sense retry might be:
* NetworkException
* NotEnoughReplicas
* NotEnoughReplicasAfterAppend
* NotController
* KafkaStorageError
* FencedLeaderEpoch
* UnknownLeaderEpoch
* PreferredLeaderNotAvailable
* EligibleLeadersNotAvailable
* RequestTimedOut
* ListenerNotFound
* FetchSessionIDNotFound
* InvalidFetchSessionEpoch
* ElectionNotNeeded

In the segment lib they added a `Temporary() bool` method to errors which returns true
if those ops can be retried.

https://github.com/segmentio/kafka-go/blob/0b5e1f680537e6d0b6b754601f808e366f7df2ba/error.go#L138

All the errors I mentioned above are in the `Temporary() bool` so we could use that
but please consider that that method contains also non-producer errors and two errors
that are not retryable, at least according to the Kafka documentation:
* NotLeaderForPartition <- not retryable
* NoReassignmentInProgress <- not retryable

We could:
1. not care that `Temporary() bool` also has errors not returned by producers, open
   an issue on Segment asking for those two errors that can't be retried and simply
   use `if err.Temporary()` to determine whether we can retry the message
2. build our own `Temp` method

I'd say 1 but it requires more time. WDYT?

services/streammanager/kafka/kafkamanager.go
fracasula: The dial timeout is already used and injected via `client.Config` and honoured by `Ping()` but I'm going to build a context with the same timeout just in case.

services/streammanager/kafka/client/producer.go
atzoum: In the context of the `Producer.Publish` method I would say we shouldn't care about retries. Rudder-server will take care of aborting or retrying if necessary. Regarding whether an error is abortable or not, we need to honour what we are currently doing in `kafkaerror.go`

services/streammanager/kafka/kafkamanager.go
atzoum: @fracasula 
please have a look how we are dealing with it already in `bqstreammanager.go`

services/streammanager/kafka/client/producer.go
fracasula: Done :+1:

services/streammanager/kafka/client/producer.go
fracasula: It is done internally already by the `kafka.Writer`, I set up the `MaxAttempts` and it is working. If you guys are OK with that please resolve.

services/streammanager/kafka/kafkamanager.go
fracasula: @atzoum done, can you have another look?

services/streammanager/kafka/client/producer.go
```diff
@@ -12,7 +12,9 @@ import (
 type ProducerConfig struct {
 	ClientID string
 	WriteTimeout,
-	ReadTimeout time.Duration
+	ReadTimeout,
+	DefaultOpTimeout time.Duration
```
atzoum, line 16: major: we need to make sure this is never 0, i.e. set a default timeout of 60 seconds in `defaults()`
minor: `ProduceTimeout` sounds more straightforward to me.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -354,78 +379,73 @@ func Produce(jsonData json.RawMessage, producer interface{}, destConfig interfac
 		return makeErrorResponse(err) //returning 500 for retrying, in case of bad configuration
 	}
 
-	topic := conf.Topic
+	if conf.Topic == "" {
+		return makeErrorResponse(fmt.Errorf("invalid destination configuration: no topic"))
+	}
 
+	ctx := context.TODO()
```
atzoum, line 386: I fear we are using too many contexts without it being necessary. Why do we need a context in `kafkaclient.Producer.Publish()` since we set the publish timeout in its configuration?

lvrach: Context is not only used for timeout, the main concept behind context is being able to cancel IO/blocking operations in a hierarchical way.

As a rule of thumb, I would say every method/function should have a context, except if it non-blocking IO and we don't expect it to be in the future.

fracasula: Yes I agree with @lvrach:
> As a rule of thumb, I would say every method/function should have a context, except if it non-blocking IO and we don't expect it to be in the future.

services/streammanager/kafka/kafkamanager.go
fracasula: I'm thinking of removing our own `abortableErrors` slice and do this instead:

```go
// ...
if client.IsProducerErrTemporary(err) {
    statusCode = 400
    break
}
// ...
```

Then add the `isTemporary()` method to the producer and use the same logic as the segment lib retrier:

```go
var tempError interface{ Temporary() bool }

func IsProducerErrTemporary(err error) bool {
	isTransientNetworkError := errors.Is(err, io.ErrUnexpectedEOF) ||
		errors.Is(err, syscall.ECONNREFUSED) ||
		errors.Is(err, syscall.ECONNRESET) ||
		errors.Is(err, syscall.EPIPE)
	if isTransientNetworkError {
		return true
	} 
	if errors.As(err, &tempError) {
		return tempError.Temporary()
	}
	return false
}
```

I think it would make sense to remove `MaxRetries` as well and use only one retry logic. Thoughts?

docker_test.go
fracasula: Given that this work is already done and it would take too long to retest all configuration with different images, I think we should just wait till bitnami supports it: https://github.com/bitnami/charts/issues/7305

In the meantime what you are suggesting should do the trick (i.e. `if runtime.GOARCH == arm64 { t.Skip() }`.

About the tests being disabled on CI I think it is a mistake to do so. My guess is that all those tests running with `t.Parallel()` take a serious toll on the CI virtual that has limited capacity. I will test it without the parallelization (the tests will take longer to run but a lot less likely to timeout).

services/streammanager/kafka/kafkamanager.go
```diff
@@ -354,86 +399,75 @@ func Produce(jsonData json.RawMessage, producer interface{}, destConfig interfac
...
-	statusCode := getStatusCodeFromError(err) //400
-	errorMessage := err.Error()
+	returnMessage := fmt.Sprintf("%s error occurred.", err)
 	pkgLogger.Error(returnMessage)
-	return statusCode, returnMessage, errorMessage
+	return getStatusCodeFromError(err), returnMessage, err.Error()
 }
 
 // getStatusCodeFromError parses the error and returns the status so that event gets retried or failed.
 func getStatusCodeFromError(err error) int {
```
atzoum, line 468: Did some quick analysis and I can share the following:

These errors were previously considered to be non-retryable but now will be retried:

1. InvalidMessage
2. UnknownTopicOrPartition
3. NotLeaderForPartition
4. RequestTimedOut
5. NotEnoughReplicas


These errors were previously considered to be retryable but now will not be retried:

1. Unknown
2. OffsetOutOfRange
3. ReplicaNotAvailable
4. StaleControllerEpoch
5. RecordListTooLarge
6. InvalidRequiredAcks
7. IllegalGeneration
8. InconsistentGroupProtocol
9. RebalanceInProgress
10. InvalidCommitOffsetSize
11. InvalidTimestamp
12. UnsupportedSASLMechanism
13. IllegalSASLState
14. UnsupportedVersion
15. TopicAlreadyExists
16. InvalidPartitionNumber
17. InvalidReplicationFactor
18. InvalidReplicaAssignment
19. InvalidConfiguration
20. UnsupportedForMessageFormat
21. PolicyViolation
22. OutOfOrderSequenceNumber
23. DuplicateSequenceNumber
24. InvalidProducerEpoch
25. InvalidTransactionState
26. InvalidProducerIDMapping
27. InvalidTransactionTimeout
28. ConcurrentTransactions
29. TransactionCoordinatorFenced
30. TransactionalIDAuthorizationFailed
31. SecurityDisabled
32. BrokerAuthorizationFailed
33. LogDirNotFound
34. SASLAuthenticationFailed
35. UnknownProducerId
36. ReassignmentInProgress
37. DelegationTokenAuthDisabled
38. DelegationTokenNotFound
39. DelegationTokenOwnerMismatch
40. DelegationTokenRequestNotAllowed
41. DelegationTokenAuthorizationFailed
42. DelegationTokenExpired
43. InvalidPrincipalType
44. NonEmptyGroup
45. GroupIdNotFound
46. TopicDeletionDisabled
47. UnsupportedCompressionType
48. StaleBrokerEpoch
49. MemberIDRequired
50. GroupMaxSizeReached
51. FencedInstanceID
52. InvalidRecord 

I cannot find anything really concerning with such a change business-wise, nevertheless it is possible that such a change may disturb some customers, even temporarily.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -83,268 +153,243 @@ func loadConfig() {
...
+		cancel()
+		if err != nil {
+			continue
 		}
-		config.Net.SASL.Mechanism = sarama.SASLTypeSCRAMSHA256
-	default:
-		return fmt.Errorf("[Kafka] invalid SASL type %s", destConfig.SaslType)
+
+		p, err := c.NewProducer(destConfig.Topic, client.ProducerConfig{
+			WriteTimeout: kafkaWriteTimeout,
```
atzoum, line 228: in the [previous implementation](https://github.com/rudderlabs/rudder-server/blob/master/services/streammanager/kafka/kafkamanager.go#L102) we were using `kafkaWriteTimeout` for setting the client's `ReadTimeout` as well. Is this omitted intentionally in the new implementation?

atzoum (commented):
@fracasula I have left my last set of comments. Thank you for bearing with me and for the good job!

Since this pull request practically introduces a new kafka client library in `rudder-server`, replacing sarama, it would be useful to provide some more information on the pull request's description, such as:

1. Reasoning, why did we perform such a change.
2. Known behaviour changes that we are introducing, e.g. different treatment of errors pertaining to retry/abort logic. 
3. Whatever else you consider to be useful for the audience (future team members, opensource customers, etc.)

services/streammanager/kafka/kafkamanager.go
```diff
@@ -83,268 +153,243 @@ func loadConfig() {
...
 	for _, data := range batch {
-		message, err := json.Marshal(data["message"])
+		message, ok := data["message"]
+		if !ok {
+			// TODO bump metric
+			pkgLogger.Errorf("batch from topic %s is missing the message attribute", topic)
+			continue
+		}
+		userID, ok := data["userId"].(string)
+		if !ok && !allowReqsWithoutUserIDAndAnonymousID {
```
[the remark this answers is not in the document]
fracasula: Yes it is, we talked about here: https://rudderlabs.slack.com/archives/C01HTT66UMB/p1652080510264379?thread_ts=1651842112.134499&cid=C01HTT66UMB

services/streammanager/kafka/client/producer.go
```diff
@@ -117,5 +126,22 @@ func (p *Producer) Publish(ctx context.Context, msgs ...Message) error {
 			Headers: headers,
 		}
 	}
+
 	return p.writer.WriteMessages(ctx, messages...)
 }
+
+var tempError interface{ Temporary() bool }
+
+func IsProducerErrTemporary(err error) bool {
```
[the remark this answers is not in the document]
fracasula: > and consider the kafka.WriteErrors as temporary only if all contained errors are temporary themselves.

I'm not sure about this because if only one is temporary then we would return 400 and that message won't be retried leading to data loss, correct?

services/streammanager/kafka/kafkamanager.go
fracasula: Yes, I'm thus applying the default. If we want it I can add a real setting for it in the kafkamanager.

services/streammanager/kafka/client/producer.go
```diff
@@ -117,5 +126,22 @@ func (p *Producer) Publish(ctx context.Context, msgs ...Message) error {
...
+		errors.Is(err, syscall.ECONNREFUSED) ||
+		errors.Is(err, syscall.ECONNRESET) ||
+		errors.Is(err, syscall.EPIPE)
+	if isTransientNetworkError {
+		return true
+	}
+	if errors.As(err, &tempError) {
+		return tempError.Temporary()
+	}
+	return false
```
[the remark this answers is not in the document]
fracasula: I think the alternative is to treat all errors as retryable, even errors where a message really can't be processed, for example if you push a 20MB message and the `message.max.bytes` on Kafka was lower. This would be a new behaviour that could add sensible stress to the system in certain situations. Consider that we retry messages for up to 3 hours before dropping them so it could also make us susceptible to attacks. In my opinion it is best to stick to the Kafka documentation and treat only the errors that they say are temporary as such.

services/streammanager/kafka/client/producer.go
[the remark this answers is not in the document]
atzoum: correct, retrying as long as there is at least 1 temporary errors would be better 👍

services/streammanager/kafka/kafkamanager.go
atzoum: The current readTimeout is 2s, whereas we will now be using 10s. If this extra delay of 8s is worthy and won't cause unwanted delays in the stream's processing, then it's fine :)

fracasula: I'm not sure 2s makes a lot of sense, I'll ask around about the rationale behind it just to be sure, but it feels like 10s is more appropriate especially since some Kafkas might not be on premise.

services/streammanager/kafka/client/producer.go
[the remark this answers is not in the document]
fracasula: Fix pushed :+1:

services/streammanager/kafka/kafkamanager.go
fracasula: `kafkaReadTimeout` introduced, if we want we can change it accordingly via configuration now.

lvrach:
@fracasula I would change the title to `feat: use new client for kafkamanager to support timeouts`

fracasula:
@atzoum @lvrach additional info about manual tests is available [here](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1652274725706709).

services/streammanager/kafka/client/producer.go
```diff
@@ -61,13 +65,15 @@ func (c *Client) NewProducer(topic string, producerConf ProducerConfig) (p *Prod
...
 	p = &Producer{
+		config: producerConf,
 		writer: &kafka.Writer{
 			Addr:                   c,
 			Topic:                  topic,
 			Balancer:               &kafka.ReferenceHash{}, // TODO replace: https://github.com/segmentio/kafka-go/pull/906
 			BatchTimeout:           time.Nanosecond,
 			WriteTimeout:           producerConf.WriteTimeout,
 			ReadTimeout:            producerConf.ReadTimeout,
+			MaxAttempts:            1,
```
atzoum, line 76: Although previously we were [using 3](https://github.com/Shopify/sarama/blob/main/config.go#L478), I am fine with 1 as well!

atzoum (approved):

services/streammanager/kafka/kafkamanager.go
```diff
@@ -1,25 +1,27 @@
...
 	"time"
 
-	"github.com/Shopify/sarama"
 	"github.com/tidwall/gjson"
-	"github.com/xdg/scram"
 
 	"github.com/rudderlabs/rudder-server/config"
-	"github.com/rudderlabs/rudder-server/utils/logger"
+	"github.com/rudderlabs/rudder-server/services/streammanager/kafka/client"
+	rslogger "github.com/rudderlabs/rudder-server/utils/logger"
```
cisse21, line 16: Any specific reason we had to name it as rslogger?

fracasula: Clash with local interface.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -31,320 +33,385 @@ type Config struct {
...
-	case "sha512":
-		config.Net.SASL.SCRAMClientGeneratorFunc = func() sarama.SCRAMClient {
-			return &xdgSCRAMClient{HashGeneratorFcn: SHA512}
+	var c *client.Client
+	// The only component that supports multiple hosts is the consumer which we are not using here.
+	// It is also worth mentioning that internally the list of brokers should be populated accordingly anyway given
+	// that Kafka uses the "advertised.listeners" property to let clients know where they can connect.
+	// see: https://kafka.apache.org/documentation/#brokerconfigs_advertised.listeners
+	for _, host := range strings.Split(destConfig.HostName, ",") {
+		c, err = client.New("tcp", host+":"+destConfig.Port, clientConf)
```
cisse21, line 235: Here we are only connecting to the first available host and attempting to send events to the host. If the host goes down all the events would fail since we call NewProducer only once for every destination type right. Is this part of the code consistent with the sarama implementation?

fracasula: @cisse21 have you read the comment right above the line you're commenting on? 

Consider that these servers are just used for the initial connection to discover the full cluster membership (which may change dynamically) so what we pass there does not need to contain the full set of servers as it will be discovered anyway. The problem would be if we can't connect  and don't try the others but that's not what we're doing here, right?

cisse21: I read the comment but just wanted to make sure what I understood was right. Thanks for the detailed reply 👍

cisse21 (approved):

services/streammanager/kafka/kafkamanager.go
fracasula: @cisse21 can you have another looksie please? I made a small change to reintroduce the brokers list [here](https://github.com/rudderlabs/rudder-server/pull/1920/commits/f438dd0a7124df1af59a2794bf8582b528311624).

PR #1922, opened 2022-05-06
fix(tooling): add prometheus endpoints for OSS users

mcchran:
# Description

To enable metrics to while deploying locally through docker compose
we added statsd-to-prometheus metrics exporter. In this sense we can
have prometheus metrics at: `localhost:9102/metrics`.

Also include env variables to project's docker-compose file.

## Notion Ticket

https://www.notion.so/rudderstacks/Provide-or-disable-metrics-for-open-source-users-c6abd45ba9b24acea0a80d042f1be5ff

## Security

### DIFF
--- a/rudder-docker.yml
+++ b/rudder-docker.yml
@@ -12,6 +12,7 @@ services:
   backend:
     depends_on:
       - db
+      - metrics-exporter
       - d-transformer
     image: rudderlabs/rudder-server:1-alpine
     entrypoint: sh -c '/wait-for db:5432 -- /rudder-server'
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -21,15 +22,28 @@ services:
...
     environment:
       - JOBS_DB_HOST=db
+      - JOBS_DB_USER=rudder
+      - JOBS_DB_PORT=5432
+      - JOBS_DB_DB_NAME=jobsdb
+      - JOBS_DB_PASSWORD=password
+      - DEST_TRANSFORM_URL=http://d-transformer:9090
+      - CONFIG_BACKEND_URL=https://api.rudderlabs.com
+      - WORKSPACE_TOKEN=<your_workspace_token>
+      - STATSD_SERVER_URL=metrics-exporter:9125
```
lvrach, line 32: Why do we need those? Since they are specified here: `build/docker.env`

mcchran: I think we should do not maintain both rudder-docker.yaml and docker-compose.yaml. This is only to trigger that discussion. I think we should keep these two files aligned and eventually to get rid of rudder-docker.yaml in favour of docker-compose.yaml. Plus for local development we should introduce a docker-compose.dev.yaml according to the best practices and what we do wit the value files in helm charts.

mcchran: to ease the process I reverted these changes since they may be considered out of the scope of the specific PR

atzoum (approved):
@mcchran is this a fix, feat or chore? :)

mcchran:
@atzoum that's a chore that fixes things ... hard to say but let's do that a chore!

rudder-docker.yml
```diff
@@ -26,23 +27,28 @@ services:
...
-
-
-
-  
\ No newline at end of file
+  metrics-exporter:
+    image:
+      prom/statsd-exporter:v0.22.4
+    ports:
+      - "9102:9102"
```
chandumlg, line 54: Shouldn't this be 9125?

mcchran: This is to get the metrics --> `localhost:9125/metrics` into the browser to be screpable from a promethrus instance out of the specific docker-compose network. 
9125 is the destiantion of udp metrics coming from the docker specified "services" i.e. server, transformer. No need to be exported. Moreover is should't because we need that  never conflict with 9125 port of the host.

chandumlg: Got it.

chandumlg (approved):

PR #1923, opened 2022-05-09
chore(warehouse): remove tracking plan error logs.

achettyiitr:
# Description

Remove error logs from tracking plan as it will be going to create a lot of fuss in the logs.

## Notion Ticket

https://www.notion.so/rudderstacks/Remove-error-logs-from-tracking-plan-as-it-will-be-going-to-create-a-lot-of-fuss-in-the-logs-7161ca7e0032492da906be0af17e0531

## Security

### DIFF
--- a/processor/trackingplan.go
+++ b/processor/trackingplan.go
@@ -27,9 +27,6 @@ func reportViolations(validateEvent *transformer.TransformerResponseT, trackingP
 	validationErrors := validateEvent.ValidationErrors
 	output := validateEvent.Output
 
-	pkgLogger.Errorf("%d errors reported", len(validationErrors))
-	pkgLogger.Error(validationErrors)
-
 	eventContext, castOk := output["context"].(map[string]interface{})
 	if castOk {
 		eventContext["trackingPlanId"] = trackingPlanId


cisse21 (approved):

abhimanyubabbar (approved):
Following questions on the error logs:
1. What's the current frequency of error logging ?
2. Would removing the error logs cause issues in our ability to detect errors or resolve issues quickly ? ( Mainly trying to understand how would we RCA the issue ? )

achettyiitr:
> Following questions on the error logs:
> 
> 1. What's the current frequency of error logging ?
> For this I am nor sure, But this logging is going to take place for all the events which will fail for the validation which can be hugs. We added this as part of debugging during the development and kind of missed it I guess.

> 2. Would removing the error logs cause issues in our ability to detect errors or resolve issues quickly ? ( Mainly trying to understand how would we RCA the issue ? )
> We already have the some form of stats and logs around rudder-transformer. Some more stats around the rudder-server as well. If we need to debug, we send these violation errors in the downstream side as well(proc errors and customers warehouse). It would be sufficient to debug the issue I guess.
> https://github.com/rudderlabs/rudder-transformer/blob/498dab25f286da59fe620189786fb745b3f9fc7d/versionedRouter.js#L190
> https://github.com/rudderlabs/rudder-server/blob/f900211d56cb578152940922e18f088269b40cfd/processor/trackingplan.go#L78

PR #1925, opened 2022-05-11
chore(multi-tenant): revert changes related to EnableMultitenancy config variable

kuldeep0020:
# Description
Earlier in PR#1882, we removed the `EnableMultitenancy` config variable and modified the `UseFairPickup` method to not use that. Reverting these changes in order to maintain backward compatibility. 

## Related PR

[PR](https://github.com/rudderlabs/rudder-server/commit/6c04f9913e19d68ff5cd3fbb720ee1e7568faf76)

## Security


config/config.yaml
```diff
@@ -246,3 +246,4 @@ PgNotifier:
   retriggerCount: 500
   trackBatchInterval: 2s
   maxAttempt: 3
+EnableMultitenancy: true
```
atzoum, line 249: we don't need the old config, just the support for the old config in our code

cisse21 (approved):

lvrach (approved):

config/config.yaml
chandumlg: removed.

chandumlg (approved):

PR #1928, opened 2022-05-11
fix repetitive bugsnag alert for a single crash

saurav-malani:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/utils/misc/misc.go
+++ b/utils/misc/misc.go
@@ -47,10 +47,13 @@ import (
 	"github.com/thoas/go-funk"
 )
 
-var AppStartTime int64
-var errorStorePath string
-var reservedFolderPaths []*RFP
-var jsonfast = jsoniter.ConfigCompatibleWithStandardLibrary
+var (
+	AppStartTime        int64
+	errorStorePath      string
+	reservedFolderPaths []*RFP
+	jsonfast            = jsoniter.ConfigCompatibleWithStandardLibrary
+	notifyOnce          sync.Once
+)
 
 const (
 	// RFC3339Milli with milli sec precision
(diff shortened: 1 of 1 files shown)


utils/misc/misc.go
```diff
@@ -47,10 +47,13 @@ import (
...
-var AppStartTime int64
-var errorStorePath string
-var reservedFolderPaths []*RFP
-var jsonfast = jsoniter.ConfigCompatibleWithStandardLibrary
+var (
+	AppStartTime        int64
+	errorStorePath      string
+	reservedFolderPaths []*RFP
+	jsonfast            = jsoniter.ConfigCompatibleWithStandardLibrary
+	notifyOnce          sync.Once
```
atzoum, line 55: Under which scenario do we get more than 1 bugsnag notifications per crash?

Do we need a different `sync.Once` per team?

utils/misc/misc.go
```diff
@@ -1247,16 +1250,18 @@ func WithBugsnagForWarehouse(fn func() error) func() error {
...
 	return func() {
 		if r := recover(); r != nil {
-			defer bugsnag.AutoNotify(ctx, bugsnag.SeverityError, bugsnag.MetaData{
-				"GoRoutines": {
-					"Number": runtime.NumGoroutine(),
-				},
-				"Team": {
-					"Name": team,
-				},
+			notifyOnce.Do(func() {
```
atzoum, line 1253: missing defer

saurav-malani: since, we have moved `RecordAppError` inside, I don't think we need defer anymore.

utils/misc/misc.go
saurav-malani: not sure, but I don't think we need more than 1 bugsnag notification per crash. My line of thought is, currently we can have two values for `Team` `Warehouse` & `Core`. Both, seems completely independent.

utils/misc/misc.go
atzoum: fine, as long as `RecordAppError` doesn't panic!

atzoum (approved):

PR #1937, opened 2022-05-12
fix: enabling earlier usage of stats in other components

fracasula:
# Description

Fix to allow usage of stats in Kafka manager. `kafka.Init()` was called after `stats.Setup()` causing a `panic`.

## Security


atzoum (approved):

kuldeep0020 (approved):
LGTM

PR #1938, opened 2022-05-12
feat: introducing rsources package, JobService interface

Sidddddarth:
# Description

First installment of rsources package. Introducing JobService.

## Notion Ticket

[Introduce rsources.JobService](https://www.notion.so/rudderstacks/Introduce-rsources-JobService-b9f0dd9c8cf444939132f7ddda3cc2f8)

## Security


services/rsources/mapper.go
```diff
@@ -0,0 +1,66 @@
...
+				ID: key.TaskRunId,
+			})
+			taskRunIdIndex[key.TaskRunId] = idx
+		} else {
+			idx = taskRunIdIndex[key.TaskRunId]
+		}
+
+		var source_idx int
+		if _, ok := sourceIdIndex[key.SourceId]; !ok {
+			source_idx = len(status.TasksStatus[idx].SourcesStatus)
```
lvrach, line 26: [minor] Avoid python case. (Didn't the linter complaint)
```suggestion
		var sourceIDx int
		if _, ok := sourceIdIndex[key.SourceId]; !ok {
			sourceIDx = len(status.TasksStatus[idx].SourcesStatus)
```

services/rsources/mapper_test.go
```diff
@@ -0,0 +1,198 @@
+package rsources
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestStatusFromQueryResult(t *testing.T) {
```
lvrach, line 10: [non-blocking] Consider using table tests for this kind of test.

services/rsources/multitenant.go
```diff
@@ -0,0 +1,50 @@
+package rsources
+
+import (
+	"context"
+	"database/sql"
+)
+
+type multitenantExtension struct {
```
lvrach, line 8: [minor] Just wondering why we called `Extension` and not `Repository`

services/rsources/extension.go
```diff
@@ -0,0 +1,130 @@
+package rsources
+
```
lvrach, line 2: [minor] To keep things tight I would create a new folder / sub-package named `extensions` or `repos` / `repositories`.

Then `extension.go` and `multitenant.go` could be moved there.

services/rsources/extension.go
```diff
@@ -0,0 +1,130 @@
...
+	localDB *sql.DB
+}
+
+func (r *defaultExtension) getReadDB() *sql.DB {
+	return r.localDB
+}
+
+func (r *defaultExtension) createStatsTable(ctx context.Context, jobRunId string) error {
+	dbHost := config.GetEnv("JOBS_DB_HOST", "localhost")
+	jobRunIdStatTableName := fmt.Sprintf("%s%s", tablePrefix, jobRunId)
```
lvrach, line 32: What is the cardinality of `jobRunId` ?
How many tables do we expect to have at any given point in time?

services/rsources/extension.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (r *defaultExtension) doCleanupTables(ctx context.Context) error {
+	db := r.localDB
+	tables, err := getAllTableNames(ctx, db)
+	if err != nil {
+		return err
+	}
+
+	before := time.Now().Add(-24 * time.Hour)
```
lvrach, line 73: I believe this should be configurable

services/rsources/handler.go
```diff
@@ -0,0 +1,154 @@
...
+		return JobStatus{}, err
+	}
+	defer rows.Close()
+
+	statMap := make(map[JobTargetKey]Stats)
+	for rows.Next() {
+		var statKey JobTargetKey
+		var stat Stats
+		err := rows.Scan(
+			&statKey.SourceId, &statKey.DestinationId, &statKey.TaskRunId,
```
lvrach, line 74: [minor] is more idiomatic in Go to capitalise all the leters of an acronym:
`SQL` instead of `Sql` , `ID` instead of `Id` etc ..

```suggestion
			&statKey.SourceID, &statKey.DestinationID, &statKey.TaskRunID,
```

services/rsources/handler.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// IncrementStats checks for stats table and upserts the stats
+func (sh *sourcesHandler) IncrementStats(ctx context.Context, tx *sql.Tx, jobRunId string, key JobTargetKey, stats Stats) error {
+	err := sh.checkForTable(ctx, jobRunId)
+	if err != nil {
+		return err
+	}
+
+	jobRunIdStatTableName := fmt.Sprintf("%s%s", tablePrefix, jobRunId)
```
lvrach, line 96: This could lead to an SQL injection. Let's make sure we sanitise and validate `jobRunId` and test the attack surface. Using user-provided data as table names is tricky

lvrach (changes requested):
I am a bit sceptical regarding one table per jobID approach. I want to make sure that this number is going to be bounded, and Postgres will be ok handling the number of different tables. We have seen performance issues around Postgres caching when having multiple tables active.

Also, jobRunId can be used for an SQL injection attack

services/rsources/extension.go
atzoum: https://rudderlabs.slack.com/archives/C017H2XKZ9R/p1653390164299659

services/rsources/handler.go
atzoum: yes, as long as we make sure we remove any double quotes from the string, I don't see any other possibility for SQL injection.

services/rsources/multitenant.go
atzoum: it is not about everything related to the database/repository, it is just about specific extension points. E.g. the select statement remains the same, it is just that it gets executed against a different database. Incrementing stats is also not part of the extension since it is common, but `createStatsTable` is, since we need to do more stuff in the multitenant scenario.

services/rsources/extension.go
atzoum: I propose to use this structure: `rsources/internal/extensions/extension.go`
We would need to export the `extension` interface and its producer methods, so let's use the `internal` keyword to highlight that it is not intended for actual use by third-parties.

services/rsources/extension.go
Sidddddarth: simply replaced it with `config.GetDuration("JOBS_CLEANUP_INTERVAL", 24, time.Hour)`

atzoum (approved):

lvrach (approved):

PR #1939, opened 2022-05-12
chore(ci): add actions to check .enterprise-commit

lvrach:
# Description

## Issue

When making changes to `rudder-server-enterprise` we need to make sure to:
1. Set `.enterpise-commit` to a commit that points at the branch we working on, otherwise the tests wont pass.
2. Merge the enterprise first
3. Update `.enterpise-commit`
4. Then merge `rudder-server`

step 3, is what a lot of us forgot to do. Causing trouble for other PRs.

### CI
Add a check to ensure that our `.enterpise-commit` points to the `origin/master`.

This is the output when things go wrong:

Screenshot 2022-05-12 at 11 15 31 PM

### Local

`make enterprise-is-at-master` can be use locally to get the same information

## Notion Ticket

[Notion Link ](https://www.notion.so/rudderstacks/Check-enterpise-commit-907566aced5e49c5b58ec84078e2f4a0)

## Security


.enterprise/scripts/is-at-master.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
```
atzoum, line 1: isn't it simpler to get the latest commit in master and then compare it against the `.enterprise-commit` ?

```
git ls-remote $ENTERPRISE_REPO | grep refs/heads/master | cut -f 1
```

lvrach: this can be done as well, I am not sure how to get a nice output in this case, however.

lvrach: I will actually make it this way since it is faster and simpler. I can sacrifice the nice error output.

atzoum: so if I am working on a feature branch with changes in the enterprise repo, this check will fail until the very last moment?
I only wish we were able to do fast-forward merges on GitHub's pull requests...! :)

chandumlg (approved):

.enterprise/scripts/is-at-master.sh
lvrach: Ok, I made the changes. Yes, it will fail until the last moment the way we currently do it.

It is a separate action and it takes 5sec to run so it shouldn't create any more trouble other than the ❌ 


We can discuss ff merges but given the situation I think we need this

atzoum: it is not possible to have ff merges in GitHub according to the docs.

lvrach: good point, will check

lvrach: We are using [pull_request](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) as a trigger, it is using a merged ref to run the actions on.

fracasula (approved):

PR #1943, opened 2022-05-13
feat(multi-tenant): cluster manager gateway int

kuldeep0020:
# Description

Introduced dynamic cluster manager in Gateway App also which was needed to start the backend config. 

## Notion Ticket

[Notion](https://www.notion.so/rudderstacks/Include-Cluster-Manager-in-Gateway-App-Handler-e3d4dfd8dd6d4a29a278237bfa1d6b98)

## Security

### DIFF
--- a/app/apphandlers/gatewayAppHandler.go
+++ b/app/apphandlers/gatewayAppHandler.go
@@ -95,7 +95,6 @@ func (gatewayApp *GatewayApp) StartRudderCore(ctx context.Context, options *app.
 	dm := cluster.Dynamic{
 		Provider:         modeProvider,
 		GatewayComponent: true,
-		BackendConfig:    backendconfig.DefaultBackendConfig,
 	}
 	g.Go(func() error {
 		return dm.Run(ctx)
--- a/app/apphandlers/gatewayAppHandler.go
+++ b/app/apphandlers/gatewayAppHandler.go
@@ -3,6 +3,10 @@ package apphandlers
 import (
 	"context"
 	"fmt"
+	"github.com/rudderlabs/rudder-server/app/cluster"
+	"github.com/rudderlabs/rudder-server/app/cluster/state"
+	"github.com/rudderlabs/rudder-server/utils/types/deployment"
+	"github.com/rudderlabs/rudder-server/utils/types/servermode"
 	"net/http"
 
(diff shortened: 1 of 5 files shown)


app/apphandlers/gatewayAppHandler.go
```diff
@@ -65,6 +69,38 @@ func (gatewayApp *GatewayApp) StartRudderCore(ctx context.Context, options *app.
...
+			modeProvider = state.NewStaticProvider(servermode.DegradedMode)
+		}
+	default:
+		return fmt.Errorf("unsupported deployment type: %q", deploymentType)
+	}
+
+	dm := cluster.Dynamic{
+		Provider:         modeProvider,
+		GatewayComponent: true,
+		BackendConfig:    backendconfig.DefaultBackendConfig,
```
lvrach, line 98:
```suggestion
```

config/backend-config/backend-config.go
```diff
@@ -217,6 +217,13 @@ type CommonBackendConfig struct {
 	ctx              context.Context
 	cancel           context.CancelFunc
 	blockChan        chan struct{}
+	once             sync.Once
+}
+
+func (bc *CommonBackendConfig) init() {
+	bc.once.Do(func() {
+		bc.eb = pubsub.NewPublishSubscriber(context.TODO())
```
fracasula, line 225: Being this a dependency, shouldn't be passed/created when the backend config is created? This way it would be easier to mock and we will also avoid locking on a mutex each time we call `StartWithIDs` (see `Once.Do()`). Thoughts?

kuldeep0020: This is a temporary solution, this needs to be refactored more. cc: @lvrach

kuldeep0020: [Related ticket](https://www.notion.so/rudderstacks/Refactor-Common-Backend-Config-to-cater-Publish-Subscriber-in-a-better-way-and-make-it-more-compatib-0da64a6b37974b69ade1a8e583f01d4d)

fracasula (approved):

config/backend-config/backend-config.go
lvrach: Agree with @fracasula, we can do it with the refactoring.

cisse21 (approved):

PR #1948, opened 2022-05-16
refactor: pubsub .Subscribe() adding ctx and channel ownership 

lvrach:
# Description

## Context

A refactor in pubsub/backend config `.Subscribe` method was long due, since the implementation of a graceful shutdown.

The reasons behind this refactoring:

### context
Improve `context` and cancellation support. 

The existing `ctx` only stopped the goroutine, but was not closing the channel. Also, it was not possible to cancel/stop a single subscription.

### channel ownership

Channel ownership. In the existing implementation, the API consumer had to create and pass the channel. The problems with that:
1. **Control over channel type**. We can not prevent someone from creating a buffered channel, but is important for our implementation to use an unbuffered channel.
2. **Responsibility for closing the channel**. In our use case, it was only safe for the pubsub to close the channel. However, having one component to be responsible for channel creation and the other for closing is an anti-pattern. 
3. **Stylistic** / Readability. An extra line of code was needed every time

### New interface 

```go 
Subscribe(ctx, topic) ch
```

## Notion Ticket

https://www.notion.so/rudderstacks/Refactor-pubsub-0e427490fb0441189920e5690e0240f4

## Security


app/apphandlers/setup.go
```diff
@@ -176,40 +174,34 @@ func monitorDestRouters(ctx context.Context, routerFactory *router.Factory, batc
 	routerFactory.RouterDB.DeleteExecuting()
 	batchRouterFactory.RouterDB.DeleteExecuting()
 
-loop:
-	for {
-		select {
-		case <-ctx.Done():
-			break loop
-		case config := <-ch:
```
fracasula, line 184: It looked more correct in the previous version with the `select`. How come you decided to just range over `ch`?
What happens if the context gets canceled and there is no value on `ch`? Won't we be hanging?

app/apphandlers/setup.go
```diff
@@ -176,40 +174,34 @@ func monitorDestRouters(ctx context.Context, routerFactory *router.Factory, batc
...
-					} else {
-						_, ok := dstToRouter[destination.DestinationDefinition.Name]
-						if !ok {
-							pkgLogger.Info("Starting a new Destination ", destination.DestinationDefinition.Name)
-							router := routerFactory.New(destination.DestinationDefinition)
-							router.Start()
-							cleanup = append(cleanup, router.Shutdown)
-							dstToRouter[destination.DestinationDefinition.Name] = router
-						}
+	for config := range ch {
```
fracasula, line 177: [minor] the variable `config` collides with the imported package, could be annoying.

utils/pubsub/pubsub.go
```diff
@@ -117,17 +150,18 @@ func (r *subPublisher) publish(data *DataEvent) {
...
+		r.lastValueLock.Unlock()
+		if v != nil {
+			r.channel <- *v
 		}
 	}
+	close(r.channel)
+}
+
+func (r *subPublisher) close() {
+	r.bgCtx.Done()
```
atzoum, line 165: does calling `bgCtx.Done()` here have any effect?

utils/pubsub/pubsub.go
```diff
@@ -98,6 +130,7 @@ type subPublisher struct {
 // internal goroutine if it is not already started
 func (r *subPublisher) publish(data *DataEvent) {
 	r.startedOnce.Do(func() {
+		r.bgCtx = context.Background()
```
atzoum, line 133: what is the point of having a background context which cannot be canceled? this case will never materialize with a background context:
```
case <-r.bgCtx.Done():
```

utils/pubsub/pubsub_test.go
```diff
@@ -9,116 +9,81 @@ func TestPubSub_two_consumers_slow_and_fast(t *testing.T) {
...
-		t.Errorf("Expected normal consumer to have consumed 2, instead got %d.", v)
-	}
-
-	bus.Publish(topic, 3)
-	v = <-normal.out
-	if v != 3 {
-		t.Errorf("Expected normal consumer to have consumed 3, instead got %d.", v)
+	v := <-normal
+	if v.Data.(int) != 1 {
+		t.Errorf("Expected slow consumer to have consumed 5, instead got %d.", v.Data.(int))
```
atzoum, line 20:
```suggestion
		t.Errorf("Expected normal consumer to have consumed 1, instead got %d.", v.Data.(int))
```

utils/pubsub/pubsub_test.go
```diff
@@ -9,116 +9,81 @@ func TestPubSub_two_consumers_slow_and_fast(t *testing.T) {
...
 	}
-
-	bus.Publish(topic, 4)
-	v = <-normal.out
-	if v != 4 {
-		t.Errorf("Expected normal consumer to have consumed 4, instead got %d.", v)
+	v = <-slow
+	if v.Data.(int) != 1 {
+		t.Errorf("Expected slow consumer to have consumed 5, instead got %d.", v.Data.(int))
```
atzoum, line 24:
```suggestion
		t.Errorf("Expected slow consumer to have consumed 1, instead got %d.", v.Data.(int))
```

utils/pubsub/pubsub_test.go
```diff
@@ -9,116 +9,81 @@ func TestPubSub_two_consumers_slow_and_fast(t *testing.T) {
...
-	if v != 5 {
-		t.Errorf("Expected normal consumer to have consumed 5, instead got %d.", v)
+	t.Log("normal consumer gets ahead by 4 publishes")
+	numbers := []int{2, 3, 4, 5}
+	for _, n := range numbers {
+		bus.Publish(topic, n)
+		// normal consumer should consume 1, 2, 3, 4 & 5
+		v := <-normal
+		if v.Data.(int) != n {
+			t.Errorf("Expected normal consumer to have consumed 1, instead got %d.", v.Data.(int))
```
atzoum, line 34:
```suggestion
			t.Errorf("Expected normal consumer to have consumed %n, instead got %d.", n, v.Data.(int))
```

utils/pubsub/pubsub_test.go
```diff
@@ -9,116 +9,81 @@ func TestPubSub_two_consumers_slow_and_fast(t *testing.T) {
...
 	}
 
 }
 
-type testConsumer struct {
-	name string
-	in   chan DataEvent
-	out  chan int
-}
+func TestPubSub_Close(t *testing.T) {
```
atzoum, line 67: It would be useful to add another test where we are cancelling the subscription's context

app/apphandlers/setup.go
fracasula: > I understand that the expectation would be that when `ctx.Done`, `ch` will be closed as well

Yeah I understand that, what I don't get is why we are doing this change. What do we gain by doing it this way beside just deleting a few lines of code? `Subscribe` lives in a different package and a different package might even be in another repo or have bugs we don't know about as far as I'm concerned. So unless we gain something by not honouring the context and hoping somebody else will, I would keep it simple and honour the context.

fracasula: Additionally, the fact that you feel you need a comment on `Subscribe` to document this I think it supports my argument above.

utils/pubsub/pubsub.go
```diff
@@ -15,18 +13,8 @@ type DataEvent struct {
 // DataChannel is a channel which can accept an DataEvent
 type DataChannel chan DataEvent
```
atzoum, line 14:
```suggestion
type DataChannel <-chan DataEvent
```

utils/pubsub/pubsub.go
```diff
@@ -55,13 +43,15 @@ func (eb *publishSubscriber) Publish(topic string, data interface{}) {
...
 
-func (eb *publishSubscriber) Subscribe(topic string, ch DataChannel) {
+func (eb *PublishSubscriber) Subscribe(ctx context.Context, topic string) DataChannel {
 	eb.subscriptionsMutex.Lock()
 	defer eb.subscriptionsMutex.Unlock()
 	eb.lastEventMutex.RLock()
 	defer eb.lastEventMutex.RUnlock()
 
-	newSubPublisher := &subPublisher{channel: ch, ctx: eb.ctx}
+	ch := make(DataChannel)
```
atzoum, line 52:
```suggestion
	ch := make(chan DataEvent)
```

app/apphandlers/setup.go
lvrach: I believe this implementation is simpler and safer. Not everyone will no to use a 
```go
select {
		case <-ctx.Done():
			break loop
		case config := <-ch:
}
```

When reading from a channel. I much prefer the default and correct way of using this to be as simple as possible.

lvrach: Btw I was inspired by [etcd API](https://github.com/etcd-io/etcd/blob/client/v3.5.4/client/v3/watch.go#L75) and I preferred all the other approaches.

My other option would be a blocking call with a callback as a parameter.

fracasula: Resource leaking is a valid argument but my expectation generally when I see things like `.Subscribe` is that you would that object like most clients, for example:

```go
client := NewClient()
defer client.Shutdown(shutdownCtx) //  in main.go

ch, err := client.Subscribe(ctx, "something")
if err != nil {
    // ...
}

for {
    select {
        case <-ctx.Done():
            return
        case msg := <-ch:
            // do something with msg
    }
}
```

If there is no "shutdown" then I see your approach as valid to avoid leaking. Either way I think they are both valid approaches :+1:

atzoum:
@kuldeep0020 [this]([app/gomock_reflect_2556226072/prog.bin](https://github.com/rudderlabs/rudder-server/pull/1948/files#diff-891b4300fcb9980a1fdd09bf6e2547cbcabd63fb5bfecd2d2e1144bf7758b64c)) should not be included in git, correct?

atzoum (changes requested):

utils/tests/async-helper.go
```diff
@@ -60,6 +60,19 @@ func (helper *AsyncTestHelper) ExpectAndNotifyCallbackWithName(name string) func
 	}
 }
 
+// ExpectAndNotifyCallback Adds one to this helper's WaitGroup, and provides a callback that calls Done on it.
+// Should be used for gomock Do calls that trigger via mocked functions executed in a goroutine.
+func (helper *AsyncTestHelper) ExpectAndNotifyCallbackWithNameOnce(name string) func(...interface{}) {
+	var s sync.Once
+	f := helper.ExpectAndNotifyCallbackWithName(name)
+	return func(...interface{}) {
```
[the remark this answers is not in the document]
kuldeep0020: done

utils/tests/async-helper.go
```diff
@@ -60,6 +60,19 @@ func (helper *AsyncTestHelper) ExpectAndNotifyCallbackWithName(name string) func
 	}
 }
 
+// ExpectAndNotifyCallback Adds one to this helper's WaitGroup, and provides a callback that calls Done on it.
+// Should be used for gomock Do calls that trigger via mocked functions executed in a goroutine.
+func (helper *AsyncTestHelper) ExpectAndNotifyCallbackWithNameOnce(name string) func(...interface{}) {
```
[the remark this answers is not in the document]
kuldeep0020: done

router/router.go
```diff
@@ -2331,8 +2330,7 @@ func (rt *HandleT) Shutdown() {
 }
 
 func (rt *HandleT) backendConfigSubscriber() {
-	ch := make(chan pubsub.DataEvent)
-	rt.backendConfig.Subscribe(ch, backendconfig.TopicBackendConfig)
+	ch := rt.backendConfig.Subscribe(context.TODO(), backendconfig.TopicBackendConfig)
 	for {
 		config := <-ch
```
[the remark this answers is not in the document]
kuldeep0020: done

router/customdestinationmanager/customdestinationmanager.go
```diff
@@ -324,6 +326,7 @@ func (customManager *CustomManagerT) backendConfigSubscriber() {
 				}
 			}
 		}
+		customManager.mapLock.Unlock()
```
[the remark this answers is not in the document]
kuldeep0020: done

router/customdestinationmanager/customdestinationmanager.go
```diff
@@ -312,10 +315,9 @@ func (customManager *CustomManagerT) BackendConfigInitialized() <-chan struct{}
 
 func (customManager *CustomManagerT) backendConfigSubscriber() {
 	var once sync.Once
-	ch := make(chan pubsub.DataEvent)
-	backendconfig.Subscribe(ch, "backendConfig")
-	for {
-		config := <-ch
+	ch := backendconfig.Subscribe(context.TODO(), "backendConfig")
+	for config := range ch {
+		customManager.mapLock.Lock()
```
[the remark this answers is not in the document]
kuldeep0020: done

router/router.go
```diff
@@ -2331,10 +2330,9 @@ func (rt *HandleT) Shutdown() {
 }
 
 func (rt *HandleT) backendConfigSubscriber() {
-	ch := make(chan pubsub.DataEvent)
-	rt.backendConfig.Subscribe(ch, backendconfig.TopicBackendConfig)
-	for {
-		config := <-ch
+	ch := rt.backendConfig.Subscribe(context.TODO(), backendconfig.TopicBackendConfig)
+	for config := range ch {
+		//config := <-ch
```
atzoum, line 2335:
```suggestion
		
```

utils/pubsub/pubsub_test.go
atzoum, line 20:
```suggestion
		t.Errorf("Expected normal consumer to have consumed 1, instead got %d.", v.Data.(int))
```

atzoum (approved):

PR #1954, opened 2022-05-17
fix(multi-tenant): etcd Key case fix

kuldeep0020:
# Description
Made the ETCD keys upper case. 

## Notion Ticket

[<Notion Link >](https://www.notion.so/rudderstacks/Change-workspace-ID-change-ETCD-Key-74b92e2a62664132816a4f4475563c60)

## Security

### DIFF
--- a/app/cluster/state/etcd_test.go
+++ b/app/cluster/state/etcd_test.go
@@ -125,11 +125,11 @@ func Test_ServerMode(t *testing.T) {
 	provider := state.ETCDManager{
 		Config: &state.ETCDConfig{
 			Endpoints:   etcdHosts,
-			Namespace:   "test",
+			ReleaseName: "test",
 			ServerIndex: "0",
 		},
 	}
-	modeRequestKey := fmt.Sprintf("/%s/SERVER/%s/MODE", provider.Config.Namespace, provider.Config.ServerIndex)
+	modeRequestKey := fmt.Sprintf("/%s/SERVER/%s/MODE", provider.Config.ReleaseName, provider.Config.ServerIndex)
 	defer provider.Close()
 
 	t.Run("key is missing initially", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


config/backend-config/multitenant_workspaces.go
```diff
@@ -74,6 +74,11 @@ func (workspaceConfig *MultiTenantWorkspacesConfig) Get(workspaces string) (Conf
 
 // getFromApi gets the workspace config from api
 func (workspaceConfig *MultiTenantWorkspacesConfig) getFromAPI(workspaceArr string) (ConfigT, bool) {
+	// added this to avoid unnecessary calls to backend config and log better until workspace IDs are not present
+	if workspaceArr == workspaceConfig.Token {
```
lvrach, line 78: maybe we need to init with an empty string instead of `workspaceConfig.Token`.
This check here takes some time to understand and would be easy to break with a refactor

fracasula (approved):

config/backend-config/multitenant_workspaces.go
kuldeep0020: done

app/cluster/state/etcd_test.go
```diff
@@ -249,14 +249,14 @@ func Test_Workspaces(t *testing.T) {
...
-				Namespace:   "test_ack_timeout",
+				ReleaseName: "test_ack_timeout",
 				ServerIndex: "0",
 				ACKTimeout:  time.Duration(1),
 			},
 		}
 		defer provider.Close()
-		appType := strings.ToUpper(config.GetEnv("APP_TYPE", app.PROCESSOR))
-		requestKey := fmt.Sprintf("/%s/server/%s/%s/workspaces", provider.Config.Namespace,
+		appType := strings.ToLower(config.GetEnv("APP_TYPE", app.PROCESSOR))
```
lvrach, line 258: This need to be a lower or upper ?

config/backend-config/multitenant_workspaces.go
cisse21: This will never be equal right ? Since Token is essentially `HOSTED_MULTITENANT_SERVICE_SECRET` ?

app/cluster/state/etcd_test.go
kuldeep0020: This is lower 😭 😭

config/backend-config/multitenant_workspaces.go
kuldeep0020: right, so the first time when `startWithIds` gets called from main then we call it with the token (due to 3 type of backend config possibilities).

cisse21: So the setup call from main would be having this right? and if it's the same then we would skip

cisse21 (approved):

lvrach (approved):

PR #1958, opened 2022-05-17
refactor(jobsdb): return extended query information

saurav-malani:
# Description
When querying jobsdb for jobs, apart from the list of jobs, more information is returned now such as:

1. **LimitsReached**: `true` if any of the provided query limits has been reached (jobs, events, payload), `false` otherwise.
2. **EventsCount** : The number of events contained within the returned jobs.
3. **PayloadSize**: The total payload size of the returned jobs.

Based on this above information, it is now possible for clients to stop querying if a limit has been reached, otherwise, easily update the `QueryParameter` limits for the subsequent query.

## Notion Ticket

https://www.notion.so/rudderstacks/Refactor-JobsDB-to-return-rich-query-information-7335becbec5c40e1afa7025379c5939a

## Security


router/batchrouter/batchrouter.go
```diff
@@ -1572,7 +1575,8 @@ func (worker *workerT) workerProcess() {
 							PayloadSizeLimit:              brt.payloadLimit,
 						},
 					)
-					unprocessedList := brt.jobsDB.GetUnprocessed(
+					// TODO: shouldn't we be decremening JobsLimit & PayloadSizeLimit for the second query?
```
atzoum, line 1578: @chandumlg @cisse21 is this intentional here?

lvrach (approved):

atzoum (approved):

router/batchrouter/batchrouter.go
```diff
@@ -1563,29 +1566,25 @@ func (worker *workerT) workerProcess() {
...
+						ParameterFilters:              parameterFilters,
+						IgnoreCustomValFiltersInQuery: true,
+						PayloadSizeLimit:              brt.payloadLimit,
+					}
+					toRetry := brt.jobsDB.GetToRetry(queryParams)
+					combinedList = toRetry.Jobs
+					if !toRetry.LimitsReached {
+						queryParams.JobsLimit -= len(toRetry.Jobs)
+						if queryParams.PayloadSizeLimit > 0 {
+							queryParams.PayloadSizeLimit -= toRetry.PayloadSize
```
chandumlg, line 1581: How will jobsdb behave if PayloadSizeLimit turns out -ve?

(or there won't be any such scenario?)

atzoum: This will not happen in practive since the `PayloadSize` of the previous response cannot be greater (or equal) than the `PayloadSizeLimit` without `LimitsReached` being `true`.
Setting `PayloadSizeLimit` to any valye `<= 0` effectively disables the limit

chandumlg: 👍

chandumlg (approved):

PR #1962, opened 2022-05-18
chore(ci): add stale branch clean up support

mcchran:
# Description

Automatically delete stale branches.
Stale branch: 
    - 2 months with no commit
    - no open PR relates
    - non main/master/release prefixed branches

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/.github/workflows/housekeeping.yaml
+++ b/.github/workflows/housekeeping.yaml
@@ -34,7 +34,7 @@ jobs:
             repo_token: ${{ github.token }}
             date: '2 months ago'
             dry_run: false
-            delete_tags: true
-            extra_protected_branch_regex: ^(main|master|release|rudder-saas)
+            delete_tags: false
+            extra_protected_branch_regex: ^(main|master|release.*|rudder-saas)$
             exclude_open_pr_branches: true
 
--- a/.github/workflows/housekeeping.yaml
+++ b/.github/workflows/housekeeping.yaml
@@ -5,7 +5,8 @@ on:
     - cron: '42 1 * * *'
 
 jobs:
-  stale:
+  prs:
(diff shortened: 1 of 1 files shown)


.github/workflows/housekeeping.yaml
```diff
@@ -21,3 +22,19 @@ jobs:
...
+      steps:
+        - name: Checkout repository
+          uses: actions/checkout@v2
+        - name: Run delete-old-branches-action
+          uses: beatlabs/delete-old-branches-action@v0.0.9
+          with:
+            repo_token: ${{ github.token }}
+            date: '2 months ago'
+            dry_run: false
+            delete_tags: true
```
atzoum, line 37: what does `delete_tags` mean? We should never delete git tags.

.github/workflows/housekeeping.yaml
```diff
@@ -21,3 +22,19 @@ jobs:
...
+            extra_protected_branch_regex: ^(main|master|release|rudder-saas)
```
atzoum, line 38:
```suggestion
            extra_protected_branch_regex: ^(main|master|release.*|rudder-saas)$
```

.github/workflows/housekeeping.yaml
mcchran: oops true story.... What it does is it treats all tags as branches -- meaning it does delete all tags older than 2 months. I am disabling that for now but we can consider performing a tag clean up using the 
`extra_protected_tag_regex` in order to protect only release tags (we rarely perform a `git push --tags` ...  but our open-source community can do so right?)

atzoum (approved):

lvrach (approved):

PR #1963, opened 2022-05-18
fix(jobsdb): sanitize jsonb values before inserting them to the database

atzoum:
# Description

Removing null characters `\u0000` from json messages before storing them to the database, since PostgreSQL cannot handle them.
More information about this known issue can be found [here](https://www.commandprompt.com/blog/null-characters-workarounds-arent-good-enough/).

## Performance
Using `bytes.ReplaceAll` due to its performance characteristics compared to `strings.ReplaceAll` and `regexp.Regexp.ReplaceAllString`. 

```
BenchmarkSanitizeJson/SanitizeUsingStrings_string_of_size_4000_with_null_characters-10         	  178659	      5813 ns/op	   15104 B/op	       4 allocs/op
BenchmarkSanitizeJson/SanitizeUsingBytes_string_of_size_4000_with_null_characters-10           	  247652	      4760 ns/op	    7552 B/op	       2 allocs/op
BenchmarkSanitizeJson/SanitizeUsingRegexp_string_of_size_4000_with_null_characters-10          	   63981	     18764 ns/op	   26695 B/op	      14 allocs/op
BenchmarkSanitizeJson/SanitizeUsingStrings_string_of_size_4000_without_null_characters-10      	  604234	      1978 ns/op	   12288 B/op	       3 allocs/op
BenchmarkSanitizeJson/SanitizeUsingBytes_string_of_size_4000_without_null_characters-10        	  981765	      1330 ns/op	    8192 B/op	       2 allocs/op
BenchmarkSanitizeJson/SanitizeUsingRegexp_string_of_size_4000_without_null_characters-10       	  360448	      3275 ns/op	   20762 B/op	       6 allocs/op

BenchmarkSanitizeJson/SanitizeUsingStrings_string_of_size_4000000_with_null_characters-10         	     835	   1264664 ns/op	16023588 B/op	       4 allocs/op
BenchmarkSanitizeJson/SanitizeUsingBytes_string_of_size_4000000_with_null_characters-10           	    1880	    647296 ns/op	 8011792 B/op	       2 allocs/op
BenchmarkSanitizeJson/SanitizeUsingRegexp_string_of_size_4000000_with_null_characters-10          	     460	   2522459 ns/op	35741791 B/op	      55 allocs/op
BenchmarkSanitizeJson/SanitizeUsingStrings_string_of_size_4000000_without_null_characters-10      	    1272	    923880 ns/op	12017696 B/op	       3 allocs/op
BenchmarkSanitizeJson/SanitizeUsingBytes_string_of_size_4000000_without_null_characters-10        	    1735	    727387 ns/op	 8011791 B/op	       2 allocs/op
BenchmarkSanitizeJson/SanitizeUsingRegexp_string_of_size_4000000_without_null_characters-10       	     864	   1351200 ns/op	20036100 B/op	      14 allocs/op
```

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/Sanitize-jsonb-values-before-inserting-them-to-the-database-545a004e96eb4ae988e2b9da3fc5e807)

## Security


jobsdb/jobsdb_test.go
```diff
@@ -303,3 +312,94 @@ var dsListInMemory = []dataSetT{
...
+		input = input[:oidx] + c + input[oidx:]
+	}
+	return input
+}
+
+func sanitizedJsonUsingStrings(input json.RawMessage) string {
+	return strings.Replace(string(input), `\u0000`, "", -1)
+}
+
+func sanitizedJsonUsingBytes(input json.RawMessage) string {
```
lvrach, line 397: I believe we are being "unfair" to bytes benchmark here.

In your tests, you convert it back to a string. However in the actual implementation, you return `json.RawMessage`, which is a []byte.

Not very important, I just expected bytes.ReplaceAll to be better and then check the benchs :)

jobsdb/jobsdb.go
```diff
@@ -2205,36 +2215,59 @@ func (*HandleT) copyJobsDSInTx(txHandler transactionHandler, ds dataSetT, jobLis
...
+	const (
+		savepointSql = "SAVEPOINT doStoreJobsInTx"
+		rollbackSql  = "ROLLBACK TO " + savepointSql
+	)
+	if _, err := txHandler.Exec(savepointSql); err != nil {
+		return err
+	}
+	err := store()
+	var e *pq.Error
+	if err != nil && errors.As(err, &e) {
```
lvrach, line 2250: I would expect the following to work:
```go
if err != nil && errors.As(err, pq.Error)
```

hasn't that the case ?

lvrach (approved):

jobsdb/jobsdb_test.go
atzoum: Yes you are correct, but even after being fair, `strings` are still winning when there are no null characters

```
BenchmarkSanitizeJson/SanitizeUsingStrings_string_of_size_4000000_with_null_characters-10         	    1186	    972290 ns/op	12017684 B/op	       3 allocs/op
BenchmarkSanitizeJson/SanitizeUsingBytes_string_of_size_4000000_with_null_characters-10           	    1834	    643793 ns/op	 8011793 B/op	       2 allocs/op
BenchmarkSanitizeJson/SanitizeUsingRegexp_string_of_size_4000000_with_null_characters-10          	     544	   2163382 ns/op	31734351 B/op	      51 allocs/op
BenchmarkSanitizeJson/SanitizeUsingStrings_string_of_size_4000000_without_null_characters-10      	    2082	    571437 ns/op	 8011797 B/op	       2 allocs/op
BenchmarkSanitizeJson/SanitizeUsingBytes_string_of_size_4000000_without_null_characters-10        	    1610	    731277 ns/op	 8011789 B/op	       2 allocs/op
BenchmarkSanitizeJson/SanitizeUsingRegexp_string_of_size_4000000_without_null_characters-10       	    1012	   1190263 ns/op	16029709 B/op	      12 allocs/op
```

jobsdb/jobsdb.go
atzoum: it is according to the [docs](https://pkg.go.dev/errors)

> As unwraps its first argument sequentially looking for an error that can be assigned to its second argument, which must be a pointer. If it succeeds, it performs the assignment and returns true. Otherwise, it returns false. The form

```
var perr *fs.PathError
if errors.As(err, &perr) {
	fmt.Println(perr.Path)
}
```

chandumlg (approved):

PR #1964, opened 2022-05-18
chore(warehouse): fixed load parallel table configuration nor working

achettyiitr:
# Description

Fix parallel load table configuration not getting populated properly because we were trying to populate from the config even before config is getting initialized

## Notion Ticket

https://www.notion.so/rudderstacks/Bug-fix-for-warehouse-not-picking-parallel-table-configuration-6475a47f000e4bb8aa23ef4f661d65f4

## Security


warehouse/warehouse.go
```diff
@@ -139,6 +139,8 @@ type ErrorResponseT struct {
 func Init4() {
 	loadConfig()
 	pkgLogger = logger.NewLogger().Child("warehouse")
+
+	setMaxParallelLoads()
```
sivashanmukh, line 143: Why moving it to the end of the inits, why not at the start of Init1. What if some of the inits require these configs.

satishrudderstack (approved):
LGTM

Jayachand (approved):

PR #1965, opened 2022-05-19
feat(warehouse): adding date single partition support for glue by def…

Jayachand:
…ault

# Description

single partition requirement for glue

## Notion Ticket

https://www.notion.so/rudderstacks/Dice-S3-Datalake-Glue-Partitioning-d597c4d24e6e49a2a8807d366b300c9f#70d98afbe1f4454b9bc14247a8fc7c54

## Security


warehouse/azure-synapse/azure-synapse.go
```diff
@@ -854,3 +854,7 @@ func (as *HandleT) LoadTestTable(location string, tableName string, payloadMap m
 func (as *HandleT) SetConnectionTimeout(timeout time.Duration) {
 	as.ConnectTimeout = timeout
 }
+
+func (*HandleT) RefreshPartitions(_ string, _ []warehouseutils.LoadFileT) (err error) {
```
achettyiitr, line 858: It's not related to the corresponding changes. It's something in General. 
I don't think we should expose this to the WareHouse Manager since on DataLake is only implementing this and others are not implementing this. 
I guess, we can opt for some other design pattern here for these kind of changes going forward.

sivashanmukh: @achettyiitr That's a great comment. It makes sense for only datalakes to implement this. A separate interface for this would make more sense.

PR #1982, opened 2022-05-24
fix: reporting the drained events to the reports table

raman2805:
# Description

The events which got drained by the router were not being reported to the *reports* table inside the
jobsdb database. This PR fixes that 

## Notion Ticket

https://www.notion.so/rudderstacks/Report-drained-jobs-to-reporting-service-c8ce3d048e2945a797be3e31f8542ac2
## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -552,6 +552,7 @@ var (
 	//Valid, Terminal
 	Succeeded   = jobStateT{isValid: true, isTerminal: true, State: "succeeded"}
 	Aborted     = jobStateT{isValid: true, isTerminal: true, State: "aborted"}
+	Drained     = jobStateT{isValid: true, isTerminal: true, State: "drained"}
 	Migrated    = jobStateT{isValid: true, isTerminal: true, State: "migrated"}
 	WontMigrate = jobStateT{isValid: true, isTerminal: true, State: "wont_migrate"}
 )
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -552,7 +552,6 @@ var (
 	//Valid, Terminal
 	Succeeded   = jobStateT{isValid: true, isTerminal: true, State: "succeeded"}
 	Aborted     = jobStateT{isValid: true, isTerminal: true, State: "aborted"}
-	Drained     = jobStateT{isValid: true, isTerminal: true, State: "drained"}
 	Migrated    = jobStateT{isValid: true, isTerminal: true, State: "migrated"}
 	WontMigrate = jobStateT{isValid: true, isTerminal: true, State: "wont_migrate"}
 )
--- a/router/router.go
+++ b/router/router.go
@@ -583,63 +583,7 @@ func (worker *workerT) workerProcess() {
 }
 
 func (worker *workerT) processDestinationJobs() {
-	// TODO - this is the only place handleWorkerDestinationJobs is being called
-	// better to move the code here than having another function call
-	worker.handleWorkerDestinationJobs(context.TODO())
-	//routerJobs/destinationJobs are processed. Clearing the queues.
-	worker.routerJobs = make([]types.RouterJobT, 0)
-	worker.destinationJobs = make([]types.DestinationJobT, 0)
-	worker.jobCountsByDestAndUser = make(map[string]*destJobCountsT)
-}
-
-func (worker *workerT) canSendJobToDestination(prevRespStatusCode int, failedUserIDsMap map[string]struct{}, destinationJob types.DestinationJobT) bool {
-	if prevRespStatusCode == 0 {
-		return true
-	}
-
-	if !worker.rt.guaranteeUserEventOrder {
-		//if guaranteeUserEventOrder is false, letting the next jobs pass
-		return true
-	}
-
-	//If batching is enabled, we send the request only if the previous one succeeds
-	if worker.rt.enableBatching {
-		return isSuccessStatus(prevRespStatusCode)
-	}
-
-	//If the destinationJob has come through router transform,
-	//drop the request if it is of a failed user, else send
-	for _, metadata := range destinationJob.JobMetadataArray {
-		if _, ok := failedUserIDsMap[metadata.UserID]; ok {
-			return false
-		}
-	}
-
-	return true
-}
-
-func getIterableStruct(payload []byte, transformAt string) ([]integrations.PostParametersT, error) {
-	var err error
-	var response integrations.PostParametersT
-	responseArray := make([]integrations.PostParametersT, 0)
-	if transformAt == "router" {
-		err = json.Unmarshal(payload, &response)
-		if err != nil {
-			err = json.Unmarshal(payload, &responseArray)
-		} else {
-			responseArray = append(responseArray, response)
-		}
-	} else {
-		err = json.Unmarshal(payload, &response)
-		if err == nil {
-			responseArray = append(responseArray, response)
-		}
-	}
-
-	return responseArray, err
-}
-
-func (worker *workerT) handleWorkerDestinationJobs(ctx context.Context) {
+	ctx := context.TODO()
 	worker.batchTimeStat.Start()
 
 	var respContentType string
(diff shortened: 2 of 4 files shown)


router/router.go
```diff
@@ -1627,7 +1641,15 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
 		}
 		//Update the status
 		err := rt.jobsDB.WithUpdateSafeTx(func(tx jobsdb.UpdateSafeTx) error {
-			err := rt.jobsDB.UpdateJobStatusInTx(tx, statusList, []string{rt.destName}, nil)
+			// The status of drained jobs (inside the rt_job_status* table) is
+			// updated inside the readAndProcess function
+			var nonDrainedStatusList []*jobsdb.JobStatusT
```
atzoum, line 1646: Is this kind of filtering necessary? Under which conditions do we end up calling `commitStatusList` for drained job statuses?
As far as I can tell, we do not push drained job statuses in the `responseQ` channel.

_**Update:**_ just went a bit further and understood that we are explicitly sending now drained job statuses to the `responseQ` channel, not sure this is the best design choice though.

jobsdb/jobsdb.go
```diff
@@ -552,6 +552,7 @@ var (
 	//Valid, Terminal
 	Succeeded   = jobStateT{isValid: true, isTerminal: true, State: "succeeded"}
 	Aborted     = jobStateT{isValid: true, isTerminal: true, State: "aborted"}
+	Drained     = jobStateT{isValid: true, isTerminal: true, State: "drained"}
```
atzoum, line 555: We should refrain from introducing new states unless:
1. We have a very good reason
2. We don't use it ad hoc, but make sure we adapt all affected areas 

Please explain the reasoning behind introducing this new state

router/router.go
```diff
@@ -583,6 +583,8 @@ func (worker *workerT) workerProcess() {
 }
 
 func (worker *workerT) processDestinationJobs() {
+	// TODO - this is the only place handleWorkerDestinationJobs is being called
```
atzoum, line 586: Either perform such a refactoring, if necessary, or be silent and remove the TODO 😄

router/router.go
```diff
@@ -1993,6 +2020,7 @@ func (rt *HandleT) readAndProcess() int {
 
 			rt.timeGained += latenciesUsed[job.WorkspaceId]
 
+			rt.responseQ <- jobResponseT{status: &status, worker: nil, userID: job.UserID, JobT: job}
```
atzoum, line 2023: why don't we simply update the reporting stats here instead of passing drained jobs to the `responseQ` channel and having to deal with all the associated (extra) complexities?

I don't believe that at this point we care too much about some extra reporting code duplication, as long as we keep the rest of the design cleaner. We should be revisiting the whole reporting design sometime soon anyway!

atzoum (changes requested):

jobsdb/jobsdb.go
raman2805: I would agree I didn't have a very strong reason to do that. My only motivation was to be able to identify the jobs which were aborted due to such drainage by a different marker in the reports table. It also made it easy for me to filter out such jobs inside the commitStatusList. I will defer to your judgement on this. In this implementation, we can use other pieces of information such as the reason to filter out the jobs

router/router.go
raman2805: I am going to add another commit to inline this code

router/router.go
raman2805: My initial implementation did involve copying over the relevant parts of the code from the commitStatusList function to manually commit the information to the report. However, the code duplication ended up being more that I felt comfortable with. I will push another commit with that implementation for you guys to review and take a call between the 2 implementation.

router/router.go
raman2805: Update: In the latest commits, I have changed the implementation to directly update the report inside the readAndProcess function

jobsdb/jobsdb.go
raman2805: The new state which was added has been removed

router/router.go
```diff
@@ -1956,6 +1954,7 @@ func (rt *HandleT) readAndProcess() int {
 
 	rt.throttledUserMap = make(map[string]struct{})
 	throttledAtTime := time.Now()
+	reportMetrics := make([]*utilTypes.PUReportedMetric, 0)
```
atzoum, line 1957: it would be better to limit the scope of this variable inside the `if drain {}` branch

router/router.go
```diff
@@ -1994,6 +1993,31 @@ func (rt *HandleT) readAndProcess() int {
 			rt.timeGained += latenciesUsed[job.WorkspaceId]
 
 			rt.MultitenantI.CalculateSuccessFailureCounts(job.WorkspaceId, rt.destName, false, true)
+			/*
```
atzoum, line 1996: I would replace this comment with what appears to be the standard comment markers for reporting, e.g.

```
//REPORTING - ROUTER - START

//REPORTING - ROUTER - END
```

router/router.go
```diff
@@ -1994,6 +1993,31 @@ func (rt *HandleT) readAndProcess() int {
...
+
+			eventName := gjson.GetBytes(job.Parameters, "event_name").String()
+			eventType := gjson.GetBytes(job.Parameters, "event_type").String()
+
+			reportMetric := &utilTypes.PUReportedMetric{
+				ConnectionDetails: *utilTypes.CreateConnectionDetail(parameters.SourceID, parameters.DestinationID, parameters.SourceBatchID, parameters.SourceTaskID, parameters.SourceTaskRunID, parameters.SourceJobID, parameters.SourceJobRunID, parameters.SourceDefinitionID, parameters.DestinationDefinitionID, parameters.SourceCategory),
+				PUDetails:         *utilTypes.CreatePUDetails(inPu, utilTypes.ROUTER, true, false),
+				StatusDetail:      utilTypes.CreateStatusDetail(status.JobState, 1, 200, string(status.ErrorResponse), job.EventPayload, eventName, eventType),
+			}
+			reportMetrics = append(reportMetrics, reportMetric)
```
atzoum, line 2020: 1. Instead of generating a single statistic for every drained job, I believe we should group statistics using the same composite key that we are using elsewhere else in the router
2. We need to redact the sample event payload if the sourceID belongs to a transient source.

router/router.go
```diff
@@ -1994,6 +1993,31 @@ func (rt *HandleT) readAndProcess() int {
...
+				inPu = utilTypes.EVENT_FILTER
+			}
+
+			eventName := gjson.GetBytes(job.Parameters, "event_name").String()
+			eventType := gjson.GetBytes(job.Parameters, "event_type").String()
+
+			reportMetric := &utilTypes.PUReportedMetric{
+				ConnectionDetails: *utilTypes.CreateConnectionDetail(parameters.SourceID, parameters.DestinationID, parameters.SourceBatchID, parameters.SourceTaskID, parameters.SourceTaskRunID, parameters.SourceJobID, parameters.SourceJobRunID, parameters.SourceDefinitionID, parameters.DestinationDefinitionID, parameters.SourceCategory),
+				PUDetails:         *utilTypes.CreatePUDetails(inPu, utilTypes.ROUTER, true, false),
+				StatusDetail:      utilTypes.CreateStatusDetail(status.JobState, 1, 200, string(status.ErrorResponse), job.EventPayload, eventName, eventType),
```
atzoum, line 2018: does it make sense to use a 200 HTTP status code for drained events? Using for instance `410 Gone` would be more appropriate.

router/router.go
raman2805: reportMetrics aggregates all the metric for all jobs and it has a scope outside not only the if condition, but the surrounding **for** loop

router/router.go
raman2805: Done

router/router.go
raman2805: Will do

chandumlg (approved):

router/utils/utils.go
```diff
@@ -17,6 +17,10 @@ var (
 	JobRetention time.Duration
 )
 
+const (
+	DRAIN_ERROR_CODE int = 401
```
atzoum, line 21: any specific reason why we chose HTTP 401 - Unauthorized for drained jobs?

atzoum:
@chandumlg @lvrach  shouldn't we simply delete `rudder-transformer` files and add it in `.gitignore`?

router/router.go
```diff
@@ -1994,6 +1995,36 @@ func (rt *HandleT) readAndProcess() int {
...
+				cd := utilTypes.CreateConnectionDetail(parameters.SourceID, parameters.DestinationID, parameters.SourceBatchID, parameters.SourceTaskID, parameters.SourceTaskRunID, parameters.SourceJobID, parameters.SourceJobRunID, parameters.SourceDefinitionID, parameters.DestinationDefinitionID, parameters.SourceCategory)
+				connectionDetailsMap[key] = cd
+				transformedAtMap[key] = parameters.TransformAt
+			}
+
+			sd, ok := statusDetailsMap[key]
+			if !ok {
+				sampleEvent := job.EventPayload
+				if rt.transientSources.Apply(parameters.SourceID) {
+					sampleEvent = []byte(`{}`)
```
cisse21, line 2021: Probably have the empty payload as a global variable?

router/router.go
```diff
@@ -2028,7 +2059,38 @@ func (rt *HandleT) readAndProcess() int {
...
+		//REPORTING - ROUTER - END
+
+		err = rt.jobsDB.WithUpdateSafeTx(func(tx jobsdb.UpdateSafeTx) error {
+			err := rt.jobsDB.UpdateJobStatusInTx(tx, drainList, []string{rt.destName}, nil)
+			if err != nil {
+				rt.logger.Errorf("[Router] :: Error occurred while updating %s jobs statuses. Panicking. Err: %v", rt.destName, err)
+				return err
+			}
+			rt.Reporting.Report(reportMetrics, tx.Tx())
+			return nil
```
cisse21, line 2091: same here

cisse21 (changes requested):

router/router.go
```diff
@@ -1994,6 +1995,36 @@ func (rt *HandleT) readAndProcess() int {
 			rt.timeGained += latenciesUsed[job.WorkspaceId]
 
 			rt.MultitenantI.CalculateSuccessFailureCounts(job.WorkspaceId, rt.destName, false, true)
+
+			// REPORTING - ROUTER - START
+			var parameters JobParametersT
+			err := json.Unmarshal(job.Parameters, &parameters)
+			if err != nil {
+				rt.logger.Error("Unmarshal of job parameters failed. ", string(job.Parameters))
```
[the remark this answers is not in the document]
raman2805: makes sense, will do

router/router.go
```diff
@@ -2028,7 +2059,38 @@ func (rt *HandleT) readAndProcess() int {
...
+				reportMetrics = append(reportMetrics, m)
+			}
+		}
+		//REPORTING - ROUTER - END
+
+		err = rt.jobsDB.WithUpdateSafeTx(func(tx jobsdb.UpdateSafeTx) error {
+			err := rt.jobsDB.UpdateJobStatusInTx(tx, drainList, []string{rt.destName}, nil)
+			if err != nil {
+				rt.logger.Errorf("[Router] :: Error occurred while updating %s jobs statuses. Panicking. Err: %v", rt.destName, err)
+				return err
```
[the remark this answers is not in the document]
raman2805: we are not returning out of the outermost function here, but the inner lambda function which is passed to rt.jobsDB.WithUpdateSafeTx as the parameter

router/router.go
raman2805: ditto

router/utils/utils.go
raman2805: No, I think I made a mistake, instead of using 410 that you suggested earlier, I ended up using 401. Let me change it to 410.

cisse21 (approved):

chandumlg:
> @chandumlg @lvrach shouldn't we simply delete `rudder-transformer` files and add it in `.gitignore`?

rudder-transformer is added as a submodule. We might have to revisit our documentation if we do that.

atzoum:
> rudder-transformer is added as a submodule. We might have to revisit our documentation if we do that.

We can still keep it as a submodule, simply add the submodule's folder in .gitignore so that we don't see changes in submodule files on the PR diffs.
E.g.
https://github.com/rudderlabs/rudder-server/pull/2025

atzoum (approved):

PR #1983, opened 2022-05-25
feat(warehouse):add table suffix in load file path in gcs_datalake

Jayachand:
# Description

> feat(warehouse):add table suffix in load file path in gcs_datalake
> giving this build to beacon to get more feedback

## Notion Ticket

https://www.notion.so/rudderstacks/Beacon-custom-file-format-layout-requirement-for-gcs_datalake-db9da732b69a4562923ec12a11952658

## Security

### DIFF
--- a/warehouse/utils/utils.go
+++ b/warehouse/utils/utils.go
@@ -909,3 +909,18 @@ func GetLoadFileFormat(whType string) string {
 		return "csv.gz"
 	}
 }
+
+func GetLoadFilePrefix(timeWindow time.Time, warehouse WarehouseT) (timeWindowFormat string) {
+	whType := warehouse.Type
+	switch whType {
+	case GCS_DATALAKE:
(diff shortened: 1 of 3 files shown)


warehouse/utils/utils.go
```diff
@@ -909,3 +909,18 @@ func GetLoadFileFormat(whType string) string {
 		return "csv.gz"
 	}
 }
+
```
achettyiitr, line 912: Not priority since we need to release it urgently. But it's a good contender for Unit tests. Can we add a TODO to create Unit tests for this.
wdyt?

sivashanmukh: @Jayachand Shouldn't be too hard to write test cases. Unless you think otherwise, please add unit tests as well for this func

sivashanmukh:
Closing in favour of https://github.com/rudderlabs/rudder-server/pull/2015/files

PR #1986, opened 2022-05-26
chore(jobsdb): fix migration sequence number for multitenant migration

cisse21:
# Description

With the new multi-tenant migrator approach we will have tables created with index 0 which might undergo internal migrations but the current new calculation assumes that we cannot have _0_1 indices and can have only _0_0_1 indices. This PR aims to generate the _0_1 indices behind a flag 

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/Don-t-create-0_0_1-index-for-tables-fcc14c3da3cf47c6a0e9da79dde99b51)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -1377,6 +1377,9 @@ func computeIdxForClusterMigration(tablePrefix string, dList []dataSetT, insertB
 //Tries to give a slice between before and after by incrementing last value in before. If the order doesn't maintain, it adds a level and recurses.
 func computeInsertVals(before, after []string) ([]string, error) {
 	for {
+		if before == nil || after == nil {
+			return nil, fmt.Errorf("before or after is nil")
+		}
 		//Safe check: In the current jobsdb implementation, indices don't go more
 		//than 3 levels deep. Breaking out of the loop if the before is of size more than 4.
 		if len(before) > 4 {
@@ -1391,7 +1394,6 @@ func computeInsertVals(before, after []string) ([]string, error) {
 		}
 		//Just increment the last value of the index as a possible candidate
 		calculatedVals[len(calculatedVals)-1] = fmt.Sprintf("%d", lastVal+1)
-
 		var equals bool
 		if len(calculatedVals) == len(after) {
 			equals = true
@@ -1402,7 +1404,6 @@ func computeInsertVals(before, after []string) ([]string, error) {
 				equals = false
 			}
 		}
-
 		if !equals {
 			comparison, err := dsComparitor(calculatedVals, after)
 			if err != nil {
@@ -1412,11 +1413,17 @@ func computeInsertVals(before, after []string) ([]string, error) {
 				return calculatedVals, fmt.Errorf("computed index is invalid. before: %v, after: %v, calculatedVals: %v", before, after, calculatedVals)
 			}
 		}
-
-		//Only when the index starts with 0, we allow three levels. This would be when we have to insert an internal migration DS between two import DSs
+		//Only when the index starts with 0, we allow three levels and when we are using the legacy migration
 		//In all other cases, we allow only two levels
-		if (before[0] == "0" && len(calculatedVals) == 3) ||
-			(before[0] != "0" && len(calculatedVals) == 2) {
+		var comparatorBool bool
+		if skipZeroAssertionForMultitenant {
+			comparatorBool = len(calculatedVals) == 2
+		} else {
+			comparatorBool = (before[0] == "0" && len(calculatedVals) == 3) ||
+				(before[0] != "0" && len(calculatedVals) == 2)
+		}
+
+		if comparatorBool {
 			if equals {
 				return calculatedVals, fmt.Errorf("calculatedVals and after are same. computed index is invalid. before: %v, after: %v, calculatedVals: %v", before, after, calculatedVals)
 			} else {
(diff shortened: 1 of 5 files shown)


jobsdb/jobsdb.go
```diff
@@ -1412,11 +1414,16 @@ func computeInsertVals(before, after []string) ([]string, error) {
 				return calculatedVals, fmt.Errorf("computed index is invalid. before: %v, after: %v, calculatedVals: %v", before, after, calculatedVals)
 			}
 		}
-
+		var comparatorBool bool
+		if skipZeroAssertionForMultitenant {
+			comparatorBool = len(calculatedVals) == 2
+		} else {
+			comparatorBool = (before[0] == "0" && len(calculatedVals) == 3) ||
+				(before[0] != "0" && len(calculatedVals) == 2)
```
fracasula, line 1422: I think this could lead to a `panic` given that up above we are only checking if `before` lenght is `> 4` but never if it has at least one element. I understand this is called by `computeInsertIdx` only at the moment but every function is a "unit" and at the moment it looks like this unit can break?

```go
func TestPanic(t *testing.T) {
	_, _ = computeInsertVals(nil, nil)
}
```

Gives:

```
=== RUN   TestPanic
--- FAIL: TestPanic (0.00s)
panic: runtime error: index out of range [-1] [recovered]
	panic: runtime error: index out of range [-1]

goroutine 85 [running]:
testing.tRunner.func1.2({0x11b92a0, 0xc000122fa8})
	/usr/local/go/src/testing/testing.go:1389 +0x24e
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1392 +0x39f
panic({0x11b92a0, 0xc000122fa8})
	/usr/local/go/src/runtime/panic.go:838 +0x207
github.com/rudderlabs/rudder-server/jobsdb.computeInsertVals({0x0?, 0x51e9b4?, 0x0?}, {0x0?, 0x0, 0x0})
	/home/francesco/Code/rudderstack/rudder-server/jobsdb/jobsdb.go:1392 +0x639
github.com/rudderlabs/rudder-server/jobsdb.TestPanic(0x0?)
	/home/francesco/Code/rudderstack/rudder-server/jobsdb/jobsdb_test.go:26 +0x29
testing.tRunner(0xc000503520, 0x12de2d0)
	/usr/local/go/src/testing/testing.go:1439 +0x102
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1486 +0x35f


Process finished with the exit code 1
```

It should instead return an error instead of killing the whole service. WDYT?

Also, it would be nice to see some fuzzy testing or if that's too much for now just test the same unit with bad data to make sure it doesn't break.

fracasula (changes requested):

jobsdb/jobsdb.go
```diff
@@ -1412,11 +1414,16 @@ func computeInsertVals(before, after []string) ([]string, error) {
...
+		}
 		//Only when the index starts with 0, we allow three levels. This would be when we have to insert an internal migration DS between two import DSs
```
atzoum, line 1424: this comment doesn't seem to apply to the `skipZeroAssertion` case

jobsdb/jobsdb.go
```diff
@@ -1412,11 +1414,16 @@ func computeInsertVals(before, after []string) ([]string, error) {
 				return calculatedVals, fmt.Errorf("computed index is invalid. before: %v, after: %v, calculatedVals: %v", before, after, calculatedVals)
 			}
 		}
-
+		var comparatorBool bool
+		if skipZeroAssertionForMultitenant {
+			comparatorBool = len(calculatedVals) == 2
```
atzoum, line 1419: why are we ommitting the `(before[0] == "0" && len(calculatedVals) == 3)` condition?

jobsdb/jobsdb.go
```diff
@@ -73,6 +73,10 @@ func IsMasterBackupEnabled() bool {
 	return masterBackupEnabled
 }
 
+func SetskipZeroAssertionForMultitenant(b bool) {
```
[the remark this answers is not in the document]
cisse21: Moved to `jobsdb_test.go` ... and did the same in `processor.go` code too

jobsdb/jobsdb.go
cisse21: Added a check to ensure neither of the inputs is a nil array. Probably should add more validations to ensure we are not panicking

jobsdb/jobsdb.go
cisse21: This would convert the `_0` tables to `_0_0_1` tables and other tables like `_1` would become `_1_1`. Just to be consistent irrespective of the index we can skip the length 3 condition for `multi-tenant` migrations and to ensure we are backwards compatible have the old flow too

jobsdb/jobsdb.go
cisse21: modified the comment

jobsdb/jobsdb.go
fracasula: That is great, thanks @cisse21 :+1: 
I think we just need to cover them now

Screenshot from 2022-05-26 13-56-55

cisse21: I will add tests for these cases too

fracasula (approved):

BonapartePC (approved):

PR #1987, opened 2022-05-26
chore: backup jobs pickup based on total pickup payload size, instead of `backupRowsBatchSize`

saurav-malani:
# Description
Currently, if backup is enabled, we backup based on `backupRowsBatchSize`, but if payload size is larger than usual, this sometimes leads to OOM kill. So, in this PR we have modified the pickup limit based on "total pickup payload size", to avoid OOM kill.

## Notion Ticket

https://www.notion.so/rudderstacks/Introduce-limit-based-on-payload-size-during-backup-e527d5b1138741e0a5c20471ef20c178


lvrach (commented):
Could you devise a test for the code that has been changed, similar to integration tests we have for get methods

atzoum (changes requested):

jobsdb/jobsdb.go
```diff
@@ -3119,14 +3122,159 @@ func (jd *HandleT) getBackUpQuery(backupDSRange *dataSetRangeT, isJobStatusTable
...
+						LIMIT 
+							%[7]d
+						OFFSET
+							%[8]d
+						) jobs
+					) subquery 
+				WHERE 
+					subquery.running_payload_size <= %[9]d
+				) AS failed_jobs
+		  `, backupDSRange.ds.JobStatusTable, "job_status", backupDSRange.ds.JobTable, "job", Failed.State, Aborted.State, backupRowsBatchSize, offset, backupMaxTotalPayloadSize)
```
atzoum, line 3228: since we are already using the `job` and `job_status` table aliases in the query, we shouldn't pass them as parameters

jobsdb/jobsdb.go
```diff
@@ -3234,13 +3382,14 @@ func (jd *HandleT) backupTable(ctx context.Context, backupDSRange *dataSetRangeT
...
-
 	var offset, batchCount int64
+	var jobCount int64
 	for {
 		stmt := jd.getBackUpQuery(backupDSRange, isJobStatusTable, offset)
 		var rawJSONRows json.RawMessage
 		row := jd.dbHandle.QueryRow(stmt)
-		err = row.Scan(&rawJSONRows)
+
+		err = row.Scan(&rawJSONRows, jobCount)
```
atzoum, line 3392:
```suggestion
		err = row.Scan(&rawJSONRows, &jobCount)
```

atzoum (changes requested):

jobsdb/jobsdb.go
```diff
@@ -3119,14 +3122,159 @@ func (jd *HandleT) getBackUpQuery(backupDSRange *dataSetRangeT, isJobStatusTable
 	var stmt string
 	if jd.BackupSettings.FailedOnly {
```
atzoum, line 3123: It is weird that, depending on this flag, the json that we export has a different set of fields...
@chandumlg @cisse21 @lvrach do we use these exports as input anywhere else?

atzoum:
@saurav-malani let's add some tests, otherwise LGTM

jobsdb/jobsdb.go
```diff
@@ -652,6 +654,7 @@ func loadConfig() {
 	config.RegisterIntConfigVariable(10, &maxMigrateDSProbe, true, 1, "JobsDB.maxMigrateDSProbe")
 	config.RegisterInt64ConfigVariable(300, &maxTableSize, true, 1000000, "JobsDB.maxTableSizeInMB")
 	config.RegisterInt64ConfigVariable(1000, &backupRowsBatchSize, true, 1, "JobsDB.backupRowsBatchSize")
+	config.RegisterInt64ConfigVariable(64*bytesize.MB, &backupMaxTotalPayloadSize, true, 1, "JobsDB.maxBackupTotalPayloadSize")
```
cisse21, line 657: I think we should let the user provide the Input in MB and use the multiplication factor(4th parameter) to convert the integer provided to MB

jobsdb/jobsdb.go
cisse21: I dont think we are exporting these fields anywhere else but for the other question where you say the JSON we export has different fields, Do you mean we store a different JSON schema in S3 based on the flag?

atzoum: yes

jobsdb/jobsdb.go
atzoum: I don't believe this is necessary, but if we do any change on this one we should do the same change on all other config properties that deal with payload size limits

cisse21: Was just thinking about it from a usability perspective as adding an input as bytes might be a bit painful but it's not something which is blocking

jobsdb/jobsdb.go
cisse21: I can't think of any reason why they should be different. We don't use these table backups even for replay, The backups which we use are `proc_error` which is not a part of this flow and `gw` which doesn't have any Failed jobs. So I feel we can use the same schema

atzoum: What are we using these backups for?

cisse21: I guess most likely they are used for debugging purposes where we need the status table and jobs table to understand what happened. Can't think of any other use case

atzoum: let's leave them as they are then, @saurav-malani I believe you can even skip the trick we are doing for duplicate columns

jobsdb/jobsdb_test.go
```diff
@@ -455,6 +482,300 @@ var dsListInMemory = []dataSetT{
...
+		Tag:        "latest",
+		Cmd:        []string{"server", "/data"},
+		Env: []string{
+			fmt.Sprintf("MINIO_ACCESS_KEY=%s", accessKeyId),
+			fmt.Sprintf("MINIO_SECRET_KEY=%s", secretAccessKey),
+			fmt.Sprintf("MINIO_SITE_REGION=%s", region),
+		},
+	})
+	if err != nil {
+		panic(fmt.Errorf("Could not start resource: %s", err))
```
atzoum, line 519: please replace panics with test failures

jobsdb/jobsdb_test.go
```diff
@@ -455,6 +482,300 @@ var dsListInMemory = []dataSetT{
...
+			return err
+		}
+		if resp.StatusCode != http.StatusOK {
+			return fmt.Errorf("status code not OK")
+		}
+		return nil
+	}); err != nil {
+		log.Fatalf("Could not connect to docker: %s", err)
+	}
+	fmt.Println("minio is up & running properly")
```
atzoum, line 543: please replace all occurences of `fmt.Println` with `t.Log` statements

atzoum:
`goldenDirectory` should be renamed to `testdata`

> Directory and file names that begin with "." or "_" are ignored by the go tool, as are directories named "testdata".

jobsdb/jobsdb_test.go
saurav-malani: done

jobsdb/jobsdb_test.go
saurav-malani: done

atzoum (approved):

jobsdb/jobsdb.go
```diff
@@ -3328,30 +3490,28 @@ func (jd *HandleT) backupTable(ctx context.Context, backupDSRange *dataSetRangeT
...
+		defer rows.Close()
 
-		if _, err := gzWriter.Write(rawJSONRows); err != nil {
-			return false, fmt.Errorf("writing to gz file %q: %w", path, err)
+		for rows.Next() {
+			err = rows.Scan(&rawJSONRows)
+			if err != nil {
+				panic(fmt.Errorf("Scanning row failed with error : %w", err))
+			}
+			rawJSONRows = append(rawJSONRows, '\n') // appending '\n'
```
lvrach, line 3507: `[very minor]` not sure about the benefit of this comment
```suggestion
			rawJSONRows = append(rawJSONRows, '\n') 
```

lvrach (approved):
Approve, with a minor comment regarding error handling.

jobsdb/jobsdb.go
```diff
@@ -3328,30 +3490,28 @@ func (jd *HandleT) backupTable(ctx context.Context, backupDSRange *dataSetRangeT
...
+		rows, err := jd.dbHandle.Query(stmt)
+		jd.assertError(err)
+		defer rows.Close()
 
-		if _, err := gzWriter.Write(rawJSONRows); err != nil {
-			return false, fmt.Errorf("writing to gz file %q: %w", path, err)
+		for rows.Next() {
+			err = rows.Scan(&rawJSONRows)
+			if err != nil {
+				panic(fmt.Errorf("Scanning row failed with error : %w", err))
```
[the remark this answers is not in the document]
atzoum: Yes indeed the method's signature supports it, but the caller of the method doesn't handle the error properly. We simply log the error and proceed by dropping the DS.
Is this what we want?

PR #1989, opened 2022-05-26
chore: appropriate response at Gateway to incoming request

saurav-malani:
# Description
List of response changes at Gateway:-
- previously we were not using `statusMap` which has list of all possible response code and body. And, irrespective of the issue type with the request we simply returned `400` for all, except `429` for too many request. In this PR, we modified this behaviour get appropriate response code and body by making use of `statusMap` in `response.go`.
- `SourceDisabled` response is introduced. modified response for "disabled sources" from `401` to `404`, by storing the config for those source in `map`.
- `InvalidWebhookSource` status code changed from `400` to `404`.

## Notion Ticket
https://www.notion.so/rudderstacks/Revisit-http-status-codes-returned-by-gateway-0cdda93d350b437589872237c7405af2

## Security

### DIFF
--- a/gateway/admin.go
+++ b/gateway/admin.go
@@ -19,8 +19,8 @@ var prefix = "gw_jobs_"
 func (g *GatewayAdmin) Status() interface{} {
 	configSubscriberLock.RLock()
 	defer configSubscriberLock.RUnlock()
-	writeKeys := make([]string, 0, len(enabledWriteKeysSourceMap))
-	for k := range enabledWriteKeysSourceMap {
+	writeKeys := make([]string, 0, len(writeKeysSourceMap))
+	for k := range writeKeysSourceMap {
 		writeKeys = append(writeKeys, k)
 	}
 
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -715,9 +715,7 @@ func (gateway *HandleT) eventSchemaWebHandler(wrappedFunc func(http.ResponseWrit
 }
 
 func (gateway *HandleT) getPayloadFromRequest(r *http.Request) ([]byte, error) {
-	// fmt.Println("---request body: ", r.Body)
 	if r.Body == nil {
-		fmt.Println("---r.Body is nil---")
 		return []byte{}, errors.New(response.RequestBodyNil)
 	}
 
@@ -1091,13 +1089,7 @@ func (gateway *HandleT) webRequestHandler(rh RequestHandler, w http.ResponseWrit
 	var errorMessage string
 	defer func() {
 		if errorMessage != "" {
-			// if strings.Contains(errorMessage, response.GetStatus(response.TooManyRequests)) {
-			// 	gateway.logger.Infof("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, http.StatusTooManyRequests, errorMessage)
-			// 	http.Error(w, errorMessage, response.GetStatusCode(errorMessage))
-			// 	return
-			// }
 			gateway.logger.Infof("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, response.GetStatusCode(errorMessage), errorMessage)
-			fmt.Println("---status code: ", response.GetStatusCode(errorMessage))
 			http.Error(w, errorMessage, response.GetStatusCode(errorMessage))
 		}
 	}()
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -81,7 +81,7 @@ var (
 	webPort, maxUserWebRequestWorkerProcess, maxDBWriterProcess, adminWebPort         int
 	maxUserWebRequestBatchSize, maxDBBatchSize, MaxHeaderBytes, maxConcurrentRequests int
 	userWebRequestBatchTimeout, dbBatchWriteTimeout                                   time.Duration
-	enabledWriteKeysSourceMap                                                         map[string]backendconfig.SourceT
+	writeKeysSourceMap                                                                map[string]backendconfig.SourceT
 	enabledWriteKeyWebhookMap                                                         map[string]string
 	enabledWriteKeyWorkspaceMap                                                       map[string]string
 	sourceIDToNameMap                                                                 map[string]string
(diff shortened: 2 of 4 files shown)


gateway/response/response.go
```diff
@@ -77,17 +79,18 @@ func loadStatusMap() {
...
 	statusMap[SourceTransformerFailed] = ResponseStatus{message: SourceTransformerFailed, code: http.StatusBadRequest}
 	statusMap[SourceTransformerResponseErrorReadFailed] = ResponseStatus{message: SourceTransformerResponseErrorReadFailed, code: http.StatusBadRequest}
 	statusMap[SourceTransformerFailedToReadOutput] = ResponseStatus{message: SourceTransformerFailedToReadOutput, code: http.StatusBadRequest}
 	statusMap[SourceTransformerInvalidResponseFormat] = ResponseStatus{message: SourceTransformerInvalidResponseFormat, code: http.StatusBadRequest}
 	statusMap[SourceTransformerInvalidOutputFormatInResponse] = ResponseStatus{message: SourceTransformerInvalidOutputFormatInResponse, code: http.StatusBadRequest}
 	statusMap[SourceTransformerInvalidOutputJSON] = ResponseStatus{message: SourceTransformerInvalidOutputJSON, code: http.StatusBadRequest}
 	statusMap[NonIdentifiableRequest] = ResponseStatus{message: NonIdentifiableRequest, code: http.StatusBadRequest}
-	statusMap[ErrorInMarshal] = ResponseStatus{message: ErrorInMarshal, code: http.StatusBadRequest}
+	statusMap[ErrorInMarshal] = ResponseStatus{message: ErrorInMarshal, code: http.StatusInternalServerError}
```
cisse21, line 93: Wouldn't this lead to retries from SDK which would eventually fail anyways with the same error

gateway/gateway.go
```diff
@@ -1073,13 +1091,14 @@ func (gateway *HandleT) webRequestHandler(rh RequestHandler, w http.ResponseWrit
...
 	defer func() {
 		if errorMessage != "" {
-			if strings.Contains(errorMessage, response.GetStatus(response.TooManyRequests)) {
-				gateway.logger.Infof("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, http.StatusTooManyRequests, errorMessage)
-				http.Error(w, errorMessage, http.StatusTooManyRequests)
-				return
-			}
-			gateway.logger.Infof("IP: %s -- %s -- Response: 400, %s", misc.GetIPFromReq(r), r.URL.Path, errorMessage)
-			http.Error(w, errorMessage, 400)
+			// if strings.Contains(errorMessage, response.GetStatus(response.TooManyRequests)) {
```
cisse21, line 1094: Remove the commented lines

gateway/gateway.go
```diff
@@ -1073,13 +1091,14 @@ func (gateway *HandleT) webRequestHandler(rh RequestHandler, w http.ResponseWrit
...
+			// 	gateway.logger.Infof("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, http.StatusTooManyRequests, errorMessage)
+			// 	http.Error(w, errorMessage, response.GetStatusCode(errorMessage))
+			// 	return
+			// }
+			gateway.logger.Infof("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, response.GetStatusCode(errorMessage), errorMessage)
+			fmt.Println("---status code: ", response.GetStatusCode(errorMessage))
```
cisse21, line 1100: remove the println statement, make it a log if needed

gateway/gateway.go
```diff
@@ -699,7 +715,9 @@ func (gateway *HandleT) eventSchemaWebHandler(wrappedFunc func(http.ResponseWrit
 }
 
 func (gateway *HandleT) getPayloadFromRequest(r *http.Request) ([]byte, error) {
+	// fmt.Println("---request body: ", r.Body)
```
cisse21, line 718: remove the commented line

gateway/gateway.go
```diff
@@ -699,7 +715,9 @@ func (gateway *HandleT) eventSchemaWebHandler(wrappedFunc func(http.ResponseWrit
...
 	if r.Body == nil {
+		fmt.Println("---r.Body is nil---")
```
cisse21, line 720: remove the println statement, make it a log if needed

cisse21 (changes requested):

gateway/response/response.go
lvrach: @saurav-malani what is your motivation for changing this? Isn't the client responsible for such errors?

saurav-malani: the reason I changed it from `400` to `500` was, in general we marshal the result we want to respond with. So, based on this assumption I made with `500` thinking if all the checks passed when we got the request and we encountered error when we tried marshalling the `resp`. But, looking at the place where it is being used (webhook), I think what you are saying makes sense.

saurav-malani: reverted it back.

cisse21:
Remove `ErrorInMarshal status code changed from 400 to 500.` from PR description too

lvrach (approved):

gateway/admin.go
```diff
@@ -19,8 +19,8 @@ var prefix = "gw_jobs_"
 func (g *GatewayAdmin) Status() interface{} {
 	configSubscriberLock.RLock()
 	defer configSubscriberLock.RUnlock()
-	writeKeys := make([]string, 0, len(enabledWriteKeysSourceMap))
-	for k := range enabledWriteKeysSourceMap {
+	writeKeys := make([]string, 0, len(writeKeysSourceMap))
+	for k := range writeKeysSourceMap {
 		writeKeys = append(writeKeys, k)
```
atzoum, line 24: Since the field's name is `enabled-write-keys`, we should only include enabled write keys.
If we are interested in including in the status response disabled key too, then I suppose we should include an additional `disabled-write-keys` field.

gateway/response/response.go
```diff
@@ -77,9 +79,10 @@ func loadStatusMap() {
 	statusMap[RequestBodyReadFailed] = ResponseStatus{message: RequestBodyReadFailed, code: http.StatusBadRequest}
 	statusMap[RequestBodyTooLarge] = ResponseStatus{message: RequestBodyTooLarge, code: http.StatusRequestEntityTooLarge}
 	statusMap[InvalidWriteKey] = ResponseStatus{message: InvalidWriteKey, code: http.StatusUnauthorized}
+	statusMap[SourceDisabled] = ResponseStatus{message: SourceDisabled, code: http.StatusNotFound}
 	statusMap[InvalidJSON] = ResponseStatus{message: InvalidJSON, code: http.StatusBadRequest}
 	// webhook specific status
-	statusMap[InvalidWebhookSource] = ResponseStatus{message: InvalidWebhookSource, code: http.StatusBadRequest}
+	statusMap[InvalidWebhookSource] = ResponseStatus{message: InvalidWebhookSource, code: http.StatusNotFound}
```
atzoum, line 85: I cannot find where we are using this error after all

gateway/response/response.go
```diff
@@ -77,9 +79,10 @@ func loadStatusMap() {
 	statusMap[RequestBodyReadFailed] = ResponseStatus{message: RequestBodyReadFailed, code: http.StatusBadRequest}
 	statusMap[RequestBodyTooLarge] = ResponseStatus{message: RequestBodyTooLarge, code: http.StatusRequestEntityTooLarge}
 	statusMap[InvalidWriteKey] = ResponseStatus{message: InvalidWriteKey, code: http.StatusUnauthorized}
+	statusMap[SourceDisabled] = ResponseStatus{message: SourceDisabled, code: http.StatusNotFound}
```
atzoum, line 82: I would argue that since the source exists but it is disabled, we should use `403 - Forbidden` instead of `404 - Not Found`

gateway/response/response.go
saurav-malani: yep as of now, I this is not being used anywhere.

gateway/admin.go
saurav-malani: but, that is the reason why I have renamed it to `writeKeysSourceMap` from `enableWriteKeysSourceMap`. Since, now it will have all the write keys (enabled & disabled both).

PR #1990, opened 2022-05-27
chore: identify transient sources through a top-level configuration field

atzoum:
# Description

Instead of relying to the source's generic config, use a static, top-level field for identifying transient sources.

## Notion Ticket

[Identify transient sources through a top-level configuration field](https://www.notion.so/rudderstacks/Identify-transient-sources-through-a-top-level-configuration-field-2d03022b25564dbaac90421ec2f21afd)

## Security


mcchran (approved):
LGTM.

PR #1994, opened 2022-05-27
refactor(jobsdb): support lazy transaction initiation

atzoum:
# Description

Instead of starting a transaction eagerly, the default behaviour now defers it creation until the transaction is actually requested by the receiver funtion.
This way, we no longer have outstanding transactions waiting their turn to come in the read & write queues of jobsDB.

## Notion Ticket

[Support lazy transaction initiation in jobsdb commands](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=8085e64339274e72b975479c6f35f0b6)

## Security

### DIFF
--- a/jobsdb/tx.go
+++ b/jobsdb/tx.go
@@ -1,0 +1,106 @@
+package jobsdb
+
+import (
+	"database/sql"
+	"errors"
+	"sync"
+)
+
+// ErrEmptyxTx sentinel error indicating an Empty Tx
+var ErrEmptyxTx = errors.New("jobsdb: empty tx")
+
+// TxGetter provides access to an sql.Tx
+type TxGetter interface {
+	// Tx retrieves the tx or returns an error if the tx could not be retrieved (started)
+	Tx() (*sql.Tx, error)
+	// MustTx retrieves the tx or panics in case of an error
+	MustTx() *sql.Tx
+}
+
+type noTx struct{}
+
+// EmptyTx returns an empty interface usable only for tests
+func EmptyTx() TxGetter {
+	return &noTx{}
+}
+
+func (*noTx) Tx() (*sql.Tx, error) {
+	return nil, ErrEmptyxTx
+}
+
+func (*noTx) MustTx() *sql.Tx {
+	return nil
+}
+
+type eagerTx struct {
+	tx *sql.Tx
+}
+
+func (r *eagerTx) Tx() (*sql.Tx, error) {
+	return r.tx, nil
+}
+
+func (r *eagerTx) MustTx() *sql.Tx {
+	return r.tx
+}
+
+type lazyTx struct {
+	dbHandle *sql.DB
+	once     sync.Once
+	tx       *sql.Tx
+	err      error
+}
+
+func (r *lazyTx) Tx() (*sql.Tx, error) {
+	r.once.Do(func() {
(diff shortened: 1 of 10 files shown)


lvrach:
I am not sure if the solution is to have lazy transactions or refactor the code in such a way that only transaction-related stuff is inside a transaction block. Even if we do it lazily there is still code between SQL statements taking valuable time.

atzoum:
> I am not sure if the solution is to have lazy transactions or refactor the code in such a way that only transaction-related stuff is inside a transaction block. Even if we do it lazily there is still code between SQL statements taking valuable time.

During a transaction multiple db operations can performed and it is normal to have other calculations happening in between them.
This change simply defers the start of the transaction until it is actually requested.
It is particularly valuable for our current usage patterns, where the 1st operation of the transaction is always writing something to the jobsdb through the writer queue. It also helps with current code reuse patterns e.g. `Store` is reusing `WithStoreSafeTx` & `StoreInTx` internally.

IMO, there is not much we can control beyond that at this stage.

jobsdb/tx.go
```diff
@@ -0,0 +1,106 @@
+package jobsdb
+
+import (
+	"database/sql"
+	"errors"
+	"sync"
+)
+
+// ErrEmptyxTx sentinel error indicating an Empty Tx
+var ErrEmptyxTx = errors.New("jobsdb: empty tx")
```
fracasula, line 10:
```suggestion
var ErrEmptyTx = errors.New("jobsdb: empty tx")
```

jobsdb/tx.go
```diff
@@ -0,0 +1,106 @@
...
+
+// TxGetter provides access to an sql.Tx
+type TxGetter interface {
+	// Tx retrieves the tx or returns an error if the tx could not be retrieved (started)
+	Tx() (*sql.Tx, error)
+	// MustTx retrieves the tx or panics in case of an error
+	MustTx() *sql.Tx
```
fracasula, line 17: qq: do you know when `Tx()` would actually return an error?

jobsdb/tx.go
```diff
@@ -0,0 +1,106 @@
...
+func (r *lazyTx) MustTx() *sql.Tx {
+	tx, err := r.Tx()
+	if err != nil {
+		panic(err)
+	}
+	return tx
+}
+
+// StoreSafeTx sealed interface
+type StoreSafeTx interface {
```
fracasula, line 70: It is not clear to me what a "store safe" transaction is. Can you explain?
If I look at the (huge) `JobsDB` interface it says:

```go
	// WithStoreSafeTx prepares a store-safe environment and then starts a transaction
	// that can be used by the provided function.
```
What is a store-safe environment?

atzoum: it makes sure all necessary locks have been acquired

jobsdb/tx.go
atzoum: if a database driver error is returned

atzoum:
@lvrach @fracasula @chandumlg you can find an alternative approach in #2209 
Let me know if you prefer it in order to proceed with adding unit/integration tests.

atzoum:
keep

PR #1996, opened 2022-05-28
feat(multi-tenant): use cached backend config

kuldeep0020:
# Use cached config backend to reduce the load on config backend. [Reference](https://github.com/rudderlabs/rudder-config-backend/blob/18169c409dc9b0dce137c7467da99b26112af758/src/swagger3/paths/dataPlanePaths.yaml#L166)

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/Use-cached-endpoint-for-hosted-configs-11a9ce59c6fa4917a94ad2a363c733ba)

## Security

### DIFF
--- a/config/backend-config/hosted_workspaces.go
+++ b/config/backend-config/hosted_workspaces.go
@@ -79,7 +79,12 @@ func (multiWorkspaceConfig *HostedWorkspacesConfig) GetWorkspaceLibrariesForWork
 
 //Get returns sources from all hosted workspaces
 func (multiWorkspaceConfig *HostedWorkspacesConfig) Get(_ string) (ConfigT, bool) {
-	url := fmt.Sprintf("%s/hostedWorkspaceConfig?fetchAll=true", configBackendURL)
+	var url string
+	if config.GetBool("BackendConfig.cachedHostedWorkspaceConfig", true) {
+		url = fmt.Sprintf("%s/cachedHostedWorkspaceConfig", configBackendURL)
+	} else {
+		url = fmt.Sprintf("%s/hostedWorkspaceConfig?fetchAll=true", configBackendURL)
+	}
 
 	var respBody []byte
 	var statusCode int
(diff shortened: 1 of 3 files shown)


config/backend-config/multitenant_workspaces.go
```diff
@@ -82,7 +82,11 @@ func (workspaceConfig *MultiTenantWorkspacesConfig) getFromAPI(workspaceArr stri
 	var url string
 	// TODO: hacky way to get the backend config for multi tenant through older hosted backed config
 	if config.GetBool("BackendConfig.useHostedBackendConfig", false) {
-		url = fmt.Sprintf("%s/hostedWorkspaceConfig?fetchAll=true", configBackendURL)
+		if config.GetBool("BackendConfig.cachedHostedWorkspaceConfig", true) {
```
cisse21, line 85: Let's also have this check in hosted_workspaces.go

cisse21 (changes requested):

config/backend-config/hosted_workspaces.go
```diff
@@ -79,7 +79,12 @@ func (multiWorkspaceConfig *HostedWorkspacesConfig) GetWorkspaceLibrariesForWork
 
 //Get returns sources from all hosted workspaces
 func (multiWorkspaceConfig *HostedWorkspacesConfig) Get(_ string) (ConfigT, bool) {
-	url := fmt.Sprintf("%s/hostedWorkspaceConfig?fetchAll=true", configBackendURL)
+	var url string
+	if config.GetBool("BackendConfig.cachedHostedWorkspaceConfig", true) {
+		url = fmt.Sprintf("%s/cachedHostedWorkspaceConfig", configBackendURL)
```
atzoum, line 84: Why do we need a separate URL for that? We should use standard HTTP caching constructs, e.g. use the same URL and control caching through the `cache-control` request header.

cisse21 (approved):

config/backend-config/hosted_workspaces.go
cisse21: Not sure why but a new endpoint has been implemented by the control-plane team, Ref : [here](https://rudderlabs.slack.com/archives/C02AASYS6R1/p1653293448145139?thread_ts=1653292691.828539&cid=C02AASYS6R1)

lvrach: That would be ideal, but the control-plane team has already implemented the endpoint like this. We have started a discussion to improve those endpoints overall, but for now, is important to utilise the existing endpoint to elevate the load from them.

lvrach: cc @fxenik

PR #1998, opened 2022-05-29
chore(router): improve tests for router pickup algorithm

cisse21:
# Description

This PR aims to add more test cases to the existing router pick up algorithm

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=c853fed058a941a58743cdac3372bba6)

## Security


services/multitenant/tenantstats_test.go
```diff
@@ -35,6 +36,10 @@ var (
 	workspaceID3 = uuid.Must(uuid.NewV4()).String()
 )
 
+func almostEqual(a, b float64) bool {
+	return math.Abs(a-b) <= float64EqualityThreshold
+}
```
fracasula, line 41: I think that there shoud be a `require.InDelta()` function that does this.

services/multitenant/tenantstats_test.go
```diff
@@ -44,117 +49,280 @@ var _ = Describe("tenantStats", func() {
...
+					//Making the failure rate for workspaceID1 high
+					tenantStats.CalculateSuccessFailureCounts(workspaceID1, destType1, false, false)
+				}
+				routerPickUpJobs, usedLatencies := tenantStats.GetRouterPickupJobs(destType1, noOfWorkers, routerTimeOut, 10000, timeGained)
+				// Jobs are picked from Input Loop
+				// Router In Rate was very low so the jobs had to be picked up from PileUp
+				// workspaceID1 has the highest failure rate so the least priority in PileUp
+				// The algorithm tries to pick the jobs likely to succeed from PileUp
+				// So the other two workspaces had the maximum picked up jobs(1000)
+				fmt.Println(routerPickUpJobs)
```
fracasula, line 316: If output is needed we can use `t.Log()` which respect `testing.Verbose()`.

fracasula (approved):

services/multitenant/tenantstats_test.go
fracasula: https://pkg.go.dev/github.com/stretchr/testify/require#InDelta

services/multitenant/tenantstats_test.go
cisse21: I missed removing the print statements post debugging. Let me remove this

services/multitenant/tenantstats_test.go
cisse21: Let me check this out and incorporate this

services/multitenant/tenantstats.go
```diff
@@ -18,7 +18,7 @@ var (
 	pkgLogger logger.LoggerI
 )
 
-type MultitenantStatsT struct {
+type StatsT struct {
```
lvrach, line 21: More idiomatic and readable like this :) 

I would go a step further and 
```suggestion
	t := Stats{}

File: processor/priorityqueue.go [hunk 4]
```diff
@@ -2,73 +2,9 @@ package processor
```
lvrach, line 2: After those changes, does the filename still remains relevant?

processor/manager.go
```diff
@@ -56,7 +56,8 @@ func (proc *LifecycleManager) Start() {
 func (proc *LifecycleManager) Stop() {
 	proc.currentCancel()
 	proc.HandleT.Shutdown()
-	proc.waitGroup.Wait()
+	//TODO : Handle the errors cleanly
+	_ = proc.waitGroup.Wait()
```
lvrach, line 60: No errors are expected from `waitGroup`, this is why the error is safe to be ignored. Although we should have been implicit about it.

cisse21: The linter pointed it out to handle the error so I added a TODO. Will remove this

services/multitenant/tenantstats.go
fracasula: +1 to that :slightly_smiling_face:

processor/priorityqueue.go
cisse21: We can rename it to transformQueue.go?

services/multitenant/tenantstats.go
fracasula: It means that there is no need for this naming convention of having the types end with a capital `T` e.g. `StatsT`. You can just call it `Stats` which is way more idiomatic in Golang. If it's a type or an interface is already determined by, well, its type :slightly_smiling_face:

processor/manager.go
fracasula: It looks like something that can be done in a separate PR, in my head when I saw this was happy with the TODO because otherwise we would need `Stop()` to also return an `error` and then have it handled all the way up (e.g. `func (proc *LifecycleManager) Stop() error`).

processor/priorityqueue.go
lvrach: all lowercase to be safe

lvrach (approved):

PR #1999, opened 2022-05-30
fix(jobsdb): mark MIGRATE_COPY done and POST_MIGRATE_DS_OP start atomically

atzoum:
# Description

Recovery doesn't delete new tables if a crash happens after the `MIGRATE_COPY` operation has completed, but before
`POST_MIGRATE_DS_OP` operation has started. As a result both old and new tables exist both containing the same jobs

- `MIGRATE_COPY` creates new tables 
- `POST_MIGRATE_DS_OP` drops or backups old ones

## Additional changes

- Replaced some panics by returning errors instead
- Functions requiring a `dsListLock` to be acquired have been refactored in order to enforce callers acquire such a lock instead of relying exclusively on good intention. For that purpose, `DSListLockToken` and `DSListLocker` have been introduced

## Notion Ticket

[Internal migration not recovering from crash happening after MIGRATE_COPY operation has completed, but before
POST_MIGRATE_DS_OP operation has started](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=68d1088e74804d448aacafd88c700a48)

[Update getDS(Range)List API to enforce proper locking](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=b9f77faa87c74a74bf15b5cbabde8f1d)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/cenkalti/backoff"
 	"github.com/rudderlabs/rudder-server/admin"
+	"github.com/rudderlabs/rudder-server/jobsdb/internal/lock"
 	"github.com/rudderlabs/rudder-server/jobsdb/prebackup"
 	"github.com/rudderlabs/rudder-server/utils/logger"
 	"github.com/tidwall/gjson"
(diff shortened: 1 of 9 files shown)


jobsdb/jobsdb.go
```diff
@@ -2118,13 +2124,13 @@ func (jd *HandleT) doClearCache(ds dataSetT, CVPMap map[string]map[string]map[st
 
 func (jd *HandleT) GetPileUpCounts(statMap map[string]map[string]int) {
 	jd.dsMigrationLock.RLock()
-	jd.dsListLock.RLock()
 	defer jd.dsMigrationLock.RUnlock()
-	defer jd.dsListLock.RUnlock()
+	// TODO: why do we need a full lock here?
```
atzoum, line 2128: @cisse21 is a full lock required?

cisse21 (approved):

lvrach (approved):

PR #2009, opened 2022-05-31
chore: apply release 2.0 for oss images

lvrach:
# Description

This is the second part of transitioning rudder-server to release 2.0.

In this part,  the GitHub actions files from [release-demo](https://github.com/rudderlabs/release-demo) have been introduced, but only for OSS images.

This is done to ensure that the following is working properly without risking breaking or delaying enterprise builds.

Enterprise builds are going to be introduced, as a next step, once OSS release flow is verified.

## Notion Ticket

[Notion link](https://www.notion.so/rudderstacks/release-2-0-for-oss-images-c26b66dbf5ef4bc69f250cc287d26657)

## Security


.github/workflows/apply-hot-fixes.yml
```diff
@@ -0,0 +1,16 @@
+name: Apply hot fixes to master
+on:
+  push:
+    branches:
+      - release/v*.*.*
+jobs:
+  auto-pull-request:
+    name: PullRequestAction
+    runs-on: ubuntu-latest
```
cisse21, line 9: not an expert here but should we be using ubuntu-latest here? or should we stick to a version

lvrach: It is not possible to specify an exact [version for your os](https://github.com/actions/virtual-environments)

The difference between `ubuntu-latest` and `ubuntu-20.04` is at some point `ubuntu-latest` is going to change to the next version `ubuntu-22.04`.

I think this is fine, as all the most popular actions we are kept updated to date with those changes. We might have a broken action when `ubuntu-latest` changes, but this is a once a year thing.

Most importantly, the libraries/tools of `ubuntu-latest` are not used to build our code, our code is built inside a more controlled docker environment.

.github/workflows/builds.yml
```diff
@@ -0,0 +1,54 @@
...
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            name=rudderlabs/develop-rudder-server
+            name=rudderlabs/rudder-server,enable=${{ github.ref == format('refs/heads/{0}', 'master') }}
```
atzoum, line 26: We are consciously omitting OSS image builds for patch releases, correct?

atzoum (approved):
LGTM

.github/workflows/builds.yml
lvrach: That was not intentional actually, I need to enable non-develop images for `release/*` branches as well.

.github/workflows/builds.yml
```diff
@@ -0,0 +1,54 @@
...
+            type=raw,value=${{ github.head_ref }}
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+
+      -
+        name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: rudderlabs
```
fracasula, line 39: [minor] This could be a secret as well as showed in the first example [here](https://github.com/marketplace/actions/docker-login).

.github/workflows/builds.yml
```diff
@@ -0,0 +1,54 @@
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: Build and push
+        uses: docker/build-push-action@v3
```
fracasula, line 43: If you want to speed this up you could use the `buildx` cache like I did [here](https://github.com/rudderlabs/rudder-scheduler/blob/29632ab300b79a1f488cdb08a43231540d58a7de/.github/workflows/tests.yaml#L37) (see `name: Cache Docker layers` and then the `cache-from` and `cache-to` in the `docker/build-push-action@v3`).

fracasula (approved):

PR #2011, opened 2022-06-01
feat(WH): skip namespace snake casing based on config

Jayachand:
# Description

skip namespace snake casing based on configuration

## Notion Ticket

https://www.notion.so/rudderstacks/kogan-skip-namespace-snake-casing-413aefa86ccc4e3596db22af5115a651

## Security


abhimanyubabbar (commented):
The changes look good to me. I was wondering if along with change, we can quickly add base test case for the change as well ?

Wdyt ?

PR #2018, opened 2022-06-02
feat(multi-tenant): multitenant support for embedded app

kuldeep0020:
# Description

Enabling support to use Embedded App in the multi tenant deployment. In this case the ETCD based cluster manager will be used. 

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/Support-for-Embedded-in-Multi-Tenant-environment-b1ff065508534fdfa30b6bab4b795948)

## Security

### DIFF
--- a/app/cluster/dynamic.go
+++ b/app/cluster/dynamic.go
@@ -187,12 +187,12 @@ func (d *Dynamic) handleWorkspaceChange(ctx context.Context, workspaces string) 
 }
 
 func (d *Dynamic) handleModeChange(newMode servermode.Mode) error {
-	if !newMode.Valid() {
-		return fmt.Errorf("unsupported mode: %s", newMode)
-	}
 	if d.GatewayComponent {
 		d.logger.Info("Not transiting the server because this is only Gateway App")
 		return nil
+	}
+	if !newMode.Valid() {
+		return fmt.Errorf("unsupported mode: %s", newMode)
 	}
 
 	if d.currentMode == newMode {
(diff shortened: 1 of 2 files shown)


app/cluster/dynamic.go
```diff
@@ -187,13 +187,13 @@ func (d *Dynamic) handleWorkspaceChange(ctx context.Context, workspaces string)
...
 
 func (d *Dynamic) handleModeChange(newMode servermode.Mode) error {
-	if !newMode.Valid() {
-		return fmt.Errorf("unsupported mode: %s", newMode)
-	}
 	if d.GatewayComponent {
 		d.logger.Info("Not transiting the server because this is only Gateway App")
 		return nil
 	}
+	if !newMode.Valid() {
```
atzoum, line 194: why did we move it down? Shouldn't we start with a valid mode for gateway too?

kuldeep0020: In case of production environment, where gateway and processor runs separately, any mode change (valid/invalid) does not affect the gateway app, so if it is a gateway app, we skip any mode change. So in order to do early exit from this method, I reordered these conditions check. 
And as per discussion with @lvrach and @cisse21, it's least likely that we will ever introduce the mode change to Gateway app, and if we do then there will be more changes, ticket to which is [this](https://www.notion.so/rudderstacks/Improve-Cluster-Manager-to-Introduce-Gateway-App-Properly-5ca903e68f9b454697a1164488241485)

cisse21 (approved):

PR #2020, opened 2022-06-03
fix(warehouse): use correct locks for dateformat in processor

achettyiitr:
# Description

Concurrent map writes fix for getting storage date format

## Notion Ticket

https://www.notion.so/rudderstacks/Adhoc-Instamart-concurrent-map-issue-b78f52924ee44d6fb41b872df934747a

## Security


cisse21 (approved):

abhimanyubabbar (commented):
Should we think about adding a benchmark test to this PR ? If not then we can create a notion ticket and handle it separately.

abhimanyubabbar (approved):
LGTM !

PR #2031, opened 2022-06-06
chore: execute ANALYZE after copying a large number of jobs or job statuses

atzoum:
# Description

Assisting query planning in case a large number of jobs or job statuses are copied (i.e. batchrouter)

## Notion Ticket

[Execute ANALYZE after a bulk store/update threshold has been reached](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=a05d41a4f89044939aac8ff2016c7147)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -423,6 +423,7 @@ type HandleT struct {
 	maxReaders                    int
 	maxWriters                    int
 	maxOpenConnections            int
+	analyzeThreshold              int
 	MaxDSSize                     *int
 	queryFilterKeys               QueryFiltersT
 	backgroundCancel              context.CancelFunc
@@ -870,6 +871,8 @@ func (jd *HandleT) workersAndAuxSetup() {
 	config.RegisterIntConfigVariable(3, &jd.maxReaders, false, 1, maxReadersKeys...)
 	maxOpenConnectionsKeys := []string{"JobsDB." + jd.tablePrefix + "." + "maxOpenConnections", "JobsDB." + "maxOpenConnections"}
 	config.RegisterIntConfigVariable(20, &jd.maxOpenConnections, false, 1, maxOpenConnectionsKeys...)
+	analyzeThresholdKeys := []string{"JobsDB." + jd.tablePrefix + "." + "analyzeThreshold", "JobsDB." + "analyzeThreshold"}
+	config.RegisterIntConfigVariable(30000, &jd.analyzeThreshold, false, 1, analyzeThresholdKeys...)
 
 }
 
@@ -2229,12 +2232,12 @@ func (*HandleT) copyJobsDSInTx(txHandler transactionHandler, ds dataSetT, jobLis
 	return err
 }
 
-func (*HandleT) doStoreJobsInTx(txHandler transactionHandler, ds dataSetT, jobList []*JobT) error {
+func (jd *HandleT) doStoreJobsInTx(tx *sql.Tx, ds dataSetT, jobList []*JobT) error {
 	store := func() error {
 		var stmt *sql.Stmt
 		var err error
 
-		stmt, err = txHandler.Prepare(pq.CopyIn(ds.JobTable, "uuid", "user_id", "custom_val", "parameters", "event_payload", "event_count", "workspace_id"))
+		stmt, err = tx.Prepare(pq.CopyIn(ds.JobTable, "uuid", "user_id", "custom_val", "parameters", "event_payload", "event_count", "workspace_id"))
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


lvrach:
What is the performance impact on running analyze ?

atzoum:
> What is the performance impact on running analyze ?

In a production postgresql pod:
- jobs table (163MB) with 100000 jobs: 900ms
- job status table (40MB) with 200000 entries: 250ms

I would say it is proportional to the table's size

lvrach:
> * 900ms

Those times could have a significant impact on the throughput of the rudder-server, I mostly worry about the processor were storing more than 30k jobs to router and/or batch router could be common. 

Adding 900ms or double that could reduce our throughput in a significant way (I don't have hard numbers, but we need to test)

atzoum:
> Adding 900ms or double that could reduce our throughput in a significant way (I don't have hard numbers, but we need to test)

I would argue against that, since we need to take more parameters into consideration:
1. The numbers I presented are worst case scenarios, the 1st bulk insert of 30000+ jobs will require `300ms` overhead, the 2nd `600ms`, the third `900ms` & after that a new table will be created so we will reset.
2. We shouldn't factor out the drastic improvement that this will have on query performance, since data might be stored once, but it is queried more than once.
3. The choice of the default threshold of 30000 appears to be correct and to the point. Here are two examples from actual production environments that can support this argument:
   a. [Example 1](https://hostedmt-grafana.rudderstack.com/grafana/d/hRrW_4Cnz/processor-events-store?orgId=1&from=now-30d&to=now) - here we are running fine and we never go beyond the threshold, so no overhead will be introduced.
   b. [Example 2](https://grafana-magicedenhfx.dataplane.rudderstack.com/grafana/d/hRrW_4Cnz/processor-events-store?orgId=1&from=now-30d&to=now) - here we are facing problems and we desperately need ANALYSE beyond the 30000 threshold.

fracasula:
> What is the performance impact on running analyze ?

I'm interested in the opposite. What is the performance impact of **NOT** running analyze?
Do you have some numbers showing how much `ANALYZE` (more or less) could help us when we run it like this vs a daily vacuum?

atzoum:
> I'm interested in the opposite. What is the performance impact of **NOT** running analyze?

There is no single answer to that, but without proper statistics the query planner takes bad decisions. 
They can be that bad, up to the point of bringing our system to a halt!

lvrach:
> * The numbers I presented are worst case scenarios, the 1st bulk insert of 30000+ jobs will require `300ms` overhead, the 2nd `600ms`, the third `900ms` & after that a new table will be created so we will reset.
> * We shouldn't factor out the drastic improvement that this will have on query performance, since data might be stored once, but it is queried more than once.

Query performance is improved only under specific cases, for example, if there are no errors we write once and read once. Under normal cases, we have no or single status per job_id.

Given how our pipeline works, I would say write performance is more important than read. Except if read is 10x - 100x more

atzoum:
> if there are no errors we write once and read once
Unfortunately it is a bit more complicated than that, e.g. for the jobs table we read at least twice (`getToRetry` & `getUnprocessed`) multiplied by destinationType. For job status table, we read once multiplied by destinationType.
Additionally, every single query mentioned above may cascade to more than one datasets, depending on the jobsdb's state and event distribution.

atzoum:
> potential impact of this change

The performance impact in our typical scenarios is negligible since we don't reach the 30000 threshold.

The truth is that there is not one single query that gets executed everytime or a single distribution of data that gets inserted in the tables, that's why I question the usefulness of a controlled test. Otherwise, yes we can perform a controlled test but what is it going to really answer about the actual behaviour of the system in production? So far, we don't have an example in production where we constantly insert more than 30K rows and we don't end up with our system halting, so for the time being it is not a matter of performance, but survival.

If we do find such an example, it is easy to switch it on and off (use 1000000 as threshold) and compare performance, eg. using ANALYZE vs using another query which performs worse than the current one under normal conditions, but it doesn't bring our system to a halt when executed with bad statistics (what Leo is proposing)

fracasula:
I like the idea of switching it on and off with ease. If you already have all metrics in place to measure the impact of these changes then I think we should go ahead.

Given that `ANALYZE` gets a read lock, do we have metrics/slow-query logs or something that would highlight a prolonged wait of writes for example? Just wondering if we got everything in place to detect possible anomalies.

lvrach (approved):
> The performance impact in our typical scenarios is negligible since we don't reach the 30000 threshold.

Our maximum pickup size in the processor is 10k, assuming someone has 3 router or batch router destinations, it can just reach this threshold for jobsdb. 

I agree this is an edge case, and we can just make the threshold 30k + 1 to avoid this scenario.

I also have the opposite concern. Is 30k low enough? What if instead of a big 50k status update, we have many smaller ones? Assume 5 router destinations with equal distribution of events all failing. But again, we have control we can play around it.

> being it is not a matter of performance, but survival
Given that we have provided a hotfix for the issue, I don't believe things are so critical we should rush with this fix.

All that being said, since:
* It is recommended by postgres to run an analyze after a batch insert 
* We have a way to effectively disable this
* The negative impact would be minimal / restricted 
* We never going to learn otherwise

fracasula (approved):

PR #2032, opened 2022-06-06
chore: limit scope of customdestinationmanager maplock

atzoum:
# Description

## Notion Ticket

[Limit scope of customdestinationmanager maplock](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=0a4fb8aab9854b94904e08c4612bd531)

## Security


router/customdestinationmanager/customdestinationmanager.go
```diff
@@ -283,18 +287,22 @@ func New(destType string, o Opts) DestinationManager {
 		}
```
fracasula, line 287: Forgive the silly question, but where are we writing into `customManagerMap`? :thinking:

router/customdestinationmanager/customdestinationmanager.go
```diff
@@ -317,7 +329,9 @@ func (customManager *CustomManagerT) backendConfigSubscriber() {
 				}
 			}
 		}
-		customManager.mapLock.Unlock()
+		once.Do(func() {
+			customManager.backendConfigInitialized <- struct{}{}
+		})
```
fracasula, line 334: To me now this method looks like "it's fine if you call it multiple times but we'll tell you when the intialization is done only the first time".
I wonder if the whole initilization should be "onced" then and not just the last bit? wdyt?

router/router.go
```diff
@@ -2306,6 +2305,9 @@ func (rt *HandleT) Start() {
 	ctx := rt.backgroundCtx
 	rt.backgroundGroup.Go(func() error {
 		<-rt.backendConfigInitialized
+		if rt.customDestinationManager != nil {
+			<-rt.customDestinationManager.BackendConfigInitialized()
```
fracasula, line 2309: Given that we got a `ctx` here, wouldn't it make sense to `select` here to avoid any possible leak?

router/customdestinationmanager/customdestinationmanager.go
```diff
@@ -220,6 +222,8 @@ func (customManager *CustomManagerT) refreshClient(destID string) error {
 }
 
 func (customManager *CustomManagerT) onNewDestination(destination backendconfig.DestinationT) error { // skipcq: CRT-P0003
+	customManager.mapLock.Lock()
+	defer customManager.mapLock.Unlock()
```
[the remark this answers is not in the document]
atzoum: yes it is hard, apparently it protects multiple resources, will call it `stateMu`

router/customdestinationmanager/customdestinationmanager.go
atzoum: it is not about calling the method more than one, it is about reading from the channel more than once, so a new `sync.Once` would be required for what you describe, correct?

router/customdestinationmanager/customdestinationmanager.go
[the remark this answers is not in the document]
fracasula: Can you please add a comment saying what resources it is meant to protect? Thanks.

router/customdestinationmanager/customdestinationmanager.go
fracasula: I was under the impression that it doesn't make sense to call `customManager.backendConfigSubscriber()` more than once because once subscribed I guess it'll keep getting configs as they come from the backend config. So I was proposing to have the entire content of that method inside `once.Do()` or, if possible, it would be even better to remove the `sync.Once` altogether because it seems to me that we're invoking this method only once per `customManager` instance inside the `New()` constructor [here](https://github.com/rudderlabs/rudder-server/blob/fca1c1d12fd7a026e442c83e7914354f6b8a4ae6/router/customdestinationmanager/customdestinationmanager.go#L304).

But perhaps I'm missing something, let me know if I'm not making sense.

fracasula (approved):
Looks good, although some tests would be welcome :+1:

cisse21 (approved):

PR #2033, opened 2022-06-07
fix(warehouse): columns need to be provided during copy command for deltalake

achettyiitr:
# Description

Handle extra columns when creating COPY command for deltalake destination because since we are using auto generated columns we are forced to do it.
Added corresponding integration test at [here](https://github.com/rudderlabs/rudder-server/blob/959c850294365c4a6a798d7baa17fdff278caa1e/warehouse/deltalake/deltalake_test.go).

## Notion Ticket

https://www.notion.so/rudderstacks/Adhoc-Deltalake-columns-issue-8c10aa27a58a422dbcf2495c072fd076

## Security

### DIFF
--- a/warehouse/deltalake/deltalake.go
+++ b/warehouse/deltalake/deltalake.go
@@ -429,7 +429,7 @@ func (dl *HandleT) sortedColumnNames(tableSchemaInUpload warehouseutils.TableSch
 		formatString := warehouseutils.JoinWithFormatting(sortedColumnKeys, format, ",")
 		if len(diff.ColumnMap) > 0 {
 			diffCols := make([]string, 0, len(diff.ColumnMap))
-			for key, _ := range diff.ColumnMap {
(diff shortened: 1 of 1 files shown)


dhawal1248 (approved):

warehouse/deltalake/deltalake.go
```diff
@@ -466,6 +478,19 @@ func (dl *HandleT) getLoadFolder(tableName string, location string) (loadFolder
 	return
 }
 
+func getTableSchemaDiff(tableSchemaInUpload, tableSchemaAfterUpload warehouseutils.TableSchemaT) (diff warehouseutils.TableSchemaDiffT) {
```
satishrudderstack, line 481: @achettyiitr Can we rename tableSchemaAfterUpload to tableSchemaInWarehouse just to be consistent ?

PR #2040, opened 2022-06-08
fix(gateway): getUsersPayload optimizations

fracasula:
# Description

We're getting some big payloads in the `POST /v1/import` [endpoint](https://github.com/rudderlabs/rudder-server/blob/7f4704cc4f2f3260417a14b3cfeaf5d7ae23f04f/gateway/gateway.go#L722) so I looked into how we could optimize the algorithm there.

I ended up with different versions of the `getUsersPayload` algorithm which I all benchmarked.

## Benchmark results with real customer data (1.2MB payload)

```
goos: linux
goarch: amd64
pkg: github.com/rudderlabs/rudder-server/gateway
BenchmarkGetUsersPayload
    benchmark_getUsersPayload_test.go:102: Running benchmark with 1.2 MiB payload
BenchmarkGetUsersPayload/original
BenchmarkGetUsersPayload/original-24				1	1870702191 ns/op	 9774432 B/op	   20315 allocs/op
BenchmarkGetUsersPayload/easyjson-double-alloc
BenchmarkGetUsersPayload/easyjson-double-alloc-24		3	 467994466 ns/op	 8588016 B/op	   16823 allocs/op
BenchmarkGetUsersPayload/easyjson-single-alloc-md5-cache
BenchmarkGetUsersPayload/easyjson-single-alloc-md5-cache-24	3	 461353689 ns/op	 4721061 B/op	    9019 allocs/op
BenchmarkGetUsersPayload/easyjson-no-md5-cache
BenchmarkGetUsersPayload/easyjson-no-md5-cache-24		3	 464432509 ns/op	 4551504 B/op	    8977 allocs/op
BenchmarkGetUsersPayload/fastjson
BenchmarkGetUsersPayload/fastjson-24				273	   4256768 ns/op	13703286 B/op	   26105 allocs/op
BenchmarkGetUsersPayload/gjson-revised
BenchmarkGetUsersPayload/gjson-revised-24			224	   5418472 ns/op	 4321157 B/op	    6071 allocs/op
PASS
```

## Benchmark results with generated payload (1.5MB)

An important difference between this benchmark and the one with real customer data, is that the generated payload has only unique userIDs (they never repeat).

```
goos: linux
goarch: amd64
pkg: github.com/rudderlabs/rudder-server/gateway
BenchmarkGetUsersPayload
    benchmark_getUsersPayload_test.go:99: Running benchmark with 1.5 MiB payload
BenchmarkGetUsersPayload/original
BenchmarkGetUsersPayload/original-24				1	26952707888 ns/op	23821448 B/op	  400372 allocs/op
BenchmarkGetUsersPayload/easyjson-double-alloc
BenchmarkGetUsersPayload/easyjson-double-alloc-24		1	8904738294 ns/op	23974352 B/op	  320905 allocs/op
BenchmarkGetUsersPayload/easyjson-single-alloc-md5-cache
BenchmarkGetUsersPayload/easyjson-single-alloc-md5-cache-24	1	8783045655 ns/op	18585528 B/op	  181511 allocs/op
BenchmarkGetUsersPayload/easyjson-no-md5-cache
BenchmarkGetUsersPayload/easyjson-no-md5-cache-24		1	8861490915 ns/op	15394672 B/op	  160977 allocs/op
BenchmarkGetUsersPayload/fastjson
BenchmarkGetUsersPayload/fastjson-24				44	  23067574 ns/op	52971111 B/op	  221139 allocs/op
BenchmarkGetUsersPayload/gjson-revised
BenchmarkGetUsersPayload/gjson-revised-24			82	  13937215 ns/op	11428351 B/op	   81090 allocs/op
PASS
```

The `fastjson` and `gjson-revised` versions seems to be a lot faster (273 and 224 times faster when testing with customer data) but the `fastjson` one uses a lot more memory than the original one (see results above).

The `gjson-revised` seems to be faster than `fastjson` with generated payloads. The difference with a generated payloads is that the events all have unique userIDs (they never repeat throughout the whole JSON).

## Running the new algo with a big payload (~77MB) from a real customer:

```go
func TestBigPayload(t *testing.T) {
	validBody, err := os.ReadFile("./testdata/output.json")
	require.NoError(t, err)

	_, err = getUsersPayloadGJsonRevised(validBody)
	require.NoError(t, err)
}
```

```
=== RUN   TestBigPayload
--- PASS: TestBigPayload (0.42s)
PASS
```

I get the same result (i.e. ~0.42s) with both `fastjson` and `gjson-revised`.

**IMPORTANT NOTE:** The above **NEVER COMPLETES** with the original algorithm, I left it running for more than 20 minutes then I had to kill the process.

## Reasonable payload

```
goos: linux
goarch: amd64
pkg: github.com/rudderlabs/rudder-server/gateway
BenchmarkFindReasonablePayload
    benchmark_getUsersPayload_test.go:210: Processing 7.5 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 72.837591ms
    benchmark_getUsersPayload_test.go:210: Processing 15.3 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 143.432382ms
    benchmark_getUsersPayload_test.go:210: Processing 23.1 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 240.87901ms
    benchmark_getUsersPayload_test.go:210: Processing 31.0 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 298.260138ms
    benchmark_getUsersPayload_test.go:210: Processing 38.8 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 436.473772ms
    benchmark_getUsersPayload_test.go:210: Processing 46.6 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 519.007825ms
    benchmark_getUsersPayload_test.go:210: Processing 54.4 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 558.771341ms
    benchmark_getUsersPayload_test.go:210: Processing 62.2 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 646.498111ms
    benchmark_getUsersPayload_test.go:210: Processing 70.1 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 795.531954ms
    benchmark_getUsersPayload_test.go:210: Processing 77.9 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 895.189155ms
    benchmark_getUsersPayload_test.go:210: Processing 86.0 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 989.841795ms
    benchmark_getUsersPayload_test.go:210: Processing 94.1 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 1.047628733s
    benchmark_getUsersPayload_test.go:210: Processing 102.2 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 1.136047053s
    benchmark_getUsersPayload_test.go:210: Processing 110.3 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 1.184535819s
    benchmark_getUsersPayload_test.go:210: Processing 118.4 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 1.249934745s
    benchmark_getUsersPayload_test.go:210: Processing 126.5 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 1.32951889s
    benchmark_getUsersPayload_test.go:210: Processing 134.6 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 1.38898445s
    benchmark_getUsersPayload_test.go:210: Processing 142.7 MiB payload
    benchmark_getUsersPayload_test.go:220: Processing took 1.71509973s
    benchmark_getUsersPayload_test.go:210: Processing 150.8 MiB payload
    benchmark_getUsersPayload_test.go:223: Payload of 150.8 MiB took more than 2s to process
BenchmarkFindReasonablePayload-24    	       1	16507293728 ns/op
PASS
```

We should now be able to process ~143MB payloads in less than 2 seconds :+1: 

## Memory usage with real customer data (payload ~77MB) and fastjson

Screenshot from 2022-06-08 16-02-35

Although way faster, it looks like `fastjson` is a bit too famished when it comes to memory consumption.
Even though we might very well be able to run this, the memory consumption is indeed worrying with very large payloads and could probably lead to OOM kills.

## Memory usage with real customer data (payload ~77MB) and gjson-revised

Screenshot from 2022-06-09 08-29-19

## Conclusion

I'm therefore proposing the `gjson-revised` to be used. Even though it seems slightly slower than the `fastjson` counterpart in some scenarios, the delta is in my opinion negligible and it still allows us to process big payloads in a very reasonable time. The memory consumption of the `gjson-revised` version though is much better than `fastjson`.

## Notion Ticket

[< Notion Link >](https://www.notion.so/rudderstacks/Optimise-import-endpoint-for-large-payloads-868c94eb305a4a9e80ceb986d887ddf1)

## Security


atzoum:
@fracasula just for fun using jsoniter
```
pkg: github.com/rudderlabs/rudder-server/gateway
BenchmarkGetUsersPayload/original-10         	       1	31505665375 ns/op	23820224 B/op	  400392 allocs/op
BenchmarkGetUsersPayload/jsoniter-10         	       1	25948272584 ns/op	6590603616 B/op	600790500 allocs/op
BenchmarkGetUsersPayload/fastjson-10         	      57	  20193770 ns/op	52970692 B/op	  221137 allocs/op
```
```
func getUsersPayloadJsoniter(requestPayload []byte) (map[string][]byte, error) {
	batch := jsoniter.Get(requestPayload, "batch")
	if batch.LastError() != nil {
		return nil, errors.New(response.InvalidJSON)
	}
	if batch.ValueType() != jsoniter.ArrayValue {
		return nil, errors.New(response.InvalidJSON)
	}
	events := batch.Size()

	type userState struct {
		count  int
		buf    *bytes.Buffer
		writer *bufio.Writer
		stream *jsoniter.Stream
	}
	var (
		userMap = make(map[string]*userState)
	)

	for i := 0; i < events; i++ {
		evt := batch.Get(i)
		userID := evt.Get("userId")
		anonymousID := evt.Get("anonymousId")
		var userIDStr, anonymousIDStr string
		if userID.ValueType() == jsoniter.StringValue {
			userIDStr = userID.ToString()
		}
		if anonymousID.ValueType() == jsoniter.StringValue {
			anonymousIDStr = anonymousID.ToString()
		}
		rudderID, err := misc.GetMD5UUID(userIDStr + ":" + anonymousIDStr)
		if err != nil {
			continue
		}

		uuidStr := rudderID.String()
		us, ok := userMap[uuidStr]
		if !ok {
			var b bytes.Buffer
			writer := bufio.NewWriter(&b)
			stream := jsoniter.NewStream(jsoniter.ConfigDefault, writer, 4096)
			us = &userState{
				buf:    &b,
				writer: writer,
				stream: stream,
			}
			stream.WriteObjectStart()
			stream.WriteObjectField("batch")
			stream.WriteArrayStart()
			userMap[uuidStr] = us

		}
		if us.count > 0 {
			us.stream.WriteMore()
		}
		evt.WriteTo(us.stream)
		us.count++
	}

	finalMap := make(map[string][]byte, len(userMap))
	for k, us := range userMap {
		us.stream.WriteArrayEnd()
		us.stream.WriteObjectEnd()
		us.stream.Flush()
		us.writer.Flush()
		finalMap[k] = us.buf.Bytes()
	}

	return finalMap, nil
}
```

fracasula:
> @fracasula just for fun using jsoniter

Cool, thanks! :+1: 

I think the main issue with the original approach was the [gjson.GetBytes](https://github.com/rudderlabs/rudder-server/blob/6996d7eeae8f0c85c5c89b51264671e3d6ba68d7/gateway/gateway.go#L1176) because as we go deeper and deeper, each call is basically starting from the beginning killing the performance.

Anyway, I think I found another way to do it with `gjson`, the performance is quite similar as the `fastjson` version but I haven't had the time to complete the memory profiling. I'll add it first thing tomorrow morning :+1:

fracasula:
@atzoum 

```
BenchmarkGetUsersPayload/fastjson
BenchmarkGetUsersPayload/fastjson-20                                	      43	  29372565 ns/op	26198313 B/op	  110577 allocs/op
BenchmarkGetUsersPayload/gjson-revised
BenchmarkGetUsersPayload/gjson-revised-20                           	      76	  15099971 ns/op	 5708752 B/op	   40529 allocs/op
```

This is what I meant above. Will do the memory profiling tomorrow but it looks very promising! :slightly_smiling_face:

gateway/gateway.go
```diff
@@ -1146,50 +1144,51 @@ func (irh *ImportRequestHandler) ProcessRequest(gateway *HandleT, w *http.Respon
 		gateway.addToWebRequestQ(w, r, done, "batch", usersPayload[key], writeKey)
```
cisse21, line 1144: Not related to the scope of the PR but I guess we should also be overriding the AnonymousID in the header so that we can route each user's payload to the appropriate worker

cisse21 (approved):

go.mod
```diff
@@ -54,6 +55,7 @@ require (
 	github.com/thoas/go-funk v0.9.1
 	github.com/tidwall/gjson v1.10.2
 	github.com/tidwall/sjson v1.0.4
+	github.com/valyala/fastjson v1.6.3
```
atzoum, line 58: After we reach to a conclusion regarding which approach we are going to follow and before merging this, we shall cleanup the dependencies here.
If we need to keep the benchmarks for posterity, we shall do so in a separate module, either here or in [rudder-tests](https://github.com/rudderlabs/rudder-tests)

fracasula: Yes I totally agree. Before merging I'll remove everything. I was thinking maybe a link to this PR for posterity in the `getUsersPayload` would be enough.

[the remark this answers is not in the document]
fracasula: https://github.com/rudderlabs/rudder-server/pull/2040#discussion_r893226887

atzoum (approved):

PR #2042, opened 2022-06-09
feat[integration]: Putting support for AVRO in Kafka.

ItsSudip:
# Description
 
AVRO is a system that helps in data serialization. We are using this for serializing Kafka data. We are taking the help of `goavro` by [LinkedIn](https://github.com/linkedin/goavro) to serialize Kafka data to AVRO. 
We are using the `Conver To AVRO` button in Webapp to decide if we need to convert the data or not. There we are taking an array of schemas and trying to convert the data using the schema one by one.

## Notion Ticket

https://www.notion.so/rudderstacks/d5d15ce4be354e3caae11d4ea9f41477?v=2ad23dce1aeb431a82f783e6607d3cb0&p=344b266f78904c4190f0ffd7efe54ade

## Security

### DIFF
--- a/services/streammanager/kafka/kafkamanager.go
+++ b/services/streammanager/kafka/kafkamanager.go
@@ -9,16 +9,21 @@ import (
 	"strings"
 	"time"
 
-	"github.com/tidwall/gjson"
-
+	"github.com/linkedin/goavro"
 	"github.com/rudderlabs/rudder-server/config"
 	"github.com/rudderlabs/rudder-server/services/stats"
 	"github.com/rudderlabs/rudder-server/services/streammanager/kafka/client"
 	rslogger "github.com/rudderlabs/rudder-server/utils/logger"
+	"github.com/tidwall/gjson"
 )
 
 type Opts struct {
 	Timeout time.Duration
+}
+
+// Schema is the AVRO schema required to convert the data to AVRO
+type Schema struct {
+	AVROSchema string
 }
 
 // configuration is the config that is required to send data to Kafka
@@ -32,6 +37,8 @@ type configuration struct {
 	SaslType      string
 	Username      string
 	Password      string
+	ConvertToAvro bool
+	AVROSchema    []Schema
 }
 
 func (c *configuration) validate() error {
@@ -394,9 +401,48 @@ func prepareMessage(topic, key string, message []byte, timestamp time.Time) clie
 	}
 }
 
-func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timestamp time.Time) (
+// This function is used to create the error string of idividual schema
+func createErrString(err error, index int) string {
+	return fmt.Sprintf(" Schema[%d]: ", index+1) + err.Error() + "."
+}
+
+// This function is used to serialize the binary data according to the AVROSchema.
+// It iterate over the schemas provided by the customer and try to serialize the data.
+// If it ables to serialize the data then it returns the converted data otherwise it throws error.
+// We are using linkedin goavro library to serialize the data. For ref: https://github.com/linkedin/goavro
+
+func serialize(value []byte, AVROSchema []Schema) ([]byte, error) {
+	var errorStr string
+	for i := 0; i < len(AVROSchema); i++ {
+		codec, err := goavro.NewCodec(AVROSchema[i].AVROSchema)
+		if err != nil {
+			errorStr += createErrString(err, i)
+			continue
+		}
+		native, _, err := codec.NativeFromTextual(value)
+		if err != nil {
+			errorStr += createErrString(err, i)
+			continue
+		}
+		binary, err := codec.BinaryFromNative(nil, native)
+		if err != nil {
+			errorStr += createErrString(err, i)
+			continue
+		}
+		if err == nil {
+			return binary, nil
+		} else {
+			errorStr += createErrString(err, i)
+		}
+	}
+	return nil, fmt.Errorf("unable convert the event with any of the given schema." + errorStr)
+}
+
+func prepareBatchOfMessages(conf configuration, batch []map[string]interface{}, timestamp time.Time) (
 	[]client.Message, error,
 ) {
+	convertToAvro := conf.ConvertToAvro
+	AVROSchema := conf.AVROSchema
 	start := now()
 	defer func() { kafkaStats.prepareBatchTime.SendTiming(since(start)) }()
 
@@ -405,21 +451,26 @@ func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timest
 		message, ok := data["message"]
 		if !ok {
 			kafkaStats.missingMessage.Increment()
-			pkgLogger.Errorf("batch from topic %s is missing the message attribute", topic)
+			pkgLogger.Errorf("batch from topic %s is missing the message attribute", conf.Topic)
 			continue
 		}
 		userID, ok := data["userId"].(string)
 		if !ok && !allowReqsWithoutUserIDAndAnonymousID {
 			kafkaStats.missingUserID.Increment()
-			pkgLogger.Errorf("batch from topic %s is missing the userId attribute", topic)
+			pkgLogger.Errorf("batch from topic %s is missing the userId attribute", conf.Topic)
 			continue
 		}
-
 		marshalledMsg, err := json.Marshal(message)
 		if err != nil {
 			return nil, err
 		}
-		messages = append(messages, prepareMessage(topic, userID, marshalledMsg, timestamp))
+		if convertToAvro {
+			marshalledMsg, err = serialize(marshalledMsg, AVROSchema)
+			if err != nil {
+				return messages, err
+			}
+		}
+		messages = append(messages, prepareMessage(conf.Topic, userID, marshalledMsg, timestamp))
 	}
 	return messages, nil
 }
@@ -470,13 +521,13 @@ func Produce(jsonData json.RawMessage, pi interface{}, destConfig interface{}) (
 	ctx, cancel := context.WithTimeout(context.TODO(), p.getTimeout())
 	defer cancel()
 	if kafkaBatchingEnabled {
-		return sendBatchedMessage(ctx, jsonData, p, conf.Topic)
+		return sendBatchedMessage(ctx, jsonData, p, conf)
 	}
 
-	return sendMessage(ctx, jsonData, p, conf.Topic)
+	return sendMessage(ctx, jsonData, p, conf)
 }
 
-func sendBatchedMessage(ctx context.Context, jsonData json.RawMessage, p producer, topic string) (int, string, string) {
+func sendBatchedMessage(ctx context.Context, jsonData json.RawMessage, p producer, conf configuration) (int, string, string) {
 	var batch []map[string]interface{}
 	err := json.Unmarshal(jsonData, &batch)
 	if err != nil {
@@ -484,7 +535,7 @@ func sendBatchedMessage(ctx context.Context, jsonData json.RawMessage, p produce
 	}
 
 	timestamp := time.Now()
-	batchOfMessages, err := prepareBatchOfMessages(topic, batch, timestamp)
+	batchOfMessages, err := prepareBatchOfMessages(conf, batch, timestamp)
 	if err != nil {
 		return 400, "Failure", "Error while preparing batched message: " + err.Error()
 	}
@@ -497,8 +548,9 @@ func sendBatchedMessage(ctx context.Context, jsonData json.RawMessage, p produce
 	returnMessage := "Kafka: Message delivered in batch"
 	return 200, returnMessage, returnMessage
 }
-
-func sendMessage(ctx context.Context, jsonData json.RawMessage, p producer, topic string) (int, string, string) {
+func sendMessage(ctx context.Context, jsonData json.RawMessage, p producer, conf configuration) (int, string, string) {
+	convertToAvro := conf.ConvertToAvro
+	AVROSchema := conf.AVROSchema
 	parsedJSON := gjson.ParseBytes(jsonData)
 	messageValue := parsedJSON.Get("message").Value()
 	if messageValue == nil {
@@ -513,12 +565,18 @@ func sendMessage(ctx context.Context, jsonData json.RawMessage, p producer, topi
 
 	timestamp := time.Now()
 	userID, _ := parsedJSON.Get("userId").Value().(string)
-	message := prepareMessage(topic, userID, value, timestamp)
+	if convertToAvro {
+		value, err = serialize(value, AVROSchema)
+		if err != nil {
+			return makeErrorResponse(err)
+		}
+	}
+	message := prepareMessage(conf.Topic, userID, value, timestamp)
 	if err = publish(ctx, p, message); err != nil {
 		return makeErrorResponse(err)
 	}
 
-	returnMessage := fmt.Sprintf("Message delivered to topic: %s", topic)
+	returnMessage := fmt.Sprintf("Message delivered to topic: %s", conf.Topic)
 	return 200, returnMessage, returnMessage
 }
 
--- a/go.mod
+++ b/go.mod
@@ -176,3 +176,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+require github.com/linkedin/goavro v2.1.0+incompatible
+
+require gopkg.in/linkedin/goavro.v1 v1.0.5 // indirect
--- a/go.sum
+++ b/go.sum
@@ -919,6 +919,8 @@ github.com/lib/pq v1.8.0/go.mod h1:AlVN5x4E4T544tWzH6hKfbfQvm3HdbOxrmggDNAPY9o=
 github.com/lib/pq v1.10.0/go.mod h1:AlVN5x4E4T544tWzH6hKfbfQvm3HdbOxrmggDNAPY9o=
 github.com/lib/pq v1.10.4 h1:SO9z7FRPzA03QhHKJrH5BXA6HU1rS4V2nIVrrNC1iYk=
 github.com/lib/pq v1.10.4/go.mod h1:AlVN5x4E4T544tWzH6hKfbfQvm3HdbOxrmggDNAPY9o=
+github.com/linkedin/goavro v2.1.0+incompatible h1:DV2aUlj2xZiuxQyvag8Dy7zjY69ENjS66bWkSfdpddY=
+github.com/linkedin/goavro v2.1.0+incompatible/go.mod h1:bBCwI2eGYpUI/4820s67MElg9tdeLbINjLjiM2xZFYM=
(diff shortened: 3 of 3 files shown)


fracasula (changes requested):
Hey @ItsSudip, thanks for the changes!
I requested some modifications, feel free to reach out if you need assistance.

Beside the above, I think we should have some tests as well. How are you planning to test the Avro support that you're introducing? Have you seen the integration tests that we have in `kafkamanager_test.go`?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -394,9 +401,48 @@ func prepareMessage(topic, key string, message []byte, timestamp time.Time) clie
...
+
+// This function is used to serialize the binary data according to the AVROSchema.
+// It iterate over the schemas provided by the customer and try to serialize the data.
+// If it ables to serialize the data then it returns the converted data otherwise it throws error.
+// We are using linkedin goavro library to serialize the data. For ref: https://github.com/linkedin/goavro
+
+func serialize(value []byte, AVROSchema []Schema) ([]byte, error) {
+	var errorStr string
+	for i := 0; i < len(AVROSchema); i++ {
+		codec, err := goavro.NewCodec(AVROSchema[i].AVROSchema)
```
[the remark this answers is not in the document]
ItsSudip: Okay, we can move the codec declaration to the `NewProducer` section.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -9,18 +9,23 @@ import (
...
 	rslogger "github.com/rudderlabs/rudder-server/utils/logger"
+	"github.com/tidwall/gjson"
 )
 
 type Opts struct {
 	Timeout time.Duration
 }
 
+// Schema is the AVRO schema required to convert the data to AVRO
+type Schema struct {
```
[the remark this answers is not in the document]
ItsSudip: Yes from config-backend we are getting array of objects.

services/streammanager/kafka/kafkamanager.go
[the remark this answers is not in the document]
fracasula: Global variables are not a good practice most of the time so unless you have a very good reason please try to avoid them. What do you think about passing it around via `producerImpl`?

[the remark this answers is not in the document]
ItsSudip: Yes I have passed it through `producerImpl`

services/streammanager/kafka/kafkamanager.go
```diff
@@ -414,11 +469,17 @@ func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timest
 			pkgLogger.Errorf("batch from topic %s is missing the userId attribute", topic)
 			continue
 		}
-
 		marshalledMsg, err := json.Marshal(message)
 		if err != nil {
 			return nil, err
 		}
+		codec := p.getCodecs()
+		if codec != nil {
```
fracasula, line 477: Instead of checking for not `nil` we can use `len` and cover empty non-nil slices as well:
`if len(codecs) > 0 {`. FYI `len` returns `0` if the slice is `nil`.

fracasula, line 477: Instead of checking for not `nil` we can use `len` and cover empty non-nil slices as well:
`if len(codecs) > 0 {`. FYI `len` returns `0` if the slice is `nil`.

ItsSudip:
> 
> With `1`, do you mean if the user provides an invalid codec (hence that codec is not created in the constructors)?
> 
> Additional questions:
> 
> * don't we want to add Avro support for Confluent Cloud and Azure Event Hubs as well?
> * shouldn't we have a limit on the number of schemas the user sets?
>   
>   * if they were to provide a lot of schemas it could seriously degrade performance given that you don't know what codec to use for a given message, you are left with iterating and trying to encode until you succeed. worst case scenario, with 100 schemas and a batch of 10k messages that are all the same, if the "matching" schema is the last one then we try to encode 10k messages * 100 times = 1,000,000 encoding attempts for 10k messages 🤯
> 
> Maybe you can give me a bit more context so that I can understand how this feature is going to be used?

Yes if the user is providing the wrong schema then the codec will not be created.

* As of now we are only supporting Avro in Kafka due to customer requirements, will add support for others as well if required.
Serializing something to Avro is strongly dependent on the schema. So here we have to provide more than one schema to support different types of event calls. Schema for identify calls will be different for track or page or group calls.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -9,18 +9,23 @@ import (
...
 )
 
 type Opts struct {
 	Timeout time.Duration
 }
 
+// schema is the AVRO schema required to convert the data to AVRO
+type schema struct {
+	avroSchema string
+}
```
mcchran, line 27: why do we need this struct ?
can't we do something like:
`type avroSchema string`
plus the name `schema` is a little bit missleading

services/streammanager/kafka/kafkamanager.go
```diff
@@ -32,6 +37,8 @@ type configuration struct {
 	SaslType      string
 	Username      string
 	Password      string
+	ConvertToAvro bool
+	avroSchema    []schema
```
mcchran, line 41: `avroSchemas []avroSchema`
let's use plural on iterables ... or a notation to denote that this is an iterable

services/streammanager/kafka/kafkamanager.go
```diff
@@ -394,7 +419,38 @@ func prepareMessage(topic, key string, message []byte, timestamp time.Time) clie
...
+	return fmt.Sprintf(" Schema[%d]: ", index+1) + err.Error() + "."
+}
+
+// This function is used to serialize the binary data according to the avroSchema.
+// It iterates over the schemas provided by the customer and tries to serialize the data.
+// If it's able to serialize the data then it returns the converted data otherwise it returns an error.
+// We are using the LinkedIn goavro library for data serialization. Ref: https://github.com/linkedin/goavro
+func serialize(value []byte, codecs []goavro.Codec) ([]byte, error) {
+	var errorStr string
+	for i, codec := range codecs {
```
mcchran, line 433: as @fracasula raised this may end up to some serious performance degradation ...

fracasula:
> we should limit the number of schemas per destination. Is there really a use-case for having more than 1 schema per destination?

That's what I would like to understand as well @atzoum, perhaps the customer can define all types in a single schema and we get rid of the slice? WDYT @ItsSudip?

Multiple types can be defined in a schema like:

```json
{
  "oneOf": [
     { "$ref": "Customer.schema.json" },
     { "$ref": "Product.schema.json" },
     { "$ref": "Order.schema.json" }
  ]
}
```

Ref: https://www.confluent.io/blog/multiple-event-types-in-the-same-kafka-topic/

services/streammanager/kafka/kafkamanager.go
fracasula: We could do it without the struct I think. About the name, I think it is very common to refer to them as schemas.

services/streammanager/kafka/kafkamanager.go
ItsSudip: Implemented.

services/streammanager/kafka/kafkamanager.go
ItsSudip: Removed it and added support of schemaId which will be available with every message and based on the schemaId we will use the codec.

ItsSudip:
> > we should limit the number of schemas per destination. Is there really a use-case for having more than 1 schema per destination?
> 
> That's what I would like to understand as well @atzoum, perhaps the customer can define all types in a single schema and we get rid of the slice? WDYT @ItsSudip?
> 
> Multiple types can be defined in a schema like:
> 
> ```json
> {
>   "oneOf": [
>      { "$ref": "Customer.schema.json" },
>      { "$ref": "Product.schema.json" },
>      { "$ref": "Order.schema.json" }
>   ]
> }
> ```
> 
> Ref: https://www.confluent.io/blog/multiple-event-types-in-the-same-kafka-topic/

I have updated the code according to the discussion in slack. Here is the [link](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1655221091348979) for reference. @fracasula please have a look and let me know if any changes required. ( FYI I will add the testcases once the implementation is approved.)

services/streammanager/kafka/kafkamanager.go
```diff
@@ -281,7 +294,20 @@ func NewProducer(destConfigJSON interface{}, o Opts) (*producerImpl, error) { //
...
+	avroSchemas := destConfig.AvroSchemas
+	var codecs map[string]goavro.Codec
+	if convertToAvro {
+		codecs = make(map[string]goavro.Codec, len(avroSchemas))
+		for _, avroSchema := range avroSchemas {
+			newCodec, err := goavro.NewCodec(avroSchema.Schema)
+			if err != nil {
+				return nil, fmt.Errorf("unable to create codec for schemaId:%+v, with error: %w", avroSchema.SchemaId, err)
+			}
+			codecs[avroSchema.SchemaId] = *newCodec
```
fracasula, line 311: What do we expect the `SchemaId` to be? Is it going to be an `int`? Or a `uuid` or something else? In any case we should at least check that it's not an empty string, right?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -414,11 +456,27 @@ func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timest
...
 		}
-
 		marshalledMsg, err := json.Marshal(message)
 		if err != nil {
 			return nil, err
 		}
+		codecs := p.getCodecs()
+		if len(codecs) > 0 {
+			schemaId, _ := data["schemaId"].(string)
+			fmt.Println(schemaId)
```
fracasula, line 470:
```suggestion
```

services/streammanager/kafka/kafkamanager.go
```diff
@@ -414,11 +456,27 @@ func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timest
...
+			if schemaId == "" {
+				return nil, fmt.Errorf("schemaId is not available for this event")
```
fracasula, line 472: Which "this event"? Can you please include some information in the error that would help people pinpoint the troublesome event?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -414,11 +456,27 @@ func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timest
...
+			}
+			if codec, ok := codecs[schemaId]; ok {
+				marshalledMsg, err = serialize(marshalledMsg, codec)
+				if err != nil {
+					return nil, fmt.Errorf("unable to convert the event with schemaId: %v, with error: %s", schemaId, err)
```
fracasula, line 477: Same here and possible in the rest of this function. Some info to pinpoint the event would be great. If available a `messageId` or otherwise the array position. WDYT?

fracasula (commented):
The approach seems fine @ItsSudip, thanks for the changes :+1: 
I left a few comments. Looking forward to seeing the tests.

services/streammanager/kafka/kafkamanager.go
ItsSudip: Updated the error message with messageId.

services/streammanager/kafka/kafkamanager.go
ItsSudip: Upfated this error message with the index.

services/streammanager/kafka/kafkamanager.go
ItsSudip: It can be any of `int` or `UUID`. As it is customer-specific, I don't think it will create any issues.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -248,6 +248,24 @@ func NewProducer(destConfigJSON interface{}, o Opts) (*producerImpl, error) { //
...
+	var codecs map[string]goavro.Codec
+	if convertToAvro {
+		codecs = make(map[string]goavro.Codec, len(avroSchemas))
+		for i, avroSchema := range avroSchemas {
+			if len(avroSchema.SchemaId) > 0 {
+				newCodec, err := goavro.NewCodec(avroSchema.Schema)
+				if err != nil {
+					return nil, fmt.Errorf("unable to create codec for schemaId:%+v, with error: %w", avroSchema.SchemaId, err)
+				}
+				codecs[avroSchema.SchemaId] = *newCodec
```
fracasula, line 262: How come you're storing it by value here?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -248,6 +248,24 @@ func NewProducer(destConfigJSON interface{}, o Opts) (*producerImpl, error) { //
 		return nil, fmt.Errorf("invalid configuration: %w", err)
 	}
 
+	convertToAvro := destConfig.ConvertToAvro
+	avroSchemas := destConfig.AvroSchemas
+	var codecs map[string]goavro.Codec
+	if convertToAvro {
+		codecs = make(map[string]goavro.Codec, len(avroSchemas))
+		for i, avroSchema := range avroSchemas {
+			if len(avroSchema.SchemaId) > 0 {
```
fracasula, line 257: Optional semplification, you could remove the `if/else` and do it with a single `if`:

```go
if len(avroSchema.SchemaId) == 0 {
    return nil, fmt.Errorf("unable to create codec for schemaId:%+v, with error: %w", avroSchema.SchemaId, err)
}

newCodec, err := goavro.NewCodec(avroSchema.Schema)
// ...
```

services/streammanager/kafka/kafkamanager.go
```diff
@@ -469,14 +474,13 @@ func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timest
 		codecs := p.getCodecs()
 		if len(codecs) > 0 {
 			schemaId, _ := data["schemaId"].(string)
-			fmt.Println(schemaId)
 			if schemaId == "" {
-				return nil, fmt.Errorf("schemaId is not available for this event")
+				return nil, fmt.Errorf("schemaId is not available for the event of index:%d", i)
```
fracasula, line 478:
```suggestion
				return nil, fmt.Errorf("schemaId is not available for the event with index: %d", i)
```

services/streammanager/kafka/kafkamanager.go
```diff
@@ -398,14 +431,30 @@ func prepareMessage(topic, key string, message []byte, timestamp time.Time) clie
 	}
 }
 
-func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timestamp time.Time) (
+// This function is used to serialize the binary data according to the avroSchema.
+// It iterates over the schemas provided by the customer and tries to serialize the data.
+// If it's able to serialize the data then it returns the converted data otherwise it returns an error.
+// We are using the LinkedIn goavro library for data serialization. Ref: https://github.com/linkedin/goavro
+func serialize(value []byte, codec goavro.Codec) ([]byte, error) {
```
fracasula, line 438:
```suggestion
func serializeAvroMessage(value []byte, codec goavro.Codec) ([]byte, error) {
```

services/streammanager/kafka/kafkamanager.go
```diff
@@ -398,14 +431,30 @@ func prepareMessage(topic, key string, message []byte, timestamp time.Time) clie
...
+
+func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timestamp time.Time, p producer) (
 	[]client.Message, error,
 ) {
 	start := now()
 	defer func() { kafkaStats.prepareBatchTime.SendTiming(since(start)) }()
 
 	var messages []client.Message
-	for _, data := range batch {
+	for i, data := range batch {
```
fracasula, line 457: Given the existing logic of this function, please check with product if you should return in this `for` loop when there are codec errors instead of doing `continue`. As you can see, if there is a missing `message` or a missing `userId`, we log and carry on so that we can process the rest of the batch. On the other hand, when you come across a codec error you just return and ignore the rest of the messages in the batch.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -418,11 +467,26 @@ func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timest
...
+				return nil, fmt.Errorf("schemaId is not available for the event of index:%d", i)
+			}
+			if codec, ok := codecs[schemaId]; ok {
+				marshalledMsg, err = serialize(marshalledMsg, codec)
+				if err != nil {
+					return nil, fmt.Errorf("unable to serialize the event of index: %d, with error: %s", i, err)
+				}
+			} else {
+				return nil, fmt.Errorf("unable to find schema with schemaId: %v", schemaId)
+			}
```
fracasula, line 487: This can also be simplified by removing the `if/else`. It's doable with a single `if`.

services/streammanager/kafka/kafkamanager.go
ItsSudip: Here the two error messages are different though I can omit else.

services/streammanager/kafka/kafkamanager.go
ItsSudip: Anyhow we are discarding the entire batch for Marshall error in line no.471. @arnab-p WDYT?

services/streammanager/kafka/kafkamanager.go
ItsSudip: I didn't get your question.

services/streammanager/kafka/kafkamanager.go
ItsSudip: Here two error messages are different. But I have removed the `else`.

services/streammanager/kafka/kafkamanager.go
ItsSudip: Anyhow, we are discarding the entire batch for any marshal error. @arnab-p WDYT?

services/streammanager/kafka/kafkamanager.go
ItsSudip: Updated.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -516,6 +579,22 @@ func sendMessage(ctx context.Context, jsonData json.RawMessage, p producer, topi
...
+			return makeErrorResponse(fmt.Errorf("schemaId is not available for event with messageId: %s", messageId))
+		}
+		if codec, ok := codecs[schemaId]; ok {
+			value, err = serialize(value, codec)
+			if err != nil {
+				return makeErrorResponse(fmt.Errorf("unable to serialize event with messageId: %s, with error %s", messageId, err))
+			}
+		} else {
+			return makeErrorResponse(fmt.Errorf("unable to find schema with schemaId: %v", schemaId))
+		}
```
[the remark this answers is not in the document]
ItsSudip: Updated.

services/streammanager/kafka/kafkamanager.go
ItsSudip: So you are asking me to declare `codecs` as a map of `goavro.Codec` type pointer instead of values? I don't know how to do that.

fracasula:
```go
codecs = make(map[string]*goavro.Codec, len(avroSchemas))
```

PR #2049, opened 2022-06-13
test(destination): transformer proxy test for correctness

sanpj2292:
# Description

https://www.notion.so/rudderstacks/Transformer-Proxy-Correctness-Test-Plan-4170d336e019459aad9f6c3fadfcb0be?d=8e13d5d47e884fd7988c840b7ee762ca#9953ba11dda9470f91a32680260786e5

## Notion Ticket

[Transformer Proxy Issues](https://www.notion.so/rudderstacks/Transformer-Proxy-network-module-issues-5f32e23f7d13410bbaa3d32bebac5207)

## Security


router/transformer/transformer.go
```diff
@@ -322,6 +325,223 @@ func (trans *HandleT) makeHTTPRequest(ctx context.Context, url string, payload [
...
+			case "FORM":
+				formValues := url.Values{}
+				for key, val := range bodyValue {
+					formValues.Set(key, fmt.Sprint(val)) // transformer ensures top level string values, still val.(string) would be restrictive
+				}
+				payload = strings.NewReader(formValues.Encode())
+			default:
+				return nil, fmt.Errorf("bodyFormat: %s is not supported", bodyFormat)
+			}
+		}
```
saikumarrs, line 396: Why are we re-implementing all of this logic? We were supposed to reuse the existing logic/module in the router that is already doing it so that we can compare its output with the transformer.

saikumarrs: https://github.com/rudderlabs/rudder-server/blob/a37ccd97d6bfd6dcf7369ce6afaa7055c2a870bd/router/network.go#L62

saikumarrs: The idea is to make the proxy request with the prepared destination request and input the delivery payload without sending it to the actual destination.

router/transformer/transformer.go
```diff
@@ -322,6 +325,223 @@ func (trans *HandleT) makeHTTPRequest(ctx context.Context, url string, payload [
...
+	Endpoint string `json:"endpoint"`
+	// The headers for the event delivery request
+	Headers interface{} `json:"headers"`
+}
+
+type ProxyTestReqBody struct {
+	// The event delivery request payload formed by Router
+	RouterDeliveryPayload RouterDelPayload
+	// The payload which will be used by transformer to form event delivery request payload
+	ProxyRequestPayload integrations.PostParametersT
```
saikumarrs, line 495: Can we name them a bit better? I've suggested the names I'm expecting in the transformer but feel free to change them.
```suggestion
	// The event delivery request payload formed by Router
	destinationRequestPayload RouterDelPayload
	// The payload which will be used by the transformer to form the event delivery request payload
	deliveryPayload integrations.PostParametersT
```

router/transformer/transformer.go
sanpj2292: > Why are we re-implementing all of this logic? We were supposed to reuse the existing logic/module in the router that is already doing it so that we can compare its output with the transformer.

1. The internal logic is same as what was there in original(`network.go`)
2. Since it is tightly bound to `SendPost` method, relocated the method to another file for clarity
3. Basically when `transformerProxy` is _enabled_ it actually sends request to `proxyTest` endpoint only & the request will not be sent to the destination at all
4. The destination request( of `http.Request` type) cannot be marshalled directly to easily send it as part of `proxyTest` request, so we had to send the data in a different data-structure so that marshalling takes place &

saikumarrs: Let's refactor the original code to accommodate our needs. Having duplicate logic even if it is identical would not be ideal.

How about, rather than activating the proxy path, we can use the same router <-> destination path but instead of sending it to the destination, we'd send it to the proxy test route? We can control it with the same config variable if needed.

I feel this is ideal cause we have both the resources (router input payload and destination request) that we can package and send appropriately to the transformer.

router/transformer/transformer.go
sanpj2292: We can add json definitions for them like how they are in `RouterDelPayload`
Do you think they have to be interchanged ?
`RouterDeliveryPayload` -> ``json:"deliveryPayload"``
`ProxyRequestPayload` -> ``json:"destinationRequestPayload"``

router/transformer/transformer.go
sanpj2292: > Let's refactor the original code to accommodate our needs. Having duplicate logic even if it is identical would not be ideal.

The thing is, why should the original code be disturbed to accommodate something temporary ?
I don't understand why this is not a clean approach to get what we need. Can you explain your perspective around this ?

router/transformer/transformer.go
saikumarrs: I named them as per the below notion:

1. Request payload send to the destination - `destinationRequestPayload`
2. Input data which router operates on - `deliveryPayload`. This sounds a bit generic, maybe we can name it something like `transformerProcResponse`?

The combined DS can be called `ProxyTestRequestPayload`.

router/transformer/transformer.go
saikumarrs: Though these changes sound temporary now, we can reuse them for integration testing later. Having duplicate code is always messy.
The approach I'm suggesting is along the lines of the transformer changes we made (decouple request payload preparation from the transport logic).

router/transformer/transformer.go
sanpj2292: Done

router/router.go
```diff
@@ -713,35 +713,13 @@ func (worker *workerT) processDestinationJobs() {
...
-									rdl_time := time.Now()
-									resp := worker.rt.netHandle.SendPost(sendCtx, val)
-									respStatusCode, respBodyTemp, respContentType = resp.StatusCode, string(resp.ResponseBody), resp.ResponseContentType
-									// stat end
-									worker.routerDeliveryLatencyStat.SendTiming(time.Since(rdl_time))
-								}
-								// transformer proxy end
+								rdl_time := time.Now()
+								jobId := destinationJob.JobMetadataArray[0].JobID
+								resp := worker.rt.netHandle.SendPost(sendCtx, val, worker.rt.destName, jobId, worker.rt.transformerProxy)
```
saikumarrs, line 718: I believe there are no changes required to this method. 
When the proxy test configuration is enabled, rather than enabling this path, can we enable the destination delivery path itself but inform it of the configuration value?

PR #2053, opened 2022-06-14
feat: integration tags

fracasula:
# Description

I'm adding the possibility to run unit tests without the burden of bootstrapping containers that are setup just because in the same package we have a `TestMain` that is creating them.

Here I'm proposing to differentiate between the two by using build tags.

Additionally, I added a `slow` CLI argument to for `docker_test.go`.

## Example

At the moment we cannot run [these](https://github.com/rudderlabs/rudder-server/blob/10913e65702f26e5e4ab591d072a15dd4ea0e64d/app/cluster/dynamic_test.go) unit tests without also executing the `TestMain` from [here](https://github.com/rudderlabs/rudder-server/blob/10913e65702f26e5e4ab591d072a15dd4ea0e64d/app/cluster/integration_test.go#L50).

## Additional notes

If you like the approach I can add the tag to all tests that are using Docker.

## Security

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,7 +26,7 @@ jobs:
     - run: go version
     - run: go mod download # Not required, used to segregate module download vs test times
     - name: Integration test
-      run: go test -v ./docker_test.go -integration -count 1
+      run: go test -v ./docker_test.go -slow -tags=integration -count 1
       env:
         RSERVER_ENABLE_MULTITENANCY: false
         BIGQUERY_INTEGRATION_TEST_USER_CRED: ${{ secrets.BIGQUERY_INTEGRATION_TEST_USER_CRED }}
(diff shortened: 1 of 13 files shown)


mcchran (approved):
LGTM.

atzoum (approved):
This is a good reason why we should look for alternatives to `TestMain`.
It is not just the fact that in the same package you might have tests that don't depend on the containers that are being started in `TestMain`.
There is also another problematic scenario, having tests in the same package, each requiring different containers to be started.

A proposed approach to this problem could be:
- when we are using Ginkgo we can use `BeforeAll` & `AfterAll` in the appropriate context.
- when we are using the standard go testing library, we should move towards creating containers within the tests themselves and use subtests.

fracasula:
> A proposed approach to this problem could be:
> 
> * when we are using ginkgo we can use `BeforeAll` & `AfterAll` in the appropriate `Context`.
> * when we are using the standard go testing library, we should move towards creating containers within the tests themselves and use subtests.

You are making good points @atzoum and that is indeed what I like to do with the standard Go library. I tend to create a test with sub-tests where the setup is there with very little assertions. On the other hand, when feasible I also use build tags like I'm proposing here if it doesn't make things too complex (in the `/app/cluster` case for example). In the `jobsdb` package the files are intertwined in a weird way and build tags there are not optimal in my opinion so we should make a better usage of either subtests or gingko contexts :+1: 

Personally I'd like to add one more reason to avoid `TestMain` and is that we don't have a `t *testing.T` available which is useful for lots of things (e.g. cleanups, logging, defining helpers, erroring instead of panicking, etc...).

PR #2056, opened 2022-06-14
chore: make lint

fracasula:
# Description

Adding `make` recipe to run `golangci-lint` locally. I'm using Docker so that we don't have to do weird `go install`s or use the `.ONESHELL` in the Makefile given how certain GNU systems handle shells in Makefiles.

## Security


lvrach:
I would rename the PR title to a `chore: `

atzoum (approved):

lvrach (approved):

PR #2058, opened 2022-06-15
chore: introduce pipeline_processed_events metric in router and batch router

atzoum:
# Description

While events are being processed in router and batch router, the actual outcome of each event cannot be guaranteed, since it might fail, succeed, be aborted, drained etc.
At the same time, we need a metric that indicates whether the system's modules (router & batch router) are processing events, regardless of what their actual outcome is.
Such a metric is especially useful for detecting stuck processing pipelines in the system, i.e. scenarios where we do have outstanding events in our queues, but the relevant system modules don't appear to be processing anything.

Unfortunately, the current alerts that detect such stuck processing pipelines rely on partial measurements, leading to false positives, more specifically:
- **router** - we are currently using `router_response_counts` which doesn't capture aborted jobs
- **batch router** - we are currently using `event_status module=batch_router` which doesn't capture drained jobs and async destinations

After introducing this new metric and releasing the changes, all relevant alerts will be adapted:
- `stuck-processing-pipeline-rt`
- `stuck-processing-pipeline-batch-rt`

## Notion Ticket

[Improved metric for capturing total router & batch router output to be used in stuck-processing-pipeline alert
](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=d2de5836f6ff4ceabe2a619e6d9113a9)

## Security

### DIFF
--- a/router/batchrouter/batchrouter.go
+++ b/router/batchrouter/batchrouter.go
@@ -267,8 +267,8 @@ func sendDestStatusStats(batchDestination *DestinationT, jobStateCounts map[stri
 	}
 
 	for jobState, countByAttemptMap := range jobStateCounts {
+		tags["job_state"] = jobState
 		for attempt, count := range countByAttemptMap {
-			tags["job_state"] = jobState
 			tags["attempt_number"] = attempt
 			if count > 0 {
 				stats.NewTaggedStat("event_status", stats.CountType, tags).Count(count)
@@ -437,7 +437,7 @@ func (brt *HandleT) pollAsyncStatus(ctx context.Context) {
 										if err != nil {
 											panic(err)
 										}
-
+										brt.updateProcessedEventsMetrics(statusList)
 										continue
 									}
 									for _, job := range importingList {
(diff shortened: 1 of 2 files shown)


cisse21 (approved):

router/batchrouter/batchrouter.go
```diff
@@ -2374,3 +2378,18 @@ func (brt *HandleT) updateRudderSourcesStats(ctx context.Context, tx jobsdb.Upda
...
+	countPerState := map[string]int{}
+	for i := range statusList {
+		state := statusList[i].JobState
+		countPerState[state]++
+	}
+	for state, count := range countPerState {
+		stats.NewTaggedStat(`pipeline_processed_events`, stats.CountType, stats.Tags{
+			"module":   "batch_router",
+			"destType": brt.destType,
+			"state":    state,
```
lvrach, line 2392: [question] are we only capturing terminal states here?

lvrach: I understand we can capture non-terminal states as well. What if we introduce a boolean flag `terminal`.

This will give us the option to measure flow in/ out without the noise of retries. If we want to know that the system is working, then we can filter without that flag

lvrach (approved):
Approve, with non-blocking comment

router/batchrouter/batchrouter.go
atzoum: we capture any state, if we want to measure terminal states, it is possible, we can add a state filter in the query

PR #2059, opened 2022-06-15
chore: removing TestMain from docker_test.go

fracasula:
# Description

I need to write more integration tests that involve running `main` and unfortunately that is not possible from another package since `main` packages are not importable:

```go
package something

import (
	"context"
	"testing"

	main "github.com/rudderlabs/rudder-server"
)

func TestSomethingInMain(t *testing.T) {
	main.Run(context.TODO())
}
```

Produces:

> import "github.com/rudderlabs/rudder-server" is a program, not an importable package

Having said that and considering that I need to write additional tests that setup different containers, I had to get rid of `TestMain`.

## Additional changes

* made the `Cleanup` struct similar to `testing.T.Cleanup` to simplify some tests that don't really need the `Cleanup` struct and can use `testing.T`
* removed arguments in favour of environment variables (see `SLOW`) due to the fact that the `flag` package complains if we have unknown arguments which makes it compatible with `TestMain` but not with regular tests (see below)

```go
package something

import (
	"flag"
	"testing"
)

var hold bool

func TestSomething(t *testing.T) {
	flag.BoolVar(&hold, "hold", false, "hold environment clean-up after test execution until Ctrl+C is provided")
	flag.Parse()
}
```

Produces:

```
$ go test . -v -count 1 -hold                                                                                                                                                                                             
flag provided but not defined: -hold
FAIL	github.com/rudderlabs/rudder-server/testhelper/something	0.001s
FAIL
```

## Notion Ticket

[< Notion Link >](https://www.notion.so/rudderstacks/Test-case-for-rudder-server-c3212a02955b416a81e1e962ba5fa2bf#66ef8d48b6d64bc08ca746cdba3edc09)

## Security

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,7 +26,7 @@ jobs:
     - run: go version
     - run: go mod download # Not required, used to segregate module download vs test times
     - name: Integration test
-      run: go test -v ./docker_test.go -slow -tags=integration -count 1
+      run: SLOW=1 go test -v ./docker_test.go -tags=integration -count 1
       env:
         RSERVER_ENABLE_MULTITENANCY: false
         BIGQUERY_INTEGRATION_TEST_USER_CRED: ${{ secrets.BIGQUERY_INTEGRATION_TEST_USER_CRED }}
@@ -62,7 +62,7 @@ jobs:
     - run: go mod download # Not required, used to segregate module download vs test times
     - run: make enterprise-init
     - name: Integration test
-      run: go test -v ./docker_test.go -slow -tags=integration -count 1
+      run: SLOW=1 go test -v ./docker_test.go -tags=integration -count 1
       env:
         RSERVER_ENABLE_MULTITENANCY: ${{ matrix.MULTITENANCY }}
         BIGQUERY_INTEGRATION_TEST_USER_CRED: ${{ secrets.BIGQUERY_INTEGRATION_TEST_USER_CRED }}
--- a/docker_test.go
+++ b/docker_test.go
@@ -8,8 +8,6 @@
 package main_test
 
 import (
-	"bufio"
-	"bytes"
 	"context"
 	"database/sql"
 	b64 "encoding/base64"
@@ -19,14 +17,11 @@ import (
 	"log"
 	"math/rand"
 	"net/http"
-	"net/http/httptest"
-	"net/http/httputil"
 	"os"
 	"os/signal"
 	"runtime"
 	"strconv"
 	"strings"
-	"sync"
 	"syscall"
 	"testing"
 	"text/template"
@@ -46,6 +41,7 @@ import (
 	"github.com/rudderlabs/rudder-server/services/streammanager/kafka/client/testutil"
 	"github.com/rudderlabs/rudder-server/testhelper/destination"
 	wht "github.com/rudderlabs/rudder-server/testhelper/warehouse"
+	whUtil "github.com/rudderlabs/rudder-server/testhelper/webhook"
 	"github.com/rudderlabs/rudder-server/utils/logger"
 	"github.com/rudderlabs/rudder-server/utils/types/deployment"
 	bq "github.com/rudderlabs/rudder-server/warehouse/bigquery"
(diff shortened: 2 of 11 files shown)


.github/workflows/tests.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
     - run: go version
     - run: go mod download # Not required, used to segregate module download vs test times
     - name: Integration test
-      run: go test -v ./docker_test.go -slow -tags=integration -count 1
+      run: SLOW=1 go test -v ./docker_test.go -tags=integration -count 1
```
atzoum, line 29: It might be better to use the opposite semantics, i.e. without you overriding any environment variable all tests should run, no matter if they are slow, integration, etc.

A user can then opt to exclude some tests by setting the relevant env variable overrides.

The only reason why the above solution might be better, is that it eliminates the risk of misconfiguration (forgetting to set an env variable) in our ci workflows, resulting in some tests being skipped inadvertently.

docker_test.go
```diff
@@ -146,270 +88,566 @@ type Event struct {
...
-		case <-ctx.Done():
-			return
-		case <-timeout:
-			log.Panicf("application was not ready after %s, for the end point: %s, caller: %s\n", atMost, endpoint,
-				caller)
-		case <-probe.C:
-			resp, err := http.Get(endpoint)
-			if err != nil {
-				continue
+	t.Run("webhook", func(t *testing.T) {
```
atzoum, line 116: for some of these sub-tests we could even turn on the parallel flag

atzoum (approved):

docker_test.go
fracasula: Yes although there is still no perfect isolation, I'd like to achieve that one first. Despite decoupling the sending of the events I'm still not able to have the "kafka" test pass alone for example. once I'm able to achieve that I'll try to turn the parallelism on.

One additional note on parallelism is that I had that on my kafka tests and I was able to run them in less than 30s on my desktop, but it was making them extremely flaky on CI so I had to remove it. It seems that with the 2 cores of the default CI env, too much parallelism might cause timeouts. So even though I'm not against it might require a bit of tuning to get it right.

.github/workflows/tests.yaml
fracasula: I do like the idea, we could have `SLOW` as `true` by default and then we set it to `false` in `make test`.

docker_test.go
atzoum: We have started containers once, even if the tests are still coupled, i.e. operate on the same source, at least using parallelism we can have them all fail/pass faster.

testhelper/cleanup.go
```diff
@@ -1,20 +1,22 @@
...
-	fns []func() error
+	fns []func()
 }
 
-func (c *Cleanup) Defer(fn func() error) {
+func (*Cleanup) Log(a ...interface{}) {
+	fmt.Println(a...)
+}
+
+func (c *Cleanup) Cleanup(fn func()) {
```
mcchran, line 13: Suggestion:

`func (c * Cleanup) Push (fn func()) {`

testhelper/destination/kafka.go
```diff
@@ -119,7 +119,7 @@ func (k *KafkaResource) Destroy() error {
 	return g.Wait()
 }
 
-func SetupKafka(pool *dockertest.Pool, d deferer, opts ...Option) (*KafkaResource, error) {
+func SetupKafka(pool *dockertest.Pool, cln cleaner, opts ...Option) (*KafkaResource, error) {
```
mcchran, line 122: here we have a "Public" signature to require a private type `cleaner`, right?

testhelper/cleanup.go
fracasula: I named it `Cleanup` for a specific reason, see PR description:

Screenshot from 2022-06-17 13-38-37

This way, if we are in a test we can just pass `t *testing.T` instead of using the `Cleanup` struct. Ideally we should get to a point where we're not using the struct and we can remove it but we're still discussing it as it's part of the "avoid TestMain" discussion.

testhelper/destination/kafka.go
```diff
@@ -61,8 +60,9 @@ type logger interface {
 	Log(...interface{})
 }
 
-type deferer interface {
-	Defer(func() error)
+type cleaner interface {
```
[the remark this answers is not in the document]
fracasula: `cleanuper` is not an English word, `cleaner` is. I think I'll keep it as `cleaner` if you don't mind.

testhelper/destination/kafka.go
fracasula: We don't require a private type, interfaces are just a contract. By not exporting the interface we're just saying "this is what we need but we offer no implementation so you'll have to choose one of your own". Meaning you have to use either the `*Cleanup` struct or `*testing.T` (or something that you can create yourself, as long as it has those two methods).

lvrach (approved):

PR #2061, opened 2022-06-16
fix: properly start a new badgerdb transaction on ErrTxnTooBig

atzoum:
# Description

Using `badgerDB.NewTransaction` instead of `badgerDB.Update` which results in a **_Trying to commit a discarded txn_** error whenever a `ErrTxnTooBig` is returned by the `SetEntry` method.

## Notion Ticket

[BadgerDB misuse](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=edc933e2190143119e369fedcdb244eb)

## Security

### DIFF
--- a/services/dedup/dedup.go
+++ b/services/dedup/dedup.go
@@ -18,7 +18,7 @@ import (
 
 type DedupI interface {
 	FindDuplicates(messageIDs []string, allMessageIDsSet map[string]struct{}) (duplicateIndexes []int)
-	MarkProcessed(messageIDs []string)
+	MarkProcessed(messageIDs []string) error
 	PrintHistogram()
 	Close()
 }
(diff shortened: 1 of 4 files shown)


lvrach (approved):

services/dedup/dedup.go
```diff
@@ -169,27 +169,27 @@ func (d *DedupHandleT) gcBadgerDB() {
...
-			e := badger.NewEntry([]byte(messageID), nil).WithTTL(*d.window)
-			if err := txn.SetEntry(e); err == badger.ErrTxnTooBig {
-				_ = txn.Commit()
-				txn = d.badgerDB.NewTransaction(true)
-				_ = txn.SetEntry(e)
+func (d *DedupHandleT) writeToBadger(messageIDs []string) error {
+	txn := d.badgerDB.NewTransaction(true)
+	for _, messageID := range messageIDs {
+		e := badger.NewEntry([]byte(messageID), nil).WithTTL(*d.window)
+		if err := txn.SetEntry(e); err == badger.ErrTxnTooBig {
```
lvrach, line 176: We should also check for err != nil. 

Unfortunately badger db examples don't contain proper error handling and this one is easy to miss

lvrach (changes requested):

services/dedup/dedup.go
atzoum: yes indeed, missed that

PR #2065, opened 2022-06-16
fix: update failedJobIDMap after draining a job in router

atzoum:
# Description

After draining a job from router's jobsdb we also need to remove the job's userID from the failedJobIDMap, if present.

## Router pipeline
The following diagram captures the current router pipeline design (not including changes introduced by this fix)

router drawio

## Notion Ticket

[Drained jobs should be properly removed from failedJobIDMap
](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=94188b8363e04b708996d351bec622e1)
## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -2138,17 +2138,12 @@ func (rt *HandleT) readAndProcess() int {
 				partition := rt.getWorkerPartition(drainedUserID)
 				drainWorker := rt.workers[partition]
 
-				drainWorker.failedJobIDMutex.RLock()
+				drainWorker.failedJobIDMutex.Lock()
 				lastJobID, ok := drainWorker.failedJobIDMap[drainedUserID]
-				drainWorker.failedJobIDMutex.RUnlock()
-
 				if ok && lastJobID == drainedJob.JobID {
-
-					drainWorker.failedJobIDMutex.Lock()
 					delete(drainWorker.failedJobIDMap, drainedUserID)
-					drainWorker.failedJobIDMutex.Unlock()
-
 				}
+				drainWorker.failedJobIDMutex.Unlock()
 			}
 		}
 
--- a/router/router.go
+++ b/router/router.go
@@ -1491,7 +1491,7 @@ func (worker *workerT) canBackoff(job *jobsdb.JobT, userID string) (shouldBackof
 }
 
 func (rt *HandleT) getWorkerPartition(userID string) int {
-	return int(math.Abs(float64(misc.GetHash(userID) % rt.noOfWorkers)))
+	return misc.GetHash(userID) % rt.noOfWorkers
 }
 
 func (rt *HandleT) shouldThrottle(destID, userID string, throttledAtTime time.Time) (canBeThrottled bool) {
--- a/router/router.go
+++ b/router/router.go
@@ -92,7 +92,6 @@ type HandleT struct {
 	failedEventsListMutex                  sync.RWMutex
 	failedEventsList                       *list.List
 	failedEventsChan                       chan jobsdb.JobStatusT
-	isEnabled                              bool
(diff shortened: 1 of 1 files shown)


router/router.go
```diff
@@ -1491,6 +1490,10 @@ func (worker *workerT) canBackoff(job *jobsdb.JobT, userID string) (shouldBackof
 	return false
 }
 
+func (rt *HandleT) getWorkerPartition(userID string) int {
+	return int(math.Abs(float64(misc.GetHash(userID) % rt.noOfWorkers)))
```
lvrach, line 1494: [question] why do we need abs here? Can `misc.GetHash(userID) % rt.noOfWorkers` be possibly negative ?

atzoum: you are correct, the hash actually return an unsigned integer...

router/router.go
```diff
@@ -2136,6 +2130,28 @@ func (rt *HandleT) readAndProcess() int {
...
+				lastJobID, ok := drainWorker.failedJobIDMap[drainedUserID]
+				drainWorker.failedJobIDMutex.RUnlock()
+
+				if ok && lastJobID == drainedJob.JobID {
+
+					drainWorker.failedJobIDMutex.Lock()
+					delete(drainWorker.failedJobIDMap, drainedUserID)
+					drainWorker.failedJobIDMutex.Unlock()
+
+				}
```
lvrach, line 2151: There is no guarantee that between:
```go
drainWorker.failedJobIDMutex.RUnlock()
```
and
```go
drainWorker.failedJobIDMutex.Lock()
```

A different goroutine can modify the content of `drainWorker.failedJobIDMap[drainedUserID]` 

Maybe it is prevented by another mechanism and/or the probability is small.

However, given that drains are infrequent and the duration of the lock very small, I suggest taking the safest way here:

```suggestion
			   drainWorker.failedJobIDMutex.Lock()
				lastJobID, ok := drainWorker.failedJobIDMap[drainedUserID]

				if ok && lastJobID == drainedJob.JobID {
					delete(drainWorker.failedJobIDMap, drainedUserID)
				}
			    drainWorker.failedJobIDMutex.Unlock()

```

lvrach:
How we could possibly test this? 🤔

router/router.go
atzoum: > Maybe it is prevented by another mechanism and/or the probability is small.

Yes, by design, `readAndProcess` is the only one removing `userID`s from the `failedJobIDMap` and it does so under the assumption that there is no other outstanding job already in the pipeline for the same `userID`, otherwise ordering guarantees would be violated.

Nevertheless, the performance penalty of what you are proposing is minimal and promotes the correct transitioning pattern from a read lock to a write lock!

lvrach (approved):

fracasula (approved):
> How we could possibly test this? :thinking: 

The `commitStatusList` is such a monolithic function, perhaps if we split it up a bit it would be easier to at least build a unit test. On the other hand, integration wise, should be doable although time consuming?

atzoum:
> > How we could possibly test this? 🤔
> 
> The `commitStatusList` is such a monolithic function, perhaps if we split it up a bit it would be easier to at least build a unit test. On the other hand, integration wise, should be doable although time consuming?

It is not a testable piece of code, I have manually verified it but my steps could be automated by an e2e scenario: 

1. Set a small job retention, e.g. `1min`.
2. Setup a source with a failing destination
3. Send 1 event for `userID1`
4. Wait `30sec`, send another event for `userID1`
5. Wait until both events are aborted in the database
6. Verify that there are failed attempts recorded for the 2nd event

PR #2069, opened 2022-06-20
chore: update readme and docker compose

lvrach:
# Description

* Replace AWS build badges with GitHub actions
* Change open-source image tag from `1-alpine` to `latest`
* Tag release images with `1-alpine` for existing references

## Notion Ticket

[Minor-OSS-release-improvements](https://www.notion.so/rudderstacks/Minor-OSS-release-improvements-a203a6e2087d4b7d96a15fe11fb7db88) 

## Security

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -26,6 +26,7 @@ jobs:
             name=rudderlabs/rudder-server,enable=${{ github.ref == format('refs/heads/{0}', 'master') || github.event_name == 'release' }}
           tags: |
             type=ref,event=branch
+            type=raw,value=1-alpine,enable=${{ github.ref == format('refs/heads/{0}', 'master') || github.event_name == 'release' }}
             type=raw,value=latest,enable={{is_default_branch}}
             type=raw,value=${{ github.head_ref }}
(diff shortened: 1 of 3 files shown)


cisse21:
We would need a PR [here](https://github.com/rudderlabs/rudderstack-helm/blob/master/values.yaml#L34) too

cisse21:
Also, should we update our documentation to ensure that we let the users know they will always be pointed to the latest version somewhere [here](https://www.rudderstack.com/docs/rudderstack-open-source/installing-and-setting-up-rudderstack/kubernetes/#rudderstack-hosted-control-plane)?

lvrach:
> Also, should we update our documentation to ensure that we let the users know they will always be pointed to the latest version somew[here](https://www.rudderstack.com/docs/rudderstack-open-source/installing-and-setting-up-rudderstack/kubernetes/#rudderstack-hosted-control-plane) here?

`latest` in this case, is the latest release - not the latest master commit. It is equivalent to `1-alpine`. As long as we are on `v0.*.*`,  there is no point to advertise other options.

atzoum (approved):

kuldeep0020 (approved):

PR #2078, opened 2022-06-20
test: multi-tenant gateway integration tests

fracasula:
# Description

Here I'm testing the Gateway in multi-tenant mode and:
1. I'm triggering a workspace change via ETCD. Everything works but the service is not healthy until we hit it with a valid workspace change. I've therefore created [this task](https://www.notion.so/rudderstacks/Review-health-checkpoint-probes-ec33b45c1b7541f3bf802f3276667920) to better handle the situation.
2. I send an event and then I check if it ends up in the jobs table (i.e. `gw_jobs_1`)
3. I put the Gateway in `DEGRADED` mode to check if it still receives events. It does.
4. I trigger a workspace change with an invalid configuration. This before wasn't doing anything, it was just logging the error and then it kept on polling with the wrong configuration. RS didn't even send an ack back, leaving the scheduler in a bad situation. The way it works now is that it stops polling because there is no point (since the configuration supplied is not valid). It also acks the message letting the scheduler know that something bad happened. In the future this means that the scheduler can orchestrate a revert. To avoid having the RS stuck in a crash loop (due to the bad configuration), we're doing changes to the scheduler so that on "ack error" it can delete the "bad" key from ETCD so that the RS can resume working.

## Changes introduced as a result of the new tests
* backend config now returns an error on `WaitForConfig` if the configuration is invalid. before my changes it would keep polling and returning the same error over and over. this also did not allow the configuration error to be properly propagated.
* etcd support for acking a workspace change as `ERROR` instead of `RELOADED`
* dynamic clustering can now acknowledge a "workspace change" event via ETCD if the configuration is invalid
* multitenant backend config now complains if all workspaces IDs provided do not lead to a call after trimming/validating
* context propagation in more parts of the backend config
* added new HTTP `NewRequestWithContext` method to allow for requests cancelations
* minor improvements to some tests
* moved some backend config types into dedicated file to increase readability (i.e. `types.go`)
* new faster version of `rand.String` to remove duplicated code in tests
* new ETCD test helper to spawn up an ETCD container
* more minor test helpers
* make multi-tenant backend-config to fetch the configuration for all workspaces like in hosted

## What's coming in later PRs
* make server-mode changes fault tolerant so that we can pass from normal->normal or degraded->degraded with a simple warning but without having RS in a crash loop
* add ackWithError to server-mode changes

# WARNING

~**DO NOT MERGE** until [this PR](https://github.com/rudderlabs/rudder-server/pull/2131) is merged first because I need to make [this changes](https://github.com/rudderlabs/rudder-server-enterprise/commit/1e7ceb245a4ccdc03c8c095cb8205dee9174e18a) in the Enterprise code. Hence do not mind the CI complaining about the Enterprise commit.~

## Notion Ticket

[< Notion Link >](https://www.notion.so/rudderstacks/Integration-test-docker_test-to-test-rudder-server-with-etcd-237ade605e704593a2dbc7624aaf3fae)

## Security

### DIFF
--- a/mt_gateway_test.go
+++ b/mt_gateway_test.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/ory/dockertest/v3"
-	"github.com/phayes/freeport"
 	"github.com/stretchr/testify/require"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"golang.org/x/sync/errgroup"
@@ -88,9 +87,9 @@ func TestMultiTenantGateway(t *testing.T) {
 	t.Logf("BackendConfig server listening on: %s", backendConfigSrv.URL)
 	t.Cleanup(backendConfigSrv.Close)
 
-	httpPort, err := freeport.GetFreePort()
+	httpPort, err := testhelper.GetFreePort()
 	require.NoError(t, err)
-	httpAdminPort, err := freeport.GetFreePort()
+	httpAdminPort, err := testhelper.GetFreePort()
 	require.NoError(t, err)
 
 	rudderTmpDir, err := os.MkdirTemp("", "rudder_server_*_test")
--- a/testhelper/freeport.go
+++ b/testhelper/freeport.go
@@ -1,0 +1,33 @@
+package testhelper
+
+import (
+	"sync"
+
+	"github.com/phayes/freeport"
+)
+
+var (
+	usedPorts   map[int]struct{}
+	usedPortsMu sync.Mutex
+)
+
+func GetFreePort() (int, error) {
+	for {
+		port, err := freeport.GetFreePort()
+		if err != nil {
+			return 0, err
+		}
+
+		usedPortsMu.Lock()
+		if usedPorts == nil {
+			usedPorts = make(map[int]struct{})
+		}
+		if _, used := usedPorts[port]; used {
+			usedPortsMu.Unlock()
+			continue
+		}
+		usedPorts[port] = struct{}{}
+		usedPortsMu.Unlock()
+		return port, nil
+	}
+}


testhelper/workspaceConfig/workspaceConfig.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func CreateTempFile(t *testing.T, templatePath string, values map[string]string) string {
+	t.Helper()
+	tpl, err := template.ParseFiles(templatePath)
+	require.NoError(t, err)
+
+	tmpFile, err := os.CreateTemp("", "workspaceConfig.*.json")
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = tmpFile.Close() })
```
atzoum, line 18: the file could be closed earlier, after this function completes with a defer statement, correct?

.enterprise-commit
```diff
@@ -1 +1 @@
-174dcd0babaf1a1d4fb052fe7cbb378b5bcf0037
+1e7ceb245a4ccdc03c8c095cb8205dee9174e18a
```
fracasula, line 1: As per PR description, this change will be gone once [this PR](https://github.com/rudderlabs/rudder-server/pull/2131) is merged first. Then I'll apply [this change](https://github.com/rudderlabs/rudder-server-enterprise/commit/1e7ceb245a4ccdc03c8c095cb8205dee9174e18a) here. Hence do not mind the CI complaining about the Enterprise commit.

testhelper/etcd/etcd.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+type Resource struct {
+	Client *etcd.Client
+	Hosts  []string
+	Port   int
+}
+
+func Setup(pool *dockertest.Pool, cln cleaner) (*Resource, error) {
+	container, err := pool.Run("bitnami/etcd", "3.4", []string{
```
atzoum, line 24: This is not arm64 friendly, correct?

regulation-worker/internal/destination/destination.go
```diff
@@ -77,13 +78,12 @@ func (d *DestMiddleware) getDestDetails(ctx context.Context) (backendconfig.Conf
...
-	var ok bool
 	if err = backoff.Retry(func() error {
 		pkgLogger.Debugf("Fetching backend-config...")
 		// TODO : Revisit the Implementation for Regulation Worker in case of MultiTenant Deployment
-		destConf, ok = d.Dest.Get(config.GetWorkspaceToken())
-		if !ok {
-			return fmt.Errorf("error while getting destination details")
+		destConf, err = d.Dest.Get(ctx, config.GetWorkspaceToken())
+		if err != nil {
+			return fmt.Errorf("error while getting destination details: %v", err)
```
atzoum, line 86: shouldn't we generally wrap errors?
```suggestion
			return fmt.Errorf("error while getting destination details: %w", err)
```

atzoum (commented):
Performed an initial, quick review round. Changes are enough, so will need to follow-up with a more detailed review round later

testhelper/etcd/etcd.go
fracasula: I just moved the existing code into this helper because now we're using it in more than one place but I didn't check whether it is arm64 friendly :slightly_smiling_face:

atzoum (commented):
Please accept my 2nd review round. More rounds shall follow

app/cluster/dynamic_test.go
```diff
@@ -194,15 +194,23 @@ func TestDynamicCluster(t *testing.T) {
 
 	t.Run("Update workspaceIDs", func(t *testing.T) {
```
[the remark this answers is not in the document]
fracasula: yeah sure. I covered that in the integration tests but it might be useful :+1:

config/backend-config/backend-config.go
```diff
@@ -277,10 +124,15 @@ func filterProcessorEnabledDestinations(config ConfigT) ConfigT {
...
-func configUpdate(eb *pubsub.PublishSubscriber, statConfigBackendError stats.RudderStats, workspaces string) {
-	sourceJSON, ok := backendConfig.Get(workspaces)
-	if !ok {
+func (bc *CommonBackendConfig) configUpdate(ctx context.Context, statConfigBackendError stats.RudderStats, workspaces string) {
+	sourceJSON, err := backendConfig.Get(ctx, workspaces)
+	if err != nil {
 		statConfigBackendError.Increment()
+		select {
+		case <-ctx.Done():
+		case bc.pollingErrors <- err:
```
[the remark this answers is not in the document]
fracasula: Good point, I'll review the logic and let you know what I can come up with :+1:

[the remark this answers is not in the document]
fracasula: I've added a leaky bucket approach since I don't really care if we got more errors in a row, I just want to terminate on the first one. I'll do some more tests to check things and add more coverage :+1:

config/backend-config/backend-config.go
```diff
@@ -289,26 +141,26 @@ func configUpdate(eb *pubsub.PublishSubscriber, statConfigBackendError stats.Rud
...
-		initializedLock.Lock()
-		defer initializedLock.Unlock()
-		initialized = true
-		LastSync = time.Now().Format(time.RFC3339)
-		eb.Publish(string(TopicBackendConfig), sourceJSON)
-		eb.Publish(string(TopicProcessConfig), filteredSourcesJSON)
+		bc.initializedLock.Lock()
+		defer bc.initializedLock.Unlock()
+		bc.initialized = true
+		LastSync = time.Now().Format(time.RFC3339) // TODO fix concurrent access
```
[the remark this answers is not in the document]
atzoum: I don't think it is that important if the health handler reports marginally out-of-sync data.. we could skip locks in this case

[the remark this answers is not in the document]
fracasula: Well, sooner or later it will come up again. The objective is to run all tests with the `-race` flag on anyway.

mt_gateway_test.go
```diff
@@ -0,0 +1,369 @@
...
+		require.Equal(t, "messageId_1", m["messageId"])
+	}
+
+	// Only the Gateway is running, so we don't expect any destinations to be hit.
+	require.EqualValues(t, 0, webhook.RequestsCount(), "webhook should have no requests because there is no processor")
+
+	serverModeReqKey := getETCDServerModeReqKey(t, serverInstanceID)
+	t.Logf("Server mode ETCD key: %s", serverModeReqKey)
+
+	{ // Trigger degraded mode, the Gateway should still work
```
atzoum, line 239: is there a way to split this test into sub-tests and at the same time make sure that sub-tests will run in order and not in parallel?
E.g. something similar to Ginkgo's [Ordered](https://onsi.github.io/ginkgo/#ordered-containers)

mt_gateway_test.go
```diff
@@ -0,0 +1,369 @@
...
+			).Scan(&count)
+			if err != nil {
+				return false
+			}
+			return count == 2
+		}, time.Minute, 50*time.Millisecond)
+		require.NoError(t, json.Unmarshal([]byte(eventPayload), &message))
+	}
+
+	{ // Checking that Workspace Changes errors are handled
```
atzoum, line 263: What shall happen if you send an event for a workspace that is not part of the node's workspaceIDs? Shall the gateway accept it?

mt_gateway_test.go
```diff
@@ -0,0 +1,369 @@
+package main
```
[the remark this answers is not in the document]
fracasula: I wish I could :slightly_smiling_face: 
That's how I initially wanted to have these tests but `main` is not a package, it's a program thus it is not importable.

I briefly explained this in an earlier PR of mine where I had to remove `TestMain` from `docker_test.go` (see [this PR](https://github.com/rudderlabs/rudder-server/pull/2059)) in order to have this test coexist in the same folder.

I'm doing it this way to avoid building an image of RudderServer and have to spawn yet another container, which is the only other solution that I can think of. The alternative is what I'm doing here i.e.:

```go
	done := make(chan struct{})
	go func() {
		defer close(done)
		Run(ctx)
	}()
```

:point_up: much cheaper than building a Docker image and running it as a container.

If you know of another way of course I'm all :ear: :ear:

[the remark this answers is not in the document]
atzoum: Keeping only `func main()` and `version` vars in `main` and moving everything else to another package would work?

If it is risky or too much work don't bother

mt_gateway_test.go
fracasula: I can do a `t.Run()`, they will run sequentially. I didn't do it because I prefer not to give the impression that you could run them separately. For example the main test ends because we fire an invalid workspace change. I can move it in `t.Run()` but then I'll have to add a `cancel()` to the main test in case you don't run the "invalid workspace change" scenario.

It feels to me that given that it's just doing 3 things it's easier to just follow them as is. It makes the main test a bit longer but easier to follow since it's clear that we're just sequentially changing state and not resetting anything between tests.

So yeah, I could split it but I don't think it'll make our life easier :thinking:

atzoum: How do multiple `t.Run` on the same test behave with ginkgo's -parallel flag?

fracasula: > We run tests with ginkgo. How do multiple `t.Run` on the same test behave with ginkgo's -parallel flag?

Do you mean with `--procs`? If you don't put `t.Parallel()` they are not executed in parallel.

```go
package ciao

import "testing"

func TestSomething(t *testing.T) {
	var count int
	t.Run("aris-1", func(t *testing.T) {
		t.Parallel()
		count++
	})
	t.Run("aris-2", func(t *testing.T) {
		t.Parallel()
		count++
	})
}
```

Run the above with `ginkgo --procs=24 --race ./ciao` and you get a data race warning. Do the same without the two `t.Parallel()` and it will be fine :+1: 

Btw, I will at least remove the extra scope: https://deepsource.io/gh/rudderlabs/rudder-server/run/ed1f1626-8c90-4ceb-a79a-1ba445e73774/go/CRT-A0008 :man_facepalming:

atzoum: thanks for taking the time to verify :) I would keep the extra scope and add a `// skipcq: CRT-A0008` instead, it is useful, to me!

fracasula: Did some reorganization, I think it's better now :+1: 
The tests can only be fired singularly, the only thing we can't do is move the last one up since it'll cause a termination so we have to keep it at the bottom (also no parallelism allowed obviously).

.enterprise-commit
fracasula: Enterprise commit removed after rebasing. This is now done.

mt_gateway_test.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+	backendConfRouter.HandleFunc("/hostedWorkspaceConfig", func(w http.ResponseWriter, r *http.Request) {
+		t.Fatal("Unexpected call to /hostedWorkspaceConfig given that hosted=false")
+	}).Methods("GET")
+	backendConfRouter.HandleFunc("/multitenantWorkspaceConfig", func(w http.ResponseWriter, r *http.Request) {
+		require.Equal(t, `["`+workspaceID+`"]`, r.FormValue("workspaceIds"))
+		_, _ = w.Write(marshalledWorkspaces)
+	}).Methods("GET")
+
+	backendConfigSrv := httptest.NewServer(backendConfRouter)
```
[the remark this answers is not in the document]
fracasula: I wouldn't bother tbh, it's just using `httptest` with a single working handler, the other one is mostly there to go fatal in case it gets called.

mt_gateway_test.go
fracasula: The gateway is quite dumb in this regard and I think it's fine for now. The one that knows about this is the proxy if I'm not mistaken where we keep a `writeKey` to workspace Map in the form of a Lua table. If there is no entry in the map then we reject it at the proxy level.

This is mostly what happens with IAM services as well. Requests coming from the outside can hit only "exposed" services (e.g. IAM service) that authenticate the request and let it through to the internal services only if it's valid. Then the internal services can avoid having the validation/authentication logic copied/duplicated in each µsvc because they can be hit from internal calls only so they can trust the requests.

I think the only case when the GW can be hit from outside is for self-hosted (open-source) deployments (cc @cisse21).

atzoum: The actual question was mostly referring to the scenario of a request with a valid workspaceID (writeKey), but not belonging to the node's list of workspaceIDs.
I understand that since the node keeps in memory the configuration from all workspaces, it will accept the request and store the event in the `gw` table, even though this is an event for a workspace not assigned to it.

fracasula: I see what you mean and although useful to help detect misbehaviour it still feels a bit like throw away code at this time, given that we haven't taken a decision on whether a node should fetch the configuration for all WS or not. I'd like to get back to this once I know more about that aspect if you don't mind.

atzoum: One could argue that if `rudder-server` doesn't respect the list of workspaceIDs allotted to it, there is no actual reason for keeping this list in `rudder-server`

fracasula: I'm setting up a call to decide what we want RS to do.

fracasula: We decided to simplify and have it fetch all workspaces configuration till the protocol is at least a bit more mature.

app/cluster/state/etcd.go
```diff
@@ -261,7 +262,31 @@ func (manager *ETCDManager) unmarshalWorkspace(raw []byte) workspace.ChangeEvent
 				manager.logger.Errorf("Failed to acknowledge workspace ID change for key: %s", req.AckKey)
 			}
 			return err
-		})
+		},
+		func(ctx context.Context, ackErr error) error {
```
lvrach, line 266: What if we merge those two callbacks ?

This signature: 
```go
func(ctx context.Context, ackErr error) error
```

can cover both ack and ack with error, if `ackErr is nil` we assume everything is ok

config/backend-config/backend-config.go
```diff
@@ -439,23 +276,53 @@ func Setup(configEnvHandler types.ConfigEnvI) (err error) {
...
+	bc.initializedLock.Lock()
+	bc.initialized = false
+	if bc.waitForConfigErrs != nil {
+		close(bc.waitForConfigErrs)
+	}
+	bc.initializedLock.Unlock()
+}
+
+// WaitForConfig waits until backend config has been initialized
+func (bc *CommonBackendConfig) WaitForConfig(ctx context.Context) error {
```
lvrach, line 303: `[question]` Why we have override this one ?

lvrach, line 303: `[question]` What the reason/motivation behind defining this method here ?

mt_gateway_test.go
```diff
@@ -0,0 +1,382 @@
...
+		return err
+	})
+	require.NoError(t, group.Wait())
+
+	webhook := whUtil.NewRecorder()
+	t.Cleanup(webhook.Close)
+
+	writeKey := rand.String(27)
+	workspaceID := rand.String(27)
+	workspaces := map[string]backendconfig.ConfigT{
```
lvrach, line 69: I would prefer to see workspace config in a json file, inside testdata folder.

lvrach (approved):
Overall looks good to me. Have a few nice to have comments and questions.

config/backend-config/backend-config.go
fracasula: It's always been there I think. I just moved it to the bottom next to `StartWithIDs` and `Stop` to make the file more readable and make more clear that they go together due to the fact that they use `bc.blockChan`, `bc.initialized` etcetera.

app/cluster/state/etcd.go
fracasula: Big fan of simpler APIs :+1: Pushed.

fracasula: Big fan of simpler APIs :+1: Pushed.

fracasula: Big fan of simpler APIs :+1: Pushed.

mt_gateway_test.go
fracasula: @lvrach there you [go](https://github.com/rudderlabs/rudder-server/pull/2078/commits/4e7754bb432716adf307b1451e09997fdbc2de69)

config/backend-config/multitenant_workspaces.go
```diff
@@ -68,44 +69,33 @@ func (workspaceConfig *MultiTenantWorkspacesConfig) GetWorkspaceLibrariesForWork
...
+	return workspaceConfig.getFromAPI(ctx, workspaces)
 }
 
 // getFromApi gets the workspace config from api
-func (workspaceConfig *MultiTenantWorkspacesConfig) getFromAPI(workspaceArr string) (ConfigT, bool) {
+func (workspaceConfig *MultiTenantWorkspacesConfig) getFromAPI(
+	ctx context.Context, workspaceArr string,
+) (ConfigT, error) {
 	// added this to avoid unnecessary calls to backend config and log better until workspace IDs are not present
 	if workspaceArr == workspaceConfig.Token {
```
raman2805, line 81: minor: not related to this commit, but the name workspaceArr is confusing (i thought it referred to a list of workspaces). Is Arr short for array or refers to something else? Its type seems to be a string. if its something else, maybe we should rename it ?

fracasula: I think it refers to `array`. Feel free to open a PR to address the issue :+1:

config/backend-config/multitenant_workspaces.go
```diff
@@ -68,44 +69,33 @@ func (workspaceConfig *MultiTenantWorkspacesConfig) GetWorkspaceLibrariesForWork
...
-		if err != nil {
-			pkgLogger.Errorf("Error fetching config: preparing request URL: %v", err)
-			return ConfigT{}, false
-		}
-		url = fmt.Sprintf("%s/multitenantWorkspaceConfig?workspaceIds=%s", configBackendURL, encodedWorkspaces)
-		url = url + "&fetchAll=true"
+		url = fmt.Sprintf("%s/hostedWorkspaceConfig?fetchAll=true", configBackendURL)
 	}
-	var respBody []byte
-	var statusCode int
```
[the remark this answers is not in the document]
fracasula: I'm not sure I follow @raman2805. If `cachedHostedWorkspaceConfig` is `true` we call `/cachedHostedWorkspaceConfig`. If it's `false` we call `/hostedWorkspaceConfig`. Why do you think is obsolete?

config/backend-config/backend-config_test.go
```diff
@@ -98,7 +99,7 @@ func initBackendConfig() {
 
 var _ = Describe("newForDeployment", func() {
 	It("supports single workspace config", func() {
-		os.Setenv("WORKSPACE_TOKEN", "password")
+		Expect(os.Setenv("WORKSPACE_TOKEN", "password")).To(BeNil())
```
cisse21, line 102: Minor but can't we be using t.SetEnv here?

config/backend-config/types.go
```diff
@@ -0,0 +1,167 @@
...
+type DgSourceTrackingPlanConfigT struct {
+	SourceId            string                            `json:"sourceId"`
+	SourceConfigVersion int                               `json:"version"`
+	Config              map[string]map[string]interface{} `json:"config"`
+	MergedConfig        map[string]interface{}            `json:"mergedConfig"`
+	Deleted             bool                              `json:"deleted"`
+	TrackingPlan        TrackingPlanT                     `json:"trackingPlan"`
+}
+
+func (dgSourceTPConfigT *DgSourceTrackingPlanConfigT) GetMergedConfig(eventType string) map[string]interface{} {
```
cisse21, line 135: Should we be having these functions in the `types.go` file?

docker_test.go
```diff
@@ -575,6 +578,9 @@ func TestMainFlow(t *testing.T) {
 
 func setupMainFlow(svcCtx context.Context, t *testing.T) <-chan struct{} {
 	setupStart := time.Now()
+	if testing.Verbose() {
+		t.Setenv("LOG_LEVEL", "DEBUG")
```
cisse21, line 582: Do we still need this Log level?

mt_gateway_test.go
```diff
@@ -0,0 +1,333 @@
...
+		require.Equal(t, marshalledWorkspaces.Len(), n)
+	}).Methods("GET")
+
+	backendConfigSrv := httptest.NewServer(backendConfRouter)
+	t.Logf("BackendConfig server listening on: %s", backendConfigSrv.URL)
+	t.Cleanup(backendConfigSrv.Close)
+
+	httpPort, err := freeport.GetFreePort()
+	require.NoError(t, err)
+	httpAdminPort, err := freeport.GetFreePort()
```
cisse21, line 93: Is there no chance that both `httpPort` and `httpAdminPort` end up being same?

cisse21 (approved):

config/backend-config/multitenant_workspaces.go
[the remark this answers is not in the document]
raman2805: my bad actually. Somehow (and I don't know how), when I was reviewing the code, this condition seemed to show up as being deleted in the new code. Still trying to wrap my head around reviewing stuff inside vs-code/github.

mt_gateway_test.go
lvrach: good point/catch, free port will close the port so it can be used again. So there is no guarantee this won't return the same port.

config/backend-config/types.go
fracasula: The rationale behind this code "relocation" is that the `backend-config.go` was getting way too big, so I decided to move all (or almost all?) types along with their methods in `types.go`.

`GetMergedConfig` and `fetchEventConfig` don't depend much on anything else, they seemed pretty self explanatory and just tied to the `DgSourceTrackingPlanConfigT` type so I put them there as well. If you want them back then I'll have to also move the `DgSourceTrackingPlanConfigT` type along with them :thinking:

mt_gateway_test.go
fracasula: Yes because we should use `httpPort` first and then ask for another one but there is very little chance that it would happen, and if it happens it'll just make the test fail. The alternative is to do a helper around it. I'll have a stab see if I can do a fix because they are both used by RS later on so we can't really use one and then ask for another. We can only keep on asking until we get a different one.

docker_test.go
fracasula: I was thinking to keep it because these tests fail quite often on CI and it would be useful to have more info. The test length time seems to be the same. WDYT?

config/backend-config/backend-config_test.go
fracasula: Pardon my ignorance, how do I get `t *testing.T` in a ginkgo test? :sweat_smile:

PR #2079, opened 2022-06-20
fix: removed the output of goRoutine count from the HTTP GET / of the server

Kannaj:
When we call HTTP GET / for the rudder server - we get the below output

```json
{"appType": "EMBEDDED", "server":"UP", "db":"UP","acceptingEvents":"TRUE","routingEvents":"TRUE","mode":"NORMAL","goroutines":"636", "backendConfigMode": "API", "lastSync":"2022-06-18T14:58:11Z", "lastRegulationSync":""}
```

Exposing values such as `goroutines` might be helpful information for harmful agents. This PR removes it from being shown by the server.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -148,6 +148,6 @@ func HealthHandler(w http.ResponseWriter, r *http.Request, jobsDB jobsdb.JobsDB)
 	}
 
 	appTypeStr := strings.ToUpper(config.GetEnv("APP_TYPE", EMBEDDED))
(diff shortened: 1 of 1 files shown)


cisse21:
@Kannaj  Thanks for the PR, Can you sign this https://github.com/rudderlabs/rudder-server/blob/master/CONTRIBUTING.md so that we can go ahead and merge this change

cisse21 (approved):

Kannaj:
@cisse21 thanks

lvrach:
Hey @Kannaj, one more thing. Because we are following conventional commits. Can you rename your commit to much the title (it needs a `fix:` prefix)? 

Or push another commit, empty works as well.

PR #2087, opened 2022-06-21
feat(core): add support for managing failed records in rsources.JobService

Sidddddarth:
# Description

rsources.JobService now provides support for adding/deleting and getting failed records.

## Notion Ticket

[Failed records support](https://www.notion.so/rudderstacks/Failed-records-support-in-rsources-JobService-6a44ebf12aba4fd7b5b9099d334fd28a)

## Security

### DIFF
--- a/services/rsources/handler.go
+++ b/services/rsources/handler.go
@@ -117,7 +117,7 @@ func (*sourcesHandler) AddFailedRecords(ctx context.Context, tx *sql.Tx, jobRunI
 	if err != nil {
 		return err
 	}
-	defer stmt.Close()
+	defer func() { _ = stmt.Close() }()
 
 	for i := range records {
 		_, err = stmt.ExecContext(
@@ -165,7 +165,7 @@ func (sh *sourcesHandler) GetFailedRecords(ctx context.Context, jobRunId string,
 	if err != nil {
 		return FailedRecords{}, err
 	}
-	defer rows.Close()
+	defer func() { _ = rows.Close() }()
 	for rows.Next() {
 		var failedRecord FailedRecord
 		err := rows.Scan(
@@ -316,7 +316,7 @@ func setupFailedKeysTable(ctx context.Context, db *sql.DB, defaultDbName string)
 		_ = tx.Rollback()
 		return err
 	}
-	_, err = tx.ExecContext(ctx, `create index if not exists failed_keys_job_run_id_idx on "rsources_failed_keys" (job_run_id)`)
+	_, err = tx.ExecContext(ctx, `create index if not exists rsources_failed_keys_job_run_id_idx on "rsources_failed_keys" (job_run_id)`)
 	if err != nil {
 		_ = tx.Rollback()
 		return err
--- a/services/rsources/handler.go
+++ b/services/rsources/handler.go
@@ -106,11 +106,26 @@ func (*sourcesHandler) IncrementStats(ctx context.Context, tx *sql.Tx, jobRunId 
 	return err
 }
 
-func (*sourcesHandler) AddFailedRecords(_ context.Context, _ *sql.Tx, _ string, _ JobTargetKey, _ []json.RawMessage) error {
-	return nil
+func (*sourcesHandler) AddFailedRecords(ctx context.Context, tx *sql.Tx, jobRunId string, key JobTargetKey, recordID []json.RawMessage) error {
+	sqlStatement := `insert into "failed_keys" (
+		job_run_id,
+		task_run_id,
+		source_id,
+		destination_id,
+		record_id,
+		created_at
+	) values ($1, $2, $3, $4, $5, $6)`
+
+	_, err := tx.ExecContext(
+		ctx,
+		sqlStatement,
+		jobRunId, key.TaskRunID, key.SourceID, key.DestinationID,
+		recordID, time.Now())
+
+	return err
 }
 
-func (sh *sourcesHandler) GetFailedRecords(ctx context.Context, tx *sql.Tx, jobRunId string, filter JobFilter) (FailedRecords, error) {
+func (*sourcesHandler) GetFailedRecords(ctx context.Context, tx *sql.Tx, jobRunId string, filter JobFilter) (FailedRecords, error) {
 	filterParams := []interface{}{}
 	filters := `WHERE job_run_id = $1`
 	filterParams = append(filterParams, jobRunId)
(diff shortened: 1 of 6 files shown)


services/rsources/rsources.go
```diff
@@ -101,7 +111,7 @@ type JobService interface {
 	AddFailedRecords(ctx context.Context, tx *sql.Tx, jobRunId string, key JobTargetKey, records []json.RawMessage) error
 
 	// TODO: future extension
```
atzoum, line 113: need to adapt the doc here as well

services/rsources/handler.go
```diff
@@ -106,17 +106,91 @@ func (*sourcesHandler) IncrementStats(ctx context.Context, tx *sql.Tx, jobRunId
 	return err
 }
 
-func (*sourcesHandler) AddFailedRecords(_ context.Context, _ *sql.Tx, _ string, _ JobTargetKey, _ []json.RawMessage) error {
+func (*sourcesHandler) AddFailedRecords(ctx context.Context, tx *sql.Tx, jobRunId string, key JobTargetKey, records []json.RawMessage) error {
+	sqlStatement := `insert into "failed_keys" (
```
atzoum, line 110: using a prepared statement here could be beneficial as well

services/rsources/handler.go
```diff
@@ -106,17 +106,91 @@ func (*sourcesHandler) IncrementStats(ctx context.Context, tx *sql.Tx, jobRunId
...
+	}
+	if err := rows.Err(); err != nil {
+		return FailedRecords{}, err
+	}
+
+	return failedRecords, nil
 }
 
 func (sh *sourcesHandler) Delete(ctx context.Context, jobRunId string) error {
 	sqlStatement := `delete from "rsources_stats" where job_run_id = $1`
```
atzoum, line 185: should perform both statements as an atomic action in the same transaction

services/rsources/handler.go
```diff
@@ -106,17 +106,91 @@ func (*sourcesHandler) IncrementStats(ctx context.Context, tx *sql.Tx, jobRunId
...
 	_, err := sh.localDB.ExecContext(ctx, sqlStatement, jobRunId)
+	if err != nil {
+		return err
+	}
+
+	sqlStatement = `delete from "failed_keys" where job_run_id = $1`
+	_, err = sh.localDB.ExecContext(ctx, sqlStatement, jobRunId)
+
 	return err
 }
```
atzoum, line 196: do we need `DropStats` method?

services/rsources/handler.go
```diff
@@ -106,17 +106,91 @@ func (*sourcesHandler) IncrementStats(ctx context.Context, tx *sql.Tx, jobRunId
...
 
 func (sh *sourcesHandler) Delete(ctx context.Context, jobRunId string) error {
 	sqlStatement := `delete from "rsources_stats" where job_run_id = $1`
 	_, err := sh.localDB.ExecContext(ctx, sqlStatement, jobRunId)
+	if err != nil {
+		return err
+	}
+
+	sqlStatement = `delete from "failed_keys" where job_run_id = $1`
```
atzoum, line 191: cleanup loop should be adapted as well. We don't want to keep old failed keys

services/rsources/handler_test.go
```diff
@@ -63,6 +64,38 @@ var _ = Describe("Using sources handler", func() {
...
+				[]byte(`{"record-2": "id-2"}`),
+			})
+			jobFilters := JobFilter{
+				SourceID:  []string{"source_id"},
+				TaskRunID: []string{"task_run_id"},
+			}
+
+			failedRecords, err := sh.GetFailedRecords(context.Background(), jobRunId, jobFilters)
+			Expect(err).NotTo(HaveOccurred(), "it should be able to get failed records")
+			expcetdRecords := FailedRecords{
```
atzoum, line 80: small typo
```suggestion
			expectedRecords := FailedRecords{
```

services/rsources/handler.go
```diff
@@ -241,5 +388,12 @@ func (sh *sourcesHandler) setupLogicalReplication(ctx context.Context) error {
 			return fmt.Errorf("failed to create subscription on shared database: %w", err)
 		}
 	}
+
+	refreshSubscriptionQuery := fmt.Sprintf(`ALTER SUBSCRIPTION %s REFRESH PUBLICATION`, subscriptionName)
```
atzoum, line 392: 👍

atzoum (approved):

services/rsources/handler.go
```diff
@@ -106,24 +106,107 @@ func (*sourcesHandler) IncrementStats(ctx context.Context, tx *sql.Tx, jobRunId
...
+	filterParams := []interface{}{}
+	filters := `WHERE job_run_id = $1`
+	filterParams = append(filterParams, jobRunId)
+
+	if len(filter.TaskRunID) > 0 {
+		filters += fmt.Sprintf(` AND task_run_id  = ANY ($%d)`, len(filterParams)+1)
+		filterParams = append(filterParams, pq.Array(filter.TaskRunID))
+	}
+
+	if len(filter.SourceID) > 0 {
```
cisse21, line 148: What is the use case where we might have to filter based on multiple sourceID's?

atzoum: It is part of the agreed contract. I guess, sourceID might change when a job is reconfigured

services/rsources/handler.go
```diff
@@ -106,24 +106,107 @@ func (*sourcesHandler) IncrementStats(ctx context.Context, tx *sql.Tx, jobRunId
 	return err
 }
 
-func (*sourcesHandler) AddFailedRecords(_ context.Context, _ *sql.Tx, _ string, _ JobTargetKey, _ []json.RawMessage) error {
-	return nil
+func (*sourcesHandler) AddFailedRecords(ctx context.Context, tx *sql.Tx, jobRunId string, key JobTargetKey, records []json.RawMessage) (err error) {
+	stmt, err := tx.Prepare(`insert into "rsources_failed_keys" (
```
lvrach, line 110: I am wondering if an insert with multiple rows would be a better choice here, both in terms of performance and consistency (no partial failures).

Depending on the domain partial failures might not be appropriate

lvrach (approved):

services/rsources/handler.go
atzoum: `pq` doesn't appear to have support for multi-row insert, plus we will also need to cater for the limit of 1000 rows per multi-row insert...
That being said, I don't believe it is worth the trouble

PR #2090, opened 2022-06-23
feat: expose robots.txt to disable crawling

lvrach:
# Description

We want data-plane URLs to stop appearing in Google searches and avoid being crawled in general.

The common way to do this is using [robots.txt](https://www.cloudflare.com/en-gb/learning/bots/what-is-robots.txt/):
The simplest rules that exclude all:
```
User-agent: * 
Disallow: / 
```
Any bot is not allowed to scrape paths under `/` -> everything.

## Notion Ticket

[Avoid-data-plane-URL-indexing](https://www.notion.so/rudderstacks/Avoid-data-plane-URL-indexing-b4dd98584cd344c3921ebf633d21262a)

## Security

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -1332,6 +1332,11 @@ func (gateway *HandleT) healthHandler(w http.ResponseWriter, r *http.Request) {
 	app.HealthHandler(w, r, gateway.jobsDB)
 }
 
+// Robots prevents robots from crawling the gateway endpoints
+func (gateway *HandleT) robots(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


cisse21 (approved):

atzoum (approved):
This is something that can be done at revproxy level as well

lvrach:
> This is something that can be done at revproxy level as well

Yeah, it could be done. To avoid unnecessarily HTTP requests to our gateway, but since rudder-server can be used without out revproxy from our open-source users we must have it here as well.

I think that the `/robots.txt` would be so few that wouldn't have any impact on the load of the system.

PR #2106, opened 2022-06-24
feat(warehouse): enable postgres query execution

achettyiitr:
# Description

Enable Query execution for warehouse postgres for debugging purpose.

## Notion Ticket

https://www.notion.so/rudderstacks/Added-debug-logs-for-delte-Query-for-EXPLAIN-Use-wrapper-for-this-381272db0c4a4653ac792ec5aa9b7a3d

## Security


warehouse/postgres/postgres.go
```diff
@@ -147,6 +148,7 @@ func loadConfig() {
 	stagingTablePrefix = "rudder_staging_"
 	config.RegisterBoolConfigVariable(false, &skipComputingUserLatestTraits, true, "Warehouse.postgres.skipComputingUserLatestTraits")
 	config.RegisterDurationConfigVariable(30, &txnRollbackTimeout, true, time.Second, "Warehouse.postgres.txnRollbackTimeout")
+	config.RegisterBoolConfigVariable(false, &enableSQLStatementExecutionPlan, true, "Warehouse.postgres.enableSQLStatementExecutionPlan")
```
sivashanmukh, line 151: Lets default this to false

warehouse/postgres/postgresQueryBuilder.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	if e.enableWithQueryPlan {
+		var rows *sql.Rows
+		if e.txn != nil {
+			rows, err = e.txn.Query(sqlStatement)
+		} else if e.db != nil {
+			rows, err = e.db.Query(sqlStatement)
+		} else {
+			err = fmt.Errorf("[WH][POSTGRES] Not able to handle query execution for statement: %s", sqlStatement)
```
sivashanmukh, line 33: Let the error say that its because both txn and db are nil

warehouse/postgres/postgresQueryBuilder.go
```diff
@@ -0,0 +1,63 @@
...
+		var response []string
+		for rows.Next() {
+			var s string
+			if err = rows.Scan(&s); err != nil {
+				err = fmt.Errorf("[WH][POSTGRES] Error occurred while processing destination revisionID query %+v with err: %w", e, err)
+				return
+			}
+			response = append(response, s)
+		}
+		pkgLogger.Infof("[WH][POSTGRES] Execution Query plan for statement: %s is %s", sqlStatement, strings.Join(response, "\n"))
```
sivashanmukh, line 51: Is it okay to have `\n` in logs, would it make it hard to capture in Loki?

sivashanmukh (changes requested):

warehouse/postgres/postgres.go
achettyiitr: Yes. It's false by default. Here true is hot reloadable.

warehouse/postgres/postgresQueryBuilder.go
achettyiitr: Done.

warehouse/postgres/postgresQueryBuilder.go
achettyiitr: Although logs in loki looks good. But it's coming in different line. It's better we should output it in a single line. 
Screenshot 2022-06-24 at 5 13 57 PM

achettyiitr: Using space separated.

achettyiitr: Using Raw string literal instead of \n now.

achettyiitr: But still this will comes as the corresponding output only. Explored other options like YAML, XML, JSON but this is the best options we have.

warehouse/postgres/postgres.go
sivashanmukh: 👍

warehouse/postgres/postgresQueryBuilder.go
sivashanmukh: 👍

sivashanmukh (approved):
LGTM

warehouse/postgres/postgresQueryBuilder.go
```diff
@@ -0,0 +1,62 @@
...
+				err = fmt.Errorf("[WH][POSTGRES] Error occurred while processing destination revisionID query %+v with err: %w", e, err)
+				return
+			}
+			response = append(response, s)
+		}
+		pkgLogger.Infof(fmt.Sprintf(`[WH][POSTGRES] Execution Query plan for statement: %s is %s`, sqlStatement, strings.Join(response, `
+`)))
+	}
+	if e.txn != nil {
+		result, err = e.txn.Exec(sqlStatement)
```
abhimanyubabbar, line 54: It is using `Exec`  in the main query which means that it can only be used for INSERT, UPDATE, DELETE clauses right and not for generic SQL query. If yes, I think the name of function is misnomer, we can rename it to `handleExec`. Wdyt ?

warehouse/postgres/postgresQueryBuilder.go
```diff
@@ -0,0 +1,62 @@
+package postgres
+
+import (
+	"database/sql"
+	"fmt"
+	"strings"
+)
+
+type QueryExecution struct {
```
abhimanyubabbar, line 9: Should we rename it to `QueryParams` or maybe `Params` ?

warehouse/postgres/postgresQueryBuilder.go
```diff
@@ -0,0 +1,62 @@
...
+	txn                 *sql.Tx
+	db                  *sql.DB
+	query               string
+	enableWithQueryPlan bool
+}
+
+// handleQueryExecution
+// Print execution plan if enableWithQueryPlan is set to true else return result set.
+// Currently, these statements are supported by EXPLAIN ANALYZE
+// Any SELECT, INSERT, UPDATE, DELETE, VALUES, EXECUTE, DECLARE, CREATE TABLE AS, or CREATE MATERIALIZED VIEW AS statement, whose execution plan you wish to see.
```
abhimanyubabbar, line 19: SELECT clause can't be sent in right ? The sql.Result interface doesn't contain rows. Should we remove this from comment then ?

abhimanyubabbar (changes requested):
Some comments from first initial look at the change. Let me know what you think ?

warehouse/postgres/postgresQueryBuilder.go
achettyiitr: QueryParams suites good.

abhimanyubabbar:
Also the name of the file `postgresQueryBuilder.go` is a stutter meaning that it is already under the `postgres` package. So we can name it to `query_builder.go`. Wdyt ?

warehouse/postgres/postgresQueryBuilder.go
```diff
@@ -0,0 +1,62 @@
...
+		sqlStatement := "EXPLAIN " + e.query
+
+		var rows *sql.Rows
+		if e.txn != nil {
+			rows, err = e.txn.Query(sqlStatement)
+		} else if e.db != nil {
+			rows, err = e.db.Query(sqlStatement)
+		} else {
+			err = fmt.Errorf("[WH][POSTGRES] Not able to handle query execution for statement: %s as both txn and db are nil", sqlStatement)
+			return
```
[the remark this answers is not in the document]
achettyiitr: Done.

warehouse/postgres/postgresQueryBuilder.go
achettyiitr: Yes.

abhimanyubabbar (commented):
LGTM 👍🏼

PR #2108, opened 2022-06-24
fix: calculate consistent tagStr cache keys in stats

atzoum:
# Description

Previously we were calculating a cache key by iterating over the map using the default random order of the map. This meant that for 1 distinct set of 10 tags we ended up with 10! = 3628800 different key combinations in our cache map for the same thing.

Here I'm therefore introducing `stats.Tags#Strings` and `stats.Tags#String` methods. Both of them return tag key value pairs sorted by increasing key order.

Fixes an issue with inconsistent (unsorted) cache keys being calculated while populating an in-memory cache holding tagged stats clients.

## Notion Ticket

[Calculate consistent tagStr cache keys in stats](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=80a1a01c2df4434da3b347f5177fef86)

## Security

### DIFF
--- a/services/stats/stats.go
+++ b/services/stats/stats.go
@@ -70,7 +70,6 @@ type Tags map[string]string
 
 // Strings returns all key value pairs as an ordered list of strings, sorted by increasing key order
 func (t Tags) Strings() []string {
-
 	res := make([]string, 0, len(t)*2)
 	// sorted by tag name (!important for consistent map iteration order)
 	tagNames := make([]string, 0, len(t))
--- a/services/stats/stats.go
+++ b/services/stats/stats.go
@@ -3,6 +3,7 @@ package stats
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


services/stats/stats.go
```diff
@@ -67,6 +68,27 @@ func Init() {
 
 type Tags map[string]string
 
+// Strings returns all key value pairs as an ordered list of strings, sorted by increasing key order
+func (t Tags) Strings() []string {
+	res := make([]string, 0, len(t)*2)
```
fracasula, line 73:
```suggestion
func (t Tags) Strings() []string {
	if len(t) == 0 {
		return nil
	}
	res := make([]string, 0, len(t)*2)
```

fracasula (commented):
LGTM, just a few questions since I don't have enough context here. How often would this `Strings()` method with the sorting be called? Do we sort once then send the stats over and clear up the memory? Some more context would be appreciated.

Also I see that this is related to a "high memory usage due to stats" task. Does this resolve the high memory usage? If yes, can you explain how?

services/stats/stats.go
fracasula: OK it makes sense. I would update the PR description though, this makes it more clear :+1:

fracasula (approved):
Performance wise I don't think the sorting would create issues and not having it seems to cause bigger problems. Also, I'm assuming that we usually don't handle too many tags (e.g. 10? correct me if I'm wrong) so :+1:

services/stats/stats.go
atzoum: open to proposals :)

fracasula: Done.

fracasula: Also you didn't comment on the suggestion, do you think it doesn't make sense to check the length first? In case it's empty it saves us a few invocations.

atzoum: To be frank I do not consider it critical, but will do so :)

lvrach (approved):

PR #2110, opened 2022-06-27
chore(multi-tenant): mt flow to dos

kuldeep0020:
# Description
Following changes are introduced:
1. Earlier the three members `ctx` `cancel` and `blockChan` of `CommonBackendConfig` were being populated into the struct during the `CommonBackendConfig.StartWithIDs()` function call. This is not intuitive and expected behaviour from the `CommonBackendConfig.StartWithIDs()` function. As populating these should be the part of creation of backendConfig during the setup process, so now this is happening with the help of `Setup()` function. 
2. `WorkspaceIDs(ctx context.Context) <-chan workspace.ChangeEvent` of `StaticProvider` 
     - Previous behaviour: This method was returning `<-chan workspace.ChangeEvent` without any workspace IDs or acknowledgement function. The assumption was that `StaticProvider` will only be used for dedicated or hosted type of deployment. In hosted type of deployments, we do not need workspace IDs to poll the backend config and in dedicated we poll backend config only for one workspace ID which was being done in `main`. This was creating an anti-pattern between Static and ETCD based provider.
     - New Behaviour: Now the channel will have the workspace ID in case of dedicated deployment and that will be used to poll the backend config. 
     
     Advantage: This will make sure that the implementation is uniform for the interface (`ChangeEventProvider`) across ETCD based provider and Static provider.

3. `WorkspaceIDs(ctx context.Context) <-chan workspace.ChangeEvent` function of `ETCDManager` was not consistent every time the context gets cancelled.

## Notion Ticket

[< Notion Ticket >](https://www.notion.so/rudderstacks/Document-mt-dev-deployment-5d624eb63cc34ccd8f81c8be01018a93)

## Security

- [*] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

### DIFF
--- a/app/cluster/state/etcd.go
+++ b/app/cluster/state/etcd.go
@@ -270,11 +270,11 @@ func (manager *ETCDManager) WorkspaceIDs(ctx context.Context) <-chan workspace.C
 	}
 
 	appTypeStr := strings.ToUpper(config.GetEnv("APP_TYPE", app.PROCESSOR))
-	modeRequestKey := fmt.Sprintf(workspacesRequestsKeyPattern, manager.Config.ReleaseName, manager.Config.ServerIndex, appTypeStr)
-	manager.logger.Infof("Workspace ID Lookup Key: %s", modeRequestKey)
+	workspaceIdLookupKey := fmt.Sprintf(workspacesRequestsKeyPattern, manager.Config.ReleaseName, manager.Config.ServerIndex, appTypeStr)
+	manager.logger.Infof("Workspace ID Lookup Key: %s", workspaceIdLookupKey)
 
 	resultChan := make(chan workspace.ChangeEvent, 1)
-	resp, err := manager.Client.Get(ctx, modeRequestKey)
+	resp, err := manager.Client.Get(ctx, workspaceIdLookupKey)
 	if err != nil {
 		return errChWorkspacesRequest(err)
 	}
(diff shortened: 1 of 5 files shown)


app/cluster/state/etcd.go
```diff
@@ -297,18 +297,14 @@ func (manager *ETCDManager) WorkspaceIDs(ctx context.Context) <-chan workspace.C
 			for _, event := range watchResp.Events {
 				switch event.Type {
 				case mvccpb.PUT:
-					select {
-					case resultChan <- manager.unmarshalWorkspace(event.Kv.Value):
-					case <-ctx.Done():
-					}
+					resultChan <- manager.unmarshalWorkspace(event.Kv.Value)
```
fracasula, line 300: This change would introduce a bug when:
1. the `ctx` gets canceled
2. `resultChan` is either full or its readers get terminated due to the same `ctx` cancelation or other reasons

This would result in an edge case where we would block indefinitely when readers cannot drain the channel and the context gets canceled at the same time (or right before). Thus the context will not get honoured.

app/cluster/state/static.go
```diff
@@ -37,10 +38,11 @@ func (s *StaticProvider) ServerMode(ctx context.Context) <-chan servermode.Chang
 }
 
 // WorkspaceIDs returns an empty channel, since we don't expect workspaceIDs updates with static provider
-// TODO: This method should return proper workspaceIDs for backend config, even with static provide.
 func (s *StaticProvider) WorkspaceIDs(ctx context.Context) <-chan workspace.ChangeEvent {
-	ch := make(chan workspace.ChangeEvent)
-
+	ch := make(chan workspace.ChangeEvent, 1)
+	wId := backendconfig.DefaultBackendConfig.AccessToken()
```
fracasula, line 43: We should avoid accessing objects globally. It makes tests really hard. Also, what if we want to use a non-default backend config? I think the backendConfig should be a `StaticProvider` attribute inside the `struct`.

PR #2114, opened 2022-06-27
feat(core): introduce NoCopy struct through PublishSubscriber

kuldeep0020:
# Description

Introducing the `NoCopy` Struct to make sure that we do not copy struct unnecessarily. This will enforce us to use the pointers to the struct as and when needed instead of copying the struct and will save hours of developer's time debugging the issues introduced because of copying the structs. This still allows us to return the structs and does not obstruct other use-cases. 

**References:** 

[Proposal in Golang Forum(was not accepted though but this explains the use-case)](https://github.com/golang/go/issues/30450) 
[use-case](https://stackoverflow.com/questions/52494458/nocopy-minimal-example)
[Documentation of the library](https://pkg.go.dev/github.com/savsgio/gotils/nocopy#NoCopy)
[Github Repo of the library implementation](https://github.com/savsgio/gotils/tree/master/nocopy)

## Notion Ticket

[Notion ticket](https://www.notion.so/rudderstacks/Introduce-NoCopy-Struct-bf31d6c28d3e41d1a1165d7267d3234c)

## Security

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -174,3 +174,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20220512140231-539c8e751b99 // indirect
 )
+
+require github.com/savsgio/gotils v0.0.0-20220530130905-52f3993e8d6d
--- a/go.mod
+++ b/go.mod
@@ -174,5 +174,3 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20220512140231-539c8e751b99 // indirect
 )
-
-require github.com/savsgio/gotils v0.0.0-20220530130905-52f3993e8d6d
--- a/utils/noCopy/guard.go
+++ b/utils/noCopy/guard.go
@@ -1,6 +1,0 @@
-package noCopy
-
-type Locker interface {
-	Lock()
-	Unlock()
-}
--- a/utils/noCopy/guard.go
+++ b/utils/noCopy/guard.go
@@ -1,0 +1,6 @@
+package noCopy
+
+type Locker interface {
+	Lock()
+	Unlock()
+}
--- a/utils/pubsub/pubsub.go
+++ b/utils/pubsub/pubsub.go
@@ -17,7 +17,7 @@ type DataChannel <-chan DataEvent
 
 // PublishSubscriber stores the information about subscribers interested for a particular topic
 type PublishSubscriber struct {
-	noCopy.NoCopy
+	noCopy.CopyRestrictor
 	lastEventMutex sync.RWMutex
 	// lastEvent holds the last event for each topic so that we can send it to new subscribers
 	lastEvent map[string]*DataEvent
--- a/utils/pubsub/pubsub.go
+++ b/utils/pubsub/pubsub.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"sync"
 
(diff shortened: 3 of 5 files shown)


atzoum:
please describe what are we guarding against here that [this](https://github.com/golangci/govet/blob/master/copylock.go) cannot cover.

go.mod
```diff
@@ -174,3 +174,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20220512140231-539c8e751b99 // indirect
 )
+
+require github.com/savsgio/gotils v0.0.0-20220530130905-52f3993e8d6d
```
lvrach, line 178: It might be better to copy paste the implementation rather than importing this package.

lvrach (approved):
Approve, with a minor concern regarding importing

fracasula (approved):

cisse21 (approved):

utils/noCopy/guard.go
```diff
@@ -0,0 +1,6 @@
+package noCopy
+
+type Locker interface {
+	Lock()
+	Unlock()
+}
```
fracasula, line 6: Judging by [this answer](https://stackoverflow.com/questions/52494458/nocopy-minimal-example) it looks like the linter does a dumb check on just the two method names (i.e. `Lock()` and `Unlock()`) so I don't think we need this interface. Just having the `struct` with those two empty methods should be enough :+1:

kuldeep0020: Make sense.

utils/pubsub/pubsub.go
```diff
@@ -15,6 +17,7 @@ type DataChannel <-chan DataEvent
 
 // PublishSubscriber stores the information about subscribers interested for a particular topic
 type PublishSubscriber struct {
+	noCopy.CopyRestrictor
```
atzoum, line 20: We are already using a `sync.RWMutex` in PublishSubscriber. Embedding a `noCopy.CopyRestrictor` appears to be redundant, since go vet will give an error for `sync.RWMutex` even if we don't use `noCopy.CopyRestrictor`.

This is was what I was trying to pinpoint in my [previous comment](https://github.com/rudderlabs/rudder-server/pull/2114#issuecomment-1168214138) as well

kuldeep0020: This is indeed redundant for `PublishSubscriber` for the obvious reason that `sync.RWMutex` internally uses `NoCopy` struct. The whole motive of this is to start using the `CopyRestrictor` if we are copying some struct which has say a map or slice, then we might not want it to be copied or be careful at that time. If that struct have `CopyRestrictor` then govet will throw the warning. I am not able to recall a direct use-case but I am sure I have faced something similar earlier in rudder-server code.

atzoum: Fine, but let's introduce it when the need arises, wdyt?

kuldeep0020: If someone else will keep this in mind then it is fine by me. I though I should introduce it before moving out to other team. I just want that developer's time should not be wasted by such silly mistakes. :)

fracasula: I didn't notice that the linter was complaining anyway when copying the `PublishSubscriber` struct, I took for granted it wasn't and we were trying to address that. Given that it does I completely agree with @atzoum

fracasula: @kuldeep0020 perhaps instead of throwing away the PR you know of some other code that should not be copied and that it is not using mutexes?

PR #2118, opened 2022-06-28
chore(warehouse): event schema performance improvements

abhimanyubabbar:
# Description
This PR limits the frequency counters we are creating and storing for event models and schema versions.

## Notion Ticket
[Notion - PR](https://www.notion.so/rudderstacks/Event-Schema-OOM-Kill-Issue-64b6d7384d704f1a987696012c6fd86a)

## Security

## Benchmark Result
```
name                      old time/op    new time/op    delta
EventSchemaHandleEvent-8     2.55s ± 0%     2.38s ± 0%   ~     (p=1.000 n=1+1)

name                      old alloc/op   new alloc/op   delta
EventSchemaHandleEvent-8    8.52MB ± 0%    4.46MB ± 0%   ~     (p=1.000 n=1+1)

name                      old allocs/op  new allocs/op  delta
EventSchemaHandleEvent-8     55.1k ± 0%     31.6k ± 0%   ~     (p=1.000 n=1+1)
```


event-schema/counter_manager.go
```diff
@@ -14,14 +14,32 @@ func init() {
...
+		count++
 	}
 	countersCache[schemaHash] = frequencyCountersMap
 }
 
+// populateFrequencyCounters is responsible for capturing the frequency counters
+// for a particular schema version into memory for fast computation.
+func populateFrequencyCounters(schemaHash string, frequencyCounters []*FrequencyCounter) {
+	populateFrequencyCountersBounded(schemaHash, frequencyCounters, frequencyCounterLimit)
+	return
```
lvrach, line 40: I am wondering why we have introduced a new method, instead of keeping the name the same or renaming it. 

Since the function is not exported we are not breaking any API

event-schema/counter_manager.go
```diff
@@ -36,13 +54,41 @@ func getAllFrequencyCounters(schemaHash string) []*FrequencyCounter {
 	return frequencyCounters
 }
 
-func getFrequencyCounter(schemaHash, key string) *FrequencyCounter {
+func getFrequencyCounterBounded(schemaHash, key string, bound int) *FrequencyCounter {
 	schemaVersionCounters, ok := countersCache[schemaHash]
 	if !ok {
 		schemaVersionCounters = make(map[string]*FrequencyCounter)
 		countersCache[schemaHash] = schemaVersionCounters
 	}
```
lvrach, line 62: Are those operations thread/concurrent safe?
`countersCache` is used by multiple methods, if any of the calls are concurrent it is going to be an issue

Also `counterCache` is a global variable, it would make more sense to create a struct around it and use methods instead of functions to edit it. I understand this is beyond the scope of this PR.

```go

type FrequencyCounter struct {
    // maybe a mutext
    countersCache map[string]map[string]*FrequencyCounter
}


func (f *FrequencyCounter) func getFrequencyCounterBounded(schemaHash, key string, bound int) *FrequencyCounter {

...
}
```

event-schema/counter_manager.go
```diff
@@ -36,13 +54,41 @@ func getAllFrequencyCounters(schemaHash string) []*FrequencyCounter {
...
+
+		toDelete := -1 * diff
+		for k := range schemaVersionCounters {
+			if toDelete > 0 {
+				delete(schemaVersionCounters, k)
+				toDelete--
+			} else {
+				break
+			}
+		}
```
lvrach, line 85: I feel a bit unexpected that a get method, has also a delete side-effect. 

Could that happen during `populateFrequencyCounters` or another method that runs periodically?

event-schema/test_input.json
```diff
@@ -0,0 +1 @@
...2,"993":993,"994":994,"995":995,"996":996,"997":997,"998":998,"999":999,"type":"track","event":"Demo Track","sentAt":"2019-08-12T05:08:30.909Z","channel":"android-sdk","context":{"app":{"name":"RudderAndroidClient","build":"1","version":"1.0","namespace":"com.rudderlabs.android.sdk"},"device":{"id":"49e4bdd1c280bc00","name":"generic_x86","model":"Android SDK built for x86","manufacturer":"Google"},"locale":"en-US","screen":{"width":1080,"height":1794,"density":420},"traits":{"anonymousId":"49e4bdd1c280bc00"},"library":{"name":"com.rudderstack.android.sdk.core"},"network":{"carrier":"Android"},"user_agent":"Dalvik/2.1.0 (Linux; U; Android 9; Android SDK built for x86 Build/PSR1.180720.075)"},"rudderId":"90ca6da0-292e-4e79-9880-f8009e0ae4a3","messageId":"a82717d0-b939-47bd-9592-59bbbea66c1a","properties":{"label":"Demo Label","value":5,"testMap":{"t1":"a","t2":4},"category":"Demo Category","floatVal":4.501,"testArray":[{"id":"elem1","value":"e1"},{"id":"elem2","value":"e2"}]},"anonymousId":"anon_id","integrations":{"All":true},"originalTimestamp":"2019-08-12T05:08:30.909Z"}],"writeKey":"29grCPTQ6XrVOSK76M7vfE5BWjb","requestIP":"127.0.0.1","receivedAt":"2022-06-15T13:51:08.754+05:30"}
```
lvrach, line 1: `[minor]` It is convectional in Go to put such test payload instead a `testdata` folder

https://dave.cheney.net/2016/05/10/test-fixtures-in-go

event-schema/event_schema.go
```diff
@@ -48,6 +48,18 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/timeutil"
 )
 
+func init() {
+	fmt.Println("Initialized the data structures")
```
lvrach, line 52: Is this a debug only?
```suggestion
```

event-schema/event_schema.go
```diff
@@ -48,6 +48,18 @@ import (
...
+
+	// Following data structures store events and schemas since last flush
+	updatedEventModels = make(map[string]*EventModelT)
+	updatedSchemaVersions = make(map[string]*SchemaVersionT)
+	offloadedEventModels = make(map[string]map[string]*OffloadedModelT)
+	offloadedSchemaVersions = make(map[string]map[string]*OffloadedSchemaVersionT)
+	archivedEventModels = make(map[string]map[string]*OffloadedModelT)
+	archivedSchemaVersions = make(map[string]map[string]*OffloadedSchemaVersionT)
+}
```
lvrach, line 61: `non-actionable` It would be better if those variables are not global, for better readability, ownership and testability. This is beyond the scope of this PR, I just mention it.

event-schema/counter_manager_test.go
```diff
@@ -0,0 +1,51 @@
...
+	countersCache[hash.String()] = map[string]*FrequencyCounter{
+		"key1": {Name: "key1"},
+		"key2": {Name: "key2"},
+	}
+
+	// Updated frequencyCounterLimit bound
+	frequencyCounterLimit := 2
+	// This should ideally result in a new key being added to the countersCache
+	// but since the bound is already reached, it simply returns nil for the freq counter
+	fq := getFrequencyCounterBounded(hash.String(), "new-key", frequencyCounterLimit)
```
lvrach, line 38: `[minor]` Maybe returning an error along with `nil` would be better here.

event-schema/event_schema.go
```diff
@@ -349,6 +364,8 @@ func (manager *EventSchemaManagerT) handleEvent(writeKey string, event EventT) {
 			}
 			stats.NewTaggedStat("reload_offloaded_event_model", stats.CountType, stats.Tags{"module": "event_schemas", "writeKey": eventModel.WriteKey, "eventIdentifier": eventModel.EventIdentifier}).Increment()
 		} else if wasArchived {
+			// If we saw event from an archived event model, reload the model into memory
+			// and archive the oldest model then. TODO: A test case for this ?
```
cisse21, line 368: `Query`: With regards to the line `374` why would we have to create the model again? We populate from the Db right so we should be able to reload?

event-schema/counter_manager.go
```diff
@@ -14,14 +14,32 @@ func init() {
 	}
 }
 
-func populateFrequencyCounters(schemaHash string, frequencyCounters []*FrequencyCounter) {
+// populateFrequencyCountersBounded is responsible for capturing the frequency counters which
+// are available in the db and store them in memory but in a bounded manner.
+func populateFrequencyCountersBounded(schemaHash string, frequencyCounters []*FrequencyCounter, bound int) {
 	frequencyCountersMap := make(map[string]*FrequencyCounter)
+	count := 0
```
fracasula, line 21: You don't really need `count` here, you could just use the slice index.

As you can see from [this example](https://go.dev/play/p/oaO92Vcnlqy) I created, no matter how you "slice it", when ranging the index always starts from zero. Meaning you could use `index` instead of count.

event-schema/counter_manager_test.go
```diff
@@ -0,0 +1,51 @@
+package event_schema
```
fracasula, line 1: Can we have a `TestConcurrentGetFrequencyCounterBounded` where multiple go routines are invoking the function? Please remember to run it with the `-race` flag like `go test -race ...`.

event-schema/event_schema.go
```diff
@@ -618,6 +636,27 @@ func getPrivateDataJSON(schemaHash string) []byte {
 }
 
 func (manager *EventSchemaManagerT) flushEventSchemas() {
+	// This will run forever. If you want to quit in between, change it to ticker and call stop()
+	// Otherwise the ticker won't be GC'ed
+	ticker := time.Tick(flushInterval)
+	for range ticker {
```
fracasula, line 642: Can we propagate a context here please? If it's too much then you can always provide a better way to stop this. a `for/select` should do the trick here.

event-schema/event_schema.go
```diff
@@ -618,6 +636,27 @@ func getPrivateDataJSON(schemaHash string) []byte {
...
+
+		// TODO: Why we are throwing panics from the statements beneath us ?
+		// TODO: Refactored to use only a single dbHandle rather than creating the handle everytime. Ask Akash of the use case in first place ?
+		if !areEventSchemasPopulated {
```
fracasula, line 646: The access (both reads and write) to this variable seem to happen from multiple routines, it doesn't seem "data race" safe.

event-schema/event_schema.go
```diff
@@ -628,138 +667,121 @@ func (manager *EventSchemaManagerT) flushEventSchemas() {
...
+	updatedSchemaVersions = make(map[string]*SchemaVersionT)
+	toDeleteEventModelIDs = []string{}
+	toDeleteSchemaVersionIDs = []string{}
 
-		updatedEventModels = make(map[string]*EventModelT)
-		updatedSchemaVersions = make(map[string]*SchemaVersionT)
-		toDeleteEventModelIDs = []string{}
-		toDeleteSchemaVersionIDs = []string{}
+	manager.schemaVersionLock.Unlock()
+	manager.eventModelLock.Unlock()
```
fracasula, line 782: Can we move these two on top as `defer`s right after:
```go
	if len(updatedEventModels) == 0 && len(schemaVersionsInCache) == 0 {
		manager.eventModelLock.Unlock()
		manager.schemaVersionLock.Unlock()
		return nil
	}
```

I'm asking this as to decrease the chances of messing up with the locks if someone where to add a return in between. It would also increase readability since we'll notice with a glance where we lock and where we unlock.

event-schema/event_schema.go
```diff
@@ -628,138 +667,121 @@ func (manager *EventSchemaManagerT) flushEventSchemas() {
...
-		if len(updatedEventModels) == 0 && len(schemaVersionsInCache) == 0 {
-			manager.eventModelLock.Unlock()
-			manager.schemaVersionLock.Unlock()
-			continue
-		}
+	flushDBHandle = createDBConnection()
 
-		flushDBHandle = createDBConnection()
+	txn, err := flushDBHandle.Begin()
+	assertError(err)
```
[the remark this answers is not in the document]
fracasula: A panic will shut the whole thing down and it could happen easily since in this case maybe the DB is just not responding. I also see that `flushEventSchemasToDB` actually returns an `error` so why not returning and removing the `recover` on `defer`?

Another question, if we cannot flush, would it make sense to retry or we actually want to throw away everything and restart from scratch? Can you explain a bit the requirements here please?

event-schema/event_schema_test.go
```diff
@@ -1,13 +1,202 @@
...
+	"github.com/rudderlabs/rudder-server/utils/logger"
+	"github.com/stretchr/testify/require"
 )
 
-func TestAlert(t *testing.T) {
-	RegisterFailHandler(Fail)
-	RunSpecs(t, "Alert Suite")
+var pgResource *destination.PostgresResource
+
+func TestMain(m *testing.M) {
```
fracasula, line 25: Can you please avoid using `TestMain` and simply do one of:
1. nested test like we do in `docker_test.go`
2. use ginkgo with contexts

If you use `TestMain` then it will be impossible to run a unit test in the same package without executing this code that is completely unrelated and it is creating actual Docker containers (i.e. psql).

event-schema/event_schema_test.go
```diff
@@ -1,13 +1,202 @@
...
+	var privateData PrivateDataT
+	err = json.Unmarshal(privateDataRaw, &privateData)
+	if err != nil {
+		return nil, err
+	}
+
+	return privateData.FrequencyCounters, nil
+}
+
+func BenchmarkEventSchemaHandleEvent(b *testing.B) {
```
fracasula, line 171: What has this benchmark revealed? Have you done a before and after so that we can compare results?

event-schema/counter_manager.go
abhimanyubabbar: Still exploring the event schema domain space. I agree to the point that given they are unexported we can have only one. I'll make the change.

event-schema/counter_manager.go
abhimanyubabbar: This is a good point. Let me think about better structuring it so the function don't have unintended side-effects

event-schema/event_schema.go
```diff
@@ -48,6 +48,18 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/timeutil"
 )
 
+func init() {
+	fmt.Println("Initialized the data structures")
+
+	// Following data structures store events and schemas since last flush
+	updatedEventModels = make(map[string]*EventModelT)
```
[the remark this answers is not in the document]
abhimanyubabbar: Missed that part, will take a look thanks 👍🏼

event-schema/event_schema.go
abhimanyubabbar: Good point. I'll dig up a bit to see why that's the case.

event-schema/event_schema_test.go
abhimanyubabbar: Good point, let me change it up a bit.

event-schema/test_input.json
abhimanyubabbar: Good catch. 🙌🏼

event-schema/counter_manager.go
abhimanyubabbar: The way it is written originally is to take a very wide lock and then perform operations on all global data structures. This works okay for currently functionality as code is a bit limited in scope but it is bound to create issues as mentioned above given multiple workers will be working on the same data structures.

event-schema/counter_manager.go
```diff
@@ -36,15 +44,49 @@ func getAllFrequencyCounters(schemaHash string) []*FrequencyCounter {
 	return frequencyCounters
 }
 
-func getFrequencyCounter(schemaHash, key string) *FrequencyCounter {
+// pruneFrequencyCounters brings the frequency counters back to desired bound.
+func pruneFrequencyCounters(schemaHash string, bound int) {
```
fracasula, line 48: Can we have a test to check that the prune happens as expected?

event-schema/counter_manager_test.go
fracasula: Can you at least create a Notion ticket to check the race conditions here? There are lots of accesses to the same variables (e.g. `countersCache`) but the mutexes that are supposed to protect access are not even in the same file. We should look into this.

event-schema/event_schema.go
fracasula: What's the case then?

event-schema/event_schema.go
```diff
@@ -617,149 +635,184 @@ func getPrivateDataJSON(schemaHash string) []byte {
...
+	defer flushDBHandle.Close()
+
+	txn, err := flushDBHandle.Begin()
+	if err != nil {
+		return fmt.Errorf("unable to begin txn to flush event schemas, err: %w", err)
+	}
+
+	if len(updatedEventModels) > 0 {
+		if err := flushEventModels(ctx, txn, updatedEventModels); err != nil {
+			txn.Rollback()
```
fracasula, line 692: If we can't rollback, shouldn't we return an error like you do when you can't begin a transaction?

event-schema/event_schema.go
```diff
@@ -617,149 +635,184 @@ func getPrivateDataJSON(schemaHash string) []byte {
...
+	for uid := range updatedSchemaVersions {
+		versionIDs = append(versionIDs, uid)
+	}
 
-				archiveVersionsForArchivedModelsSQL := fmt.Sprintf(`UPDATE %s SET archived=%t WHERE event_model_id IN ('%s')`, SCHEMA_VERSIONS_TABLE, true, strings.Join(toDeleteEventModelIDs, "', '"))
-				_, err = txn.Exec(archiveVersionsForArchivedModelsSQL)
-				assertTxnError(err, txn)
-			}
+	deleteOldVersionsSQL := fmt.Sprintf(`DELETE FROM %s WHERE uuid IN ('%s')`, SCHEMA_VERSIONS_TABLE, strings.Join(versionIDs, "', '"))
+	_, err := txn.Exec(deleteOldVersionsSQL)
```
fracasula, line 725: Why not use `ExecContext` instead? It looks like `flushSchemaVersions` does have a context but nobody is using it.

event-schema/event_schema.go
```diff
@@ -853,6 +907,9 @@ func assertError(err error) {
 	}
 }
 
+func logTxnError(err error, txn *sql.Tx) {
+}
```
fracasula, line 911: ?

event-schema/event_schema_test.go
```diff
@@ -1,13 +1,218 @@
...
+	"github.com/rudderlabs/rudder-server/testhelper/destination"
+	"github.com/rudderlabs/rudder-server/utils/logger"
 )
 
-func TestAlert(t *testing.T) {
-	RegisterFailHandler(Fail)
-	RunSpecs(t, "Alert Suite")
+// TODO: Refactor this TestMain so that it doesn't get called everytime ?
+// TODO: Can we use stretchr's testify suite library ?
+func SetupDB(cleanup *testhelper.Cleanup) *destination.PostgresResource {
```
fracasula, line 27:
```suggestion
func setupDB(cleanup *testhelper.Cleanup) *destination.PostgresResource {
```

event-schema/event_schema_test.go
```diff
@@ -1,13 +1,218 @@
...
+
+	mg := &migrator.Migrator{
+		Handle:                     pgResource.DB,
+		MigrationsTable:            "node_migrations",
+		ShouldForceSetLowerVersion: config.GetBool("SQLMigrator.forceSetLowerVersion", true),
+	}
+
+	err = mg.Migrate("node")
+	if err != nil {
+		log.Fatalf("Unable to run the migrations for the node.")
```
fracasula, line 48: If you go fatal then you don't execute the cleanup procedures. Please consider that when doing things like `destination.SetupPostgres` you're passing up a `cleanup` object. When using `fatal` function you're basically forcing an exit with `os.Exit(1)`, thus the containers will not be cleaned up. Whether that might not matter much on CI at the moment it matters when developing locally and will also be quite important in case we were to decide to run the CI builds on dedicated machines in the future.

event-schema/event_schema_test.go
```diff
@@ -1,13 +1,218 @@
...
+	return pgResource
+}
+
+func jobsDBInit(pgResource *destination.PostgresResource) {
+	// Setup env for jobsdb.
+	os.Setenv("JOBS_DB_HOST", pgResource.Host)
+	os.Setenv("JOBS_DB_USER", pgResource.User)
+	os.Setenv("JOBS_DB_PASSWORD", pgResource.Password)
+	os.Setenv("JOBS_DB_DB_NAME", pgResource.Database)
+	os.Setenv("JOBS_DB_PORT", pgResource.Port)
```
fracasula, line 71: You should use either `t.Setenv()` if you are using `t *testing.T` or `GinkgoT().Setenv()` if you're using Ginkgo.

event-schema/event_schema_test.go
fracasula: I can't find the report, have you added it somewhere?

event-schema/event_schema.go
abhimanyubabbar: Removed the unused functions.

event-schema/event_schema.go
abhimanyubabbar: I should have done that. Thanks for pointing out ! 👍🏼

event-schema/event_schema_test.go
abhimanyubabbar: Done 👍🏼

event-schema/counter_manager.go
abhimanyubabbar: I'll add that testcase as well. Thanks 👍🏼

event-schema/event_schema_test.go
abhimanyubabbar: I agree. We should be very careful with the Fatal function call in golang. Thank you for catching this 👍🏼

fracasula (approved):

achettyiitr (approved):
LGTM

PR #2131, opened 2022-07-01
chore: merge enterprise code

lvrach:
# Description

Merge private rudder-server-enterprise repository in this one to avoid maintenance overhead. 

## Change summary

* Copy/paste from rudder-server-enterprise into enterprise folder (first commit)
* [remove/clean-up all](https://github.com/rudderlabs/rudder-server/pull/2131/commits/c8c974be313d3511175a386165d4ee343f2db300) enterprise related scripts 
* Introduce [SSLP](https://github.com/rudderlabs/rudder-server/pull/2131/commits/1364838be8b76141babb41e7596218921f5fc81e) license
* noop instead of nil checks for enterprise features
* Use enterpriseToken to disable/enable enterprise
* Fix tests
 

## Notion Ticket

[Move enterprise code to rudder-server repo
](https://www.notion.so/rudderstacks/Move-enterprise-code-to-rudder-server-repo-519520f4c88a4e20b1ba8c61d858436e
)
## Security


enterprise/replay/README.md
```diff
@@ -0,0 +1,87 @@
+### ENV Variables
```
cisse21, line 1: We should probably remove this readme? and point it to a private notion link

fracasula:
@lvrach It LGTM but it's very hard to review as you can imagine, but I'm assuming this is mostly a cut and paste kind of job so we shouldn't expect major issues, right? On the other hand I see some failures here (both tests and deepsource).

lvrach:
> @lvrach It LGTM but it's very hard to review as you can imagine, but I'm assuming this is mostly a cut and paste kind of job so we shouldn't expect major issues, right? On the other hand I see some failures here (both tests and deepsource).

The first commit, is copy/paste. If you ignore that and take the one by one. It should be easier.

enterprise/replay/fileuploader/s3downloader.go
```diff
@@ -15,10 +15,21 @@ import (
...
-	downloadSession := session.Must(session.NewSession(&aws.Config{
-		Region:      aws.String(region),
-		Credentials: credentials.NewStaticCredentials(config.GetEnv("AWS_ACCESS_KEY_ID", ""), config.GetEnv("AWS_SECRET_ACCESS_KEY", ""), ""),
-	}))
+	if err != nil {
+		return 0, err
+	}
+
+	awsAccessKeyId := config.GetEnv("AWS_ACCESS_KEY_ID", "")
+	awsSecretAccessKey := config.GetEnv("AWS_SECRET_ACCESS_KEY", "")
```
fracasula, line 23: Wouldn't it make sense to just return an error if these are empty since logging in with an empty key or secret is simply not possible?

main.go
```diff
@@ -108,7 +108,7 @@ func loadConfig() {
 	config.RegisterDurationConfigVariable(15, &gracefulShutdownTimeout, false, time.Second, "GracefulShutdownTimeout")
 	config.RegisterIntConfigVariable(524288, &MaxHeaderBytes, false, 1, "MaxHeaderBytes")
 
-	config.RegisterStringConfigVariable(enterpriseToken, &enterpriseToken, false, "enterpriseToken")
+	enterpriseToken = config.GetEnv("ENTERPRISE_TOKEN", enterpriseToken)
```
fracasula, line 111: I find this a bit confusing since `enterpriseToken` is empty to begin with, maybe do:
```go
	enterpriseToken = config.GetEnv("ENTERPRISE_TOKEN", "")
```

fracasula (approved):
Left some minor comments, overall looks good :+1:

app/app.go
```diff
@@ -52,8 +57,13 @@ func (a *App) Setup() {
 		a.initCPUProfiling()
 	}
 
-	// initialize enterprise features, if available
-	a.initEnterpriseFeatures()
+	if a.options.EnterpriseToken == "" {
```
atzoum, line 60: Why do we need a token (string) instead of a boolean?

main.go
lvrach: enterpriseToken is not always empty, it can be defined as a build arg

app/app.go
lvrach: Because we want control-plane service to verify that the token is valid. This will happen at a second stage.

enterprise/replay/fileuploader/s3downloader.go
fracasula: Oh OK, that's good to know :+1: Btw, I meant that it's not possible to login to AWS if you don't have both an access key and a secret. Sounds right to you?

main.go
fracasula: What I meant is that it's always empty when you are there, on line 111. Right? If not then my bad :slightly_smiling_face:

fracasula: Oh wow :smile: Never seen `ldflags` used to pass credentials before. I guess that there's always a first time.

PR #2141, opened 2022-07-04
chore(router): improve collect metrics performance

cisse21:
# Description

Moving from `sjson` to `jsonfast` gives us both memory and performance improvement. This is to address growing memory 
footprint of this function

```
goos: darwin
goarch: arm64
pkg: github.com/rudderlabs/rudder-server/router

Benchmark_SJSON_SET-8             144646              7980 ns/op           15336 B/op         64 allocs/op
Benchmark_JSON_MARSHAL-8          469032              2776 ns/op            2376 B/op         39 allocs/op
Benchmark_FASTJSON_MARSHAL-8      458791              2827 ns/op            2416 B/op         27 allocs/op
```

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=c8b968f0ec424e8b9cba64979bd55289)

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -16,6 +16,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	jsoniter "github.com/json-iterator/go"
+
 	"github.com/cenkalti/backoff/v4"
 	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
 	"github.com/rudderlabs/rudder-server/processor/integrations"
@@ -32,7 +34,6 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/bytesize"
 	utilTypes "github.com/rudderlabs/rudder-server/utils/types"
 	"github.com/tidwall/gjson"
-	"github.com/tidwall/sjson"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/rudderlabs/rudder-server/config"
@@ -98,7 +99,7 @@ type HandleT struct {
 	failureMetricLock                      sync.RWMutex
 	diagnosisTicker                        *time.Ticker
 	requestsMetric                         []requestMetric
-	failuresMetric                         map[string][]failureMetric
+	failuresMetric                         map[string]map[string]int
 	customDestinationManager               customdestinationmanager.DestinationManager
 	throttler                              throttler.Throttler
 	guaranteeUserEventOrder                bool
@@ -231,6 +232,8 @@ var (
 	disableEgress                                                 bool
 )
 
+var jsonfast = jsoniter.ConfigCompatibleWithStandardLibrary
+
 type requestMetric struct {
 	RequestRetries       int
 	RequestAborted       int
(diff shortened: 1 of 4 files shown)


Sidddddarth (approved):

router/router.go
```diff
@@ -1805,22 +1808,21 @@ func (rt *HandleT) collectMetrics(ctx context.Context) {
 		// failureMetric struct
 		rt.failureMetricLock.RLock()
 		for key, values := range rt.failuresMetric {
```
atzoum, line 1810: A surge of failed events could cause high memory usage as well (large number of `failureMetric` structs stored in `failuresMetric` map) since we are only collecting metrics on standard time intervals (60s) without caring about their accumulation size.

router/router.go
```diff
@@ -1805,22 +1808,21 @@ func (rt *HandleT) collectMetrics(ctx context.Context) {
 		// failureMetric struct
 		rt.failureMetricLock.RLock()
```
atzoum, line 1809: A full lock would be appropriate here, since at the end we are ressetting the `failuresMetric` map

atzoum:
>This is to address growing memory footprint of this function
Any allocations performed by the function should be garbage collected, i.e. they shouldn't contribute at the end of the day to the memory growth, since what we are ultimately retaining is a `string`. I am afraid that high memory usage should be caused by something else.

router/router.go
```diff
@@ -2447,11 +2441,11 @@ func (rt *HandleT) HandleOAuthDestResponse(params *HandleDestOAuthRespParamsT) (
 			return http.StatusInternalServerError, trRespBody
 		}
 	}
-	// By default send the status code & response from transformed response directly
+	// By default, send the status code & response from transformed response directly
 	return trRespStatusCode, trRespBody
 }
 
-func (rt *HandleT) ExecDisableDestination(destinationJob types.DestinationJobT, workspaceId, destResBody, rudderAccountId string) (int, string) {
+func (rt *HandleT) ExecDisableDestination(destinationJob *types.DestinationJobT, workspaceId, destResBody, rudderAccountId string) (int, string) {
```
atzoum, line 2448: we dont need a `types.DestinationJobT` but a `backendconfig.DestinationT`, correct?
```suggestion
func (rt *HandleT) ExecDisableDestination(destination backendconfig.DestinationT, workspaceId, destResBody, rudderAccountId string) (int, string) {
```

router/router.go
```diff
@@ -1594,8 +1589,11 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
 				}
 
 				rt.failureMetricLock.Lock()
-				failureMetricVal := failureMetric{RouterDestination: rt.destName, UserId: resp.userID, RouterAttemptNum: resp.status.AttemptNum, ErrorCode: resp.status.ErrorCode, ErrorResponse: resp.status.ErrorResponse}
-				rt.failuresMetric[event] = append(rt.failuresMetric[event], failureMetricVal)
+				_, ok := rt.failuresMetric[event][string(resp.status.ErrorResponse)]
```
atzoum, line 1592: truncating the `ErrorResponse` to a specific limit, e.g. 1000 characters would be useful as well. We don't care about it being a parseable json, do we?

router/router.go
```diff
@@ -1594,8 +1589,11 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
...
+				if !ok {
+					rt.failuresMetric[event] = make(map[string]int)
+				}
```
atzoum, line 1595:
```suggestion
				if _, ok := rt.failuresMetric[event][string(resp.status.ErrorResponse)]; !ok {
					rt.failuresMetric[event] = make(map[string]int)
				}
```

router/router.go
cisse21: I even thought of doing the same but since these end up in snowflake I just let the response be a whole JSON rather than a truncated object in case they need to be queried

lvrach (approved):

PR #2142, opened 2022-07-04
multi-tenant jobsdb requirements

Sidddddarth:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security


jobsdb/jobsdb.go
```diff
@@ -1562,14 +1564,39 @@ func (jd *HandleT) setSequenceNumber(l lock.DSListLockToken, newDSIdx string) da
...
-		_, err := jd.dbHandle.Exec(sqlStatement)
-		jd.assertError(err)
+		newDSMin = dRangeList[len(dRangeList)-1].maxJobID + 1
+	} else {
+		newDSMin = 1
 	}
+	tableName := jd.tablePrefix + `_jobs_` + newDSIdx
+	sequenceName := tableName + `_job_id_seq`
+	// drop sequence
+	sqlStatement := fmt.Sprintf(`DROP SEQUENCE IF EXISTS "%s"`, sequenceName)
```
atzoum, line 1576: According to the [documentation ](https://www.postgresql.org/docs/8.1/datatype.html#DATATYPE-SERIAL) it might be easier to create first the sequence and then the table, but what do we really gain by doing this? 

The `BIGSERIAL` data type is practically a `BIGINT` bundled with a sequence.

Or, why aren't we using the identity column syntax which supports sequence initialization as well? e.g.

```
job_id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY (START WITH 1000)
```

What about the job_status table? We have the same BIGSERIAL datatype there as well.

As a sidenote, I would find it appropriate to perform all operations within `createDS` as part of a single transaction

sql/migrations/jobsdb/000001_create_tables.down.tmpl
```diff
@@ -1,7 +1,7 @@
 -- Drop any existing datasets
 {{range .Datasets}}
-    DROP TABLE {{$.Prefix}}_jobs_{{.}};
-    DROP TABLE {{$.Prefix}}_job_status_{{.}};
+    DROP TABLE "{{$.Prefix}}_jobs_{{.}}";
```
atzoum, line 3: Is there an impact by modifying these? Will it cause them to run again? Do we have any kind of documentation on these migrations and how they work?

jobsdb/jobsdb.go
```diff
@@ -1544,6 +1544,8 @@ func (jd *HandleT) createDS(newDS dataSetT, l lock.DSListLockToken) {
 
 	// In case of a migration, we don't yet update the in-memory list till
 	// we finish the migration
+
+	jd.logger.Info(l, l != nil)
```
atzoum, line 1548:
```suggestion

```

sql/migrations/jobsdb/000001_create_tables.down.tmpl
Sidddddarth: No impact from some preliminary tests. Much of the jobsdb code has been modified to include `"`s as well, so I didn't think much into it.

jobsdb/jobsdb.go
Sidddddarth: > According to the [documentation ](https://www.postgresql.org/docs/8.1/datatype.html#DATATYPE-SERIAL)it might be easier to create first the sequence and then the table,

Indeed, this is what the migrator does. Over here I did it in this order just so there's least amount of discrepancy from the original code.
As for why `BIGINT`, iirc there was some issue with -ve job_ids with a `BIGSERIAL` column(can't be negative). which is why we resorted to BIGINT. The status tables won't have -ve ids, so `BIGSERIAL` suffices.
> appropriate to perform all operations within createDS as part of a single transaction

agreed

jobsdb/jobsdb.go
```diff
@@ -1563,9 +1563,9 @@ func (jd *HandleT) setSequenceNumber(l lock.DSListLockToken, newDSIdx string) da
 
 	// Now set the min JobID for the new DS just added to be 1 more than previous max
 	if len(dRangeList) > 0 {
-		newDSMin := dRangeList[len(dRangeList)-1].maxJobID
+		newDSMin := dRangeList[len(dRangeList)-1].maxJobID + 1
 		// jd.assert(newDSMin > 0, fmt.Sprintf("newDSMin:%d <= 0", newDSMin))
-		sqlStatement := fmt.Sprintf(`SELECT setval(pg_get_serial_sequence('"%s_jobs_%s"', 'job_id'), %d)`,
+		sqlStatement := fmt.Sprintf(`ALTER SEQUENCE "%[1]s_jobs_%[2]s_job_id_seq" MINVALUE %[3]d START %[3]d RESTART %[3]d`,
```
atzoum, line 1568: can we introduce a test scenario for the specific use case we are trying to cover?

atzoum (approved):

PR #2144, opened 2022-07-06
fix(router): transformer proxy alerts

sanpj2292:
# Description

- Changed the tag to `destType` to make it easier to group on the same tag for better alerting
- Changed the info log to debug log to not spam our logs with transformer proxy related details

## Notion Ticket

[Proxy Alert Parameters](https://www.notion.so/rudderstacks/Setting-Transformer-Proxy-Alert-s-Parameters-5c32ef6abbc941448fa63b0d1a89d350)

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -747,7 +747,7 @@ func (worker *workerT) processDestinationJobs() {
 								"workspace":     workspaceID,
 							}).Count(len(result))
 
-							pkgLogger.Infof(`[TransformerProxy] (Dest-%v) {Job - %v} Input Router Events: %v, Out router events: %v`, worker.rt.destName,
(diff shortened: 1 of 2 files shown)


cisse21 (approved):

koladilip (commented):
destination to destType, with this change what happens to existing data?

koladilip:
@sanpj2292 how long data will be kept in influxdb? now old data to be queried differently from new data right?

sanpj2292:
@koladilip ,
> how long data will be kept in influxdb? 

Not sure about this part actually

> now old data to be queried differently from new data right?

Yes, the old data would need to queried differently with this change

PR #2152, opened 2022-07-11
test(destination): workload identity support for google destinations

sanpj2292:
Signed-off-by: Sai Sankeerth <sanpj2292@github.com>

# Description

We would like to check the following
- If the config output after adding service account to the identity pool created works with the existing package of `google/go/x/oauth2`
- If we send the service credentials as part of `config.Credentials` will it take the service account credentials and work as well
- For workload identity enabled GKE clusters, we need to define the `workloadIdentity` configuration in the data-plane deployment with all required details. After correctly defining the configuration, a check for the same will happen while initialising a new client for the google resources to identify a valid `config.Credentials`

## Notion Ticket

POC for [Google Workload Identity](https://www.notion.so/rudderstacks/Google-Workload-Identity-PRD-7421bef28a54445bb89c1d4ad4de87fa?d=1c1e4bbc62f84ac4bdc602cec55809a7#71b831c39ce040a7922c0f8b7a4ebf8c)

## Security


services/streammanager/bqstream/bqstreammanager.go
```diff
@@ -66,24 +66,19 @@ func NewProducer(destinationConfig interface{}, o Opts) (*Client, error) {
...
-	if err != nil {
-		return nil, createErr(err, "error in BQStream while unmarshalling credentials json")
-	}
 	opts := []option.ClientOption{
-		option.WithCredentialsJSON([]byte(config.Credentials)),
 		option.WithScopes([]string{
 			gbq.BigqueryInsertdataScope,
 		}...),
 	}
+	if !googleutils.IsEmptyCredentials(config.Credentials) {
```
atzoum, line 74: We will most probably need an additional `Config` option, e.g. `WorkloadIdentity bool`, so that we don't allow users to skip `Credentials`, unless they have explicitly opted for using GKE Workload Identity as an authentication method.

sanpj2292: The thing is that, the workload identity changes are being made for a customer. They desire to deploy our data-plane in Google cloud. Since they are waiting for this feature for a long-time, we are testing this feature in AWS. Hence we don't want to add any deployment related stuff to control plane dashboard as of now.

@krishna2020 @saikumarrs @koladilip correct me if I'm wrong here

atzoum: TL;DR it needs to be evident for future generations of developers why we are allowing empty credentials here

Maybe another runtime flag should be passed, eg. through an environment variable for controlling whether we shall allow users to skip credentials or not, e.g. GKE_DEPLOYMENT=true.
I am saying this because if we leave it this way,  a future refactoring might not take this case (GKE) into consideration and change the system's behaviour to always require `Credentials`, breaking workload identity support.

koladilip: @sanpj2292 I agree with @atzoum second suggestion, we should keep control plane agnostic to data plane deployment as it can change right. If we can detect or supply using environment variable about data plane and make decisions about code paths based on it would be good :)

utils/googleutils/googleutils.go
```diff
@@ -6,9 +6,21 @@ import (
...
 		return fmt.Errorf("Google Developers Console client_credentials.json file is not supported")
 	}
 	return nil
 }
+
+/*
+The credentials are deemed to be empty when either the field credentials is
+sent as empty string or when the field is set with "{}"
+*/
+func IsEmptyCredentials(credentials string) bool {
```
saikumarrs, line 24: Just a better name.
```suggestion
func IsCredentialsStringEmpty(credentials string) bool {
```

utils/googleutils/googleutils.go
```diff
@@ -6,9 +6,21 @@ import (
 	"golang.org/x/oauth2/google"
 )
 
+const (
+	EMPTY_CREDS = "{}"
+)
+
 func CompatibleGoogleCredentialsJSON(jsonKey []byte) error {
```
saikumarrs, line 13: [Nitpicking]
```suggestion
func CheckCredentialsJSONCompatibility(jsonKey []byte) error {
```

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -60,7 +60,7 @@ func NewProducer(destinationConfig interface{}, o Opts) (*PubsubClient, error) {
 		return nil, fmt.Errorf("invalid configuration provided, missing projectId")
 	}
 	var client *pubsub.Client
-	if config.Credentials != "" { // Normal configuration requires credentials
+	if !googleutils.IsEmptyCredentials(config.Credentials) { // Normal configuration requires credentials
```
saikumarrs, line 63: The client initialization flow is not consistent across the destinations. It'd be easier with the introduction of the server env variable now. I'd imagine the flow to be something like the below:

```
// Server environment variable takes precedence 
// in controlling the flow
if (<server env> != true) {
  // validate the creds and throw error on failure
}
// initialize the client
```

saikumarrs (commented):
Reviewed only a partial set of files.

services/streammanager/bqstream/bqstreammanager.go
```diff
@@ -66,24 +66,19 @@ func NewProducer(destinationConfig interface{}, o Opts) (*Client, error) {
...
-		return nil, createErr(err, "credentials not being sent")
-	}
-	if err = googleutils.CompatibleGoogleCredentialsJSON([]byte(config.Credentials)); err != nil {
-		return nil, createErr(err, "incompatible credentials")
-	}
-	confCreds = []byte(config.Credentials)
-	err = json.Unmarshal(confCreds, &credentialsFile)
-	if err != nil {
-		return nil, createErr(err, "error in BQStream while unmarshalling credentials json")
-	}
```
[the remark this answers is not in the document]
sanpj2292: Kind of, yes. As we are currently planning to allow empty string(`""`) and `{}` as valid credentials string

services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -72,7 +72,12 @@ func NewProducer(destinationConfig interface{}, o Opts) (*PubsubClient, error) {
 			return nil, err
 		}
 	} else { // No configuration
-		return nil, fmt.Errorf("invalid configuration provided, missing credentials")
+		if config.Credentials == "" {
```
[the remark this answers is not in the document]
sanpj2292: This condition probably doesn't make sense anymore 🤔

services/streammanager/bqstream/bqstreammanager.go
sanpj2292: Made the changes. Can you guys check and confirm if they're ok ? @aris1009 @koladilip

utils/googleutils/googleutils.go
sanpj2292: Done

services/streammanager/googlepubsub/googlepubsubmanager.go
sanpj2292: Refactored this, kindly check and confirm if it is ok

services/streammanager/googlepubsub/googlepubsubmanager.go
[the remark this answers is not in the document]
sanpj2292: Refactored the client initialisation flow, can you check and confirm ?

utils/googleutils/googleutils.go
sanpj2292: @aris1009 can you let us know if this name change makes more sense ?

koladilip (approved):

services/streammanager/googlepubsub/googlepubsubmanager.go
saikumarrs: I didn't understand the new changes completely. Are we not going to use the environment variables?
Moreover, the flow still seems to be inconsistent across different destinations.

sanpj2292: The environment variables check is abstracted into a method which is defined in `googleutils.go`. You can check [here](https://github.com/rudderlabs/rudder-server/blob/53a4c1a53030c9114559552c25231fca50751994/utils/googleutils/googleutils.go#L22).
What flow are you talking about here ?

sanpj2292: The client initialisation flow is fairly common in all the google destinations(now atleast). Can you check [this](https://github.com/rudderlabs/rudder-server/blob/53a4c1a53030c9114559552c25231fca50751994/services/streammanager/bqstream/bqstreammanager.go#L74-L81) and let me know if I'm missing something here ?

saikumarrs: I meant, the variables are defined in the configuration file, config.yaml. Isn't the idea to drive it from the deployment YAML file?

utils/googleutils/googleutils.go
```diff
@@ -3,12 +3,55 @@ package googleutils
...
+			workloadIdentity.projectId -> The project which contains the workload identity pool in GKE
+			workloadIdentity.gsaName -> The google IAM Service account name
+			workloadIdentity.gsaProjectId -> The project in which the mentioned google IAM service account is present
+
+			Note: We are currently only validating the schema rather than the values
+		*/
+		workloadProjectId := config.GetString(fmt.Sprintf("%s.gwiProjectId", WI_CONFIG_KEY), "")
+		workloadGsaName := config.GetString(fmt.Sprintf("%s.gsaName", WI_CONFIG_KEY), "")
+		workloadGsaProjectId := config.GetString(fmt.Sprintf("%s.gsaProjectId", WI_CONFIG_KEY), "")
+		return workloadProjectId != "" && workloadGsaName != "" && workloadGsaProjectId != ""
```
saikumarrs, line 54: Why validate these? I'd just check the workload's enabled status (missing) and its type.

services/streammanager/googlepubsub/googlepubsubmanager.go
saikumarrs: Got it, but we should remove them before merging the changes.

utils/googleutils/googleutils.go
sanpj2292: > check the workload's enabled status (missing) and its type.

Workload enabled is not present in the configuration recommended by you right ?! Also it didn't make sense to be available there.
The additional configuration is to make sure they supply all these values in the configuration as they'd be required for the deployment

saikumarrs: I meant from `serverConfigOverride`.
```
serverConfigOverride:
  workloadIdentityEnabled: true
workloadIdentity:
  type: GKE # EKS for Amazon support
  ...
```

I disagree that the additional parameters should be validated here. The service consuming them should bother about them.

sanpj2292: These values are being used while creating the rudderstack deployment. Any insight into this would be good
It was provided as a suggestion by you to keep it enabled by default right ?! 
Why is `workloadIdentityEnabled` is being put in `serverConfigOverride` ?

sanpj2292: Removed the whole thing, kindly check and let me know @saikumarrs

utils/googleutils/googleutils.go
```diff
@@ -3,12 +3,40 @@ package googleutils
...
+	sent as empty string or when the field is set with "{}"
+
+	Note: This is true only for workload identity enabled rudderstack data-plane deployments
+*/
+func IsCredentialsStringEmpty(credentials string) bool {
+	return (credentials == "" || credentials == EMPTY_CREDS)
+}
+
+/*
+	We would check for rudder-server configuration for workload identity for google destinations
```
atzoum, line 37: In go function documentation should start with the function's name

```suggestion
	IsGKEEnabledWorkload  checks against rudder-server configuration to find if workload identity for google destinations is enabled
```

services/streammanager/bqstream/bqstreammanager.go
```diff
@@ -57,7 +57,7 @@ func init() {
 
 func NewProducer(destinationConfig interface{}, o Opts) (*Client, error) {
 	var config Config
-	var credentialsFile Credentials
+	// var credentialsFile Credentials
```
atzoum, line 60:
```suggestion
```

utils/googleutils/googleutils.go
```diff
@@ -3,12 +3,40 @@ package googleutils
...
+*/
+func IsGKEEnabledWorkload() bool {
+	workloadType := config.GetString(fmt.Sprintf("%s.type", WI_CONFIG_KEY), "")
```
atzoum, line 40: If `workloadIdentity.type` has a value, is it possible to have any other value than `GKE`?

sanpj2292: It is possible to have another value called `EKS` but not sure about it as of this moment

utils/googleutils/googleutils.go
```diff
@@ -3,12 +3,40 @@ package googleutils
...
+	return IsGKEEnabledWorkload() && IsCredentialsStringEmpty(credentials)
+}
+
+/*
+	The credentials are deemed to be empty when either the field credentials is
+	sent as empty string or when the field is set with "{}"
+
+	Note: This is true only for workload identity enabled rudderstack data-plane deployments
+*/
+func IsCredentialsStringEmpty(credentials string) bool {
```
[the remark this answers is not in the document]
sanpj2292: Done

utils/googleutils/googleutils.go
atzoum: I was under the assumption that workload identity was a GKE thing only. If not and there are differences in workload identity configuration between GKE & EKS fine, otherwise a boolean config option would be more appropriate, e.g. `config.GetBool("WorkloadIdentityEnabled", false)`

sanpj2292: > If not and there are differences in workload identity configuration between GKE & EKS fine

Yes, there is a difference in workload identity configuration between GKE and EKS

atzoum (dismissed):
LGTM

atzoum (commented):
will need to review this again

atzoum (changes requested):

utils/googleutils/googleutils.go
```diff
@@ -3,12 +3,41 @@ package googleutils
...
+}
+
+/*
+IsCredentialsStringEmpty checks for empty credentials.
+The credentials are deemed to be empty when either the field credentials is
+sent as empty string or when the field is set with "{}"
+
+Note: This is true only for workload identity enabled rudderstack data-plane deployments
+*/
+func IsCredentialsStringEmpty(credentials string) bool {
```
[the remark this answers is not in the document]
sanpj2292: Done

utils/googleutils/googleutils.go
```diff
@@ -3,12 +3,41 @@ package googleutils
...
+	return (credentials == "" || credentials == EMPTY_CREDS)
+}
+
+/*
+	IsGKEEnabledWorkload  checks against rudder-server configuration to find if workload identity for google destinations is enabled
+*/
+func IsGKEEnabledWorkload() bool {
```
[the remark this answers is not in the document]
sanpj2292: Done

services/streammanager/bqstream/bqstreammanager.go
```diff
@@ -66,24 +65,18 @@ func NewProducer(destinationConfig interface{}, o Opts) (*Client, error) {
...
-	if config.Credentials == "" {
-		return nil, createErr(err, "credentials not being sent")
-	}
-	if err = googleutils.CompatibleGoogleCredentialsJSON([]byte(config.Credentials)); err != nil {
-		return nil, createErr(err, "incompatible credentials")
-	}
-	confCreds = []byte(config.Credentials)
-	err = json.Unmarshal(confCreds, &credentialsFile)
-	if err != nil {
-		return nil, createErr(err, "error in BQStream while unmarshalling credentials json")
```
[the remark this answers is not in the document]
sanpj2292: Basically the client would throw an error by throwing an error indicating what are the missing fields(one at a time)
Added the test-case by putting invalid credentials. Kindly take a look at it and let me know

utils/googleutils/googleutils.go
saikumarrs: Why do you want to defer a single method name change? I'd defer it if it results in several changes in the code.

services/streammanager/bqstream/bqstreammanager.go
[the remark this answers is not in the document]
saikumarrs: Your response to @atzoum's question is not clear to me. If the user provides the `client_credentials.json` file and we skip this check, would that be caught during the new producer state? If yes, I agree that we should remove it.

saikumarrs (approved):

utils/googleutils/googleutils.go
sanpj2292: This PR is related to changes being made for GKE or workload identity and refactoring is being done only on the logic involved with it but this seems more like clean-up  wherein we are not changing anything related to that method

services/streammanager/bqstream/bqstreammanager.go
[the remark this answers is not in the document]
sanpj2292: If the user provides a `client_credentials.json` in credentials field, that would be passed into `config.Credentials` object. The method `!googleutils.ShouldSkipCredentialsInit(config.Credentials)` would return `true` as we have an input containing credentials json. The schema of the json provided would be validated inside the `bigquery.NewClient`. Hence it is ok to remove the unmarshalling logic

Also the unmarshalling only validates if the credentials passed would be the service account credentials json and not the other types of credentials by the google SDK client

Please let me know if it makes sense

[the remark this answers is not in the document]
atzoum: thank you @sanpj2292 for creating a test case for this! Could you please delete the Credentials struct from line 26 as well? Nobody seems to be using it now

atzoum (approved):
thanks for addressing the comments :)

services/streammanager/bqstream/bqstreammanager.go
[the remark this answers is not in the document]
sanpj2292: Done

utils/googleutils/googleutils.go
saikumarrs: Ok. Raise another PR then.

PR #2157, opened 2022-07-12
chore(warehouse): refactor db interactions

abhimanyubabbar:
# Description

We need to refactor our interactions with the database. At this moment, the domain logic is directly interacting with the database queries. This needs to be refactored for better separation of concerns and ability to test / refactor in a better way.

The purpose of this change is to kickoff the refactoring by creating a separate db module and drive interactions through the function.

## Notion Ticket
https://www.notion.so/rudderstacks/Refactor-warehouse-database-interactions-through-a-separate-module-e2bdb18ec7d149769839c34f1baf8086

## Security

### DIFF
--- a/warehouse/db.go
+++ b/warehouse/db.go
@@ -14,23 +14,18 @@ type DB struct {
 	handle *sql.DB
 }
 
-func New(conn string) (*DB, error) {
-	handle, err := sql.Open("psql", conn)
-	if err != nil {
-		return nil, fmt.Errorf("unable to open connection to underlying database, err: %w", err)
-	}
-
-	return &DB{handle}, nil
+func NewWarehouseDB(handle *sql.DB) *DB {
+	return &DB{handle}
 }
 
-func (db *DB) GetLatestUploadStatus(ctx context.Context, destType, sourceID, destinationID string) (int64, string, int64, error) {
+func (db *DB) GetLatestUploadStatus(ctx context.Context, destType, sourceID, destinationID string) (int64, string, int, error) {
 	query := fmt.Sprintf(`SELECT id, status, COALESCE(metadata->>'priority', '100')::int FROM %[1]s WHERE %[1]s.destination_type='%[2]s' AND %[1]s.source_id='%[3]s' AND %[1]s.destination_id='%[4]s' ORDER BY id DESC LIMIT 1`, warehouseutils.WarehouseUploadsTable,
 		destType, sourceID, destinationID)
 
 	var (
 		uploadID int64
 		status   string
-		priority int64
+		priority int
 	)
 
 	err := db.handle.QueryRowContext(ctx, query).Scan(&uploadID, &status, &priority)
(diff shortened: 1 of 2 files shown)


warehouse/db.go
```diff
@@ -0,0 +1,50 @@
...
+	"context"
+	"database/sql"
+	"fmt"
+
+	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+)
+
+// DB encapsulate interactions of warehouse operations
+// with the database.
+type DB struct {
```
achettyiitr, line 13: Can we expose an interface here instead of calling it from a struct? wdyt
Later we can easily mock the interface.

warehouse/warehouse.go
```diff
@@ -116,6 +116,7 @@ type HandleT struct {
 	destType                          string
 	warehouses                        []warehouseutils.WarehouseT
 	dbHandle                          *sql.DB
+	warehouseDBHandle                 *DB
```
achettyiitr, line 119: Going forward we will not be having dbHandle?

warehouse/db.go
```diff
@@ -0,0 +1,50 @@
...
+	handle *sql.DB
+}
+
+func NewWarehouseDB(handle *sql.DB) *DB {
```
achettyiitr, line 17: Can we expose an interface here instead of calling it from a struct? wdyt
Later we can easily mock the interface.

warehouse/warehouse.go
achettyiitr, line 119: Going forward we will not be having dbHandle?

achettyiitr: If yes, Can we add some comments here? Since there are now 2 handles so some confusion will be there.

achettyiitr (changes requested):

warehouse/db.go
achettyiitr: Also, Can we move this into different package? Because since we will be moving all queries and these files becomes big?

achettyiitr (changes requested):

warehouse/db.go
abhimanyubabbar: Good point about the interface. The reason I didn't add the interface was because it's a starting point and for interface addition, we might need to break the queries further.

Similar point about a separate module. Given it grows beyond a certain point and many supporting structs are needed, it might be good to get it into a separate module.

warehouse/warehouse.go
abhimanyubabbar: Yes, going forward, the domain services need not have access to db Handle. from past experience we might get use case for having db handle but I see it getting wrapped up in warehouseDB going forward completely. I'll add a comment on the same.

warehouse/db.go
achettyiitr: Thanks @abhimanyubabbar for sharing your thoughs.

achettyiitr (approved):
LGTM

Jayachand (approved):
LGTM

PR #2177, opened 2022-07-17
chore(warehouse): added stats for rudder missing datatypes in warehouse schema

Jayachand:
# Description

> Stat to capture rudder missing datatypes
> Currently these datatype properties are omitted by rudderstack

## Notion Ticket

https://www.notion.so/rudderstacks/Stat-to-capture-missing-datatypes-in-warehouse-schema-ce93cdac8fe94f4491da994fdfb32d77

## Security


warehouse/azure-synapse/azure-synapse.go
```diff
@@ -783,6 +783,8 @@ func (as *HandleT) FetchSchema(warehouse warehouseutils.WarehouseT) (schema ware
 		}
 		if datatype, ok := mssqlDataTypesMapToRudder[cType]; ok {
 			schema[tName][cName] = datatype
+		} else {
+			warehouseutils.WHCounterStat(warehouseutils.RUDDER_MISSING_DATATYPE, &as.Warehouse, warehouseutils.Tag{Name: "datatype", Value: cType}).Count(1)
```
achettyiitr, line 787: Should we consider moving this to stats.go

warehouse/utils/utils.go
```diff
@@ -784,6 +789,20 @@ func NewCounterStat(name string, extraTags ...Tag) stats.RudderStats {
 	return stats.NewTaggedStat(name, stats.CountType, tags)
 }
 
+func WHCounterStat(name string, warehouse *WarehouseT, extraTags ...Tag) stats.RudderStats {
```
achettyiitr, line 792: Should we consider moving this to stats.go instead of creating it in utits.go

warehouse/utils/utils.go
```diff
@@ -784,6 +789,20 @@ func NewCounterStat(name string, extraTags ...Tag) stats.RudderStats {
...
+	tags := map[string]string{
+		"module":      WAREHOUSE,
+		"destType":    warehouse.Type,
+		"destID":      warehouse.Destination.ID,
+		"whNamespace": warehouse.Namespace,
```
achettyiitr, line 797: Can we cross verify whNamespace because in the alerts we are using namespace.
https://github.com/rudderlabs/rudderstack-operator/blob/4581315f2e65c71747ee8d868dc92e48e27c46b1/helm-charts/rudderstack-monitor/charts/kapacitor/alerts/warehouse-missing-datatype.yaml#L14

achettyiitr (changes requested):

warehouse/utils/utils.go
Jayachand: This `namespace` is kubernetes namespace. Added by operator by default.

warehouse/utils/utils.go
achettyiitr: let's have this for now because it now has a dependency on utils. Ideally, this should reside in a separate package where we can other functionalities for this struct only.

achettyiitr (approved):
LGTM

abhimanyubabbar (approved):
LGTM 👍🏼

PR #2187, opened 2022-07-19
chore(router): skip reporting and sending internal failures to Config backend

cisse21:
# Description

Rudder Server has a few internal status codes ( `599` and `1113`) which are not relevant to the end user. This PR aims to skip reporting such failures to the observability platforms

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=3fac0b34e4384d2092a7c1dcf149b2d9)

## Security


router/router.go
```diff
@@ -761,7 +761,7 @@ func (worker *workerT) processDestinationJobs() {
 				}
 				ch <- struct{}{}
 				timeTaken := time.Since(startedAt)
-				if respStatusCode != types.RouterTimedOutStatusCode {
+				if respStatusCode != types.RouterTimedOutStatusCode && respStatusCode != types.RouterUnMarshalErrorCode {
```
atzoum, line 764: Why is it bad to include these in the workspace latency calculation algorithm and it is not bad to include any other >500 response code?

atzoum:
@cisse21 please rebase this branch to resolve conflicts

router/router.go
cisse21: So Internal Status codes would mean that we have not made any attempt to send the Job to the destination, So this metric will reduce the latency value for the workspace giving us a false sense of faster destination

atzoum: ok, thanks. I was under the impression that `RouterTimedOutStatusCode` would increase it, but maybe I don't have a good understanding yet of under which conditions we fail with a `RouterTimedOutStatusCode`

cisse21: So whenever we pick some jobs from `router` we assign a time after which all those jobs should have been attempted. If we encounter any jobs from that batch which haven't been attempted yet we give this error `RouterTimedOutStatusCode` and simple drop the job

atzoum (approved):

PR #2188, opened 2022-07-20
feat(core): secure customer s3 bucket access for backups using cross-account role

psrikanth88:
# Description
If `BACKUP_IAM_ROLE_ARN` env is specified, sts creds are generated assuming the role for backups to customer bucket.  `BACKUP_IAM_ROLE_ARN` is the ARN of the role the customer creates with a trust policy principal being our AWS acccount and allowing necessary S3 actions on their bucket

More details: https://rudderlabs.slack.com/archives/C03KSBL7JM7/p1657156293344469

## Notion Ticket
https://www.notion.so/rudderstacks/Secure-connection-to-customer-S3-buckets-for-backups-47fa67c7cd1c40fc9edbdf9118a39703

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -3215,7 +3215,7 @@ func (jd *HandleT) getBackUpQuery(backupDSRange *dataSetRangeT, isJobStatusTable
 	if jd.BackupSettings.FailedOnly {
 		// check failed and aborted state, order the output based on destination, job_id, exec_time
 		stmt = fmt.Sprintf(
-			`SELECT 
+			`SELECT
 				json_build_object(
 					'job_id', failed_jobs.job_id,
 					'workspace_id',failed_jobs.workspace_id,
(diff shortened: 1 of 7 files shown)


services/filemanager/s3manager.go
```diff
@@ -190,27 +191,28 @@ func (manager *S3Manager) getSession(ctx context.Context) (*session.Session, err
...
-		}))
+	awsConfig := aws.Config{
+		Region:                        aws.String(region),
+		CredentialsChainVerboseErrors: aws.Bool(true),
+		Endpoint:                      manager.Config.Endpoint,
+		S3ForcePathStyle:              manager.Config.S3ForcePathStyle,
+		DisableSSL:                    manager.Config.DisableSSL, // TODO: get rid of this options as its risky
+	}
+	if manager.Config.AccessKey != "" && manager.Config.AccessKeyID != "" {
+		pkgLogger.Debug("Credentials found in the destination's config")
```
chandumlg, line 204: We can remove this log I feel.

services/filemanager/s3manager.go
```diff
@@ -190,27 +191,28 @@ func (manager *S3Manager) getSession(ctx context.Context) (*session.Session, err
...
+		awsConfig.Credentials = credentials.NewStaticCredentials(manager.Config.AccessKeyID, manager.Config.AccessKey, "")
+	} else if manager.Config.IAMRoleARN != "" {
+		pkgLogger.Debugf("Assuming - %s found in the filemanager config", manager.Config.IAMRoleARN)
```
chandumlg, line 207: Remove this log as well? (we are logging a part of config)

services/filemanager/filemanager.go
```diff
@@ -106,6 +110,12 @@ func GetProviderConfigFromEnv() map[string]interface{} {
 		providerConfig["accessKeyID"] = config.GetEnv("AWS_ACCESS_KEY_ID", "")
 		providerConfig["accessKey"] = config.GetEnv("AWS_SECRET_ACCESS_KEY", "")
 		providerConfig["enableSSE"] = config.GetEnvAsBool("AWS_ENABLE_SSE", false)
+		providerConfig["iamRoleArn"] = config.GetEnv("BACKUP_IAM_ROLE_ARN", "")
+		if providerConfig["iamRoleArn"] != "" {
+			// TODO: move this check to startup
+			backendconfig.WaitForConfig(context.TODO())
```
chandumlg, line 116: Need a better place for this and remove todo.

jobsdb/jobsdb.go
```diff
@@ -3215,7 +3215,7 @@ func (jd *HandleT) getBackUpQuery(backupDSRange *dataSetRangeT, isJobStatusTable
 	if jd.BackupSettings.FailedOnly {
 		// check failed and aborted state, order the output based on destination, job_id, exec_time
 		stmt = fmt.Sprintf(
-			`SELECT 
+			`SELECT
```
chandumlg, line 3218: Not using [gofumpt](https://github.com/mvdan/gofumpt)?

chandumlg (changes requested):

services/filemanager/filemanager.go
cisse21: Didn't understand the reason for doing this? 
Why do we try to get workspaceId for empty write key here?`backendconfig.GetWorkspaceIDForWriteKey("")`

services/filemanager/s3manager.go
psrikanth88: Since we had this debug log from earlier, did not remove it as might serve some purpose debugging this workflow

services/filemanager/s3manager.go
psrikanth88: Yes, we are logging the customers ARN but that should not be a secret. That should be ok I think, WDYT?

services/filemanager/filemanager.go
psrikanth88: From what I could infer, the only method we have in the `BackendConfig` interface to get the workspaceId without getting the whole config is `GetWorkspaceIDForWriteKey` with an empty write key (see it being used in similar fashion in couple of other places). In case of hosted/multitenant it would return an empty string and that should be ok as this feature would not anyways apply for those deployments in current format (reading from env instead of something like a org settings set by customer on UI)

jobsdb/jobsdb.go
psrikanth88: Tried configuring using gofumpt with gopls too. It still removes the whitesapce as done in the commit

services/filemanager/s3manager.go
```diff
@@ -190,27 +191,28 @@ func (manager *S3Manager) getSession(ctx context.Context) (*session.Session, err
...
+		stsSession := session.Must(session.NewSession(&awsConfig))
+		awsConfig.Credentials = stscreds.NewCredentials(stsSession, manager.Config.IAMRoleARN, func(p *stscreds.AssumeRoleProvider) {
+			p.ExternalID = aws.String(manager.Config.ExternalID)
```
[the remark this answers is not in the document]
psrikanth88: added this. thanks!

services/filemanager/filemanager.go
psrikanth88: @chandumlg Can I remove this check completely assuming the failed backups would be retried anyways in case of race condition where backup is triggered before fetching config

chandumlg: We are not fetching the config for every attempt. So, retries would fail too because of the old config.

jobsdb/jobsdb.go
```diff
@@ -3315,70 +3315,70 @@ func (jd *HandleT) getBackUpQuery(backupDSRange *dataSetRangeT, isJobStatusTable
...
 					*
-				FROM 
-					"%[1]s" 
-				ORDER BY 
-					job_id ASC 
-				LIMIT 
-					%[2]d 
-				OFFSET 
+				FROM
+					"%[1]s"
```
fracasula, line 3323: Please refrain from doing `"%s"` because DeepSource will complain about it. You can use `%q` without quotes instead.

services/filemanager/s3manager.go
```diff
@@ -190,27 +193,27 @@ func (manager *S3Manager) getSession(ctx context.Context) (*session.Session, err
...
+			p.RoleSessionName = crossAccountSTSRoleSessionName
+		})
 	}
+
+	sess = session.Must(session.NewSession(&awsConfig))
+
 	return sess, nil
```
fracasula, line 217: Given that this method can return errors, wouldn't it make more sense to avoid panicking with must and just propagate the error up?

services/filemanager/filemanager.go
fracasula: @chandumlg doing a `GetWorkspaceIDForWriteKey` here is cool in my opinion. This doesn't seem the place where we should "wait for config". I think the polling with `WaitForConfig()` should be done upstream where we want to bootstrap things (e.g. `main.go` or in the app handlers/dynamic.go, etc...). The BackendConfig will keep on polling on the background even after `WaitForConfig()` returns.

In my head it should work like this:
1. when initializing the service the first time, make sure all is bootstrapped and we got a valid configuration and the polling has started. we wait for a valid config and then say that the service is healthy and proceed with serving requests
2. when needed we can call BackendConfig methods like `GetWorkspaceIDForWriteKey` which will read what we have in memory at the time via a mutex

We have to make sure that we reach `2` only if we were able to start with a valid configuration. Does it make sense?

chandumlg:
Closing this and opening another one.

services/filemanager/s3manager.go
chandumlg: https://github.com/rudderlabs/rudder-server/pull/2251

jobsdb/jobsdb.go
chandumlg: https://github.com/rudderlabs/rudder-server/pull/2252

PR #2194, opened 2022-07-21
feat: get configuration for all workspaces in a namespace

lvrach:
# Description

The motivation behind this PR was to introduce a new backend config implementation that uses the newly created namespace endpoint. However, many fixes, as well as refactoring, took place as well. 

About workspace namespace: https://www.notion.so/rudderstacks/New-config-endpoints-2b357f233c9f4a9c99dd56b9136a2e1f

Most of the issues come up while testing:

* if the gateway were to shutdown before receiving config, it would panic 😮 
* Shutdown timeout in rudder-server would occur even if the shutdown was successful, making debugging harder. Fix by a [small refactor](https://github.com/rudderlabs/rudder-server/pull/2194/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R281), should be more readable as well.
* etcd server would shutdown before rudder-server shutdown, polluting the logs with irrelevant errors
* etcd docker image `bitnami/etcd:3.4` didn't run properly on macOS M1, I switch back to `bitnami/etcd:3` which uses `3.5`. 

Others were legacy/hacks:
* For instance, initialising backend config with a token instead of a workspaceID. Check if workspace == token to skip the initial fetch.
* Skipping fetch if there are no workspaces no longer makes sense since namespace endpoint will bring everything either way.
* Using `IsConfigured` method to check if the `SetUp` was successful. Remove the `IsConfigured` method and  `SetUp`  returns an error.

## namespace config
The implementation can be found [here](config/backend-config/namespace_config.go). I started by copying the multitenant file, but then I did a lot of refactoring on top of it.

## concerns

My main concern is around the three environment variables that are need for authorisation with control plane:

```
CONTROL_PLANE_BASIC_AUTH_USERNAME
CONTROL_PLANE_BASIC_AUTH_PASSWORD 
HOSTED_MULTITENANT_SERVICE_SECRET
```

Ideally, we should use only one. The basic auth is not really useful for dataplane<->control-plane communication.

## Notion Ticket

[Task](https://www.notion.so/rudderstacks/Fetch-backend-config-with-cluster-filter-9e837c1b2f354df7be15d6224558bb42)

## Security


fracasula:
Superseded by https://github.com/rudderlabs/rudder-server/pull/2232

PR #2195, opened 2022-07-21
fix: clean up structures on config backend stop

chandumlg:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/config/backend-config/backend-config.go
+++ b/config/backend-config/backend-config.go
@@ -62,7 +62,6 @@ type CommonBackendConfig struct {
 	ctx                      context.Context
 	cancel                   context.CancelFunc
 	blockChan                chan struct{}
-	waitForConfigErrs        chan error
 	initializedLock          sync.RWMutex
 	initialized              bool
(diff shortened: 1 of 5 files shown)


config/backend-config/backend-config.go
```diff
@@ -164,6 +166,14 @@ func (bc *CommonBackendConfig) configUpdate(ctx context.Context, statConfigBacke
 }
 
 func (bc *CommonBackendConfig) pollConfigUpdate(ctx context.Context, workspaces string) {
+	bc.pollConfigUpdatefunclock.Lock()
```
lvrach, line 169: `pollConfigUpdate` should not be called twice in the first place, but it makes sense to make the system more robust either way.

I would suggest [sync.Once](https://pkg.go.dev/sync#Once) as a more elegant way to do this.

chandumlg: Yeah. Was in a hurry. Will clean it up.

PR #2197, opened 2022-07-25
refactor: use path pkg in s3 manager to make s3 filename

koladilip:
This avoids manually checking and if conditions.
Mainly started with a small change to get an idea about the process.

# Description
Improves S3 file name generation logic.

## Notion Ticket
This minor refactor, no particular bug fix, do we still need a notion ticket?

## Security


fracasula (commented):
Can you add a test to verify this change?

koladilip:
> Can you add a test to verify this change?

Added and refactored others too.

fracasula (approved):

saikumarrs (approved):

PR #2200, opened 2022-07-25
fix: backend config refactoring

fracasula:
# Description

This PR is mostly to fix the Backend Config `WaitForConfig()` logic and to make MultiTenant work when there are no workspaces. We'll be relying on the proxy for now to do the routing. RudderServer will not be aware of what workspaces can and should serve.

## Additional changes
* `WaitForConfig` shouldn't return an error and it can now be called from multiple routines without unwanted side effects
* Gateway `Start()` and `Setup()` now return an error
* Same with all the implementation of lifecycle interface, now `Setup()` and `Start()` return an error
* Removed error from `WaitForConfig()`, now `backendConfig.Setup()` is supposed to return an error if you start with an invalid configuration
* Better error management in MT mode with dynamic and backend config
* Removed backendConfig `IsConfigured`

## Security


raman2805 (commented):
LGTM

app/apphandlers/processorAppHandler.go
```diff
@@ -272,8 +273,7 @@ func startHealthWebHandler(ctx context.Context) error {
 	// Port where Processor health handler is running
 	pkgLogger.Infof("Starting in %d", webPort)
 	srvMux := mux.NewRouter()
-	srvMux.HandleFunc("/health", healthHandler)
-	srvMux.HandleFunc("/", healthHandler)
```
cisse21, line 276: I am not 100% sure but I guess we use this end point at few places too for alerting. Can we not remove this ?

app/app.go
```diff
@@ -115,15 +115,15 @@ func (a *App) Stop() {
...
+		_ = a.cpuprofileOutput.Close()
 	}
 
 	if a.options.Memprofile != "" {
 		f, err := os.Create(a.options.Memprofile)
 		if err != nil {
 			panic(err)
 		}
-		defer f.Close()
+		defer func() { _ = f.Close() }()
```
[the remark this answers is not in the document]
fracasula: To make it explicit for linters that we want to ignore the returned error.

processor/manager.go
```diff
@@ -35,30 +34,37 @@ type LifecycleManager struct {
...
 	currentCtx, cancel := context.WithCancel(context.Background())
 	proc.currentCancel = cancel
 
-	g, ctx := errgroup.WithContext(currentCtx)
-	proc.waitGroup = g
-	g.Go(func() error {
-		proc.HandleT.Start(ctx)
-		return nil
-	})
+	var wg sync.WaitGroup
```
cisse21, line 50: Why would we need a errgroup or a waitgroup when we are not waiting for the function to return? Can't we just get rid of the wg here?

fracasula: Have you checked the rest of the changes in this file? Look at the next line (i.e. 51) where I do `proc.waitGroup = &wg`. The `Wait()` is used in the `Stop()` method, I didn't want to change the existing logic and waiting when stopping makes sense to me.

cisse21: Somehow I missed this.. Thanks for the explaining

cisse21 (approved):

lvrach (approved):

PR #2205, opened 2022-07-26
fix(gateway): make gateway user logic synonymous to proxy logic

cisse21:
# Description

Right now the way we store `userID` in the Gateway table is `<anonIDFromPayload><DELIMITER><userIDFromPayload>`. For cases where `anonIDFromPayload` is empty, the request is routed by considering `userIDFromPayload` as `anonID` in proxy [here](https://github.com/rudderlabs/rudder-multinode-proxy/blob/master/lib/mylibrary.lua#L212). This is causing an issue in the migrator which always migrates the data based on `anonIDFromPayload` in the `gw` table. To avoid this we try to store it in the same way the request is routed at proxy

## Notion Ticket

[Notion Link ](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=0e6640505fd44f399fc2c0cf6c2711dc&pm=s)

## Security


gateway/gateway.go
```diff
@@ -510,7 +510,12 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 				anonIDFromReq := strings.TrimSpace(vjson.Get("anonymousId").String())
 				userIDFromReq := strings.TrimSpace(vjson.Get("userId").String())
 				if builtUserID == "" {
```
BonapartePC, line 512: Will there ever be a case when it won't enter into this "if" condition?

cisse21: We initialise `builtUserID` with `""` so this if is always entered

BonapartePC: Does it makes sense to remove it to make it cleaner?

cisse21: It wont enter the if case when each payload has multiple events and the userId is already built

BonapartePC (approved):

gateway/gateway.go
```diff
@@ -510,7 +510,12 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
...
-					builtUserID = anonIDFromReq + DELIMITER + userIDFromReq
+					if anonIDFromReq != "" {
+						builtUserID = anonIDFromReq + DELIMITER + userIDFromReq
+					} else {
+						// Proxy gets the userID from body if there is no anonymousId in body/header
+						builtUserID = userIDFromReq + DELIMITER + anonIDFromReq
```
atzoum, line 517: Why are we using `anonIDFromReq` as the second part? Isn't using `userIDFromReq` in place of `anonIDFromReq` the only change that we need?
```suggestion
						builtUserID = userIDFromReq + DELIMITER + userIDFromReq
```

cisse21: Yeah true... we are not concerned about what the second value is in most of the places

atzoum: does the migrator only parse the 1st part?

cisse21: Yes,we use this function to determine which partition the user belongs to `pg_catalog.PartitionIndex(split_part("jobs"."user_id", '<<>>', 1)`  in the SQL query

atzoum: Don't we need to introduce an SQL migration script as well so that UserIDs of existing events in jobsdb tables are renamed accordingly? Otherwise we might experience ordering inconsistencies, between old and new events. 
What about backed-up data? Won't we observe ordering inconsistencies if we replay half events with the old format and half events with the new format? Is it worth to address this as well?

atzoum: As an alternative, the migrator could use the first non-empty part of the composite key, instead of always using the 1st part, wdyt?

cisse21: I guess we wouldn't need to have migration scripts to address this, Also this would take a bit of time before we migrate enterprise customers into the migrator setup and we back up only 1-month of data so I guess data would get clean by then and for pro customers, we do support complete event replay and only replay errors. So I guess event ordering would be lost there by default

atzoum: So after deploying the new rudder-server version containing this change, we accept the fact that events with no anonymousId which already existed in jobsdb tables might be processed along with new events arriving to the server, in mixed order, due to this change, correct?

BonapartePC (approved):

atzoum (approved):

PR #2213, opened 2022-07-27
fix: fetch filemanagers timeout from config instead of registering

chandumlg:
# Description

fetch filemanagers timeout from config instead of registering

## Notion Ticket

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=ccb681c3df8246c3a67db35289d015ce&pm=s

## Security


atzoum:
@chandumlg can you please provide some more context in the description, e.g. what is `dispose`, when it is called, why we need to deregister config vars, etc?

config/config.go
```diff
@@ -480,6 +504,29 @@ func RegisterInt64ConfigVariable(defaultValue int64, ptr *int64, isHotReloadable
...
 
+func DeregisterConfigVariableWithPtr(ptr interface{}) {
+	configVarLock.Lock()
+	defer configVarLock.Unlock()
+
+	for key, configVars := range hotReloadableConfig {
+		for _, configvar := range configVars {
+			if configvar.value == ptr {
+				delete(hotReloadableConfig, key)
+				break
```
cisse21, line 515: Why do we break here instead of returning? Do we have a case where a pointer is both hotreloadable and not hotreloadable at the same time?

cisse21: Also, we are deleting the whole key... Shouldn't we just get rid of the pointer requested to be deleted?

cisse21 (approved):

services/filemanager/filemanager.go
```diff
@@ -44,6 +44,7 @@ type FileManager interface {
 	ListFilesWithPrefix(ctx context.Context, prefix string, maxItems int64) (fileObjects []*FileObject, err error)
 	GetConfiguredPrefix() string
 	SetTimeout(timeout *time.Duration)
+	Dispose()
```
atzoum, line 47: `Close` is a more common method name for performing necessary cleanup operations before disposing an object

jobsdb/jobsdb.go
```diff
@@ -3557,6 +3557,9 @@ func (jd *HandleT) backupUploadWithExponentialBackoff(ctx context.Context, file
 	if err != nil {
 		return filemanager.UploadOutput{}, err
 	}
+
+	defer fileUploader.Dispose()
```
atzoum, line 3561:
```suggestion
```
`jobsdb.HandleT` is supposed to have a `jobsFileUploader` field for caching a single instance of a `filemanager.FileManager`. Disposing a cached fileUploader every time this method gets called is not desirable.

Btw, after looking at `getFileUploader()` method, I believe that it needs some changes... e.g.
```go
// getFileUploader get a file uploader
func (jd *HandleT) getFileUploader() (filemanager.FileManager, error) {
	var err error
	if jd.jobsFileUploader == nil {
		jd.jobsFileUploader, err = filemanager.DefaultFileManagerFactory.New(&filemanager.SettingsT{
			Provider: config.GetEnv("JOBS_BACKUP_STORAGE_PROVIDER", "S3"),
			Config:   filemanager.GetProviderConfigFromEnv(),
		})
	}
	return jd.jobsFileUploader, err
}
```

config/config.go
```diff
@@ -32,6 +32,30 @@ type ConfigVar struct {
...
+
+	return total
+}
+
+func GetTotalNonHotReloadableConfigVars() int {
+	configVarLock.RLock()
+	defer configVarLock.RUnlock()
+
+	total := 0
+	for _, configVars := range nonHotReloadableConfig {
```
atzoum, line 52: It seems to me that the only reason why we retain pointers to non-hot-reloadable config options is for doing `exit(1)` when `RESTART_ON_CONFIG_CHANGE=true` and a configuration update affects some non hot reloadable config option pointer.

It also seems to me that `RESTART_ON_CONFIG_CHANGE` is a feature that we don't use, ever.

Based on the above, I would propose to change the scope of this PR a bit: 
Remove `nonHotReloadableConfig`, drop support for `RESTART_ON_CONFIG_CHANGE` and we won't have to worry about properly closing our `FileManager`s anymore :)

atzoum (changes requested):

config/config.go
cisse21: We introduced this `RESTART_ON_CONFIG_CHANGE` because previously there were cases when people changed the config and assumed it would reload automatically, So to avoid the confusion we introduced this change but as you said this env isn't passed now via operator

atzoum: If we need to retain the `RESTART_ON_CONFIG_CHANGE` feature we don't need pointers, just the config keys of registered non-hot-reloadable options. But I don't see an actual need for retaining this feature

config/config.go
chandumlg: Based on the fact that a variable can be either hotreloadable or hotreloadable and not both, return makes sense. Changed.

chandumlg: There is no point in keeping the config var in the map anymore after deregistering. So deleting it.

services/filemanager/filemanager.go
chandumlg: done

config/config.go
chandumlg: Agreed we don't need to save pointers. 

Only other advantage I can think of nonHotReloadableConfig map is, if we ever want to inspect what values the variables are currently set to (say via rudder-cli), we can just print the value pointer are pointing to.

We can debate on this and take up in another pr. For now, resolving this. :)

jobsdb/jobsdb.go
chandumlg: Good catch. Fixed.

config/config.go
atzoum: Keeping the map but replacing the pointer with an empty struct should be enough to simplify this and we wouldn't need to introduce a `Close` method in filemanager anymore. You might have introduced `Close` where necessary now, but any further usage of filemanager should also deal with closing it. A design where you don't having to remember to use close before disposing the filemanager will always better than a design where you have.
That being said, I would debate that such a simplification is desirable & safe, thus we should do it in the context of this PR already.
If everybody else feels differently though, I agree to take it up in another PR :)

lvrach:
I would like some further explanation about this PR. What is the motivation behind this ?

config/config.go
chandumlg: Just realised, timeout variables were not hotreloadable. 

It is desirable to make them hotreloadable. But instead of adding Close overhead, fetching the value from config will make things simple.

chandumlg:
> I would like some further explanation about this PR. What is the motivation behind this ?

https://rudderlabs.slack.com/archives/C01HTT66UMB/p1659070330047979

config/config.go
atzoum: During testing, there occasions where we want to override the timeout manually, e.g. [here](https://github.com/rudderlabs/rudder-server/blob/e242c7d5d94ecc903f2565639a1b43b4a3d31ffa/warehouse/configuration_testing/helper.go#L62). Therefore, config value should be used only if timeout is not manually set.

services/filemanager/filemanager.go
```diff
@@ -135,9 +123,11 @@ func GetProviderConfigFromEnv() map[string]interface{} {
...
 
-func getSafeTimeout(timeout *time.Duration) time.Duration {
-	if timeout == nil || *timeout == 0 {
-		return time.Second * 120
+func getBatchRouterDurationConfig(key, destType string, defaultValueInTimescaleUnits int64, timeScale time.Duration) time.Duration {
+	destOverrideFound := config.IsSet("BatchRouter." + destType + "." + key)
+	if destOverrideFound {
+		return config.GetDuration("BatchRouter."+destType+"."+key, defaultValueInTimescaleUnits, timeScale)
+	} else {
+		return config.GetDuration("BatchRouter."+key, defaultValueInTimescaleUnits, timeScale)
```
atzoum, line 131: This function accepts a lot of arguments that don't actually change. We could use this instead
```suggestion
func getTimeout(destType string) time.Duration {
	key := fmt.Sprintf("BatchRouter.%s.timeout", destType)
	destOverrideFound := config.IsSet(key)
	if destOverrideFound {
		return config.GetDuration(key, 120, time.Second)
	} else {
		return config.GetDuration("BatchRouter.timeout", 120, time.Second)
	}
}
```

chandumlg: The only problem with this is different managers can't have different defaults.

atzoum: wasn't aware we had such a requirement, but we can still include the default `time.Duration` in the function's arguments :)

atzoum (approved):

PR #2214, opened 2022-07-28
feat: introduce retention for datasets

chandumlg:
# Description

With the new data retention policy, the goal is to restrict the life of the successfully processed jobs to maxDSRetention period(3hrs default). This is achieved by 
a. Rotating datasets every 90 mins (90 mins because, we dont compact the right two datasets. After 3 hrs, there will be three datasets and the left most will be eligible for compaction)
b. Migration logic is updated to allow compaction

## Notion Ticket

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=f3f5ae7f42e64036b819b38a70680f52&pm=s

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -396,7 +396,8 @@ type HandleT struct {
 	datasetRangeList              []dataSetRangeT
 	dsListLock                    lock.DSListLocker
 	dsMigrationLock               sync.RWMutex
-	dsRetentionPeriod             time.Duration
+	MinDSRetentionPeriod          time.Duration
+	MaxDSRetentionPeriod          time.Duration
 	dsEmptyResultCache            map[dataSetT]map[string]map[string]map[string]map[string]cacheEntry // DS -> workspace -> customVal -> params -> state -> cacheEntry
 	dsCacheLock                   sync.Mutex
 	BackupSettings                *backupSettings
@@ -434,9 +435,10 @@ type HandleT struct {
 	// TODO: Remove this flag once we have test setup that uses real database
 	skipSetupDBSetup bool
 
-	// TriggerAddNewDS is useful for triggering addNewDS to run from tests.
+	// TriggerAddNewDS, TriggerMigrateDS is useful for triggering addNewDS to run from tests.
 	// TODO: Ideally we should refactor the code to not use this override.
-	TriggerAddNewDS func() <-chan time.Time
+	TriggerAddNewDS  func() <-chan time.Time
+	TriggerMigrateDS func() <-chan time.Time
 
 	lifecycle struct {
 		mu      sync.Mutex
@@ -705,12 +707,6 @@ func WithClearDB(clearDB bool) OptsFunc {
 	}
 }
 
-func WithRetention(period time.Duration) OptsFunc {
-	return func(jd *HandleT) {
-		jd.dsRetentionPeriod = period
-	}
-}
-
 func WithQueryFilterKeys(filters QueryFiltersT) OptsFunc {
 	return func(jd *HandleT) {
 		jd.queryFilterKeys = filters
@@ -757,7 +753,6 @@ func newOwnerType(ownerType OwnerType, tablePrefix string, opts ...OptsFunc) *Ha
 			migrationMode: "",
 			importLock:    &sync.RWMutex{},
 		},
-		dsRetentionPeriod: 0,
 	}
 
 	for _, fn := range opts {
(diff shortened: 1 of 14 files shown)


jobsdb/jobsdb.go
```diff
@@ -1170,17 +1168,18 @@ func (jd *HandleT) checkIfMigrateDS(ds dataSetT) (bool, int) {
...
-	sqlStatement = fmt.Sprintf(`SELECT MAX(created_at) from "%s"`, ds.JobTable)
+	var minCreatedAt, maxCreatedAt time.Time
+	sqlStatement = fmt.Sprintf(`SELECT MIN(created_at), MAX(created_at) from "%s"`, ds.JobTable)
 	row = jd.dbHandle.QueryRow(sqlStatement)
-	err = row.Scan(&lastUpdate)
+	err = row.Scan(&minCreatedAt, &maxCreatedAt)
 	jd.assertError(err)
-	if jd.dsRetentionPeriod > time.Duration(0) && time.Since(lastUpdate) < jd.dsRetentionPeriod {
+
+	if jd.minDSRetentionPeriod > time.Duration(0) && time.Since(maxCreatedAt) < jd.minDSRetentionPeriod {
```
cisse21, line 1177: Why do we need these checks? Just to protect ourselves from a bad input?
```
jd.minDSRetentionPeriod > time.Duration(0)
jd.maxDSRetentionPeriod > time.Duration(0)
```

jobsdb/jobsdb.go
```diff
@@ -1210,6 +1209,17 @@ func (jd *HandleT) getTableSize(jobTable string) int64 {
 }
 
 func (jd *HandleT) checkIfFullDS(ds dataSetT) bool {
+	var minJobCreatedAt time.Time
+	sqlStatement := fmt.Sprintf(`SELECT created_at FROM "%s" ORDER BY "job_id" ASC LIMIT 1`, ds.JobTable)
```
cisse21, line 1213: Can't we use the same query to get the min id? Probably add a helper function to get min and max?

`sqlStatement = fmt.Sprintf(`SELECT MIN(created_at), MAX(created_at) from "%s"`, ds.JobTable)`

jobsdb/jobsdb.go
atzoum, line 1177: Is the order of precedence correct? Is `minDSRetentionPeriod` more powerful than `maxDSRetentionPeriod`?

atzoum:
@chandumlg it should be possible add some test scenarios by using a small value, e.g. `maxDSRetentionPeriod=1ms`

- Scenario 1 (addNewDSLoop): create a jobsdb, add 1 job to it, wait a bit and make sure a new dataset was created
- Scenario 2 (migrateDSLoop): create a jobsdb with a MaxDSSize of 2, add 2 jobs to it, trigger addNewDSLoop to create a new DS, complete 1 job out of two, verify that migration was run and this 1 completed job is no longer present in the migrated dataset.

atzoum (commented):
@chandumlg it should be possible add some test scenarios by using a small value, e.g. `maxDSRetentionPeriod=1ms`

- Scenario 1 (addNewDSLoop): create a jobsdb, add 1 job to it, wait a bit and make sure a new dataset was created
- Scenario 2 (migrateDSLoop): create a jobsdb with a MaxDSSize of 2, add 2 jobs to it, trigger addNewDSLoop to create a new DS, complete 1 job out of two, verify that migration was run and this 1 completed job is no longer present in the migrated dataset.

jobsdb/jobsdb.go
chandumlg: Yes

chandumlg: Giving priority to retaining first. So checking `minDSRetentionPeriod` first.

jobsdb/jobsdb.go
```diff
@@ -774,14 +768,13 @@ Setup is used to initialize the HandleT structure.
 clearAll = True means it will remove all existing tables
 tablePrefix must be unique and is used to separate
 multiple users of JobsDB
-dsRetentionPeriod = A DS is not deleted if it has some activity
+minDSRetentionPeriod = A DS is not deleted if it has some activity
```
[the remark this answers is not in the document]
chandumlg: done

jobsdb/jobsdb.go
chandumlg: done

cisse21 (approved):

chandumlg:
Tests added.

jobsdb/jobsdb.go
```diff
@@ -1058,7 +1061,7 @@ Function to return an ordered list of datasets and datasetRanges
 Most callers use the in-memory list of dataset and datasetRanges
 Caller must have the dsListLock readlocked
 */
-func (jd *HandleT) getDSList() []dataSetT {
+func (jd *HandleT) GetDSList() []dataSetT {
```
atzoum, line 1064: Not that comfortable with exporting this method, let alone when it has an unexported return type.

What about creating a `HandleInspector` in `jobsdb` package, e.g.

```
type HandleInspector struct {
	*HandleT
}

func (h *HandleInspector) DSListSize() int {
	return len(h.HandleT.getDSList())
}

```

Then you could use this inspector from your tests, e.g.

```
jobDBInspector := jobsdb.HandleInspector{&jobDB}
require.Equal(t, (2, jobDBInspector.DSListSize())
```

chandumlg: I wasn't comfortable too exposing that function. But din't have an option. The above wrapping is neat. Will make the change.

chandumlg: done

atzoum (approved):

PR #2221, opened 2022-07-31
fix(warehouse): use correct config for destination name

achettyiitr:
# Description
Use Correct Destination Name when we set for configs. Currently, we are using DisplayName which seems to be the correct candidate.

## Notion Ticket

https://www.notion.so/rudderstacks/use-correct-config-for-destination-name-c1b1f44ab93c4def9c9811af97095ca1

## Security


abhimanyubabbar (approved):
LGTM 👍🏼 :shipit:

warehouse/utils/utils_test.go
```diff
@@ -18,6 +19,20 @@ import (
...
 )
 
+func TestDestinationConfigKeys(t *testing.T) {
+	for _, whType := range WarehouseDestinations {
+		t.Run(whType, func(t *testing.T) {
+			require.Contains(t, WHDestNameMap, whType)
+
+			whName := WHDestNameMap[whType]
+			configKey := fmt.Sprintf("Warehouse.%s.feature", whName)
+			got := config.TransformKey(configKey)
```
Jayachand, line 29: This unit test is testing only func transformKey. Looks redundant to apply to WHDestNameMap.

Jayachand (approved):

PR #2225, opened 2022-08-01
fix(core): cache backend config and block processor

Sidddddarth:
# Description
1. A new routine that subscribes to config, and stores it to database(after encrypting).
2. An endless loop in proc-transformer if it encounters a particular response from the transformer.
3. Return to normal operation once control plane is back up.

## Notion Ticket

[Pause Processor](https://www.notion.so/rudderstacks/Pause-processor-pipeline-in-case-of-control-plane-unavailability-f0434be38cf9449eb1e7990010b9711b)

## Security


config/backend-config/config_cache.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json"
+
+	"github.com/rudderlabs/rudder-server/jobsdb"
+)
+
+type Cache interface {
```
atzoum, line 12: we need test coverage for this new component

config/backend-config/internal/cache/cache.go
```diff
@@ -0,0 +1,139 @@
...
+// Encrypt and store the config to the database
+func (db *cacheStore) set(ctx context.Context, config interface{}) error {
+	configBytes, err := json.Marshal(config)
+	if err != nil {
+		return err
+	}
+	// write to config table
+	_, err = db.ExecContext(
+		ctx,
+		`INSERT INTO config_cache (key, config) VALUES ($1, pgp_sym_encrypt($2, $3))
```
atzoum, line 72: at some point we need to delete old/previous keys. Or only keep the latest key in the table

config/backend-config/internal/cache/cache.go
```diff
@@ -0,0 +1,139 @@
...
+		pkgLogger.Errorf("failed to open db: %v", err)
+		return nil, err
+	}
+	err = db.Ping()
+	if err != nil {
+		pkgLogger.Errorf("failed to ping db: %v", err)
+		return nil, err
+	}
+	// create table if it doesn't exist
+	_, err = db.Exec(`CREATE TABLE IF NOT EXISTS config_cache (
```
atzoum, line 122: this could be performed using golang-migrate for simplifying cache's code

config/backend-config/backend-config.go
```diff
@@ -164,8 +202,9 @@ func (bc *backendConfigImpl) configUpdate(ctx context.Context, statConfigBackend
 
 func (bc *backendConfigImpl) pollConfigUpdate(ctx context.Context, workspaces string) {
 	statConfigBackendError := stats.DefaultStats.NewStat("config_backend.errors", stats.CountType)
+	statCachedConfigGauge := stats.DefaultStats.NewStat("config_from_cache", stats.GaugeType)
```
atzoum, line 205: we need a new (recoverable) alert defined for this metric

config/backend-config/backend_config_test.go
```diff
@@ -84,8 +95,86 @@ var sampleBackendConfig2 = ConfigT{
...
+
+func TestMain(m *testing.M) {
+	flag.BoolVar(&hold, "hold", false, "hold environment clean-up after test execution until Ctrl+C is provided")
+	flag.Parse()
+
+	// hack to make defer work, without being affected by the os.Exit in TestMain
+	os.Exit(run(m))
+}
+
+func run(m *testing.M) int {
```
atzoum, line 111: for tests it would be great if we had a way to disable cache, or use a mock cache so that we don't need to have a real database as a prerequisite for every test scenario.

config/backend-config/internal/cache/cache.go
```diff
@@ -0,0 +1,173 @@
...
+	// if err != nil {
+	// 	pkgLogger.Errorf("failed to create table: %v", err)
+	// 	return nil, err
+	// }
+	// // create encryption extension pgcrypto
+	// _, err = db.Exec(`CREATE EXTENSION IF NOT EXISTS "pgcrypto"`)
+	// if err != nil {
+	// 	pkgLogger.Errorf("failed to create extension: %v", err)
+	// 	return nil, err
+	// }
```
atzoum, line 154:
```suggestion
```

processor/transformer/transformer.go
```diff
@@ -296,72 +299,43 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
-			if convErr != nil {
-				transformerAPIVersion = 0
+	// endless retry if transformer-controlplane connection is down
+	outerBackoff := backoff.NewExponentialBackOff()
+	outerBackoff.MaxElapsedTime = 0
+	// endless backoff loop, only nil error or panics inside
+	_ = backoff.RetryNotify(
+		func() error {
+			respData, statusCode = trans.doPost(ctx, rawJSON, url, statsTags(data[0]))
+			if statusCode == StatusCPDown {
```
atzoum, line 319: another metric with a relevant alert would be valuable here too

config/backend-config/internal/cache/cache.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// apply migrations
+	err = migrate(dbConn)
+	if err != nil {
+		pkgLogger.Errorf("failed to migrate db: %v", err)
+		return &dbStore
+	}
+
+	// clear config for other keys
+	err = dbStore.clear(ctx)
```
atzoum, line 60: Is it enough to clear other keys during startup? Can a long running multi-tenant node create multiple different keys and entries in the cache during its lifetime?

atzoum (commented):
Overall looks good! 
Some improvement comments, with the alerts' & cache entry proliferation ones being more important

config/backend-config/internal/cache/cache.go
Sidddddarth: It can
but every time there's a workspace change
this functions is called again
so it'll cklear all the keys during the server's lifetime.
Didn't really find it necessary to have another routine that clears the table up.

processor/transformer/transformer.go
```diff
@@ -140,6 +144,7 @@ func (trans *HandleT) Setup() {
 	trans.logger = pkgLogger
 	trans.sentStat = stats.DefaultStats.NewStat("processor.transformer_sent", stats.CountType)
 	trans.receivedStat = stats.DefaultStats.NewStat("processor.transformer_received", stats.CountType)
+	trans.cpDownGauge = stats.DefaultStats.NewStat("processor.control_plane_down_gauge", stats.GaugeType)
```
atzoum, line 147:
```suggestion
	trans.cpDownGauge = stats.DefaultStats.NewStat("processor.control_plane_down", stats.GaugeType)
```

processor/transformer/transformer.go
```diff
@@ -296,72 +301,46 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
-		break
-	}
+			return nil
+		},
+		endlessBackoff,
+		func(err error, t time.Duration) {
+			trans.logger.Errorf("JS HTTP connection error: URL: %v Error: %+v", url, err)
+		})
+	// control plane back up
+	trans.cpDownGauge.Gauge(0)
```
atzoum, line 332: can't we move this to before line 325?

processor/transformer/transformer.go
atzoum, line 147: Please prepare the relevant operator changes for the 2 new alerts that we are introducing

processor/transformer/transformer.go
```diff
@@ -296,72 +301,46 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
-			transformerAPIVersion, convErr := strconv.Atoi(resp.Header.Get("apiVersion"))
-			if convErr != nil {
-				transformerAPIVersion = 0
+	// endless retry if transformer-controlplane connection is down
+	endlessBackoff := backoff.NewExponentialBackOff()
+	endlessBackoff.MaxElapsedTime = 0 // no max time -> ends only when no error
+	// endless backoff loop, only nil error or panics inside
+	_ = backoff.RetryNotify(
+		func() error {
+			respData, statusCode = trans.doPost(ctx, rawJSON, url, statsTags(data[0]))
```
atzoum, line 320: **Note:** we need to make sure that transformer uses this status code in its response and doesn't "hide" it inside the json response, masking it in a HTTP 200 response, i.e. using the `statusCode` field of transformed messages.

This functionality also needs to be tested using the relevant transformer release that includes the new status code.

config/backend-config/backend-config.go
```diff
@@ -42,8 +45,19 @@ var (
 	pkgLogger            = logger.NewLogger().Child("backend-config")
 	IoUtil               = sysUtils.NewIoUtil()
 	Diagnostics          diagnostics.DiagnosticsI
+	cacheOverride        cache.Cache
 )
 
+func setCacheOverride() {
```
atzoum, line 51:
```suggestion
func disableCache() {
```

config/backend-config/backend-config.go
```diff
@@ -262,6 +318,12 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
 	bc.ctx = ctx
 	bc.cancel = cancel
 	bc.blockChan = make(chan struct{})
+	if cacheOverride == nil {
+		cacheKey := fmt.Sprintf(`%x`, sha1.Sum([]byte(workspaces))) // using a fixed size key	// skipcq: GSC-G401, GO-S1025
+		bc.cache = cache.Start(ctx, bc.AccessToken(), cacheKey, bc.Subscribe(ctx, TopicBackendConfig))
+	} else {
+		bc.cache = cacheOverride
+	}
```
atzoum, line 326: minor
```suggestion
    bc.cache = cacheOverride
	if bc.cache == nil {
		cacheKey := fmt.Sprintf(`%x`, sha1.Sum([]byte(workspaces))) // using a fixed size key	// skipcq: GSC-G401, GO-S1025
		bc.cache = cache.Start(ctx, bc.AccessToken(), cacheKey, bc.Subscribe(ctx, TopicBackendConfig))
	}
```

atzoum (approved):
LGTM, it is important to perform an e2e test as soon as the new transformer version is ready

utils/misc/dbutils.go
```diff
@@ -3,32 +3,51 @@ package misc
...
+	// Application Name can be any string of less than NAMEDATALEN characters (64 characters in a standard PostgreSQL build).
+	// There is no need to truncate the string on our own though since PostgreSQL auto-truncates this identifier and issues a relevant notice if necessary.
+	appName = DefaultString("rudder-server").OnError(os.Hostname())
+}
+
+// GetConnectionString Returns Jobs DB connection configuration
+func GetConnectionString() string {
+	once.Do(func() {
+		loadConfig()
+	})
```
lvrach, line 37: The usage of once here might make testing harder, depending on how this function is used. What was the intention of its usage here?

config/backend-config/backend-config.go
```diff
@@ -262,6 +318,11 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
 	bc.ctx = ctx
 	bc.cancel = cancel
 	bc.blockChan = make(chan struct{})
+	bc.cache = cacheOverride
+	if bc.cache == nil {
+		cacheKey := fmt.Sprintf(`%x`, sha1.Sum([]byte(workspaces))) // using a fixed size key	// skipcq: GSC-G401, GO-S1025
```
lvrach, line 323: Why not use a sha256 ?

config/backend-config/backend_config_test.go
```diff
@@ -340,9 +363,167 @@ func TestWaitForConfig(t *testing.T) {
...
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		defer atomic.AddInt32(&calls, 1)
+		t.Log("Server got called")
+		w.WriteHeader(http.StatusBadRequest)
+	}))
+	defer server.Close()
+
+	// parsedURL, err := url.Parse(server.URL)
+	_, err := url.Parse(server.URL)
+	require.NoError(t, err)
```
lvrach, line 379:
```suggestion
```

This is probably not needed in your case.

utils/misc/dbutils.go
```diff
@@ -3,32 +3,51 @@ package misc
...
+	once                                           sync.Once
 )
 
 func loadConfig() {
 	host = config.GetEnv("JOBS_DB_HOST", "localhost")
 	user = config.GetEnv("JOBS_DB_USER", "ubuntu")
+	dbname = config.GetEnv("JOBS_DB_DB_NAME", "ubuntu")
 	port, _ = strconv.Atoi(config.GetEnv("JOBS_DB_PORT", "5432"))
 	password = config.GetEnv("JOBS_DB_PASSWORD", "ubuntu") // Reading secrets from
 	sslmode = config.GetEnv("JOBS_DB_SSL_MODE", "disable")
```
lvrach, line 27: We should consider dropping (in a backwards-compatible way) the `JOBS` prefix since we are using it for multiple functionalities of rudder-server.

processor/transformer/transformer_test.go
```diff
@@ -128,3 +129,87 @@ func Test_Transformer(t *testing.T) {
 		require.Equal(t, expectedResponse, rsp)
 	}
 }
+
+func Test_endlessLoopIf809(t *testing.T) {
```
lvrach, line 133: [super minor :P ]
```suggestion
func Test_EndlessLoopIf809(t *testing.T) {
```

processor/transformer/transformer_test.go
```diff
@@ -128,3 +129,87 @@ func Test_Transformer(t *testing.T) {
...
+		},
+		rsp,
+	)
+	require.Equal(t, ft.retryCount, ft.maxRetryCount)
+}
+
+type endlessLoopTransformer struct {
+	retryCount    int
+	maxRetryCount int
+}
```
lvrach, line 186:
```suggestion
type endlessLoopTransformer struct {
	retryCount    int
	maxRetryCount int
	t *testing.T
}
```

processor/transformer/transformer_test.go
```diff
@@ -128,3 +129,87 @@ func Test_Transformer(t *testing.T) {
...
+				Metadata:   reqBody[i].Metadata,
+				StatusCode: statusCode,
+				Error:      "",
+			}
+		}
+	}
+	w.Header().Set("apiVersion", "2")
+	if err := json.NewEncoder(w).Encode(resps); err != nil {
+		panic(err)
+	}
```
lvrach, line 214:
```suggestion
	require.NoError(t,  json.NewEncoder(w).Encode(resps))
```

processor/transformer/transformer.go
```diff
@@ -375,16 +354,68 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
+			}
+			if reqErr != nil {
+				return reqErr
+			}
+
+			return nil
+		},
+		backoff.WithMaxRetries(backoff.NewExponentialBackOff(), uint64(maxRetry)),
+		func(err error, t time.Duration) {
+			trans.logger.Errorf("JS HTTP connection error: URL: %v Error: %+v", url, err)
```
lvrach, line 400: Given we keep retrying, let's make this a Warning

processor/transformer/transformer.go
```diff
@@ -375,16 +354,68 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
+		if convErr != nil {
+			transformerAPIVersion = 0
+		}
+		if types.SUPPORTED_TRANSFORMER_API_VERSION != transformerAPIVersion {
+			trans.logger.Errorf("Incompatible transformer version: Expected: %d Received: %d, URL: %v", types.SUPPORTED_TRANSFORMER_API_VERSION, transformerAPIVersion, url)
+			panic(fmt.Errorf("Incompatible transformer version: Expected: %d Received: %d, URL: %v", types.SUPPORTED_TRANSFORMER_API_VERSION, transformerAPIVersion, url))
+		}
+	}
+	if retryCount > 0 {
+		trans.logger.Errorf("Failed request succeeded after %v retries, URL: %v", retryCount, url)
```
lvrach, line 418: I would put that as a warning or even info. Maybe it is redundant since we already log it when retrying

processor/transformer/transformer.go
```diff
@@ -375,16 +354,68 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
+
+			if reqErr == nil {
+				// If no err returned by client.Post, reading body.
+				// If reading body fails, retrying.
+				respData, reqErr = io.ReadAll(resp.Body)
+				resp.Body.Close()
+			}
+			if reqErr != nil {
+				return reqErr
+			}
```
lvrach, line 394: I find this way a bit [more readable](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88):

```suggestion
                        
			if reqErr != nil {
				return reqErr
			}
			defer resp.Body.Close()

			// If no err returned by client.Post, reading body.
			// If reading body fails, retrying.
			respData, reqErr = io.ReadAll(resp.Body)
							
```

processor/transformer/transformer.go
```diff
@@ -375,16 +354,68 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
+
+			return nil
+		},
+		backoff.WithMaxRetries(backoff.NewExponentialBackOff(), uint64(maxRetry)),
+		func(err error, t time.Duration) {
+			trans.logger.Errorf("JS HTTP connection error: URL: %v Error: %+v", url, err)
+			retryCount++
+		})
+	if err != nil {
+		panic(err)
```
lvrach, line 404: Why do we panic here instead of propagating the error? Assume transformation has an issue It should be handled in the same way if we get a `StatusCPDown`.

processor/transformer/transformer.go
```diff
@@ -296,72 +301,46 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
-	if !(resp.StatusCode == http.StatusOK ||
-		resp.StatusCode == http.StatusBadRequest ||
-		resp.StatusCode == http.StatusNotFound ||
-		resp.StatusCode == http.StatusRequestEntityTooLarge) {
-		trans.logger.Errorf("Transformer returned status code: %v", resp.StatusCode)
+	if !(statusCode == http.StatusOK ||
+		statusCode == http.StatusBadRequest ||
+		statusCode == http.StatusNotFound ||
+		statusCode == http.StatusRequestEntityTooLarge) {
+		trans.logger.Errorf("Transformer returned status code: %v", statusCode)
```
lvrach, line 339: It can be simplified by using a switch case:

```suggestion
        switch statusCode {
        case http.StatusOK, http.StatusBadRequest, http.StatusNotFound,http.StatusRequestEntityTooLarge:
		trans.logger.Errorf("Transformer returned status code: %v", statusCode)
```
It is a matter of preference ofc

processor/transformer/transformer.go
```diff
@@ -375,16 +354,68 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
 	return transformerResponses
 }
+
+func (trans *HandleT) doPost(ctx context.Context, rawJSON []byte, url string, tags stats.Tags) ([]byte, int) {
+	var (
+		retryCount int
+		resp       *http.Response
+		respData   []byte
+	)
+	err := backoff.RetryNotify(
```
lvrach, line 377: I find the usage of nested retries a bit confusing, and I think it can be avoided if HTTP client errors are handled as transformer unavailability cases. I would also propose to do JSON parsing only if the status code is the expected one.

config/backend-config/internal/cache/cache.go
```diff
@@ -0,0 +1,156 @@
...
+	// write to config table
+	_, err = db.ExecContext(
+		ctx,
+		`INSERT INTO config_cache (key, config) VALUES ($1, pgp_sym_encrypt($2, $3))
+		on conflict (key)
+		do update set
+		config = pgp_sym_encrypt($2, $3),
+		updated_at = NOW()`,
+		db.key,
+		configBytes,
```
lvrach, line 95: It might be slightly better if the encryption was happening client side:
1. The connection to DB is not happening on a secure channel (SSL is not used at the moment); an network ease dropper could get access to configs
2. It would put less pressure on DB doing the compute here, although sym encryption is relatively light.
3. If a bug appears in the encryption function is easier to upgrade the code than the DB.

config/backend-config/internal/cache/cache.go
```diff
@@ -0,0 +1,156 @@
...
+		db.secret,
+	)
+	return err
+}
+
+// Fetch the cached config when needed
+func (db *cacheStore) Get(ctx context.Context) ([]byte, error) {
```
lvrach, line 102: I would consider creating a simple CLI script around this method, allowing debugging by checking what is stored in DB. 

Alternatively, we can have a set of instructions to do it manually.

lvrach (changes requested):
I like the approach overall. 

One blocking concern is the encryption of configs in DB.

The transformer code can also get simplified, I believe. Plus some minor stuff

config/backend-config/backend-config.go
atzoum: we don't really care about security here, what we really care about is to have a cache key of fixed size, since the list of workspaces can be pretty big. Think about git commit hashes, they are sha1 too and it is ok

processor/transformer/transformer.go
atzoum: you are missing a default case statement where the log statement should belong though, right? :)

processor/transformer/transformer.go
atzoum: They are retrying for different purposes: transformer unavailability, client errors, issues reading the body are handled by the finite retry mechanism, whereas special error code indicating that backend config is unavailable (not transformer) are handled by the other, infinite retry mechanism.
What could be changed?

processor/transformer/transformer.go
atzoum: the proposal is to retry endlessly regardless of the type of the problem?

processor/transformer/transformer.go
atzoum: Will require a repetition of the following at the end:
````
if reqErr != nil {
	return reqErr
}
```

processor/transformer/transformer.go
Sidddddarth: Agreed.
The panic wasn't introduced now, but just left in place as is to not introduce too many changes in behaviour at once.

processor/transformer/transformer_test.go
Sidddddarth: nice

config/backend-config/internal/cache/cache.go
Sidddddarth: one suggestion was to use [AES Encryption/Decryption in GoLang](https://golangdocs.com/aes-encryption-decryption-in-golang) to encrypt the payload in the server and then persist to database.
Any other approaches you have in mind @lvrach ?
cc @atzoum

Sidddddarth:
pertinent transformer [PR](https://github.com/rudderlabs/rudder-transformer/pull/1415)

config/backend-config/backend-config.go
lvrach: I understand, still using 256 would not hurt. btw git is moving to sha256.  

I would suggest against using multiple `workspaceIDs` and thus the need for hashing. 

Given how fast the set of workspaceIDs might change in some environments, we would end up with a lot of unneeded entries, plus increasing the chances of downtime in-between workspaceID changes.

I would propose using: `{deploymentType}/{Identifier}`
* `dedicated/{workspaceID}`
* `namespace/{namespaceName}`
 
This way our key will be the same as long as environment configurations do not change. Some workspaces might be missing if downtime happens before the new workspaceID is fetched but thus is better than not having any configs.

lvrach (changes requested):
I have proposed a change for the cache key to better work on busy namespace environments.

config/backend-config/backend-config.go
Sidddddarth: >we would end up with a lot of unneeded entries

every time `workspaces` changes, the previous entries are cleared.
>This way our key will be the same

And if we want to make the key immutable for a server's lifetime, we don't even need one imo. Just one column(`config`) would suffice for this purpose. We could avoid using `sha#`  entirely.

atzoum: we will be using the Identifier.ID as a cache key (workspaceID for single-tenant/namespace for multi-tenant)

utils/misc/dbutils.go
atzoum: doing it in another PR here:
https://github.com/rudderlabs/rudder-server/blob/f55e59f0e92857601235ab702d447a749a1d3341/config/config.go#L271-L279

config/backend-config/backend-config.go
```diff
@@ -92,59 +108,116 @@ func Init() {
...
+// 		noOfDestinations := 0
+// 		for _, source := range curConfig.Sources {
+// 			noOfDestinations += len(source.Destinations)
+// 		}
+// 		Diagnostics.Track(diagnostics.ConfigProcessed, map[string]interface{}{
+// 			diagnostics.SourcesCount:      noOfSources,
+// 			diagnostics.DesitanationCount: noOfDestinations,
+// 		})
+// 	}
+// }
```
atzoum, line 131: why was this commented out?

config/backend-config/backend-config.go
```diff
@@ -92,59 +108,116 @@ func Init() {
...
+
+func filterProcessorEnabledDestinations(configToFilter map[string]ConfigT) map[string]ConfigT {
```
atzoum, line 133: please revisit according to original logic

config/backend-config/backend-config.go
```diff
@@ -255,6 +333,29 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
...
+		// to ensure that they cache key is diffrent from the password.
+		//
+		// If the workspaceToken is known, the key and secret can be generated from it.
+		// However, knowning the key can not give you the password or the workspaceToken.
+		//
+		// A cryptographycally secure one way hash function is need to ensure the above property.
+		//
+		u, _ := bc.Identity().BasicAuth()
+		secret := sha256.Sum256([]byte(u + "_secret"))
+		cacheKey := fmt.Sprintf(`%x`, sha256.Sum256([]byte(u+"_key")))
```
atzoum, line 351: It is ok to even have an empty key for single workspace deployments, see [here](https://github.com/rudderlabs/rudder-server/blob/f4838faefe83376fc9f935f80095fb749652b5e5/services/validators/envValidator.go#L260)
```suggestion
		cacheKey := bc.Identity().ID()
```

config/backend-config/backend-config.go
```diff
@@ -255,6 +333,29 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
...
+		// Thus we are using the workspaceToken with a salt,
+		// to ensure that they cache key is diffrent from the password.
+		//
+		// If the workspaceToken is known, the key and secret can be generated from it.
+		// However, knowning the key can not give you the password or the workspaceToken.
+		//
+		// A cryptographycally secure one way hash function is need to ensure the above property.
+		//
+		u, _ := bc.Identity().BasicAuth()
+		secret := sha256.Sum256([]byte(u + "_secret"))
```
atzoum, line 350:
```suggestion
		secret := sha256.Sum256([]byte(u))
```

services/transientsource/transientsource.go
```diff
@@ -90,13 +90,15 @@ func (r *service) updateLoop(ctx context.Context, config backendconfig.BackendCo
...
-		newSourceIds := transientSourceIds(&c)
-		r.sourceIds = newSourceIds
-		r.sourceIdsMap = asMap(newSourceIds)
-		r.onceInit.Do(func() {
-			close(r.init)
-		})
+		configs := ev.Data.(map[string]backendconfig.ConfigT)
+		for _, c := range configs {
+			newSourceIds := transientSourceIds(&c)
+			r.sourceIds = newSourceIds
```
atzoum, line 96: we should be appending, not replacing

config/backend-config/single_workspace.go
```diff
@@ -124,21 +97,17 @@ func (wc *singleWorkspaceConfig) getFromAPI(ctx context.Context, _ string) (map[
 	}
 	workspaceID := sourcesJSON.WorkspaceID
 
-	wc.workspaceIDLock.Lock()
-	wc.workspaceID = workspaceID
-	wc.workspaceIDToLibrariesMap = make(map[string]LibrariesT)
-	wc.workspaceIDToLibrariesMap[workspaceID] = sourcesJSON.Libraries
-	wc.workspaceIDLock.Unlock()
+	wc.doOnce.Do(func() {
+		wc.workspaceID = workspaceID
```
atzoum, line 101: `workspaceID` will not be initialised properly here in case of a cached configuration. We need to take care of that too.
One way would be to "override" `BackendConfig.Identity` and if we are `usingCache`handle and the `Identifier.ID` is empty we should intervene by providing an augmented `identity.Identifier`.

services/filemanager/filemanager.go
```diff
@@ -102,7 +102,7 @@ func GetProviderConfigForBackupsFromEnv(ctx context.Context) map[string]interfac
 		providerConfig["iamRoleArn"] = config.GetString("BACKUP_IAM_ROLE_ARN", "")
 		if providerConfig["iamRoleArn"] != "" {
 			backendconfig.DefaultBackendConfig.WaitForConfig(ctx)
-			providerConfig["externalId"] = backendconfig.DefaultBackendConfig.GetWorkspaceIDForWriteKey("")
+			providerConfig["externalId"] = backendconfig.DefaultBackendConfig.Identity().ID()
```
lvrach, line 105: Note: As discussed, we can not use a namespace for `externalId`. This is of course outside the scope of this PR, but we need a reminder to support externId in multitenant environments.

lvrach (approved):
Nice job! It passed through multiple iterations, but the final result is worth it.

PR #2229, opened 2022-08-02
feat(destination): onboarding Lambda destination

sandeepdsvs:
# Description

Added support for new custom destination AWS Lambda

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/AWS-Lambda-cloud-mode-destination-1fcb663d84f046388191d6531c428a56)

## Security


services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type Opts struct {
+	Timeout time.Duration
+}
+
+var pkgLogger logger.LoggerI
+
+func init() {
+	pkgLogger = logger.NewLogger().Child("streammanager").Child("firehose")
```
utsabc, line 31: `firehose` is incorrect here

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+	var s *session.Session
+	httpClient := &http.Client{
+		Timeout: o.Timeout,
+	}
+
+	if config.AccessKeyID == "" || config.AccessKey == "" {
+		s = session.Must(session.NewSession(&aws.Config{
+			Region:     aws.String(config.Region),
+			HTTPClient: httpClient,
+		}))
```
utsabc, line 54: without `AccessKeyId` and `Accesskey` will the `lambdaClient` work?

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+	} else {
+		s = session.Must(session.NewSession(&aws.Config{
+			HTTPClient:  httpClient,
+			Region:      aws.String(config.Region),
```
utsabc, line 58: If `Region` is empty string or `nil` how are we handling it?

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+var pkgLogger logger.LoggerI
+
+func init() {
+	pkgLogger = logger.NewLogger().Child("streammanager").Child("firehose")
+}
+
+// NewProducer creates a producer based on destination config
+func NewProducer(destinationConfig interface{}, o Opts) (lambda.Lambda, error) {
```
utsabc, line 35: This function return signature should have the pointer to the client check other `streamManger` `newProducer` code they are doing same

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+	var config Config
+	jsonConfig, err := json.Marshal(destinationConfig)
+	if err != nil {
+		return lambda.Lambda{}, fmt.Errorf("[Lambda] Error while marshalling destination config :: %w", err)
+	}
+	err = json.Unmarshal(jsonConfig, &config)
+	if err != nil {
+		return lambda.Lambda{}, fmt.Errorf("[Lambda] error  :: error in Lambda while unmarshelling destination config:: %w", err)
```
utsabc, line 43:
```suggestion
		return nil, fmt.Errorf("[Lambda] error  :: error in Lambda while unmarshelling destination config:: %w", err)
```

utsabc (changes requested):

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Rohit from server team informed me that there are customers who directly give access to our pod and neccessary keys will be picked from the environment variables. If this doesn't work, error will be thrown while invoking the lambda function.

utsabc: Ok

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+	var lambdaClient *lambda.Lambda = lambda.New(s)
+	return *lambdaClient, nil
+}
+
+// Produce creates a producer and send data to Lambda.
+func Produce(jsonData json.RawMessage, producer, destConfig interface{}) (statusCode int, respStatus, responseMessage string) {
+	client, ok := producer.(lambda.Lambda)
+	if !ok {
+		respStatus = "Failure"
+		responseMessage = "[Lambda] error :: Could not create producer"
```
utsabc, line 71:
```suggestion
		responseMessage = "[Lambda] error :: Could not create client"
```

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+		pkgLogger.Errorf("[Lambda] error :: Could not create producer")
+		return 400, respStatus, responseMessage
+	}
+	parsedJSON := gjson.ParseBytes(jsonData)
+	data := parsedJSON.Get("message").String()
```
utsabc, line 76: There can be a case where `message` is not present in `parsedJSON` hich might cause some issue

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+	//clientContext := parsedJSON.Get("clientContext").String()
+	invocationType := parsedJSON.Get("invocationType").String()
+	lambdaFunction := parsedJSON.Get("lambda").String()
```
utsabc, line 79: Lets handle same here

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+	input.InvocationType = aws.String(invocationType)
+
+	res, err := client.Invoke(&input)
+	if err != nil {
+		statusCode = int(*res.StatusCode)
+		respStatus = "Failure"
+		responseMessage = "[Lambda] error while Invoking the function ::" + err.Error()
+	}
+	statusCode = int(*res.StatusCode)
+	respStatus = "Success"
```
utsabc, line 93: If there is non 2XX statuscode here we would still send `Success` to jobsDb

utsabc (changes requested):

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+
+// NewProducer creates a producer based on destination config
+func NewProducer(destinationConfig interface{}, o Opts) (lambda.Lambda, error) {
+	var config Config
+	jsonConfig, err := json.Marshal(destinationConfig)
+	if err != nil {
+		return lambda.Lambda{}, fmt.Errorf("[Lambda] Error while marshalling destination config :: %w", err)
+	}
+	err = json.Unmarshal(jsonConfig, &config)
+	if err != nil {
```
lvrach, line 42: What is the reason for `Marshal` and then `Unmarshal` the same config?

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+		return lambda.Lambda{}, fmt.Errorf("[Lambda] error  :: error in Lambda while unmarshelling destination config:: %w", err)
+	}
+	var s *session.Session
+	httpClient := &http.Client{
+		Timeout: o.Timeout,
+	}
+
+	if config.AccessKeyID == "" || config.AccessKey == "" {
+		s = session.Must(session.NewSession(&aws.Config{
```
lvrach, line 51:
```suggestion
		s, err = session.NewSession(&aws.Config{
```
It is safer to remove `session.Must`, which will panic if there is an error in creating the new session and do proper error handling here

services/streammanager/lambda/lambdamanager.go
lvrach, line 35:
```suggestion
func NewProducer(destinationConfig interface{}, o Opts) (*lambda.Lambda, error) {
```

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,96 @@
...
+		}))
+	} else {
+		s = session.Must(session.NewSession(&aws.Config{
+			HTTPClient:  httpClient,
+			Region:      aws.String(config.Region),
+			Credentials: credentials.NewStaticCredentials(config.AccessKeyID, config.AccessKey, ""),
+		}))
+	}
+	var lambdaClient *lambda.Lambda = lambda.New(s)
+	return *lambdaClient, nil
```
lvrach, line 63:
```suggestion
	return lambda.New(s), nil
```

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: We are following this in every stream manager destination, not sure why though. Want to know is this any standard convention that we use for any reason or can I update it by using type assertion ?

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Corrected

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Updated

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Updated

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Updated

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Refactored code with validations

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Updated

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Updated

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Updated

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Updated

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Updated

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Updated

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Updated to `mapstructure.Decode`

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/aws/aws-sdk-go/service/lambda"
+	"github.com/mitchellh/mapstructure"
+	"github.com/rudderlabs/rudder-server/services/streammanager/common"
+	"github.com/rudderlabs/rudder-server/utils/awsutils"
+	"github.com/rudderlabs/rudder-server/utils/logger"
+	"github.com/tidwall/gjson"
+)
+
+// Config is the config that is required to send data to Lambda
+type Config struct {
```
atzoum, line 18: this internal structure doesn't need to be exported

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,92 @@
...
+	invokeInput.SetInvocationType("Event") // Setting InvocationType as Event(Asynchronous Invocation)
+	if config.ClientContext != "" {
+		invokeInput.SetClientContext(config.ClientContext)
+	}
+
+	if err = invokeInput.Validate(); err != nil {
+		return simpleErrorResponse("Invalid invokeInput :: " + err.Error())
+	}
+
+	_, err = client.Invoke(&invokeInput)
```
atzoum, line 78: Isn't it necessary to verify that the invocation was indeed successful by looking into the returned `lambda.InvokeOutput` `FunctionError` and `StatusCode` fields?

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		statusCode, respStatus, responseMessage := common.ParseAWSError(err)
+		pkgLogger.Errorf("[Lambda] Invocation error :: %d : %s : %s", statusCode, respStatus, responseMessage)
+		return statusCode, respStatus, responseMessage
+	}
+
+	return 200, "Success", "Event delivered to Lambda :: " + config.Lambda
+}
+
+func simpleErrorResponse(message string) (int, string, string) {
```
cisse21, line 96: Rename the function to `lambdaErrorResponse` ?

cisse21 (approved):

services/streammanager/lambda/lambdamanager.go
atzoum: terminal error too, since a status code `400` means no retry :)

sandeepdsvs: @atzoum This function is called only when we need to send an error response with 400 status code (no retry).

services/streammanager/lambda/lambdamanager.go
```diff
@@ -0,0 +1,100 @@
...
+	destConfig := gjson.GetBytes(jsonData, "destConfig").String()
+	if destConfig == "" {
+		return simpleErrorResponse("Invalid Destination Config")
+	}
+
+	var config Config
+	err := json.Unmarshal([]byte(destConfig), &config)
+	if err != nil {
+		return simpleErrorResponse("error while unmarshalling destination config :: " + err.Error())
+	}
```
lvrach, line 72: `[non-blocking]` I have a few issues with the usage of `gjson`:
1. Performance: Every time we use GetBytes, we have to scan a significant portion of the byte array.
2. Readability: It is hard to understand the structure of the expected payload, you have to scan the code to find all the expected properties. In this case, is not that bad because we only have two properties in the same area, but in other situations, it can become really ugly.

I have a suggestion that maintains the same behaviour. 

NOTE: I am using pointers to check whether `destConfig` is set or not 

Assuming:
```go
type inputEvent struct {
         Payload string `json:"payload"`
         Config *Config  `json:"destConfig"`
}
```

```suggestion
	err := json.Unmarshal([]byte(destConfig), &event)
	if err != nil {
		return simpleErrorResponse("error while unmarshalling :: " + err.Error())
	}
	
	if event.Payload == "" {
			return simpleErrorResponse("Invalid payload")
	}
	if event.Config == nil {
		return simpleErrorResponse("Invalid Destination Config")
	}
```

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: Among the two possible type of errors, Invocation errors & function runtime errors, Invocation error will always have a non nil err which is being handled.
As we've decided to use only asynchronous invocation, lambda won't be returning any runtime errors in the response.

services/streammanager/lambda/lambdamanager.go
lvrach: The package name is already `lambda`, so adding it as a prefix won't be of any benefit.

Unrelated to this PR, I believe we need to introduce custom error types with the following interface:

```go
type DestinationError interface {
    Error() string // for returnMessage
    Status() string
    Retriable() bool 
}

// HTTPDestinationError contains additional information for destinations that use HTTP protocol
type HTTPDestinationError interface {
    DestinationError
    HTTPCode() int
}
```

A single error can be returned in all cases, instead of the triple, plus knowing when to retry or not and decoupling from HTTP status codes.

services/streammanager/lambda/lambdamanager_test.go
```diff
@@ -0,0 +1,154 @@
+package lambda
```
lvrach, line 1: [minor & opinionated] Testing from the perspective of a consumer. Requires exposure of LambdaClient.

```suggestion
package lambda_test
```

lvrach (approved):
Nothing blocking. I only have one concern regarding the performance of `gjson` since it will run on a critical path.

I suggested a small change and usage of fastjson.

atzoum (approved):

utsabc (approved):

services/streammanager/lambda/lambdamanager.go
sandeepdsvs: updated

PR #2230, opened 2022-08-02
fix(core): filter failed events in processor based on response codes

Sidddddarth:
# Description

Filter failed events and simply drop in case of special response codes.
Cases include `GA - server side identify not on` etc.

## Notion Ticket

[Filter identify if serverside identify is not on](https://www.notion.so/rudderstacks/Filter-identify-if-serverside-identify-is-not-on-3e415f20abab41f1a644cfb932ddfb55)

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -951,7 +951,7 @@ func (proc *HandleT) getFailedEventJobs(response transformer.ResponseT, commonMe
 	failedCountMap := make(map[string]int64)
 	var failedEventsToStore []*jobsdb.JobT
 	for _, failedEvent := range response.FailedEvents {
-		if failedEvent.StatusCode == transformer.DROP_STATUS_CODE {
+		if failedEvent.StatusCode == transformer.DropStatusCode {
 			continue
 		}
 		var messages []types.SingularEventT
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -945,13 +945,21 @@ func (proc *HandleT) updateMetricMaps(countMetadataMap map[string]MetricMetadata
 }
 
 func (proc *HandleT) getFailedEventJobs(response transformer.ResponseT, commonMetaData transformer.MetadataT, eventsByMessageID map[string]types.SingularEventWithReceivedAt, stage string, transformationEnabled, trackingPlanEnabled bool) ([]*jobsdb.JobT, []*types.PUReportedMetric, map[string]int64) {
-	failedMetrics := make([]*types.PUReportedMetric, 0)
-	connectionDetailsMap := make(map[string]*types.ConnectionDetails)
-	statusDetailsMap := make(map[string]*types.StatusDetail)
-	failedCountMap := make(map[string]int64)
-	var failedEventsToStore []*jobsdb.JobT
+	var (
+		failedMetrics        = make([]*types.PUReportedMetric, 0)
+		connectionDetailsMap = make(map[string]*types.ConnectionDetails)
+		statusDetailsMap     = make(map[string]*types.StatusDetail)
+		failedCountMap       = make(map[string]int64)
+		failedEventsToStore  []*jobsdb.JobT
+	)
 	for _, failedEvent := range response.FailedEvents {
 		if failedEvent.StatusCode == transformer.DropStatusCode {
+			procDroppedEventsStat := stats.NewTaggedStat("proc_dropped_event_counts", stats.CountType, stats.Tags{
+				"destName":   commonMetaData.DestinationType,
+				"statusCode": strconv.Itoa(failedEvent.StatusCode),
+				"stage":      stage,
+			})
+			procDroppedEventsStat.Increment()
 			continue
 		}
 		var messages []types.SingularEventT
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1765,7 +1765,18 @@ func (proc *HandleT) transformSrcDest(
 				inCountMap[key] = 0
 			}
 			if _, ok := inCountMetadataMap[key]; !ok {
-				inCountMetadataMap[key] = MetricMetadata{sourceID: event.Metadata.SourceID, destinationID: event.Metadata.DestinationID, sourceBatchID: event.Metadata.SourceBatchID, sourceTaskID: event.Metadata.SourceTaskID, sourceTaskRunID: event.Metadata.SourceTaskRunID, sourceJobID: event.Metadata.SourceJobID, sourceJobRunID: event.Metadata.SourceJobRunID, sourceDefinitionID: event.Metadata.SourceDefinitionID, destinationDefinitionID: event.Metadata.DestinationDefinitionID, sourceCategory: event.Metadata.SourceCategory}
+				inCountMetadataMap[key] = MetricMetadata{
+					sourceID:                event.Metadata.SourceID,
+					destinationID:           event.Metadata.DestinationID,
+					sourceBatchID:           event.Metadata.SourceBatchID,
+					sourceTaskID:            event.Metadata.SourceTaskID,
+					sourceTaskRunID:         event.Metadata.SourceTaskRunID,
+					sourceJobID:             event.Metadata.SourceJobID,
+					sourceJobRunID:          event.Metadata.SourceJobRunID,
+					sourceDefinitionID:      event.Metadata.SourceDefinitionID,
+					destinationDefinitionID: event.Metadata.DestinationDefinitionID,
+					sourceCategory:          event.Metadata.SourceCategory,
+				}
 			}
 			inCountMap[key] = inCountMap[key] + 1
 		}
@@ -1880,7 +1891,14 @@ func (proc *HandleT) transformSrcDest(
 			}
 		}
 
-		diffMetrics := getDiffMetrics(inPU, types.EVENT_FILTER, inCountMetadataMap, inCountMap, successCountMap, failedCountMap)
+		diffMetrics := getDiffMetrics(
+			inPU,
+			types.EVENT_FILTER,
+			inCountMetadataMap,
+			inCountMap,
+			successCountMap,
+			failedCountMap,
+		)
 		reportMetrics = append(reportMetrics, successMetrics...)
 		reportMetrics = append(reportMetrics, failedMetrics...)
 		reportMetrics = append(reportMetrics, diffMetrics...)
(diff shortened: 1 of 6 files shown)


processor/processor.go
```diff
@@ -980,6 +980,10 @@ func (proc *HandleT) getFailedEventJobs(response transformer.ResponseT, commonMe
 			failedEvent.StatusCode, commonMetaData.SourceID, commonMetaData.DestinationID, failedEvent.Error,
 		)
 
+		if failedEvent.StatusCode == transformer.DROP_STATUS_CODE {
+			continue
```
atzoum, line 984: Are these events reported by the reporting service after this change? If not, should they?

Sidddddarth: They are.
They shouldn't.👍 
Should be part of the diff metric.

lvrach:
@Sidddddarth would you mind giving a bit more context about this PR? What is the motivation behind this change?

Sidddddarth:
> @Sidddddarth would you mind giving a bit more context about this PR? What is the motivation behind this change?

https://rudderlabs.slack.com/archives/C0218DFG57D/p1657798156457899
https://rudderlabs.slack.com/archives/C02B0A38QQG/p1657721887565259
We're basically trying to add another layer of filtering to "bad" events. Similar to filtering events based on unsupported message types.
The issue in question comes up because we filter out unsupported messages based on destination definition. But the actual destination configuration might differ from that, leading to some events pass along because they're supported(but the destination doesn't accept them).
Example: GA - server side identify can be toggled while configuring the destination. But according to the GA destination definition, identify is a supported message type. In cases where these identify events reach the destination, we're alerted because of the aborts that this causes.
This PR essentially tries to filter out such events. This is done based on the response code that the transformer returns.

router/router.go
```diff
@@ -1119,7 +1121,9 @@ func (worker *workerT) postStatusOnResponseQ(respStatusCode int, respBody string
 
 		if respStatusCode >= 500 {
```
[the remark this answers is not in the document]
Sidddddarth: please take another look

router/router.go
```diff
@@ -1117,9 +1117,22 @@ func (worker *workerT) postStatusOnResponseQ(respStatusCode int, respBody string
 
 		worker.rt.failedEventsChan <- *status
 
-		if respStatusCode >= 500 {
-			timeElapsed := time.Since(firstAttemptedAtTime)
-			if respStatusCode != types.RouterTimedOutStatusCode && respStatusCode != types.RouterUnMarshalErrorCode {
+		switch respStatusCode {
+		case transformer.DROP_STATUS_CODE:
+			status.JobState = jobsdb.Aborted.State
+		case types.RouterTimedOutStatusCode:
```
koladilip, line 1123: case types.RouterTimedOutStatusCode, types.RouterUnMarshalErrorCode:
 			addToFailedMap = false
 			addToAbortMap = false
Thanks for simplifying it.

koladilip (approved):

router/router.go
Sidddddarth: done
Thanks!
TIL

processor/processor.go
```diff
@@ -949,6 +949,9 @@ func (proc *HandleT) getFailedEventJobs(response transformer.ResponseT, commonMe
 	failedCountMap := make(map[string]int64)
 	var failedEventsToStore []*jobsdb.JobT
 	for _, failedEvent := range response.FailedEvents {
+		if failedEvent.StatusCode == transformer.DROP_STATUS_CODE {
+			continue
```
fracasula, line 953: Would it make sense to bump some metric here to keep track of these?

processor/transformer/transformer.go
```diff
@@ -29,6 +29,7 @@ const (
 	EventFilterStage            = "event_filter"
 	DestTransformerStage        = "dest_transformer"
 	TrackingPlanValidationStage = "trackingPlan_validation"
+	DROP_STATUS_CODE            = 722
```
fracasula, line 32:
```suggestion
	DropStatusCode            = 722
```

> See https://go.dev/doc/effective_go#mixed-caps. This applies even when it breaks conventions in other languages. For example an unexported constant is `maxLength` not `MaxLength` or `MAX_LENGTH`.

https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

router/router.go
```diff
@@ -1117,9 +1117,19 @@ func (worker *workerT) postStatusOnResponseQ(respStatusCode int, respBody string
...
-		if respStatusCode >= 500 {
-			timeElapsed := time.Since(firstAttemptedAtTime)
-			if respStatusCode != types.RouterTimedOutStatusCode && respStatusCode != types.RouterUnMarshalErrorCode {
+		switch respStatusCode {
+		case transformer.DROP_STATUS_CODE:
+			status.JobState = jobsdb.Aborted.State
+		case types.RouterTimedOutStatusCode, types.RouterUnMarshalErrorCode:
+			addToFailedMap = false
+			addToAbortMap = false
+		case 429:
```
fracasula, line 1126: [optional] just to make the code more readable.

```suggestion
		case http.StatusTooManyRequests:
```

fracasula (commented):
Can we get some test coverage?

processor/processor.go
Sidddddarth: It would.

router/router.go
```diff
@@ -1505,10 +1521,30 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
...
+		key := fmt.Sprintf(
+			"%s:%s:%s:%s:%s:%s:%s",
+			parameters.SourceID,
+			parameters.DestinationID,
+			parameters.SourceBatchID,
+			resp.status.JobState,
+			resp.status.ErrorCode,
+			eventName,
+			eventType,
+		)
```
lvrach, line 1533: [minor] The use of [string.Join](https://pkg.go.dev/strings#Join) might be a good idea here:

```suggestion
		key := strings.Join(
			parameters.SourceID,
			parameters.DestinationID,
			parameters.SourceBatchID,
			resp.status.JobState,
			resp.status.ErrorCode,
			eventName,
			eventType,
			
			":",
		)
```

router/router.go
```diff
@@ -1505,10 +1521,30 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
...
+				parameters.DestinationID,
+				parameters.SourceBatchID,
+				parameters.SourceTaskID,
+				parameters.SourceTaskRunID,
+				parameters.SourceJobID,
+				parameters.SourceJobRunID,
+				parameters.SourceDefinitionID,
+				parameters.DestinationDefinitionID,
+				parameters.SourceCategory,
+			)
```
lvrach, line 1547: [minor] I think it is safer and more readable to create the struct, rather than use a function given the number of arguments we need to pass.

Sidddddarth: Sure. The same with statusDetail as well.
Let's take this up in another PR because it'll introduce changes in too many places which might cause oversight.

router/router.go
```diff
@@ -1120,9 +1120,19 @@ func (worker *workerT) postStatusOnResponseQ(respStatusCode int, respBody string
 
 		worker.rt.failedEventsChan <- *status
 
-		if respStatusCode >= 500 {
-			timeElapsed := time.Since(firstAttemptedAtTime)
-			if respStatusCode != types.RouterTimedOutStatusCode && respStatusCode != types.RouterUnMarshalErrorCode {
+		switch respStatusCode {
+		case transformer.DropStatusCode:
+			status.JobState = jobsdb.Aborted.State
```
atzoum, line 1125: shouldn't we skip adding the userID to both failed & abort map?

router/router.go
```diff
@@ -1231,7 +1234,12 @@ func (worker *workerT) sendEventDeliveryStat(destinationJobMetadata *types.JobMe
 
 func (*workerT) sendDestinationResponseToConfigBackend(payload json.RawMessage, destinationJobMetadata *types.JobMetadataT, status *jobsdb.JobStatusT, sourceIDs []string) {
 	// Sending destination response to config backend
-	if status.ErrorCode != fmt.Sprint(types.RouterUnMarshalErrorCode) && status.ErrorCode != fmt.Sprint(types.RouterTimedOutStatusCode) {
+	switch status.ErrorCode {
+	// no live event in case of internal errors
+	case strconv.Itoa(transformer.DropStatusCode),
+		strconv.Itoa(types.RouterUnMarshalErrorCode),
+		strconv.Itoa(types.RouterTimedOutStatusCode):
```
atzoum, line 1241:
```suggestion
	case strconv.Itoa(transformer.DropStatusCode),
		strconv.Itoa(types.RouterUnMarshalErrorCode),
		strconv.Itoa(types.RouterTimedOutStatusCode):
```

router/router.go
```diff
@@ -1540,12 +1590,15 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
...
 			sd.Count++
-			rt.MultitenantI.CalculateSuccessFailureCounts(workspaceID, rt.destName, false, true)
 			routerAbortedJobs = append(routerAbortedJobs, resp.JobT)
 			PrepareJobRunIdAbortedEventsMap(resp.JobT.Parameters, jobRunIDAbortedEventsMap)
 			completedJobsList = append(completedJobsList, resp.JobT)
+			routerWorkspaceJobStatusCount[workspaceID] += 1
+			rt.MultitenantI.CalculateSuccessFailureCounts(workspaceID, rt.destName, false, true)
+		case utilTypes.DiffStatus:
+			sd.Count--
+			routerWorkspaceJobStatusCount[workspaceID] += 1
```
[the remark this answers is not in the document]
Sidddddarth: Saving the failed record for such a job would spur the sources pod to retry such an event wouldn't it?
adding to `jobRunIDAbortedEventsMap` would cause this.
The `completedJobsList` would be required because it's used for `rsources_stats`, but that would involve the above problem as well.
This happens in processor as well btw, shouldn't we be doing the same here...?

router/router.go
Sidddddarth: agreed

router/router.go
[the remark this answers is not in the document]
Sidddddarth: yeah
we can only add the job to `completedJobsList`.
`jobRunIDAbortedEventsMap` is not relevant for the stats anyway.

router/router.go
atzoum: I am not seeing the `addToFailedMap = false` thoug

Sidddddarth: it's done further down
https://github.com/rudderlabs/rudder-server/pull/2230/files/6a09470fed0e412a3a38561b82896afcb1843c4d#diff-c399d26e7f23b713ce01946ac9e0e6f1fc94d692e8a086180191901da91b6964R1152

atzoum:
IMO suppressing events from router turns out to not being as straightforward as it is in processor. Processor is an event producer at heart, whereas router is an event consumer, so there is no easy way to suppress a message in router after it has been produced by processor. Moreover, processor already has a filtering stage.

Can we reconsider our strategy & instead of relying on transformer for filtering out such events, either in processor or in router, do it always in processor during the filtering stage?

atzoum:
We are moving forward with a different approach, as in #2211

PR #2233, opened 2022-08-02
feat: cpRouter Multitenant Support

cisse21:
# Description
Currently, the warehouse sync retry functionality through the UI happens via the CP router. The CP router maintains a 1-1 connection with the data plane based on the workspaceID as the unique value. but when we move to multi-tenant mode, we will have a single data plane serving multiple workspaces.

We already have a hosted namespace where there is a single connection maintained for multiple workspaces and we would have to extend the same functionality for multi-tenant.

On the config backend DB, we have a table which maps workspace ID to associated namespaces and we would have the Server creating a connection with the namespace as a unique identifier and consequently, every request from UI upon reaching the config backend, would be checked if they are present in a multi-tenant namespace, and if they are the request will be routed to the multi-tenant namespace and if it is not then we route it via the older flow

## Notion Ticket
https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=56dc5445b47940ff92f5de7be3a11d80&pm=s

## Security


warehouse/api.go
```diff
@@ -87,27 +89,50 @@ type UploadAPIT struct {
...
+	case deployment.DedicatedType:
+		connectionIdentifier = config.GetWorkspaceToken()
+	case deployment.MultiTenantType:
+		isMultiWorkspace = true
+		isNamespaced := config.IsEnvSet("WORKSPACE_NAMESPACE")
+		if isNamespaced {
+			connectionIdentifier, err = config.GetEnvErr("WORKSPACE_NAMESPACE")
+			if err != nil {
+				pkgLogger.Errorf("error getting workspace namespace: %s", err.Error())
+				return err
```
fracasula, line 115:
```suggestion
				err = fmt.Errorf("error getting workspace namespace: %w", err)
				pkgLogger.Error(err.Error())
				return err
```

warehouse/api.go
```diff
@@ -87,27 +89,50 @@ type UploadAPIT struct {
...
 }
 
 var UploadAPI UploadAPIT
 
-func InitWarehouseAPI(dbHandle *sql.DB, log logger.LoggerI) {
-	workspaceToken := config.GetWorkspaceToken()
-	isMultiWorkspace := config.GetEnvAsBool("HOSTED_SERVICE", false)
-	if isMultiWorkspace {
-		workspaceToken = config.GetEnv("HOSTED_SERVICE_SECRET", "password")
+func InitWarehouseAPI(dbHandle *sql.DB, log logger.LoggerI) error {
```
fracasula, line 97: Can we have a test for this function?

warehouse/api.go
```diff
@@ -87,27 +89,50 @@ type UploadAPIT struct {
...
+		return err
+	}
+	var connectionIdentifier string
+	var isMultiWorkspace bool
+	switch deploymentType {
+	case deployment.DedicatedType:
+		connectionIdentifier = config.GetWorkspaceToken()
+	case deployment.MultiTenantType:
+		isMultiWorkspace = true
+		isNamespaced := config.IsEnvSet("WORKSPACE_NAMESPACE")
```
fracasula, line 110: Why are we reading environment variables from a completely different domain here in warehouse? Shouldn't we get what we need by accessing some kind of API (API as in go package)? WDYT?

warehouse/warehouse.go
```diff
@@ -790,9 +789,9 @@ func (wh *HandleT) getUploadsToProcess(availableWorkers int, skipIdentifiers []s
 						%s t
 					WHERE
 						t.destination_type = '%s' and t.in_progress=%t and t.status != '%s' and t.status != '%s' %s and COALESCE(metadata->>'nextRetryTime', now()::text)::timestamptz <= now()
-				) grouped_uplaods
+				) grouped_uploads
```
fracasula, line 792: Has the column in the database also been renamed? How do we handle deployments that require schema changes?

fracasula (changes requested):

proto/common/dpauth.proto
```diff
@@ -5,14 +5,15 @@ package proto;
 option go_package = ".;proto";
 
 service DPAuthService{
-  rpc GetWorkspaceToken( GetWorkspaceTokenRequest ) returns ( GetWorkspaceTokenResponse );
+  rpc GetConnectionIdentifier( GetConnectionIdentifierRequest ) returns ( GetConnectionIdentifierResponse );
```
achettyiitr, line 8: This is going to break in terms of backward compatibility. Because we will be going to update the cp-router first. Also, all data planes will not be updated at once. 
I guess we need to keep the existing implementation as well for Backward compatibility.

warehouse/api.go
```diff
@@ -87,27 +89,50 @@ type UploadAPIT struct {
...
-	isMultiWorkspace := config.GetEnvAsBool("HOSTED_SERVICE", false)
-	if isMultiWorkspace {
-		workspaceToken = config.GetEnv("HOSTED_SERVICE_SECRET", "password")
+func InitWarehouseAPI(dbHandle *sql.DB, log logger.LoggerI) error {
+	deploymentType, err := deployment.GetFromEnv()
+	if err != nil {
+		pkgLogger.Errorf("error getting deployment type: %s", err.Error())
+		return err
+	}
+	var connectionIdentifier string
```
achettyiitr, line 103: Can we move this to a separate function?

achettyiitr (changes requested):

warehouse/api.go
abhimanyubabbar: +1

warehouse/api.go
BonapartePC: Added a test for fetching deployment type. Do we need a function for warehouse functions as well?

proto/common/dpauth.proto
BonapartePC: Added backwards compatibility. Can you review again?

warehouse/api.go
```diff
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/rudderlabs/rudder-server/utils/types/deployment"
+
```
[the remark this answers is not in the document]
BonapartePC: go linter did this. Should be fine?

utils/types/deployment/deployment.go
```diff
@@ -33,3 +38,39 @@ func (t Type) Valid() bool {
 	}
 	return false
 }
+
+func GetConnectionToken() (string, bool, error) {
+	deploymentType, err := GetFromEnv()
+	if err != nil {
+		pkgLogger.Errorf("error getting deployment type: %s", err.Error())
```
fracasula, line 45:
```suggestion
		pkgLogger.Errorf("error getting deployment type: %s", err)
```

utils/types/deployment/deployment.go
```diff
@@ -33,3 +38,39 @@ func (t Type) Valid() bool {
...
+	switch deploymentType {
+	case DedicatedType:
+		connectionToken = config.GetWorkspaceToken()
+	case MultiTenantType:
+		isMultiWorkspace = true
+		isNamespaced := config.IsEnvSet("WORKSPACE_NAMESPACE")
+		if isNamespaced {
+			connectionToken, err = config.GetEnvErr("WORKSPACE_NAMESPACE")
+			if err != nil {
+				pkgLogger.Errorf("error getting workspace namespace: %s", err.Error())
```
fracasula, line 59:
```suggestion
				pkgLogger.Errorf("error getting workspace namespace: %s", err)
```

utils/types/deployment/deployment.go
```diff
@@ -3,25 +3,30 @@ package deployment
...
 )
 
 type Type string // skipcq: RVV-B0009
 
 const (
-	DedicatedType   Type = "DEDICATED"
-	MultiTenantType Type = "MULTITENANT"
+	DedicatedType   Type   = "DEDICATED"
+	MultiTenantType Type   = "MULTITENANT"
+	HostedNamespace string = "free-us-1" // Having it here to support legacy cp-router hosted
```
fracasula, line 16: Is this used outside this package? If not please unexport it since we can export it later when needed.

utils/types/deployment/deployment.go
```diff
@@ -33,3 +38,39 @@ func (t Type) Valid() bool {
...
+			if err != nil {
+				pkgLogger.Errorf("error getting workspace namespace: %s", err.Error())
+				return "", false, err
+			}
+			if connectionToken == HostedNamespace {
+				// CP Router still has some things hardcoded for hosted
+				// which needs to be supported
+				connectionToken = config.GetEnv("HOSTED_SERVICE_SECRET", "")
+			}
+		} else {
```
fracasula, line 67: [optional] This block is very long, you could do a `break` at the end of the above `if` and remove this `else`.

utils/types/deployment/deployment.go
fracasula: Same with other occurrences in this PR.

utils/types/deployment/deployment_test.go
```diff
@@ -84,3 +89,68 @@ func Test_GetFromEnv(t *testing.T) {
...
+			expectedBool:       true,
+		},
+		{
+			name:               "MultiTenantType",
+			dType:              "badType",
+			secret:             "secret",
+			namespace:          "free-us-1",
+			expectedIdentifier: "",
+			expectedBool:       false,
+			err:                fmt.Errorf("invalid deployment type: \"badType\""),
```
fracasula, line 140: [suggestion]
```suggestion
			err:                fmt.Errorf(`invalid deployment type: "badType"`),
```

utils/types/deployment/deployment_test.go
```diff
@@ -137,7 +137,7 @@ func Test_GetConnectionToken(t *testing.T) {
 			namespace:          "free-us-1",
 			expectedIdentifier: "",
 			expectedBool:       false,
-			err:                fmt.Errorf("invalid deployment type: \"badType\""),
+			err:                fmt.Errorf(`invalid deployment type: \"badType\"`),
```
fracasula, line 140: Is the escaping of the double quotes needed now that you're using backtickers?

```suggestion
			err:                fmt.Errorf(`invalid deployment type: \"badType\"`),
```

fracasula (approved):
Approved, left a minor comment about escaping double quotes since I'm not sure that was your intention.

controlplane/auth.go
```diff
@@ -7,19 +7,40 @@ import (
...
+	ConnectionToken string
+	InstanceID      string
 }
 
 type authService struct {
 	authInfo AuthInfo
 	proto.UnimplementedDPAuthServiceServer
 }
 
+func (a *authService) GetConnectionToken(ctx context.Context, request *proto.GetConnectionTokenRequest) (*proto.GetConnectionTokenResponse, error) {
```
fracasula, line 20: To make DeepSource happy. Same in the other method.

```suggestion
func (a *authService) GetConnectionToken(_ context.Context, _ *proto.GetConnectionTokenRequest) (*proto.GetConnectionTokenResponse, error) {
```

achettyiitr (approved):

PR #2234, opened 2022-08-02
chore(warehouse): additional config for deduping users table

Jayachand:
# Description

> custom build to separate dedup config for users table

## Notion Ticket

https://www.notion.so/rudderstacks/Add-config-to-enable-dedup-for-only-users-table-grafana-customer-6bb863e7e593424a96b0b42e945427ce

## Security


warehouse/bigquery/bigquery.go
```diff
@@ -591,8 +593,9 @@ func (bq *HandleT) LoadUserTables() (errorMap map[string]error) {
 			return
 		}
 	}
-
-	if !isDedupEnabled {
+	// set isDedupEnabled to true to enable dedup for all tables
```
achettyiitr, line 596: Can we also move this description where the configs are created as well? This will help people give more context to the descriptions of the config.

achettyiitr (changes requested):

warehouse/bigquery/bigquery.go
Jayachand: Updated

warehouse/bigquery/bigquery.go
```diff
@@ -543,7 +543,9 @@ func (bq *HandleT) LoadUserTables() (errorMap map[string]error) {
 			errorMap[warehouseutils.UsersTable] = status.Err()
 			return
 		}
-		defer bq.dropStagingTable(identifyLoadTable.stagingTableName)
+		if isDedupEnabled {
```
achettyiitr, line 546: Why do we need this extra condition for deleting the staging table.

achettyiitr (changes requested):

warehouse/bigquery/bigquery.go
Jayachand: when isDedupEnabled is set `false`, temporary table is not created for identify table, it is just partitioned table i.e. identifies$2022-08-12. Cannot be deleted.

achettyiitr: Got clarified. Thanks.

achettyiitr (approved):

abhimanyubabbar (approved):
LGTM 👍🏼

PR #2235, opened 2022-08-02
chore: don't allow gateway to ack mode change

cisse21:
# Description

Ensuring server mode change is not acknowledged by the gateway component.

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=c8b3adb89c194395875a256d749f7b96&pm=s)

## Security


app/cluster/dynamic.go
```diff
@@ -108,15 +108,17 @@ func (d *Dynamic) Run(ctx context.Context) error {
...
-			err := d.handleModeChange(req.Mode())
-			if err != nil {
-				d.logger.Error(err)
-				return err
-			}
-			d.logger.Debugf("Acknowledging the mode change")
-
-			if err := req.Ack(ctx); err != nil {
-				return fmt.Errorf("ack mode change: %w", err)
+			if !d.GatewayComponent {
```
fracasula, line 111: Wouldn't it be easier to follow if we just `continue` if it's a gateway? Like:

```go
if d.GatewayComponent {
    continue
}

// do the rest
```

router/router.go
```diff
@@ -1819,6 +1814,7 @@ func (rt *HandleT) generatorLoop(ctx context.Context) {
 	for {
 		select {
 		case <-ctx.Done():
+			rt.stopWorkers()
```
fracasula, line 1817: I'm not sure about the way we're handling the start and stop here. The workers are initialized only once and that happens on `Setup()`, not on `Start()`. Then the `generatorLoop` (this method) is started by `Start()` and the `ctx` is canceled in `Shutdown()`. So correct me if I'm wrong but it looks like if we do:
1. setup
2. start
3. stop
4. start
5. stop // it breaks because we close a channel that was already closed?

So maybe along with this change it could make sense moving the `initWorkers` to `Start()`.

multi_tenant_test.go
```diff
@@ -234,15 +234,6 @@ func testMultiTenantByAppType(t *testing.T, appType string) {
...
 
-		select {
-		case ack := <-etcdContainer.Client.Watch(ctx, "test-ack/", clientv3.WithPrefix()):
-			require.Len(t, ack.Events, 1)
-			require.Equal(t, "test-ack/2", string(ack.Events[0].Kv.Key))
-			require.Equal(t, `{"status":"DEGRADED"}`, string(ack.Events[0].Kv.Value))
-		case <-time.After(20 * time.Second):
-			t.Fatal("Timeout waiting for server-mode test-ack")
-		}
-
```
fracasula, line 245: Isn't this also affecting the EMBEDDED mode? The multi tenant test runs the whole thing with GATEWAY as the app type first and then with EMBEDDED, but I think that you want to ack with embedded so just removing the assertions doesn't seem like the best approach. WDYT?

cisse21: Yeah... I was just testing something out with respect to tests ... I will reinstate them back

app/cluster/dynamic.go
cisse21: Yeah this makes sense...

router/router.go
cisse21: Yeah this sounds correct actually, @lvrach Are we missing anything here?

router/manager/manager.go
```diff
@@ -27,7 +27,7 @@ type LifecycleManager struct {
 	waitGroup     *errgroup.Group
 }
 
-func (*LifecycleManager) Run(ctx context.Context) error {
+func (*LifecycleManager) Run(_ context.Context) error {
 	return nil
```
fracasula, line 31: [optional]

```suggestion
func (*LifecycleManager) Run(ctx context.Context) error {
	return ctx.Err()
```

router/router.go
```diff
@@ -785,10 +785,10 @@ func (worker *workerT) processDestinationJobs() {
 		_destinationJob := destinationJob
 
 		for _, destinationJobMetadata := range _destinationJob.JobMetadataArray {
-			handledJobMetadatas[destinationJobMetadata.JobID] = &destinationJobMetadata
+			_destinationJobMetadata := destinationJobMetadata
```
fracasula, line 788: [optional] It should fine, right? This way it is more idiomatic.

```suggestion
			destinationJobMetadata := destinationJobMetadata
```

Alternatively we could even avoid creating a copy and do:

```go
for i := range _destinationJob.JobMetadataArray {
    handledJobMetadatas[destinationJobMetadata.JobID] = &_destinationJob.JobMetadataArray[i]
    // ...
```

router/router.go
lvrach: The way it currently works is a bit confusing, but if you look at https://github.com/rudderlabs/rudder-server/blob/5690a383abc4ce1fae93e8553d2b20819571369d/router/manager/manager.go#L67

our flow actually is:
1. setup
2. start
....
3. shutdown
4. setup
5. start
....

So after every shutdown we create a new router and call start, we don't reuse it.

lvrach (approved):

router/router.go
cisse21: Resolving this for now, We can pick up the refactoring over a separate PR

app/cluster/integration_test.go
```diff
@@ -285,10 +285,16 @@ func TestDynamicClusterManager(t *testing.T) {
 	require.Eventually(t, func() bool {
 		return dCM.Mode() == servermode.NormalMode
 	}, time.Second, time.Millisecond)
+
+	// Letting the router start function to return
+	// else stop would be waiting forever on the wait-group
+	time.Sleep(10 * time.Second)
```
fracasula, line 291: The test is not passing anyway in my machine. I'll debug this a bit and let you know.

fracasula: @cisse21 I pushed a fix for the router manager [here](https://github.com/rudderlabs/rudder-server/pull/2235/commits/a8a24673e189d6598e76e84ab0acccec39619cc6). Now the test passes with no sleep nor delays in my machine. Let me know what you think. Now I'm going to create a ticket to tackle the outstanding router issues.

fracasula: Additional fix pushed [here](https://github.com/rudderlabs/rudder-server/pull/2235/commits/512d31c13c624899cac0237b156475e146942c35).

fracasula: Ticket [here](https://www.notion.so/rudderstacks/BackendConfig-is-initialized-when-credentials-are-bad-e0c2d684739948afa5b99a2c1c48afae).

chandumlg (approved):

PR #2236, opened 2022-08-03
fix(destination): oauth expired secret type change to json.RawMessage from string

sanpj2292:
Signed-off-by: Sai Sankeerth <sanpj2292@github.com>

# Description

The expired secret information is a json-object. Initially we used to send this information as a string to control-plane for refreshing the token. But due to a potential bug, we've now decided to send this information as is to control-plane for refreshing the token.

## Notion Ticket

https://www.notion.so/rudderstacks/Object-comparison-bug-Auth-service-c0ec7ee1c6ae4d969f8e1c10e37c4120

## Security


atzoum:
What is the bug here? The only difference I can spot between the two types is marshalling a nil []byte to `null` vs marshalling an empty string to `""`.

koladilip:
> What is the bug here? The only difference I can spot between the two types is marshalling a nil []byte to `null` vs marshalling an empty string to `""`.
From: https://pkg.go.dev/net/http#pkg-constants
NoBody is an io.ReadCloser with no bytes. Read always returns EOF and Close always returns nil. It can be used in an outgoing client request to explicitly signal that a request has zero bytes. An alternative, however, is to simply set Request.Body to nil. 

Looks like both can used. @sanpj2292 are we getting any errors in pass nil?

sanpj2292:
> > From: https://pkg.go.dev/net/http#pkg-constants
> > NoBody is an io.ReadCloser with no bytes. Read always returns EOF and Close always returns nil. It can be used in an outgoing client request to explicitly signal that a request has zero bytes. An alternative, however, is to simply set Request.Body to nil.
> 
> Looks like both can used. @sanpj2292 are we getting any errors in pass nil?

This change is attributed to the suggestion provided in DeepSource. When I put in `nil` it was suggesting to replace it with `http.NoBody`

sanpj2292:
> What is the bug here? The only difference I can spot between the two types is marshalling a nil []byte to `null` vs marshalling an empty string to `""`.

The bug is that we are sending the expiredSecret(which is actually a json) as a string. Through this PR, we are changing it to `[]byte`.

koladilip (approved):

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -460,7 +460,7 @@ func (authErrHandler *OAuthErrResHandler) cpApiCall(cpReq *ControlPlaneRequestT)
 		reqBody = bytes.NewBufferString(cpReq.Body)
 		req, err = http.NewRequest(cpReq.Method, cpReq.Url, reqBody)
 	} else {
```
saikumarrs, line 462: Naive question. Would there be any issue if we skip this check completely?

saikumarrs (approved):

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: This is to support `FetchToken` type of calls where in no body is to be sent.

saikumarrs: That's ok. Would there be an issue sending the empty body as it is?

```
...
var req *http.Request
var err error
// would it fail at this step?
reqBody = bytes.NewBufferString(cpReq.Body)

req, err = http.NewRequest(cpReq.Method, cpReq.Url, reqBody)
if err != nil {
  authErrHandler.logger.Errorf("[%s request] :: destination request failed: %+v\n", loggerNm, err)
  // Abort on receiving an error in request formation
  return http.StatusBadRequest, err.Error()
}
...
```

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -103,8 +103,8 @@ const (
 )
 
 type RefreshTokenBodyParams struct {
```
atzoum, line 105: This struct only exists for marshalling and sending the payload to control plane. Whether the type of `ExpiredSecret` is a `string` or a `[]byte` is not that much important, but changing seems to be appropriate since the underlying is a json

atzoum (approved):

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: @saikumarrs Want to understand if you feel this if-else is complicated or is there something else that you feel is not right ?! Basically want to understand why the change is being discussed.

router/oauthResponseHandler/oauthResponseHandler.go
sanpj2292: The `ExpiredSecret` is actually a json column in control-plane DB. Initially we were stringifying this json in rudder-server, it didn't feel right as an extra processing would be required in control-plane to compare existing secret with the expired secret. Hence we are making this change to preserve the original data-type of `ExpiredSecret` and send it to control-plane

Let me know if it makes sense

PR #2241, opened 2022-08-03
refactor: use interfaces and common utils for streaming destinations

koladilip:
# Description
    
Used Interfaces for better testability.
Added utils to handle AWS session creation.
Added basic IAM role support.

## Notion Ticket
https://www.notion.so/rudderstacks/AWS-FTR-for-streaming-destination-PoC-78b63869ac904ba49b0018a527ccb343

## Security


lvrach (commented):
Very useful job here! It was high time to do it.
I left a few minor comments.

A small warning here, whenever we refactor this part of the code, they're some edge cases configuration in production that we end up breaking. For this I would suggest using this tool: 
https://github.com/rudderlabs/rudder-tests/tree/master/backtest-destinations

Further more, if we could introduce Ping methods for those destinations, so that we tests that all production configuration works without side-effects.

services/streammanager/common/common.go
```diff
@@ -0,0 +1,49 @@
+package common
```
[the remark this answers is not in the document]
koladilip: It is added to avoid import cycle problem as streammanger package depends on destination specific managers so destination managers shouldn't import streammanager.

utils/googleutils/googleutils.go
```diff
@@ -8,7 +8,7 @@ import (
 
 func CompatibleGoogleCredentialsJSON(jsonKey []byte) error {
 	if _, err := google.ConfigFromJSON(jsonKey); err == nil {
-		return fmt.Errorf("Google Developers Console client_credentials.json file is not supported")
+		return fmt.Errorf("google developers console client_credentials.json file is not supported")
```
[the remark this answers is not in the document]
koladilip: We are constructing this custom error message when err := google.ConfigFromJSON(jsonKey), err is nil as we want to check if the user is using console credentials. If we get error means they are not using console credentials which is good for us.

services/streammanager/eventbridge/eventbridgemanager.go
```diff
@@ -84,21 +55,16 @@ func Produce(jsonData json.RawMessage, producer, destConfig interface{}) (int, s
...
 
-		// fetching status code from response
-		if reqErr, ok := err.(awserr.RequestFailure); ok {
-			statusCode = reqErr.StatusCode()
-		}
-
-		return statusCode, err.Error(), err.Error()
+	if err != nil {
+		statusCode, respStatus, responseMessage := common.ParseAWSError(err)
+		pkgLogger.Errorf("[EventBridge] error  :: %s : %s", respStatus, responseMessage)
```
saikumarrs, line 66: Can we log the status code too here?

services/streammanager/firehose/firehosemanager.go
```diff
@@ -1,159 +1,86 @@
...
-	err = json.Unmarshal(jsonConfig, &config)
-	if err != nil {
-		respStatus = "Failure"
-		responseMessage = "[FireHose] error  :: " + err.Error()
-		pkgLogger.Errorf("[FireHose] error  :: %v", err)
-		statusCode := 400
-		return statusCode, respStatus, responseMessage
+	client := producer.client
+	if client == nil {
+		return 400, "Failure", "[FireHose] error :: Could not create producer"
```
saikumarrs, line 45: Wouldn't it be better to validate this at the "NewProducer" stage?

saikumarrs (approved):

services/streammanager/common/common.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func mapErrorMessageToStatusCode(errorMessage string, defaultStatusCode int) int {
+
+	if strings.Contains(errorMessage, "Throttling") {
+		// aws returns  "ThrottlingException"
+		// for throttling requests server will retry
+		return 429
+	}
+	if strings.Contains(errorMessage, " RequestExpired") {
```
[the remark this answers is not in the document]
koladilip: Sure, will fix this

services/streammanager/eventbridge/eventbridgemanager.go
koladilip: Sure, will fix this

services/streammanager/firehose/firehosemanager.go
koladilip: We can but it is not necessary that we only use NewProducer to create the producer right, especially for testing purposes we pass our own mock client so to catch that this check is added.

saikumarrs: Makes sense.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -506,17 +507,17 @@ func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timest
 }
 
 // CloseProducer closes a given producer
-func CloseProducer(ctx context.Context, pi interface{}) error {
+func (producer *KafkaProducer) CloseProducer() error {
```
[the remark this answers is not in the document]
koladilip: Sure

utils/awsutils/session.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+		Region:      aws.String(config.Region),
+		Credentials: createCredentails(config),
+	}))
+}
+
+func NewSessionConfig(destinationConfig interface{}, timeout time.Duration, serviceName string) (*SessionConfig, error) {
+	sessionConfig := SessionConfig{}
+
+	jsonConfig, err := json.Marshal(destinationConfig)
```
cisse21, line 70: Any reason why we would need to do both json marshal and unmarshal instead of just type checking the input interface?

koladilip: Downstream components is sending destinationConfig as interface{} and it cannot be type casted to SessionConfig but it has same fields as session config so marshalling to json and then unmarshalling to SessionConfig, actually this is an existing code just moved here. Ideal would be taking destination config as map[string]interface{} and manually extract required fields but this itself is bit long change as we need to write custom builders for each time from map type so currently not planning to address as part of this PR.

koladilip:
@lvrach I have compared no difference in both iterations. I have files with me, I will share in our internal slack channel.

cisse21 (approved):

lvrach (approved):

PR #2243, opened 2022-08-04
fix(warehouse): marshall pg_notifier_queue payload during the time of inserting the data

achettyiitr:
# Description
We are seeing a huge spike in memory when we are trying to generate load files from staging files. This is happening because in our current system we are storing the payload job which we are sending to pg_notifier_queue in memory and one of the major components that it is keeping is UploadSchme/MergeSchema. So In order to avoid this, we are doing JsonMarshalling during the time of inserting the data into the slaves.

## Notion Ticket

https://www.notion.so/rudderstacks/PG-Notifier-payload-during-Create-Load-Files-is-causing-memory-to-shoot-up-c5f596e1e41c41438addbd9246b1d6d3

## Security


services/pgnotifier/pgnotifier.go
```diff
@@ -335,12 +335,22 @@ func (notifier *PgNotifierT) Publish(jobs []JobPayload, priority int) (ch chan [
...
+	}()
 
 	stmt, err := txn.Prepare(pq.CopyIn(queueName, "batch_id", "status", "payload", "workspace", "priority"))
 	if err != nil {
 		return
 	}
-	defer func() { _ = stmt.Close() }()
+	defer func() {
+		_ = stmt.Close()
+	}()
```
abhimanyubabbar, line 353: Why not just `defer stmt.Close()` ?

services/pgnotifier/pgnotifier.go
```diff
@@ -355,6 +365,31 @@ func (notifier *PgNotifierT) Publish(jobs []JobPayload, priority int) (ch chan [
...
+		SET
+		  payload = payload || $1
+		where
+		  batch_id = $2;`
+	_, err = txn.Exec(sqlStatement, []interface{}{
+		uploadSchemaJSON,
+		batchID,
+	}...)
+	if err != nil {
+		return
```
abhimanyubabbar, line 390: When returning, it's always good to add some context, so maybe:

`return fmt.Errorf("Unable to execute transaction to update notifier queue, err: %w", err)`

services/pgnotifier/pgnotifier.go
```diff
@@ -355,6 +365,31 @@ func (notifier *PgNotifierT) Publish(jobs []JobPayload, priority int) (ch chan [
...
 		return
 	}
+
+	uploadSchemaJSON, err := json.Marshal(struct {
+		UploadSchema whUtils.SchemaT
+	}{
+		UploadSchema: *schema,
+	})
+	if err != nil {
+		return
```
abhimanyubabbar, line 375: Same as below to add some context when returning the error. Wdyt ?

warehouse/upload.go
```diff
@@ -1776,29 +1775,28 @@ func (job *UploadJobT) createLoadFiles(generateAll bool) (startLoadFileID, endLo
...
 
-			// set merged schema as upload schema if the load file type is parquet
-			if job.upload.LoadFileType == warehouseutils.LOAD_FILE_TYPE_PARQUET {
-				payload.UploadSchema = job.upload.MergedSchema
-			}
-
 			payloadJSON, err := json.Marshal(payload)
 			if err != nil {
 				panic(err)
 			}
```
abhimanyubabbar, line 1788: Outside the purview of the PR, but I saw an unnecessary panic here, should we handle it gracefully ?

warehouse/upload.go
```diff
@@ -1776,29 +1775,28 @@ func (job *UploadJobT) createLoadFiles(generateAll bool) (startLoadFileID, endLo
...
 			messages = append(messages, payloadJSON)
 		}
+		var schema *warehouseutils.SchemaT
+		if job.upload.LoadFileType == warehouseutils.LOAD_FILE_TYPE_PARQUET {
+			schema = &job.upload.MergedSchema
+		} else {
+			schema = &job.upload.UploadSchema
+		}
```
abhimanyubabbar, line 1796: I would rephrase it like this to better represent the intent and save 1 extra indentation,
```
schema := &job.upload.UploadSchema

if job.upload.LoadFileType == warehouseutils.LOAD_FILE_TYPE_PARQUET {
    schema = &job.upload.MergedSchema
}
```

Wdyt ? ( Just an opinion and not a hard stance )

services/pgnotifier/pgnotifier.go
```diff
@@ -335,12 +335,22 @@ func (notifier *PgNotifierT) Publish(jobs []JobPayload, priority int) (ch chan [
 	if err != nil {
 		return
 	}
+	defer func() {
+		if err != nil {
+			if rollbackErr := txn.Rollback(); rollbackErr != nil {
+				pkgLogger.Error(err.Error())
+				pkgLogger.Error(rollbackErr.Error())
+			}
```
abhimanyubabbar, line 343: What happens if rollbackErr is null ? Do we not want to print the `err` in that case ?

abhimanyubabbar (commented):
Some comments from initial cursory review.

warehouse/upload.go
achettyiitr: Ideally yes. But we need some more changes as well. So, didn't touch it.

services/pgnotifier/pgnotifier.go
achettyiitr: Done.

services/pgnotifier/pgnotifier.go
achettyiitr: Done.

services/pgnotifier/pgnotifier.go
achettyiitr: Done.

satishrudderstack (approved):
LGTM

abhimanyubabbar (approved):
Some minor suggestions, please fix them if you feel it makes sense. The PR looks good to me semantically. 🚀 :shipit: 👍🏼

PR #2251, opened 2022-08-04
chore: propogate s3manager error up

chandumlg:
# Description

Propagating s3 getSession errors to the caller avoiding panic.

## Security


atzoum (approved):

lvrach (approved):

saurav-malani:
We are using `session.Must` is various part of reply code & awsutils. We might want to avoid panic & propagate the error there too.

PR #2254, opened 2022-08-05
fix(warehouse): reverting schemaExisting changes for snowflake.go. Using without query parameter implementation.

achettyiitr:
# Description

Reverting schemaExisting changes for snowflake.go. Using without query parameter implementation.

## Notion Ticket

< Replace with Notion Link >

## Security


achettyiitr:
This got deployed to customer.

PR #2258, opened 2022-08-05
fix(destination): transformer proxy respond with timeout for slow or unresponsive dests.

sanpj2292:
# Description

- For unresponsive or slow destinations, we will now respond with `504`(timeout status-code) and an appropriate error message
- We will be reducing backoff count to a single digit number so that after initial retries the router retries as per it's availability

## Security


router/transformer/transformer.go
```diff
@@ -64,7 +64,7 @@ func loadConfig() {
 	config.RegisterIntConfigVariable(30, &maxRetry, true, 1, "Processor.maxRetry")
 	config.RegisterDurationConfigVariable(100, &retrySleep, true, time.Millisecond, []string{"Processor.retrySleep", "Processor.retrySleepInMS"}...)
 	config.RegisterDurationConfigVariable(30, &timeoutDuration, false, time.Second, []string{"HttpClient.timeout"}...)
-	config.RegisterInt64ConfigVariable(15, &retryWithBackoffCount, true, 1, "Router.transformerProxyRetryCount")
+	config.RegisterInt64ConfigVariable(3, &retryWithBackoffCount, true, 1, "Router.transformerProxyRetryCount")
```
saikumarrs, line 67: Let's remove the retry logic completely and rely on the router's retry mechanism. It'll also align with the previous behavior.

sanpj2292: Yes, working on it 🧨

router/router.go
```diff
@@ -2235,7 +2236,7 @@ func (rt *HandleT) Setup(backendConfig backendconfig.BackendConfig, jobsDB jobsd
 	rt.routerResponseTransformStat = stats.NewTaggedStat("response_transform_latency", stats.TimerType, stats.Tags{"destType": rt.destName})
 
 	rt.transformer = transformer.NewTransformer()
-	rt.transformer.Setup()
+	rt.transformer.Setup(rt.netClientTimeout)
```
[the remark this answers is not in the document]
sanpj2292: This timeout is being set here in [`network.go`](https://github.com/rudderlabs/rudder-server/blob/master/router/router.go#L2177) at destination-level only

router/transformer/transformer.go
```diff
@@ -266,39 +246,56 @@ func (trans *HandleT) ProxyRequest(ctx context.Context, responseData integration
 }
 
 // is it ok to use same client for network and transformer calls? need to understand timeout setup in router
-func (trans *HandleT) Setup() {
+func (trans *HandleT) Setup(netClientTimeout time.Duration) {
 	trans.logger = pkgLogger
 	trans.tr = &http.Transport{}
 	trans.client = &http.Client{Transport: trans.tr, Timeout: timeoutDuration}
```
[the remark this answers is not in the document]
sanpj2292: Currently this client is being used for `routerTransform` so it is being undisturbed

router/transformer/transformer.go
```diff
@@ -266,39 +246,56 @@ func (trans *HandleT) ProxyRequest(ctx context.Context, responseData integration
...
+	// This client is to be used for proxy request
+	trans.tfProxyClient = &http.Client{Transport: trans.tr, Timeout: 2 * netClientTimeout}
```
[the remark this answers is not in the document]
sanpj2292: We wanted to provide a higher timeout(with a good margin) to the http client in server. Do you feel we need to probably lower it down a bit ?

router/transformer/transformer.go
```diff
@@ -42,9 +48,9 @@ type HandleT struct {
 
 // Transformer provides methods to transform events
 type Transformer interface {
-	Setup()
+	Setup(timeout time.Duration)
 	Transform(transformType string, transformMessage *types.TransformMessageT) []types.DestinationJobT
-	ProxyRequest(ctx context.Context, responseData integrations.PostParametersT, destName string, jobId int64) (statusCode int, respBody string)
+	ProxyRequest(ctx context.Context, responseData integrations.PostParametersT, destName string, jobId int64, varargs ...interface{}) (statusCode int, respBody string)
```
saikumarrs, line 53: The variable arguments changes made to support testing sounded vague to me. Can you find alternatives?

router/transformer/transformer.go
```diff
@@ -266,39 +251,56 @@ func (trans *HandleT) ProxyRequest(ctx context.Context, responseData integration
 }
 
 // is it ok to use same client for network and transformer calls? need to understand timeout setup in router
-func (trans *HandleT) Setup() {
+func (trans *HandleT) Setup(netClientTimeout time.Duration) {
 	trans.logger = pkgLogger
 	trans.tr = &http.Transport{}
 	trans.client = &http.Client{Transport: trans.tr, Timeout: timeoutDuration}
+	trans.tfProxyTimeout = 2 * netClientTimeout
```
saikumarrs, line 258: How about `<timeout used in processor stage> + <destination timeout>`?

saikumarrs: We should also split both the timeout values based on the above comment.

router/router.go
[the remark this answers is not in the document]
saikumarrs: The timeout value is set just a few statements above from the config,
https://github.com/rudderlabs/rudder-server/blob/5aea3192c32cfc3689ab1c7757ab4a2669be93b3/router/router.go#L2165

Destination-specific value takes precedence over the default timeout.

router/transformer/transformer.go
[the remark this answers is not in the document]
koladilip: For now here having higher timeout is not really harm as transformer is our own internal service but having stat will help if we notice any delays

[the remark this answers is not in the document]
saikumarrs: Why not use the timeout value used for destination transformation? It's usually 30s which is a good value to start with.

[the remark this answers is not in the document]
sanpj2292: Any reason you think destination transformatoin timeout value would be beneficial here ? @saikumarrs

router/transformer/transformer.go
sanpj2292: It is fine to split the timeout into `x + destination_timeout`. The thing is, I feel this `x` can be a constant or some dynamic value. Why can't this `x` be `destination_timeout` ? Do you feel the timeout will be too high or something ?!

router/transformer/transformer.go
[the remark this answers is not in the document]
saikumarrs: It's a proven safe timeout for server-transformer interaction. Unless we've seen values more than that, I'd start with it.

router/transformer/transformer.go
saikumarrs: `x` should be independent of destinations. If we want to increase the timeout for a given destination to say `1 min`, we'd inadvertently increase the server->transformer timeout also to `1 min`, which may not be acceptable.

router/transformer/transformer.go
saikumarrs: Looks good, but unfortunately, it adds a `marshal` operation. How is this problem solved in other unit tests? Is this unique to this situation?
`payload, err := jsonfast.Marshal(proxyReqParams.ResponseData)`

router/transformer/transformer.go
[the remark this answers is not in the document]
koladilip: Usually destinations should respond within 30 seconds for most of the requests but if there is file upload or something then we need to wait for longer time period. The positive side of the low timeout (30s) is we fail faster but negative side is: there could requests which are successful after 30 s at server end but we terminated the connection. Do we know typical response times of each destinations?

router/transformer/transformer.go
sanpj2292: The marshalling was already present even before this refactor

saikumarrs: My bad.

router/transformer/transformer.go
sanpj2292: The changes have been made. Can you verify ?

router/transformer/transformer.go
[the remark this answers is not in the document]
saikumarrs: My recommendation is `X + <destination timeout>`, where `X` can start with the timeout value we use for processor->transformer interaction, which is usually `30` secs. 
The total router->transformer proxy timeout is `30 + 30` seconds for most of the destinations (`30` seconds is also the default for destination timeout value).

Overall, the minimum timeout would be `30` seconds unless we configure the `X` as <30 secs.

router/transformer/transformer.go
```diff
@@ -266,39 +255,65 @@ func (trans *HandleT) ProxyRequest(ctx context.Context, responseData integration
...
 		trans.logger.Errorf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Client.Do Failure for %[1]v, with %[3]v`, destName, jobId, err.Error())
-		return []byte{}, http.StatusBadRequest, err
+		return []byte{}, http.StatusGatewayTimeout, err
+	} else if err != nil {
+		// This was an error, but not a timeout
+		trans.logger.Errorf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Client.Do Failure for %[1]v, with %[3]v`, destName, jobId, err.Error())
+		return []byte{}, http.StatusInternalServerError, err
 	}
 
+	// if err != nil {
```
koladilip, line 312: Delete this comment

router/transformer/transformer_test.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+}
+
+// A kind of mock for transformer proxy endpoint in transformer
+func proxyHandlerFunc(w http.ResponseWriter, req *http.Request) {
+	vars := mux.Vars(req)
+	destName, ok := vars["destName"]
+	if !ok {
+		// Modify this case ?
+		panic(fmt.Errorf("Wrong url being sent"))
```
koladilip, line 217: Instead of panicking, can we use writer to return error response?
use return http.Error(...), same for all panics in this functions

router/transformer/transformer_test.go
```diff
@@ -0,0 +1,232 @@
...
+			}
+			// Logic to include context timing out
+			ctx := context.TODO()
+			var cancelFunc context.CancelFunc
+			if tc.context.Timeout.Milliseconds() > 0 {
+				ctx, cancelFunc = context.WithTimeout(context.TODO(), tc.context.Timeout)
+				defer cancelFunc()
+			} else if tc.context.Cancel {
+				ctx, cancelFunc = context.WithCancel(context.TODO())
+				cancelFunc()
```
koladilip, line 190: Should we use `go cancelFunc()` to mimic it happened concurrently ?

sanpj2292: Not sure if it makes any noticeable impact. Do you really think, it would help us if we did this ?

router/transformer/transformer.go
sanpj2292: Done

router/transformer/transformer_test.go
sanpj2292: Done

router/transformer/transformer.go
```diff
@@ -307,13 +305,15 @@ func (trans *HandleT) makeTfProxyRequest(ctx context.Context, proxyReqParams *Pr
 		// This was an error, but not a timeout
 		trans.logger.Errorf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Client.Do Failure for %[1]v, with %[3]v`, destName, jobId, err.Error())
 		return []byte{}, http.StatusInternalServerError, err
+	} else if resp.StatusCode == http.StatusNotFound {
```
koladilip, line 308: os.IsTimeout(err) there is already a function to check if the errors is timeout or not

router/transformer/transformer_test.go
```diff
@@ -200,7 +226,7 @@ func TestProxyRequest(t *testing.T) {
 			assert.Equal(t, tc.ExpectedStCode, stCd)
 			if gjson.GetBytes([]byte(resp), "message").Raw != "" {
 				require.JSONEq(t, tc.ExpectedBody, resp)
-			} else if tc.ExpectedStCode == http.StatusGatewayTimeout {
+			} else if tc.ExpectedStCode == http.StatusGatewayTimeout || tc.ExpectedStCode == http.StatusNotFound {
```
koladilip, line 229: These test cases are getting complex with many if conditions and is it possible to implement them as separate test case instead of looping over them?

router/transformer/transformer.go
sanpj2292: Are you referring to the line where in the error is being casted to net.Error and recommending to change it ? or are you saying the line on which the comment is made needs to be changed ?

router/router.go
```diff
@@ -676,8 +676,14 @@ func (worker *workerT) processDestinationJobs() {
...
 									jobId := destinationJob.JobMetadataArray[0].JobID
 									pkgLogger.Debugf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Request started`, worker.rt.destName, jobId)
+									proxyReqparams := &transformer.ProxyRequestParams{
+										DestName:     worker.rt.destName,
+										JobId:        jobId,
+										ResponseData: val,
+									}
 									rtlTime := time.Now()
-									respStatusCode, respBodyTemp = worker.rt.transformer.ProxyRequest(ctx, val, worker.rt.destName, jobId)
+									// Should we change this context to sendCtx and probably adjust the timeout for the context ?
```
cisse21, line 685: Yes I think it makes much more sense to use `sendCtx` than the global `ctx` over here

router/transformer/transformer.go
```diff
@@ -183,43 +195,18 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
+func (trans *HandleT) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequestParams) (int, string) {
+	stats.NewTaggedStat("transformer_proxy.delivery_request", stats.CountType, stats.Tags{"destination": proxyReqParams.DestName}).Increment()
+	trans.logger.Debugf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Proxy Request starts - %[1]v`, proxyReqParams.DestName, proxyReqParams.JobId)
 
-	backoffWithMaxRetry := backoff.WithMaxRetries(backoff.NewExponentialBackOff(), uint64(retryWithBackoffCount))
-	err = backoff.RetryNotify(operation, backoffWithMaxRetry, func(err error, t time.Duration) {
-		pkgLogger.Errorf("[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Request for proxy to URL:: %[3]v, Error:: %+[4]v retrying after:: %[5]v,", destName, jobId, url, err, t)
-		stats.NewTaggedStat("transformer_proxy.retries", stats.CountType, stats.Tags{"destination": destName}).Increment()
-	})
+	rdl_time := time.Now()
```
cisse21, line 202: use camel case variables instead of `_`

router/transformer/transformer.go
```diff
@@ -183,43 +195,18 @@ func (trans *HandleT) Transform(transformType string, transformMessage *types.Tr
...
+	respData, respCode, requestError := trans.makeTfProxyRequest(ctx, proxyReqParams)
+	reqSuccessStr := strconv.FormatBool(requestError == nil)
+	stats.NewTaggedStat("transformer_proxy.request_latency", stats.TimerType, stats.Tags{"requestSuccess": reqSuccessStr, "destination": proxyReqParams.DestName}).SendTiming(time.Since(rdl_time))
+	stats.NewTaggedStat("transformer_proxy.request_result", stats.CountType, stats.Tags{"requestSuccess": reqSuccessStr, "destination": proxyReqParams.DestName}).Increment()
 
-	if err != nil {
-		panic(fmt.Errorf("[TransformerProxy] Proxy request failed after max retries Error:: %+v", err))
+	if requestError != nil {
+		return respCode, requestError.Error()
```
cisse21, line 209: `http.DetectContentType(respData)`

Why do we do this by default? Isn't this a costly operation

router/transformer/transformer.go
sanpj2292: Done

router/transformer/transformer_test.go
sanpj2292: I changed the test-cases a bit, can you check and let me know if you still feel they're complicated

router/transformer/transformer.go
cisse21: Also if there is a timeout won't the go HTTP client return `http.StatusGatewayTimeout` , do we need to explicitly send it

sanpj2292: It seems like the `resp` object will return `nil` and only error which says `context deadline exceeded (Client.Timeout exceeded while awaiting headers)` is being sent from the client

router/transformer/transformer_test.go
koladilip: http.StatusInternalServerError is covered?

koladilip: I got the logic but it is taking time to understand see if you can simplify further.

router/router.go
sanpj2292: It was changed as part of some kind of review comment or something a long-time ago. Did we see any performance impact due to using `DetectContentType` in general ?

router/transformer/transformer.go
sanpj2292: It was changed as part of some kind of review comment or something a long-time ago. Did we see any performance impact due to using `DetectContentType` in general ?

router/transformer/transformer.go
koladilip: > Are you referring to the line where in the error is being casted to net.Error and recommending to change it ? or are you saying the line on which the comment is made needs to be changed ?

Right it is line before it. It works in generic way so better.

sanpj2292: Done

router/router.go
sanpj2292: Done

koladilip (approved):

saikumarrs (approved):

router/router.go
```diff
@@ -672,16 +672,29 @@ func (worker *workerT) processDestinationJobs() {
 							} else {
 								// stat start
 								pkgLogger.Debugf(`responseTransform status :%v, %s`, worker.rt.transformerProxy, worker.rt.destName)
-								sendCtx, cancel := context.WithTimeout(ctx, worker.rt.netClientTimeout)
-								defer cancel()
+								var sendCtx context.Context
+								var cancel context.CancelFunc
```
atzoum, line 676: Don't see any need for these to be declared beforehands
```suggestion
```

router/router.go
```diff
@@ -672,16 +672,29 @@ func (worker *workerT) processDestinationJobs() {
...
+										DestName:     worker.rt.destName,
+										JobId:        jobId,
+										ResponseData: val,
+									}
+									// For http.client for proxy we are using
+									// backendProxyTimeout + netClientTimeout as the timeout value
+									// we are adding 20ms extra here to make sure the context timeout is at higher value
+									// Should we reduce the value from 20 to 10 ms ?
+									prxCtxTimeout := worker.rt.backendProxyTimeout + worker.rt.netClientTimeout + 20*time.Millisecond
+									sendCtx, cancel = context.WithTimeout(ctx, prxCtxTimeout)
```
atzoum, line 691:
```suggestion
									sendCtx, cancel := context.WithTimeout(ctx, prxCtxTimeout)
```

router/router.go
```diff
@@ -696,6 +709,8 @@ func (worker *workerT) processDestinationJobs() {
 										})
 									}
 								} else {
+									sendCtx, cancel = context.WithTimeout(ctx, worker.rt.netClientTimeout)
```
atzoum, line 712:
```suggestion
									sendCtx, cancel := context.WithTimeout(ctx, worker.rt.netClientTimeout)
```

router/router.go
```diff
@@ -672,16 +672,29 @@ func (worker *workerT) processDestinationJobs() {
...
+										JobId:        jobId,
+										ResponseData: val,
+									}
+									// For http.client for proxy we are using
+									// backendProxyTimeout + netClientTimeout as the timeout value
+									// we are adding 20ms extra here to make sure the context timeout is at higher value
+									// Should we reduce the value from 20 to 10 ms ?
+									prxCtxTimeout := worker.rt.backendProxyTimeout + worker.rt.netClientTimeout + 20*time.Millisecond
+									sendCtx, cancel = context.WithTimeout(ctx, prxCtxTimeout)
+									defer cancel()
```
atzoum, line 692: usage of `defer` here is incorrect, we don't want `cancel` to be called after `processDestinationJobs` is done, we want it to be called after the http request code block finishes.
```suggestion
```

router/router.go
```diff
@@ -672,16 +672,29 @@ func (worker *workerT) processDestinationJobs() {
...
 									rtlTime := time.Now()
-									respStatusCode, respBodyTemp = worker.rt.transformer.ProxyRequest(ctx, val, worker.rt.destName, jobID)
+									// Should we change this context to sendCtx and probably adjust the timeout for the context ?
+									respStatusCode, respBodyTemp, respContentType = worker.rt.transformer.ProxyRequest(sendCtx, proxyReqparams)
```
atzoum, line 695:
```suggestion
									respStatusCode, respBodyTemp, respContentType = worker.rt.transformer.ProxyRequest(sendCtx, proxyReqparams)
									cancel()
```

router/router.go
```diff
@@ -696,6 +709,8 @@ func (worker *workerT) processDestinationJobs() {
 										})
 									}
 								} else {
+									sendCtx, cancel = context.WithTimeout(ctx, worker.rt.netClientTimeout)
+									defer cancel()
```
atzoum, line 713:
```suggestion
```

router/router.go
```diff
@@ -696,6 +709,8 @@ func (worker *workerT) processDestinationJobs() {
...
 									rdlTime := time.Now()
 									resp := worker.rt.netHandle.SendPost(sendCtx, val)
```
atzoum, line 715:
```suggestion
									resp := worker.rt.netHandle.SendPost(sendCtx, val)
									cancel()
```

router/transformer/transformer_test.go
```diff
@@ -0,0 +1,323 @@
...
+func TestProxyRequest(t *testing.T) {
+	t.Setenv("RSERVER_HTTP_CLIENT_TIMEOUT", "1s")
+	Initialization()
+
+	srvMux := mux.NewRouter()
+	srvMux.HandleFunc("/v0/destinations/{destName}/proxy", proxyHandlerFunc)
+	srv := httptest.NewServer(srvMux)
+	defer srv.Close()
+
+	for destName, tc := range proxyResponseTcs {
```
lvrach, line 236: [minor] Using a map for test cases has the side effect of randomising test execution.

Which might create some confusion when debugging a test.

router/router.go
```diff
@@ -2236,7 +2251,7 @@ func (rt *HandleT) Setup(backendConfig backendconfig.BackendConfig, jobsDB jobsd
 	rt.routerResponseTransformStat = stats.NewTaggedStat("response_transform_latency", stats.TimerType, stats.Tags{"destType": rt.destName})
 
 	rt.transformer = transformer.NewTransformer()
-	rt.transformer.Setup()
+	rt.transformer.Setup(rt.netClientTimeout)
```
atzoum, line 2254: Please adapt `trackStuckDelivery` method too.

We shouldn't have two different ways of initializing each timeout. Either pass them both of them as arguments or initialize them both during loadConfig. Same applies for their scope (global var vs struct field)

router/transformer/transformer_test.go
```diff
@@ -0,0 +1,323 @@
...
+			tr := rtTf.NewTransformer()
+			// Logic for executing test-cases not manipulating test-cases
+			if tc.rtTimeout.Milliseconds() > 0 {
+				tr.Setup(tc.rtTimeout)
+			} else {
+				// Just a default value
+				tr.Setup(2 * time.Millisecond)
+			}
+			// Logic to include context timing out
+			ctx := context.TODO()
```
lvrach, line 251: [minor] Is there a plan to provide a parent context? If not, TODO is not need:
```suggestion
			ctx := context.Background()
```

router/transformer/transformer.go
```diff
@@ -32,19 +34,37 @@ const (
...
 }
 
+type ProxyRequestParams struct {
+	ResponseData integrations.PostParametersT
+	DestName     string
+	JobId        int64
+	BaseUrl      string
+}
+
+type HTTPProxyResponse struct {
```
atzoum, line 57: this is an internal structure, no need for it to be exported
```suggestion
type httpProxyResponse struct {
```

router/router.go
```diff
@@ -672,16 +672,29 @@ func (worker *workerT) processDestinationJobs() {
...
+									// For http.client for proxy we are using
+									// backendProxyTimeout + netClientTimeout as the timeout value
+									// we are adding 20ms extra here to make sure the context timeout is at higher value
+									// Should we reduce the value from 20 to 10 ms ?
+									prxCtxTimeout := worker.rt.backendProxyTimeout + worker.rt.netClientTimeout + 20*time.Millisecond
+									sendCtx, cancel = context.WithTimeout(ctx, prxCtxTimeout)
+									defer cancel()
 									rtlTime := time.Now()
-									respStatusCode, respBodyTemp = worker.rt.transformer.ProxyRequest(ctx, val, worker.rt.destName, jobID)
+									// Should we change this context to sendCtx and probably adjust the timeout for the context ?
```
atzoum, line 694:
```suggestion
```

router/router.go
```diff
@@ -672,16 +672,29 @@ func (worker *workerT) processDestinationJobs() {
...
+									pkgLogger.Debugf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Request started`, worker.rt.destName, jobId)
+									proxyReqparams := &transformer.ProxyRequestParams{
+										DestName:     worker.rt.destName,
+										JobId:        jobId,
+										ResponseData: val,
+									}
+									// For http.client for proxy we are using
+									// backendProxyTimeout + netClientTimeout as the timeout value
+									// we are adding 20ms extra here to make sure the context timeout is at higher value
+									// Should we reduce the value from 20 to 10 ms ?
```
atzoum, line 689:
```suggestion
```

router/transformer/transformer_test.go
```diff
@@ -0,0 +1,323 @@
...
+	tc := proxyResponseTcs[destName]
+	// sleep is being used to mimic the waiting in actual transformer response
+	if tc.Proxy.TimeoutValue.Milliseconds() > 0 {
+		time.Sleep(tc.Proxy.TimeoutValue)
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(tc.Proxy.StCode)
+	// Lint error fix
+	checkAndSendResponse(&w, []byte(tc.Proxy.Response))
+}
```
lvrach, line 316: I would suggest a small change here:
1. to avoid `proxyResponseTcs` need to be a global variable, you can define it inside the test now
2. no need for test cases to be a map
3. improve test isolation by creating a server only to handle a specific test case.

```suggestion
func mockProxy(tc) http.Handler {
	srvMux := mux.NewRouter()
	srvMux.HandleFunc("/v0/destinations/" + tc.destName "/proxy", func() { 
	vars := mux.Vars(req)
	destName, ok := vars["destName"]
	if !ok {
		// This case wouldn't occur I guess
		http.Error(w, "Wrong url being sent", http.StatusInternalServerError)
		return
	}
	// sleep is being used to mimic the waiting in actual transformer response
	if tc.Proxy.TimeoutValue.Milliseconds() > 0 {
		time.Sleep(tc.Proxy.TimeoutValue)
	}
	w.Header().Set("Content-Type", "application/json")
	w.WriteHeader(tc.Proxy.StCode)
	// Lint error fix
	checkAndSendResponse(&w, []byte(tc.Proxy.Response))

}
```

router/router.go
```diff
@@ -672,16 +672,29 @@ func (worker *workerT) processDestinationJobs() {
...
 								pkgLogger.Debugf(`responseTransform status :%v, %s`, worker.rt.transformerProxy, worker.rt.destName)
-								sendCtx, cancel := context.WithTimeout(ctx, worker.rt.netClientTimeout)
-								defer cancel()
+								var sendCtx context.Context
+								var cancel context.CancelFunc
 								// transformer proxy start
 								if worker.rt.transformerProxy {
-									jobID := destinationJob.JobMetadataArray[0].JobID
-									pkgLogger.Debugf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Request started`, worker.rt.destName, jobID)
+									jobId := destinationJob.JobMetadataArray[0].JobID
```
atzoum, line 679: Go idioms :) 
```suggestion
									jobID := destinationJob.JobMetadataArray[0].JobID
```

router/transformer/transformer_test.go
```diff
@@ -0,0 +1,323 @@
...
+	// ExpectedBody   string
+	// proxyResponse  string
+	// ExpectedStCode int
+	// // The response status code from proxy endpoint in transformer
+	// proxyStatusCode int
+	// // The delay in response from proxy endpoint in transformer
+	// proxyTimeoutValue time.Duration
+	// The timeout we have set from router
+	// Router will have the timeout of rtTimeout + <timeout_at_router_transform>
+	// For http client timeout scenarios, we need to have a proxyTimeout which is > rtTimeout + <timeout_at_router_transform>
```
atzoum, line 54: are those needed?

router/router.go
```diff
@@ -672,16 +672,29 @@ func (worker *workerT) processDestinationJobs() {
...
+									prxCtxTimeout := worker.rt.backendProxyTimeout + worker.rt.netClientTimeout + 20*time.Millisecond
+									sendCtx, cancel = context.WithTimeout(ctx, prxCtxTimeout)
+									defer cancel()
 									rtlTime := time.Now()
-									respStatusCode, respBodyTemp = worker.rt.transformer.ProxyRequest(ctx, val, worker.rt.destName, jobID)
+									// Should we change this context to sendCtx and probably adjust the timeout for the context ?
+									respStatusCode, respBodyTemp, respContentType = worker.rt.transformer.ProxyRequest(sendCtx, proxyReqparams)
 									worker.routerProxyStat.SendTiming(time.Since(rtlTime))
-									pkgLogger.Debugf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Request ended`, worker.rt.destName, jobID)
+									pkgLogger.Debugf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Request ended`, worker.rt.destName, jobId)
```
atzoum, line 697:
```suggestion
									pkgLogger.Debugf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Request ended`, worker.rt.destName, jobID)
```

router/transformer/transformer_test.go
lvrach: I would invert the logic of these tests, making the expectations more powerful.

```go
else if tc.ExpectedBody != "" {
				expectedBodyStr := fmt.Sprintf(tc.ExpectedBody, srv.URL)
				require.Equal(t, expectedBodyStr, resp)
}
```

If ExpectedBody is defined, we need to make the check, we don't want additional business logic on your testing here.

If there is some condition you don't care about or you don't expect a body just don't define it.

Same goes for the test above, which is quite dangerous:
```go
if gjson.GetBytes([]byte(resp), "message").Raw != "" {
				require.JSONEq(t, tc.ExpectedBody, resp)
			}
```
If you don't find the message in your response you are not doing the comparison, unexpected behaviour can slip in from that.

router/router.go
sanpj2292: sendCtx, cancel are being used for both SendPost(Router -> destination), ProxyRequest(Router -> proxy). So I think we should declare it beforehand. Let me know your thoughts

router/router.go
lvrach, line 691: I am unsure if the introduction of `sendCtx` is a good idea here.

Even the way we compute the timeout portraits we would probably need two different ctx with timeouts:
```
									prxCtxTimeout := worker.rt.backendProxyTimeout + worker.rt.netClientTimeout + 20*time.Millisecond
```
My suggestion would be to introduce this timeout and context inside the `transformer.ProxyRequest` and the other on `.SendPost` (if I have understood the code correctly)

And altogether remove `sendCtx` from the router code.

router/router.go

router/transformer/transformer_test.go
sanpj2292: I didn't completely understand what you meant here. Here's the change I guess you wanted me to make, kindly correct my understanding
```go
			parentctx := context.Background()
			var ctx context.Context
			var cancelFunc context.CancelFunc
			if tc.context.Timeout.Milliseconds() > 0 {
				ctx, cancelFunc = context.WithTimeout(parentctx, tc.context.Timeout)
				defer cancelFunc()
			} else if tc.context.Cancel {
				ctx, cancelFunc = context.WithCancel(parentctx)
				cancelFunc()
			} else {
				ctx = parentctx
			}
```

router/router.go
atzoum: Neither of them makes sense outside the if else block, thus each block can simply define its own same variables

router/transformer/transformer_test.go
sanpj2292: I made a changes, can you once check and let me know if it feels good ?

router/transformer/transformer_test.go
sanpj2292: Changed it

router/transformer/transformer_test.go
sanpj2292: @lvrach can you once see if this is the way we want to implement ?

router/transformer/transformer_test.go
sanpj2292: @lvrach can you once check and confirm if the changes are satisfactory ?

router/router.go
sanpj2292: Done!

router/router.go
sanpj2292: Done!

router/router.go
sanpj2292: Done

router/router.go
sanpj2292: Agreed. Done

router/transformer/transformer.go
sanpj2292: Done

router/router.go

router/router.go

router/router.go

router/transformer/transformer_test.go
sanpj2292: Done

router/router.go

router/router.go
sanpj2292: Made changes according to our discussion. Can you once verify ?

atzoum (approved):

lvrach (approved):

PR #2259, opened 2022-08-05
fix: backend config auth

fracasula:
# Description

When the BackendConfig API was returning a non successful response we were not erroring and that caused the BackendConfig to result as initialized. That happened for example if we were to start the BC with bad credentials.

## Additional changes
* I started removing some global variables
* extra `workspaceConfig` interface to simplify tests
* instead of having `CommonBackendConfig` embedded in the `workspaceConfig` structs I'm using the opposite approach. this simplifies tests and makes the dependencies in the `workspaceConfig` structs more clear and isolated
* new `logger.NOP` to use in tests where we don't want to be forced to express expectations about the logger, because those are quite often not useful and prone to useless changes.
* regulation-worker cmd `main_test.go` cleanup
* removed Ginkgo from `/config/backend-config` due to poor isolation, excessive use of mocks and tests not returning without the ability to tell which one didn't return (see [here](https://github.com/rudderlabs/rudder-server/runs/7705098570?check_suite_focus=true))
* unexporting structs that are not meant to be initialized on their own i.e. `multiTenantWorkspacesConfig`, `singleWorkspaceConfig` and `namespaceConfig`

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/BackendConfig-is-initialized-when-credentials-are-bad-e0c2d684739948afa5b99a2c1c48afae)

## Security


config/backend-config/backend-config.go
```diff
@@ -193,37 +189,38 @@ Available topics are:
...
 	case deployment.MultiTenantType:
 		isNamespaced := config.IsEnvSet("WORKSPACE_NAMESPACE")
 		if isNamespaced {
-			backendConfig = &NamespaceConfig{
-				CommonBackendConfig: CommonBackendConfig{
-					configEnvHandler: configEnvHandler,
-					eb:               pubsub.New(),
-				},
+			backendConfig.workspaceConfig = &namespaceConfig{
+				ConfigBackendURL: parsedConfigBackendURL,
```
chandumlg, line 216: Any reason why this struct has ConfigBackendURL of type *url.URL and other two structs (singleWorkspaceConfig & multiTenantWorkspacesConfig) have configBackendURL of type string?

chandumlg (approved):
LGTM

config/backend-config/backend-config.go
```diff
@@ -291,6 +288,33 @@ func (bc *CommonBackendConfig) WaitForConfig(ctx context.Context) {
 	}
 }
 
+func (bc *commonBackendConfig) SetUp() error        { return bc.workspaceConfig.SetUp() }
```
atzoum, line 291: we could get away with all these method declarations by embedding `workspaceConfig` interface in `commonBackendConfig` struct

```
type commonBackendConfig struct {
    workspaceConfig
    ....
}
```

config/backend-config/backend-config.go
```diff
@@ -23,52 +25,56 @@ import (
...
 
-type CommonBackendConfig struct {
-	eb               *pubsub.PublishSubscriber
-	configEnvHandler types.ConfigEnvI
-	ctx              context.Context
-	cancel           context.CancelFunc
-	blockChan        chan struct{}
-	initializedLock  sync.RWMutex
-	initialized      bool
+type commonBackendConfig struct {
```
atzoum, line 64: its name could be further simplified to `backendConfig`

config/backend-config/backend-config.go
```diff
@@ -23,52 +25,56 @@ import (
...
+	workspaceConfig   workspaceConfig
+	eb                *pubsub.PublishSubscriber
+	ctx               context.Context
+	cancel            context.CancelFunc
+	blockChan         chan struct{}
+	initializedLock   sync.RWMutex
+	initialized       bool
+	curSourceJSON     ConfigT
+	curSourceJSONLock sync.RWMutex
```
atzoum, line 73: moving the lock above the resource it protects is preferred

config/backend-config/backend-config.go
fracasula: It was born that way (the namespace config) and I didn't refactor the other two to be consistent. I'll have a look and see if I can refactor them easily once we merge @lvrach [PR](https://github.com/rudderlabs/rudder-server/pull/2269). I expect some conflicts upon merging, so I'll have to do some more work here anyway.

config/backend-config/multitenant_workspaces.go
```diff
@@ -136,41 +139,46 @@ func (workspaceConfig *MultiTenantWorkspacesConfig) getFromAPI(ctx context.Conte
...
+		return nil, http.StatusBadRequest, err
 	}
 
-	var respBody []byte
-	if resp != nil && resp.Body != nil {
-		respBody, _ = IoUtil.ReadAll(resp.Body)
-		defer func() { _ = resp.Body.Close() }()
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode >= 300 {
```
atzoum, line 180: since we appear to be performing error logic based on the status code here, does it make sense to remove the status code from this method's return values too?

config/backend-config/multitenant_workspaces.go
```diff
@@ -136,41 +139,46 @@ func (workspaceConfig *MultiTenantWorkspacesConfig) getFromAPI(ctx context.Conte
...
+		return nil, resp.StatusCode, getNotOKError(respBody, resp.StatusCode)
 	}
 
 	return respBody, resp.StatusCode, nil
```
atzoum, line 184: Question: previously we were not handling HTTP response codes properly, but we were still trying to unmarshall the response to the proper Config struct. Didn't this step fail?

config/backend-config/admin.go
```diff
@@ -19,7 +19,7 @@ func (bca *BackendConfigAdmin) RoutingConfig(filterProcessor bool, reply *string
 		}
 	}()
 
-	outputJSON := GetConfig()
+	outputJSON := DefaultBackendConfig.GetConfig()
```
atzoum, line 22: Aren't we at the same package? Can we unexport the `GetConfig` method?

regulation-worker/cmd/main.go
```diff
@@ -45,8 +46,12 @@ func main() {
 }
 
 func Run(ctx context.Context) {
+	admin.Init()
```
atzoum, line 49: question: why do we need this here now?

config/backend-config/backend-config.go
fracasula: @chandumlg it was super easy, it's done now :+1:

config/backend-config/backend-config.go
fracasula: where did you get this? as long as they are 1 line closed to each other it's cool imho :thinking:

config/backend-config/multitenant_workspaces.go
fracasula: yes!

config/backend-config/admin.go
fracasula: Oh yes, good point, I think I can remove it now, thanks :+1:

regulation-worker/cmd/main.go
fracasula: Because the backend config needs to register an admin handler, see [here](https://github.com/rudderlabs/rudder-server/blob/35238ebf0d426d0e21a7fb0e827e603f95180a66/config/backend-config/backend-config.go#L247).

config/backend-config/backend-config.go
atzoum: It seems to be a typical pattern in Go's own repository, e.g [here](https://github.com/golang/go/blob/2580d0e08d5e9f979b943758d3c49877fb2324cb/src/net/rpc/jsonrpc/server.go#L31), [here](https://github.com/golang/go/blob/2580d0e08d5e9f979b943758d3c49877fb2324cb/src/syscall/js/func.go#L12) & [here](https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/net/textproto/pipeline.go#L29), unless it is a test where they seem to be more relaxed, e.g. [here](https://github.com/golang/go/blob/d6d2ebb7b8194eb20d95b84bae39f7a5837e9f72/test/typeparam/genembed2.go#L22).
I remember reading it initially in some article and it made sense. I don't remember the exact article, but I could find a similar one [here](https://medium.com/hackernoon/dancing-with-go-s-mutexes-92407ae927bf#3426) as well.

regulation-worker/cmd/main.go
lvrach: Maybe we can:
1. Move this registration outside and only do this register for rudder-server
2. Or ensure that registration can happen without admin.Init() - some kind of self init way

lvrach (approved):

config/backend-config/multitenant_workspaces.go
fracasula: If we don't fail by checking the status code first here, we could get some unexpected results when unmarshalling as it's not reliable in some cases. For example when unmarshalling after a bad credentials response we get this: `{"message":"Invalid workspace token"}` (which is a valid JSON). That could trick the marshaller and accept it as valid.

```go
func TestAris(t *testing.T) {
	data := []byte(`{"message":"Invalid workspace token"}`)
	var sourcesJSON backendconfig.ConfigT
	err := json.Unmarshal(data, &sourcesJSON)
	require.NoError(t, err)
}
```

The above test passes.

config/backend-config/backend-config.go
fracasula: Interesting stuff, thanks. Not sure I feel the need though :man_shrugging:

regulation-worker/cmd/main.go
fracasula: Makes sense, although I would prefer to move it once we need the backend config to function without the admin package. I'm saying this because I have done lots of changes here already. The more I keep doing the more likely this PR could break something.

atzoum (approved):

PR #2266, opened 2022-08-08
fix(warehouse): reset priority if earlier upload is already in progress

dhawal1248:
# Description
Reset priority if uploads are already in progress

## Notion Ticket

https://www.notion.so/rudderstacks/Reset-priority-for-if-uploads-are-in-progress-fb4cb2af50484b069ac1cff30585feb3

## Security

### DIFF
--- a/warehouse/warehouse.go
+++ b/warehouse/warehouse.go
@@ -620,16 +620,17 @@ func (wh *HandleT) createJobs(warehouse warehouseutils.WarehouseT) (err error) {
 	}
 
 	wh.areBeingEnqueuedLock.Lock()
-	uploadID, uploadStatus, priority := wh.getLatestUploadStatus(&warehouse)
+
+	priority := 0
+	uploadID, uploadStatus, uploadPriority := wh.getLatestUploadStatus(&warehouse)
 	if uploadStatus == Waiting {
 		// If it is present do nothing else delete it
 		if _, inProgess := wh.isUploadJobInProgress(warehouse, uploadID); !inProgess {
 			wh.deleteWaitingUploadJob(uploadID)
-		} else {
-			// reset priority if the upload is already in progress
-			priority = 0
+			priority = uploadPriority // copy the priority from the latest upload job.
 		}
 	}
+
 	wh.areBeingEnqueuedLock.Unlock()
 
 	stagingFilesFetchStat := stats.DefaultStats.NewTaggedStat("wh_scheduler.pending_staging_files", stats.TimerType, stats.Tags{
--- a/warehouse/warehouse.go
+++ b/warehouse/warehouse.go
@@ -625,6 +625,9 @@ func (wh *HandleT) createJobs(warehouse warehouseutils.WarehouseT) (err error) {
 		// If it is present do nothing else delete it
 		if _, inProgess := wh.isUploadJobInProgress(warehouse, uploadID); !inProgess {
 			wh.deleteWaitingUploadJob(uploadID)
+		} else {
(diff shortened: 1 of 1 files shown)


warehouse/warehouse.go
```diff
@@ -625,6 +625,9 @@ func (wh *HandleT) createJobs(warehouse warehouseutils.WarehouseT) (err error) {
 		// If it is present do nothing else delete it
 		if _, inProgess := wh.isUploadJobInProgress(warehouse, uploadID); !inProgess {
 			wh.deleteWaitingUploadJob(uploadID)
+		} else {
+			// reset priority if the upload is already in progress
+			priority = 0
```
achettyiitr, line 630: Do we need to set this value to default (100) for resetting? Because processor queries for ascending order while picking up the JOB.

Jayachand (approved):

warehouse/warehouse.go
achettyiitr, line 630: Should we reset in the case of nonwaiting jobs as well? Because there also we don't need it.

achettyiitr (changes requested):

warehouse/warehouse.go
dhawal1248: non waiting as in? anything that is in progress, is non waiting right?

warehouse/warehouse.go
```diff
@@ -625,6 +625,9 @@ func (wh *HandleT) createJobs(warehouse warehouseutils.WarehouseT) (err error) {
 		// If it is present do nothing else delete it
```
abhimanyubabbar, line 625: Me and @sivashanmukh had a call yesterday, we discussed mainly about the following:
To have a defaultPriority and reset the priority in case you have a waiting job which is !inProgress.
 
 so
```
defaultPriority := 0
uploadID, uploadStatus, priority := wh.getLatestUploadStatus(&warehouse)
if uploadStatus == Waiting {
    if _, inProgess := wh.isUploadJobInProgress(warehouse, uploadID); !inProgess {
		wh.deleteWaitingUploadJob(uploadID)
        defaultPriority = priority
    }
}

...

wh.createUploadJobsFromStagingFiles(warehouse, whManager, stagingFilesList, defaultPriority, uploadStartAfter)

```

abhimanyubabbar: Wdyt ? @dhawal1248

achettyiitr: > Me and @sivashanmukh had a call yesterday, we discussed mainly about the following: To have a defaultPriority and reset the priority in case you have a waiting job which is !inProgress.
> 
> so
> 
> ```
> defaultPriority := 0
> uploadID, uploadStatus, priority := wh.getLatestUploadStatus(&warehouse)
> if uploadStatus == Waiting {
>     if _, inProgess := wh.isUploadJobInProgress(warehouse, uploadID); !inProgess {
> 		wh.deleteWaitingUploadJob(uploadID)
>         defaultPriority = priority
>     }
> }
> 
> ...
> 
> wh.createUploadJobsFromStagingFiles(warehouse, whManager, stagingFilesList, defaultPriority, uploadStartAfter)
> ```

+1

warehouse/warehouse.go
achettyiitr: I mean for jobs which we exported for those also we don't need to do it.

warehouse/warehouse.go
dhawal1248: Makes sense @abhimanyubabbar

sivashanmukh (approved):
LGTM

PR #2267, opened 2022-08-08
refactor: destination config from interface{} to map[string]interface{}

koladilip:
# Description
We were using slightly inefficient way to populate SessionConfig from
destination config so in order to improve this, we need to take destination
config as map[string]interface{}.

Rohith suggested this change in the PR#2241

## Notion Ticket

[FTR for Streaming Destinations](https://www.notion.so/rudderstacks/AWS-FTR-for-streaming-destination-78b63869ac904ba49b0018a527ccb343)

## Security

### DIFF
--- a/utils/awsutils/session.go
+++ b/utils/awsutils/session.go
@@ -70,8 +70,10 @@ func NewSessionConfig(destinationConfig map[string]interface{}, timeout time.Dur
 		return nil, errors.New("destinationConfig should not be nil")
 	}
 	sessionConfig := SessionConfig{}
-	mapstructure.Decode(destinationConfig, &sessionConfig)
-
+	err := mapstructure.Decode(destinationConfig, &sessionConfig)
+	if err != nil {
+		return nil, errors.New("unable to populate session config using destinationConfig")
+	}
 	sessionConfig.Timeout = timeout
 	sessionConfig.Service = serviceName
 	return &sessionConfig, nil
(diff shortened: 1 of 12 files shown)


cisse21 (approved):

utils/awsutils/session.go
```diff
@@ -64,16 +64,36 @@ func CreateSession(config *SessionConfig) *session.Session {
...
+func NewSessionConfig(destinationConfig map[string]interface{}, timeout time.Duration, serviceName string) (*SessionConfig, error) {
+	if destinationConfig == nil {
+		return nil, errors.New("destinationConfig should not be nil")
+	}
 	sessionConfig := SessionConfig{}
 
-	jsonConfig, err := json.Marshal(destinationConfig)
-	if err != nil {
-		return nil, fmt.Errorf("[%s] Error while marshalling destination config :: %w", serviceName, err)
+	if region, ok := destinationConfig["Region"].(string); ok {
```
saikumarrs, line 73: Would this (with some tweaks) be a more elegant way of doing the same?
https://stackoverflow.com/a/26746461/3967127

It's better to avoid code like this.

koladilip: Updated the code to use mapstructure. Initially thought about using it but I have been told that we avoid using open sources unless absolutely needed so written manually, but mapstructure simplifies this process.

saikumarrs (approved):

PR #2275, opened 2022-08-09
chore(warehouse): warehouse retry uploads count

achettyiitr:
# Description

Provide the flexibility to count the failed syncs either by providing the below items:

- Retry interval.
- Uploads ids.

## Notion Ticket

Engineering doc => https://www.notion.so/rudderstacks/Multi-Retry-admin-api-9d11979938624263bcc7e6b4989177ad
Task => https://www.notion.so/rudderstacks/Warehouse-multi-admin-ffe0e0ff340a48be87bfa167efdc0875

## Security


warehouse/retry.go
```diff
@@ -122,6 +177,35 @@ func (retryReq *RetryRequest) retryUploads(sourceIDs []string) (rowsAffected int
...
+		SELECT
+		  count(*)
+		FROM
+		  wh_uploads
+		WHERE %[1]s`,
+		clausesQuery,
+	)
+
+	// Executing the statement
+	err = retryReq.API.dbHandle.QueryRowContext(ctx, preparedStatement, clausesArgs...).Scan(&count)
```
abhimanyubabbar, line 205: I would refactor this query to the dbModule we created and call that over the retryReq.API.dbHandle.GetUploadsCount(...filterClause string). Wdyt ?

It will allow us to have a generic function call to count the entries in the wh_uploads with filter params and can be used by other candidates as well.

abhimanyubabbar (changes requested):

warehouse/retry.go
achettyiitr: Thanks for pointing this out. It's a very good motivation to start moving the Queries to DB module.

achettyiitr: Done.

warehouse/db.go
```diff
@@ -3,11 +3,17 @@ package warehouse
...
 	"database/sql"
+	"errors"
 	"fmt"
+	"strings"
 
 	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
 )
 
+const (
+	queryPlaceHolder = "<noop>"
```
achettyiitr, line 14: I guess not need to have this with the current implementation.

Jayachand (approved):

warehouse/retry.go
```diff
@@ -64,108 +64,133 @@ func (retryReq *RetryRequest) RetryWHUploads() (response RetryResponse, err erro
...
+		if err != nil {
+			retryReq.API.log.Errorf(`WH: Error occurred while fetching upload jobs to retry for workspaceId: %s, sourceId: %s, destinationId: %s with error: %s`,
+				retryReq.WorkspaceID,
+				retryReq.SourceID,
+				retryReq.DestinationID,
+				err.Error(),
+			)
+			response = RetryResponse{
+				Message:    err.Error(),
+				StatusCode: 400,
```
abhimanyubabbar, line 93: Would every outcome from this function be interpreted as `BadRequest` ?

warehouse/retry.go
```diff
@@ -64,108 +64,133 @@ func (retryReq *RetryRequest) RetryWHUploads() (response RetryResponse, err erro
...
-			updated_at = NOW()
-		WHERE %[1]s`,
-		clausesQuery,
-	)
+	// Getting the corresponding sourceIDs from workspaceID (sourceIDsByWorkspace)
+	// Also, validating if the sourceIDs list is empty or provided sourceId is present in the
+	// sources list ot not.
+	sourceIDs := retryReq.getSourceIDs()
+	if len(sourceIDs) == 0 {
+		err = fmt.Errorf("unauthorized request")
```
abhimanyubabbar, line 107: Why unauthorized request here ? Just Curious

abhimanyubabbar (approved):
Except for couple of comments, it looks good to me 👍🏼

warehouse/retry.go
achettyiitr: Currently, for every error, we are returning 400 for all Grpc calls.

achettyiitr: Although it doesn't matter what status code we send because cp-router always sends 500 for errors.

PR #2276, opened 2022-08-09
fix(router): handle transformer in/out inconsistencies

atzoum:
# Description

During router transform, the transformer may erroneously return more than 1 job per input jobID. This results in: 
- sending the same event multiple times to the destination, 
- messing with our statistics (negative pending events & rsources in/out inconsistency)

This change introduces two additional validations during router's `transformer.Transform`:
1. Validate that job numbers match (`in` equals `out`)
2. Validate that job ids match between `in` and `out` jobs

If any of the above validations fail, all jobs of this transformation request will be marked as failed (HTTP 500) so that they can be retried later.
Additionally, a new counter has been introduced `router.transformer.invalid.response` against which an alert can be configured to get notified about such errors.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=b3ef63c8b066467ba9dd6dcaf3bef09b&pm=s)

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -316,22 +316,6 @@ func (worker *workerT) batch(routerJobs []types.RouterJobT) []types.DestinationJ
 	worker.rt.batchOutputCountStat.Count(len(destinationJobs))
 	worker.recordStatsForFailedTransforms("batch", destinationJobs)
 
-	var totalJobMetadataCount int
-	for _, destinationJob := range destinationJobs {
-		totalJobMetadataCount += len(destinationJob.JobMetadataArray)
-	}
-
-	if inputJobsLength != totalJobMetadataCount {
-		worker.rt.batchInputOutputDiffCountStat.Count(inputJobsLength - totalJobMetadataCount)
-
-		worker.rt.logger.Errorf("[%v Router] :: Total input jobs count:%d did not match total job metadata count:%d returned from batch transformer", worker.rt.destName, inputJobsLength, totalJobMetadataCount)
-		jobIDs := make([]string, len(routerJobs))
-		for idx, routerJob := range routerJobs {
-			jobIDs[idx] = fmt.Sprintf("%v", routerJob.JobMetadata.JobID)
-		}
-		worker.rt.logger.Errorf("[%v Router] :: Job ids : %s", worker.rt.destName, strings.Join(jobIDs, ", "))
-	}
-
 	return destinationJobs
 }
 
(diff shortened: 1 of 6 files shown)


router/transformer/transformer_test.go
```diff
@@ -0,0 +1,185 @@
...
+			{JobMetadata: types.JobMetadataT{JobID: 3}},
+		},
+	}
+	transformerResponse := tr.Transform(BATCH, &transformMessage)
+	normalizeErrors(transformerResponse, expectedErrorTxt)
+	require.NotNil(t, transformerResponse)
+	require.Equal(t, expectedTransformerResponse, transformerResponse)
+}
+
+func TestTransformValidationJobIDMismatchError(t *testing.T) {
```
lvrach, line 125: [minor/suggestion] You could have used table-driven tests here, to reduce the setup overhead (although small in this case) and make it easier to write multiple scenarios.

Great job writing those tests.

router/router.go
```diff
@@ -2236,7 +2196,6 @@ func (rt *HandleT) Setup(backendConfig backendconfig.BackendConfig, jobsDB jobsd
 	rt.routerResponseTransformStat = stats.NewTaggedStat("response_transform_latency", stats.TimerType, stats.Tags{"destType": rt.destName})
 
 	rt.transformer = transformer.NewTransformer()
-	rt.transformer.Setup()
```
lvrach, line 2239: 👍

lvrach (commented):
LGTM so far, I like the refactor touches as well

router/transformer/transformer_test.go
atzoum: I find it more readable this way & the overhead is not too much (IMO)

lvrach (approved):

router/transformer/transformer_test.go
```diff
@@ -325,20 +321,184 @@ func mockProxyHandler(tc *ProxyResponseTc) *mux.Router {
...
+	mockRudderStats := mock_stats.NewMockRudderStats(ctrl)
+
+	mockStats.EXPECT().NewTaggedStat(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes().Return(mockRudderStats)
+	mockStats.EXPECT().NewStat(gomock.Any(), gomock.Any()).AnyTimes().Return(mockRudderStats)
+	mockRudderStats.EXPECT().SendTiming(gomock.Any()).AnyTimes()
+	mockRudderStats.EXPECT().Increment().AnyTimes()
+
+	stats.DefaultStats = mockStats
+
+	pkgLogger = logger.NOP{}
```
fracasula, line 494: :heart:

fracasula (approved):

PR #2240, opened 2022-08-03
chore: remove hosted deployment type

lvrach:
# Description

Deployment type Hosted is no longer needed since the successful introduction of multi-tenant.

## Notion Ticket

[notion link
](https://www.notion.so/rudderstacks/Cleanup-hosted-deployment-type-b8d78304e6ee42158816a5e6bed455cb)
## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #2278, opened 2022-08-10
chore: fix issues during cleanup

cisse21:
# Description

When we cleaned up hosted workspace config, We missed a few things very specific to hosted deployment [here](https://github.com/rudderlabs/rudder-server/pull/2240/files#diff-4f799306a3605b051353f0aa7463da845b95309e09520a954272f41541097a69L131) . Adding the missing pieces

## Notion Ticket

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=3a0652cc416042a081dc308ba8b28972&pm=s

## Security

### DIFF
--- a/config/backend-config/backend-config.go
+++ b/config/backend-config/backend-config.go
@@ -27,6 +27,7 @@ import (
 var (
 	// environment variables
 	configBackendURL                      string
+	cpRouterURL                           string
 	pollInterval, regulationsPollInterval time.Duration
 	configJSONPath                        string
 	configFromFile                        bool
@@ -38,8 +39,8 @@ var (
 
 	// DefaultBackendConfig will be initialized be Setup to either a WorkspaceConfig or MultiWorkspaceConfig.
 	DefaultBackendConfig BackendConfig
-	pkgLogger            logger.LoggerI   = logger.NewLogger().Child("backend-config")
-	IoUtil               sysUtils.IoUtilI = sysUtils.NewIoUtil()
+	pkgLogger            = logger.NewLogger().Child("backend-config")
+	IoUtil               = sysUtils.NewIoUtil()
 	Diagnostics          diagnostics.DiagnosticsI
 )
 
@@ -74,6 +75,7 @@ type backendConfigImpl struct {
 
 func loadConfig() {
 	configBackendURL = config.GetEnv("CONFIG_BACKEND_URL", "https://api.rudderlabs.com")
+	cpRouterURL = config.GetEnv("CP_ROUTER_URL", "https://cp-router.rudderlabs.com")
 	config.RegisterDurationConfigVariable(5, &pollInterval, true, time.Second, []string{"BackendConfig.pollInterval", "BackendConfig.pollIntervalInS"}...)
 	config.RegisterDurationConfigVariable(300, &regulationsPollInterval, true, time.Second, []string{"BackendConfig.regulationsPollInterval", "BackendConfig.regulationsPollIntervalInS"}...)
 	config.RegisterStringConfigVariable("/etc/rudderstack/workspaceConfig.json", &configJSONPath, false, "BackendConfig.configJSONPath")
@@ -100,7 +102,7 @@ func trackConfig(preConfig, curConfig ConfigT) {
 		noOfSources := len(curConfig.Sources)
 		noOfDestinations := 0
 		for _, source := range curConfig.Sources {
-			noOfDestinations = noOfDestinations + len(source.Destinations)
+			noOfDestinations += len(source.Destinations)
 		}
 		Diagnostics.Track(diagnostics.ConfigProcessed, map[string]interface{}{
 			diagnostics.SourcesCount:      noOfSources,
@@ -137,7 +139,7 @@ func (bc *backendConfigImpl) configUpdate(ctx context.Context, statConfigBackend
 
 	// sorting the sourceJSON.
 	// json unmarshal does not guarantee order. For DeepEqual to work as expected, sorting is necessary
-	sort.Slice(sourceJSON.Sources[:], func(i, j int) bool {
+	sort.Slice(sourceJSON.Sources, func(i, j int) bool {
 		return sourceJSON.Sources[i].ID < sourceJSON.Sources[j].ID
 	})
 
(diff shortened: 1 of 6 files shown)


Sidddddarth (approved):

config/backend-config/multitenant_workspaces_test.go
```diff
@@ -126,7 +126,9 @@ func TestMultiTenantWorkspacesConfig_Get(t *testing.T) {
 		}
 		conf, err := wc.Get(context.Background(), "")
 		require.NoError(t, err)
-		require.Equal(t, sampleBackendConfig, conf)
+		multiTenantWorkspacesConfig := sampleBackendConfig
+		multiTenantWorkspacesConfig.ConnectionFlags = ConnectionFlags{URL: "", Services: map[string]bool{"warehouse": true}}
```
fracasula, line 130: You could do a `t.Setenv("CP_ROUTER_URL")` and verify that the `ConnectionFlags.URL` gets that value instead of just `""`.

fracasula (changes requested):

config/backend-config/multitenant_workspaces.go
```diff
@@ -143,6 +143,9 @@ func (wc *multiTenantWorkspacesConfig) getFromAPI(ctx context.Context, _ string)
 		}
 		sourcesJSON.Sources = append(sourcesJSON.Sources, workspaceConfig.Sources...)
 	}
+	sourcesJSON.ConnectionFlags.URL = config.GetEnv("CP_ROUTER_URL", "")
```
fracasula, line 146: Instead of using `config.GetEnv` directly inside a method that shouldn't depend nor be aware of `config` or environment variables, we should use struct attributes that are way easier to tests and that decouple the struct from the above. Also it is best not to spread "construction" logic everywhere. I think we should move `config.GetEnv("CP_ROUTER_URL, "")` in the `loadConfig()` function and then have the constructor (i.e. `newForDeployment`) inject the `cpRouterURL` into the `multiTenantWorkspacesConfig`:

```go
backendConfig.workspaceConfig = &multiTenantWorkspacesConfig{
	configBackendURL: parsedConfigBackendURL,
	configEnvHandler: configEnvHandler,
	cpRouterURL: cpRouterURL,
}
```

cisse21: Done

config/backend-config/single_workspace.go
```diff
@@ -22,6 +22,7 @@ type singleWorkspaceConfig struct {
 	configBackendURL *url.URL
 	configJSONPath   string
 	configEnvHandler types.ConfigEnvI
+	cpRouterURL      string
```
fracasula, line 25: This doesn't seem to be needed. How come you added it?

config/backend-config/namespace_config.go
```diff
@@ -154,7 +154,9 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context, _ string) (ConfigT, e
 		}
 		sourcesJSON.Sources = append(sourcesJSON.Sources, nc.Sources...)
 	}
-
+	sourcesJSON.ConnectionFlags.URL = nc.cpRouterURL
+	// always set connection flags to true for hosted and multi-tenant warehouse service
+	sourcesJSON.ConnectionFlags.Services = map[string]bool{"warehouse": true}
```
fracasula, line 159: I don't see a test in `namespace_config_test.go` about this change. Can you please add it?

config/backend-config/multitenant_workspaces_test.go
```diff
@@ -97,11 +97,13 @@ func TestMultiTenantWorkspacesConfig_Get(t *testing.T) {
...
 			secretToken            = "multitenantWorkspaceSecret"
-			workspaceId            = "testWorkspaceId"
+			workspaceID            = "testWorkspaceId"
 			sampleWorkspaceSources = map[string]ConfigT{
-				workspaceId: sampleBackendConfig,
+				workspaceID: sampleBackendConfig,
 			}
+			cpRouterURL = "mockCPRouterURL"
 		)
+		t.Setenv("CP_ROUTER_URL", cpRouterURL)
```
fracasula, line 106: Is this needed at all now that we got the URL as a struct attribute?

config/backend-config/backend-config.go
```diff
@@ -208,19 +210,22 @@ func newForDeployment(deploymentType deployment.Type, configEnvHandler types.Con
 			configJSONPath:   configJSONPath,
 			configBackendURL: parsedConfigBackendURL,
 			configEnvHandler: configEnvHandler,
+			cpRouterURL:      cpRouterURL,
```
fracasula, line 213: It doesn't look like we're using this one.

fracasula (changes requested):

config/backend-config/multitenant_workspaces_test.go
cisse21: Removed

config/backend-config/namespace_config.go
cisse21: Added the test

config/backend-config/single_workspace.go
cisse21: Wanted to remain consistent across the three modules but removed it now

config/backend-config/backend-config.go
cisse21: Removed this too

config/backend-config/namespace_config_test.go
```diff
@@ -80,6 +82,10 @@ func Test_Namespace_Get(t *testing.T) {
 	require.Equal(t, workspaceID1, client.GetWorkspaceIDForSourceID("2CCgpZlqlXRDRz8rChhQKtuwqKA"))
 	require.Equal(t, workspaceID2, client.GetWorkspaceIDForSourceID("2CChOtDTWeXIQiRmHMU56C3htPf"))
 
+	t.Log("Connection Flaf Asserts")
```
fracasula, line 85: [optional] as there seems to be a typo

```suggestion
```

fracasula (approved):

PR #2281, opened 2022-08-10
chore(warehouse): batching clickhouse cluster alter statements

achettyiitr:
# Description

Some build in GA is failing with the following error: 

> code: 517, message: Metadata on replica is not up to date with common metadata in Zookeeper. It means that this replica still not applied some of previous alters. Probably too many alters executing concurrently (highly not recommended). You can retry this error

Ref: https://github.com/rudderlabs/rudder-server/runs/7763129218?check_suite_focus=true#step:8:1657

# Solution
ZooKeeper is not getting updated if we are running several alter statements for the same table for multiple columns at the same time and we are getting this error. In order to avoid this, we can merge several ALTER statements into a single one for the same table.
Ref: https://stackoverflow.com/questions/69257245/clickhouse-cannot-alter-columns-throws-dbexception-metadata-on-replica-is-not/69259570#69259570

## Notion Ticket

https://www.notion.so/rudderstacks/Warehouse-Integration-Tests-aceb9911fe574e62a3f3e6037947fd3f

## Security

### DIFF
--- a/warehouse/clickhouse/clickhouse_test.go
+++ b/warehouse/clickhouse/clickhouse_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/rudderlabs/rudder-server/warehouse/clickhouse"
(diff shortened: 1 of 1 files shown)


Jayachand (approved):
LGTM

satishrudderstack (approved):
LGTM

PR #2282, opened 2022-08-10
chore: gateway response code fixes

chandumlg:
# Description

We were returning 200 in some error cases as well (Ex: jobsdb store failures). This PR fixes those cases.

## Security


gateway/response/response.go
```diff
@@ -105,7 +105,7 @@ func GetStatusCode(key string) int {
 	if status, ok := statusMap[key]; ok {
 		return status.code
 	}
-	return http.StatusOK
+	return http.StatusInternalServerError
```
chandumlg, line 108: StatusInternalServerError is not always right. We might want to send StatusBadRequest also. 

This needs a discussion.

gateway/response/response.go
```diff
@@ -105,9 +105,9 @@ func GetStatusCode(key string) int {
...
```
atzoum, line 108: Since we only need to call this method for errors, we might as well rename it to `GetErrorStatusCode` to make it clear

gateway/response/response.go
```diff
@@ -105,9 +105,9 @@ func GetStatusCode(key string) int {
...
 }
 
 func MakeResponse(msg string) string {
-	return fmt.Sprintf(`{"msg": "%s"}`, msg)
+	return fmt.Sprintf("str: %q", msg)
```
atzoum, line 112: this doesn't look right & why would we want to change it?
```suggestion
	return fmt.Sprintf(`{"msg": %q}`, msg)
```

lvrach:
Why no tests were affected by those changes? Could we introduce some tests to cover those as well?

gateway/response/response.go
chandumlg: Deepsource :)

chandumlg: Oh Damn! I didn't realise it suggested the changed in the key. Will change it.

gateway/response/response.go
chandumlg: We make a call to this function with response.Ok key also. But only a few times compared to error responses.

atzoum: couldn't find such occurrences, can you please point me to them?

atzoum (approved):

cisse21 (approved):

PR #2115, opened 2022-06-27
fix(warehouse): support skipping escaping of reserved keywords

Jayachand:
# Description

- added skipping reserved keywords escaping for warehouse destinations
- reserved words must be enclosed in ("") or (``) to get it recognised as identifiers

## Notion Ticket

https://www.notion.so/rudderstacks/Fix-reserved-keywords-escaping-issue-ea609956e89a400d9aeeb4ced5bc334a

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #2291, opened 2022-08-12
chore(warehouse): adding support for random schema generation during Warehouse Integration Tests

achettyiitr:
# Description

Currently, during the Warehouse Integration Test, we are using the fixed schema. Issues are:

- Bloating of tables present in the fixed schema.
- Any changes in these schemas like in the case of https://github.com/rudderlabs/rudder-server/pull/2115 PR will affect all other tests as well. 
- If multiple of these tests are running parallel for the same tables where Warehouse does not support it, it might fail like in the case of Redshift, Deltalake.
- We might not be able to check CreateSchema scenarios as well.
- We are using uudigen (6CBEC569-3BE4-4620-AFA6-F0F209EE6A75).

## Notion Ticket

https://www.notion.so/rudderstacks/Use-Isolated-Schema-for-Warehouse-Integration-Tests-a99eec696aee4f2dabc8d37a9fd72428

## Security


Jayachand (approved):
LGTM

abhimanyubabbar (approved):
This looks good to me 👍🏼

PR #2294, opened 2022-08-12
chore: fix flaky integration test

cisse21:
# Description

Fix deadlock happening in concurrent start-stop scenarios during the integration tests

## Notion Ticket

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=f89bbfd437584ec7938f523f599195e4&pm=s

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -126,11 +126,10 @@ type HandleT struct {
 	sourceIDWorkspaceMap                   map[string]string
 	maxDSQuerySize                         int
 
-	backgroundGroup      *errgroup.Group
-	backgroundCtx        context.Context
-	backgroundCancel     context.CancelFunc
-	backgroundWait       func() error
-	generatorLoopStarted chan struct{}
+	backgroundGroup  *errgroup.Group
+	backgroundCtx    context.Context
+	backgroundCancel context.CancelFunc
+	backgroundWait   func() error
 
 	resultSetMeta    map[int64]*resultSetT
 	resultSetLock    sync.RWMutex
@@ -2260,7 +2259,6 @@ func (rt *HandleT) Setup(backendConfig backendconfig.BackendConfig, jobsDB jobsd
 	rt.backgroundGroup = g
 	rt.backgroundCancel = cancel
 	rt.backgroundWait = g.Wait
-	rt.generatorLoopStarted = make(chan struct{})
 	rt.initWorkers()
 
 	g.Go(misc.WithBugsnag(func() error {
@@ -2302,7 +2300,6 @@ func (rt *HandleT) Start() {
 			}
 		}
 		rt.generatorLoop(ctx)
-		close(rt.generatorLoopStarted)
 		return nil
 	})
 }
@@ -2310,7 +2307,6 @@ func (rt *HandleT) Start() {
 func (rt *HandleT) Shutdown() {
 	rt.logger.Infof("Shutting down router : %s", rt.destName)
 	rt.backgroundCancel()
-	<-rt.generatorLoopStarted
 	_ = rt.backgroundWait()
 }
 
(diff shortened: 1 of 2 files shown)


fracasula (approved):
This makes sense. So the `ctx` used in the go routine spawned by `Start()` is the `backgroundCtx` which is the same context as the very same group. And in `Shutdown()` we are waiting for that group to return gracefully but it will never return even if we cancel due to the circular dependency. Nice catch :+1:

router/router.go
```diff
@@ -2213,31 +2212,19 @@ func (rt *HandleT) Start() {
 	rt.logger.Infof("Starting router : %s", rt.destName)
```
atzoum, line 2212: shouldn't we use a lifecycle lock for controlling concurrent access to Start/Stop methods? E.g. similar to what we are doing in jobsdb?

lvrach: Currently, there is no need for concurrent support for Start/Stop. Following the standard library paradigm, I suggest not adding concurrent support unless needed. We can be explicit about concurrency support in the method documentation.

router/router.go
```diff
@@ -2213,31 +2212,19 @@ func (rt *HandleT) Start() {
...
 	ctx := rt.backgroundCtx
 	rt.backgroundGroup.Go(func() error {
-		select {
-		case <-ctx.Done():
-			rt.logger.Infof("Router : %s start goroutine exited", rt.destName)
-			return nil
-		case <-rt.backendConfigInitialized:
-			// no-op, just wait
-		}
+		<-rt.backendConfigInitialized
```
lvrach, line 2215: Why did we remove the select between `ctx.Done()` and `rt.backendConfigInitialized` are we sure router will Shutdown if backendconfig is not initialized?

atzoum: `Start` performs its operations asynchronously, through a go routine, effectively returning before it is actually done. So the error we are fixing here occurred because we were calling `Stop` before the router was actually started, right?

router/router.go
atzoum: Start performs its operations asynchronously, through a go routine, effectively returning before it is actually done. So the error we are fixing here occurred because we were calling Stop before the router was actually started, right?

router/router.go
cisse21: Without the generator loop starting the shutdown would not happen because for the shutdown to happen successfully we would need to close the worker channels. If we have `ctx.Done()` here and the `stop` function is called before the generator loop gets to start we would have the router in a stuck state which was happening in the tests

router/router.go
lvrach: Oh know I understood the exact issue.

Start should return only once it is ready to be stopped. A loop like:

```go
for {
   router.Start()
   router.Stop()
}
```
Should work like charm.

router/router.go
fracasula: The context is just the group context and this routine is part of the group so it doesn't make sense to wait for this context specifically, right?

router/router.go
fracasula: Yes but I don't think that should stop this PR from being merged. Its purpose is not fixing the setup/start/shutdown flow (remember that setup is also involved here). For that I had created a separate task [here](https://www.notion.so/rudderstacks/Router-setup-start-stop-80425af6cf2645b0a3d53620b8aca2f9). So I think this PR was just meant to improve the test situation with the plan to then fix the router setup/start/shutdown properly later.

Not strong on it though, just wanted to point it out. We can do it all together as well.

atzoum: IMO since the changes here are affecting non-test code, this is the time to make sure that start & stop behave correctly, as they should ;)

fracasula: Happy with that, @cisse21 what do you think?

cisse21: Yeah makes that sense too. I just raised this to unblock people from flaky tests causing them to rerun

PR #2295, opened 2022-08-12
chore(destination): change dest def name to destType for stats

sanpj2292:
# Description

In this PR, we want to change the stats which have the destination definition name detail to be `destType` as this is the convention followed in `rudder-server`

## Notion Ticket

https://www.notion.so/rudderstacks/Transformer-alerts-Revamp-5af8ab9e12bc4247b24515504dd74f9f

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -611,7 +611,7 @@ func (worker *workerT) processDestinationJobs() {
 									jobID := destinationJob.JobMetadataArray[0].JobID
 									pkgLogger.Debugf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Request started`, worker.rt.destName, jobID)
 									proxyReqparams := &transformer.ProxyRequestParams{
-										DestName:     strings.ToLower(worker.rt.destName),
+										DestName:     worker.rt.destName,
 										JobID:        jobID,
 										ResponseData: val,
 									}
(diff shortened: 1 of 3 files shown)


cisse21 (approved):

atzoum (approved):

router/router.go
```diff
@@ -666,7 +666,7 @@ func (worker *workerT) processDestinationJobs() {
 							)
 
 							stats.NewTaggedStat("transformer_proxy.output_events_count", stats.CountType, stats.Tags{
-								"destType":      worker.rt.destName,
+								"destType":      strings.ToLower(worker.rt.destName),
```
atzoum, line 669: why are we using lowercase?

sanpj2292: Basically the destination names are lower-cased in transformer, since we are sending a lower-cased value for `destType` tag for the stats send from transformer.
When we setup any alerts we need to group using `destType` which needs to in-sync with what we send from transformer

lvrach (approved):

router/router.go
lvrach: Not related to this PR, but we need to ensure that destType is represented in the same form across all of our services. It is used as an identifier in many instances, and having it in mixed cases will cause problems. My suggestion would be that transformer is changed to work with normal-casing, but not sure regarding the effort behind this.

atzoum (approved):

PR #2300, opened 2022-08-16
feat: use different http client timeouts for different components

BonapartePC:
# Description

This increases reporting's default timeout from `30s` to `60s` since most of the requests take more than `30s`
Various formatting changes introduced too, after running `make fmt`

## Notion Ticket

https://www.notion.so/rudderstacks/Increase-reporting-client-timeout-from-30s-to-60s-63f5c7dc84644a01a7f7aae71de90608

## Security

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -297,15 +297,6 @@ func checkAndHotReloadConfig(configMap map[string][]*ConfigVar) (hasConfigChange
 	return hasConfigChanged
 }
 
-func GetFirstBool(defaultValue bool, keys ...string) (value bool) {
-	for _, key := range keys {
-		if IsSet(key) {
-			return GetBool(key, defaultValue)
-		}
-	}
-	return defaultValue
-}
-
 // GetBool is a wrapper for viper's GetBool
 func GetBool(key string, defaultValue bool) (value bool) {
 	envVal := GetEnv(TransformKey(key), "")
@@ -317,15 +308,6 @@ func GetBool(key string, defaultValue bool) (value bool) {
 		return defaultValue
 	}
 	return viper.GetBool(key)
-}
-
-func GetFirstInt(defaultValue int, keys ...string) (value int) {
-	for _, key := range keys {
-		if IsSet(key) {
-			return GetInt(key, defaultValue)
-		}
-	}
-	return defaultValue
 }
 
 // GetInt is wrapper for viper's GetInt
@@ -605,15 +587,6 @@ func RegisterStringSliceConfigVariable(defaultValue []string, ptr *[]string, isH
 	}
 }
 
-func GetFirstInt64(defaultValue int64, keys ...string) (value int64) {
-	for _, key := range keys {
-		if IsSet(key) {
-			return GetInt64(key, defaultValue)
-		}
-	}
-	return defaultValue
-}
-
 // GetInt64 is wrapper for viper's GetInt
 func GetInt64(key string, defaultValue int64) (value int64) {
 	envVal := GetEnv(TransformKey(key), "")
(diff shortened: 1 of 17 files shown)


atzoum (approved):

enterprise/reporting/reporting.go
```diff
@@ -295,7 +295,7 @@ func (handle *HandleT) getAggregatedReports(reports []*types.ReportByStatus) []*
 
 func (handle *HandleT) mainLoop(ctx context.Context, clientName string) {
 	tr := &http.Transport{}
-	netClient := &http.Client{Transport: tr, Timeout: config.GetDuration("HttpClient.timeout", 30, time.Second)}
+	netClient := &http.Client{Transport: tr, Timeout: config.GetDuration("HttpClient.timeout", 60, time.Second)}
```
atzoum, line 298: we are using this config property in other places too. Should we provide a special config option for reporting?

saurav-malani: adding to @atzoum  comment. We might want to modify `GetDuration` func to take multiple keys, just like we do for  `RegisterDurationConfigVariable` func.
here: https://github.com/rudderlabs/rudder-server/blob/06440975a2be136d77ab09928c03e5638481f379/config/config.go#L483

or maybe create a new func that takes multiple keys.

config/config.go
```diff
@@ -310,6 +319,15 @@ func GetBool(key string, defaultValue bool) (value bool) {
 	return viper.GetBool(key)
 }
 
+func GetFirstInt(defaultValue int, keys ...string) (value int) {
```
chandumlg, line 322: Do we really need these functions?

Can't we keep it simple like this? https://github.com/rudderlabs/rudder-server/blob/581aef4bc02c06d99a7f1f637d150391bb294988/router/config.go#L7

chandumlg (changes requested):
Changes requested

config/config.go
atzoum: Simple, it depends one your perspective. Repeating this piece of code everywhere required won't keep it simple or safe. IMO it is simpler to provide functions that encapsulate such common logic, tested once, used everywhere.

lvrach: I do find the introduction of `GetFirst...` methods a bit too much, and it could make readability and understanding harder. Especially in emergencies when you try to understand which configuration is applied. Having many ways to define a configuration can confuse.

After thinking about passing, `HttpClient.timeout` is not helpful. Every client has its own needs and needs to be configured separately. It also makes sense to have a default value hardcoded as `30sec` for most, but changing the underlying base value for all HTTP clients has no use case (other than keeping backwards compatibility).

So, my suggestion is to drop the `GetFirst...` and rename `HttpClient.timeout` to `HttpClient.<client-name>.timeout`. A good idea to do it now, before introducing v1.0.0, as this is going to break existing behaviour for configurations.

atzoum: e.g. isn't this simpler?

```
func getRouterConfigBool(key, destType string, defaultValue bool) bool {
	return config.GetFirstBool(defaultValue, "Router."+destType+"."+key, "Router."+key)
}

func getRouterConfigInt(key, destType string, defaultValue int) int {
	return config.GetFirstInt(defaultValue, "Router."+destType+"."+key, "Router."+key)
}

```

atzoum: @chandumlg are you ok with what @lvrach is proposing?

chandumlg: yeah sounds good.

cisse21 (approved):

lvrach (approved):

chandumlg (approved):

PR #2305, opened 2022-08-16
chore: support for startAfter key for S3

BonapartePC:
# Description

StartAfterKey support was removed in the Previous PR. This PR brings back the support for the same. startAfter env need not be added and is being removed here.

Previous PR was mostly inspired by our rudder-server-enterprise PR (https://github.com/rudderlabs/rudder-server-enterprise/pull/90). This bug was present in enterprise PR by mistake based on the understanding that by using startAfter you could get keys modified after a given time but in reality, startAfter is used to get keys which come after the given key when all the keys are ordered lexicographically (https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_RequestSyntax). Due to this, a bug was introduced which got rectified in this PR.

## Notion Ticket

https://www.notion.so/rudderstacks/Integrate-Filemanager-into-replay-code-2aef13bc2e8c4d248be41dcfc1595bd3

## Security


atzoum:
fixing a defect within the release? The type in the title of the pr could be renamed to `chore` instead of `fix`.

fracasula:
@BonapartePC can you please add more context to the PR description?
* why was `StartAfterKey` removed in the previous PR
* why are we adding back `StartAfterKey`
* what is `startAfter` used for?
* anything else that might be relevant

BonapartePC:
> fixing a defect within the release? The type in the title of the pr could be renamed to `chore` instead of `fix`.

Done

BonapartePC:
> @BonapartePC can you please add more context to the PR description?
> 
> * why was `StartAfterKey` removed in the previous PR
> * why are we adding back `StartAfterKey`
> * what is `startAfter` used for?
> * anything else that might be relevant

Updated the PR description

atzoum:
Is it tested? Any way to include tests to guard us from doing the same mistake in the next refactoring?

lvrach:
Is this a `fix:` or a `chore:`?

atzoum:
> Is this a `fix:` or a `chore:`?

Interesting! If it is fixing an unreleased defect I would say it is a `chore`, otherwise a `fix`

BonapartePC:
> Is it tested? Any way to include tests to guard us from doing the same mistake in the next refactoring?

Yes, I did sanity testing to see if the replay files are coming correctly with a start time. In this PR, we just introduced the startAfter variable to the function and most of them are not being used in the function. This variable will be used in the next PR where the tests will also be introduced.

services/filemanager/filemanager.go
```diff
@@ -44,7 +44,7 @@ type FileManager interface {
 	GetObjectNameFromLocation(string) (string, error)
 	GetDownloadKeyFromFileLocation(location string) string
 	DeleteObjects(ctx context.Context, keys []string) error
-	ListFilesWithPrefix(ctx context.Context, prefix string, maxItems int64) (fileObjects []*FileObject, err error)
+	ListFilesWithPrefix(ctx context.Context, startAfter, prefix string, maxItems int64) (fileObjects []*FileObject, err error)
```
lvrach, line 47: Given that `startAfter` is not always required, it is better to place it as the last argument.

```suggestion
	ListFilesWithPrefix(ctx context.Context, prefix string, maxItems int64startAfter, startAfter string) (fileObjects []*FileObject, err error)
```

cisse21 (approved):

services/filemanager/filemanager.go
BonapartePC: It is not required now, but the next PR (Replay supporting all object storage) will start using it

lvrach (approved):
fixes will be made in a follow up PR

PR #2309, opened 2022-08-17
fix: group errors swallowed + honouring ctx

fracasula:
# Description

When using an `errgroup.Group`, if more than one routine returns an error it is swallowed: https://go.dev/play/p/5f6nHLHEEwu
I'm therefore making sure that we log all errors in each routine just in case more than one would return an error.
This is also useful in case one of them gets stuck and doesn't return while another prints an error immediately. In fact if the context gets canceled and one of the routines doesn't return, we would just get some log entries with a `context canceled` error printed by other components, but the `g.Wait()` would never get a chance to run so we would not know why the context was canceled in the first place.

Additionally, I removed a `time.Sleep(time.Second)` in the Warehouse `Setup()` method and I replaced it with a retry policy with an exponential backoff. While doing so I also propagated the context further down in some functions.

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/main-go-errgroup-error-f8dfc3f52ce541ebaf140b64c35be43e) >

## Security

### DIFF
--- a/main.go
+++ b/main.go
@@ -252,7 +252,7 @@ func Run(ctx context.Context) int {
 	g.Go(func() (err error) {
 		if err = admin.StartServer(ctx); err != nil {
 			if !errors.Is(err, context.Canceled) {
-				pkgLogger.Errorf("Error starting admin server: %v", err)
+				pkgLogger.Errorf("Error in Admin server routine: %v", err)
 			}
 		}
 		return err
@@ -261,7 +261,7 @@ func Run(ctx context.Context) int {
 	g.Go(func() (err error) {
 		p := &profiler.Profiler{}
 		if err = p.StartServer(ctx); err != nil {
-			pkgLogger.Errorf("Error starting profiler server: %v", err)
+			pkgLogger.Errorf("Error in Profiler server routine: %v", err)
 		}
 		return err
 	})
@@ -271,7 +271,7 @@ func Run(ctx context.Context) int {
 		appHandler.HandleRecovery(options)
 		g.Go(misc.WithBugsnag(func() (err error) {
 			if err = appHandler.StartRudderCore(ctx, options); err != nil {
-				pkgLogger.Errorf("Error starting Rudder Core: %v", err)
+				pkgLogger.Errorf("Error in Rudder Core routine: %v", err)
 			}
 			return err
 		}))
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -248,42 +249,52 @@ func Run(ctx context.Context) int {
 	}
 
 	g, ctx := errgroup.WithContext(ctx)
-	g.Go(func() error {
-		return admin.StartServer(ctx)
+	g.Go(func() (err error) {
+		if err = admin.StartServer(ctx); err != nil {
+			if !errors.Is(err, context.Canceled) {
+				pkgLogger.Errorf("Error starting admin server: %v", err)
```
atzoum, line 255: minor: should we prefer starting with a lowercase character for logs as we do with errors?

main.go
```diff
@@ -248,42 +249,52 @@ func Run(ctx context.Context) int {
...
+		return err
 	})
 
 	misc.AppStartTime = time.Now().Unix()
 	if canStartServer() {
 		appHandler.HandleRecovery(options)
-		g.Go(misc.WithBugsnag(func() error {
-			return appHandler.StartRudderCore(ctx, options)
+		g.Go(misc.WithBugsnag(func() (err error) {
+			if err = appHandler.StartRudderCore(ctx, options); err != nil {
```
atzoum, line 273: An error can happen even after rudder core has started, e.g. if during runtime some goroutine panics.
Based on the above, would it be better to remove "starting" from all similar error log messages and make them more generic? 

e.g.
pkgLogger.Errorf("error in rudder core routine: %v", err)
```

```

main.go
fracasula: We don't have a standard at the moment in this regard but I would say that it is not necessary. For errors it makes more sense IMO because we concatenate them so you might not want `error 1: Error 2` but want `error 1: error 2`. Either way no big deal, as long as we decide on one we can stick to it.

main.go
fracasula: If a goroutine panics it is not returned by `StartRudderCore` though, is it? The `StartRudderCore` method should only return errors related to starting the component. Is there a specific case you came across to?

atzoum: Not verified anything, but by reading the code, e.g. wouldn't [this runtime error](https://github.com/rudderlabs/rudder-server/blob/16195455dc7a38a00c9a0e0c997f72f3cda3dce7/app/cluster/dynamic.go#L138) result in the rudder core group returning an error?

fracasula: Right, your suggestion makes sense because it has a blocking API and it doesn't just start things and then it keeps them running asynchronously as the name would imply. At least for me when I read `Start` I don't think of a call that would block till the execution finishes. Personally I'd prefer `RunRudderCore` because A) it's a blocking call and B) there is no `Stop()` which is done via `context`. In any case given the current behaviour it makes sense to change the log entry :+1:

atzoum (approved):

sivashanmukh (approved):
LGTM

PR #2315, opened 2022-08-18
feat: add support for sending custom payload to webhook source 

koladilip:
# Description
Monday.com does verification of webhook using challenge so we need to respond to source when challenge event sent and currently that is possible so I have added support for sending custom payload. This change is backward compatible so we can deploy independently. For more details please refer to notion ticket.

## Notion Ticket
https://www.notion.so/rudderstacks/Monday-Source-Integration-9f06aa54dcd44756a68893a244c3e356

## Security


fracasula (commented):
Can we have a test for these changes?

gateway/webhook/webhookTransformer.go
```diff
@@ -83,7 +86,7 @@ func (bt *batchWebhookTransformerT) transform(events [][]byte, sourceType string
 			}
 		]
 	*/
-	var responses []interface{}
+	var responses []transformerResponseT
```
atzoum, line 89: it would be useful to include this new scenario as a message sample in the comment above

gateway/webhook/webhookTransformer.go
```diff
@@ -13,10 +13,13 @@ import (
...
+// transformerResponseT will be populated using JSON unmarshall
+// so we need to make fields public
 type transformerResponseT struct {
-	output     []byte
-	err        string
-	statusCode int
+	Output         map[string]interface{} `json:"output"`
+	Err            string                 `json:"error"`
+	StatusCode     int                    `json:"statusCode"`
+	OutputToSource map[string]interface{} `json:"outputToSource"`
```
atzoum, line 22: Are we confident that we will never have to return anything else other than a json object in our response? If not, this should be converted to a `[]byte`

gateway/webhook/webhook.go
```diff
@@ -179,25 +174,36 @@ func (webhook *HandleT) RequestHandler(w http.ResponseWriter, r *http.Request) {
 		r.Header.Set("Content-Type", "application/json")
 	}
 
-	done := make(chan webhookErrorRespT)
+	done := make(chan transformerResponseT)
```
atzoum, line 177: This would be a good chance to rename `transformerResponseT` to `transformerResponse` now that we are refactoring it. We have agreed that the `T` suffix is not something we should follow when naming structs

gateway/webhook/webhook.go
```diff
@@ -289,53 +295,53 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
...
-
 		bt.stats.sourceStats[breq.sourceType].numOutputEvents.Count(len(batchResponse.responses))
 
 		for idx, resp := range batchResponse.responses {
 			webRequest := webRequests[idx]
-			output := resp.output
-			if resp.err != "" {
-				webRequests[idx].done <- webhookErrorRespT{err: resp.err, statusCode: resp.statusCode}
-				continue
+			if resp.Err == "" && resp.Output != nil {
```
atzoum, line 317: Can we introduce a set of integration tests for the webhook RequestHandler, using an httptest Server for mocking the transformer?
It should be possible to test and verify all three major scenarios (error, success, response to client)

The only way to catch up, is by start writing these missing test scenarios when we are touching some uncovered functionality.

koladilip:
> Can we have a test for these changes?

Done

gateway/webhook/webhook.go
koladilip: Done

gateway/webhook/webhookTransformer.go
koladilip: Done

gateway/webhook/webhookTransformer.go
koladilip: Done

gateway/webhook/webhook_test.go
```diff
@@ -0,0 +1,228 @@
...
+		whReq := <-webhookHandler.requestQ[sourceDefName]
+		whReq.done <- transformerResponse{
+			Err:        sampleError,
+			StatusCode: http.StatusBadRequest,
+		}
+	}()
+	mockGW.EXPECT().TrackRequestMetrics(sampleError).Times(1)
+	webhookHandler.RequestHandler(w, req)
+
+	assert.Equal(t, w.Result().StatusCode, http.StatusBadRequest)
```
atzoum, line 90: can't we use the existing testify library for that, e.g. `require.Equal`?

go.mod
```diff
@@ -175,4 +175,5 @@ require (
 	gopkg.in/linkedin/goavro.v1 v1.0.5 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20220512140231-539c8e751b99 // indirect
+	gotest.tools v2.2.0+incompatible
```
atzoum, line 178:
```suggestion
```

gateway/webhook/webhook_test.go
```diff
@@ -29,8 +29,9 @@ const (
 )
 
 var (
-	whStats *webhookStatsT
-	once    sync.Once
+	whStats           *webhookStatsT
+	once              sync.Once
+	transformerServer = httptest.NewServer(http.HandlerFunc(transformMockHandler))
```
atzoum, line 34: isn't it simpler each test to have its own server? Tests run in paraller during ci

gateway/webhook/webhookTransformer.go
```diff
@@ -74,16 +82,17 @@ func (bt *batchWebhookTransformerT) transform(events [][]byte, sourceType string
 							userId: "U123"
 						}
 					]
+				},
+				outputToSource: {
```
atzoum, line 86: instead of having a single example response with all fields filled, which according to my understanding is not a valid scenario, wouldn't it make sense to have 3 example responses, 1 for each main scenario?

gateway/webhook/webhook_test.go
koladilip: problem here is, serverUrl is passed using common variable: sourceTransformerURL so we can't have server per test so I have written in generic way test.

gateway/webhook/webhook_test.go
```diff
@@ -0,0 +1,228 @@
...
+func transformMockHandler(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+	reqBody, _ := io.ReadAll(r.Body)
+	_, err := w.Write(reqBody)
+	if err != nil {
+		pkgLogger.Error(err)
+	}
+}
+
+func TestWebhookRequestHandlerErrorCase(t *testing.T) {
```
atzoum, line 67: First 3 test cases don't actually use the httptest server?
Would it be possible to verify behaviour for all test cases in an end-2-end manner? e.g. given that transformer responds A, gateway responds B.

gateway/webhook/webhook_test.go
atzoum: we may refactor `sourceTransformerURL` and convert it to a `batchWebhookTransformerT` field, correct?

gateway/webhook/webhook.go
```diff
@@ -289,53 +290,53 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
...
+				outputPayload, err := json.Marshal(resp.Output)
+				if err != nil {
+					webRequests[idx].done <- bt.markRepsonseFail(response.SourceTransformerInvalidOutputFormatInResponse)
```
saikumarrs, line 315:
```suggestion
					webRequest.done <- bt.markRepsonseFail(response.SourceTransformerInvalidOutputFormatInResponse)
```

gateway/webhook/webhook.go
```diff
@@ -289,53 +290,53 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
...
-	if err == nil {
-		errorMessage = webhook.gwHandle.ProcessWebRequest(req.writer, req.request, "batch", payload, req.writeKey)
-	} else {
-		errorMessage = err.Error()
+	if err != nil {
+		return err.Error()
 	}
-
-	// Wait for batcher process to be done
-	req.done <- webhookErrorRespT{err: errorMessage}
```
saikumarrs, line 338: Why isn't this required anymore?

gateway/webhook/webhookTransformer.go
```diff
@@ -13,22 +13,30 @@ import (
...
-	err        string
-	statusCode int
+type outputToSource struct {
+	Body        []byte `json:"body"`
+	ContentType string `json:"contentType"`
+}
+
+// transformerResponse will be populated using JSON unmarshall
+// so we need to make fields public
+type transformerResponse struct {
```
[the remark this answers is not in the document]
atzoum: I believe this has been discussed [in the past](https://www.notion.so/rudderstacks/Go-idiomatic-naming-79e44d023b5e48bd8dd9b4a89e07b3cb) and was agreed that this convention is not idiomatic, thus shouldn't be followed. Therefore, whenever we are refactoring old code, we also try to take care of stripping the `T` suffix as well.

gateway/webhook/webhook_test.go
fracasula: +1 for removing the global variable or simply injecting it as a struct attribute at construction time.

gateway/webhook/setup.go
```diff
@@ -1,3 +1,5 @@
+//go:generate mockgen --build_flags=--mod=mod -destination=../../../mocks/gateway/webhook/mock_webhook.go -package mock_webhook github.com/rudderlabs/rudder-server/gateway/webhook GatewayI
```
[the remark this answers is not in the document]
koladilip: This is for generating mocks, intentional.

gateway/webhook/webhook.go
koladilip: handled in caller

gateway/webhook/webhook.go
koladilip: Done

gateway/webhook/webhook_test.go
koladilip: Done, please check if this is reasonable, for tests I had to add variadic params so that we can dynamically set sourceTransformerURL.

saikumarrs (approved):

gateway/webhook/webhook.go
```diff
@@ -179,25 +177,31 @@ func (webhook *HandleT) RequestHandler(w http.ResponseWriter, r *http.Request) {
...
 		code := http.StatusBadRequest
-		if resp.statusCode != 0 {
-			code = resp.statusCode
+		if resp.StatusCode != 0 {
+			code = resp.StatusCode
 		}
-		pkgLogger.Infof("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, code, resp.err)
-		http.Error(w, resp.err, code)
+		pkgLogger.Infof("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, code, resp.Err)
+		http.Error(w, resp.Err, code)
```
lvrach, line 196: [minor] an early return here can make the code more readable.

```suggestion
		http.Error(w, resp.Err, code)
		return
```
(no need for else case)

koladilip: Done

lvrach (approved):

gateway/webhook/webhook.go
```diff
@@ -179,26 +177,33 @@ func (webhook *HandleT) RequestHandler(w http.ResponseWriter, r *http.Request) {
...
-	} else {
-		pkgLogger.Debugf("IP: %s -- %s -- Response: 200, %s", misc.GetIPFromReq(r), r.URL.Path, response.GetStatus(response.Ok))
-		_, _ = w.Write([]byte(response.GetStatus(response.Ok)))
+		pkgLogger.Infof("IP: %s -- %s -- Response: %d, %s", misc.GetIPFromReq(r), r.URL.Path, code, resp.Err)
+		http.Error(w, resp.Err, code)
+		return
 	}
+
+	payload := []byte(response.Ok)
+	if resp.OutputToSource != nil {
```
atzoum, line 201: My understanding is that this is test-specific code that shouldn't be needed. Can't we simply mock the http server and send the responses we want through this channel only?
Adding a separate test-only branch here has as an effect to cover the test-only branch with tests, leaving the actual code that runs in the actual system uncovered.

atzoum (approved):
thank you @koladilip for addressing all comments, great work!

PR #2316, opened 2022-08-18
chore: fix RetryWith logic

atzoum:
# Description

Retry based on context error, not on returned error

## Security

### DIFF
--- a/utils/misc/retry.go
+++ b/utils/misc/retry.go
@@ -24,7 +24,7 @@ func RetryWith(parentContext context.Context, timeout time.Duration, maxAttempts
 			return nil
 		}
 		// only retry if the child context's deadline was exceeded
-		if !errors.Is(err, context.DeadlineExceeded) || parentContext.Err() != nil {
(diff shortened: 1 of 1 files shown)


utils/misc/retry.go
```diff
@@ -24,7 +24,7 @@ func RetryWith(parentContext context.Context, timeout time.Duration, maxAttempts
 			return nil
 		}
 		// only retry if the child context's deadline was exceeded
-		if !errors.Is(err, context.DeadlineExceeded) || parentContext.Err() != nil {
+		if !errors.Is(ctx.Err(), context.DeadlineExceeded) || parentContext.Err() != nil {
 			return err
```
lvrach, line 28: in this case, shouldn't we: 
```suggestion
			return ctx.Err()
```

atzoum: no, the error returned by the function should be returned to the caller. What `ctx.Err()` controls here is only the retry logic

saurav-malani (approved):

cisse21 (approved):

PR #2321, opened 2022-08-20
fix(router): trying to stop the router before the generatorLoop is started causes router to hang

atzoum:
# Description
Stopping the router too early causes workers getting stopped properly and `Start`'s completion channel to never close.

Additional improvements to various multi-tenant related tests

**TestMultiTenant** changes:
- Spin-up a transformer at the beginning for the processor to use
- Use a unique release key in etcd for each test
- First switch to `NORMAL` mode before requesting to switch to `DEGRADED` mode
- Verify message reaches to the destination after switching to `NORMAL` mode
- Require that no errors happen when we are sending messages to the server
- Fail the test when the spawn server process fails unexpectedly
- Perform gw table verifications in real-time (stop using `require.Eventually`)

**TestDynamicClusterManager** changes:

- always teardown jobsDBs

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=0e93e67e9ff3454d80c934dbe5b5f497&pm=s)

## Security


multi_tenant_test.go
```diff
@@ -214,13 +220,16 @@ func testMultiTenantByAppType(t *testing.T, appType string) {
 		t.Fatal("Timeout waiting for test-ack/1")
 	}
 
+	cleanupGwJobs := func() {
+		_, _ = postgresContainer.DB.ExecContext(ctx, `DELETE FROM gw_job_status_1 WHERE job_id in (SELECT job_id from gw_jobs_1 WHERE workspace_id = $1)`, workspaceID)
+		_, _ = postgresContainer.DB.ExecContext(ctx, `DELETE FROM gw_jobs_1 WHERE workspace_id = $1`, workspaceID)
+	}
```
fracasula, line 226: :+1:

docker_test.go
```diff
@@ -427,14 +427,6 @@ func setupMainFlow(svcCtx context.Context, t *testing.T) <-chan struct{} {
 	})
 	require.NoError(t, containersGroup.Wait())
 
-	health.WaitUntilReady(
-		context.Background(), t,
-		fmt.Sprintf("%s/health", transformerContainer.TransformURL),
-		time.Minute,
-		time.Second,
-		"transformer",
-	)
```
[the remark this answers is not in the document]
atzoum: yes, this is done in the `SetupTransformer` function

router/router.go
```diff
@@ -125,13 +125,13 @@ type HandleT struct {
...
-	backgroundCancel     context.CancelFunc
-	backgroundWait       func() error
-	generatorLoopStarted chan struct{}
-	lastQueryRunTime     time.Time
-	timeGained           float64
+	backgroundGroup  *errgroup.Group
+	backgroundCtx    context.Context
+	backgroundCancel context.CancelFunc
+	backgroundWait   func() error
+	startEnded       chan struct{}
```
fracasula, line 132: It might be useful to have a test with a succession of starts and stops. I suggest running it with the `-race` flag on as well.

atzoum: indirectly we have been doing that in `TestDynamicCluster`, but we were ignoring the intermittent failures and we were accusing the test as flaky

fracasula (approved):

cisse21 (approved):

PR #2323, opened 2022-08-22
feat: honour anonId only from header

cisse21:
# Description

https://www.notion.so/rudderstacks/userID-anonymousID-ec2d067531ac42aba5f84533f32b9135

## Notion Ticket

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=97fe15a422c84dd7b906552a54983550&pm=s

## Security


gateway/gateway.go
```diff
@@ -515,10 +518,10 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 				userIDFromReq := strings.TrimSpace(vjson.Get("userId").String())
 				if builtUserID == "" {
 					if anonIDFromReq != "" {
-						builtUserID = anonIDFromReq + DELIMITER + userIDFromReq
+						builtUserID = userIDHeader + DELIMITER + anonIDFromReq + DELIMITER + userIDFromReq
 					} else {
 						// Proxy gets the userID from body if there is no anonymousId in body/header
-						builtUserID = userIDFromReq + DELIMITER + userIDFromReq
+						builtUserID = userIDHeader + DELIMITER + userIDFromReq + DELIMITER + userIDFromReq
 					}
```
lvrach, line 525: How much effort would it require to write tests for this section? Maybe isolating this to a function would make it possible or easier.

fracasula (commented):
Can we have a test that given an HTTP request with the correct header we check that the job in the DB has `user_id` populated accordingly?

BonapartePC (approved):

gateway/gateway.go
cisse21: We don't have such tests for gateway :/ I can isolate this part and add it though if it makes sense

fracasula: Can't we copy from `multi_tenant_test.go` and have the app start in GATEWAY mode only, send an event with the correct header and then check the `user_id` on the DB? Or you think it's going to be more complex?

gateway/gateway.go
```diff
@@ -515,10 +518,10 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 				userIDFromReq := strings.TrimSpace(vjson.Get("userId").String())
 				if builtUserID == "" {
 					if anonIDFromReq != "" {
-						builtUserID = anonIDFromReq + DELIMITER + userIDFromReq
+						builtUserID = userIDHeader + DELIMITER + anonIDFromReq + DELIMITER + userIDFromReq
```
atzoum, line 521: I am not sure I understand why we need a composite key containing three different keys, coming both from the request header and the message's content

multi_tenant_test.go
```diff
@@ -314,6 +314,13 @@ func testMultiTenantByAppType(t *testing.T, appType string) {
 			).Scan(&count)
 			require.NoError(t, err)
 			require.Equal(t, 1, count)
+
+			var userId string
```
atzoum, line 318: tests should be single purpose too :)
You can imagine what can happen as we are adding more and more scenarios in a single test

cisse21: Yeah I am adding integration tests under the gateway folder. Will clean this up from here

PR #2324, opened 2022-08-22
feat: perform ADD_DS operation in a single transaction

saurav-malani:
# Description

Currently, `ADD_DS` operation to add new jobs table & job status table doesn't happen in a single transaction. Since, we are journaling, so any crash in middle of the addition would be reverted during recovery. But, this might create a problem in case of multiple writer to a single DB, i.e. in case of Gateway HA. So, we here we are performing the complete operation in a single transaction to avoid any problem in case of multiple writers.

## Notion Ticket

https://www.notion.so/rudderstacks/Perform-ADD_DS-operation-in-a-single-transaction-4cc30f0bdd17473ab1f2acba59afe653

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -1044,28 +1044,6 @@ func (jd *HandleT) Close() {
 	jd.dbHandle.Close()
 }
 
-// removeExtraKey : removes extra key present in map1 and not in map2
-// Assumption is keys in map1 and map2 are same, except that map1 has one key more than map2
-func removeExtraKey(map1, map2 map[string]string) string {
-	var deleteKey, key string
-	for key = range map1 {
-		if _, ok := map2[key]; !ok {
-			deleteKey = key
-			break
-		}
-	}
-
-	if deleteKey != "" {
-		delete(map1, deleteKey)
-	}
-
-	return deleteKey
-}
-
-func remove(slice []string, idx int) []string {
-	return append(slice[:idx], slice[idx+1:]...)
-}
-
 /*
 Function to return an ordered list of datasets and datasetRanges
 Most callers use the in-memory list of dataset and datasetRanges
@@ -1109,6 +1087,7 @@ func (jd *HandleT) refreshDSRangeList(l lock.DSListLockToken) []dataSetRangeT {
 
 	// At this point we must have write-locked dsListLock
 	dsList := jd.refreshDSList(l)
+
 	jd.datasetRangeList = nil
 
 	for idx, ds := range dsList {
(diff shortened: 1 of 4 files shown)


jobsdb/jobsdb.go
```diff
@@ -1044,28 +1044,6 @@ func (jd *HandleT) Close() {
 	jd.dbHandle.Close()
 }
 
-// removeExtraKey : removes extra key present in map1 and not in map2
-// Assumption is keys in map1 and map2 are same, except that map1 has one key more than map2
-func removeExtraKey(map1, map2 map[string]string) string {
```
saurav-malani, line 1049: this method is not used anywhere, so removed.

jobsdb/jobsdb.go
```diff
@@ -1044,28 +1044,6 @@ func (jd *HandleT) Close() {
...
-	}
-
-	if deleteKey != "" {
-		delete(map1, deleteKey)
-	}
-
-	return deleteKey
-}
-
-func remove(slice []string, idx int) []string {
```
saurav-malani, line 1065: method not used anywhere.

jobsdb/jobsdb.go
```diff
@@ -1533,11 +1513,27 @@ type transactionHandler interface {
 }
 
 func (jd *HandleT) createDS(newDS dataSetT, l lock.DSListLockToken) {
```
saurav-malani, line 1515: test for the same can be found here: https://github.com/rudderlabs/rudder-server/blob/e0117268aa350d05b7a8fbadce444340ae79cebf/jobsdb/integration_test.go#L875

jobsdb/jobsdb_backup_test.go
```diff
@@ -191,6 +191,9 @@ func (*backupTestCase) insertRTData(t *testing.T, jobs []*JobT, statusList []*Jo
 		}
 		return jobsDB.copyJobStatusDS(context.Background(), tx, rtDS2, statusList, []string{}, nil)
 	})
+	jobsDB.dsListLock.WithLock(func(l lock.DSListLockToken) {
```
atzoum, line 194: why do we need this?

jobsdb/jobsdb_utils.go
```diff
@@ -45,25 +45,6 @@ func getDSList(jd assertInterface, dbHandle *sql.DB, tablePrefix string) []dataS
 
 	sortDnumList(jd, dnumList)
 
-	// If any service has crashed while creating DS, this may happen. Handling such case gracefully.
```
atzoum, line 48: 👍

jobsdb/jobsdb_backup_test.go
saurav-malani: nope, since we are anyways refreshing after the transaction. We don't need it anymore. Removed

atzoum (approved):

Sidddddarth (approved):

PR #2327, opened 2022-08-22
chore: migrate small tables

fracasula:
# Description

Small tables may result, after a table has been migrated at least once, leaving it with a low number of non-terminal state jobs.
So far, we are not defragmenting such tables, since the migration loop is not considering they should be migrated.

Tables with a low number of rows, e.g. < 5000 or less than 5% of the max job capacity of a table should be eligible for migration.

## Additional changes
* `Makefile` test recipe proceeds with the deletion of files even when tests fail
* Ability to use `MinDSRetentionPeriod` to avoid having the same table migrated in an infinite loop without delays (see `we avoid multiple migrations in a row` test)
* using the same package for `integration_test.go` to simplify mocking
* optimizing valid and non valid terminal states slice (no need to build it each time, I'm now doing it once in `init()`)
* removed `defer` from `for` loop as per DeepSource (see `writeBackupToGz` lambda)
* moved some vars inside `HandleT` to simplify mocking
* some linting/deepsource fixes

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/Migrate-small-tables-ad6da7d804d949ca9630351d9da5e287) >

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -240,7 +240,7 @@ type JobsDB interface {
 	// GetToRetry finds jobs in failed state
 	GetToRetry(ctx context.Context, params GetQueryParamsT) (JobsResult, error)
 
-	// GetToRetry finds jobs in waiting state
+	// GetWaiting finds jobs in waiting state
 	GetWaiting(ctx context.Context, params GetQueryParamsT) (JobsResult, error)
 
 	// GetExecuting finds jobs in executing state
@@ -285,7 +285,7 @@ var (
 
 /*
 UpdateJobStatusInTx updates the status of a batch of jobs in the passed transaction
-customValFilters[] is passed so we can efficinetly mark empty cache
+customValFilters[] is passed so we can efficiently mark empty cache
 Later we can move this to query
 IMP NOTE: AcquireUpdateJobStatusLocks Should be called before calling this function
 */
@@ -581,6 +581,9 @@ var (
 	Aborted     = jobStateT{isValid: true, isTerminal: true, State: "aborted"}
 	Migrated    = jobStateT{isValid: true, isTerminal: true, State: "migrated"}
 	WontMigrate = jobStateT{isValid: true, isTerminal: true, State: "wont_migrate"}
+
+	validTerminalStates    []string
+	validNonTerminalStates []string
 )
 
 // Adding a new state to this list, will require an enum change in postgres db.
@@ -610,22 +613,17 @@ const (
 	ReadWrite OwnerType = ""
 )
 
-func getValidTerminalStates() (validTerminalStates []string) {
+func init() {
 	for _, js := range jobStates {
-		if js.isValid && js.isTerminal {
-			validTerminalStates = append(validTerminalStates, js.State)
+		if !js.isValid {
+			continue
 		}
-	}
-	return
-}
-
-func getValidNonTerminalStates() (validNonTerminalStates []string) {
-	for _, js := range jobStates {
-		if js.isValid && !js.isTerminal {
+		if js.isTerminal {
+			validTerminalStates = append(validTerminalStates, js.State)
+		} else {
 			validNonTerminalStates = append(validNonTerminalStates, js.State)
 		}
 	}
-	return
 }
 
 var (
@@ -636,18 +634,20 @@ var (
 var (
 	maxDSSize, maxMigrateOnce, maxMigrateDSProbe int
 	maxTableSize                                 int64
-	jobDoneMigrateThres, jobStatusMigrateThres   float64
-	migrateDSLoopSleepDuration                   time.Duration
-	addNewDSLoopSleepDuration                    time.Duration
-	refreshDSListLoopSleepDuration               time.Duration
-	backupCheckSleepDuration                     time.Duration
-	cacheExpiration                              time.Duration
-	useJoinForUnprocessed                        bool
-	backupRowsBatchSize                          int64
-	backupMaxTotalPayloadSize                    int64
-	pkgLogger                                    logger.LoggerI
-	useNewCacheBurst                             bool
-	skipZeroAssertionForMultitenant              bool
+	jobDoneMigrateThres,
+	jobStatusMigrateThres,
+	jobMinRowsMigrateThres float64
+	migrateDSLoopSleepDuration      time.Duration
+	addNewDSLoopSleepDuration       time.Duration
+	refreshDSListLoopSleepDuration  time.Duration
+	backupCheckSleepDuration        time.Duration
+	cacheExpiration                 time.Duration
+	useJoinForUnprocessed           bool
+	backupRowsBatchSize             int64
+	backupMaxTotalPayloadSize       int64
+	pkgLogger                       logger.LoggerI
+	useNewCacheBurst                bool
+	skipZeroAssertionForMultitenant bool
 )
 
 // Loads db config and migration related config from config file
@@ -665,6 +665,9 @@ func loadConfig() {
 	/*Migration related parameters
 	jobDoneMigrateThres: A DS is migrated when this fraction of the jobs have been processed
 	jobStatusMigrateThres: A DS is migrated if the job_status exceeds this (* no_of_jobs)
+	jobMinRowsMigrateThres: A DS with a low number of rows should be eligible for migration if the number of rows are
+							less than jobMinRowsMigrateThres percent of maxDSSize (e.g. if jobMinRowsMigrateThres is 5
+							then DSs that have less than 5% of maxDSSize are eligible for migration)
 	maxDSSize: Maximum size of a DS. The process which adds new DS runs in the background
 			(every few seconds) so a DS may go beyond this size
 	maxMigrateOnce: Maximum number of DSs that are migrated together into one destination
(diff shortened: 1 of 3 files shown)


jobsdb/jobsdb.go
```diff
@@ -395,15 +395,20 @@ HandleT is the main type implementing the database for implementing
...
-	MaxDSRetentionPeriod          time.Duration
+	dbHandle         *sql.DB
+	ownerType        OwnerType
+	tablePrefix      string
+	datasetList      []dataSetT
+	datasetRangeList []dataSetRangeT
+	dsListLock       lock.DSListLocker
+	dsMigrationLock  sync.RWMutex
+	// dsRetentionMap is protected by dsMigrationLock
+	dsRetentionMap       map[string]time.Time
```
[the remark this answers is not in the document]
fracasula: The reason I created it is to keep in-memory the real timestamp of when the table was created instead of relying on fields inside like `created_at` that could be susceptible to external changes and requires an additional query. I thought I'd propose it to simplify a bit, wdyt?

jobsdb/jobsdb.go
```diff
@@ -793,6 +802,16 @@ func (jd *HandleT) Setup(
...
+	jd.dsRetentionMap = make(map[string]time.Time)
+	if jd.jobDoneMigrateThres == 0 {
+		jd.jobDoneMigrateThres = jobDoneMigrateThres
+	}
+	if jd.jobStatusMigrateThres == 0 {
+		jd.jobStatusMigrateThres = jobStatusMigrateThres
+	}
+	if jd.jobMinRowsMigrateThres == 0 {
+		jd.jobMinRowsMigrateThres = jobMinRowsMigrateThres
+	}
```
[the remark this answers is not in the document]
fracasula: Right, and we're not invoking `Setup` when hot-reloading. I was trying to get rid of some global variables but I'll leave this for another day. Doing it with the hot-reload is too much effort for this task so I'll just put them back :+1:

jobsdb/jobsdb.go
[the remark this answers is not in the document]
atzoum: jobsDB current usage patterns don't promote the use of this map:

1. jobsDB operates in 3 modes: reader, writer, read-writer
2. We might have multiple jobsDB instances for the same set of tables, e.g. 1 reader & 1 writer
3. Writers only create (new) tables
4. Reader both create (migrated) tables and delete tables

Based on the above we risk ending up with:

- jobsDBs with split brain, i.e. different entries in this map
- writer jobsDBs with a map of ever-increasing entries

Finally, the min & max retention isn't really concerned about when the table was created, it has to do with the actual data that the table holds, e.g. a migrated table may be created now, but it will hold past data.

jobsdb/jobsdb.go
```diff
@@ -676,6 +681,7 @@ func loadConfig() {
 	*/
 	config.RegisterFloat64ConfigVariable(0.8, &jobDoneMigrateThres, true, "JobsDB.jobDoneMigrateThres")
 	config.RegisterFloat64ConfigVariable(5, &jobStatusMigrateThres, true, "JobsDB.jobStatusMigrateThres")
+	config.RegisterFloat64ConfigVariable(0.5, &jobMinRowsMigrateThres, true, "JobsDB.jobMinRowsMigrateThres")
```
atzoum, line 684: Please excuse me for misleading you with my initial comment with respect to this value... assuming that we consider as small tables, tables that hold less than `5%` of the max DS size, then this value should be `0.05`. Hope I didn't mess it up again :)

jobsdb/jobsdb.go
```diff
@@ -1176,18 +1190,35 @@ func (jd *HandleT) checkIfMigrateDS(ds dataSetT) (bool, int) {
...
-		return true, totalCount - delCount
+
+	if jd.MaxDSRetentionPeriod > 0 && time.Since(minCreatedAt) > jd.MaxDSRetentionPeriod {
+		return true, false, recordsLeft
 	}
 
 	if (float64(delCount)/float64(totalCount) > jobDoneMigrateThres) ||
 		(float64(statusCount)/float64(totalCount) > jobStatusMigrateThres) {
-		return true, totalCount - delCount
+		return true, false, recordsLeft
```
atzoum, line 1213: We should be honest here about whether this is a small table or not. 
As mentioned earlier, we wouldn't want to migrate this table by itself if it is small, even though the `jobDoneMigrateThres` or `jobStatusMigrateThres` have been surpassed.

Please let me provide some examples that we would like to avoid:

Assuming that 
- jobDoneMigrateThres: 0.8
- jobStatusMigrateThres: 5

1. We don't want to migrate a small table alone with 9/10 jobs completed (0.9 > 0.8)
2. We don't want to migrate a small table alone with 10 failed jobs for which we have written 60 job statuses (6>5)

jobsdb/jobsdb.go
```diff
@@ -4590,3 +4642,8 @@ func QueryWorkspacePileupWithRetries(parentContext context.Context, timeout time
 	}
 	return res.(map[string]map[string]int), err
 }
+
+type smallDS struct {
+	ds          dataSetT
+	recordsLeft int
+}
```
atzoum, line 4649: **minor:** I would prefer declaring the `smallDS` type inside `migrateDSLoop`, as it really is private to that method

atzoum:
thanks @fracasula it looks good! Can you please refresh from latest master and resolve conflicts?

atzoum (approved):

Sidddddarth:
on a tangent: 
instead of rudder-server getting jobs and then writing them to another table, can't we just run a query which can bypass this and just move jobs in the database from `migrateFrom` tables to `migrateTo` tables?

atzoum:
> on a tangent: instead of rudder-server getting jobs and then writing them to another table, can't we just run a query which can bypass this and just move jobs in the database from `migrateFrom` tables to `migrateTo` tables?

true, this is a separate task we can pick up later

jobsdb/jobsdb.go
```diff
@@ -2526,7 +2546,7 @@ func (jd *HandleT) isEmptyResult(ds dataSetT, workspace string, stateFilters, cu
 			return false
 		}
 
-		pVals := []string{}
+		var pVals []string
```
raman2805, line 2549: [question] how is one better than the other (just so that I keep in mind for future purposes)

raman2805 (approved):

raman2805 (approved):

raman2805 (approved):

jobsdb/jobsdb.go
```diff
@@ -1039,7 +1043,7 @@ func (jd *HandleT) TearDown() {
 //
 //	Stop should be called before Close.
 func (jd *HandleT) Close() {
-	jd.dbHandle.Close()
+	_ = jd.dbHandle.Close()
```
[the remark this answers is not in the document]
lvrach: Linters complain when errors are not handled. This is a way to say: "Yes, I am aware, and I chose to ignore the error"

jobsdb/jobsdb.go
lvrach: `var pVals []string` doesn't initialise the slice. Thus doesn't allocate memory. The major reason to use `var pVals []string` is that is more idiomatic. 

To be honest I am more used to doing `pVals := []string{}`, but I try to follow the idiomatic way.

lvrach (approved):

chandumlg (changes requested):
If there are multiple tables eligible for migration, isSmall is irrelevant. 

If there is a single table eligible for migration, and that is a small, we are not migrating at all. So, this PR instead of migrating small tables, prevents migration of small tables. 

Is the understanding correct? Please correct if not.

chandumlg (commented):
One major advantage happening I see now is small tables are eligible for migration.

So, to summarise
a. if there is only a single small table eligible for migration, we prevent it from migrating.
b. if there are multiple tables for migration (small or big), they will be migrated.

Right?

chandumlg (commented):
One major advantage happening I see now is small tables are eligible for migration.

So, to summarise
a. if there is only a single small table eligible for migration, we prevent it from migrating.
b. if there are multiple tables for migration (small or big), they will be migrated.

Right?

atzoum:
> If there are multiple tables eligible for migration, isSmall is irrelevant.
> 
> If there is a single table eligible for migration, and that is a small, we are not migrating at all. So, this PR instead of migrating small tables, prevents migration of small tables.
> 
> Is the understanding correct? Please correct if not.

In case the table is eligible for migration due to job completion percentage or a large job status number, but it is small, then it falls in the small category as any small table and we will only migrate it if another table is eligible for migration too.
The reason why this was done is e.g. to avoid migrating a table with 5 jobs if 4 of them are completed, or there are 30 job statuses recorded in the status table. These events can happen very fast and frequent, resulting in constantly migrating the same small table again and again.
So, if it is a small table we need at least 2 tables to proceed with migration.
Please also note that for judging if a table is small or not, we also consult its job status size.

jobsdb/jobsdb.go
fracasula: Yes @lvrach is correct. With `var pVals []string` there is no allocation but both `len` and `append` would work as expected so you don't have to worry about that.

PR #2329, opened 2022-08-22
fix(router): event ordering algorithm with proper aborted job limiting support

atzoum:
# Description
Fixed a bug with respect to event ordering when there are aborted jobs. The issue was that the algorithm was modifying its relevant `abortedUserIDMap` state in the wrong place at the wrong time, instead of waiting for the next iteration of `generatorLoop` to start, which is the pipeline's entry point and the proper synchronisation barrier.

Additionally
- Introduced an integration test to verify that ordering guarantees are respected in practice.

Side jobs
- Did some minor modification in logs in order to mute some false positive errors such as **_"Server closed"_**
- Modified tenantstats algorithm to make it respect the `jobQueryBatchSize` limit in case of a single workspace

**_Note_**
We are redesigning the event order algorithm in #2344

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=6f2b5742f6964d20adf673407888f9fb&pm=s)

## Security

### DIFF
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -44,6 +44,7 @@ import (
 	"runtime/pprof"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/spf13/viper"
 
@@ -260,8 +261,10 @@ func StartServer(ctx context.Context) error {
 		pkgLogger.Fatal("listen error:", e) // @TODO return?
 	}
 	defer func() {
-		if err := l.Close(); err != nil {
-			pkgLogger.Warn(err)
+		if l != nil {
+			if err := l.Close(); err != nil {
+				pkgLogger.Warn(err)
+			}
 		}
 	}()
 
@@ -269,7 +272,7 @@ func StartServer(ctx context.Context) error {
 	srvMux := http.NewServeMux()
 	srvMux.Handle(rpc.DefaultRPCPath, instance.rpcServer)
 
-	srv := &http.Server{Handler: srvMux}
+	srv := &http.Server{Handler: srvMux, ReadHeaderTimeout: 3 * time.Second}
 	go func() {
 		<-ctx.Done()
 		_ = srv.Shutdown(context.Background()) // @TODO no wait nor timeout on shutdown
--- a/config/backend-config/single_workspace.go
+++ b/config/backend-config/single_workspace.go
@@ -17,6 +17,7 @@ import (
 )
 
 type singleWorkspaceConfig struct {
+	logOnce          sync.Once
 	Token            string
 	workspaceID      string
 	configBackendURL *url.URL
@@ -129,7 +130,9 @@ func (wc *singleWorkspaceConfig) getFromAPI(ctx context.Context, _ string) (Conf
 
 // getFromFile reads the workspace config from JSON file
 func (wc *singleWorkspaceConfig) getFromFile() (ConfigT, error) {
-	pkgLogger.Info("Reading workspace config from JSON file")
+	wc.logOnce.Do(func() {
+		pkgLogger.Info("Reading workspace config from JSON file")
+	})
 	data, err := IoUtil.ReadFile(wc.configJSONPath)
 	if err != nil {
 		pkgLogger.Errorf("Unable to read backend config from file: %s with error : %s", wc.configJSONPath, err.Error())
--- a/eventorder_test.go
+++ b/eventorder_test.go
@@ -1,0 +1,401 @@
+//go:build integration
+
+package main_test
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"math/rand"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/ory/dockertest/v3"
+	"github.com/phayes/freeport"
+	main "github.com/rudderlabs/rudder-server"
+	"github.com/rudderlabs/rudder-server/testhelper/destination"
+	trand "github.com/rudderlabs/rudder-server/testhelper/rand"
+	"github.com/rudderlabs/rudder-server/testhelper/workspaceConfig"
+	"github.com/rudderlabs/rudder-server/utils/types/deployment"
+	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
+	"golang.org/x/sync/errgroup"
+)
+
+// TestEventOrderGuarantee tests that order delivery guarantees are honoured by the server
+// regardless the behaviour of the destination.
+// To achieve that we generate a number of jobs per user with each job having a prescribed number
+// of responses until it is done (succeeded or aborted)
+
+// e.g.
+//
+//	user1, job1, responses: [500, 500, 200]
+//	user1, job2, responses: [200]
+//	user1, job3, responses: [400]
+//	user1, job4, responses: [500, 400]
+
+// After sending the jobs to the server, we verify that the destination has received the jobs in the
+// correct order. We also verify that the server has not sent any job twice.
+func TestEventOrderGuarantee(t *testing.T) {
+	const (
+		users         = 50                   // how many different userIDs we will send jobs for
+		jobsPerUser   = 40                   // how many jobs per user we will send
+		batchSize     = 10                   // how many jobs for the same user we will send in each batch request
+		responseDelay = 0 * time.Millisecond // how long we will the webhook wait before sending a response
+	)
+	var (
+		m      eventOrderMethods
+		passed bool
+	)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// generate the spec we are going to use
+	// this will create a number of jobs for a number of users
+	// and prescribe the sequence of status codes that the webhook will return for each one e.g. 500, 500, 200
+	spec := m.newTestSpec(users, jobsPerUser)
+	spec.responseDelay = responseDelay
+
+	t.Logf("Starting docker services (postgres & transformer)")
+	var (
+		postgresContainer    *destination.PostgresResource
+		transformerContainer *destination.TransformerResource
+		gatewayPort          string
+	)
+	pool, err := dockertest.NewPool("")
+	require.NoError(t, err)
+	containersGroup, _ := errgroup.WithContext(ctx)
+	containersGroup.Go(func() (err error) {
+		postgresContainer, err = destination.SetupPostgres(pool, t)
+		return err
+	})
+	containersGroup.Go(func() (err error) {
+		transformerContainer, err = destination.SetupTransformer(pool, t)
+		return err
+	})
+	require.NoError(t, containersGroup.Wait())
+	t.Logf("Started docker services")
+
+	t.Logf("Starting webhook destination server")
+	webhook := m.newWebhook(t, spec)
+	defer webhook.Server.Close()
+
+	t.Logf("Preparing rudder-server config")
+	writeKey := trand.String(27)
+	workspaceID := trand.String(27)
+	templateCtx := map[string]string{
+		"webhookUrl":  webhook.Server.URL,
+		"writeKey":    writeKey,
+		"workspaceId": workspaceID,
+	}
+	configJsonPath := workspaceConfig.CreateTempFile(t, "testdata/eventOrderTestTemplate.json", templateCtx)
+	t.Setenv("RSERVER_BACKEND_CONFIG_CONFIG_FROM_FILE", "true")
+	t.Setenv("RSERVER_BACKEND_CONFIG_CONFIG_JSONPATH", configJsonPath)
+
+	t.Logf("Starting rudder-server")
+	t.Setenv("DEPLOYMENT_TYPE", string(deployment.DedicatedType))
+
+	t.Setenv("JOBS_DB_PORT", postgresContainer.Port)
+	t.Setenv("JOBS_DB_USER", postgresContainer.User)
+	t.Setenv("JOBS_DB_DB_NAME", postgresContainer.Database)
+	t.Setenv("JOBS_DB_PASSWORD", postgresContainer.Password)
+	t.Setenv("DEST_TRANSFORM_URL", transformerContainer.TransformURL)
+
+	t.Setenv("RSERVER_WAREHOUSE_MODE", "off")
+	t.Setenv("RSERVER_ENABLE_STATS", "false")
+	t.Setenv("RSERVER_JOBS_DB_BACKUP_ENABLED", "false")
+
+	// generatorLoop
+	t.Setenv("RSERVER_ROUTER_JOB_QUERY_BATCH_SIZE", "500")
+	t.Setenv("RSERVER_ROUTER_READ_SLEEP", "10ms")
+	t.Setenv("RSERVER_ROUTER_MIN_RETRY_BACKOFF", "5ms")
+	t.Setenv("RSERVER_ROUTER_MAX_RETRY_BACKOFF", "10ms")
+	t.Setenv("RSERVER_ROUTER_ALLOW_ABORTED_USER_JOBS_COUNT_FOR_PROCESSING", "1")
+
+	// worker
+	t.Setenv("RSERVER_ROUTER_NO_OF_JOBS_PER_CHANNEL", "100")
+	t.Setenv("RSERVER_ROUTER_JOBS_BATCH_TIMEOUT", "100ms")
+
+	// statusInsertLoop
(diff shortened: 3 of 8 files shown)


services/multitenant/tenantstats.go
```diff
@@ -241,6 +241,7 @@ func (t *Stats) GetRouterPickupJobs(destType string, noOfWorkers int, routerTime
 					timeRequired = 0
 				}
 				runningTimeCounter = runningTimeCounter - timeRequired
+				workspacePickUpCount[workspaceKey] = misc.MinInt(workspacePickUpCount[workspaceKey], runningJobCount)
```
atzoum, line 244: This was required to enforce the jobQueryBatchSize  limit in case a single workspace exists.
Overall, I think we are missing concise test scenarios and proper coverage for this critical piece of code

admin/admin.go
```diff
@@ -260,16 +261,18 @@ func StartServer(ctx context.Context) error {
...
+			}
 		}
 	}()
 
 	pkgLogger.Info("Serving on admin interface @ ", sockAddr)
 	srvMux := http.NewServeMux()
 	srvMux.Handle(rpc.DefaultRPCPath, instance.rpcServer)
 
-	srv := &http.Server{Handler: srvMux}
+	srv := &http.Server{Handler: srvMux, ReadHeaderTimeout: 3 * time.Second}
```
atzoum, line 275: deepsource was complaining about some attack vector, I think this won't hurt anyway here :)

admin/admin.go
```diff
@@ -260,16 +261,18 @@ func StartServer(ctx context.Context) error {
 		pkgLogger.Fatal("listen error:", e) // @TODO return?
 	}
 	defer func() {
-		if err := l.Close(); err != nil {
-			pkgLogger.Warn(err)
+		if l != nil {
```
atzoum, line 264: if `net.Listen` fails to listen, `l` will be of course... `nil`!

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,218 @@
+package eventorder
```
atzoum, line 1: This package contains the new event order algorithm implementation. Although it is not used by router yet, it is still tested under similar conditions in `eventorder_simulation_test.go`

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,218 @@
...
+		failedJob := *barrier.failedJobID
+		return jobID != failedJob // wait if this is not the failed job
+	}
+	// no faild job, don't wait
+	return false
+}
+
+// EnableConcurrencyLimiter enables the concurrency limiter for this barrier.
+// Should be called as soon as a job has failed in an unrecoverable manner (to be recorded as aborted)
+func (b *Barrier) EnableConcurrencyLimiter(userID string) {
```
lvrach, line 74: `[totaly minor]` Following the barrier abstraction, a simpler name _"creating a barrier for a user"_ sounds better to me.
 
```suggestion
func (b *Barrier) Create(userID string) {
```

```suggestion
func (b *Barrier) Set(userID string) {
```

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,218 @@
...
+	if !ok {
+		barrier = &barrierInfo{}
+		b.barriers[userID] = barrier
+	}
+	barrier.mu.Lock()
+	defer barrier.mu.Unlock()
+	if barrier.concurrentJobs == nil {
+		barrier.concurrentJobs = make(map[int64]struct{})
+	}
+}
```
lvrach, line 87: I believe:
```go
	if barrier.concurrentJobs == nil {
		barrier.concurrentJobs = make(map[int64]struct{})
	}
``` 
Can be avoided and instead:
1.  init the map when creating the struct
```suggestion
	if !ok {
		barrier = &barrierInfo{
		    concurrentJobs: make(map[int64]struct{})
		}
		b.barriers[userID] = barrier
	}
	barrier.mu.Lock()
	defer barrier.mu.Unlock()
	if barrier.concurrentJobs == nil {
		barrier.
	}
}
```
2. have a newBarrier() method that returns an initialised `barrierInfo` 
3. check and initialise the concurrentJobs JIT when it is first used.

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,218 @@
...
+
+// Flush applies any enqueued events to the barrier's state. It should be called at the beginning of every new iteration of the main loop
+func (s *Barrier) Flush() int {
```
lvrach, line 90: [minor] A better name would be more apt, since usually we flush stuff at the end :P

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,218 @@
...
+	var command command
+	jsCmd := &cmd{userID: userID, jobID: jobID}
+	switch state {
+	case "succeeded":
+		command = &jobSucceededCmd{jsCmd}
+	case "failed":
+		command = &jobFailedCommand{jsCmd}
+	case "aborted":
+		command = &JobAbortedCommand{jsCmd}
+	default:
```
lvrach, line 116: `[minor]` Could we introduce constants or "enums" for those?

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,218 @@
...
+	barrier, ok := b.barriers[userID]
+	if !ok {
+		return true, nil
+	}
+	if barrier.failedJobID != nil {
+		failedJob := *barrier.failedJobID
+		return jobID == failedJob, &failedJob
+	}
+
+	// concurrency limiter
```
lvrach, line 37: Should we mix concurrency limit logic with event ordering?

main.go
```diff
@@ -293,7 +293,7 @@ func Run(ctx context.Context) int {
 	shutdownDone := make(chan struct{})
 	go func() {
 		err := g.Wait()
-		if err != nil && err != context.Canceled {
+		if err != nil && err != context.Canceled && ctx.Err() == nil {
```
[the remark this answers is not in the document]
atzoum: not sure, can't it be that not the parent context was canceled, but another?

lvrach (commented):
Review part 1:  I do like the work on the event order abstraction a lot, since we are introducing a new concept I put some focus into the naming of things.

eventorder_test.go
```diff
@@ -0,0 +1,403 @@
+//go:build integration
```
lvrach, line 1:
```suggestion
```

I would prefer the usage of `SLOW` environment variable for those tests. build tags have some practical issues. A lot of go tools and IDEs / code editors ignore those files if not being explicit setup. The problem is that they are skipped silently without any indication. 

This is why we are using the `SLOW` environment variable; let's be consistent on this - or discuss it again in the next go-chapter meeting.

router/router.go
```diff
@@ -66,21 +66,25 @@ type HandleDestOAuthRespParamsT struct {
...
+	successCount          uint64
+	failCount             uint64
+	failedEventsListMutex sync.RWMutex
+	failedEventsList      *list.List
+	failedEventsChan      chan jobsdb.JobStatusT
+
+	eventOrderSyncMu        sync.Mutex // protects all maps below
+	toClearFailJobIDMap     map[int][]string
+	toClearAbortedJobIDMap  map[int][]string
+	toRemoveAbortedJobIDMap map[int]map[string]map[int64]struct{}
```
lvrach, line 86: Can you add a small description regarding the purpose of those maps?

main.go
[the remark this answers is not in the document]
lvrach: If it is not the parent context that was canceled, then it might be an actual issue we want to report.

config/backend-config/single_workspace.go
```diff
@@ -129,7 +130,9 @@ func (wc *singleWorkspaceConfig) getFromAPI(ctx context.Context, _ string) (Conf
 
 // getFromFile reads the workspace config from JSON file
 func (wc *singleWorkspaceConfig) getFromFile() (ConfigT, error) {
-	pkgLogger.Info("Reading workspace config from JSON file")
+	wc.logOnce.Do(func() {
```
lvrach, line 133: 👍

router/router.go
atzoum: no description will be good enough, but yes :)

router/router.go
```diff
@@ -1306,19 +1288,22 @@ func (rt *HandleT) findWorker(job *jobsdb.JobT, throttledAtTime time.Time) (toSe
 		// if yes returning worker only for 1 job
 		worker.abortedUserMutex.Lock()
 		defer worker.abortedUserMutex.Unlock()
-		if count, ok := worker.abortedUserIDMap[userID]; ok {
-			if count >= rt.allowAbortedUserJobsCountForProcessing {
+		if runningJobIDs, ok := worker.abortedUserIDMap[userID]; ok {
+			count := len(runningJobIDs)
+			_, hasJobID := runningJobIDs[job.JobID]
+
+			if !hasJobID && count >= rt.allowAbortedUserJobsCountForProcessing {
```
cisse21, line 1295: @atzoum  Did you find any use for this variable? `allowAbortedUserJobsCountForProcessing`

eventorder_test.go
```diff
@@ -0,0 +1,403 @@
...
+	// worker
+	t.Setenv("RSERVER_ROUTER_NO_OF_JOBS_PER_CHANNEL", "100")
+	t.Setenv("RSERVER_ROUTER_JOBS_BATCH_TIMEOUT", "100ms")
+
+	// statusInsertLoop
+	t.Setenv("RSERVER_ROUTER_UPDATE_STATUS_BATCH_SIZE", "100")
+	t.Setenv("RSERVER_ROUTER_MAX_STATUS_UPDATE_WAIT", "10ms")
+
+	// hack
+	// t.Setenv("RSERVER_ROUTER_GUARANTEE_USER_EVENT_ORDER", "false")
```
pChondros, line 132: [question] is this a leftover or base for future implementation?

eventorder_test.go
```diff
@@ -0,0 +1,403 @@
...
+
+	// find free ports for gateway and admin http server to listen on
+	httpPortInt, err := freeport.GetFreePort()
+	require.NoError(t, err)
+	gatewayPort = strconv.Itoa(httpPortInt)
+	// httpAdminPort, err := freeport.GetFreePort()
+	// require.NoError(t, err)
```
pChondros, line 139: [question] is this a leftover?

eventorder_test.go
```diff
@@ -0,0 +1,403 @@
...
+
+	t.Logf("waiting rudder-server to start properly")
+	require.Eventually(t, func() bool {
+		url := fmt.Sprintf("http://localhost:%s/health", gatewayPort)
+		if resp, err := http.Get(url); err == nil {
+			resp.Body.Close()
+			return resp.StatusCode == 200
+		}
+		return false
+	}, 20*time.Second, 100*time.Millisecond, "server should be able to start")
```
pChondros, line 168: [question] do you feel it makes sense to standardise the timeouts and polling periods by adding a central value for such timeouts and polling periods? this can make us wait for the same amount of time for server to start in different tests.

pChondros (approved):

router/router.go
lvrach: The difference between Clear and Remove is semantically very close. To understand the difference, you need to go back and forth inside the code. This is why I am suggesting a short explanation here. Ofc with the new abstraction, that won't be a problem.

lvrach (approved):

eventorder_test.go
```diff
@@ -0,0 +1,403 @@
+//go:build integration
+
+package main_test
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io/ioutil"
```
lvrach, line 9: This package is deprecated we should use `io` instead, I am wondering why linters don't complain

router/router.go
atzoum: Well this is the cause of evil 😆 it is the concurrency limit imposed to a specific userID after a job has been aborted 

Can you please elaborate on your point?

atzoum:
> Review part 1: I do like the work on the event order abstraction a lot, since we are introducing a new concept I put some focus into the naming of things.

@lvrach have moved development of the new abstraction to #2344.

eventorder_test.go
atzoum: if we standardize, this timeout needs to be adequately large to cover the worst case scenario. Please note that not all server configurations will require the same amount of time to startup.
What we do need is a test abstraction to start a rudder-server, like what we have now for all the rest (rudder-transformer, destinations, etc.). But for this to happen we need to move `Run` method outside the `main` package.

eventorder_test.go
atzoum: I am not fond of build tags either, but we need to change stuff in various places for the proposed changes to happen.
FYI, we have a lot of other integration tests, apart from docker_test which might be classified as SLOW :)

router/router.go
cisse21: I have always felt this was a variable which is not necessary :( why do we need to impose a concurrency limit at a user level if his job is aborted?

router/router.go
```diff
@@ -1544,21 +1529,49 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
...
+				if status == jobsdb.Succeeded.State {
+					// clear the map in the next generatorLoop
+					_, ok := rt.toClearAbortedJobIDMap[worker.workerID]
+					if !ok {
+						rt.toClearAbortedJobIDMap[worker.workerID] = make([]string, 0)
+					}
+					rt.toClearAbortedJobIDMap[worker.workerID] = append(rt.toClearAbortedJobIDMap[worker.workerID], userID)
+				} else { // any other state
+					if _, ok = runningJobs[resp.status.JobID]; ok {
+						// remove the jobID from the concurrency map during the next generatorLoop
```
cisse21, line 1560: Why would these job ID's even end up in the concurrency map?

router/router.go
atzoum: I didn't ever came to question this behaviour, so I suppose we have some good reason for doing it.
For sure this introduces a delay in processing for aborting users, but I suppose we are doing it for giving more room to successful jobs so that they can occupy most of the workers' time.

router/router.go
cisse21: Yeah I got it

eventorder_test.go
atzoum: yes will cleanup

eventorder_test.go
atzoum: uncomment it and the test will fail :)

router/router.go
atzoum: if they entered the pipeline while the concurrency limiter was on, they were added in router.go#1306

eventorder_test.go
lvrach: You are right, I would suggest we start with [removing](https://github.com/rudderlabs/rudder-server/pull/2346) build tags and then we can identify and mark slow ones

router/router.go
chandumlg: Explanation for why `allowAbortedUserJobsCountForProcessing` was introduced.

This is done to prevent job status table blow up with waiting statuses. Assume generator is pickup jobs of users whose jobs are getting aborted after multiple retries. Every time a job is aborted, a max of 10K jobs will be pushed to the worker. Of these 10K jobs, first one will be failed and rest 9999 are going to be marked as waiting. 

To prevent this, if we find that a users jobs are getting aborted, we control how many jobs of that user are sent downstream to the worker for processing. By default, the value of the variable  `allowAbortedUserJobsCountForProcessing` is 1. Which means, only one job of the aborted user will be sent down to the worker for processing.

chandumlg: Quick question: While the check `!hasJobID` has no harm in having. Is it really required?

atzoum: in a perfect world no this check wouldn't be required. This only guards us from forgetting to remove the jobID from the map, i.e. bug in the mechanism

chandumlg: Another question: Isn't just the count of jobs in flight enough as opposed to keeping all the jobids? Not a big deal, just thinking.

chandumlg (commented):
Minor questions

router/router.go
atzoum: no, if you notice, there is a small detail. Once a job is aborted, we might have other jobs for the same userID already accepted in the pipeline. These jobs are not counted for, so they shouldn't decrement the counter. Practically the concurrency limiter applies for new jobs only

chandumlg: Got it

chandumlg (approved):

PR #2336, opened 2022-08-23
fix: failedJobIDMutex stale write

fracasula:
# Description

Fixing wrong usage of `RLock` followed by `Lock`. By the time we lock and write into `failedJobIDMap`, the value of `isPrevFailedUser` might not longer be the same. I'm therefore locking once, checking the condition and writing before unlocking.

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/failedJobIDMutex-usage-in-router-919aee34ed994544828aadf300ba492c) >

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -349,7 +349,7 @@ func (worker *workerT) workerProcess() {
 			var parameters JobParametersT
 			err := json.Unmarshal(job.Parameters, &parameters)
 			if err != nil {
-				worker.rt.logger.Error("Unmarshal of job parameters failed. ", string(job.Parameters))
+				worker.rt.logger.Errorf("Unmarshal of job parameters failed: %s", string(job.Parameters))
 			}
 
 			var isPrevFailedUser bool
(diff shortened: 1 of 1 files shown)


router/router.go
```diff
@@ -1054,14 +1054,14 @@ func (worker *workerT) postStatusOnResponseQ(respStatusCode int, respBody string
 		if worker.rt.guaranteeUserEventOrder {
 			if addToFailedMap {
 				//#JobOrder (see other #JobOrder comment)
-				worker.failedJobIDMutex.RLock()
+				worker.failedJobIDMutex.Lock()
```
cisse21, line 1057: Any thoughts on this? 

```
			worker.failedJobIDMutex.Lock()
			if addToFailedMap {
				//#JobOrder (see other #JobOrder comment)
                //Do something
            } else {
                // Do something
             }
			worker.failedJobIDMutex.Unlock()
```

cisse21 (approved):

router/router.go
```diff
@@ -1054,14 +1054,14 @@ func (worker *workerT) postStatusOnResponseQ(respStatusCode int, respBody string
 		if worker.rt.guaranteeUserEventOrder {
 			if addToFailedMap {
 				//#JobOrder (see other #JobOrder comment)
-				worker.failedJobIDMutex.RLock()
```
atzoum, line 1057: IMO there is practically no way for a race to happen, since the only one who can write it is the go-routine itself.

router/router.go
fracasula: Are you suggesting that we should check `addToFailedMap` within the mutex lock? Isn't that beyond the scope of that mutex?

router/router.go
fracasula: Even if `postStatusOnResponseQ` doesn't get invoked concurrently the usage of the mutex there is wrong and should be fixed, also because it adds unnecessary risks. If someone were to add an additional routine that writes there it would surely break. Do you think it's best not to merge this change?

atzoum: Please keep in mind that the goal is to move away from the current implementation
#2329 fixes various practical issues with event ordering with regards to the current implementation and introduces a new algorithm/approach. Although the new approach won't be used immediately, the plan is to replace the old one in due time.

That being said, I will introduce your changes to #2329 in order to avoid unnecessary merging conflicts :)

router/router.go
cisse21: Yeah it is outside the scope but I was just wondering if we can avoid doing `worker.failedJobIDMutex.Unlock()` twice

router/router.go
lvrach: Those changes are on #2329, @fracasula are you cool closing this?

PR #2344, opened 2022-08-24
refactor(router): new event ordering algorithm with proper aborted job limiting

atzoum:
# Description

- Introduced a new `eventorder.Barrier` abstraction for implementing the event ordering algorithm.
- Replaced previous event ordering implementation with the new one in `router` component.

## Note
The following behavioral change is introduced by the new algorithm: 
**_The concurrency limiter will only be enabled if an `aborted `state is recorded for a job whose previous state was `failed`._**
In the previous implementation, the concurrency limiter was enabled whenever an `aborted `state was encountered for a job, regardless of the job's previous state history.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=d5a293e7a4c34b6399ab914bb1e879ae&pm=s)

## Security

### DIFF
--- a/eventorder_test.go
+++ b/eventorder_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"path"
 	"strconv"
 	"strings"
 	"sync"
@@ -100,6 +101,7 @@ func TestEventOrderGuarantee(t *testing.T) {
 
 	t.Logf("Starting rudder-server")
 	t.Setenv("DEPLOYMENT_TYPE", string(deployment.DedicatedType))
+	t.Setenv("RSERVER_RECOVERY_STORAGE_PATH", path.Join(t.TempDir(), "/recovery_data.json"))
 
 	t.Setenv("JOBS_DB_PORT", postgresContainer.Port)
 	t.Setenv("JOBS_DB_USER", postgresContainer.User)
@@ -192,7 +194,7 @@ func TestEventOrderGuarantee(t *testing.T) {
 			t.Logf("%d/%d done", done, total)
 		}
 		return done == total
-	}, 180*time.Second, 2*time.Second, "webhook should receive all events and process them till the end")
+	}, 300*time.Second, 2*time.Second, "webhook should receive all events and process them till the end")
 
 	require.False(t, t.Failed(), "webhook shouldn't have failed")
 
@@ -253,7 +255,7 @@ func (m eventOrderMethods) newTestSpec(users, jobsPerUser int) *eventOrderSpec {
 func (eventOrderMethods) randomStatus() (status int, terminal bool) {
 	// playing with probabilities: 50% HTTP 500, 40% HTTP 200, 10% HTTP 400
 	statuses := []int{
-		http.StatusBadRequest, http.StatusBadRequest, http.StatusBadRequest, http.StatusBadRequest, http.StatusBadRequest,
+		http.StatusBadRequest, http.StatusBadRequest, http.StatusBadRequest, http.StatusBadRequest,
 		http.StatusOK, http.StatusOK, http.StatusOK, http.StatusOK,
 		http.StatusInternalServerError,
 	}
@@ -299,7 +301,7 @@ func (eventOrderMethods) newWebhook(t *testing.T, spec *eventOrderSpec) *eventOr
 			if len(wh.spec.doneOrdered[userID]) > 0 {
 				lastDoneId = wh.spec.doneOrdered[userID][len(wh.spec.doneOrdered[userID])-1]
 			}
-			require.Greater(t, jobID, lastDoneId, "received out-of-order event for job %d after job %d", jobID, lastDoneId)
+			require.Greater(t, jobID, lastDoneId, "received out-of-order event for user %s: job %d after jobs %+v", userID, jobID, wh.spec.doneOrdered[userID])
 			wh.spec.done[jobID] = struct{}{}
 			wh.spec.doneOrdered[userID] = append(wh.spec.doneOrdered[userID], jobID)
 			// t.Logf("job %d done", jobID)
--- a/router/internal/eventorder/eventorder.go
+++ b/router/internal/eventorder/eventorder.go
@@ -1,0 +1,260 @@
+package eventorder
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/rudderlabs/rudder-server/jobsdb"
+)
+
+var ErrUnsupportedState = errors.New("unsupported state")
+
+// NewBarrier creates a new properly initialized Barrier
+func NewBarrier(abortConcurrencyLimit int) *Barrier {
+	return &Barrier{
+		concurrencyLimit: abortConcurrencyLimit,
+		barriers:         make(map[string]*barrierInfo),
+	}
+}
+
+// Barrier is an abstraction for applying event ordering guarantees in the router.
+//
+// Events for the same userID need to be processed in order, thus when an event fails but will be retried later by the router,
+// we need to put all subsequent events for that userID on hold until the failed event succeeds or fails
+// in a terminal way.
+//
+// A barrier controls the concurrency of events in two places:
+//
+// 1. At entrance, before the event enters the pipeline.
+//
+// 2. Before actually trying to send the event, since events after being accepted by the router, they are
+// processed asynchronously through buffered channels by separate goroutine(s) aka workers.
+type Barrier struct {
+	concurrencyLimit int
+	mu               sync.RWMutex
+	queue            []command
+	barriers         map[string]*barrierInfo
+}
+
+// Enter the barrier for this userID and jobID. If there is not already a barrier for this userID
+// returns true, otherwise false along with the previous failed jobID if this is the cause of the barrier.
+// Another scenario where a barrier might exist for a user is when the previous job has failed in an unrecoverable manner and the concurrency limiter is enabled.
+func (b *Barrier) Enter(userID string, jobID int64) (accepted bool, previousFailedJobID *int64) {
+	b.mu.RLock()
+	defer b.mu.RUnlock()
+	barrier, ok := b.barriers[userID]
+	if !ok {
+		return true, nil // no barrier, accept
+	}
+
+	// if there is a failed job in the barrier, only this job can enter the barrier
+	if barrier.failedJobID != nil {
+		failedJob := *barrier.failedJobID
+		return jobID == failedJob, &failedJob
+	}
+
+	// if the concurrency limiter is enabled, only the configured number of concurrent jobs can enter the barrier
+	if barrier.concurrencyLimiter != nil {
+		barrier.mu.Lock()
+		defer barrier.mu.Unlock()
+		if _, ok := barrier.concurrencyLimiter[jobID]; ok {
+			return true, nil // if the job is already in the concurrent jobs map, accept it (poor job... it forgot to notify the barrier before leaving!)
+		}
+		if len(barrier.concurrencyLimiter) >= b.concurrencyLimit {
+			return false, nil // if the concurrent jobs map is full, reject it
+		}
+
+		// add the job to the concurrent jobs map and accept it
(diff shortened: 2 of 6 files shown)


eventorder_test.go
```diff
@@ -100,6 +100,7 @@ func TestEventOrderGuarantee(t *testing.T) {
 
 	t.Logf("Starting rudder-server")
 	t.Setenv("DEPLOYMENT_TYPE", string(deployment.DedicatedType))
+	t.Setenv("RSERVER_MODE", "normal")
```
cisse21, line 103: Why did we have to set this env explicitly?  Isn't it in normal mode by default for Dedicated Deployment

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,255 @@
...
+	"errors"
+	"fmt"
+	"strings"
+	"sync"
+)
+
+type State string
+
+const (
+	SucceededState State = "succeeded"
```
cisse21, line 13: Probably we can use the jobsdb states here rather than re initialising them again

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,255 @@
...
+	FailedState    State = "failed"
+	AbortedState   State = "aborted"
+	WaitingState   State = "waiting"
+)
+
+var ErrUnsupportedState error = errors.New("unsupported state")
```
cisse21, line 19:
```suggestion
var ErrUnsupportedState = errors.New("unsupported state")
```

We can get rid of type declaration here

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,255 @@
...
+	if command.enqueue(s.barriers) {
+		s.queue = append(s.queue, command)
+	} else {
+		command.execute(s.barriers)
+	}
+	return nil
+}
+
+// Sync applies any enqueued commands to the barrier's state. It should be called at the beginning of every new iteration of the main loop
+func (s *Barrier) Sync() int {
```
cisse21, line 122: Any reason why we are using two receiver names for the same struct?

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,255 @@
...
+		return false, nil // no barrier, don't wait
+	}
+	if barrier.failedJobID != nil {
+		failedJob := *barrier.failedJobID
+		if failedJob > jobID {
+			panic(fmt.Errorf("previousFailedJob:%d > jobID:%d", failedJob, jobID))
+		}
+		return jobID > failedJob, &failedJob // wait if this is not the failed job
+	}
+	// no faild job, don't wait
```
cisse21, line 87:
```suggestion
	// no failed job, don't wait
```
Minor Typo

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,255 @@
...
+
+// JobAbortedCommand is a command that is executed when a job has aborted.
+type JobAbortedCommand struct {
+	*cmd
+}
+
+// Creates a concurrent jobs map if none exists. Also removes the jobID from the concurrent jobs map if it exists there
+func (c *JobAbortedCommand) execute(barriers map[string]*barrierInfo) {
+	if barrier, ok := barriers[c.userID]; ok {
+		if barrier.failedJobID == nil {
```
cisse21, line 240: Why do we not need a read lock over here?

eventorder_test.go
atzoum: At some point while running tests and having the server panic, I think the server could no longer start in normal mode, that's why I had to fix it

router/internal/eventorder/eventorder.go
atzoum: nope, thanks for spotting it 👍

router/internal/eventorder/eventorder.go
atzoum: We have already acquired a full `Barrier` lock in the caller, see `Barrier#StateChanged` and `Barrier#Sync`.
In case you are referring to the other lock, `barrierInfo.mu`, this lock is only there to protect `barrierInfo.concurrencyLimiter` as mentioned in its definition comment

router/internal/eventorder/eventorder.go
atzoum: The initial intention was to keep it clean and not introduce any other dependency apart from the standard libraries. Keeping it safe from import cycles too.

eventorder_test.go
cisse21: A better way would be to set the `recovery.storagePath` and clean up the directory every time?

router/internal/eventorder/eventorder.go
cisse21: Probably then move it from jobs db too to another common types folder. We might add a few more states and might miss to update it here

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,255 @@
...
+	if barrier, ok := barriers[c.userID]; ok {
+		if barrier.failedJobID != nil && *barrier.failedJobID != c.jobID { // out-of-sync command (failed commands get executed immediately)
+			return
+		}
+	}
+	delete(barriers, c.userID)
+}
+
+// JobAbortedCommand is a command that is executed when a job has aborted.
+type JobAbortedCommand struct {
```
cisse21, line 233: Do we need it exported? If not we can convert it to lower case

router/internal/eventorder/eventorder_simulation_test.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+}
+
+func (gl *generatorLoop) putBack(jobs []*job) {
+	gl.pendingMu.Lock()
+	defer gl.pendingMu.Unlock()
+	gl.pending = append(gl.pending, jobs...)
+	sort.Slice(gl.pending, func(i, j int) bool {
+		return gl.pending[i].id < gl.pending[j].id
+	})
```
cisse21, line 156:
```suggestion
func (g *generatorLoop) putBack(jobs []*job) {
	g.pendingMu.Lock()
	defer g.pendingMu.Unlock()
	g.pending = append(g.pending, jobs...)
	sort.Slice(g.pending, func(i, j int) bool {
		return g.pending[i].id < g.pending[j].id
	})
```

router/internal/eventorder/eventorder_simulation_test.go
```diff
@@ -0,0 +1,313 @@
...
+
+func (wp *workerProcess) processJobs() {
+	for _, job := range wp.jobs {
+
+		if wait, _ := wp.barrier.Wait(job.user, job.id); wait {
+			job.states = append([]eventorder.State{eventorder.WaitingState}, job.states...)
+			wp.out <- job
+			continue
+		}
+		// wp.logger.Logf("job %d processed %q (loop: %d)", job.id, job.states[0], job.loop)
```
cisse21, line 194:
```suggestion
```

router/internal/eventorder/eventorder_simulation_test.go
```diff
@@ -0,0 +1,313 @@
...
+			}
+		}
+	}
+}
+
+func (cl *commitStatusLoop) commit() {
+	var putBack []*job
+	for _, job := range cl.jobs {
+
+		// cl.logger.Logf("committing %d: %s (loop: %d)", job.id, job.states[0], job.loop)
```
cisse21, line 249:
```suggestion
```

router/router.go
```diff
@@ -357,22 +350,32 @@ func (worker *workerT) workerProcess() {
...
-
-				// mark job as waiting if prev job from same user has not succeeded yet
-				if isPrevFailedUser {
-					markedAsWaiting := worker.handleJobForPrevFailedUser(job, userID, previousFailedJobID)
-					if markedAsWaiting {
-						worker.rt.logger.Debugf(`Decrementing in throttle map for destination:%s since job:%d is marked as waiting for user:%s`, parameters.DestinationID, job.JobID, userID)
-						worker.rt.throttler.Dec(parameters.DestinationID, userID, 1, worker.throttledAtTime, throttler.ALL_LEVELS)
-						continue
+				if wait, previousFailedJobID := worker.barrier.Wait(userID, job.JobID); wait {
+					previousFailedJobIDStr := "<nil>"
```
cisse21, line 355: move this variable and initialise at global level?

cisse21: makes sense

cisse21 (approved):

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,248 @@
...
+	queue            []command
+	barriers         map[string]*barrierInfo
+}
+
+// Enter the barrier for this userID and jobID. If there is not already a barrier for this userID
+// returns true, otherwise false along with the previous failed jobID if this is the cause of the barrier.
+// Another scenario where a barrier might exist for a user is when the previous job has failed in an unrecoverable manner and the concurrency limiter is enabled.
+func (b *Barrier) Enter(userID string, jobID int64) (accepted bool, previousFailedJobID *int64) {
+	b.mu.RLock()
+	defer b.mu.RUnlock()
```
fracasula, line 34: Why not releasing this lock sooner? Is this done on purpose because you want to finish the operations below anyway before anybody can write in `b.barriers`?

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,248 @@
...
+
+// NewBarrier creates a new properly initialized Barrier
+func NewBarrier(abortConcurrencyLimit int) *Barrier {
+	return &Barrier{
+		concurrencyLimit: abortConcurrencyLimit,
+		barriers:         make(map[string]*barrierInfo),
+	}
+}
+
+type Barrier struct {
```
fracasula, line 22: We could have a comment here explaining what a `Barrier` is for.

router/internal/eventorder/eventorder.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	// if there is a failed job in the barrier, only this job can enter the barrier
+	if barrier.failedJobID != nil {
+		failedJob := *barrier.failedJobID
+		return jobID == failedJob, &failedJob
+	}
+
+	// if the concurrency limiter is enabled, only the configured number of concurrent jobs can enter the barrier
+	if barrier.concurrencyLimiter != nil {
```
fracasula, line 47: [optional] early return
```suggestion
	if barrier.concurrencyLimiter == nil {
		return true, nil
	}
```

router/internal/eventorder/eventorder.go
atzoum: correct, even though `Enter` & `Sync` methods are not going to be called by multiple threads, they are synchronisation points

router/internal/eventorder/eventorder.go
atzoum: yes, how did I miss it :)

router/internal/eventorder/eventorder.go
fracasula: Sounds like a good candidate for a comment then :wink:

router/internal/eventorder/eventorder_test.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func Test_Job_Abort_then_Fail(t *testing.T) {
+	barrier := NewBarrier(2)
+
+	enter, previousFailedJobID := barrier.Enter("user1", 1)
+	require.True(t, enter, "job 1 for user1 should be accepted since no barrier exists")
+	require.Nil(t, previousFailedJobID)
+
+	require.False(t, firstBool(barrier.Wait("user1", 1)), "job 3 for user1 shouldn't wait")
```
fracasula, line 127: I think here you meant "job 1 for user1 shouldn't wait".

router/internal/eventorder/eventorder.go
atzoum: actually, access to all barrier's methods (commands) should be synchronised, thus added a relevant comment in the mutex's declaration

fracasula (approved):

PR #2346, opened 2022-08-24
chore: remove build tags from tests

lvrach:
# Description

The usage of build tags to enable/disable specific tests files `//go:build integration` is problematic:
1. By default, most go tools will skip those files if not configured correctly. Linting and formatting tools won't fix or report issues on those files. This applies to cli invocations, IDEs and services like deepsource.
2. Testsuites do not warn/ complain about those files being skipped; you might run `go test` or `gingko` it will return without reporting any problem. This can create confusion, or tests end up skipped at worst.

When running `make test` we were already running the tests under `integration` tag.

## Notion Ticket

https://www.notion.so/rudderstacks/remove-integration-tag-2fa2eb23004146b4a19b5821a6f6ca8a

## Security

### DIFF
--- a/docker_test.go
+++ b/docker_test.go
@@ -235,7 +235,7 @@ func TestMainFlow(t *testing.T) {
 		messages, errors := consume(t, c, topics)
 
 		signals := make(chan os.Signal, 1)
-		signal.Notify(signals, os.Interrupt, os.Kill) // Get signal for finish
+		signal.Notify(signals, os.Interrupt, syscall.SIGTERM) // Get signal for finish
 
 		var (
 			msgCount      = 0 // Count how many message processed
--- a/docker_test.go
+++ b/docker_test.go
@@ -3,8 +3,6 @@
 // It then runs the service ensuring it is configured to use the dependencies.
 // Finally, it sends events and observe the destinations expecting to get the events back.
 
-//go:build integration
-
 package main_test
 
 import (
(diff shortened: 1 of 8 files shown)


docker_test.go
```diff
@@ -237,7 +234,7 @@ func TestMainFlow(t *testing.T) {
 		messages, errors := consume(t, c, topics)
 
 		signals := make(chan os.Signal, 1)
-		signal.Notify(signals, os.Interrupt, os.Kill) // Get signal for finish
+		signal.Notify(signals, os.Interrupt, syscall.SIGTERM) // Get signal for finish
```
fracasula, line 237: [optional]

```suggestion
		signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM) // Get signal for finish
```

fracasula (approved):

atzoum:
@lvrach please search for `-tags=integration` and remove it from the `Makefile` & `.yaml` descriptors

lvrach:
> @lvrach please search for -tags=integration and remove it from the Makefile & .yaml descriptors
oups, I have done it but then I break the PR in half incorrectly

docker_test.go
lvrach: I like `os.Interrupt`, it is os agnostic. Unfortunately there is no similar thing for `syscall.SIGTERM`

chandumlg (approved):

PR #2347, opened 2022-08-24
chore: upgrade golangci and linting fixes

lvrach:
# Description

We upgrade golangci to the latest version, which enabled a lot of linters previously disabled due to lack of go1.18 support.

## Notion Ticket

https://www.notion.so/rudderstacks/Upgrade-golangci-and-fix-new-issues-8c3286adfbac4872b58408341b3610e9

## Security


enterprise/migrator/migrator.go
```diff
@@ -61,6 +61,7 @@ func New(mode string, jobsDB *jobsdb.HandleT, pf pathfinder.ClusterStateT) Migra
 
 func loadConfig() {
 	dbReadBatchSize = config.GetInt("Migrator.dbReadBatchSize", 10000)
+	config.RegisterDurationConfigVariable(0, &ReadHeaderTimeout, false, time.Second, []string{"ReadHeaderTimeout", "ReadHeaderTimeoutInSec"}...)
```
atzoum, line 64: shouldn't this be prefixed with `Migrator.` ?

.golangci.yml
```diff
@@ -36,3 +36,19 @@ issues:
...
+    # False positive .Close behind if
+    - path: 'processor/transformer/transformer.go'
+      linters:
+        - bodyclose
+    
+    # False positive httptest.NewRecorder
+    - path: 'gateway/gateway_test.go'
+      linters:
+        - bodyclose
+
```
lvrach, line 54: bodyclose linter is very useful, but has a lot of false positives. I try to skip specific false positives without polluting the code.

jobsdb/jobsdb.go
```diff
@@ -1522,7 +1521,7 @@ func (jd *HandleT) createDS(newDS dataSetT, l lock.DSListLockToken) {
 	if l != nil {
 		// to get the updated DS list in the cache after createDS transaction has been committed.
 		_ = jd.refreshDSList(l)
```
atzoum, line 1523: `refreshDSRangeList` calls `refreshDSList` internally
```suggestion
```

jobsdb/jobsdb.go
```diff
@@ -1598,7 +1597,7 @@ func (jd *HandleT) createDSInTx(tx *sql.Tx, newDS dataSetT, l lock.DSListLockTok
 	return nil
 }
 
-func (jd *HandleT) setSequenceNumberInTx(tx *sql.Tx, l lock.DSListLockToken, newDSIdx string) error {
+func (jd *HandleT) setSequenceNumberInTx(tx *sql.Tx, _ lock.DSListLockToken, newDSIdx string) error {
```
atzoum, line 1600: we are missing an assertion here that `l` is not `nil`

jobsdb/jobsdb_export.go
```diff
@@ -182,6 +183,9 @@ func (jd *HandleT) getNonMigratedJobsFromDS(ds dataSetT, count int) ([]*JobT, er
 
 	jd.logger.Info(sqlStatement)
 	rows, err = jd.dbHandle.Query(sqlStatement)
+	if err != nil {
+		return nil, fmt.Errorf("query: %w", err)
+	}
 	jd.assertError(err)
```
atzoum, line 189: what is the impact of returning an error instead of panicking?
if we should indeed return an error instead of panicking, what's the point of the assertion in line 189?

The above comment stands for the whole method. Do we use the method? Do we have tests for it? Business-wise I cannot judge whether this change is safe or not.

jobsdb/jobsdb.go
lvrach: Do you mean this method should check that `l` is not `nil` instead of ignoring it?

enterprise/migrator/migrator.go
lvrach: yeah, also `ReadHeaderTimeoutInSec` is not needed, the `Sec` suffix is deprecated cc @BonapartePC - I will fix this

jobsdb/jobsdb_export.go
lvrach: I refactor the method to return an error instead of assert/panic. The process that calls this will panic instead with `.Fatalf`.   

`jobsdb_export` is used for scaling out rudder-server instances. Multitenant code will deprecate the jobsdb `export` and `import` needs.


Unfortunately, there are no tests unit or integrations for this functionality. The fact that we are going to deprecate this and the effort required for refactoring/testing make it a low priority.

I can see two options:
  * skip `jobsdb_export`, `jobsdb_import` for deepsource, golangci
  * make those few safe changes to make the linters passing
  
 Correctly understanding, fixing and testing export/import will require a lot of effort, which is better spent on delivering multitenant.

services/filemanager/filemanager.go
```diff
@@ -133,7 +133,11 @@ func GetProviderConfigForBackupsFromEnv(ctx context.Context) map[string]interfac
 	return providerConfig
 }
 
-func getBatchRouterDurationConfig(key, destType string, defaultValueInTimescaleUnits int64, timeScale time.Duration) time.Duration {
+func getBatchRouterTimeoutConfig(destType string) time.Duration {
```
atzoum, line 136: I agree with this change, since I have proposed the same thing in the past, however I remember @chandumlg having a different opinion on that

app/cluster/integration_test.go
```diff
@@ -103,6 +98,12 @@ func run(m *testing.M) int {
 		return 1
 	}
 
+	defer func() {
+		if err := pool.Purge(resourcePostgres); err != nil {
+			log.Printf("Could not purge resource: %s \n", err)
+		}
+	}()
```
fracasula, line 105: How come you're moving this afterwards? Shouldn't we purge it even if we couldn't connect?

enterprise/migrator/export.go
```diff
@@ -43,6 +43,7 @@ type exporterT struct {
 var (
 	dbReadBatchSize              int
 	exportDoneCheckSleepDuration time.Duration
+	ReadHeaderTimeout            time.Duration
```
fracasula, line 46: Do we need this exported?

jobsdb/jobsdb.go
```diff
@@ -1521,8 +1520,7 @@ func (jd *HandleT) createDS(newDS dataSetT, l lock.DSListLockToken) {
 	// In case of a migration, we don't yet update the in-memory list till we finish the migration
 	if l != nil {
 		// to get the updated DS list in the cache after createDS transaction has been committed.
-		_ = jd.refreshDSList(l)
```
fracasula, line 1524: Can you explain why this line was removed?

jobsdb/jobsdb_export.go
```diff
@@ -177,12 +175,13 @@ func (jd *HandleT) getNonMigratedJobsFromDS(ds dataSetT, count int) ([]*JobT, er
...
 
-	jd.assert(count > 0, fmt.Sprintf("count should be greater than 0, but count = %d", count))
 	sqlStatement += fmt.Sprintf(" LIMIT %d", count)
 
 	jd.logger.Info(sqlStatement)
-	rows, err = jd.dbHandle.Query(sqlStatement)
-	jd.assertError(err)
+	rows, err := jd.dbHandle.Query(sqlStatement)
+	if err != nil {
+		return nil, fmt.Errorf("query: %w", err)
```
fracasula, line 183:
```suggestion
		return nil, fmt.Errorf("getNonMigratedJobsFromDS query: %w", err)
```

jobsdb/jobsdb_export.go
```diff
@@ -196,9 +195,9 @@ func (jd *HandleT) getNonMigratedJobsFromDS(ds dataSetT, count int) ([]*JobT, er
 			&sqlJobStatusT.ExecTime, &sqlJobStatusT.RetryTime,
 			&sqlJobStatusT.ErrorCode, &sqlJobStatusT.ErrorResponse)
 		if err != nil {
-			jd.logger.Info(err)
+			return nil, fmt.Errorf("scan rows: %w", err)
```
fracasula, line 198: Perhaps add somecontext about the function that ran this. WDYT?

fracasula: Same for other errors, where it applies.

jobsdb/jobsdb_renameDs_test.go
```diff
@@ -30,9 +30,9 @@ func Test_mustRenameDS(t *testing.T) {
 
 		// And I have jobs and job status tables with events from 3 sources
 		createTables(t, dbHandle, jobsTable, jobStatusTable)
-		addJob(t, dbHandle, jobsTable, jobStatusTable, "one", "succeeded")
-		addJob(t, dbHandle, jobsTable, jobStatusTable, "two", "failed")
-		addJob(t, dbHandle, jobsTable, jobStatusTable, "three", "aborted")
+		addJob(t, dbHandle, "one", "succeeded")
```
fracasula, line 33:
```suggestion
		addJob(t, dbHandle, "one", Succeeded.State)
```

I believe we can do that for the 2 ones below too. Optional of course.

fracasula (approved):

jobsdb/jobsdb.go
atzoum: I can, since I requested it :) 
It is redundant, the next line calls another method which internally performs a refresh too.

atzoum (approved):

jobsdb/jobsdb.go
fracasula: Makes sense, thanks :+1:

PR #2348, opened 2022-08-24
fix(warehouse): upload validations should happen when validator is set

achettyiitr:
# Description

Destination validation should happen only if validators are set

## Notion Ticket

https://www.notion.so/rudderstacks/destination-validation-should-happen-only-if-validators-are-set-14b50c9ab667454c81b18cee66c93940

## Security


abhimanyubabbar (approved):
LGTM 👍🏼

cisse21 (approved):

PR #2356, opened 2022-08-25
feat(jobsdb): thread safe addNewDSLoop

saurav-malani:
# Description
Currently, `addNewDSLoop` that is responsible for continuously checking if latest DS is full or not & creates a new one if it is full is not thread safe, i.e if two instance of JobsDB are running in parallel & both found that the last DS is full then both the instance will try to add a new DS & one will succeed & other will panic. 

In this PR, we solved the above described problem & made it thread safe.

TLDR:

- Essentially, to coordinate between multiple jobsDB instances we are using advisory lock, i.e. to add a new DS, jobsDB has to first acquire an advisory lock `JobsDBAddDsAdvisoryLock`. Once acquired, only then it can add new DS. 
- Also, to make sure that no one else adds a new DS from the time we checked if the latest DS is full & actually add a DS we are doing the following:-  
   1. start a transaction
   2. acquire `JobsDBAddDsAdvisoryLock` lock
   3. get refreshed DS
   4. check if latest DS is full
   5. if yes, add new DS
   6. end transaction
   7. refresh cache to get new DS. 
   8. release `JobsDBAddDsAdvisoryLock` lock
   
## Notion Ticket
 https://www.notion.so/rudderstacks/Thread-safe-addNewDSLoop-5955828ee79f4f96b5a6685f3b10b613

### DIFF
--- a/jobsdb/internal/lock/lock.go
+++ b/jobsdb/internal/lock/lock.go
@@ -30,6 +30,21 @@ func (r *DSListLocker) WithLock(f func(l DSListLockToken)) {
 	f(&listLockToken{})
 }
 
+// AsyncLock acquires a lock unitl the token is returned to the receiving channel
+func (r *DSListLocker) AsyncLock() (DSListLockToken, chan<- DSListLockToken) {
+	acquireDsListLock := make(chan DSListLockToken)
+	releaseDsListLock := make(chan DSListLockToken)
+
+	go func() {
+		r.WithLock(func(l DSListLockToken) {
+			acquireDsListLock <- l
+			<-releaseDsListLock
+		})
+	}()
+	dsListLock := <-acquireDsListLock
+	return dsListLock, releaseDsListLock
+}
+
 type listLockToken struct{}
 
 func (*listLockToken) listLockToken() {
--- a/jobsdb/internal/lock/lock.go
+++ b/jobsdb/internal/lock/lock.go
@@ -30,7 +30,7 @@ func (r *DSListLocker) WithLock(f func(l DSListLockToken)) {
 	f(&listLockToken{})
 }
 
-// AsyncLock acquires a lock unitl the token is returned to the receiving channel
+// AsyncLock acquires a lock until the token is returned to the receiving channel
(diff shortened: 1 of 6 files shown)


jobsdb/jobsdb.go
```diff
@@ -1583,22 +1569,26 @@ func (jd *HandleT) createDSInTx(tx *sql.Tx, newDS dataSetT, l lock.DSListLockTok
...
+		return err
 	}
 
 	err = jd.journalMarkDoneInTx(tx, opID)
 	if err != nil {
 		return err
 	}
 
+	// Tracking time interval between new ds creations. Hence calling end before start
+	if jd.isStatNewDSPeriodInitialized {
```
saurav-malani, line 1583: previously, we were were not tracking time, if new DS was added because of migration. But, with this change we will always be tracking the time interval between new DS creations. Is that fine?

saurav-malani: kindly ignore, this is not valid any anymore.

atzoum (approved):

utils/misc/dbutils.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/rudderlabs/rudder-server/config"
 )
 
+type AdvisoryLock int
+
+const (
+	JobsDBAddDsAdvisoryLock = 11
```
saurav-malani, line 16:
```suggestion
	JobsDBAddDsAdvisoryLock = 11
```
        JobsDBAddDsAdvisoryLock AdvisoryLock = 11

jobsdb/internal/lock/lock.go
```diff
@@ -30,6 +30,21 @@ func (r *DSListLocker) WithLock(f func(l DSListLockToken)) {
 	f(&listLockToken{})
 }
 
+// AsyncLock acquires a lock unitl the token is returned to the receiving channel
```
saurav-malani, line 33:
```suggestion
// AsyncLock acquires a lock until the token is returned to the receiving channel
```

Sidddddarth (approved):

PR #2357, opened 2022-08-25
chore: generate router per destination id

cisse21:
# Description

In the current implementation we spin up a router at the destination type level, This is to ensure event ordering across users for the same destination. But again this is not applicable for WEBHOOK destinations as they are pretty generic and each WEBHOOK might serve a different use case. This PR aims to set up a router per destinationID for WEBHOOK destinations

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=54e42282047f463fa8496961ae889531&pm=s)

## Security

### DIFF
--- a/app/apphandlers/setup.go
+++ b/app/apphandlers/setup.go
@@ -199,13 +199,13 @@ func monitorDestRouters(ctx context.Context, routerFactory *router.Factory, batc
 						dstToBatchRouter[destination.DestinationDefinition.Name] = brt
 					}
 				} else {
-					_, ok := dstToRouter[destination.DestinationDefinition.Name]
+					_, ok := dstToRouter[misc.GetRouterIdentifier(destination.ID, destination.DestinationDefinition.Name)]
 					if !ok {
 						pkgLogger.Info("Starting a new Destination ", destination.DestinationDefinition.Name)
-						router := routerFactory.New(destination.DestinationDefinition)
+						router, identifier := routerFactory.New(*destination)
 						router.Start()
 						cleanup = append(cleanup, router.Shutdown)
-						dstToRouter[destination.DestinationDefinition.Name] = router
+						dstToRouter[identifier] = router
 					}
 				}
 			}
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1019,7 +1019,7 @@ func (proc *HandleT) getFailedEventJobs(response transformer.ResponseT, commonMe
 			Parameters:   marshalledParams,
 			CreatedAt:    time.Now(),
 			ExpireAt:     time.Now(),
-			CustomVal:    commonMetaData.DestinationType,
+			CustomVal:    misc.GetRouterIdentifier(commonMetaData.DestinationID, commonMetaData.DestinationType),
 			UserID:       failedEvent.Metadata.RudderID,
 			WorkspaceId:  failedEvent.Metadata.WorkspaceID,
 		}
@@ -2064,7 +2064,7 @@ func (proc *HandleT) transformSrcDest(
 				Parameters:   marshalledParams,
 				CreatedAt:    time.Now(),
 				ExpireAt:     time.Now(),
-				CustomVal:    destType,
+				CustomVal:    misc.GetRouterIdentifier(destID, destType),
 				EventPayload: destEventJSON,
 				WorkspaceId:  workspaceId,
 			}
--- a/router/factory.go
+++ b/router/factory.go
@@ -1,6 +1,8 @@
 package router
 
 import (
+	"github.com/rudderlabs/rudder-server/utils/misc"
+
 	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
 	"github.com/rudderlabs/rudder-server/jobsdb"
 	"github.com/rudderlabs/rudder-server/services/rsources"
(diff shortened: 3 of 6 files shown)


lvrach (commented):
Even as a hack I don't like the fact that custom val can take either destination ID or definition ID. 

Some jobs might get stuck when we turn isolation flag on and off, right ?

atzoum:
> Even as a hack I don't like the fact that custom val can take either destination ID or definition ID.
> 
> Some jobs might get stuck when we turn isolation flag on and off, right ?

I agree that changing the semantics of `custom_val` may open the door for new problems, some of which we might not even be aware of right now. One thing is certain, that we all assume now that the `custom_val` column, even though its name "sounds" generic, it contains the destination's type. This is strong assumption and we need to be careful as to what the implications might be of breaking it now.
We already have the `destinationID` in the `parameters` column, if that is not enough we could introduce a new column for further segregating jobs. But then another problem arises, that of picking up jobs for the router instance that is processing webhook jobs for all other destinations except for the isolated ones: how are we making sure that we will be excluding isolated jobs from this router's generator loop? Maybe a separate column would be more appropriate for that purpose. But still this would affect other areas, e.g. jobsdb & its caching mechanism

IMO we need to further elaborate on the whole solution and prepare a design document containing a complete proposal and covering all aspects and associated gaps, e.g.
- What are our available solutions with their pros and cons?
- How does the proposed solution affect the fair pickup algorithm (`GetRouterPickupJobs`) and the way that we collect stats for it?
- How does this affect the router component? Do we even need a fair pickup algorithm when we are targeting a single destination?
- How do we handle relevant isolation configuration changes gracefully (turning isolation on and off), making sure that no jobs will be lost or processed out-of-sequence?

lvrach:
@atzoum i don't disagree with your comment, but there is a context for this PR that you probably are not aware and we haven't discussed in the standup yet as we weren't sure we would go with this solution.

There is an active customer issue for which we want provide a quick solution.

That of course doesn't mean we are not going to address it property. There many things to be consider for destination isolation what you mentioned plus multi tenancy.

Please have in mind in reviewing this draft that is meant as a hot fix. Once we deliver this to the customer, we will start working on a design document.

cisse21:
@atzoum  The context for doing this PR
https://rudderlabs.slack.com/archives/C02AP9GSYBT/p1661253763537919

atzoum:
Quick-n-dirty or not, even with a single workspace, the last 3 bullets need to be answered/handled. As soon as we merge this functionality we will start using it for customers.
> - How does the proposed solution affect the fair pickup algorithm (GetRouterPickupJobs) and the way that we collect stats for it?
> - How does this affect the router component? Do we even need a fair pickup algorithm when we are targeting a single destination?
> - How do we handle relevant isolation configuration changes gracefully (turning isolation on and off), making sure that no jobs will be lost or processed out-of-sequence?

cisse21:
> How does the proposed solution affect the fair pickup algorithm (GetRouterPickupJobs) and the way that we collect stats for it?

These both won't go hand in hand for now. PickUpJobs algorithm cannot be applied at a destinationId level as there is no fairness associated for a single destinationID

> How does this affect the router component? Do we even need a fair pickup algorithm when we are targeting a single destination?

We would not need a fair pick up algorithm for the short term but in the long term we would need to utilise the algorithm at a `workspaceId-destId` level rather than a `workspaceId-destType `level

> How do we handle relevant isolation configuration changes gracefully (turning isolation on and off), making sure that no jobs will be lost or processed out-of-sequence?

For the customer in question right now, we have already disabled the event ordering and this flag would be turned on by intimating the customers with the risk involved of losing event ordering for a short duration

atzoum:
Can we introduce a new integration test to verify the new behaviour?

1. Start postgres & transformer containers
2. Start 2 webhook destinations (controlled by the test)
3. Setup a source with 2 destinations (1 destination for each webhook)
4. Start server with this configuration & with isolation enabled for webhooks
5. Switch webhook-1 mode to fail requests by responding with HTTP 500
6. Send events for the same userID for both webhooks (e.g. 1000 events)
7. Verify that all events have reached webhook 2 successfully
8. Switch webhook 1 to accept requests
9. Verify that all events have reached webhook 1 successfully

A similar integration test can be found in `docker_test.go` or [here](https://github.com/rudderlabs/rudder-server/pull/2344/files#diff-961a4b58170845ff103f726a396c154fc13999450eb6c4b83e65aebc933a00fd)

atzoum:
At least, we need to make sure that no jobs will be lost when we turn isolation on & off.

atzoum:
have we considered using 1 router, but having a different set of workers per destinationId?

cisse21:
> have we considered using 1 router, but having a different set of workers per destinationId?

Having separate workers would not solve the issue if the first 10,000 jobs belong to single failing destination id. We would also have to change the way we query by using parameter filters instead of custom val with that approach

atzoum:
Indeed, the current job filtering design during the generatorLoop can cause worker starvation easily. We should come up with a smarter approach in the near future.

cisse21:
We can re-use the fair pick-up algorithm to avoid such starvation but we would need to come to a consensus on that and will require extensive changes to the current code flow

cisse21:
> start with isolation turned off, have webhook-1 rejecting requests with HTTP 500, and send some events
> switch isolation to on, have webhook-1 accepting requests, verify all previous events have been processed successfully
> Repeat the same for the inverse sequence, start with isolation turned on, then turn it off

I don't think with the current approach we can have an automated way to easily switch the flag on and off and ensure everything, that would require us to make changes in jobsdb too which I am trying to avoid as we aim for a temporary fix. In the current implementation switching the flag off would require some manual intervention to change the custom val on the database tables

atzoum:
> We can re-use the fair pick-up algorithm to avoid such starvation but we would need to come to a consensus on that and will require extensive changes to the current code flow

IMO this is a problem that can be solved without adapting the fair pickup algorithm and have application both for non multitenant deployments.

PR #2358, opened 2022-08-26
feat: addition of Azure SAS Tokens for Azure Object storage destinations 

achettyiitr:
# Description

> Grant limited access to Azure Storage resources using shared access signatures (SAS). This will be going to provide limited access to RudderStack while writing the files.

## Notion Ticket

https://www.notion.so/rudderstacks/Add-SAS-Tokens-support-for-Azure-Storage-af27e5cfb1d742989d83b45194de54ae

## Security


services/filemanager/azureBlobStoragemanager.go
```diff
@@ -14,10 +14,10 @@ import (
 	"github.com/Azure/azure-storage-blob-go/azblob"
 )
 
-func supressMinorErrors(err error) error {
+func suppressMinorErrors(err error) error {
 	if err != nil {
-		if serr, ok := err.(azblob.StorageError); ok { // This error is a Service-specific
-			switch serr.ServiceCode() { // Compare serviceCode to ServiceCodeXxx constants
+		if storageError, ok := err.(azblob.StorageError); ok { // This error is a Service-specific
```
abhimanyubabbar, line 19: Appreciate your effort towards clean code. 🎉

services/filemanager/azureBlobStoragemanager.go
```diff
@@ -265,9 +313,11 @@ type AzureBlobStorageConfig struct {
 	Prefix         string
 	AccountName    string
 	AccountKey     string
+	SASToken       string
 	EndPoint       *string
 	ForcePathStyle *bool
 	DisableSSL     *bool
+	UseSASTokens   *bool
```
abhimanyubabbar, line 320: To keep things simple, can we treat it as bool and treat the default value `False` as to be the base case ?

services/filemanager/azureBlobStoragemanager.go
```diff
@@ -77,6 +75,24 @@ func (manager *AzureBlobStorageManager) getContainerURL() (azblob.ContainerURL,
...
+	if accountName == "" || accountKey == "" {
+		return nil, errors.New("either the AccountName or AccountKey is not correct")
+	}
+
+	// Create a default request pipeline using your storage account name and account key.
+	return azblob.NewSharedKeyCredential(accountName, accountKey)
+}
+
+func (manager *AzureBlobStorageManager) useSASTokens() bool {
+	return manager.Config.UseSASTokens != nil && *manager.Config.UseSASTokens
```
abhimanyubabbar, line 93: We can simplify this check greatly in case we use bool

abhimanyubabbar (changes requested):
Some comments regarding the datatype used for sas tokens

services/filemanager/azureBlobStoragemanager.go
achettyiitr: The reason we started using *bool is to handle if the config key is not present. In that case pointer will points to nil.

achettyiitr: And we can handle that case as well.

services/filemanager/azureBlobStoragemanager.go
achettyiitr: The reason we started using *bool is to handle if the config key is not present. In that case, the pointer will point to nil.
And we can handle that case as well instead of going for default as false.

services/filemanager/azureBlobStoragemanager.go
```diff
@@ -105,10 +123,26 @@ func (manager *AzureBlobStorageManager) Upload(ctx context.Context, file *os.Fil
...
+func (manager *AzureBlobStorageManager) blobLocation(blobURL *azblob.BlockBlobURL) string {
+	if !manager.useSASTokens() {
+		return blobURL.String()
+	}
+
+	// Reset SAS Query parameters
+	blobURLParts := azblob.NewBlobURLParts(blobURL.URL())
+	blobURLParts.SAS = azblob.SASQueryParameters{}
+	newBlobURL := blobURLParts.URL()
+	return newBlobURL.String()
```
atzoum, line 142: minor and subjective
```suggestion
	return blobURLParts.URL().String()
```

services/filemanager/azureBlobStoragemanager.go
```diff
@@ -105,10 +123,26 @@ func (manager *AzureBlobStorageManager) Upload(ctx context.Context, file *os.Fil
 		return UploadOutput{}, err
 	}
 
-	return UploadOutput{Location: blobURL.String(), ObjectName: fileName}, nil
+	return UploadOutput{Location: manager.blobLocation(&blobURL), ObjectName: fileName}, nil
 }
 
-func (manager *AzureBlobStorageManager) ListFilesWithPrefix(ctx context.Context, startAfter, prefix string, maxItems int64) (fileObjects []*FileObject, err error) {
+func (manager *AzureBlobStorageManager) shouldCreateContainer() bool {
```
atzoum, line 129: minor and subjective
```suggestion
func (manager *AzureBlobStorageManager) createContainer() bool {
```

services/filemanager/azureBlobStoragemanager.go
atzoum: I agree that if this new flag had the opposite default value, things would be different. But in this specific scenario isn't it enough to assume a default value of `false`?

achettyiitr: Yes @atzoum you are correct. Even if the flag is not present it would be correct to assume that it should be false.

services/filemanager/azureBlobStoragemanager.go
achettyiitr: Will make the changes.

services/filemanager/azureBlobStoragemanager.go
achettyiitr: Will make the changes.

achettyiitr: Done.

services/filemanager/azureBlobStoragemanager.go
achettyiitr: Done.

services/filemanager/azureBlobStoragemanager.go
achettyiitr: It's giving this error `Cannot call a pointer method on 'blobURLParts.URL()'`

achettyiitr: Leaving it as it is.

abhimanyubabbar (approved):
This PR looks good to me 👍🏼

Jayachand (approved):

services/filemanager/azureBlobStoragemanager.go
```diff
@@ -77,6 +75,20 @@ func (manager *AzureBlobStorageManager) getContainerURL() (azblob.ContainerURL,
...
 }
 
+func (manager *AzureBlobStorageManager) getCredentials() (azblob.Credential, error) {
+	if manager.Config.UseSASTokens {
+		return azblob.NewAnonymousCredential(), nil
+	}
+
+	accountName, accountKey := manager.Config.AccountName, manager.Config.AccountKey
+	if accountName == "" || accountKey == "" {
+		return nil, errors.New("either the AccountName or AccountKey is not correct")
```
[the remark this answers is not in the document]
achettyiitr: Done.

PR #2359, opened 2022-08-26
chore(warehouse): remove tokens from logs for redshift and snowflake

achettyiitr:
# Description

Tokens are getting exposed in logs. It shouldn't be. Removing those from logs for Redshift and Snowflake.

## Notion Ticket

https://www.notion.so/rudderstacks/Tokens-are-getting-exposed-in-logs-e5ecac1ab38b42ca99823bc5faf4333a

## Security

### DIFF
--- a/warehouse/redshift/redshift.go
+++ b/warehouse/redshift/redshift.go
@@ -332,6 +332,7 @@ func (rs *HandleT) loadTable(tableName string, tableSchemaInUpload, tableSchemaA
 	sanitisedSQLStmt, regexErr := misc.ReplaceMultiRegex(sqlStatement, map[string]string{
 		"ACCESS_KEY_ID '[^']*'":     "ACCESS_KEY_ID '***'",
 		"SECRET_ACCESS_KEY '[^']*'": "SECRET_ACCESS_KEY '***'",
+		"SESSION_TOKEN '[^']*'":     "SESSION_TOKEN '***'",
 	})
 	if regexErr == nil {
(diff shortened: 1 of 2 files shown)


warehouse/redshift/redshift.go
```diff
@@ -332,6 +332,7 @@ func (rs *HandleT) loadTable(tableName string, tableSchemaInUpload, tableSchemaA
 	sanitisedSQLStmt, regexErr := misc.ReplaceMultiRegex(sqlStatement, map[string]string{
 		"ACCESS_KEY_ID '[^']*'":     "ACCESS_KEY_ID '***'",
 		"SECRET_ACCESS_KEY '[^']*'": "SECRET_ACCESS_KEY '***'",
+		"SESSION_TOKEN '[^']*'":     "SESSION_TOKEN '***'",
```
Jayachand, line 335: session token will be active only for a while like an hour. That could be reason why it was omitted

achettyiitr: Let's remove it because it is sensitive. wdyt?

Jayachand: `When SESSION_TOKEN is specified, you must also use ACCESS_KEY_ID and SECRET_ACCESS_KEY to provide temporary access key credentials`
Not required. wdyt?

achettyiitr: Adding this in logs also doesn't make it productive. Since it is customer specific tokens which are sensitive information it's better to not log it.

abhimanyubabbar (approved):
LGTM 👍🏼

Jayachand (approved):

PR #2364, opened 2022-08-29
feat: isolate destID in router

cisse21:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -99,6 +99,8 @@ type HandleT struct {
 	jobsDBCommandTimeout                   time.Duration
 	jobdDBMaxRetries                       int
 	enableBatching                         bool
+	isolateDestID                          bool
+	queryFilters                           jobsdb.QueryFiltersT
 	transformer                            transformer.Transformer
 	configSubscriberLock                   sync.RWMutex
 	destinationsMap                        map[string]*routerutils.BatchDestinationT // destinationID -> destination
@@ -211,13 +213,14 @@ var (
 	jobQueryBatchSize, updateStatusBatchSize, noOfJobsPerChannel  int
 	failedEventsCacheSize                                         int
 	readSleep, minSleep, maxStatusUpdateWait, diagnosisTickerTime time.Duration
+	QueryFilters                                                  jobsdb.QueryFiltersT
 	minRetryBackoff, maxRetryBackoff, jobsBatchTimeout            time.Duration
 	pkgLogger                                                     logger.LoggerI
 	Diagnostics                                                   diagnostics.DiagnosticsI
 	fixedLoopSleep                                                time.Duration
 	toAbortDestinationIDs                                         string
-	QueryFilters                                                  jobsdb.QueryFiltersT
 	disableEgress                                                 bool
+	readPerDestination                                            bool
 )
 
 var jsonfast = jsoniter.ConfigCompatibleWithStandardLibrary
@@ -260,6 +263,8 @@ func loadConfig() {
 	config.RegisterDurationConfigVariable(48, &failedKeysExpire, true, time.Hour, "Router.failedKeysExpire")
 	config.RegisterDurationConfigVariable(24, &failedKeysCleanUpSleep, true, time.Hour, "Router.failedKeysCleanUpSleep")
 	failedKeysEnabled = config.GetBool("Router.failedKeysEnabled", false)
+
+	config.RegisterBoolConfigVariable(true, &readPerDestination, false, "Router.readPerDestination")
 }
 
 func (worker *workerT) trackStuckDelivery() chan struct{} {
@@ -1230,7 +1235,6 @@ func (rt *HandleT) initWorkers() {
 		rt.logger.Debugf("[%v Router] :: closing responseQ", rt.destName)
 		close(rt.failedEventsChan)
 		rt.logger.Debugf("[%v Router] :: closing failedEventsChan", rt.destName)
-
 		return err
 	})
 }
@@ -1729,7 +1733,6 @@ func (rt *HandleT) generatorLoop(ctx context.Context) {
 	rt.logger.Info("Generator started")
 
 	generatorStat := stats.NewTaggedStat("router_generator_loop", stats.TimerType, stats.Tags{"destType": rt.destName})
-	countStat := stats.NewTaggedStat("router_generator_events", stats.CountType, stats.Tags{"destType": rt.destName})
 
 	timeout := time.After(10 * time.Millisecond)
 	for {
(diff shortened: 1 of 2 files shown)


router/router.go
```diff
@@ -1760,7 +1770,56 @@ func (rt *HandleT) generatorLoop(ctx context.Context) {
...
+			Value:    rt.getValueForParameter(batchDest, key),
+			Optional: false,
+		}
+		parameterFilters = append(parameterFilters, parameterFilter)
+	}
+
+	return parameterFilters
+}
+
+func (rt *HandleT) constructQueryFilters(destination ...backendconfig.DestinationT) jobsdb.GetQueryParamsT {
```
atzoum, line 1796: for sanity, can we please split this method to two? e.g.

```
func (rt *HandleT) getQueryParams() jobsdb.GetQueryParamsT

func (rt *HandleT) getQueryParamsForDest(destination backendconfig.DestinationT) jobsdb.GetQueryParamsT
```

router/router.go
```diff
@@ -1760,7 +1770,56 @@ func (rt *HandleT) generatorLoop(ctx context.Context) {
...
+		parameterFilter := jobsdb.ParameterFilterT{
+			Name:     key,
+			Value:    rt.getValueForParameter(batchDest, key),
+			Optional: false,
+		}
+		parameterFilters = append(parameterFilters, parameterFilter)
+	}
+
+	return parameterFilters
+}
```
atzoum, line 1794: can't we drop `rt.queryFilters` & `getValueForParameter` altogether and simplify this further?
Even better we can skip this method too and construct `parameterFilters` directly in `getQueryParamsForDest` method that was proposed before.

```suggestion
func (rt *HandleT) getParameterFilters(destination backendconfig.DestinationT) []jobsdb.ParameterFilterT {
	return []jobsdb.ParameterFilterT{{
		Name:     "destination_id",
		Value:    destination.ID,
		Optional: false,
	}}
}
```

router/router.go
```diff
@@ -1742,9 +1753,8 @@ func (rt *HandleT) generatorLoop(ctx context.Context) {
 
 			generatorStat.Start()
 
-			processCount := rt.readAndProcess()
-
-			countStat.Count(processCount)
+			rt.cleanUp()
```
atzoum, line 1756: It is unsafe to perform this operation here, since we are introducing a(nother) new async stage through the `readProcessChan`. Please move it to its original place, as the first step of the `processRouterJobs` method which is the one that decides (synchronously) what enters the worker channel and what not.
If possible, don't extract the if block in its own method, simply add it in the method.

FYI, I am refactoring the whole event ordering algorithm in #2344

router/router.go
```diff
@@ -64,6 +64,11 @@ type HandleDestOAuthRespParamsT struct {
 	secret         json.RawMessage
 }
 
+type readProcessStruct struct {
+	combinedList  []*jobsdb.JobT
+	latenciesUsed map[string]float64
+}
+
```
atzoum, line 71: please avoid using the type in the name, e.g. I would call it 
```suggestion
type jobsWithLatencies struct {
	jobs  []*jobsdb.JobT
	latencies map[string]float64
}

```

router/router.go
```diff
@@ -79,6 +84,7 @@ type HandleT struct {
 	failedEventsListMutex sync.RWMutex
 	failedEventsList      *list.List
 	failedEventsChan      chan jobsdb.JobStatusT
+	readProcessChan       chan readProcessStruct
```
atzoum, line 87: we are done reading, just processing now :)
```suggestion
	processChan       chan jobsWithLatencies
```

router/router.go
```diff
@@ -260,6 +269,8 @@ func loadConfig() {
 	config.RegisterDurationConfigVariable(48, &failedKeysExpire, true, time.Hour, "Router.failedKeysExpire")
 	config.RegisterDurationConfigVariable(24, &failedKeysCleanUpSleep, true, time.Hour, "Router.failedKeysCleanUpSleep")
 	failedKeysEnabled = config.GetBool("Router.failedKeysEnabled", false)
+
+	config.RegisterBoolConfigVariable(true, &readPerDestination, false, "Router.readPerDestination")
```
atzoum, line 273: I will be frank, I don't like that we need to set one extra configuration option for the whole feature to work properly. This is fragile, time passes by and one might forget to turn this on, but enable isolation for some destination type nevertheless.

atzoum (commented):
I was expecting to see us following the separate router per `destinationID` approach, but using an additional parameterFilter for the `destinationID` instead of the previous approah where we were changing the value of the `custom_val` column.

With the current approach, we will have failures of any `userID` in any destination to block jobs for the same `userID` in all destinations.

router/router.go
cisse21: I don't like this either.. I put it as a hack for now, to just test the functionality. I intend to remove this at the end

cisse21:
> With the current approach, we will have failures of any userID in any destination to block jobs for the same userID in all destinations.

Yeah this issue might occur in a multi-tenant scenario and had crossed my mind. My idea to counter this was to append destionationId to the userID in order to mitigate this, thoughts on that?

atzoum:
> > With the current approach, we will have failures of any userID in any destination to block jobs for the same userID in all destinations.
> 
> Yeah this issue might occur in a multi-tenant scenario and had crossed my mind. My idea to counter this was to append destionationId to the userID in order to mitigate this, thoughts on that?

IMO, this issue might occur in any scenario, not multi-tenant, whenever a customer has more than one destinations of the same type and isolation is enabled. Going forward with the current approach, will introduce a lot of changes in router's internal design thus I consider it to be unsafe.

router/router.go
```diff
@@ -2080,11 +2123,19 @@ func (*HandleT) crashRecover() {
...
 	loadConfig()
+	setQueryFilters()
 	pkgLogger = logger.NewLogger().Child("router")
-	QueryFilters = jobsdb.QueryFiltersT{CustomVal: true}
 	Diagnostics = diagnostics.Diagnostics
 }
 
+func setQueryFilters() {
+	if readPerDestination {
+		QueryFilters = jobsdb.QueryFiltersT{CustomVal: true, ParameterFilters: []string{"destination_id"}}
```
atzoum, line 2133: After having a quick look to our cache implementation in jobsDB, my impression is that we are going to introduce the following problem:

- Router jobsDB's cache will be operating with parameter filters in its maps, i.e. when it creates cache keys it will include the destination id in them.
- For routers where destination isolation is on, we will be querying jobsDB using destination id as a parameter filter, so cache will work as expected, we will have cache hits.
- For all other routers where destination isolation is off, we will be querying jobsDB without using destination id as a parameter filter, thus the cache will never find a hit, i.e. it will be as if we won't be using any cache at all.

Based on the above, I believe that we cannot be selective on destination isolation, either we need to enable it for all destination types or for none.

cc @Sidddddarth to further comment on the cache's behaviour if necessary

atzoum: actually, it might be even more complex than the description above, nevertheless the point is that it is unsafe to use a cache using a different filter configuration than the actual queries...

PR #2365, opened 2022-08-29
chore: log improvements

lvrach:
# Description

A couple of minor improvements were introduced regarding logs.

**Motivation**: rudder-server was terminating without printing a helpful log. As well as some unhelpful want/err logs when shutting down.

## Notion Ticket

https://www.notion.so/rudderstacks/Minor-logs-improvements-5e9fbfc8dfa3495aa38ffa6e707585e0

## Security


config/backend-config/namespace_config.go
```diff
@@ -122,12 +122,14 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context, _ string) (ConfigT, e
...
-	backoffWithMaxRetry := backoff.WithMaxRetries(backoff.NewExponentialBackOff(), 3)
+	backoffWithMaxRetry := backoff.WithContext(backoff.WithMaxRetries(backoff.NewExponentialBackOff(), 3), ctx)
 	err := backoff.RetryNotify(operation, backoffWithMaxRetry, func(err error, t time.Duration) {
 		nc.Logger.Warnf("Failed to fetch config from API with error: %v, retrying after %v", err, t)
 	})
 	if err != nil {
-		nc.Logger.Errorf("Error sending request to the server: %v", err)
+		if ctx.Err() == nil {
+			nc.Logger.Errorf("Error sending request to the server: %v", err)
+		}
```
lvrach, line 132: I did this to avoid unrelated errors on context cancellation. Plus, it doesn't retry after ctx is cancelled.

main.go
```diff
@@ -252,51 +251,50 @@ func Run(ctx context.Context) int {
...
-			if !errors.Is(err, context.Canceled) {
-				pkgLogger.Errorf("Error in Admin server routine: %v", err)
-			}
+	g.Go(func() error {
+		if err := admin.StartServer(ctx); err != nil {
+			return fmt.Errorf("admin server routine: %w", err)
 		}
-		return err
+		return nil
 	})
```
lvrach, line 259: There is no need to check for `ctx.Err()`, since the ctx belongs to the errgroup.

1. `ctx` will be cancelled immediately when the first error returns
2. if ctx is canceled errgroup is not going to propagate any errors

main.go
```diff
@@ -252,51 +251,50 @@ func Run(ctx context.Context) int {
...
 	}
 
 	shutdownDone := make(chan struct{})
 	go func() {
 		err := g.Wait()
-		if err != nil && ctx.Err() == nil {
-			pkgLogger.Error(err)
+		if err != nil {
+			pkgLogger.Errorf("Terminal error: %v", err)
 		}
```
lvrach, line 295: The `ctx` here is the errgroup error, which is cancelled when an error is returned.

Essentially this condition `err != nil && ctx.Err() == nil` would never be true and thus nothing will be logged.

I find it easier to print a single "Terminal Error" as late as possible.

main.go
```diff
@@ -252,51 +251,50 @@ func Run(ctx context.Context) int {
...
 		}))
 	}
 
 	shutdownDone := make(chan struct{})
 	go func() {
 		err := g.Wait()
-		if err != nil && ctx.Err() == nil {
-			pkgLogger.Error(err)
+		if err != nil {
+			pkgLogger.Errorf("Terminal error: %v", err)
```
atzoum, line 294: I think that the only side effect is that we will log an error during graceful shudown, which is a bit misleading, e.g.

```
2022-08-30T16:31:25.711+0300	ERROR	rudder-server/main.go:294	Terminal error: profiler server routine: debug server: http: Server closed
2022-08-30T16:31:25.711+0300	INFO	rudder-server/main.go:297	Attempting to shutdown gracefully
2022-08-30T16:31:25.711+0300	INFO	rudder-server/main.go:308	Graceful terminal after 2.625274209s, with 154 go-routines
```

atzoum (approved):
looks like we need to chose between muting valuable errors and erroneously logging errors during normal shutdown

main.go
lvrach: True, I will make sure Server closed errors are not propagated if ctx is cancelled

utils/httputil/server.go
```diff
@@ -0,0 +1,50 @@
...
+func GracefulListenAndServe(ctx context.Context, server *http.Server, shutdownTimeout ...time.Duration) error {
+	return gracefulFunc(ctx, server, server.ListenAndServe, shutdownTimeout...)
+}
+
+func GracefulServe(ctx context.Context, server *http.Server, l net.Listener, shutdownTimeout ...time.Duration) error {
+	fn := func() error {
+		return server.Serve(l)
+	}
+	return gracefulFunc(ctx, server, fn, shutdownTimeout...)
+}
```
lvrach, line 23: The reason I've introduced those two methods is: 
1. We are using graceful shutdown methods in many places
2. Hard and repetitive to test shutdown functionality when it is not in isolation.
3. It is easy to get the shutdown wrong; some of the implementations in our code were correct, and some were not (funny, I think I have introduced them all :D )

**Example of incorrect implementation:**
```go
	go func() {
		<-ctx.Done()
		_ = srv.Shutdown(context.Background())
	}()

	if err := srv.ListenAndServe(); err != nil {
		return fmt.Errorf("debug server: %w", err)
	}
	return nil
```
> When `Shutdown` is called `ListenAndServe` returns immediately, always with an error.

Two problems: 
1. The shutdown is the one that blocks and the one we should be waiting for.
2. An error is returned when the context gets cancelled.

semi-correct:
```go
	g.Go(func() error {
		<-ctx.Done()
		return srv.Shutdown(context.Background())
	})
	g.Go(func() error {
		return srv.ListenAndServe()
	})

	return g.Wait()
```
Here we wait for both go routines to return, but the `ListenAndServe` is still there.

The new implementation fixes both issues along with providing tests.

services/validators/envValidator.go
```diff
@@ -153,7 +153,7 @@ func closeDBConnection(handle *sql.DB) {
 }
 
 func killDanglingDBConnections(db *sql.DB) {
-	rows, err := db.Query(`SELECT PID, QUERY_START, COALESCE(WAIT_EVENT_TYPE,''), COALESCE(WAIT_EVENT, ''), STATE, QUERY, PG_TERMINATE_BACKEND(PID)
+	rows, err := db.Query(`SELECT PID, QUERY_START, COALESCE(WAIT_EVENT_TYPE,''), COALESCE(WAIT_EVENT, ''), COALESCE(STATE, ''), QUERY, PG_TERMINATE_BACKEND(PID)
```
lvrach, line 156: I got an error in my test suite that `STATE` was null

utils/httputil/server.go
fracasula: Sounds good to me, although I'm not sure we need the `Graceful` part in the name. What do you think about just having it as `ListenAndServe` since we don't want an ungraceful one?

config/backend-config/backend-config.go
```diff
@@ -145,7 +145,12 @@ func (bc *backendConfigImpl) configUpdate(ctx context.Context, statConfigBackend
 
 	bc.curSourceJSONLock.Lock()
 	if !reflect.DeepEqual(bc.curSourceJSON, sourceJSON) {
-		pkgLogger.Infof("Workspace Config changed: %s", workspaces)
+		if len(workspaces) > 0 {
+			pkgLogger.Infof("Workspace Config changed: %d", len(workspaces))
+		} else {
+			pkgLogger.Infof("Workspace Config changed")
+		}
```
fracasula, line 152:
```suggestion
		pkgLogger.Infof("Workspace Config changed: %d", len(workspaces))
```

It's OK if it prints "Workspace Config changed: 0", right?

config/backend-config/namespace_config.go
fracasula, line 132: I think this change would be useful in the other backend configs as well.

utils/httputil/server.go
```diff
@@ -0,0 +1,50 @@
...
+
+func gracefulFunc(ctx context.Context, g grace, fn func() error, shutdownTimeout ...time.Duration) error {
```
fracasula, line 25: How come `shutdownTimeout` is a variadic parameter?

utils/httputil/server.go
```diff
@@ -0,0 +1,50 @@
...
+	go func() {
+		errCh <- fn()
+	}()
+	select {
+	case err := <-errCh:
+		return err
+	case <-ctx.Done():
+		var (
+			shutdownCtx context.Context
+			cancel      context.CancelFunc
```
fracasula, line 36: [optional] You don't need cancel nor to invoke it if a timeout is not specified:

```go
var shutdownCtx context.Context
if shutdownTimeout > 0 {
    var cancel func()
    shutdownCtx, cancel = context.WithTimeout(context.Background(), shutdownTimeout)
    defer cancel()
} else {
    shutdownTimeout = context.Background()
}
```

fracasula (approved):
Approved with some suggestions :+1:

utils/httputil/server.go
fracasula: If it's to make it optional, shouldn't you check whether it is `> 0` anyway? Seems bad to pass a 0 there, it would just be doomed to fail :sweat_smile:

app/cluster/state/etcd.go
```diff
@@ -112,7 +112,8 @@ func (manager *ETCDManager) init() error {
 			},
 		})
 		if err != nil {
-			manager.initErr = err
+			endpoints := strings.Join(manager.Config.Endpoints, `,`)
+			manager.initErr = fmt.Errorf("etcd client connect (%q): %w", endpoints, err)
```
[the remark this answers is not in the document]
lvrach: Interestingly etcd is eager to connect, I had as creation - but found that the errors will be related with connection and thus be more useful

config/backend-config/backend-config.go
lvrach: Yeah, but I found it confusing to get this long in single-tenant setups.

utils/httputil/server.go
lvrach: Yeah, make sense :)

utils/httputil/server.go
lvrach: The reason a made variadic is that in some cases is not possible to figure out what is the correct `shutdownTimeout`. 

You might have a lot of application-specific logic that dictates the timeout per handler, and figuring out the correct value on the top level is impossible.

In those cases, the responsibility is on the handler to use context and timeout properly, and a Shutdown should have no timeout.

An improper shutdown will report all running go routines - so it can be examined.

lvrach: I think I should make `0` zero helpful value here, and cancel the server instead of calling shutdown

config/backend-config/backend-config.go
fracasula: Why long? We're just printing the number of workspaces, it's not like it used to be anymore (list of workspaces with `%s`). In any case up to you, it's a super minor thing :+1:

utils/httputil/server.go
fracasula: Sounds good :+1:

config/backend-config/backend-config.go
lvrach: sorry: s/long/log

fracasula: Gotcha :+1:

main.go
lvrach: I have addressed this issue by introducing a new function that takes care of `http: Server closed`

PR #2368, opened 2022-08-30
feat: thread-safe job storage

saurav-malani:
# Description

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -116,7 +116,10 @@ type statTags struct {
 	StateFilters     []string
 }
 
-var getTimeNowFunc = time.Now
+var (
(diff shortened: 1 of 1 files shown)


saurav-malani:
There is a problem with using `advisory lock` here. Since, we want to make sure that max(table_i)<min(table_i+1). 
The scenario that won't be covered with `advisory lock` solution:-
If say table-1 is full but jobsDB-1 `addNewDSLoop` hasn't checked for it yet. So, `jobsDB1.Store()` will store in table-1.
But, at the same time say another instance of jobsDB, jobsdb2, checked for full and added new DS. So, there is a possibility that we had new DS but, Data wasn't inserted in it, which is an issue.

PR #2369, opened 2022-08-30
feat: isolate destID in router

cisse21:
# Description

- Spinning up a new router per destination ID for the assigned destType based on a flag
- Update the cache to rely on query rather than queryFilters

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=54e42282047f463fa8496961ae889531&pm=s)

## Security


atzoum:
First quick comment is that due to the amount of changes, this will require thorough review. 
Thus, we cannot consider it to be a safe quick-fix as it opens an additional topic, that of the overall design of the caching algorithm :) 
Please allow the time for a proper review, I personally consider this as 2 separate tasks:
- Caching algorithm refactoring (prerequisite)
- Router destination isolation

utils/misc/misc.go
```diff
@@ -1426,3 +1426,10 @@ func CopyStringMap(originalMap map[string]string) map[string]string {
 	}
 	return newMap
 }
+
+func GetRouterIdentifier(id, name string) string {
```
atzoum, line 1430: it would help to transform this into a `DestinationT` receiver function, e.g.

```
func (d DestinationT) GetRouterIdentifier() string {
	if config.GetBool(fmt.Sprintf("Router.%s.isolateDestID", d.DestinationDefinition.Name), false) {
		return d.ID
	}
	return d.DestinationDefinition.Name
}
```

app/apphandlers/setup.go
```diff
@@ -199,13 +199,13 @@ func monitorDestRouters(ctx context.Context, routerFactory *router.Factory, batc
...
 				} else {
-					_, ok := dstToRouter[destination.DestinationDefinition.Name]
+					_, ok := dstToRouter[misc.GetRouterIdentifier(destination.ID, destination.DestinationDefinition.Name)]
 					if !ok {
-						pkgLogger.Info("Starting a new Destination ", destination.DestinationDefinition.Name)
-						router := routerFactory.New(destination.DestinationDefinition)
-						router.Start()
-						cleanup = append(cleanup, router.Shutdown)
-						dstToRouter[destination.DestinationDefinition.Name] = router
+						pkgLogger.Infof("Starting a new Destination: %s", destination.DestinationDefinition.Name)
```
atzoum, line 204: shouldn't we use the identifier here too?

router/factory.go
```diff
@@ -17,11 +35,12 @@ type Factory struct {
 	RsourcesService  rsources.JobService
 }
 
-func (f *Factory) New(destinationDefinition backendconfig.DestinationDefinitionT) *HandleT {
+func (f *Factory) New(destination *backendconfig.DestinationT) (*HandleT, string) {
```
atzoum, line 38: We don't need to return the identifier as a second return value

router/manager/manager.go
```diff
@@ -93,8 +93,10 @@ loop:
 			}
 			sources := config.Data.(backendconfig.ConfigT)
 			enabledDestinations := make(map[string]bool)
-			for _, source := range sources.Sources {
-				for _, destination := range source.Destinations { // TODO skipcq: CRT-P0006
+			for i := range sources.Sources {
+				source := &sources.Sources[i]
+				for k := range source.Destinations { // TODO skipcq: CRT-P0006
```
atzoum, line 98:
```suggestion
				for k := range source.Destinations {
```

router/router.go
```diff
@@ -126,6 +127,7 @@ type HandleT struct {
 	oauth                                  oauth.Authorizer
 	transformerProxy                       bool
 	saveDestinationResponseOverride        bool
+	isolateDestID                          bool
```
atzoum, line 130:
```suggestion
```

router/router.go
```diff
@@ -2159,6 +2181,7 @@ func (rt *HandleT) Setup(backendConfig backendconfig.BackendConfig, jobsDB jobsd
 	config.RegisterBoolConfigVariable(false, &rt.transformerProxy, true, transformerProxyKeys...)
 	config.RegisterBoolConfigVariable(false, &rt.saveDestinationResponseOverride, true, saveDestinationResponseOverrideKeys...)
 	rt.allowAbortedUserJobsCountForProcessing = getRouterConfigInt("allowAbortedUserJobsCountForProcessing", destName, 1)
+	config.RegisterBoolConfigVariable(false, &rt.isolateDestID, false, []string{"Router." + rt.destName + "." + "isolateDestID", "Router.isolateDestID"}...)
```
atzoum, line 2184:
```suggestion
```

router/router.go
```diff
@@ -1760,6 +1760,32 @@ func (rt *HandleT) generatorLoop(ctx context.Context) {
...
+func (rt *HandleT) getQueryParams(pickUpCount int) jobsdb.GetQueryParamsT {
+	if rt.destinationId != rt.destName {
+		parameterFilters := getParameterFilters(rt.destinationId)
+		return jobsdb.GetQueryParamsT{
+			CustomValFilters:              []string{rt.destName},
+			ParameterFilters:              parameterFilters,
+			IgnoreCustomValFiltersInQuery: true,
+			PayloadSizeLimit:              rt.payloadLimit,
+			JobsLimit:                     pickUpCount,
+		}
```
atzoum, line 1780:
```suggestion
		return jobsdb.GetQueryParamsT{
			CustomValFilters: []string{rt.destName},
			ParameterFilters: []jobsdb.ParameterFilterT{{
				Name:     "destination_id",
				Value:    rt.destinationId,
				Optional: false,
			}},
			IgnoreCustomValFiltersInQuery: true,
			PayloadSizeLimit:              rt.payloadLimit,
			JobsLimit:                     pickUpCount,
		}
```

router/router.go
```diff
@@ -1760,6 +1760,32 @@ func (rt *HandleT) generatorLoop(ctx context.Context) {
 	}
 }
 
+func getParameterFilters(destinationID string) []jobsdb.ParameterFilterT {
+	return []jobsdb.ParameterFilterT{{
+		Name:     "destination_id",
+		Value:    destinationID,
+		Optional: false,
+	}}
+}
```
atzoum, line 1769:
```suggestion
```

router/router.go
```diff
@@ -2081,16 +2103,15 @@ func (*HandleT) crashRecover() {
...
 // Setup initializes this module
-func (rt *HandleT) Setup(backendConfig backendconfig.BackendConfig, jobsDB jobsdb.MultiTenantJobsDB, errorDB jobsdb.JobsDB, destinationDefinition backendconfig.DestinationDefinitionT, transientSources transientsource.Service, rsourcesService rsources.JobService) {
+func (rt *HandleT) Setup(backendConfig backendconfig.BackendConfig, jobsDB jobsdb.MultiTenantJobsDB, errorDB jobsdb.JobsDB, destinationConfig destinationConfig, transientSources transientsource.Service, rsourcesService rsources.JobService) {
 	rt.backendConfig = backendConfig
 	rt.workspaceSet = make(map[string]struct{})
 
-	destName := destinationDefinition.Name
+	destName := destinationConfig.Name
 	rt.logger = pkgLogger.Child(destName)
 	rt.logger.Info("Router started: ", destName)
```
atzoum, line 2116:
```suggestion
	rt.logger.Info("Router started: ", destinationConfig. DestinationID)
```

router/factory.go
```diff
@@ -5,8 +5,26 @@ import (
...
+
+func getRouterConfig(destination *backendconfig.DestinationT) destinationConfig {
+	var destConfig destinationConfig
+
+	destConfig.Name = destination.DestinationDefinition.Name
+	destConfig.DestinationID = misc.GetRouterIdentifier(destination.ID, destination.DestinationDefinition.Name)
+	destConfig.Config = destination.DestinationDefinition.Config
+	destConfig.ResponseRules = destination.DestinationDefinition.ResponseRules
+	return destConfig
+}
```
atzoum, line 26: We could live without this method, or the `destinationConfig` struct. All these can be extracted during `Setup` by passing the `backendconfig.DestinationT`

Sidddddarth:
```
/*
* If a query returns empty result for a specific dataset, we cache that so that
* future queries don't have to hit the DB.
* markClearEmptyResult() when mark=True marks dataset,customVal,state as empty.
* markClearEmptyResult() when mark=False clears a previous empty mark
 */
func (jd *HandleT) markClearEmptyResult(
```
This needs some change

jobsdb/jobsdb.go
```diff
@@ -2459,13 +2426,21 @@ func (jd *HandleT) markClearEmptyResult(ds dataSetT, workspace string, stateFilt
 			jd.dsEmptyResultCache[ds][workspace][cVal][pVal] = map[string]cacheEntry{}
 		}
```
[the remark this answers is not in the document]
Sidddddarth: or just an empty string

utils/misc/misc.go
cisse21: Do we really need this in misc? I wouldn't want to import backend-config into misc , so should I move this to backend-config file?

router/factory.go
cisse21: Yeah true but passing the whole destinationT struct didn't feel right if we are going to use only a few parameters in it

utils/misc/misc.go
atzoum: of course this would belong inside backend-config :)

router/factory.go
atzoum: `backendconfig.DestinationT` is a config struct, so IMO it is relevant and it would help keep things simpler, but I don't insist. Can we convert `destinationConfig` fields to lowercase please?

jobsdb/jobsdb.go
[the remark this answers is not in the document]
cisse21: Done

config/backend-config/backend-config.go
```diff
@@ -308,3 +308,10 @@ func getNotOKError(respBody []byte, statusCode int) error {
 	}
 	return fmt.Errorf("backend config request failed with %d%s", statusCode, errMsg)
 }
+
+func (d *DestinationT) GetRouterIdentifier() string {
+	if config.GetBool(fmt.Sprintf("Router.%s.isolateDestID", d.DestinationDefinition.Name), false) {
```
atzoum, line 313: Changing this configuration option during runtime shouldn't be supported as it could have undesirable effects with respect to the number of routers we would end up having active.
Thus, we should read this configuration option only once at startup.

app/apphandlers/setup.go
```diff
@@ -178,9 +180,10 @@ func monitorDestRouters(ctx context.Context, routerFactory *router.Factory, batc
 	routerFactory.RouterDB.DeleteExecuting()
 	batchRouterFactory.RouterDB.DeleteExecuting()
 
-	for config := range ch {
-		sources := config.Data.(backendconfig.ConfigT)
+	for workspaceConfig := range ch {
+		sources := workspaceConfig.Data.(backendconfig.ConfigT)
 		enabledDestinations := make(map[string]bool)
+		isolateFlagMap := make(map[string]bool)
```
atzoum, line 186: this should be defined outside the range loop, correct?

router/manager/manager.go
```diff
@@ -86,15 +88,18 @@ loop:
...
 			if !open {
 				pkgLogger.Infof("Router monitor stopped, Config Channel Closed")
 				break loop
 			}
-			sources := config.Data.(backendconfig.ConfigT)
+			sources := workspaceConfig.Data.(backendconfig.ConfigT)
 			enabledDestinations := make(map[string]bool)
-			for _, source := range sources.Sources {
-				for _, destination := range source.Destinations { // TODO skipcq: CRT-P0006
+			isolateFlagMap := make(map[string]bool)
```
atzoum, line 98: same here, shouldn't this be defined outside the loop?

app/apphandlers/setup.go
```diff
@@ -178,9 +180,10 @@ func monitorDestRouters(ctx context.Context, routerFactory *router.Factory, batc
 	routerFactory.RouterDB.DeleteExecuting()
 	batchRouterFactory.RouterDB.DeleteExecuting()
 
-	for config := range ch {
-		sources := config.Data.(backendconfig.ConfigT)
+	for workspaceConfig := range ch {
```
atzoum, line 183: **Question:** why do we need both `apphandlers.monitorDestRouters` and `LifecycleManager.monitorDestRouters`?

config/backend-config/backend-config.go
```diff
@@ -308,3 +308,10 @@ func getNotOKError(respBody []byte, statusCode int) error {
 	}
 	return fmt.Errorf("backend config request failed with %d%s", statusCode, errMsg)
 }
+
+func (d *DestinationT) GetRouterIdentifier(isolateFlagMap map[string]bool) string {
```
atzoum, line 312: instead of passing the map as an argument, can't we have this method maintain a thread-safe global var map?

app/apphandlers/setup.go
cisse21: Yeah I should do that

cisse21: It is outside the range loop right?

config/backend-config/backend-config.go
cisse21: I wanted to avoid introducing newer mutexes ... do you prefer having a global var over this?

atzoum: I admit I don't like global variables, but here for some reason I prefer it over the existing solution :) Global variable is used as a singleton config state and we only need to initialise these config variables once

cisse21: Modified it to use global var

config/backend-config/backend-config.go
```diff
@@ -308,3 +310,17 @@ func getNotOKError(respBody []byte, statusCode int) error {
 	}
 	return fmt.Errorf("backend config request failed with %d%s", statusCode, errMsg)
 }
+
+func (d *DestinationT) GetRouterIdentifier() string {
```
lvrach, line 314: Are we adding here so RouterIdentifier in backend-config so later we can get the config from control-plane ?

Minor suggestion, as it is more idiomatic in go to avoid the `Get`
```suggestion
func (d *DestinationT) GetRouterIdentifier() string {
```

router/utils/utils.go
```diff
@@ -123,3 +118,10 @@ func GetRudderAccountId(destination *backendconfig.DestinationT) string {
 	}
 	return ""
 }
+
+func PopulateIsolateMap(isolateMap map[string]bool, destination *backendconfig.DestinationT) {
+	_, ok := isolateMap[destination.DestinationDefinition.Name]
+	if !ok {
+		isolateMap[destination.DestinationDefinition.Name] = config.GetBool(fmt.Sprintf("Router.%s.isolateDestID", destination.DestinationDefinition.Name), false)
+	}
+}
```
lvrach, line 127: Why is this global? Can we make it part of a struct?

jobsdb/jobsdb.go
```diff
@@ -2144,60 +2118,49 @@ func (jd *HandleT) internalStoreWithRetryEachInTx(ctx context.Context, tx *sql.T
...
-		}
+	var vals []string
+	for _, key := range ParameterFilters {
+		val := gjson.GetBytes(job.Parameters, key).String()
+		vals = append(vals, fmt.Sprintf("%s##%s", key, val))
+	}
+
+	key := strings.Join(vals, "::")
+	if _, ok := CVPMap[job.WorkspaceId][job.CustomVal][key]; !ok {
+		CVPMap[job.WorkspaceId][job.CustomVal][key] = struct{}{}
```
lvrach, line 2141: Since we are making changes here, it is an opportunity to introduce https://pkg.go.dev/expvar#Map, so we can inspect the contents of the map easily by hitting the debug/expvar endpoint.

router/factory.go
lvrach: I saw this comment after adding my own about `RouterIdentifier`. I believe it is not the responsibility of backend-config to return the identifier used in our router logic. Yes, the destination as an abstraction is a good idea, but I would not put it inside `backendconfig.Destination`.

config/backend-config/backend-config.go
lvrach: I don't think a global variable is a good idea for caching configs. It makes testing with different parameters tricky and can be hard to debug. 

I believe these smells point to the fact that we need a different abstraction to for making this mapping

jobsdb/jobsdb.go
cisse21: Not sure how easy this would be right? We would have to do insertions multiple times, and right now we expose this through rudder-cli already

cisse21: Can't think of an easy solution where these are updated to exp var and do not impact the performance of jobsdb

router/manager/manager.go
```diff
@@ -56,11 +62,27 @@ func (r *LifecycleManager) Stop() {
...
+	}
+}
+
+func (r *LifecycleManager) RouterIdentifier(destinationID, destinationType string) string {
+	r.isolateRouterMapLock.Lock()
+	defer r.isolateRouterMapLock.Unlock()
+	_, ok := r.isolateRouterMap[destinationType]
+	if !ok {
+		r.isolateRouterMap[destinationType] = config.GetBool(fmt.Sprintf("Router.%s.isolateDestID", destinationType), false)
+	}
```
atzoum, line 80: minor 
```suggestion
	if _, ok := r.isolateRouterMap[destinationType]; !ok {
		r.isolateRouterMap[destinationType] = config.GetBool(fmt.Sprintf("Router.%s.isolateDestID", destinationType), false)
	}
```

router/factory.go
```diff
@@ -7,6 +7,23 @@ import (
...
+
+func getRouterConfig(destination *backendconfig.DestinationT, identifier string) destinationConfig {
+	var destConfig destinationConfig
+
+	destConfig.name = destination.DestinationDefinition.Name
+	destConfig.destinationID = identifier
+	destConfig.config = destination.DestinationDefinition.Config
+	destConfig.responseRules = destination.DestinationDefinition.ResponseRules
+	return destConfig
```
atzoum, line 24: minor
```suggestion
	return destinationConfig{
		name:          destination.DestinationDefinition.Name,
		destinationID: identifier,
		config:        destination.DestinationDefinition.Config,
		responseRules: destination.DestinationDefinition.ResponseRules,
	}
```

jobsdb/jobsdb.go
```diff
@@ -2458,6 +2420,12 @@ func (jd *HandleT) markClearEmptyResult(ds dataSetT, workspace string, stateFilt
 		jd.invalidCacheKeyStat.Increment()
 	}
 
+	for _, parameterFilter := range parameterFilters {
+		if !misc.ContainsString(ParameterFilters, parameterFilter.Name) {
+			return
+		}
+	}
```
atzoum, line 2427:
```suggestion
	// Skip the cache if a parameter filter is provided that does not belong to the caching key
	for _, parameterFilter := range parameterFilters {
		if !misc.ContainsString(ParameterFilters, parameterFilter.Name) {
			return
		}
	}
```

jobsdb/jobsdb.go
```diff
@@ -2164,60 +2138,49 @@ func (jd *HandleT) internalStoreWithRetryEachInTx(ctx context.Context, tx *sql.T
 	return
 }
 
+var ParameterFilters = []string{"destination_id"}
```
atzoum, line 2141:
```suggestion
// the parameter filters that will be part of the cache key
var CacheKeyParameterFilters = []string{"destination_id"}
```

jobsdb/jobsdb.go
atzoum: Isn't it more appropriate to test whether all parameter filters are provided or not?

E.g. if we were using 2 parameter filters for the caching key and the user provided only 1 of them as a parameter filter in the query, the cache should be skipped.

This test is now failing

```
t.Run("Test cache with two less parameter filters (destination_id & source_id)", func(t *testing.T) {
	previousParameterFilters := ParameterFilters
	ParameterFilters = []string{"destination_id", "source_id"}
	defer func() {
		ParameterFilters = previousParameterFilters
	}()
	jobsDB := NewForReadWrite("two_params_cache_query_less")
	require.NoError(t, jobsDB.Start())
	defer jobsDB.TearDown()

	destinationID := "destinationID"

	res, err := jobsDB.getUnprocessed(context.Background(), GetQueryParamsT{CustomValFilters: []string{customVal}, ParameterFilters: []ParameterFilterT{{Name: "destination_id", Value: destinationID}}, JobsLimit: 100})
	require.NoError(t, err)
	require.Equal(t, 0, len(res.Jobs), "jobsDB should report 0 unprocessed jobs when using destination_id as filter")

	require.NoError(t, jobsDB.Store(context.Background(), generateJobs(2, destinationID)))
	res, err = jobsDB.getUnprocessed(context.Background(), GetQueryParamsT{CustomValFilters: []string{customVal}, ParameterFilters: []ParameterFilterT{{Name: "destination_id", Value: destinationID}}, JobsLimit: 100})
	require.NoError(t, err)
	require.Equal(t, 2, len(res.Jobs), "jobsDB should report 2 unprocessed jobs when using destination_id as filter, after we added 2 jobs")
})
```

atzoum (approved):
one comment about cache hits and parameter filters, although our current setup and usage patterns are not really affected by it

lvrach (approved):

PR #2375, opened 2022-08-31
chore(router): include input payload in inconsistent transformer response error

atzoum:
# Description

Including the payload we send to the transformer for easier debugging

## Security

### DIFF
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -185,7 +185,7 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
 			invalidResponseError = fmt.Sprintf("Transformer returned invalid response: %s for input: %s", string(respData), string(rawJSON))
 		} else if len(in) != len(out) {
 			invalidResponseReason = "in out mismatch"
-			invalidResponseError = fmt.Sprintf("Transformer returned invalid output size: %d for input size: %d", len(out), len(in))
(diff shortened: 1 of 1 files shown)


lvrach:
Do we want to merge this in master or use it just for debugging?

atzoum:
> Do we want to merge this in master or use it just for debugging?

I guess for now it would suffice to use it for debugging, but maybe we should deploy a release that contains this log, for future problems. The current PR targets the `release/1.0.x` branch

lvrach:
> we should deploy a release that contains this log

Two problems with printing raw JSON payloads:
1. Things space to multiple lines; breaking parses and tools like loki and grep becomes harder.
2. There is a limit on the line size - each tool has one limit, and it can decide to either omit anything after the limit or push it to the next line.

For the reasons above @fracasula had created a tool that automatically upload big payloads to an s3 storage (with a lot of optimisations not impose performance problems)

fracasula:
> For the reasons above @fracasula had created a tool that automatically upload big payloads to an s3 storage (with a lot of optimisations not impose performance problems)

Another case is that multiple logs can happen at the same time so you could get your JSON output mixed with other entries making it increasingly difficult to get the whole output.

I had implemented a GCP uploader with a leaky bucket approach so that only one upload would take place at a time, any other call to `upload()` would be discarded. It was very useful for sampling/debugging (plus you don't risk as much to incur in big bills or storage limits on GCP due to the leaky bucket approach).

```go
select {
case upload<-file:
    doUpload()
default:
    // discard
}
```

PR #2377, opened 2022-08-31
feat: introduce features/settings client

lvrach:
# Description

This PR is meant to introduce functionality that will inform the control plane about the features of rudder-server. Control-plane will make decisions based on what rudder-server supports or not.

More about this can be read in [notion](https://www.notion.so/d5e7e041c02543e99e7bbcfcec63e9e4#cf2ee37a64084794a91e3d930b1715d5).

## Registry
I was puzzled about the best way to capture features of multiple components as they are added in rudder-server, **global vs injected**

One approach is to DI the functionality in every component, which seemed overkill for two reasons:
* This information is relatively static; it will be modified only when new features are added and won't change based on logic.
* The amount of code needed to change for DI to be adequately introduced would be extensive.

I end up using a global registry approach, for now, similar to what we do for configs, stats and logs. This should be revised in the future, especially for logs and stats DI will be handy for testing.

## Client
I wanted to decouple the client for Registry. Both for testing and extensibility. Registry in a simple in-memory store. While client just takes the Registry and sends the payload to the control plane.

Registry assumes all features will be added before a client call is made. We might want to revisit the decision if a feature should register based on configuration (feature flags). However, a system that waits until all registrations have occurred would require more complexity, so it is deferred until needed.

With the current implementations, features need to be registered in an `init` method.

With this approach, I wanted to avoid:
1.  Sending a message every time we register a feature.
2. Use a debounce logic, where we wait for an amount of time to gather and batch multiple features.
3. Use pre-registration and a lock/channel mechanism waiting for everything to be registered before we send it.

Some of the rejected approaches above might need to revisit as our requirements, code pollution, and usage patterns become more advanced.

## Identity

I have also introduced an abstraction called identity. I wanted this abstraction to be under control-plane-specific libraries and address the two different ways a rudder-server can be identified by control-plane **workspace** and **namespace**.

The differences between workspace and namespace are minor, with a slightly different endpoint and a different authorisation mechanism. Instead of having two different clients for workspace and namespace, I've delegated/abstracted the differences to the identity interface.

I also want to introduce identity in backend config and Oauth client, and remove the `AccessToken` method, which doesn't encapsulate the auth mechanism and the type of identity we are using. For now, I've marked as deprecated as replacing AccessToken would be beyond the scope of this task.

## Removed legacy multitenant code

Backend config had an additional mechanism two support multitenant without namespaces. This mechanism didn't support `/settings` as it was deprecated. I didn't want to perplex things anymore with this legacy code, so I've removed it.

## Notion Ticket

https://www.notion.so/rudderstacks/Reporting-features-to-control-plane-dd4976cb288544cf979f1d1e85913c1b

## Security


app/apphandlers/setup.go
```diff
@@ -107,6 +108,15 @@ func rudderCoreBaseSetup() {
 
 	processor.RegisterAdminHandlers(&readonlyProcErrorDB)
 	router.RegisterAdminHandlers(&readonlyRouterDB, &readonlyBatchRouterDB)
+
+	go func() {
+		backendconfig.DefaultBackendConfig.WaitForConfig(context.Background())
```
lvrach, line 113: Waiting for config to send is a workaround since we don't have the workspaceID, until config is fetched. 

There is a plan to remove this need in the `v2` API.

config/backend-config/namespace_config.go
```diff
@@ -10,12 +10,26 @@ import (
...
 	"github.com/rudderlabs/rudder-server/config"
+	"github.com/rudderlabs/rudder-server/services/controlplane/features"
+	"github.com/rudderlabs/rudder-server/services/controlplane/identity"
 	"github.com/rudderlabs/rudder-server/utils/logger"
 	"github.com/rudderlabs/rudder-server/utils/types"
 )
 
+var jsonfast = jsoniter.ConfigCompatibleWithStandardLibrary
+
+func init() {
```
atzoum, line 24: the package is `backendconfig`. Won't this be called as soon as the `backendconfig` package is imported, regardless if `namespaceConfig` is used or not?

Based on the above, will package init be enough for being able to properly register features during server startup, before the client tries to send them to the control plane?

atzoum: **Question:** after reading the document, what I understood is that the component might be any of the following
- `server`
- `sources`
- `transformer`
- `warehouse`

According to that, I wouldn't expect to see here any component other than `server` or `warehouse`. Why do we register a `backend-config` component?

services/controlplane/features/client.go
```diff
@@ -0,0 +1,123 @@
...
+	Name     string   `json:"name"`
+	Features []string `json:"features"`
+}
+
+func New(identity identity.Identifier, fns ...OptFn) *Client {
+	c := &Client{
+		client: &http.Client{
+			Timeout: DefaultTimeout,
+		},
+		url:      config.GetEnv("CONFIG_BACKEND_URL", "https://api.rudderlabs.com"),
```
atzoum, line 63: shouldn't the `url` be externally configurable as well?

services/controlplane/identity/identity.go
```diff
@@ -0,0 +1,61 @@
+package identity
+
+import "net/http"
+
+type Identifier interface {
```
atzoum, line 5: I know that names and concepts are subjective and always the hardest problem to solve, thus I am not insisting...

That being said, in my mental model what we are really abstracting here is control plane's different resource identifiers (urls). The authentication method always seems to be the same so far (basic with only a username).

So...
```
package controlplane

type HttpResource interface {
  URL() string
  BasicAuth() string
}

type workspaceResource struct {
  baseUrl string
  key string
  basicAuth string
}

func(nr *workspaceResource) URL() string {
 return path.Join(nr.baseUrl, "workspaces", nr.key)
}

type namespaceResource struct {
  baseUrl string
  key string
  basicAuth string
}

func(nr *namespaceResource) URL() string {
 return path.Join(nr.baseUrl, "namespaces", nr.key)
}
```

Finally, there seems to be some HTTP resource identifier inconsistency in control plane that needs to be resolved...

/data-plane/namespaces/:namespace/settings (plural without v1)
/data-plane/v1/namespaces/:namespace/config (plural with v1)
/data-plane/workspace/:workspaceId/settings (singular without v1)

config/backend-config/namespace_config.go
lvrach: This is something I want to get some feedback on from the control plane team as well. 

> Won't this be called as soon as the backendconfig package is imported, regardless if namespaceConfig is used or not?

Yes, it will. I mention some alternatives approach that will increase the complexity. My understanding so far is that for the first use cases, we just want to know if a feature is implemented in the current rudder-server version. Not if it is enabled or not.

I will ask for feedback again and revisit my approach if needed.

> Why do we register a backend-config component?

Regarding `backend-config, namespace`, I added it as an example, and It will not be used by the control plane. I think it is better to change this into something more useful.

services/controlplane/features/client.go
lvrach: But it is right? 
 https://github.com/rudderlabs/rudder-server/pull/2377/files/503743e24501d634a0f9b17e800d77f2af357282#diff-df67db07b56f23e5559aced20214af1134daefe6eec3bca096ef3807ed04e91cR25

atzoum: oops!

services/controlplane/identity/identity.go
lvrach: > abstracting here is control plane's different resource identifiers (urls). 

I get your point, but I had in my head a different kind of abstraction. 

Workspace and namespace are two ways the control plane can identify a rudder-server instance. In the feature, it might abstract things like auth scope and different auth mechanisms. The identifier will also be use for connecting using gRPC, for example. 

So the usage of Resource or Collection does not encapsulate what I have in mind for this.

I am happy to discuss this further, as I want Identifier to be used in communications with control-plane.

I hear the idea about `BasicAuth()`, although I don't see any problem with abstracting it from now.

lvrach: >Finally, there seems to be some HTTP resource identifier inconsistency in control plane that needs to be resolved...

I was waiting for a PR to merge in order to fix this

config/backend-config/namespace_config.go
atzoum: >My understanding so far is that for the first use cases, we just want to know if a feature is implemented in the current rudder-server version. Not if it is enabled or not.

In such a case, a statically initialized map would be sufficient and we don't really need a registry

```
features := map[string][]string {
	"server": ["featureA", "featureB"],
	"warehouse": ["featureC", "featureD"],
}
```

services/controlplane/identity/identity.go
atzoum: >I get your point, but I had in my head a different kind of abstraction.

As long as the abstraction is properly documented I am fine

config/backend-config/namespace_config.go
lvrach: You are right, as an initial approach this seems to be the most sensible way.

multi_tenant_test.go
```diff
@@ -41,6 +41,18 @@ func TestMultiTenant(t *testing.T) {
...
 
+func requireAuth(t *testing.T, secret string, hander http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		u, _, ok := r.BasicAuth()
+		require.True(t, ok, "Auth should be present")
+		require.Equalf(t, secret, u,
+			"Expected HTTP basic authentication to be %q, got %q instead",
+			secret, u)
+
+		hander(w, r)
```
fracasula, line 52:
```suggestion
func requireAuth(t *testing.T, secret string, handler http.HandlerFunc) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		u, _, ok := r.BasicAuth()
		require.True(t, ok, "Auth should be present")
		require.Equalf(t, secret, u,
			"Expected HTTP basic authentication to be %q, got %q instead",
			secret, u)

		handler(w, r)
```

fracasula (commented):
Preliminary review, I'll do another round later.

services/controlplane/features/client_test.go
```diff
@@ -0,0 +1,118 @@
...
+		err := c.Send(context.Background(), features.PerComponent{
+			"test": {"feature1", "feature2"},
+		})
+		require.Error(t, err)
+
+		require.Equalf(t, int64(4), atomic.LoadInt64(&count), "retry %d times", features.MaxRetries)
+	})
+
+	t.Run("timeout", func(t *testing.T) {
+		t.Skip("test is too slow")
```
fracasula, line 97: Why having a test if we skip it? Also, given that you set the timeout to one millisecond, why is the test slow? With the backoff included we are below 3secs. Additionally we could inject the retry strategy in the `*Client` and have it default to the backoff one. WDYT?

fracasula (commented):
2nd round.

services/controlplane/features/client_test.go
fracasula: If you do something like [this](https://github.com/rudderlabs/rudder-server/commit/c67635452140ffb0391c9c5a47f97a72ea9da7f8) the test goes down to 0.01s of execution plus we get a new feature :+1:

services/controlplane/features/client.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	return hostname
+}
+
+func New(identity identity.Identifier, fns ...OptFn) *Client {
+	c := &Client{
+		client: &http.Client{
+			Timeout: DefaultTimeout,
+		},
+		url:      config.GetEnv("CONFIG_BACKEND_URL", "https://api.rudderlabs.com"),
```
fracasula, line 74: Can we please avoid spreading the config package everywhere and have dependencies properly injected? We're duplicating also the default (i.e. `https://api.rudderlabs.com`) each time. It would be great if we could have `url` as a parameter since it's not optional and remove the `WithURL` and return an error with the provided `url` is empty.

services/controlplane/features/client.go
```diff
@@ -0,0 +1,147 @@
...
+	for _, fn := range fns {
+		fn(c)
+	}
+
+	return c
+}
+
+type PerComponent map[string][]string
+
+func (c *Client) Send(ctx context.Context, components PerComponent) error {
```
fracasula, line 88: Not sure it's worth having a type for `components` here. Not strong on it but consider that we won't be able to decouple much when using this stuff (e.g. possibility to create local interface without depending on this package):

```go
type client interface {
     Send(ctx context.Context, components features.PerComponent) error
}
```

:point_up: still depends on `features` versus...

```go
type client interface {
     Send(ctx context.Context, components map[string][]string) error
}
```

services/controlplane/features/client.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	for name, features := range components {
+		payload.Components = append(payload.Components, component{
+			Name:     name,
+			Features: features,
+		})
+	}
+
+	// sort by name, so that the order of components is always the same in every call
```
fracasula, line 111: How come is the order important? :thinking:

services/controlplane/features/client.go
```diff
@@ -0,0 +1,147 @@
...
+
+		req.SetBasicAuth(c.identity.BasicAuth())
+
+		resp, err := c.client.Do(req)
+		if err != nil {
+			return err
+		}
+		defer resp.Body.Close()
+
+		b, err := io.ReadAll(resp.Body)
```
fracasula, line 137:
```suggestion
		b, err := io.ReadAll(resp.Body)
		_ = resp.Body.Close()
```

Optional of course but I wanted to share to gather your thoughts given that in this case it seems very easy to avoid deferring.

```go
package features

import (
	"testing"
)

func doNoDefer(t *int) {
	func() {
		*t++
	}()
}

func doDefer(t *int) {
	defer func() {
		*t++
	}()
}

func BenchmarkDefer(b *testing.B) {
	b.Run("yes", func(b *testing.B) {
		t := 0
		for i := 0; i < b.N; i++ {
			doDefer(&t)
		}
	})

	b.Run("no", func(b *testing.B) {
		t := 0
		for i := 0; i < b.N; i++ {
			doNoDefer(&t)
		}
	})
}
```

With go 1.19 it gives me:

```
goos: linux
goarch: amd64
pkg: github.com/rudderlabs/rudder-server/services/controlplane/features
cpu: 12th Gen Intel(R) Core(TM) i9-12900HK
BenchmarkDefer
BenchmarkDefer/yes
BenchmarkDefer/yes-20         	588573512	         2.031 ns/op
BenchmarkDefer/no
BenchmarkDefer/no-20          	1000000000	         0.2673 ns/op
PASS
```

Perhaps I'm not measuring it appropriately but it seems like some overhead is there :thinking: 

(I'm not saying do not use `defer`, just think twice if it's not really needed. If you have multiple `if/else` with returns then paying the overhead makes sense for the sake of keeping the complexity under control, in any case again totally optional).

main.go
```diff
@@ -275,6 +277,13 @@ func Run(ctx context.Context) int {
 			}
 			return nil
 		}))
+		go func() {
```
fracasula, line 280: Are you sure we don't need any orchestration here? Why not wait for this routine to return and terminate gracefully?

warehouse/warehouse.go
```diff
@@ -1811,6 +1813,15 @@ func Start(ctx context.Context, app app.Interface) error {
 		}))
 	}
 
+	// Report warehouse features
+	go func() {
+		backendconfig.DefaultBackendConfig.WaitForConfig(context.Background())
+		err := features.New(backendconfig.DefaultBackendConfig.Identity()).Send(context.Background(), info.WarehouseFeatures)
```
fracasula, line 1819: Same comments as the routine in `main.go`. If it makes sense to orchestrate and use the context from above.

services/controlplane/features/client_test.go
lvrach: Good point; I've overridden the retries to one. It solves the problem without introducing coupling. retries are tested with a different test either way.

utils/httputil/client.go
```diff
@@ -0,0 +1,24 @@
+package httputil
+
+import "net/http"
+
+// RetriableStatus returns true if the HTTP status code should be retried.
+// 4xx codes are considered non-retiable except for 408 - request timeout and 429 - too many requests.
```
fracasula, line 6:
```suggestion
// 4xx codes are considered non-retriable except for 408 - request timeout and 429 - too many requests.
```

services/controlplane/features/client.go
fracasula: It looks like this one was removed. Resolving :+1:

services/controlplane/features/client.go
```diff
@@ -0,0 +1,147 @@
...
+			Features: features,
+		})
+	}
+
+	// sort by name, so that the order of components is always the same in every call
+	sort.Slice(payload.Components, func(i, j int) bool { return payload.Components[i].Name < payload.Components[j].Name })
+
+	body, err := json.Marshal(payload)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
fracasula: Ping :stuck_out_tongue:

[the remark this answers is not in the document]
fracasula: [optional but nice] The same would apply to the other errors in this package if you think it can be useful.

services/controlplane/features/client.go
lvrach: Interesting, I would not expect this to be that big.

However, the performance overhead of using defer in this case should be minor. Indeed, in this case, it can be easily avoided, but I like it as a consistent pattern across the code. I expect a defer close as soon as a successful request has returned.

For me, the readability/predictability ways the performance when IO is involved.

fracasula: Yes I agree with you, having the defer makes it easier to follow if you're coming from a Go background.

It is a micro optimization on a piece of code that is not really giving us problems so like yourself I think that readability should take precedence for the sake of code maintenability.

I wanted to share this anyway because sometimes I avoid using `defer` myself for this reason (when the code is quite simple and doesn't have returns) plus sharing is caring so :smile:

config/backend-config/backend-config.go
```diff
@@ -46,16 +47,19 @@ var (
...
 	GetWorkspaceIDForWriteKey(string) string
 	GetWorkspaceIDForSourceID(string) string
 	GetWorkspaceLibrariesForWorkspaceID(string) LibrariesT
+	Identity() identity.Identifier
 }
 
 type BackendConfig interface {
 	workspaceConfig
 	WaitForConfig(ctx context.Context)
+	Identity() identity.Identifier
```
atzoum, line 62: don't need it here since `BackendConfig` embeds `workspaceConfig`, right?
```suggestion
```

warehouse/warehouse.go
```diff
@@ -1811,6 +1813,24 @@ func Start(ctx context.Context, app app.Interface) error {
 		}))
 	}
 
+	// Report warehouse features
+	g.Go(func() error {
+		backendconfig.DefaultBackendConfig.WaitForConfig(ctx)
+
+		c := features.New(
```
atzoum, line 1820: it took me a while to understand that we are creating a new features client here, thus I propose to rename `features.New` to `features.NewClient`

info/features.go
```diff
@@ -0,0 +1,22 @@
...
+type Component struct {
+	Name     string
+	Features []string
+}
+
+// ServerComponent describes server component info, like implemented features.
+var ServerComponent = Component{
+	Name: "server",
+	Features: []string{
+		"features",
```
atzoum, line 12: how and when is this going to be populated?

utils/httputil/client.go
```diff
@@ -0,0 +1,24 @@
...
+
+import "net/http"
+
+// RetriableStatus returns true if the HTTP status code should be retried.
+// 4xx codes are considered non-retriable except for 408 - request timeout and 429 - too many requests.
+func RetriableStatus(statusCode int) bool {
+	// 2xx, 3xx
+	if statusCode < 400 {
+		return true
+	}
```
atzoum, line 11: why would anyone want to retry a 2xx or 3xx?

info/build.go
```diff
@@ -0,0 +1,6 @@
+package info
+
+var (
+	Version                                                                  = "Not an official release. Get the latest release from the github repo."
+	Major, Minor, Commit, BuildDate, BuiltBy, GitURL, Patch, EnterpriseToken string
+)
```
atzoum, line 6: nobody seems to be using these...

lvrach: I will remove the file and introduce it in another PR.

utils/httputil/client.go
lvrach: Why not :P Ok, initially, this method has called NonRetriableStatus. By default, everything should be retried except 4xx (with two exceptions). Retries happen only after an error, so if your logic determines there is an error on a 2xx or 3xx you should be able to retry.

I understand that it creates confusion, so I will make them non-retriable as well.

info/features.go
lvrach: They are going to be populated manually for now. 

When a new feature that the control plane needs to be aware of is developed, it will be added here.

Following your comment here https://github.com/rudderlabs/rudder-server/pull/2377#discussion_r964753658, I found this approach to be the most straightforward and cleanest for now.

The dynamic enabled/disable of features based on config is not supported yet, but it can be easily implemented if required by something like this:

```go
func ServerComponentFromConfig(conf config.Config) Component {

  comp := Component{
	Name: "server",
	Features: []string{},
  }
  if config.IsSet("feature.enabled") {
  	comp.Features = append(comp.Features, "features")
  }
  
  return comp
}
```

atzoum (approved):

PR #2379, opened 2022-08-31
feat(router): avoid worker starvation during job pickup

atzoum:
# Description

Introducing `jobiterator` package in `router`, responsible for performing additional queries against jobsDB in order to fetch more jobs in case that some of the initially picked-up jobs get discarded (e.g. due to job ordering barrier, throttling or backoff). The following (configurable) limitations apply:

1. Iterator will stop querying after `Router.jobIterator.maxQueries` (default: 10). Setting this to 1, effectively disables the feature.
2. Iterator will stop querying if the percentage of discarded jobs for the running query is less than or equal to `Router.jobIterator.discardedPercentageTolerance` (default: 10%)

**Note:** an additional limitation applies when `JobsDB.fairPickup` is enabled: A maximum number of `Router.maxDSQuery` datasets can be queried at any time (default: 10).

Additionally, the following changes/improvements have been introduced:
1. Include `destinationID` in the job order key along with `userID`.
4. Remove `latenciesUsed` & `timeGained` from `GetRouterPickupJobs`.
5. Remove `JobsDB.useJoinForUnprocessed`.
6. Move server startup logic away from package `main`, to the `runner` package, so that we can move integration test files in the packages that they belong to.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=b8054d57db2048db8e3e52674f86b784&pm=s)

## Security


jobsdb/jobsdb.go
```diff
@@ -704,7 +705,6 @@ func loadConfig() {
 	config.RegisterDurationConfigVariable(5, &refreshDSListLoopSleepDuration, true, time.Second, []string{"JobsDB.refreshDSListLoopSleepDuration", "JobsDB.refreshDSListLoopSleepDurationInS"}...)
 	config.RegisterDurationConfigVariable(5, &backupCheckSleepDuration, true, time.Second, []string{"JobsDB.backupCheckSleepDuration", "JobsDB.backupCheckSleepDurationIns"}...)
 	config.RegisterDurationConfigVariable(5, &cacheExpiration, true, time.Minute, []string{"JobsDB.cacheExpiration"}...)
-	useJoinForUnprocessed = config.GetBool("JobsDB.useJoinForUnprocessed", true)
```
atzoum, line 707: removing this config option and always use a join for unprocessed

jobsdb/unionQueryLegacy.go
```diff
@@ -1,53 +1,84 @@
...
+	"context"
+	"fmt"
+)
 
 type MultiTenantLegacy struct {
 	*HandleT
 }
 
-func (mj *MultiTenantLegacy) GetAllJobs(ctx context.Context, workspaceCount map[string]int, params GetQueryParamsT, _ int) ([]*JobT, error) { // skipcq: CRT-P0003
+type legacyMoreToken struct {
```
atzoum, line 12: Explain: for legacy query we need to keep track of each subQuery's maximum job ID

jobsdb/unionQuery.go
```diff
@@ -18,8 +18,26 @@ type MultiTenantHandleT struct {
...
+type workspacePickup struct {
+	Limit        int
+	PayloadLimit int64
+	AfterJobID   *int64
+}
+
+type (
+	MoreToken interface{}
+	moreToken struct {
+		afterJobIDs map[string]*int64
```
atzoum, line 30: **Explain:** for fair pickup algorithm we need to keep track of all the max job IDs per workspace

jobsdb/jobsdb.go
```diff
@@ -2484,41 +2483,44 @@ func (jd *HandleT) getProcessedJobsDS(ctx context.Context, ds dataSetT, params G
...
-	queryStat.Start()
-	defer queryStat.End()
+	start := time.Now()
+	defer jd.getTimerStat("processed_ds_time", &tags).Since(start)
 
-	// We don't reset this in case of error for now, as any error in this function causes panic
-	jd.markClearEmptyResult(ds, allWorkspaces, stateFilters, customValFilters, parameterFilters, willTryToSet, nil)
-
-	var stateQuery, customValQuery, limitQuery, sourceQuery string
+	skipCacheResult := params.AfterJobID != nil
```
atzoum, line 2489: Explain: we don't update the cache if the query contains an `AfterJobID` parameter

main.go
```diff
@@ -2,361 +2,34 @@ package main
...
-	alert.Init()
-	multitenant.Init()
-	oauth.Init()
-	Init()
-}
-
 func main() {
 	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
-	exitCode := Run(ctx)
+	r := runner.New(runner.ReleaseInfo{
```
atzoum, line 21: Explain: moved running logic away from `main`, to the `runner` package, so that we can move integration test files in the packages that they belong.

router/router_test.go
```diff
@@ -1471,139 +1419,3 @@ func Benchmark_FASTJSON_MARSHAL(b *testing.B) {
 		_ = string(val)
 	}
 }
-
-func Test_RouterDestIsolation(t *testing.T) {
```
atzoum, line 1475: **Explain:** moved to its own file

chandumlg:
> Include destinationID in the job order key along with userID.
👍

router/internal/jobiterator/jobiterator.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// try to fetch some more jobs
+	var err error
+	var jobsLimit int
+	for _, limit := range ji.state.pickupMap {
+		if limit > 0 {
+			jobsLimit += limit
+		}
+	}
+	ji.params.JobsLimit = jobsLimit
```
cisse21, line 120: Can we use `ji.params.JobsLimit` here and skip the for loop above?
```suggestion

```

router/router.go
```diff
@@ -362,9 +363,8 @@ func (worker *workerT) workerProcess() {
 			userID := job.UserID
 
 			var parameters JobParametersT
-			err := json.Unmarshal(job.Parameters, &parameters)
-			if err != nil {
-				worker.rt.logger.Error("Unmarshal of job parameters failed. ", string(job.Parameters))
+			if err := json.Unmarshal(job.Parameters, &parameters); err != nil {
+				panic(fmt.Errorf("unmarshalling of job parameters failed (%s): %w", string(job.Parameters), err))
```
chandumlg, line 367: Instead of panicking, let's log relevant information, raise alert and abort the job?

router/internal/jobiterator/jobiterator.go
```diff
@@ -0,0 +1,166 @@
...
+
+// New returns a new job iterator
+//   - pickupMap: a map of workspaceID to number of jobs to be fetched from that workspace
+//   - params: jobsDB query parameters
+//   - getJobsFn: the function to fetch jobs from jobsDB
+//   - opts: optional iterator options
+func New(pickupMap map[string]int, params jobsdb.GetQueryParamsT, getJobsFn func(context.Context, map[string]int, jobsdb.GetQueryParamsT, jobsdb.MoreToken) (*jobsdb.GetAllJobsResult, error), opts ...IteratorOptFn) *Iterator {
+	ji := &Iterator{
+		params:                       params,
+		maxQueries:                   100,
```
chandumlg, line 73: Any reason why this default is more than, what router sets by default?

router/internal/jobiterator/jobiterator.go
```diff
@@ -0,0 +1,166 @@
...
+	ji.state.pickupMap = pickupMap
+	ji.state.previousJob = map[string]*jobsdb.JobT{}
+	for _, opt := range opts {
+		opt(ji)
+	}
+	return ji
+}
+
+// HasNext returns true when there are more jobs to be fetched by Next(), false otherwise.
+func (ji *Iterator) HasNext() bool {
```
chandumlg, line 87: [optional] Let's add a stat on how many queries were run by iterator in one generator loop. Just for information.

Not sure how easy it will be in visualising on grafana.

router/router.go
atzoum: Failing to unmarshal bytes stored in a jsonb column is highly unlikely, practically impossible. Thus I don't think we should care too much about it, to let it complicate router's logic even further.
It is just that logging an error instead of panicking in such a fundamental issue seemed to be weird in the first place.

router/internal/jobiterator/jobiterator.go
atzoum: Just picked a value high enough not to be limiting by default.

router/internal/jobiterator/jobiterator.go
atzoum: good idea, done

Sidddddarth:
LGTM
Only reservation I have is resetting the pickupMap in the `iterator.HasNext()` between successive iterations.
Because this way we'd only be picking up `discarded` number of jobs in further iterations.
In case we don't find as many jobs in the first iteration and several more are stored(enough to fill the pickupmap counts), we'd end up not picking them up soon enough.

Sidddddarth (approved):

atzoum:
> Because this way we'd only be picking up `discarded` number of jobs in further iterations.

This behaviour is intentional. The iterator only compensates for discarded jobs to avoid ending up with an unproductive pickup loop.

> In case we don't find as many jobs in the first iteration and several more are stored(enough to fill the pickupmap counts), we'd end up not picking them up soon enough.

After an iterator completes, the next pickup loop will create a fresh iterator with fresh limits. The time between two pickup loops should remain small enough and shouldn't cause any significant delays in picking up newly arrived jobs.

Sidddddarth:
One other thought is that since we query by `job_id` as well now, we could use 
```
type dataSetRangeT struct {
	minJobID  int64
	maxJobID  int64
	startTime int64
	endTime   int64
	ds        dataSetT
}
```
to check whether or not to query a DS(except the last DS).

atzoum:
> One other thought is that since we query by `job_id` as well now, we could use
> 
> ```
> type dataSetRangeT struct {
> 	minJobID  int64
> 	maxJobID  int64
> 	startTime int64
> 	endTime   int64
> 	ds        dataSetT
> }
> ```
> 
> to check whether or not to query a DS(except the last DS).

Yes this would be an option, however, since job_id is an indexed column, the sql query itself should be sufficiently quick and lightweight, and should handle equally efficiently the absence of results due to the job_id condition.

Sidddddarth:
We'd also want to have a memory limit over how the `jobIterator` fetches jobs right..?
Because every iteration it could fetch many jobs and `discard` them and all these jobs would be in memory till they get garbage collected..

atzoum:
> till they get garbage collected..

Since discarded jobs will be eligible for garbage collection, we can rely on that, if memory resources become scarce, they will get garbage collected.

router/router.go
```diff
@@ -1793,50 +1778,40 @@ func (rt *HandleT) readAndProcess() int {
...
 		return 0
-	}
-
-	sort.Slice(combinedList, func(i, j int) bool {
-		return combinedList[i].JobID < combinedList[j].JobID
-	})
-
-	if len(combinedList) > 0 {
+	} else {
 		rt.logger.Debugf("[%v Router] :: router is enabled", rt.destName)
```
chandumlg, line 1810: This log doesn't make sense. We can remove it.

chandumlg (approved):

chandumlg:
Awesome work! 🎉

PR #2382, opened 2022-09-01
feat(warehouse): detailed warehouse reporting

dparveez-ruddr:
# Description

Change reporting behaviour in warehouse.

- Will populate missing values like event name, source job details, etc. wherever possible.
- Will send success and failure reports per event name.

## Notion Ticket

https://www.notion.so/rudderstacks/Reporting-populate-empty-values-in-wh-reports-handle-negative-counts-e2de2a63c5b04445b8aae0ef752b5984

## Security

### DIFF
--- a/warehouse/upload.go
+++ b/warehouse/upload.go
@@ -547,7 +547,7 @@ func (job *UploadJobT) run() (err error) {
 
 		uploadStatusOpts := UploadStatusOpts{Status: newStatus}
 		if newStatus == ExportedData {
-			reportingMetrics := make([]types.PUReportedMetric, len(loadFilesTableEventCountMap))
+			reportingMetrics := make([]types.PUReportedMetric, 0, len(loadFilesTableEventCountMap))
 
 			for table, eventCount := range loadFilesTableEventCountMap {
 				reportingMetric := types.PUReportedMetric{
@@ -1251,10 +1251,10 @@ func (job *UploadJobT) setUploadStatus(statusOpts UploadStatusOpts) (err error) 
 		}
 
 		if config.GetBool("Reporting.enabled", types.DEFAULT_REPORTING_ENABLED) {
-			metricsArr := make([]*types.PUReportedMetric, len(statusOpts.ReportingMetrics))
+			metricsArr := make([]*types.PUReportedMetric, 0, len(statusOpts.ReportingMetrics))
 
-			for _, reportingMetric := range statusOpts.ReportingMetrics {
-				metricsArr = append(metricsArr, &reportingMetric)
+			for i := range statusOpts.ReportingMetrics {
+				metricsArr = append(metricsArr, &statusOpts.ReportingMetrics[i])
 			}
 
 			application.Features().Reporting.GetReportingInstance().Report(metricsArr, txn)
@@ -1477,7 +1477,7 @@ func (job *UploadJobT) setUploadError(statusError error, state string) (string, 
 		return "", fmt.Errorf("unable to change upload columns: %w", err)
 	}
 
-	failedTableMap, _ := job.getTablesToSkip()
+	_, currentJobSucceededTables := job.getTablesToSkip()
 	_, loadFilesTableEventCountMap, err := job.getLoadFilesTablePresenceAndEventCountMaps()
 
 	if err != nil {
@@ -1485,7 +1485,13 @@ func (job *UploadJobT) setUploadError(statusError error, state string) (string, 
 		loadFilesTableEventCountMap = make(map[tableNameT]int)
 	}
 
-	for tableName, _ := range failedTableMap {
+	for tableName, _ := range job.upload.UploadSchema {
+		_, ok := currentJobSucceededTables[string(tableName)]
+
+		if ok {
+			continue
+		}
+
 		failCount := int64(loadFilesTableEventCountMap[tableNameT(tableName)])
 
 		reportingStatus := jobsdb.Failed.State
--- a/warehouse/upload.go
+++ b/warehouse/upload.go
@@ -454,13 +454,12 @@ func (job *UploadJobT) run() (err error) {
 
 		case ExportedData:
 			newStatus = nextUploadState.failed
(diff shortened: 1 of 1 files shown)


warehouse/upload.go
```diff
@@ -1239,7 +1255,13 @@ func (job *UploadJobT) setUploadStatus(statusOpts UploadStatusOpts) (err error)
 		}
 
 		if config.GetBool("Reporting.enabled", types.DEFAULT_REPORTING_ENABLED) {
-			application.Features().Reporting.GetReportingInstance().Report([]*types.PUReportedMetric{&statusOpts.ReportingMetric}, txn)
+			metricsArr := make([]*types.PUReportedMetric, 0, len(statusOpts.ReportingMetrics))
+
+			for i := range statusOpts.ReportingMetrics {
+				metricsArr = append(metricsArr, &statusOpts.ReportingMetrics[i])
```
abhimanyubabbar, line 1261: You are appending to the array `metricsArr` which already has a default length specified when created. You will be unecessary holding blank values in it. Can you double check if length of the two arrays is similar ?

abhimanyubabbar (changes requested):

warehouse/upload.go
```diff
@@ -545,29 +547,39 @@ func (job *UploadJobT) run() (err error) {
...
+					},
+					PUDetails: types.PUDetails{
+						InPU:       types.BATCH_ROUTER,
+						PU:         types.WAREHOUSE,
+						TerminalPU: true,
+					},
+					StatusDetail: &types.StatusDetail{
+						Status:     jobsdb.Succeeded.State,
+						StatusCode: 200,
+						// Count:       job.getTotalRowsInStagingFiles(),
```
achettyiitr, line 572: If it is not required. Can we remove this?

warehouse/upload.go
```diff
@@ -1459,65 +1481,58 @@ func (job *UploadJobT) setUploadError(statusError error, state string) (string,
...
-		reportingMetrics = append(reportingMetrics, &types.PUReportedMetric{
+	job.resetTableUploadStatuses()
+	_, currentJobSucceededTables := job.getTablesToSkip()
+	_, loadFilesTableEventCountMap, err := job.getLoadFilesTablePresenceAndEventCountMaps()
+	if err != nil {
+		loadFilesTableEventCountMap = make(map[tableNameT]int)
+	}
+
+	for tableName := range job.upload.UploadSchema {
+		_, ok := currentJobSucceededTables[string(tableName)]
```
achettyiitr, line 1492: UploadSchema is basically map[string]map[string]string. Let's not wrap it again with string. Not required.

warehouse/upload.go
```diff
@@ -1459,65 +1481,58 @@ func (job *UploadJobT) setUploadError(statusError error, state string) (string,
...
-			Count:          failCount,
-			SampleEvent:    []byte("{}"),
-			SampleResponse: string(serializedErr),
-		},
-	}}
-	if outputCount > 0 {
-		reportingMetrics = append(reportingMetrics, &types.PUReportedMetric{
+	job.resetTableUploadStatuses()
+	_, currentJobSucceededTables := job.getTablesToSkip()
+	_, loadFilesTableEventCountMap, err := job.getLoadFilesTablePresenceAndEventCountMaps()
```
achettyiitr, line 1486: We are using global variable and closure variable `loadFilesTableEventCountMap` with the same name,

warehouse/upload.go
```diff
@@ -745,6 +757,10 @@ func (job *UploadJobT) fetchPendingUploadTableStatus() []*TableUploadStatusT {
 	return tableUploadStatuses
 }
 
+func (job *UploadJobT) resetTableUploadStatuses() {
```
achettyiitr, line 760: Any specific reason why we have `resetTableUploadStatuses`?

warehouse/upload.go
```diff
@@ -1589,13 +1604,14 @@ func (job *UploadJobT) setStagingFilesStatus(stagingFiles []*StagingFileT, statu
...
-func (job *UploadJobT) getLoadFilesTableMap() (loadFilesMap map[tableNameT]bool, err error) {
+func (job *UploadJobT) getLoadFilesTablePresenceAndEventCountMaps() (loadFilesMap map[tableNameT]bool, loadFilesEventCountMaps map[tableNameT]int, err error) {
 	loadFilesMap = make(map[tableNameT]bool)
+	loadFilesEventCountMaps = make(map[tableNameT]int)
 
 	sourceID := job.warehouse.Source.ID
 	destID := job.warehouse.Destination.ID
 
-	sqlStatement := fmt.Sprintf(`SELECT distinct table_name FROM %s WHERE ( source_id = $1 AND destination_id = $2 AND id >= $3 AND id <= $4 );`,
+	sqlStatement := fmt.Sprintf(`SELECT table_name, sum(total_events) FROM %s WHERE ( source_id = $1 AND destination_id = $2 AND id >= $3 AND id <= $4 ) group by table_name;`,
```
achettyiitr, line 1614: Can we format this SQL statement?

warehouse/upload.go
dparveez-ruddr: This is being used to send success reports right?

dparveez-ruddr: > If it is not required. Can we remove this?

This is being used to send success reports right?

warehouse/upload.go
dparveez-ruddr: Doesn't this format of make ensure 0 length and capacity=len(statusOpts.ReportingMetrics)? This should avoid any chance of blanks right?

warehouse/upload.go
achettyiitr: I mean commented line. Do we need to keep those.

warehouse/upload.go
abhimanyubabbar: you are right, this becomes the cap.  Due to indentation, I was not looking at the initial `0`.

PR #2386, opened 2022-09-02
fix: use workspace id as aws iam role external id

koladilip:
# Description
This is to enhance AWS IAM role support we are forcing users set external id as workspace id, this done by propagating workspace id using destination struct to all the destination handlers.

Currently user can define their own external ID but then it is not secure and leads following problem:
Ref: https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html

## Notion Ticket
https://www.notion.so/rudderstacks/AWS-FTR-Force-AWS-role-to-use-workspace-token-as-external-ID-c64382423217478ab02a942498f38486

## Security


sandeepdsvs (approved):

saikumarrs (approved):

utils/awsutils/session.go
```diff
@@ -110,5 +111,13 @@ func NewSessionConfig(destinationConfig map[string]interface{}, timeout time.Dur
 	}
 	sessionConfig.Timeout = timeout
 	sessionConfig.Service = serviceName
+	if sessionConfig.IAMRoleARN != "" {
+		/**
+		In order prevent confused deputy problem, we are using
+		workspace token as external ID.
+		Ref: https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html
+		*/
+		sessionConfig.ExternalID = config.GetWorkspaceToken()
```
cisse21, line 120: `GetWorkspaceToken` would return the same value for all the customers on free deployment and for the new multi-tenant deployments. Is this okay to have it same for every customer?

saikumarrs: No, that'll be an issue. How to get each customer's workspace token?

koladilip: We decided to propagate workspace ID to destination handlers and use it as external id. @saikumarrs

sandeepdsvs (approved):

config/backend-config/types.go
```diff
@@ -48,6 +48,7 @@ type DestinationT struct {
 	DestinationDefinition DestinationDefinitionT
 	Config                map[string]interface{}
 	Enabled               bool
+	WorkspaceID           string
```
atzoum, line 51: Is this a new field returned by the control plane?

router/batchrouter/batchrouter.go
```diff
@@ -700,6 +700,7 @@ func (brt *HandleT) copyJobsToStorage(provider string, batchJobs *BatchJobsT, is
 			Config:           batchJobs.BatchDestination.Destination.Config,
 			UseRudderStorage: useRudderStorage,
 		}),
+		WorkspaceID: batchJobs.BatchDestination.Destination.WorkspaceID,
```
atzoum, line 703: Shouldn't we move the `WorkspaceID` field in `ObjectStorageOptsT` instead?
By doing it, `GetObjectStorageConfig` function would then be able to set the appropriate (new) entry in the `objectStorageConfigMap` if necessary for S3 and `GetS3Config` would simply use this new entry to populate `S3Config` with a new field, e.g. `ExternalID`.
Finally, `S3Manager#getSessionConfig` wouldn't need any `WorkspaceID` field, it would have all necessary information in its `S3Config`

config/backend-config/types.go
koladilip: It was there always, we are not using it before. No change done in CP on this.

router/customdestinationmanager/customdestinationmanager.go
```diff
@@ -121,12 +122,13 @@ func (customManager *CustomManagerT) newClient(destID string) error {
...
 }
 
-func (customManager *CustomManagerT) send(jsonData json.RawMessage, destType string, client, config interface{}) (int, string) {
+func (customManager *CustomManagerT) send(jsonData json.RawMessage, destType string, client interface{}, config map[string]interface{}) (int, string) {
 	var statusCode int
 	var respBody string
 	switch customManager.managerType {
 	case STREAM:
-		statusCode, _, respBody = streammanager.Produce(jsonData, destType, client, config)
+		streamProducer, _ := client.(common.StreamProducer)
```
atzoum, line 130: why did we have to move this type assertion out of `streammanager.Produce` and why we no longer test that the type assertion is a valid one?

koladilip: You mean, we should check for err rather than ignoring it? in the stream manager want to avoid taking interface{} but  we should test here too.

services/filemanager/s3manager.go
```diff
@@ -193,7 +197,12 @@ func (manager *S3Manager) getSession(ctx context.Context) (*session.Session, err
 		manager.Config.Region = aws.String(region)
 	}
 
-	return awsutils.CreateSession(manager.getSessionConfig())
+	var err error
+	manager.session, err = awsutils.CreateSession(manager.getSessionConfig())
```
atzoum, line 201: As it turns out, previously we were not reusing the session, but creating a new one every time. How do we verify now that reusing the session is safe and won't cause any regression?

services/streammanager/streammanager.go
```diff
@@ -65,15 +69,14 @@ func Close(producer interface{}, destType string) error {
...
-func Produce(jsonData json.RawMessage, destType string, producer, config interface{}) (int, string, string) {
+func Produce(jsonData json.RawMessage, destType string, producer common.StreamProducer, config map[string]interface{}) (int, string, string) {
 	switch destType {
 	case "KINESIS", "KAFKA", "AZURE_EVENT_HUB", "CONFLUENT_CLOUD", "PERSONALIZE",
 		"FIREHOSE", "EVENTBRIDGE", "GOOGLEPUBSUB", "GOOGLESHEETS", "BQSTREAM", "LAMBDA":
-		streamProducer, ok := producer.(common.StreamProducer)
-		if !ok {
-			return 400, "Could not create stream producer", "Could not create stream producer"
+		if producer == nil {
+			return 500, "producer is nil", "producer is nil"
```
atzoum, line 77: why are changing this to a retryable error?

koladilip: This is not user data error but rudder server went into a bad state so kept 500, should we panic in this case? what is the best way to handle it

utils/awsutils/session.go
```diff
@@ -96,19 +97,30 @@ func CreateSession(config *SessionConfig) (*session.Session, error) {
...
+func NewSessionConfigForDestination(destination *backendconfig.DestinationT, timeout time.Duration, serviceName string) (*SessionConfig, error) {
+	if destination == nil {
+		return nil, errors.New("destination should not be nil")
 	}
 	sessionConfig := SessionConfig{}
-	if err := mapstructure.Decode(destinationConfig, &sessionConfig); err != nil {
+	if err := mapstructure.Decode(destination.Config, &sessionConfig); err != nil {
 		return nil, fmt.Errorf("unable to populate session config using destinationConfig: %w", err)
 	}
+	if sessionConfig.Region == "" {
```
atzoum, line 108: just need a confirmation that this is really required :)

services/streammanager/streammanager.go
atzoum: how did rudder server went into a bad state and what makes us believe that on next try things will be better?

utils/awsutils/session.go
koladilip: mostly not required as streaming dest handlers have region as mandatory param. S3 is where I think we are making region optional but in s3 we are not using this method. In one of the destination it was there so I pull it here but can be removed.

services/filemanager/s3manager.go
koladilip: // This token provider will be called when ever the assumed role's
    // credentials need to be refreshed. Within the context of service clients
    // all sharing the same session the SDK will ensure calls to the token
    // provider are atomic. When sharing a token provider across multiple
    // sessions additional synchronization logic is needed to ensure the
    // token providers do not introduce race conditions. It is recommend to
    // share the session where possible.
https://docs.aws.amazon.com/sdk-for-go/api/aws/session/

atzoum: Please explain how is this documentation excerpt relevant? My understanding is that it has to do with MFA. How is it possible for us to be using MFA without third-party (i.e. user input)?

services/streammanager/streammanager.go
atzoum: First thing first, we need to make sure this cannot happen in production because we cannot really recover from such a panic without altering the codebase to fix the underlying bug.
As long as we can ensure that, whether we return 500, 400 or panic is not really that much important to me :)
Now, excluding test cases and after going through the code I don't see how this can be nil in practice, so it should be ok to panic

utils/awsutils/session.go
atzoum: just want to make sure that there is no possibility for any existing configuration to break due to this. Maybe using the backtesting tool would be enough here?

services/filemanager/s3manager.go
koladilip: Sorry my bad, here is right excerpt:
Sessions are safe to use concurrently as long as the Session is not being modified. Sessions should be cached when possible, because creating a new Session will load all configuration values from the environment, and config files each time the Session is created. Sharing the Session value across all of your service clients will ensure the configuration is loaded the fewest number of times possible.
https://docs.aws.amazon.com/sdk-for-go/api/aws/session/

utils/awsutils/session.go
koladilip: Sure will do that

router/customdestinationmanager/customdestinationmanager.go
koladilip: Currently decided to not handle this as this is not going to happen in product at all as SendData won't even call send if client is not initialised.

router/batchrouter/batchrouter.go
koladilip: Implemented.

services/filemanager/s3manager.go
koladilip: @atzoum can we resolve this?

cisse21 (approved):

services/filemanager/s3manager.go
atzoum: Fine, let's have our eyes open though for any possible regression due to this session reuse, hopefully there won't be any @rudderlabs/server_team

atzoum (approved):

services/filemanager/s3manager.go
koladilip: Sure

saikumarrs (approved):

koladilip:
Did backtest, and all is fine. We will do testing in the dev cluster in the coming week.

PR #2391, opened 2022-09-06
chore: add destType in tags and remove from name

cisse21:
# Description
Batch Router stats have the destinationType in the stat name causing difficulty in grouping by destType on Grafana. This PR attempts to add them as tags.

Also removed a redundant table count stat from jobsdb 

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=6059f6a105144d47b59d7be7d4c3a839&pm=s)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -408,8 +408,8 @@ type HandleT struct {
 	dsCacheLock                   sync.Mutex
 	BackupSettings                *backupSettings
 	jobsFileUploader              filemanager.FileManager
-	statTableCount                stats.RudderStats
 	statDSCount                   stats.RudderStats
+	statTableCount                stats.RudderStats
 	statNewDSPeriod               stats.RudderStats
 	invalidCacheKeyStat           stats.RudderStats
 	isStatNewDSPeriodInitialized  bool
@@ -861,6 +861,7 @@ func (jd *HandleT) workersAndAuxSetup() {
 
 	jd.logger.Infof("Connected to %s DB", jd.tablePrefix)
 
+	// TODO: Get rid of statTableCount after changing the alerts appropriately
 	jd.statTableCount = stats.DefaultStats.NewStat(fmt.Sprintf("jobsdb.%s_tables_count", jd.tablePrefix), stats.GaugeType)
 	jd.statDSCount = stats.NewTaggedStat("jobsdb.tables_count", stats.GaugeType, stats.Tags{"customVal": jd.tablePrefix})
 	jd.tablesQueriedStat = stats.NewTaggedStat("tables_queried_gauge", stats.GaugeType, stats.Tags{
--- a/router/batchrouter/batchrouter.go
+++ b/router/batchrouter/batchrouter.go
@@ -30,7 +30,7 @@ import (
 	"github.com/thoas/go-funk"
 	"golang.org/x/sync/errgroup"
 
-	uuid "github.com/gofrs/uuid"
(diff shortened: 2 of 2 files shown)


Sidddddarth (approved):

Sidddddarth:
Approved
Maybe the `module` tag in the stats is unnecessary if the name already consists of the module name.

router/batchrouter/batchrouter.go
```diff
@@ -1797,7 +1797,7 @@ func (worker *workerT) workerProcess() {
 			rruntime.Go(func() {
 				switch {
 				case misc.ContainsString(objectStorageDestinations, brt.destType):
-					destUploadStat := stats.DefaultStats.NewStat(fmt.Sprintf(`batch_router.%s_dest_upload_time`, brt.destType), stats.TimerType)
+					destUploadStat := stats.NewTaggedStat(`batch_router.dest_upload_time`, stats.TimerType, map[string]string{"destType": brt.destType})
```
atzoum, line 1800: We need to have both measurements (old & new) and only remove the old measurement as soon as operator changes have been deployed. these measurement are already used in grafana dashboards

atzoum (changes requested):

jobsdb/jobsdb.go
```diff
@@ -861,7 +860,6 @@ func (jd *HandleT) workersAndAuxSetup() {
 
 	jd.logger.Infof("Connected to %s DB", jd.tablePrefix)
 
-	jd.statTableCount = stats.DefaultStats.NewStat(fmt.Sprintf("jobsdb.%s_tables_count", jd.tablePrefix), stats.GaugeType)
```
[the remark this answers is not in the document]
cisse21: Sorry missed it but any reason why we are not using the group by query in the operator?

router/batchrouter/batchrouter.go
cisse21: Which panel is using these? I don't see any on hosted grafana

jobsdb/jobsdb.go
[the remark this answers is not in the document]
atzoum: there is no reason, but changing it is a long process :)

router/batchrouter/batchrouter.go
atzoum: https://github.com/rudderlabs/rudderstack-operator/blob/85a7eaf55b07db21b3faa7001fb7f586bc9039aa/helm-charts/renders/onprem/rudderstack-monitor/charts/grafana/templates/ruddermetric-dashboards-json-configmap.yaml#L4693

jobsdb/jobsdb.go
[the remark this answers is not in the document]
chandumlg: In the very beginning, table prefix was part of the metric. Alerts were defined on that metric. Later we defined new metrics and used table prefix as tag. But we never removed the old metric because we had to update dashboards and alert definitions. I think dashboards were updated but alert definitions are not. We need a task to do that. Till that time a TODO in the code makes sense.

BTW, should this PR be seen as a breaking major change? or minor change?

router/batchrouter/batchrouter.go
atzoum: the panel's name is **Batch Router destination (S3, MinIO, etc.)**

PR #2396, opened 2022-09-07
refactor: use generics for contains function

lvrach:
# Description

After officially upgrading to go1.18 and having a grace period of a few weeks, it is time to start introducing generics in our code base.

The case of `Contains` function was low-hanging fruit. With the simple use of comparable constraint, we could get rid of the different type variables of the methods and let go compiler figure out the rest. 

```go
Contains[K comparable](slice []K, item K) bool
```
Any type that satisfies comparable will work.

Please note, if you mix types, you are going to get a compiler error:
```go
Contains([]string{"a", "b"}, 1)
```

## Notion Ticket

https://www.notion.so/rudderstacks/Use-generics-for-contains-func-7c1eb55d135b4850864a3dbcefd063cd

## Security


warehouse/bigquery/bigquery.go
```diff
@@ -790,8 +790,11 @@ func (bq *HandleT) TestConnection(warehouse warehouseutils.WarehouseT) (err erro
 
 func (bq *HandleT) LoadTable(tableName string) error {
 	var getLoadFileLocFromTableUploads bool
-	if misc.ContainsString([]string{warehouseutils.IdentityMappingsTable, warehouseutils.IdentityMergeRulesTable}, tableName) {
```
lvrach, line 793: NOTE: here, I found the usage of Contains unnecessary so I removed it.

atzoum (approved):
There is a lot of boilerplate (and bugs) that we can avoid by using generic functions!
Until the standard library finally catches up and starts providing its own proper utility functions over collections and maps, we might benefit by adopting an external library, e.g. [this](https://github.com/samber/lo)

PR #2402, opened 2022-09-07
chore(warehouse): syncs issues for no uploads present

achettyiitr:
# Description

- Handle no syncs case when the warehouse tries to merge existing syncs with the new ones thereby users do not get the existing syncs and we are sending a 204 response code to handle it.
- Come code cleanup around the warehouse Grpc.

## Notion Ticket
https://www.notion.so/rudderstacks/Syncs-dashboard-fixes-4d3c454856d844e09cc15a73012a37e1

## Security


achettyiitr:
Moved this changes to [here](https://github.com/rudderlabs/rudder-server/pull/2732).

PR #2403, opened 2022-09-07
feat: thread safe Job Storage

saurav-malani:
# Description

> Currently Storage of jobs is not thread safe i.e. it should never happen that we insert in the second last DS and not last (latest) DS. For eg. there is a possibility that jobsDB-1 created a new DS but, during that time other instance of jobsDB is still inserting in the last DS, which shouldn't be the case. 

So in this PR, we addressed the above problem. Essentially, to make sure that all the instance of jobsDB add jobs only to last DS, we do the following:-
1. we acquire a Full Lock on last DS (to make sure that no one inserts any job in the process of adding new DS). 
2. Once full lock is acquired, we create new DS.
3. If creating new DS succeeds, we set a `trigger` on latest full table(now penultimate), which will throw custom exception code "RS001", if insert operation is tried on it.
4. This exception code is essentially used to convey other instances of jobsDB that new DS is added & they need to refresh their DS cache. So, `Store` func smartly handles this exception to refresh DS cache & tires storing again.

## Notion Ticket

https://www.notion.so/rudderstacks/Thread-safe-job-storage-62c193190578438489fc4695012fc175


jobsdb/jobsdb_test.go
```diff
@@ -797,3 +798,84 @@ func TestThreadSafeAddNewDSLoop(t *testing.T) {
 		time.Second, time.Millisecond,
 		"expected only one DS to be added, even though both jobsDB-1 & jobsDB-2 are triggered to add new DS (dsLen1: %d, dsLen2: %d)", dsLen1, dsLen2)
 }
+
+func TestThreadSafeJobStorage(t *testing.T) {
```
atzoum, line 802: The test appears to be incomplete. A concurrent scenario we need to test is one where 1 jobsDB creates a new dataset and the other jobsDB tries to store jobs before being aware of this change

jobsdb/jobsdb.go
```diff
@@ -3031,7 +3062,30 @@ func (jd *HandleT) updateJobStatusDSInTx(ctx context.Context, tx *sql.Tx, ds dat
...
+	_, err := jd.dbHandle.Exec(sqlStatement)
+
+	var e *pq.Error
+	if err != nil && errors.As(err, &e) {
+		if e.Code == postgresDuplicateFuncErr {
+			return nil
+		}
+	}
+	return err
+}
```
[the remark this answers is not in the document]
saurav-malani: "`CREATE OR REPLACE FUNCTION` will either create a new function, or replace an existing definition. To replace the current definition of an existing function, use `CREATE OR REPLACE  FUNCTION`."
ref: https://www.postgresql.org/docs/current/sql-createfunction.html

Since, create or replace replaces the existing func, do you think it is a good idea to unnecessarily overwrite the func again and again?

jobsdb/jobsdb.go
```diff
@@ -2373,8 +2378,21 @@ func (jd *HandleT) doStoreJobsInTx(ctx context.Context, tx *sql.Tx, ds dataSetT,
 		return err
 	}
 	err := store()
+
 	var e *pq.Error
 	if err != nil && errors.As(err, &e) {
+
+		if e.Code == lockTableInsertExceptionCode {
+			var updatedList []dataSetT
+			jd.dsListLock.WithLock(func(l lock.DSListLockToken) {
```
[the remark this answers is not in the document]
saurav-malani: great catch, it completely got out of my mind that we are calling it inside `WithStoreSafeTx` and this can create a deadlock situation. 
BTW, what will be an ideal (systematic) approach to avoid mistakes like this in future?

jobsdb/jobsdb_test.go
saurav-malani: I have added another test to check to verify store() always stores the jobs in latest DS.

jobsdb/jobsdb.go
```diff
@@ -2390,26 +2409,26 @@ func (jd *HandleT) doStoreJobsInTx(ctx context.Context, tx *sql.Tx, ds dataSetT,
...
 	sqlStatement := fmt.Sprintf(`INSERT INTO %q (uuid, user_id, custom_val, parameters, event_payload, workspace_id)
-	                                   VALUES ($1, $2, $3, $4, $5, $6) RETURNING job_id`, ds.JobTable)
+		VALUES ($1, $2, $3, $4, $5, $6) RETURNING job_id`, ds.JobTable)
 	stmt, err := tx.PrepareContext(ctx, sqlStatement)
-	jd.assertError(err)
-	defer func() { _ = stmt.Close() }()
+	if err != nil {
+		return err
+	}
+	defer stmt.Close()
```
raman2805, line 2417: I have seen other people usually follow the convention of using 
defer func() { _ = stmt.Close() }()
I have been told this makes it clear that we are ignoring the return value and is the preferable style. We should probably stick to that convention. You might want to check with @fracasula  (who might be able to advise better)

jobsdb/jobsdb.go
```diff
@@ -2390,26 +2409,26 @@ func (jd *HandleT) doStoreJobsInTx(ctx context.Context, tx *sql.Tx, ds dataSetT,
...
+	if err != nil && errors.As(err, &e) {
+		errCode := string(e.Code)
 		if _, ok := dbInvalidJsonErrors[errCode]; ok {
 			return errors.New("invalid JSON")
 		}
 	}
+
+	// Empty customValFilters means we want to clear for all
+	jd.markClearEmptyResult(ds, allWorkspaces, []string{}, []string{}, nil, hasJobs, nil)
+	jd.markClearEmptyResult(ds, job.WorkspaceId, []string{}, []string{}, nil, hasJobs, nil)
```
raman2805, line 2431: [question] The order of the statements has been reversed, hope this is intended

jobsdb/jobsdb.go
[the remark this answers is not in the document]
raman2805: Either ways, I guess we should move this to startup to avoid doing this on every DS creation (with CREATE / CREATE or REPLACE) functions

atzoum (approved):

atzoum (approved):

jobsdb/jobsdb.go
saurav-malani: done.

jobsdb/jobsdb.go
saurav-malani: reverted back.

raman2805 (approved):

jobsdb/jobsdb.go
```diff
@@ -2031,9 +2043,24 @@ func (jd *HandleT) WithStoreSafeTx(f func(tx StoreSafeTx) error) error {
...
+		err = op()
+		if err != nil && errors.Is(err, errStaleDsList) {
+			jd.logger.Errorf("[JobsDB] :: Store failed: %v. Retrying after refreshing DS cache", errStaleDsList)
+			jd.dsListLock.WithLock(func(l lock.DSListLockToken) {
+				_ = jd.refreshDSList(l)
+			})
+		} else {
+			break
+		}
+	}
```
saurav-malani, line 2062: In case of high load or when `maxDSSize` is less, there is a possibility that even the newly added DS gets filled by the time we refresh & try again. So, we keep refreshing DS until `Store()` succeeds or some other error is returned. 
And, given we call it with `Context.WithTimeout()` it will return in bounded time.

jobsdb/jobsdb.go
```diff
@@ -2031,9 +2043,24 @@ func (jd *HandleT) WithStoreSafeTx(f func(tx StoreSafeTx) error) error {
...
+	return err
 }
```
atzoum, line 2064: minor simplification
```suggestion
	for {
		err := op()
		if err != nil && errors.Is(err, errStaleDsList) {
			jd.logger.Errorf("[JobsDB] :: Store failed: %v. Retrying after refreshing DS cache", errStaleDsList)
			jd.dsListLock.WithLock(func(l lock.DSListLockToken) {
				_ = jd.refreshDSList(l)
			})
		} else {
			return err
		}
	}
}
```

saurav-malani: modified.

atzoum (approved):

PR #2410, opened 2022-09-08
chore: disable parallel test execution and refactor jobsdb tests

atzoum:
# Description
As long as we are depending on global and environment variables for initialization, it is not safe to execute tests in parallel, e.g.
1. One test might pick up the configuration of another test
2. One test might override the configuration of another test
3. One test's initialization (config, logging, admin, etc.) might cause a panic to another running test.

Even though parallel test execution is desirable, however it is unsafe and constantly requires effort for debugging, retrying, etc. Thus disabling it.

Furthermore, the following refactoring was performed in jobsdb tests
- Using a common function for starting the postgresql container, setting required environment variables and initializing global variables across go tests, benchmarks & ginkgo tests
- Using a different postgresql container in most test cases
- Using a random table prefix for each separate test
- Various other small fixes

## Security

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,10 +11,10 @@ mocks: install-tools ## Generate all mocks
 
 test: ## Run all unit tests
 ifdef package
-	SLOW=0 $(GINKGO) -p --randomize-all --randomize-suites --fail-on-pending --cover  \
+	SLOW=0 $(GINKGO) --fail-on-pending --cover  \
 		-coverprofile=profile.out -covermode=atomic --trace -keep-separate-coverprofiles $(package) ;
 else
-	SLOW=0 $(GINKGO) -p --randomize-all --randomize-suites --fail-on-pending --cover  \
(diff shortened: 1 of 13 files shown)


jobsdb/setup.go
```diff
@@ -37,13 +37,12 @@ func (jd *HandleT) setupDatabaseTables(l lock.DSListLockToken, clearAll bool) {
 		"Prefix":   jd.tablePrefix,
 		"Datasets": datasetIndices,
 	}
-
 	psqlInfo := GetConnectionString()
+	var err error
```
chandumlg, line 41: minor: Need this?

chandumlg (approved):

jobsdb/setup.go
atzoum: nope, leftover since the time I was fighting with parallel tests, thanks for spotting!

fracasula (approved):

PR #2411, opened 2022-09-08
chore: using go test

fracasula:
# Description

Running tests with `go test` instead of `ginkgo` due to some issues that the team has been experiencing:
* ginkgo doesn't seem to considering `testing.Verbose()`
* when running tests in parallel sometimes failures are not reported and we have to manually re-run the tests without the `-p`
* some tests seem to fail more often with ginkgo that with `go test`

If after some time we see that the situation has not improved we can easily go back by changing the `Makefile`.

## Additional changes

* fixing some flaky tests
* I made some changes around the `test` target in the `Makefile` to make sure that we always tear down removing coverage files.

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/Try-go-test-instead-of-ginkgo-d757557227ed42c3b3aa436a60ad34dc) >

## Security


atzoum (approved):

jobsdb/integration_test.go
```diff
@@ -1295,3 +1296,10 @@ func getPayloadSize(t *testing.T, jobsDB JobsDB, job *JobT) (int64, error) {
 
 	return size, err
 }
+
+func getRefreshedDSList(jobDB *HandleT) []dataSetT {
+	jobDB.dsListLock.WithLock(func(l lock.DSListLockToken) {
+		jobDB.refreshDSList(l)
+	})
+	return jobDB.getDSList()
+}
```
saurav-malani, line 1305: I don't think making this change & every time calling `jobDB.refreshDSList()` would be helpful. Given that, in `addNewDSLoop` func we refresh the DS cache after adding new DS.

fracasula: @saurav-malani I added this because I kept getting an empty DS list in my test, this is how I solved it.

What alternative do you suggest?

The failing test was `should migrate the datasets after maxDSRetentionPeriod (except the right most one)`.

fracasula: To be more precise, I was getting a length of zero [here](https://github.com/rudderlabs/rudder-server/blob/2134c54d1b10fe40a15abeac153f15982b0509ad/jobsdb/integration_test.go#L583).

atzoum: getting this error after turning off parallel tests? Was getting the same error, but only with parallel test execution.

testhelper/ignore_err.go
```diff
@@ -0,0 +1,6 @@
+package testhelper
+
+// IgnoreErr is a helper function to ignore errors in tests, useful when closing bodies for example
+func IgnoreErr(f func() error) {
```
saurav-malani, line 4: we do ignore error at many other places apart from test. So, making it accessible globally would make sense, maybe in `utils/misc`.

jobsdb/integration_test.go
fracasula: Yes because the first thing I did was to replace `ginkgo` with the `go test` you see in the `Makefile`, so I was definitely getting that without parallelism. The only thing I'm doing now is shuffling, doesn't seem related though.

atzoum: Note that the go test runs the tests for different packages in parallel by default, unless the -p 1 is passed to it

multi_tenant_test.go
```diff
@@ -196,7 +197,7 @@ func testMultiTenantByAppType(t *testing.T, appType string) {
 	require.ErrorContains(t, err, "connection refused")
 	require.Nil(t, resp)
 	if err == nil {
-		defer resp.Body.Close()
+		defer func() { _ = resp.Body.Close() }()
```
saurav-malani, line 200: might want to use `IgnoreErr` func that you introduced in `testhelper` pkg.

jobsdb/integration_test.go
fracasula: And "different packages" fine in this case is not fine? You think it should be disabled?

atzoum: no, different packages are not fine :)

multi_tenant_test.go
fracasula: In this case I'm getting a complaint by golangci-lint because it can't detect that we're actually closing the body. I will get rid of this helper.

testhelper/ignore_err.go
fracasula: I'll just get rid of it due to linting issues: https://github.com/rudderlabs/rudder-server/pull/2411#discussion_r967192780

saurav-malani (approved):
Overall looks good, added a few non-blocker comments.

jobsdb/integration_test.go
fracasula: Enforcing `-p=1` for now till we got rid of global variables throughout the codebase (e.g. logger, config, etc...).

PR #2412, opened 2022-09-09
chore: fix cache bug

cisse21:
# Description

Fix minor bug to use cache for no parameter filters scenarios

## Notion Ticket

[Notion Link ](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=54e42282047f463fa8496961ae889531&pm=s)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -2464,6 +2464,7 @@ func (jd *HandleT) markClearEmptyResult(ds dataSetT, workspace string, stateFilt
 		}
 		sort.Strings(pVals)
 		pVal := strings.Join(pVals, "_")
+		pvalArr := []string{pVal, cValDefaultFilter}
 
 		_, ok = jd.dsEmptyResultCache[ds][workspace][cVal][pVal]
 		if !ok {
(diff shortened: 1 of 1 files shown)


jobsdb/jobsdb.go
```diff
@@ -2421,7 +2419,7 @@ func (jd *HandleT) markClearEmptyResult(ds dataSetT, workspace string, stateFilt
 	}
 
 	// Safe Check , All parameter filters must be provided explicitly
-	if len(parameterFilters) != len(CacheKeyParameterFilters) {
+	if len(parameterFilters) != len(CacheKeyParameterFilters) && len(parameterFilters) != 0 {
```
atzoum, line 2422: is length alone enough? are we checking that keys match elswhere?

atzoum:
@cisse21 can you remove this too?
https://github.com/rudderlabs/rudder-server/blob/0708615155cc388ef97071f06bdcbb7fee77c05b/jobsdb/jobsdb.go#L1093

jobsdb/jobsdb.go
cisse21: We check it in the below loop. If the length matches and the keys match then we go ahead

jobsdb/jobsdb.go
```diff
@@ -1089,8 +1089,6 @@ func (jd *HandleT) refreshDSRangeList(l lock.DSListLockToken) {
 		}
 		_ = rows.Close()
 		jd.logger.Debug(sqlStatement, minID, maxID)
-
-		_ = rows.Close()
```
cisse21, line 1093: @atzoum  This is what you wanted me to remove right?

atzoum (approved):

jobsdb/jobsdb.go
```diff
@@ -2478,15 +2477,16 @@ func (jd *HandleT) markClearEmptyResult(ds dataSetT, workspace string, stateFilt
 			jd.dsEmptyResultCache[ds][workspace][cVal][cValDefaultFilter] = map[string]cacheEntry{}
 		}
 
-		for _, st := range stateFilters {
-			previous := jd.dsEmptyResultCache[ds][workspace][cVal][pVal][st]
-			if checkAndSet == nil || *checkAndSet == previous.Value {
-				cache := cacheEntry{
-					Value: value,
-					T:     time.Now(),
+		for _, pf := range pvalArr {
```
atzoum, line 2480: what was the issue here?

cisse21: So there are two places where cache value is changed

When we do a store we change entries in the cache map:
- For every job which we get we store cache for the default cache value(cval_cval) and with (destID_xxxx)

Then whenever we do a get we check `isEmptyResult` we check dynamically based on how the query is being made
- with cval_cval is parameter filters are empty
- with destID_xxxx if parameter filters are passed on

Post this after a successful query if we end up having no jobs, then we would need to update the cache for both cval_cval and destID_xxxx, but the problem is the query might send either valid param filters or empty filters and so we need to account the updation for both the cases

lvrach (approved):

lvrach:
Do we need rudder-transformer changes?

jobsdb/jobsdb.go
atzoum: no I don't think there is a way to properly test it, I was just trying to make a stronger point of why the current cache implementation needs to be redesigned!

cisse21:
No transformer changes should be necessary for this

PR #2413, opened 2022-09-09
chore: introduce tooling for local development

lvrach:
# Description

In this PR, I've introduced a CLI tool for assisting in local dev testing.

## Motivation

The introduction of multi-tenant architecture makes rudder-server more complex to run in a non-k8s environment.

For technical and legal reasons, some components responsible for starting rudder-server can not be added to the open source repo or run locally efficiently.

## Approach

For this reason, I have introduced a tool that emulates the scheduler's behaviour. Change mode and workspaceIDs. This also helps with debugging the system in isolation.

In addition to etcd, I have added to the tool two more commands that are essential for testing locally: an event sender and a webhook. Currently, we are using `./scripts` folder and the `test-dest` repo. I believe this tool can replace the script folder making it easier to work with and more extendable.

It is worth noting that: `make run-mt` can be used as a single command to run rudder-server in multitenant mode. in addition with `docker compose --profile multi-tenant up` and setting up `WORKSPACE_NAMESPACE`, `HOSTED_SERVICE_SECRET`

Additional documentation will be added to the notion.

## Notion Ticket

https://www.notion.so/rudderstacks/Run-multitenant-locally-9eadbd20f21f4dc684ffb7240f80990c

## Security


chandumlg (approved):

docker-compose.yml
```diff
@@ -24,10 +24,8 @@ services:
 #   Uncomment the following lines to mount workspaceConfig file
 #   volumes:
 #     - <absolute_path_to_workspace_config>:/etc/rudderstack/workspaceConfig.json
-  d-transformer:
-    build:
-      context: ./rudder-transformer/
-      dockerfile: Dockerfile
+  transformer:
+    image: "rudderlabs/develop-rudder-transformer:master"
```
cisse21, line 28: with this change can we also completely get rid of the transformer submodule in the rudder-server repo right?

lvrach: Yeah, I think it makes more sense from rudder-server to use the docker image.

Not sure if it create inconvenience for people working with transformer

fracasula (approved):
LGTM, just not sure whether this belongs here or in some other repo like `rudder-tests` for example. Either way it looks useful, thanks :+1:

PR #2422, opened 2022-09-12
chore: modified regulation-worker response code handling from 404 to 204, when no job is available

saurav-malani:
# Description

> Currently in case of Data Regulation Service, if no job is found, regulation-manager returns 404 statusCode with `ErrNoRunnableJob` body. But, Control-Plane get alerts if it returns 404 for any request. So, we are modifying the status code returned from `404` to `204` in regulation-manager & modifying here (in regulation-worker) to handle the same. (regulation-manager PR: https://github.com/rudderlabs/data-regulation-service/pull/50)

## Notion Ticket

https://www.notion.so/rudderstacks/For-regulation-worker-regulation-manager-modify-statusCode-for-no-jobs-found-from-404-to-204-95b3e80b87a64d3599e95c1588d4269e

### DIFF
--- a/regulation-worker/internal/client/client.go
+++ b/regulation-worker/internal/client/client.go
@@ -64,6 +64,11 @@ func (j *JobAPI) Get(ctx context.Context) (model.Job, error) {
 	// if successful
 
 	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		if resp.StatusCode == http.StatusNoContent {
+			pkgLogger.Debugf("no runnable job found")
+			return model.Job{}, model.ErrNoRunnableJob
+		}
+
 		var jobSchema jobSchema
 		if err := json.NewDecoder(resp.Body).Decode(&jobSchema); err != nil {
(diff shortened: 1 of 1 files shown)


regulation-worker/internal/client/client.go
```diff
@@ -82,9 +87,6 @@ func (j *JobAPI) Get(ctx context.Context) (model.Job, error) {
 		pkgLogger.Debugf("obtained job: %v", job)
 		return job, nil
 
-	} else if resp.StatusCode == http.StatusNotFound {
```
atzoum, line 85: don't we need both old and new client behaviours? How are we supposed to rollout such a breaking change otherwise?

fracasula: Yeah sounds about right to me. Have both then delete the deprecated one in a later deployment, right?

saurav-malani: yup, I have added it back, with a note.

atzoum (approved):

PR #2424, opened 2022-09-12
chore: fix flaky tests

atzoum:
# Description

- Properly starting and stopping the gateway on each test
- Fixing a bug where we were sending an empty batch request to the worker during gateway shutdown
- acquire a read lock on the jobsdb DS list before reading it
- avoid unnecessarily setting jobsdb DS list to nil when refreshing it

## Security

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -273,8 +273,9 @@ func (gateway *HandleT) userWorkerRequestBatcher() {
 		select {
 		case userWorkerBatchRequest, hasMore := <-gateway.userWorkerBatchRequestQ:
 			if !hasMore {
-				breq := batchUserWorkerBatchRequestT{batchUserWorkerBatchRequest: userWorkerBatchRequestBuffer}
-				gateway.batchUserWorkerBatchRequestQ <- &breq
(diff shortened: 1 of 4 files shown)


fracasula (approved):
Makes sense to me, thanks Aris :+1:

gateway/gateway.go
```diff
@@ -418,12 +419,6 @@ func (gateway *HandleT) getSourceTagFromWriteKey(writeKey string) string {
 //
 // Finally sends responses(error) if any back to the webRequests over their `done` channels
 func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWebRequestWorkerT) {
-	if userWebRequestWorker == nil {
```
chandumlg, line 421: Why panics removed?

fracasula: Yeah I introduced them recently for debugging.

saurav-malani (approved):

gateway/gateway.go
chandumlg: 👍

chandumlg (approved):

PR #2430, opened 2022-09-13
chore: add reason tag in failed requests and failed events stats

satishrudderstack:
Currently, we have stats for failed requests and failed events. With this change we will also be capturing the reason for the failure. Example reasons are invalidJson, invalidWriteKey


gateway/gateway.go
```diff
@@ -667,14 +689,14 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 		// update stats request wise
 		gateway.updateSourceStats(sourceStats, "gateway.write_key_requests", sourceTagMap)
 		gateway.updateSourceStats(sourceSuccessStats, "gateway.write_key_successful_requests", sourceTagMap)
-		gateway.updateSourceStats(sourceFailStats, "gateway.write_key_failed_requests", sourceTagMap)
+		gateway.updateFailedSourceStats(sourceFailStats, "gateway.write_key_failed_requests", sourceTagMap)
```
lvrach, line 692: I find it peculiar having 3 different metrics that capture similar things, with an introduction of a tag we can simplify things.

What if `gateway.write_key_requests` had two extra tags:
* `failed`: true/false 
* `reason` - same to what you introduce here.

Using the `failed` tag above you can get any metric you like.

With this, we need can simplify our code and make our stats more compact.

We still need to keep:

```go
		gateway.updateSourceStats(sourceSuccessStats, "gateway.write_key_successful_requests", sourceTagMap)
		gateway.updateFailedSourceStats(sourceFailStats, "gateway.write_key_failed_requests", sourceTagMap)
```

For a few release for backwards compatibility, but after that, they can be removed.

lvrach (approved):
Approve, but with a suggestion that would make our stats a bit cleaner.

satishrudderstack:
@lvrach Added a new commit to this PR. https://github.com/rudderlabs/rudder-server/pull/2430/commits/53779e3cbe1d662bfaa516f88b54487bfcce8028. Please take a look

gateway/gateway.go
satishrudderstack: @lvrach The similar metrics exist for gateway events, `gateway.write_key_events`, `gateway.write_key_successful_events`, `gateway.write_key_failed_events`

Let's take this a separate change ?

Also, I think we might need to revisit all the metrics once. I think we need to add rudderstack resourceIDs (`sourceID`, `destinationID`, `transformationID`, `trackingPlanID`) in the tags wherever possible. Going forward, this would enable us to send more notifications to the customers. With your changes to filter out these tags in hosted multi-tenant this should not be an issue ?

gateway/gateway.go
```diff
@@ -443,7 +459,9 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 		for _, req := range breq.batchRequest {
 			writeKey := req.writeKey
 			sourceTag := gateway.getSourceTagFromWriteKey(writeKey)
```
cisse21, line 461: Do we still need this tag now that we are using sourceID?

chandumlg: This is used in grafana dashboards now. Let's keep it for now.

gateway/gateway.go
```diff
@@ -210,6 +211,7 @@ func (gateway *HandleT) updateFailedSourceStats(sourceStats map[string]int, buck
 			"reqType":     sourceTagMap["reqType"],
 			"workspaceId": sourceTagMap["workspaceId"],
 			"reason":      sourceTagMap["reason"],
+			"sourceID":    sourceTagMap["sourceID"],
```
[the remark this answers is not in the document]
chandumlg: Skipping sourceID for free [here](https://github.com/rudderlabs/rudder-server/pull/2430/commits/a91654ea2679c345eadfa8975e12f86a0dc7b26a)

chandumlg (approved):

gateway/gateway.go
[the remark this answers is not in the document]
fracasula: A comment here would be great though, like to exclude tags please refer to `statsExcludedTags`.

gateway/gateway.go
fracasula: +1 for tags

main.go
```diff
@@ -244,7 +244,7 @@ func Run(ctx context.Context) int {
 	}
 	// TODO: remove as soon as we update the configuration with statsExcludedTags where necessary
```
fracasula, line 245: Can someone explain this TODO to me?

cisse21: Was discussing the same with @chandumlg ... So the basic idea is that for free customers we don't want to send so many tags to influx as it gets bloated up but essentially we should remove the redundant tags for every metric

PR #2435, opened 2022-09-13
chore(warehouse): warehouse api to capture task run ID when calculating the pending uploads.

abhimanyubabbar:
# Description
For sources to know that the task run has no transient uploads in the system, we need to capture the task run id in the endpoint and then call the pendingUploadCount with this filterBy clause.

## Notion Ticket

https://www.notion.so/rudderstacks/Warehouse-Pending-Events-API-a3e22ec2834c427fa3a7b8ed02c8e63e

## Security

### DIFF
--- a/warehouse/warehouse.go
+++ b/warehouse/warehouse.go
@@ -1361,6 +1361,11 @@ func pendingEventsHandler(w http.ResponseWriter, r *http.Request) {
 	// TODO : respond with errors in a common way
 	pkgLogger.LogRequest(r)
 
+	if r.Method != "POST" {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
 	// read body
 	body, err := io.ReadAll(r.Body)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


warehouse/warehouse.go
```diff
@@ -1506,21 +1513,24 @@ func getPendingStagingFileCount(sourceOrDestId string, isSourceId bool) (fileCou
...
-func getPendingUploadCount(sourceOrDestId string, isSourceId bool) (uploadCount int64, err error) {
-	sourceOrDestColumn := ""
-	if isSourceId {
-		sourceOrDestColumn = "source_id"
-	} else {
-		sourceOrDestColumn = "destination_id"
+func getPendingUploadCount(filters ...warehouseutils.FilterBy) (uploadCount int64, err error) {
+	pkgLogger.Debugf("Fetching pending upload count with filters: %v", filters)
+
+	query := fmt.Sprintf(`
```
achettyiitr, line 1519: Can we format the Query.

achettyiitr (changes requested):

warehouse/warehouse.go
```diff
@@ -1361,6 +1361,11 @@ func pendingEventsHandler(w http.ResponseWriter, r *http.Request) {
 	// TODO : respond with errors in a common way
 	pkgLogger.LogRequest(r)
 
+	if r.Method != "POST" {
```
achettyiitr, line 1364: Instead of using this can we move this to the server mutex?
`mux := mux.NewRouter()`
`mux.HandleFunc("/v1/warehouse/pending-events", pendingEventsHandler).Methods("POST")`

warehouse/warehouse.go
achettyiitr: There is a gap between % [1]s in the format string.

warehouse/warehouse.go
```diff
@@ -1678,7 +1691,7 @@ func startWebHandler(ctx context.Context) error {
 			backendconfig.DefaultBackendConfig.WaitForConfig(ctx)
 			mux.HandleFunc("/v1/process", processHandler)
 			// triggers uploads only when there are pending events and triggerUpload is sent for a sourceId
-			mux.HandleFunc("/v1/warehouse/pending-events", pendingEventsHandler)
+			mux.HandleFunc("/v1/warehouse/pending-events", pendingEventsHandler).Methods("GET")
```
achettyiitr, line 1694: We don't have a Methods function for ServerMux. We would need to convert it to Router.
```
mux := mux.NewRouter()
mux.HandleFunc("/v1/warehouse/pending-events", pendingEventsHandler).Methods("POST")
```

warehouse/warehouse.go
achettyiitr: We would not need this right.

achettyiitr (changes requested):

Jayachand (approved):

achettyiitr (approved):

PR #2436, opened 2022-09-13
monitoringQueryRoutine

Sidddddarth:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security


app/apphandlers/setup.go
```diff
@@ -242,5 +243,23 @@ func NewRsourcesService(deploymentType deployment.Type) (rsources.JobService, er
...
 	}
 
-	return rsources.NewJobService(rsourcesConfig)
+	service, err := rsources.NewJobService(rsourcesConfig)
+	if err != nil {
+		return nil, err
+	}
+	if rsourcesConfig.SharedConn != "" {
+		ctx := context.TODO()
+		lagStatus := service.MonitorLagLoop(ctx)
```
atzoum, line 252: can we do what we are doing now for the cleanupLoop? Don't start it by default, but let the caller decide when it will be started (so that we don't have 2 loops running, 1 in gateway and 1 in processor)

services/rsources/handler.go
```diff
@@ -404,3 +407,35 @@ func sqlFilters(jobRunId string, filter JobFilter) (fragment string, params []in
 	}
 	return filters, filterParams
 }
+
+func (sh *sourcesHandler) MonitorLagLoop(ctx context.Context) <-chan LagStatus {
+	lagCh := make(chan LagStatus)
+	logicalReplicationTrigger := func() <-chan time.Time {
+		return time.After(config.GetDuration("Rsources.stats.logRep.monitoring.interval", 10, time.Second))
```
cisse21, line 414: Can we shorten the config variable name?

cisse21 (approved):

services/rsources/handler.go
Sidddddarth: going with `Rsources.stats.logRepMonitoringInterval`

services/rsources/handler.go
```diff
@@ -404,3 +407,35 @@ func sqlFilters(jobRunId string, filter JobFilter) (fragment string, params []in
...
+			select {
+			case <-ctx.Done():
+				close(lagCh)
+			case <-logicalReplicationTrigger():
+				var lag int64
+				err := sh.localDB.QueryRowContext(
+					ctx,
+					`select
+					pg_wal_lsn_diff(pg_current_wal_lsn(), replay_lsn)::numeric as total_lag
+					from pg_stat_replication;`).Scan(&lag)
```
atzoum, line 432: Let's use replay_lag in seconds: 
```
select EXTRACT(EPOCH FROM replay_lag) from pg_stat_replication
```

app/apphandlers/processorAppHandler.go
```diff
@@ -267,6 +268,19 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 		return rsourcesService.CleanupLoop(ctx)
 	})
 
+	g.Go(func() error {
+		lagCh := rsourcesService.MonitorLagLoop(ctx)
+		replicationLagStat := stats.DefaultStats.NewStat("rsources_log_replication_lag", stats.CountType)
```
atzoum, line 273: let's add another counter for this query
```
select count(*) from pg_replication_slots
```

services/rsources/handler.go
```diff
@@ -404,3 +406,43 @@ func sqlFilters(jobRunId string, filter JobFilter) (fragment string, params []in
...
+
+func (sh *sourcesHandler) Monitor(ctx context.Context, lagGauge, replicationSlotGauge Gauger) {
+	if sh.sharedDB == nil {
+		sh.log.Warn("shared database is not configured, skipping logical replication monitoring")
+		return
+	}
+	logicalReplicationTrigger := func() <-chan time.Time {
+		return time.After(config.GetDuration("Rsources.stats.monitoringInterval", 10, time.Second))
+	}
+	go func() {
```
saurav-malani, line 418: why do we really need to put this under a `go routine`?

app/apphandlers/processorAppHandler.go
```diff
@@ -267,6 +268,13 @@ func (processor *ProcessorApp) StartRudderCore(ctx context.Context, options *app
 		return rsourcesService.CleanupLoop(ctx)
 	})
 
+	g.Go(func() error {
+		replicationLagStat := stats.DefaultStats.NewStat("rsources_log_replication_lag", stats.GaugeType)
+		replicationSlotStat := stats.DefaultStats.NewStat("rsources_log_replication_slot", stats.GaugeType)
```
saurav-malani, line 273: won't it be better if we move this to `monitor` func, instead of declaring it here and then needlessly passing them as arguments?

atzoum:
Thank you @Sidddddarth. Is it possible to add a test scenario for it as well?

services/rsources/handler.go
Sidddddarth: we don't

app/apphandlers/processorAppHandler.go
Sidddddarth: didn't want to import the entire package over there.
Something that `Gauge`s is all we need inside the package under consideration.

services/rsources/handler_test.go
```diff
@@ -698,6 +699,104 @@ var _ = Describe("Using sources handler", Ordered, func() {
...
+				_ = pool.Client.RemoveNetwork(network.ID)
+			}
+			if pgA.resource != nil {
+				purgeResource(pool, pgA.resource, pgB.resource)
+			}
+		})
+
+		It("should always have one replication slot on the localdb for replication", func() {
+			var replicationSlot int64
+			err := pgA.db.QueryRow(`select count(*) from pg_replication_slots;`).Scan(&replicationSlot)
```
atzoum, line 749: why are we not monitoring the gauge?

PR #2439, opened 2022-09-14
refactor: config package

atzoum:
# Description

Refactoring the `config` package and introducing the following features:

1. Support for multiple, independent config instances
2. Thread-safe initialization & reinitialization of config instances (including the default, singleton instance). No more nil pointer dereference panics.
3. Rely exclusively on viper's precedence order (see [doc](https://github.com/rudderlabs/rudder-server/blob/2d9233d62d649c23a57add4868557e71b9b9f2ac/config/config.go#L1-L16) - explicit overrides have higher precedence than environment variables)
4. Support overriding config at runtime, without having to use environment variables, [example](https://github.com/rudderlabs/rudder-server/blob/2d9233d62d649c23a57add4868557e71b9b9f2ac/eventorder_test.go#L106-L135) (thread-safe, see [config.Set](https://github.com/rudderlabs/rudder-server/blob/2d9233d62d649c23a57add4868557e71b9b9f2ac/config/config.go#L238))
5. Remove warehouse-specific code
6. Test coverage is now >80%

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=7968e584d23444028eb75351255f4b30&pm=s)

## Security

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -54,13 +54,12 @@ func New() *Config {
 
 // Config is the entry point for accessing configuration
 type Config struct {
-	vLock                  sync.RWMutex // protects reading and writing to the config (viper is not thread-safe)
-	v                      *viper.Viper
-	mapLock                sync.RWMutex // protects maps holding registered config keys
-	hotReloadableConfig    map[string][]*configValue
-	nonHotReloadableConfig map[string][]*configValue
-	envsLock               sync.RWMutex // protects the envs map below
-	envs                   map[string]string
+	vLock                   sync.RWMutex // protects reading and writing to the config (viper is not thread-safe)
+	v                       *viper.Viper
+	hotReloadableConfigLock sync.RWMutex // protects map holding hot reloadable config keys
+	hotReloadableConfig     map[string][]*configValue
+	envsLock                sync.RWMutex // protects the envs map below
+	envs                    map[string]string
 }
 
 // GetBool gets bool value from config
--- a/config/config.go
+++ b/config/config.go
@@ -27,7 +27,7 @@ import (
 	"github.com/spf13/viper"
 )
 
-// regular expression matching lowecase letter followed by an uppercase letter
+// regular expression matching lowercase letter followed by an uppercase letter
 var camelCaseMatch = regexp.MustCompile("([a-z0-9])([A-Z])")
 
 // regular expression matching uppercase letters contained in environment variable names
@@ -56,10 +56,10 @@ func New() *Config {
 type Config struct {
 	vLock                  sync.RWMutex // protects reading and writing to the config (viper is not thread-safe)
 	v                      *viper.Viper
-	mapLock                sync.RWMutex // protects maps holding registered condig keys
+	mapLock                sync.RWMutex // protects maps holding registered config keys
 	hotReloadableConfig    map[string][]*configValue
 	nonHotReloadableConfig map[string][]*configValue
-	envLock                sync.RWMutex // protects the envs map below
+	envsLock               sync.RWMutex // protects the envs map below
 	envs                   map[string]string
 }
 
@@ -98,7 +98,7 @@ func MustGetInt(key string) (value int) {
 	return c.MustGetInt(key)
 }
 
-// MustGetString gets int value from config or panics if the config doesn't exist
+// MustGetInt gets int value from config or panics if the config doesn't exist
 func (c *Config) MustGetInt(key string) (value int) {
 	c.vLock.RLock()
 	defer c.vLock.RUnlock()
@@ -166,11 +166,6 @@ func (c *Config) MustGetString(key string) (value string) {
 		panic(fmt.Errorf("config key %s not found", key))
 	}
 	return c.v.GetString(key)
-}
-
-// GetStringSlice gets string slice value from config
-func GetStringSlice(key string, defaultValue []string) (value []string) {
-	return c.GetStringSlice(key, defaultValue)
 }
 
 // GetStringSlice gets string slice value from config
@@ -245,17 +240,17 @@ func (c *Config) Set(key string, value interface{}) {
 func (c *Config) bindEnv(key string) {
 	envVar := key
 	if !upperCaseMatch.MatchString(key) {
-		envVar = ConfigKeyToEnvRudder(key)
+		envVar = ConfigKeyToEnv(key)
 	}
 	// bind once
-	c.envLock.RLock()
+	c.envsLock.RLock()
 	if _, ok := c.envs[key]; !ok {
-		c.envLock.RUnlock()
-		c.envLock.Lock() // don't really care about race here, setting the same value
+		c.envsLock.RUnlock()
+		c.envsLock.Lock() // don't really care about race here, setting the same value
 		c.envs[strings.ToUpper(key)] = envVar
-		c.envLock.Unlock()
+		c.envsLock.Unlock()
 	} else {
-		c.envLock.RUnlock()
+		c.envsLock.RUnlock()
 	}
 }
 
@@ -264,8 +259,8 @@ type envReplacer struct {
 }
 
 func (r *envReplacer) Replace(s string) string {
-	r.c.envLock.RLock()
-	defer r.c.envLock.RUnlock()
+	r.c.envsLock.RLock()
+	defer r.c.envsLock.RUnlock()
 	if v, ok := r.c.envs[s]; ok {
 		return v
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -34,15 +34,15 @@ var camelCaseMatch = regexp.MustCompile("([a-z0-9])([A-Z])")
 var upperCaseMatch = regexp.MustCompile("^[A-Z0-9_]+$")
 
 // default, singleton config instance
-var c *Config
+var defaultConfig *Config
 
 func init() {
-	c = New()
+	defaultConfig = New()
 }
 
 // Reset resets configuration
 func Reset() {
-	c = New()
+	defaultConfig = New()
 }
 
 // New creates a new config instance
@@ -64,7 +64,7 @@ type Config struct {
 
 // GetBool gets bool value from config
 func GetBool(key string, defaultValue bool) (value bool) {
-	return c.GetBool(key, defaultValue)
+	return defaultConfig.GetBool(key, defaultValue)
 }
 
 // GetBool gets bool value from config
@@ -79,7 +79,7 @@ func (c *Config) GetBool(key string, defaultValue bool) (value bool) {
 
 // GetInt gets int value from config
 func GetInt(key string, defaultValue int) (value int) {
-	return c.GetInt(key, defaultValue)
+	return defaultConfig.GetInt(key, defaultValue)
 }
 
 // GetInt gets int value from config
@@ -94,7 +94,7 @@ func (c *Config) GetInt(key string, defaultValue int) (value int) {
 
 // MustGetInt gets int value from config or panics if the config doesn't exist
 func MustGetInt(key string) (value int) {
-	return c.MustGetInt(key)
+	return defaultConfig.MustGetInt(key)
 }
 
 // MustGetInt gets int value from config or panics if the config doesn't exist
@@ -109,7 +109,7 @@ func (c *Config) MustGetInt(key string) (value int) {
 
 // GetInt64 gets int64 value from config
 func GetInt64(key string, defaultValue int64) (value int64) {
-	return c.GetInt64(key, defaultValue)
+	return defaultConfig.GetInt64(key, defaultValue)
 }
 
 // GetInt64 gets int64 value from config
@@ -124,7 +124,7 @@ func (c *Config) GetInt64(key string, defaultValue int64) (value int64) {
 
 // GetFloat64 gets float64 value from config
 func GetFloat64(key string, defaultValue float64) (value float64) {
-	return c.GetFloat64(key, defaultValue)
+	return defaultConfig.GetFloat64(key, defaultValue)
 }
 
 // GetFloat64 gets float64 value from config
@@ -139,7 +139,7 @@ func (c *Config) GetFloat64(key string, defaultValue float64) (value float64) {
 
 // GetString gets string value from config
 func GetString(key, defaultValue string) (value string) {
-	return c.GetString(key, defaultValue)
+	return defaultConfig.GetString(key, defaultValue)
 }
 
 // GetString gets string value from config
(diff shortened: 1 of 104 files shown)


config/config.go
```diff
@@ -1,675 +1,208 @@
...
 	"time"
 
-	"github.com/fsnotify/fsnotify"
-	"github.com/joho/godotenv"
-	"github.com/spf13/cast"
 	"github.com/spf13/viper"
 )
 
-var matchAllCap = regexp.MustCompile("([a-z0-9])([A-Z])")
+// regular expression matching lowecase letter followed by an uppercase letter
```
fracasula, line 30:
```suggestion
// regular expression matching lowercase letter followed by an uppercase letter
```

config/config.go
```diff
@@ -1,675 +1,208 @@
...
-				}
-			}
-		}
-	}
-	return hasConfigChanged
+// Config is the entry point for accessing configuration
+type Config struct {
+	vLock                  sync.RWMutex // protects reading and writing to the config (viper is not thread-safe)
+	v                      *viper.Viper
+	mapLock                sync.RWMutex // protects maps holding registered condig keys
```
fracasula, line 59:
```suggestion
	mapLock                sync.RWMutex // protects maps holding registered config keys
```

config/config.go
```diff
@@ -1,675 +1,208 @@
...
+	hotReloadableConfig    map[string][]*configValue
+	nonHotReloadableConfig map[string][]*configValue
+	envLock                sync.RWMutex // protects the envs map below
```
fracasula, line 62:
```suggestion
	envsLock                sync.RWMutex // protects the envs map below
```

config/config.go
```diff
@@ -1,675 +1,208 @@
...
-				isSet = true
-				*ptr = GetBool(key, defaultValue)
-				break
-			}
-		}
-	}
-
-	if !isSet {
-		*ptr = defaultValue
+// MustGetString gets int value from config or panics if the config doesn't exist
```
fracasula, line 101:
```suggestion
// MustGetInt gets int value from config or panics if the config doesn't exist
```

config/config.go
```diff
@@ -1,675 +1,208 @@
...
+	if !c.IsSet(key) {
+		panic(fmt.Errorf("config key %s not found", key))
 	}
-	return viper.GetString(key)
+	return c.v.GetString(key)
 }
 
-// GetStringSlice is wrapper for viper's GetStringSlice
+// GetStringSlice gets string slice value from config
 func GetStringSlice(key string, defaultValue []string) (value []string) {
```
fracasula, line 172: Unused?

config/config.go
```diff
@@ -1,675 +1,208 @@
...
-	}()
-	configVarLock.RLock()
-	defer configVarLock.RUnlock()
-	_ = checkAndHotReloadConfig(hotReloadableConfig)
-	isChanged := checkAndHotReloadConfig(nonHotReloadableConfig)
-	if isChanged && GetEnvAsBool("RESTART_ON_CONFIG_CHANGE", false) {
-		os.Exit(1)
-	}
+// New creates a new config instance
+func New() *Config {
```
fracasula, line 49: If we want to stick to the singleton pattern, do we even need this constructor? It also seems to be unused, perhaps we can remove it and just stick to the fact that we'll call the functions (e.g. `func GetBool`) instead of the methods (e.g. `func (c *Config) GetBool`).

config/config.go
```diff
@@ -679,159 +212,73 @@ func GetDuration(key string, defaultValueInTimescaleUnits int64, timeScale time.
...
+	}
+	// bind once
+	c.envLock.RLock()
+	if _, ok := c.envs[key]; !ok {
+		c.envLock.RUnlock()
+		c.envLock.Lock() // don't really care about race here, setting the same value
+		c.envs[strings.ToUpper(key)] = envVar
+		c.envLock.Unlock()
+	} else {
+		c.envLock.RUnlock()
```
fracasula, line 258: [optional] Even though you don't care about the race it can make sense to simplify this bit and just `Lock()` and `defer Unlock()` instead of mixing.

config/config_test.go
```diff
@@ -2,25 +2,246 @@ package config
...
+	tc.Set("string", "string")
+	require.Equal(t, "string", tc.MustGetString("string"), "it should return the key value")
+	func() {
+		defer func() {
+			if r := recover(); r == nil {
+				t.Errorf("The code did not panic")
+			}
+		}()
+		tc.MustGetString("other")
+	}()
```
fracasula, line 54: I think we can replace these blocks with the recover like this:

```suggestion
	require.Panics(t, func() { tc.MustGetString("other") })
```

fracasula (commented):
1st round :+1:

config/env.go
```diff
@@ -0,0 +1,35 @@
...
+
+// TODO: everything in this file should be either removed or unexported
+import (
+	"os"
+	"strconv"
+	"strings"
+)
+
+// ConfigKeyToEnvRudder as package method to get the formatted env from a given string
+func ConfigKeyToEnvRudder(s string) string {
```
fracasula, line 11: Why do we need `Rudder` in the name?

config/config.go
atzoum: the singleton pattern and package functions are there only because it is not possible at this stage to proceed without them. The goal is to move away from a singleton config instance, for the reasons we have already discussed (hard to test, parallelise, etc. )

config/env.go
atzoum: we don't! To my defence, there used to be a `ConfigKeyToEnvViper` once, but decided it was not needed after all. Will rename

config/config.go
atzoum: For a strange reason I tend to prefer this kind of (biased) optimisation, considering that in most cases the key will be already in the map, so we don't really need to be pessimistic and block competing config readers.

config/config.go
fracasula: We are not moving away from it now, in fact the constructor is not used. I'd refrain from adding code meant for future use until needed.

Anyway, can you explain what is blocking us from using `config` only during bootstrapping and removing it from each package?

In my head the ideal scenario is to stop spreading `config` in all packages and read the configuration in places like `main.go`, app handlers etcetera only. Then pass the read data as parameters when instantiating things. That way we remove the need of having something thread-safe and we decouple from config altogether. What do you think are the blockers to go towards that? Hot reloads?

config/config_test.go
```diff
@@ -2,25 +2,239 @@ package config
...
+	tc.Set("int", 0)
+	require.Equal(t, 0, tc.MustGetInt("int"), "it should return the key value")
+	func() {
+		defer func() {
+			if r := recover(); r == nil {
+				t.Errorf("The code did not panic")
+			}
+		}()
+		tc.MustGetInt("other")
+	}()
```
fracasula, line 58: I think you missed a `require.Panics()` :slightly_smiling_face:

config/env.go
```diff
@@ -0,0 +1,35 @@
...
+	"os"
+	"strconv"
+	"strings"
+)
+
+// ConfigKeyToEnv gets the env variable name from a given config key
+func ConfigKeyToEnv(s string) string {
+	if upperCaseMatch.MatchString(s) {
+		return s
+	}
```
fracasula, line 14: [optional] Some stuff doesn't seem to be covered as reported by codecov here. Just pointing it out because you did a great job with the tests and these seem fairly easy to add.

config/load.go
```diff
@@ -0,0 +1,210 @@
...
+	v.WatchConfig()
+
+	c.v = v
+}
+
+func (c *Config) onConfigChange() {
+	defer func() {
+		if r := recover(); r != nil {
+			err := fmt.Errorf("cannot update Config Variables: %v", r)
+			fmt.Println(err)
```
fracasula, line 46: How are we going to catch this when it happens?

config/env.go
atzoum: I intentionally left `GetXXX` package-level functions uncovered, since they simply delegate the call to the singleton instance and the intention is to remove them at a later stage. Will try to cover others

config/load.go
atzoum: Used this piece at its original form, just moved it to another file.
I wouldn't worry that much about it though, since config files don't change randomly in our managed environments. We do changes purposefully  and then monitor, expecting to verify that our changes have actually been applied. Failure to do so, might trigger someone to have a look at the logs :)

config/config.go
atzoum: 1. the `New` function is being used already by `config.init`, `config.Reset` & `config_tests`.
2. We need to refactor a lot of codebase to stop using the config package as we are doing now. Codebase in packages that are already using global variables for their configuration purposes. IMO this is something that needs to be done gradually and with care, not in bulk, since each case might have its own unique details we need to pay attention to. This also involves getting rid of the [various init functions](https://github.com/rudderlabs/rudder-server/blob/master/main.go#L140) that we have hanging around :)

config/load.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}()
+	c.vLock.RLock()
+	defer c.vLock.RUnlock()
+	c.mapLock.RLock()
+	defer c.mapLock.RUnlock()
+	_ = c.checkAndHotReloadConfig(c.hotReloadableConfig)
+	isChanged := c.checkAndHotReloadConfig(c.nonHotReloadableConfig)
+	if isChanged && getEnvAsBool("RESTART_ON_CONFIG_CHANGE", false) {
+		os.Exit(1)
```
atzoum, line 56: @chandumlg @cisse21 do we really still need this functionality?

cisse21: We can get rid of it... we never actually used this and should be easy to add again if needed

eventorder_test.go
```diff
@@ -98,48 +103,46 @@ func TestEventOrderGuarantee(t *testing.T) {
 		"workspaceId": workspaceID,
 	}
 	configJsonPath := workspaceConfig.CreateTempFile(t, "testdata/eventOrderTestTemplate.json", templateCtx)
-	t.Setenv("RSERVER_BACKEND_CONFIG_CONFIG_FROM_FILE", "true")
-	t.Setenv("RSERVER_BACKEND_CONFIG_CONFIG_JSONPATH", configJsonPath)
+	config.Set("BackendConfig.configFromFile", true)
```
cisse21, line 106: what is the  reason we are not using `t.Setenv `anymore ?

atzoum: We cannot consider having tests running in parallel at any point in the future if we continue using environment variables for setting up our test environments. `t.Setenv` sets an environment variable and unsets it when the test is done, but this doesn't mean that the environment variable is specific for the test, it will affect all other tests running in parrallel at the same time.

config/config.go
fracasula: Yes I understand, I'm suggesting mostly to simplify the logic. I don't see that much of a performance impact in this case so perhaps simplyfing can be good. Up to you of course :+1:

config/load.go
fracasula: Makes sense at least for now it shouldn't cause major issues given that we apply and check :+1: 

I wonder what we could do to improve these bits, perhaps structured logging might be useful in these cases cc @lvrach

config/config.go
fracasula: Agreed :+1: In regards to `1` I know but I mentiond it because those are all within the `config` package. In any case, let's tackle this in the long run, it is a sensible approach :+1:

fracasula (approved):

services/validators/envValidator.go
```diff
@@ -258,7 +257,7 @@ func CheckAndValidateWorkspaceToken() {
 
 	pkgLogger.Warn("Previous workspace token is not same as the current workspace token. Parking current jobsdb aside and creating a new one")
 
-	dbName := config.GetEnv("JOBS_DB_DB_NAME", "ubuntu")
+	dbName := config.GetString("DB.name", "ubuntu")
```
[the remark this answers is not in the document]
atzoum: yes see [here](https://github.com/rudderlabs/rudder-server/blob/fc5b2ddb651835edd545a0e9c3031971d04f7905/config/config.go#L270-L277)

eventorder_test.go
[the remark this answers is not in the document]
atzoum: Starting a service as a separate process comes with disadvantages too. Please see [this comment](https://github.com/rudderlabs/rudder-server/pull/2439#discussion_r978286582) and let me know if you wish me to elaborate further

config/config.go
```diff
@@ -1,675 +1,208 @@
...
-)
-
-// TransformKey as package method to get the formatted env from a give string
-func TransformKey(s string) string {
-	snake := matchAllCap.ReplaceAllString(s, "${1}_${2}")
-	snake = strings.ReplaceAll(snake, ".", "_")
-	return "RSERVER_" + strings.ToUpper(snake)
-}
+// default, singleton config instance
+var c *Config
```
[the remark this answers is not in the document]
atzoum: I borrowed this pattern from [viper](https://github.com/spf13/viper/blob/master/viper.go) itself. This kind of shadowing acts as a safeguard, since you wouldn't ever want to have access to the singleton instance and its methods from within the `Config`'s methods.
But if others believe that this is confusing too, we can rename it to something unique.

eventorder_test.go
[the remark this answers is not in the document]
fracasula: @atzoum it sure has disadvantages like harder to debug (setting breakpoints) and code verbosity (setting up the exec, piping output, etc...). have you come across more issues? I'd like to know because in some places it helped me with isolation.

config/config.go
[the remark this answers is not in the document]
fracasula: Personally I don't feel the need for the protection especially if done via shadowing. If possible I'd go for a more descriptive name like `defaultConfig` (or `singletonConfig`, etc...).

eventorder_test.go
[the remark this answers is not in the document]
atzoum: Practically its only advantage is isolation, everything else could be considered as a disadvantage, e.g. a non-restrictive list:
1. debugging (practically impossible)
2. extensibility (unit testing components, i.e. non-programms)
3. intuitiveness (a test case executing `go run` causes some people to say WTF)
4. configuration (expecting that the version of the go binary executing the test is the same as the go binary available in the path)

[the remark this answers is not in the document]
fracasula: Good point about the configuration. About extensibility one thing is to do an integration test with it, one thing a unit test. The latter definitely not recommended :slightly_smiling_face:

cisse21 (approved):

achettyiitr (approved):

PR #2441, opened 2022-09-15
chore: added pre_drop_table count metrics

saurav-malani:
# Description

Added metrics for `pre_drop_table` count.

## Notion Ticket

https://www.notion.so/rudderstacks/add-metrics-alert-for-pre_drop-table-count-dddf58fd73c940f39b113a54bc6f06f6

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -869,8 +869,7 @@ func (jd *HandleT) workersAndAuxSetup() {
 	jd.registerBackUpSettings()
 
 	jd.logger.Infof("Connected to %s DB", jd.tablePrefix)
-
-	jd.statPreDropTableCount = stats.DefaultStats.NewStat("jobsdb.pre_drop_tables_count", stats.GaugeType)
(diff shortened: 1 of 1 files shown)


jobsdb/jobsdb.go
```diff
@@ -863,6 +864,7 @@ func (jd *HandleT) workersAndAuxSetup() {
 
 	jd.logger.Infof("Connected to %s DB", jd.tablePrefix)
 
+	jd.statPreDropTableCount = stats.DefaultStats.NewStat("jobsdb.pre_drop_tables_count", stats.GaugeType)
```
atzoum, line 867: don't we need tags for table prefix?

chandumlg: Yes. Use NewTaggedStat with table prefix as the tag.

chandumlg (changes requested):

jobsdb/jobsdb.go
saurav-malani: done.

atzoum (approved):

chandumlg (approved):

PR #2445, opened 2022-09-15
chore: added rudder-server config metrics with version and other details as tag

saurav-malani:
## Description

Added metrics with following rudder-server config details as tag:-  

- Version
- Major
- Minor
- Patch
- Commit
- BuildDate
- BuiltBy
- GitUrl
- TransformerVersion
- DatabricksVersion

## Notion Ticket

https://www.notion.so/rudderstacks/Report-version-to-metrics-60b766e8186d48adb45621fbdee7e6be

### DIFF
--- a/main.go
+++ b/main.go
@@ -225,7 +225,8 @@ func Run(ctx context.Context) int {
 	appTypeStr := strings.ToUpper(config.GetEnv("APP_TYPE", app.EMBEDDED))
 	appHandler = apphandlers.GetAppHandler(application, appTypeStr, versionHandler)
 
-	version := versionInfo()
+	versionDetail := versionInfo()
+	stats.NewTaggedStat("rudder_server_config", stats.GaugeType, stats.Tags{"version": version, "major": major, "minor": minor, "patch": patch, "commit": commit, "buildDate": buildDate, "builtBy": builtBy, "gitUrl": gitURL, "TransformerVersion": transformer.GetVersion(), "DatabricksVersion": misc.GetDatabricksVersion()}).Gauge(1)
 	bugsnag.Configure(bugsnag.Configuration{
 		APIKey:       config.GetEnv("BUGSNAG_KEY", ""),
 		ReleaseStage: config.GetEnv("GO_ENV", "development"),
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -107,6 +107,7 @@ func Init() {
 }
 
 func versionInfo() map[string]interface{} {
+	stats.NewTaggedStat("rudder_server_config", stats.GaugeType, stats.Tags{"version": version, "Major": major, "Minor": minor, "Patch": patch, "Commit": commit, "BuildDate": buildDate, "BuiltBy": builtBy, "GitUrl": gitURL, "TransformerVersion": transformer.GetVersion(), "DatabricksVersion": misc.GetDatabricksVersion()}).Gauge(1)
```
atzoum, line 110: we only need to push this metric once during startup, right?
```suggestion
	stats.NewTaggedStat("rudder_server_config", stats.GaugeType, stats.Tags{"version": version, "major": major, "minor": minor, "patch": patch, "commit": commit, "buildDate": buildDate, "builtBy": builtBy, "gitUrl": gitURL, "TransformerVersion": transformer.GetVersion(), "DatabricksVersion": misc.GetDatabricksVersion()}).Gauge(1)
```

saurav-malani: Here, I am assuming that to change any of these we need to restart the server. So, pushing it only once.

chandumlg: The function `versionInfo` is called whenever someone hits /version endpoint. So, there is not guarantee that this will be pushed once.

main.go
```diff
@@ -107,6 +107,7 @@ func Init() {
 }
 
 func versionInfo() map[string]interface{} {
+	stats.NewTaggedStat("rudder_server_config", stats.GaugeType, stats.Tags{"version": version, "major": major, "minor": minor, "patch": patch, "commit": commit, "buildDate": buildDate, "builtBy": builtBy, "gitUrl": gitURL, "TransformerVersion": transformer.GetVersion(), "DatabricksVersion": misc.GetDatabricksVersion()}).Gauge(1)
```
atzoum, line 110: please make sure we publish this stat only once, during startup

saurav-malani: moved it to `Run()` which is called only once.

atzoum (approved):

chandumlg (approved):

PR #2446, opened 2022-09-15
chore: limit queries to ds

BonapartePC:
# Description

Two changes are made in this PR:
1. Add stats for the number of DS hits made per query
2. Limit the number of DS hits per query

## Notion Ticket

https://www.notion.so/rudderstacks/Add-stats-for-the-number-of-DS-per-query-41c957dab7eb414397cd7653a7c65701
https://www.notion.so/rudderstacks/Limit-the-number-of-DS-queries-per-jobsdb-Get-4db41289310940a8b49ae2af7b36f6f3

## Security


enterprise/replay/replay.go
```diff
@@ -152,6 +154,7 @@ func (handle *Handler) Setup(ctx context.Context, dumpsLoader *dumpsLoaderHandle
 	handle.noOfWorkers = config.GetEnvAsInt("WORKERS_PER_SOURCE", 4)
 	handle.dumpsLoader = dumpsLoader
 	handle.dbReadSize = config.GetEnvAsInt("DB_READ_SIZE", 10)
+	handle.dsQueryLimit = config.GetEnvAsInt("DS_QUERY_LIMIT", 10)
```
atzoum, line 157: imposing a limit is not something we would like to do normally, thus let's start with no limit: 0

jobsdb/jobsdb.go
```diff
@@ -105,6 +105,10 @@ type GetQueryParamsT struct {
 	// Limit the total number of jobs.
 	// A value less than or equal to zero will return no results
 	JobsLimit int
+	// Limit the number of Datasets to query.
+	// A value less than or equal to zero will disable this limit (no limit),
+	// only values greater than zero are considered as valid limits.
+	DSLimit int
```
atzoum, line 111: let's not expose this and use the ds limit internally in jobsdb's unexported methods

BonapartePC: If we don't expose ds limit, we won't be able to set value in processor, router and batch router according to their respective env right? Example: https://github.com/rudderlabs/rudder-server/pull/2446/files#diff-63264a27291123a1413bf5e86a8dcd368b9214681cafd30426fafbdaa2a182faR2203

atzoum: Each component uses a different jobsdb so we may set different limits during jobsdb initialisation using another `OptsFunc`.

atzoum: The idea behind it is that jobsdb shouldn't expose anything in its public API with respect to datasets. Datasets are an internal detail, they control how jobs are stored internally, and callers should worry about them when using the API.

jobsdb/unionQuery.go
```diff
@@ -89,37 +88,35 @@ func (mj *MultiTenantHandleT) GetAllJobs(ctx context.Context, workspaceCount map
...
 		outJobs = append(outJobs, jobs...)
-		if len(jobs) > 0 {
-			tablesQueried++
-		}
-
-		if len(workspaceCount) == 0 {
-			break
-		}
-		if tablesQueried >= maxDSQuerySize {
-			break
```
atzoum, line 106: Propose to leave the old logic intact. The intention is not to change unionQuery's behaviour

atzoum:
@BonapartePC let's add a test verifying this new jobsDB configuration option

jobsdb/unionQuery.go
```diff
@@ -61,7 +61,7 @@ func (*MultiTenantHandleT) getSingleWorkspaceQueryString(workspace string, jobsL
 
 // All Jobs
 
-func (mj *MultiTenantHandleT) GetAllJobs(ctx context.Context, workspaceCount map[string]int, params GetQueryParamsT, maxDSQuerySize int) ([]*JobT, error) { // skipcq: CRT-P0003
```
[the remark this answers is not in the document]
atzoum: @cisse21 do we need this default limit of 10 datasets here, or we may use a single config option across all query strategies?

jobsdb/jobsdb.go
```diff
@@ -4328,11 +4343,19 @@ func (jd *HandleT) getUnprocessed(ctx context.Context, params GetQueryParamsT) (
 	}
 
 	var completeUnprocessedJobs JobsResult
+	dsQueryCount := 0
+	dsLimit := jd.dsLimit
 	for _, ds := range dsList {
-		unprocessedJobs, err := jd.getUnprocessedJobsDS(ctx, ds, true, params)
+		if dsLimit > 0 && dsQueryCount >= dsLimit {
```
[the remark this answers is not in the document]
atzoum: the default is no limit

jobsdb/unionQuery.go
BonapartePC: maxDSQuerySize is the same as DSLimit. So I removed it. Should we support both of them?

jobsdb/unionQuery.go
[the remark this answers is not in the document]
cisse21: We already enforce a limit of 10 when we do a union query to reduce the load on Postgres. Would like that to be default while using unless we have a strong concern

jobsdb/unionQuery.go
atzoum: according to @cisse21 response [here](https://github.com/rudderlabs/rudder-server/pull/2446#discussion_r974485327), it is still desirable to use a default limit of 10 for unionQuery, thus we shouldn't consider the two limits (`maxDSQuerySize` & `dsLimit`) as same.

BonapartePC: In that case, we should keep the default limit of dsLimit as 10 for unionQuery right? Or we should get rid of `dsLimit` altogether

atzoum: let's treat them as different things, see [here](https://github.com/rudderlabs/rudder-server/pull/2446#discussion_r974068382)

atzoum: @cisse21 they may look the same but they are practically not, since we don't want to enforce a default limit of 10 datasets for queries other than unionDS.
IMO by default there should be no limit to the number of datasets that we can query as this limit might hinder proper job pickup and processing.
I see this limit only as a tool to use when things are going wrong and there is a lot of pressure to the database, where we can temporarily impose such a limit, if necessary, but still we need to revert it after the incident is resolved.

atzoum (approved):
thank you @BonapartePC for addressing all comments!

jobsdb/jobsdb.go
```diff
@@ -4536,11 +4555,22 @@ func (jd *HandleT) GetProcessed(ctx context.Context, params GetQueryParamsT) (Jo
...
+			break
+		}
+		processedJobs, dsHit, err := jd.getProcessedJobsDS(ctx, ds, false, params)
 		if err != nil {
 			return JobsResult{}, err
 		}
+		if dsHit {
+			dsQueryCount++
+		}
```
lvrach, line 4573: We only increase `dsQueryCount` when a dsHit takes place, and the decision to stop is based on that. 

What if dsQueryCount:
1.  was measured regardless of queries hit ?
2. we stopped early once the dsQueryCount was reached and we had non-zero data.

Letting 

The current approach makes sense, but I feel it doesn't penalise queries early enough when we have sparse entries.

Assume we have 200 tables and data are only on in table 3:
* first one
* one in the middle 
* one in the end 

Now, this algorithm will scan all 200 tables twice until the cache kicks in and avoid the lookup in some tables.

With the approaches I am suggesting, it will require 5 queries for the tables to be scanned, but those queries will be bounded.

atzoum: @lvrach a `dsHit` happens whenever we query the dataset, regardless if the query returns results or not.
The only occasions where `dsHit` will be returned as `false` are:
1. If we have a cache hit (no jobs & no query)
2. An error is returned for whatever reason

cisse21 (approved):

jobsdb/jobsdb.go
lvrach: Got it, my bad it wasn't clear from the diff. 

A small comment about what `getProcessedJobsDS` returns would be useful here.

PR #2453, opened 2022-09-18
chore: support for opt out from reporting sample event and response

chandumlg:
# Description

Adds data retention support.
a. Filter columns before sending reports to reporting service based on user settings.
b. Use storage settings from workspace config.

Data retention notion doc: https://www.notion.so/rudderstacks/Data-Retention-8293118d1665485cbf11fee62b60b908

## Notion Ticket

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=0c07cea477ad4f55812f10fda68589f7&pm=s

## Security

### DIFF
--- a/config/backend-config/backend-config.go
+++ b/config/backend-config/backend-config.go
@@ -51,6 +51,7 @@ type workspaceConfig interface {
 	GetWorkspaceIDForWriteKey(string) string
 	GetWorkspaceIDForSourceID(string) string
 	GetWorkspaceLibrariesForWorkspaceID(string) LibrariesT
+	GetDataRetentionSettingsForWorkspaceID(string) DataRetention
 }
 
 type BackendConfig interface {
--- a/config/backend-config/backend-config.go
+++ b/config/backend-config/backend-config.go
@@ -51,7 +51,7 @@ type workspaceConfig interface {
 	GetWorkspaceIDForWriteKey(string) string
 	GetWorkspaceIDForSourceID(string) string
 	GetWorkspaceLibrariesForWorkspaceID(string) LibrariesT
-	GetDataRetentionSettingsForWorkspaceID(string) DataRetention
+	DataRetentionSettings(workspaceID string) DataRetention
 }
 
 type BackendConfig interface {
--- a/config/backend-config/multitenant_workspaces.go
+++ b/config/backend-config/multitenant_workspaces.go
@@ -25,14 +25,15 @@ type WorkspacesT struct {
 }
 
 type multiTenantWorkspacesConfig struct {
-	Token                     string
-	configBackendURL          *url.URL
-	configEnvHandler          types.ConfigEnvI
-	cpRouterURL               string
-	writeKeyToWorkspaceIDMap  map[string]string
-	sourceToWorkspaceIDMap    map[string]string
-	workspaceIDToLibrariesMap map[string]LibrariesT
-	workspaceWriteKeysMapLock sync.RWMutex
+	Token                         string
+	configBackendURL              *url.URL
+	configEnvHandler              types.ConfigEnvI
+	cpRouterURL                   string
+	writeKeyToWorkspaceIDMap      map[string]string
+	sourceToWorkspaceIDMap        map[string]string
+	workspaceIDToLibrariesMap     map[string]LibrariesT
+	workspaceIDToDataRetentionMap map[string]DataRetention
+	workspaceWriteKeysMapLock     sync.RWMutex
 }
 
 func (wc *multiTenantWorkspacesConfig) SetUp() error {
@@ -82,6 +83,16 @@ func (wc *multiTenantWorkspacesConfig) GetWorkspaceLibrariesForWorkspaceID(works
 		return workspaceLibraries
 	}
 	return LibrariesT{}
+}
+
+func (wc *multiTenantWorkspacesConfig) GetDataRetentionSettingsForWorkspaceID(workspaceID string) DataRetention {
+	wc.workspaceWriteKeysMapLock.RLock()
+	defer wc.workspaceWriteKeysMapLock.RUnlock()
+
+	if workspaceDataRetentionSettings, ok := wc.workspaceIDToDataRetentionMap[workspaceID]; ok {
+		return workspaceDataRetentionSettings
+	}
+	return DataRetention{}
 }
 
 // Get returns sources from the workspace
@@ -134,6 +145,7 @@ func (wc *multiTenantWorkspacesConfig) getFromAPI(ctx context.Context, _ string)
 	writeKeyToWorkspaceIDMap := make(map[string]string)
 	sourceToWorkspaceIDMap := make(map[string]string)
 	workspaceIDToLibrariesMap := make(map[string]LibrariesT)
+	workspaceIDToDataRetentionMap := make(map[string]DataRetention)
 	sourcesJSON := ConfigT{}
 	sourcesJSON.Sources = make([]SourceT, 0)
 	for workspaceID, workspaceConfig := range workspaces.WorkspaceSourcesMap {
@@ -141,6 +153,7 @@ func (wc *multiTenantWorkspacesConfig) getFromAPI(ctx context.Context, _ string)
 			writeKeyToWorkspaceIDMap[source.WriteKey] = workspaceID
 			sourceToWorkspaceIDMap[source.ID] = workspaceID
 			workspaceIDToLibrariesMap[workspaceID] = workspaceConfig.Libraries
+			workspaceIDToDataRetentionMap[workspaceID] = workspaceConfig.Settings.DataRetention
 		}
 		sourcesJSON.Sources = append(sourcesJSON.Sources, workspaceConfig.Sources...)
 	}
@@ -151,6 +164,7 @@ func (wc *multiTenantWorkspacesConfig) getFromAPI(ctx context.Context, _ string)
 	wc.writeKeyToWorkspaceIDMap = writeKeyToWorkspaceIDMap
 	wc.sourceToWorkspaceIDMap = sourceToWorkspaceIDMap
 	wc.workspaceIDToLibrariesMap = workspaceIDToLibrariesMap
+	wc.workspaceIDToDataRetentionMap = workspaceIDToDataRetentionMap
 	wc.workspaceWriteKeysMapLock.Unlock()
 
 	return sourcesJSON, nil
--- a/config/backend-config/multitenant_workspaces.go
+++ b/config/backend-config/multitenant_workspaces.go
@@ -85,7 +85,7 @@ func (wc *multiTenantWorkspacesConfig) GetWorkspaceLibrariesForWorkspaceID(works
 	return LibrariesT{}
 }
 
-func (wc *multiTenantWorkspacesConfig) GetDataRetentionSettingsForWorkspaceID(workspaceID string) DataRetention {
+func (wc *multiTenantWorkspacesConfig) DataRetentionSettings(workspaceID string) DataRetention {
 	wc.workspaceWriteKeysMapLock.RLock()
 	defer wc.workspaceWriteKeysMapLock.RUnlock()
 
--- a/config/backend-config/namespace_config.go
+++ b/config/backend-config/namespace_config.go
@@ -17,12 +17,13 @@ import (
 )
 
 type namespaceConfig struct {
-	configEnvHandler          types.ConfigEnvI
-	mapsMutex                 sync.RWMutex
-	writeKeyToWorkspaceIDMap  map[string]string
-	workspaceIDToLibrariesMap map[string]LibrariesT
-	sourceToWorkspaceIDMap    map[string]string
-	cpRouterURL               string
+	configEnvHandler              types.ConfigEnvI
+	mapsMutex                     sync.RWMutex
+	writeKeyToWorkspaceIDMap      map[string]string
+	workspaceIDToLibrariesMap     map[string]LibrariesT
+	workspaceIDToDataRetentionMap map[string]DataRetention
+	sourceToWorkspaceIDMap        map[string]string
+	cpRouterURL                   string
 
 	Logger logger.LoggerI
 	Client *http.Client
@@ -102,6 +103,17 @@ func (nc *namespaceConfig) GetWorkspaceLibrariesForWorkspaceID(workspaceID strin
 	return LibrariesT{}
 }
 
+// GetDataRetentionSettingsForWorkspaceID returns dataRetention settings for workspaceID
+func (nc *namespaceConfig) GetDataRetentionSettingsForWorkspaceID(workspaceID string) DataRetention {
+	nc.mapsMutex.RLock()
+	defer nc.mapsMutex.RUnlock()
+
+	if workspaceDataRetentionSettings, ok := nc.workspaceIDToDataRetentionMap[workspaceID]; ok {
+		return workspaceDataRetentionSettings
+	}
+	return DataRetention{}
+}
+
 // Get returns sources from the workspace
 func (nc *namespaceConfig) Get(ctx context.Context, workspaces string) (ConfigT, error) {
 	return nc.getFromAPI(ctx, workspaces)
@@ -147,6 +159,7 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context, _ string) (ConfigT, e
 	writeKeyToWorkspaceIDMap := make(map[string]string)
 	sourceToWorkspaceIDMap := make(map[string]string)
 	workspaceIDToLibrariesMap := make(map[string]LibrariesT)
+	workspaceIDToDataRetentionMap := make(map[string]DataRetention)
 	sourcesJSON := ConfigT{}
 	sourcesJSON.Sources = make([]SourceT, 0)
 	for workspaceID, nc := range workspaces.WorkspaceSourcesMap {
@@ -155,6 +168,7 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context, _ string) (ConfigT, e
 			writeKeyToWorkspaceIDMap[source.WriteKey] = workspaceID
 			sourceToWorkspaceIDMap[source.ID] = workspaceID
 			workspaceIDToLibrariesMap[workspaceID] = nc.Libraries
+			workspaceIDToDataRetentionMap[workspaceID] = nc.Settings.DataRetention
 		}
 		sourcesJSON.Sources = append(sourcesJSON.Sources, nc.Sources...)
 	}
@@ -165,6 +179,7 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context, _ string) (ConfigT, e
 	nc.writeKeyToWorkspaceIDMap = writeKeyToWorkspaceIDMap
 	nc.sourceToWorkspaceIDMap = sourceToWorkspaceIDMap
 	nc.workspaceIDToLibrariesMap = workspaceIDToLibrariesMap
+	nc.workspaceIDToDataRetentionMap = workspaceIDToDataRetentionMap
 	nc.mapsMutex.Unlock()
 
 	return sourcesJSON, nil
--- a/config/backend-config/namespace_config.go
+++ b/config/backend-config/namespace_config.go
@@ -103,8 +103,8 @@ func (nc *namespaceConfig) GetWorkspaceLibrariesForWorkspaceID(workspaceID strin
 	return LibrariesT{}
 }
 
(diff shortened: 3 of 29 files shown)


config/backend-config/types.go
```diff
@@ -87,13 +87,29 @@ type ConfigT struct {
...
 	URL      string          `json:"url"`
 	Services map[string]bool `json:"services"`
 }
 
+type Settings struct {
+	DataRetention DataRetention `json:"dataRetention"`
+}
+
+type DataRetention struct {
+	EnableReportingPii     bool          `json:"enableReportingPii"`
```
lvrach, line 103:
```suggestion
	EnableReportingPII     bool          `json:"enableReportingPii"`
```

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

config/backend-config/backend-config.go
```diff
@@ -51,6 +51,7 @@ type workspaceConfig interface {
 	GetWorkspaceIDForWriteKey(string) string
 	GetWorkspaceIDForSourceID(string) string
 	GetWorkspaceLibrariesForWorkspaceID(string) LibrariesT
+	GetDataRetentionSettingsForWorkspaceID(string) DataRetention
```
lvrach, line 54: We can make our methods a bit more readable and idiomatic by:
* remove `Get`
* `ForWorkspaceID` we can assume workspaceID is the default means of lookup and only specify methods that don't follow the standard, example`ForWriteKey`.  The name param can provide additional documentation here.

```suggestion
	DataRetentionSettings(workspaceID string) DataRetention
```

For consistency we should rename the other methods as well.

cisse21 (approved):

config/backend-config/backend-config.go
lvrach: I am also wondering whether we need a new method/map for  `DataRetentionSettings` or we can just use `ConfigT`.

On the one hand, using specific data structs reduces the coupling for the components that depend on the config; on the other hand, the config becomes polluted with all the specific components we want to configure.

Without getting outside of the scope of this PR, I would suggest merging:
```go
	wc.workspaceIDToLibrariesMap = workspaceIDToLibrariesMap
	wc.workspaceIDToDataRetentionMap = workspaceIDToDataRetentionMap
 ```
 into a
 ```go
  map[string]ConfigT
  ```
  
  While maintaining the existing method signatures.

**Review by lvrach — COMMENTED** [review]
I have a few comments specific to backend config changes. I have not review the other components yet.

File: enterprise/reporting/reporting.go [hunk 3]
```diff
@@ -432,7 +432,7 @@ func isMetricPosted(status int) bool {
 }
 
 func getPIIColumnsToExclude() []string {
-	piiColumnsToExclude := strings.Split(config.GetString("REPORTING_PII_COLUMNS_TO_EXCLUDE", ""), ",")
+	piiColumnsToExclude := strings.Split(config.GetString("Reporting.piiColumnstoExclude", "sample_event,sample_response"), ",")
```
lvrach, line 435: Was this used for debugging, or should it be committed in the code?
```go
"sample_event,sample_response"
```

services/filemanager/filemanager.go
```diff
@@ -87,8 +87,27 @@ func (factory *FileManagerFactoryT) New(settings *SettingsT) (FileManager, error
...
+func GetBackupStorageConfig(ctx context.Context) (string, map[string]interface{}) {
+	backendconfig.DefaultBackendConfig.WaitForConfig(ctx)
+	// TODO: Adapt for multi-tenancy
+	workspaceID := backendconfig.DefaultBackendConfig.GetWorkspaceIDForWriteKey("")
+	useRudderStorage := backendconfig.DefaultBackendConfig.GetDataRetentionSettingsForWorkspaceID(workspaceID).UseRudderServerStorage
+
+	if useRudderStorage {
+		return getProviderAndConfigForBackupsFromEnv()
+	} else {
+		return getProviderAndConfigForBackupsFromWorkspaceSettings(backendconfig.DefaultBackendConfig.GetDataRetentionSettingsForWorkspaceID(workspaceID).StorageBucket)
```
fracasula, line 100: Why getting (thus Rlocking) twice here?

```suggestion
	retentionSettings := backendconfig.DefaultBackendConfig.GetDataRetentionSettingsForWorkspaceID(workspaceID)
	useRudderStorage := retentionSettings.UseRudderServerStorage

	if useRudderStorage {
		return getProviderAndConfigForBackupsFromEnv()
	} else {
		return getProviderAndConfigForBackupsFromWorkspaceSettings(retentionSettings.StorageBucket)
```

services/filemanager/filemanager.go
```diff
@@ -87,8 +87,27 @@ func (factory *FileManagerFactoryT) New(settings *SettingsT) (FileManager, error
...
+	}
+}
+
+// getProviderAndConfigForBackupsFromWorkspaceSettings returns the provider config from the workspace settings
+func getProviderAndConfigForBackupsFromWorkspaceSettings(bucket backendconfig.StorageBucket) (string, map[string]interface{}) {
```
fracasula, line 105: Is this function needed? Why not doing this in `GetBackupStorageConfig`:

```go
    return retentionSettings.StorageBucket.Provider, retentionSettings.StorageBucket.Config
```

services/filemanager/filemanager.go
```diff
@@ -87,8 +87,27 @@ func (factory *FileManagerFactoryT) New(settings *SettingsT) (FileManager, error
...
 }
 
-// GetProviderConfigForBackupsFromEnv returns the provider config
-func GetProviderConfigForBackupsFromEnv(ctx context.Context) map[string]interface{} {
+// GetBackupStorageConfig returns the backup storage config based on the user settings
+func GetBackupStorageConfig(ctx context.Context) (string, map[string]interface{}) {
+	backendconfig.DefaultBackendConfig.WaitForConfig(ctx)
+	// TODO: Adapt for multi-tenancy
+	workspaceID := backendconfig.DefaultBackendConfig.GetWorkspaceIDForWriteKey("")
+	useRudderStorage := backendconfig.DefaultBackendConfig.GetDataRetentionSettingsForWorkspaceID(workspaceID).UseRudderServerStorage
```
lvrach, line 95: What if the control-plane runs on an older version, were this configuration is not defined?

Also, what if a bug is introduced in control-plane and this configuration is missing? We will interpret it as false, is this a safe option here?

Otherwise, we might want to modify 
```
GetDataRetentionSettingsForWorkspaceID(workspaceID)
```
to return an error if the setting is missing. To detect if Setting is missing we need to use a pointer when parsing the JSON.

lvrach (commented):
Added some concerns regarding error handling and forward compatibility

config/backend-config/namespace_config.go
```diff
@@ -102,6 +103,17 @@ func (nc *namespaceConfig) GetWorkspaceLibrariesForWorkspaceID(workspaceID strin
...
 
+// GetDataRetentionSettingsForWorkspaceID returns dataRetention settings for workspaceID
+func (nc *namespaceConfig) GetDataRetentionSettingsForWorkspaceID(workspaceID string) DataRetention {
+	nc.mapsMutex.RLock()
+	defer nc.mapsMutex.RUnlock()
+
+	if workspaceDataRetentionSettings, ok := nc.workspaceIDToDataRetentionMap[workspaceID]; ok {
+		return workspaceDataRetentionSettings
+	}
+	return DataRetention{}
```
fracasula, line 114: Not sure it's a good idea to just return `DataRetention{}` here. Wouldn't it be more clear to return `(DataRetention, bool)` instead so that we would be sure that the setting isn't there?

If the zero values of the struct are always enough then I guess it can be OK although it might make things a bit more difficult if those defaults were to change. WDYT?

config/backend-config/namespace_config_test.go
```diff
@@ -99,6 +99,12 @@ func Test_Namespace_Get(t *testing.T) {
 		)
 	}
 
+	t.Log("correct workspaceID to dataRetention settings mapping")
+	w1DataRetention := DataRetention{EnableReportingPii: true}
+	w2DataRetention := DataRetention{EnableReportingPii: false}
+	require.Equal(t, w1DataRetention, client.GetDataRetentionSettingsForWorkspaceID(workspaceID1))
+	require.Equal(t, w2DataRetention, client.GetDataRetentionSettingsForWorkspaceID(workspaceID2))
```
fracasula, line 106: Given your implementation this would also be true:

```go
require.Equal(t, w2DataRetention, client.GetDataRetentionSettingsForWorkspaceID("NON-EXISTENT"))
```

Again, it might not be a problem after all, just pointing it out.

enterprise/reporting/noop.go
```diff
@@ -10,7 +10,7 @@ import (
 // NOOP reporting implementation that does nothing
 type NOOP struct{}
 
-func (n *NOOP) Report(metrics []*types.PUReportedMetric, txn *sql.Tx) {
+func (n *NOOP) Report(ctx context.Context, metrics []*types.PUReportedMetric, txn *sql.Tx) {
```
fracasula, line 13:
```suggestion
func (*NOOP) Report(_ context.Context, _ []*types.PUReportedMetric, _ *sql.Tx) {
```

Possibly DeepSource might be complaining about the method receiver being there but not being used anyway.

enterprise/reporting/reporting.go
```diff
@@ -445,13 +445,17 @@ func transformMetricForPII(metric types.PUReportedMetric, piiColumns []string) t
 			metric.StatusDetail.SampleEvent = []byte(`{}`)
 		} else if col == "sample_response" {
 			metric.StatusDetail.SampleResponse = ""
+		} else if col == "event_name" {
+			metric.StatusDetail.EventName = ""
+		} else if col == "event_type" {
+			metric.StatusDetail.EventType = ""
```
fracasula, line 451: [optional] It looks like a good candidate for a `switch`.

router/batchrouter/batchrouter.go
```diff
@@ -1366,7 +1366,7 @@ func (brt *HandleT) setJobStatus(batchJobs *BatchJobsT, isWarehouse bool, errOcc
 				router.GetFailedEventsManager().SaveFailedRecordIDs(jobRunIDAbortedEventsMap, tx.Tx())
 			}
 			if brt.reporting != nil && brt.reportingEnabled {
-				brt.reporting.Report(reportMetrics, tx.Tx())
+				brt.reporting.Report(context.TODO(), reportMetrics, tx.Tx())
```
fracasula, line 1369: Can't we use the above `ctx` here?

services/filemanager/filemanager.go
fracasula: [Here](https://github.com/rudderlabs/rudder-server/pull/2453#discussion_r973991488) I proposed `(DataRetention, bool)` instead of the pointer as it seemed easy enough to copy across the stack instead of spreading pointers. The point is the same anyway, I'm cool with both options.

services/filemanager/filemanager.go
```diff
@@ -101,7 +120,6 @@ func GetProviderConfigForBackupsFromEnv(ctx context.Context) map[string]interfac
 		providerConfig["regionHint"] = config.GetEnv("AWS_S3_REGION_HINT", "us-east-1")
 		providerConfig["iamRoleArn"] = config.GetEnv("BACKUP_IAM_ROLE_ARN", "")
 		if providerConfig["iamRoleArn"] != "" {
-			backendconfig.DefaultBackendConfig.WaitForConfig(ctx)
```
koladilip, line 104: What if in the future this method is called when backendConfig is not ready?

koladilip (approved):
Looks good

enterprise/reporting/reporting.go
```diff
@@ -445,13 +445,17 @@ func transformMetricForPII(metric types.PUReportedMetric, piiColumns []string) t
 			metric.StatusDetail.SampleEvent = []byte(`{}`)
 		} else if col == "sample_response" {
 			metric.StatusDetail.SampleResponse = ""
+		} else if col == "event_name" {
```
cisse21, line 448: Screenshot 2022-09-19 at 3 23 42 PM

config/backend-config/multitenant_workspaces.go
```diff
@@ -134,13 +145,15 @@ func (wc *multiTenantWorkspacesConfig) getFromAPI(ctx context.Context, _ string)
...
 	workspaceIDToLibrariesMap := make(map[string]LibrariesT)
+	workspaceIDToDataRetentionMap := make(map[string]DataRetention)
 	sourcesJSON := ConfigT{}
 	sourcesJSON.Sources = make([]SourceT, 0)
 	for workspaceID, workspaceConfig := range workspaces.WorkspaceSourcesMap {
 		for _, source := range workspaceConfig.Sources {
 			writeKeyToWorkspaceIDMap[source.WriteKey] = workspaceID
 			sourceToWorkspaceIDMap[source.ID] = workspaceID
 			workspaceIDToLibrariesMap[workspaceID] = workspaceConfig.Libraries
+			workspaceIDToDataRetentionMap[workspaceID] = workspaceConfig.Settings.DataRetention
```
atzoum, line 156: @Sidddddarth with respect to changes in #2225, this kind of initialization only happens upon successful retrieval of the config by the http client, whereas we would need these to be initialized if the config was cached too, correct?

That being said, I believe we should initialize them outside this method, as a second step, e.g. maybe expose a `workspaceConfig.ConfigUpdated` method for that purpose to be called by backend-config.

config/backend-config/types.go
```diff
@@ -87,13 +87,29 @@ type ConfigT struct {
...
+	UseRudderServerStorage bool          `json:"useRudderServerStorage"`
```
atzoum, line 104: If the default value for both is `false` we are fine. But I think that the default value for UseRudderServerStorage is true, correct? Maybe it would be wiser to reverse this config option and change its name accordingly

services/filemanager/filemanager.go
```diff
@@ -87,8 +87,27 @@ func (factory *FileManagerFactoryT) New(settings *SettingsT) (FileManager, error
...
+	if useRudderStorage {
+		return getProviderAndConfigForBackupsFromEnv()
+	} else {
+		return getProviderAndConfigForBackupsFromWorkspaceSettings(backendconfig.DefaultBackendConfig.GetDataRetentionSettingsForWorkspaceID(workspaceID).StorageBucket)
+	}
+}
+
+// getProviderAndConfigForBackupsFromWorkspaceSettings returns the provider config from the workspace settings
+func getProviderAndConfigForBackupsFromWorkspaceSettings(bucket backendconfig.StorageBucket) (string, map[string]interface{}) {
+	return bucket.Provider, bucket.Config
```
atzoum, line 106: Are we confident all required parameters will be there and with the appropriate keys?
E.g. @koladilip recently added support for IAM roles in S3 with #2386 and is using the `workspaceID` as a value for `externalId` parameter.

services/filemanager/filemanager.go
atzoum: we could inline both methods in the calling method (`GetBackupStorageConfig`)

services/filemanager/filemanager.go
```diff
@@ -87,8 +87,27 @@ func (factory *FileManagerFactoryT) New(settings *SettingsT) (FileManager, error
 	return nil, fmt.Errorf("%w: %s", rterror.InvalidServiceProvider, settings.Provider)
 }
 
-// GetProviderConfigForBackupsFromEnv returns the provider config
-func GetProviderConfigForBackupsFromEnv(ctx context.Context) map[string]interface{} {
+// GetBackupStorageConfig returns the backup storage config based on the user settings
+func GetBackupStorageConfig(ctx context.Context) (string, map[string]interface{}) {
```
atzoum, line 91: An integration test with a mock CP service would be nice here to verify 2 possible scenarios:
1. config retrieved through environment variables
2. config retrieved through control plane

config/backend-config/types.go
chandumlg: done

services/filemanager/filemanager.go
chandumlg: Thought so. Anyways removed now.

services/filemanager/filemanager.go
chandumlg: done

config/backend-config/backend-config.go
chandumlg: Addressed first comment. Second one, we can do all of them together later.

config/backend-config/namespace_config_test.go
chandumlg: Added this check also.

router/batchrouter/batchrouter.go
chandumlg: Thanks. Missed it. Fixed.

enterprise/reporting/reporting.go
chandumlg: Left them in the code in case we may need to filter event_name & event_type in the future. Right now, we are not. https://rudderlabs.slack.com/archives/C03MJ8AP78E/p1661884205170179?thread_ts=1661867090.472649&cid=C03MJ8AP78E

config/backend-config/multitenant_workspaces.go
Sidddddarth:
```
type ConfigT struct {
	EnableMetrics   bool            `json:"enableMetrics"`
	WorkspaceID     string          `json:"workspaceId"`
	Sources         []SourceT       `json:"sources"`
	Libraries       LibrariesT      `json:"libraries"`
	ConnectionFlags ConnectionFlags `json:"flags"`
}
```

We could change the `Libraries` field to a map(`map[workspaceID]LibrariesT`).
This would remove the need for `GetWorkspaceLibrariesForWorkspaceID(workspaceID)` and other similar functions(ex: `GetWorkspaceIDForSourceID` and `GetWorkspaceIDForWriteKey`) that look redundant really.

That would also address the initialisation problem you describe above.

config/config.yaml
```diff
@@ -250,3 +250,5 @@ PgNotifier:
   retriggerCount: 500
   trackBatchInterval: 2s
   maxAttempt: 3
+Reporting:
+  piiColumnstoExclude: sample_event,sample_response
```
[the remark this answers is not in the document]
chandumlg: Just more control on the columns we would like to filter.

We are not sure yet if event_name & event_type needs to be filtered. For now, we are not. If we decide to filter them too, adding them here would suffice.

enterprise/reporting/reporting.go
chandumlg: done

services/filemanager/filemanager.go
chandumlg: Required parameters might be present. But some optional parameters might miss. Do you think enriching on the server make sense?

services/filemanager/filemanager.go
chandumlg: [This idea](https://github.com/rudderlabs/rudder-server/pull/2453#discussion_r974265156) seems better to me instead of checking for nested nils. We can work with control plane team to invert the names. What do you guys think?

services/filemanager/filemanager.go
atzoum: I don't really know, but I believe we shouldn't release this until we can properly support the same features regardless if we are using control-plane or environment-based config

config/backend-config/multitenant_workspaces.go
atzoum: It is not only about libraries, it is about the following exposed interface methods:
https://github.com/rudderlabs/rudder-server/blob/master/config/backend-config/backend-config.go#L51-L53

Changing the struct's fields shouldn't be an option as this would break backwards compatibility of control plane's response schema.

services/filemanager/filemanager.go
fracasula: It's not a bad idea and I'm OK with that, but personally I find it less robust as it might not be immediate for devs to understand the behaviour. On the other hand `(DataRetention, bool)` is unequivocally clear. It gives you the freedom to rename things as you want and don't depend on the control plane. Also you don't rely on zero values that sometimes can be missed or not taken into consideration. If you guys prefer zero-values or pointers it is also cool of course, just expressing a preference.

config/backend-config/multitenant_workspaces.go
Sidddddarth: > following exposed interface methods

Yes. I'm going thru the calls to these functions, and they're redundant really. We could do without them too.

PR #2457, opened 2022-09-19
feat: changes to support variadic keys for deletion from API dest.

saurav-malani:
# Description
>Currently, when we sent request to transformer to further call down stream destination to perform deletion, the `userAttributes` section of the request can only have `userID` (required), `phone` (optional) & `email` (optional). We are modifying it to support variadic keys with only `userID` as required field.

## Notion Ticket

https://www.notion.so/rudderstacks/data-regulation-worker-support-key-value-user-attributes-1771ea8b38ac43d1a09e6ada68f2ce83

### DIFF
--- a/regulation-worker/internal/client/client.go
+++ b/regulation-worker/internal/client/client.go
@@ -164,14 +164,14 @@ func prepURL(url string, params ...string) string {
 }
 
 func mapPayloadToJob(wjs jobSchema, workspaceID string) (model.Job, error) {
-	usrAttribute := make([]model.UserAttribute, len(wjs.UserAttributes))
+	usrAttribute := make([]model.User, len(wjs.UserAttributes))
 	for i, usrAttr := range wjs.UserAttributes {
-		usrAttribute[i].Opts = make(map[string]string)
+		usrAttribute[i].Attributes = make(map[string]string)
 		for key, value := range usrAttr {
 			if key == "userId" {
-				usrAttribute[i].UserID = value
+				usrAttribute[i].ID = value
 			} else {
-				usrAttribute[i].Opts[key] = value
+				usrAttribute[i].Attributes[key] = value
 			}
 		}
 	}
--- a/regulation-worker/internal/client/client.go
+++ b/regulation-worker/internal/client/client.go
@@ -182,10 +182,10 @@ func mapPayloadToJob(wjs jobSchema, workspaceID string) (model.Job, error) {
 	}
 
 	return model.Job{
-		ID:             jobID,
-		WorkspaceID:    workspaceID,
-		DestinationID:  wjs.DestinationID,
-		Status:         model.JobStatusRunning,
-		UserAttributes: usrAttribute,
+		ID:            jobID,
+		WorkspaceID:   workspaceID,
+		DestinationID: wjs.DestinationID,
+		Status:        model.JobStatusRunning,
+		Users:         usrAttribute,
 	}, nil
 }
--- a/regulation-worker/internal/client/client.go
+++ b/regulation-worker/internal/client/client.go
@@ -165,11 +165,14 @@ func prepURL(url string, params ...string) string {
 
 func mapPayloadToJob(wjs jobSchema, workspaceID string) (model.Job, error) {
 	usrAttribute := make([]model.UserAttribute, len(wjs.UserAttributes))
-	for i := 0; i < len(wjs.UserAttributes); i++ {
-		usrAttribute[i] = model.UserAttribute{
-			UserID: wjs.UserAttributes[i].UserID,
-			Phone:  wjs.UserAttributes[i].Phone,
-			Email:  wjs.UserAttributes[i].Email,
+	for i, usrAttr := range wjs.UserAttributes {
+		usrAttribute[i].Opts = make(map[string]string)
+		for key, value := range usrAttr {
+			if key == "userId" {
+				usrAttribute[i].UserID = value.(string)
+			} else {
+				usrAttribute[i].Opts[key] = value.(string)
+			}
 		}
 	}
 	jobID, err := strconv.Atoi(wjs.JobID)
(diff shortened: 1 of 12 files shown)


regulation-worker/internal/model/model.go
```diff
@@ -36,8 +36,7 @@ type Job struct {
 
 type UserAttribute struct {
 	UserID string
-	Phone  *string
-	Email  *string
+	Opts   map[string]string
```
lvrach, line 39: In terms of readability, the following approaches sound better to me:

(Opts seems like user options, which is not exactly what we expect here to be)

```go
type User struct {
	ID string
	Attributes   map[string]string
}
```

Or 
```go
type UserAttribute struct {
	UserID string
	Optional   map[string]string
```

```go
type UserAttribute struct {
	UserID string
	KeyValue   map[string]string
```

lvrach (approved):

regulation-worker/internal/client/client.go
```diff
@@ -164,12 +164,15 @@ func prepURL(url string, params ...string) string {
...
-	for i := 0; i < len(wjs.UserAttributes); i++ {
-		usrAttribute[i] = model.UserAttribute{
-			UserID: wjs.UserAttributes[i].UserID,
-			Phone:  wjs.UserAttributes[i].Phone,
-			Email:  wjs.UserAttributes[i].Email,
+	usrAttribute := make([]model.User, len(wjs.UserAttributes))
+	for i, usrAttr := range wjs.UserAttributes {
+		usrAttribute[i].Attributes = make(map[string]string)
+		for key, value := range usrAttr {
+			if key == "userId" {
```
atzoum, line 171: how do we guarantee that `userId` field will always be present in the response? Or is it acceptable to have a `User` with an empty `ID`?

A test scenario would be desirable to disclose the desired system behaviour if this happens

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -338,28 +338,28 @@ func (b *Batch) upload(ctx context.Context, uploadFileAbsPath, actualFileName, a
...
+		if users.Attributes["email"] != "" {
 			searchObject = append(searchObject, "/"...)
 			searchObject = append(searchObject, "\"email\": *\""...)
-			searchObject = append(searchObject, []byte(*users.Email)...)
+			searchObject = append(searchObject, []byte(users.Attributes["email"])...)
 			searchObject = append(searchObject, "\"/d;"...)
 		}
 
-		if users.Phone != nil {
+		if users.Attributes["phone"] != "" {
```
atzoum, line 359: Why don't we search by other attributes and use only email and phone?

regulation-worker/internal/delete/kvstore/kvstore.go
```diff
@@ -28,11 +28,11 @@ func (kv *KVDeleteManager) Delete(ctx context.Context, job model.Job, destConfig
 	fileCleaningTime := stats.NewTaggedStat("file_cleaning_time", stats.TimerType, stats.Tags{"jobId": fmt.Sprintf("%d", job.ID), "workspaceId": job.WorkspaceID, "destType": "kvstore", "destName": destName})
 	fileCleaningTime.Start()
 	defer fileCleaningTime.End()
-	for _, user := range job.UserAttributes {
-		key := fmt.Sprintf("user:%s", user.UserID)
+	for _, user := range job.Users {
+		key := fmt.Sprintf("user:%s", user.ID)
 		err = kvm.DeleteKey(key)
```
atzoum, line 33: other attributes are not important for kvstore?

regulation-worker/internal/delete/batch/batch.go
saurav-malani: because for `batch` & `kv` destination type we only support deletion via `userId`, `Phone` & `email`. And, the reason behind this is, if the API caller gave a very common key & value that is present at rows which they didn't intended to delete can create problem. So, we are restricting to these three keys.

regulation-worker/internal/client/client.go
saurav-malani: That's a good point. `userId` is a required field. Although the reason we don't have a check for it is, regulation-manager throws error if any user facing API request doesn't have userId. So, we assumed in regulation-worker that regulation-manager itself will take care of it. So, we don't really need to worry about it.

saurav-malani: I am personally confused, if we should have a redundant check in regulation-worker or not?

One line of thought is, since it is checked at regulation-manager so it doesn't makes sense to check again here.
But another line of thought is, we shouldn't depend of regulation-manager (a different component) to make sure that it will send appropriate request with all the fields as per our requirement. 

But, at the same time, everything in itself could of thought of as a component, so redundantly who many times should we do the same check again & again. As in, the client here passes `job` to `service` (another component) which further passes job to `deleter` (another component). So, Where do we draw the line?

regulation-worker/internal/delete/kvstore/kvstore.go
saurav-malani: while writing `kvstore` deleter we checked how it is stored it redis. And, we found that key there is `user:$USERID` . So, we are using it to delete a user data in case of `kvstore`.

regulation-worker/internal/client/client.go
atzoum: I guess we could filter a response entry if userID is not present in it. But if we are OK to fail when sending the request to the service and we handle it gracefully, I' m fine with that too.

atzoum:
Are we going to include this in a patch release or it will be simply included in the next scheduled release?

cisse21:
We can patch this and make a release for the specific customer who needs it

atzoum:
> We can patch this and make a release for the specific customer who needs it

A patch release for a feature is not something we can really do. Our options as I see it are:
1. Mask it as a fix and change the target branch to release/1.1.x. Follow the patch release process (`1.1.2`) from there on.
2. Merge it normally in the main branch and cherry-pick it in a customer-specific branch (e.g. `1.1.1-<customername>-hotfix.1`) on top of tag 1.1.1, 
The first approach doesn't really make much sense to me

lvrach:
I would go with our normal flow. If it is urgent and since the worker is building as a separate binary and runs on a separate pod, we can deploy the master tag to the worker pod of the specific customer

atzoum (approved):

PR #2462, opened 2022-09-21
chore: add resource tags in metrics

satishrudderstack:
add resource tags in metrics


router/router.go
```diff
@@ -1079,6 +1079,7 @@ func (worker *workerT) sendRouterResponseCountStat(status *jobsdb.JobStatusT, de
 		"destType":       worker.rt.destName,
 		"respStatusCode": status.ErrorCode,
 		"destination":    destinationTag,
+		"destId":         destination.ID,
```
lvrach, line 1082: Similar to my comment here:
https://github.com/rudderlabs/rudder-server/pull/2430#discussion_r978345020

As destination is of high granularity, we should exclude it in free multitenant cluster.

lvrach (approved):

achettyiitr (approved):

router/router.go
chandumlg: https://github.com/rudderlabs/rudder-server/pull/2555/files

PR #2465, opened 2022-09-22
Fix/json out of bounds

lvrach:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security


router/batchrouter/batchrouterBenchmark_test.go
```diff
@@ -39,3 +44,68 @@ func Benchmark_GetStorageDateFormat(b *testing.B) {
...
+		})
+
+		g.Go(func() error {
+			for i := range jobs {
+				var params JobParametersT
+				_ = TryUnmarshalJSON(jobs[i].Parameters, &params)
+				// if err != nil {
+				// 	b.Fatal(err)
+				// }
+
```
fracasula, line 103:
```suggestion
```

router/batchrouter/batchrouterBenchmark_test.go
```diff
@@ -39,3 +44,68 @@ func Benchmark_GetStorageDateFormat(b *testing.B) {
 func randomString() string {
 	return strings.ReplaceAll(uuid.Must(uuid.NewV4()).String(), "-", "")
 }
+
+func Benchmark_JSONUmarshal(b *testing.B) {
```
fracasula, line 48:
```suggestion
// Benchmark_JSONUmarshal tries to reproduce a panic encountered with jsoniter
func Benchmark_JSONUmarshal(b *testing.B) {
```

router/batchrouter/batchrouter.go
```diff
@@ -2410,3 +2410,24 @@ func (brt *HandleT) updateProcessedEventsMetrics(statusList []*jobsdb.JobStatusT
...
+	c := cap(data)
+	l := len(data)
+
+	startingData := make([]byte, l)
+	copy(startingData, data)
+
+	defer func() {
+		if r := recover(); r != nil {
+			pkgLogger.Warnf("Panic while unmarshalling json: %v", r)
+			pkgLogger.Warnf("	starting slice: [%d:%d]: %q", l, c, string(startingData))
```
fracasula, line 2424: [optional] Not sure `%q` makes sense here, it'd probably generate confusion at least for me since you got two more double quotes around. Same below.

router/batchrouter/batchrouter.go
```diff
@@ -2410,3 +2410,24 @@ func (brt *HandleT) updateProcessedEventsMetrics(statusList []*jobsdb.JobStatusT
 		}).Count(count)
 	}
 }
+
+func TryUnmarshalJSON(data []byte, v interface{}) (err error) {
```
fracasula, line 2414:
```suggestion
func tryUnmarshalJSON(data []byte, v interface{}) (err error) {
```

fracasula (approved):
It's OK as a temporary solution, I'll try to understand the root cause and see if we can remove the `copy`.

router/batchrouter/batchrouterBenchmark_test.go
```diff
@@ -39,3 +44,68 @@ func Benchmark_GetStorageDateFormat(b *testing.B) {
...
+	config.Load()
```
fracasula, line 49: Do we still need the `config.Load()`? :thinking:

router/batchrouter/batchrouter.go
```diff
@@ -1191,7 +1191,7 @@ func (brt *HandleT) setJobStatus(batchJobs *BatchJobsT, isWarehouse bool, errOcc
 		}
 
 		var parameters JobParametersT
-		err = json.Unmarshal(job.Parameters, &parameters)
+		err = TryUnmarshalJSON(job.Parameters, &parameters)
```
lvrach, line 1194:
```suggestion
		err = tryUnmarshalJSON(job.Parameters, &parameters)
```

router/batchrouter/batchrouterBenchmark_test.go
lvrach: This PR is based on the `release/1.1.x` branch

saurav-malani (approved):

router/batchrouter/batchrouter.go
```diff
@@ -2410,3 +2412,25 @@ func (brt *HandleT) updateProcessedEventsMetrics(statusList []*jobsdb.JobStatusT
...
+func tryUnmarshalJSON(jobID int64, data []byte, v interface{}) (err error) {
+	c := cap(data)
+	l := len(data)
+
+	startingData := make([]byte, l)
+	copy(startingData, data)
+
+	defer func() {
+		if r := recover(); r != nil {
+			pkgLogger.Warnf(
```
atzoum, line 2425: How are we going to be alerted and make sure this piece of code won't be left there forever, in oblivion?

router/batchrouter/batchrouter.go
```diff
@@ -2434,3 +2436,25 @@ func (brt *HandleT) updateProcessedEventsMetrics(statusList []*jobsdb.JobStatusT
 		}
 	}
 }
+
+func tryUnmarshalJSON(jobID int64, data []byte, v interface{}) (err error) {
+	c := cap(data)
+	l := len(data)
+
+	startingData := make([]byte, l)
+	copy(startingData, data)
```
chandumlg, line 2445: Does this add any significant overhead?

router/batchrouter/batchrouter.go
fracasula: We can do this: https://github.com/rudderlabs/rudder-devops/blob/e64392172c6beddb7991adcc5907435806177853/helm-charts/loki-stack/values.infra1-us.yaml#L73

router/batchrouter/batchrouter.go
fracasula: Not particularly concerned about the overhead but not sure how useful it can be as a metric given that the race could happen exactly when we copy rather than we unmarshal. I'll get rid of it. The plan is to get the configuration used to trigger this and the jobID to further investigate, I don't think the copy will solve much. @lvrach any thoughts since you added this?

atzoum: Please keep in mind that it is possible for the job to get lost pretty quickly from the table, without leaving other trace, since AFAIK we don't keep backups of all jobs in hosted environment

lvrach: > race could happen exactly when we copy rather than we unmarshal
Do you think that this going to cause a panic? Well, the defer will deal with that I believe. if we manage to capture `startingData` != `data` it would be very useful.

router/batchrouter/batchrouter.go
atzoum: Since this issue is not frequent (2 occurences so far), it might be easier for us to detect and debug if we didn't shallow the panic, but simply log some more stuff, either in the panic message itself or in the logs.

atzoum (approved):

PR #2477, opened 2022-09-24
refactor: logger package

atzoum:
# Description

Refactoring the `logger` package and introducing the following features:

1. Support for multiple, independent `logger.Factory` instances
2. Thread-safe initialization & reinitialization of `logger.Factory` instances (including the default, singleton instance). No more nil pointer dereference panics.
5. Support for structured logging, see [Logger.With](https://github.com/rudderlabs/rudder-server/blob/39a4cbf98c3efc6563ae1abb06ceaebfe0863c32/utils/logger/logger.go#L102), [Logger.Infow](https://github.com/rudderlabs/rudder-server/blob/39a4cbf98c3efc6563ae1abb06ceaebfe0863c32/utils/logger/logger.go#L67), [Logger.Errorw](https://github.com/rudderlabs/rudder-server/blob/39a4cbf98c3efc6563ae1abb06ceaebfe0863c32/utils/logger/logger.go#L85), etc
4. Include logger name in log output (configuration option available for opting out)
5. Test coverage is now >90%

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=ac57434350c94d39af6eac097a323748&pm=s)

## Security


utils/logger/nop.go
```diff
@@ -1,21 +0,0 @@
-package logger
-
-import "net/http"
-
-var _ LoggerI = &NOP{}
-
-type NOP struct{}
```
atzoum, line 7: _**Note:**_ there is no longer a need for a no-op logger implementation since, there will always be a non-nil default logger factory available, and logger is immune from constant and concurrent re-initializations.

lvrach: A question, what if we don't want to print any logs? What the non-nil default logger will do?

app/app.go
```diff
@@ -43,7 +43,7 @@ type Interface interface {
 	Features() *Features // Get this application's enterprise features
 }
 
-var pkgLogger logger.LoggerI
+var pkgLogger logger.Logger
 
 func Init() {
 	pkgLogger = logger.NewLogger().Child("app")
```
lvrach, line 49: I believe we could do the following:

```go
	var pkgLogger = logger.NewLogger().Child("app")
```

And avoid having an `Init()` method here; it doesn't provide any more benefit as there is always a default logger initialised.

config/backend-config/namespace_config_test.go
```diff
@@ -38,7 +38,7 @@ func Test_Namespace_SetUp(t *testing.T) {
 
 func Test_Namespace_Get(t *testing.T) {
 	config.Reset()
-	logger.Init()
+	logger.Reset()
```
lvrach, line 41: I don't think we are using any config-specific thing in our logs apart from the log level. 

We can set it to debug in tests, and if spam is a concern, start injecting the logger in components.

 Thus should be able to safely remove `logger.Reset()`

utils/logger/nop.go
lvrach: I think a better approach is to keep the NOP{} logger for now, and introduce a particular "mock" logger that would:

a) define the log level 
b) use t.Logf
c) inspect/grep the log for specific output
d) have the option to mute logs

lvrach (approved):
Overall am ok with the changes. 

I would go a step further and 

A. eliminate logger from the `Init` method:
```go
func Init() {
	pkgLogger = logger.NewLogger().Child("...")
}
```
No practical value is provided, and it will significantly improve the code quality.

B. Not use `logger.Reset()`; my understanding is that it offers only the ability to change the log level in tests, but given the introduction's dirtiness, I prefer having to debug level everywhere the global logger is used. An incentive to DI your logger if you don't want to get spam in your tests.

C. let's not replace noop logger in the existing test yet. Using logger.NewLogger introduces spam and makes the output less readable. Test cases by themself should explain what is wrong (for unit tests). I would keep it until we have a better option.

config/backend-config/namespace_config_test.go
atzoum: As long as the logger package depends on the config package, we may keep it until we reach to the point where we no longer need to depend on reseting the config. I'm not in favour of enabling debug logging in tests, e.g. by exporting some env variable in `make test`. IMO, it should be possible for any test to use it's own config options, including defining the `LOG_LEVEL`.

utils/logger/nop.go
atzoum: To disable logs from the console, one would need to `config.Set("Logger.enableConsole", false)`, but why one would want to disable logs completely?

atzoum:
> B. Not use `logger.Reset()`

please beware, that most packages will need to have a `package.Reset` function for some time more, until we reach to the point that we don't need them anymore.

utils/logger/nop.go
atzoum: Revived the `logger.NOP` struct, since it appears to be useful for some of us :)

chandumlg (approved):

PR #2481, opened 2022-09-26
[WIP] chore: resolve WAREHOUSE_URL in gateway

atzoum:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security


chrikar:
@atzoum do we need this after all?

atzoum:
> @atzoum do we need this after all?

You don't want to have to manually override these values in each and every namespace that you want to use the new `/job-status` endpoint, do you?

PR #2484, opened 2022-09-27
feat(warehouse): batching of alter add statements 

achettyiitr:
# Description

Batching of `ALTER ADD` statements in order to optimize the Queries that run on the warehouse. 

Currently, we are following running single `ALTER ADD` statement changes for new properties of the Schema. If we batch these, we don't need to run many Queries for these.

## Notion Ticket

https://www.notion.so/rudderstacks/Use-batching-to-run-several-of-alter-statements-81ae26d5db6f4b2ca8e0258e3ce0cba6

## Security


warehouse/deltalake/deltalake.go
```diff
@@ -868,11 +868,9 @@ func (dl *HandleT) DropTable(tableName string) (err error) {
 	return
 }
 
-// AddColumn adds column for column name and type
-func (dl *HandleT) AddColumn(name, columnName, columnType string) (err error) {
-	tableName := fmt.Sprintf(`%s.%s`, dl.Namespace, name)
-	sqlStatement := fmt.Sprintf(`ALTER TABLE %v ADD COLUMNS ( %s %s );`, tableName, columnName, getDeltaLakeDataType(columnType))
-	pkgLogger.Infof("%s Adding column in delta lake with SQL:%v", dl.GetLogIdentifier(tableName, columnName), sqlStatement)
+func (dl *HandleT) AddColumns(tableName string, columnsInfo warehouseutils.ColumnsInto) (err error) {
+	sqlStatement := addColumnsSQLStatement(dl.Namespace, tableName, columnsInfo)
```
lvrach, line 872: I am not sure if the `addColumnsSQLStatement` is necessary since it need only from this method, it would be better for readability for this code to be in one place:

```suggestion
	format := func(_ int, columnInfo warehouseutils.ColumnInfoT) string {
		return fmt.Sprintf(`%s %s`, columnInfo.Name, getDeltaLakeDataType(columnInfo.Type))
	}
	sqlStatement =  fmt.Sprintf(`
		ALTER TABLE
		%s.%s
		ADD COLUMNS ( %s );`,
		namespace,
		tableName,
		columnsInfo.JoinColumns(format, ","),
	)
```

The same comments applies in other files as well.

achettyiitr: In favor of this. Once @abhimanyubabbar is back, we will finalize on this.

warehouse/azure-synapse/azure-synapse.go
```diff
@@ -653,9 +645,46 @@ func (as *HandleT) DropTable(tableName string) (err error) {
...
-func (as *HandleT) AddColumn(tableName, columnName, columnType string) (err error) {
-	err = as.addColumn(as.Namespace+"."+tableName, columnName, columnType)
-	return err
+func (as *HandleT) AddColumns(tableName string, columnsInfo warehouseutils.ColumnsInto) (err error) {
+	var (
+		format       func(_ int, columnInfo warehouseutils.ColumnInfoT) string
+		sqlStatement string
+	)
+
+	format = func(_ int, columnInfo warehouseutils.ColumnInfoT) string {
```
achettyiitr, line 654: Try to remove anonymous functions.

achettyiitr: Try to use parameterization.

achettyiitr: Table-driven tests for verifying for normalization of the queries?

achettyiitr: Sanitize the date. Don't need to be dependent for transtransformer.

achettyiitr:
Try to remove anonymous functions.

warehouse/identity/identity.go
```diff
@@ -11,7 +11,7 @@ import (
 	"strings"
 	"time"
 
-	uuid "github.com/gofrs/uuid"
```
lvrach, line 14: Note: we should use google's uuid, as gofrs can cause [performance issues](https://github.com/rudderlabs/rudder-server/pull/1606)

warehouse/utils/utils.go
```diff
@@ -1062,6 +1062,14 @@ type FilterBy struct {
 	Value interface{}
 }
 
+type ColumnInfoT struct {
+	Name  string
+	Value interface{}
+	Type  string
+}
+
+type ColumnsInfo []ColumnInfoT
```
lvrach, line 1071: * let's avoid the T suffix
* In go it is not very idiomatic to define a type for slices of other types. We can simply use `[]ColumnInfo`, which is more readable as well, because it is clear it is a slice, and you now that is a slice of `ColumnInfo` without further inspection. Further more it only cost an extra character if you think about it `s` vs `[]`


```suggestion
type ColumnInfo struct {
	Name  string
	Value interface{}
	Type  string
}
```

warehouse/azure-synapse/azure-synapse.go
```diff
@@ -653,9 +645,44 @@ func (as *HandleT) DropTable(tableName string) (err error) {
...
+	query += fmt.Sprintf(`
+		ALTER TABLE
+		  %s.%s
+		ADD`,
+		as.Namespace,
+		tableName,
+	)
+
+	for _, columnInfo := range columnsInfo {
+		query += fmt.Sprintf(` %s %s,`, columnInfo.Name, rudderDataTypesMapToMssql[columnInfo.Type])
```
lvrach, line 677: What if a column name contains spaces ? Is it already sanitised in transformations?

warehouse/bigquery/bigquery.go
```diff
@@ -826,15 +809,36 @@ func (bq *HandleT) LoadTable(tableName string) error {
 	return err
 }
 
-func (bq *HandleT) AddColumn(tableName, columnName, columnType string) (err error) {
-	err = bq.addColumn(tableName, columnName, columnType)
+func (bq *HandleT) AddColumns(tableName string, columnsInfo warehouseutils.ColumnsInfo) (err error) {
+	pkgLogger.Infof("BQ: Adding columns in table %s in bigquery dataset: %s in project: %s", tableName, bq.namespace, bq.projectID)
```
lvrach, line 813: I am wondering, are those logs useful? It seems to me that it is something frequent that might lead to noise.

lvrach (approved):
The changes look good to me. Note: I don't have a lot of context on how those different warehouse work. I rely on automation tests to verify those changes.

I made a few non-blocking comments/suggestions

warehouse/identity/identity.go
achettyiitr: Thanks, @lvrach for sharing this PR.

achettyiitr: Although it was out of the scope of the PR, so thinking to make this change in a separate PR.

achettyiitr: Also, any specific reason why we haven't migrated all existing implementation everywhere from `uuid.Must(uuid.NewV4())` to `misc.FastUUID().String()`?

achettyiitr: Created a draft https://github.com/rudderlabs/rudder-server/pull/2598 to discuss the same.

warehouse/utils/utils.go
achettyiitr: Done.

warehouse/azure-synapse/azure-synapse.go
achettyiitr: Yes. For properties are already sanitized in [here](https://github.com/rudderlabs/rudder-transformer/blob/586f771f8e0733ac2f79ea4741bb155eb24910ca/warehouse/v0/util.js#L82).

warehouse/bigquery/bigquery.go
achettyiitr: Whenever new properties are added, this will happen. Earlier we used to log for every insert statement. But now since we are batching for every 50 columns, we will see only a few logs.
Since this is just a one-time thing that will happen when new properties are added, I think it should be fine.

achettyiitr:
> The changes look good to me. Note: I don't have a lot of context on how those different warehouse work. I rely on automation tests to verify those changes.

Except for the Azure synapse and S3 Datalake (Glue), all destinations are covered with multiple columns for alter. So we are good here.

deepakrai9185720 (approved):

achettyiitr:
> > The changes look good to me. Note: I don't have a lot of context on how those different warehouse work. I rely on automation tests to verify those changes.
> 
> Except for the Azure synapse and S3 Datalake (Glue), all destinations are covered with multiple columns for alter around Integration tests. So we are good here.

For Azure synapse, tested the changes locally.

PR #2485, opened 2022-09-27
fix: remove custom credentials chain to use default

koladilip:
# Description
OSS user reported that custom credentials chain is causing issues so removing to use default chain provided by SDK.

I have mistakenly configured AWS creds chain thinking about requirements from Kubernetes stand point and missed other types deployment like ECS so it is causing failures in ECS tasks when users want to task role instead of access keys from destination config. When no chain is mentioned default chain is taking care of this use case without any issues.

## Notion Ticket

https://www.notion.so/rudderstacks/AWS-FTR-fix-OSS-issues-d64b93d9ed1d4c00a8d7e5573f7992c3

## Security


saikumarrs (approved):

utils/awsutils/session.go
```diff
@@ -46,6 +44,9 @@ func createDefaultSession(config *SessionConfig) (*session.Session, error) {
 }
 
 func createCredentailsForRole(config *SessionConfig) (*credentials.Credentials, error) {
+	if config.ExternalID == "" {
+		return nil, errors.New("externalID is required fo IAM role")
```
psrikanth88, line 48: typo: fo -> for

psrikanth88 (approved):

PR #2489, opened 2022-09-27
refactor: stats package

atzoum:
# Description

Refactoring the stats package and introducing the following improvements:

1.  Thread-safe initialization of `stats.Stats` instances (including the default, singleton instance). No more global variables.
2. Using `Stats.Start` & `Stats.Stop` for controlling the component's lifecycle. During tests, the default stats instance is initialised, but not started.
3. Proper handling of muted measurements, i.e. `Measurement` (fka `RudderStats`) instances created before the `Stats` instance was able to successfully create a statsd client remained **muted** in the previous implementation. A special test case verifies the correct behaviour (see [stats_internal_test.go](https://github.com/rudderlabs/rudder-server/blob/65bd41d779f0622d7fb765c08a518adbb799649b/services/stats/stats_internal_test.go)).
4. Test coverage is now >90%.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=ab2ed69ae644454db5a7a9bf42ebbcb3&pm=s)

## Security


services/stats/misc.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"strings"
+
+	"github.com/rudderlabs/rudder-server/config"
+	"github.com/rudderlabs/rudder-server/utils/types/deployment"
+)
+
+// TODO: Do we need this?
+func CleanupTagsBasedOnDeploymentType(tags Tags, key string) Tags {
```
atzoum, line 11: @chandumlg I remember @cisse21 mentioning that we shall push all tags (unfiltered) and infra shall take care of removing tags where necessary. Do I remember correctly?

services/stats/config.go
```diff
@@ -0,0 +1,70 @@
...
+func (c *statsdConfig) statsdDefaultTags() statsd.Option {
+	tags := []string{"instanceName", c.instanceID}
+	if len(config.GetKubeNamespace()) > 0 {
+		tags = append(tags, "namespace", config.GetKubeNamespace())
+	}
+	return statsd.Tags(tags...)
+}
+
+// statsdtagsFormat returns the tags format to use for statsd
+func (c *statsdConfig) statsdtagsFormat() statsd.Option {
```
lvrach, line 38: [minor] statsdTagsFormat

services/stats/misc.go
lvrach: Let's keep this like this for now. There technical reasons were filter on infra might not be feasible.

services/stats/stats_internal_test.go
```diff
@@ -0,0 +1,141 @@
...
+	//  3. Create a stats measurement and try to post some stats
+	counter := s.NewTaggedStat("test", CountType, Tags{"tag1": "value1"})
+	counter.Increment()
+
+	// 4. No statsd client should have been created yet, thus we expect pendingClients to be non-empty
+	require.Equal(t, 1, len(s.state.pendingClients))
+	require.Equal(t, "", lastMetric, "statsd server shouldn't have received any metrics yet")
+
+	// 5. Make statsd server's domain name resolvable to localhost
+	srv, _ := mockdns.NewServer(map[string]mockdns.Zone{
```
lvrach, line 61: This is interesting, why it was necessary?

lvrach (approved):
Amazing work on the tests, approved with some minor comments

lvrach (approved):

services/stats/stats_internal_test.go
atzoum: Since we are using udp (connectionless protocol), even if nothing is listening on the target port, the statsd client creation will succeed, as long as the provided address is resolvable (e.g. if we are using an ip address such as 127.0.0.1). The only way for the statsd client creation to fail is to provide a non-existent domain in the address field.
Next step, after we verify that client creation has failed, is to make the domain name resolvable so that statsd client creation can succeed when retried. To make the domain name resolvable we need to intercept go's dns resolution and use our own (mock) DNS server.

lvrach: Brilliant! 

May I suggest adding this paragraph (without the meme :P )  as a comment?

services/stats/misc.go
atzoum: I will keep it, but in another form, decoupling the stats package from this logic through a config option (`excludedTags`)

main.go
```diff
@@ -239,6 +237,18 @@ func Run(ctx context.Context) int {
...
 
+	deploymentType, err := deployment.GetFromEnv()
+	if err != nil {
+		pkgLogger.Errorf("failed to get deployment type: %w", err)
+		return 1
+	}
+	// TODO: remove as soon as we update the configuration with statsExcludedTags where necessary
+	if !config.IsSet("statsExcludedTags") && deploymentType == deployment.MultiTenantType && (!config.IsSet("WORKSPACE_NAMESPACE") || strings.Contains(config.GetString("WORKSPACE_NAMESPACE", ""), "free")) {
+		config.Set("statsExcludedTags", []string{"workspaceId"})
+	}
```
atzoum, line 248: @chandumlg we should really set this through the config. I think it is only needed in one place, right?

services/stats/stats.go
```diff
@@ -93,327 +78,190 @@ func (t Tags) String() string {
...
-			tagVals := tags.Strings()
-			if !connEstablished {
-				taggedClientPendingTags = append(taggedClientPendingTags, taggedClientKey)
-				taggedClientPendingKeys = append(taggedClientPendingKeys, tagVals)
-			}
-			taggedClient = client.Clone(conn, statsd.TagsFormat(getTagsFormat()), defaultTags(), statsd.Tags(tagVals...), statsd.SampleRate(samplingRate))
-			taggedClientsMap[taggedClientKey] = taggedClient
+func (s *statsdStats) collectPeriodicStats() {
+	gaugeFunc := func(key string, val uint64) {
+		if s.state.client.statsd != nil {
```
chandumlg, line 176: Do we need this nil check?

services/stats/stats.go
```diff
@@ -93,327 +78,190 @@ func (t Tags) String() string {
...
+	if s.conf.periodic.enabled {
+		var wg sync.WaitGroup
+		wg.Add(2)
+		go func() {
+			defer wg.Done()
+			s.state.rc.run()
+		}()
+		go func() {
+			defer wg.Done()
+			s.state.mc.run()
```
chandumlg, line 196: Is it safe to put metrics collector behind `s.conf.periodic.enabled` flag?

main.go
chandumlg: Yes.

BTW, do we need !config.IsSet("statsExcludedTags") and !config.IsSet("WORKSPACE_NAMESPACE") checks?

atzoum: ` !config.IsSet("statsExcludedTags")` is there to ignore this fragment in case external configuration has been provided for `statsExcludedTags`.

services/stats/stats.go
atzoum: yes, `statsd` client might be nil in case of errors. But I will replace it with `s.state.client.ready()`

services/stats/stats.go
atzoum: Yes it was like that previously. The idea is that both routines collect periodical stats at the same frequency

main.go
chandumlg: Got it.

services/stats/stats.go
chandumlg: My only fear was if someone disables runtime metrics (not wanting cpu/memory usage metrics), pending_events metrics will go away. But if it was like that previously as well, we can leave this as it is.

services/stats/stats.go
chandumlg: `statsd` client will never be nil right?

atzoum: statsd client will be nil if `Stats` is not [started](https://github.com/rudderlabs/rudder-server/blob/a8b1ab147a3264b9038000ff107f20dd554f8bf7/services/stats/stats.go#L117). This is how all tests, other than stats’ own tests, run: stats is initialized but not started

atzoum: but you are correct, this specific function will only be called after a statsd client has been successfully created...

PR #2496, opened 2022-09-28
feat: add support iam role support for warehouse destinations

koladilip:
# Description
As part of AWS FTR we need to support IAM roles instead of access keys for all AWS interactions. We already supported for file and streaming destinations using common utils. Now I am doing the same for warehouse destinations where there is dependency on S3 for Snowflake, deltalake, Redshift and using common utils we can automatically enable support for role based authentication.

In that process I have update glue client creation to use the same common utils to support role based auth.

## Notion Ticket

https://www.notion.so/rudderstacks/AWS-FTR-for-Redshift-e7dc80ee847a49d6b7bb7fc03aeeffc2

## Security


lvrach:
@koladilip can you provide additional details and context regarding the changes in this PR?

achettyiitr (approved):

koladilip:
> @koladilip can you provide additional details and context regarding the changes in this PR?

I have updated the issue description, let me know if you more details are needed.

koladilip:
Tested with redshift and glue,  it is working fine.

utils/awsutils/session.go
```diff
@@ -111,3 +113,15 @@ func NewSessionConfigForDestination(destination *backendconfig.DestinationT, tim
...
 }
+
+func NewSessionConfigForDestination(destination *backendconfig.DestinationT, timeout time.Duration, serviceName string) (*SessionConfig, error) {
+	sessionConfig, err := NewSimpleSessionConfigForDestination(destination, serviceName)
+	if err != nil {
+		return nil, err
+	}
+	if sessionConfig.Region == "" {
+		return nil, errors.New("could not find region configuration")
+	}
```
fracasula, line 124: I see that this was part of `NewSimpleSessionConfigForDestination` before. Just for me to understand, what is the use case for a session config with no region?

utils/misc/misc.go
```diff
@@ -923,6 +923,16 @@ func FastUUID() uuid.UUID {
 	return uuid.FromBytesOrNil(b)
 }
 
+func HasAWSRoleARNInConfig(configMap map[string]interface{}) bool {
+	if configMap["iamRoleARN"] == nil {
+		return false
+	}
+	if configMap["iamRoleARN"].(string) == "" {
```
fracasula, line 930: To avoid a potential panic in case this isn't a string, maybe it makes more sense to do:
```suggestion
	iamRoleARN, ok := configMap["iamRoleARN"].(string)
	if !ok {
		return false
	}
	if iamRoleARN == "" {
```

warehouse/utils/utils.go
```diff
@@ -707,14 +706,44 @@ func JoinWithFormatting(keys []string, format func(idx int, str string) string,
...
+		sessionConfig = &awsutils.SessionConfig{
+			AccessKeyID: accessKeyID,
+			AccessKey:   accessKey,
+			Service:     serviceName,
+		}
+	} else {
+		sessionConfig, err = awsutils.NewSimpleSessionConfigForDestination(destination, serviceName)
+	}
+	return sessionConfig, err
+}
```
fracasula, line 726: [optional]
```suggestion
func CreateAWSSessionConfig(destination *backendconfig.DestinationT, serviceName string) (*awsutils.SessionConfig, error) {
	if !misc.IsConfiguredToUseRudderObjectStorage(destination.Config) {
		return awsutils.NewSimpleSessionConfigForDestination(destination, serviceName)
	}
	accessKeyID, accessKey := misc.GetRudderObjectStorageAccessKeys()
	return &awsutils.SessionConfig{
		AccessKeyID: accessKeyID,
		AccessKey:   accessKey,
		Service:     serviceName,
	}, nil
}
```

utils/awsutils/session.go
koladilip: In some case we derive from S3 buckets and also sometimes for warehouse we are using to generate STS tokens that doesn't require region.

fracasula (approved):

PR #2503, opened 2022-09-29
[WIP] fix: don't treat numebers in message payloads as floats

atzoum:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -3,7 +3,6 @@ package gateway
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -21,6 +20,7 @@ import (
 	"github.com/bugsnag/bugsnag-go/v2"
 	"github.com/gofrs/uuid"
 	"github.com/gorilla/mux"
+	jsoniter "github.com/json-iterator/go"
 	"github.com/rs/cors"
 	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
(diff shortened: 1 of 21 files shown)


gateway/gateway.go
```diff
@@ -1670,3 +1632,112 @@ func WithContentType(contentType string, delegate http.HandlerFunc) http.Handler
...
+		return true // keep iterating
+	})
+
+	res, _ = sjson.SetBytes(body, "batch", out)
+	return
+}
+
+func injectIdentifiersV2(userIDHeader string, body []byte) (res []byte, notIdentifiable, nonRudderEvent, containsAudienceList bool, builtUserID string) {
+	type batchMessage struct {
+		Batch []map[string]interface{} `json:"batch"`
```
lvrach, line 1689: We can also test:
```suggestion
        type usefulFields struct {
            AnonymousID string 
            UserID string
            Type string
        }
		Batch []usefulFields `json:"batch"`
```
It is more memory efficient for sure, but maybe faster for the CPU as well

lvrach: Ignore, for now, I will try to give a proposal that works end 2 end

lvrach: I was quick to comment without reading the rest of the function:

here is my suggestion:
https://github.com/rudderlabs/rudder-server/pull/2506/files#diff-17cb8b37eda9018fe1c6cdb5f96b3fc948fc8ba49bc516987b8269576db9fcd4R1812

atzoum: v4 looks nice :)

PR #2507, opened 2022-09-29
suggestion for #2501

lvrach:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security


warehouse/snowflake/snowflake_test.go
```diff
@@ -47,121 +37,109 @@ func (t *TestHandle) VerifyConnection() error {
...
+	credentials, err := testhelper.SnowflakeCredentials()
+	if err != nil {
+		return err
+	}
+	
+	testcase := []struct {
+		name string
+		dbName string
+		schema string
+	}{
```
lvrach, line 69: Using table test to avoid code repetition, but also to highlight the attributes we are testing every time

warehouse/snowflake/snowflake_test.go
```diff
@@ -47,121 +37,109 @@ func (t *TestHandle) VerifyConnection() error {
 		return err
 	}
 
-	return testhelper.WithConstantBackoff(func() (err error) {
```
lvrach, line 50: I understand that verify the connection is used to wait for the snowflake to be available, thus is not needed to test both `db_name`s .

warehouse/snowflake/snowflake_test.go
```diff
@@ -199,19 +177,3 @@ func TestSnowflakeConfigurationValidation(t *testing.T) {
...
-	//	return
-	//}
-	//
-	//handle = &TestHandle{
-	//	WriteKey:             "2eSJyYtqwcFiUILzXv2fcNIrWO7",
-	//	CaseSensiiveWriteKey: "2eSJyYtqwcFYUILzXv2fcNIrWO7",
-	//	Schema:               testhelper.Schema(warehouseutils.SNOWFLAKE, testhelper.SnowflakeIntegrationTestSchema),
-	//	CaseSensitiveSchema:  fmt.Sprintf("%s_%s", testhelper.Schema(warehouseutils.SNOWFLAKE, testhelper.SnowflakeIntegrationTestSchema), "cs"),
-	//	Tables:               []string{"identifies", "users", "tracks", "product_track", "pages", "screens", "aliases", "groups"},
-	//}
```
lvrach, line 215: Not sure why those things are needed here

warehouse/snowflake/snowflake_test.go
achettyiitr: Yes.

warehouse/snowflake/snowflake_test.go
achettyiitr: Yes. Thant's true.

warehouse/snowflake/snowflake_test.go
achettyiitr: Since we have moved to Table Driven Tests these are not required here.

achettyiitr:
Thanks @lvrach for suggesting these changes. 

- We were able to remove the global handler for TestHandle.
- Remove duplicate code using Table Driven Tests.

warehouse/snowflake/snowflake_test.go
achettyiitr: Even the global handler is not required.

PR #2508, opened 2022-09-29
chore: metric deprecation

lvrach:
# Description

## Metric deprecation

With this PR, I would like to introduce a way to deprecate metrics. Removing or renaming a metric can break dashboards or alerts. Moreover, server and dashboard/alert deployments happen in multiple and different phases. 

Thus a process for depreciating metrics is needed. I propose to use the `Deprecated` functions that have a two-fold usage:
 * They make it evident in the code that this metric needs to be removed.
 * It is also an indication for people using those metrics that are deprecated and should be avoided.
 * We can easily query our metric storage to determine the transition of a change quickly. For example, if all metrics are coming with a deprecated flag, we can transition to the new metric.
 
## Notion Ticket

https://www.notion.so/rudderstacks/stats-depreciation-05ea828db9094473a27d415f58a0312f

## Security


fracasula (commented):
The PR looks good to me and I'm not against this implementation. Just a small comment though.

What do you think of adding options as variadic to `NewStat`, `NewTaggedStat` and `NewSampledTaggedStat`? Too much work? I'd prefer it compared to your proposal.

It would give us more flexibility to add more options in the future without breaking the signature and we don't fatten up the interface either.

```go
NewStat(name, statType string, options ...Option) (m Measurement)
```

services/stats/stats.go
```diff
@@ -217,12 +236,30 @@ func (s *statsdStats) NewStat(name, statType string) (m Measurement) {
 	return newStatsdMeasurement(s.conf, name, statType, s.state.client)
 }
 
-func (s *statsdStats) NewTaggedStat(Name, StatType string, tags Tags) (m Measurement) {
-	return s.internalNewTaggedStat(Name, StatType, tags, 1)
+func (s *statsdStats) DeprecatedStat(name, statType string) (m Measurement) {
+	return s.internalNewTaggedStat(name, statType, Tags{deprecatedTag: "true"}, 1)
```
atzoum, line 240:
```suggestion
	return s.internalNewTaggedStat(name, statType, Tags{deprecated: "true"}, 1)
```

atzoum:
Can you describe the proposed process for replacing old measurements, with respect to dashboards, alerts in combination with this newly introduced `_deprecated` tag?

PR #2512, opened 2022-09-30
refactor: dataset index calculation

atzoum:
# Description

After removing the old migrator, the algorithm for calculating dataset indices can be simplified.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=e7ebbf8478c743e3ad0904b9a179bc8c&pm=s)

## Security

### DIFF
--- a/jobsdb/internal/dsindex/dsindex.go
+++ b/jobsdb/internal/dsindex/dsindex.go
@@ -1,0 +1,128 @@
+package dsindex
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// Index represents a sortable dataset index, e.g. 1 < 1_1 < 1_1_1 < 1_2 < 2
+type Index struct {
+	segments []int
+}
+
+// MustBump returns the next index that is greater than the current one,
+// but still less than the index provided in the parameter. Panics on error
+func (idx *Index) MustBump(previous *Index) *Index {
+	res, err := idx.Bump(previous)
+	if err != nil {
+		panic(err)
+	}
+	return res
+}
+
+// Bump returns the next index that is greater than the current one,
+// but still less than the index provided in the parameter.
+//
+// Bump doesn't increment the first, major segment of the index, it starts from the second segment instead.
+func (idx *Index) Bump(before *Index) (*Index, error) {
+	if !idx.Less(before) {
+		return nil, fmt.Errorf("%s is not before %s", idx, before)
+	}
+	var segment int // never increasing the major segment (index 0)
+	for {
+		segment++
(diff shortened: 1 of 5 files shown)


jobsdb/jobsdb.go
```diff
@@ -1350,94 +1349,23 @@ func (jd *HandleT) doComputeNewIdxForAppend(dList []dataSetT) string {
...
-	calculatedInsertVals, err := computeInsertVals(beforeVals, afterVals)
+	result, err := before.Bump(after)
 	if err != nil {
-		return "", fmt.Errorf("Failed to calculate InserVals with error: %w", err)
+		return "", fmt.Errorf("could not compute insert index: %w", err)
 	}
-	calculatedIdx := strings.Join(calculatedInsertVals, "_")
-	if len(calculatedInsertVals) > 3 {
-		return "", fmt.Errorf("We don't expect a ds to be computed to Level3. We got %s while trying to insert between %s and %s", calculatedIdx, beforeIndex, afterIndex)
+	if result.Length() > 2 {
```
atzoum, line 1365: Why should we explicitly forbid a level 3 index?
In practice we know that this will not happen, but what would be the problem even if it happened?
As long as we can guarantee that the calculated index is still smaller than the `afterIndex` and greater than the `beforeIndex` we should accept the generated index.

cisse21 (approved):

jobsdb/jobsdb.go
cisse21: Yes… it would read the smallest DS index and create tables accordingly

atzoum: then even a 10 lvl index shouldn't complicate things, would it?

chandumlg (approved):

jobsdb/internal/dsindex/dsindex.go
```diff
@@ -0,0 +1,128 @@
...
+	var lastSegment int
+	if idx.Length() > segment {
+		lastSegment = idx.segments[segment]
+	}
+	res.segments = append(res.segments, lastSegment+1)
+	return res, nil
+}
+
+// Less returns true if this dataset index is Less than the other dataset index
+func (idx *Index) Less(other *Index) bool {
```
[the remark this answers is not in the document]
atzoum: Since comparing or bumping a nil Index cannot yield any meaningful result other than a panic, no extra branch is needed for checking against it. A nil pointer dereference panic is guaranteed and should be enough :)

PR #2514, opened 2022-09-30
feat(destination): add new tags to router-aborted-count alert definition

sanpj2292:
# Description

The integrations are planning to release new alerts for integrations which is using measurement `integration_failure_detailed` for proxy enabled destinations(only) and all destinations with any type of transformation(processor or router or batch). Since, to enable transformerProxy for all the destinations, we need to do a benchmarking of transformer on whole to get to know several parameters like scaling, throughput etc., which would take time as we have other important things(CDK V2) to finish off.
Hence we want the alert definition(`router-aborted-count`) to produce alerts for the destinations which don't have proxy enabled.

To help us in the process of co-existing with new alert defs, we wanted to add new tags so that we can control allowing of the old alert for transformations(router or batch) or proxy 

This PR adds the new tags to `router_abort_events` and `router_response_counts` measurements involved in `router-aborted-count` alert definition

The tags are `alert` and `errorAt`. There is in-code documentation what they are and how are we planning to use them

**Note**:
- We would like to get this released as a `hot-fix`
- The alert definition would require a change to include only if the tag `alert` is `true` only then an alert should be triggered

## Notion Ticket

https://www.notion.so/rudderstacks/Transformer-alerts-Revamp-5af8ab9e12bc4247b24515504dd74f9f

## Security


router/router.go
```diff
@@ -977,17 +997,40 @@ func (worker *workerT) updateReqMetrics(respStatusCode int, diagnosisStartTime *
...
+}
+
+func (worker *workerT) updateAbortedMetrics(destinationID, statusCode, errorAt string) {
+	allowedTags := worker.getAllowedRtTags(errorAt)
 	eventsAbortedStat := stats.Default.NewTaggedStat(`router_aborted_events`, stats.CountType, stats.Tags{
 		"destType":       worker.rt.destName,
 		"respStatusCode": statusCode,
 		"destId":         destinationID,
+		"allowTf":        allowedTags.allowTf,
+		"allowDel":       allowedTags.allowDel,
```
atzoum, line 1026: Why do we need 2 tags for alerting, wouldn't a single `alert` tag suffice? I also find the choice of names a bit unintuitive

router/router.go
```diff
@@ -122,6 +131,8 @@ type HandleT struct {
 	savePayloadOnError                     bool
 	oauth                                  oauth.Authorizer
 	transformerProxy                       bool
+	allowRtAbortAlertForDelv               bool // represents if transformation(router or batch) should be alerted via router-aborted-count alert def
+	allowRtAbortForTf                      bool // represents if event delivery(via transformerProxy) should be alerted via router-aborted-count alert def
```
atzoum, line 135: I am a bit confused with the comments of these fields and how they are actually being used in `getAllowedRtTags`.

E.g. it seems to me that their names should be:
`allowRtAbortForTf` --> `alertOnTransformationAbort // send an alert if an event is aborted due to a transformation error`
`allowRtAbortAlertForDelv` --> `alertOnTransformerProxyAbort  // send an alert if an event is aborted by transformer proxy`

router/router.go
```diff
@@ -551,6 +562,7 @@ func (worker *workerT) processDestinationJobs() {
 
 	for _, destinationJob := range worker.destinationJobs {
 		var attemptedToSendTheJob bool
+		var errorAt string
```
atzoum, line 565: Maintaining an `errorAt` variable without an error having necessarily happened can be confusing for someone reading the code. Would it be better to rename it to `routerStep`?
Same for the constants, e.g. `ERROR_AT_TF` -> `TRANSFORMATION_STEP`

router/router.go
sanpj2292: The detail of name(s) chosen
`allowRtAbortForTf` - allow router-aborted-count alert for transformation
`allowRtAbortForDelv` - allow router-aborted-count alert for event Delivery
These are specific to controlling alert through `router-aborted-count` alert definition

**Note**:
- `Abort` in the names chosen doesn't reflect the (aborted)status of the job/event


Let me know if it makes sense

router/router.go
sanpj2292: The intention here is to specifically send the step at which error occurred while processing an event in router. Hence the naming was chosen to be `errorAt` as it communicates the same thing. But if it is not intuitive, it can be renamed. Please let me know if you think it is not very intuitive

atzoum: It is a subjective comment, since intuitiveness is not something objective. I am not insisting as long as I am the only one with this consideration :)

router/router.go
atzoum: We already have 2 configuration options in the server for controlling when an alert should be sent, why do we need two flags for the alert definition itself too?

Another point is that the purpose of an `alert` tag would be pretty obvious to anyone, whereas I still find it hard to be able to easily decode the meaning of `allowTf` & `allowDel` tags.

sanpj2292: >why do we need two flags for the alert definition itself too?

Basically we want the old alert definition to co-exist with the new one. In that case, we don't want duplicate alerts to be triggered. Hence by sending these tags to the measurement, we are making sure that only alerts that cannot be alerted from new definition should be alerted from older one.

These tags will be used for older alert definition(in operator) so that only those alerts that need to be sent via the older one would be used. Let me know if it clarifies the question

> allowTf & allowDel tags

Does `alertOnTransformation` and `alertOnDelivery` communicate the purpose & is easily understood ?

router/router.go
```diff
@@ -1073,14 +1109,19 @@ func (worker *workerT) postStatusOnResponseQ(respStatusCode int, respBody string
 	}
 }
 
-func (worker *workerT) sendRouterResponseCountStat(status *jobsdb.JobStatusT, destination *backendconfig.DestinationT) {
+func (worker *workerT) sendRouterResponseCountStat(status *jobsdb.JobStatusT, destination *backendconfig.DestinationT, errorAt string) {
 	destinationTag := misc.GetTagName(destination.ID, destination.Name)
+	alert := worker.allowRouterAbortedAlert(errorAt)
 	routerResponseStat := stats.Default.NewTaggedStat("router_response_counts", stats.CountType, stats.Tags{
```
atzoum, line 1115: Should we make sure that we don't erroneously set the `alert` & `errorAt` tags if the status is successful?
```
if status.JobState == jobsdb.Succeeded.State {
   alert = false
   errorAt = ""
}
```

sanpj2292: `router_response_counts` is the input measurement in `router-aborted-count` definition. 
We should probably judge alert in below way when job succeeds
```go
if status.JobState == jobsdb.Succeeded.State {
  alert = !worker.rt.skipRtAbortAlertForTransformation || !worker.rt.skipRtAbortAlertForDelivery
  errorAt = ""
}
```
WDYT ?

atzoum: Why would the `alert` value ever need to be `true` when job's state is `succeeded`? Do we want to alert on successes too?

sanpj2292: Basically this `router_response_counts` is a measurement which is being used in `router-aborted-count`. The alert def looks like this
```md
router_aborted_events/router_response_counts
```
where router_response_counts contains events which are sent to router(as a whole success, failure included)
So it would make sense to include alert tag appropriately so that the definition of percentage still remains intact(in terms of `alert` tag)

Basically, if `errorAt` tag is empty, it means that the event was successful & when `alert` tag is `true` it means we would include this event's response while alerting

atzoum: I believe that `errorAt` might be non-empty even if the job's state is `succeeded`, but you do have a point about the rate calculation.

atzoum (approved):

router/router.go
sanpj2292: Yes, to remove the possibility of success state having an `errorAt`, I have made changes to override `errorAt` with empty string in case of job-success. Can you once verify and let me know if the changes are good ?

router/router.go
```diff
@@ -56,6 +56,15 @@ type tenantStats interface {
 	UpdateWorkspaceLatencyMap(destType, workspaceID string, val float64)
 }
 
+const (
+	// transformation(router or batch)
```
cisse21, line 60: [Optional] Probably move these constants to `router/utils`

cisse21 (approved):

saikumarrs (approved):

router/router.go
sanpj2292: Done. resolving this comment

PR #2515, opened 2022-09-30
feat(warehouse): added base support for s3 datalake deletion as part of regulation API.

abhimanyubabbar:
# Description

As per the regulation api, we are adding support to remove the user information data from s3 datalake.

## Notion Ticket
https://www.notion.so/rudderstacks/S3-Datalake-Data-Deletion-Request-Dev-tasks-dfdba07974da4ffdb4b40af163bfdefc

## Security


regulation-worker/internal/client/client.go
```diff
@@ -33,12 +33,10 @@ func (j *JobAPI) Get(ctx context.Context) (model.Job, error) {
...
 	defer cancel()
 
-	method := "GET"
-	genEndPoint := "/dataplane/workspaces/{workspace_id}/regulations/workerJobs"
-	url := fmt.Sprint(j.URLPrefix, prepURL(genEndPoint, j.WorkspaceID))
+	url := fmt.Sprintf("%s/dataplane/workspaces/%s/regulations/workerJobs", j.URLPrefix, j.WorkspaceID)
 	pkgLogger.Debugf("making GET request to URL: %v", url)
 
-	req, err := http.NewRequestWithContext(ctx, method, url, nil)
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
```
lvrach, line 39: [minor]
```suggestion
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -31,64 +29,35 @@ import (
...
-	}
-}
-
 // returns list of all .json.gz files.
 // NOTE: assuming that all of batch destination have same file system as S3, i.e. flat.
 func (b *Batch) listFiles(ctx context.Context) ([]*filemanager.FileObject, error) {
 	pkgLogger.Debugf("getting a list of files from destination")
+
+	// TODO: What happens if we have files greater than 1000 ?
 	fileObjects, err := b.FM.ListFilesWithPrefix(ctx, "", "", listMaxItem)
```
lvrach, line 54: Valid point, we need to make sure we keep iterating

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -144,115 +113,106 @@ func (*Batch) updateStatusTrackerFile(absStatusTrackerFileName, fileName string)
...
+		}
+	} else {
+		lines := strings.Split(string(byt), "\n")
+		// check if our <jobID> matches with the one in file.
+		// if not, then truncate the file & write new current jobID.
 
-	return decompressedFileName, nil
-}
+		// TODO: Why would this happen ? ( remanent of the past )
+		if lines[0] != jobID {
```
lvrach, line 158: As we discussed, this might be a problem if two jobs are assigned on the same bucket and:
1. job 1 starts, but gets interrupted (gets marked as pending and will be retried later)
2. then job 2 gets pickup 

We need two re-consider this case.

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -144,115 +113,106 @@ func (*Batch) updateStatusTrackerFile(absStatusTrackerFileName, fileName string)
...
+			if err != nil {
+				return nil, fmt.Errorf("error while creating temporary directory: %w", err)
+			}
 
-	cleanCompressedFilePtr, err := os.OpenFile(fileName, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o644)
-	if err != nil {
-		return fmt.Errorf("error while opening file, %w", err)
-	}
-	defer cleanCompressedFilePtr.Close()
+			f, err = os.OpenFile(filepath.Join(statusTrackerTmpDir, StatusTrackerFileName), os.O_CREATE|os.O_RDWR, 0o644)
```
lvrach, line 166: Since we removed `os.O_TRUNC`, if the file already exists, we are going to error, correct? 

It is safer this way, I just wanted to make sure this was the intention behind this change.

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -144,115 +113,106 @@ func (*Batch) updateStatusTrackerFile(absStatusTrackerFileName, fileName string)
...
+			if err != nil {
+				return nil, fmt.Errorf("error while opening file to create status tracker, %w", err)
```
lvrach, line 168: [minor] more idiomatic error messages
```suggestion
				return nil, fmt.Errorf("create status tracker: %w", err)
```

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -144,115 +113,106 @@ func (*Batch) updateStatusTrackerFile(absStatusTrackerFileName, fileName string)
...
-	_, err = cleanCompressedFilePtr.Write(buffer.Bytes())
-	if err != nil {
-		return fmt.Errorf("error while writing cleaned & compressed data:%w", err)
+			if _, err := io.WriteString(f, jobID+"\n"); err != nil {
+				return nil, fmt.Errorf("error while writing to status tracker file:%s, err: %w", StatusTrackerFileName, err)
+			}
+
+		} else {
+			// if yes, then simply read it.
+			response = lines[1 : len(lines)-1]
```
lvrach, line 177: This assumes that the file contains at least one line. We should valid the file before making the assumption, or use a common format for storing this information. 

(not introduced in this PR)

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -144,115 +113,106 @@ func (*Batch) updateStatusTrackerFile(absStatusTrackerFileName, fileName string)
...
 
-	err = b.compress(targetFile, out)
+	absPath, err := filepath.Abs(tmpFilePtr.Name())
 	if err != nil {
-		return fmt.Errorf("error while compressing file: %w", err)
+		return "", fmt.Errorf("error while getting absolute path: %w", err)
 	}
 
-	return nil
+	err = b.FM.Download(context.TODO(), tmpFilePtr, completeFileName)
```
lvrach, line 206: We can pass the context here:
```suggestion
	err = b.FM.Download(ctx, tmpFilePtr, completeFileName)
```

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -338,62 +299,44 @@ func (b *Batch) upload(_ context.Context, uploadFileAbsPath, actualFileName, abs
...
-	_, err = PatternFilePtr.Write(searchObject)
-	if err != nil {
-		pkgLogger.Errorf("error while writing pattern: %v", err)
-		return "", fmt.Errorf("error while writing pattern:%w", err)
+	// S3_DATALAKE is a warehouse destination, so in order
+	// to send events into a warehouse destination, we simply snake_cased
+	// so the gziphandler needs to be created with Snakecasing `user_id` in mind
+	if destType == "S3_DATALAKE" {
+		gzipHandler = filehandler.NewGZIPLocalFileHandler(filehandler.SnakeCase)
 	}
```
lvrach, line 322: What if we introduce additional destinations, will the default path work for them or should we error ?

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -550,13 +435,63 @@ func (bm *BatchManager) Delete(ctx context.Context, job model.Job, destConfig ma
...
+	if err := handler.Read(ctx, sourceFile); err != nil {
+		return fmt.Errorf("unable to parse contents of local file: %s, err: %w", sourceFile, err)
+	}
+
+	if err := handler.RemoveIdentity(ctx, attributes); err != nil {
+		return fmt.Errorf("unable to handle identity removal for attributes: %v, err: %w", nil, err)
+	}
+
+	if err := handler.Write(ctx, targetFile); err != nil {
+		return fmt.Errorf("unable to write to local file: %s, err: %w", targetFile, err)
```
lvrach, line 479: [minor] idiomatic errors
```suggestion
		return fmt.Errorf("parse contents of local file: %s, err: %w", sourceFile, err)
	}

	if err := handler.RemoveIdentity(ctx, attributes); err != nil {
		return fmt.Errorf("identity removal for attributes: %v, err: %w", nil, err)
	}

	if err := handler.Write(ctx, targetFile); err != nil {
		return fmt.Errorf("write to local file: %s, err: %w", targetFile, err)
```

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -550,13 +435,63 @@ func (bm *BatchManager) Delete(ctx context.Context, job model.Job, destConfig ma
...
+func maxRoutines() (int, error) {
+	procAllocated, err := strconv.Atoi(config.GetString("GOMAXPROCS", "32"))
+	if err != nil {
+		return 0, fmt.Errorf("fetching maximum number of go routines: %w", err)
+	}
+
+	maxGoRoutine := 8 * procAllocated
+	return maxGoRoutine, nil
+}
+
```
lvrach, line 494: If I understand your intention here is better to use 

https://pkg.go.dev/runtime#GOMAXPROCS
```suggestion
func maxRoutines() (int) {
	maxGoRoutine := 8 * runtime.GOMAXPROCS(0)
	return maxGoRoutine
}
```
> If n < 1, it does not change the current setting
https://pkg.go.dev/runtime#GOMAXPROCS

Note: we use https://github.com/uber-go/automaxprocs library to determine the correct number of CPU allocated, which considers the cgroup restriction when running in k8s.

Given this I would advice against `config.GetString("GOMAXPROCS", "32")`

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -31,64 +29,35 @@ import (
...
 	StatusTrackerFileName = "rudderDeleteTracker.txt"
-	supportedDestinations = []string{"S3"}
+	supportedDestinations = []string{"S3", "S3_DATALAKE"}
 )
 
 const listMaxItem int64 = 1000
 
-type deleteManager interface {
-	delete(ctx context.Context, patternFilePtr, targetFilePtr string) ([]byte, error)
+type PatternDeleter interface {
```
lvrach, line 38: Is this interface used?

lvrach (commented):
minor or issues not introduced in this PR, but I will need a second pass

regulation-worker/internal/client/client.go
abhimanyubabbar: 👍🏼

regulation-worker/internal/delete/batch/batch.go
abhimanyubabbar: I checked and we do capture the token for next page every-time we list the maximum files first time.

regulation-worker/internal/delete/batch/batch.go
abhimanyubabbar: I get it now. I'll make sure we have a ticket for properly reviewing this.

regulation-worker/internal/delete/batch/batch.go
abhimanyubabbar: Nope, I don't this so. I would remove it now.

regulation-worker/internal/delete/batch/batch.go
abhimanyubabbar: I'll improve this check. I think you are right, we need to handle this correctly. But the problem is we have a check before on the destination we are handling. This check nonetheless needs to be improved.

regulation-worker/internal/delete/batch/batch.go
abhimanyubabbar: I haven't changed this implementation.

https://github.com/rudderlabs/rudder-server/blob/master/regulation-worker/internal/delete/batch/batch.go#L473

abhimanyubabbar: Probably did a bit more code movement, so it's fuzzing the boundaries.

regulation-worker/internal/delete/batch/batch.go
abhimanyubabbar: Old habits 😄

regulation-worker/internal/delete/batch/batch.go
abhimanyubabbar: Good point, I'll come up with a plan for fixing it. Potential nil pointer exception.

services/filemanager/filemanager.go
```diff
@@ -63,6 +63,10 @@ func init() {
 // New returns FileManager backed by configured provider
 func (*FileManagerFactoryT) New(settings *SettingsT) (FileManager, error) {
 	switch settings.Provider {
+	case "S3_DATALAKE":
```
achettyiitr, line 66: Why do we need a duplicate fileMangaer for S3 Datalake when we can use S3?

abhimanyubabbar: The provider comes in from above and it has S3_DATALAKE in it hardcoded. Let me see what change can I make !

lvrach (approved):

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -31,64 +28,28 @@ import (
 
 var (
 	pkgLogger             = logger.NewLogger().Child("batch")
-	regexRequiredSuffix   = regexp.MustCompile(".json.gz$")
 	StatusTrackerFileName = "rudderDeleteTracker.txt"
-	supportedDestinations = []string{"S3"}
+	supportedDestinations = []string{"S3", "S3_DATALAKE"}
 )
 
 const listMaxItem int64 = 1000
```
saurav-malani, line 35: I believe since we have defined `FilesLimit` in `BatchManager` struct, we don't need this anymore.

regulation-worker/internal/delete/batch/filehandler/gzip.go
```diff
@@ -0,0 +1,105 @@
...
+
+	_, err = gw.Write(h.records)
+	if err != nil {
+		return fmt.Errorf("error while writing cleaned & compressed data:%w", err)
+	}
+
+	return nil
+}
+
+func (h *GZIPLocalFileHandler) RemoveIdentity(ctx context.Context, attributes []model.User) error {
```
saurav-malani, line 67: Given that we have changed the logic of `RemoveIdentity`  and since this is going to be one of the most time consuming part of the loop during deletion. I strongly feel we should benchmark the old & new deletion logic. 

I have a feeling that if `sed` knows about all the users that needs to be matched for, it will do some kind of optimization do avoid literally iterating over each `OR` condition of pattern file and look for a match in each line.

regulation-worker/internal/delete/batch/batch.go
abhimanyubabbar: I'll fix it.

regulation-worker/internal/delete/batch/filehandler/local_file.go
```diff
@@ -0,0 +1,21 @@
+package filehandler
```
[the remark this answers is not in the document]
abhimanyubabbar: Usually names are camelcased in golang but filenames are snakecased in popular repository. The name local_file indicates the interface of local file handling. Let me know if you wanna have an offline discussion.

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -130,133 +91,132 @@ func removeCleanedFiles(files []*filemanager.FileObject, cleanedFiles []string)
...
+	if len(lines) >= 1 {
+		return lines[1:], nil
 	}
 
-	decompressedFileName, err := filepath.Abs(decompressedFilePtr.Name())
-	if err != nil {
-		return "", err
+	if err := f.Close(); err != nil {
+		return nil, fmt.Errorf("closing file: %w", err)
 	}
```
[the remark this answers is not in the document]
abhimanyubabbar: This is somewhat a different way to cleanly handle closing of file. Usually we wanna handle the error that comes with file closing, so we handle it multiple times in such a fashion.

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -130,133 +91,132 @@ func removeCleanedFiles(files []*filemanager.FileObject, cleanedFiles []string)
...
+	}
+
 	return nil
 }
 
-// downloads `fileName` locally. And returns empty file, if file not found.
-// Note: download happens concurrently in 5 go routine by default.
-func (b *Batch) download(_ context.Context, completeFileName string) (string, error) {
-	pkgLogger.Debugf("downloading file: %v", completeFileName)
+func (_ *Batch) cleanedFiles(_ context.Context, path string, job *model.Job) ([]string, error) {
```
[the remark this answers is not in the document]
abhimanyubabbar:
```Go doesn't provide automatic support for getters and setters. There's nothing wrong with providing getters and setters yourself, and it's often appropriate to do so, but it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner. The use of upper-case names for export provides the hook to discriminate the field from the method. A setter function, if needed, will likely be called SetOwner.
```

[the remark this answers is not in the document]
abhimanyubabbar: I was reading this in effective go a while back where I came across this advice to not place `get` in the name of the function.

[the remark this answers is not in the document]
abhimanyubabbar: Also why would a pointer be beneficial ? In go I think, array's are passed by value and cost is pretty minimal right ?

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -31,64 +28,28 @@ import (
...
 
 const listMaxItem int64 = 1000
 
-type deleteManager interface {
-	delete(ctx context.Context, patternFilePtr, targetFilePtr string) ([]byte, error)
-}
-
 type Batch struct {
 	mu         sync.Mutex
 	FM         filemanager.FileManager
```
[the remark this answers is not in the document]
abhimanyubabbar: The thing is filehandler needs to be created on processing of every file as they store the intermediate results to allow for cleaner interface. We have added a factory to instantiate a filehandler everytime we see a file of that sorts. So this might not work.

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -338,210 +299,154 @@ func (b *Batch) upload(_ context.Context, uploadFileAbsPath, actualFileName, abs
...
-						err = fmt.Errorf("error while writing to statusTrackerFile: %w", err)
-						return "", err
-					}
-				} else {
-					// if yes, then simply read it.
-					cleanedFiles = lines[1 : len(lines)-1]
-				}
-			}
-			absPath, err := filepath.Abs(statusTrackerFilePtr.Name())
+		fName, err := batch.download(ctx, filepath.Join(prefix, StatusTrackerFileName))
```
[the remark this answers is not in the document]
abhimanyubabbar: Usually within a function name which clearly defines the intent, we can shorten the name to indicate popular structure. I like to indicate files through `f` etc. Wdyt ?

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -338,210 +299,154 @@ func (b *Batch) upload(_ context.Context, uploadFileAbsPath, actualFileName, abs
...
 }
 
-// Delete users corresponding to input userAttributes from a given batch destination
-func (bm *BatchManager) Delete(ctx context.Context, job model.Job, destConfig map[string]interface{}, destName string) model.JobStatus {
-	pkgLogger.Debugf("deleting job: %v", job, "from batch destination: %v", destName)
+func getLocalFileHandlers(destType string) map[string]filehandler.LocalFileHandler {
+	switch destType {
+	case "S3":
+		return map[string]filehandler.LocalFileHandler{
+			".json.gz": filehandler.NewGZIPLocalFileHandler(filehandler.CamelCase),
```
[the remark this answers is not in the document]
abhimanyubabbar: Let me think about it. That can be done.

regulation-worker/internal/delete/batch/filehandler/gzip.go
abhimanyubabbar: Based on the man of sed, -f is just reading the commands from the file and appending it to the list of commands to sed. -e which we have replaced it with does the same thing as well. 

The main approach we have changed is letting sed completely handle the file from which the data needs to be removed and we are doing it in memory. This is something which needs to be benchmarked and checked against.

sundernagesh18:
This feature is good to go. QA Approved.

PR #2516, opened 2022-10-02
Chore.run all tool

lvrach:
# Description

When working locally, we usually do `go run main.go` or `docker-compose up`. 

Those command run rudder-server in embedded app type, with all the components running under a single process.

However, rudder-server can run those components independently. Moreover, this is how we run them in Kubernetes environment. To debug/test how this setup, you need to deploy on your dev kubernetes namespace, which might be a hustle for debugging and minor changes.

So, I have introduced a new command in dev tool that runs all the components locally while taking care of port conflicts.

Screenshot 2022-10-03 at 12 13 06 PM

We should also support the ability to do this with docker-compose and also test with all the components separated. However, that is outside the scope of the PR.

## Notion Ticket

< Replace with Notion Link >

## Security


lvrach:
No stale

PR #2519, opened 2022-10-03
feat: add support for role based auth flag

koladilip:
# Description

Currently we are not able to delete already stored secrets from Config BE DB so we are adding for support roleBasedAuth flag so that users can easily switch between AWS Access Keys and Roles. This enables to revert back to Access keys if users are getting any troubles after migrating to the AWS roles.

## Notion Ticket

https://www.notion.so/AWS-FTR-Support-RoleBasedAuth-Flag-5cfd191922ae4f6eb4a8fc75dca40ac3

## Security


koladilip:
Tested on dev cluster and changes are working as expected.

saikumarrs (approved):

cisse21 (approved):

PR #2521, opened 2022-10-04
fix: jobsDB backup query time

saurav-malani:
# Description

> currently, we see `pre_drop` table pileup if a lot of jobs are failing. Since, we have a limit on number of rows read at a time to 1000 (by default). And, when failure is high we have a lot of columns in status table. Because of while taking backup of a single table requires querying DB a lot of time, which takes really really long time. 

So, to avoid this, we are setting the default value of `LIMIT` to a large value `100,000` since, we are already guarded by `running_payload_size`, to avoid OOM kill.

## Notion Ticket

https://www.notion.so/rudderstacks/remove-LIMIT-on-row-count-from-getBackUpQuery-bde1f44331d444f7a33e3b3f99904470

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -657,7 +657,7 @@ func loadConfig() {
 	config.RegisterIntConfigVariable(10, &maxMigrateOnce, true, 1, "JobsDB.maxMigrateOnce")
 	config.RegisterIntConfigVariable(10, &maxMigrateDSProbe, true, 1, "JobsDB.maxMigrateDSProbe")
 	config.RegisterInt64ConfigVariable(300, &maxTableSize, true, 1000000, "JobsDB.maxTableSizeInMB")
-	//Since, we are guarded by running payload size, setting default value of backupRowsBatchSize to large value.
+	// Since, we are guarded by running payload size, setting default value of backupRowsBatchSize to large value.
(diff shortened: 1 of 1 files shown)


jobsdb/jobsdb.go
```diff
@@ -657,7 +657,8 @@ func loadConfig() {
 	config.RegisterIntConfigVariable(10, &maxMigrateOnce, true, 1, "JobsDB.maxMigrateOnce")
 	config.RegisterIntConfigVariable(10, &maxMigrateDSProbe, true, 1, "JobsDB.maxMigrateDSProbe")
 	config.RegisterInt64ConfigVariable(300, &maxTableSize, true, 1000000, "JobsDB.maxTableSizeInMB")
-	config.RegisterInt64ConfigVariable(1000, &backupRowsBatchSize, true, 1, "JobsDB.backupRowsBatchSize")
+	// Since, we are guarded by running payload size, setting default value of backupRowsBatchSize to large value.
```
chandumlg, line 660: Remove the comment. Unnecessary I feel.

jobsdb/jobsdb.go
```diff
@@ -657,7 +657,8 @@ func loadConfig() {
...
+	config.RegisterInt64ConfigVariable(100000, &backupRowsBatchSize, true, 1, "JobsDB.backupRowsBatchSize")
```
chandumlg, line 661: Let's do 10K?

chandumlg (approved):

atzoum (approved):

cisse21 (approved):

PR #2529, opened 2022-10-06
refactor(warehouse): utility for staging table name and some refactoring

achettyiitr:
# Description

Since staging table logic is distributed across packages. Instead of code repeating for constructing the staging table name use a single utility to get it.
Some code cleanup.
Some utility test cases.

## Notion Ticket

https://www.notion.so/rudderstacks/Refactor-staging-table-name-03572926ad034438b70342b889fea70f

## Security


warehouse/redshift/redshift.go
```diff
@@ -534,9 +523,19 @@ func Connect(cred RedshiftCredentialsT) (*sql.DB, error) {
...
+		  table_name
+		from
+		  information_schema.tables
+		where
+		  table_schema = '%s'
+		  AND table_name like '%s%s';
+	`,
+		rs.Namespace,
+		warehouseutils.StagingTablePrefix(provider),
+		"%",
```
lvrach, line 537: You can use `%%` if you want a literal `%` https://pkg.go.dev/fmt

I find this a bit more readable, although I would prefer using prepared statements here.
```suggestion
		  AND table_name like '%s%%';
	`,
		rs.Namespace,
		warehouseutils.StagingTablePrefix(provider),
```

warehouse/utils/utils.go
```diff
@@ -139,11 +170,22 @@ var DiscardsSchema = map[string]string{
...
 
+var tableNameLimitMap = map[string]int{
+	AZURE_SYNAPSE: 127,
+	BQ:            127,
+	DELTALAKE:     127,
+	MSSQL:         127,
+	POSTGRES:      63,
+	RS:            127,
+	SNOWFLAKE:     127,
+}
```
lvrach, line 181: Ideally we want those values encapsulated on their packages

achettyiitr: Good point.

warehouse/redshift/redshift.go
achettyiitr: Will make the changes.

achettyiitr: Done.

warehouse/utils/utils.go
achettyiitr: Done.

Jayachand (approved):

lvrach (approved):

PR #2531, opened 2022-10-07
chore: include matched template path in gateway's request metrics

atzoum:
# Description

Using the matched template's path instead the url's raw path in order to keep measurement cardinality under control, and be able to group requests for paths that contain path parameters, such as:

* /v1/job-status/{job_run_id}
* /schemas/event-model/{EventID}/key-counts
* /schemas/event-version/{VersionID}/metadata

Screenshot 2022-10-07 at 10 14 14 AM

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=cbebdcd5d77e4a8f84151563e7483019&pm=s)

## Security

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -1368,7 +1368,7 @@ func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
 
 	srvMux := mux.NewRouter()
 	srvMux.Use(
-		middleware.StatMiddleware(ctx),
+		middleware.StatMiddleware(ctx, srvMux),
 		middleware.LimitConcurrentRequests(maxConcurrentRequests),
 	)
 	srvMux.HandleFunc("/v1/batch", gateway.webBatchHandler).Methods("POST")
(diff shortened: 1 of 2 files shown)


gateway/gateway.go
```diff
@@ -1368,7 +1368,7 @@ func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
 
 	srvMux := mux.NewRouter()
 	srvMux.Use(
-		middleware.StatMiddleware(ctx),
+		middleware.StatMiddleware(ctx, srvMux),
```
saurav-malani, line 1371: the idea of passing a `router` to it's own middleware func as an argument looks weird to me. But, I am unable to think of an alternative to `router.Match`.

saurav-malani (approved):

lvrach (approved):

PR #2537, opened 2022-10-07
chore: fix defect and use noCache if database setup fails during cache.Start

Sidddddarth:
# Description

https://app.squadcast.com/incident/634022812e29e30e4da5b511

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/config/backend-config/backend-config.go
+++ b/config/backend-config/backend-config.go
@@ -56,7 +56,7 @@ func disableCache() {
 type noCache struct{}
 
 func (*noCache) Get(context.Context) ([]byte, error) {
-	return nil, nil
+	return nil, fmt.Errorf(`noCache: cache disabled`)
 }
 
 type workspaceConfig interface {
@@ -319,6 +319,7 @@ func Setup(configEnvHandler types.ConfigEnvI) (err error) {
 }
 
 func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string) {
+	var err error
 	ctx, cancel := context.WithCancel(ctx)
 	bc.ctx = ctx
 	bc.cancel = cancel
@@ -329,13 +330,24 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
 		u, _ := identifier.BasicAuth()
 		secret := sha256.Sum256([]byte(u))
 		cacheKey := identifier.ID()
-		bc.cache = cache.Start(
+		bc.cache, err = cache.Start(
 			ctx,
 			secret,
 			cacheKey,
-			bc.Subscribe(ctx, TopicBackendConfig),
+			func() pubsub.DataChannel { return bc.Subscribe(ctx, TopicBackendConfig) },
 		)
+		if err != nil {
+			// the only reason why we should resume by using no cache,
+			// would be if no database configuration has been set
+			if config.IsSet("DB.host") {
+				panic(fmt.Errorf("error starting backend config cache: %w", err))
+			} else {
+				pkgLogger.Warnf("Failed to start backend config cache, no cache will be used: %w", err)
+				bc.cache = &noCache{}
+			}
+		}
 	}
+
 	rruntime.Go(func() {
 		bc.pollConfigUpdate(ctx, workspaces)
 		close(bc.blockChan)
--- a/config/backend-config/internal/cache/cache.go
+++ b/config/backend-config/internal/cache/cache.go
@@ -36,40 +36,40 @@ type cacheStore struct {
 // key is the key to use to store and fetch the config from the cache store
 //
 // ch is the channel to listen on for config updates and store them
-func Start(ctx context.Context, secret [32]byte, key string, ch pubsub.DataChannel) Cache {
+func Start(ctx context.Context, secret [32]byte, key string, channelProvider func() pubsub.DataChannel) (Cache, error) {
 	var (
 		err    error
 		dbConn *sql.DB
 	)
 	// setup db connection
 	dbConn, err = setupDBConn()
+	if err != nil {
+		pkgLogger.Errorf("failed to setup db: %v", err)
+		return nil, err
+	}
 	dbStore := cacheStore{
 		dbConn,
 		secret,
 		key,
-	}
-	if err != nil {
-		pkgLogger.Errorf("failed to setup db: %v", err)
-		return &dbStore
 	}
 
 	// apply migrations
 	err = migrate(dbConn)
 	if err != nil {
 		pkgLogger.Errorf("failed to apply db migrations: %v", err)
-		return &dbStore
+		return nil, err
 	}
 
 	// clear config for other keys
 	err = dbStore.clear(ctx)
 	if err != nil {
 		pkgLogger.Errorf("failed to clear previous config: %v", err)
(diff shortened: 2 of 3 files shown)


config/backend-config/backend-config.go
```diff
@@ -323,19 +323,27 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
...
+		bc.Subscribe(ctx, TopicBackendConfig),
+	)
+
+	if dbCache != nil {
+		bc.cache = dbCache
+	} else {
+		pkgLogger.Info(`couldn't setup config cache in database, disabling cache`)
+		disableCache()
+		bc.cache = cacheOverride
 	}
```
fracasula, line 345: To avoid running unnecessary code if `bc.cache` is set:

```suggestion
	// setup cache
	if bc.cache == nil {
		identifier := bc.Identity()
		u, _ := identifier.BasicAuth()
		secret := sha256.Sum256([]byte(u))
		cacheKey := identifier.ID()
		bc.cache = cache.Start(
			ctx,
			secret,
			cacheKey,
			bc.Subscribe(ctx, TopicBackendConfig),
		)
		if bc.cache == nil {
			bc.cache = cacheOverride
			pkgLogger.Info("Couldn't setup config cache in database, disabling cache")
			disableCache()
		}
	}
```

config/backend-config/backend-config.go
```diff
@@ -323,19 +323,27 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
...
-		secret := sha256.Sum256([]byte(u))
-		cacheKey := identifier.ID()
-		bc.cache = cache.Start(
-			ctx,
-			secret,
-			cacheKey,
-			bc.Subscribe(ctx, TopicBackendConfig),
-		)
+
+	// setup cache
```
atzoum, line 327: why are we reversing the `cacheOverride` semantics?

config/backend-config/internal/cache/cache.go
```diff
@@ -43,28 +43,28 @@ func Start(ctx context.Context, secret [32]byte, key string, ch pubsub.DataChann
 	)
 	// setup db connection
 	dbConn, err = setupDBConn()
+	if err != nil {
```
atzoum, line 46: `Start` should return an error as well

Sidddddarth: should we also be closing the pubsub channel..?

config/backend-config/backend-config.go
```diff
@@ -329,13 +330,19 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
...
+		bc.cache, err = cache.Start(
 			ctx,
 			secret,
 			cacheKey,
 			bc.Subscribe(ctx, TopicBackendConfig),
 		)
+		if err != nil {
+			pkgLogger.Errorf("Error starting cache: %v", err)
+			disableCache()
+			bc.cache = cacheOverride
```
atzoum, line 342:
```suggestion
			bc.cache = &noCache{}
```

config/backend-config/backend-config.go
```diff
@@ -329,13 +330,19 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
 		u, _ := identifier.BasicAuth()
 		secret := sha256.Sum256([]byte(u))
 		cacheKey := identifier.ID()
-		bc.cache = cache.Start(
+		bc.cache, err = cache.Start(
 			ctx,
 			secret,
 			cacheKey,
 			bc.Subscribe(ctx, TopicBackendConfig),
```
atzoum, line 337:
```suggestion
			func() pubsub.DataChannel { return bc.Subscribe(ctx, TopicBackendConfig) },
```

config/backend-config/internal/cache/cache.go
```diff
@@ -36,35 +36,35 @@ type cacheStore struct {
 // key is the key to use to store and fetch the config from the cache store
 //
 // ch is the channel to listen on for config updates and store them
-func Start(ctx context.Context, secret [32]byte, key string, ch pubsub.DataChannel) Cache {
+func Start(ctx context.Context, secret [32]byte, key string, ch pubsub.DataChannel) (Cache, error) {
```
atzoum, line 39: We may only subscribe if a connection has been acquired, so in line 69:
```
ch := channelProvider()
```

```suggestion
func Start(ctx context.Context, secret [32]byte, key string, channelProvider func() pubsub.DataChannel) (Cache, error) {
```

config/backend-config/backend-config.go
```diff
@@ -329,13 +330,19 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
...
 		)
+		if err != nil {
```
atzoum, line 339: @Sidddddarth is it really a valid scenario to have rudder-server running in an environment without a jobsdb?
From what I can tell, at least in production, all pods, including warehouse, have `JOBS_DB` environment variables set.
So the question is, should we proceed by disabling cache, or panic?

config/backend-config/backend-config.go
atzoum, line 342:
```suggestion
			if config.IsSet("DB.host") {
				panic(fmt.Errorf("error starting backend config cache: %w", err))
			} else {
				pkgLogger.Warnf("Failed to start backend config cache, no cache will be used: %v", err)
				bc.cache = &noCache{}
			}
```

config/backend-config/backend-config.go
Sidddddarth: I think we can leave the panic to other components.
But I agree, the premise around the entire thing was that without jobsdb the server crashes anyway, so we never really pursued this idea.

config/backend-config/backend-config.go
Sidddddarth: nice

config/backend-config/backend-config.go
```diff
@@ -329,13 +330,22 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
...
-		bc.cache = cache.Start(
+		bc.cache, err = cache.Start(
 			ctx,
 			secret,
 			cacheKey,
-			bc.Subscribe(ctx, TopicBackendConfig),
+			func() pubsub.DataChannel { return bc.Subscribe(ctx, TopicBackendConfig) },
 		)
+		if err != nil {
+			if config.IsSet("DB.host") {
```
atzoum, line 340:
```suggestion
			if config.IsSet("DB.host") { // the only reason why we should resume by using no cache, would be if no database configuration has been set
```

config/backend-config/backend-config.go
```diff
@@ -329,13 +330,22 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
...
+				panic(fmt.Errorf("error starting backend config cache: %w", err))
+			} else {
+				pkgLogger.Warnf("Failed to start backend config cache, no cache will be used: %w", err)
```
atzoum, line 343: can we add test scenarios for both failures?

atzoum (approved):
one minor comment about adding a comment :)

fracasula (approved):

PR #2538, opened 2022-10-08
test(warehouse): add coverage for warehouse configuration validations

achettyiitr:
# Description

Add test coverage for warehouse configuration validations.

## Notion Ticket

https://www.notion.so/rudderstacks/Test-coverage-for-warehouse-validations-123f5d20034d4c1e92c79d4f30b1a42a

## Security


warehouse/configuration_testing/configuration_testing.go
```diff
@@ -19,16 +19,38 @@ var (
 	fileManagerTimeout      time.Duration
 )
```
lvrach, line 21: It is more idiomatic for the filename to be `configuration_test.go` instead of `configuration_testing.go`

lvrach (approved):

warehouse/configuration_testing/configuration_testing.go
achettyiitr: Unfortunately, it's my bad. This is a package `configuration_test` which doesn't follow the [go package convention](https://go.dev/blog/package-names). Renaming the package name to something more specific.

achettyiitr: Making this package validations.

Jayachand (approved):

PR #2546, opened 2022-10-10
fix(router): preserve event order while draining a previously failed job

atzoum:
# Description

When a job needs to be drained from the router, we cannot drain it early, but we need to push it to the worker's queue, so that all other (waiting) jobs for the same userID have time to flush from the pipeline's buffers, before the job is actually drained.

## Notion Ticket

[Link](https://www.notion.so/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=ca310400e89541129789bb20d104b93b&pm=s)

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -1339,7 +1339,7 @@ func (rt *HandleT) findWorker(job *jobsdb.JobT, throttledAtTime time.Time) (toSe
 
 	if !rt.guaranteeUserEventOrder {
 		// if guaranteeUserEventOrder is false, assigning worker randomly and returning here.
-		toSendWorker = rt.workers[rand.Intn(rt.noOfWorkers)]
+		toSendWorker = rt.workers[rand.Intn(rt.noOfWorkers)] // skipcq: GSC-G404
 		return
 	}
 
--- a/router/router.go
+++ b/router/router.go
@@ -173,8 +173,6 @@ type workerMessageT struct {
 	job                *jobsdb.JobT
 	throttledAtTime    time.Time
 	workerAssignedTime time.Time
-	drain              bool
-	drainReason        string
 }
 
 // workerT a structure to define a worker for sending events to sinks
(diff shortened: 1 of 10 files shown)


router/router.go
```diff
@@ -1271,7 +1311,7 @@ func (rt *HandleT) findWorker(job *jobsdb.JobT, throttledAtTime time.Time) (toSe
 	userID := job.UserID
 	// checking if the user is in throttledMap. If yes, returning nil.
 	// this check is done to maintain order.
-	if _, ok := rt.throttledUserMap[userID]; ok {
+	if _, ok := rt.throttledUserMap[userID]; ok && rt.guaranteeUserEventOrder {
```
atzoum, line 1314: found this was missing

router/router.go
```diff
@@ -1288,9 +1328,16 @@ func (rt *HandleT) findWorker(job *jobsdb.JobT, throttledAtTime time.Time) (toSe
 		return nil
 	}
 
+	defer func() {
+		if toSendWorker == nil {
+			rt.throttler.Dec(parameters.DestinationID, userID, 1, throttledAtTime, throttler.ALL_LEVELS)
+		}
+	}()
```
atzoum, line 1335: Added this since we need to decrement the throttler if no worker is found for the job.

As a general note, the current throttling logic (incrementing/decrementing) appears to be fragile.

cc @fracasula

chandumlg (approved):
Looks good.

Only comment is, to avoid draining in two places (by checking canDrainEarly), we can drain all kinds of jobs in the worker itself. True?

atzoum:
> Looks good.
> 
> Only comment is, to avoid draining in two places (by checking canDrainEarly), we can drain all kinds of jobs in the worker itself. True?

@chandumlg the initial intention was to leave it there as an optimization, i.e. sparing the extra loop for expired jobs that won't be picked up due to another previously failed job. But you are correct, this minor optimization doesn't justify all this extra code. Will simplify it by using a single draining strategy. Thanks!

router/router.go
```diff
@@ -1748,7 +1787,7 @@ func (rt *HandleT) readAndProcess() int {
 
 	rt.lastQueryRunTime = time.Now()
 
-	pickupMap, latenciesUsed := rt.MultitenantI.GetRouterPickupJobs(rt.destName, rt.noOfWorkers, timeOut, jobQueryBatchSize, rt.timeGained)
+	pickupMap, _ := rt.MultitenantI.GetRouterPickupJobs(rt.destName, rt.noOfWorkers, timeOut, jobQueryBatchSize, rt.timeGained)
```
atzoum, line 1790: @cisse21 `latenciesUsed` doesn't appear to be needed anymore after removing job draining from pickupLoop. Please confirm this looks good

cisse21 (approved):

cisse21:
LGTM

PR #2548, opened 2022-10-11
chore: additional deepsource lints 

lvrach:
# Description

Solve the linting issue mentioned in:
* https://deepsource.io/gh/rudderlabs/rudder-server/issues/?category=bug-risk
* https://deepsource.io/gh/rudderlabs/rudder-server/issues/?category=antipattern&page=1
* https://deepsource.io/gh/rudderlabs/rudder-server/issue/GO-S2112/occurrences?listindex=0

## Notion Ticket

https://www.notion.so/rudderstacks/warehouse-linting-fixes-0ae5ee91d9be42ac801248107807903f

## Security


warehouse/client/client.go
```diff
@@ -55,6 +55,8 @@ func (cl *Client) sqlQuery(statement string) (result warehouseutils.QueryResult,
 			switch t := values[i].(type) {
 			case []uint8:
 				values[i] = string(t)
+			default:
+				return result, fmt.Errorf("unexpected type %T", t)
```
lvrach, line 59: @achettyiitr @sivashanmukh, what do you think of this?

achettyiitr: This would cause an issue I think for other datatypes. We added this only because in the later step we are doing `fmt.Sprintf("%+v", values[i])` which tends to print byte array which is difficult to read.
```
       for rows.Next() {
		for i := 0; i < colCount; i++ {
			valuePtrs[i] = &values[i]
		}

		err = rows.Scan(valuePtrs...)
		if err != nil {
			return result, err
		}

		var stringRow []string
		for i := 0; i < colCount; i++ {
			switch t := values[i].(type) {
			case []uint8:
				stringRow = append(stringRow, fmt.Sprintf("%+v", string(t)))
			default:
				stringRow = append(stringRow, fmt.Sprintf("%+v", values[i]))
			}
		}
		result.Values = append(result.Values, stringRow)
	}
```

achettyiitr: Although on a future note, I think that we should get rid of this package and delegate the query to their corresponding destination packages because here we can only run.

achettyiitr: This would cause an issue I think for other datatypes. We added this only because in the later step we are doing `fmt.Sprintf("%+v", values[i])` which tends to print byte array which is difficult to read.
I would like to refactor it in this way.
```
       for rows.Next() {
		for i := 0; i < colCount; i++ {
			valuePtrs[i] = &values[i]
		}

		err = rows.Scan(valuePtrs...)
		if err != nil {
			return result, err
		}

		var stringRow []string
		for i := 0; i < colCount; i++ {
			switch t := values[i].(type) {
			case []uint8:
				stringRow = append(stringRow, fmt.Sprintf("%+v", string(t)))
			default:
				stringRow = append(stringRow, fmt.Sprintf("%+v", values[i]))
			}
		}
		result.Values = append(result.Values, stringRow)
	}
```

warehouse/warehouse.go
```diff
@@ -2028,8 +2028,9 @@ func startWebHandler(ctx context.Context) error {
 	}
 
 	srv := &http.Server{
-		Addr:    fmt.Sprintf(":%d", webPort),
-		Handler: bugsnag.Handler(mux),
+		Addr:              fmt.Sprintf(":%d", webPort),
+		Handler:           bugsnag.Handler(mux),
+		ReadHeaderTimeout: 3 * time.Second,
```
achettyiitr, line 2033: I am curious, how did we come up with this number? Should we make this configurable and keep the default values as 3 seconds?

warehouse/testhelper/setup.go
```diff
@@ -167,17 +167,17 @@ func setUpJobsDB() *JobsDBResource {
 
 	var err error
 	if jobsDB.DB, err = postgres.Connect(*pgCredentials); err != nil {
-		log.Fatalf("could not connect to jobsDb with error: %s", err.Error())
+		log.Panicf("could not connect to jobsDb with error: %s", err.Error())
```
achettyiitr, line 170: [log. Fatal vs log. Panic](https://quasilyte.dev/blog/post/log-fatal-vs-log-panic/)

> The log.Panic vs log.Fatal is essentially panic vs os.Exit(1). The latter will not let deferred calls to run. Most of the time, it’s not what you want. It also makes testing much harder. It’s quite simple to stop panic in test by the means of recover. It’s much harder to cope with a code that does os.Exit somewhere while you’re trying to do end2end testing without loosing coverage info.

Wanted to understand if is it because of the above reason we converted it.

PR #2549, opened 2022-10-11
chore: backup support for new multitenant system

BonapartePC:
# Description

Supporting backup upload for the multitenant system. The following are the changes:
1. We support using self-storage by the customer now
2. In a table with multiple customers, we might have to send different customers to different storage. With this PR, it should be possible.
3. Every gz file will now have a workspace id at the end attached to the file name.
4. Depending on the user preferences, proc_errors and gw dumps will be uploaded if they are set to true.

## Notion Ticket

https://www.notion.so/rudderstacks/Allow-Customers-to-use-self-storage-26f45f75e38b4ef6bd872bea6187aa06
https://www.notion.so/rudderstacks/Divide-customers-events-in-jobs-table-and-upload-them-in-respective-storages-04dbf3dbd7bd435786e6a3808bed73ed

## Security


config/backend-config/types.go
```diff
@@ -88,6 +88,27 @@ type ConfigT struct {
 	Sources         []SourceT       `json:"sources"`
 	Libraries       LibrariesT      `json:"libraries"`
 	ConnectionFlags ConnectionFlags `json:"flags"`
+	Settings        SettingsT       `json:"settings"`
+}
+
+type SettingsT struct {
```
fracasula, line 94: Why do we keep creating structs with the `T` suffix? I thought we agreed that moving forward we would have avoided this.

config/backend-config/types.go
```diff
@@ -88,6 +88,27 @@ type ConfigT struct {
...
+	DataRetention DataRetentionT `json:"dataRetention"`
+}
+
+type DataRetentionT struct {
+	UseSelfStorage     bool                `json:"useSelfStorage"`
+	StorageBucket      StorageBucketT      `json:"storageBucket"`
+	StoragePreferences StoragePreferencesT `json:"storagePreferences"`
```
fracasula, line 101: [optional] the JSON tags here (and in the new structs you created) are redundant: https://go.dev/play/p/F9SAkkWcL2j

If you get `useSelfStorage` or even `useSELFStorage` it doesn't matter, it will be parsed. However you will need a JSON tag if the field in the JSON has more characters like `use_self_storage`.

atzoum:
@BonapartePC #2542 introduced a major refactoring on most backup functions. Furthermore, all backup-related functions are now moved in a separate file, [backup.go](https://github.com/rudderlabs/rudder-server/blob/master/jobsdb/backup.go). @saurav-malani may provide you with some more information regarding the scope of this aforementioned refactoring.
Please resolve all relevant conflicts, in order to proceed with the review.

services/backup/backup.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"context"
+	"sync"
+
+	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
+)
+
+type StorageSettings map[string]backendconfig.SettingsT
+
+type StorageService interface {
```
atzoum, line 12: This looks more like a data retention settings provider, rather than a backup storage service.

IMO, we shouldn't need anything more than the following function in jobsdb.

```
func (jd *HandleT) getBackupFileUploader(ctx context.Context, workspaceID string) (filemanager.FileManager, error)
```

The above function could use the data retention settings provider in order to resolve the latest data retention configuration for this workspace.

fracasula: I agree, this seems tightly coupled to the backend config and this specific use case. It doesn't seem like having it as a lib here in `services` is giving it any reusability. Perhaps it can be simplified a bit and be part of jobsdb instead.

BonapartePC: I have separated this from jobsdb because storage settings need to be updated from the config backend by subscribing to it.

chandumlg:
@BonapartePC Are proc errors (stash.go) not using these storage settings?

jobsdb/backup.go
```diff
@@ -129,50 +189,85 @@ func (jd *HandleT) cleanStatusTable(backupDSRange *dataSetRangeT) error {
...
-	err = jd.createTableDump(getFailedOnlyBackupQueryFn(backupDSRange), path, totalCount)
+	workspaces, err := jd.getAllWorkspaces(backupDSRange.ds.JobTable)
 	if err != nil {
-		return fmt.Errorf("error while creating table dump: %w", err)
+		return err
 	}
-	defer func() { _ = os.Remove(path) }()
 
-	err = jd.uploadTableDump(ctx, path)
+	for i := 0; i < len(workspaces); i++ {
```
saurav-malani, line 220: I don't think it is a good idea to handle each workspace separately, i.e. in current implementation to backup a table, we have to query postgres separately for to deal with each workspace, which is quite inefficient. Instead, I would suggest to handle each workspace in a single DB query maybe by `GROUP BY` or something else. This will reduce the DB load significantly and will also be optimal in terms of time.

jobsdb/backup.go
```diff
@@ -129,50 +189,85 @@ func (jd *HandleT) cleanStatusTable(backupDSRange *dataSetRangeT) error {
...
+		workspace := workspaces[i]
+		totalCount, err := getRowCount(workspace)
+		if err != nil {
+			return err
+		}
+
+		if totalCount == 0 {
+			jd.logger.Infof("[JobsDB] :: Skipping backup for workspace/table since there are no rows: %s/%s", workspace, tableName)
+			return nil
```
saurav-malani, line 229: we shouldn't `return` here, instead `continue`

saurav-malani (changes requested):
In 1st pass just looked at things from optimization perspective. I have highlighted the same in comments.

jobsdb/backup.go
BonapartePC: True

BonapartePC:
> @BonapartePC Are proc errors (stash.go) not using these storage settings?

Currently no. Will add support for them as well

jobsdb/backup.go
BonapartePC: Resolving this since it is outdated and we don't deal with each workspace anymore

services/fileuploader/gzwriter.go
```diff
@@ -0,0 +1,52 @@
...
+
+		err := os.MkdirAll(filepath.Dir(path), os.ModePerm)
+		if err != nil {
+			return 0, fmt.Errorf("creating gz file %q: %w", path, err)
+		}
+
+		g.gzWriters[workspaceID], err = misc.CreateGZ(path)
+		g.dumps[workspaceID] = path
+		if err != nil {
+			return 0, err
```
fracasula, line 39: We could use some context error here considering the above.

services/fileuploader/gzwriter.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/rudderlabs/rudder-server/utils/misc"
+)
+
+type MultiFileWriter interface {
+	Write(workspaceID, path string, data []byte) (int, error)
```
fracasula, line 12: This is a multi file writer, why bothering with workspaceIDs at all? We could rename it to `key` instead and keep it generic. Thoughts?

services/fileuploader/fileuploader.go
```diff
@@ -0,0 +1,150 @@
...
+		GatewayDumps:     true,
+		ProcErrorDumps:   true,
+		RouterDumps:      true,
+		BatchRouterDumps: true,
+	}, nil
+}
+
+func getDefaultBucket(ctx context.Context) backendconfig.StorageBucket {
+	return backendconfig.StorageBucket{
+		Type:   config.GetString("JOBS_BACKUP_STORAGE_PROVIDER", "S3"),
```
fracasula, line 130: Why are we reading env vars here at this stage? Can't we find another way to get this? Like passing the storage provider in the `New` constructor?

services/fileuploader/gzwriter.go
```diff
@@ -0,0 +1,52 @@
...
+		dumps:     make(map[string]string),
+	}
+}
+
+func (g *GZFileHandler) Write(workspaceID, path string, data []byte) (count int, err error) {
+	if _, ok := g.gzWriters[workspaceID]; !ok {
+
+		err := os.MkdirAll(filepath.Dir(path), os.ModePerm)
+		if err != nil {
+			return 0, fmt.Errorf("creating gz file %q: %w", path, err)
```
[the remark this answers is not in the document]
BonapartePC: Made the change

services/fileuploader/gzwriter.go
BonapartePC: Removed dependency on the workspaceID

services/fileuploader/fileuploader.go
BonapartePC: We are mostly using env for testing purposes now and only when customer doesn't provide bucket details

services/fileuploader/gzwriter.go
BonapartePC: Sorry, you want me to check if context has been cancelled here?

atzoum:
> Base: 43.87% // Head: 43.64% // Decreases project coverage by -0.23% ⚠️

we shouldn't be decreasing coverage

atzoum (commented):
My only concern is test coverage.
Such large refactorings should increase overal code coverage instead of decreasing it.

BonapartePC:
> My only concern is test coverage. Such large refactorings should increase overal code coverage instead of decreasing it.

The coverage is a little better now. Most of the impacted files are the ones where the fileuploader is just passed but there are no tests that result in decreasing coverage a bit.

services/fileuploader/fileuploader.go
```diff
@@ -0,0 +1,160 @@
...
+func (p *provider) GetFileManager(workspaceID string) (filemanager.FileManager, error) {
+	<-p.init
+	settings, ok := p.storageSettings[workspaceID]
+	if !ok {
+		return nil, fmt.Errorf("no storage settings found for workspace: %s", workspaceID)
+	}
+	return filemanager.DefaultFileManagerFactory.New(&filemanager.SettingsT{
+		Provider: settings.Bucket.Type,
+		Config:   settings.Bucket.Config,
+	})
```
atzoum, line 69: please add a test scenario for this function

services/fileuploader/fileuploader.go
```diff
@@ -0,0 +1,160 @@
...
+
+// NewStaticProvider creates a new provider that operates against a predefined storage settings.
+// Useful for tests.
+func NewStaticProvider(storageSettings map[string]StorageSettings) Provider {
+	s := &provider{
+		init:            make(chan struct{}),
+		storageSettings: storageSettings,
+	}
+	close(s.init)
+	return s
```
atzoum, line 44: please add a test scenario for this function

services/fileuploader/fileuploader.go
```diff
@@ -0,0 +1,160 @@
...
+			}
+		}
+		p.onceInit.Do(func() {
+			close(p.init)
+		})
+		p.storageSettings = settings
+	}
+
+	p.onceInit.Do(func() {
+		close(p.init)
```
atzoum, line 114: why this appears to be not covered with tests?

services/fileuploader/fileuploader.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+type defaultProvider struct{}
+
+func (*defaultProvider) GetFileManager(_ string) (filemanager.FileManager, error) {
+	defaultConfig := getDefaultBucket(context.Background(), config.GetString("JOBS_BACKUP_STORAGE_PROVIDER", "S3"))
+	return filemanager.DefaultFileManagerFactory.New(&filemanager.SettingsT{
+		Provider: defaultConfig.Type,
+		Config:   defaultConfig.Config,
+	})
```
atzoum, line 125: please add a test scenario for this function

jobsdb/backup.go
```diff
@@ -68,7 +70,6 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
 		opID := jd.JournalMarkStart(backupDSOperation, opPayload)
 		err = jd.backupDS(ctx, backupDSRange)
 		if err != nil {
-			stats.Default.NewTaggedStat("backup_ds_failed", stats.CountType, stats.Tags{"customVal": jd.tablePrefix, "provider": config.GetString("JOBS_BACKUP_STORAGE_PROVIDER", "S3")}).Increment()
```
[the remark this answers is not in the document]
BonapartePC: I had to remove this here because at the point where we are making this stat we can't determine the `provider` tag. This will be different for each workspace which was previously taken from the env.

jobsdb/backup.go
```diff
@@ -147,41 +166,39 @@ func (jd *HandleT) failedOnlyBackup(ctx context.Context, backupDSRange *dataSetR
...
+	err = jd.uploadDumps(ctx, dumps)
 	if err != nil {
-		jd.logger.Errorf("[JobsDB] :: Failed to upload table %v", tableName)
-		return err
+		return fmt.Errorf("error while uploading dumps for table: %s: %w", tableName, err)
 	}
-
 	stats.Default.NewTaggedStat("total_TableDump_TimeStat", stats.TimerType, stats.Tags{"customVal": jd.tablePrefix}).Since(start)
 	return nil
 }
```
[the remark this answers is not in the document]
BonapartePC: We decided to leave it this way so it is clearly shown as different steps being taken

jobsdb/backup.go
```diff
@@ -127,6 +128,24 @@ func (jd *HandleT) cleanStatusTable(backupDSRange *dataSetRangeT) error {
 	return err
 }
 
+func (jd *HandleT) uploadDumps(ctx context.Context, dumps map[string]string) error {
+	g, _ := errgroup.WithContext(ctx)
+	g.SetLimit(config.GetInt("JobsDB.JobsBackupUploadWorkers", 100))
```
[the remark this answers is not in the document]
BonapartePC: Makes sense. added

jobsdb/backup.go
```diff
@@ -147,41 +166,39 @@ func (jd *HandleT) failedOnlyBackup(ctx context.Context, backupDSRange *dataSetR
...
+	dumps, err := jd.createTableDumps(getFailedOnlyBackupQueryFn(backupDSRange), getFileName, totalCount)
 	if err != nil {
 		return fmt.Errorf("error while creating table dump: %w", err)
 	}
-	defer func() { _ = os.Remove(path) }()
-
-	err = jd.uploadTableDump(ctx, path)
+	defer func() {
+		for _, filePath := range dumps {
+			os.Remove(filePath)
```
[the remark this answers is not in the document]
BonapartePC: Made the change

services/fileuploader/fileuploader.go
BonapartePC: Added

services/fileuploader/fileuploader.go
BonapartePC: Done

services/fileuploader/fileuploader.go
BonapartePC: Added a test for that

services/fileuploader/fileuploader.go
BonapartePC: Done

services/fileuploader/gzwriter.go
```diff
@@ -0,0 +1,62 @@
+package fileuploader
```
saurav-malani, line 1: why is the rational behind creating this `services/fileuploader/gzwriter`? I mean it seems to be doing things very specific to our jobsDB backup (atleast as of now). Creating a gzFileHandler which has a map of path & gzipwriter. Where else are we going to use it? If no where, then won’t it be better to move it to jobsdb backup?

BonapartePC: The idea is it can be used anywhere not just jobsdb so I kept it in services folder

saurav-malani (approved):

atzoum (approved):

fracasula (approved):

PR #2554, opened 2022-10-12
fix: to prevent migration of tables every loop after maxdsretention is passed

chandumlg:
# Description

This pr fixes the bug, because of which, tables are migrated in every loop after maxdsretention time is passed. (as long as there are >=1 terminal jobs).

In the new logic, a table which has >= 1 terminal jobs will be eligible for migration only after maxdsretention is passed (apart from other migration conditions)

## Notion Ticket

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=08b5bdb9444d4c75ae89a5930dfc01cd&pm=s

## Security


jobsdb/jobsdb.go
```diff
@@ -1422,8 +1432,8 @@ func (jd *HandleT) createDSInTx(tx *sql.Tx, newDS dataSetT) error {
                                      job_id BIGINT REFERENCES %q(job_id),
                                      job_state VARCHAR(64),
                                      attempt SMALLINT,
-                                     exec_time TIMESTAMP,
-                                     retry_time TIMESTAMP,
+                                     exec_time TIMESTAMP NOT NULL DEFAULT NOW(),
+                                     retry_time TIMESTAMP NOT NULL DEFAULT NOW(),
```
chandumlg, line 1436: Don't feel database migration files are necessary because
a. we inject exec_time & retry_time from our code.
b. these tables are transient.

enterprise/replay/replay.go
```diff
@@ -97,8 +97,8 @@ func (handle *Handler) generatorLoop(ctx context.Context) {
 			status := jobsdb.JobStatusT{
 				JobID:         job.JobID,
 				JobState:      jobsdb.Executing.State,
-				ExecTime:      time.Now(),
-				RetryTime:     time.Now(),
+				ExecTime:      time.Now().UTC(),
```
chandumlg, line 100: To keep parity with the created_at times in jobs table, I had to convert all exec_times & retry_times to UTC.

created_at times in jobs table are assigned by postgres (because of default value) & postgres deals in UTCs.

jobsdb/jobsdb.go
atzoum: why do we have the need to remove these default values?

enterprise/replay/replay.go
atzoum: Do we also parse those dates as UTC when we read rows from the database too?

cisse21 (approved):

jobsdb/jobsdb.go
```diff
@@ -1128,21 +1131,31 @@ func (jd *HandleT) checkIfMigrateDS(ds dataSetT) (
...
 	}
 
-	if jd.MaxDSRetentionPeriod > 0 && time.Since(minCreatedAt) > jd.MaxDSRetentionPeriod {
-		return true, false, recordsLeft
+	if jd.MaxDSRetentionPeriod > 0 {
+		var terminalJobsExist sql.NullBool
+		sqlStatement = fmt.Sprintf(`SELECT EXISTS (
+									SELECT id
+										FROM %q
+										WHERE job_state IN ('%s') and now() - exec_time > interval '%d second' LIMIT 1)`,
```
atzoum, line 1151:
```suggestion
										WHERE job_state IN ('%s') and exec_time < $2)`,
```
Passing `time.Now().Add(jd.MaxDSRetentionPeriod)` as a second argument.
I also believe there is no need to use `limit 1` as the `exists` clause will take care of the necessary optimizations

jobsdb/jobsdb.go
```diff
@@ -1128,21 +1131,31 @@ func (jd *HandleT) checkIfMigrateDS(ds dataSetT) (
...
-		return false, false, recordsLeft
+		if time.Since(maxCreatedAt) < jd.MinDSRetentionPeriod {
+			return false, false, recordsLeft
+		}
 	}
 
-	if jd.MaxDSRetentionPeriod > 0 && time.Since(minCreatedAt) > jd.MaxDSRetentionPeriod {
-		return true, false, recordsLeft
+	if jd.MaxDSRetentionPeriod > 0 {
+		var terminalJobsExist sql.NullBool
```
atzoum, line 1147: is there a case for an exists clause to return null?

jobsdb/integration_test.go
```diff
@@ -495,7 +495,7 @@ func TestJobsDB(t *testing.T) {
 
 		jobDB.MaxDSRetentionPeriod = time.Second
 
-		jobs := genJobs(defaultWorkspaceID, customVal, 1, 1)
+		jobs := genJobs(defaultWorkspaceID, customVal, 10000, 1)
```
atzoum, line 498: why did we have to increase it to such a large number?

chandumlg: To make sure that the migration happened due to maxdsretention check and not because of other conditions.

jobsdb/jobsdb.go
chandumlg: No. Used NullBool and .Valid check as a safe thing.

enterprise/replay/replay.go
chandumlg: 1. True.
2. No
3. Don't think so

We can use `timestamp with time zone` too. But we need to change the way we create jobs table then.

jobsdb/integration_test.go
atzoum: not sure I can follow, can't we achieve the same with a lower number of jobs and different configuration options?

jobsdb/jobsdb.go
atzoum: I don't believe this is needed though

enterprise/replay/replay.go
atzoum: Indeed, we need to adapt our create table statements for jobs & job status tables

chandumlg: done

jobsdb/jobsdb.go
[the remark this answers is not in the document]
chandumlg: done

jobsdb/jobsdb.go
chandumlg: removed

jobsdb/jobsdb.go
chandumlg: It should be `time.Now().Add(-1*jd.MaxDSRetentionPeriod)`. Done the change.

jobsdb/integration_test.go
chandumlg: reduced to meaningful number (10).

jobsdb/jobsdb.go
atzoum: you are correct, sorry for the misleading proposal

atzoum:
Thank you @chandumlg for addressing comments, it looks great!

PR #2555, opened 2022-10-12
chore: excluding destId tag for free customers

chandumlg:
# Description

Exclude destId tag for free tier to prevent tags blowup.

## Security

### DIFF
--- a/main.go
+++ b/main.go
@@ -244,7 +244,7 @@ func Run(ctx context.Context) int {
 	}
 	// TODO: remove as soon as we update the configuration with statsExcludedTags where necessary
 	if !config.IsSet("statsExcludedTags") && deploymentType == deployment.MultiTenantType && (!config.IsSet("WORKSPACE_NAMESPACE") || strings.Contains(config.GetString("WORKSPACE_NAMESPACE", ""), "free")) {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -244,7 +244,7 @@ func Run(ctx context.Context) int {
 	}
 	// TODO: remove as soon as we update the configuration with statsExcludedTags where necessary
 	if !config.IsSet("statsExcludedTags") && deploymentType == deployment.MultiTenantType && (!config.IsSet("WORKSPACE_NAMESPACE") || strings.Contains(config.GetString("WORKSPACE_NAMESPACE", ""), "free")) {
-		config.Set("statsExcludedTags", []string{"workspaceId", "sourceID"})
+		config.Set("statsExcludedTags", []string{"workspaceId", "sourceID", "destId"})
```
atzoum, line 247: so we are not going to move them as config overrides after all?

chandumlg: This [pr](https://github.com/rudderlabs/rudder-server/pull/2462/files) added a new tag. To prevent blow up, added it to statsExcludedTags as a safe check. We can serve it from config as well.

atzoum (approved):

PR #2557, opened 2022-10-12
chore: minor error message change

deepakrai9185720:
# Description
minor error refactoring

## Notion Ticket
https://www.notion.so/rudderstacks/Refactoring-error-message-for-object-storage-validation-9fc7d4329da54a608ebd42bce6432332

## Security

### DIFF
--- a/warehouse/api.go
+++ b/warehouse/api.go
@@ -772,7 +772,7 @@ func validateObjectStorage(ctx context.Context, request *ObjectStorageValidation
 	factory := &filemanager.FileManagerFactoryT{}
 	fileManager, err := factory.New(getFileManagerSettings(request.Type, request.Config))
 	if err != nil {
-		return fmt.Errorf("unable to create file manager: %s", err.Error())
+		return fmt.Errorf("unable to create file manager: \n%s", err.Error())
 	}
 
 	req := validations.DestinationValidationRequest{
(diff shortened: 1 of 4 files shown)


achettyiitr (approved):

warehouse/api.go
```diff
@@ -772,7 +772,7 @@ func validateObjectStorage(ctx context.Context, request *ObjectStorageValidation
 	factory := &filemanager.FileManagerFactoryT{}
 	fileManager, err := factory.New(getFileManagerSettings(request.Type, request.Config))
 	if err != nil {
-		return fmt.Errorf("unable to create file manager: %s", err.Error())
+		return fmt.Errorf("unable to create file manager: \n%s", err.Error())
```
shekhar-rudder, line 775: Is the UI supposed to break the string based on \n and then render them 1 by 1 ?

deepakrai9185720: The \n divide the string internally  into 1+ lines. 1st line represent the explanation of error from our side and then rest line will be the exact error we are getting.

shekhar-rudder: \n won't automatically create line breaks in UI, I need to update the CSS (trivial change). But this will also cause more line breaks if \n is part of the actual error (for example bucket name).

deepakrai9185720: Yeap it will create one extra line, but except first line everything will be the exact error

deepakrai9185720: Also, we can have multiple line break(bucket name case you mentioned)

shekhar-rudder: I mean if the bucket name is "shashank\n", then "\n" won't show up in the error message

deepakrai9185720: Yeap printing the string won't show it up

deepakrai9185720: Hey but we are escaping the string once. That means it will show up. I am 
Screenshot 2022-10-12 at 6 02 46 PM

shekhar-rudder: As long as you are confident that this change won't mess up other parts of the message, I'm good

abhimanyubabbar (approved):
LGTM 👍🏼

PR #2561, opened 2022-10-13
chore: cleanup unused code

cisse21:
# Description

Cleaning up unused code left over from deleting Migrator code

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=0a191ae269f24a34b459f0ec6f15fd00&pm=s)

## Security

### DIFF
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -132,7 +132,6 @@ func (embedded *EmbeddedApp) StartRudderCore(ctx context.Context, options *app.O
 		}))
 	}
 
-	enableGateway := true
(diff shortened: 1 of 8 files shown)


koladilip (commented):
LGTM, but why are we removing enableGateway flag?

cisse21:
It's by default true right , I don't think there is any case in gatewayAppHandler and embeddedAppHandler where we use enableGateway as false

atzoum (approved):

BonapartePC (approved):

PR #2564, opened 2022-10-13
exp(warehouse): mocks and tests for the warehouse module

achettyiitr:
# Description

This is just an experimental branch to verify how many changes we would need to have mocking and testing for the warehouse module:

## Concerns

- **The creation of the mocks was easy because we just need to abstract it out to interface and provide a way of creating of struct.**
- **One of the major issues I faced is with a lot of global in place.**
- **We need to put queries behind proper repositories. DB seems to be a good candidate but I feel that having separate repositories would be beneficial in the longer run.**
- **We need proper structuring of code.**
- **Addition of processor and scheduler.**
- **Since everything resides in 1 package, was getting issues with circular dependency.**
- **Faced issues with Schemahandle since it exposes responsibilities and uses the state as well.**
- **Upload package contains too many responsibilities. We should definitely break it.**
- **Warehouse util contains a lot of this and structs types. For every circular dependency, we used to move it to utils and now it contains too much.**

- **Injection of configs for parallel tests.**

- **QueryBuilders(clean code)?**

## Mocked for the corresponding packages:

1. pgNotifier
2. manager
3. DB
4. Processor
5. Scheduler
6. SchemaHandle
7. TableUpload
8. Upload
9. Validations

## Notion Ticket

https://www.notion.so/rudderstacks/Enabled-errCheck-and-unParam-linters-for-warehouse-746d741345a748ea9354778268ff86c5

## Security


achettyiitr:
Configs needs to be globall injected from top down manner.

PR #2567, opened 2022-10-13
fix(jobsdb): update cache after transaction completes

atzoum:
# Description

By updating (invalidating) the cache entries before the transaction commits and for the period until the transaction actually commits, we are risking to have other select queries who arrive during this timeframe to populate the cache with incorrect values: until the transaction commits, changes are not visible to other queries, thus they may erroneously put "no result" entries in the cache.

The above can have as a side-effect to: 
1. Delay processing of whole datasets until cache entries expire (observed with `proc_error` tables)
2. Cause out-of-order event delivery

This fix introduces a `jobsdb.Tx` that wraps a `sql.Tx`, adding support for registering success listeners. Now, all cache entries are updated after the transaction has been successfully committed.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=624bf10d7947440c90b7399e8f190086&pm=s)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -701,7 +701,7 @@ func loadConfig() {
 	config.RegisterDurationConfigVariable(5, &addNewDSLoopSleepDuration, true, time.Second, []string{"JobsDB.addNewDSLoopSleepDuration", "JobsDB.addNewDSLoopSleepDurationInS"}...)
 	config.RegisterDurationConfigVariable(5, &refreshDSListLoopSleepDuration, true, time.Second, []string{"JobsDB.refreshDSListLoopSleepDuration", "JobsDB.refreshDSListLoopSleepDurationInS"}...)
 	config.RegisterDurationConfigVariable(5, &backupCheckSleepDuration, true, time.Second, []string{"JobsDB.backupCheckSleepDuration", "JobsDB.backupCheckSleepDurationIns"}...)
-	config.RegisterDurationConfigVariable(60, &cacheExpiration, true, time.Minute, []string{"JobsDB.cacheExpiration"}...)
+	config.RegisterDurationConfigVariable(5, &cacheExpiration, true, time.Minute, []string{"JobsDB.cacheExpiration"}...)
 	useJoinForUnprocessed = config.GetBool("JobsDB.useJoinForUnprocessed", true)
 }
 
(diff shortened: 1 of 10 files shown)


jobsdb/jobsdb.go
```diff
@@ -668,7 +701,7 @@ func loadConfig() {
 	config.RegisterDurationConfigVariable(5, &addNewDSLoopSleepDuration, true, time.Second, []string{"JobsDB.addNewDSLoopSleepDuration", "JobsDB.addNewDSLoopSleepDurationInS"}...)
 	config.RegisterDurationConfigVariable(5, &refreshDSListLoopSleepDuration, true, time.Second, []string{"JobsDB.refreshDSListLoopSleepDuration", "JobsDB.refreshDSListLoopSleepDurationInS"}...)
 	config.RegisterDurationConfigVariable(5, &backupCheckSleepDuration, true, time.Second, []string{"JobsDB.backupCheckSleepDuration", "JobsDB.backupCheckSleepDurationIns"}...)
-	config.RegisterDurationConfigVariable(60, &cacheExpiration, true, time.Minute, []string{"JobsDB.cacheExpiration"}...)
+	config.RegisterDurationConfigVariable(5, &cacheExpiration, true, time.Minute, []string{"JobsDB.cacheExpiration"}...)
```
atzoum, line 704: 5 minutes should be more than enough and could potentially alleviate the consequences of having an incorrectly populated cache for too long.

jobsdb/jobsdb.go
```diff
@@ -124,51 +124,84 @@ type statTags struct {
...
+
+// AddSuccessListener registers a listener to be executed after the transaction has been committed successfully.
+func (tx *Tx) AddSuccessListener(listener func()) {
+	tx.successListeners = append(tx.successListeners, listener)
+}
+
+// Commit commits the transaction and executes all listeners.
+func (tx *Tx) Commit() error {
+	err := tx.Tx.Commit()
+	if err == nil {
```
cisse21, line 142: Any reason why we are not checking the error early and returning?

cisse21 (approved):

jobsdb/jobsdb.go
atzoum: In the future we might want to introduce other listeners as well, such as `errorListeners` and `completionListeners`. In such a case returning early wouldn't be relevant.
Now it can be done, but I preferred having one less loc instead :)

```
if err != nil {
	return err
}
for _, successListener := range tx.successListeners {
	successListener()
}
return nil
```

chandumlg (approved):

PR #2569, opened 2022-10-14
feat: support initialisation of producer without topic for kafka, azure event hub and confluent cloud

utsabc:
# Description

The scope of this PR is to support KAFKA, Azure event hub and confluent cloud producer with multiple topic from a single client. We want to generate the kafka client without the context of producer where in we will be passing topic information at individual message level.

Transformer takes the responsibility of allocating correct topic information to messages

To support backward compatibility the server client falls back to default topic present in destination config if the message level topic is not present.

Also from transformer to support fair event ordering for topics the logic that is added are as follows

------------------------------------------

events --> server --> transformer (transform each message) ---> server ---> client --> delivery
         [rudderId = anonymousId <<>>userID ]                  [rudderId = anonymousId<<>>userId<<>>topic]
                 
## Notion Ticket

[Notion Ticket](https://www.notion.so/rudderstacks/Kafka-Support-for-multiple-topics-d0e1f2edbe274d13a7b52530a344ea2e)

## Security


services/streammanager/kafka/client/producer.go
```diff
@@ -36,7 +36,7 @@ type Producer struct {
 }
 
 // NewProducer instantiates a new producer. To use it asynchronously just do "go p.Publish(ctx, msgs)".
-func (c *Client) NewProducer(topic string, producerConf ProducerConfig) (p *Producer, err error) { // skipcq: CRT-P0003
+func (c *Client) NewProducer(topic string, producerConf ProducerConfig, skipTopicInitialisation bool) (p *Producer, err error) { // skipcq: CRT-P0003
```
koladilip, line 39: We don't need to change this, we can pass empty topic name and rest of the logic will adjust automatically

services/streammanager/kafka/kafkamanager.go
```diff
@@ -311,11 +313,14 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (*Produ
 	if err = c.Ping(ctx); err != nil {
 		return nil, fmt.Errorf("could not ping: %w", err)
 	}
-
+	skipTopicInitialisation := false
```
koladilip, line 316: no need of this.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -581,7 +594,7 @@ func sendBatchedMessage(ctx context.Context, jsonData json.RawMessage, p produce
 	return 200, returnMessage, returnMessage
 }
 
-func sendMessage(ctx context.Context, jsonData json.RawMessage, p producerManager, topic string) (int, string, string) {
+func sendMessage(ctx context.Context, jsonData json.RawMessage, p producerManager, topic string, config configuration) (int, string, string) {
```
koladilip, line 597: remove topic argument, just handle logic using config

krishna2020:
Notion link: https://www.notion.so/rudderstacks/Kafka-Destination-with-Multiple-Topics-85cbce26706d49eca6853b972e5497b0

services/streammanager/kafka/client/producer.go
```diff
@@ -132,6 +124,19 @@ func (p *Producer) Publish(ctx context.Context, msgs ...Message) error {
 
 var tempError interface{ Temporary() bool }
 
+func processHeaderForEachMessage(msg Message) (headers []kafka.Header) {
```
atzoum, line 127: minor
```suggestion
func headers(msg Message) (headers []kafka.Header) {
```

services/streammanager/kafka/client/producer.go
```diff
@@ -109,17 +109,9 @@ func (p *Producer) Close(ctx context.Context) error {
...
-		if l := len(msgs[i].Headers); l > 0 {
-			headers = make([]kafka.Header, l)
-			for k := range msgs[i].Headers {
-				headers[k] = kafka.Header{
-					Key:   msgs[i].Headers[k].Key,
-					Value: msgs[i].Headers[k].Value,
-				}
-			}
-		}
+		headers := processHeaderForEachMessage(msgs[i])
```
atzoum, line 112:
```suggestion
		headers := headers(msgs[i])
```

services/streammanager/kafka/client/producer.go
```diff
@@ -109,17 +109,9 @@ func (p *Producer) Close(ctx context.Context) error {
...
 		messages[i] = kafka.Message{
+			Topic:   msgs[i].Topic,
```
atzoum, line 114: `client.Producer` should know if it has been initialized with a topic or not and only include the message's topic in the latter case.
Additionally, this function should return an error if the producer is not initialized with a topic and no topic is included in any of the messages.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -450,14 +451,17 @@ func serializeAvroMessage(value []byte, codec goavro.Codec) ([]byte, error) {
...
 	[]client.Message, error,
 ) {
 	start := now()
 	defer func() { kafkaStats.prepareBatchTime.SendTiming(since(start)) }()
-
 	var messages []client.Message
+	topic := conf.Topic
 	for i, data := range batch {
+		if conf.MultiTopicSupport {
+			topic = data["topic"].(string)
```
atzoum, line 463: shouldn't we fallback to the default topic if the message doesn't contain one?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -611,7 +617,11 @@ func sendMessage(ctx context.Context, jsonData json.RawMessage, p producerManage
 			return makeErrorResponse(fmt.Errorf("unable to serialize event with messageId: %s, with error %s", messageId, err))
 		}
 	}
+	if config.MultiTopicSupport {
+		topic = parsedJSON.Get("topic").Value().(string)
```
atzoum, line 621: type cast will panic here if the field is absent or not a string. Additionally, shouldn't we fallback to the default topic if the message doesn't contain one?

atzoum (commented):
We need some test cases for all the new scenarios introduced for `client.Producer`:

1. Initialize a producer with a topic and send a message containing a topic, the message's topic shall be ignored
2. Initialize a producer without a topic and try to send a message containing no topic, an error should be returned

Furthermore, we need test scenarios for kafkamanager as well covering

1. Reading the topic from the message
2. Falling back to the default topic if no topic is provided by the message
3. Handling invalid topic fields (e.g. number, object, nil)

services/streammanager/kafka/kafkamanager.go
```diff
@@ -312,7 +313,7 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (*Produ
 		return nil, fmt.Errorf("could not ping: %w", err)
 	}
 
-	p, err := c.NewProducer(destConfig.Topic, client.ProducerConfig{
+	p, err := c.NewProducer("", client.ProducerConfig{
```
fracasula, line 316: If this is how you want to handle topics then why keeping the argument in the constructor? It looks like you'll be passing it each time anyway even for old configurations.

fracasula:
> We need some test cases for all the new scenarios introduced for `client.Producer`:
> 
> 1. Initialize a producer with a topic and send a message containing a topic, the message's topic shall be ignored
> 2. Initialize a producer without a topic and try to send a message containing no topic, an error should be returned

To simplify a bit I'd remove the `topic` argument from `NewProducer` and always demand it on `Publish`. I'm suggesting this because of the approach proposed in this PR it looks like we would always pass it on `Publish` anyway so we can clean up the constructor and have only a single scenario to test there.

services/streammanager/kafka/client/producer.go
```diff
@@ -36,7 +36,7 @@ type Producer struct {
 }
 
 // NewProducer instantiates a new producer. To use it asynchronously just do "go p.Publish(ctx, msgs)".
-func (c *Client) NewProducer(topic string, producerConf ProducerConfig, skipTopicInitialisation bool) (p *Producer, err error) { // skipcq: CRT-P0003
+func (c *Client) NewProducer(topic string, producerConf ProducerConfig) (p *Producer, err error) { // skipcq: CRT-P0003
```
koladilip, line 39: we can get rid of topic params as per @fracasula suggestion

services/streammanager/kafka/client/producer.go
utsabc: this is addressed

services/streammanager/kafka/kafkamanager.go
utsabc: this is addressed

services/streammanager/kafka/kafkamanager.go
```diff
@@ -418,7 +423,7 @@ func NewProducerForConfluentCloud(destination *backendconfig.DestinationT, o com
 	p, err := c.NewProducer(destConfig.Topic, client.ProducerConfig{
```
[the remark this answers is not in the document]
utsabc: this is addressed

services/streammanager/kafka/kafkamanager.go
utsabc: this is addressed

services/streammanager/kafka/client/producer.go
```diff
@@ -109,17 +113,26 @@ func (p *Producer) Close(ctx context.Context) error {
...
+			Headers: headers,
 		}
+	}
+
+	return p.writer.WriteMessages(ctx, messages...)
+}
+
+// PublishWithTopic allows the production of one or more message to Kafka
+// with a specific topic.
+func (p *Producer) PublishWithTopic(ctx context.Context, msgs ...Message) error {
```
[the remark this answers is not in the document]
utsabc: this is addressed

services/streammanager/kafka/client/producer.go
utsabc: This is refactored and addressed

services/streammanager/kafka/client/producer.go
utsabc: This is refactored and addressed

services/streammanager/kafka/client/producer.go
utsabc: We plan to go with the approach where we always initialize the producer without client and we intend to always send each message with topic information. That proper assignment of topic to each message would be handled at transformer

services/streammanager/kafka/kafkamanager.go
utsabc: This is handled with transformer being the responsible service for assigning to topic to each message

services/streammanager/kafka/kafkamanager.go
utsabc: Handled similar vulnerabilities

services/streammanager/kafka/kafkamanager.go
utsabc: Yes we are planning the same approach updated for the same

services/streammanager/kafka/client/producer.go
utsabc: Addressed

services/streammanager/kafka/client/producer.go
```diff
@@ -132,6 +123,19 @@ func (p *Producer) Publish(ctx context.Context, msgs ...Message) error {
 
 var tempError interface{ Temporary() bool }
 
+func headers(msg Message) (headers []kafka.Header) {
```
koladilip, line 126: rename to prepareHeaders

koladilip (commented):
Looks like tests are failing, check once, also make sure to add unit test cases for the code you have modified

services/streammanager/kafka/kafkamanager.go
```diff
@@ -547,27 +549,31 @@ func (p *ProducerManager) Produce(jsonData json.RawMessage, destConfig interface
 		return makeErrorResponse(err) // returning 500 for retrying, in case of bad configuration
 	}
 
-	if conf.Topic == "" {
+	parsedJSON := gjson.ParseBytes(jsonData)
```
krishna2020, line 552: Here, parsedJSON can be single message or batchMessage. For batchMessage, will topic be available at the outer json ?

utsabc: Yes for batch it will be available at outer json we will batch the messages based on topic

services/streammanager/kafka/kafkamanager.go
```diff
@@ -450,14 +451,21 @@ func serializeAvroMessage(value []byte, codec goavro.Codec) ([]byte, error) {
...
 	defer func() { kafkaStats.prepareBatchTime.SendTiming(since(start)) }()
-
 	var messages []client.Message
+
 	for i, data := range batch {
+		topic := data["topic"].(string)
+
+		// we are guranatee presence of topic in each payload this is a fall back handler
+		if topic == "" {
+			return nil, fmt.Errorf("topic is empty for message %d", i)
```
koladilip, line 466: please make sure that this error is treated at 400 abortable error

utsabc: Yes this will be 400 outside where its handled

services/streammanager/kafka/kafkamanager.go
```diff
@@ -450,14 +451,21 @@ func serializeAvroMessage(value []byte, codec goavro.Codec) ([]byte, error) {
...
+func prepareBatchOfMessages(batch []map[string]interface{}, timestamp time.Time, p producerManager) (
 	[]client.Message, error,
 ) {
 	start := now()
 	defer func() { kafkaStats.prepareBatchTime.SendTiming(since(start)) }()
-
 	var messages []client.Message
+
 	for i, data := range batch {
+		topic := data["topic"].(string)
```
fracasula, line 462: Let's not be this relaxed and try to avoid panics:

```go
topic, ok := data["topic"].(string)
if !ok {
    // handle error
}
```

services/streammanager/kafka/kafkamanager.go
koladilip: ref: https://play.golang.com/p/YTTFP5ZSc_C
What about following?
```go
topic, ok := data["topic"].(string)
if !ok || topic == "" {
  return nil, fmt.Errorf("topic is invalid for message %d", i)
}
```

services/streammanager/kafka/kafkamanager.go
```diff
@@ -529,45 +546,32 @@ func (p *ProducerManager) Publish(ctx context.Context, msgs ...client.Message) e
 
 // Produce creates a producer and send data to Kafka.
 func (p *ProducerManager) Produce(jsonData json.RawMessage, destConfig interface{}) (int, string, string) {
+	fmt.Println("KAFKA KAFKA KAFKA" + string(jsonData))
```
koladilip, line 549: Should be removed

services/streammanager/kafka/kafkamanager.go
```diff
@@ -594,11 +598,11 @@ func sendMessage(ctx context.Context, jsonData json.RawMessage, p producerManage
...
 
 	timestamp := time.Now()
-	userID, _ := parsedJSON.Get("userId").Value().(string)
+	userID := parsedJSON.Get("userId").String()
 	codecs := p.getCodecs()
 	if len(codecs) > 0 {
-		schemaId, _ := parsedJSON.Get("schemaId").Value().(string)
-		messageId, _ := parsedJSON.Get("message.messageId").Value().(string)
+		schemaId := parsedJSON.Get("schemaId").String()
+		messageId := parsedJSON.Get("message.messageId").String()
```
koladilip, line 605: what happens if parsedJSON.Get("schemaId") is not string or doesn't exist?

services/streammanager/kafka/kafkamanager.go
krishna2020: Error message can be extended to "topic is invalid for the event, it should be non-empty string %d"

testhelper/destination/transformer.go
```diff
@@ -21,15 +21,15 @@ func SetupTransformer(pool *dockertest.Pool, d cleaner) (*TransformerResource, e
 	// pulls an image first to make sure we don't have an old cached version locally,
 	// then it creates a container based on it and runs it
 	err := pool.Client.PullImage(docker.PullImageOptions{
-		Repository: "rudderlabs/rudder-transformer",
-		Tag:        "latest",
+		Repository: "rudderlabs/develop-rudder-transformer",
```
krishna2020, line 24: will this get reverted post the transformer PR merge ?

utsabc: Yes its added now for testing purpose

utsabc:
> It this ready for review now? I still see some failures in the PR checks. Some linting issues and test failures.

Not yet will be cleaning up and pushing the changes by EOD

services/streammanager/kafka/kafkamanager.go
utsabc: yes addressed

services/streammanager/kafka/kafkamanager.go
utsabc: It will be handled as  `""` if not present if its other type that is also handled here
https://github.com/tidwall/gjson/blob/8bf80a300b41c9f7c43406ee300606b217cf08c2/gjson.go#L72

koladilip (approved):

services/streammanager/kafka/kafkamanager_test.go
```diff
@@ -214,7 +214,7 @@ func TestNewProducerForAzureEventHubs(t *testing.T) {
 
 		require.Eventually(t, func() bool {
 			ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
-			err = p.p.Publish(ctx, client.Message{Key: []byte("key-01"), Value: []byte("message-01")})
+			err = p.p.Publish(ctx, client.Message{Key: []byte("key-01"), Value: []byte("message-01"), Topic: destConfig["topic"].(string)})
```
crownedlake, line 217: @utsabc, I saw the changes and could see you have used a single-letter variable name for the `Producer` struct. Generally, you can try to use more meaningful names. Single var names are okay if you are creating a library but if this logic will be worked upon in future also then it is very difficult to read something like `p.p` when compared to `someProducer.produce` or `someProducer.childProducer`. (I couldn't understand what the second p is doing).

services/streammanager/kafka/kafkamanager.go
utsabc: This is addressed

services/streammanager/kafka/kafkamanager_test.go
utsabc: Yeah this was done as part of original implementation, I didn't want to make refactor there as it was not part of the original scope. But I see your point

krishna2020 (approved):

services/streammanager/kafka/kafkamanager.go
```diff
@@ -533,41 +544,27 @@ func (p *ProducerManager) Produce(jsonData json.RawMessage, destConfig interface
...
-
 	ctx, cancel := context.WithTimeout(context.TODO(), p.getTimeout())
 	defer cancel()
 	if kafkaBatchingEnabled {
-		return sendBatchedMessage(ctx, jsonData, p, conf.Topic)
+		return sendBatchedMessage(ctx, jsonData, p)
 	}
-	return sendMessage(ctx, jsonData, p, conf.Topic)
+
+	return sendMessage(ctx, jsonData, p)
```
atzoum, line 556: How do we guarantee that we are backwards compatible? We are totally ignoring the Topic in the destination's configuration...

I am personally afraid of relying exclusively to the transformer for setting the topic, without having a fallback in the server, since bugs in transformer are not uncommon. They way it seems now, a message will be aborted, aka lost, if transformer fails to populate the topic properly.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -29,17 +29,18 @@ type avroSchema struct {
...
+	Topic             string
+	HostName          string
+	Port              string
+	SslEnabled        bool
+	CACertificate     string
+	UseSASL           bool
+	SaslType          string
+	Username          string
+	Password          string
+	MultiTopicSupport bool
```
atzoum, line 41: Why are we introducing this config option if nobody is to use it?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -450,14 +451,24 @@ func serializeAvroMessage(value []byte, codec goavro.Codec) ([]byte, error) {
 	return binary, nil
 }
 
-func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timestamp time.Time, p producerManager) (
+func prepareBatchOfMessages(batch []map[string]interface{}, timestamp time.Time, p producerManager) (
```
atzoum, line 454: Can we returned message batches grouped by topic? Even if transformer might do something similar, it won't hurt doing it here. In the worst case we will be returning only 1 group.
see

atzoum (changes requested):

services/streammanager/kafka/kafkamanager.go
atzoum: opensource users don't always update their transformers, do they? Plus I don't feel safe with this decision (have commented about it elswhere)

services/streammanager/kafka/kafkamanager.go
utsabc: I see your point we can fall back on the config topic here as well. As black box we can assume that transformer might not always send the topic as part of message

services/streammanager/kafka/client/producer.go
fracasula: Given that we decided to go down this road, can you please make it mandatory and raise an error here if the topic is empty?

```go
	for i := range msgs {
		if msgs[i].Topic == "" {
			return fmt.Errorf("no topic provided for message %d", i)
		}
```

I think we should also have a test for this (don't specify topic and check that an error is returned).

services/streammanager/kafka/kafkamanager.go
fracasula: Please remove this, it's not being used.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -450,14 +451,24 @@ func serializeAvroMessage(value []byte, codec goavro.Codec) ([]byte, error) {
...
-
 	var messages []client.Message
+
 	for i, data := range batch {
+		topicI, ok := data["topic"]
+		if !ok {
+			return nil, fmt.Errorf("topic is empty for message %d", i)
+		}
+
+		topic, ok := topicI.(string)
```
fracasula, line 467:
```suggestion
		topic, ok := data["topic"].(string)
		if !ok || topic == "" {
			return nil, fmt.Errorf("empty or invalid topic for message %d", i)
		}
```

services/streammanager/kafka/kafkamanager.go
fracasula: I see your point @atzoum but I'm not sure that having some extra logic here just to safeguard us from a bug from another component (i.e. transformer) that we control is such a good idea. Can't we just have a test about this in the transformer codebase instead?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -533,41 +544,27 @@ func (p *ProducerManager) Produce(jsonData json.RawMessage, destConfig interface
...
-	}
-	err = json.Unmarshal(jsonConfig, &conf)
-	if err != nil {
-		return makeErrorResponse(err) // returning 500 for retrying, in case of bad configuration
-	}
-
-	if conf.Topic == "" {
-		return makeErrorResponse(fmt.Errorf("invalid destination configuration: no topic"))
-	}
-
```
fracasula, line 553: You're removing these sanity checks from here but where are we checking if `destConfig` is a valid JSON now? If we're planning to drop this check please rename `destConfig` to `_` (function parameter).

services/streammanager/kafka/kafkamanager_test.go
fracasula: I wrote the original implementation this way because I find the pattern common in Go codebases where one struct is just a wrapper of another so the second `p` is just the internal producer.

If it's not readable for you we can rename it in another PR to something like `p.client` or `p.internalClient`.

services/streammanager/kafka/kafkamanager.go
fracasula: I was under the impression that this was not needed. Please @utsabc clarify what it's done on the transformer side. Can you provide a full example with a data sample of an input coming to the transformer and the output produced by the transformer?

fracasula (changes requested):

fracasula:
@utsabc you got some DeepSource failures, can you please look into them?

services/streammanager/kafka/kafkamanager.go
utsabc: I have done what @atzoum has asked but I haven't pushed it yet @koladilip @krishna2020 would like to hear your thoughts too

atzoum: @fracasula we need to be backwards compatible too with respect to the transformer versions we support

services/streammanager/kafka/kafkamanager.go
utsabc: Yes sure this is the [function](https://github.com/rudderlabs/rudder-transformer/blob/f76b4c18825fd94116dfba21b9e9fc213c1d8e71/v0/destinations/kafka/transform.js#L22) where during batch phase we are receiving events form server and we are grouping them by topic. The batched request for all events having same topics are sent as one request from transformer, we did this so that failure in batch for a topic has no side effects on events for another topic in batch scenario

This is an input test [case](https://github.com/rudderlabs/rudder-transformer/blob/master/__tests__/data/kafka_batch_input.json) for batch function I shared above 
and this is the [output](https://github.com/rudderlabs/rudder-transformer/blob/master/__tests__/data/kafka_batch_output.json)
Essentially this is our unit test case for batch scneario

services/streammanager/kafka/kafkamanager.go
fracasula: If we have to be compatible with prior versions of the transformer then yeah it makes total sense :+1:

services/streammanager/kafka/kafkamanager.go
utsabc: This is addressed

services/streammanager/kafka/kafkamanager.go
utsabc: This is addressed

services/streammanager/kafka/kafkamanager.go
utsabc: This is addressed

services/streammanager/kafka/kafkamanager.go
utsabc: The configs are used now, added the sanity test cases as well

services/streammanager/kafka/client/producer.go
```diff
@@ -109,17 +108,12 @@ func (p *Producer) Close(ctx context.Context) error {
...
-		if l := len(msgs[i].Headers); l > 0 {
-			headers = make([]kafka.Header, l)
-			for k := range msgs[i].Headers {
-				headers[k] = kafka.Header{
-					Key:   msgs[i].Headers[k].Key,
-					Value: msgs[i].Headers[k].Value,
-				}
-			}
+		if msgs[i].Topic == "" {
+			return fmt.Errorf("no topic provided for message %d", i)
```
atzoum, line 112: looks like we are missing coverage for this scenario

services/streammanager/kafka/kafkamanager.go
```diff
@@ -450,14 +450,19 @@ func serializeAvroMessage(value []byte, codec goavro.Codec) ([]byte, error) {
...
 ) {
 	start := now()
 	defer func() { kafkaStats.prepareBatchTime.SendTiming(since(start)) }()
-
 	var messages []client.Message
+
 	for i, data := range batch {
+		topic, ok := data["topic"].(string)
+		if !ok || topic == "" {
+			topic = defaultTopic
```
atzoum, line 463: missing test coverage

utsabc: I have added this test case here already https://github.com/rudderlabs/rudder-server/blob/560fa4ef61f6d4cd7578c9309528c1e2bb2be047/services/streammanager/kafka/kafkamanager_test.go#L817

utsabc: This is addressed

atzoum:
@utsabc if this is actually the final product, can you please take care of the PR title and description accordingly?

utsabc:
> @utsabc if this is actually the final product, can you please take care of the PR title and description accordingly?

Yes

services/streammanager/kafka/client/producer.go
fracasula: I think codecov complains here because it is covered from another package. We don't have a test about this in `client_test.go` where it is its natural fit. It's OK anyway I guess since it is covered.

services/streammanager/kafka/kafkamanager_test.go
```diff
@@ -800,6 +838,17 @@ func TestSendMessage(t *testing.T) {
 	})
 }
 
+func TestPublish(t *testing.T) {
+	t.Run("no topic publish", func(t *testing.T) {
```
fracasula, line 842: The main test is `TestPublish`, no need to repeat `publish`:

```suggestion
	t.Run("no topic", func(t *testing.T) {
```

services/streammanager/kafka/kafkamanager_test.go
```diff
@@ -800,6 +838,17 @@ func TestSendMessage(t *testing.T) {
...
+		kafkaStats.publishTime = getMockedTimer(t, gomock.NewController(t))
+		p := &client.Producer{}
+		pm := &ProducerManager{p: p}
+		message := prepareMessage("", "userId", []byte(`{"test": "value"}`), time.Now())
+		err := publish(context.Background(), pm, message)
+		require.Equal(t, "no topic provided for message 0", err.Error())
```
fracasula, line 848: FYI, there is a `require.ErrorContains` helper too which takes an error instead of a string.

services/streammanager/kafka/kafkamanager_test.go
```diff
@@ -583,6 +597,30 @@ func TestSendBatchedMessage(t *testing.T) {
...
+		require.Equal(t, "Kafka: Message delivered in batch", err)
+		require.Len(t, p.calls, 1)
+		require.Len(t, p.calls[0], 1)
+		require.Equal(t, []byte("123"), p.calls[0][0].Key)
+		require.Equal(t, []byte(`"ciao"`), p.calls[0][0].Value)
+		require.Equal(t, "some-topic", p.calls[0][0].Topic)
+		require.InDelta(t, time.Now().Unix(), p.calls[0][0].Timestamp.Unix(), 1)
+	})
+
+	t.Run("default-topic-test", func(t *testing.T) {
```
fracasula, line 619: Same as the other tests:

```suggestion
	t.Run("default topic test", func(t *testing.T) {
```

The missing characters are replaced by Go. Spaces are replaced with `_` but you don't have to remember that, just let it do its thing.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -450,14 +450,19 @@ func serializeAvroMessage(value []byte, codec goavro.Codec) ([]byte, error) {
 	return binary, nil
 }
 
-func prepareBatchOfMessages(topic string, batch []map[string]interface{}, timestamp time.Time, p producerManager) (
+func prepareBatchOfMessages(batch []map[string]interface{}, timestamp time.Time, p producerManager, defaultTopic string) (
```
fracasula, line 453: Instead of passing the `defaultTopic` around, an alternative approach is to use the existing pattern of including such information in the `producerManager`:

```go
type producerManager interface {
	io.Closer
	publisher
	getDefaultTopic() string
	getTimeout() time.Duration
	getCodecs() map[string]*goavro.Codec
}
```

That should be preferred given that it is already in place. I've attached a git diff for you to consider this approach (e.g. `git apply yourcoworkers.diff`).

[kafka.diff.zip](https://github.com/rudderlabs/rudder-server/files/9950299/kafka.diff.zip)

services/streammanager/kafka/kafkamanager.go
fracasula: Looks good to me, it looks like you used different topics as well in your test. @atzoum any thoughts?

services/streammanager/kafka/kafkamanager.go
utsabc: Can we take this up in the next iteration of enhancement I can create a backlog item in our board for this. The reason I am asking is I need to update and add some test cases for this approach as we are dangerously close to release this might take more time :/

services/streammanager/kafka/kafkamanager_test.go
utsabc: This is addressed

services/streammanager/kafka/kafkamanager_test.go
utsabc: This is addressed

services/streammanager/kafka/kafkamanager_test.go
utsabc: This is addressed

atzoum (approved):

services/streammanager/kafka/kafkamanager.go
fracasula: Sure no problem :+1:

fracasula (approved):

PR #2570, opened 2022-10-14
chore(warehouse): correct exported for handle parameters

achettyiitr:
# Description

Use correct exported for destination handler parameters.

## Notion Ticket

https://www.notion.so/rudderstacks/Correct-exported-parameter-for-handle-98888bdd3d024ee299c74a465f205fd7

## Security


achettyiitr:
Not required now because of 2 reasons:
1. Since during testing also some things need to be exported so that they can be injected for testing.
2. If some other packages need to use it and set it then also they require it.

PR #2582, opened 2022-10-18
fix: router destinationsMap access

Sidddddarth:
# Description

Accessing `destination` without checking for its existence first. Fixes that.

## Notion Ticket

[fix: router destination map access](https://www.notion.so/rudderstacks/fix-router-destination-map-access-2834a4985d0c4886aa080abe77174053)

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -446,7 +446,6 @@ func (worker *workerT) workerProcess() {
 
 			worker.rt.configSubscriberLock.RLock()
 			batchDestination, ok := worker.rt.destinationsMap[parameters.DestinationID]
-			destination := batchDestination.Destination
 			worker.rt.configSubscriberLock.RUnlock()
 			if !ok {
 				status := jobsdb.JobStatusT{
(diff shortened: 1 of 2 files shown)


atzoum:
@Sidddddarth we may need to fail events for which no configuration can be found instead of aborting them. A test scenario for this would be desirable too.

cisse21 (approved):

chandumlg (approved):

router/router.go
```diff
@@ -446,23 +446,23 @@ func (worker *workerT) workerProcess() {
...
 			worker.rt.configSubscriberLock.RLock()
 			batchDestination, ok := worker.rt.destinationsMap[parameters.DestinationID]
-			destination := batchDestination.Destination
 			worker.rt.configSubscriberLock.RUnlock()
 			if !ok {
 				status := jobsdb.JobStatusT{
 					JobID:         job.JobID,
 					AttemptNum:    job.LastJobStatus.AttemptNum,
-					JobState:      jobsdb.Aborted.State,
+					JobState:      jobsdb.Failed.State,
```
atzoum, line 454: @Sidddddarth we need to update the status of the job in the barrier, since failed jobs are notified at this point

PR #2598, opened 2022-10-19
chore(warehouse): use fastUUID with google UUID generation

achettyiitr:
# Description

**Use google's uuid, as gofrs can cause [performance issues](https://github.com/rudderlabs/rudder-server/pull/1606)** 

**Also, we have introduced one more utility called RandHex which generated a random string of 32 characters from UUID.** 
(The benefit here is that at some places where after getting the UUID string of 36 characters we used to replace `-`. Since we are doing extra computation of preparing the UUID string and replacing the `-`.  The same functionality is being provided by RandHex.)

## Notion Ticket

https://www.notion.so/rudderstacks/Fasf-UUID-4683365544654681aa04f69f790e05ce

## Security


warehouse/validations/helper.go
```diff
@@ -59,7 +58,7 @@ func parseOptions(req json.RawMessage, v interface{}) error {
 }
 
 func randomString() string {
-	return strings.ReplaceAll(uuid.Must(uuid.NewV4()).String(), "-", "")
+	return strings.ReplaceAll(misc.FastUUID().String(), "-", "")
```
lvrach, line 61: There is probably a faster way to do this by converting the UUID bytes directly to runes. If this is used in a performance-critical path, it is probably worth it ( a benchmark will be advised)

lvrach: We can use our own, modified version of encodeHex(dst []byte, uuid UUID) https://github.com/google/uuid/blob/44b5fee7c49cf3bcdf723f106b36d56ef13ccc88/uuid.go#L200

A single call to [hex.Encode](https://pkg.go.dev/encoding/hex#Encode) should be enough.

lvrach:
```suggestion
	var buf [32]byte
	 hex.Encode(buf[:], misc.FastUUID())
	return string(buf[:])
```

achettyiitr: Make sense. Since we are ignoring `-`. It's better to use this.

achettyiitr: Done.

warehouse/utils/utils.go
```diff
@@ -1074,8 +1074,15 @@ func StagingTablePrefix(provider string) string {
...
 
 func StagingTableName(provider, tableName string, tableNameLimit int) string {
-	randomNess := strings.ReplaceAll(uuid.Must(uuid.NewV4()).String(), "-", "")
+	randomNess := RandStringWithoutQuotes()
 	prefix := StagingTablePrefix(provider)
 	stagingTableName := fmt.Sprintf(`%s%s_%s`, prefix, tableName, randomNess)
 	return misc.TruncateStr(stagingTableName, tableNameLimit)
 }
+
+func RandStringWithoutQuotes() string {
```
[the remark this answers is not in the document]
achettyiitr: Done.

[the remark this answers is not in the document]
achettyiitr: Done.

warehouse/bigquery/bigquery.go
```diff
@@ -276,7 +276,7 @@ func (bq *HandleT) DeleteBy(tableNames []string, params warehouseutils.DeleteByP
 }
 
 func partitionedTable(tableName, partitionDate string) string {
-	return fmt.Sprintf(`%s$%v`, tableName, strings.ReplaceAll(partitionDate, "-", ""))
+	return fmt.Sprintf(`%s$%v`, tableName, warehouseutils.RandStringWithoutQuotes())
```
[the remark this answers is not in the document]
achettyiitr: Thanks, @lvrach for pointing it out. It happened by mistake.

lvrach (approved):
Thanks for addressing my comments

warehouse/utils/utils.go
```diff
@@ -1075,8 +1075,16 @@ func StagingTablePrefix(provider string) string {
...
 func StagingTableName(provider, tableName string, tableNameLimit int) string {
-	randomNess := strings.ReplaceAll(uuid.Must(uuid.NewV4()).String(), "-", "")
+	randomNess := RandHex()
 	prefix := StagingTablePrefix(provider)
 	stagingTableName := fmt.Sprintf(`%s%s_%s`, prefix, tableName, randomNess)
 	return misc.TruncateStr(stagingTableName, tableNameLimit)
 }
+
+// RandHex returns a random hex string of length 32
+func RandHex() string {
```
abhimanyubabbar, line 1085: Is there any improvement in this function in terms of speed or uniqueness over the other function ? Just curious to understand the motivation behind it.

abhimanyubabbar (approved):
Overall the changes looks good. 👍🏼 LGTM
I have just one comment on the motivation behind creation of RandomHex() function.

warehouse/utils/utils.go
achettyiitr: Yes. Actually, UUID.string() is processing for 36 bytes. We are ignoring the `-` present by replacing it using `strings.ReplaceAll(uuid.Must(uuid.NewV4()).String(), "-", "")`. We can just probably call `hex.Encode` which can just populate the 32 bytes.

```
const Size = 16
type UUID [Size]byte
```

```
func (u UUID) String() string {
	buf := make([]byte, 36)

	hex.Encode(buf[0:8], u[0:4])
	buf[8] = '-'
	hex.Encode(buf[9:13], u[4:6])
	buf[13] = '-'
	hex.Encode(buf[14:18], u[6:8])
	buf[18] = '-'
	hex.Encode(buf[19:23], u[8:10])
	buf[23] = '-'
	hex.Encode(buf[24:], u[10:])

	return string(buf)
}
```

PR #2604, opened 2022-10-20
feat: user suppressions adaptations for namespaces

Sidddddarth:
# Description

server can poll new namespaces endpoint to fetch the suppression for workspaces that constitute a namespace.

## Notion Ticket

[suppress-user adaptations for multitenant](https://www.notion.so/rudderstacks/suppress-user-adaptations-for-multitenant-08545073c1e2443ab4787f1b563c14c9)

## Security


enterprise/suppress-user/setup.go
```diff
@@ -40,11 +42,28 @@ func (m *Factory) Setup(backendConfig backendconfig.BackendConfig) (types.UserSu
 	loadConfig()
 	ctx := context.TODO()
 	backendConfig.WaitForConfig(ctx)
-	workspaceId := backendConfig.Identity().ID()
+	ID := backendConfig.Identity().ID()
```
atzoum, line 45: We may use the `identity.Identifier` for everything and set is as a field in `SuppressRegulationHandler`

1. Introduce a new method in `identity.Identifier`: `Type() deployment.Type`
2. `getSourceRegulationsFromRegulationService` can construct the url from `identity.Identifier`
3. `getSourceRegulationsFromRegulationService` can delegate basic authentication to `identity.Identifier`

atzoum (approved):

services/controlplane/identity/identity.go
```diff
@@ -1,11 +1,14 @@
...
 
+import "github.com/rudderlabs/rudder-server/utils/types/deployment"
+
 // Identifier abstracts how data-plane can be identified to the control-plane.
 //
 //	Including both a unique identifier and the authentication method.
 type Identifier interface {
 	ID() string
 	BasicAuth() (string, string)
+	Type() deployment.Type
```
lvrach, line 11: Why not use reflection to determine the type? 

It is quite an idiomatic way, the only downside is performance.

enterprise/suppress-user/suppressUser.go
```diff
@@ -47,29 +49,41 @@ type apiResponse struct {
...
-	SourceIDs []string `json:"sourceIds"`
+	Canceled    bool     `json:"canceled"`
+	WorkspaceID string   `json:"workspaceId"`
+	UserID      string   `json:"userId"`
+	SourceIDs   []string `json:"sourceIds"`
 }
 
 func (suppressUser *SuppressRegulationHandler) setup(ctx context.Context) {
+	suppressUser.RegulationBackendURL = configBackendURL
+	switch suppressUser.ID.Type() {
```
lvrach, line 60: I would suggest [this](https://github.com/rudderlabs/rudder-server/blob/master/services/controlplane/features/client.go#L93) approach mainly for consistency. I also find it more idiomatic, but that is up for discussion.

lvrach (approved):
LGTM, a non-blocking comment regarding type.

chandumlg (approved):

enterprise/suppress-user/suppressUser.go
atzoum: Usually, it is considered to be a bad practice requiring from the consumer of an exported type to have to know both about the generic interface type and the concrete implementation types, since it beats the original purpose of the interface, that is abstracting from the underlying implementation(s).

PR #2606, opened 2022-10-21
feat: distributed throttling

fracasula:
# Description

I'm introducing a new throttler that supports different algorithms to do rate limiting. The idea was to have it support at least one in-memory and one distributed (with Redis) but having worked earlier on it with benchmarks I had other algorithms GCRA based so I introduced the support for them as well.

The proposal is to have it configurable and switch from one algo to another via feature flags. Here's a list of the supported algorithms and their use cases:
1. in-memory go-rate
   * it's a wrapper around [golang.org/x/time/rate](https://pkg.go.dev/golang.org/x/time/rate)
   * it is slower than the in-memory GCRA counterpart
   * uses more memory than the in-memory GCRA counterpart 
   * it's more precise than GCRA in high concurrency environments
   * it supports returning tokens
2. in-memory GCRA
   * usually faster and with very low memory footprint
   * it tends to loose precision (sometimes a lot) in high concurrency environments
   * it does not support returning tokens
3. Redis GCRA
   * like it's in-memory counterpart it tends to loose precision
   * very low memory footprint
   * it does not support returning tokens
4. Redis SortedSet (defaul)
   * custom algorithm wrote by myself (although inspired by bits and pieces around the www, many had bugs and did not support returning tokens)
   * it consumes more memory than GCRA because it keeps track of all tokens in the bucket
   * most precise of all 4 algorithms
   * faster than the Redis GCRA counterpart
   * it supports returning tokens

The occasional loss of precision made the tests very brittle. This is the reason why in the throttler tests you will see two tuning parameters: `concurrency` and `additionalErrorMargin`.

By default I'm applying a ~5% error margin but some algos like GCRA are particularly susceptible to concurrency so I had to introduce two more parameters to make the test more tolerant in those cases.

## Benchmarks without persistence

All the benchmarks below were done with:
* default Redis server settings
* no persistence
* rate 10k
* window 60s
* duration 3mins
* concurrency 5k

### Benchmark with AWS Elasticache (Redis v6.2.6, 1 master + 1 replica)

```
2022/10/26 08:54:19 Trying to connect to Redis at: throttling-2-001.012ut9.0001.use1.cache.amazonaws.com:6379
2022/10/26 08:54:19 Initiating benchmark on key: xednnfuRFS
2022/10/26 08:57:19 Timer triggered
2022/10/26 08:57:19 Waiting on all goroutines to finish...
2022/10/26 08:57:20 Number of requests: 4505973
2022/10/26 08:57:20 Elapsed time: 3m0.198631755s
2022/10/26 08:57:20 RPS: 25005.589421602097
```

### Benchmark with Redis in same K8s cluster (Redis v7.0.4, 1 master + 3 replicas)

```
2022/10/26 08:50:26 Trying to connect to Redis at: redis-throttling-master.fcasula.svc.cluster.local:6379
2022/10/26 08:50:26 Initiating benchmark on key: UQlVjYHZYG
2022/10/26 08:53:26 Timer triggered
2022/10/26 08:53:26 Waiting on all goroutines to finish...
2022/10/26 08:53:26 Number of requests: 7705874
2022/10/26 08:53:26 Elapsed time: 3m0.120130569s
2022/10/26 08:53:26 RPS: 42781.858838637985
```

### Benchmark with AWS Elasticache 10Gbit network (Redis v6.2.6, 1 master + 2 replica)

```
2022/10/26 11:28:54 Trying to connect to Redis at: throttling-persistent-cluster-10gbit-001.012ut9.0001.use1.cache.amazonaws.com:6379
2022/10/26 11:28:54 Initiating benchmark on key: ZMVRMxncLH
2022/10/26 11:31:54 Timer triggered
2022/10/26 11:31:54 Waiting on all goroutines to finish...
2022/10/26 11:31:54 Number of requests: 9704567
2022/10/26 11:31:54 Elapsed time: 3m0.096971429s
2022/10/26 11:31:54 RPS: 53885.23151165732
```

### Benchmark with local Docker container (Redis 7.0.5-debian-11-r12, 1 master)

```
2022/10/26 10:59:12 Trying to connect to Redis at: localhost:6379
2022/10/26 10:59:12 Initiating benchmark on key: pCIlrIGPyf
2022/10/26 11:02:12 Timer triggered
2022/10/26 11:02:12 Waiting on all goroutines to finish...
2022/10/26 11:02:12 Number of requests: 24007970
2022/10/26 11:02:12 Elapsed time: 3m0.048058627s
2022/10/26 11:02:12 RPS: 133342.00981159465

```

## Benchmarks with persistence

All the benchmarks below were done with:
* default Redis server settings
* `save 1 1` and `appendonly=yes`
* rate 10k
* window 60s
* duration 3mins
* concurrency 5k

### Benchmark with local Docker container (Redis 7.0.5-debian-11-r12, 1 master)

```
2022/10/26 13:44:01 Trying to connect to Redis at: localhost:6379
2022/10/26 13:44:01 Initiating benchmark on key: rvZVVLGPBh
2022/10/26 13:47:01 Timer triggered
2022/10/26 13:47:01 Waiting on all goroutines to finish...
2022/10/26 13:47:01 Number of requests: 23924718
2022/10/26 13:47:01 Elapsed time: 3m0.048432922s
2022/10/26 13:47:01 RPS: 132879.34591668783
```

### Benchmark with Redis in same K8s cluster (Redis v7.0.4, 1 master and no replicas)

```
2022/10/26 12:00:36 Trying to connect to Redis at: redis-throttling-master.fcasula.svc.cluster.local:6379
2022/10/26 12:00:36 Initiating benchmark on key: mgfFqoSutX
2022/10/26 12:03:36 Timer triggered
2022/10/26 12:03:36 Waiting on all goroutines to finish...
2022/10/26 12:03:36 Number of requests: 8478004
2022/10/26 12:03:36 Elapsed time: 3m0.142655478s
2022/10/26 12:03:36 RPS: 47062.723581508326
```
### Benchmark with AWS Elasticache 

Persistence not supported on AWS Elasticache.

## Notion Ticket

* < [Notion Link](https://www.notion.so/rudderstacks/Distributed-throttling-implementation-1c0a955a40a54ab9b81d4bec2ad4b711) >
* < [Notion Docs](https://www.notion.so/rudderstacks/Distributed-Throttling-93774160b8a34836b83a986ac9700b25) >

## Security


router/router.go
```diff
@@ -411,14 +415,46 @@ func (worker *workerT) workerProcess() {
...
+			var (
+				err            error
+				limited        bool
+				throttlingCost int64
+				tokensReturner throttling.TokenReturner
+				throttler      = worker.rt.getThrottler(parameters.DestinationID)
+			)
+			if len(worker.routerJobs) > 0 {
+				throttlingCost = worker.getThrottlingCost()
+			}
```
atzoum, line 431: It seems like we are calculating the throttling cost of all existing router jobs for every job, before adding this job to the router jobs. Why?

router/router.go
```diff
@@ -520,6 +568,33 @@ func (worker *workerT) workerProcess() {
 	}
 }
 
+func (worker *workerT) getThrottlingCost() (cost int64) {
```
atzoum, line 571: Instead of calculating this cost everytime for each message, can't we calculate it once, when creating the throttler?
Each router of every destination type, already keeps a map with all its destination ids and their configuration.

router/router.go
```diff
@@ -438,14 +474,16 @@ func (worker *workerT) workerProcess() {
 			batchDestination, ok := worker.rt.destinationsMap[parameters.DestinationID]
 			worker.rt.configSubscriberLock.RUnlock()
 			if !ok {
+				// we won't be using the throttling tokens anymore
+				go worker.rt.returnTokens(context.TODO(), tokensReturner)
```
atzoum, line 478: If this is the only place where we need to return tokens, then by simply moving this check before performing the throttling, we effectively eliminate the need for returning tokens altogether, correct?

internal/throttling/memory_gcra.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/throttled/throttled/v2/store/memstore"
+
+	"github.com/rudderlabs/rudder-server/internal/cachettl"
+)
+
+const defaultMaxCASAttemptsLimit = 100
+
+type gcra struct {
+	mu    sync.Mutex
+	store *cachettl.Cache
```
atzoum, line 18: Not sure I understand why we need a cache with ttl for storing independent rate limiters per key, when the gcra limiter can work for multiple keys already

router/router.go
```diff
@@ -411,14 +415,46 @@ func (worker *workerT) workerProcess() {
...
+				limited        bool
+				throttlingCost int64
+				tokensReturner throttling.TokenReturner
+				throttler      = worker.rt.getThrottler(parameters.DestinationID)
+			)
+			if len(worker.routerJobs) > 0 {
+				throttlingCost = worker.getThrottlingCost()
+			}
+			if throttlingCost > 0 && throttler != nil {
+				limited, tokensReturner, err = throttler.CheckLimitReached(userID, throttlingCost)
```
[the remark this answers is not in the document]
fracasula: Yes, bold move indeed. That's the thing that I got left basically, increase coverage in `router_test.go` and manual tests on dev to understand what can go wrong here. I'm pretty confident about the rest, not so much about the router changes atm.

internal/throttling/memory_gcra.go
fracasula: The underlying GCRA limiter can work with multiple keys but you can't set a different quota per key so I had to create a new limiter each time. Meaning that with the `throttled` GCRA limiter alone we'll have to use the same quota for all destinations.

An optimization I can do is to create one limiter per quota instead of one per key, so if more than one key uses the same quota I don't create an instance for each but that would increase the expiration complexity for little advantage.

In any case, to give a bit more context, I created the `cachettl.Cache` because I did not want something to live in memory forever even if it was used only once (e.g. you get only one message a day/week/whatever for a destination, still keep the limiter in the map forever). I searched a bit for existing implementations because I didn't really want to spend time writing my own cache but I could only find caches where the TTL is set globally and not per-key. It was simple enough to convince me to write it so here we are :-)

atzoum: Each router is for a specific destination type and each destination type has its own quota configuration.
Additionally, this quota configuration doesn't appear reloadable for the time being, i.e. it cannot change.
Based on the above, a single GCRA limiter would suffice. The only place I would justify using a TTL cache and multiple limiters would be in the gorate implementation, where we need to have a different limiter per key.

router/router.go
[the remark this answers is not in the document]
atzoum: It is important to answer (and agree) why we need to apply throttling at the router, e.g. instead of relying to the destinations responding with a `HTTP 429`.

IMO, two of the most important reasons are:
1. Try to keep our workers busy with jobs that are most likely to succeed.
2. Avoid recording too many error statuses in jobsdb, whenever a job is rate limited.

The second reason might be more important than the first one, since the side-effects of having too many job statuses recorded in our jobsdb tables are multiple (and cascading):
- queries keep getting slower and slower
- the migrator is constantly busy migrating large tables in an attempt to lighten up the bloated status tables
- the backup loop is constantly busy trying to backup the never ending stream of migrated tables, it usually cannot keep up, causing a table pileup and eventually free disk space gets exhausted.

By moving the throttling gate out of the `generatorLoop` and inside each worker, I believe we are effectively cancelling both of the above points:
1. Jobs that might get rate limited will fill up our workers' queue
2. Two statuses will be recorded in jobsdb for every job that gets rate limited (`executing` & `failed`) which can end up getting bad pretty quickly.

Even though we are now trying to improve jobsdb's tolerance to bloated job status tables by introducing compaction, nevertheless it wouldn't be wise to consciously fill our job status tables with 429 errors.

internal/throttling/memory_gorate.go
```diff
@@ -0,0 +1,50 @@
...
+func (r *goRate) limit(key string, cost, rate, periodInSecs int64) *goRateReservation {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.store == nil {
+		r.store = cachettl.New()
+	}
+
+	window := time.Duration(periodInSecs) * time.Second
+	l, ok := r.store.Get(key).(*gorate.Limiter)
```
atzoum, line 26: when a limiter is used, it should update (refresh) its ttl in the cache so that it only expires due to inactivity

internal/throttling/memory_gcra.go
fracasula: This isn't just about routers. The library lives in `internal/throttling` and it is meant to be used across the board for all throttling needs. In a following PR I'm planning to replace the limiter that we're using on the gateway as well.

That said, what I'm trying to do here is to provide a simple enough interface to throttle with more than one algorithm supported. You don't have to care about which algorithm you're using, you just setup the throttler and call `.Limit()`. The throttler provides different quotas per key and so that applies to all algos, memory GCRA included. I don't think extracting the memory GCRA one makes sense just because of the TTL cache (which in my opinion is a simple enough and lightweight data structure by the way).

After all the benchmarking and the tests I don't see a risk with going live with this to be honest. After some time we can still gather some stats and decide whether we want to review this, but so far I don't see a reason to remove the TTL cache. Of course if you're concerned that this would create issues please elaborate.

testhelper/destination/redis.go
```diff
@@ -1,21 +1,62 @@
...
+}
+
 type RedisResource struct {
-	RedisAddress string
+	Addr string
+}
+
+type RedisOption interface {
+	apply(*redisConfig)
+}
```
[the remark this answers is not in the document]
fracasula: Yes, you're right, let's simplify :+1:

config/backend-config/types.go
```diff
@@ -29,9 +29,12 @@ const (
...
 type DestinationDefinitionT struct {
-	ID            string
-	Name          string
-	DisplayName   string
+	ID          string
+	Name        string
+	DisplayName string
+	// Config key "throttlingCost" is expected to have the eventType as the first key and the call type
+	// as the second key (e.g. track, identify, etc...) or default to apply the cost to all call types:
+	// dDT["config"]["throttlingCost"] = `{"eventType":{"default":1,"track":2,"identify":3}}`
```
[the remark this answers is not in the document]
fracasula: Good point, moved :+1:

internal/cachettl/cachettl.go
```diff
@@ -0,0 +1,121 @@
+package cachettl
```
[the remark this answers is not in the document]
fracasula: I'm OK either way to be honest. I created a separate package because it is generic enough after all, and it's just a data structure so I saw no harm in having it in its own package without subbing it.

internal/cachettl/cachettl.go
```diff
@@ -0,0 +1,121 @@
...
+// Cache is a double linked list sorted by expiration time (ascending order)
+// the root (head) node is the node with the lowest expiration time
+// the tail node (end) is the node with the highest expiration time
+// Cleanups are done on Get() calls so if Get() is never invoked then Nodes stay in-memory.
+type Cache struct {
+	root *node
+	mu   sync.Mutex
+	m    map[string]*node
+	now  func() time.Time
+}
```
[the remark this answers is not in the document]
fracasula: Good idea :+1:

router/router.go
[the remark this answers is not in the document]
fracasula: Makes sense, I'll revisit this and see what I can do.

fracasula:
Please hold on on the router review, I will be pushing router changes soon with dedicated throttling test as well.

router/router.go
```diff
@@ -2049,14 +1998,30 @@ func (rt *HandleT) backendConfigSubscriber() {
...
+							// Config key "throttlingCost" is expected to have the eventType as the first key and the call type
+							// as the second key (e.g. track, identify, etc...) or default to apply the cost to all call types:
+							// dDT["config"]["throttlingCost"] = `{"eventType":{"default":1,"track":2,"identify":3}}`
+							if value, ok := destination.DestinationDefinition.Config["throttlingCost"].(map[string]interface{}); ok {
+								if _, ok := rt.throttlingCosts[destination.ID]; !ok {
+									if rt.throttlingCosts == nil {
+										rt.throttlingCosts = make(map[string]types.EventTypeThrottlingCost)
+									}
+
+									rt.throttlingCosts[destination.ID] = types.NewEventTypeThrottlingCost(value)
```
atzoum, line 2022: All destinations in router share the same `DestinationDefinition`, thus we don't need to keep separate throttling costs per destination id. A single throttling cost would be sufficient

router/router.go
```diff
@@ -1346,17 +1282,28 @@ func (rt *HandleT) getWorkerPartition(userID string) int {
...
-	// No need of locks here, because this is used only by a single goroutine (generatorLoop)
-	limitReached := rt.throttler.CheckLimitReached(destID, userID, throttledAtTime)
-	if !limitReached {
-		rt.throttler.Inc(destID, userID, throttledAtTime)
+	if !limited {
+		return tokensReturner, false
 	}
-	return limitReached
+
+	return nil, true
```
atzoum, line 1306: can't we simplify this, e.g.

```
return tokensReturner, !limited
```
Will `throttler.CheckLimitReached` return a non-nil tokensReturner if limit is reached? It shouldn't right?

router/router.go
```diff
@@ -1523,8 +1470,13 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
 			userID := resp.userID
 			worker := resp.worker
 			if status != jobsdb.Failed.State {
+				if worker == nil {
+					// worker is nil when we're trying to find a worker and there is no destination configuration
+					rt.logger.Debugf("[%v Router] :: No worker was found for job %d", rt.destName, resp.status.JobID)
+					continue
+				}
```
atzoum, line 1477: Why is this needed? We shouldn't call `commitStatusList` if no worker is found

fracasula: I forgot to remove it (old refactoring), thanks for catching it!

router/throttler/throttler.go
```diff
@@ -1,173 +1,257 @@
...
 
-// HandleT is a Handle for event limiter
-type HandleT struct {
-	destinationName string
-	destLimiter     *Limiter
-	userLimiter     *Limiter
+// Client is a Handle for event limiterSettings
+type Client struct {
+	algoType      string
+	redisClient   *redis.Client
```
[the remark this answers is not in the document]
fracasula: We don't need it but the alternative would be a map with some lock around it. Also the map would be a global variable to the `throttler` package. The `redisAddr` is unique enough to be used as the map key. Then if a client for that addr is already defined we use that instead of initializing a new one. Wdyt of this approach? Or if you have a better one in mind I'm all ears :+1:

router/router.go
```diff
@@ -933,78 +936,6 @@ type JobResponse struct {
...
-				if worker.encounteredRouterTransform {
-					diff /= 2
-				}
-				pkgLogger.Debugf(`Decrementing destination level throttle map by %d for dest:%s`, diff, destID)
-				worker.rt.throttler.Dec(destID, "", diff, worker.throttledAtTime, throttler.DESTINATION_LEVEL)
-			}
-		}
-	}
-}
-
```
atzoum, line 1007: what a relief 😄

atzoum (commented):
Overall it looks good! If possible, I would like some more time to perform a third (final) review round, this time focusing more on the `internal/throttling` package. 👍

router/router.go
```diff
@@ -2201,3 +2157,32 @@ func (rt *HandleT) updateProcessedEventsMetrics(statusList []*jobsdb.JobStatusT)
...
+}
+
+func (rt *HandleT) getThrottlingCost(job *jobsdb.JobT) (cost int64) {
+	cost = 1
+	eventType := gjson.GetBytes(job.Parameters, "event_type").String()
+
+	tc, ok := rt.throttlingCosts.Load().(types.EventTypeThrottlingCost)
+	if ok {
+		cost = tc.Cost(eventType)
+	}
```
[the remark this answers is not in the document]
fracasula: Good idea, I changed it to use an `atomic.Pointer` since it supports generics already :+1:

router/router.go
```diff
@@ -1288,18 +1219,23 @@ func (rt *HandleT) findWorker(job *jobsdb.JobT, throttledUserMap map[string]stru
 	err := json.Unmarshal(job.Parameters, &parameters)
 	if err != nil {
 		rt.logger.Errorf(`[%v Router] :: Unmarshalling parameters failed with the error %v . Returning nil worker`, err)
-		return nil
+		return
 	}
 
-	if rt.shouldThrottle(parameters.DestinationID, userID, throttledAtTime) {
+	tokensReturner, shouldThrottle := rt.shouldThrottle(job, parameters)
```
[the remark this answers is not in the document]
fracasula: Not sure I follow.

router/throttler/throttler.go
```diff
@@ -1,173 +1,224 @@
...
-	Inc(destID, userID string, currentTime time.Time)
-	Dec(destID, userID string, count int64, currentTime time.Time, atLevel string)
-	IsEnabled() bool
-	IsUserLevelEnabled() bool
-	IsDestLevelEnabled() bool
+var clientsPool redisClientsPool
+
+type limiter interface {
+	// Limit should return true if the request can be done or false if the limit is reached.
+	Limit(ctx context.Context, cost, rate, window int64, key string) (
```
raman2805, line 27: missing type ?

fracasula: What missing type?

router/router.go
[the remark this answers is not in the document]
atzoum: If throttling becomes the last check inside `findWorker`, then there wouldn't be a need for returning tokens. The only thing keeping us away from making it the last check is `eventorder.Barrier`, where we need to be able for jobs to `Leave` the barrier after `Enter`.

router/throttler/throttler.go
```diff
@@ -1,173 +1,224 @@
...
 }
 
-type Limiter struct {
+type limiterSettings struct {
 	enabled     bool
-	eventLimit  int
+	eventLimit  int64
 	timeWindow  time.Duration
-	ratelimiter *ratelimiter.RateLimiter
+	rateLimiter limiter
```
atzoum, line 36: `rateLimiter` could be reused across all destinationIDs of the same router. Each destination has its own key, however we seem to be using a separate limiter per key, whereas our limiters support multiple keys.

internal/throttling/throttling.go
```diff
@@ -0,0 +1,249 @@
...
+
+type TokenReturner interface {
+	Return(context.Context) error
+}
+
+type statsCollector interface {
+	NewTaggedStat(name, statType string, tags stats.Tags) stats.Measurement
+}
+
+type Limiter struct {
```
atzoum, line 50: Why does `Limiter` need to hold the union of fields for all different algorithms/setups? Wouldn't it be more straightforward if we had (4) different limiter implementations, one for each algorithm and each one containing only the fields it requires?

internal/throttling/throttling.go
```diff
@@ -0,0 +1,249 @@
...
+	default:
+		defer l.getTimer(key, "go-rate", rate, window)()
+		return l.goRateLimit(ctx, cost, rate, window, key)
+	}
+}
+
+func (l *Limiter) redisSortedSet(ctx context.Context, cost, rate, window int64, key string) (
+	bool, TokenReturner, error,
+) {
+	redisSpeaker := l.getRedisSpeaker(key)
```
atzoum, line 135: Our current usage patterns don't justify supporting a different redis client per key. For the time being I believe we could live without a `redisKeyToClientMap` and revisit once it is actually needed.

internal/throttling/throttling.go
```diff
@@ -0,0 +1,249 @@
...
+	// redisToClientMap does not require a Mutex because it is only read from and written once in the constructor
+	redisKeyToClientMap map[string]redisSpeaker
+
+	// for in-memory configurations
+	gcra   *gcra
+	goRate *goRate
+
+	// other flags
+	useGCRA            bool
+	useGCRABurstAsRate bool
```
atzoum, line 63: this should be `true` by default

internal/throttling/throttling.go
atzoum: As a side note, if this is to support global redis limiters in the future, then we still don't need support for multiple redis clients within the same `Limiter`, since each router will either need a local or global limiter, not both at the same time

internal/throttling/throttling.go
fracasula: I like the approach I'm proposing because I think it'll be harder to have an implementation go astray from the others and I don't want the consumer side to worry too much about implementation details. With this approach there isn't much complexity anyway, we're talking about a `switch` in `Limit` and a bigger constructor. The constructor returns a pointer to the same `*Limiter` struct with no need to abstract (no interface returned).

internal/throttling/throttling.go
fracasula: We still haven't decided whether we want to pay for a big AWS Elasticache to take care of *all* requests across *all* clusters (mt, pro, enterprise) or not. We have a call on Wednesday to discuss that. My take is that 1) it would cost too much 2) SPOF.
I'd go with a generic AWS Elasticache for the "global" destinations that share limits across all clusters and then have cheap standalone Redis masters deployed in the same K8s cluster.
Let's see what comes out of the call and we'll decide whether to keep this map afterwards.

router/router.go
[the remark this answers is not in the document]
fracasula: I'll have a look.

router/factory.go
```diff
@@ -31,14 +35,87 @@ type Factory struct {
...
+	)
+
+	var (
+		err  error
+		l    *throttling.Limiter
+		opts = []throttling.Option{
+			throttling.WithStatsCollector(f.Stats),
+		}
+	)
+	switch throttlingAlgorithm {
```
atzoum, line 81: Wouldn't it be more straightforward if we slightly modified the `throttling` package's `Options` so that all necessary information was passed using a single option? e.g.

```
throttling.WithGoRate()
throttling.WithGCRA()
throttling.WithGCRABurst(burst int64)
throttling.WithRedisGCRA(redisClient *redis.Client)
throttling.WithRedisGCRABurst(redisClient *redis.Client, burst int64)
throttling.WithRedisSortedSet(redisClient)
```

router/factory.go
```diff
@@ -1,26 +1,30 @@
...
-		name:          destination.DestinationDefinition.Name,
-		destinationID: identifier,
-		config:        destination.DestinationDefinition.Config,
-		responseRules: destination.DestinationDefinition.ResponseRules,
-	}
+type limiter interface {
+	CheckLimitReached(key string, cost int64) (limited bool, retErr error)
 }
 
 type Factory struct {
```
atzoum, line 30: By providing a producer function for `Factory` and always creating factories using this function, we could perform initialisation once, without using a `sync.Once` and having to call it first thing at the beginning of each struct's method.

router/factory.go
```diff
@@ -31,14 +35,87 @@ type Factory struct {
...
+		Reporting:        f.Reporting,
+		MultitenantI:     f.Multitenant,
+		throttlerFactory: f.throttlerFactory,
 	}
 	destConfig := getRouterConfig(destination, identifier)
 	r.Setup(f.BackendConfig, f.RouterDB, f.ProcErrorDB, destConfig, f.TransientSources, f.RsourcesService)
 	return r
 }
+
+func (f *Factory) initThrottlerFactory() {
```
atzoum, line 59: Shouldn't we move all this initialization logic inside the `throttler` package?

router/throttler/throttler.go
```diff
@@ -1,173 +1,89 @@
...
-func (throttler *HandleT) Dec(destID, userID string, count int64, currentTime time.Time, atLevel string) {
-	if throttler.destLimiter.enabled && destID != "" && (atLevel == ALL_LEVELS || atLevel == DESTINATION_LEVEL) {
-		destKey := throttler.getDestKey(destID)
-		_ = throttler.destLimiter.ratelimiter.Dec(destKey, count, currentTime)
+	if config.IsSet(fmt.Sprintf(`Router.throttler.%s.%s.timeWindow`, destName, destID)) ||
+		config.IsSet(fmt.Sprintf(`Router.throttler.%s.%s.timeWindowInS`, destName, destID)) {
+		config.RegisterDurationConfigVariable(
+			0, &c.window, false, time.Second,
+			fmt.Sprintf(`Router.throttler.%s.%s.timeWindow`, destName, destID),
+			fmt.Sprintf(`Router.throttler.%s.%s.timeWindowInS`, destName, destID),
```
atzoum, line 71: All calls to `config.RegisterXXX` keep pointers to resources which we might desire to release at some point but it won't be possible, i.e. are potential sources of memory leaks.
Let's avoid them wherever possible
```suggestion
		c.window = config.GetDuration(fmt.Sprintf(`Router.throttler.%s.%s.timeWindow`, destName, destID), 0, time.Second)
```

cmd/benchmark/throttling/deployment.yaml
```diff
@@ -0,0 +1,26 @@
+# redis-throttling-master.fcasula.svc.cluster.local:6379
+# throttling-2-001.012ut9.0001.use1.cache.amazonaws.com:6379
```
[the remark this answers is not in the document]
fracasula: Nope, they are just an example, don't worry. You won't be able to do much without the key anyway.

internal/throttling/lua/gcra.lua
```diff
@@ -0,0 +1,65 @@
...
+local burst_offset = emission_interval * burst
+
+-- Redis returns time as an array containing two integers: seconds of the epoch
+-- time (10 digits) and microseconds (~5-6 digits). for convenience we need to
+-- convert them to a floating point number. The resulting number is 16 digits,
+-- bordering on the limits of a 64-bit double-precision floating point number.
+-- Adjust the epoch to be relative to Jan 1, 2017 00:00:00 GMT to avoid floating
+-- point problems. This approach is good until "now" is 2,483,228,799 (Wed, 09
+-- Sep 2048 01:46:39 GMT), when the adjusted value is 16 digits.
+local jan_1_2017 = 1483228800 * 1000 * 1000 -- in microseconds precision
```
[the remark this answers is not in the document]
fracasula: No real reason, this is just a copy over from [here](https://github.com/go-redis/redis_rate/blob/v9/lua.go#L30). I carried it as is here to carry over as little as possible. We don't really need the whole library.

internal/throttling/lua/gcra.lua
```diff
@@ -0,0 +1,65 @@
...
+else
+    tat = tonumber(tat)
+end
+tat = math.max(tat, current_time)
+
+local new_tat = tat + increment
+local allow_at = new_tat - burst_offset
+local diff = current_time - allow_at
+local remaining = diff / emission_interval
+if remaining < 0 then
```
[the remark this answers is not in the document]
fracasula: I'd rather not change the Redis GCRA Lua implementation. This one is battle tested: https://github.com/go-redis/redis_rate/blob/v9/lua.go#L30

internal/throttling/memory_gcra.go
```diff
@@ -0,0 +1,63 @@
...
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	if g.store == nil {
+		g.store = cachettl.New[string, *throttled.GCRARateLimiter]()
+	}
+
+	rl := g.store.Get(key)
+	if rl == nil {
+		store, err := memstore.New(0)
```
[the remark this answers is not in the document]
fracasula: You're gonna need an astonishing number of destinations to be able to blow this up. It's true that the throttler can be used in other places too but I don't think setting a limit here makes much sense anyway. If `0` it just means that internally it uses a `map[string]*int64` instead of a LRU data structure with a limit. Like in other places where maps and other data structures that can grow are used, I think it's sensible to pay attention on the caller side instead of setting a hard limit in the data structure itself (e.g. instead of setting a limit on how many keys a map should have it's more sensible to avoid things like loading all data from a database into a map).

router/types/types.go
```diff
@@ -72,3 +72,22 @@ func (tm *TransformMessageT) JobIDs() map[int64]struct{} {
 	}
 	return jobIDs
 }
+
+func NewEventTypeThrottlingCost(m map[string]interface{}) (v EventTypeThrottlingCost) {
```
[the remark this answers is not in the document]
fracasula: Isn't this just a type coming from a portion of a backend config response? I'm OK either way but I felt like `types` would be a good place :man_shrugging:

router/router.go
```diff
@@ -1349,17 +1269,44 @@ func (rt *HandleT) getWorkerPartition(userID string) int {
...
+		rt.throttler[parameters.DestinationID] = throttler
+	}
+	rt.throttlerMu.Unlock()
+
+	throttlingCost := rt.getThrottlingCost(job)
+	// TODO throttling key could be the combination of workspaceID + destinationID to avoid conflicts across clusters
+	// The concatenation of workspaceID and destinationID would make sense for destinations that do not have global limits
+	// but we don't have yet a way to understand which ones are those (needs configuration setting).
+	limited, err := throttler.CheckLimitReached(parameters.DestinationID, throttlingCost)
+	if err != nil {
```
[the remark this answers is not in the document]
fracasula: Good point, let's log this one :+1:

internal/throttling/return.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (r *sortedSetRedisReturn) Return(ctx context.Context) error {
+	var (
+		length = len(r.members)
+		slice  = make([]interface{}, length)
+	)
+	for i, v := range r.members {
+		slice[i] = v
+	}
+	_, err := r.remover.ZRem(ctx, r.key, slice...).Result()
```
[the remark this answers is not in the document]
fracasula: No such thing as a dumb question :slightly_smiling_face: 
It's a Redis command to remove an element from a sorted set. See [here](https://redis.io/commands/zrem/).

> Removes the specified members from the sorted set stored at key. Non existing members are ignored.

router/router.go
```diff
@@ -84,7 +87,10 @@ type HandleT struct {
 	workers                                 []*workerT
 	telemetry                               *DiagnosticT
 	customDestinationManager                customDestinationManager.DestinationManager
-	throttler                               throttler.Throttler
+	throttler                               map[string]limiter // map key is the destinationID
+	throttlerMu                             sync.Mutex
+	throttlingCosts                         atomic.Pointer[types.EventTypeThrottlingCost]
+	throttlerFactory                        *rtThrottler.Factory
```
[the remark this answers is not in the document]
fracasula: I like the idea, will make the change :+1:

router/factory.go
fracasula: Hopefully it should be more clear now.

router/factory.go
```diff
@@ -31,14 +18,39 @@ type Factory struct {
 	ProcErrorDB      jobsdb.JobsDB
 	TransientSources transientsource.Service
 	RsourcesService  rsources.JobService
+	Logger           logger.Logger
+	Stats            stats.Stats
 }
 
 func (f *Factory) New(destination *backendconfig.DestinationT, identifier string) *HandleT {
+	throttlerFactory, err := throttler.New()
```
atzoum, line 26: Why would we need a different `throttler.Factory` for each router? Wouldn't it be simpler if we created only one `throttler.Factory` while creating the `router.Factory` in `embeddedAppHandler.go` & `embeddedAppHandler.go` and fail to `StartRudderCore` if an error occurs?

router/router.go
```diff
@@ -1349,17 +1267,36 @@ func (rt *HandleT) getWorkerPartition(userID string) int {
 	return misc.GetHash(userID) % rt.noOfWorkers
 }
 
-func (rt *HandleT) shouldThrottle(destID, userID string, throttledAtTime time.Time) (canBeThrottled bool) {
-	if !rt.throttler.IsEnabled() {
+func (rt *HandleT) shouldThrottle(job *jobsdb.JobT, parameters JobParametersT, throttledUserMap map[string]struct{}) (
+	limited bool,
+) {
+	if rt.throttlerFactory == nil {
```
atzoum, line 1273: Another reason why we wouldn't want to start without a `throttlerFactory` is the surge of log messages 😄

router/throttler/throttler.go
```diff
@@ -1,173 +1,156 @@
...
+	Allow(ctx context.Context, cost, rate, window int64, key string) (bool, func(context.Context) error, error)
 }
 
-type Settings struct {
-	limit               int
-	timeWindow          int64
-	userLevelLimit      int
-	userLevelTimeWindow int64
+type Factory struct {
+	Stats        stats.Stats
```
atzoum, line 29: we don't ever seem to set this field

router/throttler/throttler.go
```diff
@@ -1,173 +1,156 @@
...
 }
 
-// HandleT is a Handle for event limiter
-type HandleT struct {
-	destinationName string
-	destLimiter     *Limiter
-	userLimiter     *Limiter
+// New constructs a new Throttler Factory
+func New() (*Factory, error) {
+	f := Factory{}
```
atzoum, line 37: Since we are using the New function, we could initialize its `throttlers` map here and skip the nil check in `Factory#Get`

router/router.go
fracasula: Converting it to `debug` level. It shouldn't happen anyway since I moved the initialization to the app handlers.

atzoum: Then we don't even need this nil check here, correct?

atzoum (approved):

cisse21 (approved):

PR #2621, opened 2022-10-26
chore: make config load files optional

lvrach:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/config/load.go
+++ b/config/load.go
@@ -13,28 +13,31 @@ func (c *Config) load() {
 	c.hotReloadableConfig = make(map[string][]*configValue)
 	c.envs = make(map[string]string)
 
-	if err := godotenv.Load(); err != nil {
-		fmt.Println("INFO: No .env file found.")
+	if c.dotEnvFiles != nil {
+		fmt.Println(c.dotEnvFiles)
+		if err := godotenv.Load(c.dotEnvFiles...); err != nil {
+			fmt.Println("INFO: No .env file found.")
+		}
 	}
 
-	configPath := getEnv("CONFIG_PATH", "./config/config.yaml")
+	c.v = viper.NewWithOptions(viper.EnvKeyReplacer(&envReplacer{c: c}))
+	c.v.AutomaticEnv()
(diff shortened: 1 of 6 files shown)


config/load.go
```diff
@@ -13,28 +13,31 @@ func (c *Config) load() {
...
-	v.SetConfigFile(configPath)
-	err := v.ReadInConfig() // Find and read the config file
-	// Don't panic if config.yaml is not found or error with parsing. Use the default config values instead
-	if err != nil {
-		fmt.Printf("[Config] :: Failed to parse config file from path %q, using default values: %v\n", configPath, err)
+		c.v.SetConfigFile(configPath)
+		err := c.v.ReadInConfig() // Find and read the config file
+		// Don't panic if config.yaml is not found or error with parsing. Use the default config values instead
+		if err != nil {
+			fmt.Printf("[Config] :: Failed to parse config file from path %q, using default values: %v\n", c.configPath, err)
```
atzoum, line 34: can you please add a test scenario to cover this case?

atzoum (commented):
Isn't this going to affect all tests in the sense that [config.yaml](https://github.com/rudderlabs/rudder-server/blob/master/config/config.yaml) overrides will no longer be used?

Can we reverse the logic, i.e. have options for ignoring dotEnv (`WithoutDotEnv`) and config files (`WithoutConfigFile`)? Then, if you choose to use `rudder-server` as a library, you may override `config.Default ` by using these options.

atzoum:
> Isn't this going to affect all tests in the sense that [config.yaml](https://github.com/rudderlabs/rudder-server/blob/master/config/config.yaml) overrides will no longer be used?
> 
> Can we reverse the logic, i.e. have options for ignoring dotEnv (`WithoutDotEnv`) and config files (`WithoutConfigFile`)? Then, if you choose to use `rudder-server` as a library, you may override your own `config.Default ` by using these options.

what I proposed might not help. Would printing these error messages to stderr help with your usage scenarios?

PR #2622, opened 2022-10-27
feat: adding metadata in the proxy request

ujjwal-ab:
# Description

adding the metadata object in the proxy request to send the destination info inside that to use it in the transformer

## Security


router/types/types.go
```diff
@@ -55,6 +55,7 @@ type JobMetadataT struct {
 	Secret             json.RawMessage `json:"secret"`
 	JobT               *jobsdb.JobT    `json:"jobsT"`
 	WorkerAssignedTime time.Time       `json:"workerAssignedTime"`
+	DestInfo           json.RawMessage `json:"destInfo,omitempty"`
```
chandumlg, line 58: Where is this used?

router/router.go
```diff
@@ -669,10 +669,28 @@ func (worker *workerT) processDestinationJobs() {
 								if worker.rt.transformerProxy {
 									jobID := destinationJob.JobMetadataArray[0].JobID
 									pkgLogger.Debugf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Request started`, worker.rt.destName, jobID)
+
+									// setting metadata
+									firstJobMetadata := destinationJob.JobMetadataArray[0]
```
saikumarrs, line 674: Are these changes backward compatible with older transformer versions?

saikumarrs (approved):

router/types/types.go
sanpj2292: This information is being set by transformer and made use by transformer

router/router.go
sanpj2292: Yes, for the older transformer, this would act as extra information being sent & this information would not be used
For the new transformer, this information would be useful

sanpj2292: Even if the server is of older version, the metadata information will not be sent. Currently handling is done appropriately to handle this case.

chandumlg (approved):

chandumlg:
Changes look okay.

But lint & tests are failing. Please address.

chandumlg (commented):
Please check the tests

router/transformer/transformer.go
```diff
@@ -48,8 +49,23 @@ type handle struct {
...
+	UserID        string          `json:"userId"`
+	SourceID      string          `json:"sourceId"`
+	DestinationID string          `json:"destinationId"`
+	WorkspaceID   string          `json:"workspaceId"`
+	Secret        json.RawMessage `json:"secret"`
+	DestInfo      json.RawMessage `json:"destInfo,omitempty"`
+}
+
+type ProxyRequestPayload struct {
+	*integrations.PostParametersT
```
atzoum, line 64: Why are we embedding a pointer instead of a value? We don't want to treat nils, do we?

sanpj2292: Basically I followed a stackoverflow thread where in combining two structs to send it as aa json was shown in this way! 😓 

```go
type A struct {
a string
a1 string
}

type B struct {
b string
}

type C struct {
 A,
 B
}

/*
Can we achieve this ?
output json should look like this
{
 a: "v1",
 a1: "v11",
 b: "v2"
}
*/
```

router/router.go
```diff
@@ -663,10 +663,28 @@ func (worker *workerT) processDestinationJobs() {
...
+										DestinationID: firstJobMetadata.DestinationID,
+										WorkspaceID:   firstJobMetadata.WorkspaceID,
+										JobID:         firstJobMetadata.JobID,
+										AttemptNum:    firstJobMetadata.AttemptNum,
+										DestInfo:      firstJobMetadata.DestInfo,
+										Secret:        firstJobMetadata.Secret,
+									}
+									// forming the proxy request payload
+									proxyReqPayload := transformer.ProxyRequestPayload{
+										PostParametersT: &result[resultInd],
```
atzoum, line 680: another reason why we wouldn't need a pointer?

router/router.go
```diff
@@ -663,10 +663,28 @@ func (worker *workerT) processDestinationJobs() {
...
+										Metadata:        proxyReqMetadata,
+									}
+									// forming proxy request method parameters
 									proxyReqparams := &transformer.ProxyRequestParams{
 										DestName:     worker.rt.destName,
 										JobID:        jobID,
-										ResponseData: val,
+										ResponseData: proxyReqPayload,
 									}
```
atzoum, line 688: we could simplify using less variables/comments

```go
									firstJobMetadata := destinationJob.JobMetadataArray[0]
jobID := firstJobMetadata.JobID
pkgLogger.Debugf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Request started`, worker.rt.destName, jobID)

proxyReqparams := &transformer.ProxyRequestParams{
	DestName: worker.rt.destName,
	JobID:    jobID,
	ResponseData: transformer.ProxyRequestPayload{
		PostParametersT: &result[resultInd],
		Metadata: transformer.ProxyRequestMetadata{
			SourceID:      firstJobMetadata.SourceID,
			DestinationID: firstJobMetadata.DestinationID,
			WorkspaceID:   firstJobMetadata.WorkspaceID,
			JobID:         firstJobMetadata.JobID,
			AttemptNum:    firstJobMetadata.AttemptNum,
			DestInfo:      firstJobMetadata.DestInfo,
			Secret:        firstJobMetadata.Secret,
		},
	},
}

```

router/transformer/transformer.go
atzoum: embedding is fine, embedding a value instead of a pointer is preferred here
```go
	integrations.PostParametersT
	Metadata ProxyRequestMetadata `json:"metadata,omitempty"`
}
```

router/router.go

router/router.go

atzoum (approved):

PR #2625, opened 2022-10-27
feat: support for config to custom destinations

utsabc:
Adding support for passing config to transformer for custom destinations

# Description

Previously we were filtering the total config for requests to transformer for custom destinations. Main motivation for it was to avoid making heavy requests as some some configs would contain would contain certificate details etc.

This PR approaches the solution with more general intnet by filtering any destination config based on predefined filters set in integrations module

## Notion Ticket

Notion link: https://www.notion.so/rudderstacks/Kafka-Destination-with-Multiple-Topics-85cbce26706d49eca6853b972e5497b0

## Security


utsabc:
Related PR: https://github.com/rudderlabs/rudder-server/pull/2569

koladilip (commented):
It looks good, but is this something that can be done from destination definitions itself? Why keep config at different places?

utsabc:
> It looks good, but is this something that can be done from destination definitions itself? Why keep config at different places?

Ideally yes for a quick fix I went with this appraoch

processor/processor.go
```diff
@@ -1391,14 +1391,9 @@ func (proc *HandleT) processJobsForDest(subJobs subJob, parsedEventList [][]type
...
 					shallowEventCopy.Metadata.DestinationType = destination.DestinationDefinition.Name
 
-					//TODO: Test for multiple workspaces ex: hosted data plane
-					/* Stream destinations does not need config in transformer. As the Kafka destination config
-					holds the ca-certificate and it depends on user input, it may happen that they provide entire
-					certificate chain. So, that will make the payload huge while sending a batch of events to transformer,
-					it may result into payload larger than accepted by transformer. So, discarding destination config from being
-					sent to transformer for such destination. */
-					if misc.Contains(customDestinations, *destType) {
-						shallowEventCopy.Destination.Config = nil
```
krishna2020, line 1401: As the check on customDestinations is removed, Is there not a need to handle other destination configs apart from Kafka to filter ?

koladilip:
> > It looks good, but is this something that can be done from destination definitions itself? Why keep config at different places?
> 
> Ideally yes for a quick fix I went with this appraoch

is this going to be tech debt again? I prefer to have this to be defined at destDef level so that it is easy to debug and understand.

processor/processor.go
utsabc: We can add fields as when needed I actually checked other custom destinations we didn't have any properties which were large as SSL cert but if we find any we can add it

utsabc:
> > > It looks good, but is this something that can be done from destination definitions itself? Why keep config at different places?
> > 
> > 
> > Ideally yes for a quick fix I went with this appraoch
> 
> is this going to be tech debt again? I prefer to have this to be defined at destDef level so that it is easy to debug and understand.

Make sense I will try to do an impl

utsabc:
> > > > It looks good, but is this something that can be done from destination definitions itself? Why keep config at different places?
> > > 
> > > 
> > > Ideally yes for a quick fix I went with this appraoch
> > 
> > 
> > is this going to be tech debt again? I prefer to have this to be defined at destDef level so that it is easy to debug and understand.
> 
> Make sense I will try to do an impl

@koladilip I have approached the problem with the way you suggested, taking the filter keys from destination definition

processor/processor.go
```diff
@@ -1390,15 +1390,9 @@ func (proc *HandleT) processJobsForDest(subJobs subJob, parsedEventList [][]type
...
-
-					//TODO: Test for multiple workspaces ex: hosted data plane
-					/* Stream destinations does not need config in transformer. As the Kafka destination config
-					holds the ca-certificate and it depends on user input, it may happen that they provide entire
-					certificate chain. So, that will make the payload huge while sending a batch of events to transformer,
-					it may result into payload larger than accepted by transformer. So, discarding destination config from being
-					sent to transformer for such destination. */
-					if misc.Contains(customDestinations, *destType) {
-						shallowEventCopy.Destination.Config = nil
+					configsToFilter := destination.DestinationDefinition.ConfigFilters
```
koladilip, line 1393: What happens to range operation if this is nil?

utsabc: `ConfigFilters` won't be `nil` as its type Array when initialized it will be an empty array. But we can do a nil check before iterating

utsabc: Screenshot 2022-10-31 at 10 38 00 AM
 Lint is marking it as not needed

krishna2020 (approved):

processor/processor.go
koladilip: Yes looks like nil is handled by range so we don't need to worry about: https://play.golang.com/p/iymc4sfOdcw

koladilip (approved):

processor/processor.go
```diff
@@ -1390,17 +1390,12 @@ func (proc *HandleT) processJobsForDest(subJobs subJob, parsedEventList [][]type
...
-					//TODO: Test for multiple workspaces ex: hosted data plane
-					/* Stream destinations does not need config in transformer. As the Kafka destination config
-					holds the ca-certificate and it depends on user input, it may happen that they provide entire
-					certificate chain. So, that will make the payload huge while sending a batch of events to transformer,
-					it may result into payload larger than accepted by transformer. So, discarding destination config from being
-					sent to transformer for such destination. */
-					if misc.Contains(customDestinations, *destType) {
-						shallowEventCopy.Destination.Config = nil
+					configsToFilter := destination.DestinationDefinition.Config["configFilters"]
+					if configsToFilter != nil && reflect.TypeOf(configsToFilter).Kind() == reflect.Slice {
```
chandumlg, line 1394: Why reflect? Can we do something like this?

processor/processor.go
```diff
@@ -1390,17 +1390,12 @@ func (proc *HandleT) processJobsForDest(subJobs subJob, parsedEventList [][]type
...
+						for _, configKey := range configsToFilter.([]interface{}) {
+							shallowEventCopy.Destination.Config[configKey.(string)] = ""
```
chandumlg, line 1396: What if configKey is not string? Avoid panics.

chandumlg (changes requested):
Please take a look at the comments.

processor/processor.go
utsabc: Addressed

processor/processor.go
utsabc: Addressed

utsabc:
> Is it possible to add tests?

Added unit test cases

processor/processor.go
```diff
@@ -2589,3 +2579,13 @@ func (proc *HandleT) updateRudderSourcesStats(ctx context.Context, tx jobsdb.Sto
 	err := rsourcesStats.Publish(ctx, tx.SqlTx())
 	return err
 }
+
+func filterConfig(eventCopy *transformer.TransformerEventT, destination *backendconfig.DestinationT) {
+	if configsToFilterI, ok := destination.DestinationDefinition.Config["configFilters"]; ok {
+		if configsToFilter, ok := configsToFilterI.([]string); ok {
```
atzoum, line 2585: In practice, casting to `[]string` has no chances for success, since this config will be populated through json unmarshalling. 
https://go.dev/play/p/BLqiaLRfgO-

- Please adapt relevant tests to use proper, unmarshalled configs
- Make sure to typecheck the next casting that will be introduced too (`k, ok := configKey.(string); ok`) as a slice of `any` type can literally hold **any type**

utsabc: Got it addressed the same, added unit test cases as well

chandumlg (approved):

processor/processor.go
```diff
@@ -509,7 +508,7 @@ func loadConfig() {
 	config.RegisterBoolConfigVariable(false, &enableEventSchemasAPIOnly, true, "EventSchemas.enableEventSchemasAPIOnly")
 	config.RegisterIntConfigVariable(10000, &maxEventsToProcess, true, 1, "Processor.maxLoopProcessEvents")
 
-	batchDestinations, customDestinations = misc.LoadDestinations()
+	batchDestinations, _ = misc.LoadDestinations()
```
atzoum, line 511: since `misc.LoadDestinations` is only used here and we don't really care about custom destinations, please simplify it by renaming it to `misc.BatchDestinations` and  only return a single value
```suggestion
	batchDestinations = misc.BatchDestinations()
```

processor/processor_test.go
```diff
@@ -2407,3 +2407,103 @@ var _ = Describe("TestSubJobMerger", func() {
...
+				Name: "test",
+				Config: map[string]interface{}{
+					"config_key":   "config_value",
+					"long_config1": "long_config1_value..................................",
+					"long_config2": map[string]interface{}{"hello": "world"},
+				},
+				DestinationDefinition: backendconfig.DestinationDefinitionT{
+					Config: map[string]interface{}{
+						"configFilters": []interface{}{"long_config1", "long_config2"},
+					},
```
atzoum, line 2425: At least once we should use the outcome of `json.Unmarshal` in our tests, by unmarshalling a json string.

processor/processor.go
utsabc: Addressed

processor/processor_test.go
utsabc: Addressed

atzoum (approved):

fracasula (commented):
@utsabc linting was not passing when you merged. can you please open a PR to fix the linting warnings? https://github.com/rudderlabs/rudder-server/actions/runs/3375866788/jobs/5602936742

PR #2626, opened 2022-10-27
chore: modified statsMiddleware to be generic

saurav-malani:
# Description
Modified `statsMiddleware` to be more generics. So, that we can import and use it at other places as well, like regulation-manager.

## Notion Ticket

https://www.notion.so/rudderstacks/make-gateway-middleware-generic-use-it-in-regulation-manager-749a8f97268e46c3bff11d8474d1ce84

## Security


atzoum (approved):

fracasula (approved):
Looks good but the code is not formatted (linting step is not passing).

chandumlg (approved):

PR #2627, opened 2022-10-27
feat(warehouse): degraded workspace id

lvrach:
# Description

## Feature

Degradation is a common concept in rudder-stack infrastructure, where a service runs in a limited way allowing some operations to happen safely. In the warehouse domain, degraded mode means:
* No processing or database updates are taking place.
* Will not accept any new request. Most APIs are not registered.

Workspace degradation is a feature that enables applying the degradation mode for specific workspaceIDs.

## Motivation

1. Migrate to multi-tenant cluster without downtime for customers that are not being migrated but are residents in the source or destination of migration.
2. Isolate problematic workspaceIDs if the need arises/
3. Set the foundations for a multi-master warehouse system.

## Implementation 
A new component is introduced, `multitenant.Manager` is responsible for:
* be the source of truth for degraded workspaceIDs
* filter degraded workspaceIDs from backend config 
* map sourceID to workspaceID

Places that have been affected:
1. HTTP API handlers will not respond with `http.StatusServiceUnavailable`, if a workspace id is degraded.
2. archiver will [not archive](https://github.com/rudderlabs/rudder-server/pull/2627/files#diff-b3734beaeca355db94d2e583a7ab81e1f1ad1dfa660d2397cbe98813683d34f3R171) degraded workspaceID 
3. [WatchConfig](https://github.com/rudderlabs/rudder-server/pull/2627/files#diff-ca619c46891b7f54338e43d99d700a883ce5ca90491a05a90179a02ad1a8a3ceR1461) has replace backendconfig.Subcribe  so that workers related to degraded mode ids won't start.

## Archive refactoring

To properly test the changes of the archiver:
* I have added assertions for minio contents and db states
* Added table test for workspaceId functionality
* Convert from gingko to go native tests, as I had trouble setting up table tests
* Small refactor of the archive, introducing struct and DI (surely further work remains)

## limitation and out-of-scope

* Degraded workspaces are statically defined. A service restart will be required every time we change them. The manager is design is a way to easily support dynamic workspace ids in the future.

* The manager is not properly injected in all the cases. To use proper DI for the `manager` significant refactoring will be required.

* async framework is not in the scope of this PR.

## Notion Ticket

https://www.notion.so/rudderstacks/workspace-degraded-mode-5f231fa3899b4a79bf230d2ee12e85ca

## Security


warehouse/multitenant/manager.go
```diff
@@ -0,0 +1,139 @@
...
+
+		for _, workspaceID := range m.DegradedWorkspaceIDs {
+			m.excludeWorkspaceIDMap[workspaceID] = struct{}{}
+		}
+		m.ready = make(chan struct{})
+	})
+}
+
+// Run is a blocking function that executes manager background logic.
+func (m *Manager) Run(ctx context.Context) error {
```
achettyiitr, line 48: Since we are always returning nil. Should we remove it?

warehouse/multitenant/manager.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// SourceToWorkspace returns the workspaceID for a given sourceID, even if workspaceID is degraded.
+// An error is returned if the sourceID is not found, or context is canceled.
+//
+//	NOTE: This function blocks until the backend config is loaded.
+func (m *Manager) SourceToWorkspace(ctx context.Context, sourceID string) (string, error) {
+	m.init()
+
+	select {
```
achettyiitr, line 95: Should we add a timeout here as well? Since these are being used in sources pending events.

warehouse/multitenant/manager.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (m *Manager) init() {
+	m.initOnce.Do(func() {
+		if m.DegradedWorkspaceIDs == nil {
+			m.DegradedWorkspaceIDs = degradedWorkspaceIDs
+		}
+
+		m.sourceIDToWorkspaceID = make(map[string]string)
+		m.excludeWorkspaceIDMap = make(map[string]struct{})
```
achettyiitr, line 38: We are not using this anywhere?

warehouse/multitenant/manager.go
lvrach: I believe `ctx` should be enough, the call can chose to provide a timeout or not.

In case for sousrce pending events endpoint, it will cancel when the HTTP requests gets canceled or timeouts

achettyiitr: My bad. Didn't see the request context being passed. Please ignore.

warehouse/multitenant/manager.go
achettyiitr: I believe we are `degradedWorkspaceIDs` as `excludeWorkspace`. Do we still need this.

achettyiitr (approved):

achettyiitr:
Kudos @lvrach. It was really nice going through this and learning a lot from it.

abhimanyubabbar (approved):
This looks really good. 👍🏼

warehouse/archive/archiver.go
```diff
@@ -57,13 +59,22 @@ type uploadRecord struct {
 	startLoadFileID    int64
 	endLoadFileID      int64
 	uploadMetdata      json.RawMessage
+	workspaceID        string
 }
 
-func backupRecords(args backupRecordsArgs) (backupLocation string, err error) {
-	pkgLogger.Infof(`Starting backupRecords for uploadId: %s, sourceId: %s, destinationId: %s, tableName: %s,`, args.uploadID, args.sourceID, args.destID, args.tableName)
+type Archiver struct {
+	DB          *sql.DB
```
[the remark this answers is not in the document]
lvrach: I prefer this approach for packages that are used as internal components. Let me explain,

If we were to introduce a NewArchive method, is going to need 5 args, and more will be added in the future. In this case, we could use opts fn or a separate struct for the options that still require validation.

In general, is common in Go to initialise struct like this (http.Server, for example).

If we need validation we can do it lazily as part of `.Run` method for example, but personally, for internal packages, I find it overkill.

PR #2632, opened 2022-10-31
feat: regulation api support for Universal Analytics

sanpj2292:
# Description

In this PR, we have included the Regulation API support for Universal Analytics(GA) which uses OAuth for processing the user deletion request
This work would unblock any other destinations that might use OAuth for processing delete user requests

## Notion Ticket

[OAuth in regulation Worker module](https://www.notion.so/rudderstacks/Integrate-OAuth-in-regulation-worker-module-of-server-dae0cf6ef89e41eb91e9212ac524cf49)

## Security


regulation-worker/cmd/main.go
```diff
@@ -21,15 +21,21 @@ import (
...
 )
 
 var pkgLogger = logger.NewLogger().Child("regulation-worker")
 
-func main() {
+func InitAll() {
 	initialize.Init()
 	backendconfig.Init()
+	oauth.Init()
+}
```
saikumarrs, line 35: I believe we shouldn't ideally be calling this function again anywhere else (even mistakenly). Hence, I recommend merging it with the `main` function for now.

Suppose the steps in the init phase grow big enough in the future. In that case, we can consider introducing a flag to track the initialization status and validate accordingly.

regulation-worker/internal/delete/api/api.go
```diff
@@ -39,6 +41,31 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
...
 	pkgLogger.Debugf("payload: %#v", bodySchema)
+	pkgLogger.Debugf("Regulation Delete API called:\n")
+
+	var tokenStatusCode int
+	var accountSecretInfo *oauth.AuthResponse
+	// identifier to know if the destination supports OAuth
+	// TODO: "rudderAccountId" has to change to "rudderUserDeleteAccountId"
+	rudderUserDeleteAccountId, delAccountIdExists := destConfig["rudderAccountId"]
+	// TODO: This needs to be changed
+	isOauthEnabled := delAccountIdExists || destName == "GA"
```
saikumarrs, line 52: We should define a separate collection for OAuth destination names and use that to validate rather than hardcoding in the code.

saikumarrs: Maybe that's what the above TODO is for.

regulation-worker/internal/delete/api/api.go
```diff
@@ -39,6 +41,31 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 
 	bodySchema := mapJobToPayload(job, strings.ToLower(destName), destConfig)
 	pkgLogger.Debugf("payload: %#v", bodySchema)
+	pkgLogger.Debugf("Regulation Delete API called:\n")
+
+	var tokenStatusCode int
+	var accountSecretInfo *oauth.AuthResponse
+	// identifier to know if the destination supports OAuth
+	// TODO: "rudderAccountId" has to change to "rudderUserDeleteAccountId"
+	rudderUserDeleteAccountId, delAccountIdExists := destConfig["rudderAccountId"]
```
saikumarrs, line 50: Isn't the presence of a valid value in `rudderUserDeleteAccountId` enough to determine that a destination supports user deletion via OAuth?

regulation-worker/internal/delete/api/api.go
```diff
@@ -39,6 +41,31 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
...
+			DestDefName:     destName,
+			EventNamePrefix: "fetch_token",
+		})
+		pkgLogger.Infof(`[%s][FetchToken] Token Fetch Method finished (statusCode, value): (%v, %+v)`, destName, tokenStatusCode, accountSecretInfo)
+		if tokenStatusCode != http.StatusOK {
+			pkgLogger.Errorf(`[%s][FetchToken] Error in Token Fetch statusCode: %d\t error: %s\n`, destName, tokenStatusCode, accountSecretInfo.Err)
+		}
+	} else {
+		pkgLogger.Errorf("[%v] Destination probably doesn't support OAuth or some issue happened while doing OAuth for deletion", destName)
+	}
```
saikumarrs, line 68: All this can be moved after Ln 81.

regulation-worker/internal/delete/api/api.go
```diff
@@ -53,6 +80,16 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 	}
 	req.Header.Set("Content-Type", "application/json")
 
+	// setting oauth related information
+	if isOauthEnabled {
+		payload, unmarshalErr := json.Marshal(accountSecretInfo.Account)
```
saikumarrs, line 85:
```suggestion
		payload, marshalErr := json.Marshal(accountSecretInfo.Account)
```

regulation-worker/internal/delete/api/api.go
```diff
@@ -53,6 +80,16 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
...
+		if unmarshalErr != nil {
+			pkgLogger.Errorf("error while unmarshalling account secret information: %v", unmarshalErr)
```
saikumarrs, line 87:
```suggestion
			pkgLogger.Errorf("error while marshalling account secret information: %v", marshalErr)
```

regulation-worker/internal/delete/api/api.go
```diff
@@ -82,6 +119,39 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 		pkgLogger.Errorf("error while decoding response body: %v", err)
 		return model.JobStatusFailed
 	}
+	// Refresh Flow Start
+	var isRefresh bool
+	for _, jobResponse := range jobResp {
+		isRefresh = strings.TrimSpace(jobResponse.AuthErrorCategory) != "" && jobResponse.AuthErrorCategory == oauth.REFRESH_TOKEN
```
saikumarrs, line 125: Isn't this enough?
```suggestion
		isRefresh = jobResponse.AuthErrorCategory == oauth.REFRESH_TOKEN
```

regulation-worker/internal/delete/api/api.go
```diff
@@ -82,6 +119,39 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
...
+		var refSecret *oauth.AuthResponse
+		var errCatStatusCode int
+		refTokenParams := &oauth.RefreshTokenParams{
+			Secret:          accountSecretInfo.Account.Secret,
+			WorkspaceId:     job.WorkspaceID,
+			AccountId:       rudderUserDeleteAccountId.(string),
+			DestDefName:     destName,
+			EventNamePrefix: "refresh_token",
+		}
+		errCatStatusCode, refSecret = api.OAuth.RefreshToken(refTokenParams)
```
saikumarrs, line 142: Just a naive question.
Would it be beneficial if we invalidate/delete the current token we have rather than generate a new one upon failure? So, the next token fetch call would force the module to generate/refresh a new one.

We never know when the next request would come (immediately or 5 mins later), and if the token is generated way early, all that time on the token is expired, doing nothing.

regulation-worker/internal/delete/api/api.go
sanpj2292: Basically this is for testing purpose only. The TODO is to change the logic of checking oauth enabling

regulation-worker/internal/delete/api/api.go
sanpj2292: The thing is we can do this check through `DestinationDefinition.Config.auth.type` check rather than `Destination.Config`. But we need to decide if going via DestinationDefinition makes sense over Destination.Config or vice-versa

regulation-worker/internal/delete/api/api.go
sanpj2292: ### Heads-up about each of the methods

- `FetchToken` only fetches what is present in the config-be DB
- `RefreshToken`
  - if the token sent(from `RefreshToken` method) matches the one present already in config-be DB, refreshes token present in the config-be DB
  - if the token sent(from `RefreshToken` method) doesn’t match the one present already in config-be DB, the token present in the config-be DB is sent
  - if the token sent(from `RefreshToken` method) is not defined or empty or null, the token present in the config-be DB is sent

So if we invalidate or delete the token on failure, the `FetchToken` call wouldn’t work & provide the expired token always(as that is present in DB)

### What happens if we replace `FetchToken` with `RefreshToken` ?

As you see in `router.go`, we send secret information after receive response from transformer proxy(`destinationJob.JobMetadataArray[0].Secret`)
- Based on this, we’d check if the token is present in server-cache is same as the one we used, then try to send this to config-be for refreshing(You can refer [this block of code](https://github.com/rudderlabs/rudder-server/blob/master/router/oauthResponseHandler/oauthResponseHandler.go#L180-L189) to understand)
-  In config-be, we check if the one present in DB is same as the one that we sent
    - If both are same, we refresh the flow, as we know that we have to refresh
    - If both are not same, the one present in DB is sent back(as valid token)

Basically, since at the start, we wouldn’t have any information about the token, the one present in cache is always sent or the one present in config-be DB is sent

It’d not be of much use for us!

regulation-worker/cmd/main.go
sanpj2292: If you keenly observe this is the same way that the `rudder-server` code is built. Not saying that this is the right way but I was being consistent with the approach we took inside the code-base, so that when we decide to apply changes, we do it in the same way across the code-base

regulation-worker/internal/delete/api/api.go
sanpj2292: Let me know if it makes sense

regulation-worker/internal/delete/api/api.go

regulation-worker/internal/delete/api/api.go

regulation-worker/internal/delete/api/api.go

regulation-worker/cmd/main.go
sanpj2292: Please check [this](https://github.com/rudderlabs/rudder-server/blob/master/main.go#L141-L208)

koladilip (commented):
Partial review

regulation-worker/cmd/main.go
saikumarrs: Got it. Should we add the flag for extra safety? Thoughts?

regulation-worker/internal/delete/api/api.go
saikumarrs: Ok. Keeping this open as a reminder to address it.

regulation-worker/internal/delete/api/api.go
saikumarrs: My suggestion was to change the OAuth flow to something like this:

If a token is expired, rather than refreshing the token, invalidate it (or delete it). So, the next token fetch request is forced to refresh it.

So, in terms of APIs, we have these two:
- `fetchToken`
  - Fetches the token from cache or DB.
  - If needed, refreshes the token and cache it for future use.
- `invalidateToken`
  - Deletes from cache and DB.

@rudderlabs/integrations-platform would love to hear your thoughts as well!

regulation-worker/internal/delete/api/api.go
```diff
@@ -82,6 +119,39 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 		pkgLogger.Errorf("error while decoding response body: %v", err)
 		return model.JobStatusFailed
 	}
+	// Refresh Flow Start
+	var isRefresh bool
+	for _, jobResponse := range jobResp {
```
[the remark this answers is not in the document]

regulation-worker/internal/delete/api/api.go
```diff
@@ -39,6 +41,31 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
...
+			WorkspaceId:     job.WorkspaceID,
+			DestDefName:     destName,
+			EventNamePrefix: "fetch_token",
+		})
+		pkgLogger.Infof(`[%s][FetchToken] Token Fetch Method finished (statusCode, value): (%v, %+v)`, destName, tokenStatusCode, accountSecretInfo)
+		if tokenStatusCode != http.StatusOK {
+			pkgLogger.Errorf(`[%s][FetchToken] Error in Token Fetch statusCode: %d\t error: %s\n`, destName, tokenStatusCode, accountSecretInfo.Err)
+		}
+	} else {
+		pkgLogger.Errorf("[%v] Destination probably doesn't support OAuth or some issue happened while doing OAuth for deletion", destName)
```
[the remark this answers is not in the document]

regulation-worker/internal/delete/api/api.go

regulation-worker/internal/delete/api/api.go
saikumarrs: Moving this discussion to an internal forum.

regulation-worker/internal/delete/api/api.go
```diff
@@ -49,11 +53,39 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
...
+		if tokenStatusCode != http.StatusOK {
+			pkgLogger.Errorf(`[%s][FetchToken] Error in Token Fetch statusCode: %d\t error: %s\n`, destDetail.Name, tokenStatusCode, accountSecretInfo.Err)
+		}
+		// setting oauth related information
+		payload, marshalErr := json.Marshal(accountSecretInfo.Account)
+		if marshalErr != nil {
+			pkgLogger.Errorf("error while marshalling account secret information: %v", marshalErr)
+			return model.JobStatusFailed
+		}
+		req.Header.Set("X-Rudder-Dest-Info", string(payload))
```
saurav-malani, line 79: would it make things more readable, if we move this complete `OAuth` fetching and marshalling thing to a separate function & set it there.
```suggestion
err:=api.setOAuthToken(req)
if err!=nil{
model.JobStatusFailed
}
```

regulation-worker/internal/delete/api/api.go
```diff
@@ -76,6 +108,22 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 		return model.JobStatusFailed
 	}
 
+	// Refresh Flow Start
+	if delAccountIdExists {
+		respParams := &handleRefreshFlowParams{
+			jobResponses: jobResp,
```
saurav-malani, line 114: previously, we were not using `jobResp`. Can you please confirm that transformer doesn't really sends response for each users in the payload that we sent?

regulation-worker/internal/delete/api/api.go
```diff
@@ -49,11 +53,39 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 	}
 	req.Header.Set("Content-Type", "application/json")
 
+	var tokenStatusCode int
+	var accountSecretInfo *oauth.AuthResponse
+	// identifier to know if the destination supports OAuth for regulation-api
+	rudderUserDeleteAccountId, delAccountIdExists := destDetail.Config["rudderDeleteAccountId"]
```
saurav-malani, line 59: To increase the readability of the code.
```suggestion
	rudderUserDeleteAccountId, isOAuthRequired := destDetail.Config["rudderDeleteAccountId"]
```

regulation-worker/internal/delete/api/api.go
```diff
@@ -76,6 +108,22 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 		return model.JobStatusFailed
 	}
 
+	// Refresh Flow Start
+	if delAccountIdExists {
```
saurav-malani, line 112: Won't it be better to know, what would be the status code returned by the transformer in case of token expiry ( I guess it will be 401). Once, we know it, we can avoid checking for each user in the batch if token for it failed or not. And, directly decide based on the status code if we need to refresh to token or not?

regulation-worker/internal/delete/api/api.go
```diff
@@ -76,6 +108,22 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
...
+		respParams := &handleRefreshFlowParams{
+			jobResponses: jobResp,
+			secret:       accountSecretInfo.Account.Secret,
+			workspaceId:  job.WorkspaceID,
+			accountId:    rudderUserDeleteAccountId.(string),
+			destName:     destDetail.Name,
+		}
+		refreshFlowStatus := api.handleRefreshFlow(respParams)
```
saurav-malani, line 120: I think this flow of handling refresh could be improved if we know the status code of the response returned.

saurav-malani (changes requested):

regulation-worker/cmd/main.go
```diff
@@ -58,6 +64,10 @@ func Run(ctx context.Context) {
 	if err != nil {
 		panic(fmt.Errorf("error while getting workspaceId: %w", err))
 	}
+	// setting up oauth
+	OAuth := oauth.NewOAuthErrorHandler(backendconfig.DefaultBackendConfig)
+	OAuth.Setup()
```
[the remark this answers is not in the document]
sanpj2292: Did we merge this kind of implementation any where else in `rudder-server` ?

regulation-worker/internal/destination/destination.go
```diff
@@ -60,6 +60,8 @@ func (d *DestMiddleware) GetDestDetails(ctx context.Context, destID string) (mod
 					destDetail.Config = dest.Config
 					destDetail.DestinationID = dest.ID
 					destDetail.Name = dest.DestinationDefinition.Name
+					// Destination Definition Config would most likely be needed
```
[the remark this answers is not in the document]
sanpj2292: Done

regulation-worker/internal/delete/kvstore/kvstore.go
```diff
@@ -21,7 +21,12 @@ func (*KVDeleteManager) GetSupportedDestinations() []string {
 	return supportedDestinations
 }
 
-func (*KVDeleteManager) Delete(_ context.Context, job model.Job, destConfig map[string]interface{}, destName string) model.JobStatus {
+func (*KVDeleteManager) Delete(_ context.Context, job model.Job, destDetail model.Destination) model.JobStatus {
+	// destConfig map[string]interface{},
+	// destName string,
```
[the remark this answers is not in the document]
sanpj2292: Done

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -315,9 +315,13 @@ func (*BatchManager) GetSupportedDestinations() []string {
 func (bm *BatchManager) Delete(
 	ctx context.Context,
 	job model.Job,
-	destConfig map[string]interface{},
-	destName string,
+	destDetail model.Destination,
 ) model.JobStatus {
+	// destConfig map[string]interface{},
+	// destName string,
```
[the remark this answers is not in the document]
sanpj2292: Done

regulation-worker/cmd/main.go
```diff
@@ -75,6 +85,8 @@ func Run(ctx context.Context) {
 			&api.APIManager{
 				Client:           &http.Client{Timeout: config.GetDuration("HttpClient.regulationWorker.transformer.timeout", 60, time.Second)},
 				DestTransformURL: config.MustGetString("DEST_TRANSFORM_URL"),
+				// OAuthResponseHandler instance
```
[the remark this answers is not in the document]
sanpj2292: Done

regulation-worker/internal/delete/api/api.go
```diff
@@ -49,11 +53,39 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
...
+		// Get Access Token Information to send it as part of the event
+		tokenStatusCode, accountSecretInfo = api.OAuth.FetchToken(&oauth.RefreshTokenParams{
+			AccountId:       rudderUserDeleteAccountId.(string),
+			WorkspaceId:     job.WorkspaceID,
+			DestDefName:     destDetail.Name,
+			EventNamePrefix: "fetch_token",
+		})
+		pkgLogger.Debugf(`[%s][FetchToken] Token Fetch Method finished (statusCode, value): (%v, %+v)`, destDetail.Name, tokenStatusCode, accountSecretInfo)
+		if tokenStatusCode != http.StatusOK {
+			pkgLogger.Errorf(`[%s][FetchToken] Error in Token Fetch statusCode: %d\t error: %s\n`, destDetail.Name, tokenStatusCode, accountSecretInfo.Err)
```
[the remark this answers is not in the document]

regulation-worker/internal/model/model.go
```diff
@@ -42,6 +42,7 @@ type User struct {
 
 type Destination struct {
 	Config        map[string]interface{}
+	DestDefConfig map[string]interface{}
```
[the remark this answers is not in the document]
sanpj2292: As discussed, we are using this data in identifying if oauth is enabled for a destination. Please verify

regulation-worker/internal/delete/api/api.go
sanpj2292: Slightly modified the implementation, please take a look

regulation-worker/internal/delete/api/api.go

regulation-worker/internal/delete/api/api.go
sanpj2292: As discussed, we are moving ahead with current implementation of checking with an enum kind of thing that we get through the response from transformer rather than relying on source-codes

regulation-worker/cmd/main.go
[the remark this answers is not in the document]
sanpj2292: Combined the oauth handler instance and setup. Can you once check and verify ?

regulation-worker/internal/delete/api/api.go
sanpj2292: The refactoring for refresh flow handling is complete. Can you please check and verify ?

utils/misc/misc.go
```diff
@@ -1354,3 +1354,34 @@ func CopyStringMap(originalMap map[string]string) map[string]string {
...
+}
+
+type GetAccountIdParams struct {
+	DestConfig map[string]interface{}
+	IdKey      string
+}
+
+func GetAccountId(params GetAccountIdParams) string {
+	idKey := params.IdKey
+	if strings.TrimSpace(idKey) == "" {
```
krishna2020, line 1378: In user delete flow, wouldn't the idKey be `rudderUserDeleteAccountId` to search in DestConfig ?

sanpj2292: Yes. We are sending the idKey as `rudderDeleteAccountId` **[here](https://github.com/rudderlabs/rudder-server/pull/2632/files#diff-a378f351d7562810ac2b3f21d979516467ca2da122de096005221d3e2218fbf8R246)**. Can you please verify ?

krishna2020: Ok, then why to default to `rudderAccountId` ? Will this not effect the eventDelivery account configuration ?

sanpj2292: When someone doesn't specifically request a specific `idKey` parameter, we assume they're probably asking the account information for event-delivery and provide the information. Is there any problem with this approach ?

regulation-worker/internal/delete/api/api.go
sanpj2292: The changes have been made, can you verify ?

regulation-worker/cmd/main.go
sanpj2292: Don't think a flag is a necessity

saikumarrs (approved):

regulation-worker/cmd/main.go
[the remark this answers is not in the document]
saurav-malani: looks good, thanks!

regulation-worker/internal/model/model.go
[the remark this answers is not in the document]
saurav-malani: yup, now I can see it being used in `oauth.OAuthParams`. So, looks good.

regulation-worker/internal/delete/api/api.go
```diff
@@ -126,3 +149,126 @@ func mapJobToPayload(job model.Job, destName string, destConfig map[string]inter
...
+
+	} else {
+		pkgLogger.Errorf("[%v] Destination probably doesn't support OAuth or some issue happened while doing OAuth for deletion [Enabled: %v]", destDetail.Name, oAuthParamsResult.Enabled)
+	}
+	return &OAuthTokenResult{
+		AccountSecretInfo: accountSecretInfo,
+		DeleteAccountId:   oAuthParamsResult.AccountId,
+		IsOAuthEnabled:    oAuthParamsResult.Enabled,
+	}, nil
+}
```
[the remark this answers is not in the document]
sanpj2292: It makes sense to think of the oauth flow as an interceptor. IMHO it seems like the ideal approach for this kind of feature

Making changes to make sure we use this as an interceptor doesn't fit in our current plan. I have outlined a couple of reasons for it
1. Timelines. We want this feature to be available for customers
2. Long-term vision. As discussed in an earlier call about regulation-worker, the OAuth flow in it's entirety was to move-away from `rudder-server`. Keeping this in mind & the changes that we need to do in order for us to make sure the interceptor approach is being used will be time-taking

But we will make sure to keep this thought in mind while we work on OAuth Flow movement itself. Please let me know if this makes sense

regulation-worker/cmd/main.go
```diff
@@ -21,16 +21,22 @@ import (
 	"github.com/rudderlabs/rudder-server/regulation-worker/internal/destination"
 	"github.com/rudderlabs/rudder-server/regulation-worker/internal/initialize"
 	"github.com/rudderlabs/rudder-server/regulation-worker/internal/service"
+	oauth "github.com/rudderlabs/rudder-server/router/oauthResponseHandler"
```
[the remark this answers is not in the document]
sanpj2292: Moved to a separate package

router/oauthResponseHandler/oauthResponseHandler.go
```diff
@@ -87,7 +88,9 @@ type ControlPlaneRequestT struct {
 
 // This function creates a new OauthErrorResponseHandler
 func NewOAuthErrorHandler(provider tokenProvider) *OAuthErrResHandler {
-	return &OAuthErrResHandler{TokenProvider: provider}
+	oAuthErrResHandler := &OAuthErrResHandler{TokenProvider: provider}
+	oAuthErrResHandler.Setup()
```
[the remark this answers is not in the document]
sanpj2292: Done

router/utils/utils.go
```diff
@@ -101,24 +101,11 @@ func IsNotEmptyString(s string) bool {
 }
 
 func GetAuthType(dest backendconfig.DestinationT) (authType string) {
```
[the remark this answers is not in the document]
sanpj2292: Removed

router/utils/utils.go
```diff
@@ -101,24 +101,11 @@ func IsNotEmptyString(s string) bool {
...
-	}
-	authType, ok := authValue.(string)
-	if !ok {
-		return ""
-	}
-	return authType
+	return misc.GetAuthType(dest.DestinationDefinition.Config)
 }
 
 func GetRudderAccountId(destination *backendconfig.DestinationT) string {
```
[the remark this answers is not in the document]
sanpj2292: Moved to `/services/oauth/oauth.go`

regulation-worker/internal/delete/api/api.go
```diff
@@ -8,35 +8,40 @@ import (
...
 type APIManager struct {
 	Client           *http.Client
 	DestTransformURL string
+	OAuth            oauth.Authorizer
 }
 
 // prepares payload based on (job,destDetail) & make an API call to transformer.
 // gets (status, failure_reason) which is converted to appropriate model.Error & returned to caller.
-func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map[string]interface{}, destName string) model.JobStatus {
+func (api *APIManager) Delete(ctx context.Context, job model.Job, destDetail model.Destination) model.JobStatus {
```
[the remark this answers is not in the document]

regulation-worker/cmd/main.go
```diff
@@ -21,16 +21,22 @@ import (
...
 	"github.com/rudderlabs/rudder-server/services/filemanager"
 	"github.com/rudderlabs/rudder-server/utils/logger"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 )
 
 var pkgLogger = logger.NewLogger().Child("regulation-worker")
 
-func main() {
+func InitAll() {
```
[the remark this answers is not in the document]
sanpj2292: Done

regulation-worker/internal/delete/api/api.go
sanpj2292: As discussed in the call, this is still kept for future needs.

saurav-malani (approved):
Assuming e-2-e testing is done. Code wise looks good to me.

sanpj2292:
>Assuming e-2-e testing is done.

Had been testing for every change all along this PR

services/oauth/oauth.go
```diff
@@ -129,6 +148,51 @@ func (authErrHandler *OAuthErrResHandler) Setup() {
...
+	if !ok {
+		return ""
+	}
+	return AuthType(authType)
+}
+
+// This function creates a new OauthErrorResponseHandler
+func NewOAuthErrorHandler(provider tokenProvider, options ...func(*OAuthErrResHandler)) *OAuthErrResHandler {
+	oAuthErrResHandler := &OAuthErrResHandler{TokenProvider: provider}
+	oAuthErrResHandler.setup()
```
atzoum, line 170: we can get rid of the `setup` method
```suggestion
	oAuthErrResHandler := &OAuthErrResHandler{
		tokenProvider:    provider,
		logger:           pkgLogger,
		tr:               &http.Transport{},
		client:           &http.Client{Timeout: config.GetDuration("HttpClient.oauth.timeout", 30, time.Second)},
		destLockMap:      make(map[string]*sync.RWMutex),
		accountLockMap:   make(map[string]*sync.RWMutex),
		lockMapWMutex:    &sync.RWMutex{},
		destAuthInfoMap:  make(map[string]*AuthResponse),
		refreshActiveMap: make(map[string]bool),
		rudderFlowType:   RudderFlow_Delivery,
	}
```

regulation-worker/internal/delete/api/api.go
```diff
@@ -76,6 +106,17 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 		return model.JobStatusFailed
 	}
 
+	if isOAuthEnabled {
+		isOAuthTokenExpired := isTokenExpired(jobResp)
+		if isOAuthTokenExpired {
+			err = api.refreshOAuthToken(destination.Name, job.WorkspaceID, oAuthDetail)
```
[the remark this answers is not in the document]
sanpj2292: We return an error. Yes, I think it will be re-tried after sometime. @saurav-malani can you provide insight here ?

regulation-worker/internal/delete/api/api.go
```diff
@@ -8,35 +8,49 @@ import (
...
 type APIManager struct {
 	Client           *http.Client
 	DestTransformURL string
+	OAuth            oauth.Authorizer
+}
+
+type OAuthDetail struct {
+	SecretToken *oauth.AuthResponse
+	ID          string
+}
```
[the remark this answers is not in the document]

regulation-worker/internal/delete/api/api.go
```diff
@@ -49,11 +63,27 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 	}
 	req.Header.Set("Content-Type", "application/json")
 
+	// check if OAuth destination
+	isOAuthEnabled := IsOAuthEnabled(destination.DestDefConfig)
```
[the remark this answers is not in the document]

regulation-worker/internal/delete/api/api.go
```diff
@@ -126,3 +163,63 @@ func mapJobToPayload(job model.Job, destName string, destConfig map[string]inter
...
+		return errors.New(marshalFailErr)
+	}
+	req.Header.Set("X-Rudder-Dest-Info", string(payload))
+	return nil
+}
+
+func IsOAuthEnabled(destDefConfig map[string]interface{}) bool {
+	authType := oauth.GetAuthType(destDefConfig)
+	return authType == oauth.OAuth
+}
```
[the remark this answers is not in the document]

services/oauth/oauth.go
```diff
@@ -67,10 +85,10 @@ type OAuthErrResHandler struct {
 	refreshActiveMap     map[string]bool // Used to check if a refresh request for an account is already InProgress
 	disableDestActiveMap map[string]bool // Used to check if a disable destination request for a destination is already InProgress
 	TokenProvider        tokenProvider
+	RudderFlowType       RudderFlow
```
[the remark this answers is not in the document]
sanpj2292: Done

services/oauth/oauth.go
```diff
@@ -112,13 +125,19 @@ type tokenProvider interface {
 	AccessToken() string
 }
 
+type OAuthParams struct {
+	DestConfig    map[string]interface{}
+	DestDefConfig map[string]interface{}
+	IdKey         string
+}
+
```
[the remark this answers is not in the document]
sanpj2292: Done

services/oauth/oauth.go
sanpj2292: Done

regulation-worker/internal/delete/api/api.go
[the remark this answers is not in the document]
atzoum: What will happen if we have only 1 job and its retry backoff becomes higher than the token's validity period? Won't we enter an endless loop?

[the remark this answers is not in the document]
saurav-malani: yes @atzoum, this is a possible known issue with the current implementation. In fact, I did raise this concern is our discussion with the transformation team when you and Leo where also in the call.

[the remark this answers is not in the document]
atzoum: is there something we are afraid of to not retry the request a second time?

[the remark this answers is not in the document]
saurav-malani: no, I think given that in response from transformer we get status corresponding to each user in the batch. It should be quite simple. And, I don't see any problem.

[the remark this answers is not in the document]
sanpj2292: Currently the userIds for which the failure happens is not being sent in the response. But it can be captured & server needs to make changes to it's response schema to capture that detail & process accordingly which I think is not in scope of this PR.

As far as retry goes, would 1 retry solve the problem or should we retry multiple times ?!

regulation-worker/internal/delete/api/api.go
```diff
@@ -75,8 +105,28 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
...
+			err = api.refreshOAuthToken(destination.Name, job.WorkspaceID, oAuthDetail)
+			if err != nil {
+				pkgLogger.Error(err)
+				return model.JobStatusFailed
+			}
+			// retry the request
+			pkgLogger.Debug("Retrying deleteRequest job for the whole batch")
+			return api.Delete(ctx, job, destination)
+		}
+	}
```
atzoum, line 123:
```suggestion
	if isOAuthEnabled && isTokenExpired(jobResp) {
		err = api.refreshOAuthToken(destination.Name, job.WorkspaceID, oAuthDetail)
		if err != nil {
			pkgLogger.Error(err)
			return model.JobStatusFailed
		}
		// retry the request
		pkgLogger.Debug("Retrying deleteRequest job for the whole batch")
		return api.Delete(ctx, job, destination)
	}
```

atzoum (approved):

regulation-worker/internal/delete/api/api.go

PR #2637, opened 2022-11-02
fix: send router transform failures to live events

cisse21:
# Description

For events which fail at router transform, we are not sending them to live events viewers resulting in customers being unable to view the errors. This PR aims to send those jobs to config backend for live events viewers

## Notion Ticket

[ Notion Link ](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=81cad161836f407b9983a68a20b9e0ce&pm=s)

## Security


router/router.go
```diff
@@ -592,7 +592,7 @@ func (worker *workerT) processDestinationJobs() {
 	})
 
 	for _, destinationJob := range worker.destinationJobs {
-		var attemptedToSendTheJob bool
+		var attemptedToSendTheJob = true
```
atzoum, line 595: seems complex and error prone :) how do we test it?

cisse21: Yeah even I was thinking how can we test these cases

cisse21: Was thinking do we need to complicate this logic and probably we should get rid of the variable all together

router/router.go
```diff
@@ -855,27 +850,22 @@ func (worker *workerT) processDestinationJobs() {
 			userToJobIDMap[destinationJobMetadata.UserID] = destinationJobMetadata.JobID
 		}
 
-		if attemptedToSendTheJob {
-			status.AttemptNum++
-		}
-
+		status.AttemptNum++
```
chandumlg, line 853: The only risk I see by removing this check is, if transformer throws a 5xx for a valid job (because of some bug), we would end up aborting after certain number of retries.

This was introduced to help in those scenarios. I am not sure if there is a proper alert mechanism to identify these bugs.

chandumlg: Any thoughts. @cisse21 @atzoum

atzoum: Even now we don't know this happens, so we could add an alert for > HTTP 500 errors?

chandumlg: Now, we will know if there is a growth in router jobs queue. If in any case, the queue doesn't grow, you are right. We aren't alerted.

Integrations team might already have alerts on router_response metrics. I am not completely sure though.

I am okay with this change. Just want to highlight above.

chandumlg (approved):

go.mod
```diff
@@ -215,3 +215,5 @@ require (
 )
 
 require github.com/samber/lo v1.33.0
+
+require gotest.tools/v3 v3.3.0 // indirect
```
atzoum, line 219: where did that come from?

cisse21: go mod tidy just added this automatically for some reason

atzoum: if you remove the `go.mod`/`go.sum` changes and run go mod tidy again, will it do the same?

atzoum (approved):

PR #2638, opened 2022-11-02
feat(warehouse): unrecognized schema in warehouse

achettyiitr:
# Description

The changes revolve around `uploadSchema` and `schemaInWarehouse` terminology. 

- `uploadScheme` It is being prepared by consolidating local schema (copy of warehouse schema) and staging files schema. 
- `schemaInWarehouse` It is being prepared by querying the warehouse destinations and mapping them with datatypes supported by Rudderstack. During the fetching of the schema from the warehouse destinations, if there are some datatypes that are not supported by RudderStack we skip those.

During the loading of the data, we calculate schema diff between the `uploadScheme` and `schemaInWarehouse` to create the necessary tables and columns. There are two things: 
1. Since `schemaInWarehouse` is skipping the unsupported datatypes, it is resulting in creating columns every time for unsupported datatypes.  Errors are being suppressed.
2. Since we started batching `ALTER` queries, we stopped suppressing the errors because one or more columns already present will result in other columns not getting altered.

So, we introduced one more parameter called `unrecognizedSchema` which will be populated during the fetching of the schema from unsupported datatypes. We are using it to ignore columns that are already present in the warehouse.

## Notion Ticket

https://www.notion.so/rudderstacks/Warehouse-unrecognized-schema-4029829a49564d5087a2186464b63e63

## Security

### DIFF
--- a/warehouse/schema.go
+++ b/warehouse/schema.go
@@ -376,7 +376,7 @@ func hasSchemaChanged(localSchema, schemaInWarehouse warehouseutils.SchemaT) boo
 	return false
 }
 
-func getTableSchemaDiff(tableName string, currentSchema, currentUnRecognizedSchema, uploadSchema warehouseutils.SchemaT) (diff warehouseutils.TableSchemaDiffT) {
+func getTableSchemaDiff(tableName string, currentSchema, uploadSchema warehouseutils.SchemaT) (diff warehouseutils.TableSchemaDiffT) {
 	diff = warehouseutils.TableSchemaDiffT{
 		ColumnMap:     make(map[string]string),
 		UpdatedSchema: make(map[string]string),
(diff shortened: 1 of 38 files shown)


Jayachand (approved):

warehouse/schema.go
```diff
@@ -13,12 +13,13 @@ import (
...
-	warehouse         warehouseutils.Warehouse
-	localSchema       warehouseutils.SchemaT
-	schemaInWarehouse warehouseutils.SchemaT
-	uploadSchema      warehouseutils.SchemaT
+	dbHandle                      *sql.DB
+	stagingFiles                  []*StagingFileT
+	warehouse                     warehouseutils.Warehouse
+	localSchema                   warehouseutils.SchemaT
+	schemaInWarehouse             warehouseutils.SchemaT
+	unRecognizedSchemaInWarehouse warehouseutils.SchemaT
```
abhimanyubabbar, line 21: Should naming be `unrecognizedSchemaInWarehouse` ?

abhimanyubabbar (approved):
nit: Simple change in naming. Nothing special. If you feel it looks good handle it otherwise it's okay.

Good Job 🥳 🎉

warehouse/schema.go
achettyiitr: Done.

PR #2644, opened 2022-11-03
fix: remove gateway db write panic

Sidddddarth:
# Description

Remove panic in `gateway.dbWriterWorkerProcess` on store failure.

## Notion Ticket

[remove gateway panic on write failure](https://www.notion.so/rudderstacks/remove-gateway-panic-on-write-timeout-f65be5e734cc42718c37a0436b449308)

## Security

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -323,8 +323,10 @@ func (gateway *HandleT) userWorkerRequestBatcher() {
 // This in turn sends the error over the done channel of each respective webRequest.
 func (gateway *HandleT) dbWriterWorkerProcess() {
 	for breq := range gateway.batchUserWorkerBatchRequestQ {
-		jobList := make([]*jobsdb.JobT, 0)
-		var errorMessagesMap map[uuid.UUID]string
+		var (
+			jobList          = make([]*jobsdb.JobT, 0)
+			errorMessagesMap = map[uuid.UUID]string{}
+		)
 
 		for _, userWorkerBatchRequest := range breq.batchUserWorkerBatchRequest {
 			jobList = append(jobList, userWorkerBatchRequest.jobList...)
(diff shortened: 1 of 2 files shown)


gateway/gateway.go
```diff
@@ -354,7 +356,9 @@ func (gateway *HandleT) dbWriterWorkerProcess() {
 		})
 		cancel()
 		if err != nil {
-			panic(err)
+			for _, job := range jobList {
+				errorMessagesMap[job.UUID] = err.Error()
+			}
 		}
 		gateway.dbWritesStat.Count(1)
```
atzoum, line 363: we cannot be confident that the error will always be `context deadline exceeded`, thus should give precedence to the context's error if there is one.
It would also be fragile to keep adding different errors in the error map, since there will always be a possibility to miss something, e.g. `pq: canceling statement due to user request`
```go
		if err != nil {
			errorMessage := err.Error()
			if ctx.Err() != nil {
				errorMessage = ctx.Err().Error()
			}
			for _, job := range jobList {
				errorMessagesMap[job.UUID] = errorMessage
			}
		}
		cancel()
```

gateway/response/response.go
```diff
@@ -52,6 +52,12 @@ const (
 	ErrorInParseMultiform = "Error during parsing multiform"
 	// NotRudderEvent = Event is not a Valid Rudder Event
 	NotRudderEvent = "Event is not a valid rudder event"
+	// ContextDeadlineExceeded - context deadline exceeded
+	ContextDeadlineExceeded = "context deadline exceeded"
+	// could not acquire a dslist read lock: context deadline exceeded
+	DSListReadLockTimeout = "could not acquire a dslist read lock: context deadline exceeded"
```
atzoum, line 58:
```suggestion
```

gateway/response/response.go
```diff
@@ -82,6 +88,8 @@ var statusMap = map[string]status{
 	ErrorInParseForm:                               {message: ErrorInParseForm, code: http.StatusBadRequest},
 	ErrorInParseMultiform:                          {message: ErrorInParseMultiform, code: http.StatusBadRequest},
 	NotRudderEvent:                                 {message: NotRudderEvent, code: http.StatusBadRequest},
+	ContextDeadlineExceeded:                        {message: GatewayTimeout, code: http.StatusGatewayTimeout},
+	DSListReadLockTimeout:                          {message: GatewayTimeout, code: http.StatusGatewayTimeout},
```
atzoum, line 92:
```suggestion
```

gateway/gateway.go
Sidddddarth: any context error would be populated here right?
`errorMessagesMap, err = gateway.jobsDB.StoreWithRetryEachInTx(ctx, tx, jobList)`

atzoum (approved):

cisse21 (approved):

PR #2654, opened 2022-11-04
chore(warehouse): handle schema change

achettyiitr:
# Description

### Schema Test
Use table-driven tests to handle schema change. The benefit here is we can be able to run these tests parallelly.

### Handle Schema Change Behaviour
This is applied to the mismatch of data types between the warehouse schema and staging files schema.

**Existing behavior**

- When we are trying to do the conversion of the data but we are not able to do it, we used to send nil values to the load files.

**Current behavior**
- Throw `ErrIncompatibleSchemaConv`,   when we are trying to do the conversion of the data but we are not able to do it.
- Throw `ErrSchemaConvNotSupported`, when we don't support the conversion of the data.

## Notion Ticket

https://www.notion.so/rudderstacks/Handle-schema-table-driven-tests-751462cd730d4e4aaf92fba22c575c34

## Security


warehouse/schema.go
```diff
@@ -22,36 +22,38 @@ type SchemaHandleT struct {
 	uploadSchema                  warehouseutils.SchemaT
 }
 
-func HandleSchemaChange(existingDataType, columnType string, columnVal interface{}) (newColumnVal interface{}, ok bool) {
+func HandleSchemaChange(existingDataType, newDataType string, value interface{}) (interface{}, bool) {
```
lvrach, line 25: [minor] go version 1.18  introduced `any` as an alias to `interface{}`. I personally, find it prettier.
```suggestion
func HandleSchemaChange(existingDataType, newDataType string, value any) (any, bool) {
```

warehouse/schema_test.go
```diff
@@ -6,222 +6,262 @@ import (
...
+			name:             "existing datatype is datetime, new datatype is json",
+			existingDatatype: "int",
+			newDataType:      "json",
+			value:            `{"json":true}`,
+			schemaChanged:    false,
+		},
+	}
+	for _, ip := range inputs {
+		t.Run(ip.name, func(t *testing.T) {
+			t.Parallel()
```
lvrach, line 255: 👍 nice touch

warehouse/schema.go
achettyiitr: Done.

warehouse/schema_test.go
```diff
@@ -6,222 +6,262 @@ import (
...
-				Expect(ok).To(BeFalse())
-			})
+func TestHandleSchemaChange(t *testing.T) {
+	inputs := []struct {
+		name             string
+		existingDatatype string
+		newDataType      string
+		value            interface{}
+		newColumnVal     interface{}
+		schemaChanged    bool
```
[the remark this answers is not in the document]
achettyiitr: Thanks, @lvrach for suggesting this. Although it's a small change it creates a great impact on code visibility.

warehouse/schema_test.go
```diff
@@ -3,225 +3,267 @@
...
+		{
+			name:             "existing datatype is boolean, new datatype is float",
+			existingDatatype: "boolean",
+			newDataType:      "float",
+			value:            1.501,
+			schemaChanged:    false,
+		},
+		{
+			name:             "existing datatype is boolean, new datatype is string",
+			existingDatatype: "boolean",
```
lvrach, line 144: Shouldn't we get an error when we try to do something like that?

warehouse/schema.go
```diff
@@ -22,36 +22,38 @@ type SchemaHandleT struct {
 	uploadSchema                  warehouseutils.SchemaT
 }
 
-func HandleSchemaChange(existingDataType, columnType string, columnVal interface{}) (newColumnVal interface{}, ok bool) {
+func HandleSchemaChange(existingDataType, currentDataType string, value any) (any, bool) {
```
lvrach, line 25: We should probably introduce an error in situations where conversion will lead into losing data, 
like `datetime` to `bool` 

or were conversion is invalid.

the bool, schema change can be replaced with a special error: ErrSchemaChange.

```suggestion
func HandleSchemaChange(existingDataType, currentDataType string, value any) (any, error) {
```

warehouse/schema_test.go
achettyiitr: Actually, it's an issue with the function signature we have, instead of with err we are handling it with boolean.
```
func HandleSchemaChange(existingDataType, currentDataType string, value any) (any, bool)
```

warehouse/schema.go
achettyiitr: Good point @lvrach

warehouse/schema.go
```diff
@@ -22,42 +22,59 @@ type SchemaHandleT struct {
...
+type SchemaType string
+
+const (
+	StringDataType  = "string"
+	BooleanDataType = "boolean"
+	IntDataType     = "int"
+	BigIntDataType  = "bigint"
+	FloatDataType   = "float"
+	JSONDataType    = "json"
+	TextDataType    = "text"
```
lvrach, line 34:
```suggestion
const (
	StringSchemaType  SchemaType = "string"
	BooleanSchemaType = "boolean"
	IntSchemaType     = "int"
	BigIntSchemaType  = "bigint"
	FloatSchemaType   = "float"
	JSONSchemaType    = "json"
	TextSchemaType    = "text"
```

warehouse/schema_test.go
achettyiitr: We are now handling error as well.

warehouse/model/schema.go
```diff
@@ -0,0 +1,13 @@
+package model
+
+type SchemaType string
+
+const (
+	StringDataType  SchemaType = "string"
```
achettyiitr, line 6: > only the first constant in this group has an explicit type

https://stackoverflow.com/questions/55282615/golangci-lint-constant-explicit-type

warehouse/model/schema.go
```diff
@@ -0,0 +1,13 @@
...
+	BooleanDataType SchemaType = "boolean"
```
achettyiitr, line 7: > only the first constant in this group has an explicit type

https://stackoverflow.com/questions/55282615/golangci-lint-constant-explicit-type

abhimanyubabbar (approved):
Minor nit.

lvrach (approved):
thanks for addressing all the comments

PR #2657, opened 2022-11-07
chore: regulation worker avoid panic in case of timeout

saurav-malani:
# Description

- Modified regulation-worker to avoid panic if regulation-manager API call times out. Instead we are simply retrying after a minute & pushing the stats for it. 
- Added stats to get metrics like number of users deleted till now, deletion time, total loop time.
- Called from main with bugsnag to get notified in case of crash.

## Notion Ticket

https://www.notion.so/rudderstacks/regulation-manager-worker-fail-safe-with-metrics-alert-a6c1eee073924651a15e9caad45750d7

### DIFF
--- a/regulation-worker/internal/client/client.go
+++ b/regulation-worker/internal/client/client.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"regexp"
 	"strconv"
@@ -30,12 +31,9 @@ type JobAPI struct {
 // which is actually returned.
 func (j *JobAPI) Get(ctx context.Context) (model.Job, error) {
 	pkgLogger.Debugf("making http request to regulation manager to get new job")
-	ctx, cancel := context.WithTimeout(ctx, time.Minute)
-	defer cancel()
 
 	url := fmt.Sprintf("%s/dataplane/workspaces/%s/regulations/workerJobs", j.URLPrefix, j.WorkspaceID)
 	pkgLogger.Debugf("making GET request to URL: %v", url)
-
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
 	if err != nil {
 		pkgLogger.Errorf("error while create new http request: %v", err)
@@ -45,12 +43,14 @@ func (j *JobAPI) Get(ctx context.Context) (model.Job, error) {
 	req.SetBasicAuth(j.WorkspaceToken, "")
 	req.Header.Set("Content-Type", "application/json")
 
-	pkgLogger.Debugf("making request: %v", req)
 	resp, err := j.Client.Do(req)
-	if err != nil {
-		pkgLogger.Errorf("http request failed with error: %v", err)
+	if err, ok := err.(net.Error); ok && err.Timeout() {
+		stats.Default.NewStat("regulation_worker.manager_request_timeout_count", stats.CountType).Count(1)
+		return model.Job{}, model.ErrRequestTimeout
+	} else if err != nil {
 		return model.Job{}, err
 	}
+
 	defer func() {
 		err := resp.Body.Close()
 		if err != nil {
@@ -133,8 +133,10 @@ func (j *JobAPI) UpdateStatus(ctx context.Context, status model.JobStatus, jobID
 	req.Header.Set("Content-Type", "application/json")
 
 	resp, err := j.Client.Do(req)
-	if err != nil {
-		pkgLogger.Errorf("error while making http request: %v", err)
+	if err, ok := err.(net.Error); ok && err.Timeout() {
+		stats.Default.NewStat("regulation_worker.manager_request_timeout_count", stats.CountType).Count(1)
+		return model.ErrRequestTimeout
+	} else if err != nil {
 		return err
 	}
 	defer func() { _ = resp.Body.Close() }()
--- a/regulation-worker/internal/client/client.go
+++ b/regulation-worker/internal/client/client.go
@@ -45,7 +45,7 @@ func (j *JobAPI) Get(ctx context.Context) (model.Job, error) {
 
 	resp, err := j.Client.Do(req)
 	if err, ok := err.(net.Error); ok && err.Timeout() {
-		stats.Default.NewStat("regulation_worker.manager_request_timeout_count", stats.CountType).Count(1)
+		stats.Default.NewStat("regulation_manager.request_timeout", stats.CountType).Count(1)
 		return model.Job{}, model.ErrRequestTimeout
 	} else if err != nil {
 		return model.Job{}, err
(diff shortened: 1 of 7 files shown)


regulation-worker/internal/delete/api/api.go
```diff
@@ -62,26 +58,24 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 	fileCleaningTime.Start()
 	defer fileCleaningTime.End()
 
-	pkgLogger.Debugf("sending request: %v", req)
 	resp, err := api.Client.Do(req)
 	if err != nil {
-		pkgLogger.Errorf("error while making http request: %v", err)
+		if errors.Is(err, context.DeadlineExceeded) {
```
atzoum, line 63: use `os.IsTimeout(err)` and add a test scenario to verify

saurav-malani: I am not sure if using`os.IsTimeout(err)` would be a good idea. This is what the comment of the func says

> // IsTimeout returns a boolean indicating whether the error is known
> // to report that a timeout occurred.
> //
> // This function predates errors.Is, and the notion of whether an
> // error indicates a timeout can be ambiguous. For example, the Unix
> // error EWOULDBLOCK sometimes indicates a timeout and sometimes does not.
> // New code should use errors.Is with a value appropriate to the call
> // returning the error, such as os.ErrDeadlineExceeded.

saurav-malani: even they recommend using `errors.Is`

atzoum: then we may use `errors.Is(os.ErrDeadlineExceeded)` but definitely need a test scenario to verify.
If I remember correctly, the http client in Go doesn't return a `context.DeadlineExceeded` if an internal timeout happens.

fracasula: +1 for the test :+1:

saurav-malani: added.

regulation-worker/internal/client/client.go
```diff
@@ -133,8 +133,10 @@ func (j *JobAPI) UpdateStatus(ctx context.Context, status model.JobStatus, jobID
 	req.Header.Set("Content-Type", "application/json")
 
 	resp, err := j.Client.Do(req)
-	if err != nil {
-		pkgLogger.Errorf("error while making http request: %v", err)
+	if err, ok := err.(net.Error); ok && err.Timeout() {
```
atzoum, line 136:
```suggestion
	if os.isTimeout(err) {
```

regulation-worker/internal/client/client.go
```diff
@@ -45,12 +43,14 @@ func (j *JobAPI) Get(ctx context.Context) (model.Job, error) {
 	req.SetBasicAuth(j.WorkspaceToken, "")
 	req.Header.Set("Content-Type", "application/json")
 
-	pkgLogger.Debugf("making request: %v", req)
 	resp, err := j.Client.Do(req)
-	if err != nil {
-		pkgLogger.Errorf("http request failed with error: %v", err)
+	if err, ok := err.(net.Error); ok && err.Timeout() {
```
atzoum, line 47: `os.IsTimeout(err)` captures this and much more, including `context.DeadlineExceeded`
```suggestion
	if os.IsTimeout(err) {
```

regulation-worker/internal/delete/api/api.go
```diff
@@ -62,26 +58,24 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destConfig map
 	fileCleaningTime.Start()
 	defer fileCleaningTime.End()
 
-	pkgLogger.Debugf("sending request: %v", req)
 	resp, err := api.Client.Do(req)
 	if err != nil {
-		pkgLogger.Errorf("error while making http request: %v", err)
+		if err, ok := err.(net.Error); ok && err.Timeout() {
```
atzoum, line 63:
```suggestion
		if os.IsTimeout(err) {
```

regulation-worker/internal/service/looper.go
```diff
@@ -38,7 +40,15 @@ func (l *Looper) Loop(ctx context.Context) error {
 				pkgLogger.Debugf("context cancelled... exiting infinite loop %v", err)
 				return nil
 			}
-
+			continue
+		}
+		// this is to make sure that we don't panic when any of the API call fails with deadline exceeded error.
+		if err == model.ErrRequestTimeout {
+			pkgLogger.Errorf("context deadline exceeded... retrying after a minute %v", err)
```
atzoum, line 47:
```suggestion
			pkgLogger.Errorf("context deadline exceeded... retrying after %d minute(): %v", retryDelay, err)
```

atzoum:
@saurav-malani @chandumlg do we want this to be included in `1.3.0`?

saurav-malani:
alert for timeouts: https://github.com/rudderlabs/rudderstack-operator/pull/906

regulation-worker/internal/client/client.go
saurav-malani: done

regulation-worker/internal/client/client.go
saurav-malani: done

regulation-worker/internal/service/looper.go
```diff
@@ -27,6 +25,10 @@ func (l *Looper) Loop(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("reading value: %s from env: %s", "INTERVAL_IN_MINUTES", err.Error())
 	}
+	retryDelay, err := getenvInt("RETRY_DELAY_IN_MINUTES", 1)
```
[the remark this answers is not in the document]
saurav-malani: modified it to second.

regulation-worker/internal/service/looper.go
```diff
@@ -38,7 +40,16 @@ func (l *Looper) Loop(ctx context.Context) error {
 				pkgLogger.Debugf("context cancelled... exiting infinite loop %v", err)
 				return nil
 			}
-
+			continue
+		}
+		//this is to make sure that we don't panic when any of the API call fails with deadline exceeded error.
+		if errors.Is(err, context.DeadlineExceeded) {
```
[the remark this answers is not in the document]
saurav-malani: added test.

regulation-worker/internal/service/looper.go
saurav-malani: modified.

saurav-malani:
@atzoum if everything looks good, then it would be great to include it in release, otherwise we will have the use the image of the PR in regulation-worker deployment.

regulation-worker/internal/client/client_test.go
```diff
@@ -44,17 +46,31 @@ func TestGet(t *testing.T) {
...
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			svr := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-				w.WriteHeader(tt.respCode)
+				if tt.respCode != 0 {
+					w.WriteHeader(tt.respCode)
+				}
+				time.Sleep(time.Duration(tt.serverDelay) * time.Second)
```
fracasula, line 62: Alternatively I think you could set the client timeout to one nanosecond and do a 1ms sleep here to get a faster test.

Not major of course but I'm proposing this because I think that 1s for such a test it's still a long time and can be easily fixed.

fracasula (approved):

regulation-worker/internal/client/client_test.go
saurav-malani: makes sense. Modified.

atzoum (approved):

PR #2661, opened 2022-11-07
fix: preferences should be false incase of invalid bucket

BonapartePC:
# Description

Setting all backup preferences to false incase there is invalid bucket provided

## Notion Ticket

https://www.notion.so/rudderstacks/Allow-Customers-to-use-self-storage-26f45f75e38b4ef6bd872bea6187aa06

## Security

### DIFF
--- a/services/fileuploader/fileuploader.go
+++ b/services/fileuploader/fileuploader.go
@@ -85,12 +85,13 @@ func (p *provider) updateLoop(ctx context.Context, backendConfig backendconfig.B
 
 	settings := make(map[string]StorageSettings)
 
-	var bucket backendconfig.StorageBucket
-	var preferences backendconfig.StoragePreferences
-
 	for ev := range ch {
 		configs := ev.Data.(map[string]backendconfig.ConfigT)
 		for _, c := range configs {
+
+			var bucket backendconfig.StorageBucket
+			var preferences backendconfig.StoragePreferences
+
 			if c.Settings.DataRetention.UseSelfStorage {
 				settings := c.Settings.DataRetention.StorageBucket
 				defaultBucket := getDefaultBucket(ctx, settings.Type)
@@ -98,7 +99,10 @@ func (p *provider) updateLoop(ctx context.Context, backendConfig backendconfig.B
 			} else {
 				bucket = getDefaultBucket(ctx, config.GetString("JOBS_BACKUP_STORAGE_PROVIDER", "S3"))
 			}
-			preferences = c.Settings.DataRetention.StoragePreferences
+			// bucket type and configuration must not be empty
+			if bucket.Type != "" && len(bucket.Config) > 0 {
+				preferences = c.Settings.DataRetention.StoragePreferences
+			}
 			settings[c.WorkspaceID] = StorageSettings{
 				Bucket:      bucket,
 				Preferences: preferences,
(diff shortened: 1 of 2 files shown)


cisse21 (approved):

services/fileuploader/fileuploader.go
```diff
@@ -99,6 +99,10 @@ func (p *provider) updateLoop(ctx context.Context, backendConfig backendconfig.B
 				bucket = getDefaultBucket(ctx, config.GetString("JOBS_BACKUP_STORAGE_PROVIDER", "S3"))
 			}
 			preferences = c.Settings.DataRetention.StoragePreferences
+			// If bucket type is empty string or configuration is empty, clearing preferences so that no backup happens for that workspace.
+			if bucket.Type == "" || len(bucket.Config) == 0 {
+				preferences = backendconfig.StoragePreferences{}
+			}
```
atzoum, line 105: alternative
```suggestion
			var preferences backendconfig.StoragePreferences
			// bucket type and configuration must not be empty
			if bucket.Type != "" && len(bucket.Config) == 0 {
				preferences = c.Settings.DataRetention.StoragePreferences
			}
```

services/fileuploader/fileuploader.go
```diff
@@ -99,6 +99,10 @@ func (p *provider) updateLoop(ctx context.Context, backendConfig backendconfig.B
 				bucket = getDefaultBucket(ctx, config.GetString("JOBS_BACKUP_STORAGE_PROVIDER", "S3"))
 			}
 			preferences = c.Settings.DataRetention.StoragePreferences
+			// If bucket type is empty string or configuration is empty, clearing preferences so that no backup happens for that workspace.
+			if bucket.Type == "" || len(bucket.Config) == 0 {
+				preferences = backendconfig.StoragePreferences{}
```
fracasula, line 104: You're saying `preferences should be false` but `preferences` is not a `boolean`. What is the side effect of having an empty `StoragePreferences` down in the workflow?

fracasula: Shouldn't we not set the map key and delete it if present? So that `settings[c.WorkspaceID]` won't exist.

This way `GetStoragePreferences` would return the proper error:

```go
	if !ok {
		return prefs, fmt.Errorf("no storage settings found for workspace: %s", workspaceID)
	}
```

fracasula (changes requested):
Can you please add a test where an invalid bucket configuration is provided and then you call `GetStoragePreferences` with the right assertions?

services/fileuploader/fileuploader.go
BonapartePC: Made the change

BonapartePC:
> Can you please add a test where an invalid bucket configuration is provided and then you call `GetStoragePreferences` with the right assertions?

Added a test

services/fileuploader/fileuploader.go
BonapartePC: > You're saying `preferences should be false` but `preferences` is not a `boolean`. What is the side effect of having an empty `StoragePreferences` down in the workflow?

By preferences being false, I meant all the boolean inside are false. When we have StoragePreferences empty - all of the dumps will be set as false by default so we will not backup them

BonapartePC: > Shouldn't we not set the map key and delete it if present? So that `settings[c.WorkspaceID]` won't exist?
> 
> This way `GetStoragePreferences` would return the proper error:
> 
> ```go
> 	if !ok {
> 		return prefs, fmt.Errorf("no storage settings found for workspace: %s", workspaceID)
> 	}
> ```

We are not deleting them, we just assume that preferences by the customer is no backup (all set to false) if we are given an invalid bucket

fracasula (approved):

chandumlg (approved):

PR #2662, opened 2022-11-07
fix: gcs manager cancels context

BonapartePC:
# Description

GCS Manager creates an iterator to fetch objects. Bug: We were passing context to it and later cancelling at the end of the function call although the iterator will be used for every call of the function

## Notion Ticket

https://www.notion.so/rudderstacks/Fix-GCS-Replay-2dd4e0f0e97942a5bcce282a5f61ea77

## Security

### DIFF
--- a/services/filemanager/gcsmanager.go
+++ b/services/filemanager/gcsmanager.go
@@ -66,9 +66,6 @@ func (manager *GCSManager) ListFilesWithPrefix(ctx context.Context, startAfter, 
 		return
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, manager.getTimeout())
-	defer cancel()
-
 	// Create GCS Bucket handle
 	if manager.Config.Iterator == nil {
 		manager.Config.Iterator = client.Bucket(manager.Config.Bucket).Objects(ctx, &storage.Query{


chandumlg (approved):

cisse21 (approved):

services/filemanager/gcsmanager.go
```diff
@@ -66,9 +66,6 @@ func (manager *GCSManager) ListFilesWithPrefix(ctx context.Context, startAfter,
 		return
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, manager.getTimeout())
-	defer cancel()
-
 	// Create GCS Bucket handle
 	if manager.Config.Iterator == nil {
```
atzoum, line 70: @BonapartePC why are we caching the iterator? Is this thread safe??

BonapartePC: The way this function is being used is: With a given manager, calling subsequent ListFilesWithPrefix() should return us continuous items one after the other. This is the reason we cache it in the manager itself.

atzoum: This is not safe, we need to revisit. Can we guarantee that filemanager will be used exclusively by one goroutine and then never reused?

BonapartePC: Currently, all of the code is written keeping that in mind - https://github.com/rudderlabs/rudder-server/blob/master/services/filemanager/s3manager.go#L207

atzoum: this is a bad api design, I am confident we can do much better than that

PR #2663, opened 2022-11-07
fix: properly escape table name when querying for failed events

atzoum:
# Description

Using double quotes and escaping the table name when querying for failed events, until we drop support for the old endpoint.

[Reference](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#:~:text=Quoted%20identifiers%20can,limitation%20still%20applies)

## Security

### DIFF
--- a/router/failed-events-manager.go
+++ b/router/failed-events-manager.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/rudderlabs/rudder-server/utils/misc"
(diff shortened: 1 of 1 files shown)


fracasula (approved):
Maybe we could use a function for escaping these :thinking: Optional of course.

router/failed-events-manager.go
```diff
@@ -55,7 +56,7 @@ func (*FailedEventsManagerT) SaveFailedRecordIDs(taskRunIDFailedEventsMap map[st
 	}
 
 	for taskRunID, failedEvents := range taskRunIDFailedEventsMap {
-		table := fmt.Sprintf(`%s_%s`, failedKeysTablePrefix, taskRunID)
+		table := `"` + strings.ReplaceAll(fmt.Sprintf(`%s_%s`, failedKeysTablePrefix, taskRunID), `"`, `""`) + `"`
```
chrikar, line 59: I know this is probably a temporary measure. It would make more sense to perform this in `failedEventsHandler()` where both `SaveFailedRecordIDs()` and `FetchFailedRecordIDs()` are called.

chandumlg (approved):

router/failed-events-manager.go
chandumlg: @chrikar The goal of this pr is to evade a security risk with respect to /v1/failed-events endpoint. `SaveFailedRecordIDs` function is not exposed outside. So, no risk.

atzoum: the HTTP handler should not know how/where failed events are stored, let alone that jobRunId needs to be escaped as if it was a postgresql identifier in the first place :)

chandumlg: Oh! Apologies. I misread the changes in this pr. I thought the escaping is done on exposed functions `FetchFailedRecordIDs` & `DropFailedRecordIDs` and felt it is not necessary to modify `SaveFailedRecordIDs`.

PR #2667, opened 2022-11-08
fix: use enterprise token from env as fallback

chandumlg:
# Description

Falling back to enterprise token read from env (if not provided as an ldflag)

## Security

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -133,9 +133,6 @@ func (r *Runner) Run(ctx context.Context, args []string) int {
 		return 0
 	}
 
-	if r.releaseInfo.EnterpriseToken == "" {
-		r.releaseInfo.EnterpriseToken = os.Getenv("ENTERPRISE_TOKEN")
-	}
 	options.EnterpriseToken = r.releaseInfo.EnterpriseToken
 
 	r.application = app.New(options)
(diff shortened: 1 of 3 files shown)


runner/runner.go
```diff
@@ -133,6 +133,9 @@ func (r *Runner) Run(ctx context.Context, args []string) int {
 		return 0
 	}
 
+	if r.releaseInfo.EnterpriseToken == "" {
```
atzoum, line 136: shouldn't we do this change it in main.go#31?
```
EnterpriseToken: config.GetString("ENTERPRISE_TOKEN", enterpriseToken),
```

chandumlg: We can do it in main too.

But we need os.Getenv and not config.GetString (we don't want RSERVER_ prefix).

atzoum: don't worry about the RSERVER_ prefix, it looks up the env without it too ;)

atzoum:
@chandumlg  is this change needed for `1.3.0`? If so, please change the target to `release/1.3.x branch`

runner/runner.go
chandumlg: Oh kool. Pushing the change.

chandumlg: Done

atzoum (approved):

cisse21 (approved):

PR #2668, opened 2022-11-08
fix: removing the backup bucket env check to determine if backup is enabled

BonapartePC:
# Description

Removing the backup bucket env check to determine if backup is enabled since bucket can be provided by customer as well

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/jobsdb/backup.go
+++ b/jobsdb/backup.go
@@ -29,7 +29,7 @@ type backupSettings struct {
 }
 
 func (b *backupSettings) isBackupEnabled() bool {
-	return masterBackupEnabled && b.instanceBackupEnabled && config.GetString("JOBS_BACKUP_BUCKET", "") != ""
+	return masterBackupEnabled && b.instanceBackupEnabled
 }
 
 func IsMasterBackupEnabled() bool {


atzoum:
@BonapartePC is this change needed for `1.3.0`? If so, please change the target to `release/1.3.x branch`

BonapartePC:
> @BonapartePC is this change needed for `1.3.0`? If so, please change the target to `release/1.3.x branch`

Made the change

jobsdb/backup.go
```diff
@@ -29,7 +29,7 @@ type backupSettings struct {
 }
 
 func (b *backupSettings) isBackupEnabled() bool {
-	return masterBackupEnabled && b.instanceBackupEnabled && config.GetString("JOBS_BACKUP_BUCKET", "") != ""
```
atzoum, line 32: don't we need a default (fallback) backup bucket anyway?

BonapartePC: True, but checking if `JOBS_BACKUP_BUCKET` env is set to determine if the backup is enabled does not seem right since we can take bucket details from the user as well right?

atzoum: Then jobsdb needs to consult another component for that?

BonapartePC: No, jobsdb only cares if master backup is enabled and instance backup is enabled and it's the job of fileuploader to provide bucket details to it right?

atzoum: > we can take bucket details from the user as well right?
But still, don't  we need to have a default bucket properly setup in a multitenant deployment, regardless of workspace-specific configurations?

BonapartePC: True, that assertion should happen independently of this right? Probably somewhere even before

saurav-malani: I agree with @atzoum. I think it is safer to keep the env check to ensure a default bucket is set for fallback, until we have a better way to do it.

atzoum (approved):

chandumlg (approved):

PR #2675, opened 2022-11-09
fix(destination): empty oauth account check

sanpj2292:
# Description

The oauth account id can be empty. Due to the changes we might do as part of the regulation-worker support for Universal Analytics.

This check serves as protection when `rudderAccountId` is missing. 

### Background:
As part of the changes for supporting OAuth in regulation-worker module. We need to have one more field in destination config known as `rudderUserDeleteAccountId` which indicates that this account information will be used for sending delete user request to destination

This user delete feature can sometimes only require OAuth and event-delivery for the same destination might not need OAuth, int this case, we don’t populate `rudderAccountId` but we populate `rudderUserDeleteAccountId` with the access token information(which will be used for user-deletion requests)

This can cause a problem while processing the transformer proxy response for OAuth destinations. We would need to add a check to prevent any errors for such things.

## Notion Ticket

https://www.notion.so/rudderstacks/Integrate-OAuth-in-regulation-worker-module-of-server-dae0cf6ef89e41eb91e9212ac524cf49

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -2091,42 +2091,44 @@ func (rt *HandleT) HandleOAuthDestResponse(params *HandleDestOAuthRespParamsT) (
 		// Check the category
 		// Trigger the refresh endpoint/disable endpoint
 		rudderAccountID := routerutils.GetRudderAccountId(&destinationJob.Destination)
-		switch destErrOutput.AuthErrorCategory {
-		case oauth.DISABLE_DEST:
-			return rt.ExecDisableDestination(&destinationJob.Destination, workspaceID, trRespBody, rudderAccountID)
-		case oauth.REFRESH_TOKEN:
-			var refSecret *oauth.AuthResponse
-			refTokenParams := &oauth.RefreshTokenParams{
-				Secret:          params.secret,
-				WorkspaceId:     workspaceID,
-				AccountId:       rudderAccountID,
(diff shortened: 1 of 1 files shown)


router/router.go
```diff
@@ -2091,42 +2091,44 @@ func (rt *HandleT) HandleOAuthDestResponse(params *HandleDestOAuthRespParamsT) (
...
-					"accountId":     refTokenParams.AccountId,
-					"destType":      refTokenParams.DestDefName,
-				}).Increment()
-				rt.logger.Errorf(`[OAuth request] Aborting the event as %v`, oauth.INVALID_REFRESH_TOKEN_GRANT)
-				return disableStCd, refSec.Err
-			}
-			// Error while refreshing the token or Has an error while refreshing or sending empty access token
-			if errCatStatusCode != http.StatusOK || routerutils.IsNotEmptyString(refSec.Err) {
-				return http.StatusTooManyRequests, refSec.Err
+		if strings.TrimSpace(rudderAccountID) != "" {
```
atzoum, line 2094: there is a lot of if-nesting happening here, wouldn't it better to return early, e.g.
```
if strings.TrimSpace(rudderAccountID) == "" {
	return trRespStatusCode, trRespBody
}
```

sanpj2292: Done. Can you check and confirm ?

atzoum (approved):

chandumlg (approved):

PR #2685, opened 2022-11-10
chore(warehouse): integration test improvements

achettyiitr:
# Description

### 1. Improvements
- Use [**table-driven tests**](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) for better maintenance and readability for the tests module.

 ### 2. Module-level dependency
- **[Problem]** **Restricts all tests** which are part of the module to be dependent. We can't be able to have independent Unit tests.
- **[Solution]** Moved the corresponding setup/teardown to their corresponding Integration Test.

### 3. Sources (Async Job Framework)
- **[Problem]** Currently, `sources tests` are not running. Although, there was a plan to make it independent from the warehouse tests because of the time consumption for running the tests.
- **[Solution]** **Addition sources async module changes as part of Integration tests.** 
- **[Problem]**  Since the async job framework is part of the warehouse, we should run the corresponding tests.
- **[Solution]** **Using separate namespace** to run these tests parallel and to reduce the overall test time if possible.

### 4. Test Files Exclusion
- **[Problem]** Makes it **difficult** to work with. 
- **[Solution]** Removed `go: build warehouse_integration`
- **[Problem]** **Build Tags Maintenance**.
- **[Solution]** Removed `-tags=warehouse_integration` and using **SLOW=1** env to handle it.

### 5. Unit vs Integration Tests
- For **running Unit tests**, use **SLOW=0** flag to **ignore Integration tests**.
- For **running Integration tests**, use **TestIntegration** or **TestConfigurationValidation** prefix.

## Notion Ticket

https://www.notion.so/rudderstacks/Test-cleanup-for-Datalakes-Integration-e6f8e3fd7b024b22aff23427d3e7d2c6

## Security


warehouse/datalake/datalake_test.go
```diff
@@ -14,145 +14,100 @@ import (
...
+	}{
+		{
+			name:          "S3Datalake",
+			writeKey:      "ZapZJHfSxUN96GTIuShnz6bv0zi",
+			sourceID:      "279L3gEKqwruNoKGsXZtSVX7vIy",
+			destinationID: "27SthahyhhqEZ7H4T4NTtNPl06V",
+			provider:      warehouseutils.S3_DATALAKE,
+			prerequisite: func(t *testing.T) {
+				testhelper.CreateBucketForMinio(t, "s3-datalake-test")
+			},
```
lvrach, line 40: ❤️

warehouse/datalake/datalake_test.go
```diff
@@ -246,10 +201,5 @@ func TestDatalakeConfigurationValidation(t *testing.T) {
 }
 
 func TestMain(m *testing.M) {
-	handle = &TestHandle{
-		S3DatalakeWriteKey:    "ZapZJHfSxUN96GTIuShnz6bv0zi",
-		GCSDatalakeWriteKey:   "9zZFfcRqr2LpwerxICilhQmMybn",
-		AZUREDatalakeWriteKey: "Hf4GTz4OiufmUqR1cq6KIeguOdC",
-	}
-	os.Exit(testhelper.Run(m, handle))
+	os.Exit(testhelper.Run(m, &TestHandle{}))
```
lvrach, line 204: We can probably get rid of `TestMain`

lvrach (approved):
Nice improvement; let's see if we can remove `TestMain` as well

warehouse/datalake/datalake_test.go
achettyiitr: @lvrach Removed module-level test dependency.

warehouse/bigquery/bigquery_source_test.go
```diff
@@ -4,162 +4,82 @@ package bigquery_test
...
-	})
-	if err != nil {
-		return fmt.Errorf("error while running test connection for bigquery with err: %s", err.Error())
-	}
-	return nil
-}
+	db, err := bigquery2.Connect(context.TODO(), &credentials)
+	require.NoError(t, err)
+
+	schema := testhelper.Schema(warehouseutils.BQ, testhelper.BigqueryIntegrationTestSchema)
```
lvrach, line 36: This schema is status right ? What if two tests suites run in parallel? Can we use a parallel schema every time ?

.github/workflows/tests.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
           username: rudderlabs
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Warehouse Integration Test
-        run: make run-warehouse-integration
+        run: make run-warehouse-integration tags=warehouse_integration
```
lvrach, line 57: This is out of the scope of the PR, but we should consider removing tags for tests.

lvrach (commented):
I like the overall changes, but I am bit concerned regarding the introduction of globals. Setting environmental variables or using global can make test flaky when running in parallel. 

Ensure that setting a global doesn't affect another tests that would run in parallel.

Run tests a couple of times to ensure they are not flaky (they should always passing)

warehouse/bigquery/bigquery_source_test.go
achettyiitr: We can't use parallel schema.

achettyiitr: I am having the assumption that there will only be one Integration test per destination.

Makefile
```diff
@@ -87,36 +87,30 @@ proto: install-tools ## Generate protobuf files
...
+	SNOWFLAKE_INTEGRATION_TEST_SCHEMA=$(call generate_namespace, swh) \
+	DATABRICKS_INTEGRATION_TEST_SCHEMA=$(call generate_namespace, dwh) \
 	docker-compose -f warehouse/docker-compose.test.yml up --build start_warehouse_integration
 
 logs-warehouse-integration:
 	docker logs wh-backend
 
 run-warehouse-integration: setup-warehouse-integration
-	if docker-compose -f warehouse/docker-compose.test.yml exec -T wh-backend go test -v ./warehouse/... -tags=warehouse_integration -p 8 -timeout 30m -count 1; then \
+	$(eval TEST_PATTERN = '^Test.*Integration$$$$$$$$' '^Test.*ConfigurationValidation$$$$$$$$')
```
lvrach, line 105: Do we need multiple `$` here? 

Also, should Integration be a prefix or postfix:

`TestIntegratioBigQuery` vs `TestBigQueryIntegration`

The second is more readable, but the first more convenient due to the hierarchy

achettyiitr:
> https://www.notion.so/rudderstacks/Test-cleanup-for-Datalakes-Integration-e6f8e3fd7b024b22aff23427d3e7d2c6

Added the fix [here](https://github.com/rudderlabs/rudder-server/pull/2731).

Makefile
achettyiitr: > Do we need multiple $ here?

Actually, we are using it in 2 eval expressions and then we are running the command.  Basically, we need to escape it 3 times.
2^3 = 8

achettyiitr: Yes, in terms of hierarchy `TestIntegratioBigQuery` is the best choice.

achettyiitr:
> there is one deep source issue https://deepsource.io/gh/rudderlabs/rudder-server/run/2284a648-4dd4-41a9-8dc3-c3d87c9e1fff/go/GO-W1024?listindex=0

This we added in here[https://github.com/rudderlabs/rudder-server/pull/2731]

warehouse/testhelper/setup.go
```diff
@@ -106,57 +72,120 @@ const (
...
+	verifyEventsInTableUploads(t, w)
+	if !w.SkipWarehouse {
+		verifyEventsInWareHouse(t, w)
+	}
+	verifyWorkspaceIDInStats(t, w.StatsToVerify...)
 }
 
-func initJobsDB() {
-	jobsDB = setUpJobsDB()
+func (w *WareHouseTest) TestScenarioTwo(t testing.TB) {
```
achettyiitr, line 160: Better name: `TestScenarioTwo`

lvrach (approved):

abhimanyubabbar (approved):

achettyiitr:
> Can't we use pq.QuoteIdentifier here ?

Yes Will add this in a subsequent PR?

PR #2688, opened 2022-11-10
chore: compact status tables before proceeding with migration

Sidddddarth:
# Description

Cleaning up the status table(delete all but the latest status per job) before proceeding with migration.
The DSs this is done to, are selected via analyzer estimates(`pg_class.reltuples`) and based on existing thresholds.
This is once again repeated before the tables are backup, just to ensure that there's only one status per job in the uploaded file.
Moved migration related logic to `jobsdb/migration.go`.

## Notion Ticket

[Perform compaction on the rightmost dataset during migration if necessary](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84)

## Security


jobsdb/jobsdb_backup_test.go
```diff
@@ -163,7 +163,7 @@ func TestBackupTable(t *testing.T) {
 	goldenStatusFile, err := tc.readGzipStatusFile(goldenFileStatusFileName)
 	require.NoError(t, err, "expected no error while reading golden status file")
 	verifyStatus(t, backedupStatus, goldenStatusFile)
-	require.Equal(t, len(goldenStatusFile)*2, len(backedupStatus), "expected status files to be same")
+	require.Equal(t, len(goldenStatusFile), len(backedupStatus), "expected status files to be same")
```
chandumlg, line 166: Don't we need a test for `clean up done before back up` scenario?

Sidddddarth: This test verifies the same.
The only change in behaviour is that two statuses per job would get uploaded earlier, whereas now only one is. These changes reflect that.

jobsdb/migration.go
```diff
@@ -0,0 +1,472 @@
...
+			); err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+}
+
+// cleanStatusTable deletes all rows except for the latest status for each job
+func (*HandleT) cleanStatusTable(ctx context.Context, tx *Tx, table string) error {
```
atzoum, line 229: let's also run an `ANALYZE` on the status table after the delete, since we will be deleting a lot of record.

jobsdb/migration.go
```diff
@@ -0,0 +1,479 @@
...
+	)
+
+	var numJobsMigrated int64
+	if err := tx.QueryRowContext(
+		ctx,
+		compactDSQuery,
+	).Scan(&numJobsMigrated); err != nil {
+		return 0, err
+	}
+	if _, err := tx.Exec(fmt.Sprintf(`ANALYZE %q, %q`, destDS.JobTable, destDS.JobStatusTable)); err != nil {
```
chandumlg, line 340: If this returns error, there will be a problem.
1. We would end up with gw_1 & gw_1_1 tables
2. There will be an entry in journal table with done as FALSE. This will cause some other problems.

Should we ignore this error? @Sidddddarth @atzoum

chandumlg (changes requested):
A comment posted. Please take a look.

chandumlg (approved):

jobsdb/migration.go
Sidddddarth: `WithTx` wrapper helps us avoid this situation.
All the suggested updates would've been rolled back.

PR #2691, opened 2022-11-10
refactor: introduce api & repo components

lvrach:
# Description

## Motivation
While preparing the degradation per workspace id functionality, it has hard to test particular components due to tight coupling. Not having the ability to do dependency injection and the usage of globals would lead to lousy test, making refactoring a prerequisite for adding tests.

Furthermore, as a team, we have recognised the importance of decoupling components and introducing a dedicated layer for accessing the database.

Finally, many database queries were not using prepared statements, opening the possibility of dependency injection.

In this PR, I tried to address the problems above but in a limited scope. The refactoring affected only the `/v1/processing` endpoint and the staging file repository. This incremental approach has been used for two reasons:
* Being an experiment and a concrete example to discuss regarding the design changes we want as a team.
* Keep the number of changes as low as possible to make the review more manageable and reduce the risk of introducing regressions.

## Proposed design

Read more about the [proposed design in notion](https://www.notion.so/rudderstacks/Decoupling-warehouse-c619dc5282c8414cac33ec5c9866055d)

## Notion Ticket

https://www.notion.so/rudderstacks/Test-refactor-warehouse-API-5f55d977215a44d4b8c9dfc94951b32e

## Security


warehouse/internal/repo/staging.go
```diff
@@ -0,0 +1,268 @@
...
+	SourceJobID           string `json:"source_job_id"`
+	SourceJobRunID        string `json:"source_job_run_id"`
+	TimeWindowYear        int    `json:"time_window_year"`
+	TimeWindowMonth       int    `json:"time_window_month"`
+	TimeWindowDay         int    `json:"time_window_day"`
+	TimeWindowHour        int    `json:"time_window_hour"`
+	DestinationRevisionID string `json:"destination_revision_id"`
+}
+
+func metadataFromStagingFile(stagingFile warehouse.StagingFileT) metadataSchema {
```
achettyiitr, line 57:
```suggestion
func metadataFromStagingFile(stagingFile *warehouse.StagingFileT) metadataSchema {
```
Should we use something like this?

warehouse/internal/api/http.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	if payload.Location == "" {
+		return warehouse.StagingFileT{}, fmt.Errorf("location is required")
+	}
+
+	if payload.BatchDestination.Source.ID == "" {
+		return warehouse.StagingFileT{}, fmt.Errorf("batchDestination.source.id is required")
+	}
+	if payload.BatchDestination.Destination.ID == "" {
```
achettyiitr, line 68: `Destination` can probably can nil as well. Should we validate that as well?

warehouse/internal/api/http.go
```diff
@@ -0,0 +1,152 @@
...
+	SourceBatchID   string
+	SourceTaskID    string
+	SourceTaskRunID string
+	SourceJobID     string
+	SourceJobRunID  string
+	TimeWindow      time.Time
+}
+
+func mapStagingFile(payload stagingFileSchema) (warehouse.StagingFileT, error) {
+	if payload.WorkspaceID == "" {
```
achettyiitr, line 57: @lvrach 
Any benefit to using [validators](https://pkg.go.dev/github.com/go-playground/validator/v10) for this?

achettyiitr (changes requested):

warehouse/internal/api/http_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	"github.com/rudderlabs/rudder-server/services/stats"
+	"github.com/rudderlabs/rudder-server/utils/logger"
+	"github.com/rudderlabs/rudder-server/warehouse"
+	"github.com/rudderlabs/rudder-server/warehouse/internal/api"
+	"github.com/rudderlabs/rudder-server/warehouse/multitenant"
+	"github.com/stretchr/testify/require"
+)
+
+type memRepo struct {
```
achettyiitr, line 23: This is very nice.

warehouse/internal/api/http.go
lvrach: I think, eventually we need to introduce them, I find them overkill for the checks we are doing so far.

warehouse/internal/api/http.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/rudderlabs/rudder-server/utils/logger"
+	"github.com/rudderlabs/rudder-server/utils/misc"
+	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
+	"github.com/rudderlabs/rudder-server/warehouse/multitenant"
+)
+
+var json = jsoniter.ConfigCompatibleWithStandardLibrary
+
+type stagingFilesRepo interface {
+	Insert(ctx context.Context, stagingFile model.StagingFile) (int64, error)
```
achettyiitr, line 23:
```suggestion
	insert(ctx context.Context, stagingFile model.StagingFile) (int64, error)
```

warehouse/internal/repo/staging.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+
+		m.SetStagingFile(&stagingFile)
+		stagingFiles = append(stagingFiles, stagingFile)
+	}
+
+	return stagingFiles, nil
+}
+
+// GetInRange returns staging file with the given ID.
```
achettyiitr, line 226:
```suggestion
// GetByID returns staging file with the given ID.
```

warehouse/warehouse.go
```diff
@@ -476,93 +412,25 @@ func (wh *HandleT) getPendingStagingFiles(warehouse warehouseutils.Warehouse) ([
...
-		WHERE
-		  ST.id > %[2]v
-		  AND ST.source_id = '%[3]s'
-		  AND ST.destination_id = '%[4]s'
-		ORDER BY
-		  id ASC;
-`,
-		warehouseutils.WarehouseStagingFilesTable,
-		lastStagingFileID,
+	stagingFilesList, err := wh.stagingRepo.GetAfterID(
```
achettyiitr, line 415: Should we consider returning `[]*model.StagingFile` as we are doing pointer conversion later?

warehouse/internal/repo/staging.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// Insert inserts a staging file into the staging files table.
+//
+// NOTE: The following fields are ignored and set by the database:
+// - ID
+// - Error
+// - CreatedAt
+// - UpdatedAt
+func (repo *StagingFiles) Insert(ctx context.Context, stagingFile model.StagingFile) (int64, error) {
```
achettyiitr, line 99:
```suggestion
func (repo *StagingFiles) Insert(ctx context.Context, stagingFile *model.StagingFile) (int64, error) {
```

achettyiitr (changes requested):

warehouse/warehouse.go
achettyiitr: @lvrach Any benefits of this?

achettyiitr (approved):

warehouse/internal/repo/staging.go
```diff
@@ -0,0 +1,288 @@
...
+	).Scan(&id)
+	if err != nil {
+		return id, fmt.Errorf("inserting staging file: %w", err)
+	}
+
+	return id, nil
+}
+
+// praseRow is a helper for mapping a row of tableColumns to a model.StagingFile.
+func (*StagingFiles) parseRows(rows *sql.Rows) ([]model.StagingFile, error) {
```
achettyiitr, line 170: We are not calling rows.close() anywhere?

achettyiitr (changes requested):

warehouse/internal/repo/staging.go
achettyiitr:
```suggestion
func (*StagingFiles) parseRows(rows *sql.Rows) ([]model.StagingFile, error) {
	var (
		stagingFiles []model.StagingFile
		err          error
	)

	defer func() {
		closeErr := rows.Close()
		if err == nil {
			if closeErr != nil {
				fmt.Println("closing rows: ", closeErr)
			}
		}
		err = closeErr
	}()
```

achettyiitr (changes requested):

warehouse/internal/repo/staging.go
achettyiitr:
```suggestion
func (*StagingFiles) parseRows(rows *sql.Rows) ([]model.StagingFile, error) {
	var (
		stagingFiles []model.StagingFile
		err          error
	)

	defer func() {
		closeErr := rows.Close()
		if err == nil {
			if closeErr != nil {
				fmt.Println("closing rows: ", closeErr)
			}
		}
		err = closeErr
	}()
```

achettyiitr: I have seen that in most of the places in our code base, we are not handling close errors. Maybe it's a good time to handle it properly.

achettyiitr: https://github.com/teivah/100-go-mistakes/blob/4fdb88c13ed09f626863d40e9532b39628f79889/07-error-management/54-defer-errors/main.go#L32

warehouse/internal/repo/staging.go
```diff
@@ -0,0 +1,288 @@
...
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/rudderlabs/rudder-server/utils/timeutil"
+	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
+	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+)
+
+const tableName = warehouseutils.WarehouseStagingFilesTable
```
achettyiitr, line 17: We can't use generic constants here, right?
It will collide with other tables later right?

achettyiitr (changes requested):

achettyiitr:
Can we use [Sentinel error](https://www.joeshaw.org/error-handling-in-go-http-applications/) for things like this?
1. can't unmarshal body
2. invalid payload:
3. Workspace is degraded

achettyiitr:
I was going through the `http` package and found that `http.Error` also [logs](https://github.com/golang/go/blob/cf93b25366aa418dea3eea49a7b85447631c2a1d/src/net/http/server.go#L2135) the error.
Since we are also logging the error before calling `http.Error`. Is there a way that it can be handled by the http package? Or is it not required to have it?

achettyiitr:
> 

Please ignore.

warehouse/internal/repo/staging.go
```diff
@@ -0,0 +1,288 @@
...
+	stagingFile.DestinationRevisionID = m.DestinationRevisionID
+}
+
+func (repo *StagingFiles) init() {
+	repo.once.Do(func() {
+		if repo.Now == nil {
+			repo.Now = timeutil.Now
+		}
+	})
+}
```
abhimanyubabbar, line 91: can you explain the rational behind this special function ?
Why not assign the function to `repo.Now` at time of its creation ?

warehouse/internal/repo/staging.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+	now := repo.Now()
+
+	schemaPayload, err := json.Marshal(stagingFile.Schema)
+	if err != nil {
+		return id, fmt.Errorf("marshaling schema: %w", err)
+	}
+
+	err = repo.DB.QueryRowContext(ctx,
+		`INSERT INTO `+tableName+` (
```
abhimanyubabbar, line 131: should we move tablename at repo itself due to close association between the two ?

abhimanyubabbar (approved):
Overall looks good and clean. minor comments.

achettyiitr:
> I was going through the `http` package and found that `http.Error` also [logs](https://github.com/golang/go/blob/cf93b25366aa418dea3eea49a7b85447631c2a1d/src/net/http/server.go#L2135) the error. Since we are also logging the error before calling `http.Error`. Is there a way that it can be handled by the http package? Or is it not required to have it?
> 
> Wanted your thoughts on this?

Please ignore this.

warehouse/internal/repo/staging.go
lvrach: The idea behind this is that we can mock time for testing purposes.

We can set `Now` during tests to return a static/predictable time. If Now is not set, it defaults to `timeutil.Now`.

The overall idea behind `init` is lazy initialisation. It is a helpful pattern used from many components in the standard library. We are using the repo struct itself for configuration. Construction implicitly happens when the first method is called.

warehouse/internal/repo/staging.go
lvrach: Do you mean adding it as part of the struct instead of a const at the repo level?

warehouse/internal/repo/staging.go
lvrach: @achettyiitr, you are right and I was missing error handling as well.

After some thought, I think this is a simpler equivalent would be:

```go
	defer func() { _ = rows.Close() }()

	for rows.Next() {
		var (
			stagingFile model.StagingFile
			metadataRaw []byte

			firstEventAt, lastEventAt sql.NullTime

			errorRaw sql.NullString
		)
		err := rows.Scan(
			&stagingFile.ID,
			&stagingFile.Location,
			&stagingFile.SourceID,
			&stagingFile.DestinationID,
			&stagingFile.Schema,
			&errorRaw,
			&stagingFile.Status,
			&firstEventAt,
			&lastEventAt,
			&stagingFile.TotalEvents,
			&stagingFile.CreatedAt,
			&stagingFile.UpdatedAt,
			&metadataRaw,
			&stagingFile.WorkspaceID,
		)
		if err != nil {
			return nil, fmt.Errorf("scanning row: %w", err)
		}

		stagingFile.CreatedAt = stagingFile.CreatedAt.UTC()
		stagingFile.UpdatedAt = stagingFile.UpdatedAt.UTC()

		if firstEventAt.Valid {
			stagingFile.FirstEventAt = firstEventAt.Time.UTC()
		}
		if lastEventAt.Valid {
			stagingFile.LastEventAt = lastEventAt.Time.UTC()
		}

		if errorRaw.Valid {
			stagingFile.Error = errors.New(errorRaw.String)
		}

		var m metadataSchema
		err = json.Unmarshal(metadataRaw, &m)
		if err != nil {
			return nil, fmt.Errorf("unmarshal metadata: %w", err)
		}

		m.SetStagingFile(&stagingFile)
		stagingFiles = append(stagingFiles, stagingFile)
	}

	if err := rows.Err(); err != nil {
		return nil, fmt.Errorf("iterating rows: %w", err)
	}
```

Reason, `.Close` will be effective only if we interrupt the iteration. 

If `.Next()` returns false it will close everything automatically; in this case, we only need to check for an error.

The only case we need to close is when scan or unmarshal fails. In that case, we would already have an error. In the example of the book, if there is already an error is not getting overwritten by a close error. I think this is the correct approach; close error is less significant than the others and combining the two errors might lead to confusion.

achettyiitr (approved):

warehouse/internal/repo/staging.go
achettyiitr: Thanks, @lvrach for sharing thoughts on this.

achettyiitr (approved):

PR #2703, opened 2022-11-15
chore: data residency adaptations

atzoum:
# Description

With data residency, multiple data planes can exist for the same namespace/workspace in different regions. Each data plane will operate against a region-specific configuration and warehouse nodes need to advertise their region to cp-router, so that control plane can perform cp-router requests targeting a specific region.

According to the above, the following `rudder-server` adaptations are required:
- If a region is set in the environment's configuration, include it as a `region` query parameter when requesting the configuration from config backend.
- If a region is set in the environment's configuration, include it as a `region` label in cp-router's authentication response.

## Relevant work

[#70](https://github.com/rudderlabs/cp-router/pull/70) adds support for labels and label-based selection of connections in `cp-router`.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=7b092def0d6f43028849cb58c1168285&pm=s)

## Security


achettyiitr (approved):

cisse21 (approved):

config/backend-config/backend-config.go
```diff
@@ -303,11 +305,12 @@ func newForDeployment(deploymentType deployment.Type, configEnvHandler types.Con
 // Setup backend config
 func Setup(configEnvHandler types.ConfigEnvI) (err error) {
 	deploymentType, err := deployment.GetFromEnv()
+	region := config.GetString("region", "")
```
chandumlg, line 308: It would be good, if this line is below err check line.

PR #2716, opened 2022-11-21
feat: redshift temporary tables and other integration improvements

achettyiitr:
# Description

### Use Temporary tables 

- This will help us not to run cleanups like `DropStagingTables` and `DropDanglingStagingTables`.

### Use DI to expose the Redshift Handle

- This will help us more towards testing.

### Use Config per Sync level

- Earlier the issue is since these configs are hot reloadable, they can have an effect only on the partially synced tables (tables that are yet to be processed after the config is updated). 
- So that any config change should affect the entire sync as a whole.

### Improvements

- Formatted a couple of Queries.
- Some code cleanup

## Notion Ticket

https://www.notion.so/rudderstacks/Redshift-temporary-tables-implementation-f76674dc8a794da9bedb11ca0b81100f

## Security


warehouse/redshift/redshift.go
```diff
@@ -21,35 +21,43 @@ import (
...
-var (
+type Config struct {
 	setVarCharMax                 bool
 	dedupWindow                   bool
 	dedupWindowInHours            time.Duration
-	pkgLogger                     logger.Logger
 	skipComputingUserLatestTraits bool
 	enableDeleteByJobs            bool
-)
+}
```
lvrach, line 30: Instead of defining a new config struct, why not use the existing `Handler` struct to store those things?

warehouse/redshift/redshift.go
```diff
@@ -21,35 +21,43 @@ import (
...
+			enableDeleteByJobs:            config.GetBool("Warehouse.redshift.enableDeleteByJobs", false),
+			dedupWindowInHours:            config.GetDuration("Warehouse.redshift.dedupWindow", 720, time.Hour),
+		},
+		Logger: pkgLogger,
+	}
 }
 
-type HandleT struct {
+type Handle struct {
 	Db             *sql.DB
```
lvrach, line 54: nit-picking
```suggestion
	DB             *sql.DB
```

warehouse/redshift/redshift.go
lvrach: Also, we would probably need to expose those fields so we can set them on our test suite

achettyiitr: I thought injecting a mock Config would be easy while testing.

achettyiitr: But we can add this in Handler as well.

warehouse/redshift/redshift.go
```diff
@@ -21,35 +21,43 @@ import (
...
+	return Handle{
+		Config: Config{
+			setVarCharMax:                 config.GetBool("Warehouse.redshift.setVarCharMax", false),
+			dedupWindow:                   config.GetBool("Warehouse.redshift.dedupWindow", false),
+			skipComputingUserLatestTraits: config.GetBool("Warehouse.redshift.skipComputingUserLatestTraits", false),
+			enableDeleteByJobs:            config.GetBool("Warehouse.redshift.enableDeleteByJobs", false),
+			dedupWindowInHours:            config.GetDuration("Warehouse.redshift.dedupWindow", 720, time.Hour),
+		},
+		Logger: pkgLogger,
+	}
```
lvrach, line 50: Our constructor is coupled with the config package.

I would suggest something like:
```suggestion
func WithConfig(h *Handle, config.Config) {
			h.SetVarCharMax =              config.GetBool("Warehouse.redshift.setVarCharMax", false)
			h.DedupWindow = config.GetBool("Warehouse.redshift.dedupWindow", false)                  
			h.SkipComputingUserLatestTraits = config.GetBool("Warehouse.redshift.skipComputingUserLatestTraits", false),
			h.EnableDeleteByJobs =          config.GetBool("Warehouse.redshift.enableDeleteByJobs", false),
			h.DedupWindowInHours =            config.GetDuration("Warehouse.redshift.dedupWindow", 720, time.Hour),
}
```
Usage:
```go
h := redshift.Handle{
   Logger: pkgLogger,
}
redshift.WithConfig(h, config.DefaultConfig)
```

achettyiitr: What do you think of adding WithConfig in the receiver itself. @lvrach

lvrach:
Two questions based on the docs:
>Keyword that creates a temporary table that is visible only within the current session. The table is automatically dropped at the end of the session in which it is created. The temporary table can have the same name as a permanent table.

What is considered a session by redshift? If every connection has its own session, is it possible that due to connection pooling, we create the temp in one connection and we try to query it in a different one? 

> By default, users have permission to create temporary tables by their automatic membership in the PUBLIC group. To deny this privilege to a user, revoke the TEMP privilege from the PUBLIC group, and then explicitly grant the TEMP privilege only to specific users or groups of users.

Should we try to create a temporary table in our verification step since it has different privileges than normal ones? Although I find it unlikely someone would disable them.

achettyiitr:
> 
> Should we try to create a temporary table in our verification step since it has different privileges than normal ones? Although I find it unlikely someone would disable them.

Yes, this concern I also had. Since we added `rudder-cli wh-test -d 'destinationID'` I planned to validated first that the current ongoing customers are not affected.

One everything looks good, we would need to add this to our validation step.

achettyiitr:
> 
>  Although I find it unlikely someone would disable them.

That's true.

achettyiitr:
We use temporary tables for [SNOWFLAKE](https://github.com/rudderlabs/rudder-server/blob/88f1ec1d31dd65f8cc4093fdb1754972ef016482/warehouse/snowflake/snowflake.go#L268) integration as well.

achettyiitr:
> 
> What is considered a session by redshift? If every connection has its own session, is it possible that due to connection pooling, we create the temp in one connection and we try to query it in a different one?

@lvrach Is there a way we can test this scenario?

PR #2718, opened 2022-11-22
chore: close http responses in a way to allow the Transport to re-use the TCP connection

atzoum:
# Description

If the http response body is not read, the underlying TCP connection cannot be reused ([ref](https://pkg.go.dev/net/http#Response:~:text=The%20default%20HTTP%20client%27s%20Transport%20may%20not%0A%09//%20reuse%20HTTP/1.x%20%22keep%2Dalive%22%20TCP%20connections%20if%20the%20Body%20is%0A%09//%20not%20read%20to%20completion%20and%20closed.)).

For that purpose, introducing `httputil.CloseResponse`

```go
// CloseResponse closes the response's body. But reads at least some of the body so if it's
// small the underlying TCP connection will be re-used. No need to check for errors: if it
// fails, the Transport won't reuse it anyway.
func CloseResponse(resp *http.Response) {
	if resp != nil && resp.Body != nil {
		const maxBodySlurpSize = 2 << 10 // 2KB
		_, _ = io.CopyN(io.Discard, resp.Body, maxBodySlurpSize)
		resp.Body.Close()
	}
}
```

**_Note_**
The usage patterns of this new function might seem strange to you at first, since instead of
```go
defer httputil.CloseResponse(resp)
// or
httputil.CloseResponse(resp)
```
we are doing
```go
// or 
func() { httputil.CloseResponse(resp) }()
```

The justification for using the above patterns is that `golangci-lint` `bodyclose` linter, doesn't seem to recognise an indirect body close operation unless it is performed within a closure!!
## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=a166d3e5516f49098bc8a71cf42a1dc9&pm=s)

## Security

### DIFF
--- a/cmd/devtool/commands/event.go
+++ b/cmd/devtool/commands/event.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/rudderlabs/rudder-server/utils/httputil"
 	"github.com/urfave/cli/v2"
 )
 
@@ -64,40 +65,42 @@ func EventSend(c *cli.Context) error {
 	}
 
 	for i := 0; i < c.Int("count"); i++ {
-		anonymousId := uuid.New().String()
+		if err := func() error {
+			anonymousId := uuid.New().String()
+			buf := bytes.NewBuffer(nil)
+			err = t.Execute(buf, map[string]string{
+				"AnonymousId": anonymousId,
+				"Timestamp":   time.Now().Format(time.RFC3339),
+			})
+			if err != nil {
+				return err
+			}
 
-		buf := bytes.NewBuffer(nil)
-		err = t.Execute(buf, map[string]string{
-			"AnonymousId": anonymousId,
-			"Timestamp":   time.Now().Format(time.RFC3339),
-		})
-		if err != nil {
+			req, err := http.NewRequestWithContext(c.Context, "POST", url, buf)
+			if err != nil {
+				return err
+			}
+
+			req.SetBasicAuth(c.String("write-key"), "")
+			req.Header.Set("Content-Type", "application/json")
+			req.Header.Set("AnonymousId", anonymousId)
+
+			resp, err := client.Do(req)
+			if err != nil {
+				return err
+			}
+			defer func() { httputil.CloseResponse(resp) }()
+			b, err := io.ReadAll(resp.Body)
+			if err != nil {
+				return err
+			}
+			if resp.StatusCode != http.StatusOK {
+				fmt.Printf("%s\n%s\n", resp.Status, b)
+				return fmt.Errorf("status code: %d", resp.StatusCode)
+			}
+			return nil
+		}(); err != nil {
 			return err
-		}
-
-		req, err := http.NewRequestWithContext(c.Context, "POST", url, buf)
-		if err != nil {
-			return err
-		}
-
-		req.SetBasicAuth(c.String("write-key"), "")
-		req.Header.Set("Content-Type", "application/json")
-		req.Header.Set("AnonymousId", anonymousId)
-
-		resp, err := client.Do(req)
-		if err != nil {
-			return err
-		}
-		b, err := io.ReadAll(resp.Body)
-		if err != nil {
-			return err
-		}
-		defer resp.Body.Close()
-
-		if resp.StatusCode != http.StatusOK {
-			fmt.Printf("%s\n%s\n", resp.Status, b)
-
-			return fmt.Errorf("status code: %d", resp.StatusCode)
 		}
 	}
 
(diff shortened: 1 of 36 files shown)


lvrach:
Two concerns:
1. I don't find the usage of `httputil.CloseResponse` everywhere very useful; if you plan to read the body anyway, what it the point of using this function? I believe it will unnecessarily complicate things. I prefer to use it only if we plan to discard the body. 
2. The usage of `const maxBodySlurpSize = 2 << 10 // 2KB` is an implementation detail which makes sense to be part of [HTTP package](https://go.googlesource.com/go/+/3b2577c/src/net/http/client.go#350), but I would refrain from hardcoding in our code. 

My suggestion is the following:

Introducing this function:
```go
func DiscardBody(resp *http.Response) {
	if resp != nil && resp.Body != nil {
		_, _ = io.Copy(io.Discard, resp.Body)
		resp.Body.Close()
	}
}
```

Use this function when we don't plan to read the body.

lvrach:
Explaining myself a bit further:

* Adding a new function call in every HTTP request introduces overhead in all development cycles. 
* The typical case should be to read the body of a request. Not reading the body is the exception.
* I don't want to enforce the usage of a call in every call to handle the exception, but I like the idea of making it easy and readable for people to discard it.

The biggest issue is that we can not enforce the usage of those functions. The ideal solution is to have a linter that requires you to read and close the body.

atzoum:
> 2\. The usage of `const maxBodySlurpSize = 2 << 10 // 2KB` is an implementation detail which makes sense to be part of [HTTP package](https://go.googlesource.com/go/+/3b2577c/src/net/http/client.go#350), but I would refrain from hardcoding in our code.

You caught me! I found it smart though, not attempting to read larger bodies which would result in more time and resources being spent compared to simply opening a new tcp connection ;)

lvrach:
> > I don't find the usage of httputil.CloseResponse everywhere very useful; if you plan to read the body anyway, what it the point of using this function? I believe it will unnecessarily complicate things. I prefer to use it only if we plan to discard the body.
> 
> I don't see any complication in always trying to read the request body just before closing it, even if it was already read before (resulting in a no-op). If this pattern is used across our codebase, it becomes easier for less senior members of the team to pick it up, copy-paste it and guard themselves from the inevitable complex branching logic which can result in not reading the body after all, even though this was the initial intention :)

That is a good point. I have no strong argument; both approaches help. I will wait for @fracasula feedback.

fracasula:
I don't have a problem with always closing the body. Consistency can actually help in making sure we forget less. We're going to forget though. In this regard I think that some kind of linter as @atzoum proposed would be helpful in this case. Worst case we end up with what we have now (non-reusable connection), but writing a linter doesn't seem complex and it's definitely know-how that can turn out to be useful moving forward.

Another thing related to this PR, I'd report the `bodyclose` issue, it's silly that we have to wrap the close in an anonymous function.

Also another thing is that some `defer`s are inside for loops and that is not a good practice (possible resource leak). Whenever possible please create a dedicated function or an anonymous one to execute the defer asap instead of having defer calls piling up for each element we iterate over.

```go
for i := 0; i < c.Int("count"); i++ {
	/* ... */

	resp, err := client.Do(req)
	if err != nil {
		return err
	}
	b, err := io.ReadAll(resp.Body)
	if err != nil {
		return err
	}

	if resp.StatusCode != http.StatusOK {
		fmt.Printf("%s\n%s\n", resp.Status, b)

		return fmt.Errorf("status code: %d", resp.StatusCode)
	}
}
```

Becomes:

```go
for i := 0; i < c.Int("count"); i++ {
	/* ... */

	resp, b, err := doReq(client, req)
	if err != nil {
		return err
	}

	if resp.StatusCode != http.StatusOK {
		fmt.Printf("%s\n%s\n", resp.Status, b)

		return fmt.Errorf("status code: %d", resp.StatusCode)
	}
}

// ...

func doReq(client *http.Client, req *http.Request) (*http.Response, []byte, error) {
	resp, err := client.Do(req)
	if err != nil {
		return nil, nil, err
	}
	b, err := io.ReadAll(resp.Body)
	if err != nil {
		return nil, nil, err
	}
	return resp, b, nil
}
```

atzoum:
> some `defer`s are inside for loops

I could spot two occurrences of defers inside for loops. Please highlight any additional issues

atzoum:
> I'd report the `bodyclose` issue

I am afraid the linter's [repo](https://github.com/timakin/bodyclose) doesn't appear to be maintained anymore. It's automated build is failing and last commit is since 2021. There are also similar issues reported [here](https://github.com/timakin/bodyclose/issues/30) and [here](https://github.com/timakin/bodyclose/issues/39) without them receiving any response from the maintainer.

IMO, our best bet would be to move forward with our own linter, forking bodyclose, but this will take me some time

fracasula:
Yeah I didn't realise it wasn't maintained anymore, that is unfortunate.

lvrach (approved):

fracasula (approved):

PR #2723, opened 2022-11-23
feat(warehouse): added support for warehouse column count limit

achettyiitr:
# Description

- Added support for warehouse column count limit using a threshold.
- Introduced stats in `UploadJob` and `JobRun` structs for testing the stats. Also to remove the Global dependency being used in stats.
- Using [memStats](https://github.com/rudderlabs/rudder-server/pull/2735) for testing.

## Notion Ticket

https://www.notion.so/rudderstacks/Send-warehouse-load-table-column-count-alert-to-customer-directly-10496242a6324e7e84ed629790efe208

## Security


warehouse/stats_test.go
```diff
@@ -59,39 +51,6 @@ var _ = Describe("Stats", Ordered, func() {
...
-		BeforeEach(func() {
-			mockStats, mockMeasurement := getMockStats(g)
-			mockStats.EXPECT().NewTaggedStat(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(mockMeasurement)
-			mockMeasurement.EXPECT().Count(gomock.Any()).AnyTimes()
-
-			stats.Default = mockStats
-		})
-
-		It("Upload status stat", func() {
-			getUploadStatusStat(statName, warehouseutils.Warehouse{
```
achettyiitr, line 80: Removed these 2 tests because it does not test any behavior.

warehouse/upload.go
```diff
@@ -1067,10 +1068,25 @@ func (job *UploadJobT) loadTable(tName string) (alteredSchema bool, err error) {
...
+					name: "tableName", value: strings.ToLower(tableName),
+				},
+				{
+					name: "columnCountLimit", value: strconv.Itoa(columnCountLimit),
+				},
+			}
+
+			job.counterStat(`warehouse_load_table_column_count`, tags...).Count(currentColumnsCount)
 		}
 	}
```
abhimanyubabbar, line 1091:
```suggestion
	if columnCountLimit, ok := columnCountLimitMap[job.warehouse.Type]; !ok {
	    return
	}
	currentColumnCount := len(job.schemaHandle.schemaInWarehouse[tableName])
	if currentColumnsCount > int(float64(columnCountLimit)*columnCountLimitThreshold) {
		tags := []tag{
			{
				name: "tableName", value: strings.ToLower(tableName),
			},
			{
				name: "columnCountLimit", value: strconv.Itoa(columnCountLimit),
			},
		}

		job.counterStat(`warehouse_load_table_column_count`,
		tags...).Count(currentColumnsCount)
	}
```

abhimanyubabbar: Just returning early to save on the indentation. Wdyt ?

abhimanyubabbar (approved):
Minor changes, overall the change looks good 👍🏼

warehouse/upload.go
achettyiitr: Done.

warehouse/jobs/types.go
```diff
@@ -91,5 +91,5 @@ const (
 	MaxQueryRetries       int = 3
 	RetryTimeInterval         = 10 * time.Second
 	MaxAttemptsPerJob     int = 3
-	WhAsyncJobTimeOut         = 10 * time.Second
+	WhAsyncJobTimeOut         = 300 * time.Second
```
Jayachand, line 94: this change is present in diff PR: https://github.com/rudderlabs/rudder-server/pull/2721

warehouse/upload.go
```diff
@@ -1067,11 +1068,31 @@ func (job *UploadJobT) loadTable(tName string) (alteredSchema bool, err error) {
...
+	if currentColumnsCount > int(float64(columnCountLimit)*columnCountLimitThreshold) {
+		tags := []tag{
+			{
+				name: "tableName", value: strings.ToLower(tableName),
+			},
+			{
+				name: "columnCountLimit", value: strconv.Itoa(columnCountLimit),
+			},
 		}
+
```
lvrach, line 1094: Would it make sense to report two metrics? 

`warehouse_load_table_column_count`

`warehouse_load_table_column_limit`

And instead of computing the threshold here, we compute it on the alert manager, by doing a similar operation

lvrach: Should we add other identifiers as well, like workspaceId and destinationId ?

abhimanyubabbar: Can you elaborate the motivation behind computing it at the alert manager rather than here ? @lvrach

achettyiitr: > Should we add other identifiers as well, like workspaceId and destinationId ?

We are adding these inside `job.counterStat(`warehouse_load_table_column_count`, tags...).Count(currentColumnsCount)`

achettyiitr: `warehouse_load_table_column_limit` This is basically a constant. We should not probably send it as a stat.

warehouse/upload.go
```diff
@@ -167,15 +168,14 @@ func setMaxParallelLoads() {
...
 		warehouseutils.DELTALAKE:  config.GetInt("Warehouse.deltalake.maxParallelLoads", 3),
 	}
-	columnCountThresholds = map[string]int{
-		warehouseutils.AZURE_SYNAPSE: config.GetInt("Warehouse.azure_synapse.columnCountThreshold", 800),
-		warehouseutils.BQ:            config.GetInt("Warehouse.bigquery.columnCountThreshold", 8000),
-		warehouseutils.CLICKHOUSE:    config.GetInt("Warehouse.clickhouse.columnCountThreshold", 800),
-		warehouseutils.MSSQL:         config.GetInt("Warehouse.mssql.columnCountThreshold", 800),
-		warehouseutils.POSTGRES:      config.GetInt("Warehouse.postgres.columnCountThreshold", 1200),
-		warehouseutils.RS:            config.GetInt("Warehouse.redshift.columnCountThreshold", 1200),
-		warehouseutils.SNOWFLAKE:     config.GetInt("Warehouse.snowflake.columnCountThreshold", 1600),
```
achettyiitr, line 177: We are not planning to add this for SNOWFLAKE because in case of SNOWFLAKE we have a limit on the row size to be 16MB and this will most likely not meet because of the payload max size we have.

warehouse/upload.go
lvrach: > Can you elaborate the motivation behind computing it at the alert manager rather than here? @lvrach

Since the threshold dictates whether an alert should be sent, it is more natural for me to have it on the alert manager. It makes it easier to change it. Also, the fact that we add `tableName` & `columnCountLimit` only if the threshold passes prevents us from doing back-testing (going to grafana dashboard and checking when the alert will be triggered with another threshold). 

> warehouse_load_table_column_limit This is a constant. We should probably not send it as a stat.
If you want to visualise the limit as a line in grafana it is easier if it's metric.

achettyiitr: > Also, the fact that we add tableName & columnCountLimit only if the threshold passes prevents us from doing back-testing (going to grafana dashboard and checking when the alert will be triggered with another threshold).

Good point.

Jayachand: > `warehouse_load_table_column_limit` This is a constant. We should probably not send it as a stat.
> 
> If you want to visualise the limit as a line in grafana it is easier if it's metric.

I still feel this information is not much useful. Recording in separate stat and adding tags warehouseId, tableName for an almost constant value (changes only on destination type) would be tedious. Adds computational cost to kapacitor or prometheus executor(future) too.
wdyt?

achettyiitr: Better for debugging and visualization.

lvrach (approved):

PR #2725, opened 2022-11-23
exp: benchmark union query vs parallel query

cisse21:
# Description

BenchmarkUnionQuery/UnionQuery-1000-8         	       1	1204381167 ns/op
BenchmarkUnionQuery/ParallelQuery-1000-8      	       1	1458239917 ns/op
BenchmarkUnionQuery/UnionQuery-10000-8        	       1	13644813875 ns/op
BenchmarkUnionQuery/ParallelQuery-10000-8     	       1	10617516792 ns/op
BenchmarkUnionQuery/UnionQuery-100000-8       	       1	122106416959 ns/op
BenchmarkUnionQuery/ParallelQuery-100000-8    	       1	109830196208 ns/op

## Notion Ticket

< Replace with Notion Link >

## Security


jobsdb/unionQuery_test.go
```diff
@@ -111,3 +114,183 @@ func TestMultiTenantHandleT_GetAllJobs(t *testing.T) {
...
+		wg.Wait()
+	}
+}
+
+func generateJobs(size int) ([]*JobT, map[string]int) {
+	workspaces := []string{"testWorkspace1", "testWorkspace2", "testWorkspace3", "testWorkspace4", "testWorkspace5", "testWorkspace6", "testWorkspace7", "testWorkspace8", "testWorkspace9", "testWorkspace10", "testWorkspace11", "testWorkspace12", "testWorkspace13", "testWorkspace14", "testWorkspace15", "testWorkspace16", "testWorkspace17", "testWorkspace18", "testWorkspace19", "testWorkspace20"}
+	jobs := make([]*JobT, size)
+	testWorkspacePickup := map[string]int{}
+	for i := 0; i < size; i++ {
+		workspaceId := workspaces[rand.Int()%len(workspaces)]
```
fracasula, line 284: I don't like that we don't control the number of workspaces in the pickup. This could potentially give very different results. If very unlucky we could end up with always the same workspace(s) in the pickup map. Can we make the number of workspaces that we want in the pickup a part of the benchmark configuration?

For example:
1. union query 1k jobs and 5 workspaces
2. union query 1k jobs and 20 workspaces
3. union query 1k jobs and 50 workspaces
4. ...

Then repeat with parallel query then repeat with 10k jobs and again 100k jobs.

jobsdb/unionQuery_test.go
```diff
@@ -111,3 +114,183 @@ func TestMultiTenantHandleT_GetAllJobs(t *testing.T) {
...
+
+BenchmarkUnionQuery/UnionQuery-100000-8                         	                                 143067570417 ns/op
+BenchmarkUnionQuery/Concurrent_Union_Query__-100000_-16-8       	                 200310106917 ns/op
+BenchmarkUnionQuery/Concurrent_Union_Query__-100000_-32-8       	                 409845463834 ns/op
+BenchmarkUnionQuery/ParallelQuery-100000-8                      	                                  118793139708 ns/op
+*/
+
+// BenchmarkUnionQuery is a benchmark for UnionQuery with 20 different workspaces
+// and jobSize is the total number of jobs split across the workspaces
+func BenchmarkUnionQuery(b *testing.B) {
```
fracasula, line 149: Can we add a benchmark where we partitition the table by workspaceID? I think it might give us some benefits in scenarios with many workspaces and parallel queries. We a configurable number of workspaces as per [this comment](https://github.com/rudderlabs/rudder-server/pull/2725/files#r1036208241) we should get an idea.

fracasula: I was thinking that we could try both `LIST` and `HASH` partitioning in the benchmark. With the `LIST` one it should fairly simple to setup a partition per workspace.

jobsdb/unionQuery_test.go
```diff
@@ -111,3 +113,230 @@ func TestMultiTenantHandleT_GetAllJobs(t *testing.T) {
...
+					//jobsDb1.TearDown()
+
+					maxDSSize := 5
+					jobsDb2 := MultiTenantLegacy{HandleT: &HandleT{
+						MaxDSSize: &maxDSSize,
+					}}
+					err := jobsDb2.Setup(ReadWrite, true, tablePrefix, true, []prebackup.Handler{}, fileuploader.NewDefaultProvider(), partitionType)
+					require.NoError(b, err)
+					err = jobsDb2.Store(context.Background(), jobs)
+					require.NoError(b, err)
```
fracasula, line 201: @cisse21 @BonapartePC beware that we're still writing first and then reading. in production we're doing both at the same time so the benchmark won't highlight the difference in performance that we would get in a real scenario.

fracasula: Let's have a minimum number of records that are already in the table when the benchmark starts and then a write frequency (e.g. start with 100k rows, then start benchmarking the reads but have X writes per second while the reads are running). We'll have the same number of writes per second when benchmarking different types of queries so the benchmark will still be valid.

PR #2730, opened 2022-11-24
fix: regulation-worker changes for oauth destinations

sanpj2292:
# Description

Initially in regulation-worker for OAuth destinations as soon as the refresh token request is complete, we would retry the regulation worker job recursively. It is possible that the recursion happens for infinite number of times(according to source-code but in actual case it can be rarely seen).
To prevent the risk of recursive calls happening infinite number of times for a job. We want to restrict it to only `once`
This PR includes this change.

Several other things included
- We have included a configurable maxRetryAttempts for OAuth case
- Included a stat to help us understand if the request to control plane(as part of the OAuth flow) is timing out

## Notion Ticket

[Notion](https://www.notion.so/rudderstacks/Integrate-OAuth-in-regulation-worker-module-of-server-to-support-GA-dae0cf6ef89e41eb91e9212ac524cf49)

## Security


regulation-worker/internal/delete/api/api.go
```diff
@@ -106,22 +105,29 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destination mo
...
 		err = api.refreshOAuthToken(destination.Name, job.WorkspaceID, oAuthDetail)
 		if err != nil {
 			pkgLogger.Error(err)
 			return model.JobStatusFailed
 		}
 		// retry the request
-		pkgLogger.Debug("Retrying deleteRequest job for the whole batch")
-		return api.Delete(ctx, job, destination)
+		pkgLogger.Debugf("Retrying deleteRequest job for the whole batch: %v", job.DestinationID)
+		attempts++
```
saikumarrs, line 118: It doesn't make sense to increment the incoming variable. You can send `attempts + 1` during invocation.

regulation-worker/internal/delete/api/api.go
```diff
@@ -41,9 +41,8 @@ func (*APIManager) GetSupportedDestinations() []string {
 	return supportedDestinations
 }
 
-// prepares payload based on (job,destDetail) & make an API call to transformer.
-// gets (status, failure_reason) which is converted to appropriate model.Error & returned to caller.
-func (api *APIManager) Delete(ctx context.Context, job model.Job, destination model.Destination) model.JobStatus {
+func (api *APIManager) delete(ctx context.Context, job model.Job, destination model.Destination, attempts int) model.JobStatus {
```
saikumarrs, line 44: As we track the no of retry attempts, `retryAttempts` is better than `attempts`.

saikumarrs, line 44: `Delete` and `delete` names for the functions would be very confusing. Please pick something else.
Suggestion: Call the same method recursively as we have a terminating condition now.

regulation-worker/internal/delete/api/api.go
```diff
@@ -106,22 +105,29 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destination mo
 		return model.JobStatusFailed
 	}
 	jobStatus := getJobStatus(resp.StatusCode, jobResp)
-	pkgLogger.Debugf("[%v] JobStatus for %v: %v", destination.Name, destination.DestinationID, jobStatus)
+	pkgLogger.Debugf("[%v] JobStatus for %v: %v", destination.Name, job.DestinationID, jobStatus)
 
-	if isOAuthEnabled && isTokenExpired(jobResp) {
+	if isOAuthEnabled && isTokenExpired(jobResp) && attempts < 1 {
```
saikumarrs, line 110: I'd make the max retry attempts configurable.

regulation-worker/internal/delete/api/api.go
koladilip: we need another function as golang doesn't support default or optional params

regulation-worker/internal/delete/api/api.go
```diff
@@ -106,22 +105,29 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destination mo
...
+		attempts++
+		return api.delete(ctx, job, destination, attempts)
 	}
 
 	return jobStatus
 }
 
+// prepares payload based on (job,destDetail) & make an API call to transformer.
+// gets (status, failure_reason) which is converted to appropriate model.Error & returned to caller.
+func (api *APIManager) Delete(ctx context.Context, job model.Job, destination model.Destination) model.JobStatus {
```
koladilip, line 127: https://pkg.go.dev/github.com/cenkalti/backoff/v4#WithMaxRetries does it make sense to use libraries like these to do the same

regulation-worker/internal/delete/api/api.go
saikumarrs: Ok, then 'Delete' and 'DeleteWithRetyAttempts'.

regulation-worker/internal/delete/api/api.go
sanpj2292: Done. Can you please verify ?

regulation-worker/internal/delete/api/api.go

regulation-worker/internal/delete/api/api.go
sanpj2292: Made it configurable. Can you please verify ?

regulation-worker/internal/delete/api/api.go
koladilip: Don't we need any gap between retries?

regulation-worker/internal/delete/api/api.go
```diff
@@ -41,9 +42,7 @@ func (*APIManager) GetSupportedDestinations() []string {
 	return supportedDestinations
 }
 
-// prepares payload based on (job,destDetail) & make an API call to transformer.
-// gets (status, failure_reason) which is converted to appropriate model.Error & returned to caller.
-func (api *APIManager) Delete(ctx context.Context, job model.Job, destination model.Destination) model.JobStatus {
+func (api *APIManager) deleteWithRetry(ctx context.Context, job model.Job, destination model.Destination, retryAttempts int) model.JobStatus {
```
koladilip, line 45: retryAttempts -> retryAttempt  or retryAttemptCount or something like makes more sense as it is indicating, up till how many attempts happened right?

saikumarrs: It can go either way:
- count how many retries happened until now
- current retry number

I'd prefer the latter.

regulation-worker/cmd/main.go
```diff
@@ -80,9 +80,10 @@ func Run(ctx context.Context) {
 				FilesLimit: config.GetInt("REGULATION_WORKER_FILES_LIMIT", 1000),
 			},
 			&api.APIManager{
-				Client:           &http.Client{Timeout: config.GetDuration("HttpClient.regulationWorker.transformer.timeout", 60, time.Second)},
-				DestTransformURL: config.MustGetString("DEST_TRANSFORM_URL"),
-				OAuth:            OAuth,
+				Client:                       &http.Client{Timeout: config.GetDuration("HttpClient.regulationWorker.transformer.timeout", 60, time.Second)},
+				DestTransformURL:             config.MustGetString("DEST_TRANSFORM_URL"),
+				OAuth:                        OAuth,
+				MaxOAuthRefreshRetryAttempts: config.GetInt("RegulationWorker.oauth.maxRefreshRetryAttempts", 1),
```
saikumarrs, line 86: I'd not tie this variable to OAuth as it's a delete retry we are doing rather than a "refresh retry".
Moreover, we might retry a delete request due to other failure scenarios that could possibly succeed in the next retry.

saikumarrs (approved):

regulation-worker/cmd/main.go
sanpj2292: The retry only is required for the OAuth case in order to prevent the case where in a failed job is retried after the token's validity period.

koladilip (approved):

regulation-worker/internal/delete/api/api.go
```diff
@@ -106,22 +105,28 @@ func (api *APIManager) Delete(ctx context.Context, job model.Job, destination mo
 		return model.JobStatusFailed
 	}
 	jobStatus := getJobStatus(resp.StatusCode, jobResp)
-	pkgLogger.Debugf("[%v] JobStatus for %v: %v", destination.Name, destination.DestinationID, jobStatus)
+	pkgLogger.Debugf("[%v] Job: %v, JobStatus: %v", destination.Name, job.ID, jobStatus)
 
-	if isOAuthEnabled && isTokenExpired(jobResp) {
+	if isOAuthEnabled && isTokenExpired(jobResp) && currentOauthRetryAttempt < api.MaxOAuthRefreshRetryAttempts {
```
atzoum, line 110: what the `jobStatus` will be when `isOAuthEnabled && isTokenExpired && currentOauthRetryAttempt >= api.MaxOAuthRefreshRetryAttempts`?

sanpj2292: Whatever is calculated as part of `getJobStatus()` that we got in the last retry(when `currentOauthRetry = api.MaxOAuthRefreshRetryAttempts - 1`)

sanpj2292: There is already a retry mechanism such that regulation-manager would provide the already failed jobs to regulation-worker to send it to destinaiton, so that way I feel it should be ok. Let me know if you think otherwise

atzoum: I will try to rephrase the question:
What will the transformer's **HTTP response code** be whenever it returns in its response body at least 1 entry with `authErrorCategory=REFRESH_TOKEN`? Is it deterministic or not?

sanpj2292: Whenever atleast one entry of `authErrorCategory = REFRESH_TOKEN` is present in response body, we will return a status-code of `500` as transformer http status-code as of today.

sanpj2292: When you say deterministic, are you asking about how do we identify perpetual refresh token failures, I mean the scenario wherein even after multiple retries(when `api.MaxOAuthRefreshRetryAttempts > 1`), we are still getting `refresh_token` error from transformer or may be it is something else ?

atzoum: Just trying to make sure that the jobStatus will be `model.JobStatusFailed` if the token refresh fails or we have reached to maximum refresh attempts :)

atzoum (approved):

PR #2732, opened 2022-11-24
fix(warehouse): syncs issues

achettyiitr:
# Description

- Handle no syncs case when the warehouse tries to merge existing syncs with the new ones thereby users do not get the existing syncs and we are sending a 204 response code to handle it.
- Added logs for warehouse Grpc for debugging purpose.

## Notion Ticket
https://www.notion.so/rudderstacks/Syncs-dashboard-fixes-4d3c454856d844e09cc15a73012a37e1

## Security


warehouse/api.go
```diff
@@ -248,18 +249,32 @@ func (uploadsReq *UploadsReqT) TriggerWhUploads() (response *proto.TriggerWhUplo
...
+	response := &proto.WHUploadResponse{}
 	err := uploadReq.validateReq()
+	//defer func() {
+	//	if err != nil {
+	//		response = &proto.WHUploadResponse{
+	//			StatusCode: http.StatusBadRequest,
+	//		}
+	//	}
+	//}()
+
```
lvrach, line 261:
```suggestion


```

lvrach (approved):

proto/warehouse/warehouse.proto
```diff
@@ -73,6 +73,7 @@ message WHUploadResponse {
   int32 duration = 14;
   repeated WHTable tables = 15;
   bool isArchivedUpload = 16;
+  int32 status_code = 17;
```
lvrach, line 76: I am a bit concern with adding HTTP semantics into GRPC, is it because we are using an HTTP API at some part of the chain ?

achettyiitr: Let me check how should we remove this.

warehouse/api.go
```diff
@@ -106,6 +110,7 @@ const (
 	TriggeredSuccessfully   = "Triggered successfully"
 	NoPendingEvents         = "No pending events to sync for this destination"
 	DownloadFileNamePattern = "downloadfile.*.tmp"
+	NoSuchSyncs             = "No such sync exist"
```
abhimanyubabbar, line 113: nit: `NoSuchSync`

abhimanyubabbar (approved):
Minor nit, overall looks good. 👍🏼

PR #2733, opened 2022-11-24
chore: completely replace gofrs with google uuid 

lvrach:
# Description

We have started using the google uuid library as it [is more performant](https://github.com/rudderlabs/rudder-server/pull/1606). This is the last iteration to altogether replace gofrs from our code.

The motivation for this PR, in particular, was to stop using two libraries for uuid in the same codebase, causing UUID type conflicts and unnecessary conversations. Also, people would use gofrs uuid in places where performance is vital due to copy-pasting. 

I've also added a [depguard](https://golangci-lint.run/usage/linters/#depguard) linting rule, to enforce the replacement.

## Notion Ticket

https://www.notion.so/rudderstacks/chore-replace-gofrs-with-google-uuid-0365427537fb4055ba5fe10228e12c4c

## Security


.golangci.yml
```diff
@@ -56,3 +57,43 @@ issues:
...
+    # Default: []
+    # ignore-file-rules:
+    #   - "**/test_*.go"
+    # # Create additional guards that follow the same configuration pattern.
+    # # Results from all guards are aggregated together.
+    additional-guards:
+      - list-type: denylist
+        include-go-root: false
+        packages:
+          - github.com/gofrs/uuid
```
fracasula, line 94: This is very useful, I didn't know about `depguard` :+1:

utils/misc/misc.go
```diff
@@ -1012,11 +1010,25 @@ func MakeRetryablePostRequest(url, endpoint string, data interface{}) (response
 
 // GetMD5UUID hashes the given string into md5 and returns it as auuid
 func GetMD5UUID(str string) (uuid.UUID, error) {
```
fracasula, line 1012: Given the changes introduced, it might make sense to write some test to make sure we're meeting the expectations.

fracasula (approved):

utils/misc/misc.go
lvrach: I have written a fuzzy test to compare this with the previous implementation. Is anything else you have in mind?

achettyiitr (approved):

atzoum (approved):

PR #2734, opened 2022-11-24
refactor: safe database initialisation and schema migration during startup

atzoum:
# Description

When running in `embedded` mode (core & warehouse) and using separate database instances for core & warehouse data, it is possible that one database's unavailability panic may cause the other database's schema migration to become dirty (corrupted) due to the asynchronous nature of the two components' startup sequence.

To eliminate this, we are now initialising both databases sequentially and stop the process gracefully in case of an error. Database initialisation for reach component happens only if the component needs to be started.

Other improvements:
- removed a handful of panics for supporting graceful termination on failure scenarios
- refactored the `apphandlers` package to remove global variables
- introduced tests for some exceptional `runner` & `apphandler` scenarios

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=bdf94aeaedb84e5c985a3b6648bf45fb&pm=s)

## Security

### DIFF
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/rudderlabs/rudder-server/config"
+	"github.com/rudderlabs/rudder-server/utils/logger"
 	"github.com/rudderlabs/rudder-server/utils/types/deployment"
 
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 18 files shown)


pChondros (approved):

app/apphandlers/embeddedAppHandler.go
```diff
@@ -34,46 +36,86 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/types/servermode"
 )
 
-// EmbeddedApp is the type for embedded type implementation
-type EmbeddedApp struct {
-	App            app.App
-	VersionHandler func(w http.ResponseWriter, r *http.Request)
+// embeddedApp is the type for embedded type implementation
+type embeddedApp struct {
```
fracasula, line 40: Nice job with unexporting the type :+1:

fracasula (approved):

chandumlg (approved):

PR #2735, opened 2022-11-24
chore: introduce in-memory stats for testing

lvrach:
# Description

This PR introduces an implementation of `stats.Stats` and `stats.Measurements`, where the metrics are kept in a simple in-memory store to assist with testing stats side-effects.

## Motivation

The current method [of testing stats](https://github.com/rudderlabs/rudder-server/pull/2723/files#diff-11f5384e9631d0f96cb207f5f03591802f7287be5a1083dbbc1080b2d0bcd39bR124) is to use an auto-generated mock:

```
	ctrl := gomock.NewController(t)
	mockStats := mock_stats.NewMockStats(ctrl)
	mockMeasurement := mock_stats.NewMockMeasurement(ctrl)
```

and latter
```go
				mockStats.EXPECT().NewTaggedStat(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(mockMeasurement)
				mockMeasurement.EXPECT().Count(3).Times(1)
```
 
Apart from this approach being a bit complex, the primary issue is the fragility and extensibility of this approach.

Imagine a new stats is introduced, which shouldn't be uncommon. The assertion above will fail and to fix it, you need to do something like:

```go
				mockStats.EXPECT().NewTaggedStat("the_counter", gomock.Any(), gomock.Any()).Times(1).Return(mockMeasurement)

	mockMeasurementSecond := mock_stats.NewMockMeasurement(ctrl)
				mockStats.EXPECT().NewTaggedStat("the_timer", gomock.Any(), gomock.Any()).Times(1).Return(mockMeasurementSecond)
				mockMeasurementSecond.EXPECT().SendTiming(now).Times(1)
```

You can easily understand how those tests can become unmanageable.

## Proposed solution

```go
   	store := &memstats.Store{}
       // inject into stats.Stats
       // run your code
       require.Equal(t, 1, store.Get("stat_name, tags.Tags{"key": "value"}).LastValue)
```
This approach has the following benefits:
1. We test exactly the expected metric we will send. `name`, and `tags` should match our expectations.
2. Easy to extend your tests to new measurements.
3. Ignore metrics we are not interested in testing.

## Limitations

* This is not meant to run in production; the implementation is not performant.

## Notion Ticket

https://www.notion.so/rudderstacks/memory-stats-for-tests-2846f156181f4843a848471660529d61

## Security


services/stats/memstats/stats.go
```diff
@@ -0,0 +1,181 @@
...
+	m.LastValue += float64(n)
+	m.Values = append(m.Values, m.LastValue)
+}
+
+// Increment implements stats.Measurement
+func (m *Measurement) Increment() {
+	if m.Type != stats.CountType {
+		panic("operation Increment not supported for measurement type:" + m.Type)
+	}
+
```
fracasula, line 56: This check is redundant.

```suggestion
```

services/stats/memstats/stats.go
```diff
@@ -0,0 +1,181 @@
...
+	m.startTime = m.now()
+}
+
+// End implements stats.Measurement
+func (m *Measurement) End() {
+	if m.Type != stats.TimerType {
+		panic("operation End not supported for measurement type:" + m.Type)
+	}
+
```
fracasula, line 103: This check is redundant.

services/stats/memstats/stats.go
```diff
@@ -0,0 +1,181 @@
...
+	m.SendTiming(m.now().Sub(m.startTime))
+}
+
+// Since implements stats.Measurement
+func (m *Measurement) Since(start time.Time) {
+	if m.Type != stats.TimerType {
+		panic("operation Since not supported for measurement type:" + m.Type)
+	}
+
```
fracasula, line 112: This check is redundant.

services/stats/memstats/stats.go
```diff
@@ -0,0 +1,181 @@
...
+	m := &Measurement{
+		Name: name,
+		Tags: tags,
+		Type: statType,
+
+		now: ms.Now,
+	}
+
+	ms.byKey[name+tags.String()] = m
```
fracasula, line 163: Probably a overkilling but I'd generate they key with a function:
```go
func (*Store) getKey(name string, tags stats.Tags) string {
    return name+tags.String()
}
```

fracasula (commented):
Overall looks good to me and I like that this would simplify our tests. Personally I find gomock to be very unfriendly and I tend to prefer [moq](https://github.com/matryer/moq) a lot more. Not sure if with moq you'd still feel the need to do all this, perhaps it is worth checking before merging this (just in case since it might help with other cases too).

So mostly two things:
1. please check moq, let me know if it can help somehow
2. beware of reading `LastValue` directly in the tests. if some code is still running the write is protected by the library via a mutex but in the test you're reading it without passing through the mutex. it's usually fine in unit tests, less fine in integration tests where the code that you're testing might still be running while you make your assertions. that could cause a failure when running tests of that kind with `-race`

services/stats/memstats/stats.go
```diff
@@ -0,0 +1,181 @@
...
+	"time"
+
+	"github.com/rudderlabs/rudder-server/services/stats"
+)
+
+var _ stats.Stats = (*Store)(nil)
+
+var _ stats.Measurement = (*Measurement)(nil)
+
+type Store struct {
```
atzoum, line 15: all fields in the `Store` struct could be unexported and intialization could be taken care once and for all by a producer function using `Opts` if necessary. This would eliminate the need for the `init` method and the `sync.Once` field e.g.

```go
type Opt func(*Store)

var WithNow = func(now func() time.Time) Opt {
	return func(s *Store) {
		s.now = now
	}
}

// New returns a new stats.Stats implementation that stores stats in-memory.
func New(opts ...Opt) *Store {
	s := &Store{
		byKey: make(map[string]*measurement),
		now:   time.Now,
	}
	for _, opt := range opts {
		opt(s)
	}
	return s
}

type Store struct {
	mu    sync.Mutex
	byKey map[string]*measurement
	now func() time.Time
}
```

and other packages could simply do
```go
store := memstats.New(
		memstats.WithNow(func() time.Time {
			return now
		}))
```

atzoum (approved):
It would be nice before merging it to use it in at least one test :)

lvrach:
> It would be nice before merging it to use it in at least one test :)

@atzoum I would added it [here](https://github.com/rudderlabs/rudder-server/pull/2723/files#diff-11f5384e9631d0f96cb207f5f03591802f7287be5a1083dbbc1080b2d0bcd39bR124) If @achettyiitr likes it.

I don't want to couple these changes in the same PR, given it addressing a different audience and they have different priorities.

services/stats/memstats/stats.go
lvrach: Your suggestion and my implementation are good patterns to initialise components in Go. I use and like both depending on the situation.

I don't see it as a problem having unexported fields for options; keep in mind most standard libraries are using this pattern.

The downside of the lazy init approach is that we have to call `.init` in every exported method. The `with` approach also has an overhead of as well but is safer indeed.

services/stats/memstats/stats.go
lvrach: > 1. please check moq, let me know if it can help somehow

Overall I prefer moq for generating mocks. I still find it will add unnecessary complexity to the alerts. The reason is stats and measurement are very coupled, the problem appears when you have to deal with multiple metrics. Essentially the logic that is implemented here should be repeated one way or the other, in all the test cases. 

* Ensuring that the correct measurement (name, tags) is changed.
* Ignoring stats we are not interested in.

Thought I could have used the moq, to build the measurement implementation.

Given the frequency and complexity of testing stats, I would go ahead with this custom implementation. Maybe we should have the same for logging.


> 2. beware of reading `LastValue` directly in the tests. if some code is still running the write is protected by the library via a mutex but in the test you're reading it without passing through the mutex. it's usually fine in unit tests, less fine in integration tests where the code that you're testing might still be running while you make your assertions. that could cause a failure when running tests of that kind with `-race`
> 
> In any case just reading the last value in some cases feels a bit fragile so perhaps a race failure is OK as it would force us to review the test. Anyhow, this could be a good starting point, imo we can add more on it if needed down the road.

Good point, it slipped my mind, I will make sure it is thread-safe.

services/stats/memstats/stats.go
lvrach: Why do you say this check is redundant?

fracasula (approved):
@lvrach I replied to the "redundant check" comment. if it makes sense please make the changes. overall again, it looks good. thanks for the PR :+1:

services/stats/memstats/stats.go
lvrach: I see; I did it this way cause the panic would be a bit misleading otherwise.

```go
m.Increment()
```

Will panic with `operation Count not supported for measurement type:" + m.Type`

achettyiitr (approved):

PR #2736, opened 2022-11-24
failing executing jobs instead of deleting them

Sidddddarth:
# Description

instead of deleting executing statuses on sever startup, they are now marked failed.
This is done so that any additional info(attempt numbers etc.) are preserver for further pickups.

## Notion Ticket

[fail executing jobs on startup](https://www.notion.so/rudderstacks/fail-executing-jobs-on-server-startup-21dd951bb376437daae2077281aec80c)

## Security


jobsdb/jobsdb.go
```diff
@@ -3385,59 +3386,31 @@ func (jd *HandleT) deleteJobStatusInTx(txHandler transactionHandler, conditions
...
-		return 0, err
-	}
-	deleteCount, err := res.RowsAffected()
-	if err != nil {
-		return 0, err
-	}
 
-	return int(deleteCount), nil
+	_, err = stmt.Exec(getTimeNowFunc())
+	return err
```
atzoum, line 3413: it could be simplified
```
_, err := txHandler.Exec(fmt.Sprintf(
		`UPDATE %[1]q SET job_state='failed'
			WHERE id = ANY(
				SELECT id from "v_last_%[1]s" where job_state='executing'
			) AND retry_time < $1`,
		ds.JobStatusTable,
	))
	return err
```

jobsdb/jobsdb.go
```diff
@@ -287,6 +287,7 @@ type JobsDB interface {
 	Status() interface{}
 	Ping() error
 	DeleteExecuting()
```
atzoum, line 289: what would the implications be if we failed executing in processor too?

Sidddddarth: The processor only looks for unprocessed jobs, and marks them succeeded in any case.
Failed jobs wouldn't be picked at all by it when the server starts up.
There was not much motivation in this particular case to preserve the statuses because each job is picked only once(unless of course the server crashes) and marked `succeeded` anyway.

atzoum (approved):

chandumlg (approved):

PR #2738, opened 2022-11-25
fix: metadata in event_schema table has TotalCount which exceeds 32bit integer limits

raman2805:
# Description

The event schema tables have a metadata column which contains a TotalCount field. This was being casted to a 32 bit integer. We exceeded those limits for one of our customers resulting in the value of TotalCount to be greater that the 32 bit integer range, and leading to a subsequent crash in the server pod. The SQL query has been modified to cast that into a 64 bit integer field.


abhimanyubabbar (approved):
Looks good 👍🏼

cisse21 (approved):

chandumlg (approved):

PR #2739, opened 2022-11-28
feat(warehouse): clickhouse s3 table engine for load tables

achettyiitr:
# Description

### [s3 Table Function](https://clickhouse.com/docs/en/sql-reference/table-functions/s3)

- This engine provides integration with [Amazon S3](https://aws.amazon.com/s3/) ecosystem. This engine is similar to the [HDFS](https://clickhouse.com/docs/en/engines/table-engines/special/file#table_engines-hdfs) engine but provides S3-specific features.

### Changes
- Added support for **S3 engine** for loading the load files into tables directly.
- Converting **Handle** struct to **Clickhouse** struct for better abbreviated naming. 
- **Removed Global configs**. Added them as part of the **Clickhouse** struct for better testability.
- **logger** as part of Clickhouse.
- Added **Test coverage** **(~75%)**.

## Notion Ticket

https://www.notion.so/rudderstacks/Use-ClickHouse-S3-Engine-to-create-staging-tables-and-use-if-for-dumping-the-data-6a2065f59ae949b899392e150b71a73d

## Security


warehouse/clickhouse/clickhouse.go
```diff
@@ -45,6 +47,7 @@ var (
 	commitTimeOutInSeconds      time.Duration
 	loadTableFailureRetries     int
 	numWorkersDownloadLoadFiles int
+	enableS3EngineForLoading    bool
```
achettyiitr, line 50: Instead of this use workspace-specific config.

warehouse/clickhouse/clickhouse.go
```diff
@@ -562,6 +585,75 @@ func (ch *HandleT) loadTable(tableName string, tableSchemaInUpload warehouseutil
...
+		`,
+		ch.Namespace,                   // 1
+		tableName,                      // 2
+		sortedColumnNames,              // 3
+		loadFolder,                     // 4
+		accessKeyID,                    // 5
+		secretAccessKey,                // 6
+		sortedColumnNamesWithDataTypes, // 7
+	)
+	_, err = ch.Db.Exec(sqlStatement)
```
achettyiitr, line 647: Regex replaces here. We should not logs the SQL statement.

warehouse/clickhouse/clickhouse.go
```diff
@@ -123,24 +115,37 @@ var clickhouseDataTypesMapToRudder = map[string]string{
 	"SimpleAggregateFunction(anyLast, Nullable(Int16))":    "int",
 	"SimpleAggregateFunction(anyLast, Nullable(Int32))":    "int",
 	"SimpleAggregateFunction(anyLast, Nullable(Int64))":    "int",
-	"SimpleAggregateFunction(anyLast, Nullable(Floats32))": "float",
-	"SimpleAggregateFunction(anyLast, Nullable(Floats64))": "float",
+	"SimpleAggregateFunction(anyLast, Nullable(Float32))":  "float",
```
achettyiitr, line 118: This seems to be a bug caught during writing Unit tests. Because of this, during the fetching of the schema, these are part of the unrecognized schema.
Data has went to rudder discards instead of their corresponding tables.

warehouse/clickhouse/clickhouse.go
```diff
@@ -196,32 +201,31 @@ func (ch *HandleT) newClickHouseStat(tableName string) *clickHouseStatT {
...
-func Connect(cred CredentialsT, includeDBInConn bool) (*sql.DB, error) {
+// ConnectToClickhouse connects to clickhouse with provided credentials
+func (ch *Handle) ConnectToClickhouse(cred Credentials, includeDBInConn bool) (*sql.DB, error) {
 	var dbNameParam string
 	if includeDBInConn {
 		dbNameParam = fmt.Sprintf(`database=%s`, cred.DBName)
 	}
 
-	url := fmt.Sprintf("tcp://%s:%s?&username=%s&password=%s&block_size=%s&pool_size=%s&debug=%s&secure=%s&skip_verify=%s&tls_config=%s&%s&read_timeout=%s&write_timeout=%s&compress=%t",
+	url := fmt.Sprintf("tcp://%s:%s?&username=%s&password=%s&block_size=%s&pool_size=%s&debug=%s&secure=%s&skip_verify=%s&tls_config=%s&%s&read_timeout=%s&write_timeout=%s&Compress=%t",
```
lvrach, line 214: Is there a reason for renaming `compress` to `Compress`?

achettyiitr: It happened by mistake. Thanks for pointing that out. reverting this.

lvrach: We could use `url.Values` for constructing the URI, it would be more readable and we won't have any issue with unescaped values.

warehouse/slave.go
```diff
@@ -485,6 +485,31 @@ func processStagingFile(job Payload, workerIndex int) (loadFileUploadOutputs []l
 			columnType := columnInfo.Type
 			columnVal := columnInfo.Value
 
+			if job.DestinationType == warehouseutils.CLICKHOUSE && model.SchemaType(columnType) == model.BooleanDataType {
```
lvrach, line 488: Did we discover this during testing?

warehouse/clickhouse/clickhouse.go
lvrach: Do you think we could use prepared statements here? For:
```
		loadFolder,                     // 4
		accessKeyID,                    // 5
		secretAccessKey,                // 6
		sortedColumnNamesWithDataTypes, // 7
		```

File: warehouse/clickhouse/clickhouse.go [hunk 6]
```diff
@@ -562,52 +599,124 @@ func (ch *HandleT) loadTable(tableName string, tableSchemaInUpload warehouseutil
...
+		return fmt.Sprintf(`%s %s`, name, rudderDataTypesMapToClickHouse[tableSchemaInUpload[name]])
+	}, ",")
+
+	csvObjectLocation, err := ch.Uploader.GetSampleLoadFileLocation(tableName)
+	if err != nil {
+		err = fmt.Errorf("failed to get sample load file location with error: %s", err.Error())
+		return
+	}
+	loadFolder := csvObjectLocation
+	loadFolder = loadFolder[:strings.LastIndex(loadFolder, "/")]
```
lvrach, line 628: Could this be helpful https://pkg.go.dev/path#Base ?

warehouse/clickhouse/clickhouse.go
```diff
@@ -123,24 +115,37 @@ var clickhouseDataTypesMapToRudder = map[string]string{
...
-type HandleT struct {
-	Db             *sql.DB
-	Namespace      string
-	ObjectStorage  string
-	Warehouse      warehouseutils.Warehouse
-	Uploader       warehouseutils.UploaderI
-	stats          stats.Stats
-	ConnectTimeout time.Duration
+type Handle struct {
+	Db                          *sql.DB
```
lvrach, line 126:
```suggestion
	DB                          *sql.DB
```

warehouse/slave.go
achettyiitr: We are mapping boolean to UInt8 datatype in clickhouse. In our previous load table strategy, when we try to insert the data, we used to convert true/false to 0/1 in [here](https://github.com/rudderlabs/rudder-server/blob/b7aa80003e237ddda370dff125898d4ac52ea820/warehouse/clickhouse/clickhouse.go#L504).

achettyiitr: Since now we are using COPY command which does the part of loading the data, We want to make sure that the load files generated are compatible with the table schema.

achettyiitr: We are adding this check here to  convert true/false to 0/1 directly in the load files itself.

achettyiitr: For the old load strategy, we are not handling both cases.
1. true/false
2. 0/1

warehouse/clickhouse/clickhouse.go
```diff
@@ -123,24 +117,37 @@ var clickhouseDataTypesMapToRudder = map[string]string{
...
 
-type HandleT struct {
-	Db             *sql.DB
-	Namespace      string
-	ObjectStorage  string
-	Warehouse      warehouseutils.Warehouse
-	Uploader       warehouseutils.UploaderI
-	stats          stats.Stats
-	ConnectTimeout time.Duration
+type Handle struct {
```
lvrach, line 127: [minor] We can rename Handle to Clickhouse. I think is more apt.

lvrach (approved):

warehouse/clickhouse/clickhouse.go
```diff
@@ -526,9 +554,26 @@ func typecastDataFromType(data, dataType string) interface{} {
...
+}
+
+func (ch *Clickhouse) UseS3CopyEngineForLoading() bool {
+	if !slices.Contains(ch.S3EngineEnabledWorkspaceIDs, ch.Warehouse.WorkspaceID) {
+		return false
+	}
+	if ch.ObjectStorage != warehouseutils.S3 && ch.ObjectStorage != warehouseutils.MINIO {
+		return false
+	}
+	return true
```
abhimanyubabbar, line 571: return `ch.ObjectStorage != warehouseutils.S3 && ch.ObjectStorage != warehouseutils.MINIO`

warehouse/slave.go
abhimanyubabbar, line 488: We can rewrite this simply by one if and switch clause for brevity. Wdyt ?

warehouse/slave.go
```diff
@@ -485,6 +485,31 @@ func processStagingFile(job Payload, workerIndex int) (loadFileUploadOutputs []l
...
+				newColumnVal := 0
+				if k, ok := columnVal.(bool); ok {
+					if k {
+						newColumnVal = 1
+					}
+				}
```
abhimanyubabbar, line 494: Should we add case where the ok is false ? Or would there ever be case like that ?

achettyiitr: We are already capturing this in the testdata we have..

abhimanyubabbar (approved):
LGTM :shipit:

PR #2741, opened 2022-11-28
feat: add support for request payloads using gzip content-encoding

atzoum:
# Description

Adding support for accepting requests using a gzipped body along with the appropriate `Content-Encoding: gzip` header.
Compressed request bodies are being uncompressed lazily and in a streaming fashion.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=e46c493fa108495cb6b9c85ad7594409&pm=s)

## Security

### DIFF
--- a/middleware/uncompress.go
+++ b/middleware/uncompress.go
@@ -1,0 +1,44 @@
+package middleware
+
+import (
+	"compress/gzip"
+	"io"
+	"net/http"
+)
+
+// UncompressMiddleware uncompresses gzipped HTTP requests carrying a 'Content-Encoding: gzip' header.
+var UncompressMiddleware = func(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Header.Get("Content-Encoding") == "gzip" {
+			r.Body = &gzipReader{body: r.Body}
+		}
+		h.ServeHTTP(w, r)
+	})
+}
+
+// gzipReader wraps a response body so it can lazily
+// call gzip.NewReader on the first call to Read
+type gzipReader struct {
+	body io.ReadCloser // underlying request body
+	zr   *gzip.Reader  // lazily-initialized gzip reader
+	zerr error         // any error from gzip.NewReader; sticky
+}
+
+func (gz *gzipReader) Read(p []byte) (n int, err error) {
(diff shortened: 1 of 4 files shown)


BonapartePC (approved):

middleware/uncompress.go
```diff
@@ -0,0 +1,44 @@
...
+var UncompressMiddleware = func(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Header.Get("Content-Encoding") == "gzip" {
+			r.Body = &gzipReader{body: r.Body}
+		}
+		h.ServeHTTP(w, r)
+	})
+}
+
+// gzipReader wraps a response body so it can lazily
```
chandumlg, line 19: wraps a response body or request body?

middleware/uncompress.go
```diff
@@ -0,0 +1,44 @@
...
+// call gzip.NewReader on the first call to Read
+type gzipReader struct {
+	body io.ReadCloser // underlying request body
+	zr   *gzip.Reader  // lazily-initialized gzip reader
+	zerr error         // any error from gzip.NewReader; sticky
+}
+
+func (gz *gzipReader) Read(p []byte) (n int, err error) {
+	if gz.zr == nil {
```
chandumlg, line 28: Should we also check if gz.zerr ! = nil also return immediately?

middleware/uncompress.go
atzoum: ~~response~~ body :)

middleware/uncompress.go
atzoum: Borrowed implementation from `net/http/transport` [link](https://go.dev/src/net/http/transport.go?s=13180:13237#:~:text=func%20(gz%20*gzipReader)%20Read(p%20%5B%5Dbyte)%20(n%20int%2C%20err%20error)%20%7B)

Found it smart, since you only need to check for `zerr` once. If we check for `zerr` early then we need to do it again after `gzip.NewReader`.

The only modification I did in transport's gzipReader was to close the `gzip.Reader` if not nil during `Close` method.

chandumlg (approved):

PR #2744, opened 2022-11-29
fix(destination): delete users support add test-cases and better error strings

sanpj2292:
# Description

- We are adding some test-cases
- We are also providing more information in error logs

## Notion Ticket

https://www.notion.so/rudderstacks/Integrate-OAuth-in-regulation-worker-module-of-server-to-support-GA-dae0cf6ef89e41eb91e9212ac524cf49

## Security


regulation-worker/internal/delete/api/api.go
```diff
@@ -198,7 +199,7 @@ func setOAuthHeader(secretToken *oauth.AuthResponse, req *http.Request) error {
 func (api *APIManager) getOAuthDetail(destDetail *model.Destination, workspaceId string) (oauthDetail, error) {
 	id := oauth.GetAccountId(destDetail.Config, oauth.DeleteAccountIdKey)
 	if strings.TrimSpace(id) == "" {
-		return oauthDetail{}, fmt.Errorf("%v is not present for %v", oauth.DeleteAccountIdKey, destDetail.Name)
+		return oauthDetail{}, fmt.Errorf("[%v] %v is not present for destination: %v", destDetail.Name, oauth.DeleteAccountIdKey, destDetail.DestinationID)
```
saikumarrs, line 202:
```suggestion
		return oauthDetail{}, fmt.Errorf("[%v] Delete account ID key (%v) is not present for destination: %v", destDetail.Name, oauth.DeleteAccountIdKey, destDetail.DestinationID)
```

saikumarrs: For more clarity.

saikumarrs (approved):

regulation-worker/internal/delete/api/api.go
```diff
@@ -223,9 +224,9 @@ func (api *APIManager) refreshOAuthToken(destName, workspaceId string, oAuthDeta
 		DestDefName:     destName,
 		EventNamePrefix: "refresh_token",
 	}
-	statusCode, _ := api.OAuth.RefreshToken(refTokenParams)
+	statusCode, refreshResponse := api.OAuth.RefreshToken(refTokenParams)
```
atzoum, line 227: `refreshResponse` is a pointer, can't it be `nil`?

sanpj2292: We are some or other response through `RefreshToken`. Except in one-case where while we're trying to marshal to send request to control-plane, we are panicking there. [Here is the source-code reference](https://github.com/rudderlabs/rudder-server/blob/fix.delete-users-oauth/services/oauth/oauth.go#L298-L301)

atzoum: would it be too much to add a nil check here?

sanpj2292: We can add it, not a problem 👍🏼

sanpj2292: Added it

regulation-worker/internal/delete/api/api.go

regulation-worker/internal/delete/api/api.go
```diff
@@ -223,9 +224,12 @@ func (api *APIManager) refreshOAuthToken(destName, workspaceId string, oAuthDeta
...
+	if refreshResponse == nil {
+		return fmt.Errorf("[%v] Failed to refresh token for destination in workspace(%v) & account(%v) due to `nil` response", destName, workspaceId, oAuthDetail.id)
+	}
 	if statusCode != http.StatusOK {
-		return fmt.Errorf("failed to refresh token for destination: %v", destName)
+		return fmt.Errorf("[%v] Failed to refresh token for destination in workspace(%v) & account(%v) with %v", destName, workspaceId, oAuthDetail.id, refreshResponse.Err)
 	}
```
atzoum, line 233: we shouldn't skip the `statusCode` check, right?
```
statusCode, refreshResponse := api.OAuth.RefreshToken(refTokenParams)
if statusCode != http.StatusOK {
	var responseErr string
	if refreshResponse != nil {
		responseErr = refreshResponse.Err
	}
	return fmt.Errorf("[%v] Failed to refresh token for destination in workspace(%v) & account(%v) with %v", destName, workspaceId, oAuthDetail.id, responseErr)
}
```

sanpj2292: Yeah, makes sense! Made the change

atzoum (approved):

PR #2747, opened 2022-11-29
feat: controlplane client support for destination history

lvrach:
# Description

## Motivation

Data Residency epic required a destination history endpoint to include the region the database is running on, when the region is available.

While looking into the existing implementation, I found that the multiple tenant/namespace environment was not adequately supported and working by accident. 

I also want to move all control-plane API-related calls inside the client component.

## Changes

I have merged the feature endpoint & destination history under a single client.

## Notion Ticket

https://www.notion.so/rudderstacks/Region-specific-destinationHistory-request-7385ef5d42f8439482d048dd1ba3adfb

## Security


services/controlplane/client/client_test.go
```diff
@@ -0,0 +1,329 @@
...
+			name: "DestinationHistory",
+			fn: func(c *client.Client) error {
+				_, err := c.DestinationHistory(context.Background(), "test")
+				return err
+			},
+		},
+	}
+
+	for _, m := range methods {
+		t.Run(m.name, func(t *testing.T) {
```
achettyiitr, line 258: We would need to capture the [range variable](https://go.dev/blog/subtests#:~:text=capture%20range%20variable) here.

services/controlplane/client/client_test.go
```diff
@@ -0,0 +1,329 @@
...
+				}))
+				defer s.Close()
+
+				c := client.NewClient(s.URL, &identity.Namespace{},
+					client.WithHTTPClient(s.Client()),
+					client.WithMaxRetries(maxRetries),
+				)
+
+				err := m.fn(c)
+				require.ErrorContains(t, err, "")
```
achettyiitr, line 301: This will always pass right.

achettyiitr:
Let's use the `destinationHistory` endpoint?
https://github.com/rudderlabs/rudder-server/blob/1ba319b973efd39dfafad9e80a2bcd2af6dd79e6/warehouse/upload.go#L1786

achettyiitr:
Let's use the endpoint?
https://github.com/rudderlabs/rudder-server/blob/1ba319b973efd39dfafad9e80a2bcd2af6dd79e6/warehouse/upload.go#L1786

services/controlplane/client/client_test.go
```diff
@@ -0,0 +1,329 @@
...
```
achettyiitr, line 278: This will always pass right.

services/controlplane/client/client_test.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	testcases := []struct {
+		name       string
+		identity   identity.Identifier
+		region     string
+		revisionID string
+
+		responseBody   string
+		responseStatus int
```
achettyiitr, line 118: We are not using this in any test case. Are we planning to use this for later purposes?

services/controlplane/client/client_test.go
```diff
@@ -0,0 +1,329 @@
...
+			identity: &identity.Workspace{
+				WorkspaceID:    "valid-workspace-id",
+				WorkspaceToken: validSecret,
+			},
+			revisionID:   "2ENkYQ1MR9f83YhqS5k7uTSe5XH",
+			responseBody: string(body),
+
+			wantDestination: expectedDestination,
+			wantPath:        "/workspaces/destinationHistory/2ENkYQ1MR9f83YhqS5k7uTSe5XH",
+			wantErr:         nil,
```
achettyiitr, line 135: Do we need to set this specifically to nil? It will be by default nil right.

achettyiitr (approved):

abhimanyubabbar (approved):
Minor nit.

abhimanyubabbar:
Should we put the client in warehouse module ? @lvrach
Going forward, we might wanna keep the development cadence and codebase of warehouse separate than server. Should we think about moving the controlplane as a client inside warehouse ? Wdyt ?

runner/runner.go
```diff
@@ -43,7 +43,7 @@ import (
 	batchrouterutils "github.com/rudderlabs/rudder-server/router/utils"
 	"github.com/rudderlabs/rudder-server/services/alert"
 	"github.com/rudderlabs/rudder-server/services/archiver"
-	"github.com/rudderlabs/rudder-server/services/controlplane/features"
+	"github.com/rudderlabs/rudder-server/services/controlplane/client"
```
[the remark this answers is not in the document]
lvrach: I want to avoid `cp` repetition in the import path. 

Having a package named just `client` is not uncommon in go. I believe it is ok since it is under the context of `controlplane`. A similar pattern is used by the [etcd client](https://pkg.go.dev/go.etcd.io/etcd/client/v3) as well.

PR #2764, opened 2022-12-01
chore: upgrade postgres image

cisse21:
# Description

Upgrading Postgres to 15 in the default docker-compose file and the relevant tests

## Notion Ticket

[ Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=97b29b8fee654a52ac58812513c06ff6&pm=s)

## Security

### DIFF
--- a/app/cluster/integration_test.go
+++ b/app/cluster/integration_test.go
@@ -66,7 +66,7 @@ func run(m *testing.M) int {
 
 	database := "jobsdb"
 	// pulls an image, creates a container based on it and runs it
-	resourcePostgres, err := pool.Run("postgres", "11-alpine", []string{
+	resourcePostgres, err := pool.Run("postgres", "15-alpine", []string{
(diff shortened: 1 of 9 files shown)


achettyiitr (approved):

lvrach (approved):

app/cluster/integration_test.go
```diff
@@ -66,7 +66,7 @@ func run(m *testing.M) int {
 
 	database := "jobsdb"
 	// pulls an image, creates a container based on it and runs it
-	resourcePostgres, err := pool.Run("postgres", "11-alpine", []string{
+	resourcePostgres, err := pool.Run("postgres", "15-alpine", []string{
```
atzoum, line 69: Since we are still using postgres-11 in various environments, shouldn't we leave it as is until we upgrade them first?

PR #2767, opened 2022-12-04
chore(warehouse): added support to fetch the ssh keys on destination level from control plane

abhimanyubabbar:
# Description
We have introduced an internal endpoint in control plane to allow for fetching the public private keypair at the destination level. As part of the change:

1. **Controlplane API Client**: We added a client for control plane API and also a cached client for the same. The purpose of the cached client is to prevent us to make network call everytime to fetch the secrets as it doesn't change that frequently.

2. **Warehouse Destination Data Augmentation**: Augmented the information of destination object by adding private and public keys into the config.

## Notion Ticket

https://www.notion.so/rudderstacks/Add-logic-to-fetch-keypair-from-control-plane-in-warehouse-while-fetching-backend-config-9e66144e05944ddfb26f782e7534c7a4

## Security

### DIFF
--- a/config/backend-config/types.go
+++ b/config/backend-config/types.go
@@ -52,6 +52,12 @@ type DestinationT struct {
 	Transformations       []TransformationT
 	IsProcessorEnabled    bool
 	RevisionID            string
+	Tunnel                *Tunnelling
+}
+
+type Tunnelling struct {
+	Type   string
+	Config map[string]string
 }
 
 type SourceT struct {
--- a/warehouse/client/controlplane/client.go
+++ b/warehouse/client/controlplane/client.go
@@ -1,0 +1,99 @@
+package controlplane
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+type PublicPrivateKeyPair struct {
+	PublicKey  string
+	PrivateKey string
+}
+
+type BasicAuth struct {
+	Username string
+	Password string
+}
+
+type CPInternalAPI interface {
+	GetDestinationSSHKeys(ctx context.Context, id string) (*PublicPrivateKeyPair, error)
+}
+
+type cachedControlPlaneAPIImpl struct {
+	cpAPI CPInternalAPI
+	cache map[string]*PublicPrivateKeyPair
+}
+
+func NewControlPlaneAPI(baseURI string, auth BasicAuth) CPInternalAPI {
+	return &controlPlaneAPIImpl{
+		baseURI: baseURI,
+		auth:    auth,
+	}
+}
+
+type controlPlaneAPIImpl struct {
+	baseURI string
+	auth    BasicAuth
+}
+
+func (api *controlPlaneAPIImpl) GetDestinationSSHKeys(ctx context.Context, id string) (*PublicPrivateKeyPair, error) {
+	url := fmt.Sprintf("%s/dataplane/admin/destinations/%s/sshKeys", api.baseURI, id)
+
+	req, err := http.NewRequestWithContext(ctx, "GET", url, http.NoBody)
+	if err != nil {
+		return nil, fmt.Errorf("creating new request with ctx: %w", err)
+	}
+
+	req.SetBasicAuth(
+		api.auth.Username,
+		api.auth.Password)
+
+	client := http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("fetching response from http client: %w", err)
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode == 404 {
+		return nil, nil
+	}
+
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("invalid status code: %d", resp.StatusCode)
+	}
+
+	decoder := json.NewDecoder(resp.Body)
+
+	keypair := PublicPrivateKeyPair{}
+	if err := decoder.Decode(&keypair); err != nil {
+		return nil, fmt.Errorf("decoding upstream response body: %w", err)
+	}
+
+	return &keypair, nil
+}
+
+func NewCachedControlPlaneAPI(baseURI string, auth BasicAuth) CPInternalAPI {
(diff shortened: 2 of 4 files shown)


config/backend-config/types.go
```diff
@@ -52,6 +52,12 @@ type DestinationT struct {
 	Transformations       []TransformationT
 	IsProcessorEnabled    bool
 	RevisionID            string
+	Tunnel                *Tunnelling
+}
+
+type Tunnelling struct {
+	Type   string
+	Config map[string]string
```
achettyiitr, line 60:
```suggestion
	Config map[string]interface
```

wdyt?

achettyiitr: Or if we are completely sure that config backend sends map[string]string then pls ignore this comment.

warehouse/client/controlplane/client.go
```diff
@@ -0,0 +1,99 @@
...
+
+func NewCachedControlPlaneAPI(baseURI string, auth BasicAuth) CPInternalAPI {
+	apiClient := NewControlPlaneAPI(baseURI, auth)
+	return &cachedControlPlaneAPIImpl{
+		cpAPI: apiClient,
+		cache: make(map[string]*PublicPrivateKeyPair),
+	}
+}
+
+func (api *cachedControlPlaneAPIImpl) GetDestinationSSHKeys(ctx context.Context, id string) (*PublicPrivateKeyPair, error) {
```
achettyiitr, line 87: Let's add test case for this?

warehouse/client/controlplane/client.go
```diff
@@ -0,0 +1,99 @@
...
+	if val, ok := api.cache[id]; ok {
```
achettyiitr, line 88: Should we set the expiry on the cache as well?

warehouse/warehouse.go
```diff
@@ -1231,6 +1254,12 @@ func (wh *HandleT) Setup(whType string) {
 	config.RegisterIntConfigVariable(1, &wh.maxConcurrentUploadJobs, false, 1, fmt.Sprintf(`Warehouse.%v.maxConcurrentUploadJobs`, whName))
 	config.RegisterBoolConfigVariable(false, &wh.allowMultipleSourcesForJobsPickup, false, fmt.Sprintf(`Warehouse.%v.allowMultipleSourcesForJobsPickup`, whName))
 
+	wh.internalAuth = cpclient.BasicAuth{
+		Username: config.GetString("INTERNAL_AUTH_USERNAME", "username"),
```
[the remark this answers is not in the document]
achettyiitr: Do we need to use `RUDDER_ADMIN_USER` and `RUDDER_ADMIN_PASSWORD`?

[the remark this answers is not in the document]
abhimanyubabbar: These are credentials necessary for interacting with the internal api.

warehouse/client/controlplane/client.go
abhimanyubabbar: The keys don't expire in general sense unless we have a breach leaking the keys. So the TTL doesn't make much sense here. Wdyt ?

warehouse/client/controlplane/client.go
```diff
@@ -0,0 +1,99 @@
...
+	req, err := http.NewRequestWithContext(ctx, "GET", url, http.NoBody)
+	if err != nil {
+		return nil, fmt.Errorf("creating new request with ctx: %w", err)
+	}
+
+	req.SetBasicAuth(
+		api.auth.Username,
+		api.auth.Password)
+
+	client := http.Client{}
```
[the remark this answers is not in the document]
abhimanyubabbar: I'll add a timeout for this.

config/backend-config/types.go
abhimanyubabbar: At this moment, to keep the struct fields tight, we are only capturing values as string. Tomorrow in case the DS changes, we will capture it through interface{}

warehouse/warehouse.go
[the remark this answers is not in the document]
abhimanyubabbar: We will be generating another set of credentials to interact with the API.

warehouse/client/controlplane/client.go
achettyiitr: In that case, restarting would be sufficient I think.

abhimanyubabbar: yes

warehouse/client/controlplane/client.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+}
+
+type controlPlaneAPIImpl struct {
+	baseURI string
+	auth    BasicAuth
+}
+
+func (api *controlPlaneAPIImpl) GetDestinationSSHKeys(ctx context.Context, id string) (*PublicPrivateKeyPair, error) {
+	url := fmt.Sprintf("%s/dataplane/admin/destinations/%s/sshKeys", api.baseURI, id)
```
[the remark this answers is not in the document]
abhimanyubabbar: Not needed for now. Confirmed with Fotis. Resolving this for now.

warehouse/client/controlplane/client.go
```diff
@@ -0,0 +1,99 @@
...
+type cachedControlPlaneAPIImpl struct {
+	cpAPI CPInternalAPI
+	cache map[string]*PublicPrivateKeyPair
+}
+
+func NewControlPlaneAPI(baseURI string, auth BasicAuth) CPInternalAPI {
+	return &controlPlaneAPIImpl{
+		baseURI: baseURI,
+		auth:    auth,
+	}
```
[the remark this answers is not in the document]
abhimanyubabbar: Done 👍🏼

warehouse/client/controlplane/client.go
```diff
@@ -0,0 +1,99 @@
+package controlplane
```
lvrach, line 1: Would it make sense to move the client [here](https://github.com/rudderlabs/rudder-server/tree/master/services/controlplane)

PR #2795, opened 2022-12-10
chore: migrate stats to otel

fracasula:
# Description

First PR to migrate towards OpenTelemetry.

The scope of this PR is:
* offer an initial `otel` package to help us avoid most of the boilerplate during the bootstrap phases
* replace statsd in the `stats` package with `otel`

If possible I'm trying not to change the metrics to avoid reworking Grafana dashboards as much as possible. Unfortunately StatsD timers were converted into summaries while now we have to use histograms. I couldn't find a way to avoid that change so we'll have to update a few dashboards.

**BEWARE** that I'm using a collector here. This means that RudderServer will offload traces and metrics as soon as possible to an OTel collector. The configuration about what to do with those traces and metrics will be in the OTel collector and **NOT** in RudderServer (see `otel-collector-config.yaml` used in the `otel_test.go`). This means that Prometheus will scrape from the OTel collectors and not from the Statsd exporter sidecar or RudderServer.

### Important notes
* I tested the metrics on `dev` and the `.` get converted to `_` (e.g. `gateway.my_metric` = `gateway_my_metric`). The metrics also maintained their type.
* the infra side is still being worked on by @BonapartePC 
   * the idea for now is to have a bunch of OTel collectors that are horizontally scalable deployed in each namespace behind a load balancer (no more sidecars)
   * OTel collector [example](https://github.com/open-telemetry/opentelemetry-go/tree/main/example/otel-collector) for reference
 * some stuff in the operator to make this work: https://github.com/rudderlabs/rudderstack-operator/blob/9a3214b569b2623c8654f7b4aae5c24ac267baaf/otel-collector-values.yaml

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=23d2a92cf290475c9c2b91cd1fc7ec1e&pm=s) >

## TODO
- [x] check that the **ALL** metrics names are the same as before
- [x] check that a statsd timer gets converted to an histogram in the statsd exporter

## Security


internal/otel/otel.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	if !c.tracerProviderConfig.enabled && !c.meterProviderConfig.enabled {
+		return nil, nil, fmt.Errorf("no trace provider or meter provider to initialize")
+	}
+
+	conn, err := grpc.DialContext(ctx, target,
+		append(c.dialOpts, grpc.WithBlock())...,
+	)
+	if err != nil {
```
koladilip, line 43: on error, should we close the `conn`?

koladilip (commented):
overall looks good, what is our strategy to test this?

internal/otel/otel.go
fracasula: https://github.com/grpc/grpc-go/blob/v1.51.0/clientconn.go#L160

Additionally, if you look at the rest of the function, when `err` is not `nil` then `conn` is `nil` so a `conn.Close` would give us a panic.

fracasula:
> overall looks good, what is our strategy to test this?

I expect the majority of metrics not to change. As I wrote in the PR description we're lucky that Statsd timers get converted to Histograms already.

Nonetheless checking that every single metric did not change can be quite cumbersome, so my idea is the following:
1. deploy this branch (i.e. `chore.otel`) on dev along with an OTel collector. have something maybe from @pChondros that runs in order to generate all metrics
2. deploy master branch on another namespace in parallel and have @pChondros stuff hit that namespace as well
3. do a text diff between the collector `/metrics` and the statsd exporter `/metrics`

Main blockers to do so:
1. on dev we don't have the statsd exporter but by looking at the rudderstack object on k8s it might be fairly easy to add
2. have a script that hits the gateway (in this case) in all possible ways (I'm talking with @pChondros about this) still requires work although it might turn out to be useful for other scenarios

Suggestions are welcome.

internal/otel/otel.go
koladilip: here yes but after that we are using it right and error might occur there right then what happens to conn?

fracasula: Oh I get what you mean. I'll look into it, perhaps expose a `Shutdown` method :+1:

internal/otel/otel.go
```diff
@@ -0,0 +1,146 @@
+package otel
```
lvrach, line 1: Why we have put this under internal, would this component be useful for other projects as well?

fracasula: True. Although I'm a fan of starting with everything private and opening it up as needed this might be a good candidate for making it public from the start.

atzoum: This doesn't seem to have any dependency on rudder-server's other packages. Would it make sense to use a separate repository for it so that other projects don't have to depend on rudder-server?

fracasula: A separate repo just for this I'd say it is overkilling. Unless we're planning to do a collection of these "utils". At Netdata we used to have a versioned `cloud-kit` repository with all these little independent packages used by different µservices and it was quite useful.

runner/runner.go
```diff
@@ -144,7 +148,48 @@ func (r *Runner) Run(ctx context.Context, args []string) int {
 	// application & backend setup should be done before starting any new goroutines.
 	r.application.Setup()
 
-	var err error
+	// START - OpenTelemetry setup
+	otelEndpoint := config.GetString("OpenTelemetry.Endpoint", "") // @TODO check these env vars with infra
+	otelMetricsExportsInterval := config.GetDuration("OpenTelemetry.MetricsExportInterval", 10, time.Second)
+	if otelEndpoint == "" {
+		r.logger.Errorf("OpenTelemetry endpoint not provided")
+		return 1
```
fracasula, line 156: This is making some tests fail, we could skip the OpenTelemetry setup altogether if an endpoint is not provided. If provided an we can't connect then we fail. I'll make the changes but I'll leave the comment here for visibility.

fracasula:
Memo to explore idea to turn the stats package into an adapter.

fracasula:
`stats` package refactoring done. 

**TODO:**
* test existing metrics
* remove `stats.Start()` and `stats.Stop()` occurrences from everywhere

gateway/gateway.go
```diff
@@ -158,16 +164,28 @@ type HandleT struct {
...
-	requestSizeStat                               stats.Measurement
-	dbWritesStat                                  stats.Measurement
-	dbWorkersBufferFullStat, dbWorkersTimeOutStat stats.Measurement
-	bodyReadTimeStat                              stats.Measurement
-	addToWebRequestQWaitTime                      stats.Measurement
-	processRequestTime                            stats.Measurement
-	emptyAnonIdHeaderStat                         stats.Measurement
-	addToBatchRequestQWaitTime                    stats.Measurement
+	// gateway metrics
+	stats                                         metric.MeterProvider
```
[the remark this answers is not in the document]
fracasula: Quick update: the statsd sidecars will be removed and Prometheus will scrape directly from the collectors.
The collectors are going to be deployed in a horizontally scaled manner behind a single load balancer and we'll have one setup per namespace so no more sidecars. @BonapartePC is working on this, we'll tests this setup with plenty of metrics to tune the memory limits of the collectors.

Outstanding tasks:
1. change summaries (old statsd timers) to histograms in the Grafana dashboards (more info below)
2. finish testing all dashboards
3. stress tests collectors once the setup by @BonapartePC is ready

### Summaries changes

For example we were querying summaries like this:
```promql
max by (quantile) (gateway_response_time{quantile =~ "0.9|0.99", instance =~ "($coreInstance).*", reqType =~ "/v1/.*"} > 0)
```

We could rewrite those queries as:
```promql
histogram_quantile(0.95, sum(rate(gateway_response_time_bucket{instanceName =~ "($coreInstance).*", reqType =~ "/v1/.*"}[5m])) by (le))
```

Example [here](https://github.com/rudderlabs/rudderstack-operator/blob/4436074fbfdf0cb838f59db85e02181e8119b72a/helm-charts/rudderstack-monitor/charts/grafana/dashboards/rudder-metrics-dashboard-prom.json#L605).

internal/otel/otel.go
fracasula: For now I moved the `otel` package in the root folder.

services/stats/stats.go
```diff
@@ -42,14 +38,8 @@ type Stats interface {
...
 	// NewSampledTaggedStat creates a tagged measurement with the configured sampling rate (statsSamplingRate)
+	// Deprecated: use NewTaggedStat instead
 	NewSampledTaggedStat(name, statType string, tags Tags) Measurement
-
-	// Start starts the stats service. It tries to connect to statsd server once and returns immediately.
-	// If it fails to connect, it will retry in the background indefinitely, however any stats published in the meantime will be lost.
-	Start(ctx context.Context)
-
-	// Stop stops the service and collection of periodic stats.
-	Stop()
```
atzoum, line 52: It might be safer to follow a phased strategy like below:

1. Keep the old `statsd` implementation and keep using it as the default `stats` implementation.
2. Use a configuration option for being able to switch to the `otel` implementation so that we can test and verify it in a production environment.
3. As we become more confident we can make the `otel` implementation the default one and deprecate `statsd`.
4. Finally, remove support for statsd 
   - delete implementation
   - replace docker container with otel in docker compose file(s)
   - remove statsd environment variables etc.

atzoum (commented):
I might be repeating my [previous comment](https://github.com/rudderlabs/rudder-server/pull/2795#discussion_r1088714717), but I was under the impression that we were ultimately going to keep both `otel` & `statsd` implementations available as a first step so that we are in a better position to verify and rollout this major change safely and in phases & also have a way to rollback if needed.

If that's not the case, are we moving towards rudder-server `v2.0.0`? I still have some reservations with respect to how we are going to rollout such a change.

fracasula:
> I might be repeating my [previous comment](https://github.com/rudderlabs/rudder-server/pull/2795#discussion_r1088714717), but I was under the impression that we were ultimately going to keep both `otel` & `statsd` implementations available as a first step so that we are in a better position to verify and rollout this major change safely and in phases & also have a way to rollback if needed.
> 
> If that's not the case, are we moving towards rudder-server `v2.0.0`? I still have some reservations with respect to how we are going to rollout such a change.

The plan is not to merge right away these changes (see label "do not merge") but to create a dedicated release for it and roll it out incrementally. We'll start with one customer only at first.

In regards to the operator, we'll keep a version (helm chart) for OSS customers that want to keep Statsd but at some point (once this is merged) they'll have to start using OTel if they want to upgrade.

To roll back a specific customer consider that once we cut a beta release with OTel there will be one that is exactly the same but without OTel, so we'll have to option to switch to that. Of course the operator version also needs to be switched. The rest of the pipeline (Prometheus/Grafana) stays together with the exception that some metrics (timers only) will be lost for the period where OTel was running. This is because timers with statsd were summaries but now we're converting them to histograms. So the data will be there but once we switch the operator version the dashboard will have the summaries in their promql. If needed  they can be changed on the fly to read from the histograms as well, although not very convenient.

atzoum:
> To roll back a specific customer consider that once we cut a beta release with OTel there will be one that is exactly the same but without OTel, so we'll have to option to switch to that

Please take into consideration the overhead (and risk for human error) of having to maintain and synchronise two branches for as long as we will be rolling out the otel changes.

regulation-worker/cmd/main.go
```diff
@@ -51,8 +56,59 @@ func Run(ctx context.Context) error {
 	misc.Init()
 	diagnostics.Init()
 	backendconfig.Init()
-
 	stats.Default.Start(ctx)
+	defer stats.Default.Stop()
+
+	// START - OpenTelemetry setup
```
atzoum, line 62: **Note:** We are using rudder-server's `stats` in [data-regulation-service](https://github.com/rudderlabs/data-regulation-service/blob/master/cmd/service/main.go#L36) too. This would mean that in the future, once the dependency version is bumped we will need to adapt the agent stack there as well.

cc @BonapartePC @saurav-malani

regulation-worker/cmd/main.go
```diff
@@ -51,8 +56,59 @@ func Run(ctx context.Context) error {
...
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			if err := otelManager.Shutdown(ctx); err != nil {
+				pkgLogger.Warnf("OpenTelemetry shutdown failed: %v", err)
+			}
+		}()
+	} else {
+		pkgLogger.Warnf("Stats are disabled")
+	}
+	// END - OpenTelemetry setup
```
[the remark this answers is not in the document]
fracasula: I'm OK with the package name (i.e. `stats`) for now.
The reason I didn't include the boilerplate of setting up a connection there is that I'm used to do dependency injection in `main` and the `stats` package was already doing a lot. If desired I can move this logic into `stats`, the main change it'll probably be `Start()` returning an `error` (in case `enableStats=true` but no connection could be established within the timeout).

runner/runner.go
```diff
@@ -140,6 +140,56 @@ func (r *Runner) Run(ctx context.Context, args []string) int {
...
+				otel.WithGlobalMeterProvider(),
+				otel.WithMeterProviderExportsInterval(
+					config.GetDuration("OpenTelemetry.Metrics.ExportInterval", 5, time.Second),
+				),
+			))
+		}
+		if _, _, err = otelManager.Setup(ctx, res, otelOpts...); err != nil {
+			r.logger.Errorf("OpenTelemetry setup failed: %v", err)
+			return 1
+		}
```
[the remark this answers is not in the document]
fracasula: Yes that is correct. That is a blocking call that retries a few times and returns an error if a connection cannot be established. I can make it optional if preferred and have RS start anyway. To avoid bloating the server memory I'll have to skip everything until we can connect.

services/stats/tags.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	return res
+}
+
+// String returns all key value pairs as a single string, separated by commas, sorted by increasing key order
+func (t Tags) String() string {
+	return strings.Join(t.Strings(), ",")
+}
+
+// otelAttributes returns all key value pairs as an ordered list of OpenTelemetry attributes
```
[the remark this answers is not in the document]
fracasula: Good catch, the order is not important, will amend the comment :+1:

services/stats/measurement.go
```diff
@@ -42,182 +43,143 @@ type Measurement interface {
...
-	c.client.statsd.Increment(c.name)
 }
 
-// statsdGauge represents a gauge stat
-type statsdGauge struct {
-	*statsdMeasurement
+// otelGauge represents a gauge stat
+type otelGauge struct {
+	*otelMeasurement
+	values  []interface{}
```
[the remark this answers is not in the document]
fracasula: You're right we can just keep the latest :+1:

atzoum:
I shall give this one last try, since I still believe that it really wouldn't cost us anything to retain the old implementation alongside the new one... that's the advantage of using an interface after all.
By having both implementations available (by default `statsd` should be enabled) we could merge these changes early, enable `otel` selectively and address any potential bugs in the main branch, without us being burdened with the risky task to maintain and synchronise two separate branches for longer periods.

fracasula:
> I shall give this one last try, since I still believe that it really wouldn't cost us anything to retain the old implementation alongside the new one... that's the advantage of using an interface after all. By having both implementations available (by default `statsd` should be enabled) we could merge these changes early, enable `otel` selectively and address any potential bugs in the main branch, without us being burdened with the risky task of maintaining and synchronising two separate branches for longer periods.

Alright, I don't see a problem with this and we already have the statsd implementation. I will update accordingly, thanks :+1:

services/stats/tags.go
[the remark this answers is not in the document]
atzoum: the doc still appears to be intact

services/stats/testdata/otel-collector-config.yaml
```diff
@@ -0,0 +1,24 @@
+receivers:
```
atzoum, line 1: **Question:** is this an alternative way for configuring otel? Are we really using it?

atzoum (changes requested):

services/stats/statsd.go
```diff
@@ -0,0 +1,267 @@
...
+			s.state.clients[taggedClientKey] = taggedClient
+		}
+		s.state.clientsLock.Unlock()
+	}
+
+	return s.newStatsdMeasurement(s.logger, name, statType, taggedClient)
+}
+
+// newStatsdMeasurement creates a new measurement of the specific type
+func (s *statsdStats) newStatsdMeasurement(log logger.Logger, name, statType string, client *statsdClient) Measurement {
```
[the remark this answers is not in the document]
fracasula: Yes, I forgot thanks :+1:

runner/runner.go
```diff
@@ -169,10 +168,19 @@ func (r *Runner) Run(ctx context.Context, args []string) int {
...
 		config.Set("statsExcludedTags", []string{"workspaceId", "sourceID", "destId"})
 	}
-	stats.Default.Start(ctx)
+	stats.Default = stats.NewStats(config.Default, logger.Default, svcMetric.Instance,
+		stats.WithServiceName(r.appType),
+		stats.WithServiceVersion(r.releaseInfo.Version),
+		stats.WithNamespace(config.GetNamespaceIdentifier()),
+		stats.WithInstanceName(config.GetString("INSTANCE_ID", "")),
+	)
+	if err := stats.Default.Start(ctx); err != nil {
```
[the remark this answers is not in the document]
fracasula: There is a test for it i.e. `TestOTelStartStopError` in `/services/stats/otel_test.go`. 
In the test I'm not providing an endpoint which makes `Start()` return an error and then I invoke `Stop()` to see if it blocks or not. Let me know if you think more cases should be added.

services/stats/testdata/otel-collector-config.yaml
fracasula: That is the agent/collector configuration. That configuration that you see there in the `otel-collector-config.yaml` file is not what we'll use in production of course. However some things are similar like the receivers. We'll have a single receiver getting metrics/traces over the gRPC connection on 4317, which is this bit:

```yaml
receivers:
  otlp:
    protocols:
      grpc:
```

If that doesn't answer your question please let me know as I'm unsure about what you were asking exactly.

services/stats/periodic.go
```diff
@@ -55,7 +66,8 @@ func newRuntimeStatsCollector(gaugeFunc gaugeFunc) runtimeStatsCollector {
 // Run gathers statistics from package runtime and outputs them to the configured GaugeFunc every
 // PauseDur. This function will not return until Done has been closed (or never if Done is nil),
 // therefore it should be called in its own goroutine.
-func (c runtimeStatsCollector) run() {
+func (c runtimeStatsCollector) run(ctx context.Context) {
+	defer close(c.done)
```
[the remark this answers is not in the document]
fracasula: Yes, I forgot to adapt after I put back the statsd implementation. Fixed :+1:

runner/runner.go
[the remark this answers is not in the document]
atzoum: In case of an error does it retry to connect behind-the-scenes? If not, we should return and not start the server. Otherwise, it is fine to simply log an error and resume.

services/stats/testdata/otel-collector-config.yaml
atzoum: Thanks, I confused collectors with exporters for a while

atzoum (approved):
Thank you for keeping the statsd implementation as the default one!

runner/runner.go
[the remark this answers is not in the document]
fracasula: If it can't connect an error is not returned unless we block but we moved away from that implementation. That's how it works by design (see `grpc.DialContext`). This is a double edged-sword since we might not know if we managed to connect after all (maybe we pass the wrong port and it's not possible to establish a connection ever).

In order to mitigate this we are setting up alerts that will warn us if no metrics are being received. The collectors expose several useful metrics such as accepted metrics, refused metrics, etcetera that I think should be a good enough indicator of the health of our monitoring stack. At least for now.

If this will prove not to be sufficient, an alternative would be to manage the lifecycle of the connection ourselves, but before going down that road and burden us with the additional complexity, I think this apprach should be reliable enough and similar to what we already have in place.

On the other end, if an error is returned, it is because we passed something really wrong in terms of configuration (e.g. you enable OpenTelemetry but then you don't provide any metrics nor traces endpoint) so it might make sense to return. Configuration wise I never managed to get an actual error from the underlying OTel framework though but I think the same should apply there.

I'll make the change to the regulation worker as well :+1:

fracasula:
> Thank you for keeping the statsd implementation as the default one!

Thank you for all your feedback @atzoum! :+1:

lvrach (approved):

achettyiitr (approved):

PR #2803, opened 2022-12-13
chore: limit live events stored in memory

cisse21:
# Description
Clean up live events code and reduce the events stored in memory

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=d8e8dd1b53cc414598a6e39c045decb0&pm=s)

## Security


services/debugger/uploader.go
```diff
@@ -51,10 +52,11 @@ func init() {
 func (uploader *Uploader) Setup() {
 	// Number of events that are batched before sending events to control plane
 	config.RegisterIntConfigVariable(32, &uploader.maxBatchSize, true, 1, "Debugger.maxBatchSize")
-	config.RegisterIntConfigVariable(1024, &uploader.maxESQueueSize, true, 1, "Debugger.maxESQueueSize")
+	config.RegisterIntConfigVariable(256, &uploader.maxESQueueSize, true, 1, "Debugger.maxESQueueSize")
 	config.RegisterIntConfigVariable(3, &uploader.maxRetry, true, 1, "Debugger.maxRetry")
 	config.RegisterDurationConfigVariable(2, &uploader.batchTimeout, true, time.Second, "Debugger.batchTimeoutInS")
 	config.RegisterDurationConfigVariable(100, &uploader.retrySleep, true, time.Millisecond, "Debugger.retrySleepInMS")
+	uploader.timingStat = stats.Default.NewStat("debugger_timing_stat", stats.TimerType)
```
atzoum, line 59:
```suggestion
	uploader.timingStat = stats.Default.NewStat("debugger_upload", stats.TimerType)
```

services/debugger/uploader.go
```diff
@@ -126,13 +130,12 @@ func (uploader *Uploader) uploadEvents(eventBuffer []interface{}) {
 			// Refresh the connection
 			continue
 		}
-		defer func() { httputil.CloseResponse(resp) }()
+		resp.Body.Close()
```
atzoum, line 133: both read and close response body always
```suggestion
		httputil.CloseResponse(resp)
```

cisse21: Doing this didn't pass the linter and so had to modify it the current way

atzoum: the following should do the trick and I will eventually treat this issue with the new linter
```
func(){httputil.CloseResponse(resp)}()
```

services/debugger/source/eventUploader.go
```diff
@@ -76,10 +76,7 @@ func recordHistoricEvents(writeKeys []string) {
 	for _, writeKey := range writeKeys {
 		historicEvents := eventsCacheMap.ReadAndPopData(writeKey)
 		for _, eventBatchData := range historicEvents {
-			var eventBatch string
-			if err := json.Unmarshal(eventBatchData, &eventBatch); err != nil {
-				panic(err)
-			}
```
atzoum, line 82: not sure why this was done... for stripping doublequotes?

services/debugger/source/eventUploader.go
```diff
@@ -97,7 +94,7 @@ func RecordEvent(writeKey, eventBatch string) bool {
 	configSubscriberLock.RLock()
 	defer configSubscriberLock.RUnlock()
 	if !misc.Contains(uploadEnabledWriteKeys, writeKey) {
-		eventBatchData, _ := json.Marshal(eventBatch)
+		eventBatchData := []byte(eventBatch)
```
atzoum, line 97: what's going on here with bytes and strings? is there an easy explanation?

cisse21: It seemed redundant to do an unmarshal and marshal again for a byte array. I didn't think of any reason why could have this been done

atzoum: do we even need to convert string to bytes?

cisse21: `eventsCacheMap` accepts a byteArray. This is common across sources, destinations and transformations. At the remaining places, we are doing an marshal to store as a byte array

services/debugger/source/eventUploader.go
atzoum: pretty unnecessary what we did before!

services/debugger/source/eventUploader.go
```diff
@@ -107,13 +102,15 @@ func RecordEvent(writeKey, eventBatch string) bool {
 }
 
 func (*EventUploader) Transform(data interface{}) ([]byte, error) {
-	eventBuffer := data.([]interface{})
+	eventBuffer, ok := data.([]GatewayEventBatchT)
```
atzoum, line 105: why are we accepting `any` in the method signature whereas we are very specific to what we expect?!

BonapartePC (approved):

atzoum (approved):

services/debugger/source/eventUploader.go
cisse21: `Uploader` module which does the final upload to config backend has access to `Transform` interface and the data type which uploader receives for source/destination/transformation live events are different. So interface has a generic  signature but the actual implementation checks for the exact data struct it needs

atzoum: This doesn't sound convincing, imo it is an indication of a weak design and poorly chosen exported API & responsibilities for these two components.

gateway/gateway.go
```diff
@@ -695,7 +695,7 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 		}
 		// Sending events to config backend
 		for _, eventBatch := range eventBatchesToRecord {
-			writeKey := gjson.Get(eventBatch, "writeKey").Str
+			writeKey := gjson.GetBytes(eventBatch, "writeKey").Str
```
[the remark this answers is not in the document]
cisse21: Good Idea... optimised it

services/debugger/source/eventUploader.go
cisse21: Anything which you have in mind for this?

atzoum:
I fear that the cache introduced by #1450 can be a greater memory hog than the uploader's queue.

It looks like we are caching the 3 latest events for each destination that doesn't have `eventDelivery` enabled and keep them  up to 30 days after the last event was encountered!
Namespaces with a large number of destinations, e.g. 1000+ will hold 3000+ events in memory for each uploader (source/destination/transformation)

PR #2804, opened 2022-12-13
feat(warehouse): implement ssh tunnelling

abhimanyubabbar:
# Description
In order for us to allow warehouse customers to connect to the destinations over `ssh`, we need to be able capture the information regarding the bastion host and key pairs required for the connection.

The changes of this PR can be categorised as follows:

1. When the backend config subscriber pulls a new config, the process to augment the tunnelling information is controlled behind the flag `enableTunnelling` controlled by env `ENABLE_TUNNELLING`. By default the value is `true` and can be flipped to `false`

2. When `enableTunnelling` is `true`, system goes over every destination and based on a boolean param set upstream, we start fetching the `keypair` from control plane required to open the ssh tunnel.

3. Added control plane client to fetch the destination ssh keys.

4. Added ability to capture `cp_internal_api_username` and `cp_internal_api_password` to allow for interacting with cp api using basic auth.

## Security


warehouse/warehouse.go
```diff
@@ -288,22 +293,54 @@ func (wh *HandleT) backendConfigSubscriber(ctx context.Context) {
...
+						}
+
+						if keypair == nil {
+							pkgLogger.Warnf("unable to locate keypair for destination: %s with valid tunnel info", destination.ID)
+						} else {
+							source.Destinations[idx].Config["privateKey"] = keypair.PrivateKey
+							source.Destinations[idx].Config["publicKey"] = keypair.PublicKey
+						}
+
+					}
```
lvrach, line 342: I would use a middleware approach for intercepting and injecting tunnel config.

Similar to how `tenantManager.WatchConfig` works. This will help with readability and testability of the newly added code, while keeping this section clean.

warehouse/warehouse.go
```diff
@@ -1230,6 +1263,12 @@ func (wh *HandleT) Setup(whType string) {
 	config.RegisterIntConfigVariable(1, &wh.maxConcurrentUploadJobs, false, 1, fmt.Sprintf(`Warehouse.%v.maxConcurrentUploadJobs`, whName))
 	config.RegisterBoolConfigVariable(false, &wh.allowMultipleSourcesForJobsPickup, false, fmt.Sprintf(`Warehouse.%v.allowMultipleSourcesForJobsPickup`, whName))
 
+	wh.internalAuth = cpclient.BasicAuth{
+		Username: config.GetString("CP_INTERNAL_API_USERNAME", ""),
+		Password: config.GetString("CP_INTERNAL_API_PASSWORD", ""),
+	}
+	wh.configBackendURL = configBackendURL
```
lvrach, line 1270:
```suggestion

       wh.CPInternalClient =  cpclient.NewInternalClientWithCache(configBackendURL, cpclient.BasicAuth{
		Username: config.GetString("CP_INTERNAL_API_USERNAME", ""),
		Password: config.GetString("CP_INTERNAL_API_PASSWORD", ""),
	})
```

It is preferable to DI the client rather than the config/credentials; it would be easier to mock the client in the feature.

However, it might be even better to inject the config middleware here instead depending on the client directly.

warehouse/warehouse.go
```diff
@@ -2090,6 +2130,11 @@ func Start(ctx context.Context, app app.App) error {
 		asyncWh = jobs.InitWarehouseJobsAPI(ctx, dbHandle, &notifier)
 		jobs.WithConfig(asyncWh, config.Default)
 
+		// Add code here to associate ssh keypair with destination data.
+		g.Go(misc.WithBugsnagForWarehouse(func() error {
+			return nil
+		}))
+
```
lvrach, line 2137:
```suggestion

```

Do we need this ?

warehouse/client/controlplane/client.go
```diff
@@ -0,0 +1,99 @@
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
+	if err != nil {
+		return nil, fmt.Errorf("creating new request with ctx: %w", err)
+	}
+
+	req.SetBasicAuth(api.auth.Username, api.auth.Password)
+	req.Header.Set("Content-Type", "application/json")
+
+	client := http.Client{}
```
lvrach, line 54: Avoid creating a new client in every request, cause it is going to establish a new tcp connection every time. 

Instead, reuse the client to enable connection pooling. We can simply do:

```go
type internalClient struct {
	baseURI string
	auth    BasicAuth
	client *http.Client
}
```

During the construction, you should also set the timeout.

```suggestion
```

warehouse/client/controlplane/client.go
```diff
@@ -0,0 +1,99 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("fetching response from http client: %w", err)
+	}
+
+	defer func() { httputil.CloseResponse(resp) }()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return nil, nil
```
lvrach, line 63: you can consider returning a custom error here

```go
const ErrKeyNotFound = errors.New("request key not found")
```

```suggestion
		return nil, ErrKeyNotFound
```

The consumer can test against this error

warehouse/client/controlplane/client.go
abhimanyubabbar: That's a good point about reusing the tcp connection. I'll make the change now.

warehouse/client/controlplane/client.go
abhimanyubabbar: Makes sense 👍🏼

warehouse/warehouse.go
abhimanyubabbar: No we don't, I'll remove this remanent of past.

sivashanmukh (approved):
LGTM

achettyiitr:
Let's revert the rudder-transformer changes.

warehouse/redshift/redshift.go
```diff
@@ -576,31 +577,51 @@ func (rs *HandleT) loadUserTables() (errorMap map[string]error) {
...
-	)
+	dsn := url.URL{
+		Scheme: "postgres",
+		User:   url.UserPassword(cred.Username, cred.Password),
+		Host:   fmt.Sprintf("%s:%s", cred.Host, cred.Port),
+		Path:   cred.DbName,
+	}
+
+	params := url.Values{}
+	params.Add("sslmode", "require")
```
achettyiitr, line 598: Any specific reason we introduced SSL mode for RS?

warehouse/tunnelling/model.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+type Type string
+type Config map[string]interface{}
+
+type TunnelInfo struct {
+	Config Config
+}
+
+func ReadSSHTunnelConfig(config map[string]interface{}) (conf *stunnel.Config, err error) {
```
achettyiitr, line 25:
```suggestion
func ReadSSHTunnelConfig(config Config) (conf *stunnel.Config, err error) {
```

warehouse/tunnelling/model.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return &stunnel.Config{
+		User:       *user,
+		Host:       *host,
+		PrivateKey: []byte(*privateKey),
+		Port:       portInt,
+	}, nil
+}
+
+func ReadString(key string, ip map[string]interface{}) (*string, error) {
```
achettyiitr, line 60:
```suggestion
func ReadString(key string, config Config) (*string, error) {
```

warehouse/tunnelling/model.go
```diff
@@ -0,0 +1,72 @@
...
+	Config Config
+}
+
+func ReadSSHTunnelConfig(config map[string]interface{}) (conf *stunnel.Config, err error) {
+
+	var (
+		user, host, port, privateKey *string
+	)
+
+	if user, err = ReadString("sshUser", config); err != nil {
```
achettyiitr, line 31: Can we use constants for sshUser, sshHost, sshPort, sshPrivateKey?

warehouse/tunnelling/model.go
```diff
@@ -0,0 +1,72 @@
...
+	"strconv"
+
+	stunnel "github.com/rudderlabs/sql-tunnels/driver/ssh"
+)
+
+var ErrMissingKey = errors.New("missing mandatory key")
+var ErrUnexpectedType = errors.New("unexpected type")
+
+const (
+	SSHForward Type = "ssh_forward"
```
achettyiitr, line 15: Is this for future reference since we are not using it now?

warehouse/clickhouse/clickhouse.go
```diff
@@ -200,6 +200,10 @@ func Init() {
 	pkgLogger = logger.NewLogger().Child("warehouse").Child("clickhouse")
 }
 
+func connectWithTunnel(cred CredentialsT, includeDBInConn bool) (*sql.DB, error) {
```
achettyiitr, line 203: Are we keeping this for future references?

warehouse/client/controlplane/client_test.go
```diff
@@ -0,0 +1,80 @@
...
+			expectedKeyPair: nil,
+		},
+	}
+
+	for _, tc := range testcases {
+
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			svc := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				_, _, ok := r.BasicAuth()
```
achettyiitr, line 59: WDYT?
Should we also check for auth?
```Go
username, password, ok := r.BasicAuth()
require.Equal(t, "username", username)
require.Equal(t, "password", password)
```

warehouse/clickhouse/clickhouse.go
abhimanyubabbar: I'll remove this. We will add it when we handle it.

warehouse/tunnelling/model.go
abhimanyubabbar: I'll remove this. Remanent of past

warehouse/tunnelling/model.go
abhimanyubabbar: Done.

warehouse/client/controlplane/client_test.go
```diff
@@ -0,0 +1,80 @@
...
+				PublicKey:  "public_key",
+				PrivateKey: "private_key",
+			},
+		},
+		{
+			name:            "fetch ssh keys returns no keys found",
+			responseBody:    ``,
+			responseCode:    http.StatusNotFound,
+			destinationID:   "id",
+			expectedError:   nil,
```
achettyiitr, line 41:
```suggestion
			expectedError:   fmt.Errorf("%w: key requested: id", cp.ErrKeyNotFound),
```

warehouse/api.go
```diff
@@ -122,6 +123,15 @@ func InitWarehouseAPI(dbHandle *sql.DB, log logger.Logger) error {
 	if region := config.GetString("region", ""); region != "" {
 		labels["region"] = region
 	}
+
+	client := cpclient.NewInternalClient(
```
achettyiitr, line 127:
```suggestion
	client := cpclient.NewInternalClientWithCache(
```
Any specific reason why we are not using cached client?

warehouse/tunnelling/model.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if host, err = ReadString(sshHost, config); err != nil {
+		return nil, err
+	}
+
+	if port, err = ReadString(sshPort, config); err != nil {
+		return nil, err
+	}
+
+	if privateKey, err = ReadString(sshPrivateKey, config); err != nil {
```
abhimanyubabbar, line 47: @lvrach This will result into err and return back before it could open tunnel.

warehouse/tunnelling/connect.go
```diff
@@ -0,0 +1,22 @@
+package tunnelling
```
lvrach, line 1: [nit-pick] I would merge this file with the `model.go`

lvrach (approved):

go.mod
```diff
@@ -220,3 +219,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+require (
+	github.com/golang-migrate/migrate/v4 v4.15.2
+	github.com/rudderlabs/sql-tunnels v0.0.0-20221213183444-465af21674b7
```
lvrach, line 225: We can use the latest version

achettyiitr (approved):

lvrach:
@abhimanyubabbar feel free to ignore the `security/snyk (rudderstack) — 1 test has failed` check.

The vulnerabilities are introduced by tested dependencies and they are not impacting us in any way.

PR #2805, opened 2022-12-13
chore: enable failed keys by default

chandumlg:
# Description

Enable the failed keys by default.

## Notion Ticket

https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=fa820a352a7542d39d9d0c90bf82bc04&pm=s

## Security


atzoum:
@koladilip rudder-sources team wants this functionality to be enabled by default, thus we are reversing the default value. Otherwise we would need to override the configuration in all k8s rudderstack objects.

cisse21 (approved):

PR #2799, opened 2022-12-12
fix: aws session creation is failing for s3 manager when roles are used

koladilip:
# Description

Unify AWS session config creation process so that it is consistent across all AWS destinations. Earlier S3 has slightly different process to create SessionConfing so when updated the session config creation in aws utils to support RoleBasedAuth flag, it got missed for S3 manager as it was using different implementation so this unification should prevent such errors in the future.

## Notion Ticket

[Warehouse AWS Role support] https://www.notion.so/rudderstacks/Warehouse-destination-config-for-Role-base-Authentication-control-plane-db9198dce2024723959c224e91a11ef4
## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #2807, opened 2022-12-13
fix: aws session creation is failing for s3 manager when roles are used

koladilip:
# Description

Redshift to use awsSession.Config.Credentials.Get() when roles are used to get temporary credentials instead of sts.GetSessionToken.

Unify AWS session config creation process so that it is consistent across all AWS destinations. Earlier S3 has slightly different process to create SessionConfing so when updated the session config creation in aws utils to support RoleBasedAuth flag, it got missed for S3 manager as it was using different implementation so this unification should prevent such errors in the future.

## Notion Ticket

[Warehouse AWS Role support] https://www.notion.so/rudderstacks/Warehouse-destination-config-for-Role-base-Authentication-control-plane-db9198dce2024723959c224e91a11ef4

## Security


koladilip:
This is already reviewed here:
https://github.com/rudderlabs/rudder-server/pull/2799
This is hotfix change merging back to master.

PR #2808, opened 2022-12-14
fix(destination): alert flag setting for router-aborted-count alert definition

sanpj2292:
# Description

The issue is that, when `transformerProxy` was enabled and `skipRtAbortAlertForDelivery` configuration was not present, the `alert` tag(or flag) was coming out to be `true`, which is not expected
We want the alert to only be triggered when the following happens during delivery phase for cloud-mode destinations
|proxy|skipRtAbortAlertForDelivery(in configuration)|alert(in router_aborted_events or router_response_counts stat)|
|--|--|--|
|0|0|1
|0|1|0
|1|0|0
|1|1|0

## Notion Ticket

https://www.notion.so/rudderstacks/Kapacitor-alerts-configuration-Prometheus-migration-a0f65d8ff61e440a9f69f9884752bf51

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -1019,7 +1019,6 @@ func (worker *workerT) updateReqMetrics(respStatusCode int, diagnosisStartTime *
 }
 
 func (worker *workerT) allowRouterAbortedAlert(errorAt string) bool {
-
 	switch errorAt {
 	case routerutils.ERROR_AT_CUST:
 		return true
@@ -1030,7 +1029,6 @@ func (worker *workerT) allowRouterAbortedAlert(errorAt string) bool {
 	default:
 		return true
 	}
-
 }
 
 func (worker *workerT) updateAbortedMetrics(destinationID, workspaceId, statusCode, errorAt string) {
--- a/router/router.go
+++ b/router/router.go
@@ -1019,12 +1019,18 @@ func (worker *workerT) updateReqMetrics(respStatusCode int, diagnosisStartTime *
 }
 
 func (worker *workerT) allowRouterAbortedAlert(errorAt string) bool {
-	// when error occur during transformation(rt or batch) or when skipRtAbortAlertForTransformation is false(default)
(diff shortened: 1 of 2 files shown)


router/router.go
```diff
@@ -1021,9 +1021,9 @@ func (worker *workerT) updateReqMetrics(respStatusCode int, diagnosisStartTime *
 func (worker *workerT) allowRouterAbortedAlert(errorAt string) bool {
 	// when error occur during transformation(rt or batch) or when skipRtAbortAlertForTransformation is false(default)
 	allowTf := errorAt == routerutils.ERROR_AT_TF && !worker.rt.skipRtAbortAlertForTransformation
-	// (when error occur during delivery and (when proxy is not enabled or when skipRtAbortAlertForDelivery is false(default)))
+	// (when error occur during delivery and (when proxy is not enabled and when skipRtAbortAlertForDelivery is false(default)))
 	// or when destination is managed custom destination manager
-	allowDel := (errorAt == routerutils.ERROR_AT_DEL && (!worker.rt.transformerProxy || !worker.rt.skipRtAbortAlertForDelivery)) || errorAt == routerutils.ERROR_AT_CUST
+	allowDel := (errorAt == routerutils.ERROR_AT_DEL && (!worker.rt.transformerProxy && !worker.rt.skipRtAbortAlertForDelivery)) || errorAt == routerutils.ERROR_AT_CUST
```
koladilip, line 1026: How about this?
```go
if errorAt == routerutils.ERROR_AT_CUST {
  return true
}
if worker.rt.skipRtAbortAlertForTransformation {
  return false
}
return errorAt == routerutils.ERROR_AT_TF ||
(errorAt == routerutils.ERROR_AT_DEL && !worker.rt.transformerProxy)
```

sanpj2292: Simplified it taking inspiration from your suggestion. Can you once take a look and let me know if this is more readable ?

koladilip (approved):

atzoum (approved):

PR #2811, opened 2022-12-14
chore(gateway): add webhook failure stats

koladilip:
# Description
This is for refactoring webhook failure stats as per comments in PR https://github.com/rudderlabs/rudder-server/pull/2809

## Notion Ticket

[Webhook failure stats](https://www.notion.so/rudderstacks/Add-alerts-on-source-transformation-10c57b78060c4eccb255321de9fc0f73)

## Security

### DIFF
--- a/gateway/webhook/webhookTransformer.go
+++ b/gateway/webhook/webhookTransformer.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/rudderlabs/rudder-server/gateway/response"
+	"github.com/rudderlabs/rudder-server/services/stats"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 )
 
@@ -47,6 +48,10 @@ func (bt *batchWebhookTransformerT) markResponseFail(sourceType, reason string) 
 func (bt *batchWebhookTransformerT) transform(events [][]byte, sourceType string) transformerBatchResponseT {
 	bt.stats.sentStat.Count(len(events))
 	bt.stats.transformTimerStat.Start()
+
+	stats.Default.NewTaggedStat("source_transform_received_jobs", stats.CountType, stats.Tags{
+		"sourceType": sourceType,
+	}).Count(len(events))
 
 	payload := misc.MakeJSONArray(events)
 	url := fmt.Sprintf(`%s/%s`, bt.sourceTransformerURL, strings.ToLower(sourceType))
--- a/gateway/webhook/webhookTransformer.go
+++ b/gateway/webhook/webhookTransformer.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 
 	"github.com/rudderlabs/rudder-server/gateway/response"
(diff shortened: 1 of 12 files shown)


gateway/webhook/webhookTransformer.go
```diff
@@ -36,15 +35,13 @@ type transformerBatchResponseT struct {
...
 		Err:        response.GetStatus(reason),
-		StatusCode: response.GetErrorStatusCode(reason),
+		StatusCode: statusCode,
 	}
 	bt.stats.failedStat.Count(1)
-	stats.Default.NewTaggedStat("source_transform_failed_jobs", stats.CountType, stats.Tags{
-		"sourceType": sourceType,
-		"statusCode": strconv.Itoa(response.GetErrorStatusCode(reason)),
-	}).Count(1)
+	countWebhookErrors(sourceType, statusCode, 1)
```
atzoum, line 44: do we really want to include this here? won't this lead to duplicates?

gateway/webhook/webhookTransformer.go
```diff
@@ -63,6 +60,8 @@ func (bt *batchWebhookTransformerT) transform(events [][]byte, sourceType string
 	bt.stats.transformTimerStat.End()
 	if err != nil {
 		err := fmt.Errorf("JS HTTP connection error to source transformer: URL: %v Error: %+v", url, err)
+		statusCode := http.StatusServiceUnavailable
+		countWebhookErrors(sourceType, statusCode, len(events))
```
atzoum, line 64: I believe it would be better for this counter to be incremented by only one method, during `batchWebhookTransformerT#batchTransformLoop`.

It is hard to reason now whether the current implementation is correct and doesn't miss or count more stats than it should.

koladilip: Yes I missed this, moved metrics counting to batchWebhookTransformerT#batchTransformLoop

gateway/webhook/webhookTransformer.go
koladilip: Yes removed from this method

atzoum (approved):
Thank you @koladilip.

An observation, is that we appear to be missing proper test coverage for the `webhook` package. E.g. most of the areas that this pull request touches are not covered with tests. Code not covered by tests is risky to maintain & evolve & more usually than not it can hide bugs and logical mistakes.

Thus, whenever we need to make changes to a piece of code that lacks proper test coverage, we also try to make sure to add missing test scenarios, at least covering the changes that we are introducing. Occasionally, this requires some refactoring to happen in parallel so that code becomes testable.

With this strategy we can at least be confident that we are not introducing more technical debt than it already exists. Additionally, we are making it easier for the next set of test scenarios to be written, i.e. dealing with debt incrementally, instead of letting it pile up to an insurmountable mountain.

koladilip:
@atzoum writing test cases for stats is bit challenging so I would need your help on this since this needs to be released as hot fix, I will merge this and start work on tests as separate task.

PR #2812, opened 2022-12-14
fix(warehouse): include exclusion window in status tracker

achettyiitr:
# Description

## Notion Ticket

https://www.notion.so/rudderstacks/Factor-in-exclusion-window-for-the-triggering-of-the-alert-for-stuck-wh-pipelines-warehouse-success-ea319bf05d54416a9dc0e593b45e9291

## Security


warehouse/tracker.go
```diff
@@ -0,0 +1,183 @@
...
+		"workspaceId": warehouse.WorkspaceID,
+		"module":      moduleName,
+		"destType":    wh.destType,
+		"warehouseID": misc.GetTagName(
+			destination.ID,
+			source.Name,
+			destination.Name,
+			misc.TailTruncateStr(source.ID, 6)),
+	}
+	wh.stats.NewTaggedStat("warehouse_successful_upload_exists", stats.CountType, tags).Count(uploaded)
```
lvrach, line 181: I would suggest that instead of reporting 0 or 1, report:

max( `now() - max(created_at, excludeWindowEndTime)` of each upload)

We can later set up the rule based on sync time

warehouse/testdata/sql/7.sql
```diff
@@ -0,0 +1,23 @@
+BEGIN;
```
lvrach, line 1: `7.sql`, this naming might get unmanageable in the future. It might be better to give the name of the test file we are using it for:
```
seed_tracker_test.sql
```

warehouse/tracker.go
```diff
@@ -0,0 +1,183 @@
...
+
+		for _, warehouse := range wh.warehouses {
+			warehouse := warehouse
+
+			rruntime.GoForWarehouse(func() {
+				defer wg.Done()
+				ch <- warehouse
+			})
+		}
+		wh.configSubscriberLock.RUnlock()
```
lvrach, line 37: I try to get my head around this, why do we spawn up go routines, and we need a channel here?

Couldn't we just copy the warehouses in a new slice and pass them to `wh.Track`?

warehouse/tracker.go
```diff
@@ -0,0 +1,183 @@
...
+	var (
+		query         string
+		queryArgs     []interface{}
+		createdAt     sql.NullTime
+		exists        bool
+		uploaded      int
+		syncFrequency = "1440"
+		Now           = timeutil.Now
+		NowSQL        = "NOW()"
+		timeWindow    = config.GetInt("Warehouse.uploadBufferTimeInMin", 180)
```
lvrach, line 69: Ideally, we should use time.Duration for `time.Window`, but this change might exceed the scope of this PR.

warehouse/tracker.go
```diff
@@ -0,0 +1,183 @@
...
+					  destination_id = $2 AND
+					  (
+						status = $3
+						OR status = $4
+						OR status LIKE $5
+					  ) AND
+					  updated_at > $6
+				  );
+	`,
+		warehouseutils.WarehouseUploadsTable,
```
lvrach, line 151: Personal preference, assuming `warehouseutils.WarehouseUploadsTable` is a constant
```suggestion
	query = `
				SELECT
				  EXISTS (
					SELECT
					  1
					FROM
					  ` + warehouseutils.WarehouseUploadsTable + `
					WHERE
					  source_id = $1 AND
					  destination_id = $2 AND
					  (
						status = $3
						OR status = $4
						OR status LIKE $5
					  ) AND
					  updated_at > $6
				  );
	`
```

warehouse/tracker.go
lvrach: @achettyiitr, what do you think of this? Can we add a metric to capture the max, avg, and sum waiting time?

warehouse/tracker.go
achettyiitr: Yeah. Initially, I was also thinking along similar lines. But seeing the implementation for the mainLoop I got biased. Making the changes.

warehouse/tracker.go
achettyiitr: Added this since we already have a fallback for when time unit is not present.

warehouse/tracker.go
achettyiitr: We will be tracking this as a separate task?

lvrach (approved):

abhimanyubabbar (approved):
The changes look good to me. 1 minor comment on the exclude window being a struct.

PR #2816, opened 2022-12-16
fix: get instance id using regex

BonapartePC:
# Description

Getting instance id using regex. This will help us fetch instance id properly for gw ha deployments as well.
Ex: for rudderstack-gw-ha-0-89gsas we now get `0-89gsas` instead of `89gsas` like before.

## Notion Ticket

https://www.notion.so/rudderstacks/Gateway-HA-support-for-multitenant-37b2158507a24ded81b5744269bc2038

## Security

### DIFF
--- a/config/misc.go
+++ b/config/misc.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"os"
+	"regexp"
 	"strings"
 )
 
@@ -35,8 +36,8 @@ func GetInstanceID() string {
 	if instance == "" {
 		return ""
 	}
-	instanceArr := strings.Split(instance, "-")
-	return instanceArr[len(instanceArr)-1]
+	re := regexp.MustCompile("-[0-9]+([A-Za-z0-9-]*)$")
+	return strings.TrimPrefix(string(re.Find([]byte(instance))), "-")
 }
 
 func GetReleaseName() string {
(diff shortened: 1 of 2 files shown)


config/misc.go
```diff
@@ -35,8 +36,8 @@ func GetInstanceID() string {
 	if instance == "" {
 		return ""
 	}
-	instanceArr := strings.Split(instance, "-")
-	return instanceArr[len(instanceArr)-1]
+	re := regexp.MustCompile("-[0-9]+([A-Za-z0-9-]*)$")
+	return strings.TrimPrefix(string(re.Find([]byte(instance))), "-")
```
atzoum, line 40: Don't we just need to find the first segment that represents a number?
```suggestion
	// split instance id by "-" and return the first number
	if instanceID, ok := lo.Find(
		strings.Split(instance, "-"),
		func(s string) bool {
			_, err := strconv.Atoi(s)
			return err == nil
		}); ok {
		return instanceID
	}
	return ""
```

config/misc.go
```diff
@@ -35,8 +36,8 @@ func GetInstanceID() string {
 	if instance == "" {
 		return ""
 	}
-	instanceArr := strings.Split(instance, "-")
-	return instanceArr[len(instanceArr)-1]
+	re := regexp.MustCompile("-[0-9]+([A-Za-z0-9-]*)$")
```
atzoum, line 39: Please add a test case for this function covering all scenarios & using some actual instance_id values:

1. Scenarios that successfully parse `INSTANCE_ID`
2.Scenarios that fail to parse `INSTANCE_ID` and return an empty string

Specifically for the case of `ETCDManager`, I understand that it is important that instance id is a non-empty string, i.e. a number, thus the least we should do is log a warning when this is empty, or even fail to start.

config/misc.go
cisse21: We are also making the scheduler changes accordingly to wait for both gateway pods to acknowledge

cisse21:
Closing this PR as @raman2805 made the changes on another PR

PR #2820, opened 2022-12-19
chore: drop support for v1 pending events

atzoum:
# Description

/v1/pending-events has now been superseded by /v1/job-status

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=fb48aa6a733b4bb7922a9a9e83fd09f1&pm=s)

## Security


app/apphandlers/setup.go
```diff
@@ -53,7 +53,7 @@ func rudderCoreWorkSpaceTableSetup() error {
 	return validators.CheckAndValidateWorkspaceToken()
 }
 
-func setupReadonlyDBs() (gw, rt, batchrt *jobsdb.ReadonlyHandleT, err error) {
+func setupReadonlyDBs() (gw *jobsdb.ReadonlyHandleT, err error) {
```
cisse21, line 56: We should probably change the function name as it only set's up GW Db

cisse21 (approved):

app/apphandlers/setup.go
atzoum: In reality it is not that simple. It sets all readonly dbs but only returns gw db... bad design

cisse21 (approved):

app/apphandlers/setup.go
```diff
@@ -62,25 +62,22 @@ func setupReadonlyDBs() (gw, rt, batchrt *jobsdb.ReadonlyHandleT, err error) {
 	var gwDB, rtDB, batchrtDB, procerrDB jobsdb.ReadonlyHandleT
 
 	if err := gwDB.Setup("gw"); err != nil {
-		return nil, nil, nil, fmt.Errorf("setting up gw readonly db: %w", err)
+		return nil, fmt.Errorf("setting up gw readonly db: %w", err)
 	}
 	gw = &gwDB
 
 	if err := rtDB.Setup("rt"); err != nil {
```
Sidddddarth, line 69: we don't need these anymore, do we?

Sidddddarth (approved):

PR #2822, opened 2022-12-19
fix(warehouse): processing available workers to be gauge stats

achettyiitr:
# Description

The available number of workers to be Guage stats.

## Notion Ticket

https://www.notion.so/rudderstacks/fix-warehouse-processing-stats-to-be-guage-for-workers-3f9ad02af83541459ae1a3434d65561a

## Security

### DIFF
--- a/services/stats/memstats/stats.go
+++ b/services/stats/memstats/stats.go
@@ -5,6 +5,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/spf13/cast"
+
 	"github.com/rudderlabs/rudder-server/services/stats"
 )
 
@@ -105,7 +107,7 @@ func (m *Measurement) Gauge(value interface{}) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
-	m.values = append(m.values, value.(float64))
+	m.values = append(m.values, cast.ToFloat64(value))
 }
 
 // Observe implements stats.Measurement
--- a/go.mod
+++ b/go.mod
@@ -188,7 +188,7 @@ require (
 	github.com/segmentio/backo-go v0.0.0-20160424052352-204274ad699c // indirect
 	github.com/sirupsen/logrus v1.9.0 // indirect
 	github.com/spf13/afero v1.9.2 // indirect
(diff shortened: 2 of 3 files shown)


services/stats/memstats/stats.go
```diff
@@ -105,7 +107,7 @@ func (m *Measurement) Gauge(value interface{}) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
-	m.values = append(m.values, value.(float64))
+	m.values = append(m.values, cast.ToFloat64(value))
```
lvrach, line 110: Just curious, why did you use the library for casting ?

lvrach (approved):

services/stats/memstats/stats.go
achettyiitr: We are sending `int` till now in Guage stats.

abhimanyubabbar: I think leo was asking, why add one more dependency. Why not in this case write the code yourself to convert to `float64` ?

abhimanyubabbar (approved):
Looks good. Minor comment onto the addition of a new library for now.

abhimanyubabbar:
I think the library was already imported indirectly. So ignore my comment.

services/stats/memstats/stats.go
achettyiitr: I think I misunderstood the comment. Since the value type is interface{} and it can be anything. So I thought of un-necessary handling the type casting by ourselves whenever the new type is being added we can use cast library.

achettyiitr:
```Go
	var v float64

	switch value.(type) {
	case int:
		v = float64(value.(int))
	case float64:
		v = value.(float64)
	default:
		panic("unsupported type for Gauge")
	}
	m.values = append(m.values, v)
```
Should I go ahead with this one?
@lvrach @abhimanyubabbar

lvrach: > Should I go ahead with this one?
For me, the current approach is ok. I just wanted to understand the problem.

PR #2823, opened 2022-12-20
feat: changes to make rudder-scheduler work with HA gateway

raman2805:
# Description

Changes to make rudder-scheduler correctly communicate with all replicas of the gateway server when running in HA 
mode. 

Currently, to reload the workspaces for a server, the scheduler writes the following key `<releaseName>/SERVER/<serverIndex>/<appType>/WORKSPACES ` to etcd with the json contents  containing another key which the server app should ack at once it finishes reloading. This key is named _ack_key_ and looks like `<some_constant_prefix>/<serverIdx>/<appType>`

This works fine for processors and gateways (without HA). However, with HA gateways, we want to wait for all the replicas of the gateway receive the request and ack it before we proceed. We are making the following change to scheduler and server to achieve this

1. The scheduler asks server to reload the workspaces by putting in the key as mentioned above
2. The GATEWAY/PROCESSOR app watches for that etcd key for changes in workspaces. Once it receives a change event and has finished processing, it writes the following 2 keys (instead of 1) to etcd 
    - {ack_key}  [This is done to handle the status quo so that the change in server is compatible with old instances of the scheduler] 
    - {ack_key}/{serverPodName}
3. The scheduler listens to the prefix {ack_key} (instead of the exact match). It continuously tries to compare all the running server (gateway) pods and compares it from the set of pods obtained from ack'ed keys (by plucking out the suffixed server name). Once both the lists match, it moves to the next step. 

## Notion Ticket
https://www.notion.so/rudderstacks/Gateway-HA-support-for-multitenant-37b2158507a24ded81b5744269bc2038

## Security


app/cluster/state/etcd.go
```diff
@@ -267,14 +267,27 @@ func (manager *ETCDManager) unmarshalWorkspace(raw []byte) workspace.ChangeEvent
...
-			_, err = manager.Client.Put(ctx, req.AckKey, ackValue)
-			if err != nil {
-				manager.logger.Errorf(
-					"Failed to acknowledge workspace ID change (error: %v) for key: %s", ackErr != nil, req.AckKey,
-				)
+			// for backward compatibility, for GATEWAY we acknowledge the request at both
+			// the incoming value contained in ack_key and the new value obtained by appending
+			// the instance ID to the ack_key. This enables the scheduler to ensure all the
+			// replicas of the gateway have acknowledged the request, when running in HA mode
+			ackKeys := []string{req.AckKey}
```
BonapartePC, line 274: I don't think we need backward compatibility for this. We have to update the scheduler before we update server so we should be good I guess. Can we get rid of this?

app/cluster/state/etcd.go
```diff
@@ -267,14 +267,27 @@ func (manager *ETCDManager) unmarshalWorkspace(raw []byte) workspace.ChangeEvent
...
+			appTypeStr := strings.ToUpper(config.GetString("APP_TYPE", app.PROCESSOR))
+			if appTypeStr == app.GATEWAY {
```
BonapartePC, line 276: key will be `<ack_key>/<instance_id>` should happen for both gateway and processor

raman2805: removing the if

cisse21 (approved):

BonapartePC (approved):

PR #2830, opened 2022-12-22
refactor: use destination config from config-be than from message

koladilip:
# Description
Earlier transformer was sending destConfig in message and it contains secrets so it is appearing in live events so refactored code to use config from CP.

## Notion Ticket

[Lambda Keys exposed in Live events](https://www.notion.so/rudderstacks/CFD-Destination-config-leaking-to-Live-Events-for-AWS-Lambda-0efdfbd385384462b0f194dca2c66bfd)

## Security

### DIFF
--- a/services/streammanager/lambda/lambdamanager.go
+++ b/services/streammanager/lambda/lambdamanager.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/service/lambda"
 	jsoniter "github.com/json-iterator/go"
+	"github.com/mitchellh/mapstructure"
 	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
 	"github.com/rudderlabs/rudder-server/services/streammanager/common"
 	"github.com/rudderlabs/rudder-server/utils/awsutils"
@@ -15,14 +16,13 @@ import (
 
 // Config is the config that is required to send data to Lambda
 type destinationConfig struct {
-	InvocationType string
-	ClientContext  string
-	Lambda         string
+	InvocationType string `json:"invocationType"`
+	ClientContext  string `json:"clientContext"`
+	Lambda         string `json:"lambda"`
 }
 
-type inputConfig struct {
-	Payload           string             `json:"payload"`
-	DestinationConfig *destinationConfig `json:"destConfig"`
+type inputData struct {
+	Payload string `json:"payload"`
 }
 
 type LambdaProducer struct {
@@ -56,13 +56,13 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (*Lambd
 }
 
 // Produce creates a producer and send data to Lambda.
-func (producer *LambdaProducer) Produce(jsonData json.RawMessage, _ interface{}) (int, string, string) {
+func (producer *LambdaProducer) Produce(jsonData json.RawMessage, destConfig interface{}) (int, string, string) {
 	client := producer.client
 	if client == nil {
 		return 400, "Failure", "[Lambda] error :: Could not create client"
 	}
 
-	var input inputConfig
+	var input inputData
 	err := jsonfast.Unmarshal(jsonData, &input)
 	if err != nil {
 		returnMessage := "[Lambda] error while unmarshalling jsonData :: " + err.Error()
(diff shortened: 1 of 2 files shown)


services/streammanager/lambda/lambdamanager.go
```diff
@@ -15,14 +16,13 @@ import (
...
 type destinationConfig struct {
-	InvocationType string
-	ClientContext  string
-	Lambda         string
+	InvocationType string `json:"invocationType"`
+	ClientContext  string `json:"clientContext"`
+	Lambda         string `json:"lambda"`
 }
 
 type inputConfig struct {
```
saikumarrs, line 24: The name `inputConfig`, is it still relevant?

services/streammanager/lambda/lambdamanager.go
```diff
@@ -71,9 +71,14 @@ func (producer *LambdaProducer) Produce(jsonData json.RawMessage, _ interface{})
...
-	if config == nil {
-		return 400, "Failure", "[Lambda] error :: Invalid destination config"
+	var config destinationConfig
+	err = mapstructure.Decode(destConfig, &config)
+	if err != nil {
+		returnMessage := "[Lambda] error while unmarshalling destConfig :: " + err.Error()
+		return 400, "Failure", returnMessage
+	}
+	if config.InvocationType == "" {
+		config.InvocationType = "Event"
```
saikumarrs, line 81: Is this change relevant? Also, is there a test case for this?

koladilip: Yes this required as this not part of the destConfig, we are hard coding it.

utsabc: We are always sending this from transformer there won't be any scenario of this being empty string

services/streammanager/lambda/lambdamanager_test.go
```diff
@@ -81,26 +82,6 @@ func TestProduceWithInvalidData(t *testing.T) {
...
-	// Invalid Destination Config
-	sampleDestConfig := map[string]interface{}{}
-	sampleEventJson, _ = json.Marshal(map[string]interface{}{
-		"payload":    sampleMessage,
-		"destConfig": "invalid dest config",
-	})
-	statusCode, statusMsg, respMsg = producer.Produce(sampleEventJson, sampleDestConfig)
-	assert.Equal(t, 400, statusCode)
-	assert.Equal(t, "Failure", statusMsg)
-	assert.Contains(t, respMsg, "[Lambda] error while unmarshalling jsonData")
```
utsabc, line 103: Why are we removing these tests?

services/streammanager/lambda/lambdamanager.go
koladilip: yes, it is to hold what transformer is sending right?

services/streammanager/lambda/lambdamanager.go
saikumarrs: @utsabc the plan is to not use the config sent from the transformer.

koladilip: We won't be sending this in the future from transformer so this required here.

utsabc: Got it

services/streammanager/lambda/lambdamanager_test.go
koladilip: Destination config is not part the input message any more so this is not required anymore

services/streammanager/lambda/lambdamanager.go
saikumarrs: It's the input payload, right? Now, we are not even sending the config.

utsabc (approved):

services/streammanager/lambda/lambdamanager.go
koladilip: https://github.com/rudderlabs/rudder-server/blob/master/services/streammanager/common/common.go#L16
we are expecting json as input in the method.
https://github.com/rudderlabs/rudder-transformer/blob/master/src/v0/destinations/lambda/transform.js#L66
We are sending this structure from transformer.

saikumarrs: I'm just talking about the name `inputConfig`. Should it be something like `inputData` at least?

koladilip: Sorry I missed reading the name part. inputData is fine

koladilip: Updated

sandeepdsvs (approved):

saikumarrs (approved):

PR #2832, opened 2022-12-26
chore: introduce region in reporting url

cisse21:
# Description
As part of data residency adaptations ... introducing the changes for encoding region into query params for reporting and live events url

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=7f3a90f7bb7a4c9aab1a66c84aa10e2e&pm=s)

## Security


enterprise/reporting/reporting.go
```diff
@@ -89,58 +90,59 @@ func NewFromEnvConfig(log logger.Logger) *HandleT {
 		whActionsOnly:             whActionsOnly,
 		sleepInterval:             sleepInterval,
 		mainLoopSleepInterval:     mainLoopSleepInterval,
+		region:                    config.GetString("region", ""),
 	}
 }
 
-func (handle *HandleT) setup(beConfigHandle backendconfig.BackendConfig) {
```
saurav-malani, line 95: if you are renaming `r`, won't it be better to rename it to `h` instead of `handle`? Since, such a large receiver name is avoided in golang.

saurav-malani (approved):

enterprise/reporting/reporting.go
cisse21: Used `r` to just represent reporting handle... can change it to `rh` though

services/debugger/uploader.go
```diff
@@ -118,6 +120,11 @@ func (uploader *uploaderImpl[E]) uploadEvents(eventBuffer []E) {
 			pkgLogger.Errorf("[Uploader] Failed to create new http request. Err: %v", err)
 			return
 		}
+		if uploader.region != "" {
+			q := req.URL.Query()
+			q.Add("region", uploader.region)
+			req.URL.RawQuery = q.Encode()
+		}
```
BonapartePC, line 127: Can we write test for this?

BonapartePC (approved):

Sidddddarth (approved):

PR #2833, opened 2022-12-26
chore: use badgerdb for live events cache

cisse21:
# Description

Currently, the live events cache uses an in-memory list to store the last 3 batches for sources/destinations/transformations. This can cause a bloat up of memory for shared multi-tenant deployments which would not be cleaned up for a long time. The storage is now moved to an embedded badgerDB which persists these events to disk with a TTL

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=f052de999e5b489caf40248107dd3f92&pm=s)

## Security


services/debugger/cache/cache.go
```diff
@@ -0,0 +1,46 @@
...
+
+//go:generate mockgen -destination=../../../mocks/services/debugger/cache/mock_cache.go -package=mocks_cache github.com/rudderlabs/rudder-server/services/debugger/cache CacheAny
+
+import (
+	"github.com/rudderlabs/rudder-server/services/debugger/cache/embeddedcache"
+	"github.com/rudderlabs/rudder-server/services/debugger/cache/listcache"
+	"github.com/rudderlabs/rudder-server/utils/logger"
+)
+
+var pkgLogger logger.Logger
```
fracasula, line 11: Can we avoid global variables please? The logger can simply be passed as a dependency ([example](https://github.com/rudderlabs/rudder-server/blob/v1.4.5/enterprise/suppress-user/repository.go#L34)).

services/debugger/cache/cache.go
```diff
@@ -0,0 +1,46 @@
...
+
+type CacheAny interface {
+	Cache[any]
+}
```
fracasula, line 15: What is the purpose of this interface? It doesn't seem to be used.

services/debugger/cache/cache.go
```diff
@@ -0,0 +1,46 @@
...
+func init() {
+	pkgLogger = logger.NewLogger().Child("cache")
+}
+
+func (FactoryImpl[E]) New(cacheType string) Cache[E] {
+	switch cacheType {
+	case "memory":
+		pkgLogger.Info("Using in-memory cache")
+		return &listcache.ListCache[E]{}
+	case "goqpue":
```
fracasula, line 39: Typo?

services/debugger/cache/cache.go
```diff
@@ -0,0 +1,46 @@
...
+		pkgLogger.Info("Using goque cache")
```
fracasula, line 40: Isn't this using badger under the hood? What is this `goque`?

services/debugger/cache/cache.go
```diff
@@ -0,0 +1,46 @@
...
+type CacheAny interface {
+	Cache[any]
+}
+
+type Cache[E any] interface {
+	Update(key string, value E)
+	ReadAndPopData(key string) []E
+}
+
+type Impl[E any] struct{}
```
fracasula, line 22: I think we can remove this `Impl` type and unexport all other types (i.e. `EmbeddedCache`, `ListCache`...) beside `Cache[E]`, right?

services/debugger/cache/cache.go
```diff
@@ -0,0 +1,46 @@
...
+
+type Factory[E any] interface {
+	New(cacheType string) Cache[E]
+}
+
+func init() {
+	pkgLogger = logger.NewLogger().Child("cache")
+}
+
+func (FactoryImpl[E]) New(cacheType string) Cache[E] {
```
fracasula, line 34: Do we even need `FactoryImpl`? Can't we just have a `New` function to return `Cache`?

Possibly I think this whole file could be reduced to this:

```go
type cacheType int8

const (
	MemoryCacheType cacheType = iota
	BadgerCacheType
)

type Cache[T any] interface {
	Update(key string, value T)
	ReadAndPopData(key string) []T
}

func New[T any](ct cacheType, l logger.Logger) (Cache[T], error) {
	switch ct {
	case MemoryCacheType:
		l.Info("Using in-memory cache")
		return &listcache.ListCache[T]{}, nil
	case BadgerCacheType:
		l.Info("Using badger cache")
		return &embeddedcache.EmbeddedCache[T]{}, nil
	default:
		return nil, fmt.Errorf("cache type %v out of the known domain", ct)
	}
}
```

fracasula (changes requested):
First round of review. I still have to look into `embeddedcache` and `listcache`.

services/debugger/cache/cache.go
cisse21: Yeah this makes much more sense...

services/debugger/cache/cache.go
cisse21: Initially I thought of implementing it using embedded queue but it felt like a pain to maintain it so ended up using badger

services/debugger/cache/cache.go
cisse21: Yeah will refactor accordingly by trying not using global variables.

services/debugger/cache/cache.go
cisse21: This is unused but the gomock we use does not seem to be having the capability to generate mocks for generic interfaces and hence this hack

fracasula: Then I'd say don't use gomock :slightly_smiling_face:
It seems very simple to mock without assistance from a library :man_shrugging:

services/debugger/cache/listcache/listCache.go
```diff
@@ -0,0 +1,107 @@
...
+		config.RegisterDurationConfigVariable(15, &c.cleanupFreq, true, time.Second, "LiveEvent.cache.clearFreq") // default clearFreq is 15 seconds
+	}
+}
+
+/*
+	Cache is an in-memory cache. Each key-value pair stored in this cache have a TTL and one goroutine removes the
+
+key-value pair form the cache which is older than TTL time.
+*/
+type ListCache[E any] struct {
```
fracasula, line 36: What do you think of using [this one](https://github.com/rudderlabs/rudder-server/blob/87a967e1a734789402db70d01b9e626699d40917/internal/throttling/cachettl/cachettl.go) instead? I designed it purposedly to avoid go routines and to increase performance. It's a double linked list with a hashmap to guarantee O(1) when getting values. The values are sorted in the double linked list by TTL to speed up the expiration process so that you don't have to iterate over all values each time (remember that we acquired a lock in order to clean up and that we don't know how many values we might have in the map. if they are sorted by ttl you can stop the moment you come across a value that has not expired). The double linked list is used also to speed up deletions and insertions.

```go
type ListCache[E any] struct {
	cache  cachettl.Cache[string, E]
	keyTTL time.Duration // Time after which the data will be expired and removed from the cache
}

/*
Update Inserts the data in the cache, This method expects a string as a key and []byte as the data
*/
func (c *ListCache[E]) Update(key string, value E) {
	c.cache.Put(key, value, c.keyTTL)
}
```

You might need to move it out of `internal` and add a `Pop` method. Let me know if it doesn't make sense and if you need assistance.

cisse21: Yeah I remember you using this in the throttling code but this has been present as legacy code for almost a year so I kept it intact but makes sense to re-use the one you implemented for redis

services/debugger/source/eventUploader.go
```diff
@@ -42,12 +43,13 @@ var (
...
 	disableEventUploads bool
 	pkgLogger           logger.Logger
-	eventsCacheMap      debugger.Cache[[]byte]
+	eventsCache         cache.Cache[[]byte]
 )
 
 func Init() {
 	loadConfig()
 	pkgLogger = logger.NewLogger().Child("debugger").Child("source")
+	eventsCache, _ = cache.New[[]byte](cache.BadgerCacheType, pkgLogger)
```
cisse21, line 52: @fracasula Is there any nicer way where I can toggle the type instead of hardcoding this? All I can think of is to have a separate function in `config` which returns the CacheType output

fracasula: Not sure I follow. Do you need this to switch from `[]byte` to something else in `Init()`?

cisse21: I want to be able to change `cache.BadgerCacheType` to something else via config

services/debugger/cache/listcache/listCache.go
cisse21: I was looking into incorporating this but it seems like we don't have support to hold only a certain number of elements in memory. Let me know if I missed something

fracasula: We don't have it although it could be added along with a few tests. What would be the policy if the limit is reached? Refuse adding or eviction of older elements?

gateway/gateway_test.go
```diff
@@ -114,12 +114,8 @@ func (c *testContext) Setup() {
 	c.mockApp = mocksApp.NewMockApp(c.mockCtrl)
 	c.mockRateLimiter = mocksRateLimiter.NewMockRateLimiter(c.mockCtrl)
 
-	tFunc := c.asyncHelper.ExpectAndNotifyCallbackWithName("process_config")
```
atzoum, line 117: might need to reinstate this functionality

router/eventorder_test.go
```diff
@@ -144,8 +144,10 @@ func TestEventOrderGuarantee(t *testing.T) {
 			// statusInsertLoop
 			config.Set("Router.updateStatusBatchSize", 100)
 			config.Set("Router.maxStatusUpdateWait", "10ms")
-
-			defer config.Reset()
+			config.Set("SourceDebugger.origin", "event_order_source_"+trand.String(100))
+			config.Set("DestinationDebugger.origin", "event_order_destination_"+trand.String(100))
+			config.Set("TransformerDebugger.origin", "event_order_transformer_"+trand.String(100))
+			config.Set("LiveEvent.cache.GCTime", "1s")
```
atzoum, line 150: are these really needed?
```suggestion
			defer config.Reset()
```

router/router_test.go
```diff
@@ -153,6 +154,7 @@ func initRouter() {
 	config.Reset()
 	admin.Init()
 	logger.Reset()
+	misc.Init()
```
atzoum, line 157: is this really needed?

services/debugger/cache/cache.go
```diff
@@ -0,0 +1,33 @@
...
+	switch ct {
+	case MemoryCacheType:
+		l.Info("Using in-memory cache")
+		return &listcache.ListCache[T]{Origin: origin}
+	case BadgerCacheType:
+		l.Info("Using badger cache")
+		return &embeddedcache.EmbeddedCache[T]{Origin: origin, Logger: l}
+	default:
+		return &listcache.ListCache[T]{Origin: origin}
+	}
```
atzoum, line 32: `case MemoryCacheType` is not needed since it is the default cache, right?
```suggestion
	switch ct {
	case BadgerCacheType:
		l.Info("Using badger cache")
		return &embeddedcache.EmbeddedCache[T]{Origin: origin, Logger: l}
	default:
		l.Info("Using in-memory cache")
		return &listcache.ListCache[T]{Origin: origin}
	}
```

router/router_test.go
cisse21: There is a global variable in misc package which get's triggered to nil pointer exception whenever a tmpdir is created is what I saw in few tests

router/eventorder_test.go
cisse21: Why do need two `defer config.Reset()`? We already have one at the start of the function right

gateway/gateway_test.go
fracasula: What was this used for?

router/eventorder_test.go
fracasula: The tests are passing for me even without these four `config.Set` lines. The `config.Reset()` and `defer config.Reset()` on top seem to work.

services/debugger/destination/eventDeliveryStatusUploader.go
```diff
@@ -27,73 +28,129 @@ type DeliveryStatusT struct {
...
+
+var WithDisableEventUploads = func(disableEventDeliveryStatusUploads bool) func(h *Handle) {
+	return func(h *Handle) {
+		h.disableEventDeliveryStatusUploads = disableEventDeliveryStatusUploads
+	}
+}
+
+func NewHandle(opts ...Opt) *Handle {
+	h := &Handle{
+		configBackendURL: config.GetString("CONFIG_BACKEND_URL", "https://api.rudderstack.com"),
```
fracasula, line 56: Can't we pass the backend url as a parameter instead of relying on `config`?

fracasula: Same for the logger.

services/debugger/cache/listcache/listCache.go
fracasula: @cisse21 if I understand correctly the plan is to remove `ListCache` once we're confident that badger works. please make sure to add a ticket for that.

services/debugger/cache/embeddedcache/embeddedCache.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+		return values, nil
+	}
+
+	res, _ := misc.QueryWithRetries(context.TODO(), e.queryTimeout, e.retries, read)
+
+	return res
+}
+
+func (e *EmbeddedCache[E]) Init() {
```
fracasula, line 130: We can unexport this, right?

services/debugger/cache/embeddedcache/embeddedCache.go
```diff
@@ -0,0 +1,192 @@
...
+	e.once.Do(func() {
+		e.loadCacheConfig()
+		badgerPathName := e.Origin + "/cache/badgerdbv3"
+		tmpDirPath, err := misc.CreateTMPDIR()
+		if err != nil {
+			e.Logger.Errorf("Unable to create tmp directory: %v", err)
+			return
```
fracasula, line 137: If we use a proper constructor we can fail there instead of logging and returning here. It'd be a more robust way to handle these cases in my opinion.

services/debugger/cache/embeddedcache/embeddedCache.go
```diff
@@ -0,0 +1,192 @@
...
+		})
+	})
+}
+
+func (e *EmbeddedCache[E]) gcBadgerDB() {
+	ticker := time.NewTicker(e.Ticker)
+	defer ticker.Stop()
+	// One call would only result in removal of at max one log file.
+	// As an optimization, you could also immediately re-run it whenever it returns nil error
+	// (this is why `goto again` is used).
```
fracasula, line 168: This is true but it is also true that we should do this in periods of low activity as per Badger documentation. Usually these things are "Stop the world" events. This means that all activities on this Badger DB instance are stopped until the operation completes.
Since we're not running this during periods of low activity we can consider running it periodically without the `goto again`. @atzoum you played with badger more, any thoughs on this?

services/debugger/cache/embeddedcache/embeddedCache.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// Update writes the entries into badger db with a TTL
+func (e *EmbeddedCache[E]) Update(key string, value E) {
+	e.Init()
+	txn := e.Db.NewTransaction(true)
+	res := e.Read(key)
+	if len(res) >= e.Limiter {
+		res = res[len(res)-e.Limiter+1:]
+	}
```
fracasula, line 70: What are we doing here exactly? Can this truncation create issues somehow? I'd like to understand more.

services/debugger/cache/embeddedcache/embeddedCache.go
```diff
@@ -0,0 +1,192 @@
...
+			txn = e.Db.NewTransaction(true)
+			err = txn.SetEntry(entry)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+	_ = misc.RetryWith(context.TODO(), e.queryTimeout, e.retries, setBadgerEntry)
+	_ = txn.Commit()
```
fracasula, line 92: How come we're not returning errors here?

services/debugger/cache/embeddedcache/embeddedCache.go
cisse21: This part of the code is copied from badgerDB documentation so I just took that as is

services/debugger/cache/embeddedcache/embeddedCache.go
cisse21: This is basically to ensure that we don't store more than x amount of events into the db at any point in time. So I would just want to remove the oldest elements(typically the starting elements) and append the newer entries into the cache

services/debugger/cache/embeddedcache/embeddedCache.go
cisse21: We can throw an error and handle it at the parent function but live events is a functionality where we are okay with the server not panicking even if something goes wrong with the cache(unable to write to cache or read from cache) but I should probably add some error logs over here

services/debugger/cache/embeddedcache/embeddedCache.go
cisse21: Exported it just for the tests but yeah we probably don't need to export this

services/debugger/destination/eventDeliveryStatusUploader.go
cisse21: With the way the code is structured right now I guess it would need a significant amount of refactoring to ensure that we can pass them as parameters and moving away from config usages across the files. With this PR already significantly large does it make sense that we have it as a followup task which can be undertaken at a later point in time when we make changes with regards to apphandlers/runner?

services/debugger/cache/embeddedcache/embeddedCache.go
fracasula: You could return an error and let the parent log. I don't see a particular reason why a logger would be needed here.

services/debugger/cache/embeddedcache/embeddedCache.go
fracasula: Makes sense. Can you please add a comment?

services/debugger/cache/embeddedcache/embeddedCache.go
fracasula: We should promote cleaner APIs, unexport this and either have the test live in the same package or make the test not rely on direct calls to `Init()` which also sounds correct to me.

services/debugger/cache/embeddedcache/embeddedCache.go
fracasula: I know but in the same docs where you got that they are suggesting to run it during periods of low activity.

services/debugger/destination/eventDeliveryStatusUploader.go
fracasula: Can you please create a task to pick it up at the first convenience?

services/debugger/cache/embeddedcache/embeddedCache.go
cisse21: Done

services/debugger/cache/embeddedcache/embeddedCache.go
cisse21: Added a comment

services/debugger/cache/embeddedcache/embeddedCache.go
cisse21: Done

services/debugger/destination/eventDeliveryStatusUploader.go
cisse21: [Added it here](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=0d205985825a4eddbc6c616c291f8fc7&pm=s)

gateway/gateway_test.go
atzoum: verifies that the `tFunc` has been called

router/eventorder_test.go
```diff
@@ -145,8 +145,6 @@ func TestEventOrderGuarantee(t *testing.T) {
 			config.Set("Router.updateStatusBatchSize", 100)
 			config.Set("Router.maxStatusUpdateWait", "10ms")
 
-			defer config.Reset()
-
```
atzoum, line 149: resetting the singleton config instance after the test is done makes sure that overridden config values are reset and won't affect other tests

services/debugger/cache/listcache/listCache.go
```diff
@@ -0,0 +1,114 @@
+package listcache
```
atzoum, line 1: IMO `memcache` would be a more appropriate name

services/debugger/cache/embeddedcache/embeddedCache.go
```diff
@@ -0,0 +1,199 @@
+package embeddedcache
```
atzoum, line 1: IMO `badgercache` would be a more appropriate name

services/debugger/destination/eventDeliveryStatusUploader_test.go
```diff
@@ -161,45 +163,54 @@ var faultyData = DeliveryStatusT{
...
 	BeforeEach(func() {
+		config.Reset()
+		config.Set("RUDDER_TMPDIR", path.Join(GinkgoT().TempDir(), rand.String(10)))
+		config.Set("LiveEvent.cache.GCTime", "1s")
+
 		c = &eventDeliveryStatusUploaderContext{}
 		c.Setup()
-		deliveryStatus = DeliveryStatusT{
+
+		h = destinationdebugger.NewHandle(destinationdebugger.WithDisableEventUploads(false))
```
atzoum, line 198: Since we have two different cache implementations, all debugger (source, destination & transformation) tests should optimally run **twice**, once for every cache implementation

services/debugger/cache/embeddedcache/embeddedCache.go
```diff
@@ -0,0 +1,199 @@
...
+			WithIndexCacheSize(16 << 20). // 16mb
+			WithNumGoroutines(1).
+			WithNumMemtables(0).
+			WithBlockCacheSize(0)
+
+		e.db, err = badger.Open(opts)
+		if err != nil {
+			e.Logger.Errorf("Error while opening badgerDB: %v", err)
+			return
+		}
```
atzoum, line 163: We shouldn't start rudder-server in case of an error. We cannot ignore this and we shouldn't panic either:

1. `cache.New` should return an error along with `Cache[T]` (`EmbeddedCache` needs a proper producer function too and move away from this call-onceinit-in-every-method pattern)
2. The various debbuger (source, destination, transformation) `Start` methods should also propagate this error

services/debugger/cache/embeddedcache/embeddedCache.go
```diff
@@ -0,0 +1,199 @@
...
+func (badgerLogger) Errorf(format string, a ...interface{}) {
+	_, _ = fmt.Fprintf(os.Stderr, format, a...)
+}
+
+func (badgerLogger) Warningf(format string, a ...interface{}) {
+	_, _ = fmt.Fprintf(os.Stderr, format, a...)
+}
+
+// Update writes the entries into badger db with a TTL
+func (e *EmbeddedCache[E]) Update(key string, value E) error {
```
atzoum, line 64: reading the key and writing the same key should happen in the same transaction.

services/debugger/cache/embeddedcache/embeddedCache.go
```diff
@@ -0,0 +1,199 @@
...
+	return nil
+}
+
+// Read fetches all the entries for a given key from badgerDB
+func (e *EmbeddedCache[E]) Read(key string) []E {
+	e.Init()
+	read := func(ctx context.Context) ([]E, error) {
+		var values []E
+		err := e.db.View(func(txn *badger.Txn) error {
+			it := txn.NewIterator(badger.DefaultIteratorOptions)
```
atzoum, line 108: why do we need an iterator and why do we consider the key to be a prefix?

cisse21: Didn't check the documentation properly and assumed this was the only way to retrieve data

services/debugger/cache/embeddedcache/embeddedCache.go
cisse21: Missed this part... thanks for pointing out. Fixed this

services/debugger/cache/embeddedcache/embeddedCache.go
cisse21: I don't agree with this actually... why. shouldn't we start rudder-server just because we are unable to initialise the cache? All of the errors in the live events module are actually recovered in a defer recover function. Cache is just an auxiliary functionality and should affect any of the existing pipeline behaviour right

fracasula:
https://deepsource.io/gh/rudderlabs/rudder-server/run/6f91d58e-2cd4-4578-beb4-a04e3273d692/go/

services/debugger/destination/eventDeliveryStatusUploader_test.go
cisse21: Done

services/debugger/cache/badgercache/badgercache.go
```diff
@@ -0,0 +1,188 @@
+package embeddedcache
```
atzoum, line 1: Proposal:

directory tree
```
cache/
├─ internal/
│  ├─ memory/
│  │  ├─ memory.go
│  ├─ badger/
│  │  ├─ badger.go
├─ cache.go
```

badger.go
```go
package badger

type Cache[E any] struct {
...
```

memory.go
```go
package memory

type Cache[E any] struct {
...
```

cache.go
```go
package cache

import (
	"github.com/rudderlabs/rudder-server/services/debugger/cache/internal/badger"
	"github.com/rudderlabs/rudder-server/services/debugger/cache/internal/memory"
)

type Type int8

const (
	MemoryType Type = iota
	BadgerType
)

type Cache[T any] interface {
	Update(key string, value T) error
	Read(key string) []T
	Stop() error
}

func New[T any](ct Type, origin string, l logger.Logger) Cache[T] {
	switch ct {
	case BadgerType:
		l.Info("Using badger cache")
		return &badger.Cache[T]{Origin: origin, Logger: l}
	default:
		l.Info("Using in-memory cache")
		return &memory.Cache[T]{Origin: origin}
	}
}

```

services/debugger/cache/embeddedcache/embeddedCache.go
atzoum: The various debugger modules (source, destination, transformation)  should propagate initialization errors. Otherwise they are hiding that they will be malfunctioning. No, design-wise logging an error is not enough. If you wish to ignore such initialization errors in the apphandlers then do it there.

services/debugger/cache/badgercache/badgercache.go
```diff
@@ -0,0 +1,188 @@
...
+	e.Init()
+	read := func(ctx context.Context) ([]E, error) {
+		var values []E
+		err := e.db.View(func(txn *badger.Txn) error {
+			item, err := txn.Get([]byte(key))
+			if err != nil {
+				return err
+			}
+			var data []byte
+			data, err = item.ValueCopy(data)
```
atzoum, line 105: what's the purpose of this copy?

services/debugger/cache/badgercache/badgercache.go
```diff
@@ -0,0 +1,188 @@
...
+			}
+			return nil
+		})
+		if err != nil {
+			return nil, err
+		}
+		return values, nil
+	}
+
+	res, _ := misc.QueryWithRetries(context.TODO(), e.queryTimeout, e.retries, read)
```
atzoum, line 121: why do we need retries here?

services/debugger/cache/badgercache/badgercache.go
```diff
@@ -0,0 +1,188 @@
...
+		if err != nil {
+			return err
+		}
+		entry := badger.NewEntry([]byte(key), data).WithTTL(e.cleanupFreq)
+		if err := txn.SetEntry(entry); err != nil {
+			return err
+		}
+		return nil
+	}
+	err := misc.RetryWith(context.TODO(), e.queryTimeout, e.retries, setBadgerEntry)
```
atzoum, line 83: why are we retrying here?

services/debugger/cache/badgercache/badgercache.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return err
+	}
+	err = txn.Commit()
```
atzoum, line 87: we should retry here if the error is `ErrConflict`

services/debugger/cache/badgercache/badgercache.go
```diff
@@ -0,0 +1,188 @@
...
+func (badgerLogger) Warningf(format string, a ...interface{}) {
+	_, _ = fmt.Fprintf(os.Stderr, format, a...)
+}
+
+// Update writes the entries into badger db with a TTL
+func (e *EmbeddedCache[E]) Update(key string, value E) error {
+	e.Init()
+	txn := e.db.NewTransaction(true)
+	setBadgerEntry := func(ctx context.Context) error {
+		res := e.Read(key)
```
atzoum, line 68: we need to do a `txn.Get()` here

atzoum (changes requested):

services/debugger/cache/badgercache/badgercache.go
cisse21: was retrying all different errors instead of just retrying `ErrConflict`. Should not be an issue right with having retries?

atzoum: retry why? we don't want retry just for the sake of it. The actions being retried now don't return an `ErrConflict`.

services/debugger/cache/embeddedcache/embeddedCache.go
cisse21: Refactored the code to propagate the errors to the top

services/debugger/cache/badgercache/badgercache.go
cisse21: Done

services/debugger/cache/badgercache/badgercache.go
cisse21: Modified it to only retry for `ErrConflict`

services/debugger/cache/badgercache/badgercache.go
cisse21: Got rid of this and used `Value`

services/debugger/cache/badgercache/badgercache.go
cisse21: Got rid of this

services/debugger/cache/badgercache/badgercache.go
cisse21: Updated

services/debugger/cache/cache.go
cisse21: Got rid of the interface and the mock

services/debugger/cache/listcache/listCache.go
cisse21: Added this [here](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=2500e96bbcc94e059b08ac2d28b795da&pm=s)

services/debugger/cache/badgercache/badgercache.go
cisse21: Done the same

atzoum (approved):

BonapartePC (approved):

fracasula (approved):

PR #2834, opened 2022-12-26
chore: use badger v3 across rudder-server

cisse21:
# Description

Use badgerV3 across the code base and get rid of the v2 version of the code.

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=eda5349d1d6a4daab190762f33faa60d&pm=s)

## Security

### DIFF
--- a/services/dedup/dedup.go
+++ b/services/dedup/dedup.go
@@ -7,8 +7,8 @@ import (
 	"sort"
 	"time"
 
-	"github.com/dgraph-io/badger/v2"
-	"github.com/dgraph-io/badger/v2/options"
+	"github.com/dgraph-io/badger/v3"
+	"github.com/dgraph-io/badger/v3/options"
 
 	"github.com/rudderlabs/rudder-server/config"
 	"github.com/rudderlabs/rudder-server/rruntime"
@@ -50,7 +50,7 @@ func (l loggerForBadger) Warningf(fmt string, args ...interface{}) {
 }
 
 func DefaultRudderPath() string {
-	badgerPathName := "/badgerdbv2"
+	badgerPathName := "/badgerdbv3"
 	tmpDirPath, err := misc.CreateTMPDIR()
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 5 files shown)


services/dedup/dedup.go
```diff
@@ -111,25 +111,17 @@ func (d *DedupHandleT) openBadger() {
 
 	opts := badger.
 		DefaultOptions(d.path).
-		WithTruncate(true).
```
saurav-malani, line 114: I can see that `v3` doesn't have `WithTruncate()` func. Can you please clarify if it changes anything for us, since we are not setting it explicitly?

services/dedup/dedup.go
```diff
@@ -111,25 +111,17 @@ func (d *DedupHandleT) openBadger() {
...
-		// Memory usage optimizations:
-		// Inspired by https://github.com/dgraph-io/badger/issues/1304#issuecomment-630078745
-		// With modifications to ensure no performance degradation for dedup.
-		opts.TableLoadingMode = options.FileIO
-		opts.ValueLogLoadingMode = options.FileIO
 		opts.NumMemtables = 3
-		opts.MaxTableSize = 16 << 20
 		opts.NumLevelZeroTables = 1
 		opts.NumLevelZeroTablesStall = 2
-		opts.KeepL0InMemory = false
```
saurav-malani, line 132: Also, over here a lot of variable are no more present. Can you explain, how these things are being handled even if we don't sent them?

services/dedup/setup.go
```diff
@@ -23,3 +28,12 @@ func GetInstance(clearDB *bool) DedupI {
 
 	return dedupManager
 }
+
+func cleanUpV2Data() {
```
saurav-malani, line 32: I am having hard time understanding the use of this function. Can you please explain it's utility?

services/dedup/dedup.go
```diff
@@ -50,7 +50,7 @@ func (l loggerForBadger) Warningf(fmt string, args ...interface{}) {
 }
 
 func DefaultRudderPath() string {
-	badgerPathName := "/badgerdbv2"
+	badgerPathName := "/badgerdbv3"
 	tmpDirPath, err := misc.CreateTMPDIR()
 	if err != nil {
 		panic(err)
```
saurav-malani, line 56: Can't we return the error to the caller and handler it appropriately there without panicking?

saurav-malani (approved):

services/dedup/dedup.go
atzoum, line 132: We would need to perform a benchmark to verify that v3 with this configuration behaves the same like v2 in terms of performance & memory usage.

services/dedup/setup.go
```diff
@@ -23,3 +28,12 @@ func GetInstance(clearDB *bool) DedupI {
...
+	// clean up badgerDBv2 data
+	tmpDirPath, err := misc.CreateTMPDIR()
+	if err != nil {
+		return
+	}
+	_ = os.RemoveAll(fmt.Sprintf(`%v/badgerdbv2`, tmpDirPath))
```
atzoum, line 38: before deleting these, shouldn't we migrate them first to the new db?

services/dedup/dedup.go
cisse21: Yeah ... I am yet to do that.... have it on hold till then

services/dedup/setup.go
cisse21: I checked to official documentation and read somewhere that since the way data is stored changed v2 to v3 carrying out the migration would be very tough. Since dedup was not being used my many customers back then I went ahead with this but now this would require a bit more rethinking since we are planning to enable it for everyone from this release

PR #2836, opened 2022-12-29
fix(warehouse): added support for testing ssh tunnelling using postgres

abhimanyubabbar:
# Description

Adding integration test for ssh tunnelling using `Postgres` as the warehouse. This integration tests verifies events flowing into the warehouse.

## Notion Ticket

https://www.notion.so/rudderstacks/Add-warehouse-integration-test-for-ssh-tunnelling-393a41dc620b4ac18cccd3a9fd86ecbb

## Security


warehouse/postgres/postgres_test.go
```diff
@@ -15,6 +18,104 @@ import (
 	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
 )
 
+func TestIntegrationPostgresThroughTunnelling(t *testing.T) {
```
lvrach, line 21: General question: How do you mock the endpoint to get the secret private key?

achettyiitr (approved):

warehouse/postgres/postgres_test.go
abhimanyubabbar: The part to get the secret key is happening when we fetch the workspace json. Here we provide the workspace json mocked ourselves so no need to mock that function.

lvrach (approved):

PR #2848, opened 2023-01-05
chore: add workspace id in wh_async_jobs table

am6010:
# Description

We need to add the workspace id in the wh_async_jobs table to allow the transition to the new multitenant setup.
The workspace id will be added in the `AsyncJobPayloadT`.
Rudder-Sources will call this endpoint to trigger a job for deduplicating singer google sheets syncs.

## Notion Ticket

[ticket](https://www.notion.so/rudderstacks/Support-wh-multi-tenant-for-wh-async-jobs-api-4c8172eb09bc4b6c9eec2efcc93d4a53)

## Security


warehouse/jobs/runner.go
```diff
@@ -45,7 +45,7 @@ func (a *AsyncJobWhT) getTableNamesBy(sourceID, destinationID, jobRunID, taskRun
 	a.logger.Infof("[WH-Jobs]: Extracting table names for the job run id %s", jobRunID)
 	var tableNames []string
 	var err error
-	query := fmt.Sprintf(`SELECT id from %s where metadata->>'%s'=$1 and metadata->>'%s'=$2 and metadata in (SELECT metadata FROM wh_uploads where source_id=$3 and destination_id=$4)`, warehouseutils.WarehouseUploadsTable, "source_job_run_id", "source_task_run_id")
+	query := fmt.Sprintf(`SELECT id from %s where metadata->>'%s'=$1 and metadata->>'%s'=$2 and metadata in (SELECT metadata FROM %s where source_id=$3 and destination_id=$4)`, warehouseutils.WarehouseUploadsTable, "source_job_run_id", "source_task_run_id", warehouseutils.WarehouseUploadsTable)
```
lvrach, line 48: [minor] I believe it would be more readable this way
```suggestion
	query :=`SELECT id from`+warehouseutils.WarehouseUploadsTable+`where metadata->>'source_job_run_id'=$1 and metadata->>'source_task_run_id'=$2 and metadata in (SELECT metadata FROM ` + warehouseutils.WarehouseUploadsTable + ` where source_id=$3 and destination_id=$4)`
```

warehouse/jobs/utils.go
```diff
@@ -32,7 +32,7 @@ func getMessagePayloadsFromAsyncJobPayloads(asyncJobPayloads []AsyncJobPayloadT)
 }
 
 func validatePayload(payload StartJobReqPayload) bool {
-	if payload.SourceID == "" || payload.JobRunID == "" || payload.TaskRunID == "" || payload.DestinationID == "" {
+	if payload.SourceID == "" || payload.JobRunID == "" || payload.TaskRunID == "" || payload.DestinationID == "" || payload.WorkspaceID == "" {
```
lvrach, line 35: This validation would require sources to be deployed before this change is released in warehouse.

 Should we consider doing it in two phases to maintain forward compatibility?
 
 Phase 1: workspaceID is optional
 Phase 2: workspaceID is required

warehouse/jobs/runner.go
```diff
@@ -45,7 +45,7 @@ func (a *AsyncJobWhT) getTableNamesBy(sourceID, destinationID, jobRunID, taskRun
...
 	a.logger.Debugf("[WH-Jobs]: Query is %s\n", query)
 	rows, err := a.dbHandle.Query(query, jobRunID, taskRunID, sourceID, destinationID)
```
lvrach, line 50: We should exclude workspaceIDs that [are in degraded mode](https://github.com/rudderlabs/rudder-server/blob/exp.coveralls/warehouse/multitenant/manager.go#L76). We can also do it in a follow up PR.

lvrach (approved):

warehouse/jobs/utils.go
am6010: We don't call the async jobs endpoints yet from sources.  
I have the [pr](https://github.com/rudderlabs/rudder-sources/pull/798) ready for sources but to be safe I will set to optional and make it mandatory in the following pr.

warehouse/jobs/runner.go
am6010: done

warehouse/jobs/runner.go
```diff
@@ -45,7 +45,9 @@ func (a *AsyncJobWhT) getTableNamesBy(sourceID, destinationID, jobRunID, taskRun
 	a.logger.Infof("[WH-Jobs]: Extracting table names for the job run id %s", jobRunID)
 	var tableNames []string
 	var err error
-	query := fmt.Sprintf(`SELECT id from %s where metadata->>'%s'=$1 and metadata->>'%s'=$2 and metadata in (SELECT metadata FROM wh_uploads where source_id=$3 and destination_id=$4)`, warehouseutils.WarehouseUploadsTable, "source_job_run_id", "source_task_run_id")
+	query := fmt.Sprintf(`SELECT id from ` + warehouseutils.WarehouseUploadsTable +
```
achettyiitr, line 48: Can we format this query for better readability?

achettyiitr:
```Go
	query := `
		SELECT
		  id
		from
		where
		  metadata->>'source_job_run_id' = $1 and
          metadata->>'source_task_run_id' = $2 and
		  metadata in (
			SELECT
			  metadata
			FROM
			where
			  source_id = $3
			  and destination_id = $4
		  );
`
```
wdyt?

warehouse/jobs/runner.go
```diff
@@ -90,8 +92,8 @@ func (a *AsyncJobWhT) addJobsToDB(ctx context.Context, payload *AsyncJobPayloadT
 	}
 	a.logger.Infof("[WH-Jobs]: Adding job to the wh_async_jobs %s for tableName: %s", payload.MetaData, payload.TableName)
 
-	sqlStatement := fmt.Sprintf(`INSERT INTO %s (source_id, destination_id, tablename, status, created_at, updated_at, async_job_type, metadata)
-	VALUES ($1, $2, $3, $4, $5, $6 ,$7, $8 ) RETURNING id`, warehouseutils.WarehouseAsyncJobTable)
+	sqlStatement := fmt.Sprintf(`INSERT INTO %s (source_id, destination_id, tablename, status, created_at, updated_at, async_job_type, workspace_id, metadata)
```
achettyiitr, line 95:
```Go
	sqlStatement := `
		INSERT INTO ` + warehouseutils.WarehouseAsyncJobTable + ` (
		  source_id, destination_id, tablename,
		  status, created_at, updated_at, async_job_type,
		  workspace_id, metadata
		)
		VALUES
		  ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING id;
`
```

wdyt?

achettyiitr (approved):

warehouse/jobs/runner.go
am6010: Done

warehouse/jobs/runner.go
am6010: good idea

warehouse/jobs/runner.go
am6010: @lvrach thanks for reviewing.   I will raise another pr to check if the workspace is in degraded mode.

PR #2853, opened 2023-01-09
feat: add `total_bytes` field in staging file model

lvrach:
# Description

For staging_files, we want to keep `total_bytes` as metadata in addition to total_events.  Total bytes will be used to decide how staging files should be merged.

In this PR:
* total_bytes are computed during the file creation (before compression) in batch_router
* batch_router sends them to warehouse process endpoint
* warehouse API captures and the staging file repo stores and loads  the `total_bytes`

In addition to the main objective, I took the opportunity to move the warehouse client into a separate component under `warehouse/client/warehouse.go`. I plan to move all client/SDK logic related to the warehouse under that package.

**Noticeable change**: The process request payload contained unnecessary information. We marshalled the full source/destination where only the IDs were needed. You can check the [changes in the payload](https://github.com/rudderlabs/rudder-server/pull/2853/files#diff-379b6888a12fa80a1a91c706f7f52306e7676615e9fe6eac8b9e30f1e75e6504
) 

## Notion Ticket

https://www.notion.so/rudderstacks/Pass-staging-file-byte-size-event-count-e3884ae7dc0f4159be94700ab1ec5626

## Security


router/batchrouter/batchrouter.go
```diff
@@ -2297,8 +2286,10 @@ func (brt *HandleT) Setup(backendConfig backendconfig.BackendConfig, jobsDB, err
 		// error is ignored as context.TODO() is passed, err is not expected.
 		_ = brt.reporting.WaitForSetup(context.TODO(), types.CoreReportingClient)
 	}
-	if brt.warehouseURL == "" {
-		brt.warehouseURL = misc.GetWarehouseURL()
+	if brt.warehouseClient == nil {
+		brt.warehouseClient = client.NewWarehouse(misc.GetWarehouseURL(), client.WithTimeout(
+			config.GetDuration("WarehouseClient.timeout", 30, time.Second),
```
achettyiitr, line 2291:
```suggestion
			config.GetDuration("Warehouse.client.timeout", 30, time.Second),
```
Should it be something like this?

warehouse/client/warehouse.go
```diff
@@ -0,0 +1,141 @@
...
+	"net/http"
+	"time"
+
+	"github.com/rudderlabs/rudder-server/utils/httputil"
+)
+
+const defaultTimeout = 10 * time.Second
+
+// StagingFile contains the require metadata to process a staging file.
+type StagingFile struct {
```
achettyiitr, line 18: Since we are using this. Can we remove the StagingFile struct in [here](https://github.com/rudderlabs/rudder-server/blob/b7aa80003e237ddda370dff125898d4ac52ea820/warehouse/utils/utils.go#L266).

warehouse/client/warehouse.go
```diff
@@ -0,0 +1,141 @@
...
+	SourceBatchID   string
+	SourceTaskID    string
+	SourceTaskRunID string
+	SourceJobID     string
+	SourceJobRunID  string
+	TimeWindow      time.Time
+}
+
+type stagingFileBatchDestination struct {
+	Source      struct{ ID string }
```
achettyiitr, line 64: TIL: Are these anonymous structs?

router/batchrouter/batchrouter.go
```diff
@@ -673,10 +673,12 @@ func (brt *HandleT) copyJobsToStorage(provider string, batchJobs *BatchJobsT, is
 		if isDestInterrupted {
 			if _, ok = interruptedEventsMap[eventID]; !ok {
 				eventsFound = true
+				totalBytes += len(string(job.EventPayload) + "\n")
```
achettyiitr, line 676: Since we are doing `string(job.EventPayload) + "\n"` twice. This will be created twice?

warehouse/client/warehouse.go
```diff
@@ -0,0 +1,141 @@
...
+	Destination struct{ ID string }
+}
+
+type Warehouse struct {
+	baseURL string
+	client  http.Client
```
achettyiitr, line 70:
```suggestion
	client  *http.Client
```
Any specific reason we didn't use a pointer?

router/batchrouter/batchrouter.go
lvrach: Semantically, I felt that `Warehouse.client` -> means a client that runs in a warehouse.

Where `WarehouseClient` is the warehouse client in general.

warehouse/client/warehouse.go
lvrach: I don't know about the correct terminology here. 

When you use the keyword type you are creating an alias. But you can skip types and use the struct definition directly.

It is not recommended, but I wanted to avoid creating more types for it as it used for backwards compatibility only.

achettyiitr (approved):

Jayachand (approved):

PR #2855, opened 2023-01-10
fix(destination): add proper stats for rt/batch transformation & proxy

sanpj2292:
# Description

For the stats router_batch_num_output_jobs and router_transform_num_output_jobs we don't have workspaceId as one of the labels. And these stats are being used to calculate the percentage of failures
Without workspaceId information, in a multi-tenant kind of environment, it would be very difficult to find out from which workspace are we getting error from.
Also, for transformerProxy enabled destinations, we are using transformer_proxy.delivery_request to calculate the percentage of failures during delivery to destination. Even this stat doesn't have workspaceId

Hence made the changes below

Add workspaceId stat for transformer_proxy_delivery_request
Add a new stat router_transform_num_jobs which tracks the jobs for batch/router transformation, includes workspaceId

## Notion Ticket

https://www.notion.so/rudderstacks/Update-Prometheus-alert-definitions-9bd6abb291b94b61abd20fd52fbce806

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -294,7 +294,7 @@ func (worker *workerT) recordStatsForFailedTransforms(transformType string, tran
 			"transformType": transformType,
 			"statusCode":    strconv.Itoa(destJob.StatusCode),
 			"workspaceId":   destJob.Destination.WorkspaceID,
-			"destination":   destJob.Destination.ID,
+			"destinationId": destJob.Destination.ID,
 		}).Count(1)
 		if destJob.StatusCode != http.StatusOK {
 			transformFailedCountStat := stats.Default.NewTaggedStat("router_transform_num_failed_jobs", stats.CountType, stats.Tags{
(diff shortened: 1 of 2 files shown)


koladilip (approved):

saikumarrs (approved):

krishna2020 (approved):

router/router.go
```diff
@@ -288,6 +288,14 @@ func (worker *workerT) trackStuckDelivery() chan struct{} {
 
 func (worker *workerT) recordStatsForFailedTransforms(transformType string, transformedJobs []types.DestinationJobT) {
 	for _, destJob := range transformedJobs {
+		// Input Stats for batch/router transformation
+		stats.Default.NewTaggedStat("router_transform_num_jobs", stats.CountType, stats.Tags{
+			"destType":      worker.rt.destName,
+			"transformType": transformType,
+			"statusCode":    strconv.Itoa(destJob.StatusCode),
+			"workspaceId":   destJob.Destination.WorkspaceID,
+			"destination":   destJob.Destination.ID,
```
koladilip, line 297: should this be destinationId?

sanpj2292: Makes sense

cisse21 (approved):

Sidddddarth (approved):

atzoum (commented):
We already have an issue with existing metric cardinality in multitenant deployments, causing missing measurements & gaps. Just noting that this is expected to amplify the aforementioned issue.

PR #2859, opened 2023-01-10
exp: sourcegraph/conc

Sidddddarth:
# Description

exp: github.com/sourcegraph/conc/pool

## Notion Ticket

[exp](https://www.notion.so/rudderstacks/Refactor-gateway-stats-82ff63ac487341f3ac8b795df39c2f04)

## Security


processor/transformer/transformer.go
```diff
@@ -205,39 +206,29 @@ func (trans *HandleT) Transform(ctx context.Context, clientEvents []TransformerE
...
-	wg.Add(len(transformResponse))
+	p := pool.New().WithMaxGoroutines(maxConcurrency)
 	for i := range transformResponse {
-		i := i
-		from := i * batchSize
-		to := (i + 1) * batchSize
-		if to > len(clientEvents) {
-			to = len(clientEvents)
-		}
-		trans.guardConcurrency <- struct{}{}
```
atzoum, line 231: But we still need to have a `HandleT`-wide concurrency guard, right? In the future, multiple go routines might be calling `HandleT#Transform`. If this doesn't happen already, it definitely will happen in the future when we will be implementing isolation at processor.

Sidddddarth: you mean having this gaurd outside `Transform`..?

PR #2862, opened 2023-01-10
feat(warehouse): databricks unity catalog

achettyiitr:
# Description

- Added support for Unity catalog to handle catalog + schema + table.(three-level namespace)
- **Removed Global configs**. Added them as part of the **Clickhouse** struct for better testability.
- **logger** and **stats** as part of Handle.
- **Test coverage** (~55%).

## Notion Ticket

https://www.notion.so/rudderstacks/WH-Services-Team-Dashboard-ff1ac7f9ebfa42aabdacabe70ebe29ac

## Security


warehouse/deltalake/databricks/databricks.go
```diff
@@ -11,21 +11,14 @@ import (
...
-	UserAgentEntry  string
+type Credentials struct {
+	Host  string
+	Port  string
+	Path  string
+	Token string
 }
 
-type DBHandleT struct {
+type DatabricksClient struct {
```
achettyiitr, line 21: For better naming convention, using `DatabricksClient` instead of `DBHandleT`

warehouse/deltalake/databricks/databricks.go
```diff
@@ -11,21 +11,14 @@ import (
 
 var pkgLogger logger.Logger
 
-type CredentialsT struct {
-	Host            string
-	Port            string
-	Path            string
-	Token           string
-	Schema          string
```
achettyiitr, line 19: These are not being used. Removing these.

warehouse/deltalake/deltalake.go
```diff
@@ -1155,40 +1165,7 @@ func GetDatabricksVersion() (databricksBuildVersion string) {
 	return
 }
 
-// GetDatabricksVersion Gets the databricks version by making a grpc call to Version stub.
```
achettyiitr, line 1158: This is not required now. Earlier this was added because we didn't have databricks deployment. Now databricks deployment is by default present.

PR #2865, opened 2023-01-11
refactor: using an exponential backoff

Sidddddarth:
# Description

remove external backoff package

## Notion Ticket

[router backoff](https://www.notion.so/rudderstacks/router-backoff-be6e80e68f35478eab4922bb2e5618ac)

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"math"
 	"math/rand"
 	"net/http"
 	"sort"
@@ -14,7 +15,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/cenkalti/backoff/v4"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/tidwall/gjson"
 	"golang.org/x/sync/errgroup"
@@ -1132,18 +1132,8 @@ func (*workerT) sendDestinationResponseToConfigBackend(payload json.RawMessage, 
 	}
 }
 
-func durationBeforeNextAttempt(attempt int) (d time.Duration) {
-	b := backoff.NewExponentialBackOff()
-	b.InitialInterval = minRetryBackoff
-	b.MaxInterval = maxRetryBackoff
-	b.RandomizationFactor = 0
-	b.MaxElapsedTime = 0
-	b.Multiplier = 2
-	b.Reset()
-	for index := 0; index < attempt; index++ {
-		d = b.NextBackOff()
-	}
-	return
+func durationBeforeNextAttempt(attempt int) time.Duration {
+	return time.Duration(math.Min(float64(maxRetryBackoff), float64(minRetryBackoff)*math.Exp2(float64(attempt-1))))
 }
 
 func (rt *HandleT) trackRequestMetrics(reqMetric requestMetric) {
--- a/router/router.go
+++ b/router/router.go
@@ -1133,6 +1133,9 @@ func (*workerT) sendDestinationResponseToConfigBackend(payload json.RawMessage, 
 }
 
 func durationBeforeNextAttempt(attempt int) time.Duration {
+	if attempt < 1 {
+		attempt = 1
+	}
 	return time.Duration(math.Min(float64(maxRetryBackoff), float64(minRetryBackoff)*math.Exp2(float64(attempt-1))))
 }
 
--- a/router/router_test.go
+++ b/router/router_test.go
@@ -159,6 +159,7 @@ func initRouter() {
 
 func TestBackoff(t *testing.T) {
 	loadConfig()
+	assert.Equal(t, 10*time.Second, durationBeforeNextAttempt(0))
 	assert.Equal(t, 10*time.Second, durationBeforeNextAttempt(1))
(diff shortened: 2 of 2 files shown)


router/router.go
```diff
@@ -1133,16 +1133,20 @@ func (*workerT) sendDestinationResponseToConfigBackend(payload json.RawMessage,
...
+			1 - 10
+			2 - 18.17
+			3 - 26.84
+			4 - 35.25
+			5 - 43.39
+			6 - 51.27
+	*/
+	exponentialMultiplier := 1 - math.Exp(-float64(attempt)*0.03125)
+	exponentialDuration := float64(maxRetryBackoff) * exponentialMultiplier
+	d = time.Duration(math.Max(constantMin, exponentialDuration))
```
atzoum, line 1149: Given `minRetryBackoff`, `maxRetryBackoff`, `multiplier: 2` and `randomizationFactor: 0`, shouldn't it look like below?

```
nextBackoff := time.Duration(math.Min(float64(maxRetryBackoff), float64(minRetryBackoff)*math.Exp2(float64(attempt))))
```

atzoum: we need some tests too :)

Sidddddarth: added some

router/router_test.go
```diff
@@ -157,6 +157,57 @@ func initRouter() {
 	InitRouterAdmin()
 }
 
+func TestBackoff(t *testing.T) {
```
fracasula, line 160: By looking at these tests it is not clear to me how `minRetryBackoff` and `maxRetryBackoff` affect the results. Can we have them as parameters?

If I run the old algorithm with `10s` and `300s` for example I get the following:

```go
func TestBackoff(t *testing.T) {
	minRetryBackoff := 10 * time.Second
	maxRetryBackoff := 300 * time.Second

	b := backoff.NewExponentialBackOff()
	b.InitialInterval = minRetryBackoff
	b.MaxInterval = maxRetryBackoff
	b.RandomizationFactor = 0
	b.MaxElapsedTime = 0
	b.Multiplier = 2
	b.Reset()
	for index := 0; index < 10; index++ {
		d := b.NextBackOff()
		t.Logf("attempt: %d, duration: %s", index+1, d)
	}
}
```

Which gives:

```
=== RUN   TestBackoff
    backoff_test.go:23: attempt: 1, duration: 10s
    backoff_test.go:23: attempt: 2, duration: 20s
    backoff_test.go:23: attempt: 3, duration: 40s
    backoff_test.go:23: attempt: 4, duration: 1m20s
    backoff_test.go:23: attempt: 5, duration: 2m40s
    backoff_test.go:23: attempt: 6, duration: 5m0s
    backoff_test.go:23: attempt: 7, duration: 5m0s
    backoff_test.go:23: attempt: 8, duration: 5m0s
    backoff_test.go:23: attempt: 9, duration: 5m0s
    backoff_test.go:23: attempt: 10, duration: 5m0s
--- PASS: TestBackoff (0.00s)
```

Sidddddarth:
TIL I had a very different understanding as to what exponential backoff meant.
Based on both comments I think it means just doubling(or `multiplier`ing) till we reach a `maxRetryBackoff`.
We can scrap this and go with what @atzoum mentioned above as it's more in line with what's expected..?

router/router.go
```diff
@@ -1133,19 +1133,22 @@ func (*workerT) sendDestinationResponseToConfigBackend(payload json.RawMessage,
...
+	/*
+		1 - e^(-multiplier*(min/max)*attempt)
+	*/
+	exponentialMultiplier := 1 - math.Exp(-float64(attempt)*2*(float64(minRetryBackoff)/float64(maxRetryBackoff)))
+	exponentialDuration := int(float64(maxRetryBackoff) * exponentialMultiplier)
+	d = getMaxDuration(minRetryBackoff, time.Duration(exponentialDuration)).Round(time.Second)
 	return
 }
 
+func getMaxDuration(durA, durB time.Duration) time.Duration {
```
cisse21, line 1145: This can be a generic function in misc.go

cisse21 (approved):

router/router.go
```diff
@@ -1132,18 +1132,8 @@ func (*workerT) sendDestinationResponseToConfigBackend(payload json.RawMessage,
...
-	b.RandomizationFactor = 0
-	b.MaxElapsedTime = 0
-	b.Multiplier = 2
-	b.Reset()
-	for index := 0; index < attempt; index++ {
-		d = b.NextBackOff()
-	}
-	return
+func durationBeforeNextAttempt(attempt int) time.Duration {
+	return time.Duration(math.Min(float64(maxRetryBackoff), float64(minRetryBackoff)*math.Exp2(float64(attempt-1))))
```
atzoum, line 1136: It is not clear if this can be called with attempt==0 or not, thus can we guard against such scenario & add a test scenario too?
```suggestion
	if attempt < 1 {
		attempt = 1
	}
	return time.Duration(math.Min(float64(maxRetryBackoff), float64(minRetryBackoff)*math.Exp2(float64(attempt-1))))
```

koladilip:
What is the reason for this change?

Sidddddarth:
> What is the reason for this change?

https://rudderlabs.slack.com/archives/C037TBN33EW/p1673421553480469

atzoum (approved):

PR #2877, opened 2023-01-13
fix: regulation-worker panics during startup

atzoum:
# Description

Initializing diagnostics and disabling it

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=133fd11d91464897b804a33fa504e9b4&pm=s)

## Security


saurav-malani (approved):
Some of the tests are failing, please take a look at them.

saurav-malani:
I believe we should remove `diagnostics.Init()` from here, so that test can capture it in future.

https://github.com/rudderlabs/rudder-server/blob/a80c902dd8990f01690e15b7d6fa275f7ce0885f/regulation-worker/cmd/main_test.go#L175

fracasula (approved):

cisse21 (approved):

cisse21:
Ideally we should not even be needing initialisation of diagnostics since regulation worker does not need this component. Probably in future we should be able to get rid of initialisation of diagnostics from the modules where it is not required

PR #2879, opened 2023-01-16
chore: added prefix to regulation worker stats

saurav-malani:
# Description

Added prefix `regulation_worker` to all the regulation-worker stats.

### DIFF
--- a/regulation-worker/internal/client/client.go
+++ b/regulation-worker/internal/client/client.go
@@ -52,7 +52,7 @@ func (j *JobAPI) Get(ctx context.Context) (model.Job, error) {
 	req.Header.Set("Content-Type", "application/json")
 	reqTime := time.Now()
 	resp, err := j.Client.Do(req)
-	stats.Default.NewStat("regulation_manager.request_time", stats.TimerType).Since(reqTime)
+	stats.Default.NewTaggedStat("regulation_manager.request_time", stats.TimerType, stats.Tags{"reqType": http.MethodGet}).Since(reqTime)
 	if os.IsTimeout(err) {
 		stats.Default.NewStat("regulation_manager.request_timeout", stats.CountType).Count(1)
 		return model.Job{}, model.ErrRequestTimeout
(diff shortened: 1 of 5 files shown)


regulation-worker/internal/delete/api/api.go
```diff
@@ -81,7 +81,7 @@ func (api *APIManager) deleteWithRetry(ctx context.Context, job model.Job, desti
 	}
 
 	defer stats.Default.NewTaggedStat(
-		"file_cleaning_time",
+		"regulation_worker_file_cleaning_time",
```
atzoum, line 84: The name of this measurement is weird. What file is it referring to?

regulation-worker/internal/client/client.go
```diff
@@ -75,7 +75,7 @@ func (j *JobAPI) Get(ctx context.Context) (model.Job, error) {
 			return model.Job{}, fmt.Errorf("error while decoding job: %w", err)
 		}
 
-		userCountPerJob := stats.Default.NewTaggedStat("user_count_per_job", stats.CountType, stats.Tags{"jobId": jobSchema.JobID, "ID": j.Identity.ID()})
+		userCountPerJob := stats.Default.NewTaggedStat("regulation_worker_user_count_per_job", stats.CountType, stats.Tags{"jobId": jobSchema.JobID, "ID": j.Identity.ID()})
```
atzoum, line 78: Can we add timers for all http requests that we perform against data-regulation-service (`Client.Do(req)`)?

regulation-worker/internal/delete/api/api.go
saurav-malani: for `batch` destination, it refers to the files that we clean during the deletion process. But, I believe this name doesn't makes sense in case of `api` & `kvstore` destination. 
Do you think `regulation_worker_cleaning_time` makes sense? with `destType` having module name?

regulation-worker/internal/client/client.go
```diff
@@ -50,8 +50,9 @@ func (j *JobAPI) Get(ctx context.Context) (model.Job, error) {
 	}
 	req.SetBasicAuth(j.Identity.BasicAuth())
 	req.Header.Set("Content-Type", "application/json")
-
+	reqTime := time.Now()
 	resp, err := j.Client.Do(req)
+	stats.Default.NewStat("regulation_manager.request_time", stats.TimerType).Since(reqTime)
```
atzoum, line 55: either use a separate measurement name, or use a tag to denote the request type

saurav-malani: added it as tag.

BonapartePC (approved):

cisse21 (approved):

PR #2881, opened 2023-01-17
refactor(processor): stop using global variables

atzoum:
# Description

Preparation steps before starting the actual work for introducing workspace isolation in processor:

- Removing global variables from the `processor` package
- Renaming `processor.HandleT` to `processor.Handle`
- Introducing `misc.AsyncInit` for streamlining how our components wait for their proper initialization before actually starting working (hint: `backendconfig.WaitForConfig` is not 100% safe)
- Make sure that processor is properly initialized before starting (using `misc.AsyncInit`)
- Fix issue with multiple successful HTTP requests performed for transformer features when starting up

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=4f93b562e40743fcad2ce119dcacdc84&pm=s)

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -46,14 +46,27 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/types"
 )
 
+const (
+	MetricKeyDelimiter = "!<<#>>!"
+	UserTransformation = "USER_TRANSFORMATION"
+	DestTransformation = "DEST_TRANSFORMATION"
+	EventFilter        = "EVENT_FILTER"
+)
+
 var jsonfast = jsoniter.ConfigCompatibleWithStandardLibrary
 
 func RegisterAdminHandlers(readonlyProcErrorDB jobsdb.ReadonlyJobsDB) {
 	admin.RegisterAdminHandler("ProcErrors", &stash.StashRpcHandler{ReadOnlyJobsDB: readonlyProcErrorDB})
 }
 
-// HandleT is a handle to this object used in main.go
-type HandleT struct {
+func NewHandle(transformer transformer.Transformer) *Handle {
+	h := &Handle{transformer: transformer}
+	h.loadConfig()
+	return h
+}
+
+// Handle is a handle to the processor module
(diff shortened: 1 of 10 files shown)


processor/processor.go
```diff
@@ -46,14 +46,27 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/types"
 )
 
+const (
+	METRICKEYDELIMITER  = "!<<#>>!"
+	USER_TRANSFORMATION = "USER_TRANSFORMATION"
+	DEST_TRANSFORMATION = "DEST_TRANSFORMATION"
+	EVENT_FILTER        = "EVENT_FILTER"
```
fracasula, line 53: Shall we do like most std library constants and use the camel-case approach?

utils/misc/asyncinit.go
```diff
@@ -0,0 +1,54 @@
...
+func NewAsyncInit(count int64) *AsyncInit {
+	a := &AsyncInit{
+		c: make(chan struct{}),
+	}
+	a.count.Store(count)
+	return a
+}
+
+// AsyncInit is a helper object to wait for multiple asynchronous initialization events.
+type AsyncInit struct {
```
fracasula, line 19: It looks like `WaitContext` is used in tests only. Why not use a `WaitGroup` instead of creating this new `AsyncInit` type?

saurav-malani (approved):

utils/misc/asyncinit.go
atzoum: Using a wait group was my initial approach, but `WaitGroup#Wait` is a blocking call, whereas listening from a channel doesn't have to be. Consider this as a `WaitGroup` using a channel for communicating completion instead of blocking.

fracasula: Makes sense, it's not too complex either so :+1:

fracasula (approved):

PR #2886, opened 2023-01-19
chore: disable batch destinations by default in regulation worker

atzoum:
# Description

Disable batch destinations by default through a configuration option

## Security

### DIFF
--- a/regulation-worker/internal/delete/batch/batch.go
+++ b/regulation-worker/internal/delete/batch/batch.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff"
+	"github.com/rudderlabs/rudder-server/config"
 	"github.com/rudderlabs/rudder-server/regulation-worker/internal/delete/batch/filehandler"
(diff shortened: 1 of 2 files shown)


abhimanyubabbar (approved):

regulation-worker/internal/delete/batch/batch.go
```diff
@@ -308,7 +309,10 @@ type BatchManager struct {
 }
 
 func (*BatchManager) GetSupportedDestinations() []string {
-	return supportedDestinations
+	if config.Default.GetBool("REGULATION_WORKER_BATCH_DESTINATIONS_ENABLED", false) {
```
saurav-malani, line 312: I think the default value for it should be `true`. Disabling it should be a special case.

atzoum: IMO we cannot enable it until we have proper guards in the worker to make sure that jobs will not take forever to complete.

lvrach (approved):

PR #2888, opened 2023-01-19
chore(warehouse): alter handling for redshift

achettyiitr:
# Description

### Adding support for altering a particular column using these steps:
- Create a staging column.
- Populate the staging column from the original column.
- Rename the original column to the deprecated column.
- Rename the staging column to the original column.
- Drop the deprecated column.
- Handling errors while dropping columns in case of views/rules attached to the table.

### Adding Alerta component:
- It is used to send `/alerta `call to the **alerta service** which is then used by the **notification service** to send notifications to the clients.

### Additional changes
- **Removed Global configs**. Added them as part of the **Redshift** struct for better testability.
- **logger** as part of Handle.

### Dependent PR
https://github.com/rudderlabs/rudder-transformer/pull/1809

## Notion Ticket

https://www.notion.so/rudderstacks/Figure-out-way-to-set-varchar-max-for-new-customers-Establish-if-it-is-ok-from-cost-impact-perfo-809f7548af144562ad0aafa4e5035f75

## Security


services/alerta/client.go
```diff
@@ -0,0 +1,218 @@
...
+	defaultSeverity   = SeverityCritical
+	defaultPriority   = PriorityP1
+	defaultTeam       = "No Team"
+)
+
+type Client interface {
+	SendAlert(ctx context.Context, resource string, service Service) error
+}
+
+type ClientImpl struct {
```
lvrach, line 66: [minor] naming suggestion:

```suggestion
type AlertSender interface {
	SendAlert(ctx context.Context, resource string, service Service) error
}

type Client struct {
```

Also consider defining the interface on the consumer.

services/alerta/client.go
```diff
@@ -0,0 +1,218 @@
...
+func NewClient(baseURL string, fns ...OptFn) *ClientImpl {
+	c := &ClientImpl{
+		url: baseURL,
+		client: &http.Client{
+			Timeout: defaultTimeout,
+		},
+
+		retries:  defaultMaxRetries,
+		severity: defaultSeverity,
+		priority: defaultPriority,
```
lvrach, line 134: severity and priority don't match with the client lifecycle. For example, should we create a new client if we want to send a different severity/priority event?

I would suggest removing them and moving them to `SendAlert` method.

```suggestion
```

achettyiitr: Great point.

warehouse/upload.go
```diff
@@ -824,19 +828,76 @@ func (job *UploadJobT) updateTableSchema(tName string, tableSchemaDiff warehouse
...
+
+		query := strings.Join(queries, "\n")
+		err := alerta.NewClient(
+			job.AlertaURL,
+			alerta.WithTeam(warehouseutils.WAREHOUSE),
+			alerta.WithTags(alerta.Tags{
+				fmt.Sprintf("destID=%s", job.upload.DestinationID),
+				fmt.Sprintf("destType=%s", job.upload.DestinationType),
+				fmt.Sprintf("query=%s", query),
+			}),
```
lvrach, line 884: Should the tags be part of the NewClient options?

Also, what if we use map[string]string instead of []string for tags, seems a better abstraction:  
```suggestion
			alerta.WithTags(alerta.Tags{
				"destID": job.upload.DestinationID,
			         "destType": job.upload.DestinationType,
			         "query": query,
			}),
```

warehouse/upload.go
```diff
@@ -824,19 +828,76 @@ func (job *UploadJobT) updateTableSchema(tName string, tableSchemaDiff warehouse
...
 
-	return err
+	if len(responseToAlerta) > 0 {
+		queries := make([]string, len(responseToAlerta))
+		for i, res := range responseToAlerta {
+			queries[i] = res.Query
+		}
+
+		query := strings.Join(queries, "\n")
+		err := alerta.NewClient(
```
lvrach, line 877: Ideally, we don't want to create `alerta.NewClient` every time in the method. We should create the client in the "main" function and pass it down with Dependency injection. (Or else use a factory and DI the factory, but I don't think a factory is needed)

A few issues:
* We will create a new HTTP client pool every time, and not reuse TCP connections, thus creating tcp connections every time.
* We can need to spin up a test http server to test it. If you used a simple interface, testing would be easier.

services/alerta/client.go
```diff
@@ -0,0 +1,203 @@
...
+	var opts backoff.BackOff
+
+	opts = backoff.NewExponentialBackOff()
+	opts = backoff.WithMaxRetries(opts, uint64(c.retries))
+	opts = backoff.WithContext(opts, ctx)
+
+	return backoff.Retry(fn, opts)
+}
+
+func (c *Client) SendAlert(ctx context.Context, resource string, service Service, severity Severity, priority Priority) error {
```
lvrach, line 137:
```suggestion
type SendAlertOpts struct {
 Service Service
 Severity Severity
 Priority Priority
 Tags Tags
}

func (c *Client) SendAlert(ctx context.Context, resource string, opts SendAlertOpts) error {
```

lvrach: OR 

```
type Alert struct {
 Resource string
 Service Service
 Severity Severity
 Priority Priority
 Tags Tags
}
```

```suggestion
func (c *Client) SendAlert(ctx context.Context, alert Alert) error {
```

lvrach (approved):
I have some non-blocking comments for the alert client.

warehouse/upload.go
achettyiitr: Actually for alerts Tags should be a list.
https://github.com/rudderlabs/alerta/blob/039c1fb04b48615efea54844ae0994404d16d4a8/alerta/models/alert.py#L121

warehouse/upload.go
```diff
@@ -824,19 +832,78 @@ func (job *UploadJobT) updateTableSchema(tName string, tableSchemaDiff warehouse
...
+	for columnName, columnType := range columnsMap {
+		res, err := job.whManager.AlterColumn(tName, columnName, columnType)
 		if err != nil {
-			pkgLogger.Errorf("Altering column %s in table: %s.%s failed. Error: %v", columnName, job.warehouse.Namespace, tName, err)
-			break
+			errs = append(errs, err)
+			continue
 		}
+
+		if res.IsDependent {
```
abhimanyubabbar, line 857: What does this check mean here ?

services/alerta/client_test.go
```diff
@@ -0,0 +1,213 @@
...
+			require.NoError(t, err)
+			require.Equal(t, wantBody, string(body))
+
+			w.WriteHeader(http.StatusCreated)
+		}))
+		defer s.Close()
+
+		ctx := context.Background()
+
+		err := alertaClient(s,
```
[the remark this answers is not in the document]
achettyiitr: We are also using s.url as well.
```Go
alertaClient := func(s *httptest.Server, o ...alerta.OptFn) alerta.AlertSender {
		var op []alerta.OptFn
		op = append(op, o...)
		op = append(op, alerta.WithHTTPClient(s.Client()))
		op = append(op, alerta.WithTeam(testTeam))
		op = append(op, alerta.WithKubeNamespace(testNamespace))
		op = append(op, alerta.WithAlertTimeout(testAlertTimeout))
		op = append(op, alerta.WithEnvironment(testEnvironment))
		op = append(op, alerta.WithConfig(c))

		return alerta.NewClient(
			s.URL,
			op...,
		)
	}
```

services/alerta/client_test.go
```diff
@@ -0,0 +1,213 @@
...
+			testResource,
+			alerta.SendAlertOpts{
+				Severity: testSeverity,
+				Priority: testPriority,
+				Service:  testService,
+				Text:     testText,
+				Tags:     testTags,
+			},
+		)
+		require.EqualError(t, err, "non retriable: unexpected status code 400: ")
```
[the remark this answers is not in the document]
achettyiitr: We are logging body in case of unexpected status code,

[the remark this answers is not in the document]
achettyiitr: Updated the test case.

warehouse/upload.go
achettyiitr:
```Go
type AlterTableResponse struct {
	IsDependent bool // true if the column is dependent on another view or rules, false otherwise
	Query       string
}
```

We have added this component which is being used when we do the alter. This is being set to true, if the dropping column is dependent on another view or rules.

warehouse/integrations/redshift/redshift.go
```diff
@@ -656,45 +653,172 @@ func (rs *HandleT) dropDanglingStagingTables() bool {
...
+		rs.Logger.Infof("RS: Skipping creating schema: %s since it already exists", rs.Namespace)
 		return
 	}
 	return rs.createSchema()
 }
 
-func (rs *HandleT) AlterColumn(tableName, columnName, columnType string) (err error) {
-	if setVarCharMax && columnType == "text" {
-		err = rs.alterStringToText(fmt.Sprintf(`%q.%q`, rs.Namespace, tableName), columnName)
+func (rs *Redshift) AlterColumn(tableName, columnName, columnType string) (model.AlterTableResponse, error) {
```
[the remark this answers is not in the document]
achettyiitr: Is there certain scenarios should I verify?

warehouse/integrations/redshift/redshift.go
```diff
@@ -656,45 +653,172 @@ func (rs *HandleT) dropDanglingStagingTables() bool {
...
+	}
+
+	// populating staging column
+	query = fmt.Sprintf(`
+		UPDATE
+		  %[1]q.%[2]q
+		SET
+		  %[3]q = CAST (%[4]q AS %[5]s)
+		WHERE
+		  %[4]q IS NOT NULL;
```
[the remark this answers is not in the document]
achettyiitr: Nothing in particular. Although if the tables contain most null, we can avoid updating them.

abhimanyubabbar (approved):

PR #2891, opened 2023-01-20
feat(processor): add support for workspace-level isolation

atzoum:
# Description

Introducing support for workspace isolation in the processor module.

## Design Doc
[Link](https://www.notion.so/rudderstacks/Design-Proposal-Processor-Isolation-2db79f6c5b0f4a9fb14465b62aaaf54a)

Untitled

### `utils` package
- `misc.ExponentialNumber` is used by processor to  sleep for exponentially increasing durations while no jobs are being picked up.
- `queue.PriorityQueue` implements a priority queue pretty close to the standard one [here](https://pkg.go.dev/container/heap#example-package-PriorityQueue).
- `sync.Limiter` provides a concurrency limiter which supports dynamic priorities (i.e. priority increases periodically as time goes by for avoiding starvation of lower priority routines) and captures useful statistics. It is inspired by [vivek-ng/concurrency-limiter](https://github.com/vivek-ng/concurrency-limiter).

### `processor` package
- `isolation.Strategy` provides both the interface and the 2 different isolation strategy implementations. This shall be the only extension point for introducing additional isolation strategies in the future.
- `processor.worker`  implements the processor worker as described in the design doc.
- `processor.workerPool` implements the worker pool as described in the design doc.
- `processor.workerHandle` is the interface that abstracts `processor.Handle` implementation from the worker by including only the required methods.
- `processor.Handle#Start` starts the pinger loop as described in the design doc.

## Workspace Priorities

By default, all workspace have the same (low) priority with respect to their chance of getting picked up by the various concurrency limiters (get jobs, preprocess, transform, store) when there is a waiting queue. It is possible though to treat a workspace with higher priority, by setting the `Processor.Limiter.<WorkspaceID>.Priority` configuration property to a value between 2-4:
- 1: low priority
- 2: medium priority
- 3: medium/high priority
- 4: high priority

To make sure that low-priority workspaces will not be starved, every second, their priority will be increasing by 1 until it reaches to 4.

## Testing & Benchmarks

To ease testing of multi-workspace deployments (a.k.a multi-tenant) without the need for a properly seeded etcd server, support has been added for forcing the use of a static `ModeProvider` through the `forceStaticModeProvider` config property.

Workspace isolation overall behaviour and performance testing is performed in `processor_isolation_test.go`.  In comparison to the "no isolation" mode, workspace isolation:
- Performs **equally well** with up to **200 active** workspaces. 
- With **500 active** workspaces it is **18% slower**.
- With **1000 active** workspaces it is **30% slower**. 
- With **10000 active** workspaces it is **150% slower**.

To provide some perspective, in our current production workloads, the average number of active workspaces is not greater than 50 & the same number for destinations is between100-150.

A visualisation dashboard which demonstrates both the new statistics that are being captured, as well as the server's performance during the benchmarks can be found [here](https://snapshots.raintank.io/dashboard/snapshot/9643HY7OqrsYae1HuAjAz167953NOCN3).

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=1bd87345dc654c26ab2fb9e92d98f7df&pm=s)

## Security

### DIFF
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/rudderlabs/rudder-server/app"
 	"github.com/rudderlabs/rudder-server/app/cluster"
-	"github.com/rudderlabs/rudder-server/app/cluster/state"
 	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
 	"github.com/rudderlabs/rudder-server/gateway"
 	"github.com/rudderlabs/rudder-server/jobsdb"
@@ -35,7 +34,6 @@ import (
 	"github.com/rudderlabs/rudder-server/services/transientsource"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 	"github.com/rudderlabs/rudder-server/utils/types"
-	"github.com/rudderlabs/rudder-server/utils/types/servermode"
 )
 
 // embeddedApp is the type for embedded type implementation
@@ -45,26 +43,20 @@ type embeddedApp struct {
 	versionHandler func(w http.ResponseWriter, r *http.Request)
 	log            logger.Logger
 	config         struct {
-		enableProcessor    bool
-		enableRouter       bool
 		enableReplay       bool
 		processorDSLimit   int
 		routerDSLimit      int
 		batchRouterDSLimit int
 		gatewayDSLimit     int
-		forceStaticMode    bool
 	}
 }
 
 func (a *embeddedApp) loadConfiguration() {
-	config.RegisterBoolConfigVariable(true, &a.config.enableProcessor, false, "enableProcessor")
 	config.RegisterBoolConfigVariable(types.DefaultReplayEnabled, &a.config.enableReplay, false, "Replay.enabled")
-	config.RegisterBoolConfigVariable(true, &a.config.enableRouter, false, "enableRouter")
 	config.RegisterIntConfigVariable(0, &a.config.processorDSLimit, true, 1, "Processor.jobsDB.dsLimit", "JobsDB.dsLimit")
 	config.RegisterIntConfigVariable(0, &a.config.gatewayDSLimit, true, 1, "Gateway.jobsDB.dsLimit", "JobsDB.dsLimit")
 	config.RegisterIntConfigVariable(0, &a.config.routerDSLimit, true, 1, "Router.jobsDB.dsLimit", "JobsDB.dsLimit")
 	config.RegisterIntConfigVariable(0, &a.config.batchRouterDSLimit, true, 1, "BatchRouter.jobsDB.dsLimit", "JobsDB.dsLimit")
-	config.RegisterBoolConfigVariable(false, &a.config.forceStaticMode, false, "forceStaticModeProvider")
 }
 
 func (a *embeddedApp) Setup(options *app.Options) error {
@@ -198,30 +190,9 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 		}))
 	}
 
-	var modeProvider cluster.ChangeEventProvider
-
-	staticModeProvider := func() cluster.ChangeEventProvider {
-		// FIXME: hacky way to determine server mode
-		if a.config.enableProcessor && a.config.enableRouter {
-			return state.NewStaticProvider(servermode.NormalMode)
-		}
-		return state.NewStaticProvider(servermode.DegradedMode)
-	}
-
-	if a.config.forceStaticMode {
-		a.log.Info("forcing the use of Static Cluster Manager")
-		modeProvider = staticModeProvider()
-	} else {
-		switch deploymentType {
-		case deployment.MultiTenantType:
-			a.log.Info("using ETCD Based Dynamic Cluster Manager")
-			modeProvider = state.NewETCDDynamicProvider()
-		case deployment.DedicatedType:
-			a.log.Info("using Static Cluster Manager")
-			modeProvider = staticModeProvider()
-		default:
-			return fmt.Errorf("unsupported deployment type: %q", deploymentType)
-		}
+	modeProvider, err := resolveModeProvider(a.log, deploymentType)
+	if err != nil {
+		return err
 	}
 
 	adaptiveLimit := payload.SetupAdaptiveLimiter(ctx, g)
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -52,6 +52,7 @@ type embeddedApp struct {
 		routerDSLimit      int
 		batchRouterDSLimit int
 		gatewayDSLimit     int
+		forceStaticMode    bool
 	}
 }
 
@@ -63,6 +64,7 @@ func (a *embeddedApp) loadConfiguration() {
 	config.RegisterIntConfigVariable(0, &a.config.gatewayDSLimit, true, 1, "Gateway.jobsDB.dsLimit", "JobsDB.dsLimit")
 	config.RegisterIntConfigVariable(0, &a.config.routerDSLimit, true, 1, "Router.jobsDB.dsLimit", "JobsDB.dsLimit")
 	config.RegisterIntConfigVariable(0, &a.config.batchRouterDSLimit, true, 1, "BatchRouter.jobsDB.dsLimit", "JobsDB.dsLimit")
+	config.RegisterBoolConfigVariable(false, &a.config.forceStaticMode, false, "forceStaticModeProvider")
 }
 
 func (a *embeddedApp) Setup(options *app.Options) error {
@@ -198,20 +200,28 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 
 	var modeProvider cluster.ChangeEventProvider
 
-	switch deploymentType {
-	case deployment.MultiTenantType:
-		a.log.Info("using ETCD Based Dynamic Cluster Manager")
-		modeProvider = state.NewETCDDynamicProvider()
-	case deployment.DedicatedType:
+	staticModeProvider := func() cluster.ChangeEventProvider {
 		// FIXME: hacky way to determine server mode
-		a.log.Info("using Static Cluster Manager")
 		if a.config.enableProcessor && a.config.enableRouter {
-			modeProvider = state.NewStaticProvider(servermode.NormalMode)
-		} else {
-			modeProvider = state.NewStaticProvider(servermode.DegradedMode)
+			return state.NewStaticProvider(servermode.NormalMode)
 		}
-	default:
-		return fmt.Errorf("unsupported deployment type: %q", deploymentType)
+		return state.NewStaticProvider(servermode.DegradedMode)
+	}
+
+	if a.config.forceStaticMode {
+		a.log.Info("forcing the use of Static Cluster Manager")
+		modeProvider = staticModeProvider()
+	} else {
+		switch deploymentType {
+		case deployment.MultiTenantType:
+			a.log.Info("using ETCD Based Dynamic Cluster Manager")
+			modeProvider = state.NewETCDDynamicProvider()
+		case deployment.DedicatedType:
+			a.log.Info("using Static Cluster Manager")
+			modeProvider = staticModeProvider()
+		default:
+			return fmt.Errorf("unsupported deployment type: %q", deploymentType)
+		}
 	}
 
 	adaptiveLimit := payload.SetupAdaptiveLimiter(ctx, g)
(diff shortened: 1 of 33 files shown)


testhelper/destination/transformer.go
```diff
@@ -21,13 +20,13 @@ func SetupTransformer(pool *dockertest.Pool, d cleaner) (*TransformerResource, e
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to pull image: %w", err)
-	}
+	// err := pool.Client.PullImage(docker.PullImageOptions{
+	// 	Repository: "rudderlabs/rudder-transformer",
+	// 	Tag:        "latest",
+	// }, docker.AuthConfiguration{})
+	// if err != nil {
+	// 	return nil, fmt.Errorf("failed to pull image: %w", err)
+	// }
```
atzoum, line 29: need to revert this change

jobsdb/jobsdb.go
```diff
@@ -1307,6 +1313,12 @@ func (jd *HandleT) createDSInTx(tx *Tx, newDS dataSetT) error {
 		return err
 	}
 
+	sqlStatement = fmt.Sprintf(`CREATE INDEX "idx_%[1]s_ws" ON %[1]q (workspace_id)`, newDS.JobTable)
+	_, err = tx.ExecContext(context.TODO(), sqlStatement)
+	if err != nil {
+		return err
+	}
```
atzoum, line 1320: migrator needs to add the same index when creating tables on its own

app/apphandlers/embeddedAppHandler.go
```diff
@@ -62,6 +63,7 @@ func (a *embeddedApp) loadConfiguration() {
 	config.RegisterIntConfigVariable(0, &a.config.gatewayDSLimit, true, 1, "Gateway.jobsDB.dsLimit", "JobsDB.dsLimit")
 	config.RegisterIntConfigVariable(0, &a.config.routerDSLimit, true, 1, "Router.jobsDB.dsLimit", "JobsDB.dsLimit")
 	config.RegisterIntConfigVariable(0, &a.config.batchRouterDSLimit, true, 1, "BatchRouter.jobsDB.dsLimit", "JobsDB.dsLimit")
+	config.RegisterBoolConfigVariable(false, &a.config.forceStaticMode, false, "forceStaticModeProvider")
```
atzoum, line 66: **_Note:_** 
For testing purposes, we need to be able to support both a static ModeProvider and a Multi-tenant DeploymentType both at the same time, so that we don't need to do the _"etcd dance"_ when testing multi-workspace scenarios.

`processor_isolation_test.go` is such an example

jobsdb/jobsdb.go
```diff
@@ -1911,6 +1923,53 @@ func (jd *HandleT) GetPileUpCounts(ctx context.Context) (map[string]map[string]i
...
+			(SELECT s.* FROM t, LATERAL(
+			  SELECT workspace_id FROM %[1]q f
+			  WHERE f.workspace_id > t.workspace_id
+			  ORDER BY workspace_id LIMIT 1) s)
+		  )
+		  SELECT * FROM t) a`, ds.JobTable))
+		} else {
+			queries = append(queries, fmt.Sprintf(`SELECT DISTINCT workspace_id from %[1]q j LEFT JOIN "v_last_%[2]s" js ON js.job_id = j.job_id
+			WHERE js.job_state is NULL or js.job_state not in ('aborted', 'succeeded', 'migrated')`, ds.JobTable, ds.JobStatusTable))
+		}
```
atzoum, line 1954: _**Note:**_ 
To assess what is more valuable, a fast query or a precise one, taking into consideration [this](https://www.notion.so/rudderstacks/Design-Proposal-Processor-Isolation-2db79f6c5b0f4a9fb14465b62aaaf54a#8bb2901bd4374bdfaaf30d2ca823b0e3), along with the fact that jobsdb's cache is not enabled for the `gw` rightmost dataset

jobsdb/jobsdb.go
atzoum: https://github.com/rudderlabs/rudder-migrator/pull/35

jobsdb/jobsdb.go
fracasula: What do you think about being redundant with the workspace_id and have it in the status table too and hit just that index without a join? I think I would've find it useful to have it there a few times.

fracasula: Also, what is the performance difference between the two? Is it negligible with the number of workspaces that we deal with?

utils/misc/exponential.go
```diff
@@ -0,0 +1,31 @@
...
+	expo.value = 0
+}
+
+// Next returns the next number, which is the previous one multiplied by 2, always abiding by the min and max provided.
+func (expo *ExponentialNumber[T]) Next(min, max T) T {
+	expo.value *= 2
+	if expo.value > max {
+		expo.value = max
+	}
+	if expo.value < min {
```
fracasula, line 27: With <=1 the function will always return `min` e.g.:

```go
var exp misc.ExponentialNumber[float64]
t.Log(exp.Next(0.9, 1)) // prints 0.9 since 0.9^2 = 0.81 which is less than min
```

I'm pointing it out only because it is a generic component and it can be used by anyone. A power of a <1 is still possible, it's just that it decreases.
Given the intended use for this, perhaps we should complain if that is attempted. Either way it shouldn't be a big deal.

utils/queue/priorityQueue.go
```diff
@@ -0,0 +1,84 @@
...
+	Value     T
+	Priority  int
+	timeStamp int64
+	index     int
+}
+
+// PriorityQueue ....
+type PriorityQueue[T any] []*Item[T]
+
+// Len: Size of the priority queue . Used to satisfy the heap interface...
```
fracasula, line 19: I got some linting stuff out of this file. Beside JetBrains complaining for `:` after `Len` saying:
> Comment should have the following format 'Len ...' (with an optional leading article) 

On a more serious note, some methods of `PriorityQueue` have method receivers, some do not. Is that intended?

utils/queue/priorityQueue.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Item stores the attributes which will be pushed to the priority queue..
+type Item[T any] struct {
+	Value     T
+	Priority  int
+	timeStamp int64
+	index     int
+}
+
+// PriorityQueue ....
```
fracasula, line 16: So meaningful :laughing:

utils/queue/priorityQueue.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Pop removes elements from the priority queue
+func (pq *PriorityQueue[T]) Pop() interface{} {
+	old := *pq
+	n := len(old)
+	if n == 0 {
+		return nil
+	}
+	item := old[n-1]
+	old[n-1] = nil
```
fracasula, line 55: Is this line necessary given that we'll reslice `pq`? :thinking:

utils/queue/priorityQueue.go
fracasula: Given that as you put it it's very close to the heap std implementation, I think it would be worth explaining in a comment how it is different from a `std` one and when to use this implementation.

utils/sync/limiter.go
```diff
@@ -0,0 +1,192 @@
...
+	l.mu.Unlock()
+
+	// no dynamic priority
+	if l.dynamicPeriod == 0 || priority == LimiterPriorityValueHigh {
+		<-w.Value
+		return
+	}
+
+	// dynamic priority (increment priority every dynamicPeriod)
+	ticker := time.NewTicker(l.dynamicPeriod)
```
fracasula, line 174:
```suggestion
	ticker := time.NewTicker(l.dynamicPeriod)
	defer ticket.Stop()
```

utils/sync/limiter_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+		statsTriggerCh := make(chan time.Time)
+		triggerFn := func() <-chan time.Time {
+			return statsTriggerCh
+		}
+
+		limiter := miscsync.NewLimiter(ctx, &wg, "test", 1, ms, miscsync.WithLimiterStatsTriggerFunc(triggerFn))
+		var counter int
+		statsTriggerCh <- time.Now()
+		statsTriggerCh <- time.Now()
```
fracasula, line 29: How come we need to trigger twice?

processor/worker.go
```diff
@@ -0,0 +1,211 @@
...
+// newWorker creates a new worker
+func newWorker(ctx context.Context, partition string, h workerHandle) *worker {
+	w := &worker{
+		handle:    h,
+		logger:    h.logger().Child(partition),
+		partition: partition,
+	}
+	w.lifecycle.ctx, w.lifecycle.cancel = context.WithCancel(ctx)
+	w.channel.ping = make(chan struct{}, 1)
+	w.channel.preproccess = make(chan subJob, w.handle.config().pipelineBufferedItems)
```
fracasula, line 25: Typo in `proccess`. It is repeated a few times.

jobsdb/jobsdb.go
atzoum: Overall performance difference is negligible when table count is low. It becomes more prominent when table count grows. With the number of workspaces we are having it should be safer to use the index-only scan, since we can tolerate false positives better than large table counts.

With respect to the status table, it is not enough on its own, since the status table doesn't hold entries for unprocessed events, which we need to be aware of.

utils/misc/exponential.go
atzoum: The algorithm is not raising the previous number to any power, it is merely multiplying the previous number by 2. Adding another test case using the scenario you mentioned for clarity
```
	t.Run("exponential float", func(t *testing.T) {
		var exp misc.ExponentialNumber[float64]
		require.EqualValues(t, 0.9, exp.Next(0.9, 5))
		require.EqualValues(t, 1.8, exp.Next(0.9, 5))
		require.EqualValues(t, 3.6, exp.Next(0.9, 5))
		require.EqualValues(t, 5, exp.Next(0.9, 5))
		require.EqualValues(t, 5, exp.Next(0.9, 5))
		exp.Reset()
		require.EqualValues(t, 0.9, exp.Next(0.9, 5))
	})
```

utils/queue/priorityQueue.go
atzoum: Yes, using pointer and non-pointer receivers per case is intended and not any different than [Go's own proposed implementation](https://pkg.go.dev/container/heap#example-package-PriorityQueue) of a priority queue.

utils/queue/priorityQueue.go
atzoum: There is no std implementation of a priority queue, at least none that I am aware of. There is only an [example](https://pkg.go.dev/container/heap#example-package-PriorityQueue) in the docs. The only standard is the `heap.Interface` that types must implement to be used for heap operations.

utils/sync/limiter_test.go
atzoum: To make sure that the first operation that was triggered has been completed

atzoum: To make sure that the first operation that was triggered has been completed

utils/queue/priorityQueue.go
atzoum: According to the [example](https://pkg.go.dev/container/heap#example-package-PriorityQueue)
```go
old[n-1] = nil  // avoid memory leak
item.index = -1 // for safety
```

fracasula (approved):
LGTM :+1:

processor/worker.go
```diff
@@ -0,0 +1,211 @@
...
+		defer w.logger.Debugf("ping loop stopped for worker: %s", w.partition)
+		for {
+			select {
+			case <-w.lifecycle.ctx.Done():
+				return
+			case <-w.channel.ping:
+			}
+
+			w.setIdleSince(time.Time{})
+			if w.pickJobs() {
```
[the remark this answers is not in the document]
atzoum: Yes but I found it risky to introduce (i.e. respect) context in this step, since all our processes, so far, have been designed using a panic-driven approach, meaning that they cannot tolerate interruptions well.

Sidddddarth (approved):

jobsdb/jobsdb.go
```diff
@@ -1307,6 +1313,12 @@ func (jd *HandleT) createDSInTx(tx *Tx, newDS dataSetT) error {
 		return err
 	}
 
+	sqlStatement = fmt.Sprintf(`CREATE INDEX "idx_%[1]s_ws" ON %[1]q (workspace_id)`, newDS.JobTable)
```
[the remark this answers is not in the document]
atzoum: Keeping both indexes until we stop supporting an isolation mode, e.g. if we stop supporting workspace isolation this index could be dropped.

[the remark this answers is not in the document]
atzoum: Just realized that for existing tables we are missing this index, so I am adding a migration script 👍

PR #2896, opened 2023-01-23
chore: add tags(sourceType, sdkVersion) to gateway stats

Sidddddarth:
# Description

add `sourceType`, `sdkVersion` tags to gateway stats.

## Notion Ticket

[new gateway stat-tags](https://www.notion.so/rudderstacks/new-gateway-stat-tags-87db19e3380449018a7da14f847190f3)

## Security

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -412,17 +412,13 @@ func (gateway *HandleT) getSourceTagFromWriteKey(writeKey string) string {
 }
 
 func (gateway *HandleT) NewSourceStat(writeKey, reqType string) *gwstats.SourceStat {
-	sourceTag := gateway.getSourceTagFromWriteKey(writeKey)
-	sourceID := gateway.getSourceIDForWriteKey(writeKey)
-	configSubscriberLock.RLock()
-	workspaceId := enabledWriteKeyWorkspaceMap[writeKey]
-	configSubscriberLock.RUnlock()
 	return &gwstats.SourceStat{
-		Source:      sourceTag,
-		SourceID:    sourceID,
+		Source:      gateway.getSourceTagFromWriteKey(writeKey),
+		SourceID:    gateway.getSourceIDForWriteKey(writeKey),
 		WriteKey:    writeKey,
 		ReqType:     reqType,
-		WorkspaceID: workspaceId,
+		WorkspaceID: gateway.getWorkspaceForWriteKey(writeKey),
+		SourceType:  gateway.getSourceDefForWriteKey(writeKey),
 	}
 }
 
@@ -514,9 +510,7 @@ func (gateway *HandleT) getJobFromRequest(req *webRequestT) (job *jobsdb.JobT, n
 	writeKey := req.writeKey
 	sourceID := gateway.getSourceIDForWriteKey(writeKey)
 	// Should be function of body
-	configSubscriberLock.RLock()
-	workspaceId := enabledWriteKeyWorkspaceMap[writeKey]
-	configSubscriberLock.RUnlock()
+	workspaceId := gateway.getWorkspaceForWriteKey(writeKey)
 	userIDHeader := req.userIDHeader
 
 	ipAddr := req.ipAddr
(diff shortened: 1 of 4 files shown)


gateway/gateway.go
```diff
@@ -540,7 +544,7 @@ func (gateway *HandleT) getJobFromRequest(req *webRequestT) (job *jobsdb.JobT, n
 			return
 		}
 	}
-	numEvents = len(gjson.GetBytes(body, "batch").Array())
+	jobData.numEvents = len(gjson.GetBytes(body, "batch").Array())
```
psrikanth88, line 547: Should we consider using `gjson.GetMany` ([link](https://github.com/tidwall/gjson#get-multiple-values-at-once)) instead of decoding the json multiple times here and below for different keys ?

Sidddddarth: I'm not entirely sure... the number of events in a batch could change and we'd have to dynamically create all the paths that we'd like to get from the batched payload.
Not that we couldn't do it, but it doesn't look as neat in my head.
I'm still thinking of a way to make this work, I'll post an update tomorrow.

gateway/gateway.go
```diff
@@ -674,18 +687,39 @@ func (gateway *HandleT) getJobFromRequest(req *webRequestT) (job *jobsdb.JobT, n
...
+	jobData.job = job
 	return
 }
 
+func (*HandleT) getSourceDefForWriteKey(writeKey string) string {
+	configSubscriberLock.RLock()
+	defer configSubscriberLock.RUnlock()
+
+	if _, ok := writeKeysSourceMap[writeKey]; ok {
+		return writeKeysSourceMap[writeKey].SourceDefinition.Category
```
atzoum, line 708: what does a `null` category refer to? Shouldn't we return a value instead of an empty string for that case?

gateway/gateway.go
```diff
@@ -674,18 +687,39 @@ func (gateway *HandleT) getJobFromRequest(req *webRequestT) (job *jobsdb.JobT, n
...
 		EventPayload: body,
-		EventCount:   numEvents,
+		EventCount:   jobData.numEvents,
 		WorkspaceId:  workspaceId,
 	}
+	jobData.job = job
 	return
 }
 
+func (*HandleT) getSourceDefForWriteKey(writeKey string) string {
```
atzoum, line 703: shouldn't the name of the method be `getSourceCategoryForWriteKey` ?

gateway/gateway.go
atzoum: performance-wise it should be no different with what we are doing now.

gateway/gateway.go
Sidddddarth: returning `eventStream` in case of null categories.

Sidddddarth: if `writeKey` not found in configMap, returns an empty string.

gateway/gateway.go
Sidddddarth: deferring this suggestion to a different PR in the future, includes some refactor and needs some more tests to be added.

atzoum (approved):

gateway/gateway.go
```diff
@@ -445,13 +441,13 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
...
 			writeKey := req.writeKey
 			sourceTag := gateway.getSourceTagFromWriteKey(writeKey)
-			if _, ok := sourceStats[sourceTag]; !ok {
-				sourceStats[sourceTag] = gateway.NewSourceStat(
-					writeKey,
-					req.reqType,
-				)
+			var sourceStat *gwstats.SourceStat
+			var ok bool
+			if sourceStat, ok = sourceStats[sourceTag]; !ok {
```
fracasula, line 446:
```suggestion
			sourceStat, ok := sourceStats[sourceTag]
			if !ok {
```

gateway/gateway.go
```diff
@@ -656,6 +660,15 @@ func (gateway *HandleT) getJobFromRequest(req *webRequestT) (job *jobsdb.JobT, n
 		"batch.0.context.sources.task_run_id",
 	).Str
 
+	jobData.version = gjson.GetBytes(
+		body,
+		"batch.0.context.library.name",
+	).Str + "/" +
+		gjson.GetBytes(
+			body,
+			"batch.0.context.library.version",
```
fracasula, line 669: Might be faster if we get `batch.0.context.library` only once and then get the name and version from it, something like:

```go
lib := gjson.GetBytes(body, "batch.0.context.library")
jobData.version = gjson.GetBytes(lib, "name")+gjson.GetBytes(lib, "version")
```

gateway/gateway.go
```diff
@@ -674,18 +687,42 @@ func (gateway *HandleT) getJobFromRequest(req *webRequestT) (job *jobsdb.JobT, n
...
+}
+
+func (*HandleT) getSourceCategoryForWriteKey(writeKey string) (category string) {
+	configSubscriberLock.RLock()
+	defer configSubscriberLock.RUnlock()
+
+	if _, ok := writeKeysSourceMap[writeKey]; ok {
+		category = writeKeysSourceMap[writeKey].SourceDefinition.Category
+		if category == "" {
+			category = "eventStream"
```
fracasula, line 710: [optional] Might be useful to have a constant for `"eventStream"`.

fracasula (approved):
Approved with a few suggestions. They should be fairly easy and quick to fix too.

PR #2898, opened 2023-01-23
chore: handling wrapped temporary errors for Kafka destinations

fracasula:
# Description

When `WriteErrors` from the Segment Kafka library are wrapped they don't result as temporary anymore. This PR aims to fix this issue.

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/Handling-wrapped-temporary-errors-for-Kafka-destinations-0959d007cbc64dcc82baaf02ec187007) >

## Security

### DIFF
--- a/services/streammanager/kafka/client/producer.go
+++ b/services/streammanager/kafka/client/producer.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"os"
 	"syscall"
 	"time"
 
@@ -152,9 +153,8 @@ func isErrTemporary(err error) bool {
 	if errors.As(err, &tempError) {
 		return tempError.Temporary()
 	}
-	var timeoutError interface{ Timeout() bool }
-	if errors.As(err, &timeoutError) {
-		return timeoutError.Timeout()
+	if os.IsTimeout(err) {
+		return true
 	}
 	return false
 }
--- a/services/streammanager/kafka/client/producer.go
+++ b/services/streammanager/kafka/client/producer.go
@@ -160,23 +160,18 @@ func isErrTemporary(err error) bool {
 }
 
 func IsProducerErrTemporary(err error) bool {
-	f := func(err error) bool {
-		if err == nil {
-			return false
-		}
-		if we, ok := err.(kafka.WriteErrors); ok {
-			for _, err := range we {
-				// if at least one was temporary then we treat the whole batch as such
-				if isErrTemporary(err) {
-					return true
-				}
+	if err == nil {
+		return false
+	}
+	var we kafka.WriteErrors
+	if errors.As(err, &we) {
+		for _, err := range we {
+			// if at least one was temporary then we treat the whole batch as such
+			if isErrTemporary(err) {
+				return true
 			}
-			return false
 		}
-		return isErrTemporary(err)
+		return false
 	}
-	if ue := errors.Unwrap(err); ue != nil {
-		return f(err) || IsProducerErrTemporary(ue)
-	}
-	return f(err)
+	return isErrTemporary(err)
 }
--- a/services/streammanager/kafka/client/producer.go
+++ b/services/streammanager/kafka/client/producer.go
@@ -139,9 +139,6 @@ func headers(msg Message) (headers []kafka.Header) {
 }
 
 func isErrTemporary(err error) bool {
-	if errors.Is(err, context.DeadlineExceeded) {
-		return true
-	}
 	isTransientNetworkError := errors.Is(err, io.ErrUnexpectedEOF) ||
 		errors.Is(err, syscall.ECONNREFUSED) ||
 		errors.Is(err, syscall.ECONNRESET) ||
--- a/services/streammanager/kafka/client/producer.go
+++ b/services/streammanager/kafka/client/producer.go
@@ -173,9 +173,8 @@ func IsProducerErrTemporary(err error) bool {
 		}
 		return isErrTemporary(err)
 	}
-	var isUnwrappedErrorTemporary bool
 	if ue := errors.Unwrap(err); ue != nil {
-		isUnwrappedErrorTemporary = IsProducerErrTemporary(ue)
+		return f(err) || IsProducerErrTemporary(ue)
 	}
-	return f(err) || isUnwrappedErrorTemporary
+	return f(err)
 }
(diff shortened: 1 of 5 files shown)


fracasula:
Please don't mind DeepSource. It is reporting all sorts of unrelated issues.

chandumlg (approved):

services/streammanager/kafka/client/producer.go
```diff
@@ -140,28 +138,45 @@ func headers(msg Message) (headers []kafka.Header) {
...
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
 	isTransientNetworkError := errors.Is(err, io.ErrUnexpectedEOF) ||
 		errors.Is(err, syscall.ECONNREFUSED) ||
 		errors.Is(err, syscall.ECONNRESET) ||
 		errors.Is(err, syscall.EPIPE)
 	if isTransientNetworkError {
 		return true
 	}
```
ItsSudip, line 150:
```suggestion
	isTransientNetworkError := errors.Is(err, io.ErrUnexpectedEOF) ||
		errors.Is(err, syscall.ECONNREFUSED) ||
		errors.Is(err, syscall.ECONNRESET) ||
		errors.Is(err, syscall.EPIPE) || errors.Is(err, context.DeadlineExceeded)
	if isTransientNetworkError {
		return true
	}
```

services/streammanager/kafka/client/producer.go
```diff
@@ -140,28 +138,45 @@ func headers(msg Message) (headers []kafka.Header) {
...
+	var tempError interface{ Temporary() bool }
 	if errors.As(err, &tempError) {
 		return tempError.Temporary()
 	}
+	var timeoutError interface{ Timeout() bool }
+	if errors.As(err, &timeoutError) {
+		return timeoutError.Timeout()
+	}
```
atzoum, line 158: we can use `os.IsTimeout(err)` which already does more or less the same thing

services/streammanager/kafka/client/producer.go
```diff
@@ -140,28 +138,45 @@ func headers(msg Message) (headers []kafka.Header) {
...
+					return true
+				}
 			}
+			return false
 		}
-		return false
+		return isErrTemporary(err)
+	}
+	if ue := errors.Unwrap(err); ue != nil {
+		return f(err) || IsProducerErrTemporary(ue)
```
atzoum, line 179: couldn't we spare this if we simply converted 
```
if we, ok := err.(kafka.WriteErrors); ok {
```
to
```
var we kafka.WriteErrors
if errors.As(err, &we) {
```
?
If we make sure all checks inside `f` properly unwrap errors, we wouldn't need to unwrap them manually here and perform the recursive operation afterwards

services/streammanager/kafka/client/producer.go
```diff
@@ -140,28 +138,45 @@ func headers(msg Message) (headers []kafka.Header) {
...
 		errors.Is(err, syscall.ECONNREFUSED) ||
 		errors.Is(err, syscall.ECONNRESET) ||
 		errors.Is(err, syscall.EPIPE)
 	if isTransientNetworkError {
 		return true
 	}
+	var tempError interface{ Temporary() bool }
 	if errors.As(err, &tempError) {
 		return tempError.Temporary()
 	}
```
atzoum, line 154: BTW, I believe that the language has moved away from this `Temporary` error pattern, at least according to [that](https://go-review.googlesource.com/c/go/+/188398)

services/streammanager/kafka/client/producer.go
fracasula: I did that way in the beginning but moved it in a separate `if` because a `context.DeadlineExceeded` could be due to things unrelated to network as well, so I don't want it to change the `isTransientNetworkError`.

services/streammanager/kafka/client/producer.go
fracasula: Yes, it should work. I copied the whole block from [here](https://github.com/segmentio/kafka-go/blob/v0.4.38/error.go#L593) and didn't realize I could use the std library for that instead :+1:

services/streammanager/kafka/client/producer.go
fracasula: Errors from the underlying segment library still follow this pattern so we might have to stick with it for now.

services/streammanager/kafka/client/producer.go
fracasula: True but it is slower given that it uses reflection on top of `Unwrap`. It's still looping (without recursion) so it works. In any case for the sake of simplicity I will change it. 

Btw, another trick is that I was doing:
```go
if ue := errors.Unwrap(err); ue != nil {
	return f(err) || IsProducerErrTemporary(ue)
}
return f(err)
```

So recursion was not used unless the parent error was not temporary already. Might even be nicer this way to avoid calling `Unwrap` unnecessarily:

```go
if f(err) {
    return true
} else if ue := errors.Unwrap(err); ue != nil {
	return IsProducerErrTemporary(ue)
}
return false
```

```
BenchmarkTempErrors
BenchmarkTempErrors/recursion
BenchmarkTempErrors/recursion-24         	 4419303	       266.6 ns/op
BenchmarkTempErrors/errors.As
BenchmarkTempErrors/errors.As-24         	 3531273	       350.4 ns/op
PASS
```

In any case, thanks for the comment! :+1:

services/streammanager/kafka/client/producer.go
```diff
@@ -140,21 +139,32 @@ func headers(msg Message) (headers []kafka.Header) {
 }
 
 func isErrTemporary(err error) bool {
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
```
atzoum, line 144: isn't this redundant since we are checking for `Temporary` or `Timeout` errors later on?

services/streammanager/kafka/client/producer.go
```diff
@@ -140,21 +139,32 @@ func headers(msg Message) (headers []kafka.Header) {
...
 	return false
 }
 
 func IsProducerErrTemporary(err error) bool {
-	if we, ok := err.(kafka.WriteErrors); ok {
+	if err == nil {
+		return false
+	}
```
atzoum, line 165: why do we need the nil check?

atzoum (approved):

PR #2902, opened 2023-01-24
chore: use a partition lock when updating badger live events cache

atzoum:
# Description

Reading and updating the same key in a single transaction concurrently in `badgerdb` results in `ErrConflict`, due to badger's serializable snapshot isolation (SSI) guarantees. Thus we need to use a partitioned (per key) mutex to synchronise the update order.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=707219e884cb4ffdac1c8310e8f99f2e&pm=s)

## Security


utils/sync/plocker.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Lock locks the lock. If the lock is locked, it waits until the lock is unlocked.
+func (p *PartitionLocker) Lock(id string) {
+	p.l.Lock()
+	defer p.l.Unlock()
+	for p.locked(id) {
+		p.c.Wait()
+	}
+	p.s[id] = struct{}{}
+}
```
lvrach, line 35: Assume the following use case:
```		
		go func() {
				locker.Lock("id1")
                 time.Sleep(1 * time.Millisecond)
				locker.Unlock("id1")
         }()
         go func() {
				locker.Lock("id2")
                 time.Sleep(1 * time.Minute)
				locker.Unlock("id2")
         }()
         go func() {
				locker.Lock("id2")
                 time.Sleep(1 * time.Millisecond)
				locker.Unlock("id2")
         }()
 ```
 Depending on the execution order, id1 can wait unnecessarily long for id2 to get locked.
 Since both will try to hold: `p.l.Lock()`
 
 Depending on your use case, this might or might not be a problem.

Comment (atzoum) [reply utils/sync/plocker.go:35]: `p.l` is not locked while `p.c.Wait` is blocking. Thus I can't see where the problem can be

File: utils/sync/plocker_bench_test.go [hunk 2]
```diff
@@ -0,0 +1,103 @@
...
+			for j := 0; j < numGoroutines; j++ {
+				key := strconv.Itoa(j % numKeys)
+				wg.Add(1)
+				go func() {
+					defer wg.Done()
+					l.Lock(key)
+					countersMu.Lock()
+					counters[key]++
+					countersMu.Unlock()
+					time.Sleep(time.Millisecond)
```
fracasula, line 43: What is the reason for this sleep here?

atzoum: just trying to introduce a bit of contention :)

utils/sync/plocker_bench_test.go
```diff
@@ -0,0 +1,103 @@
...
+// Running tool: go test -benchmem -run=^$ -bench ^BenchmarkPartitionLocker$ github.com/rudderlabs/rudder-server/utils/sync -v -count=1
+
+// goos: darwin
+// goarch: arm64
+// pkg: github.com/rudderlabs/rudder-server/utils/sync
+// BenchmarkPartitionLocker
+// BenchmarkPartitionLocker/mutex
+// BenchmarkPartitionLocker/mutex-10         	      20	  57351981 ns/op	  162918 B/op	    2121 allocs/op
+// BenchmarkPartitionLocker/cond
+// BenchmarkPartitionLocker/cond-10          	      20	  56779356 ns/op	  149106 B/op	    2030 allocs/op
```
fracasula, line 22: This is what I'm getting on my desktop :thinking: 

```
BenchmarkPartitionLocker
BenchmarkPartitionLocker/mutex
BenchmarkPartitionLocker/mutex-24         	      18	  58423921 ns/op
BenchmarkPartitionLocker/cond
BenchmarkPartitionLocker/cond-24          	      14	  75101069 ns/op
PASS
ok  	github.com/rudderlabs/rudder-server/utils/sync	6.337s
```

fracasula: The big delta happens when most of my threads (i.e. 21/24) are busy. If I free most of them up there is less difference but still it gives me cond as slower:

```
BenchmarkPartitionLocker
BenchmarkPartitionLocker/mutex
BenchmarkPartitionLocker/mutex-24         	      21	  56724928 ns/op
BenchmarkPartitionLocker/cond
BenchmarkPartitionLocker/cond-24          	      18	  63376837 ns/op
PASS
ok  	github.com/rudderlabs/rudder-server/utils/sync	5.434s
```

With a busy CPU I consistently get a big diff between the 2.

atzoum: Is the difference still comparable? E.g. is it 10-20% slower or is it orders of magnitude slower? In any case, `75ms` for having 1000 goroutines locking & unlocking 20 mutexes and sleeping for `1ms` in between is not an issue, right?

fracasula: With a busy CPU it is consistently ~30% slower :disappointed: 

```
BenchmarkPartitionLocker/mutex-24         	      20	  59212050 ns/op
BenchmarkPartitionLocker/cond-24          	      16	  77213625 ns/op
```

Screenshot from 2023-01-27 14-50-58

If it's slower and clearly also more complex, why use it?
I like the implementation, it's a smart use of `cond` but I'm not sure I get it.

atzoum: It is more cpu intensive but more memory efficient. It is slower in some architectures & scenarios but faster in others. Complexity is of course relevant, since it mostly boils down to a person's understanding of `sync.Cond` 😄 

I would argue that any of these solutions would do just fine coping with our workloads since they are comparable, but let's vote!

❤️ `sync.Cond`
🚀 map of `sync.Mutex`

fracasula: My take is that most of the time we want our CPU busy doing things and that should be the normal state in a MultiTenant architecture so we can assume that most of the time it's going to be 30% slower. The advantage in terms of memory isn't a lot to begin with i.e. ~%5. 

So to me it boils down to deciding between:
* have a 5% more memory efficient algorithm that is 30% less CPU efficient and that is more complex to understand/maintain/debug
* have a common pattern that consumes 5% more memory but it is 30% faster

> Complexity is of course relevant, since it mostly boils down to a person's understanding of sync.Cond

It is not just that! :smile: Tests are also added and benchmarks and it is more code overall in the codebase.
In your tests you're also creating wrappers around mutexes. Worst case, we might happen to have to debug some code live that is using your version of a mutex and by then the guy on call would have forgot how it works :smile:

atzoum: It depends on the scenario and the architecture I guess since with `numKeys: 100000` and `numGoroutines: 500000` I am getting the following results:

```
BenchmarkPartitionLocker/mutex
BenchmarkPartitionLocker/mutex-10         	       1	1330982500 ns/op	336118704 B/op	 2583238 allocs/op
BenchmarkPartitionLocker/cond
BenchmarkPartitionLocker/cond-10          	       1	1295121292 ns/op	127760256 B/op	 1918228 allocs/op
```
I don't see any of the two solutions as true bottlenecks, but also understand the value of simplicity! Nevertheless, this was a fun exercise!

lvrach (approved):

fracasula (approved):

cisse21 (approved):

PR #2908, opened 2023-01-26
feat: introduce upload id to associate staging files to uploads

lvrach:
# Description

In this PR, we are introducing an upload id foreign key, which will replace the start/end range approach for associating uploads to staging files.

The foreign key will be used for pending staging files at this phase. The upload_id will always be populated, but no migration is needed for existing files. The change is handling backwards compatibility by using the `use_upload_id` flag.

A visualization of the changes https://drive.google.com/file/d/1looAyHBBipDbrvQ3LmGo8KIMMPd0MBA9/view?usp=sharing

## Motivation
The reason behind this change is to enable compaction. Using ranges will not work correctly with compaction.

Screenshot 2023-01-27 at 12 23 06 PM

## Picking pending staging files

Two ways to get pending staging files:

### Legacy

Screenshot 2023-01-27 at 11 27 04 AM

### upload_id is null (optimised)

* We pick the last upload for the source/destination 
* if `use_upload_id` flag is set, continue, else Legacy flow
* get staging files for the source/destination, upload_id = NULL, after latest upload's 

Screenshot 2023-01-27 at 12 27 52 PM

## New column `upload_id`

A few notes regarding the new column creation:

```
ALTER TABLE wh_staging_files ADD COLUMN IF NOT EXISTS upload_id INTEGER REFERENCES wh_uploads (id) ON DELETE SET NULL(upload_id);
```
* the type is INTEGER to match the type of SERIAL in the `wh_uploads` table
* There is a Foreign Key constraint, upload_id must be either NULL or a primary key (id) on `wh_uploads`. You can not insert integers that are not present in the other table. Inserts with NULL upload_id id do not need to verify anything, setting the upload_id has some additional lookup cost.
* When an upload gets deleted, all FK will become NULL. This incurs costs during upload deletion. 

## Notion Ticket

https://www.notion.so/rudderstacks/Upload-StageFiles-020cbe9436c74e6a9504f616685cac40

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is used.


warehouse/internal/repo/staging.go
```diff
@@ -329,24 +337,52 @@ func (repo *StagingFiles) Pending(ctx context.Context, sourceID, destinationID s
...
+	).Scan(
+		&uploadID,
+		&lastStartStagingFileID,
+		&lastEndStagingFileID,
+		&useUploadID,
+	)
+
 	if err == sql.ErrNoRows {
-		lastStagingFileID = 0
+		lastEndStagingFileID = 0
```
achettyiitr, line 348: Should we reset `lastStartStagingFileID` as well?

achettyiitr:
We can remove `//go:build !warehouse_integration` from the test files as it is not being used.

warehouse/internal/repo/upload.go
```diff
@@ -341,6 +383,18 @@ func (uploads *Uploads) UploadJobsStats(ctx context.Context, destType string, op
 	return stats, nil
 }
 
+func (uploads *Uploads) DeleteWaiting(ctx context.Context, uploadID int64) error {
+	_, err := uploads.db.ExecContext(ctx,
```
achettyiitr, line 387: Initially, I thought we might need to reset the `upload_id` as well from the `staging files`. But since we have `ON DELETE` clause.

achettyiitr (approved):

warehouse/internal/repo/upload.go
lvrach: I should have mentioned, I have updated the PR description.

We would also need an index on upload_id to ensure the deletion won't be costly.

warehouse/internal/repo/staging.go
```diff
@@ -317,10 +317,18 @@ func (repo *StagingFiles) GetAfterID(ctx context.Context, sourceID, destinationI
...
+		lastEndStagingFileID   int64
+		useUploadID            sql.NullBool
+	)
 	err := repo.db.QueryRowContext(ctx, `
 		SELECT
-			end_staging_file_id
+			id,
+			start_staging_file_id,
+			end_staging_file_id,
+			metadata->>'use_upload_id' AS use_upload_id
```
abhimanyubabbar, line 331: Should we be fetching cross db inside repo meant for staging files ? Or should we move this call a layer above or have access to upload repo here ? Wdyt ?

warehouse/internal/repo/staging.go
abhimanyubabbar: Wouldn't this value by default be `0` ? Given there are no rows so no data would be scanned from DB. Wdyt ?

abhimanyubabbar: Should we even keep this check of `err == sql.ErrNoRows` ?

abhimanyubabbar (approved):
some minor comments, looks good otherwise

warehouse/internal/repo/staging.go
lvrach: A good point. I believe `stagingFiles.Pending` is a good abstraction, the question I present myself is: 
Should the service layer know that in order to fetch pending events, we have to get the max staging file from another file? 

For me, this is an implementation detail specific to our persistent layer that needs to be abstracted.

Moreover, once the migration phase gets completed, we can fetch pending files by just looking at a single table.

warehouse/internal/repo/staging.go
lvrach: It should have been removed indeed.

sql/migrations/warehouse/000020_add_staging_files_upload_id.up.sql
```diff
@@ -0,0 +1,8 @@
+
+--
+-- wh_staging_files
+--
+
+ALTER TABLE wh_staging_files ADD COLUMN IF NOT EXISTS upload_id INTEGER REFERENCES wh_uploads (id) ON DELETE SET NULL(upload_id);
+
+CREATE INDEX wh_staging_files_upload_id_index ON wh_staging_files(upload_id);
```
lvrach, line 8: `CREATE INDEX IF NOT EXIST`

sql/migrations/warehouse/000020_add_staging_files_upload_id.up.sql
```diff
@@ -0,0 +1,8 @@
+
+--
+-- wh_staging_files
+--
+
+ALTER TABLE wh_staging_files ADD COLUMN IF NOT EXISTS upload_id INTEGER REFERENCES wh_uploads (id) ON DELETE SET NULL(upload_id);
```
lvrach, line 6: maybe needs a default

PR #2909, opened 2023-01-27
chore: warmup gcra algorithms in throttling test

atzoum:
# Description

To achieve consistent results from GCRA rate limiting algorithms, `burst` should be set equal to `rate` and a **warmup** cycle should  be performed before the actual test, in order to fill up the `burst`. After that point the algorithm should "leak" at the expected rate.

## Security

### DIFF
--- a/internal/throttling/throttling_test.go
+++ b/internal/throttling/throttling_test.go
@@ -41,13 +41,13 @@ func TestThrottling(t *testing.T) {
 			},
 			{
 				name:        "gcra",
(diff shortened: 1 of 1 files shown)


lvrach:
Question: Is this intended to fix the flakiness?

atzoum:
> Question: Is this intended to fix the flakiness?

yes

lvrach (approved):

fracasula (approved):
It makes sense, thanks :+1:

PR #2910, opened 2023-01-27
feat: suppression backup service

saurav-malani:
# Description

Suppression Backup Service to speed up the backup process. It maintain two `badgerdb`. Both of them regularly syncs with `data-regulation-service` for `latest` & `full` suppressions and creates an export of the backup, so that a gateway during start can quickly get the list of users to be suppressed, which leads to decrease in start time of gateway.

## Notion Ticket

https://www.notion.so/rudderstacks/Speed-up-resync-of-suppress-regulations-b29f4df9d10047c7ae3ae7d80781b9bb


suppression-backup-service/api/goldenFile.proto
```diff
@@ -0,0 +1,11 @@
+message position {
```
atzoum, line 1: please move this to a `testdata` folder

suppression-backup-service/exporter/exporter.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-time.After(e.pollIntervalFn()):
+			err := syncer.Sync(ctx)
+			if err != nil {
+				e.log.Errorf("error syncing suppression list: %v", err)
```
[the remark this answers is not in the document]
saurav-malani: the reason behind logging the error and not returning was, I thought that because of some issue the sync failed. So, maybe the next time it won't. But, I think what you are saying make sense. Since, assuming that in next retry after a while it will succeed isn't appropriate. And, we should assuming that the Sync func returned error only after some retry. And, hence we should not ignore it. As ignoring might raise a scenario where sync always throw error and hence nothing is ever stored and exported.

suppression-backup-service/exporter/exporter.go
```diff
@@ -0,0 +1,223 @@
...
+	err = os.Rename(tempFile.Name(), file.Path)
+	if err != nil {
+		return fmt.Errorf("error moving temp file to final destination: %w", err)
+	}
+
+	return nil
+}
+
+func newBadgerDBInstance(baseDir string, pkgLogger logger.Logger) (suppression.Repository, error) {
+	_ = os.RemoveAll(baseDir)
```
atzoum, line 54: 1. In case of a full export, the badgerDB folders should be preserved between restarts, they shouldn't be deleted.
2. If `RemoveAll` fails, we shouldn't ignore this error.
3. Maybe this operation shouldn't be performed in this function at all, but just before it, wherever necessary.

suppression-backup-service/exporter/exporter.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	return repository, nil
+}
+
+type Exporter struct {
+	id             identity.Identifier
+	file           model.File
+	log            logger.Logger
+	pollIntervalFn func() time.Duration
+}
```
atzoum, line 73: **minor:** this struct appears to be redundant, i.e. the struct's fields could be moved to the respective functions:
- `FullExporterLoop`
- `LatestExporterLoop`

This way we could spare the `Exporter` struct, `ExporterOpt` type, `WithPollIntervalFn` & `NewExporter` functions.

suppression-backup-service/exporter/exporter.go
```diff
@@ -0,0 +1,223 @@
...
+			if err := syncer.Sync(ctx); err != nil {
+				return err
+			}
+			stats.Default.NewStat("suppression_backup_service_full_sync_time", stats.TimerType).Since(syncStart)
+			exportStart := time.Now()
+			if err = Export(repo, e.file); err != nil {
+				return err
+			}
+			stats.Default.NewStat("suppression_backup_service_full_export_time", stats.TimerType).Since(exportStart)
+			<-time.After(e.pollIntervalFn())
```
atzoum, line 138: we should respect the context here

suppression-backup-service/exporter/exporter.go
```diff
@@ -0,0 +1,223 @@
...
+					return err
+				}
+				stats.Default.NewStat("suppression_backup_service_latest_export_time", stats.TimerType).Since(exportStart)
+
+				return nil
+			}
+			if err := latestExport(); err != nil {
+				return err
+			}
+			<-time.After(e.pollIntervalFn())
```
atzoum, line 206: we should respect the context here as well

suppression-backup-service/cmd/main.go
```diff
@@ -0,0 +1,139 @@
...
+	pkgLogger.Info("http server setup done")
+
+	g.Go(func() error {
+		<-gCtx.Done()
+		pkgLogger.Info("context.Done triggered... shutting down httpserver")
+		return httpServer.Shutdown(context.Background())
+	})
+
+	if err := g.Wait(); err != nil {
+		if err != http.ErrServerClosed || ctx.Err() != nil {
```
atzoum, line 100: I would argue that `if ctx.Err() != nil` we shouldn't return an `error` under any condition, because this is the definition of a normal shutdown.

suppression-backup-service/cmd/main.go
```diff
@@ -0,0 +1,139 @@
+package main
```
atzoum, line 1: We are missing docker-related adaptations. With respect to that, I believe that we should start producing separate docker images for this and other similar processes, i.e. have a separate `Dockerfile` & push separate docker images. E.g. regulation-worker should have its own image too (in the future as a separate task).

Piggybacking on rudder-server's docker image won't get us too far.

suppression-backup-service/exporter/exporter.go
```diff
@@ -0,0 +1,223 @@
...
+					return err
+				}
+				baseDir := path.Join(tmpDir, "latestsuppression")
+				repo, err := newBadgerDBInstance(baseDir, e.log)
+				if err != nil {
+					return err
+				}
+				defer func() {
+					_ = repo.Stop()
+					_ = os.RemoveAll(baseDir)
```
atzoum, line 164: This might be redundant. We definitely need to make sure that folder doesn't exist before starting the badgerDB instead.

suppression-backup-service/exporter/exporter.go
```diff
@@ -0,0 +1,223 @@
...
+			if err := latestExport(); err != nil {
+				return err
+			}
+			<-time.After(e.pollIntervalFn())
+		}
+	}
+}
+
+func latestToken() (string, error) {
+	marshalledToken, err := json.Marshal(Token{SyncStartTime: time.Now().Add(config.GetDuration("SuppressionExporter.latestExportDuration", -30*24, time.Hour)), SyncSeqId: -1})
```
atzoum, line 212:
```suggestion
	marshalledToken, err := json.Marshal(Token{SyncStartTime: time.Now().Add(-config.GetDuration("SuppressionExporter.latestExportDuration", 30*24, time.Hour)), SyncSeqId: -1})
```

suppression-backup-service/exporter/exporter.go
saurav-malani: removed `os.RemoveAll` from here to before calling `newBadgerDBInstance`

suppression-backup-service/exporter/exporter.go
saurav-malani: good point... modified.

suppression-backup-service/api/api.go
```diff
@@ -0,0 +1,61 @@
...
+func (api *API) latestExport(w http.ResponseWriter, r *http.Request) {
+	api.latestBackup.Mu.RLock()
+	defer api.latestBackup.Mu.RUnlock()
+	http.ServeFile(w, r, api.latestBackup.Path)
+	setHeader(w, http.StatusOK)
+}
+
+func setHeader(w http.ResponseWriter, code int) {
+	w.Header().Set("Content-Type", "application/octet-stream")
+	w.Header().Set("Content-Encoding", "gzip")
```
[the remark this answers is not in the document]
saurav-malani: yup

suppression-backup-service/cmd/main.go
saurav-malani: make sense. I have added a ticket for the same on notion board.

suppression-backup-service/api/api.go
```diff
@@ -0,0 +1,61 @@
...
+	http.ServeFile(w, r, api.fullBackup.Path)
+	setHeader(w, http.StatusOK)
+}
+
+func (api *API) latestExport(w http.ResponseWriter, r *http.Request) {
+	api.latestBackup.Mu.RLock()
+	defer api.latestBackup.Mu.RUnlock()
+	http.ServeFile(w, r, api.latestBackup.Path)
+	setHeader(w, http.StatusOK)
+}
```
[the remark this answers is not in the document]
saurav-malani: good point.

suppression-backup-service/cmd/main.go
```diff
@@ -0,0 +1,139 @@
...
+	return id, nil
+}
+
+// exportPath creates a tmp dir and returns the path to it
+func exportPath() (baseDir string, err error) {
+	tmpDir, err := misc.CreateTMPDIR()
+	if err != nil {
+		return "", fmt.Errorf("could not create tmp dir: %w", err)
+	}
+	baseDir, err = os.MkdirTemp(tmpDir, "export")
```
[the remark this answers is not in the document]
saurav-malani: yup make sense, modified.

suppression-backup-service/cmd/main.go
saurav-malani: modified it to `err == http.ErrServerClosed && ctx.Err() == nil `, which means that since ctx.Err==nil means `Done()` is not closed yet. And, still server is closed. So, it is unexpected.

suppression-backup-service/api/api.go
```diff
@@ -0,0 +1,56 @@
...
+		defer file.Mu.RUnlock()
+		http.ServeFile(w, r, file.Path)
+		setHeader(w, http.StatusOK)
+	}
+}
+
+func setHeader(w http.ResponseWriter, code int) {
+	w.Header().Set("Content-Type", "application/octet-stream")
+	w.WriteHeader(code)
+}
```
atzoum, line 56: We may delete this function since:
1. setting the content-type appears to be redundant, since this is already being done in `http.ServeFile`
2. setting the status code appears to be redundant too, since `http.ServeFile` does it too

```suggestion
```

suppression-backup-service/api/api_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Log("endpoint tested:", tt.endpoint)
+			port, err := freeport.GetFreePort()
+			require.NoError(t, err)
+			baseUrl := fmt.Sprintf("http://localhost:%d", port)
+			req, err := http.NewRequest(tt.method, baseUrl+tt.endpoint, http.NoBody)
+			require.NoError(t, err)
+			req.Header.Set("Content-Type", "application/json")
```
atzoum, line 81:
```suggestion
```

suppression-backup-service/api/api_test.go
```diff
@@ -0,0 +1,120 @@
...
+	logger.Reset()
+	misc.Init()
+	latestExportPostfix := "latest-export"
+	fullExportPostfix := "full-export"
+	tests := []struct {
+		name                 string
+		endpoint             string
+		method               string
+		responseCode         int
+		expectedResponseBody string
```
atzoum, line 47: should we add an expected content type here too?

suppression-backup-service/exporter/exporter.go
saurav-malani: `WithPollIntervalFn` helps in making loop testable.

suppression-backup-service/exporter/exporter.go
```diff
@@ -0,0 +1,235 @@
...
+			if err = misc.SleepCtx(ctx, e.pollIntervalFn()); err != nil {
+				return err
+			}
+		}
+	}
+}
+
+func (e *Exporter) LatestExporterLoop(ctx context.Context) error {
+	var interval time.Duration
+	config.RegisterDurationConfigVariable(600, &interval, true, time.Second, "SuppressionExporter.latestExportInterval")
```
atzoum, line 152: We could use a larger value here, e.g. capture a latest backup every 1 hour

suppression-backup-service/exporter/exporter.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	return nil
+}
+
+func newBadgerDBInstance(baseDir string, pkgLogger logger.Logger) (suppression.Repository, error) {
+	err := os.MkdirAll(baseDir, os.ModePerm)
+	if err != nil {
+		return nil, fmt.Errorf("could not create tmp dir: %w", err)
+	}
```
atzoum, line 63: BadgerDB creates directories whenever necessary with the appropriate permissions. Do we need this?
```suggestion
	err := os.MkdirAll(baseDir, os.ModePerm)
	if err != nil {
		return nil, fmt.Errorf("could not create tmp dir: %w", err)
	}
```

suppression-backup-service/cmd/main.go
atzoum: Shouldn't we add the docker image for this new command as part of this task?

suppression-backup-service/exporter/exporter.go
atzoum: by passing the same as a parameter to the function(s)
> the struct's fields could be moved to the respective functions

suppression-backup-service/cmd/main.go
saurav-malani: okay, I thought you are suggesting to create a `Dockerfile` for both `regulation-worker` & `suppression-backup-server` both in separate ticket. Anyways, I will add the `Dockerfile` for this service in this PR only.

suppression-backup-service/exporter/exporter.go
saurav-malani: okay, I didn't know that badgerdb does that. Removed it.

suppression-backup-service/api/api_test.go
saurav-malani: added `http.DetectContentType(body)` to compare with.

suppression-backup-service/cmd/main.go
atzoum: We'll need a github action for building the image too

saurav-malani: yup

.github/workflows/builds.yml
```diff
@@ -94,3 +94,46 @@ jobs:
...
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          context: .
+          file: ./suppression-backup-service/Dockerfile
+          images: |
+            name=rudderlabs/develop-rudder-server-suppression-backup-service
+            name=rudderlabs/rudder-server-suppression-backup-service,enable=${{ github.ref == format('refs/heads/{0}', 'master') || github.event_name == 'release' }}
```
lvrach, line 112: I believe is an overkill to prefix the image name with `rudder-server` even if it is under the same repo.
```suggestion
            name=rudderlabs/develop-suppression-backup-service
            name=rudderlabs/suppression-backup-service,enable=${{ github.ref == format('refs/heads/{0}', 'master') || github.event_name == 'release' }}
```

saurav-malani: renamed

.github/workflows/builds.yml
```diff
@@ -94,3 +94,46 @@ jobs:
             COMMIT_HASH=${{ github.sha }}
             REVISION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.revision'] }}
             ENTERPRISE_TOKEN=${{ secrets.ENTERPRISE_TOKEN }}
+  docker-suppression-backup-service:
```
atzoum, line 97: It doesn't seem like the correct Dockerfile is being used [here](https://github.com/rudderlabs/rudder-server/actions/runs/4416443796/jobs/7740738131)

.github/workflows/builds.yml
```diff
@@ -94,3 +94,46 @@ jobs:
...
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          context: ../../suppression-backup-service
```
atzoum, line 108: this looks weird, why is it needed?

suppression-backup-service/.dockerignore
```diff
@@ -0,0 +1,7 @@
+**/.git
+.env
+.vscode
+.dockerignore
+cmd/benchmark/Dockerfile
+Dockerfile
+regulation-worker
```
atzoum, line 7: why do we need this file?

atzoum (approved):
LGTM

enterprise/suppress-user/syncer.go
```diff
@@ -109,40 +109,44 @@ func (s *Syncer) SyncLoop(ctx context.Context) {
...
-			s.log.Errorf("Failed to add %d suppressions to repository: %w", len(suppressions), err)
-			continue
-		}
-		if len(suppressions) != 0 {
 			goto again
 		}
+		s.log.Errorf("Failed to get token from repository: %w", err)
+		return err
 	}
+	s.log.Info("Fetching Regulations")
```
fracasula, line 136: We're printing info entries in production and we're also printing this one twice. How is this going to be used? It doesn't seem very useful.

saurav-malani: yup, I actually added it while debugging an issue. And, forgot to remove it. Thanks for pointing it out.

suppression-backup-service/README.md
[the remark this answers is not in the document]
saurav-malani: added.

fracasula (approved):

PR #2914, opened 2023-01-30
chore: enhance data sent to reporting

cisse21:
# Description

- Sending sample payload for GW to reporting table
- Sending the successful response body from router for 2xx status codes
- Getting rid of `source_batch_id` and `source_task_id` across the codebase

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=d98e215c6a1e4324b370d86748ccf62a&pm=s)

## Security


processor/processor.go
```diff
@@ -1205,6 +1191,10 @@ func (proc *Handle) processJobsForDest(subJobs subJob, parsedEventList [][]types
 			// Iterate through all the events in the batch
 			for eventIndex, singularEvent := range singularEvents {
 				messageId := misc.GetStringifiedData(singularEvent["messageId"])
+				sampleEvent, err := jsonfast.Marshal(singularEvent)
```
fracasula, line 1194: Out of curiosity, how big can this `singularEvent` be?

cisse21: It would be the event which we receive in the gateway so max 4MB size

fracasula (approved):

processor/processor.go
fracasula: With your change they end up in a map though for batching right? Have you checked whether this could create an issue in terms of sudden memory spikes? I'm asking because it looks like we had `{}` hard-coded before so what if now we get a bunch of 4MB events all in a row?

sql/migrations/reports/000003_alter_reports_remove_source_batch_id.down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE reports
+ADD COLUMN IF NOT EXISTS source_batch_id text DEFAULT '',
+ADD COLUMN IF NOT EXISTS source_task_id text DEFAULT '';
```
BonapartePC, line 3: Minor: What is the point of this migration?

BonapartePC (approved):

sql/migrations/reports/000003_alter_reports_remove_source_batch_id.down.sql
cisse21: we have the down migrations everywhere though never used in production. Just following the standard thing in the code

atzoum (changes requested):

sql/migrations/reports/000003_alter_reports_remove_source_batch_id.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE reports
+DROP COLUMN source_batch_id,
+DROP COLUMN source_task_id;
```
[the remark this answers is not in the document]
cisse21: Yes... It's only the master which does reporting and not the slaces

PR #2915, opened 2023-01-30
chore: refactored gateway marshalling

Sidddddarth:
# Description

moved around some code and removed a few `gjson.GetBytes` from `gateway.getJobDataFromRequest`.

## Notion Ticket

< Replace with Notion Link >

## Security


gateway/gateway.go
```diff
@@ -563,51 +561,85 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
+		builtUserID, userIDFromReq                            string
+		sourcesJobRunID, sourcesTaskRunID                     string
+		sdkName, sdkVersion                                   string
+		notIdentifiable, nonRudderEvent, containsAudienceList bool
+	)
 
-		eventTypeFromReq := strings.TrimSpace(vjson.Get("type").String())
+	lo.ForEach(eventsBatch, func(vjson gjson.Result, _ int) {
+		anonIDFromReq := strings.TrimSpace(vjson.Get("anonymousId").String())
+		userIDFromReq = strings.TrimSpace(vjson.Get("userId").String())
```
fracasula, line 574: We're checking these two for each events in the batch but then we're considering only the last value of `userIDFromReq` when checking later for suppressions (i.e. `if gateway.isUserSuppressed(workspaceId, userIDFromReq, sourceID) {`)?

Is this correct?

gateway/gateway.go
```diff
@@ -693,6 +701,66 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
+func isEmptyString(value interface{}) bool {
+	if value == nil {
+		return true
+	}
+	str, ok := value.(string)
+	if !ok {
+		return true
+	}
+	return strings.TrimSpace(str) == ""
+}
```
fracasula, line 762: Same logic as yours, less lines :+1: 

```suggestion
func isEmptyString(value interface{}) bool {
	str, ok := value.(string)
	return !ok || strings.TrimSpace(str) == ""
}
```

gateway/gateway.go
```diff
@@ -693,6 +701,66 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
+		workspaceID,
+		userID,
+		sourceID,
+	)
+}
+
+// checks for the presence of messageId in the event
+//
+// sets to a new uuid if not present
+func checkMessageID(event map[string]interface{}) {
```
fracasula, line 746: Might make sense to rename this function. It's not just checking but setting a key on the given map.

```suggestion
func setRandomMessageIDWhenEmpty(event map[string]interface{}) {
```

gateway/gateway.go
```diff
@@ -693,6 +701,66 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
+	messageID := misc.MapLookup(event, "messageId")
```
fracasula, line 747: I'm not sure this `MapLookup` makes sense. It's used in many places but there are no tests for it. Do people really understand what it is supposed to do when multiple keys are passed to it and that recursion only happens if multiple keys are provided?

```go
func TestMapLookup(t *testing.T) {
	m := map[string]interface{}{
		"foo": "bar",
		"baz": "qux",
	}
	t.Log(MapLookup(m, "foo", "baz")) // prints <nil>

	m = map[string]interface{}{
		"foo": map[string]interface{}{
			"baz": "qux",
		},
	}
	t.Log(MapLookup(m, "foo", "baz")) // prints qux

	m = map[string]interface{}{
		"hello": map[string]interface{}{
			"foo": "bar",
		},
	}
	t.Log(MapLookup(m, "foo")) // prints <nil>
}
```

Anyhow, if that is how the function is supposed to work then using it gives you no benefits since it looks for the `messageId` key in the first level of the map only. You might as well just do:

```go
func checkMessageID(event map[string]interface{}) {
	messageID, _ := event["messageId"].(string)
	if strings.TrimSpace(messageID) == "" {
		event["messageId"] = uuid.New().String()
	}
}
```

And then you can remove `isEmptyString`. Let me know if I'm not making sense.

Sidddddarth: True👍
makes sense

gateway/gateway.go
Sidddddarth: the earlier logic was something similar, differing only in that the `userID` from the first event of the batch was being considered when looking up suppression.
I'm making some more changes around it.
Should we consider looking up supressions for all the users in a batch..?

gateway/gateway.go
Sidddddarth: I was worried about `messageID, _ := event["messageId"].(string)` part and didn't go this way initially.
This clarified things up for me.
Thanks!

gateway/gateway.go
Sidddddarth: brought back looking if userID from first event of the batch is suppressed behaviour.
> Should we consider looking up suppressions for all the users in a batch..?

this question still stands though

fracasula: Apparently it [might be possible](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1675240518176459?thread_ts=1675239804.575179&cid=C01HTT66UMB) to get multiple users in the same batch.

It would be correct to filter all the filterable ones and then work with whatever batch is left. But let's address this in another PR :+1:

utils/misc/misc_test.go
```diff
@@ -652,3 +652,25 @@ func FileExists(path string) (bool, error) {
 
 	return !fileInfo.IsDir(), nil
 }
+
+func TestMapLookup(t *testing.T) {
```
fracasula, line 656: Would you mind updating the docs on the function itself explaining this?

```go
// MapLookup returns the value of the key in the map, or nil if the key is not present.
// If multiple keys are provided then it looks for nested maps recursively.
func MapLookup(mapToLookup map[string]interface{}, keys ...string) interface{} {
```

gateway/gateway.go
```diff
@@ -693,6 +704,55 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
+func isNonIdentifiable(anonIDFromReq, userIDFromReq string) bool {
+	if anonIDFromReq == "" && userIDFromReq == "" {
+		return !allowReqsWithoutUserIDAndAnonymousID
+	}
+	return false
+}
+
+func buildUserID(userIDHeader, anonIDFromReq, userIDFromReq string) string {
+	if anonIDFromReq != "" {
+		return userIDHeader + DELIMITER + anonIDFromReq + DELIMITER + userIDFromReq
```
fracasula, line 730: Just double checking in case it's a copy/paste issue. You meant the last one to be `userIDFromReq` and not `anonIDFromReq`, right?

gateway/gateway.go
```diff
@@ -563,51 +561,88 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
+		if !ok {
+			nonRudderEvent = true
+			return
 		}
 
-		eventTypeFromReq := strings.TrimSpace(vjson.Get("type").String())
+		anonIDFromReq, _ := toSet["anonymousId"].(string)
+		userIDFromReq, _ = toSet["userId"].(string)
+		if userID == "" {
+			userID, _ = toSet["userId"].(string)
```
fracasula, line 582:
```suggestion
			userID = userIDFromReq
```

fracasula (approved):

gateway/gateway.go
Sidddddarth: yep👍

atzoum (approved):

PR #2916, opened 2023-01-31
fix: add workspace id in wh  async jobs table

am6010:
# Description

We need to add the workspace id in the WH async jobs table for supporting multitenant deployment. 
Sources will provide the workspace when registering a new job through the API.

## Notion Ticket

[ticket](https://www.notion.so/rudderstacks/Support-wh-multi-tenant-for-wh-async-jobs-api-4c8172eb09bc4b6c9eec2efcc93d4a53)

## Security

### DIFF
--- a/warehouse/jobs/runner.go
+++ b/warehouse/jobs/runner.go
@@ -45,7 +45,19 @@ func (a *AsyncJobWhT) getTableNamesBy(sourceID, destinationID, jobRunID, taskRun
 	a.logger.Infof("[WH-Jobs]: Extracting table names for the job run id %s", jobRunID)
 	var tableNames []string
 	var err error
-	query := fmt.Sprintf(`SELECT id from %s where metadata->>'%s'=$1 and metadata->>'%s'=$2 and metadata in (SELECT metadata FROM wh_uploads where source_id=$3 and destination_id=$4)`, warehouseutils.WarehouseUploadsTable, "source_job_run_id", "source_task_run_id")
+	query := `SELECT id 
+		FROM 
+		` + warehouseutils.WarehouseUploadsTable + `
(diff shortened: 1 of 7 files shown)


achettyiitr (approved):

warehouse/jobs/runner.go
```diff
@@ -45,7 +45,19 @@ func (a *AsyncJobWhT) getTableNamesBy(sourceID, destinationID, jobRunID, taskRun
...
+		WHERE metadata->>'source_job_run_id'=$1 
+			AND metadata->>'source_task_run_id'=$2 
+			AND metadata in (
+				SELECT 
+					metadata 
+				FROM 
+				` + warehouseutils.WarehouseUploadsTable + ` 
+				WHERE source_id=$3 
+				AND destination_id=$4
+			)`
```
lvrach, line 60: I know this is not introduced here, but why are we doing:
```
			AND metadata in (
				SELECT 
					metadata 
				FROM 
				WHERE source_id=$3 
				AND destination_id=$4
			)
```
?
Would it be enough doing :

```suggestion
			AND source_id=$3 
			AND destination_id=$4
			`
```

maybe I am missing something

lvrach (commented):
Can we add some test cases or extend existing tests to include workspace_id ?

warehouse/jobs/runner.go
am6010: @lvrach I think we have decided to change this approach. There is a [ticket](https://www.notion.so/rudderstacks/Transformer-for-table-name-handling-967f49f789d64c4b92904b1f98329854) to change it.
Checking the just with the source id will give us more tables that we need. Every job can have multiple tasks that write to different tables.

am6010:
I will raise a pr with some extra tests for workspaceId

PR #2917, opened 2023-01-31
tmp

saurav-malani:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/enterprise/suppress-user/internal/badgerdb/badgerdb_test.go
+++ b/enterprise/suppress-user/internal/badgerdb/badgerdb_test.go
@@ -2,15 +2,10 @@ package badgerdb_test
 
 import (
 	"bytes"
-	"context"
-	"io"
 	"path"
 	"strings"
-	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


enterprise/suppress-user/internal/badgerdb/badgerdb_test.go
```diff
@@ -28,125 +23,160 @@ func (f readerFunc) Read(p []byte) (n int, err error) {
...
 
-		require.Equal(t, repo.Restore(nil), badger.ErrDBClosed)
+		con2 := func() bool {
+			suppressed, err := repo2.Suppressed("workspace1", "user1", "src-1")
+			require.NoError(t, err)
+			return suppressed
+		}
+		require.Eventually(t, con2, 5*time.Second, 100*time.Millisecond, "abc")
 	})
+
```
saurav-malani, line 181: this test seems to be failing at line 179, which is not expected since we are restoring, what we backed up before.

PR #2920, opened 2023-02-01
chore: remove reports down migrations

cisse21:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security


processor/transformer/transformer.go
```diff
@@ -51,7 +51,6 @@ type MetadataT struct {
 	SourceTpConfig      map[string]map[string]interface{} `json:"sourceTpConfig"`
 	MergedTpConfig      map[string]interface{}            `json:"mergedTpConfig"`
 	DestinationID       string                            `json:"destinationId"`
-	JobRunID            string                            `json:"jobRunId"`
```
atzoum, line 54: any idea why this was there? @chandumlg ?

BonapartePC (approved):

processor/transformer/transformer.go
chandumlg: Not sure. Looks like a left over. 

I don't see any value set to that variable. No point keeping it.

chandumlg (approved):

PR #2928, opened 2023-02-01
fix: limit entire transformationStatus struct when caching them

Sidddddarth:
# Description

update cache with only the `messageID`s of events that are put into the cache.
Done to avoid an issue scene recently where successful events were being displayed as `dropped` in live events.
[context](https://rudderlabs.slack.com/archives/C018SGKA998/p1674657597947759)

## Notion Ticket

[notion ticket](https://www.notion.so/rudderstacks/live-events-cache-messageIDs-fix-ebdcb3068a354861a68ded5186d82033)

## Security

### DIFF
--- a/services/debugger/destination/eventDeliveryStatusUploader.go
+++ b/services/debugger/destination/eventDeliveryStatusUploader.go
@@ -63,7 +63,7 @@ func NewHandle(backendConfig backendconfig.BackendConfig, opts ...Opt) (Destinat
 		log:              logger.NewLogger().Child("debugger").Child("destination"),
 	}
 	var err error
-	url := fmt.Sprintf("%s/dataplane/v2/eventUploads", h.configBackendURL)
+	url := fmt.Sprintf("%s/dataplane/v2/eventDeliveryStatus", h.configBackendURL)
 	eventUploader := NewEventDeliveryStatusUploader(h.log)
 	h.uploader = debugger.New[*DeliveryStatusT](url, eventUploader)
 	h.uploader.Start()
--- a/services/debugger/transformation/transformationStatusUploader.go
+++ b/services/debugger/transformation/transformationStatusUploader.go
@@ -224,6 +224,41 @@ func (h *Handle) backendConfigSubscriber(backendConfig backendconfig.BackendConf
 	close(h.done)
 }
 
+// limit the number of stored events
+func (ts *TransformationStatusT) Limit(
+	limit int,
+	transformation backendconfig.TransformationT,
+) *TransformationStatusT {
+	ts.Destination.Transformations = []backendconfig.TransformationT{transformation}
+	ts.UserTransformedEvents = lo.Slice(ts.UserTransformedEvents, 0, limit)
+	ts.FailedEvents = lo.Slice(ts.FailedEvents, 0, limit)
+	messageIDs := lo.SliceToMap(
+		append(
+			lo.Map(
+				ts.UserTransformedEvents,
+				func(event transformer.TransformerEventT, _ int) string {
+					return event.Metadata.MessageID
+				},
+			),
+			lo.Map(
+				ts.FailedEvents,
+				func(event transformer.TransformerResponseT, _ int) string {
+					return event.Metadata.MessageID
+				},
+			)...,
+		),
+		func(messageID string) (string, struct{}) {
+			return messageID, struct{}{}
+		},
+	)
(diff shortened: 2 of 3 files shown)


services/debugger/transformation/transformationStatusUploader.go
```diff
@@ -243,10 +243,30 @@ func (h *Handle) UploadTransformationStatus(tStatus *TransformationStatusT) bool
 			h.processRecordTransformationStatus(tStatus, transformation.ID)
 		} else {
 			tStatusUpdated := *tStatus
-			lo.Slice(tStatusUpdated.UserTransformedEvents, 0, h.limitEventsInMemory+1)
 			tStatusUpdated.Destination.Transformations = []backendconfig.TransformationT{transformation}
 			tStatusUpdated.UserTransformedEvents = lo.Slice(tStatusUpdated.UserTransformedEvents, 0, h.limitEventsInMemory+1)
 			tStatusUpdated.FailedEvents = lo.Slice(tStatusUpdated.FailedEvents, 0, h.limitEventsInMemory+1)
+			// while caching, we may only store the messageIDs of the events in the cache
+			// done to avoid displaying events as dropped
```
atzoum, line 250: Is this comment necessary? We just need to keep the struct consistent after trimming down the number of events it holds

services/debugger/transformation/transformationStatusUploader.go
```diff
@@ -243,10 +243,30 @@ func (h *Handle) UploadTransformationStatus(tStatus *TransformationStatusT) bool
...
+			tStatusUpdated.UniqueMessageIds = lo.SliceToMap(
```
atzoum, line 251: after updating `UniqueMessageIds` we should also take care of `EventsByMessageID` and remove any no longer existent keys

services/debugger/transformation/transformationStatusUploader.go
Sidddddarth: 👍

services/debugger/destination/eventDeliveryStatusUploader.go
```diff
@@ -63,7 +63,7 @@ func NewHandle(backendConfig backendconfig.BackendConfig, opts ...Opt) (Destinat
 		log:              logger.NewLogger().Child("debugger").Child("destination"),
 	}
 	var err error
-	url := fmt.Sprintf("%s/dataplane/v2/eventUploads", h.configBackendURL)
+	url := fmt.Sprintf("%s/dataplane/v2/eventDeliveryStatus", h.configBackendURL)
```
atzoum, line 66: this was wrong?!

Sidddddarth: Yep

atzoum: so previously we didn't upload event delivery statuses? Or both urls work?

cisse21: This went in my Badger PR.... so it is not on any customer but is present in master

cisse21: Both URL's don't work so this is a bug which was caught by QA

cisse21 (approved):

services/debugger/destination/eventDeliveryStatusUploader.go
Sidddddarth: It is available.
As part of a previous PR, there was some mixup while copy pasting.
This just corrects it.

Sidddddarth: also yes.
`/dataplane/v2/eventDeliveryStatus` is the correct endpoint here

services/debugger/transformation/transformationStatusUploader.go
```diff
@@ -224,6 +224,41 @@ func (h *Handle) backendConfigSubscriber(backendConfig backendconfig.BackendConf
...
+				ts.UserTransformedEvents,
+				func(event transformer.TransformerEventT, _ int) string {
+					return event.Metadata.MessageID
+				},
+			),
+			lo.Map(
+				ts.FailedEvents,
+				func(event transformer.TransformerResponseT, _ int) string {
+					return event.Metadata.MessageID
+				},
```
atzoum, line 247: It would be great if we could cover this with tests

atzoum (approved):

PR #2931, opened 2023-02-02
fix(warehouse): warehouse successful upload exists

achettyiitr:
# Description

- Use gauge instead of the counter
- Instead of reporting 0 or 1, we are going to report always 1 and use a flag ok=true/false to capture the problematic uploads.
- We are going to log every time a problem is detected

## Notion Ticket

https://www.notion.so/rudderstacks/Make-the-alert-for-stuck-wh-pipelines-warehouse-successful-upload-exist-as-P0-d69f679ee0d04907ae0ed76b7a482ac7

## Security

### DIFF
--- a/warehouse/tracker.go
+++ b/warehouse/tracker.go
@@ -156,7 +156,7 @@ func (wh *HandleT) Track(ctx context.Context, warehouse *warehouseutils.Warehous
 	}
 
 	if !exists {
-		wh.Logger.Warn("pending staging files not picked",
+		wh.Logger.Errorw("pending staging files not picked",
 			warehouseutils.SourceID, source.ID,
 			warehouseutils.DestinationID, destination.ID,
 			warehouseutils.WorkspaceID, warehouse.WorkspaceID,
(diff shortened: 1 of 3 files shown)


warehouse/tracker.go
```diff
@@ -156,20 +155,26 @@ func (wh *HandleT) Track(ctx context.Context, warehouse *warehouseutils.Warehous
 		return fmt.Errorf("fetching last upload status for source: %s and destination: %s: %w", source.ID, destination.ID, err)
 	}
 
-	if exists {
-		uploaded = 1
+	if !exists {
+		wh.Logger.Warn("upload is not picked for ",
+			warehouseutils.SourceID, source.ID,
+			warehouseutils.DestinationID, destination.ID,
+			warehouseutils.DestinationID, warehouse.WorkspaceID,
```
lvrach, line 162:
```suggestion
			warehouseutils.WorkspaceID, warehouse.WorkspaceID,
```

warehouse/tracker.go
```diff
@@ -156,20 +155,26 @@ func (wh *HandleT) Track(ctx context.Context, warehouse *warehouseutils.Warehous
...
+		)
+		wh.Logger.Info("upload is not picked for source: ", source.ID, " and destination: ", destination.ID)
```
lvrach, line 164:
```suggestion
```
Do we need this one as well?

lvrach (approved):

PR #2940, opened 2023-02-05
chore(warehouse): added warehouse handling for s3 with glue and other improvements

achettyiitr:
# Description

- Added S3 Glue validations.
- Use the namespace defined by the customer in the destination config. If not present use the default test namespace.
- Don't use random tables every time while doing validations. 
- Using simple components like `objectStorage`, `connections`, `createSchema`, `createAlterTable`, `fetchSchema`, `loadTable` to satisfy `Validate() error` instead of coupling it with validation steps.

# Improvements

- Avoid coupling the GRPC module with the validations module.
- Test coverage for the validations module.

# Design changes
Validations

## Notion Ticket

- https://www.notion.so/rudderstacks/S3-datalake-add-glue-checks-to-validation-flow-along-with-recent-changes-for-schemas-520855f99b3c43429dc220b8ecc42642
- https://www.notion.so/rudderstacks/Configuration-testing-changes-2a455393fe854c109a87b40465943b56

## Security


warehouse/integrations/bigquery/bigquery.go
```diff
@@ -115,23 +115,25 @@ func (bq *HandleT) DeleteTable(tableName string) (err error) {
 	return
 }
 
-func (bq *HandleT) CreateTable(tableName string, columnMap map[string]string) (err error) {
+func (bq *HandleT) CreateTable(tableName string, columnMap map[string]string) error {
```
achettyiitr, line 118: While writing the Unit test, we found that suppression is not happening properly.

warehouse/integrations/deltalake/deltalake.go
```diff
@@ -897,8 +899,22 @@ func (dl *Deltalake) AddColumns(tableName string, columnsInfo []warehouseutils.C
...
+		Config:       dl.Client.CredConfig,
+		Identifier:   dl.Client.CredIdentifier,
+		SqlStatement: query,
+	})
+	if err != nil {
+		return fmt.Errorf("add columns: %w", err)
+	}
+
+	// Handle error in case of single column
+	if len(columnsInfo) == 1 {
```
achettyiitr, line 912: We would need to suppress the errors in case of a single column alter since databricks does support `ADD COLUMN IF NOT EXISTS`

warehouse/validations/steps.go
```diff
@@ -3,61 +3,72 @@ package validations
...
+		wh := createDummyWarehouse(dest)
+		if canUseGlue := schemarepository.UseGlue(&wh); !canUseGlue {
+			break
+		}
 
-	steps = append(steps,
-		&validationStep{
-			ID:        2,
-			Name:      verifyingConnections,
-			Validator: ct.verifyingConnections,
```
achettyiitr, line 39: There was too much coupling with steps and validation functions. steps API will return only the steps. Once we get the steps, we call `NewValidator` to get the validating function.

warehouse/validations/validations_suite_test.go
```diff
@@ -1,13 +0,0 @@
-package validations_test
```
achettyiitr, line 1: Since the package does not contain Ginkgo-based tests. Removing it.

warehouse/validations/validations.go
```diff
@@ -2,83 +2,78 @@ package validations
...
-	return
+	return res, nil
 }
 
-// validationFunctions returns validating functions for validation
-func (ct *CTHandleT) validationFunctions() map[string]*validationFunc {
+func validationFunctions() map[string]*validationFunc {
 	return map[string]*validationFunc{
 		"validate": {
-			Path: "/validate",
```
achettyiitr, line 76: The `Path` is not being used. Removed it.

warehouse/warehousegrpc.go
```diff
@@ -94,12 +97,49 @@ func (*warehouseGRPC) TriggerWHUpload(_ context.Context, request *proto.WHUpload
 	return res, err
 }
 
-func (grpc *warehouseGRPC) Validate(_ context.Context, req *proto.WHValidationRequest) (*proto.WHValidationResponse, error) {
-	handleT := validations.CTHandleT{
```
achettyiitr, line 98: The domain model for the validations package should depend on the destination only, not on the internal representation of the destination like whether tunneling is enabled or not. 
Moving the logic of adding those information to the caller itself

warehouse/validations/helper.go
```diff
@@ -1,60 +0,0 @@
...
-	"encoding/json"
-	"fmt"
-
-	backendconfig "github.com/rudderlabs/rudder-server/config/backend-config"
-	"github.com/rudderlabs/rudder-server/services/filemanager"
-	"github.com/rudderlabs/rudder-server/utils/misc"
-	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
-)
-
-func warehouse(req *DestinationValidationRequest) warehouseutils.Warehouse {
```
achettyiitr, line 13: All these have been moved to validate.go

warehouse/integrations/redshift/redshift.go
```diff
@@ -255,15 +255,32 @@ func (rs *Redshift) AddColumns(tableName string, columnsInfo []warehouseutils.Co
 			columnInfo.Name,
 			getRSDataType(columnInfo.Type),
 		)
-		rs.Logger.Infof("AZ: Adding column for destinationID: %s, tableName: %s with query: %v", rs.Warehouse.Destination.ID, tableName, query)
+		rs.Logger.Infof("RS: Adding column for destinationID: %s, tableName: %s with query: %v", rs.Warehouse.Destination.ID, tableName, query)
 
 		if _, err := rs.DB.Exec(query); err != nil {
+			if CheckAndIgnoreColumnAlreadyExistError(err) {
+				pkgLogger.Infof("RS: Column already exists for destinationID: %s, tableName: %s with query: %v", rs.Warehouse.Destination.ID, tableName, query)
```
achettyiitr, line 262: We would need to suppress the errors in case of a single column alter since Redshift does support ADD COLUMN IF NOT EXISTS

warehouse/validations/helper.go
```diff
@@ -1,60 +0,0 @@
-package validations
```
achettyiitr, line 1: Moved to validate.go

warehouse/validations/types.go
```diff
@@ -1,149 +0,0 @@
-package validations
```
achettyiitr, line 1: Moved to validate.go

warehouse/validations/validate.go
```diff
@@ -17,378 +20,537 @@ import (
...
+type fetchSchema struct {
+	manager     manager.WarehouseOperations
+	destination *backendconfig.DestinationT
+}
+
+type loadTable struct {
+	manager     manager.WarehouseOperations
+	destination *backendconfig.DestinationT
+	table       string
+}
```
achettyiitr, line 54: These satisfies the validator interface to validate the particular step.

warehouse/validations/validate.go
```diff
@@ -17,378 +20,537 @@ import (
...
 
-	// Create table
-	err = ct.manager.CreateTable(stagingTableName, TestTableSchemaMap)
-	if err != nil {
-		return
+	operations.SetConnectionTimeout(warehouseutils.TestConnectionTimeout)
+
+	if err = operations.Setup(warehouse, &mockUploader{
+		dest: dest,
+	}); err != nil {
```
lvrach, line 518: I am not sure why we are doing things. Since `mockUploader` is usually for helping testing doesn't imply a lot about its functionality.

achettyiitr: True. Should I rename this to something like `dummyUploader`?

achettyiitr: We wanted 2 functionalities from Uploader interface `GetLoadFileType` and `UseRudderStorage`

achettyiitr: I am thinking of using noop uploader to avoid adding default implementations.
https://github.com/rudderlabs/rudder-server/pull/2993

achettyiitr: Once the above PR merges, we would not need this.

achettyiitr: Because of the complexity of the changes associated with `Uploader` interface. I am thinking to move to a separate PR.
I renamed it to `dummyUploader` for the current use case.

lvrach (approved):

warehouse/validations/validate.go
```diff
@@ -17,378 +22,549 @@ import (
...
+		logfield.WorkspaceID, dest.WorkspaceID,
+		logfield.DestinationValidationsStep, stepToValidate,
 	)
 
-	resp := DestinationValidationResponse{}
 	// check if req has specified a step in query params
-	if step != "" {
-		stepI, err := strconv.Atoi(step)
+	if stepToValidate != "" {
+		stepI, err := strconv.Atoi(stepToValidate)
```
cisse21, line 130: Why don't we accept this as an integer? We can get away with all this type cast right

achettyiitr: This comes as [query params](https://github.com/rudderlabs/cp-router/blob/8d929099e5cae0f707fdc0ec87d623675898003c/api/warehouse.go#L298), from config backend -> cp-router -> warehouse. Either we need to do this conversion at the cp-router or at the warehouse. Since the warehouse acts as a proxy, it's better we handle it at the warehouse.

cisse21 (approved):

PR #2950, opened 2023-02-07
fix(warehouse): warehouse validations fixed to lookup secrets using sshKeyId

abhimanyubabbar:
# Description
The warehouse validations are moved to use the sshKeyId which is injected from upstream when making a call to run validations. The benefit is to not depend on the warehouse destinationId the first time we have created a destination.

## Notion Ticket

https://www.notion.so/rudderstacks/Augment-the-warehouse-validation-flow-to-play-nicely-with-the-sql-tunnel-based-validations-eb71efa64fcc466bb28da64b2d64a0f0

## Security

### DIFF
--- a/warehouse/client/controlplane/client.go
+++ b/warehouse/client/controlplane/client.go
@@ -25,6 +25,7 @@ type BasicAuth struct {
 
 type InternalControlPlane interface {
 	GetDestinationSSHKeys(ctx context.Context, id string) (*PublicPrivateKeyPair, error)
+	GetSSHKeys(ctx context.Context, id string) (*PublicPrivateKeyPair, error)
 }
 
 type internalClientWithCache struct {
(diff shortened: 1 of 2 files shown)


warehouse/client/controlplane/client.go
```diff
@@ -46,6 +47,42 @@ type internalClient struct {
 	httpClient *http.Client
 }
 
+func (api *internalClient) GetSSHKeys(ctx context.Context, id string) (*PublicPrivateKeyPair, error) {
```
achettyiitr, line 50: Can we add test coverage for this?

warehouse/validations/validate.go
```diff
@@ -116,18 +116,34 @@ func (ct *CTHandleT) verifyingObjectStorage() (err error) {
...
+	if !warehouseutils.ReadAsBool("useSSH", config) {
+		return nil
+	}
+
+	sshKeyId, ok := ct.warehouse.Destination.Config["sshKeyId"]
+	if !ok {
+		return fmt.Errorf("missing sshKeyId in validation payload")
+	}
+
+	keys, err := ct.CPClient.GetSSHKeys(context.TODO(), sshKeyId.(string))
```
achettyiitr, line 129:
```suggestion
    	sshKey, ok := sshKeyId.(string)
	if !ok {
		return fmt.Errorf("invalid sshKeyId")
	}

	keys, err := ct.CPClient.GetSSHKeys(context.TODO(), sshKey)
```

Should we have an extra check here whether sskKeyID is of the proper format?

abhimanyubabbar: Having a different datatype would be a very big deviation from the standard. In case that happens, it's better to let us know asap, so panic might help us. Wdyt ?

achettyiitr (approved):

warehouse/client/controlplane/client.go
abhimanyubabbar: Let me see if I can add coverage for this in realistic time frame. Would it be possible to handle it as a separate PR in case the timelines exceed the cutoff time for the release ?

achettyiitr: Yes. I already approved the PR.

warehouse/client/controlplane/client.go
```diff
@@ -25,6 +25,7 @@ type BasicAuth struct {
 
 type InternalControlPlane interface {
```
lvrach, line 26: [minor] should this interface be defined only on the validations package? Are there other packages that specifically depend on `GetDestinationSSHKeys` and `GetSSHKeys`?

warehouse/validations/validate.go
lvrach: I prefer @achettyiitr suggestion, to be honest. 

Given the schema of the payload is not in our control and we don't validate against the contract when we receive the request in our system. 

I don't see it as an unlikely scenario a small change in the control plane, encoding `sshKeyId` as int and then having warehouse panic when someone tries to setup an ssh destination.

warehouse/client/controlplane/client.go
```diff
@@ -89,6 +126,20 @@ func NewInternalClientWithCache(baseURI string, auth BasicAuth) InternalControlP
 	}
 }
 
+func (cc *internalClientWithCache) GetSSHKeys(ctx context.Context, id string) (*PublicPrivateKeyPair, error) {
+	if val, ok := cc.cache.Load(id); ok {
```
lvrach, line 130: Is it ok for `GetSSHKeys` ids and `GetDestinationSSHKeys` to be stored in the same cache?

In other words, will id_1 returned by `SSHKeys` always represent the same content as id_1 from `DestinationSSHKeys`

abhimanyubabbar: The id's are different so there won't be any clash. I can move to a different cache but the chance of collision isn't very likely and we will be removing the cache lookup using destination_id in next release. 

Let me know what you think ?

lvrach: We can also prefix to avoid collisions, if suid is used it should be ok. If we plan to remove this code in the future I am ok keeping it like this.

lvrach (approved):

warehouse/client/controlplane/client.go
abhimanyubabbar: We will be removing the destination ssh keys functionality in future, so for now it seems okay. I'll track the corresponding changes in another ticket for now.

PR #2956, opened 2023-02-08
feat(warehouse): error tagging

achettyiitr:
# Description

- **Added tags for error handling for upload jobs.**

- **Added grouping for upload errors.**
```GO
const (
	PermissionError           JobErrorType = "permission_error"
	AlterColumnError          JobErrorType = "alter_column_error"
	ResourceNotFoundError     JobErrorType = "resource_not_found_error"
	ColumnCountError          JobErrorType = "column_count_error"
	ColumnSizeError           JobErrorType = "column_size_error"
	InsufficientResourceError JobErrorType = "insufficient_resource_error"
	ConcurrentQueriesError    JobErrorType = "concurrent_queries_error"
	UnknownError              JobErrorType = "unknown_error"
)
```

## Notion Ticket

https://www.notion.so/rudderstacks/Add-proper-tags-with-error-handling-for-upload-aborted-5a6da646b6e74c95819c91e53d0fabf8

## Security


warehouse/internal/model/upload.go
```diff
@@ -82,6 +94,13 @@ type UploadJob struct {
 	LoadFileGenStartTime time.Time
 }
 
+type JobErrorType string
+
+type JobError struct {
+	Type   JobErrorType
+	Format *regexp.Regexp
```
lvrach, line 101: I am thinking of the following approach to allow the format to be described using both regex, and other ways as well.

```go
type Matcher interface {
     MatchString(string) bool
}

type Contains string

func (c Contains) MatchString(substr string) bool {
      return strings.Contains(string(c), substr)
}
```

```suggestion
	Format Matcher
```

lvrach (approved):

warehouse/internal/model/upload.go
abhimanyubabbar: +1 for the matcher interface. We would abstract the underlying pattern matching and then test in whatever fashion we like.

abhimanyubabbar (approved):

PR #2959, opened 2023-02-09
fix: move recovery from scheduler to upload

lvrach:
# Description

## Context
Under some conditions, recovery can take time. We have seen cases from 30 minutes to 4 hours.

The recovery can block the scheduler loop for a particular destination type. Causing significant delays to upload creations.

## Mitigation

By moving recovery to the upload loop we mitigate the problem. A single blocking recovery process can not block other uploads from being processed. Multiple (>8) blocking recoveries can saturate the available workers, in this case, we will increase the available workers.

## Refactoring & Tests

The original was split into two parts:
* repo: querying the database to get interrupted destinations 
* service: the logic of triggering the recovery once, if it is needed

In an effort to simplify the recovery interface, the detection - querying the database -  happens just in time.

The logic is now thread-safe, expecting multiple warehouses to call for the same destinationID at the same time. This is why a map of `sync.Once` was introduced.

## Notion Ticket

https://www.notion.so/rudderstacks/Move-recovery-from-scheduler-to-upload-7617b94967cd4876b94bde6dc59f3616?pvs=4

## Security

### DIFF
--- a/warehouse/internal/service/recovery.go
+++ b/warehouse/internal/service/recovery.go
@@ -1,0 +1,79 @@
+package service
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/rudderlabs/rudder-server/warehouse/integrations/manager"
+	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+	"golang.org/x/exp/slices"
+)
+
+var crashRecoverWarehouses = []string{
+	warehouseutils.RS,
+	warehouseutils.POSTGRES,
+	warehouseutils.MSSQL,
+	warehouseutils.AZURE_SYNAPSE,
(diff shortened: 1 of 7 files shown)


sivashanmukh (approved):
LGTM

warehouse/internal/service/recovery.go
```diff
@@ -0,0 +1,86 @@
...
+	once, ok := r.inRecovery[wh.Destination.ID]
+	if !ok {
+		return nil
+	}
+
+	once.Do(func() {
+		once.err = whManager.CrashRecover(wh)
+	})
+
+	return once.err
```
achettyiitr, line 85: We should probably remove it from the `inRecovery` once the recovery is successfull?

achettyiitr (approved):

PR #2967, opened 2023-02-09
fix: handle NULL on upload timings

lvrach:
# Description

The original code didn't handle NULL correctly, but since the error has not been handled we weren't aware of the issue.
After a recent refactoring, the error code was exposed as a log.

In this PR, I have moved the SQL-related code to repo, added tests and handle the NULL case using `COALESCE(timings, '[]')::JSONB`

## Notion Ticket

https://www.notion.so/rudderstacks/Handle-NULL-timings-0e2e23f055c54ee2abb6e19420ee88e3?pvs=4

## Security

### DIFF
--- a/warehouse/internal/repo/upload.go
+++ b/warehouse/internal/repo/upload.go
@@ -377,6 +377,35 @@ func (uploads *Uploads) UploadJobsStats(ctx context.Context, destType string, op
 	return stats, nil
 }
 
+func (uploads *Uploads) UploadTimings(ctx context.Context, uploadID int64) (model.Timings, error) {
+	var (
+		rawJSON jsoniter.RawMessage
+		timings model.Timings
+	)
+
+	err := uploads.db.QueryRowContext(ctx, `
+		SELECT
+			COALESCE(timings, '[]')::JSONB
+		FROM
+			`+uploadsTableName+`
+		WHERE
+			id = $1;
+	`, uploadID).Scan(&rawJSON)
+	if err == sql.ErrNoRows {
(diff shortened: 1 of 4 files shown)


achettyiitr (approved):

warehouse/internal/repo/upload.go
```diff
@@ -377,6 +377,35 @@ func (uploads *Uploads) UploadJobsStats(ctx context.Context, destType string, op
 	return stats, nil
 }
 
+func (uploads *Uploads) UploadTimings(ctx context.Context, uploadID int64) (model.Timings, error) {
```
abhimanyubabbar, line 380: To keep the interface of the upload repo a bit tight, would it be beneficial to look for the `upload_by_id` and fetch the information from it ?

Not sure, if that would be efficient. Depends on the call count maybe. Wdyt ?

abhimanyubabbar (approved):

warehouse/internal/repo/upload.go
lvrach: It is a good point, with some refactoring in the service layer it would make a lot of sense to retrieve the full model and modify it, then updated it in the db again.

I could simplify fetching the whole model here, but I am concerned with the risk of changing too much in this particular PR.

I can add a TODO and DEPRECATED comment here, so we don't forget them.

Once we're able to move all the DB access inside the repo, it would be much easier to refactor it into an ideal design.

abhimanyubabbar: Sounds good to me 👍🏼

PR #2975, opened 2023-02-14
fix(warehouse): added support for filtering on the uploads and calculating aborted events for task_run_id

abhimanyubabbar:
# Description

The PR augments the capability of repo to count over wh_uploads using generic filters which gets used in the pending events API when calculating pending events and aborted events.

## Notion Ticket

https://www.notion.so/rudderstacks/Fix-the-Pending-Events-API-to-return-always-if-the-events-have-been-aborted-in-a-boolean-flag-e251b60a20484513ad802fe4ed50d790

## Security

### DIFF
--- a/warehouse/internal/repo/upload.go
+++ b/warehouse/internal/repo/upload.go
@@ -203,6 +203,35 @@ func (uploads *Uploads) CreateWithStagingFiles(ctx context.Context, upload model
 	return uploadID, nil
 }
 
+type FilterBy struct {
+	Key       string
+	Value     interface{}
+	NotEquals bool
+}
+
+func (uploads *Uploads) Count(ctx context.Context, filters ...FilterBy) (int64, error) {
+	query := fmt.Sprintf("SELECT COUNT(*) FROM %s WHERE 1=1", uploadsTableName)
+
+	args := make([]interface{}, 0)
+	for i, filter := range filters {
+
+		if filter.NotEquals {
+			query += fmt.Sprintf(" AND %s!=$%d", filter.Key, i+1)
+		} else {
+			query += fmt.Sprintf(" AND %s=$%d", filter.Key, i+1)
+		}
+
+		args = append(args, filter.Value)
+	}
+
+	var count int64
+	if err := uploads.db.QueryRowContext(ctx, query, args...).Scan(&count); err != nil {
+		return 0, fmt.Errorf("scanning count into local variable: %w", err)
+	}
+
+	return count, nil
+}
+
 func (uploads *Uploads) Get(ctx context.Context, id int64) (model.Upload, error) {
 	row := uploads.db.QueryRowContext(ctx, `
 		SELECT
--- a/warehouse/warehouse.go
+++ b/warehouse/warehouse.go
@@ -1161,8 +1161,8 @@ func pendingEventsHandler(w http.ResponseWriter, r *http.Request) {
 	sourceID, taskRunID := pendingEventsReq.SourceID, pendingEventsReq.TaskRunID
 	// return error if source id is empty
 	if sourceID == "" || taskRunID == "" {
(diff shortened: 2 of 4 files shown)


warehouse/warehouse.go
```diff
@@ -1195,11 +1195,10 @@ func pendingEventsHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	sourceID := pendingEventsReq.SourceID
-
+	sourceID, taskRunID := pendingEventsReq.SourceID, pendingEventsReq.TaskRunID
 	// return error if source id is empty
-	if sourceID == "" {
-		pkgLogger.Errorf("[WH]: pending-events:  Empty source id")
+	if sourceID == "" || taskRunID == "" {
+		pkgLogger.Errorf("[WH]: pending-events:  Empty source_id or task_run_id ")
```
abhimanyubabbar, line 1201: I'll make sure we have task_run_id coming always in pending events api

warehouse/internal/repo/upload.go
```diff
@@ -205,6 +205,35 @@ func (uploads *Uploads) CreateWithStagingFiles(ctx context.Context, upload model
...
+		} else {
+			query += fmt.Sprintf(" AND %s=$%d", filter.Key, i+1)
+		}
+
+		args = append(args, filter.Value)
+	}
+
+	var count int64
+	if err := uploads.db.QueryRowContext(ctx, query, args...).Scan(&count); err != nil {
+		return 0, fmt.Errorf("scanning count into local variable: %w", err)
```
achettyiitr, line 231:
```suggestion
		return 0, fmt.Errorf("count uploads: %w", err)
```
The error seems to be misleading w.r.t.  local variable. Can we update it?

warehouse/warehouse.go
achettyiitr, line 1201: Can we make it a structured log? Also, shouldn't it be a warning?

achettyiitr (approved):

warehouse/warehouse.go
```diff
@@ -1233,14 +1232,31 @@ func pendingEventsHandler(w http.ResponseWriter, r *http.Request) {
...
+		{Key: "metadata->>'source_task_run_id'", Value: taskRunID},
+		{Key: "status", NotEquals: true, Value: model.ExportedData},
+		{Key: "status", NotEquals: true, Value: model.Aborted},
+	}
+
+	pendingUploadCount, err = getFilteredCount(ctx, filters...)
+
+	if err != nil {
+		err := fmt.Errorf("getting pending uploads count: %v", err)
+		pkgLogger.Errorf("[WH]: %v", err)
```
achettyiitr, line 1246: Can we make it a structured log? Also, should we make this warning?

warehouse/warehouse.go
abhimanyubabbar: Why a warning and not error ?

achettyiitr: Since the error is not causing our application to be in trouble or involving us in the investigation immediately. I thought it was okay to mark them as a warning.
Also, since these APIs are being polled continuously by the extract, we might see too much noise as well.

warehouse/warehouse.go
```diff
@@ -1233,16 +1232,33 @@ func pendingEventsHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	filterBy := []warehouseutils.FilterBy{{Key: "source_id", Value: sourceID}}
-	if pendingEventsReq.TaskRunID != "" {
-		filterBy = append(filterBy, warehouseutils.FilterBy{Key: "metadata->>'source_task_run_id'", Value: pendingEventsReq.TaskRunID})
+	filters := []repo.FilterBy{
+		{Key: "source_id", Value: sourceID},
```
am6010, line 1236: Have we defined the table columns as constants? Maybe it would be safer to use constants.

am6010 (approved):

PR #2976, opened 2023-02-15
chore(warehouse): log improvements for total count in warehouse integrations

achettyiitr:
# Description

Since we are already logging the error at the caller function, we don't want to log it twice. This is creating too much noise for the `ERROR` logs.

## Notion Ticket

https://www.notion.so/rudderstacks/Log-improvement-for-Total-count-in-warehouse-integrations-fce52fb18a0d4b16947ea2900e1ba02e?pvs=4

## Security

### DIFF
--- a/warehouse/integrations/deltalake/deltalake.go
+++ b/warehouse/integrations/deltalake/deltalake.go
@@ -1090,7 +1090,9 @@ func (dl *Deltalake) GetTotalCountInTable(ctx context.Context, tableName string)
 	if err != nil {
 		return 0, fmt.Errorf("fetching table count: %w", err)
 	}
-
+	if response.GetErrorCode() != "" {
+		return 0, fmt.Errorf("fetching table count: %s", response.GetErrorMessage())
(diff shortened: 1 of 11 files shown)


warehouse/integrations/deltalake/deltalake.go
```diff
@@ -1070,35 +1070,30 @@ func (*Deltalake) DownloadIdentityRules(*misc.GZipWriter) (err error) {
 	return
 }
 
-// GetTotalCountInTable returns total count in tables.
-func (dl *Deltalake) GetTotalCountInTable(ctx context.Context, tableName string) (total int64, err error) {
-	fetchTotalCountExecTime := dl.Stats.NewTaggedStat("warehouse.deltalake.grpcExecTime", stats.TimerType, stats.Tags{
```
achettyiitr, line 1075: Stats are not required now. We added these during POC for the python connector. Removing this now.

lvrach (approved):

PR #2981, opened 2023-02-15
fix(warehouse): added check for nil warehouse manager during error mapping

achettyiitr:
# Description

Added nil check for warehouse manager during error mappings.

## Notion Ticket

https://www.notion.so/rudderstacks/Nil-check-for-manager-in-Error-Handling-e1d56655a88f482a8aaaedd83a859882?pvs=4

## Security


lvrach (approved):

warehouse/stats.go
```diff
@@ -152,7 +154,10 @@ func (job *UploadJobT) generateUploadAbortedMetrics() {
 
 	// Set the upload_aborted stat
 	attempts := job.getAttemptNumber()
-	tags := []Tag{{Name: "attempt_number", Value: strconv.Itoa(attempts)}}
+	tags := []Tag{
+		{Name: "attempt_number", Value: strconv.Itoa(attempts)},
+		{Name: "error_mapping", Value: string(model.Noop)},
```
abhimanyubabbar, line 159: should value be `model.Noop` here ?

abhimanyubabbar (approved):

warehouse/stats.go
achettyiitr: Thanks @abhimanyubabbar. upload_aborted metrics is being sent twice at https://github.com/rudderlabs/rudder-server/blob/dfc227302b23bf8ceb3cdc72601f1bbd94435178/warehouse/upload.go#L1747 and https://github.com/rudderlabs/rudder-server/blob/dfc227302b23bf8ceb3cdc72601f1bbd94435178/warehouse/stats.go#L135

achettyiitr: Removing it from here.

PR #2993, opened 2023-02-17
chore(warehouse): added noop uploader

achettyiitr:
# Description

- Added `noop` uploader to avoid having default implementations.
- This interface is being used in multiple places and every time we wanted to provide `default` implementation of unused functions. So having `noop` uploader will help here.
- UploadJobT -> UploadJob, UploaderI -> Uploader

## Notion Ticket

https://www.notion.so/rudderstacks/Noop-uploader-d456a95c38324b3c97f3d45574caf432?pvs=4

## Security


lvrach (commented):
It is not a bad idea, but: 
* I see we mostly used for testing
* The fact that many methods remain un-implemented is a code-smell we need to address with better interfaces and abstractions 
* Danger of extended a NOOP interface, is that you might accidentally forget to implement something, for instance `UseRudderStorage() bool`

lvrach (approved):
Non-blocking concerns, other than NOOP I am cool with other changes.

achettyiitr:
> It is not a bad idea, but:
> 
> * I see we mostly used for testing
> * The fact that many methods remain un-implemented is a code-smell we need to address with better interfaces and abstractions
> * Danger of extended a NOOP interface, is that you might accidentally forget to implement something, for instance `UseRudderStorage() bool`

Thanks @lvrach for sharing, Let me come up with a better solution than this.

PR #2998, opened 2023-02-20
refactor: prefer using lo.BufferWithTimeout

Sidddddarth:
# Description

Using `lo.BufferWithTimeout` that either waits for a number of objects to receive from a channel or a timeout whichever's first.

## Notion Ticket

[BufferWithTimeout](https://github.com/samber/lo#bufferwithtimeout)

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -1446,53 +1446,24 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
 // statusInsertLoop will run in a separate goroutine
 // Blocking method, returns when rt.responseQ channel is closed.
 func (rt *HandleT) statusInsertLoop() {
-	var responseList []jobResponseT
-
-	// Wait for the responses from statusQ
-	lastUpdate := time.Now()
-
 	statusStat := stats.Default.NewTaggedStat("router_status_loop", stats.TimerType, stats.Tags{"destType": rt.destName})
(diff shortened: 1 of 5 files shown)


router/router.go
```diff
@@ -1477,54 +1477,26 @@ func (rt *HandleT) commitStatusList(responseList *[]jobResponseT) {
...
-			lastUpdate = time.Now()
+			rt.commitStatusList(&jobResponseBuffer)
+			countStat.Count(numJobResponses)
 			statusStat.Since(start)
 		}
+		if !isResponseQOpen {
+			rt.logger.Debugf("[%v Router] :: statusInsertLoop exiting", rt.destName)
+			return
+		}
+		time.Sleep(maxStatusUpdateWait - timeToBuffer)
```
atzoum, line 1499: This doesn't take into consideration the time taken to commit. But overall, adding some extra sleep when the buffer is getting full faster than the `maxStatusUpdateWait` doesn't provide any actual benefit, on the contrary.

atzoum (commented):
nice!

router/router.go
Sidddddarth: agreed
removing it

atzoum:
@Sidddddarth is this still work in progress, or ready for review?

atzoum (approved):

cisse21 (approved):

PR #2999, opened 2023-02-20
feat(gateway): support new event type extract

debanjan97:
# Description

As part of supporting a new event type, `extract`, we are adding
a new endpoint `/internal/extract` for ingesting the event.
We are not enforcing the event payload to have either userId or anonymousId.
The same is reflected in the test case modification.

## Notion Ticket
https://www.notion.so/rudderstacks/Development-New-event-for-extract-b9e4d7abef294bd7a690acc0332a18ea?pvs=4

## Security


gateway/gateway.go
```diff
@@ -114,8 +114,13 @@ var BatchEvent = []byte(`
 const (
 	DELIMITER                 = string("<<>>")
 	eventStreamSourceCategory = "eventStream"
+	extractEvent              = "extract"
 )
 
+// EventTypesSkipUserCheck is a list of event types that
+// do not require an user id or anonymous id
+var EventTypesSkipUserCheck = []string{extractEvent}
```
fracasula, line 122: There is no need to export this variable, we should make it unexported.

gateway/gateway.go
```diff
@@ -694,7 +701,10 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
 	return
 }
 
-func isNonIdentifiable(anonIDFromReq, userIDFromReq string) bool {
+func isNonIdentifiable(anonIDFromReq, userIDFromReq, eventType string) bool {
+	if misc.Contains(EventTypesSkipUserCheck, eventType) {
```
fracasula, line 705: There is no need to use `misc.Contains` since `EventTypesSkipUserCheck` is a slice with one element only, right? We can always have a slice later once we have more than one event type that needs skipping.
```suggestion
	if eventType == extractEvent {
```

gateway/gateway.go
```diff
@@ -1266,6 +1280,7 @@ func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
 	srvMux.HandleFunc("/", WithContentType("application/json; charset=utf-8", app.LivenessHandler(gateway.jobsDB))).Methods("GET")
 	srvMux.HandleFunc("/v1/import", gateway.webImportHandler).Methods("POST")
 	srvMux.HandleFunc("/v1/audiencelist", gateway.webAudienceListHandler).Methods("POST")
+	srvMux.HandleFunc("/v1/extract", gateway.webExtractHandler).Methods("POST")
```
fracasula, line 1283: Can we do `/internal/v1/extract` instead? This way we can easily block unwanted access and make sure that this is used internally only (i.e. by making sure that everything that starts with `/internal` isn't accessible from outside).

fracasula (changes requested):

gateway/gateway.go
debanjan97: Addressed

gateway/gateway.go
debanjan97: Addressed!

gateway/gateway.go
debanjan97: Addressed!

gateway/gateway.go
```diff
@@ -1248,6 +1259,7 @@ func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
 	srvMux.HandleFunc("/", WithContentType("application/json; charset=utf-8", app.LivenessHandler(gateway.jobsDB))).Methods("GET")
 	srvMux.HandleFunc("/v1/import", gateway.webImportHandler).Methods("POST")
 	srvMux.HandleFunc("/v1/audiencelist", gateway.webAudienceListHandler).Methods("POST")
+	srvMux.HandleFunc("/internal/v1/extract", gateway.webExtractHandler).Methods("POST")
```
fracasula, line 1262: Thanks for the changes @debanjan97 ! Two more things.

1. Can we have the internals under a different mux?

```go
srvMux := mux.NewRouter()
internalMux := srvMux.PathPrefix("/internal").Subrouter()

internalMux.HandleFunc("/v1/extract", gateway.webExtractHandler).Methods("POST")
```

2. Can we have an HTTP middleware that checks if the request path begins with `/internal` and if it does return a 401 if a token in the `Authorization` header is not set? For now just checking if the token is not empty is OK, then in a later PR we can set an environment variable and compare against that.

fracasula (changes requested):

gateway/gateway.go
debanjan97: I was planning to add the middleware on the internal mux so that we can skip checking the request path.
That way, the middleware need not be restricted to the endpoints containing`/internal`; it can be used wherever we need authorization.

```go
internalMux := srvMux.PathPrefix("/internal").Subrouter() // mux for internal endpoints
internalMux.Use(
	middleware.InternalTokenMiddleware,
)
```

debanjan97: In addition to that, currently, we overload the Authorization header for fetching the write key.
We have checks (not in any middleware, though) which reject requests if the `Authorization` header is missing.
https://github.com/rudderlabs/rudder-server/blob/b71ae92859b1dd9ac8720a0f54d7e05bad0ec26f/gateway/gateway.go#L943

Unless we decouple fetching the `writeKey` from the actual authorization, we might have to look for another way to reject calls to the internal endpoints without proper authorization.
Unless we are adding the middleware to currently act as a placeholder for what is next to come.

fracasula: @debanjan97 let's just add an internal mux then. we'll address the authorization in a later PR since it is more complex.

debanjan97: addressed!

fracasula (approved):

am6010 (approved):
Nice !!

gateway/gateway.go
```diff
@@ -1256,6 +1269,9 @@ func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
 	srvMux.HandleFunc("/version", WithContentType("application/json; charset=utf-8", gateway.versionHandler)).Methods("GET")
 	srvMux.HandleFunc("/robots.txt", gateway.robots).Methods("GET")
 
+	// internal endpoints
+	internalMux.HandleFunc("/v1/extract", gateway.webExtractHandler).Methods("POST")
```
atzoum, line 1273: @cisse21 doesn't this require proxy adaptations too so that not a random node is picked?

PR #3000, opened 2023-02-20
fix(warehouse): skipping deprecated columns

achettyiitr:
# Description

- Users tables have a hard dependency on `identifies` tables to have a similar schema. 
- Since in the `ALTER` column flow, if there is a dependent column, we are leaving the dependent columns as it is asking for customers to drop it.
- This is causing differences in the schema for users and identifies and it's causing failure for the creation of staging users table in [here](https://github.com/rudderlabs/rudder-server/blob/47a12bd3a270d76da8e2509bd59746ddd078a04d/warehouse/integrations/redshift/redshift.go#L566).
- Corresponding error logs:
`pq: column "context_<prop>-deprecated-<uuid>" does not exist in rudder_staging_identifies_73ffe05a67886765c9f0ebb64fe8a1e01`

## Notion Ticket

https://www.notion.so/rudderstacks/skipping-deprecated-columns-14e79ea3e5ac4a3280b71aead9634f76?pvs=4

## Security

### DIFF
--- a/warehouse/schema.go
+++ b/warehouse/schema.go
@@ -6,10 +6,11 @@ import (
 	"fmt"
 	"reflect"
 
-	"github.com/rudderlabs/rudder-server/warehouse/integrations/manager"
+	"github.com/rudderlabs/rudder-server/warehouse/logfield"
 
 	"github.com/rudderlabs/rudder-server/utils/misc"
 	"github.com/rudderlabs/rudder-server/utils/timeutil"
+	"github.com/rudderlabs/rudder-server/warehouse/integrations/manager"
 	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
 	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
 )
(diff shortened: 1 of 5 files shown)


warehouse/utils/utils.go
```diff
@@ -135,6 +135,8 @@ const (
 	MinioSecretAccessKey = "secretAccessKey"
 )
 
+var DeprecatedColumnsRegex = regexp.MustCompile(`.*-deprecated-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`)
```
abhimanyubabbar, line 138: Is this a UUID in the end ? Also can you write a comment upwards describing what this regex matches ideally ?

abhimanyubabbar: `abc-deprecated-1234-8321..`

abhimanyubabbar (approved):
LGTM 👍🏼

abhimanyubabbar (approved):

warehouse/schema.go
```diff
@@ -6,10 +6,11 @@ import (
 	"fmt"
 	"reflect"
 
-	"github.com/rudderlabs/rudder-server/warehouse/integrations/manager"
+	"github.com/rudderlabs/rudder-server/warehouse/logfield"
```
cisse21, line 9: club this with the imports below

cisse21 (approved):

PR #3002, opened 2023-02-20
fix: inconsistent and leaky retry delay logic in router

atzoum:
# Description

Addressing router job retry backoff inconsistencies which may also cause memory leak:
- We don’t delete entries from the map in case we drain jobs or mark them as aborted in places other than `postStatusOnResponseQ`.
- In case event ordering is disabled, the worker-specific map we are using now is not really useful, since there is no guarantee that the same job will always be assigned to the same worker…

To fix these issues we are no longer using an in-memory map per worker, but the job status' `retry_time` column is used for storing the calculated backoff time. JobsDB queries for retrieving jobs don't use any conditions against the `retry_time` column, otherwise in case of a server restart this would cause jobs to be picked out-of-order as router's event-ordering algorithm doesn't persist its state, it only retains it in-memory.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=6be0e77514b241b698545cc84172b225&pm=s)

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -181,8 +181,6 @@ type workerT struct {
 	workerID                  int                 // identifies the worker
 	failedJobs                int                 // counts the failed jobs of a worker till it gets reset by external channel
 	barrier                   *eventorder.Barrier
-	retryForJobMap            map[int64]time.Time     // jobID to next retry time map
-	retryForJobMapMutex       sync.RWMutex            // lock to protect structure above
 	routerJobs                []types.RouterJobT      // slice to hold router jobs to send to destination transformer
 	destinationJobs           []types.DestinationJobT // slice to hold destination jobs
 	rt                        *HandleT                // handle to router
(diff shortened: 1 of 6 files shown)


router/router.go
```diff
@@ -1023,19 +1015,12 @@ func (worker *workerT) postStatusOnResponseQ(respStatusCode int, payload json.Ra
...
-					worker.retryForJobMap[destinationJobMetadata.JobID] = time.Now().Add(durationBeforeNextAttempt(status.AttemptNum))
-					worker.retryForJobMapMutex.Unlock()
+					status.RetryTime = time.Now().Add(durationBeforeNextAttempt(status.AttemptNum))
 				}
 			}
 		} else if respStatusCode == 429 {
-			worker.retryForJobMapMutex.Lock()
-			worker.retryForJobMap[destinationJobMetadata.JobID] = time.Now().Add(durationBeforeNextAttempt(status.AttemptNum))
-			worker.retryForJobMapMutex.Unlock()
+			status.RetryTime = time.Now().Add(durationBeforeNextAttempt(status.AttemptNum))
```
Sidddddarth, line 1023: can do this once for both cases(500 fail, 429) a few lines below after the `jobState` has been decided.

```
if status.JobState == jobsdb.Failed.State {

    status.RetryTime = time.Now().Add(durationBeforeNextAttempt(status.AttemptNum))

}
```

router/router.go
```diff
@@ -1256,14 +1240,7 @@ func (rt *HandleT) findWorker(job *jobsdb.JobT, throttledUserMap map[string]stru
 }
 
 func (worker *workerT) canBackoff(job *jobsdb.JobT) (shouldBackoff bool) {
```
Sidddddarth, line 1242:
```suggestion
func (*workerT) canBackoff(job *jobsdb.JobT) bool {
// or 
func canBackoff(job *jobsdb.JobT) bool {
```

Sidddddarth (approved):

router/router.go
atzoum: Unfortunately `jobState` can be amended later to `aborted`. Will try to refactor 👍

router/router.go
atzoum: or inline :D

router/router.go
Sidddddarth: LGTM
Only complaint is things go way beyond 80 columns in some places :-p

Sidddddarth (approved):

router/router.go
atzoum: Have you ever used a [Punch Card](https://blogs.sap.com/2022/07/26/the-magic-80-char-limit-which-used-to-exist-for-developers/)?!

cisse21 (approved):

PR #3007, opened 2023-02-20
feat: kafka over ssh

fracasula:
# Description

This PR is to add support for connecting to Kafka via SSH. It is required by a specific customer and control plane prefers not to have SSH keys on their side so for this one we're opting to configure via environment variables only.

The PR was tested manually with a Kafka over SSH on AWS (EC2).

Related [Slack conversation](https://rudderlabs.slack.com/archives/C04Q7BPRG8K/p1676701804135719).

# Sidenote

In order to make the test work I had to make changes to the way Kafka was provisioned so that the [advertised listeners](https://www.confluent.io/blog/kafka-listeners-explained/) were set correctly.

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/Add-Kafka-SSH-support-457bccbd5976451383fb2ff3c7e6d3b4) >

## Security


services/streammanager/kafka/client/config.go
```diff
@@ -51,6 +51,12 @@ type Config struct {
 	DialTimeout time.Duration
 	TLS         *TLS
 	SASL        *SASL
+	SSHConfig   *SSHConfig
+}
+
+type SSHConfig struct {
+	Key, User, Host  string
```
lvrach, line 58:
```suggestion
	PrivateKey, User, Host  string
```

services/streammanager/kafka/client/client.go
```diff
@@ -64,6 +66,33 @@ func New(network string, addresses []string, conf Config) (*Client, error) {
...
+
+			conn, err := sshClient.Dial(network, address)
+			if err != nil {
+				return nil, fmt.Errorf(
+					"cannot dial address %q over SSH (host %q): %w", address, conf.SSHConfig.Host, err,
+				)
+			}
+			return conn, nil
+		}
+	}
```
lvrach, line 94: Should we set the net Dialer with Timeout if not defined?

```suggestion
	}
	dialer.DialFunc = net.Dialer{
		Timeout: c.config.DialTimeout,
	}.DialContext
```

fracasula: I've added a timeout in the SSH config. It works as expected when dialing. Wdyt?

services/streammanager/kafka/kafkamanager.go
```diff
@@ -659,3 +660,15 @@ func getStatusCodeFromError(err error) int {
 	}
 	return 400
 }
+
+func getSSHConfig(c *config.Config) *client.SSHConfig {
```
atzoum, line 664: Would it make sense to be able to enable it globally, but also allow for destinationID-specific overrides? e.g.

```yaml
Router:
  KAFKA:
    ssh:
      enabled: false
      destinationID1:
        enabled: true
        user: "user"
        host: "host"
```

fracasula: @lvrach would this be OK or we would have issues with the SSH private key in this case as well?

lvrach: If we want to support this feature for general usage from multiple customers, it should be configurable for the control plane/app dashboard.

However, in this case, we want to experimentally (and quickly) support this for a single customer in an enterprise setup.

services/streammanager/kafka/client/client.go
```diff
@@ -64,6 +66,37 @@ func New(network string, addresses []string, conf Config) (*Client, error) {
...
+			User:    conf.SSHConfig.User,
+			Auth:    []ssh.AuthMethod{ssh.PublicKeys(signer)},
+			Timeout: conf.DialTimeout,
+		}
+		if conf.SSHConfig.AcceptAnyHostKey {
+			sshConfig.HostKeyCallback = ssh.InsecureIgnoreHostKey() // skipcq: GSC-G106
+		}
+
+		dialer.DialFunc = func(ctx context.Context, network, address string) (net.Conn, error) {
+			sshClient, err := ssh.Dial("tcp", conf.SSHConfig.Host, sshConfig)
```
snarkychef, line 85: Why do we create sshClient for every connection? Why not use single client and multiplexing connections?

snarkychef: Never mind. Maybe one broker not responding can hand the central ssh connection. And downside is not much as we create as many ssh clients as brokers, which is not very high.

atzoum (approved):

lvrach (approved):

PR #3011, opened 2023-02-20
fix: connection reset error

koladilip:
# Description

    Re-Initializing HTTP client if connection got reset.

## Security


utsabc (approved):

sanpj2292 (approved):

router/transformer/transformer.go
```diff
@@ -139,6 +141,10 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
 		}
 
 		if err != nil {
+			// Re-Initializing http client if connection got reset
+			if errors.Is(err, syscall.ECONNRESET) {
```
atzoum, line 145: I don't understand why we need to recreate the client. Please elaborate or prove the issue with a test case.

atzoum: btw, an `ECONNRESET` happens if server terminates the response abruptly while the client is trying to read the response's body, i.e. the server did reply to the request, it sent the response headers but stopped transmitting while sending the body

koladilip: You are right, Kube service is not caching; it can rotate to different when the pod restarts, so client initialization is not required.

PR #3022, opened 2023-02-22
chore: support for refreshing token during transformation and delivery

sanpj2292:
# Description

Previously we used to support refreshing of oauth token during delivery of event data to destination.
Now, we are trying to support refreshing of oauth token even during transformation as well

## Notion Ticket

https://www.notion.so/rudderstacks/OAuth-Token-expiry-during-transformation-doesn-t-refresh-access-token-c2f5f269455248d0b413a60bcc2951c1

## Security


koladilip (approved):

lvrach (approved):
The code changes do make sense to me, but I would like to explore the possibility of writing a test for this.

sanpj2292:
> I would like to explore the possibility of writing a test for this

Currently I am not seeing any test-cases for `processDestinationJobs` method.
I think the method `processDestinationJobs` has to be refactored quite a bit to make sure that we can add test-cases.
Any advice here ? @lvrach

PR #3023, opened 2023-02-22
chore: make GC discard ratio configurable

cisse21:
# Description

Have an aggressive GC for Live Events Badger DB usage and add few more options to reduce disk consumption

## Notion Ticket
https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=c29bc50da3d042dca7d109cc66447a42&pm=s

## Security

### DIFF
--- a/services/debugger/cache/internal/badger/badger.go
+++ b/services/debugger/cache/internal/badger/badger.go
@@ -27,6 +27,7 @@ func (e *Cache[E]) loadCacheConfig() {
 	config.RegisterDurationConfigVariable(5, &e.ticker, false, time.Minute, "LiveEvent.cache.GCTime")
 	config.RegisterDurationConfigVariable(15, &e.queryTimeout, false, time.Second, "LiveEvent.cache.queryTimeout")
 	config.RegisterIntConfigVariable(3, &e.retries, false, 1, "LiveEvent.cache.retries")
+	config.RegisterFloat64ConfigVariable(0.5, &e.gcDiscardRatio, false, "LiveEvent.cache.gcDiscardRatio")
 }
 
 /*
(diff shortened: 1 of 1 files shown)


services/debugger/cache/internal/badger/badger.go
```diff
@@ -27,25 +27,27 @@ func (e *Cache[E]) loadCacheConfig() {
 	config.RegisterDurationConfigVariable(5, &e.ticker, false, time.Minute, "LiveEvent.cache.GCTime")
```
atzoum, line 27: shouldn't we decrease this too, e.g. to `10s`?

services/debugger/cache/internal/badger/badger.go
```diff
@@ -172,7 +177,7 @@ func (e *Cache[E]) gcBadgerDB() {
 			return
 		case <-ticker.C:
 		again: // see https://dgraph.io/docs/badger/get-started/#garbage-collection
-			err := e.db.RunValueLogGC(0.7)
+			err := e.db.RunValueLogGC(0.5)
```
atzoum, line 180:
```suggestion
			err := e.db.RunValueLogGC(e.gcDiscardRatio)
```

services/debugger/cache/internal/badger/badger.go
cisse21: Should we reduce this even on our systems? Everything looks fine for us right so I just left this as is

lvrach:
What is the impact on CPU and IOPS have we tested it?

cisse21:
I have tested it with the previous setup without these changes and everything looked good without any spikes in CPU/Disk/Ram consumption but this PR is a result of a OSS user having unusual disk usage due to Badger

services/debugger/cache/internal/badger/badger.go
lvrach: Is there any particular reason to make GC more frequent? 

The increased frequency will hardly affect the used storage, while it might become less efficient. 

I suggest increasing the frequency if we observe long GC pauses.

lvrach (approved):

services/debugger/cache/internal/badger/badger.go
atzoum: The issue might occur in our higher-traffic deployments, e.g. `bolt`. `ValueLogGC` doesn't cause the system to pause. Since in this use case we are overwriting the same key multiple times within `5m`, it would be better to perform cleanup in higher frequencies.

lvrach: I stand corrected:  ~long GC pauses~ -> GC execution times

Within the 10s period, there might not be a log page to garbage collect and there is certainly some cost involved in calling the GC more frequently. 

We can do a quick test for `bolt` and see the impact of this change. 

I understand that this is minor, but given how important badgerDB is for us, I would like to make an informed decision about this change and increase our understanding.

PR #3031, opened 2023-02-23
feat(warehouse): snowflake roles support.

achettyiitr:
# Description

Added roles support for snowflake

## Notion Ticket

https://www.notion.so/rudderstacks/Snowflake-support-for-roles-bc2f264c6f0d4fd28a585b3424e954bd?pvs=4

## Security


lvrach (approved):
Shall we add an integration test were role is used ?

Other than that looks good to me.

achettyiitr:
> Details

Good point. Let me add this as well.

achettyiitr:
> Shall we add an integration test were role is used ?
> 
> Other than that looks good to me.

Good point. Let me add this as well.

cisse21 (approved):

PR #3035, opened 2023-02-24
chore: reduce parquet file size datalake

cisse21:
# Description

- Reduce parquet file size by not having a unified merged schema 
- Get rid of the unused `mergedschema` column
- Clean up unused `useParquetLoadFilesRS` config variable

## Notion Ticket
[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=357cfbeda3514c20ba9023e224728293&pm=s)

## Security


achettyiitr (approved):

sql/migrations/warehouse/000021_drop_merged_schema_from-wh_uploads.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE wh_uploads DROP COLUMN IF EXISTS mergedschema;
```
lvrach, line 1: Lets not rush to drop the column, we might need to revert

warehouse/internal/repo/staging_test.go
```diff
@@ -522,8 +523,10 @@ func TestStagingFileIDs(t *testing.T) {
 			ID: 3,
 		},
 	}
-
-	ids := repo.StagingFileIDs(sfs)
+	var ids []int64
+	lo.ForEach(sfs, func(sf *model.StagingFile, _ int) {
+		ids = append(ids, sf.ID)
+	})
```
lvrach, line 529: I find the usage of `lo.ForEach` hindering readability in this case. 

We can always use a for loop. 
Or use `lo.Map`.

I personally prefer `repo.StagingFileIDs` since it is a common/repeated behaviour in our code.

warehouse/upload.go
```diff
@@ -170,6 +170,10 @@ func setMaxParallelLoads() {
 }
 
 func (f *UploadJobFactory) NewUploadJob(dto *model.UploadJob, whManager manager.Manager) *UploadJobT {
+	var stagingFileIDs []int64
+	lo.ForEach(dto.StagingFiles, func(stagingFile *model.StagingFile, _ int) {
+		stagingFileIDs = append(stagingFileIDs, stagingFile.ID)
+	})
```
lvrach, line 176: Similar to my comment above.

lvrach (changes requested):
You should not drop the column just yet.

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -237,13 +234,11 @@ func (lf *LoadFileGenerator) createFromStaging(ctx context.Context, job model.Up
...
-		batchStartIdx := i
-		batchEndIdx := j
-
+		startId := chunk[0].ID
+		endId := chunk[len(chunk)-1].ID
 		g.Go(func() error {
 			responses := <-ch
-			lf.Logger.Infof("[WH]: Received responses for staging files %d:%d for %s:%s from PgNotifier", toProcessStagingFiles[batchStartIdx].ID, toProcessStagingFiles[batchEndIdx-1].ID, destType, destID)
-
+			lf.Logger.Infof("[WH]: Received responses for staging files %d:%d for %s:%s from PgNotifier", startId, endId, destType, destID)
```
lvrach, line 241: we can convert this to structured log:
```suggestion
			lf.Logger.Infow("Received responses for staging files %d:%d for %s:%s from PgNotifier", 
			"startId", startId, 
			"endID", endId,
			 logfields.DestType, destType,
			 logfields.DestID, destID,
	)
```

lvrach (commented):
Apart from the database column, we also need to consider the revertability of the code.

If we introduce this change and have to revert back, the following code will not work correctly, as the `MergedSchema` won't be defined:

```go
		schema := &job.Upload.UploadSchema
		if job.Upload.LoadFileType == warehouseutils.LOAD_FILE_TYPE_PARQUET {
			schema = &job.Upload.MergedSchema
		}
```

We need to consider if we can mitigate this situation, for example manually populate MergeSchema from upload schema in case of revert.

Or we can proactively populate both columns with `UploadSchema`, so in case we revert to a previous version a valid schema could be used.

warehouse/internal/repo/staging_test.go
cisse21: Yeah I was thinking to change this to the older code too

lvrach (approved):

PR #3041, opened 2023-02-27
fix: enabled data race

fracasula:
# Description

Don't mind the stack trace that says v1.5.0, this happens on the latest of RudderServer as well.

Quick fix for:

```
==================
WARNING: DATA RACE
Write at 0x00c000195be0 by goroutine 905:
  github.com/rudderlabs/rudder-server/services/stats.(*statsdStats).Start()
      /Users/debanjanchatterjee/.go/pkg/mod/github.com/rudderlabs/rudder-server@v1.5.0-preview.2/services/stats/stats.go:126 +0x3de
  github.com/rudderlabs/device-mode-transformation-server/cmd/server.Run()
      /Users/debanjanchatterjee/backend/device-mode-transformation-server/cmd/server/main.go:51 +0x2c3
  github.com/rudderlabs/device-mode-transformation-server/cmd/server_test.beforeEach.func1()
      /Users/debanjanchatterjee/backend/device-mode-transformation-server/cmd/server/main_test.go:51 +0x3c
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /Users/debanjanchatterjee/.go/pkg/mod/golang.org/x/sync@v0.1.0/errgroup/errgroup.go:75 +0x82

Previous read at 0x00c000195be0 by goroutine 463:
  github.com/rudderlabs/rudder-server/services/stats.(*statsdStats).collectPeriodicStats.func1()
      /Users/debanjanchatterjee/.go/pkg/mod/github.com/rudderlabs/rudder-server@v1.5.0-preview.2/services/stats/stats.go:177 +0xa4
  github.com/rudderlabs/rudder-server/services/stats.runtimeStatsCollector.outputGCStats()
      /Users/debanjanchatterjee/.go/pkg/mod/github.com/rudderlabs/rudder-server@v1.5.0-preview.2/services/stats/periodic.go:155 +0x255
  github.com/rudderlabs/rudder-server/services/stats.runtimeStatsCollector.zeroStats()
      /Users/debanjanchatterjee/.go/pkg/mod/github.com/rudderlabs/rudder-server@v1.5.0-preview.2/services/stats/periodic.go:92 +0x1b2
  github.com/rudderlabs/rudder-server/services/stats.runtimeStatsCollector.run.func1()
      /Users/debanjanchatterjee/.go/pkg/mod/github.com/rudderlabs/rudder-server@v1.5.0-preview.2/services/stats/periodic.go:59 +0x65
  runtime.deferreturn()
      /usr/local/go/src/runtime/panic.go:476 +0x32
  github.com/rudderlabs/rudder-server/services/stats.(*statsdStats).collectPeriodicStats.func2()
      /Users/debanjanchatterjee/.go/pkg/mod/github.com/rudderlabs/rudder-server@v1.5.0-preview.2/services/stats/stats.go:191 +0x117
```

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/Stats-enabled-data-race-babc5c7b1a6341ce9e1e20247a9e9ac5) >

## Security

### DIFF
--- a/services/stats/stats.go
+++ b/services/stats/stats.go
@@ -84,7 +84,6 @@ func NewStats(config *config.Config, loggerFactory *logger.Factory, metricManage
 	s := &statsdStats{
 		log: loggerFactory.NewLogger().Child("stats"),
 		conf: &statsdConfig{
-			enabled:         config.GetBool("enableStats", true),
 			tagsFormat:      config.GetString("statsTagsFormat", "influxdb"),
 			excludedTags:    config.GetStringSlice("statsExcludedTags", nil),
 			statsdServerURL: config.GetString("STATSD_SERVER_URL", "localhost:8125"),
@@ -105,6 +104,7 @@ func NewStats(config *config.Config, loggerFactory *logger.Factory, metricManage
 			pendingClients: make(map[string]*statsdClient),
 		},
 	}
+	s.conf.enabled.Store(config.GetBool("enableStats", true))
 	return s
 }
 
@@ -116,13 +116,13 @@ type statsdStats struct {
 }
 
 func (s *statsdStats) Start(ctx context.Context) {
-	if !s.conf.enabled {
+	if !s.conf.enabled.Load() {
 		return
 	}
 	s.state.conn = statsd.Address(s.conf.statsdServerURL)
 	// since, we don't want setup to be a blocking call, creating a separate `go routine`` for retry to get statsd client.
 	var err error
-	// NOTE: this is to get atleast a dummy client, even if there is a failure. So, that nil pointer error is not received when client is called.
+	// NOTE: this is to get at least a dummy client, even if there is a failure. So, that nil pointer error is not received when client is called.
 	s.state.client.statsd, err = statsd.New(s.state.conn, s.conf.statsdTagsFormat(), s.conf.statsdDefaultTags())
 	if err == nil {
 		s.state.clientsLock.Lock()
@@ -133,7 +133,7 @@ func (s *statsdStats) Start(ctx context.Context) {
 		if err != nil {
 			s.state.client.statsd, err = s.getNewStatsdClientWithExpoBackoff(ctx, s.state.conn, s.conf.statsdTagsFormat(), s.conf.statsdDefaultTags())
 			if err != nil {
-				s.conf.enabled = false
+				s.conf.enabled.Store(false)
 				s.log.Errorf("error while creating new statsd client: %v", err)
 			} else {
 				s.state.clientsLock.Lock()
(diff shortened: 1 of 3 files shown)


services/stats/stats.go
```diff
@@ -202,7 +202,7 @@ func (s *statsdStats) collectPeriodicStats() {
 func (s *statsdStats) Stop() {
 	s.state.clientsLock.RLock()
 	defer s.state.clientsLock.RUnlock()
-	if !s.conf.enabled || !s.state.connEstablished {
+	if !s.conf.enabled.Load() || !s.state.connEstablished {
```
lvrach, line 205: Since `enabled` can change to `false` while the stats run, should we always try to close the channels? 

```suggestion
	if !s.state.connEstablished {
```

services/stats/stats.go
```diff
@@ -133,7 +133,7 @@ func (s *statsdStats) Start(ctx context.Context) {
 		if err != nil {
 			s.state.client.statsd, err = s.getNewStatsdClientWithExpoBackoff(ctx, s.state.conn, s.conf.statsdTagsFormat(), s.conf.statsdDefaultTags())
 			if err != nil {
-				s.conf.enabled = false
+				s.conf.enabled.Store(false)
```
lvrach, line 136: What if we keep enabled as a simple bool, by adding a new field for example:

```suggestion
				s.err.Store(fmt.Errorf("stats client: %w", err))
```

fracasula: 1. The problem is that `enabled` is checked when skipping metrics as well. In any case, please keep in mind that the real contention with atomic here would happen only if someone is actually doing a write. Luckily we write into `enabled` only once (concurrently).
    * The value is set at the beginning and there is no concurrency.
    * Then in the go routine, after all the backoff retries have failed, we give up and write `false` to `enabled` (only once). 
    * There will be no more writes after that. Everything else is a read and an `atomic.Bool` read is super fast (unlike with a mutex). So I don't have performance concerns because there will be virtually no locking (or very little overhead anyway) considering that 99.99% of the times we only **read** `enabled` (see benchmark below).
2. We don't have this race in the OTEL version so I think that this fix is OK for now.

```go
func BenchmarkBool(b *testing.B) {
	b.Run("bool", func(b *testing.B) {
		var x bool
		for i := 0; i < b.N; i++ {
			if !x {
				go func() {
					x = true
				}()
			}
		}
	})
	b.Run("atomic", func(b *testing.B) {
		var x atomic.Bool
		for i := 0; i < b.N; i++ {
			if !x.Load() {
				go func() {
					x.Store(true)
				}()
			}
		}
	})
}
```

Gives:

```
BenchmarkBool
BenchmarkBool/bool
BenchmarkBool/bool-24         	1000000000	         0.2157 ns/op
BenchmarkBool/atomic
BenchmarkBool/atomic-24       	1000000000	         0.2178 ns/op
PASS
```

lvrach: My concern was about using and thus polluting a configuration struct with logic with a state that should belong to `statsdStats`. This is semantic since the conf struct is private, otherwise, it would be strange to ask people to pass atomics to a config.

Also, given that we are going to remove this code latter, I agree that this change is sufficient.

lvrach (approved):

atzoum (approved):

PR #3050, opened 2023-02-28
fix(warehouse): infinites retries

achettyiitr:
# Description

- Since we already had a hard limit for **x Hours and y Times** **we don’t need hasAllTablesSkipped**
    - The problem with having **hasAllTablesSkipped** is it is being **shared across multiple goroutines** during exporting **(regular tables, identifies/users tables, ID resolution)**
    - With the bug `retries causing succeeded tables to get retried again`, during export, the oldest exporting goroutines set it to true in case of successful exports for that goroutine.
    - Later this is used to decide whether we should neglect to mark the upload status as aborted.
- **Once the failed sync is retried, all succeeded tables are getting retries.**
   - We used to set [tableUploadStatuses](https://github.com/rudderlabs/rudder-server/blob/a258f7412f589b3010265ac3fa45857112954ece/warehouse/upload.go#L727) once. But this check does not become invalid because we are [initializing](https://github.com/rudderlabs/rudder-server/blob/a258f7412f589b3010265ac3fa45857112954ece/warehouse/upload.go#L188) it with an empty slice.

## Notion Ticket

https://www.notion.so/rudderstacks/infinite-retries-a61e259c0b9349bfb5485083298d6baf?pvs=4

## Security


warehouse/upload.go
```diff
@@ -963,17 +863,6 @@ func (job *UploadJobT) addColumnsToWarehouse(tName string, columnsMap map[string
...
-type TableSkipError struct {
-	tableName        string
-	previousJobID    int64
-	previousJobError string
-}
-
-func (tse *TableSkipError) Error() string {
-	return fmt.Sprintf("Skipping %s table because it previously failed to load in an earlier job: %d with error: %s", tse.tableName, tse.previousJobID, tse.previousJobError)
-}
-
```
achettyiitr, line 976: Removing this as we are now avoiding TableSkipError.

warehouse/upload_test.go
```diff
@@ -6,20 +6,20 @@ import (
 	"fmt"
 	"os"
 	"testing"
+	time2 "time"
 
 	"github.com/aws/smithy-go/time"
```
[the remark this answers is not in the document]
achettyiitr: Let me remove this. Not sure why we are using this.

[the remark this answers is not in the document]
achettyiitr: Removed this.

warehouse/upload.go
```diff
@@ -683,142 +700,39 @@ func (job *UploadJobT) exportRegularTables(specialTables []string, loadFilesTabl
...
-			&tableUploadStatus.namespace,
-			&tableUploadStatus.tableName,
-			&tableUploadStatus.status,
-			&tableUploadStatus.error,
-		)
-		if err != nil {
-			panic(err)
+		// Previous upload and table upload failed
+		exportingfailedStatus = func(status string) bool {
+			return status == TableUploadExportingFailed || status == UserTableUploadExportingFailed || status == IdentityTableUploadExportingFailed
```
lvrach, line 723: [minor] & opinionated
```suggestion
                         switch status {
                         case TableUploadExportingFailed, UserTableUploadExportingFailed, IdentityTableUploadExportingFailed:
                             return true
                          default:
                             return false
                          } 
```

lvrach: We should consider moving this logic outside of this function. Maybe it could be part of the table upload model, or introduce a separate failure flag. 

The concern I have is when we introduce a new failed state we might forget to add it here.

lvrach (approved):

warehouse/upload.go
achettyiitr: Good point.

achettyiitr: Let me make the change.

warehouse/internal/repo/upload_test.go
```diff
@@ -762,3 +762,70 @@ func TestUploads_InterruptedDestinations(t *testing.T) {
...
+		(2, 1, 1, true, 'RS', 'aborted', 'test_namespace', '{}', NOW(), NOW())
+	`)
+	require.NoError(t, err)
+
+	_, err = db.Exec(`INSERT INTO wh_table_uploads (id, wh_upload_id, table_name, status, error, created_at, updated_at)
+		VALUES
+		(1, 1, 'test_table_1','exporting_data', '{}', NOW(), NOW()),
+		(2, 1, 'test_table_2','exporting_data_failed', 'error loading data', NOW(), NOW())
+	`)
+	require.NoError(t, err)
```
achettyiitr, line 788: This has be modified with repository in this https://github.com/rudderlabs/rudder-server/pull/3087

warehouse/upload.go
achettyiitr: Looks like the other 2 flags were not being used. Removing this.

cisse21 (approved):

PR #3051, opened 2023-02-28
fix: user/anonymous id read at gateway

Sidddddarth:
# Description

Losing typecase to avoid missing out on events where `userId`/`anonymousId` is not of type string.

## Notion Ticket

[gateway userId read fix](https://www.notion.so/rudderstacks/gateway-userId-read-fix-6b190ee8f1f24f6089664295ee54174c?pvs=4)

## Security


atzoum (approved):
shouldn't we have some test cases?

lvrach (approved):

atzoum (approved):

PR #3052, opened 2023-02-28
chore: perform rss/wss-aware cgroups memory usage calculation

atzoum:
# Description

Previous calculation of memory usage under cgroups was not taking into consideration RSS & WSS memory, it was relying only to `memory.usage_in_bytes` which can be imprecise under circumstances. [This article](https://mohamedmsaeed.medium.com/memory-working-set-vs-memory-rss-in-kubernetes-which-one-you-should-monitor-8ef77bf0acee) explains how k8s actually calculates RSS & WSS memory respectively. 
With this change, we are now calculating both RSS & WSS values and consider the greatest one to be equal to the actual memory usage.

**_Note:_** verified in `prousmt` environment where this inconsistency was initially observed.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=dfc827ae9677450b9f2a5b8e7ce4a742&pm=s)

## Security

### DIFF
--- a/utils/mem/internal/cgroup/mem.go
+++ b/utils/mem/internal/cgroup/mem.go
@@ -6,20 +6,20 @@ import (
 
 // GetMemoryUsage returns cgroup (v1 or v2) memory usage
 func GetMemoryUsage(basePath string) int64 {
-	n, err := getMemStat(basePath, "memory.usage_in_bytes")
+	n, err := getMemStatCgroup1(basePath, "memory.usage_in_bytes")
 	if err == nil {
-		wss := getWSSMemory(basePath, n)
-		rss := getRSSMemory(basePath)
+		wss := getWSSMemoryCgroup1(basePath, n)
+		rss := getRSSMemoryCgroup1(basePath)
 		if wss > rss {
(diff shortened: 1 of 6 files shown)


utils/mem/internal/cgroup/mem.go
```diff
@@ -63,11 +68,51 @@ func getMemStat(basePath, statName string) (int64, error) {
...
+
+func getWSSMemory(basePath string, used int64) int64 {
+	inactive := memStat(basePath, "total_inactive_file")
+	if used < inactive {
+		return 0
+	}
+	return used - inactive
+}
+
+func getWSSMemoryV2(basePath string, used int64) int64 {
```
fracasula, line 86: Is `v2` a clear enough name in this case? In the article they use something more "evocative" i.e. `cgroups.IsCgroup2UnifiedMode()`.

fracasula (approved):

cisse21 (approved):

PR #3056, opened 2023-03-01
fix: limiter deadlock while trying to notify a dynamic priority item

atzoum:
# Description

In case of high contention, a deadlock was possible if the limiter was trying to notify a dynamic priority item which in turn was trying to increase its priority in the queue.
The limiter now releases the lock earlier, just before notifying the item's channel. The (removed) item will no longer block trying to increase its priority in the queue, it will execute this no-op and then get properly notified.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=feb16db615794204a0c3d9810073f2fb&pm=s)

## Security

### DIFF
--- a/utils/queue/priorityQueue.go
+++ b/utils/queue/priorityQueue.go
@@ -76,6 +76,9 @@ func (pq *PriorityQueue[T]) GetIndex(x interface{}) int {
 
 // Update updates the attributes of an element in the priority queue.
 func (pq *PriorityQueue[T]) Update(item *Item[T], priority int) {
+	if item.index == -1 {
+		return
+	}
 	item.Priority = priority
 	heap.Fix(pq, item.index)
 }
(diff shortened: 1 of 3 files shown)


lvrach:
Can you include parts of the trace that indicate the problem? It would be easier to understand the problem this way

lvrach:
Can you add a benchmark test that simulates a case of high contention?

atzoum:
> Can you add a benchmark test that simulates a case of high contention?

There is a benchmark test already, but didn't really managed to manifest the issue :(

atzoum:
> Can you include parts of the trace that indicate the problem? It would be easier to understand the problem this way

1. Limiter is trying to notify the channel [here](https://github.com/rudderlabs/rudder-server/blob/b7329199fc9e1db4097c6d21376deec8effb2310/utils/sync/limiter.go#L145) (has acquired the lock)
2. The goroutine that should be listening to this channel is trying to acquire a lock [here](https://github.com/rudderlabs/rudder-server/blob/b7329199fc9e1db4097c6d21376deec8effb2310/utils/sync/limiter.go#L183) in order to bump its priority
3. Deadlock!

lvrach (approved):

PR #3066, opened 2023-03-06
chore: use upload_id for staging files

cisse21:
# Description
Use `upload_id` for staging files instead of `start_staging_file_id` and `end_staging_file_id`

## Notion Ticket
[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=5e46904b02e74b4287ee196b27feedc7&pm=s)

## Security


lvrach (approved):

achettyiitr (approved):

saurav-malani (commented):
Given that we are moving to `upload_id`. Do we need `stagingFileStartID` & end id [here](https://github.com/rudderlabs/rudder-server/blob/e7c70ecb9902073bb78539250a5e28adc470025b/warehouse/internal/model/upload.go#L63-L65) & [here](https://github.com/rudderlabs/rudder-server/blob/c1d808ef8ce718f8548cef35a5c178d2dd7c3bf8/warehouse/archive/archiver.go#L218-L219)? I checked they are not used anywhere else.

PR #3076, opened 2023-03-07
chore(warehouse): naming conventions

achettyiitr:
# Description

- Removing suffix `T` from struct types.
- Removing suffix `I`  from interface types.
- Replace map[string]string with warehouseutils.TableSchema
- Replace map[string]map[string]string with warehouseutils.Schema
- Move Schema and TableSchema to the model package. https://github.com/rudderlabs/rudder-server/commit/4e8f72c94ce93abac8247e6e8c6bc644ecdffaba
- [Deepsource changes](https://github.com/rudderlabs/rudder-server/pull/3076/commits/e703268c5e6c699d02a83c4fb79107215f296794)
  1. Possibly return is missed after the http.Error call
  2. Use %q to quote a string in fmt.Sprintf format specifiers
  3. Unnecessary block

## Notion Ticket

https://www.notion.so/rudderstacks/clean-code-3f197f00beac4fe3917385af011c2902?pvs=4

## Security


lvrach (approved):
Could we move tableSchema under the model package?

achettyiitr:
> Could we move tableSchema under the model package?

Yes. This required me to move Warehouse as well into the model's package because of circular dependency.
https://github.com/rudderlabs/rudder-server/pull/3076/commits/4e8f72c94ce93abac8247e6e8c6bc644ecdffaba

cisse21 (approved):

PR #3086, opened 2023-03-09
chore: use gcra rate limiter gateway

cisse21:
# Description

Adapting the gateway rate limiting to the new GCRA algorithm

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/8aac9087df644365acdf64e28e290153?v=6e06b0a5ade24f0aa5ffe05dc2972e84&p=d53b86c7e4244e739b4faf82c205513d&pm=s)

## Security


gateway/throttler/throttler.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+type Throttler struct {
+	limiter Limiter
+	config  throttlingConfig
+}
+
+// checkLimitReached returns true if we're not allowed to process the number of event
+func (t *Throttler) checkLimitReached(key string) (limited bool, retErr error) {
+	ctx := context.TODO()
```
lvrach, line 102: [minor] since this a newly created method, no reason not to expose the interface with a context
```suggestion
func (t *Throttler) checkLimitReached(ctx context.Context, key string) (limited bool, retErr error) {
```

app/apphandlers/gatewayAppHandler.go
```diff
@@ -108,9 +109,10 @@ func (a *gatewayApp) StartRudderCore(ctx context.Context, options *app.Options)
 	})
 
 	var gw gateway.HandleT
-	var rateLimiter ratelimiter.HandleT
-
-	rateLimiter.SetUp()
+	rateLimiter, err := gwThrottler.New(stats.Default)
+	if err != nil {
+		return fmt.Errorf("failed to create rate limiter: %w", err)
+	}
```
lvrach, line 115: Should we have a way to fall back to the legacy throttler? 
What is the risk of introducing GCRA new component on the gateway? 
Could this increase the load on redis?

lvrach (commented):
I want to make sure that we have a fallback mechanism for gateway rate-limiting, either in the gateway adapter or in `internal/throttler` one that would easily allow us to, either revert to an in-memory implementation or stop throttling (noop).

app/apphandlers/embeddedAppHandler.go
```diff
@@ -253,8 +252,10 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 		MultiTenantStat: multitenantStats,
 	}
 
-	rateLimiter := ratelimiter.HandleT{}
-	rateLimiter.SetUp()
+	rateLimiter, err := gwThrottler.New(stats.Default)
+	if err != nil {
+		return fmt.Errorf("failed to create gw rate limiter: %w", err)
```
lvrach, line 257: I like that in case of an error we allow the request to pass, but we might need a metric or setup an alert based on that log.

app/apphandlers/gatewayAppHandler.go
fracasula: I don't think we're introducing the distributed one here, right @cisse21 ?

cisse21: We are not using the distributed one here... We can altogether disable the rate limiter in case we run into any issues. Currently we use this only on hosted and no other customer we have rate limiter enabled

cisse21:
> I want to make sure that we have a fallback mechanism for gateway rate-limiting, either in the gateway adapter or in `internal/throttler` one that would easily allow us to, either revert to an in-memory implementation or stop throttling (noop).

We already do have this flag `Gateway.enableRateLimit` which would disable the rate limit completely but can add a noop implementation in case that is the better approach

lvrach:
> We already do have this flag Gateway.enableRateLimit which would disable the rate limit completely but can add a noop implementation in case that is the better approach

This is enough, just wanted to ensure we have this ability. If noop approach can simplify the code it would be a nice touch.

app/apphandlers/embeddedAppHandler.go
cisse21: Here we aren't passing the requests right? We just return from the app handler and the server would restart. If you meant passing the request in gateway component yes we can set up an alert on the log/add a metric

gateway/throttler/throttler.go
```diff
@@ -0,0 +1,135 @@
...
+const (
+	throttlingAlgoTypeGCRA = "gcra"
+)
+
+type Limiter interface {
+	// Allow returns true if the limit is not exceeded, false otherwise.
+	Allow(ctx context.Context, cost, rate, window int64, key string) (bool, func(context.Context) error, error)
+}
+
+type GetThrottler interface {
```
atzoum, line 25: Why is it called `GetThrottler` instead of `Throttler`?

atzoum (approved):

lvrach (approved):

gateway/throttler/throttler.go
cisse21: Very bad with naming 😕 changing it now

PR #3088, opened 2023-03-09
feat(warehouse): staging file schema consolidation

achettyiitr:
# Description

- Consolidation of staging files schema with the warehouse schema
    - Before
      - For consolidating staging files, we iterate over all the staging files and consolidate them with warehouse schema.
    - After
      - Now we consolidate all the staging files first.
      - Later it is then consolidated with the warehouse schema. 
- Refactoring around the schema package.

## Notion Ticket

https://www.notion.so/rudderstacks/Consolidation-staging-files-schema-846379613cae451dbb5154d5903310cf?pvs=4

## Security


warehouse/errors.go
```diff
@@ -1,19 +1,13 @@
...
 
 	"github.com/rudderlabs/rudder-server/warehouse/integrations/manager"
 	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
 )
 
-var (
-	ErrIncompatibleSchemaConversion = errors.New("incompatible schema conversion")
-	ErrSchemaConversionNotSupported = errors.New("schema conversion not supported")
-)
-
```
achettyiitr, line 16: Since these are internal errors for schema, Moving it to the schema package.

warehouse/identities.go
```diff
@@ -442,24 +442,19 @@ func (wh *HandleT) populateHistoricIdentities(warehouse model.Warehouse) {
...
+		job.schemaHandler.SchemaInWarehouse, job.schemaHandler.UnrecognizedSchemaInWarehouse, err = job.schemaHandler.FetchSchemaFromWarehouse(whManager)
 		if err != nil {
 			pkgLogger.Errorf(`[WH]: Failed fetching schema from warehouse: %v`, err)
 			job.setUploadError(err, model.Aborted)
 			return
 		}
 
-		schemaHandle.SkipDeprecatedColumns(job.schemaHandle.schemaInWarehouse)
-		schemaHandle.SkipDeprecatedColumns(job.schemaHandle.unrecognizedSchemaInWarehouse)
-
```
achettyiitr, line 462: This should not be exposed to the caller whether we are skipping deprecated columns or not. Moved this inside the `FetchSchemaFromWarehouse`

warehouse/schema.go
```diff
@@ -1,381 +0,0 @@
-package warehouse
```
achettyiitr, line 1: This got moved to the schema package.

warehouse/schema/schema.go
```diff
@@ -0,0 +1,391 @@
...
+					logfield.Namespace, sh.Warehouse.Namespace,
+					logfield.TableName, tableName,
+					logfield.ColumnName, columnName,
+				)
+				delete(schema[tableName], columnName)
+				continue
+			}
+		}
+	}
+}
```
achettyiitr, line 175: No changes here.

warehouse/schema/schema.go
```diff
@@ -0,0 +1,391 @@
...
+			diff.UpdatedSchema[columnName] = columnType
+			diff.Exists = true
+		} else if columnType == "text" && currentTableSchema[columnName] == "string" {
+			diff.AlteredColumnMap[columnName] = columnType
+			diff.UpdatedSchema[columnName] = columnType
+			diff.Exists = true
+		}
+	}
+	return diff
+}
```
achettyiitr, line 391: No changes here.

lvrach (approved):

warehouse/errors.go
```diff
@@ -9,11 +9,7 @@ import (
 	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
 )
 
-var (
-	ErrIncompatibleSchemaConversion = errors.New("incompatible schema conversion")
-	ErrSchemaConversionNotSupported = errors.New("schema conversion not supported")
-	ErrCancellingStatement          = errors.New("[error] pq: canceling statement due to user request")
-)
+var ErrCancellingStatement = errors.New("[error] pq: canceling statement due to user request")
```
fracasula, line 12: I cannot find who is returning this error. Can you help me out here? @achettyiitr 

The only place where I see that this error is being used is this one:

```go
		uploadJobsToProcess, err := wh.getUploadsToProcess(ctx, availableWorkers, inProgressNamespaces)

		switch err {
		case nil:
		case context.Canceled, context.DeadlineExceeded, ErrCancellingStatement:
			break loop
```

fracasula (changes requested):
This is returning an error for me, can you please check it out? I think the `fetchSchemaFromWarehouse` method takes the `fetchSchemaRepo` interface:

```go
type fetchSchemaRepo interface {
	FetchSchema(warehouse model.Warehouse) (model.Schema, model.Schema, error)
}
```

Which as the `warehouse` parameter as the only parameter. The problem is that `whManager` is of type `Manager` which has:

```go
type Manager interface {
	FetchSchema() (model.Schema, model.Schema, error)
```

No parameter there.

warehouse/identities.go
```diff
@@ -442,24 +440,13 @@ func (wh *HandleT) populateHistoricIdentities(warehouse model.Warehouse) {
...
-		job.schemaHandle.schemaInWarehouse, job.schemaHandle.unrecognizedSchemaInWarehouse, err = whManager.FetchSchema()
+		err = job.schemaHandle.fetchSchemaFromWarehouse(whManager)
 		if err != nil {
 			pkgLogger.Errorf(`[WH]: Failed fetching schema from warehouse: %v`, err)
 			job.setUploadError(err, model.Aborted)
 			return
 		}
 
-		schemaHandle.SkipDeprecatedColumns(job.schemaHandle.schemaInWarehouse)
-		schemaHandle.SkipDeprecatedColumns(job.schemaHandle.unrecognizedSchemaInWarehouse)
```
fracasula, line 461: How come we removed these? What were they doing exactly? Can you explain this a bit?

warehouse/internal/repo/staging.go
```diff
@@ -251,22 +251,42 @@ func (repo *StagingFiles) GetByID(ctx context.Context, ID int64) (model.StagingF
...
+	schemas := make([]model.Schema, 0, len(ids))
+
+	for rows.Next() {
+		var (
+			rawSchema jsonstd.RawMessage
+			schema    model.Schema
+		)
+
+		if err := rows.Scan(&rawSchema); err != nil {
+			return nil, fmt.Errorf("scanning row: %w", err)
```
fracasula, line 273: We might be missing some context here, perhaps it makes sense to prepend the schema function in order to ease debug when coming across these errors:
```suggestion
			return nil, fmt.Errorf("cannot get schemas by ids: scanning row: %w", err)
```

fracasula: The same would apply to other errors like the ones below e.g. `fmt.Errorf("iterating rows: %w", err)` etc...

warehouse/internal/repo/staging_test.go
```diff
@@ -231,6 +226,47 @@ func TestStagingFileRepo_Many(t *testing.T) {
 			})
 		}
 	})
+
+	t.Run("GetSchemasByIDs", func(t *testing.T) {
+		t.Parallel()
```
fracasula, line 231: [out of curiousity] Do we really need this `t.Parallel()` given that all the sub tests have one? :thinking:

warehouse/schema.go
```diff
@@ -3,119 +3,145 @@ package warehouse
...
+	_, err := sh.schemaRepo.Insert(context.TODO(), &model.WHSchema{
+		UploadID:        uploadId,
+		SourceID:        sh.warehouse.Source.ID,
+		Namespace:       sh.warehouse.Namespace,
+		DestinationID:   sh.warehouse.Destination.ID,
+		DestinationType: sh.warehouse.Type,
+		Schema:          updatedSchema,
+	})
+	if err != nil {
+		return fmt.Errorf("updating schema: %w", err)
```
fracasula, line 87:
```suggestion
		return fmt.Errorf("updating local schema: %w", err)
```

warehouse/identities.go
achettyiitr: We are calling this internally instead of exposing it.
https://github.com/rudderlabs/rudder-server/blob/d1f967f05d6f738bf93fdcf1df3f8bafacd5f06d/warehouse/schema.go#L130

warehouse/internal/repo/staging_test.go
achettyiitr: It's not required here. Removing it.

warehouse/errors.go
achettyiitr: I think we added this to avoid panic. https://github.com/rudderlabs/rudder-server/pull/3105

warehouse/internal/repo/staging_test.go
```diff
@@ -166,6 +159,8 @@ func manyStagingFiles(size int, now time.Time) []*model.StagingFile {
 }
 
 func TestStagingFileRepo_Many(t *testing.T) {
+	t.Parallel()
```
achettyiitr, line 162: deepsource: should call t.Parallel on the top level as well as its subtests

warehouse/internal/repo/staging.go
```diff
@@ -251,22 +251,43 @@ func (repo *StagingFiles) GetByID(ctx context.Context, ID int64) (model.StagingF
 	return *entries[0], err
 }
 
-// GetSchemaByID returns staging file schema field the given ID.
-func (repo *StagingFiles) GetSchemaByID(ctx context.Context, ID int64) (jsonstd.RawMessage, error) {
-	query := `SELECT schema FROM ` + stagingTableName + ` WHERE id = $1`
+// GetSchemasByIDs returns staging file schemas for the given IDs.
+func (repo *StagingFiles) GetSchemasByIDs(ctx context.Context, ids []int64) ([]model.Schema, error) {
+	time.Sleep(time.Second * 5)
```
fracasula, line 256: How come you added this? What behaviour were you trying to test?

```suggestion
```

achettyiitr: I thought I removed this. I was trying to verify user cancel the request.

warehouse/internal/repo/staging.go
```diff
@@ -251,22 +251,43 @@ func (repo *StagingFiles) GetByID(ctx context.Context, ID int64) (model.StagingF
...
+		var (
+			rawSchema jsonstd.RawMessage
+			schema    model.Schema
+		)
+
+		if err := rows.Scan(&rawSchema); err != nil {
+			return nil, fmt.Errorf("cannot get schemas by ids: scanning row: %w", err)
+		}
+		if err := json.Unmarshal(rawSchema, &schema); err != nil {
+			return nil, fmt.Errorf("cannot get schemas by ids: unmarshal staging schema: %w", err)
```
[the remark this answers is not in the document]
achettyiitr: But this would require invalid JSON to be present on the schema column in DB.

warehouse/schema_test.go
```diff
@@ -272,440 +276,170 @@ func TestHandleSchemaChange(t *testing.T) {
 	}
```
fracasula, line 276: Commenting here but my question is about line 274 meaning `require.Equal(t, convError, tc.convError)`.

Wouldn't it be semantically more correct to use `require.ErrorIs` so that the code will keep working if the error gets wrapped as well?

warehouse/schema.go
```diff
@@ -3,119 +3,145 @@ package warehouse
...
+		log:                              logger.NewLogger().Child("warehouse").Child("schema"),
+		stagingFilesSchemaPaginationSize: conf.GetInt("Warehouse.stagingFilesSchemaPaginationSize", 100),
+		skipDeepEqualSchemas:             conf.GetBool("Warehouse.skipDeepEqualSchemas", false),
+		enableIDResolution:               conf.GetBool("Warehouse.enableIDResolution", false),
 	}
+}
 
-	return newColumnVal, err
+func (sh *Schema) updateLocalSchema(uploadId int64, updatedSchema model.Schema) error {
+	_, err := sh.schemaRepo.Insert(context.TODO(), &model.WHSchema{
```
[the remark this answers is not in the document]
achettyiitr: Yes.

warehouse/internal/repo/staging.go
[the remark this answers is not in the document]
achettyiitr: Will add a test case for this.

warehouse/errors.go
achettyiitr: is it ok if I try to take this as a separate task? 
https://www.notion.so/rudderstacks/GetUploadToProcess-error-handling-04fe991f6a90426195f8a9db78fd6e08?pvs=4

fracasula (approved):

warehouse/errors.go
achettyiitr: Right.

PR #3100, opened 2023-03-14
fix(warehouse): snowflake default timestamp to timestamp with time zone

achettyiitr:
# Description

- Snowflake default timestamp to timestamp with time zone

## Notion Ticket

https://www.notion.so/rudderstacks/Snowflake-default-timestamp-to-timestamp_tz-69f60cef979841a5b3fbcdc805d7ce95?pvs=4

## Security

### DIFF
--- a/warehouse/integrations/snowflake/snowflake.go
+++ b/warehouse/integrations/snowflake/snowflake.go
@@ -49,7 +49,7 @@ var dataTypesMap = map[string]string{
 	"bigint":   "number",
 	"float":    "double precision",
 	"string":   "varchar",
-	"datetime": "timestamp",
+	"datetime": "timestamp_tz",
 	"json":     "variant",
 }
 


cisse21 (approved):

Jayachand (approved):

lvrach:
Which time zone we are going to use in this case? 

Is the time zone of SDK or the server ?

achettyiitr:
> Which time zone we are going to use in this case?
> 
> Is the time zone of SDK or the server ?

Server time zone (UTC). From the [transformer](https://github.com/rudderlabs/rudder-transformer/blob/1f88222f8a5768b6d0236b833f9d0211e9a5fea1/src/warehouse/index.js#L95), we will always get UTC.

achettyiitr:
Although it doesn't matter the weather we use TIMESTAMP or TIMESTAMP_TZ because from the transformer we will always get UTC. 
Just for Snowflake partner connect we are making it to TIMESTAMP_TZ

PR #3103, opened 2023-03-15
chore: start server in normal mode when there is no instruction by scheduler

BonapartePC:
# Description

Server/Processor should start in normal mode if there is no mode set by the scheduler.
Note: The server starts in degraded mode and only goes to normal mode when there is no instruction from the scheduler via ETCD so far.

## Notion Ticket

https://www.notion.so/rudderstacks/Server-should-start-in-normal-mode-by-default-d6f155d9375e4ca6a7fa25412c93a4e6

## Security


lvrach (commented):
It would be safer if scheduler send/set the state to normal if the state is missing, instead of rudder server starting with it by default.

cisse21:
@lvrach the way scheduler works now it would feel out of place and a bit hacky to have this logic present so instead we can always start the server in normal mode by default rather than degraded mode which would be a switch in default behaviour and nothing more right?

BonapartePC:
> It does feel weird that RS starts in normal though, because the scheduler is the one orchestrating the sagas via ETCD. If RS is supposed to be in DEGRADED due to a Saga initiated by the scheduler but then RS crashes, it will start in a wrong state and it might accept events that it wasn't supposed to accept/process, right?

RS always saves its state in etcd. Even if the RS crashes, it'll always start in the same state. With this PR, our only change is new servers will start in `NORMAL` mode when we scale up rather than `DEGRADED` mode.

fracasula:
> RS always saves its state in etcd. Even if the RS crashes, it'll always start in the same state. With this PR, our only change is new servers (that doesn't have any instruction from scheduler so far) will start in `NORMAL` mode when we scale up rather than `DEGRADED` mode.

Do we have a test to ensure that interrupted sagas are not affected by this?

BonapartePC:
> Do we have a test to ensure that interrupted sagas are not affected by this?

Added a test for that as well

BonapartePC:
Note: The server starts in degraded mode and only goes to normal mode when there is no instruction from the scheduler via ETCD so far.

fracasula (approved):

PR #3108, opened 2023-03-16
chore: dedup stat size comparison

achettyiitr:
# Description

Enhance our dedup logic:
- Store the size of the event we dedup along with its id
- When a duplicate is detected, retrieve the original message size and compare it with the existing message size.
- When increasing the duplicate metric, use an additional label `size_cmp` that will be true or false, according to the previous comparison.

## Notion Ticket

https://www.notion.so/rudderstacks/Enhance-Dedup-metric-with-size-comparison-64a3dc3943f7434e9fba6ed33e1feaa6?pvs=4

## Security


services/dedup/dedup.go
```diff
@@ -192,59 +202,67 @@ func (d *DedupHandleT) writeToBadger(messageIDs []string) error {
...
 	// Dedup with badgerDB
 	err := d.badgerDB.View(func(txn *badger.Txn) error {
-		for idx, messageID := range messageIDs {
-			_, err := txn.Get([]byte(messageID))
-			if err != badger.ErrKeyNotFound {
-				toRemoveMessageIndexesSet[idx] = struct{}{}
+		for idx, info := range infos {
+			item, err := txn.Get([]byte(info.MessageID))
+			if err != nil && err != badger.ErrKeyNotFound {
+				pkgLogger.Warn("[[ Dedup ]] Error while getting item from badgerDB", err)
```
achettyiitr, line 240: Adding a warning log instead of returning it. Right now, we are not aware of the scale at which we are getting this error. Since the error is being handled as panic.

services/dedup/dedup.go
```diff
@@ -192,59 +202,67 @@ func (d *DedupHandleT) writeToBadger(messageIDs []string) error {
...
 			}
+			if err == badger.ErrKeyNotFound {
+				continue
+			}
+
+			var (
+				value   []byte
+				payload Payload
+			)
+			if value, err = item.ValueCopy(nil); err == nil && len(value) > 0 {
```
achettyiitr, line 251: `len(value) > 0` added this for backward compatibility.

services/dedup/dedup.go
```diff
@@ -17,9 +17,18 @@ import (
...
-	FindDuplicates(messageIDs []string, allMessageIDsSet map[string]struct{}) (duplicateIndexes []int)
-	MarkProcessed(messageIDs []string) error
+type Info struct {
+	MessageID string
+	Size      int
+}
+
+type Payload struct {
+	Size int `json:"size"`
+}
```
atzoum, line 27: We shouldn't need a struct for holding the payloadSize, an `int64` would suffice
```suggestion
```
For converting to-from bytes we may use `strconv`, e.g.
```go
value := strconv.FormatInt(payloadSize, 10)
payloadSize, _ = strconv.ParseInt(string(value), 10, 0)
```

services/dedup/dedup.go
```diff
@@ -17,9 +17,18 @@ import (
...
+
+type Dedup interface {
+	FindDuplicates(infos []Info, allMessageIDsSet map[string]Payload) map[int]Payload
```
atzoum, line 30: The method didn't and still doesn't appear provide a clear enough API with respect to  Dedup's responsibilities. More specifically, `allMessageIDsSet` seems a bit out of place. The caller should be responsible for performing this kind of deduplication (along with the deduplication of the infos slice itself), since it has nothing to do with state persisted in dedup's storage itself. The fact that we don't include the payload size in the aforementioned dedup scenarios is indicative of this api design weakness.

services/dedup/dedup.go
achettyiitr: Since the payload stored in badgerDB should be backward compatible, I was storing this payload.

achettyiitr: Not sure if just storing a string would work here.

atzoum: what is stored in badgerDB now? Isn't it `nil`?

achettyiitr: Yeah, right now it is nil. But in the future, if we are going to change the payload format, our current structure would not work?

atzoum: If we change it in the future, we may use a fallback if parsing the value fails

achettyiitr: But wouldn't it better to set the correct expectations from the code instead of handling the error?

atzoum: We are dropping the previous values when badgerDB starts, thus we shouldn't worry too much on the evolution of the key's value format. At this stage I would prefer focusing on simplicity and performance (compute & storage).

services/dedup/dedup.go
```diff
@@ -192,59 +202,67 @@ func (d *DedupHandleT) writeToBadger(messageIDs []string) error {
...
 
 	// Dedup with badgerDB
 	err := d.badgerDB.View(func(txn *badger.Txn) error {
-		for idx, messageID := range messageIDs {
-			_, err := txn.Get([]byte(messageID))
-			if err != badger.ErrKeyNotFound {
-				toRemoveMessageIndexesSet[idx] = struct{}{}
+		for idx, info := range infos {
+			item, err := txn.Get([]byte(info.MessageID))
+			if err != nil && err != badger.ErrKeyNotFound {
```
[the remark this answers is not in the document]
achettyiitr: Yeah, we can return the error from here itself. Since, till now we were skipping the error so I am not sure weather we should panic or not.

services/dedup/dedup.go
achettyiitr: Yeah, this will simplify the dedup logic for the processor. 😄

PR #3109, opened 2023-03-16
fix: minio error while creating bucket if already exists

achettyiitr:
# Description

- Since for every upload, we are trying to create the bucket. In case of the bucket already exists, we try to check if the bucket exists or not using `minioClient.BucketExists` but this is returning a `permission error`.
- We particularly don't need an API call to verify if the bucket exists or not. We can suppress the error from the error code itself when we get a bucket creation error.

## Notion Ticket

https://www.notion.so/rudderstacks/Minio-data-retention-object-storage-validations-is-failing-b1c961c67337476a82566a43eaed243e?pvs=4

## Security


services/filemanager/miniomanager.go
```diff
@@ -36,8 +36,8 @@ func (manager *MinioManager) Upload(ctx context.Context, file *os.File, prefixes
 	defer cancel()
 
 	if err = minioClient.MakeBucket(ctx, manager.Config.Bucket, minio.MakeBucketOptions{Region: "us-east-1"}); err != nil {
-		exists, errBucketExists := minioClient.BucketExists(ctx, manager.Config.Bucket)
-		if !(errBucketExists == nil && exists) {
+		err = suppressMinorErrors(err)
```
lvrach, line 39: [minor] not sure if wrapping error handling in a separate method is really necessary. The errors that we are ignoring here are specific to bucket creation, not sure if the function should be reused elsewhere

lvrach (approved):

cisse21 (approved):

saurav-malani (commented):
Instead of directly trying to create a bucket and then handling the error, why don't we 1st check if the bucket exist or not. And only create if it doesn't exist?

saurav-malani:
> Yeah, But it will still require 1 API call anyways. Why not just use the Create APi and suppress it in case of already exists?
Well, I believe preventing error in first place is better then handling it. 
Also, according to me creating of a bucket is like a `write` operation where as checking if it exists or not is like a `read` operation. And, read would have less overhead in comparison to write in terms of internal implementation.

achettyiitr:
> > Yeah, But it will still require 1 API call anyways. Why not just use the Create APi and suppress it in case of already exists?
> 
> Well, I believe preventing error in first place is better then handling it. Also, according to me creating of a bucket is like a `write` operation where as checking if it exists or not is like a `read` operation. And, read would have less overhead in comparison to write in terms of internal implementation. In short, the overhead of both the api call would be quite different.

Make sense.

achettyiitr:
What if we don't create a bucket at all? Ideally, it should be the customer's responsibility to create the bucket.

saurav-malani:
> What if we don't create a bucket at all? Ideally, it should be the customer's responsibility to create the bucket.

well, I am not sure if we have clearly mentioned it somewhere that they themself need to create a bucket. But, if that is the case then your point make sense. But, still I believe at the end of the it will escalate to some issue when upload would fail and then we will be notified and CS team will get in touch with them to create it. So, I personally think creating on our own if it doesn't exist would be best for us. 😄

achettyiitr:
I have reverted the Azure changes since it requires additional logic to check if the container exists or not using [storage.GetContainerReferenceFromSASURI().Exists()](https://github.com/Azure/azure-sdk-for-g) which requires a SASUri

achettyiitr:
Also, it was not the scope of the PR.

saurav-malani (approved):

PR #3110, opened 2023-03-17
chore: fix namespace bug

cisse21:
# Description
Fixing the bug introduced in the previous refactoring of `getNamespace` function which returned an empty namespace even though there was a valid namespace present in the `schema` table also adding tests for the same

## Notion Ticket
[ Notion Link ](https://www.notion.so/rudderstacks/Fix-getNamespace-bug-a21a74603ef14372bf9aa4d0c51c67b5?pvs=4)

## Security


lvrach:
Please provide more context, what is the bug and the intended fix.

lvrach (approved):

PR #3111, opened 2023-03-17
chore(router): don't include job in metadata during transformation requests

atzoum:
# Description

Job information in metadata is not of any use to transformer, thus excluding it from the request.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/Pipelines-Sprint-68c1c213e9b840b3a9b3826e3631e39a?p=655d85119e6e4b07ac4abe1c8991df3d&pm=s)

## Security

### DIFF
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -104,8 +104,7 @@ func Init() {
 // Transform transforms router jobs to destination jobs
 func (trans *handle) Transform(transformType string, transformMessage *types.TransformMessageT) []types.DestinationJobT {
 	var destinationJobs types.DestinationJobs
-	transformMessageCopy := *transformMessage
-	jobs := transformMessageCopy.Dehydrate()
+	transformMessageCopy, jobs := transformMessage.Dehydrate()
 
 	// Call remote transformation
 	rawJSON, err := jsonfast.Marshal(&transformMessageCopy)
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -106,7 +106,6 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
 	var destinationJobs types.DestinationJobs
 	transformMessageCopy := *transformMessage
 	jobs := transformMessageCopy.Dehydrate()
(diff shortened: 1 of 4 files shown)


router/transformer/transformer.go
```diff
@@ -103,8 +103,13 @@ func Init() {
 
 // Transform transforms router jobs to destination jobs
 func (trans *handle) Transform(transformType string, transformMessage *types.TransformMessageT) []types.DestinationJobT {
+	var destinationJobs types.DestinationJobs
+	transformMessageCopy := *transformMessage
```
lvrach, line 107: Is there a reason doing the shallow copy here? 🤔

During hydration we manipulate non root level fields, thus I don't see the benefit.

atzoum: Just don't want to mutate input parameters of this method. As a principle it is safer not to mutate input parameters, because you never know if and how the owner (caller) can use them later on. If a method is mutating input parameters, this introduces strong coupling & should be explicitly mentioned in the method's doc. But even so, a caller might not read the doc thus it can cause issues/bugs.

atzoum: As a sidenote, the struct that we are dehydrating is not the same as the struct we are hydrating

router/transformer/transformer.go
```diff
@@ -241,6 +245,7 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
 	}
 	func() { httputil.CloseResponse(resp) }()
 
+	destinationJobs.Hydrate(jobs)
```
cisse21, line 248: Why do we need to hydrate the jobs twice?

atzoum: we don't

cisse21 (approved):

router/transformer/transformer.go
atzoum: **_Update:_** you are correct, a shallow copy isn't enough, thus adapted the Dehydrate method to return a proper copy of the `TransformMessageT`

utsabc (approved):

PR #3118, opened 2023-03-20
feat: embed avro schema id

fracasula:
# Description

Pull request to enable customers (i.e. CARS) to optionally have the schemaID embedded in the message payload.

This is compliant with the way Confluent libraries do it but it would break the workflow of existing customers using the Avro functionality (three at the moment). That is the reason why I decided to have it behind an optional flag (i.e. `embedAvroSchemaID bool`).

I've added a new `TestAvroSchemaRegistry` where I'm testing both ways of sending Avro messages. One with the schemaID not embedded (existing customers) and one with the schemaID embedded (for CARS).

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/Serial-schema-id-into-Avro-message-fc15909ef535427d9cc1e639515231fe) >

## Security


services/streammanager/kafka/kafkamanager.go
```diff
@@ -329,7 +332,15 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (*Produ
 	if err != nil {
 		return nil, err
 	}
-	return &ProducerManager{p: p, timeout: o.Timeout, codecs: codecs}, nil
+
+	// @TODO embedAvroSchemaID should come from control plane (i.e. destination config)
```
atzoum, line 336: Shouldn't we add config-based support for this already & let it be overridden through this environment variable for the time being? That way, as soon as control plane is ready with their changes, the only thing we'll need to do is to remove the override.

atzoum (approved):

services/streammanager/kafka/kafkamanager.go
fracasula: The configuration needs to be reworked anyway and if I understand correctly the whole thing will be picked up by @saurav-malani. He'll make it proper don't worry :+1:

PR #3129, opened 2023-03-24
chore: deltalake parquet load files support

achettyiitr:
# Description

Added support for parquet load files for delta lake destination

## Notion Ticket

https://www.notion.so/rudderstacks/Deltalake-parquet-support-1465ea1f5a064932b063636cebd8eb57?pvs=4

## Security


warehouse/upload.go
```diff
@@ -1642,6 +1642,7 @@ func (job *UploadJob) setUploadError(statusError error, state string) (string, e
 	}
 
 	serializedErr, _ := json.Marshal(&uploadErrors)
+	serializedErr = warehouseutils.SanitizeJson(serializedErr)
```
achettyiitr, line 1645: We were getting this error while updating:
`unable to change upload columns: pq: unsupported Unicode escape sequence`

PR #3134, opened 2023-03-28
fix: roleBasedAuth parameter wasn't being set correctly

raman2805:
# Description

Setting the sessionConfig.RoleConfigAuth parameter before throwing an error for incompatible configs

### DIFF
--- a/utils/awsutils/session.go
+++ b/utils/awsutils/session.go
@@ -106,12 +106,12 @@ func NewSimpleSessionConfig(config map[string]interface{}, serviceName string) (
 		return nil, fmt.Errorf("unable to populate session config using destinationConfig: %w", err)
 	}
 
-	if sessionConfig.RoleBasedAuth && sessionConfig.IAMRoleARN == "" {
-		return nil, errors.New("incompatible role configuration")
-	}
-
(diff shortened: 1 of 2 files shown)


utils/awsutils/session.go
```diff
@@ -106,14 +106,14 @@ func NewSimpleSessionConfig(config map[string]interface{}, serviceName string) (
...
-		return nil, errors.New("incompatible role configuration")
-	}
-
 	if !isRoleBasedAuthFieldExist(config) {
 		sessionConfig.RoleBasedAuth = sessionConfig.IAMRoleARN != ""
 	}
 
+	if sessionConfig.RoleBasedAuth && sessionConfig.IAMRoleARN == "" {
+		return nil, errors.New("incompatible role configuration")
+	}
```
koladilip, line 115:
```suggestion
	if  sessionConfig.IAMRoleARN == "" {
		sessionConfig.RoleBasedAuth = false
	}
```

koladilip (changes requested):

koladilip (approved):

achettyiitr (approved):

PR #3145, opened 2023-03-30
chore: update staging files upload id

cisse21:
# Description

Update all the existing staging file columns with a valid upload id

## Notion Ticket

[ Notion Link ](https://www.notion.so/rudderstacks/script-Populate-upload_id-for-older-warehouse-jobs-7318c98481e94be099ee1a38d10c13f8?pvs=4)

## Security


cmd/populateUserID/populateID.go
lvrach, line 1: Maybe we can add it here https://github.com/rudderlabs/rudder-warehouse-tools, I don't expect this script to be run many times.

cisse21: will move it there and get rid of this PR

PR #3146, opened 2023-03-30
feat: schema forwarder

cisse21:
# Description
This PR introduces schema-forwarder which will act as a bridge between event schema jobsDB and the external pulsar cluster

## Notion Ticket

[ Notion Link ](https://www.notion.so/rudderstacks/jobsdb-forwarder-30d8716a0f484f2b91254be1671751b3?pvs=4)

## Security


app/cluster/dynamic.go
```diff
@@ -54,6 +56,7 @@ type Dynamic struct {
 
 	currentMode         servermode.Mode
 	currentWorkspaceIDs string
+	JobsForwarder       jobs_forwarder.Forwarder
```
atzoum, line 59:
```suggestion
	JobsForwarder       lifecycle
```

atzoum, line 59: should also move this up in the list of fields, above the non-exported ones

internal/pulsar/pulsar.go
```diff
@@ -0,0 +1,137 @@
...
+func NewClient(config *config.Config) (Client, error) {
+	log := logger.NewLogger().Child("pulsar")
+	client, err := newPulsarClient(getClientConf(config), log)
+	if err != nil {
+		return Client{}, fmt.Errorf("error creating pulsar client : %v", err)
+	}
+	return client, nil
+}
+
+func (c *Client) NewProducer(config *config.Config) (ProducerAdapter, error) {
```
atzoum, line 49: It is possible to have a single pulsar client (using the same pulsar cluster for different use cases). But for every use case we are going to need to use different topics and/or producer/consumer configurations.

E.g. for event schemas we want our producer to batch using `BatcherBuilderType: pulsar.KeyBasedBatchBuilder`. This might not stand true for another producer operating against a different topic with different requirements.

Thus, I believe that NewProducer should accept `pulsar.ProducerOptions` instead of `config.Config`.

jobs-forwarder/internal/forwarder/baseforwarder.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/rudderlabs/rudder-server/jobsdb"
+	"github.com/rudderlabs/rudder-server/utils/misc"
+)
+
+type BaseForwarder struct {
+	ctx        context.Context
+	g          *errgroup.Group
+	jobsDB     jobsdb.JobsDB
+	log        logger.Logger
+	config     *config.Config
```
atzoum, line 22: Do we really need this?
```suggestion
```

jobs-forwarder/internal/forwarder/baseforwarder.go
```diff
@@ -0,0 +1,107 @@
...
+	bf.jobsDB = schemaDB
+	bf.ctx = ctx
+	bf.g = g
+}
+
+func (bf *BaseForwarder) GetJobs(ctx context.Context) ([]*jobsdb.JobT, bool, error) {
+	if bf.jobsDB == nil {
+		bf.log.Errorf("base forwarder: jobsDB is nil")
+		return []*jobsdb.JobT{}, true, nil
+	}
```
atzoum, line 47: I believe that `jobsDB` should never be nil, let it panic
```suggestion
```

jobs-forwarder/internal/forwarder/baseforwarder.go
```diff
@@ -0,0 +1,107 @@
...
+	combinedList = append(combinedList, unprocessed.Jobs...)
+	return combinedList, unprocessed.LimitsReached, nil
+}
+
+func (bf *BaseForwarder) MarkJobStatuses(ctx context.Context, statusList []*jobsdb.JobStatusT) error {
+	if bf.jobsDB == nil {
+		bf.log.Errorf("base forwarder: jobsDB is nil")
+		return nil
+	}
```
atzoum, line 81:
```suggestion
```

jobs-forwarder/internal/forwarder/baseforwarder.go
```diff
@@ -0,0 +1,107 @@
...
+	unprocessed, err := misc.QueryWithRetriesAndNotify(ctx, bf.baseConfig.jobsDBQueryRequestTimeout, bf.baseConfig.jobsDBMaxRetries, func(ctx context.Context) (jobsdb.JobsResult, error) {
+		return bf.jobsDB.GetUnprocessed(ctx, queryParams)
+	}, bf.sendQueryRetryStats)
+	if err != nil {
+		bf.log.Errorf("base forwarder: Error while reading unprocessed from DB: %v", err)
+		return nil, false, err
+	}
+	combinedList = append(combinedList, unprocessed.Jobs...)
+	return combinedList, unprocessed.LimitsReached, nil
```
atzoum, line 74: shouldn't we sort by jobID ascending?

jobs-forwarder/internal/forwarder/baseforwarder.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (bf *BaseForwarder) sendQueryRetryStats(attempt int) {
+	bf.log.Warnf("Timeout during query jobs in processor module, attempt %d", attempt)
+	stats.Default.NewTaggedStat("jobsdb_query_timeout", stats.CountType, stats.Tags{"attempt": fmt.Sprint(attempt), "module": "jobs_forwarder"}).Count(1)
+}
+
+func (bf *BaseForwarder) generateQueryParams() jobsdb.GetQueryParamsT {
+	return jobsdb.GetQueryParamsT{
+		EventsLimit: bf.baseConfig.pickupSize,
+		JobsLimit:   bf.baseConfig.pickupSize,
```
atzoum, line 105: what about the message byte size limit?

jobs-forwarder/internal/forwarder/jobsforwarder.go
```diff
@@ -0,0 +1,150 @@
...
+							JobState:      jobsdb.Failed.State,
+							ExecTime:      time.Now(),
+							RetryTime:     time.Now(),
+							ErrorCode:     "500",
+							Parameters:    []byte{},
+							ErrorResponse: json.RawMessage(err.Error()),
+						})
+						continue
+					}
+					statusFunc := func(_ pulsarType.MessageID, _ *pulsarType.ProducerMessage, err error) {
```
atzoum, line 74: We should retry failed messages, with some retries and backoff, until all messages of the batch have been sent. After failing all the retries we can abort and then pick the next batch. The retry period though should be really big, e.g. a few hours.

It is not expected for messages to fail at this stage, only if there is a technical issue (network, pulsar unavailability etc.)

jobs-forwarder/internal/forwarder/baseforwarder.go
cisse21: Don't we do that by default in the query? Also as we have decided that we would only store terminal states in the database we can skip fetching failed jobs and only fetch unprocessed ones

atzoum (approved):

BonapartePC (approved):

PR #3148, opened 2023-03-30
feat: databricks compatible go driver

achettyiitr:
# Description

- Using databricks [Go driver](https://github.com/databricks/databricks-sql-go) for running SQL queries.
- Use `Warehouse.deltalake.useNative` to pick between current and legacy implementation.
- Deepsource (Inappropriate usage of t.Parallel())

## Notion Ticket

https://www.notion.so/rudderstacks/Use-go-driver-for-databricks-7279a67897c04fa9b4debaef625c93f8?pvs=4

## Security


lvrach:
I was thinking of a feature flag for this change, are we confident enough about do driver?

fracasula:
> I was thinking of a feature flag for this change, are we confident enough about go driver?

The last "serious" thing they fixed was back in January and it was a race condition. Overall it seems maintained and it has e2e tests. Nonetheless it's a small project and it doesn't seem used by many so :+1: for the feature flag.

cisse21:
@lvrach Let's check how many customers have this destination and if there aren't too many we can validate by using this branch on the customer one by one

achettyiitr:
@lvrach wdyt?
Once the testing is completed, We can deploy it for acorns since they are already having a smaller version of the changes. Monitor for a week and then we can merge this PR.

warehouse/integrations/deltalake/deltalake.go
```diff
@@ -31,24 +31,28 @@ const (
...
 	provider       = warehouseutils.DELTALAKE
 	tableNameLimit = 127
 )
 
-// Reference: https://docs.oracle.com/cd/E17952_01/connector-odbc-en/connector-odbc-reference-errorcodes.html
 const (
-	tableOrViewNotFound = "42S02"
-	columnNotFound      = "42000"
-	databaseNotFound    = "42000"
-	partitionNotFound   = "42000"
```
fracasula, line 46: What happened to `tableOrViewNotFound` for example? We don't detect this error anymore?

Also, is there a more robust way to detect these errors like by error code like it was before? It would be nice to have an integration test for this and see what happens if the table is not there for example.

warehouse/integrations/deltalake/deltalake.go
```diff
@@ -146,210 +149,202 @@ func NewDeltalake() *Deltalake {
...
-		return
-	}
-	if connectionResponse.GetErrorCode() != "" {
-		err = fmt.Errorf("error connecting to Delta lake with response: %s", connectionResponse.GetErrorMessage())
-		return
-	}
+	query = fmt.Sprintf(`SHOW tables from %s;`, d.Namespace)
+
+	defer func() {
+		d.Logger.Infow("fetching tables",
```
fracasula, line 208: It seems a bit too much to have these as INFO. I think it makes more sense to have them as DEBUG if you really need them and then have a metric for the query execution time.

fracasula: The same applies to similar defers in this code.

warehouse/integrations/deltalake/deltalake.go
```diff
@@ -500,164 +552,346 @@ func getTableSchemaDiff(tableSchemaInUpload, tableSchemaAfterUpload model.TableS
...
+			logfield.DestinationID, d.Warehouse.Destination.ID,
+			logfield.DestinationType, d.Warehouse.Destination.DestinationDefinition.Name,
+			logfield.WorkspaceID, d.Warehouse.WorkspaceID,
+			logfield.Namespace, d.Namespace,
+			logfield.TableName, tableName,
+			logfield.StagingTableName, stagingTableName,
+			logfield.Error, err.Error(),
+		)
+
+		return "", fmt.Errorf("creating staging table: %w", err)
```
fracasula, line 588: How about logging upstream only as ERROR instead of warning? Is the logger really necessary here? If yes then I guess it should be ERROR. Wdyt?

warehouse/integrations/deltalake/deltalake_test.go
fracasula, line 1: How are we supposed to test this now?

fracasula (changes requested):
I'm a bit concerned about the lack of tests. Can we discuss strategies to test these changes?

Sidenote: I also see an excessive (in my opinion) usage of logs at INFO level. It would be nice to see less logs (which also cause an infra cost due to having them ingested on Loki if I'm not mistaken) and see more metrics.

warehouse/integrations/deltalake/deltalake_test.go
achettyiitr: Tests that I added earlier were using mocked GRPC client. Since now we don't have GRPC, we need to rely on Integration tests.

fracasula: > we need to rely on Integration tests.

What integration tests? I don't see any.

warehouse/integrations/deltalake/deltalake.go
fracasula: That was done [last week](https://github.com/databricks/databricks-sql-go/pull/117). It might be worth chasing them to see if they manage to review it since it might break things for us. It would be a really nice addition.

warehouse/integrations/deltalake/deltalake_test.go
achettyiitr: https://github.com/rudderlabs/rudder-server/blob/8f74bbb55fdb2d172352e2dbe54c9f64e8915edd/Makefile#L106

achettyiitr:
> I'm a bit concerned about the lack of tests. Can we discuss strategies to test these changes?
> 
> Sidenote: I also see an excessive (in my opinion) usage of logs at INFO level. It would be nice to see less logs (which also cause an infra cost due to having them ingested on Loki if I'm not mistaken) and see more metrics.

There are some concerns raised by @lvrach around using metrics. Since we are using histograms, we might not see queries fitting into a bucket or only a few queries.

fracasula:
> > I'm a bit concerned about the lack of tests. Can we discuss strategies to test these changes?
> > Sidenote: I also see an excessive (in my opinion) usage of logs at INFO level. It would be nice to see less logs (which also cause an infra cost due to having them ingested on Loki if I'm not mistaken) and see more metrics.
> 
> There are some concerns raised by @lvrach around using metrics. Since we are using histograms, we might not see queries fitting into a bucket or only a few queries.

True although it's always possible to define buckets per metric. Maybe we can consider that.

achettyiitr:
> > > I'm a bit concerned about the lack of tests. Can we discuss strategies to test these changes?
> > > Sidenote: I also see an excessive (in my opinion) usage of logs at INFO level. It would be nice to see less logs (which also cause an infra cost due to having them ingested on Loki if I'm not mistaken) and see more metrics.
> > 
> > 
> > There are some concerns raised by @lvrach around using metrics. Since we are using histograms, we might not see queries fitting into a bucket or only a few queries.
> 
> True although it's always possible to define buckets per metric. Maybe we can consider that.

Setted up a call today to discuss and finalize things here.

warehouse/integrations/deltalake/deltalake_test.go
fracasula: Right, you pushed them afterwards, I couldn't see them at the time, thanks. I'll have a look :+1:

achettyiitr: Integration tests have been there for a long time. Recently just added the legacy implementation back.

warehouse/integrations/deltalake/deltalake.go
achettyiitr: Added slow query logs based on execution time.

go.mod
```diff
@@ -96,6 +96,17 @@ require (
...
+require (
+	github.com/apache/arrow/go/v11 v11.0.0-20221220213742-09453f9258f3 // indirect
+	github.com/databricks/databricks-sql-go v1.1.0
+	github.com/dnephin/pflag v1.0.7 // indirect
+	github.com/fatih/color v1.13.0 // indirect
+	github.com/mattn/go-colorable v0.1.12 // indirect
+	github.com/mattn/go-isatty v0.0.16 // indirect
+	github.com/rs/zerolog v1.28.0 // indirect
+	gotest.tools/gotestsum v1.8.2 // indirect
+)
```
saurav-malani, line 108: let's put this inside the previous `require`.

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1517 @@
...
+		if !strings.HasPrefix(tableName, warehouseutils.StagingTablePrefix(provider)) {
+			continue
+		}
+		filteredTablesNames = append(filteredTablesNames, tableName)
+	}
+
+	d.dropStagingTables(filteredTablesNames)
+}
+
+func Connect(cred Credentials) (*sql.DB, error) {
```
saurav-malani, line 959: I don't see `Connect` being used anywhere else outside of this pkg apart from test. So, won't it make more sense to create a separate test file for non-exported func and then test `Connect` there instead?

achettyiitr: We are calling this from deltalake_native_test.

saurav-malani: yup, but that is a test... right? Exporting something just for the sake of testing it, sounds weird to me. Alternative to this, can't we move tests of all unexported functions to maybe a separate file which is in `datalake_native` pkg itself.

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1367 @@
...
+				logfield.DestinationID, d.Warehouse.Destination.ID,
+				logfield.DestinationType, d.Warehouse.Destination.DestinationDefinition.Name,
+				logfield.WorkspaceID, d.Warehouse.WorkspaceID,
+				logfield.Namespace, d.Namespace,
+				logfield.StagingTableName, stagingTable,
+				logfield.Error, err.Error(),
+			)
+		}
+	}
+}
```
saurav-malani, line 374: Won't it be better to create a single `Drop` query with all tables that are needed to be dropped, instead of doing it this way?

achettyiitr: Databricks don't support running drop table query with multiple [tables](https://docs.databricks.com/sql/language-manual/sql-ref-syntax-ddl-drop-table.html).

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1367 @@
...
+		return ""
+	}
+	if externalLocation == "" {
+		return ""
+	}
+	return fmt.Sprintf("LOCATION '%s/%s/%s'", externalLocation, d.Namespace, tableName)
+}
+
+func (d *Deltalake) dropDanglingStagingTables() {
+	tableNames, _ := d.fetchTables("^rudder_staging_.*$")
```
saurav-malani, line 920: isn't dangerous to ignore error here. Since, there is a possibility that there is nothing in `tableNames` because some error occured while fetching the tables names. This will led to the accumulation of those tables. I believe we should atleast log it.

achettyiitr: Good point. let me add error logs here.

achettyiitr: Although, we are dropping staging tables after every sync. But there might be a case where we didn't have DROP permissions and since we are skipping these errors, these tables will accumulate.

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1367 @@
...
+	if err == sql.ErrNoRows {
+		return false, nil
+	}
+	if err != nil {
+		return false, fmt.Errorf("schema exists: %w", err)
+	}
+
+	exists := schema == schemaName
+	return exists, nil
+}
```
saurav-malani, line 329:
```suggestion
func (d *Deltalake) schemaExists(schemaName string) (bool, error) {
	var schema string
	if err := d.DB.QueryRow(fmt.Sprintf(`SHOW SCHEMAS LIKE '%s';`, schemaName)).Scan(&schema); err != nil {
		if err == sql.ErrNoRows {
			return false, nil
		}
		return false, fmt.Errorf("schema exists: %w", err)
	}
	return schema == schemaName, nil
}
```

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1367 @@
...
+		SELECT COUNT(*) FROM %[1]s.%[2]s;
+	`,
+		d.Namespace,
+		tableName,
+	)
+
+	err = d.DB.QueryRowContext(ctx, query).Scan(&total)
+
+	if err != nil {
+		if strings.Contains(err.Error(), schemaNotFound) {
```
saurav-malani, line 1259: I believe we should try adding tests in all such situations to make sure that the error returned is indeed what we expected.

achettyiitr: Right now we are waiting for error types to get added to the databricks SQL go driver.
https://github.com/databricks/databricks-sql-go/pull/117

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1356 @@
...
+	}
+	return warehouseutils.JoinWithFormatting(keys, format, ",")
+}
+
+func columnNames(columns []string) string {
+	return strings.Join(columns, ",")
+}
+
+func stagingColumnNames(keys []string) string {
+	format := func(idx int, str string) string {
```
pChondros, line 178: [question] why do we need the idx variable here?

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1356 @@
...
+		return fmt.Sprintf(`STAGING.%s`, str)
+	}
+	return warehouseutils.JoinWithFormatting(keys, format, ",")
+}
+
+func columnsWithValues(keys []string) string {
+	format := func(idx int, str string) string {
```
pChondros, line 185: [question] why do we need the idx variable here?

warehouse/integrations/deltalake-native/deltalake_native.go
fracasula: +1 for not exporting functions that are not meant to be used from outside. in my opinion not exporting unnecessarily symbols limits the API surface that we expose, reducing cognitive load and easing software maintenance. tests in go can use unexported symbols, there is no need to change the package name to have a test being called "integration test".

runner/runner.go
```diff
@@ -365,6 +367,7 @@ func runAllInit() {
 	redshift.Init()
 	snowflake.Init()
 	deltalake.Init()
+	deltalake_native.Init()
```
fracasula, line 370: It'd be a good idea to start moving away from this pattern and stop having to call `Init` functions of packages which is known to create issues due to the use of global variables and due to the fact that we saw cases where the order of calling `Init` packages created unexpected outcomes.

In this case it's simple enough to fix given that in the `Init` you just create a logger. Perhaps for now we can move the logger creation in the constructor. That would allow us to remove both the global variable `pkgLogger` and the `Init` function.

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1356 @@
...
+	Catalog                = "catalog"
+	UseSTSTokens           = "useSTSTokens"
+	EnableExternalLocation = "enableExternalLocation"
+	ExternalLocation       = "externalLocation"
+	UserAgent              = "Rudderstack"
+)
+
+const (
+	provider       = warehouseutils.DELTALAKE
+	tableNameLimit = 127
```
fracasula, line 42: Can we please have a comment as to what this number represents and how we came up with `127`?

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1356 @@
...
+	"date":      "date",
+	"timestamp": "datetime",
+}
+
+// excludeColumnsMap Columns you need to exclude
+// Since event_date is an auto generated column in order to support partitioning.
+// We need to ignore it during query generation.
+var excludeColumnsMap = map[string]bool{
+	"event_date": true,
+}
```
fracasula, line 100: [minor optimization] we don't really need a boolean here. using a `struct{}` instead should have a [smaller footprint](https://stackoverflow.com/a/22770744/828366):

```suggestion
var excludeColumnsMap = map[string]struct{}{
	"event_date": {},
}
```

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1356 @@
...
+
+var errorsMappings = []model.JobError{
+	{
+		Type:   model.PermissionError,
+		Format: regexp.MustCompile(`UnauthorizedAccessException: PERMISSION_DENIED: User does not have READ FILES on External Location`),
+	},
+	{
+		Type:   model.PermissionError,
+		Format: regexp.MustCompile(`SecurityException: User does not have permission CREATE on CATALOG`),
```
fracasula, line 115: It's a bit scary that we rely on this very long regular expressions to map errors. I see that in the tests we only have assertions saying `Noop` error or `Unknown` error, but there isn't one with `PermissionError`, right?

My questions:
1. can these be tested somehow?
2. if we map an error incorrectly, what is the side effect?

achettyiitr: We are only testing whether we can map the error or not, We are not testing whether it can get mapped to the correct mapping.

warehouse/integrations/deltalake-native/deltalake_native.go
achettyiitr: Will take this as a separate task to do this.

fracasula: Great, thanks @achettyiitr, please link the notion task here for posterity :+1:

achettyiitr: https://www.notion.so/rudderstacks/For-error-mappings-test-that-errors-are-mapped-with-their-corresponding-mapping-0d2740508c2b4eeba80d39a47f20a1f9?pvs=4

achettyiitr: https://www.notion.so/rudderstacks/Export-functions-for-warehouse-1e1b9177cd3a42c4a2d18fcc2d73f138?pvs=4

warehouse/integrations/deltalake-native/deltalake_native_test.go
```diff
@@ -0,0 +1,388 @@
...
+
+func TestIntegrationDeltalake(t *testing.T) {
+	if os.Getenv("SLOW") == "0" {
+		t.Skip("Skipping tests. Remove 'SLOW=0' env var to run them.")
+	}
+	if _, exists := os.LookupEnv(testhelper.DeltalakeIntegrationTestCredentials); !exists {
+		t.Skipf("Skipping %s as %s is not set", t.Name(), testhelper.DeltalakeIntegrationTestCredentials)
+	}
+
+	t.Parallel()
```
fracasula, line 37: You might get data races on `pkgLogger` here. Perhaps it is time to remove `pkgLogger` after all :slightly_smiling_face:

warehouse/integrations/deltalake-native/deltalake_native_test.go
```diff
@@ -0,0 +1,388 @@
...
+		tc := tc
+
+		t.Cleanup(func() {
+			require.NoError(
+				t,
+				testhelper.WithConstantBackoff(func() (err error) {
+					_, err = db.Exec(fmt.Sprintf(`DROP SCHEMA %[1]s CASCADE;`, tc.schema))
+					return
+				}),
+				fmt.Sprintf("Failed dropping schema %s for Deltalake", tc.schema),
```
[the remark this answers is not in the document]
fracasula: If you were to use `require.Eventually` instead of `require.NoError`+`WithConstantBackoff` it would be clearer to the reader to how long you're going to try for and how often. Plus you'd be using one function only.

warehouse/integrations/deltalake-native/deltalake_native.go
lvrach: It has to do with the generic `warehouseutils.JoinWithFormatting`, which requires a function with (index, str)

We could do:
```suggestion
	format := func(_ int, str string) string {
```

warehouse/integrations/deltalake-native/deltalake_native.go
```diff
@@ -0,0 +1,1356 @@
...
+		tableSchemaDiff   = getTableSchemaDiff(tableSchemaInUpload, tableSchemaAfterUpload)
+		sortedColumnNames = d.sortedColumnNames(tableSchemaInUpload, sortedColumnKeys, tableSchemaDiff)
+
+		query          string
+		partitionQuery string
+	)
+
+	if d.Uploader.GetLoadFileType() == warehouseutils.LOAD_FILE_TYPE_PARQUET {
+		query = fmt.Sprintf(`
+			COPY INTO %v
```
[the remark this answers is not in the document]
achettyiitr: Thanks for pointing that out. Updated other occurrences as well.

warehouse/integrations/deltalake-native/deltalake_native.go
achettyiitr: Adding comment.

warehouse/integrations/deltalake-native/deltalake_native.go
achettyiitr: Added warning logs for this.

warehouse/integrations/deltalake/deltalake.go
achettyiitr: https://github.com/rudderlabs/rudder-server/pull/3190

warehouse/integrations/deltalake-native/deltalake_native_test.go
achettyiitr: https://github.com/rudderlabs/rudder-server/pull/3202

runner/runner.go
achettyiitr: https://github.com/rudderlabs/rudder-server/pull/3202

warehouse/integrations/bigquery/bigquery_test.go
```diff
@@ -53,11 +54,15 @@ func TestIntegrationBigQuery(t *testing.T) {
...
-				fmt.Sprintf("Failed dropping dataset %s for BigQuery", dataset),
+			require.Eventually(t, func() bool {
+				if err := db.Dataset(dataset).DeleteWithContents(context.TODO()); err != nil {
+					t.Logf("dropping dataset %s: %s", dataset, err.Error())
+					return false
+				}
+				return true
+			},
+				1*time.Minute,
+				1*time.Second,
```
fracasula, line 65: [optional] We could retry more often and potentially shorten the test time. If with a shorter interval the log above bothers you it can be removed or the test can be run withouth `-v`.

Same thing on the other `Eventually` below.

fracasula (approved):

warehouse/integrations/bigquery/bigquery_test.go
achettyiitr: I was biased with using `require.Eventually`. Let me use a simple for loop for retries.

achettyiitr: There is no need for Backoff and Eventually. In case of failures, we just need to retry a few of times.

achettyiitr:
Thanks, @fracasula for the detailed feedback.

warehouse/integrations/deltalake-native/deltalake_native.go
achettyiitr: This has been fixed in the current version.

PR #3150, opened 2023-03-30
chore: updating error mappings

achettyiitr:
# Description

Update error mappings

## Notion Ticket

https://www.notion.so/rudderstacks/redshift-error-mappings-for-column-count-eafaff8afa1e4fa896e2deca9d42effa?pvs=4

## Security

### DIFF
--- a/warehouse/integrations/postgres/postgres.go
+++ b/warehouse/integrations/postgres/postgres.go
@@ -47,7 +47,7 @@ var errorsMappings = []model.JobError{
 		Format: regexp.MustCompile(`dial tcp: lookup .*: no such host`),
 	},
 	{
-		Type:   model.ResourceNotFoundError,
+		Type:   model.PermissionError,
 		Format: regexp.MustCompile(`dial tcp .*: connect: connection timed out`),
 	},
 	{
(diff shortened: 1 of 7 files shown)


warehouse/integrations/postgres/postgres.go
```diff
@@ -47,7 +47,7 @@ var errorsMappings = []model.JobError{
 		Format: regexp.MustCompile(`dial tcp: lookup .*: no such host`),
 	},
 	{
-		Type:   model.ResourceNotFoundError,
+		Type:   model.PermissionError,
 		Format: regexp.MustCompile(`dial tcp .*: connect: connection timed out`),
```
lvrach: The timeout could be for various reasons right?

achettyiitr: Yes.

achettyiitr: I was thinking of this database's firewall being one of the reasons.

achettyiitr: Also, since we don't have set connection limit on our end, it must be causing due to the connection limit set on the database side right.

lvrach: It could also be due to other network conditions as well, misconfiguration can be the most common. If after backoff retries we still get it, then the probabilities of miss-configuration are high (permission or incorrect DSN).

achettyiitr: Let me remove it.

cisse21 (approved):

PR #3151, opened 2023-03-30
chore: kafka batching

fracasula:
# Description

Making sure that batching on Kafka works as expected.

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/Increase-Kafka-batching-usage-432e3f51e50941189bb9be4981acd4da) >

## Security

### DIFF
--- a/integration_test/kafka_batching/kafka_batching_test.go
+++ b/integration_test/kafka_batching/kafka_batching_test.go
@@ -220,7 +220,7 @@ func TestKafkaBatching(t *testing.T) {
 		if err != nil {
 			return false
 		}
-		defer httputil.CloseResponse(resp)
+		defer func() { _ = resp.Body.Close() }()
 
 		var buf []byte
 		buf, err = io.ReadAll(resp.Body)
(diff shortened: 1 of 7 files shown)


lvrach (approved):

integration_test/kafka_batching/kafka_batching_test.go
```diff
@@ -0,0 +1,478 @@
+package kafka_batching
```
atzoum, line 1: **Question:** does batching work in combination with **AVRO Schema Lists** and/or **multiple topic support**? This is not evident by looking [here](https://github.com/rudderlabs/rudder-transformer/blob/develop/src/v0/destinations/kafka/transform.js), where it seems that the `process` method performs more steps than its `batch` counterpart. Maybe `batch` is called with events that have already been `process`ed in a previous step?

@utsabc @koladilip

atzoum: I think I can answer that question myself... events are first transformed in processor, then batched in router!

atzoum (approved):

PR #3152, opened 2023-04-03
fix: add externalID(workspace-id) to S3 file uploader config

Sidddddarth:
# Description

adds `externalID`(workspaceID) to S3 fileuploader config.
The `externalID` was spelled wrong([ref](https://github.com/rudderlabs/rudder-server/blob/master/utils/misc/misc.go#L931))
Also it is a required for IAM role.
Proc_error job status:
```
{
  "Error": "error starting S3 session: externalID is required for IAM role"
}
```
Going with what's being done at batch-router as jobs uploaded to destination with same credentials succeeds but dumps being uploaded to the same bucket fails.

## Notion Ticket

[data-retention role based auth](https://www.notion.so/rudderstacks/testing-data-retention-changes-0aeba1ea3a5d4c7c80b31ec55fb3fedb?pvs=4)

## Security

### DIFF
--- a/services/fileuploader/fileuploader.go
+++ b/services/fileuploader/fileuploader.go
@@ -173,8 +173,8 @@ func overrideWithSettings(defaultConfig map[string]interface{}, settings backend
 	for k, v := range settings.Config {
 		config[k] = v
 	}
-	if _, ok := config["externalId"]; ok && settings.Type == "S3" {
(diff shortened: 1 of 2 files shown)


saurav-malani (approved):
As long as we are sure that `externalID` is used by S3 and not `externalId`, it looks good to me.

PR #3153, opened 2023-04-03
fix: add externalID to S3 file uploader config

Sidddddarth:
# Description
adds externalID(workspaceID) to S3 fileuploader config.
The externalID was [spelled wrong](https://github.com/rudderlabs/rudder-server/blob/master/utils/awsutils/session.go#L26)([correct usage](https://github.com/rudderlabs/rudder-server/blob/master/utils/misc/misc.go#L931))
Also it is a required for IAM role.
Proc_error job status:
```
{
  "Error": "error starting S3 session: externalID is required for IAM role"
}
```
Going with what's being done at batch-router as jobs uploaded to destination with same credentials succeeds but dumps being uploaded to the same bucket fails.

## Notion Ticket
[data-retention role based auth](https://www.notion.so/rudderstacks/testing-data-retention-changes-0aeba1ea3a5d4c7c80b31ec55fb3fedb?pvs=4)

## Security

### DIFF
--- a/services/fileuploader/fileuploader.go
+++ b/services/fileuploader/fileuploader.go
@@ -173,8 +173,8 @@ func overrideWithSettings(defaultConfig map[string]interface{}, settings backend
 	for k, v := range settings.Config {
 		config[k] = v
 	}
-	if _, ok := config["externalId"]; ok && settings.Type == "S3" {
-		config["externalId"] = workspaceID
+	if settings.Type == "S3" && config["iamRoleArn"] != nil {
+		config["externalID"] = workspaceID
 	}
 	return backendconfig.StorageBucket{
 		Type:   settings.Type,
--- a/services/filemanager/filemanager.go
+++ b/services/filemanager/filemanager.go
@@ -101,7 +101,7 @@ func GetProviderConfigFromEnv(ctx context.Context, provider string) map[string]i
 		providerConfig["iamRoleArn"] = config.GetString("BACKUP_IAM_ROLE_ARN", "")
 		if providerConfig["iamRoleArn"] != "" {
 			backendconfig.DefaultBackendConfig.WaitForConfig(ctx)
-			providerConfig["externalId"] = backendconfig.DefaultBackendConfig.Identity().ID()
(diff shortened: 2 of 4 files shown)


services/fileuploader/fileuploader.go
```diff
@@ -173,8 +173,8 @@ func overrideWithSettings(defaultConfig map[string]interface{}, settings backend
 	for k, v := range settings.Config {
 		config[k] = v
 	}
-	if _, ok := config["externalId"]; ok && settings.Type == "S3" {
-		config["externalId"] = workspaceID
+	if settings.Type == "S3" {
```
atzoum, line 176: Shouldn't we be doing something similar to that, i.e. check for the presence of `iamRoleArn` config property?
https://github.com/rudderlabs/rudder-server/blob/master/services/filemanager/filemanager.go#L102-L105

services/fileuploader/fileuploader.go
```diff
@@ -173,8 +173,8 @@ func overrideWithSettings(defaultConfig map[string]interface{}, settings backend
...
+		config["externalID"] = workspaceID
```
atzoum, line 177: is it `externalID` or `externalId`?

Sidddddarth: The reference that I've gone with is what we do [here](https://github.com/rudderlabs/rudder-server/blob/master/utils/misc/misc.go#L931)
same for the above comment as well, just went ahead without checking for `iamRoleArn` because it's not being done in above reference.
I wasn't sure of what's required, but was sure that the batchrouter succeeds, so simply just went ahead.

atzoum: it is confusing that there are occurrences where we are using `externalID` and others where we are using `externalId`... Are both correct in their own context, or something needs fixing? @koladilip

koladilip: `externalID` is used in session util which is the one that generates role credentials using externalID. Not sure if externalId works

koladilip (approved):

atzoum (approved):

PR #3156, opened 2023-04-04
fix: archiver staging files pickup

achettyiitr:
# Description

Use `upload_id` filter while getting staging files for a particular upload instead of range.

## Notion Ticket

https://www.notion.so/rudderstacks/warehouse-archiving-staging-files-pickup-b77d508545fc4c879ce8a01695210462?pvs=4

## Security


saurav-malani (commented):
I see `_ = txn.Rollback()` , `stagingFileRows.Close()` & `loadLocationRows.Close()` quite frequently in the code. Since, we want to make sure that in case of error we handle them properly. But, still I believe it leaves room for introduction of bug. So, instead of expecting the coder to do it mindfully can't we divide these things into functions and then `defer` rollback() and close().

achettyiitr:
> Also, I little more context on why we are doing this would be really helpful. 😄

While picking staging files, earlier we were using range(start_staging_file_id, end_staging_file_id). Later, we introduced upload_id in the wh_staging_files table to pickup.
Archival should also use wh_upload_id to pick the staging files.

achettyiitr:
> I see `_ = txn.Rollback()` , `stagingFileRows.Close()` & `loadLocationRows.Close()` quite frequently in the code. Since, we want to make sure that in case of error we handle them properly. But, still I believe it leaves room for introduction of bug. So, instead of expecting the coder to do it mindfully can't we divide these things into functions and then `defer` rollback() and close(). eg for ref:
> 
> https://github.com/rudderlabs/rudder-server/blob/c1d808ef8ce718f8548cef35a5c178d2dd7c3bf8/jobsdb/jobsdb.go#L1722-L1736
> 
> Also, I little more context on why we are doing this would be really helpful. 😄

Good point. For the time being, I have reverted the rollback changes and only kept the fix. I will add a separate pr for the rollback changes.

saurav-malani (commented):
Given that we are moving to `upload_id`. Do we need `stagingFileStartID` & end id [here](https://github.com/rudderlabs/rudder-server/blob/e7c70ecb9902073bb78539250a5e28adc470025b/warehouse/internal/model/upload.go#L63-L65) & [here](https://github.com/rudderlabs/rudder-server/blob/c1d808ef8ce718f8548cef35a5c178d2dd7c3bf8/warehouse/archive/archiver.go#L218-L219)? I checked they are not used anywhere else.

achettyiitr:
> stagingFileStartID

It is still being used in at a couple of places. We haven't yet completely migrated.
https://github.com/rudderlabs/rudder-server/blob/3ec2433b671428a7f5504af8e9cb85b2f4373c51/warehouse/internal/repo/staging.go#L405

achettyiitr:
From @lvrach , 
We need to migrate all the instances where start and end staging file ids at all pieces.

PR #3163, opened 2023-04-04
fix: illegal job order with maxRetriesLimitReached

atzoum:
# Description

Out-of-order delivery was possible when `maxRetriesLimitReached` happened for a job with **batch transformation enabled** or when **transformation happened at the router (implicit batching)**: 

_A failed job that reached its max retries' limit was ultimately marked as `aborted` instead of `failed`, but jobs in the same batch for the same order key were marked as `waiting`. This caused other later jobs, already in the worker's buffer, for the same order key to be processed out-of-order, i.e. before the `waiting` jobs._

This fix, moves the process of marking such jobs as aborted to the next loop, at the beginning of the worker's process, in the same step where job draining is happening.

A side-effect of this change is that some metrics are affected too, i.e. instead of the `router_aborted_events` counter, we are now incrementing the `drained_events` counter using as reason `retry limit reached`.

**_Bonus:_** renamed `workerT` struct to `worker` and renamed `worker` method receiver to `w`.
## Notion Ticket

[Link](https://www.notion.so/rudderstacks/Pipelines-Sprint-68c1c213e9b840b3a9b3826e3631e39a?p=8ccdaf7684f44efeb7890b4867679d5c&pm=s)

## Security

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -85,7 +85,7 @@ type HandleT struct {
 	MultitenantI                            tenantStats
 	destName                                string
 	destinationId                           string
-	workers                                 []*workerT
+	workers                                 []*worker
 	telemetry                               *DiagnosticT
 	customDestinationManager                customDestinationManager.DestinationManager
 	throttlingCosts                         atomic.Pointer[types.EventTypeThrottlingCost]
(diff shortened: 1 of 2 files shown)


fracasula (approved):
Nice :+1:

koladilip (approved):

PR #3166, opened 2023-04-05
chore: snowflake slow query logs

achettyiitr:
# Description

Added slow query logs for snowflake

## Notion Ticket

https://www.notion.so/rudderstacks/snowflake-slow-query-logs-bf7f83c8e5604bd7be03a7dc61a1b814?pvs=4

## Security


saurav-malani (commented):
Since, for each query we want to make sure that we get a log if the query time is more than `x`. So, won't a cleaner way of doing this would be to wrap these db query calls with a timer and log it then an there if the query time is more. 
eg: `func QueryLogger(query func()) {
    start := time.Now()
    // code to make the query
    duration := time.Since(start)
    if duration > x {
        logger.Printf("Query1 took %s", duration)
    }
}`

achettyiitr:
> Since, for each query we want to make sure that we get a log if the query time is more than `x`. So, won't a cleaner way of doing this would be to wrap these db query calls with a timer and log it then an there if the query time is more. eg:
> 
> ```
>  func QueryLogger(query func()) {
>     start := time.Now()    
>     // code to make the query
>     duration := time.Since(start)
>     if duration > x {
>         logger.Printf("Query1 took %s", duration)
>     }
> }
> ```

Good point. Let me try this.

PR #3175, opened 2023-04-05
feat: support for incremental updates while fetching backend config

raman2805:
# Description

We have added support to fetch incremental updates from the control-plane while getting the workspace configs for a particular workspace. The way it works is when we first fetch the config from the control-plane, the config corresponding to every workspace has the **updateAt** field set. The control-plane api also supports a parameter called **updatedAfter** , which if passed, is expected to return only those workspaces which have seen a change. The semantics of the behavior is such that we keep track of the last value of the **updateAt** field in the various workspaces and pass that in subsequent request to only get the updates which are then applied on top of the previously maintained state of the configs. 

## Notion Ticket

https://www.notion.so/rudderstacks/use-test-updatedAt-for-backend-config-1701f037ed644f36b2a8fe37b143a0e2?pvs=4

### DIFF
--- a/backend-config/backend-config.go
+++ b/backend-config/backend-config.go
@@ -42,11 +42,12 @@ var (
 	LastRegulationSync string
 
 	// DefaultBackendConfig will be initialized be Setup to either a WorkspaceConfig or MultiWorkspaceConfig.
-	DefaultBackendConfig BackendConfig
-	pkgLogger            = logger.NewLogger().Child("backend-config")
-	IoUtil               = sysUtils.NewIoUtil()
-	Diagnostics          diagnostics.DiagnosticsI
-	cacheOverride        cache.Cache
+	DefaultBackendConfig        BackendConfig
+	pkgLogger                   = logger.NewLogger().Child("backend-config")
+	IoUtil                      = sysUtils.NewIoUtil()
+	Diagnostics                 diagnostics.DiagnosticsI
+	cacheOverride               cache.Cache
+	useIncrementalConfigUpdates bool
 )
 
 func disableCache() {
@@ -98,6 +99,7 @@ func loadConfig() {
 	config.RegisterBoolConfigVariable(false, &configFromFile, false, "BackendConfig.configFromFile")
 	config.RegisterIntConfigVariable(1000, &maxRegulationsPerRequest, true, 1, "BackendConfig.maxRegulationsPerRequest")
 	config.RegisterBoolConfigVariable(true, &configEnvReplacementEnabled, false, "BackendConfig.envReplacementEnabled")
+	config.RegisterBoolConfigVariable(false, &useIncrementalConfigUpdates, false, "BackendConfig.useIncrementalConfigUpdates")
 	config.RegisterBoolConfigVariable(true, &dbCacheEnabled, false, "BackendConfig.dbCacheEnabled")
 }
 
@@ -284,10 +286,11 @@ func newForDeployment(deploymentType deployment.Type, region string, configEnvHa
 		}
 	case deployment.MultiTenantType:
 		backendConfig.workspaceConfig = &namespaceConfig{
-			configBackendURL: parsedConfigBackendURL,
-			configEnvHandler: configEnvHandler,
-			cpRouterURL:      cpRouterURL,
-			region:           region,
+			configBackendURL:            parsedConfigBackendURL,
+			configEnvHandler:            configEnvHandler,
+			cpRouterURL:                 cpRouterURL,
+			region:                      region,
+			useIncrementalConfigUpdates: useIncrementalConfigUpdates,
 		}
 	default:
 		return nil, fmt.Errorf("deployment type %q not supported", deploymentType)
--- a/backend-config/backend-config.go
+++ b/backend-config/backend-config.go
@@ -42,12 +42,12 @@ var (
 	LastRegulationSync string
 
 	// DefaultBackendConfig will be initialized be Setup to either a WorkspaceConfig or MultiWorkspaceConfig.
-	DefaultBackendConfig        BackendConfig
-	pkgLogger                   = logger.NewLogger().Child("backend-config")
-	IoUtil                      = sysUtils.NewIoUtil()
-	Diagnostics                 diagnostics.DiagnosticsI
-	cacheOverride               cache.Cache
-	useIncrementalConfigUpdates bool
+	DefaultBackendConfig     BackendConfig
+	pkgLogger                = logger.NewLogger().Child("backend-config")
+	IoUtil                   = sysUtils.NewIoUtil()
+	Diagnostics              diagnostics.DiagnosticsI
+	cacheOverride            cache.Cache
+	incrementalConfigUpdates bool
 )
 
 func disableCache() {
@@ -99,7 +99,7 @@ func loadConfig() {
 	config.RegisterBoolConfigVariable(false, &configFromFile, false, "BackendConfig.configFromFile")
 	config.RegisterIntConfigVariable(1000, &maxRegulationsPerRequest, true, 1, "BackendConfig.maxRegulationsPerRequest")
 	config.RegisterBoolConfigVariable(true, &configEnvReplacementEnabled, false, "BackendConfig.envReplacementEnabled")
-	config.RegisterBoolConfigVariable(false, &useIncrementalConfigUpdates, false, "BackendConfig.useIncrementalConfigUpdates")
+	config.RegisterBoolConfigVariable(false, &incrementalConfigUpdates, false, "BackendConfig.incrementalConfigUpdates")
 	config.RegisterBoolConfigVariable(true, &dbCacheEnabled, false, "BackendConfig.dbCacheEnabled")
 }
 
@@ -286,11 +286,11 @@ func newForDeployment(deploymentType deployment.Type, region string, configEnvHa
 		}
 	case deployment.MultiTenantType:
 		backendConfig.workspaceConfig = &namespaceConfig{
-			configBackendURL:            parsedConfigBackendURL,
-			configEnvHandler:            configEnvHandler,
-			cpRouterURL:                 cpRouterURL,
-			region:                      region,
-			useIncrementalConfigUpdates: useIncrementalConfigUpdates,
+			configBackendURL:         parsedConfigBackendURL,
+			configEnvHandler:         configEnvHandler,
+			cpRouterURL:              cpRouterURL,
+			region:                   region,
+			incrementalConfigUpdates: incrementalConfigUpdates,
 		}
 	default:
 		return nil, fmt.Errorf("deployment type %q not supported", deploymentType)
--- a/backend-config/backend_config_test.go
+++ b/backend-config/backend_config_test.go
@@ -584,14 +584,6 @@ type mockIdentifier struct {
 	token string
 }
 
-func (m *mockIdentifier) ID() string {
-	return m.key
-}
-
-func (m *mockIdentifier) BasicAuth() (string, string) {
-	return m.token, ""
-}
-
-func (*mockIdentifier) Type() deployment.Type {
-	return deployment.Type(`mockType`)
-}
+func (m *mockIdentifier) ID() string                  { return m.key }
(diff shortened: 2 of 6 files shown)


backend-config/backend_config_test.go
```diff
@@ -16,7 +16,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/golang/mock/gomock"
```
Sidddddarth, line 19: lint failing here

backend-config/namespace_config.go
```diff
@@ -80,10 +86,19 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
 	if nc.namespace == "" {
 		return config, fmt.Errorf("namespace is not configured")
 	}
+	useUpdateAfter := !(nc.lastUpdatedAt.IsZero()) && nc.useIncrementalConfigUpdates
+	fmt.Println(nc.lastUpdatedAt.IsZero())
```
Sidddddarth, line 90:
```suggestion
```

backend-config/backend-config.go
```diff
@@ -42,11 +42,12 @@ var (
...
-	pkgLogger            = logger.NewLogger().Child("backend-config")
-	IoUtil               = sysUtils.NewIoUtil()
-	Diagnostics          diagnostics.DiagnosticsI
-	cacheOverride        cache.Cache
+	DefaultBackendConfig        BackendConfig
+	pkgLogger                   = logger.NewLogger().Child("backend-config")
+	IoUtil                      = sysUtils.NewIoUtil()
+	Diagnostics                 diagnostics.DiagnosticsI
+	cacheOverride               cache.Cache
+	useIncrementalConfigUpdates bool
```
Sidddddarth, line 50: benefit to have it be a part of `backendConfigImpl` instead of a global..?

raman2805: I guess for the sake of consistency it makes sense to have it here (all other variables are being declared here)

Sidddddarth: fair enough

Sidddddarth (approved):

backend-config/namespace_config_test.go
```diff
@@ -162,7 +206,19 @@ func (server *backendConfigServer) ServeHTTP(resp http.ResponseWriter, req *http
...
 	}
-
+	values := req.URL.Query()
+	for k, v := range values {
+		if k == "updatedAfter" {
+			_, err := time.Parse(updateAfterTimeFormat, v[0])
+			if err != nil {
+				resp.WriteHeader(http.StatusBadRequest)
+				_, _ = resp.Write(([]byte(`{"message":"invalid param for updatedAfter"}`)))
+			}
```
lvrach, line 216:
```suggestion
	_, err := time.Parse( req.URL.Query().Get("updatedAfter"))
				if err != nil {
				resp.WriteHeader(http.StatusBadRequest)
				_, _ = resp.Write(([]byte(`{"message":"invalid param for updatedAfter"}`)))
			}
```

backend-config/namespace_config_test.go
lvrach, line 1: How are we testing that the correct updateAt timestamp is being request?

backend-config/namespace_config.go
```diff
@@ -80,10 +86,18 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
 	if nc.namespace == "" {
 		return config, fmt.Errorf("namespace is not configured")
 	}
+	useUpdateAfter := !(nc.lastUpdatedAt.IsZero()) && nc.useIncrementalConfigUpdates
```
fracasula, line 89: No need to call `IsZero()` if `nc.useIncrementalConfigUpdates` is `false`:

```suggestion
	useUpdateAfter := nc.useIncrementalConfigUpdates && !nc.lastUpdatedAt.IsZero()
```

backend-config/namespace_config.go
```diff
@@ -105,6 +119,10 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
 		respBody = configEnvHandler.ReplaceConfigWithEnvVariables(respBody)
 	}
 
+	if !useUpdateAfter {
+		nc.workspacesConfig = make(map[string]ConfigT)
```
fracasula, line 123: This allocation would be unnecessary in case we return earlier due to an unmarshal error on 127.

We should move it right before the last `for` loop.

One question though, why do we initialize this map only if `!useUpdateAfter`? Shouldn't it simply be `if nc.workspaceConfig == nil {`?

backend-config/namespace_config.go
```diff
@@ -80,10 +86,18 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
 	if nc.namespace == "" {
```
fracasula, line 86: We should be able to delete the line above, right? i.e. `config := make(map[string]ConfigT)`.

backend-config/namespace_config_test.go
raman2805: This change the semantics a bit. We want to handle 3 scenarios
1. updateAfter is not given
2. updateAfter is given but the value is not parseable as time where we want to throw an exception 
3. updateAfter is given and is parseable as time. 

req.URL.Query().Get(..) does not throw an error. With the change you proposed, we will not be able to distinguish between 1&2

backend-config/namespace_config.go
```diff
@@ -80,10 +86,18 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
...
 	u := *nc.configBackendURL
-	u.Path = fmt.Sprintf("/data-plane/v1/namespaces/%s/config", nc.namespace)
+	if useUpdateAfter {
+		u.Path = fmt.Sprintf("/data-plane/v1/namespaces/%s/config", nc.namespace)
+		values := u.Query()
+		values.Add("updatedAfter", nc.lastUpdatedAt.Format(updateAfterTimeFormat))
+		u.RawQuery = values.Encode()
+	} else {
+		u.Path = fmt.Sprintf("/data-plane/v1/namespaces/%s/config", nc.namespace)
+	}
```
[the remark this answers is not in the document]
raman2805: Done!

backend-config/namespace_config_test.go
raman2805: removed redundant parenthesis

backend-config/namespace_config_test.go
raman2805: Added some test for updateAt , please have a look

backend-config/namespace_config.go
raman2805: (1) Yes, changed the location.
(2) No, we can't do that because if useUpdateAfter is false, we should be rebuilding the entire map (to keep the semantics of not using useUpdateAfter exactly the same as before

backend-config/namespace_config.go
raman2805: It seems it won't work because the upstream code which uses this call actually ends up using the return value even if there is an error. It might require make some changes there as well if we were to change this

backend-config/namespace_config.go
fracasula: I see. Clever although I find it confusing because you're relying on `IsZero()` above. It's not obvious at a first glance.

Personally I'd find it more immediate if the map is initialized in the constructor (i.e. `SetUp`) as it's quite expected/common thing to do. Then you could just change the `if !useUpdateAfter` into:

```go
if !nc.useIncrementalConfigUpdates {
	nc.workspacesConfig = make(map[string]ConfigT)
}
```

Not mandatory of course, just a suggestion to make the code easier to understand.

backend-config/namespace_config.go
fracasula: That's unfortunate. I suggest we rename `config` to `configOnError` to at least make clear that the map isn't really meant for anything but it's just a "zero" value that we return in case of error.

backend-config/namespace_config.go
```diff
@@ -112,14 +124,21 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
...
+	if !useUpdateAfter {
+		nc.workspacesConfig = make(map[string]ConfigT)
+	}
+
 	for workspaceID, wc := range workspacesConfig {
 		// always set connection flags to true for hosted and multi-tenant warehouse service
 		wc.ConnectionFlags.URL = nc.cpRouterURL
 		wc.ConnectionFlags.Services = map[string]bool{"warehouse": true}
-		workspacesConfig[workspaceID] = wc
+		nc.workspacesConfig[workspaceID] = wc
```
fracasula, line 135: One question. What if one of the incremental updates something gets deleted? I see that you only add to the map but never delete. I was expecting something like another loop that iterates over `nc.workspacesConfig` and deletes the entry if not present in the new `workspacesConfig`. Would control plane send us such updates?

raman2805: Yeah, thats an interesting question. Let me check with them about this

backend-config/namespace_config.go
raman2805: Done

backend-config/namespace_config.go
raman2805: Done!

backend-config/namespace_config.go
fracasula: I just had a chat with @fxenik and @ktgowtham and we agreed on the following structure:

Screenshot from 2023-04-24 09-56-02

This means that workspaces with `null` will have to be skipped since they don't have any relevant changes for us. Then we'll loop over `nc.workspacesConfig` instead and if we don't have a workspaceID at all in the received response we delete it from `nc.workspacesConfig`.

**NOTE:** This PR should be on hold until they are done with the aforementioned changes.

backend-config/namespace_config.go
```diff
@@ -109,17 +121,32 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
...
+		if wc.UpdatedAt.After(nc.lastUpdatedAt) {
+			nc.lastUpdatedAt = wc.UpdatedAt
+		}
+	}
+	for _, workspaceID := range deletedWorkspaces {
+		delete(nc.workspacesConfig, workspaceID)
 	}
```
fracasula, line 147: Are you sure this is going to work? I was expecting something like:

```go
for workspaceID := range nc.workspacesConfig {
    if _, ok := workspacesConfig[workspaceID]; !ok {
        // delete!
    }
}
```

raman2805: Yeah, I checked this (both before making the change) , and through the unit tests. Basically, if there is a key, value pair in the json with the value as **null** it gets parsed and the corresponding struct gets default initialized. 

See this snippet on go playground https://go.dev/play/p/y5OawhHI02u

fracasula: Hold on, I was referring to the third case.

1. the ID is there and the value is a valid object
2. the ID is there and the value is NULL (which is what you are referring to I guess)
3. the ID is NOT there

In regards to 2 it might help to have a `map[string]*ConfigT` instead of `map[string]ConfigT`. Have you tried that?

fracasula: For clarity this is what I mean: https://github.com/rudderlabs/rudder-server/compare/feat.useUpdateAfterForEfficientPolling...feat.useUpdateAfterForEfficientPolling-2?expand=1

If it makes sense please update the tests accordingly.

raman2805: Done

backend-config/namespace_config.go
```diff
@@ -98,28 +114,47 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
...
-		return config, err
+		return configOnError, err
+	}
+
+	if !nc.useIncrementalConfigUpdates {
+		nc.workspacesConfig = make(map[string]ConfigT)
 	}
 
-	for workspaceID, wc := range workspacesConfig {
+	for workspaceID, workspace := range requestData {
```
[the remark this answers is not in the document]
fracasula: @atzoum Fixed :+1:

atzoum (approved):

fracasula (approved):

PR #3179, opened 2023-04-06
feat: kafka compression

fracasula:
# Description

These changes add the ability to use compression in the Kafka manager.

Compression in Kafka can happen in two ways:
1. Compression enabled producer-side
2. Compression enabled broker-side

Here we are using the compression enabled producer-side. The advantage is that it doesn’t require any configuration change in the brokers or in the consumers. 

This is the reason why I believe that we can enabled it by default for everybody while still giving the ability to change it as need arises. Another approach is to have `None` as default, start enabling it for one customer at a time and then change the default once everybody is already running on it.

In order to understand which algorithm we should have enabled by default I built a benchmark to compare them. The main objective was to find an algorithm that would give a decent compression without being too cumbersome on CPU/Memory.

### How I built the benchmark

* one Kafka setup is created from scratch every time a compression type benchmark is created (i.e. one for gzip, one for snappy, etc...).
* the payloads are the same for all benchmarks, even if randomly generated they are generated only once at the beginning so that all algorithms have to work on the same sample.
* to measure the data transferred I'm connecting to a local proxy that measures the bytes received from the producer.
* in order to be able to use the proxy properly kafka has to advertise the proxy address as the broker address (thus the kafka destination changes).
* in order to measure the real transferred data I'm closing the proxy to make sure that all connections (thus all `io.Copy`s) are terminated. this is the reason why I start and stop the proxy each time.

### Why Ztsd as the default algorithm?

If you look at the [benchmark results](https://raw.githubusercontent.com/rudderlabs/rudder-server/test.kafkaCompression/services/streammanager/kafka/client/compression_benchmark_results.txt) you might notice that the algorithm that gives the best compression is indeed **Ztsd**.

At the same time is one of the most efficient as well. 

If we compare it with not having compression at all we can see for example that pushing a single message with no compression leads to 18k ops vs only 6k ops with Ztsd (`None` is ~285% faster than Ztsd). However with no compression we would have to transfer 300GB of data vs 110GB with Ztsd. So there it depends on where the bottleneck would be, if CPU or network. Either way not a huge win, but that's to expect without batching.

If we look at the other results where we are sending batches of messages, the data looks a lot more promising.

A batch of 100 messages of 1kb each with no compression, once repeated enough times to get the benchmark results, it leads to 2.4TB of data transferred and a total of 5077 ops (meaning `101.6 kb/op`).

With Ztsd we run only 4773 ops (i.e. `248584 ns/op` vs the faster without compression counterpart at `218305 ns/op`).
This time `None` is not ~285% faster anymore but ~106% faster but the data transferred with Ztsd caps at `1.206 kb/op` (vs `101.6 kb/op`). It beats even Gzip with its `1.865 kb/op`. However Gzip had to use a sizeable `348371 ns/op` to get there, basically 100k ns more than Ztsd. One thing to consider though is that Gzip there is a little easier on the memory.

**TL;DR:** 4773 ops vs 5077 in the same amount of time is surely a bit less but the amount of data not transferred is outstandingly lower (23.7GB vs 2.4TB).

All considered, when looking at the benchmark results I'd say that Ztsd wins in most scenarios and that it could be our default compression algorithm candidate.

## Sidenotes

* If you think it might be useful we can have the TCP proxy in the go-kit
* I'll change the base branch myself once [this PR](https://github.com/rudderlabs/rudder-server/pull/3151) gets merged.

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/Kafka-compression-54c80082bd564353b8ff16c6b8353a25) >

## Security


services/streammanager/kafka/kafkamanager.go
```diff
@@ -234,6 +235,19 @@ func Init() {
 		)
 	}
 
+	kafkaCompression = client.CompressionZstd
+	if kc := config.GetInt("Router.kafkaCompression", -1); kc != -1 {
```
atzoum, line 239: maybe we should use the "standard" naming pattern, e.g. `Router.KAFKA.compression`

services/streammanager/kafka/kafkamanager.go
```diff
@@ -234,6 +235,19 @@ func Init() {
 		)
 	}
 
+	kafkaCompression = client.CompressionZstd
```
atzoum, line 238: Should we enable compression by default only if batching is enabled too?

atzoum (approved):

services/streammanager/kafka/kafkamanager.go
fracasula: I had exactly the same thought when I was coming up with the name but then I realized that the batching parameter is the only one following that pattern :sweat_smile: 

The majority of the parameters use this other one so I went with the majority. Not strong on this though, I can change it if you guys prefer it :+1:

services/streammanager/kafka/kafkamanager.go
fracasula: Yeah I thought about it but I wasn't going to bother because I wanted to propose to enable batching for everybody given that it doesn't really affect the customers. They don't need to do anything special. We batch at the processor, then the transformer takes care of it and we just end up with an array in the kafka manager so I don't see why we shouldn't just enable compression plus batching for everybody.

Nonetheless the setting would be possible to use so it's a sensible choice not to compress if batching is disabled. Will do the change :+1:

saurav-malani (approved):

PR #3188, opened 2023-04-11
feat(destination): support event filtering in a scalable way for hybrid & cloud modes

sanpj2292:
# Description

**Requirement**
In hybrid mode, few messageType events go through device mode & rest through cloud mode. So the events which go through device mode should be filtered out at server & should not get propagated further down the pipeline.

Currently, server only has the capability to deal with device (filter all) & cloud (allow all). For hybrid mode, it needs to consider messageType too.

**Conclusion**: We are supporting event filtering based on `connectionMode`, `messageType` and `sourceType`

## Notion Ticket

https://www.notion.so/rudderstacks/Research-Filtering-events-in-Hybrid-Mode-709802f747184a3d91bd8934f23a7c91

## Security


backend-config/backend-config.go
```diff
@@ -407,3 +410,167 @@ func (bc *backendConfigImpl) Identity() identity.Identifier {
...
+			// TODO: Should remove this
+			pkgLogger.Infof("SupportedVals: %v -- EventType from event: %v\n", supportedVals, eventType)
+			allowEvent = lo.Contains(supportedVals, eventType) && evaluatedDefaultBehaviour
+		}
+	}
+	return allowEvent
+}
+
+func evaluateSupportedTypes[T comparable](destConfig map[string]interface{}, evalKey string, checkValue T) bool {
+	if !lo.Contains([]string{"supportedMessageTypes"}, evalKey) {
```
koladilip, line 559: required?
evalKey !="supportedMessageTypes"

backend-config/backend-config.go
```diff
@@ -407,3 +410,167 @@ func (bc *backendConfigImpl) Identity() identity.Identifier {
...
+	}
+	supportedConnectionModes := supportedConnectionModesI.(map[string]interface{})
+
+	supportedEventPropsMapI := misc.MapLookup(supportedConnectionModes, destConnectionMode, srcType)
+	if supportedEventPropsMapI == nil {
+		// Most probably sourceType would not be present in "supportedConnectionModes"
+		// Might also occur when destination connection mode is not provided in "supportedConnectionModes"
+		return evaluatedDefaultBehaviour
+	}
+	supportedEventPropsMap := supportedEventPropsMapI.(map[string]interface{})
```
koladilip, line 539: what about supportedEventPropsMapI.(map[string]map[string][]string) something along these lines

backend-config/backend-config.go
```diff
@@ -407,3 +409,178 @@ func (bc *backendConfigImpl) Identity() identity.Identifier {
...
+func (destination *DestinationT) AllowEventToDestination(source *SourceT, event types.SingularEventT) bool {
+	// We are setting this value through logic in rudder-server
+	// Ideally we need to send this information from config-backend itself
+	srcType := source.SourceDefinition.Type
+
+	// Event-Type default check
+	eventTypeI := misc.MapLookup(event, "type")
+	if eventTypeI == nil {
+		// TODO: Need to think through if this condition is correct
+		pkgLogger.Error("Event type is not being sent for the event")
```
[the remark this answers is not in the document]
sanpj2292: Yes. But I am handling so that we don't have any issues in the future

backend-config/types.go
```diff
@@ -194,6 +205,36 @@ func (dgSourceTPConfigT *DgSourceTrackingPlanConfigT) fetchEventConfig(eventType
...
 }
 
+/*
+Fills in sourceType value to make sure we have the capability to filter events
+
+TODO:
+May be if this value is sent from config-backend itself somehow, it'd be better for us!
+As the logic will be at a single place
+*/
+func (src *SourceT) FillSourceType() {
```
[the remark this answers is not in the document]
sanpj2292: We will be removing this function altogether actually. This was committed only for local-dev purposes.

backend-config/backend-config.go
sanpj2292: Initially we had `supportedSourceTypes` as well here. So the condition would look something like below
```
if !lo.Contains([]string{"supportedMessageTypes", "supportedSourceTypes"}, evalKey)
```

since `supportedSourceTypes` didn't make a lot of sense, removed it.
Hence the logic looks like this.

backend-config/backend-config.go
sanpj2292: It didn't work when I tried the same thing(even before writing this piece of source-code)

processor/processor.go
```diff
@@ -1475,13 +1477,21 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob, parsedE
...
+			}
 
 			for i := range enabledDestTypes {
 				destType := &enabledDestTypes[i]
 				enabledDestinationsList := proc.filterDestinations(
 					singularEvent,
 					proc.getEnabledDestinations(writeKey, *destType),
 				)
+				// Additional filteration based on connectionMode
+				enabledDestinationsList = lo.Filter(enabledDestinationsList, func(destination backendconfig.DestinationT, _ int) bool {
```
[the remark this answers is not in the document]
sanpj2292: Makes sense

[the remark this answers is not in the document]
sanpj2292: Updated

backend-config/backend-config.go
```diff
@@ -407,3 +409,178 @@ func (bc *backendConfigImpl) Identity() identity.Identifier {
...
+	supportedValsI := misc.MapLookup(destConfig, evalKey)
+	if supportedValsI == nil {
+		return false
+	}
+	supportedVals := ConvertToArrayOfType[T](supportedValsI)
+	return lo.Contains(supportedVals, checkValue)
+}
+
+func ConvertToArrayOfType[R any](valsI interface{}) []R {
+	valuesArrI, ok := valsI.([]interface{})
```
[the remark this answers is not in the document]
sanpj2292: Moved to a different casting technique

backend-config/backend-config.go
sanpj2292: I have simplified the condition for now!

sanpj2292:
We have changed the structure, discussed all the cases possible and made the changes accordingly
The details are [here](https://www.notion.so/rudderstacks/Research-Filtering-events-in-Hybrid-Mode-709802f747184a3d91bd8934f23a7c91?pvs=4#81bd3b9e580a49c6bcce6c906361223e)

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+
+			pkgLogger.Debugf("SupportedVals: %v -- EventType from event: %v\n", messageTypeEventPropDef.AllowedVals, messageType)
+			allowEvent = lo.Contains(messageTypeEventPropDef.AllowedVals, messageType) && evaluatedDefaultBehaviour
+		}
+	}
+	return allowEvent, nil
+}
+
+type EventPropsTypes interface {
+	~string
```
koladilip, line 270: This is new for me, can you share doc for ~string

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+func getSupportedEventPropertiesMap(destinationDefinitionConfig map[string]interface{}, srcType, destConnectionMode string, defaultBehaviour bool) supportedEventProperties {
+	supportedConnectionModesI, connModesOk := destinationDefinitionConfig["supportedConnectionModes"]
+	if !connModesOk {
+		pkgLogger.Debug("supportedConnectionModes not present, filtering event based on default behaviour")
+		return supportedEventProperties{Allow: defaultBehaviour}
+	}
+	supportedConnectionModes := supportedConnectionModesI.(map[string]interface{})
+
+	supportedEventPropsMapI, supportedEventPropsLookupErr := misc.NestedMapLookup(supportedConnectionModes, srcType, destConnectionMode)
+	if supportedEventPropsLookupErr != nil {
```
koladilip, line 383: Looks bit complicated, why don't be look for srcType first and take some decisions based on that then check for destConnectionMode

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+	}
+
+	eventProperties := getSupportedEventPropertiesMap(destination.DestinationDefinition.Config, srcType, destConnectionMode, evaluatedDefaultBehaviour)
+	if len(eventProperties.EventPropsMap) == 0 {
+		return eventProperties.Allow, eventProperties.FailedResponse
+	}
+
+	// Flag indicating to let the event pass through
+	allowEvent := evaluatedDefaultBehaviour
+	for eventProperty, supportedEventVals := range eventProperties.EventPropsMap {
```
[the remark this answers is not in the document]
sanpj2292: In the future we might require additional fields to be incorporated. Hence kept a loop.
Any logic specific to the addtional field alone has to be implemented

processor/eventfilter/eventfilter.go
sanpj2292: https://stackoverflow.com/questions/72570147/interface-for-all-numbers

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+			}
+		}
+		return result
+	}
+	return []T{}
+}
+
+type EventProperty[T EventPropsTypes] struct {
+	AllowAll    bool
+	AllowedVals []T
```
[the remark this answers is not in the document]
sanpj2292: None currently.

processor/eventfilter/eventfilter.go
sanpj2292: We can do the checking 1-by-1 it looked rather un-comfortable.

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+	}
+
+	// Flag indicating to let the event pass through
+	allowEvent := evaluatedDefaultBehaviour
+	for eventProperty, supportedEventVals := range eventProperties.EventPropsMap {
+		eventPropMap, isMap := supportedEventVals.(map[string]interface{})
+
+		if !allowEvent || !isMap {
+			pkgLogger.Debugf("Previous evaluation of allowAll is false or type assertion failed for an event property(%v), filtering event based on default behaviour", eventProperty)
+			allowEvent = evaluatedDefaultBehaviour
```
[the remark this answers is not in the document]
sanpj2292: Even though `allowEvent` is `false` or the value mentioned is not a map, we are relying on defaultBehaviour to tell us what do with the event.

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+		// When something is missing in "supportedConnectionModes" or if "supportedConnectionModes" is not defined
+		// We would be checking for below things
+		// 1. Check if the event.type value is present in destination.DestinationDefinition.Config["supportedMessageTypes"]
+		// 2. Check if the connectionMode of destination is cloud or hybrid(evaluated through `IsProcessorEnabled`)
+		// Only when 1 & 2 are true, we would allow the event to flow through to server
+		// As when this will be called, we would have already checked if event.type in supportedMessageTypes
+		DefaultBehaviour: transformerEvent.Destination.IsProcessorEnabled && isSupportedMsgType,
+	})
+
+	if !allow {
```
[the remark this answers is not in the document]
sanpj2292: Ideally `eventfilter.go` should contain all the filteration logic for source-event-destination combination.
I have intentionally left this part open for someone else to write more logic related to event-filtering here

processor/eventfilter/eventfilter.go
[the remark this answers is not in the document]
koladilip: we are setting default Behaviour initially right?

processor/eventfilter/eventfilter.go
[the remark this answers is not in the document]
koladilip: what is our vision for making this generic?

processor/eventfilter/eventfilter.go
koladilip: but level can lead to some errors and not sure if this is making code more readable

koladilip (approved):
LGTM

processor/eventfilter/eventfilter.go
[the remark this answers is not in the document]
sanpj2292: If you really see, we are doing below things after we find an event property
1. Compute the event property value(here it is already done earlier itself)
2. Type-cast into EventProperty struct
3. Check if the computed value is present in the EventProperty struct's properties(any one of [AllowAll, AllowedValues])

We have to ideally have a function that does the type-casting work easily and also checking of value's presence

So, for casting I tried to write a generic function to cover all basic types(string, int(int64,int32,uint16 etc,.) that we might get. But including ints and floats with string is causing some problems, which can be solved using go's reflect package. But go reflect package makes things slower hence for now restricted the casting logic to just strings

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+
+	destConnModeI := misc.MapLookup(destination.Config, "connectionMode")
+	if destConnModeI == nil {
+		pkgLogger.Debug("connectionMode not present, filtering event based on default behaviour")
+		return evaluatedDefaultBehaviour, nil
+	}
+	destConnectionMode, isDestConnModeString := destConnModeI.(string)
+	if !isDestConnModeString || destConnectionMode == "device" {
+		// includes Case 6
+		pkgLogger.Errorf("Provided connectionMode(%v) is in wrong format or the mode is device", destConnModeI)
```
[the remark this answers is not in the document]
sanpj2292: We can emit some stats for this case but I'm not sure if the stat will actually have any value. I mean, I'm not able to think of any actionable item if we create an alert for this use-case 💭 
Let me know if you have any inputs here

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+
+func getMessageType(event *types.SingularEventT) string {
+	eventMessageTypeI := misc.MapLookup(*event, "type")
+	if eventMessageTypeI == nil {
+		pkgLogger.Error("Message type is not being sent for the event")
+		return ""
+	}
+	eventMessageType, isEventTypeString := eventMessageTypeI.(string)
+	if !isEventTypeString {
+		pkgLogger.Errorf("Invalid message type :%v, Type assertion failed", eventMessageTypeI)
```
lvrach, line 105: [minor] Should the severity be an Error here, since it is not under our control, should we make this into Warn? 

Plus more go idiomatic way of logging [nipik] 

```suggestion
		pkgLogger.Error("Message type is not being sent for the event")
		return ""
	}
	eventMessageType, isEventTypeString := eventMessageTypeI.(string)
	if !isEventTypeString {
		pkgLogger.Errorf("Invalid message type: type assertion failed: %v", eventMessageTypeI)
```

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+		7. if eventProperty not in sourceType.connectionMode -- rely on property based default behaviour
+		8. if sourceType.connectionMode.eventProperty = {} (both allowAll, allowedValues are not included in defConfig) -- rely on default behaviour
+		9. if sourceType.connectionMode.eventProperty = { allowAll: true, allowedValues: ["track", "page"]} (both allowAll, allowedValues are included in defConfig) -- rely on default behaviour
+	*/
+
+	destConnModeI := misc.MapLookup(destination.Config, "connectionMode")
+	if destConnModeI == nil {
+		pkgLogger.Debug("connectionMode not present, filtering event based on default behaviour")
+		return evaluatedDefaultBehaviour, nil
+	}
```
lvrach, line 228: Instead of ignoring the error, we can use for better log message, plus is more idiomatic to check for absent error rather than nil

```suggestion
	destConnModeI, err := misc.MapLookup(destination.Config, "connectionMode")
	if err != nil {
		pkgLogger.Debug("connectionMode not present, filtering event based on default behaviour: %v", err)
		return evaluatedDefaultBehaviour, nil
	}
```

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+    - Not expected
+
+    - EventProperty{AllowAll:false}
+*/
+func ConvertEventPropMapToStruct[T EventPropsTypes](eventPropMap map[string]interface{}) *EventProperty[T] {
+	var eventPropertyStruct EventProperty[T]
+	for _, key := range []string{"allowAll", "allowedValues"} {
+		val, ok := eventPropMap[key]
+		if !ok {
+			pkgLogger.Debugf("'%v' not found in eventPropertiesMap(supportedConnectionModes.sourceType.connectionMode.[eventProperty])", key)
```
lvrach, line 334: [minor[
```suggestion
			pkgLogger.Debugf(" %q not found in eventPropertiesMap(supportedConnectionModes.sourceType.connectionMode.[eventProperty])", key)
```

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+
+ 4. { [eventProperty]: { }  }
+
+    - Not expected
+
+    - EventProperty{AllowAll:false}
+*/
+func ConvertEventPropMapToStruct[T EventPropsTypes](eventPropMap map[string]interface{}) *EventProperty[T] {
+	var eventPropertyStruct EventProperty[T]
+	for _, key := range []string{"allowAll", "allowedValues"} {
```
lvrach, line 331: Should we store as constants:

```go
[]string{"allowAll", "allowedValues"}
```

processor/eventfilter/eventfilter.go
```diff
@@ -69,3 +77,325 @@ func GetSupportedMessageEvents(destination *backendconfig.DestinationT) ([]strin
...
+/*
+This methods lets the caller know if we need to allow the event to proceed to destination transformation
+
+Currently this method supports below validations(executed in the same order)
+
+1. Validate if messageType sent in event is included in SupportedMessageTypes
+
+2. Validate if the event is sendable to destination based on connectionMode, sourceType & messageType
+*/
+func AllowEventToDestTransformation(transformerEvent *transformer.TransformerEventT, supportedMsgTypes []string) (bool, *transformer.TransformerResponseT) {
```
lvrach, line 120: [minor] idiomatic function comment,  should begin with function name

https://tip.golang.org/doc/comment

```suggestion
/*
AllowEventToDestTransformation lets the caller know if we need to allow the event to proceed to destination transformation.

Currently this method supports below validations(executed in the same order):
1. Validate if messageType sent in event is included in SupportedMessageTypes
2. Validate if the event is sendable to destination based on connectionMode, sourceType & messageType
*/
func AllowEventToDestTransformation(transformerEvent *transformer.TransformerEventT, supportedMsgTypes []string) (bool, *transformer.TransformerResponseT) {
```

lvrach (approved):
LGTM, with minor comments.

I am not very versed with this part of the code, but there is a good code coverage and logic makes sense.

processor/eventfilter/eventfilter.go
sanpj2292: `MapLookup` returns an `interface{}` only. Hence the check condition was written as such.

processor/eventfilter/eventfilter.go
sanpj2292: Done

processor/eventfilter/eventfilter.go
sanpj2292: Done

processor/eventfilter/eventfilter.go
sanpj2292: Moved to constants.

processor/eventfilter/eventfilter.go
sanpj2292: Except the spacing, taken care of other things

chandumlg (approved):

sanpj2292:
We would like to change the structure to the below one
```
{
	...
	"hybridModeCloudEventsFilter": {
	      "web": {
	        "messageType": ["track", "page"]
	      }
	    },
	...
}
```

As it solves the problem specific to hybrid-mode. The general structure had some problems that could again trigger a refactor.
We would take up the task of refactoring later in the next quarter(s) as per the priorities and requirements at that point of time

processor/processor.go
```diff
@@ -2426,16 +2435,11 @@ func ConvertToFilteredTransformerResponse(events []transformer.TransformerEventT
...
-				if !typOk {
-					// add to FailedEvents
-					errMessage = "Invalid message type. Type assertion failed"
-					resp = transformer.TransformerResponseT{Output: event.Message, StatusCode: 400, Metadata: event.Metadata, Error: errMessage}
-					failedEvents = append(failedEvents, resp)
-					continue
-				}
-				messageType = strings.TrimSpace(strings.ToLower(messageType))
-				if !slices.Contains(supportedTypes.values, messageType) {
+				allow, failedEvent := eventfilter.AllowEventToDestTransformation(event, supportedTypes.values)
```
atzoum, line 2438: This is called only if the destination's configuration has a `supportedMessageTypes` field. Otherwise the assumption is that all types are supported and this block is skipped.
Is this condition valid for the message type filtering scenario that is being introduced too?

sanpj2292: The logic was that way even before I added my changes. I have let it be like that for now

sanpj2292: I think we were ok with that approach and let it be that way

atzoum: Yes, the previous behavior was the desired one and we need to keep it that way.
The question is whether this new kind of message type filtering should only be applied only to destinations that have `supportedMessageTypes` in their config or to every destination.

sanpj2292: IMHO, Since the message type filtering deals with the hybrid-mode destination's cloud behaviour(in terms of filtering events), we can keep the previous behaviour as is i.e only those destination(s) which have `supportedMessageTypes` would be eligible for this kind of event filtering logic.

@krishna2020 to confirm on this

processor/transformer/transformer.go
```diff
@@ -68,7 +68,9 @@ type MetadataT struct {
 	EventType               string   `json:"eventType"`
 	SourceDefinitionID      string   `json:"sourceDefinitionId"`
 	DestinationDefinitionID string   `json:"destinationDefinitionId"`
-	SourceDefinitionType    string
+	TransformationID        string   `json:"transformationId"`
+	TransformationVersionID string   `json:"transformationVersionId"`
+	SourceDefinitionType    string   `json:"-"`
```
koladilip, line 73: any reason why  `json:"-"`

koladilip (commented):
MapLookupError is used now?

processor/transformer/transformer.go
sanpj2292: Basically it doesn't appear in json after marshalling, figured it wouldn't be needed for transformer

sanpj2292:
> MapLookupError is used now?

Not right now! But I let it be as it can be used by anyone else as needed

backend-config/types.go
```diff
@@ -40,6 +40,7 @@ type SourceDefinitionT struct {
 	ID       string
 	Name     string
 	Category string
+	Type     string // // Indicates whether source is one of {cloud, web, flutter, android, ios, warehouse, cordova, amp, reactnative, unity}. This field is not present in sources table
```
atzoum, line 43: Is this a new field that control plane is going to provide?

sanpj2292: Yes

atzoum: fyi, in the future we'll be switching to the v2 config api, now [in progress](https://github.com/rudderlabs/rudder-cp-sdk/blob/initial-setup/model/model.go#L74-L83)

koladilip (approved):

PR #3201, opened 2023-04-12
feat(processor): forward events to new events-schema jobsDB

Sidddddarth:
# Description

Processor writes jobs to a new jobsdb - eventSchemasDB as well.

## Notion Ticket

[event schemas jobs db](https://www.notion.so/rudderstacks/new-jobsdb-for-event-schema-4f3d9595f0164a4e98f46b4ca58f8b8f?pvs=4)

## Security


processor/manager_test.go
```diff
@@ -201,7 +201,7 @@ func TestProcessorManager(t *testing.T) {
 			"batch_rt": &jobsdb.MultiTenantLegacy{HandleT: brtDB},
 		},
 	}
-	processor := New(ctx, &clearDb, gwDB, rtDB, brtDB, errDB, mtStat, &reporting.NOOP{}, transientsource.NewEmptyService(), fileuploader.NewDefaultProvider(), mockRsourcesService, destinationdebugger.NewNoOpService(), transformationdebugger.NewNoOpService(),
+	processor := New(ctx, &clearDb, gwDB, rtDB, brtDB, errDB, nil, mtStat, &reporting.NOOP{}, transientsource.NewEmptyService(), fileuploader.NewDefaultProvider(), mockRsourcesService, destinationdebugger.NewNoOpService(), transformationdebugger.NewNoOpService(),
```
atzoum, line 204: let's use an actual db in this test

processor/processor.go
```diff
@@ -1350,18 +1367,57 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob, parsedE
 					source,
 				)
 
+				payloadFunc := func() json.RawMessage {
```
atzoum, line 1370: we are effectively calling this twice. We can memoize it by assigning the result to a variable and returning this variable during subsequent calls

processor/processor.go
```diff
@@ -1350,18 +1367,57 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob, parsedE
...
+					}
+					return payloadBytes
+				}
+				if proc.config.eventSchemaV2Enabled && // schemas enabled
+					// source has schemas enabled or if we override schemas for all sources
+					(source.EventSchemasEnabled || proc.config.eventSchemaV2AllSources) &&
+					// TODO: could use source.SourceDefinition.Category instead?
+					commonMetadataFromSingularEvent.SourceJobRunID == "" {
+					payload := payloadFunc()
+					if payload != nil {
```
atzoum, line 1386:
```suggestion
					if payload := payloadFunc(); payload != nil {
```

processor/processor.go
atzoum, line 1370: first call assumes `nil` as the default and the second one assumes `{}`
```suggestion
				payloadFunc := func(default json.RawMessage) json.RawMessage {
```

processor/processor.go
```diff
@@ -1410,6 +1466,27 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob, parsedE
 		statusList = append(statusList, &newStatus)
 	}
 
+	if len(eventSchemaJobs) > 0 {
+		proc.logger.Debug("[Processor] Total jobs written to event_schema: ", len(eventSchemaJobs))
```
atzoum, line 1470: this debug statement is misleading since nothing has been written here

processor/processor.go
```diff
@@ -1350,18 +1367,68 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob, parsedE
...
+								return payloadBytes
+							}()
+							called = true
+						}
+						if actualPayload == nil {
+							return defaultPayload
+						}
+						return actualPayload
+					}
+				}()
```
atzoum, line 1394: This appears to have too much code. Can we do the following?

1. We create a reusable function in rudder-go-kit
```go
package ro

func Memoize[R any](f func() R) func() R {
	var result R
	var called bool
	return func() R {
		if called {
			return result
		}
		result = f()
		return result
	}
}
```
2. We use it like so
```go
payloadFunc := ro.Memoize(func() json.RawMessage {
	if proc.transientSources.Apply(source.ID) {
		return nil
	}
	payloadBytes, err := jsonfast.Marshal(singularEvent)
	if err != nil {
		return nil
	}
	return payloadBytes
})

...

if payload := payloadFunc(); payload != nil {

...


func() json.RawMessage {
	if p := payloadFunc(); p != nil {
		return p
	}
	return []byte("{}")
},
```

Sidddddarth: [raised a PR here](https://github.com/rudderlabs/rudder-go-kit/pull/30)
Will change here after that goes in.

atzoum: https://github.com/rudderlabs/rudder-go-kit/releases/tag/v0.13.0

atzoum (approved):

cisse21 (approved):

PR #3237, opened 2023-04-24
fix: postgres support for loading users and identifies in same transactions

achettyiitr:
# Description

- Since we are using temporary tables for loading the data, loading of users and identifies table needs to be done in a single transaction.
-  Adding sqlmiddleware.WithTx to run queries in transactions.

## Notion Ticket

https://www.notion.so/rudderstacks/Postgres-temp-table-implementation-getting-a-staging-table-not-exists-On-retries-it-s-getting-succ-4b350c767266424ca2e81b9ffde889ff?pvs=4

## Security


warehouse/integrations/postgres/load.go
```diff
@@ -61,7 +67,7 @@ type LoadUsersTable struct {
 	LoadFileDownloader downloader.Downloader
 }
 
-func (lt *LoadTable) Load(ctx context.Context, tableName string, tableSchemaInUpload model.TableSchema) (string, error) {
+func (lt *LoadTable) Load(ctx context.Context, tableName string, tableSchemaInUpload model.TableSchema) (LoadTableResponse, error) {
```
atzoum, line 70: It is a preferred pattern to begin the transaction externally and pass it as an argument to this method.
Having one method creating a transaction whose lifecycle needs to be managed by the caller can be considered to be a weak, error-prone design.
Typically we prefer having clear transaction boundaries, e.g. `WithTx(func(tx *sql.Tx) error {})`, so that transaction can be consistently commited/rolledback as it is fit in all scenarios.

achettyiitr: Good point. I was also thinking of passing transactions but because of the complexity of the code, I took a short route.
Also, I have started a [discussion](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1682341076776989) around the decoupling of identifies and users table loading so that transactions won't be required.

let me see how we can pass transactions.

warehouse/integrations/middleware/sqlquerywrapper/tx.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func (tx *txWithTimeout) Rollback() error {
+	ctx, cancel := context.WithTimeout(context.Background(), tx.timeout)
+	defer cancel()
+
+	errCh := make(chan error, 1)
+	go func() {
+		errCh <- tx.Tx.Rollback()
+	}()
```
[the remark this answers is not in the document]
achettyiitr: For customers, we have seen queries getting stuck at Rollback for couple of hours. 
https://rudderlabs.slack.com/archives/C02AALVFHJN/p1649286005516699?thread_ts=1649079527.127459&cid=C02AALVFHJN

warehouse/integrations/azure-synapse/azure-synapse.go
```diff
@@ -770,11 +770,11 @@ func (as *AzureSynapse) FetchSchema(warehouse model.Warehouse) (schema, unrecogn
 	return
 }
 
-func (as *AzureSynapse) LoadUserTables() map[string]error {
+func (as *AzureSynapse) LoadUserTables(context.Context) map[string]error {
 	return as.loadUserTables()
```
fracasula, line 774: Why not propagate the context to `as.loadUserTables` as well? It looks like [it can be benefit from it](https://github.com/rudderlabs/rudder-server/blob/fix.postgres-load-table/warehouse/integrations/azure-synapse/azure-synapse.go#L511):

```go
	// BEGIN TRANSACTION
	tx, err := as.DB.BeginTx(ctx)
```

fracasula: Same with `LoadTable` where for example we have code like the following that can be updated:

```go
fileNames, err := as.LoadFileDownLoader.Download(context.TODO(), tableName)
```

warehouse/integrations/middleware/sqlquerywrapper/sql.go
```diff
@@ -108,6 +109,28 @@ func (db *DB) QueryRowContext(ctx context.Context, query string, args ...interfa
 	return row
 }
 
+func (db *DB) WithTx(fn func(*Tx) error, timeout time.Duration) error {
```
fracasula, line 112: What about having a context propagated here as well? I haven't checked how extensive the change might be but it may be worth considering. If not in this PR perhaps it can be addressed in a dedicated PR. Wdyt?

warehouse/integrations/middleware/sqlquerywrapper/sql.go
```diff
@@ -108,6 +109,28 @@ func (db *DB) QueryRowContext(ctx context.Context, query string, args ...interfa
...
+		return fmt.Errorf("begin transaction: %w", err)
+	}
+
+	txt := txWithTimeout{
+		Tx:      tx,
+		timeout: timeout,
+	}
+
+	defer func() {
+		_ = txt.Rollback()
```
fracasula, line 124: I know that transactions cannot be rolled back after they are committed in PostgreSQL but I was wondering if that is the case with other SQL DBs as well. I'm asking because this is a generic sql wrapper that can be used with other backends as well, right?

Using always the same `err` and checking in the `defer` if `err != nil` could also be an option.

warehouse/integrations/middleware/sqlquerywrapper/sql_test.go
```diff
@@ -98,6 +98,13 @@ func TestQueryWrapper(t *testing.T) {
 
 			_ = qw.QueryRowContext(ctx, query)
 			require.NoError(t, err)
+
+			_ = qw.WithTx(func(tx *Tx) error {
+				_, err := tx.Exec(query)
+				return err
+			},
+				1*time.Second,
+			)
```
fracasula, line 107: [minor] this code seems weirdly formatted :thinking:

warehouse/integrations/middleware/sqlquerywrapper/tx.go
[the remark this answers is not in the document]
fracasula: @achettyiitr are we sure that passing a context with a timeout to `BeginTx` doesn't timeout also the rollback in case it takes too long?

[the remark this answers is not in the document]
fracasula: We could test this with a `SELECT pg_sleep(X)` where `X` would be a number of seconds greater than the ones passed to the context timeout.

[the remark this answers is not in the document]
fracasula: FYI, some of these could return `pq: canceling statement due to user request` instead of `context.Canceled` or `context.DeadlineExceeded`. It might be worth checking for it (e.g. in case you got some retry policy in place in case that would affect the retry logic). Feel free to ignore if you think it won't affect the logic in any way.

warehouse/integrations/postgres/diagnostic.go
fracasula, line 1: How come this file was deleted?

warehouse/upload.go
```diff
@@ -564,7 +564,10 @@ func (job *UploadJob) run() (err error) {
 			})
 
 			rruntime.GoForWarehouse(func() {
-				specialTables := append(userTables, identityTables...)
+				var specialTables []string
```
fracasula, line 567: We could preallocate here specifying the desired `capacity`, right? Basically by using the 3rd argument of [make](https://go.dev/tour/moretypes/13) like [this](https://go.dev/play/p/8MV8RK5g2Wj).

lvrach (approved):

fracasula (approved):
LGTM, thanks for changes @achettyiitr

warehouse/integrations/middleware/sqlquerywrapper/sql.go
```diff
@@ -108,6 +116,25 @@ func (db *DB) QueryRowContext(ctx context.Context, query string, args ...interfa
...
+		if rollbackErr := tx.Rollback(); rollbackErr != nil && !errors.Is(rollbackErr, sql.ErrTxDone) {
+			keysAndValues := []any{logfield.Error, fmt.Errorf("executing transaction: %s, rollback: %s", err.Error(), rollbackErr.Error()).Error()}
+			keysAndValues = append(keysAndValues, db.keysAndValues...)
+
+			db.logger.Warnw("failed rollback transaction", keysAndValues...)
+		}
+		return fmt.Errorf("executing transaction: %w", err)
+	}
+
+	return tx.Commit()
```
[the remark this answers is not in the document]
achettyiitr: Good point. Let me add this as a separate PR.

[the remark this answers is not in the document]
achettyiitr: https://github.com/rudderlabs/rudder-server/pull/3273

PR #3243, opened 2023-04-25
feat(router): limit number of jobs with the same ordering key in worker's queue

atzoum:
# Description

Until now, a single concurrency limiter existed in router's barrier, i.e. our event ordering mechanism, which was there to ensure that when a failed job got drained (aborted), in the next router generator loop only a single job would be picked up by the router for the same ordering key. This happens so as not to flood the worker's queue with jobs which are most probably going to be marked as `waiting` later on.

We are now introducing an additional concurrency limiter, which is responsible for limiting the overall maximum number of jobs that can enter the barrier for the same ordering key, regardless of failures. This barrier will protect us from a single "chatty" ordering key saturating the worker’s queue, so that jobs for other ordering keys can have a fair chance of being picked up by router.

The new concurrency limiter can be disabled by setting the limit to 0.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/Pipelines-Sprint-68c1c213e9b840b3a9b3826e3631e39a?p=74682dae510f457b91e992cf04e8effa&pm=s)

## Security

### DIFF
--- a/router/internal/eventorder/eventorder.go
+++ b/router/internal/eventorder/eventorder.go
@@ -72,8 +72,8 @@ type Barrier struct {
 // returns true, otherwise false along with the previous failed jobID if this is the cause of the barrier.
 // Another scenario where a barrier might exist for a key is when the previous job has failed in an unrecoverable manner and the concurrency limiter is enabled.
 func (b *Barrier) Enter(key string, jobID int64) (accepted bool, previousFailedJobID *int64) {
-	b.mu.RLock()
-	defer b.mu.RUnlock()
+	b.mu.Lock()
+	defer b.mu.Unlock()
 	barrier, ok := b.barriers[key]
 	if !ok {
 		if b.concurrencyLimit == 0 { // if not concurrency limit is set accept the job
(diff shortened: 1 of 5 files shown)


cisse21 (approved):

router/internal/eventorder/eventorder.go
```diff
@@ -67,35 +76,40 @@ func (b *Barrier) Enter(key string, jobID int64) (accepted bool, previousFailedJ
...
 	if !ok {
-		return true, nil // no barrier, accept
+		if b.concurrencyLimit == 0 { // if not concurrency limit is set accept the job
+			return true, nil
+		}
+		// create a new barrier with the concurrency limiter enabled
+		barrier = &barrierInfo{
+			concurrencyLimiter: make(map[int64]struct{}),
+		}
+		b.barriers[key] = barrier
```
fracasula, line 86: Is it OK to write into `b.barriers` even though on line 75 only a read lock was acquired?

atzoum: thanks, need to change the lock type to full

fracasula (approved):

PR #3248, opened 2023-04-26
feat(batchrouter): introduce isolation levels and concurrency limiters

atzoum:
# Description

Introducing isolation strategies in batchrouter, following the pattern we are already using in processor.

Batchrouter Isolation

The following isolation modes are now supported in batchrouter:

1. `none` - a single worker retrieves and processes jobs for all workspaces & destinations
2. `workspace` - using a separate worker per workspace (default mode in multi-tenant deployments)
3. `destination` - using a separate worker per destination (default mode in non-multi-tenant deployments)

To avoid proliferation of busy goroutines, we are introducing 3 limiters:
1. `read` - for reading jobs from jobsdb (limit: 20 goroutines/destType)
2. `process` -  for controlling concurrency while processing jobs grouped by destination (limit: 20 goroutines/destType)
3. `upload` - for controlling concurrency while uploading files grouped by source & destination (limit: 50 goroutines/destType)

## Upload frequency adaptations
In the new design, upload frequency is still honoured, but only as long as jobsdb's read limits are not being reached (limited by the number of events and/or total payload size). If a worker detects that jobsdb's limits are reached, it disregards upload frequency restrictions, until the backlog is flushed.

**Side-effects of upload frequency adaptations in the context of isolation modes:**

- `none`: destinations with sparse jobs will be processed more frequently when limits are reached for any other destination.
- `workspace`: destinations with sparse jobs will be processed more frequently when limits are reached for any other destination beloning to the same workspace.
- `destination`: none, only destinations with large backlog will be processed more frequently when limits are reached.

## Performance

Similarly to what has already been observed in processor, both isolation modes (`workspace` & `destination`) perform equally well with `none` up to a cardinality of 200. After that, throughput starts deteriorating, which is an acceptable toll to pay, taking into consideration the other advantages associated with isolation (i.e. one customer/destination not affecting the other).

The relevant benchmark can be found inside `batchrouter_isolation_test.go` and a visualisation of one execution of this benchmark is available [here](https://snapshots.raintank.io/dashboard/snapshot/r3Yfqat9xfyLvhUFK6jTtH1FZ81S6duh).  

## Other notable changes

- Extracted worker pool & worker lifecycle management logic to a common utility package, dubbed `workerpool`, so that it can be shared between processor and batchrouter.
- Reorganised package's code in separate, logically-grouped files, trying to make it easier to read and follow.
- Removed global variables and `Init` function from the package.
- Introduced the db utility function `unionjobsdbmetadata` which is a faster version of `unionjobsdb`, since it doesn't include the payloads in the resultsets.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/Pipelines-Sprint-68c1c213e9b840b3a9b3826e3631e39a?p=8267e0cee9a147bca10fc388e784494b&pm=s)

## Security


processor/worker.go
```diff
@@ -58,38 +52,18 @@ type worker struct {
 
 // start starts the various worker goroutines
 func (w *worker) start() {
-	// ping loop
+	if !w.handle.config().enablePipelining {
```
cisse21, line 55: We should get rid of the `enablePipelining` config probably in future as we stopped using it

processor/worker.go
```diff
@@ -13,15 +13,14 @@ import (
...
+// newProcessorWorker creates a new worker
+func newProcessorWorker(partition string, h workerHandle) *worker {
 	w := &worker{
 		handle:    h,
 		logger:    h.logger().Child(partition),
 		partition: partition,
 	}
-	w.lifecycle.ctx, w.lifecycle.cancel = context.WithCancel(ctx)
-	w.channel.ping = make(chan struct{}, 1)
+	w.lifecycle.ctx, w.lifecycle.cancel = context.WithCancel(context.Background())
```
cisse21, line 23: Curious why did we stop propagating the context from parent

router/batchrouter/batchrouter.go
```diff
@@ -1,2415 +1,14 @@
...
-	config.RegisterBoolConfigVariable(true, &readPerDestination, false, "BatchRouter.readPerDestination")
-	config.RegisterStringConfigVariable("", &toAbortDestinationIDs, true, "BatchRouter.toAbortDestinationIDs")
-	config.RegisterDurationConfigVariable(10, &netClientTimeout, false, time.Second, "BatchRouter.httpTimeout")
-	transformerURL = config.GetString("DEST_TRANSFORM_URL", "http://localhost:9090")
-	config.RegisterStringConfigVariable("", &datePrefixOverride, true, "BatchRouter.datePrefixOverride")
-	dateFormatLayouts = map[string]string{
+	json                                      = jsoniter.ConfigCompatibleWithStandardLibrary
+	objectStoreDestinations []string          = []string{"S3", "GCS", "AZURE_BLOB", "MINIO", "DIGITAL_OCEAN_SPACES"}
+	asyncDestinations       []string          = []string{"MARKETO_BULK_UPLOAD"}
+	dateFormatLayouts       map[string]string = map[string]string{
```
cisse21, line 9: Minor

```suggestion
	objectStoreDestinations = []string{"S3", "GCS", "AZURE_BLOB", "MINIO", "DIGITAL_OCEAN_SPACES"}
	asyncDestinations       = []string{"MARKETO_BULK_UPLOAD"}
	dateFormatLayouts                = map[string]string{
```

router/batchrouter/batchrouter.go
cisse21, line 1: Does it still make sense to call this file `batchrouter.go` when it has nothing but constants

router/batchrouter/batchrouter_test.go
```diff
@@ -149,36 +149,38 @@ var _ = Describe("BatchRouter", func() {
...
 
 	BeforeEach(func() {
+		config.Reset()
 		router_utils.JobRetention = time.Duration(175200) * time.Hour // 20 Years(20*365*24)
 		c = &testContext{}
 		c.Setup()
 	})
 
 	AfterEach(func() {
+		config.Reset()
```
cisse21, line 159: Do we need to reset it both in `BeforeEach` and `AfterEach` ?

router/batchrouter/batchrouter_test.go
```diff
@@ -259,7 +283,18 @@ var _ = Describe("BatchRouter", func() {
...
 
 			<-batchrouter.backendConfigInitialized
-			batchrouter.readAndProcess()
+			batchrouter.mainLoopSleep = time.Microsecond
+			ctx, cancel := context.WithCancel(context.Background())
+			var wg sync.WaitGroup
+			wg.Add(1)
+			go func() {
+				batchrouter.mainLoop(ctx)
+				GinkgoRecover()
```
cisse21, line 292: Probably a remanent after debugging?

processor/worker.go
atzoum: For controlling the worker's lifecycle exclusively through `Stop`. If the parent context is canceled, the worker's stop will eventually be called by the pool

router/batchrouter/batchrouter_test.go
atzoum: it depends, does the previous/next test in the suite resets? who knows!

router/batchrouter/batchrouter_test.go
atzoum: oops!

router/batchrouter/handle.go
```diff
@@ -0,0 +1,841 @@
...
+		}
+	}
+
+	outputFile, err := os.Open(gzipFilePath)
+	if err != nil {
+		panic(err)
+	}
+
+	brt.logger.Debugf("BRT: Starting upload to %s", provider)
+	folderName := ""
```
cisse21, line 380: minor
```suggestion
	var folderName string
```

router/batchrouter/handle.go
```diff
@@ -0,0 +1,841 @@
...
+		folderName = config.GetString("WAREHOUSE_STAGING_BUCKET_FOLDER_NAME", "rudder-warehouse-staging-logs")
+	} else {
+		folderName = config.GetString("DESTINATION_BUCKET_FOLDER_NAME", "rudder-logs")
+	}
+
+	var datePrefixLayout string
+	if brt.datePrefixOverride != "" {
+		datePrefixLayout = brt.datePrefixOverride
+	} else {
+		dateFormat, _ := brt.dateFormatProvider.GetFormat(brt.logger, uploader, batchJobs.Connection, folderName)
```
cisse21, line 391: @achettyiitr Do we still need this part of the code? I remember this was something temporary done a while back

pChondros (approved):

PR #3249, opened 2023-04-26
chore: added logs to help debug suppression backup service issue

saurav-malani:
# Description

added logs to help debug suppression backup service issue.


enterprise/suppress-user/factory.go
```diff
@@ -82,6 +82,7 @@ func (m *Factory) Setup(ctx context.Context, backendConfig backendconfig.Backend
 
 			subCtx, latestSyncCancel := context.WithCancel(ctx)
 			rruntime.Go(func() {
+				m.Log.Infof("Starting latest sync")
```
atzoum, line 85:
```suggestion
				m.Log.Infof("Starting latest suppression sync")
```

enterprise/suppress-user/factory.go
```diff
@@ -91,6 +92,7 @@ func (m *Factory) Setup(ctx context.Context, backendConfig backendconfig.Backend
 				if err != nil {
 					m.Log.Errorf("Latest Sync failed: could not remove repo: %w", err)
 				}
+				m.Log.Info("Latest sync stopped")
```
atzoum, line 95:
```suggestion
				m.Log.Info("Latest suppression sync stopped")
```

enterprise/suppress-user/factory.go
```diff
@@ -115,14 +117,17 @@ func (m *Factory) Setup(ctx context.Context, backendConfig backendconfig.Backend
 				}
 				repo.Switch(fullRepo)
 				latestSyncCancel()
+				m.Log.Info("Full sync started")
 				fullSyncer.SyncLoop(ctx)
 				err = fullRepo.Stop()
 				if err != nil {
 					m.Log.Warnf("Full Sync failed: could not stop repo: %w", err)
 				}
+				m.Log.Info("Full sync stopped")
```
atzoum, line 126:
```suggestion
```

enterprise/suppress-user/factory.go
```diff
@@ -115,14 +117,17 @@ func (m *Factory) Setup(ctx context.Context, backendConfig backendconfig.Backend
 				}
 				repo.Switch(fullRepo)
 				latestSyncCancel()
+				m.Log.Info("Full sync started")
```
atzoum, line 120:
```suggestion
				repo.Switch(fullRepo)
				m.Log.Info("Switched to full suppression repository")
				latestSyncCancel()
```


We also need a log statement around lines 110-112 to know when the 1st full sync started and ended
```go
m.Log.Info("First full suppression sync started")
...
m.Log.Info("First full suppression sync done")
```

atzoum: We also need a log statement around lines 110-112 to know when the 1st full sync started and ended

```go
m.Log.Info("First full suppression sync started")
...
m.Log.Info("First full suppression sync done")
```

saurav-malani: added.

atzoum (approved):
Thanks!

enterprise/suppress-user/factory.go
```diff
@@ -123,6 +128,7 @@ func (m *Factory) Setup(ctx context.Context, backendConfig backendconfig.Backend
 			})
```
Sidddddarth, line 128: maybe another `stopped` log like you've added for `latestRepo.Stop()` above?
Also not being fully aware, I hope there's distinction between `...sync stopped` and `...sync done` and that they're used appropriately

saurav-malani: `fullRepo` will be stopped only during shutdown of rudder-server. So, I don't think would be of much help. And, the reason we added log for `latestRepo` is because we are swapping `latestRepo` with `fullRepo` once `latestRepo` has all the suppression regulations.

saurav-malani: yes, `sync done` is when we are done with syncing with `suppression-backup-service`. And, `stop` is more like closing or getting rid of it.

Sidddddarth: fair enough.
>fullRepo will be stopped only during shutdown of rudder-server

In any case, it'd be nice to have one imo.

Sidddddarth (approved):

PR #3253, opened 2023-04-28
chore: improve pending events api

achettyiitr:
# Description

We can merge the two queries into a single query that uses a subquery to retrieve the maximum value of end_staging_file_id from the uploads table and then uses it as a parameter to filter the staging_files table.

## Notion Ticket

https://www.notion.so/rudderstacks/Improve-pending-events-api-123bda09cd2e43f5aff949213bfb15e4?pvs=4

## Security


warehouse/internal/repo/staging.go
```diff
@@ -342,23 +342,10 @@ func (repo *StagingFiles) CountPendingForDestination(ctx context.Context, destin
 }
 
 func (repo *StagingFiles) countPending(ctx context.Context, query string, value interface{}) (int64, error) {
-	var id sql.NullInt64
+	var count int64
 	err := repo.db.QueryRowContext(ctx,
-		`SELECT MAX(end_staging_file_id) FROM `+uploadsTableName+` WHERE `+query,
+		`SELECT COUNT(*) FROM `+stagingTableName+` WHERE `+query+` AND id > (SELECT COALESCE(MAX(end_staging_file_id), 0) FROM `+uploadsTableName+` WHERE `+query+`)`,
```
atzoum, line 347: Is `end_staging_file_id` indexed? Should we expect considerable performance degradation with the new query?

achettyiitr: Currently, we don't have indexes present.

atzoum (approved):

fracasula (approved):

PR #3251, opened 2023-04-27
chore: gateway stores singular events

Sidddddarth:
# Description

gateway now stores singular events to jobsdb.

gateway receives:
```
{
  "batch": [
    {event-1},
    {event-2}
  ]
}
```
gateway stores:
```
jobid - 1:
{event-1}

jobid - 2:
{event-2}
```

## Notion Ticket

[relevant slack thread](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1682402259238799)

## Security

- [ ] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #3256, opened 2023-04-29
chore: gateway stores singular event batches

Sidddddarth:
# Description

gateway opens a batch of event, stores each event inside as its own batch.

gateway receives:
```
{
  "batch": [
    {event-1},
    {event-2}
  ]
}
```
gateway stores:
```
jobid - 1:
{
  "batch": [
    {event-1}
  ]
}

jobid - 2:
{
  "batch": [
    {event-2}
  ]
}
```
Tangential to another [PR](https://github.com/rudderlabs/rudder-server/pull/3251) where gateway simply stores the singular event to gateway table.
The current approach avoids changing things at processor, and does not bear the burden of transitioning from previous behaviour which is evident from the other PR.

## Notion Ticket

[relevant slack thread](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1682402259238799)

## Security

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -76,14 +76,6 @@ type webRequestT struct {
 	errors         []string
 }
 
-func (wr *webRequestT) respond() {
-	if len(wr.errors) > 0 {
-		wr.done <- wr.errors[0]
-	} else {
-		wr.done <- ""
-	}
-}
-
 type batchWebRequestT struct {
 	batchRequest []*webRequestT
 }
@@ -490,9 +482,7 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 			} else {
 				sourceStats[sourceTag].RequestEventsSucceeded(len(batch))
 			}
-		}
-		for _, req := range breq.batchRequest {
-			req.respond()
+			jobIDReqMap[batch[0].UUID].done <- err
 		}
 		// Sending events to config backend
 		for _, eventBatch := range eventBatchesToRecord {
@@ -694,14 +684,13 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
 		)
 	}
 	if !allowBatchSplitting {
-		events := lo.FilterMap(out, func(userEvent jobObject, _ int) (map[string]interface{}, bool) {
-			return userEvent.events[0], true
-		})
-		userID := out[0].userID
+		// instead of multiple jobs with one event, create one job with all events
 		out = []jobObject{
 			{
-				userID: userID,
-				events: events,
+				userID: out[0].userID,
+				events: lo.Map(out, func(userEvent jobObject, _ int) map[string]interface{} {
+					return userEvent.events[0]
+				}),
 			},
 		}
 	}
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -448,10 +448,18 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 				}
 				continue
 			}
-			jobList = append(jobList, jobData.job)
-			jobIDReqMap[jobData.job.UUID] = req
-			jobSourceTagMap[jobData.job.UUID] = sourceTag
-			eventBatchesToRecord = append(eventBatchesToRecord, sourceDebugger{data: jobData.job.EventPayload, writeKey: writeKey})
+			jobList = append(jobList, jobData.jobs...)
+			for _, job := range jobData.jobs {
+				jobIDReqMap[job.UUID] = req
+				jobSourceTagMap[job.UUID] = sourceTag
+				eventBatchesToRecord = append(
+					eventBatchesToRecord,
+					sourceDebugger{
+						data:     job.EventPayload,
+						writeKey: writeKey,
+					},
+				)
+			}
 		}
 
 		errorMessagesMap := make(map[uuid.UUID]string)
@@ -463,6 +471,7 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 			errorMessagesMap = <-userWebRequestWorker.reponseQ
 		}
 
+		reqOKMap := make(map[*webRequestT]string)
 		for _, job := range jobList {
 			err, found := errorMessagesMap[job.UUID]
 			sourceTag := jobSourceTagMap[job.UUID]
@@ -471,7 +480,12 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 			} else {
 				sourceStats[sourceTag].RequestEventsSucceeded(job.EventCount)
 			}
-			jobIDReqMap[job.UUID].done <- err
+			if reqOKMap[jobIDReqMap[job.UUID]] == "" {
+				reqOKMap[jobIDReqMap[job.UUID]] = err
+			}
+		}
+		for req, err := range reqOKMap {
+			req.done <- err
 		}
 		// Sending events to config backend
 		for _, eventBatch := range eventBatchesToRecord {
@@ -493,7 +507,7 @@ var (
 )
 
 type jobFromReq struct {
-	job       *jobsdb.JobT
+	jobs      []*jobsdb.JobT
 	numEvents int
 	version   string
 }
@@ -552,10 +566,10 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
 
 	var (
 		// map to hold modified/filtered events of the batch
-		out []map[string]interface{}
+		out = make(map[string][]map[string]interface{})
 
+		marshalledParams []byte
 		// values retrieved from first event in batch
-		firstUserID                                 string
 		firstSourcesJobRunID, firstSourcesTaskRunID string
 
 		// facts about the batch populated as we iterate over events
@@ -583,7 +597,6 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
 		}
 
 		if idx == 0 {
-			firstUserID = buildUserID(userIDHeader, anonIDFromReq, userIDFromReq)
 			firstSourcesJobRunID, _ = misc.MapLookup(
 				toSet,
 				"context",
@@ -636,7 +649,8 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
 			containsAudienceList = true
 		}
 
-		out = append(out, toSet)
+		userID := buildUserID(userIDHeader, anonIDFromReq, userIDFromReq)
+		out[userID] = append(out[userID], toSet)
 	}
 
 	if len(out) == 0 && suppressed {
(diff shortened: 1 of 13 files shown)


gateway/gateway.go
```diff
@@ -447,10 +447,18 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
...
+				jobIDReqMap[job.UUID] = req
+				jobSourceTagMap[job.UUID] = sourceTag
+				eventBatchesToRecord = append(
+					eventBatchesToRecord,
+					sourceDebugger{
+						data:     job.EventPayload,
+						writeKey: writeKey,
+					},
+				)
+			})
```
lvrach, line 461: [minor] What is the benefit of using `lo.ForEach` instead of `for`?

```suggestion
			for _, job := range jobData.jobs {
				jobIDReqMap[job.UUID] = req
				jobSourceTagMap[job.UUID] = sourceTag
				eventBatchesToRecord = append(
					eventBatchesToRecord,
					sourceDebugger{
						data:     job.EventPayload,
						writeKey: writeKey,
					},
				)
			}
```

gateway/gateway.go
```diff
@@ -670,17 +678,40 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
-		WorkspaceId:  workspaceId,
-	}
-	jobData.job = job
+	jobs := make([]*jobsdb.JobT, len(out))
+	for idx, event := range out {
+		var payload json.RawMessage
+		{
+			type SingularEventBatch struct {
+				Batch      []types.SingularEventT `json:"batch"`
+				RequestIp  string                 `json:"requestIP"`
```
lvrach, line 687: [minor]
```suggestion
				RequestIP  string                 `json:"requestIP"`
```

gateway/gateway.go
```diff
@@ -670,17 +678,40 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
+		}
+
+		jobs[idx] = &jobsdb.JobT{
+			UUID:         uuid.New(),
+			UserID:       firstUserID,
+			Parameters:   marshalledParams,
+			CustomVal:    CustomVal,
+			EventPayload: payload,
+			EventCount:   1,
+			WorkspaceId:  workspaceId,
```
lvrach, line 711: Should we also populate the payload size field?
```suggestion
			EventCount:   1,
			PayloadSize: len(payload),
			WorkspaceId:  workspaceId,
```

Also, we might want to populate the userID of the payload instead of `firstUserID`. Although it breaks compatibility with our current approach and might introduce performance downsides.
.

gateway/gateway.go
```diff
@@ -470,7 +479,12 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 			} else {
 				sourceStats[sourceTag].RequestEventsSucceeded(job.EventCount)
 			}
-			jobIDReqMap[job.UUID].done <- err
+			if reqOKMap[jobIDReqMap[job.UUID]] == "" {
+				reqOKMap[jobIDReqMap[job.UUID]] = err
+			}
+		}
+		for req, err := range reqOKMap {
+			req.done <- err
```
lvrach, line 487: can you explain the logic/motivation behind this?

Sidddddarth: Each request can have multiple events.
We only want one message to be posted on the request's done channel, it being `err` if at least one event(now a `*jobsb.JobT`) fails storing. or an `OK` in case all events(`[]*jobsdb.JobT`) have stored successfully.

gateway/gateway.go
Sidddddarth: True
We consider each individual event's userID for suppression purposes anyway.
Elaborate further on the performance downsides.. I don't see a drastic effect due to this at all.

gateway/gateway.go
Sidddddarth: no benefit really, I've just become used to it.

gateway/gateway.go
Sidddddarth: adopted👍

atzoum:
Would it be easy to introduce this behaviour behind a configuration flag (enabled by default), i.e. be able to disable it in case we observed any undesired side-effects?

gateway/gateway.go
atzoum: Either all or no jobs from the batch should be stored in case of an error. We might require adaptations in `internalStoreWithRetryEachInTx` for this

gateway/gateway.go
```diff
@@ -670,17 +677,42 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
+					Batch:      []types.SingularEventT{event},
+					RequestIP:  ipAddr,
+					WriteKey:   writeKey,
+					ReceivedAt: time.Now().Format(misc.RFC3339Milli),
+				}
+				payload, err = json.Marshal(singularEventBatch)
+				if err != nil {
+					err = errors.New(response.InvalidJSON)
+					return
+				}
```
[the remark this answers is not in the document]
Sidddddarth: true
`panic`ing now

gateway/gateway.go
```diff
@@ -481,6 +502,8 @@ func (gateway *HandleT) userWebRequestWorkerProcess(userWebRequestWorker *userWe
 		gateway.batchSizeStat.Observe(float64(len(breq.batchRequest)))
 
 		for _, v := range sourceStats {
+			gateway.logger.Info(`@@@@@@@`)
+			gateway.logger.Info(v)
```
lvrach, line 506: is this a debug artefact?

```suggestion
```

jobsdb/jobsdb.go
```diff
@@ -2981,6 +2984,95 @@ func (jd *HandleT) StoreWithRetryEachInTx(ctx context.Context, tx StoreSafeTx, j
...
+		}
+		res, _ = jd.executeDbRequest(newWriteDbRequest("store_each_batch_retry", nil, command)).(map[uuid.UUID]string)
+		return err
+	}
+	if tx.storeSafeTxIdentifier() != jd.Identifier() {
+		_ = jd.inStoreSafeCtx(ctx, storeCmd)
+		return res, err
+	}
+	_ = storeCmd()
+	return res, err
```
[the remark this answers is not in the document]
Sidddddarth: assigning the error returned below to `err`(line#2999)
`res, err = jd.internalStoreEachBatchRetryInTx(ctx, tx.Tx(), dsList[len(dsList)-1], jobBatches)`

jobsdb/jobsdb.go
```diff
@@ -235,6 +235,9 @@ type JobsDB interface {
 	//    })
 	StoreWithRetryEachInTx(ctx context.Context, tx StoreSafeTx, jobList []*JobT) (map[uuid.UUID]string, error)
```
atzoum, line 236: `StoreWithRetryEachInTx` & `StoreWithRetryEach` are being replaced by `StoreEachBatchRetryInTx`? Why aren't we removing the previoue methods?

atzoum: why didn't we keep the old methods and change their signature, since their only purpose was to serve gateway anyway :)

jobsdb/jobsdb.go
```diff
@@ -2983,6 +2986,95 @@ func (jd *HandleT) StoreWithRetryEachInTx(ctx context.Context, tx StoreSafeTx, j
...
+		return
+	}
+	if errors.Is(err, errStaleDsList) {
+		return nil, err
+	}
+	_, err = tx.ExecContext(ctx, rollbackSql)
+	if err != nil {
+		return failAll(err), nil
+	}
+	jd.logger.Errorf("Copy In command failed with error %v", err)
```
atzoum, line 3050: do we need this?

gateway/gateway.go
```diff
@@ -73,6 +73,15 @@ type webRequestT struct {
...
 	userIDHeader   string
+	errors         []string
+}
+
+func (wr *webRequestT) respond() {
+	if len(wr.errors) > 0 {
+		wr.done <- wr.errors[0]
+	} else {
+		wr.done <- ""
+	}
```
atzoum, line 84: I notice you are performing some special handling in jobsdb, where you only include an error for the 1st job of the batch (not sure I like it). In that context, is this still necessary?

Sidddddarth: >include an error for the 1st job of the batch

this felt simpler. I thought of introducing a jobsdb.JobBatch kind of struct, but it'd have been more work, and could be taken up at a later point considering I didn't any breaking issue going with the first job's UUID like I've done here.

jobsdb/jobsdb.go
Sidddddarth: can be removed yeah.
I will 👍

gateway/gateway.go
atzoum: This detail definitely needs to be included in the josbdb method's doc, since one would naturally expect that all errors for all uuids would be returned, instead of the first one.

gateway/gateway.go
```diff
@@ -671,17 +693,55 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
-		CustomVal:    CustomVal,
-		EventPayload: body,
-		EventCount:   jobData.numEvents,
-		WorkspaceId:  workspaceId,
-	}
-	jobData.job = job
+	if !allowBatchSplitting {
+		events := lo.FilterMap(out, func(userEvent jobObject, _ int) (map[string]interface{}, bool) {
+			return userEvent.events[0], true
+		})
```
atzoum, line 699:
```suggestion
		events := lo.Map(out, func(userEvent jobObject, _ int) (map[string]interface{}) {
			return userEvent.events[0]
		})
```

gateway/gateway.go
```diff
@@ -671,17 +693,55 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
...
+		userID := out[0].userID
+		out = []jobObject{
+			{
+				userID: userID,
+				events: events,
+			},
+		}
```
atzoum, line 706:
```suggestion
		out = []jobObject{
			{
				userID: out[0].userID,
				events: lo.Map(out, func(userEvent jobObject, _ int) map[string]interface{} {
					return userEvent.events[0]
				}),
			},
		}
```

gateway/gateway.go
Sidddddarth: yeah, realised that above. Adding that part in docs right now.

jobsdb/jobsdb.go
```diff
@@ -2983,6 +2904,70 @@ func (jd *HandleT) StoreWithRetryEachInTx(ctx context.Context, tx StoreSafeTx, j
...
+			errorMessagesMap[jobBatches[i][0].UUID] = err.Error()
+		}
+		return errorMessagesMap
+	}
+	defer jd.getTimerStat("store_jobs_retry_each_batch", nil).RecordDuration()()
+	_, err = tx.ExecContext(ctx, savepointSql)
+	if err != nil {
+		return failAll(err), nil
+	}
+	err = jd.internalStoreJobsInTx(ctx, tx, ds, lo.Flatten(jobBatches))
```
atzoum, line 2931: one side-effect of using this here is that the `store_jobs` metric will be published additionally to `store_jobs_retry_each`

Sidddddarth: Thankfully because the `table_prefix` is only `gateway` where this happens, we can excuse this(at least while inferring job storage times from the stats, but redundant stats would still exist).
We could pull out the stat part inside `internalStoreJobsInTx`, and have the caller publish said stat(maybe a wrapper?)

atzoum (approved):

gateway/gateway_test.go
```diff
@@ -496,7 +497,9 @@ var _ = Describe("Gateway", func() {
 
 		// common tests for all web handlers
 		It("should accept valid requests on a single endpoint (except batch), and store to jobsdb", func() {
-			workspaceID := getWorkspaceID(WriteKeyEnabled)
+			workspaceID := gateway.getWorkspaceForWriteKey(WriteKeyEnabled)
+			fmt.Println(`################## - `, workspaceID, ` - ##################`)
+			fmt.Println(workspaceID)
 			for handlerType, handler := range allHandlers(gateway) {
```
fracasula, line 503:
```suggestion
			for handlerType, handler := range allHandlers(gateway) {
```

fracasula (approved):
I'm not too comfortable with code that access slice elements returned by other functions without checking for its length, but if you're positive there's always at least one element when `err` is not `nil` then go ahead.

i.e. when you do `jobIDReqMap[jobData.jobs[0].UUID] = req` and `jobData` is returned by `gateway.getJobDataFromRequest(req)`.

Sidddddarth:
> I'm not too comfortable with code that access slice elements returned by other functions without checking for its length, but if you're positive there's always at least one element when `err` is not `nil` then go ahead.
> 
> i.e. when you do `jobIDReqMap[jobData.jobs[0].UUID] = req` and `jobData` is returned by `gateway.getJobDataFromRequest(req)`.

True.
Returning `Invalid JSON` if length is 0.

PR #3272, opened 2023-05-04
chore: fix deadlock issue uploader

cisse21:
# Description
Fixes the Deadlock described below as described clearly by @atzoum 

- goroutine-1 acquires a read lock [here](https://github.com/rudderlabs/rudder-server/blob/0ad29a3222b68d7cb82a6395672b91f12d363840/services/debugger/destination/eventDeliveryStatusUploader.go#L127) successfully!
- goroutine-2 tries to acquire a full lock [here](https://github.com/rudderlabs/rudder-server/blob/0ad29a3222b68d7cb82a6395672b91f12d363840/services/debugger/destination/eventDeliveryStatusUploader.go#L185), but is waiting since goroutine-1 already has a read lock.
- goroutine-1 proceeds and calls HasUploadEnabled, which in turn tries to acquire a read lock again [here](https://github.com/rudderlabs/rudder-server/blob/0ad29a3222b68d7cb82a6395672b91f12d363840/services/debugger/destination/eventDeliveryStatusUploader.go#L143). This read lock is now blocked by the full lock request issued by goroutine-2 earlier.
- deadlock!

## Notion Ticket

[ Notion Link ](https://www.notion.so/rudderstacks/Fix-deadlock-eventUploader-cf0be292f6484fe5ad313ad5f735ed06?pvs=4)

## Security

### DIFF
--- a/services/debugger/destination/eventDeliveryStatusUploader.go
+++ b/services/debugger/destination/eventDeliveryStatusUploader.go
@@ -124,8 +124,6 @@ func (h *Handle) RecordEventDeliveryStatus(destinationID string, deliveryStatus 
 	}
 	<-h.initialized
 	// Check if destinationID part of enabled destinations, if not then push the job in cache to keep track
-	h.uploadEnabledDestinationIDsMu.RLock()
-	defer h.uploadEnabledDestinationIDsMu.RUnlock()
 	if !h.HasUploadEnabled(destinationID) {
 		err := h.eventsDeliveryCache.Update(destinationID, deliveryStatus)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


atzoum:
@cisse21 can you please review the other 2 debuggers (source & transformation) too for similar issues?

lvrach (approved):

cisse21:
> can you please review the other 2 debuggers (source & transformation) too for similar issues?

I already reviewed the other two and they seemed fine to me but I will have a look at them again

atzoum (approved):
thanks!

PR #3273, opened 2023-05-04
chore: added threshold for commit for sqlwrapper

achettyiitr:
# Description

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/warehouse/integrations/middleware/sqlquerywrapper/sql.go
+++ b/warehouse/integrations/middleware/sqlquerywrapper/sql.go
@@ -28,6 +28,7 @@ type DB struct {
 	keysAndValues      []any
 	slowQueryThreshold time.Duration
 	rollbackThreshold  time.Duration
+	commitThreshold    time.Duration
 	secretsRegex       map[string]string
 }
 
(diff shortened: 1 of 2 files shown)


lvrach (approved):

warehouse/integrations/middleware/sqlquerywrapper/sql.go
```diff
@@ -217,3 +219,12 @@ func (tx *Tx) Rollback() error {
 	}
 	return err
 }
+
+func (tx *Tx) Commit() error {
+	startedAt := time.Now()
+	err := tx.Tx.Rollback()
```
fracasula, line 225: Rollback? :sweat_smile:

fracasula (changes requested):

warehouse/integrations/middleware/sqlquerywrapper/sql.go
fracasula: I guess this could have been easily spotted if we had a spy in our test and included an expectation on `Commit` being called.

fracasula (approved):

warehouse/integrations/middleware/sqlquerywrapper/sql.go
achettyiitr: Added a test for round trip?

PR #3279, opened 2023-05-04
feat: wh Endpoint To Fetch Tables Per Connection

am6010:
# Description

A new endpoint to get all the tables for all the connections.
`POST /v1/warehouse/fetch-tables`
Request Body
```
{
  "connections":  [
    {
      "source_id": "<source id>",
      "destination_id": "<destination id>"
    },
   {
     "source_id": "<source_id>",
     "destination_id": "<destination_id>"
   },
  ...
  ]
}
```

Response Body 
```
{
  "tables_per_connection":  [
    {
      "source_id": "<source id>",
      "destination_id": "<destination id>",
      "tables": ["table1", "table2", .... ]
    },
   {
     "source_id": "<source_id>",
     "destination_id": "<destination_id>",
     "tables": ["table1", "table2", .... ]
   },
  ...
  ]
}
```

Tested: 
- add some unit tests for the function `GetTablesForConnection`
- called the endpoint manually.
```
curl -X GET http://localhost:8082/v1/warehouse/fetch-tables \
   -H 'Content-Type: application/json' \
   -d '{"connections": []}'

curl -X GET http://localhost:8082/v1/warehouse/fetch-tables \
   -H 'Content-Type: application/json' \
   -d '{"connections": [{"source_id": "2PGrgf2kZOs1WlK5wBfi2Zes3f0", "destination_id":"1yrEf5DSPKEZLuUcFAmoGRWttZG"}, 
   {"source_id":"2PDxn82icdBJS1TlXk05OSvtzzN", "destination_id":"1yrEf5DSPKEZLuUcFAmoGRWttZG"}]}'
```

## Notion Ticket

[notion ticket](https://www.notion.so/rudderstacks/Filter-Library-Project-Based-On-Metadata-and-Customer-Connections-aa41f8097f4a48e184d771c56fee2c2e?pvs=4#45c59c87b136418db2de5e797d047d9d)

## Security


warehouse/internal/repo/schema.go
```diff
@@ -192,3 +193,56 @@ func (repo *WHSchema) GetNamespace(ctx context.Context, sourceID, destID string)
...
+	if len(connections) == 0 {
+		return nil, fmt.Errorf("no source id and destination id pairs provided")
+	}
+
+	// select max id for each source id and destination id pair
+	innerQueryTemplate := `SELECT max(id) FROM ` + whSchemaTableName + `
+		WHERE
+			(source_id, destination_id) IN (%s)
+		AND
+			namespace <> ''
```
lvrach, line 207: [Question] why do you ignore empty namespaces ?

warehouse/internal/repo/schema.go
```diff
@@ -192,3 +193,56 @@ func (repo *WHSchema) GetNamespace(ctx context.Context, sourceID, destID string)
...
+		GROUP BY id
+	`
+	sourceIdDestinationIdArr := make([]string, len(connections))
+	for idx, connection := range connections {
+		pair := fmt.Sprintf("'%s','%s'", connection.SourceId, connection.DestinationId)
```
lvrach, line 212: Although it is used by an internal endpoint now, I would prefer to write this query in a way to avoid SQL injections. We are not sure how this repository method can be used in the future.

I haven't tried `pq.Array` with nested elements, but it might work.

Another approach could be to compare a concatenated version of them, with single dimension pq.Array

warehouse/warehouse.go
```diff
@@ -1543,6 +1581,9 @@ func startWebHandler(ctx context.Context) error {
 			srvMux.HandleFunc("/v1/warehouse/jobs", asyncWh.AddWarehouseJobHandler).Methods("POST")          // FIXME: add degraded mode
 			srvMux.HandleFunc("/v1/warehouse/jobs/status", asyncWh.StatusWarehouseJobHandler).Methods("GET") // FIXME: add degraded mode
 
+			// fetch schema info
+			srvMux.HandleFunc("/v1/warehouse/fetch-tables", fetchTablesHandler).Methods("POST")
```
lvrach, line 1585: would get be more appropriate here?
```suggestion
			srvMux.HandleFunc("/v1/warehouse/fetch-tables", fetchTablesHandler).Methods("GET")
```

warehouse/utils/utils.go
```diff
@@ -305,6 +305,20 @@ type TriggerUploadRequest struct {
 	DestinationID string `json:"destination_id"`
 }
 
+type Connection struct {
+	SourceId      string   `json:"source_id"`
+	DestinationId string   `json:"destination_id"`
```
lvrach, line 310: minor: for go idiomatic reasons and consistency
```suggestion
	SourceID      string   `json:"source_id"`
	DestinationID string   `json:"destination_id"`
```

warehouse/utils/utils.go
```diff
@@ -305,6 +305,20 @@ type TriggerUploadRequest struct {
...
+	Tables        []string `json:"tables"`
+}
+
+type ConnectionsTablesRequest struct {
+	Connections []Connection `json:"connections"`
+}
```
achettyiitr, line 316: Let's not use Connection here it gives the impression that Tables are passed as requests but we are using it only for response.

achettyiitr:
```GO

type SourceIDDestinationID struct {
	SourceID      string `json:"source_id"`
	DestinationID string `json:"destination_id"`
}

type ConnectionsTablesRequest struct {
	Connections []SourceIDDestinationID `json:"connections"`
}
``

File: warehouse/utils/utils.go [hunk 6]
```diff
@@ -305,6 +305,20 @@ type TriggerUploadRequest struct {
...
+
+type ConnectionsTablesResponse struct {
```
achettyiitr, line 318: Also the response,

```GO
type FetchTableInfo struct {
	SourceID      string   `json:"source_id"`
	DestinationID string   `json:"destination_id"`
	Tables        []string `json:"tables"`
}

type FetchTablesResponse struct {
	ConnectionsTables []FetchTableInfo `json:"tables_per_connection"`
}
```

warehouse/internal/repo/schema.go
achettyiitr: +1

achettyiitr: Can we try something like this?

```GO
func (repo *WHSchema) GetTablesForConnection(ctx context.Context, connections []warehouseutils.SourceIDDestinationID) ([]warehouseutils.FetchTableInfo, error) {
	if len(connections) == 0 {
		return nil, fmt.Errorf("no source id and destination id pairs provided")
	}

	sourceIDDestinationIDPairs := make([]string, len(connections))
	var parameters []interface{}
	for i, conn := range connections {
		sourceIDDestinationIDPairs[i] = fmt.Sprintf("($%d, $%d)", i*2+1, i*2+2)
		parameters = append(parameters, conn.SourceID, conn.DestinationID)
	}

	// select all rows with max id for each source id and destination id pair
	query := `SELECT` + whSchemaTableColumns + `FROM ` + whSchemaTableName + `
		WHERE (source_id, destination_id) IN (` + strings.Join(sourceIDDestinationIDPairs, ", ") + `)`

	rows, err := repo.db.QueryContext(
		ctx,
		query,
		parameters...
	)
	if err != nil {
		return nil, fmt.Errorf("querying schema: %w", err)
	}

	entries, err := repo.parseRows(rows)
	if err != nil {
		return nil, fmt.Errorf("parsing rows: %w", err)
	}

	var tables []warehouseutils.FetchTableInfo
	for _, entry := range entries {
		var allTablesOfConnections []string
		for tableName := range entry.Schema {
			allTablesOfConnections = append(allTablesOfConnections, tableName)
		}
		tables = append(tables, warehouseutils.FetchTableInfo{
			SourceID:      entry.SourceID,
			DestinationID: entry.DestinationID,
			Tables:        allTablesOfConnections,
		})
	}

	return tables, nil
}
```

warehouse/internal/repo/schema.go
```diff
@@ -192,3 +193,56 @@ func (repo *WHSchema) GetNamespace(ctx context.Context, sourceID, destID string)
...
+		}
+		tables = append(tables, warehouseutils.Connection{
+			SourceId:      entry.SourceID,
+			DestinationId: entry.DestinationID,
+			Tables:        allTablesOfConnections,
+		})
+	}
+
+	return tables, nil
+}
```
achettyiitr, line 248: Can we try something like this?


```GO
func (repo *WHSchema) GetTablesForConnection(ctx context.Context, connections []warehouseutils.SourceIDDestinationID) ([]warehouseutils.FetchTableInfo, error) {
	if len(connections) == 0 {
		return nil, fmt.Errorf("no source id and destination id pairs provided")
	}

	sourceIDDestinationIDPairs := make([]string, len(connections))
	var parameters []interface{}
	for i, conn := range connections {
		sourceIDDestinationIDPairs[i] = fmt.Sprintf("($%d, $%d)", i*2+1, i*2+2)
		parameters = append(parameters, conn.SourceID, conn.DestinationID)
	}

	// select all rows with max id for each source id and destination id pair
	query := `SELECT` + whSchemaTableColumns + `FROM ` + whSchemaTableName + `
		WHERE (source_id, destination_id) IN (` + strings.Join(sourceIDDestinationIDPairs, ", ") + `)`

	rows, err := repo.db.QueryContext(
		ctx,
		query,
		parameters...
	)
	if err != nil {
		return nil, fmt.Errorf("querying schema: %w", err)
	}

	entries, err := repo.parseRows(rows)
	if err != nil {
		return nil, fmt.Errorf("parsing rows: %w", err)
	}

	var tables []warehouseutils.FetchTableInfo
	for _, entry := range entries {
		var allTablesOfConnections []string
		for tableName := range entry.Schema {
			allTablesOfConnections = append(allTablesOfConnections, tableName)
		}
		tables = append(tables, warehouseutils.FetchTableInfo{
			SourceID:      entry.SourceID,
			DestinationID: entry.DestinationID,
			Tables:        allTablesOfConnections,
		})
	}

	return tables, nil
}
```

warehouse/internal/repo/schema.go
am6010: My idea was that if the namespace was empty the schema column will be empty as well. 
I have changed this condition to a direct check `schema::text <> '{}'::text`

warehouse/internal/repo/schema.go
am6010: @lvrach  @achettyiitr Thank you for reviewing
I have followed the suggestion from @achettyiitr to avoid SQL injection.

warehouse/internal/repo/schema.go
achettyiitr: namespace will never be empty i think

achettyiitr (approved):

lvrach (approved):

PR #3282, opened 2023-05-05
chore: enable setting of `embedAvroSchemaID` from config

saurav-malani:
# Description

Enable setting of `embedAvroSchemaID` from config in addition to setting it via env.

### DIFF
--- a/services/streammanager/kafka/kafkamanager.go
+++ b/services/streammanager/kafka/kafkamanager.go
@@ -32,17 +32,18 @@ type avroSchema struct {
 
 // configuration is the config that is required to send data to Kafka
 type configuration struct {
-	Topic         string
-	HostName      string
-	Port          string
-	SslEnabled    bool
-	CACertificate string
-	UseSASL       bool
-	SaslType      string
-	Username      string
-	Password      string
-	ConvertToAvro bool
-	AvroSchemas   []avroSchema
+	Topic             string
+	HostName          string
+	Port              string
+	SslEnabled        bool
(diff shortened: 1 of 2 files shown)


fracasula (approved):

BonapartePC (approved):

services/streammanager/kafka/kafkamanager.go
```diff
@@ -362,8 +363,10 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (*Produ
 		return nil, err
 	}
 
-	// @TODO embedAvroSchemaID should come from control plane (i.e. destination config)
 	embedAvroSchemaID := config.GetBool("ROUTER_KAFKA_EMBED_AVRO_SCHEMA_ID_"+strings.ToUpper(destination.ID), false)
+	if !embedAvroSchemaID {
```
atzoum, line 367: The default value is false right? Wouldn't that break backwards compatibility?

fracasula: It's OK we have this inside a Kubernetes secret only. If you pass `true` in the Kubernetes secret that is going to be honoured and we won't read from control plane.

If you pass `false` or omit it then we try to read from control plane. If you get `true` from control plane it means that the "new way" of configuring this is live so it's OK.

The only ones that have this as `true` is Cars on their `dev` and they are waiting for this to be configurable from CP so it's OK :+1:

atzoum (approved):

PR #3319, opened 2023-05-12
chore: support to skip caching backendConfig

cisse21:
# Description

Adding the support to be able to skip caching backendConfig based on an environment variable

## Notion Ticket

[Notion Link](https://www.notion.so/rudderstacks/Skip-caching-backend-config-56fe33e447f44c44b917689742e81378?pvs=4)

## Security

### DIFF
--- a/backend-config/backend-config.go
+++ b/backend-config/backend-config.go
@@ -1,7 +1,7 @@
 package backendconfig
 
-// go:generate mockgen -destination=../mocks/backend-config/mock_backendconfig.go -package=mock_backendconfig github.com/rudderlabs/rudder-server/backend-config BackendConfig
-// go:generate mockgen -destination=./mock_workspaceconfig.go -package=backendconfig -source=./backend-config.go workspaceConfig
+//go:generate mockgen -destination=../mocks/backend-config/mock_backendconfig.go -package=mock_backendconfig github.com/rudderlabs/rudder-server/backend-config BackendConfig
+//go:generate mockgen -destination=./mock_workspaceconfig.go -package=backendconfig -source=./backend-config.go workspaceConfig
 
 import (
 	"context"
@@ -71,7 +71,7 @@ type BackendConfig interface {
 	WaitForConfig(ctx context.Context)
 	Subscribe(ctx context.Context, topic Topic) pubsub.DataChannel
 	Stop()
-	StartWithIDs(ctx context.Context, workspaces string)
+	StartWithIDs(ctx context.Context, workspaces string, canUseCache bool)
 }
 
 type backendConfigImpl struct {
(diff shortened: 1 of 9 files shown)


backend-config/backend-config.go
```diff
@@ -71,7 +71,7 @@ type BackendConfig interface {
 	WaitForConfig(ctx context.Context)
 	Subscribe(ctx context.Context, topic Topic) pubsub.DataChannel
 	Stop()
-	StartWithIDs(ctx context.Context, workspaces string)
+	StartWithIDs(ctx context.Context, workspaces string, canUseCache bool)
```
atzoum, line 74: this looks too obtrusive

backend-config/backend-config.go
```diff
@@ -340,6 +340,8 @@ func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string
 				bc.cache = &noCache{}
 			}
 		}
+	} else {
```
achettyiitr, line 343: else needs to be done only in case of canUseCache as false right?

backend-config/backend-config.go
```diff
@@ -312,14 +312,14 @@ func Setup(configEnvHandler types.ConfigEnvI) (err error) {
...
-func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string) {
+func (bc *backendConfigImpl) StartWithIDs(ctx context.Context, workspaces string, canUseCache bool) {
 	var err error
 	ctx, cancel := context.WithCancel(ctx)
 	bc.ctx = ctx
 	bc.cancel = cancel
 	bc.blockChan = make(chan struct{})
 	bc.cache = cacheOverride
-	if bc.cache == nil {
+	if bc.cache == nil && canUseCache {
```
atzoum, line 322: we already have a `noCache` implementation, why don't we want to use this?

backend-config/backend-config.go
achettyiitr: Otherwise, it will override the cache when it it initialized and canUseCache is set to true?

backend-config/backend-config.go
cisse21: So another way would be to read the config in the `StartWithIDs` function right? Or do you something else in mind?

backend-config/backend-config.go
```diff
@@ -1,7 +1,7 @@
 package backendconfig
 
-// go:generate mockgen -destination=../mocks/backend-config/mock_backendconfig.go -package=mock_backendconfig github.com/rudderlabs/rudder-server/backend-config BackendConfig
-// go:generate mockgen -destination=./mock_workspaceconfig.go -package=backendconfig -source=./backend-config.go workspaceConfig
+//go:generate mockgen -destination=../mocks/backend-config/mock_backendconfig.go -package=mock_backendconfig github.com/rudderlabs/rudder-server/backend-config BackendConfig
+//go:generate mockgen -destination=./mock_workspaceconfig.go -package=backendconfig -source=./backend-config.go workspaceConfig
 
 import (
```
atzoum, line 6: Do we need anything more than adding the following in `loadConfig`?

```go
	if config.GetBool("CONFIG_BACKEND_DISABLE_CACHE", false) {
		disableCache()
	}
```

PR #3326, opened 2023-05-12
chore: pass warehouse context

achettyiitr:
# Description

- Pass context from UploadJob down to the Integrations package. One context is created during the UploadJob and passed down to the integrations.

## Notion Ticket

https://www.notion.so/rudderstacks/make-sure-context-works-as-intended-1d128f7c18fa49acaf007ab82b20394b?pvs=4

## Security


warehouse/api.go
```diff
@@ -874,20 +875,20 @@ func validateObjectStorage(ctx context.Context, request *ObjectStorageValidation
...
-	overrideWithEnv(settings)
+	overrideWithEnv(ctx, settings)
 	return settings
 }
 
 // overrideWithEnv overrides the config keys in the fileManager settings
 // with fallback values pulled from env. Only supported for S3 for now.
-func overrideWithEnv(settings *filemanager.SettingsT) {
-	envConfig := filemanager.GetProviderConfigFromEnv(context.TODO(), settings.Provider)
+func overrideWithEnv(ctx context.Context, settings *filemanager.SettingsT) {
```
fracasula, line 890: The context seems to be passed to `backendConfig.WaitForConfig()` only. What if it gets canceled or it times out though?  Do we want `GetProviderConfigFromEnv` to return an error?

warehouse/admin.go
```diff
@@ -78,7 +79,7 @@ func (*WarehouseAdmin) Query(s QueryInput, reply *warehouseutils.QueryResult) er
 	if err != nil {
 		return err
 	}
-	client, err := whManager.Connect(warehouse)
+	client, err := whManager.Connect(context.TODO(), warehouse)
```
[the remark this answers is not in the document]
achettyiitr: It's used from rudder-cli. 
https://github.com/rudderlabs/rudder-cli/blob/master/warehouse/warehouse.go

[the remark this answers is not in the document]
achettyiitr: It's being used over rpc.

warehouse/api.go
cisse21: Do we even need a context propagated to this function?

fracasula: I guess we do, in case of cancelation it would stop waiting for the backend config which is a blocking call, right?

cisse21: Yeah true but that part of code kinda feels weird to me and probably something which no one would ever use.

Because from the UI he won't even be able to create a S3 destination without filling the required fields but we are waiting for config to get the workspaceID for a destination which isn't there. Not sure if I put it out here properly

achettyiitr: @fracasula I am also aligned with passing the context.

fracasula (approved):

PR #3338, opened 2023-05-17
fix: include source definition type when destination has UT

sanpj2292:
# Description

Source definition type was being sent as empty string("") when user transformation is connected to a destination.
As `SourceDefinitionType` was not being populated in the metadata(`commonMetadata`) before sending to user transformation.

We are aiming to fix.

## Notion Ticket

https://www.notion.so/rudderstacks/GA4-page-calls-error-is-displaying-in-live-events-tab-for-hybrid-mode-connection-52c58ee1924941f1b037fb8f63fc2445?pvs=4

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -2052,14 +2052,15 @@ func (proc *Handle) transformSrcDest(
 	destination := &eventList[0].Destination
 	workspaceID := eventList[0].Metadata.WorkspaceID
 	commonMetaData := &transformer.MetadataT{
-		SourceID:        sourceID,
-		SourceType:      eventList[0].Metadata.SourceType,
-		SourceCategory:  eventList[0].Metadata.SourceCategory,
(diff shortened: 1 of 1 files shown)


sanpj2292:
The destination(GA4) is operating in `Hybrid` mode

We can see `page` & `track` in source
Screenshot 2023-05-17 at 07 53 36

But at GA4, we can see only `track` & `page` event is getting filtered out (expected)
Screenshot 2023-05-17 at 07 53 43

cisse21 (approved):

koladilip (approved):

sanpj2292:
> Can we add a test case for this failure scenario? The test can be added on different branches based on the master branch.

Sure, will look to add test-case for this scenario in a different branch based out of `master`

lvrach (approved):

PR #3351, opened 2023-05-18
chore: gateway health endpoint should return 503 if db is down

BonapartePC:
# Description

`/health` returns 503 if db is down

## Notion Ticket

https://www.notion.so/rudderstacks/Improve-gateway-health-endpoint-86bae24700cc424ca7e44d7b9cde4bd9

## Security

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -138,14 +138,20 @@ func New(options *Options) App {
 // LivenessHandler is the http handler for the Kubernetes liveness probe
 func LivenessHandler(jobsDB jobsdb.JobsDB) http.HandlerFunc {
 	return func(w http.ResponseWriter, _ *http.Request) {
-		_, _ = w.Write([]byte(getHealthVal(jobsDB)))
+		healthy, responsePayload := getHealthVal(jobsDB)
+		if !healthy {
+			http.Error(w, "Cannot connect to db", http.StatusServiceUnavailable)
+			return
+		}
+		_, _ = w.Write([]byte(responsePayload))
 	}
 }
 
-func getHealthVal(jobsDB jobsdb.JobsDB) string {
+func getHealthVal(jobsDB jobsdb.JobsDB) (bool, string) {
 	dbService := "UP"
 	if jobsDB.Ping() != nil {
 		dbService = "DOWN"
+		return false, ""
 	}
 	enabledRouter := "TRUE"
 	if !config.GetBool("enableRouter", true) {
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -138,14 +138,20 @@ func New(options *Options) App {
 // LivenessHandler is the http handler for the Kubernetes liveness probe
 func LivenessHandler(jobsDB jobsdb.JobsDB) http.HandlerFunc {
 	return func(w http.ResponseWriter, _ *http.Request) {
-		_, _ = w.Write([]byte(getHealthVal(jobsDB)))
+		healthy, responsePayload := getHealthVal(jobsDB)
+		if !healthy {
+			http.Error(w, "Cannot connect to db", http.StatusServiceUnavailable)
+			return
+		}
+		_, _ = w.Write([]byte(responsePayload))
```
atzoum, line 146: What are we doing to introduce this change gracefully?

cisse21: `    Liveness:   tcp-socket :8080 delay=120s timeout=1s period=10s #success=1 #failure=3`

There is nothing to be changed with respect to probes as we are not concerned about the health end point but this will help with SemaText alerts which is an external service pinging the data-plane. We can discuss at later point in time on how to combine the liveness and readiness probes with the health endpoint

atzoum: please elaborate, I fail to understand the purpose of this change

BonapartePC: With this change, we want to improve the health endpoint to return 200 only when the db is up. We are also planning to add this to readiness probes so that traffic won't be routed unless the db is up

atzoum: the ticket mentions other things as well, e.g. jobsdb is ready as well (tables are there in case of writer)
I need some lessons on readiness & liveliness probes

BonapartePC: You can read a bit about probes here: https://medium.com/@AADota/kubernetes-liveness-and-readiness-probes-difference-1b659c369e17
I have a set up a meeting tomorrow to discuss this

atzoum (commented):
I expect to see another PR which will be changing readiness and liveliness probes in rudderstack-operator

BonapartePC:
> I expect to see another PR which will be changing readiness and liveliness probes in rudderstack-operator

Operator PR: https://github.com/rudderlabs/rudderstack-operator/pull/1501

atzoum (approved):

cisse21 (approved):

PR #3354, opened 2023-05-18
fix: added retry to avoid gateway from crash because of transient error

saurav-malani:
# Description

Added retry mechanism in jobsDB queries to make sure that `gateway` doesn't crashes in case of any transient error because of which `postgres` is temporarily unreachable.

NOTE: Retry mechanism is added only to the parts that `gw` might touch. 

## Notion Ticket

https://www.notion.so/rudderstacks/Gateway-shall-not-panic-on-transient-DNS-errors-ff53bcb47f6045b1a409aecea927841c?pvs=4


atzoum:
I have a different opinion with respect to how we should approach this problem. I believe it is best to treat it as a feature instead of a fix, i.e. add support for jobsdb's maintenance routines to gracefully handle database unavailability:

atzoum:
I have a different opinion with respect to how we should approach this problem. I believe it is best to treat it as a larger feature rather than a hotfix, i.e. add support for jobsdb's maintenance routines to gracefully handle database's unavailability:
- backup loop
- migration loop
- add ds loop

So, instead of adding exponential backoffs to all our database calls that are relevant to the aforementioned routines, I would prefer us refactoring relevant functions to propagate such errors (instead of panicking themselves). Then its will be up to the calling routine how it will treat such errors, e.g. recover, retry or event panic.

For instance, gw jobsdb could be configured to ignore maintenance routine failures, whereas other jobsdbs shouldn't

PR #3378, opened 2023-05-23
fix: external table support for databricks validations

achettyiitr:
# Description

- In case of databricks external table, since we are using the same table and performing alter on it, We can't create the table again because there will be a mismatch in the schema. It is basically returning the following error:
  ```
  The specified schema does not match the existing schema at s3://acorns-rudderstack-s3-sink-staging/rudder/delta/tables/rudderstack_setup_test/setup_test_staging.
  
  == Specified ==
  root
  |-- id: long (nullable = true)
  |-- val: string (nullable = true)
  
  
  == Existing ==
  root
  |-- id: long (nullable = true)
  |-- val: string (nullable = true)
  |-- val_alter: string (nullable = true)
  
  
  == Differences ==
  - Specified schema is missing field(s): val_alter
  
  If your intention is to keep the existing schema, you can omit the
  schema from the create table command. Otherwise please ensure that
  the schema matches.
  ```
- Therefore adding random `uuid` to the table in case of **databricks external location configured**.

## Notion Ticket

https://www.notion.so/rudderstacks/Deltalake-validations-fails-configured-with-external-location-e299460400a8464ab4deeb25261e552a?pvs=4

## Security

### DIFF
--- a/warehouse/validations/validate.go
+++ b/warehouse/validations/validate.go
@@ -233,7 +233,7 @@ func NewValidator(step string, dest *backendconfig.DestinationT) (Validator, err
 			return nil, fmt.Errorf("create manager: %w", err)
 		}
 		return &createAlterTable{
-			table:   Table,
+			table:   getTable(dest),
 			manager: operations,
 		}, nil
 	case model.VerifyingFetchSchema:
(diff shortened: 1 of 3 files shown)


warehouse/validations/validate.go
```diff
@@ -251,7 +251,7 @@ func NewValidator(step string, dest *backendconfig.DestinationT) (Validator, err
 		return &loadTable{
 			destination: dest,
 			manager:     operations,
-			table:       Table,
```
fracasula, line 254: Unrelated to this PR, but `Table` could be unexported and it could be a `const`.

PR #3379, opened 2023-05-23
feat(router): support for isolation modes using limiters

atzoum:
# Description

Introducing isolation strategies in `router`, following the same pattern that we are already using in `processor` & `batchrouter`.

Router Isolation drawio

Router is one of a kind, since it now uses not 1 but **2 different types of workers**:
1. **partition worker:** multiple partition workers are being created by the worker pool according to the configured isolation strategy.
2. **(internal) worker:** each partition worker has a pool of (internal) workers which ultimately process the jobs and deliver them to their destination. By default, 64 internal workers are spawn for every single partition worker.

## Isolation modes

1. `none` - a single partition worker picks up and processes jobs for all workspaces & destinations.
3. `workspace` - using a separate partition worker per active workspace (default mode in multi-tenant deployments).
4. `destination` - using a separate partition worker per active destination (default mode in non-multi-tenant deployments).

## Limiters

1. `pickup` - for picking up jobs from jobsdb and assigning them to workers (limit: 50 goroutines/destType)
2. `transform` -  for controlling concurrency while transforming jobs (limit: 100 goroutines/destType)
3. `batch` - for controlling concurrency while batching jobs (limit: 100 goroutines/destType)
5. `process` - for controlling concurrency while processing jobs within workers (limit: 100 goroutines/destType)

## Dynamic limiter priorities

Router uses dynamic priorities for executing tasks inside its limiters. Each partition keeps track of it's throughput and error rate for every limiter-enabled operation, by using `partition.Stats`. These statistics are used to calculate a performance score between 0-100, which in turn gets mapped to a limiter priority between 1-4. This priority is especially important when there is congestion in the limiter (goroutines requesting more slots than the limiter's total cap). The limiter itself has a mechanism to avoid starvation of lower-priority jobs, by periodically incrementing their priority (by default every 1 sec, see `WithLimiterDynamicPeriod`).

## Future cleanup work

The existing fair pickup algorithm has been retained and will be used only when `JobsDB.fairPickup` is enabled and isolation mode  is set to `none`. This is to be able to revert in case an unexpected issue happens during the rollout. However, the final plan is to remove `MultiTenantJobsDB - unionQuery.go` and `MultiTenantI - tenantstats.go` in the next release, after router isolation has been released, tested and proven with production workloads. 

## Performance

It comes as no surprise to say that performance is again on par (and even better) with the previous solution up to 200 partitions. After that threshold, throughput starts deteriorating.

The relevant benchmark can be found inside `router_isolation_test.go` and a visualisation of one execution of this benchmark is available [here](https://snapshots.raintank.io/dashboard/snapshot/CLX01r5Nixc3XCrU2P2s3LMF0ZYr0bv5).  

## Other notable changes

- Reorganised package's code in separate, logically-grouped files, trying to make it easier to read and follow.
- Removed global variables and `Init` function from the package.
- `AdaptivePayloadLimiter.enabled` is now `true` by default.
- Setting a soft memory limit (aka `GOMEMLIMIT`) to 80% of the container's available memory.
- Using `bytesize`, `queue`, `mem`, `profiler `and `sync` packages from `rudder-go-kit`.
- Capturing pipeline delay statistics through `pipeline_delay_min` & `pipeline_delay_max`.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/Pipelines-Sprint-68c1c213e9b840b3a9b3826e3631e39a?p=005360d059e64a71bd84472dc77b9ae5&pm=s)

## Security

### DIFF
--- a/gateway/integration_test.go
+++ b/gateway/integration_test.go
@@ -22,7 +22,7 @@ import (
 	"github.com/ory/dockertest/v3"
 	"github.com/stretchr/testify/require"
 
-	kit_helper "github.com/rudderlabs/rudder-go-kit/testhelper"
+	kithelper "github.com/rudderlabs/rudder-go-kit/testhelper"
 	"github.com/rudderlabs/rudder-go-kit/testhelper/docker/resource"
 	"github.com/rudderlabs/rudder-go-kit/testhelper/rand"
 	"github.com/rudderlabs/rudder-server/app"
@@ -117,11 +117,11 @@ func testGatewayByAppType(t *testing.T, appType string) {
 	t.Logf("BackendConfig server listening on: %s", backendConfigSrv.URL)
 	t.Cleanup(backendConfigSrv.Close)
 
-	httpPort, err := kit_helper.GetFreePort()
+	httpPort, err := kithelper.GetFreePort()
 	require.NoError(t, err)
-	httpAdminPort, err := kit_helper.GetFreePort()
+	httpAdminPort, err := kithelper.GetFreePort()
 	require.NoError(t, err)
-	debugPort, err := kit_helper.GetFreePort()
+	debugPort, err := kithelper.GetFreePort()
 	require.NoError(t, err)
 
 	rudderTmpDir, err := os.MkdirTemp("", "rudder_server_*_test")
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -45,7 +45,7 @@ import (
 	"github.com/spf13/viper"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
-	kit_httputil "github.com/rudderlabs/rudder-go-kit/httputil"
+	kithttputil "github.com/rudderlabs/rudder-go-kit/httputil"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 )
@@ -170,5 +170,5 @@ func StartServer(ctx context.Context) error {
 
 	srv := &http.Server{Handler: srvMux, ReadHeaderTimeout: 3 * time.Second}
 
-	return kit_httputil.Serve(ctx, srv, l, time.Second)
+	return kithttputil.Serve(ctx, srv, l, time.Second)
 }
--- a/app/apphandlers/processorAppHandler.go
+++ b/app/apphandlers/processorAppHandler.go
@@ -20,7 +20,7 @@ import (
 
 	"github.com/bugsnag/bugsnag-go/v2"
 
-	kit_httputil "github.com/rudderlabs/rudder-go-kit/httputil"
+	kithttputil "github.com/rudderlabs/rudder-go-kit/httputil"
 	"github.com/rudderlabs/rudder-go-kit/stats"
 	"github.com/rudderlabs/rudder-server/app"
 	"github.com/rudderlabs/rudder-server/app/cluster"
@@ -315,5 +315,5 @@ func (a *processorApp) startHealthWebHandler(ctx context.Context, db *jobsdb.Han
 		MaxHeaderBytes:    a.config.http.MaxHeaderBytes,
 	}
 
-	return kit_httputil.ListenAndServe(ctx, srv)
+	return kithttputil.ListenAndServe(ctx, srv)
 }
--- a/backend-config/namespace_config.go
+++ b/backend-config/namespace_config.go
@@ -13,7 +13,7 @@ import (
 	jsoniter "github.com/json-iterator/go"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
-	kit_httputil "github.com/rudderlabs/rudder-go-kit/httputil"
+	kithttputil "github.com/rudderlabs/rudder-go-kit/httputil"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-server/services/controlplane/identity"
 	"github.com/rudderlabs/rudder-server/utils/types"
@@ -183,7 +183,7 @@ func (nc *namespaceConfig) makeHTTPRequest(req *http.Request) ([]byte, error) {
 		return nil, err
 	}
 
-	defer func() { kit_httputil.CloseResponse(resp) }()
+	defer func() { kithttputil.CloseResponse(resp) }()
 	respBody, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return nil, err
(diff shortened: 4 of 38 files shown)


jobsdb/unionQuery.go
```diff
@@ -33,10 +33,12 @@ type (
...
 
 type GetAllJobsResult struct {
-	Jobs []*JobT
-	More MoreToken
+	Jobs          []*JobT
+	More          MoreToken
+	LimitsReached bool
 }
 
+// TODO: delete this once we remove the old fair pickup algorithm and move MultiTenantLegacy#GetAllJobs inside JobsDB
```
atzoum, line 41: **[note]** added some todos for cleaning up in the next release

main.go
```diff
@@ -18,6 +21,11 @@ var (
 )
 
 func main() {
+	if memStat, err := mem.Get(); err == nil {
+		memoryLimit := int64(80 * memStat.Total / 100)
+		logger.NewLogger().Infow("Setting memory limit to", "limit", memoryLimit)
+		debug.SetMemoryLimit(memoryLimit)
+	}
```
atzoum, line 28: **[note]** setting a soft memory limit (aka `GOMEMLIMIT`) to 80% of the container's available memory.

utils/workerpool/worker_pool.go
```diff
@@ -93,12 +93,24 @@ func (wp *workerPool) PingWorker(partition string) {
...
+	wg.Add(len(wp.workers))
 	for _, w := range wp.workers {
-		w.Stop()
+		w := w
+		go func() {
+			wstart := time.Now()
+			w.Stop()
+			wg.Done()
+			wp.logger.Debugf("worker %s stopped in %s", w.partition, time.Since(wstart))
+		}()
```
atzoum, line 107: **[note]** stopping workers in parallel

utils/payload/limiter_setup.go
```diff
@@ -16,7 +16,7 @@ type AdaptiveLimiterFunc func(int64) int64
 // SetupAdaptiveLimiter creates a new AdaptiveLimiter, starts its RunLoop in a goroutine and periodically collects statistics.
 func SetupAdaptiveLimiter(ctx context.Context, g *errgroup.Group) AdaptiveLimiterFunc {
 	var freeMem FreeMemory
-	if config.GetBool("AdaptivePayloadLimiter.enabled", false) {
+	if config.GetBool("AdaptivePayloadLimiter.enabled", true) {
```
atzoum, line 19: **[note]** `AdaptivePayloadLimiter.enabled` is now `true` by default.

utils/misc/misc.go
```diff
@@ -1289,7 +1268,7 @@ func Unique(stringSlice []string) []string {
 }
 
 func UseFairPickup() bool {
-	return config.GetBool("JobsDB.fairPickup", false) || config.GetBool("EnableMultitenancy", false)
+	return config.GetBool("JobsDB.fairPickup", false) && config.GetString("Router.isolationMode", "default") == "none"
```
atzoum, line 1271: [note] fair pickup will be used only if it is enabled and isolation mode is `none`

router/types.go
```diff
@@ -0,0 +1,95 @@
...
+	jobID                  int64
+	destinationJob         *types.DestinationJobT
+	destinationJobMetadata *types.JobMetadataT
+	respStatusCode         int
+	respBody               string
+	errorAt                string
+	status                 *jobsdb.JobStatusT
+}
+
+type reloadableConfig struct {
```
atzoum, line 72: **[note]** since we will be starting new routers after switching from degraded to normal mode, moving all reloadable configuration to a separate struct, so that there is minimal leakage! 
And promise to prepare a solution for non-leaking reloadable configuration in rudder-go-kit in the future!

.github/workflows/tests.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
-          go-version: '~1.20.3'
```
lvrach, line 21: btw since we are using `~` it is not necessary to change minor versions. `'~1.20.3'` will use anything greater or equal to `1.20.3`
https://devhints.io/semver

atzoum: Yes, in this context, it is true. Nevertheless, it shouldn't do any harm to increase the minimum version allowed.

Excluding golang and only focusing on other packages' versions, years of experience have proven that you shall be better off locking your dependencies and not using `~`. Breaking changes can sneak in patch releases, when you'll expect it the least. Thus, having strict control can be preferable compared to letting it looser. You'll spare some debugging hours at the expense of not being always up-to-date!

gateway/integration_test.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/ory/dockertest/v3"
 	"github.com/stretchr/testify/require"
 
-	kitHelper "github.com/rudderlabs/rudder-go-kit/testhelper"
+	kit_helper "github.com/rudderlabs/rudder-go-kit/testhelper"
```
cisse21, line 25: Aren't we adhering to the outcome of the poll  :P

atzoum: enjoy!

cisse21: We voted for the same thing but lost :'(

router/worker.go
```diff
@@ -289,7 +249,7 @@ func (w *worker) WorkerProcess() {
 			}
 
 		case <-timeout:
-			timeout = time.After(jobsBatchTimeout)
+			timeout = time.After(w.rt.reloadableConfig.jobsBatchTimeout)
```
cisse21, line 252: Why do have this both here and at line 60?

router/worker.go
```diff
@@ -696,6 +639,11 @@ func (w *worker) processDestinationJobs() {
 		status.ErrorResponse = routerutils.EnhanceJSON(routerutils.EmptyPayload, "response", routerJobResponse.respBody)
 		status.ErrorCode = strconv.Itoa(respStatusCode)
 
+		if isSuccessStatus(respStatusCode) {
+			successCount++
+		} else {
+			errorCount++ // TODO: shall we treat aborted as errors?
```
cisse21, line 645: I think if we are flushing the jobs out of the system we shouldn't be much concerned about limiting them or circuit breaking

cisse21 (approved):

router/worker.go
atzoum: further refactoring is required indeed, but will refrain from doing it now as changes are already enough :)

PR #3289, opened 2023-05-08
chore: dedup based on message ID

BonapartePC:
# Description

Enhance our dedup logic:

- Store the size of the event we dedup along with its id
- When a duplicate is detected, retrieve the original message size and compare it with the existing message size.
- When increasing the duplicate metric, use an additional label `equalSize` that will be true or false, according to the previous comparison.
- Introduced cache together with badgerDB which keeps track of messageIDs before they are stored in badgerDB. Once stored in badgerDB,  entries in cache will be deleted.

## Notion Ticket

https://www.notion.so/rudderstacks/Enhance-Dedup-metric-with-size-comparison-64a3dc3943f7434e9fba6ed33e1feaa6?pvs=4

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #3395, opened 2023-05-25
fix: processor panicking during shutdown

BonapartePC:
# Description

Processor must call shutdown after waitgroup returns. Issue surfaced after changes in #3289 where badgerDB is being shut down.

## Security


atzoum (changes requested):
please rebase the branch properly to include only desired changes

PR #3417, opened 2023-05-30
chore: fix flaky dedup test

cisse21:
# Description

Fixing the flaky dedup test where the logger would be a nil pointer in misc package if misc hasn't been initialised

## Notion Ticket

[ Notion Link ](https://www.notion.so/rudderstacks/upgrade-to-badgerV4-86d09c352df94fd5b6c6c109eafc1271?pvs=4)

## Security

### DIFF
--- a/services/dedup/dedup_test.go
+++ b/services/dedup/dedup_test.go
@@ -61,6 +61,7 @@ func Test_Dedup(t *testing.T) {
 func Test_Dedup_Window(t *testing.T) {
 	config.Reset()
 	logger.Reset()
+	misc.Init()
 
 	dbPath := os.TempDir() + "/dedup_test"
 	defer func() { _ = os.RemoveAll(dbPath) }()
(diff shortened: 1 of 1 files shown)


services/dedup/dedup_test.go
```diff
@@ -61,6 +61,7 @@ func Test_Dedup(t *testing.T) {
 func Test_Dedup_Window(t *testing.T) {
 	config.Reset()
 	logger.Reset()
+	misc.Init()
```
fracasula, line 64: [optional] We are only using `misc.CreateTMPDIR()` and `misc` is a poor package name anyway so I propose we move the function in a proper package and remove the dependency from `misc` here.

fracasula: Sounds good to me, +1 for moving it to the kit :+1:

atzoum: test keeps failing frequently, let's do the quick fix first

atzoum (approved):

services/dedup/dedup_test.go
fracasula: Sounds good :+1:

fracasula (approved):

PR #3428, opened 2023-05-31
chore(processor): support multiple jobsdb writers when source isolation is enabled

atzoum:
# Description

Since multiple goroutines might be trying to store jobs in router's jobsdb for the same destinationID and to avoid job ordering guarantee panics (router encountering a job with id less than a job it has previously encountered), we are introducing a partition locker to serialise write access.
Now, before writing a batch of jobs in router, the goroutine in processor will acquire all required partition locks (using an ascending lock order for avoiding deadlocks) and release them after store is completed.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/Pipelines-Sprint-68c1c213e9b840b3a9b3826e3631e39a?p=326809a9d1dd409d9786fb70ac9389ce&pm=s)

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -22,6 +22,7 @@ import (
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/ro"
 	"github.com/rudderlabs/rudder-go-kit/stats"
+	kit_sync "github.com/rudderlabs/rudder-go-kit/sync"
 	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 	eventschema "github.com/rudderlabs/rudder-server/event-schema"
 	"github.com/rudderlabs/rudder-server/jobsdb"
@@ -139,6 +140,7 @@ type Handle struct {
 	}
 
 	adaptiveLimit func(int64) int64
+	storePlocker  kit_sync.PartitionLocker
 }
 type processorStats struct {
 	statGatewayDBR                stats.Measurement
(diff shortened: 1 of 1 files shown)


Sidddddarth:
wouldn't adding the sourceID to the barrier's `orderKey` have sufficed?

atzoum:
> wouldn't adding the sourceID to the barrier's `orderKey` have sufficed?

Thought about it [here](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1684421606972329?thread_ts=1684421266.046459&cid=C01HTT66UMB) as well, but I believe that still there could be a slight possibility for things to go wrong with job iterator.

processor/processor.go
```diff
@@ -1894,41 +1902,50 @@ func (proc *Handle) Store(partition string, in *storeMessage) {
...
+			// committing at different timestamps which can cause events with lower jobIDs to appear after events with higher ones.
+			// For that purpose, before storing, we lock the relevant destination IDs (in sorted order to avoid deadlocks).
+			if len(in.routerDestIDs) > 0 {
+				destIDs := lo.Uniq(in.routerDestIDs)
+				slices.Sort(destIDs)
+				for _, destID := range destIDs {
+					proc.storePlocker.Lock(destID)
+					defer proc.storePlocker.Unlock(destID)
+				}
+			}
```
lvrach, line 1916: Would it be simpler if we had just one lock? 

I haven't any benchmarks, but concurrent writes on the same table are not beneficial from a performance standpoint.

lvrach (approved):

processor/processor.go
atzoum: We have observed major improvements in wait times when we have increased the number of writers, for some scenarios

Sidddddarth (approved):

PR #3433, opened 2023-06-02
chore: timeout for warehouse sql queries

Sidddddarth:
# Description

introduces timeouts for warehouse's sql queries.

## Notion Ticket

[warehouse timeouts](https://www.notion.so/rudderstacks/Timeout-for-all-SQL-queries-in-warehouse-7fb3196d38684ef483079f971c7719f4?pvs=4)

## Security

### DIFF
--- a/warehouse/integrations/middleware/sqlquerywrapper/sql.go
+++ b/warehouse/integrations/middleware/sqlquerywrapper/sql.go
@@ -71,7 +71,6 @@ func (r *Row) Scan(dest ...interface{}) error {
 }
 
 func (r *Row) Err() error {
-	defer r.CancelFunc()
 	return r.Row.Err()
 }
 
--- a/warehouse/integrations/middleware/sqlquerywrapper/sql.go
+++ b/warehouse/integrations/middleware/sqlquerywrapper/sql.go
@@ -70,12 +70,11 @@ func WithTimeout(timeout time.Duration) Opt {
 
 func New(db *sql.DB, opts ...Opt) *DB {
 	s := &DB{
-		DB:                 db,
-		since:              time.Since,
-		slowQueryThreshold: 300 * time.Second,
-		rollbackThreshold:  30 * time.Second,
-		commitThreshold:    30 * time.Second,
-		logger:             rslogger.NOP,
+		DB:                db,
+		since:             time.Since,
+		rollbackThreshold: 30 * time.Second,
+		commitThreshold:   30 * time.Second,
+		logger:            rslogger.NOP,
 	}
 	for _, opt := range opts {
 		opt(s)
@@ -144,6 +143,9 @@ func (db *DB) WithTx(ctx context.Context, fn func(*Tx) error) error {
 }
 
 func (db *DB) logQuery(query string, elapsed time.Duration) {
+	if db.slowQueryThreshold <= 0 {
+		return
+	}
 	if elapsed < db.slowQueryThreshold {
 		return
 	}
@@ -159,31 +161,25 @@ func (db *DB) logQuery(query string, elapsed time.Duration) {
 	db.logger.Infow("executing query", keysAndValues...)
 }
 
+func (tx *Tx) GetTx() *sql.Tx {
+	return tx.Tx
+}
+
 // Begin starts a transaction.
 //
 // Use BeginTx to pass context and options to the underlying driver.
 func (db *DB) Begin() (*Tx, error) {
-	ctx, cancel := queryContextWithTimeout(context.Background(), db.queryTimeout)
-	defer cancel()
-	if tx, err := db.DB.BeginTx(ctx, nil); err != nil {
-		return nil, err
-	} else {
-		return &Tx{tx, db}, nil
-	}
-}
-
-func (tx *Tx) GetTx() *sql.Tx {
-	return tx.Tx
+	return db.BeginTx(context.Background(), nil)
 }
 
 func (db *DB) BeginTx(ctx context.Context, opts *sql.TxOptions) (*Tx, error) {
 	ctx, cancel := queryContextWithTimeout(ctx, db.queryTimeout)
 	defer cancel()
-	if tx, err := db.DB.BeginTx(ctx, opts); err != nil {
+	tx, err := db.DB.BeginTx(ctx, opts)
+	if err != nil {
 		return nil, err
-	} else {
-		return &Tx{tx, db}, nil
 	}
+	return &Tx{tx, db}, nil
 }
 
 func (tx *Tx) Exec(query string, args ...interface{}) (sql.Result, error) {
--- a/warehouse/integrations/middleware/sqlquerywrapper/sql.go
+++ b/warehouse/integrations/middleware/sqlquerywrapper/sql.go
@@ -172,6 +172,10 @@ func (db *DB) Begin() (*Tx, error) {
 	}
 }
 
+func (tx *Tx) GetTx() *sql.Tx {
+	return tx.Tx
+}
+
 func (db *DB) BeginTx(ctx context.Context, opts *sql.TxOptions) (*Tx, error) {
 	ctx, cancel := queryContextWithTimeout(ctx, db.queryTimeout)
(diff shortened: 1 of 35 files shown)


warehouse/internal/repo/load.go
```diff
@@ -43,6 +45,11 @@ func NewLoadFiles(db *sql.DB, opts ...Opt) *LoadFiles {
 
 // DeleteByStagingFiles deletes load files associated with stagingFileIDs.
 func (repo *LoadFiles) DeleteByStagingFiles(ctx context.Context, stagingFileIDs []int64) error {
+	ctx, cancel := context.WithTimeout(
+		ctx,
+		config.GetDuration("Warehouse.deleteLoadFilesTimeout", 5, time.Minute),
```
lvrach, line 50: Having such fine-grain control over the timeouts might make it harder to change the configuration during runtime. Assuming there is a problem with the database that affects all timing, we have to track down all the queries.

 Having such great timeouts I find it hard to require different configs for each of our queries.

warehouse/internal/repo/staging.go
```diff
@@ -49,11 +51,10 @@ type metadataSchema struct {
...
 func StagingFileIDs(stagingFiles []*model.StagingFile) []int64 {
-	var stagingFileIDs []int64
-	for _, stagingFile := range stagingFiles {
-		stagingFileIDs = append(stagingFileIDs, stagingFile.ID)
-	}
-	return stagingFileIDs
+	return lo.Map(
+		stagingFiles,
+		func(stagingFile *model.StagingFile, _ int) int64 { return stagingFile.ID },
+	)
```
lvrach, line 57: Subjective comment, but not in favour of using `lo` here. I find the previous code simple enough.

We could do, for a minor performance improvement:
```suggestion
        stagingFileIDs := make([]int64, len(stagingFiles)
	for i, stagingFile := range stagingFiles {
		stagingFileIDs[i] = stagingFile.ID
	}
```

warehouse/upload.go
```diff
@@ -2126,17 +2237,29 @@ func (job *UploadJob) RefreshPartitions(loadFileStartID, loadFileEndID int64) er
...
+					StartID: loadFileStartID,
+					EndID:   loadFileEndID,
+				})
+				batches := lo.Chunk(loadFiles, job.refreshPartitionBatchSize)
+				for _, batch := range batches {
+					if err = repository.RefreshPartitions(ctx, tableName, batch); err != nil {
+						return fmt.Errorf("refresh partitions: %w", err)
+					}
+				}
+				return nil
```
lvrach, line 2259: Why not add the timeouts to individual queries as we did above?

warehouse/validations/validate.go
```diff
@@ -527,7 +527,10 @@ func createManager(ctx context.Context, dest *backendconfig.DestinationT) (manag
 		return nil, fmt.Errorf("getting manager: %w", err)
 	}
 
-	operations.SetConnectionTimeout(warehouseutils.TestConnectionTimeout)
+	operations.SetConnectionTimeout(warehouseutils.GetConnectionTimeout(
+		destType,
+		dest.ID,
+	))
```
achettyiitr, line 533: We should probably keep this as it is since it is used for destination validations from the control plane.

warehouse/warehouse.go
```diff
@@ -206,6 +209,7 @@ func loadConfig() {
 	config.RegisterIntConfigVariable(8, &maxParallelJobCreation, true, 1, "Warehouse.maxParallelJobCreation")
 	config.RegisterBoolConfigVariable(false, &enableJitterForSyncs, true, "Warehouse.enableJitterForSyncs")
 	config.RegisterDurationConfigVariable(30, &tableCountQueryTimeout, true, time.Second, []string{"Warehouse.tableCountQueryTimeout", "Warehouse.tableCountQueryTimeoutInS"}...)
+	config.RegisterDurationConfigVariable(5, &dbHanndleTimeout, true, time.Minute, []string{"Warehouse.dbHanndleTimeout", "Warehouse.dbHanndleTimeoutInS"}...)
```
achettyiitr, line 212:
```suggestion
	config.RegisterDurationConfigVariable(5, &dbHanndleTimeout, true, time.Minute, []string{"Warehouse.dbHanndleTimeout", "Warehouse.dbHanndleTimeoutInMin"}...)
```

achettyiitr:
Should we use the same middleware for [schema](https://github.com/rudderlabs/rudder-server/blob/f9ce758eb2f86d01944097f3d9870e3174f265d8/warehouse/internal/repo/schema.go#L33) and [staging](https://github.com/rudderlabs/rudder-server/blob/aba7cb7dcd7e089ba73abb414eff084dfb36c08a/warehouse/internal/repo/staging.go#L77) repo as well.

warehouse/integrations/middleware/sqlquerywrapper/sql.go
```diff
@@ -153,27 +161,29 @@ func (db *DB) logQuery(query string, elapsed time.Duration) {
...
-		return nil, err
-	} else {
-		return &Tx{tx, db}, nil
-	}
+	return db.BeginTx(context.Background(), nil)
 }
 
 func (db *DB) BeginTx(ctx context.Context, opts *sql.TxOptions) (*Tx, error) {
-	if tx, err := db.DB.BeginTx(ctx, opts); err != nil {
+	ctx, cancel := queryContextWithTimeout(ctx, db.queryTimeout)
```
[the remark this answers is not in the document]
Sidddddarth: using `db.transactionTimeout` now, which also defaults to zero. Meaning no timeout applied to passed context.

warehouse/internal/repo/upload.go
```diff
@@ -137,9 +139,8 @@ func (uploads *Uploads) CreateWithStagingFiles(ctx context.Context, upload model
 		}
 	}()
 
-	row := tx.QueryRowContext(
-		ctx,
-
+	var uploadID int64
+	err = tx.QueryRow(
```
lvrach, line 143: Why we have removed ctx from here ?

Sidddddarth: because it's the same context that's used to create the transaction.

warehouse/integrations/middleware/sqlquerywrapper/sql.go
```diff
@@ -137,78 +205,101 @@ func (db *DB) WithTx(ctx context.Context, fn func(*Tx) error) error {
...
-	} else {
-		return &Tx{tx, db}, nil
-	}
+	return db.BeginTx(context.Background(), nil)
 }
 
 func (db *DB) BeginTx(ctx context.Context, opts *sql.TxOptions) (*Tx, error) {
-	if tx, err := db.DB.BeginTx(ctx, opts); err != nil {
+	ctx, cancel := queryContextWithTimeout(ctx, db.transactionTimeout)
+	defer cancel()
```
Sidddddarth, line 244: wonder how the tests passed considering the following documentation for `*sql.DB.BeginTx`:

```
func (*sql.DB).BeginTx(ctx context.Context, opts *sql.TxOptions) (*sql.Tx, error)
BeginTx starts a transaction.

The provided context is used until the transaction is committed or rolled back. 
If the context is canceled, the sql package will roll back the transaction. 
Tx.Commit will return an error if the context provided to BeginTx is canceled.

The provided TxOptions is optional and may be nil if defaults should be used. 
If a non-default isolation level is used that the driver doesn't support, an error will be returned.
```

Sidddddarth: made changes so context is cancelled when commit/rollback returns.

Sidddddarth: > wonder how the tests passed

We weren't setting `db.transactionTimeout`.
so `defer cancel()` wouldn't cancel the context.
We'd have seen errors if we'd set that timeout.

Adding tests to verify that.

achettyiitr (approved):

cisse21 (approved):

PR #3265, opened 2023-05-03
feat: error detail reporting

sanpj2292:
# Description

We want to introduce a new table `error_detail_reports` to have details of precise error message and error code for each event that is sent to transformer.

## Notion Ticket

https://www.notion.so/rudderstacks/Design-for-errorCode-and-errorMessage-in-reports-table-3d5db2e4d485459f9b4f98b601c7a5e5?pvs=4

## Security

- [ ] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #3439, opened 2023-06-05
feat: facade implementation for reporting

sanpj2292:
# Description

We are implementing a facade or mediator for reporting to make it easier for us to integrate

This is how things look at high-level
Screenshot 2023-06-01 at 9 06 40 AM

We have created a new component that is responsible for:
 - calling multiple error reporting implementations
 - only call them if they are enabled

Advantages:
- This will simplify the code in the processor, router, and warehouse
- They will just call a single interface, and no need to check if reporting is enabled or not

**Credits**: @lvrach for the image, the following content

## Notion Ticket
https://www.notion.so/rudderstacks/Design-for-errorCode-and-errorMessage-in-reports-table-3d5db2e4d485459f9b4f98b601c7a5e5?pvs=4

## Security


sanpj2292:
This PR will be merged to the branch in PR #3265

koladilip (approved):

PR #3443, opened 2023-06-05
chore: upgrade golangci version and lint fixes

lvrach:
# Description

I have upgraded golangci to `version: v1.52`, I have only specified up to minor versions so that patch upgrades will happen automatically.

With the upgrade, the following linters were reenabled again (disabled before due to generics)
* [rowserrcheck](https://github.com/golangci/golangci-lint/pull/3691)
* [wastedassign](https://github.com/golangci/golangci-lint/pull/3689)

## Linting fixes

rowserrcheck: After `rows.Next()` iteration we need to check for `rows.Err()`.
wastedassign: Assign and value and not use it. In most cases `x := 0` -> `var x int`

## Notion Ticket

https://www.notion.so/rudderstacks/Upgrade-golangci-version-lint-fixes-7abe589fec944d16b16ca378363ac8d1?pvs=4

## Security

### DIFF
--- a/jobsdb/admin.go
+++ b/jobsdb/admin.go
@@ -12,11 +12,10 @@ Ping returns health check for pg database
 func (jd *HandleT) Ping() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
-	rows, err := jd.dbHandle.QueryContext(ctx, `SELECT 'Rudder DB Health Check'::text as message`)
+	_, err := jd.dbHandle.ExecContext(ctx, `SELECT 'Rudder DB Health Check'::text as message`)
 	if err != nil {
 		return err
 	}
-	_ = rows.Close()
 	return nil
 }
 
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -1067,7 +1067,6 @@ func (jd *HandleT) refreshDSRangeList(l lock.LockToken) {
 		sqlStatement := fmt.Sprintf(`SELECT MIN(job_id), MAX(job_id) FROM %q`, ds.JobTable)
 		// Note: Using Query instead of QueryRow, because the sqlmock library doesn't have support for QueryRow
 		row := jd.dbHandle.QueryRow(sqlStatement)
-		jd.assertError(row.Err())
 
 		err := row.Scan(&minID, &maxID)
 		jd.assertError(err)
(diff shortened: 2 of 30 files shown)


jobsdb/admin.go
```diff
@@ -12,11 +12,10 @@ Ping returns health check for pg database
 func (jd *HandleT) Ping() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
-	rows, err := jd.dbHandle.QueryContext(ctx, `SELECT 'Rudder DB Health Check'::text as message`)
+	_, err := jd.dbHandle.ExecContext(ctx, `SELECT 'Rudder DB Health Check'::text as message`)
```
lvrach, line 15: No need to call `QueryContext` if we are not interested in result

jobsdb/jobsdb.go
```diff
@@ -1066,14 +1066,12 @@ func (jd *HandleT) refreshDSRangeList(l lock.LockToken) {
 		jd.assert(ds.Index != "", "ds.Index is empty")
 		sqlStatement := fmt.Sprintf(`SELECT MIN(job_id), MAX(job_id) FROM %q`, ds.JobTable)
 		// Note: Using Query instead of QueryRow, because the sqlmock library doesn't have support for QueryRow
-		rows, err := jd.dbHandle.Query(sqlStatement)
+		row := jd.dbHandle.QueryRow(sqlStatement)
+		jd.assertError(row.Err())
+
+		err := row.Scan(&minID, &maxID)
```
lvrach, line 1072: Using `QueryRow` since we only wanted to read a single row 

```suggestion
		row := jd.dbHandle.QueryRow(sqlStatement)

		err := row.Scan(&minID, &maxID)
```

services/multitenant/tenantstats.go
```diff
@@ -273,7 +273,7 @@ func (t *Stats) GetRouterPickupJobs(destType string, noOfWorkers int, routerTime
 			break
 		}
 
-		pickUpCount := 0
+		var pickUpCount int
```
lvrach, line 276: linter was complaining: variable defined but got overwritten before used.

warehouse/integrations/deltalake-native/deltalake.go
```diff
@@ -905,6 +905,8 @@ func (d *Deltalake) partitionQuery(ctx context.Context, tableName string) (strin
 	}
 	defer func() { _ = rows.Close() }()
 
+	_ = rows.Err() // ignore error
```
lvrach, line 908: We need to read Columns, but not the rows, no need to check for errors.

I prefer using `_ = rows.Err()`, instead of `//nolint:rowserrcheck` because it works across all linters.

achettyiitr (approved):

achettyiitr:
I believe we introduced a couple of panics while handling the errors. But I think we can take it as a separate PR to handle it.

lvrach:
> I believe we introduced a couple of panics while handling the errors for the warehouse module. But I think we can take it as a separate PR to handle it.

My strategy was to mimic the way errors were handled in the function. Otherways the scope of PR could have been much bigger.

atzoum (approved):

cisse21 (approved):

PR #3446, opened 2023-06-05
chore: add resp body status check

cisse21:
# Description

Handle non 2xx response codes in processor transformer and retry the events

## Notion Ticket

[ Notion Link ](https://www.notion.so/rudderstacks/Retry-non-2xx-responses-processor-transformer-58ee7768907c4a61bc48f40b63653dc9?pvs=4)

## Security

### DIFF
--- a/processor/transformer/transformer.go
+++ b/processor/transformer/transformer.go
@@ -118,6 +118,13 @@ func Init() {
 	pkgLogger = logger.NewLogger().Child("processor").Child("transformer")
 }
 
+func isJobTerminated(status int) bool {
+	if status == http.StatusTooManyRequests {
+		return false
+	}
+	return status >= http.StatusOK && status < http.StatusInternalServerError
+}
+
 func loadConfig() {
 	config.RegisterIntConfigVariable(200, &maxConcurrency, false, 1, "Processor.maxConcurrency")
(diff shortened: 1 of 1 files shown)


processor/transformer/transformer.go
```diff
@@ -389,6 +393,9 @@ func (trans *HandleT) doPost(ctx context.Context, rawJSON []byte, url string, ta
 			if reqErr != nil {
 				return reqErr
 			}
+			if !isJobTerminated(resp.StatusCode) {
+				return fmt.Errorf("transformer returned status code: %v", resp.StatusCode)
+			}
 			defer func() { httputil.CloseResponse(resp) }()
```
Sidddddarth, line 399: return after this deferred body close..?

Sidddddarth (approved):

processor/transformer/transformer.go
```diff
@@ -118,6 +118,10 @@ func Init() {
 	pkgLogger = logger.NewLogger().Child("processor").Child("transformer")
 }
 
+func isJobTerminated(status int) bool {
+	return status >= 200 && status < 500
```
atzoum, line 122: what about `429`?

cisse21: Do we want to handle `429` from transformer? because right now it doesn't return any `429`

atzoum: why shouldn't we handle it already?

atzoum (approved):

PR #3451, opened 2023-06-06
fix: transform timeout crash

saurav-malani:
# Description
This PR makes rudder-server panic configurable incase of transformer's request timeout.

## Notion Ticket

https://www.notion.so/rudderstacks/Hosted-abort-jobs-timed-out-during-UT-c21526789be046b3bec857bd239b8ae3?pvs=4

## Security


atzoum (changes requested):
This task shouldn't affect router's transformer code. User transformations are always happening at processor.

saurav-malani:
> This task shouldn't affect router's transformer code. User transformations are always happening at processor.

@atzoum 
but won't [this](https://github.com/rudderlabs/rudder-server/blob/f53a42cfec78c34c15b971eb5a4f21c1d000499c/processor/transformer/transformer.go#L381-L406) part of the code take care of not panicing in case of timeout after multiple retries i.e. by setting `"Processor.Transformer.failOnError"` to `true`?

processor/transformer/transformer_test.go
```diff
@@ -48,6 +49,36 @@ func (t *fakeTransformer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	events[0] = transformer.TransformerEventT{
+		Metadata: transformer.MetadataT{
+			MessageID: "messageID-1",
+		},
+		Message: map[string]interface{}{
+			"src-key-1": "messageID-1",
+		},
+	}
+	rsp := tr.Transform(context.TODO(), events, svr.URL, 1)
+	require.Equal(t, transformer.TransformerRequestTimeoutAbort, rsp.FailedEvents[0].StatusCode, "actual response code different than expected")
```
lvrach, line 79: Another, more controlled way of doing this:

```suggestion
	transformer.Init()
	ch := make(chan struct{})
	svr := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("apiVersion", "2")
		<- ch
	}))

	defer svr.Close()

	tr := transformer.NewTransformer()
	tr.Setup()
	events := make([]transformer.TransformerEventT, 1)

	events[0] = transformer.TransformerEventT{
		Metadata: transformer.MetadataT{
			MessageID: "messageID-1",
		},
		Message: map[string]interface{}{
			"src-key-1": "messageID-1",
		},
	}
	rsp := tr.Transform(context.TODO(), events, svr.URL, 1)
	close(ch)
	require.Equal(t, transformer.TransformerRequestTimeoutAbort, rsp.FailedEvents[0].StatusCode, "actual response code different than expected")
```

processor/transformer/transformer_test.go
```diff
@@ -48,6 +49,36 @@ func (t *fakeTransformer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func TestTransformerTimeoutAbort(t *testing.T) {
+	config.Reset()
+	logger.Reset()
+	config.Set("HttpClient.procTransformer.timeout", 1)
```
lvrach, line 55: Is it possible to do:
```suggestion
	config.Set("HttpClient.procTransformer.timeout", "1ms")
```

saurav-malani: damn... I didn't know we can do something like this. Thanks, modified.

processor/transformer/transformer_test.go
saurav-malani: modified to use channel

processor/transformer/transformer.go
```diff
@@ -409,7 +413,10 @@ func (trans *HandleT) doPost(ctx context.Context, rawJSON []byte, url string, ta
 			trans.logger.Warnf("JS HTTP connection error: URL: %v Error: %+v after %v tries", url, err, retryCount)
 		})
 	if err != nil {
-		if config.GetBool("Processor.Transformer.failOnError", false) {
+		if netErr, ok := err.(net.Error); ok && netErr.Timeout() && config.GetBool("Processor.Transformer.abortOnTimeout", false) {
```
atzoum, line 416: Using `os.IsTimeout` is preferred

atzoum, line 416: We only want this behaviour to happen during user transformations. Timed-out dest transformations should never result in jobs getting aborted

processor/transformer/transformer.go
```diff
@@ -350,31 +353,32 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
 			transformerResponses = append(transformerResponses, resp)
 		}
+		return transformerResponses
 	}
+
+	integrations.CollectIntgTransformErrorStats(respData)
+
+	trace.Logf(ctx, "Unmarshal", "response raw size: %d", len(respData))
+	trace.WithRegion(ctx, "Unmarshal", func() {
+		err = jsonfast.Unmarshal(respData, &transformerResponses)
```
cisse21, line 370: Why are doing an unmarshal when the status code isn't 200. This would mean that the response wasn't sent by transformer and there is no need for it to be of `TransformerResponseT` type

processor/transformer/transformer.go
```diff
@@ -350,31 +353,32 @@ func (trans *HandleT) request(ctx context.Context, url string, data []Transforme
...
+	})
+	// This is returned by our JS engine so should  be parsable
+	// but still handling it
+	if err != nil {
+		trans.logger.Errorf("Data sent to transformer : %v", string(rawJSON))
+		trans.logger.Errorf("Transformer returned : %v", string(respData))
+		respData = []byte(fmt.Sprintf("Failed to unmarshal transformer response: %s", string(respData)))
+		transformerResponses = nil
+		statusCode = 400
```
cisse21, line 379: This doesn't seem to be needed as it isn't used anywhere post here
```suggestion
```

processor/transformer/transformer.go
saurav-malani: I think you missed the `}` in line 364. So, we are handling non-200 case before and this part of the code is for 200. Essentially this unmarshal is done for 200 cases.

PR #3460, opened 2023-06-07
fix: flag for warehouse altering

achettyiitr:
# Description

Added flag to skip altering columns for warehouse

## Notion Ticket

https://www.notion.so/rudderstacks/Warehouse-altering-columns-flag-3e99c9a29656474596eddfc7375e1bdb?pvs=4

## Security

### DIFF
--- a/warehouse/upload.go
+++ b/warehouse/upload.go
@@ -104,6 +104,7 @@ type UploadJob struct {
 	RefreshPartitionBatchSize int
 	RetryTimeWindow           time.Duration
 	MinRetryAttempts          int
+	DisableAlter              bool
 
 	ErrorHandler ErrorHandler
 }
@@ -193,6 +194,7 @@ func (f *UploadJobFactory) NewUploadJob(dto *model.UploadJob, whManager manager.
 		RefreshPartitionBatchSize: config.GetInt("Warehouse.refreshPartitionBatchSize", 100),
 		RetryTimeWindow:           retryTimeWindow,
 		MinRetryAttempts:          minRetryAttempts,
+		DisableAlter:              config.GetBool("Warehouse.disableAlter", false),
 
 		AlertSender: alerta.NewClient(
 			config.GetString("ALERTA_URL", "https://alerta.rudderstack.com/api/"),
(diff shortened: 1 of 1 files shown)


warehouse/upload.go
```diff
@@ -781,6 +783,19 @@ func (job *UploadJob) UpdateTableSchema(tName string, tableSchemaDiff warehouseu
 }
 
 func (job *UploadJob) alterColumnsToWarehouse(tName string, columnsMap model.TableSchema) error {
+	if !job.AlterEnabled {
+		pkgLogger.Debugw("skipping alter columns to warehouse",
+			logfield.SourceID, job.warehouse.Source.ID,
+			logfield.SourceType, job.warehouse.Source.SourceDefinition.Name,
```
fracasula, line 789: Shouldn't this be `SourceDefinition.Type`? :thinking:

warehouse/upload.go
```diff
@@ -781,6 +783,19 @@ func (job *UploadJob) UpdateTableSchema(tName string, tableSchemaDiff warehouseu
...
+			logfield.DestinationID, job.warehouse.Destination.ID,
+			logfield.DestinationType, job.warehouse.Destination.DestinationDefinition.Name,
+			logfield.WorkspaceID, job.warehouse.WorkspaceID,
+			logfield.TableName, tName,
+			"columns", fmt.Sprint(columnsMap),
```
fracasula, line 794: `fmt.Sprintf` is probably not needed. just saying though, it's not required to remove it.

fracasula (approved):

lvrach (approved):

warehouse/upload.go
fracasula: I noticed because you're preceding it with `SourceType`. I just wanted to double check with you in case you would expect some data in the debug log entry but ended up with something else. No need to make changes if this is the expected behaviour.

warehouse/upload_test.go
```diff
@@ -381,6 +381,7 @@ func TestUploadJobT_UpdateTableSchema(t *testing.T) {
 						AlertSender: &mockAlertSender{
 							mockError: tc.mockAlertError,
 						},
+						AlterEnabled: true,
```
fracasula, line 384: We could have the attribute be called `DisableAlter` instead so that the default (i.e. `false`) would make more sense.

PR #3465, opened 2023-06-08
chore: kafka manager errors

fracasula:
# Description

Improving some errors to include the schema ID.

## Security


cisse21 (approved):

services/streammanager/kafka/kafkamanager.go
```diff
@@ -722,11 +725,14 @@ func sendMessage(ctx context.Context, jsonData json.RawMessage, p producerManage
...
 		codec, ok := codecs[schemaId]
 		if !ok {
-			return makeErrorResponse(fmt.Errorf("unable to find schema with schemaId: %v", schemaId))
+			return makeErrorResponse(fmt.Errorf("unable to find schema with ID %v", schemaId))
 		}
 		value, err = serializeAvroMessage(schemaId, p.getEmbedAvroSchemaID(), value, *codec)
 		if err != nil {
-			return makeErrorResponse(fmt.Errorf("unable to serialize event with messageId: %s, with error %s", messageId, err))
+			return makeErrorResponse(fmt.Errorf(
+				"unable to serialize event with schemaId %q and messageId %s: %s",
```
Sidddddarth, line 733: when to use `%w` as opposed to `%s` for an `error` type.

Sidddddarth (approved):

Sidddddarth (commented):
looks like the test files need said changes too

services/streammanager/kafka/kafkamanager.go
fracasula: In this case it doesn't matter, it gets converted right away into a string by `makeErrorResponse`.

Otherwise, wrapping error makes sense since they can be unwrapped manually or used via `errors.Is` and other utils. This gives you the ability to add some context to the error but it still lets the caller check the underlying error like:

```go
func someDatabaseOperation() error {
    return fmt.Errorf("empty result set while doing some database operation: %w", sql.ErrNoRows)
}

err := someDatabaseOperation()
if errors.Is(err, sql.ErrNoRows) { // this would still work
    // return 404
}
```

Sidddddarth (approved):

Sidddddarth (approved):

PR #3478, opened 2023-06-10
chore: avoid querying a dataset in case AfterJobID falls after said dataset

Sidddddarth:
# Description

Avoid querying a dataset if `params.AfterJobID != nil && *params.AfterJobID > dsRange.maxJobID`.
Added said logic to `jobsdb.GetProcessed(...)` and `jobsdb.GetUnprocessed(...)`

## Notion Ticket

[update archive status](https://www.notion.so/rudderstacks/jobsdb-support-archive-state-a9c85931f711407a9cc0415f2c75ed83?pvs=4)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -3105,7 +3105,7 @@ func (jd *HandleT) getUnprocessed(ctx context.Context, params GetQueryParamsT) (
 	}
 	defer jd.dsListLock.RUnlock()
 
-	dsList := jd.getDSList()
+	dsRangeList := jd.getDSRangeList()
 
 	limitByEventCount := false
 	if params.EventsLimit > 0 {
@@ -3124,11 +3124,17 @@ func (jd *HandleT) getUnprocessed(ctx context.Context, params GetQueryParamsT) (
 	if jd.dsLimit != nil {
 		dsLimit = *jd.dsLimit
 	}
-	for _, ds := range dsList {
+	for _, dsRange := range dsRangeList {
+		if params.AfterJobID != nil {
+			if dsRange.maxJobID < *params.AfterJobID {
+				// cacheHitCount++ ?
+				continue
+			}
+		}
 		if dsLimit > 0 && dsQueryCount >= dsLimit {
 			break
 		}
-		unprocessedJobs, dsHit, err := jd.getUnprocessedJobsDS(ctx, ds, params)
+		unprocessedJobs, dsHit, err := jd.getUnprocessedJobsDS(ctx, dsRange.ds, params)
 		if err != nil {
 			return JobsResult{}, err
 		}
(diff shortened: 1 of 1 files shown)


jobsdb/jobsdb.go
```diff
@@ -3124,11 +3124,17 @@ func (jd *HandleT) getUnprocessed(ctx context.Context, params GetQueryParamsT) (
 	if jd.dsLimit != nil {
 		dsLimit = *jd.dsLimit
 	}
-	for _, ds := range dsList {
+	for _, dsRange := range dsRangeList {
+		if params.AfterJobID != nil {
+			if dsRange.maxJobID < *params.AfterJobID {
```
atzoum, line 3129: AFAIK the last dataset doesn't have a `maxJobID`

atzoum, line 3129:
```suggestion
			if *params.AfterJobID > dsRange.maxJobID {
```

Sidddddarth: true
please take a look again

jobsdb/jobsdb.go
```diff
@@ -3124,7 +3125,14 @@ func (jd *HandleT) getUnprocessed(ctx context.Context, params GetQueryParamsT) (
 	if jd.dsLimit != nil {
 		dsLimit = *jd.dsLimit
 	}
-	for _, ds := range dsList {
+	for idx, ds := range dsList {
+		if params.AfterJobID != nil {
+			if idx < len(dsRangeList) {
```
atzoum, line 3130: would be nice to add a comment of why we are excluding the last dataset

jobsdb/jobsdb.go
```diff
@@ -3105,6 +3105,7 @@ func (jd *HandleT) getUnprocessed(ctx context.Context, params GetQueryParamsT) (
 	}
 	defer jd.dsListLock.RUnlock()
 
+	dsRangeList := jd.getDSRangeList()
```
atzoum, line 3108: In line 1678 we might need to call `refreshDSRangeList` instead of `refreshDSList` otherwise there is a possibility to ignore the latest dataset for a brief period of time.

atzoum (approved):

cisse21 (approved):

jobsdb/jobsdb.go
Sidddddarth: imo we wouldn't be ignoring a dataset, we'd just not be applying

```
if *params.AfterJobID > dsRangeList[idx].maxJobID {
	continue
}
```

condition for the penultimate dataset which means we'd query it even if above condition was true.

But I do see your point, shouldn't we be changing `refreshDSList` to `refreshDSRangeList` in multiple places then?

atzoum: I don't believe it is needed elswhere, but let's do this change in another PR

atzoum: yes, another pr :)

PR #3500, opened 2023-06-15
chore: nil check for health handler

achettyiitr:
# Description

Handle nil check for health handler. In case of error, we were trying to close the rows which will cause an issue.

## Notion Ticket

https://www.notion.so/rudderstacks/nil-check-during-health-a6e8f42fcba74233b5e63fa97e04f800?pvs=4

## Security


warehouse/warehouse.go
```diff
@@ -1493,31 +1476,60 @@ func clearTriggeredUpload(wh model.Warehouse) {
...
+
 	_, _ = w.Write([]byte(healthVal))
 }
 
+func CheckPGHealth(ctx context.Context, db *sql.DB) bool {
+	if db == nil {
+		return false
+	}
+
+	rows, err := db.QueryContext(ctx, `SELECT 'Rudder Warehouse DB Health Check'::text as message;`)
```
cisse21, line 1524: Why did we remove the `rows.Err()` check

cisse21 (approved):

warehouse/warehouse.go
achettyiitr: Since we are not doing scan, no need to check for Err() as well?

achettyiitr: Although, since we are enforced to adding `rows.Err()` due to the upgrade of `golangci` version, ignoring error for `rows.Err()`

fracasula: Are you sure this is good enough? You're not even checking the result of the query.

Anyway, I like the idea implemented by this library [here](https://github.com/dimiro1/health/blob/master/db/checker.go#L28). Maybe we could do the same, `SELECT VERSION()` and then check that we actually got something valid out of it.

lvrach: We can do `QueryRowContext` if we care about reading the response or just `db.ExecContext` if we ignore results
```suggestion
	err := db.QueryRowContext(ctx, "SELECT  'Rudder Warehouse DB Health Check'::text as message;").Scan(&msg)

```

achettyiitr: Should I just compare the msg with `Rudder Warehouse DB Health Check`?

achettyiitr: Comparing msg now.

lvrach (approved):

PR #3504, opened 2023-06-15
chore: add support for detecting bot events in gateway

achettyiitr:
# Description

- Add support for detecting bot events during gateway injection.
- Using bot user agents from the [crawler-user-agents.json file](https://raw.githubusercontent.com/monperrus/crawler-user-agents/master/crawler-user-agents.json
  - Out of 500+ patterns shared, 60% of them are resolved with bot, crawler and spider.

## Notion Ticket

https://www.notion.so/rudderstacks/Observe-bot-traffic-in-gateway-metrics-f13fc91b5d1d43509a38ea60b84af439?pvs=4

## Security


gateway/internal/stats/stats.go
```diff
@@ -100,6 +105,7 @@ func (ss *SourceStat) Report(s stats.Stats) {
 	}
 
 	s.NewTaggedStat("gateway.write_key_requests", stats.CountType, tags).Count(ss.requests.total)
+	s.NewTaggedStat("gateway.write_key_bot_requests", stats.CountType, tags).Count(ss.requests.bot)
```
lvrach, line 108: what if we keep the same metric name, but add a bot label? 
```suggestion
	s.NewTaggedStat("gateway.write_key_requests", stats.CountType, tags + {bot=true}).Count(ss.requests.bot)
```

achettyiitr: We would need to change in Grafana as well then, right?

achettyiitr:
Adding a benchmark test.
Adding some more tests as well.

achettyiitr:
With 235 keywords => 
```
goos: darwin
goarch: amd64
pkg: github.com/rudderlabs/rudder-server/gateway/internal/bot
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkIsBotUserAgent
BenchmarkIsBotUserAgent/Complex_User_Agent
BenchmarkIsBotUserAgent/Complex_User_Agent-12         	   14808	     68921 ns/op
BenchmarkIsBotUserAgent/Simple_User_Agent
BenchmarkIsBotUserAgent/Simple_User_Agent-12          	 5578768	       207.0 ns/op
PASS
```


With 3 keywords => 
```
goos: darwin
goarch: amd64
pkg: github.com/rudderlabs/rudder-server/gateway/internal/bot
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkIsBotUserAgent
BenchmarkIsBotUserAgent/Complex_User_Agent
BenchmarkIsBotUserAgent/Complex_User_Agent-12         	   65400	     18208 ns/op
BenchmarkIsBotUserAgent/Simple_User_Agent
BenchmarkIsBotUserAgent/Simple_User_Agent-12          	 5751565	       206.7 ns/op
PASS
```

gateway/gateway.go
```diff
@@ -603,33 +607,32 @@ func (gateway *HandleT) getJobDataFromRequest(req *webRequestT) (jobData *jobFro
 			return
 		}
 
+		eventContext, _ := toSet["context"].(map[string]interface{})
```
Sidddddarth, line 610: if `context` is not present in `toSet`, I think this may cause an issue, because if the type assertion doesn't hold the nil value for a map would be nil.


```suggestion
		eventContext, ok := misc.MapLookup(toSet, "context").(map[string]interface{})
		if ok {
				further lookups
		}
```

Sidddddarth (approved):
LGTM

PR #3522, opened 2023-06-20
feat: save webhook (/source transformation) failures to proc errors

chandumlg:
# Description

Save webhook (/source transformation) failures to proc errors.

To make this happen, I had to create two instances of proc error jobsdbs (one in read mode and one in write mode). Router and Batch router take write error jobsdb, gateway needs write error jobsdb & processor needs both read and write jobsdbs.

**Multitenant scenario**: This change should not create any problems in multi-tenant deployment (since proc error jobsdb behavior will be similar to gw jobsdb)

## Notion Ticket

https://www.notion.so/rudderstacks/d5d15ce4be354e3caae11d4ea9f41477?v=2ad23dce1aeb431a82f783e6607d3cb0&p=334353dc6e204af188569cb5646f13d4&pm=c

## Security


app/apphandlers/embeddedAppHandler.go
```diff
@@ -219,7 +232,7 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 		Reporting:        reportingI,
 		BackendConfig:    backendconfig.DefaultBackendConfig,
 		RouterDB:         routerDB,
-		ProcErrorDB:      errDB,
+		ProcErrorDB:      errDBForRead,
```
atzoum, line 235: Except for processor, which both reads (stash loop) and writes, all other components are only writing to `proc_errors`, correct?

app/apphandlers/processorAppHandler.go
```diff
@@ -169,14 +169,23 @@ func (a *processorApp) StartRudderCore(ctx context.Context, options *app.Options
...
 		jobsdb.WithClearDB(options.ClearDB),
 		jobsdb.WithPreBackupHandlers(prebackupHandlers),
 		jobsdb.WithDSLimit(&a.config.processorDSLimit),
 		jobsdb.WithFileUploaderProvider(fileUploaderProvider),
 		jobsdb.WithSkipMaintenanceErr(config.GetBool("Processor.jobsDB.skipMaintenanceError", false)),
 	)
+	defer errDBForRead.Close()
+	errDBForWrite := jobsdb.NewForWrite(
+		"proc_error",
+		jobsdb.WithClearDB(options.ClearDB),
```
atzoum, line 183: I think this needs to be included too
```
jobsdb.WithSkipMaintenanceErr(config.GetBool("Processor.jobsDB.skipMaintenanceError", false)),
```

app/apphandlers/gatewayAppHandler.go
```diff
@@ -89,6 +89,20 @@ func (a *gatewayApp) StartRudderCore(ctx context.Context, options *app.Options)
 	}
 	defer gatewayDB.Stop()
 
+	errDB := jobsdb.NewForWrite(
+		"proc_error",
+		jobsdb.WithClearDB(options.ClearDB),
+		jobsdb.WithDSLimit(&a.config.gatewayDSLimit), // TODO: use better config
+		jobsdb.WithSkipMaintenanceErr(config.GetBool("Gateway.jobsDB.skipMaintenanceError", false)),
```
atzoum, line 96:
```suggestion
		jobsdb.WithSkipMaintenanceErr(config.GetBool("Gateway.jobsDB.skipMaintenanceError", true)),
```

app/apphandlers/gatewayAppHandler.go
```diff
@@ -89,6 +89,20 @@ func (a *gatewayApp) StartRudderCore(ctx context.Context, options *app.Options)
 	}
 	defer gatewayDB.Stop()
 
+	errDB := jobsdb.NewForWrite(
+		"proc_error",
+		jobsdb.WithClearDB(options.ClearDB),
+		jobsdb.WithDSLimit(&a.config.gatewayDSLimit), // TODO: use better config
```
atzoum, line 95: `dsLimit` is not of much use for a writer jobsdb. This is an option that affects queries

gateway/gateway.go
```diff
@@ -1491,6 +1494,38 @@ func (*HandleT) GetWebhookSourceDefName(writeKey string) (name string, ok bool)
 	return
 }
 
+// SaveErrors saves errors to the error db
+func (gateway *HandleT) SaveWebhookFailures(reqs []*model.FailedWebhookPayload) error {
```
atzoum, line 1498: does this method needs to be exported?

gateway/gateway.go
```diff
@@ -1491,6 +1494,38 @@ func (*HandleT) GetWebhookSourceDefName(writeKey string) (name string, ok bool)
...
+		if err != nil {
+			gateway.logger.Errorf("[Gateway] Failed to marshal parameters map. Parameters: %+v", params)
+			marshalledParams = []byte(`{"error": "rudder-server gateway failed to marshal params"}`)
+		}
+
+		jobs = append(jobs, &jobsdb.JobT{
+			UUID:         uuid.New(),
+			UserID:       uuid.New().String(), // TODO remind for a better value
+			Parameters:   marshalledParams,
+			CustomVal:    CustomVal,
```
atzoum, line 1517: Does this need to be `GW` or something else? How will the current replay process know how to exclude these events in case we are replaying gateway events @cisse21?

app/apphandlers/embeddedAppHandler.go
chandumlg: Oops. Yes, this must be errDBForWrite.

gateway/gateway.go
chandumlg: In general, we use both `customVal` column and `stage` field in `parameters` column to replay. 

In this case, `stage` is set to webhook and `sourceType` is set to source definition name, which can be used for replays. 

If this requires a better `customVal`, pls mention. I am open to change it.

gateway/gateway.go
chandumlg: Yes. Can't pass `gateway` to webhook.Setup (in gateway.go) without exporting.

Also, there were similar functions in that interface exported.

app/apphandlers/gatewayAppHandler.go
chandumlg: Removed.

app/apphandlers/processorAppHandler.go
chandumlg: Actually, writer doesn't perform any maintenance, so not required right?

app/apphandlers/gatewayAppHandler.go
chandumlg: Removing this line altogether. Since writer doesn't perform any maintenance. (same as above comment)

app/apphandlers/embeddedAppHandler.go
chandumlg: Oops. Yes you are right.

Both rt & brt should recive errDBForWrite and not Read. Corrected.

gateway/gateway.go
chandumlg: Let's use `WEBHOOK` or source definition name?

app/apphandlers/embeddedAppHandler.go
```diff
@@ -243,7 +256,7 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 		GatewayDB:       gwDBForProcessor,
 		RouterDB:        routerDB,
 		BatchRouterDB:   batchRouterDB,
-		ErrorDB:         errDB,
+		ErrorDB:         errDBForRead,
```
atzoum, line 259: Doesn't it need both dbs, read & write?

atzoum, line 259: @cisse21 does migrator migrate jobs from `proc_errors`? I assume not, correct?

app/apphandlers/processorAppHandler.go
atzoum: `dsMigrationLoop` is part of jobsDB's maintenance

gateway/gateway.go
atzoum: @cisse21 ?

cisse21: When we replay from `proc_errors` we use a field called stage as chandu mentioned we should be good enough, also as we don't have any runbook to replay these events for now we should be good too

CustomVal in the existing replay shouldn't matter because GW replay and proc_error replay use different files altogether with different file formats and file structures but setting it as WEBHOOK makes more sense for someone looking at the backed up entries

app/apphandlers/embeddedAppHandler.go
cisse21: Yes migrator does migrate jobs from proc_errors

atzoum: @chandumlg maybe we need another jobsdb for gw errors

cisse21: Let me know if this doesn't make sense. We can discuss this is over a call

atzoum: so no issue?

app/apphandlers/gatewayAppHandler.go
atzoum: please see my reply above :)

app/apphandlers/processorAppHandler.go
atzoum: Sorry, my previous comment was inaccurate. Here is a list of maintenance operations:
- **Writer** jobsdb: `addNewDSLoop`
- **Reader** jobsdb: `backupDSLoop`, `refreshDSListLoop`, `migrateDSLoop`

gateway/gateway.go
```diff
@@ -1491,6 +1494,38 @@ func (*HandleT) GetWebhookSourceDefName(writeKey string) (name string, ok bool)
...
+			gateway.logger.Errorf("[Gateway] Failed to marshal parameters map. Parameters: %+v", params)
+			marshalledParams = []byte(`{"error": "rudder-server gateway failed to marshal params"}`)
+		}
+
+		jobs = append(jobs, &jobsdb.JobT{
+			UUID:         uuid.New(),
+			UserID:       uuid.New().String(), // Using a random userid for these failures. There is no notion of user id for these events.
+			Parameters:   marshalledParams,
+			CustomVal:    "WEBHOOK",
+			EventPayload: req.Payload,
```
cisse21, line 1518: Just thinking out loud... what if the req.Payload is not a JSON but rather XML data? or some other format

atzoum: Correct, payload should be stored as a valid `json`, otherwise store would be retried twice and ultimately fail

atzoum: A gateway test scenario using an actual jobsdb would be useful here

chandumlg: We validate the incoming payload to be JSON [here](https://github.com/rudderlabs/rudder-server/blob/a8abd1db5b6a9b434dd3363b3f959302ecdfc60b/gateway/webhook/webhook.go#L348)

Isn't that sufficient for now? In worse case scenario, if the store fails in first attempt, we log the error and continue. Server is not interrupted in any way.

atzoum (approved):

cisse21 (approved):

PR #3542, opened 2023-06-23
chore: send stats for the time taken for suppression

BonapartePC:
# Description

send stats for the time taken for suppression to be used from its creation

## Notion Ticket

https://www.notion.so/rudderstacks/Historical-suppression-stats-8638e9abab974ff5b2e62d182946aa99

## Security


gateway/gateway.go
```diff
@@ -788,11 +788,19 @@ func (gateway *HandleT) isUserSuppressed(workspaceID, userID, sourceID string) b
...
-	return gateway.suppressUserHandler.IsSuppressedUser(
+	isSuppressedUser := gateway.suppressUserHandler.IsSuppressedUser(
 		workspaceID,
 		userID,
 		sourceID,
 	)
+	if isSuppressedUser {
+		createdAt := gateway.suppressUserHandler.GetCreatedAt(workspaceID, userID, sourceID)
+		if !createdAt.IsZero() {
+			userSuppressionTimer := gateway.stats.NewTaggedStat("gateway.user_suppression_timer", stats.TimerType, stats.Tags{"workspaceID": workspaceID, "userID": userID, "sourceID": sourceID})
```
cisse21, line 799: Cardinality on userID would blow up the metrics

BonapartePC: Will remove it

cisse21: Another idea if we really need to capture details about userID could be that we can log if we get data from the suppressed userID after x number of day... say 60

@lvrach WDYT?

gateway/gateway.go
```diff
@@ -788,11 +788,19 @@ func (gateway *HandleT) isUserSuppressed(workspaceID, userID, sourceID string) b
...
+	isSuppressedUser := gateway.suppressUserHandler.IsSuppressedUser(
 		workspaceID,
 		userID,
 		sourceID,
 	)
+	if isSuppressedUser {
+		createdAt := gateway.suppressUserHandler.GetCreatedAt(workspaceID, userID, sourceID)
+		if !createdAt.IsZero() {
+			userSuppressionTimer := gateway.stats.NewTaggedStat("gateway.user_suppression_timer", stats.TimerType, stats.Tags{"workspaceId": workspaceID, "userID": userID, "sourceID": sourceID})
+			userSuppressionTimer.Since(createdAt)
```
saurav-malani, line 800:
```suggestion
			gateway.stats.NewTaggedStat("gateway.user_suppression_age", stats.TimerType, stats.Tags{"workspaceId": workspaceID, "userID": userID, "sourceID": sourceID}).Since(createdAt)
```

gateway/gateway.go
BonapartePC: @lvrach @atzoum

enterprise/suppress-user/handler.go
```diff
@@ -30,3 +31,12 @@ func (h *handler) IsSuppressedUser(workspaceID, userID, sourceID string) bool {
...
 }
+
+func (h *handler) GetCreatedAt(workspaceID, userID, sourceID string) time.Time {
+	h.log.Debugf("GetCreatedAt called for workspace: %s, user %s, source %s", workspaceID, userID, sourceID)
+	createdAt, err := h.r.GetCreatedAt(workspaceID, userID, sourceID)
+	if err != nil && !errors.Is(err, model.ErrRestoring) {
+		h.log.Errorf("GetCreatedAt failed for workspace: %s, user: %s, source: %s: %w", workspaceID, userID, sourceID, err)
+	}
+	return createdAt
+}
```
atzoum, line 42: I believe we don't need a new method for that

enterprise/suppress-user/internal/badgerdb/badgerdb.go
```diff
@@ -41,6 +42,10 @@ func WithMaxSeedWait(maxSeedWait time.Duration) Opt {
 	}
 }
 
+type suppressionMetadata struct {
+	CreatedAt string
+}
+
```
atzoum, line 48: this private type would be better to be placed towards the end of the file

enterprise/suppress-user/internal/badgerdb/badgerdb.go
```diff
@@ -176,7 +247,17 @@ func (b *Repository) Add(suppressions []model.Suppression, token []byte) error {
...
+					CreatedAt: suppression.CreatedAt.Format(time.RFC3339Nano),
+				}
+				value, err := json.Marshal(metadata)
+				if err != nil {
+					return fmt.Errorf("could not marshal suppression metadata: %w", err)
+				}
+				err = wb.Set([]byte(key), value)
+				if err != nil {
+					return fmt.Errorf("could not add key %s (canceled:%t) in write batch: %w", key, suppression.Canceled, err)
+				}
```
atzoum, line 260:
```suggestion
```

enterprise/suppress-user/internal/badgerdb/badgerdb.go
```diff
@@ -146,6 +151,72 @@ func (b *Repository) Suppressed(workspaceID, userID, sourceID string) (bool, err
 	return true, nil
 }
 
+func (b *Repository) GetCreatedAt(workspaceID, userID, sourceID string) (time.Time, error) {
```
atzoum, line 154: Instead of going through all this trouble, we may simply modify the Suppressed method signature to return a pointer to suppression metadata if it exists.

Note: backwards compatibility should be carefully thought out

atzoum (changes requested):
Any state maintained by suppression backup service needs to be reset as well during rollout, since so far this service has been creating backups without `createdAt`  information

enterprise/suppress-user/handler.go
BonapartePC: Removed it

enterprise/suppress-user/internal/badgerdb/badgerdb.go
BonapartePC: Done

enterprise/suppress-user/internal/badgerdb/badgerdb.go
```diff
@@ -333,3 +346,29 @@ func (l blogger) Warningf(fmt string, args ...interface{}) {
...
+	if err != nil {
+		return nil, fmt.Errorf("could not parse createdAt: %w", err)
+	}
+	return &model.Metadata{
+		CreatedAt: createdAt,
+	}, nil
+}
+
+type badgerDBValue struct {
+	CreatedAt string
```
atzoum, line 373: why not `time.Time` instead of `string`?

atzoum: After having a closer look, why do we even need the `badgerDBValue` struct?

gateway/gateway.go
```diff
@@ -788,11 +788,17 @@ func (gateway *HandleT) isUserSuppressed(workspaceID, userID, sourceID string) b
...
 		workspaceID,
 		userID,
 		sourceID,
 	)
+	if metadata != nil {
+		createdAt := metadata.CreatedAt
+		userSuppressionTimer := gateway.stats.NewTaggedStat("gateway.user_suppression_age", stats.TimerType, stats.Tags{"workspaceId": workspaceID, "sourceID": sourceID})
+		userSuppressionTimer.Since(createdAt)
+	}
+	return metadata != nil
```
atzoum, line 801:
```suggestion
	if metadata := gateway.suppressUserHandler.GetSuppressedUser(workspaceID, userID, sourceID); metadata != nil {
		gateway.stats.NewTaggedStat("gateway.user_suppression_age", stats.TimerType, stats.Tags{
			"workspaceId": workspaceID,
			"sourceID":    sourceID,
		}).Since(metadata.CreatedAt)
		return true
	}
	return false
```

enterprise/suppress-user/factory_test.go
```diff
@@ -118,3 +125,60 @@ func getSuppressionFromManager(w http.ResponseWriter, _ *http.Request) {
...
+func generateBackupFiles(t *testing.T, dir, backupFilename string, repo *badgerdb.Repository, suppressions []model.Suppression) {
+	token := []byte("__token__")
+
+	repo.Add(suppressions, token)
+
+	f, err := os.Create(path.Join(dir, backupFilename))
+	defer func() { _ = f.Close() }()
+	require.NoError(t, err)
+	require.NoError(t, repo.Backup(f))
+}
```
atzoum, line 175: can this be converted into a function local to `generateTestData`?

suppression-backup-service/cmd/main_test.go
```diff
@@ -53,7 +53,7 @@ func verifyBackup(t *testing.T, filePath, data string) {
 	require.NoError(t, err)
 	isSuppressed, err := repo.Suppressed("workspace-1", "user-1", "src-1")
 	require.NoError(t, err)
-	require.True(t, isSuppressed)
+	require.NotNil(t, isSuppressed)
```
atzoum, line 56: We need to be careful while rolling this out:
-Suppression backup service needs to be updated before gateways. 
- The service's pvc needs to be deleted as well so that no previous backup file is retained

atzoum: Another option would be to change paths for the full backup & export files so that the service doesn't pickup the old ones.

enterprise/suppress-user/internal/badgerdb/badgerdb.go
BonapartePC: I used this struct for changing metadata into bytes and vice versa. I used this because badbgerdb can only save byte array as value.

enterprise/suppress-user/factory_test.go
BonapartePC: Done

gateway/gateway.go
BonapartePC: Done

suppression-backup-service/cmd/main_test.go
BonapartePC: Made change in this commit: https://github.com/rudderlabs/rudder-server/pull/3542/commits/de8d3e3538284164ba15eeca75c80140209548f0

enterprise/suppress-user/internal/badgerdb/badgerdb.go
atzoum: wouldn't a marshalled/unmarshalled `model.Metadata` suffice and take away the need for a new struct and custom data parsing?

BonapartePC: Yeah, made the change

enterprise/suppress-user/internal/badgerdb/badgerdb.go
```diff
@@ -333,3 +346,21 @@ func (l blogger) Warningf(fmt string, args ...interface{}) {
...
+	if len(itemValue) == 0 {
+		return nil, nil
+	}
+	err = json.Unmarshal(itemValue, &metadata)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal metadata: %w", err)
+	}
+	return &model.Metadata{
+		CreatedAt: metadata.CreatedAt,
+	}, nil
```
atzoum, line 365: do we really need to create yet-another-`model.Metadata` here?

suppression-backup-service/cmd/main_test.go
atzoum: Changing the path for exported files is one aspect, changing the path for badgerDB is another :) 

BadgerDB for `fullsuppression` survives restarts, whereas badgerDB for `latestsuppression` doesn't.

enterprise/suppress-user/internal/badgerdb/badgerdb.go
atzoum, line 365:
```suggestion
	return &metadata, nil
```

BonapartePC: Done

enterprise/suppress-user/internal/badgerdb/badgerdb.go
```diff
@@ -333,3 +346,19 @@ func (l blogger) Warningf(fmt string, args ...interface{}) {
...
+
+func getMetadataFromBadgerItem(item *badger.Item) (*model.Metadata, error) {
+	itemValue, err := item.ValueCopy(nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not copy item value: %w", err)
+	}
+	var metadata model.Metadata
+	if len(itemValue) == 0 {
+		return nil, nil
+	}
```
atzoum, line 358: A previous badgerDB version will have an empty string as a value. In that case we need to return a zero-value metadata, not `nil`
```suggestion
	if len(itemValue) == 0 { // backwards compatibility
		return &metadata, nil
	}
```

BonapartePC: Done

atzoum (approved):

enterprise/suppress-user/internal/badgerdb/badgerdb.go
```diff
@@ -176,7 +181,15 @@ func (b *Repository) Add(suppressions []model.Suppression, token []byte) error {
 			if suppression.Canceled {
 				err = wb.Delete([]byte(key))
 			} else {
-				err = wb.Set([]byte(key), []byte(""))
+				var value []byte
+				metadata := model.Metadata{
+					CreatedAt: suppression.CreatedAt,
+				}
+				value, err = json.Marshal(metadata)
```
cisse21, line 188: We could probably use the jsoniter library which we have been using everywhere

WDYT @atzoum

cisse21 (approved):

PR #3552, opened 2023-06-26
chore: load file upload

achettyiitr:
# Description

- Remove globals: numLoadFileUploadWorkers, slaveUploadTimeout, pkgLogger.
- Minor refactoring for uploadLoadFiles.

## Notion Ticket

https://www.notion.so/rudderstacks/Metrics-with-same-name-and-label-values-16f1ee308ee545848bc12ad63f990f96?pvs=4

## Security


warehouse/slave.go
```diff
@@ -130,45 +148,45 @@ func (job *Payload) getFileManager(config interface{}, useRudderStorage bool) (f
...
 
 		err = downloader.Download(ctx, file, job.StagingFileLocation)
 		if err != nil {
-			pkgLogger.Errorf("[WH]: Failed to download file")
+			jr.logger.Errorf("[WH]: Failed to download file")
 			return err
 		}
 		file.Close()
-		jobRun.timerStat("download_staging_file_time").Since(downloadStart)
+		jr.timerStat("download_staging_file_time").Since(downloadStart)
```
fracasula, line 174: Although of minimal impact, recreating the same metric over and over has a small cost due to the usage of [mutexes](https://github.com/rudderlabs/rudder-go-kit/blob/v0.15.1/stats/otel.go#L329). I suggest we initialize this metrics only once. This one in particular shouldn't even be a `jobRun` one in my opinion, but it can be attached to something else upstream (e.g. like a `warehouseApp` for example :stuck_out_tongue:).

warehouse/slave.go
```diff
@@ -189,126 +207,170 @@ func (job *Payload) getDiscardsTable() string {
...
+
+	type result struct {
+		uploadOutput loadFileUploadOutput
+		err          error
+	}
+
+	type uploadJob struct {
+		tableName  string
+		outputFile encoding.LoadFileWriter
+	}
```
fracasula, line 238: My preference would be to have these outside the `uploadLoadFiles` method, which is already quite long.

warehouse/slave.go
```diff
@@ -189,126 +207,170 @@ func (job *Payload) getDiscardsTable() string {
...
+		}
+	}()
+
+	generate := func() <-chan *uploadJob {
+		jobStream := make(chan *uploadJob, len(jr.outputFileWritersMap))
+		go func() {
+			defer close(jobStream)
+
+			for tableName, loadFile := range jr.outputFileWritersMap {
 				select {
```
fracasula, line 256: With this pattern we are not honouring the context. We're doing:

```go
select {
case <-ctx.Done():
	return
default:
	jobStream <- &uploadJob{
		tableName:  tableName,
		outputFile: loadFile,
	}
}
```

We check `ctx.Done`, if not done we proceed with the `default`. In the `default` we try to push into the `jobStream` channel. The problem is that we cannot foresee how long the pushing is going to take. The context might get canceled before we have the chance to push into the channel. What's even worse is that, due to the same context cancelation, what if the reader of `jobStream` stops reading while the channel happened to be full? We would have a leak here since we would be stuck in the `default` unable to proceed with the execution in the routine.

In order to avoid the leak, all channel operations must be done via the `select`:

```go
select {
case <-ctx.Done():
	return
case jobStream <- &uploadJob{
	tableName:  tableName,
	outputFile: loadFile,
}:
}
```

It is true that any of the two cases of the `select` can be picked but that's unavoidable and not a problem in my opinion. This means that if both the context is done and `jobStream` is not closed (and has buffer available), we might push into the channel instead of returning due to the context cancelation. There is no way to prevent that but it's OK because A) we would still return and not block and B) right after pushing at the next iteration (or if there are no more iterations) we close `jobStream`.

fracasula (commented):
Do you already know if this fixes the "metrics with same name and label values" issue or it's just an attempt?

achettyiitr:
> Do you already know if this fixes the "metrics with same name and label values" issue or it's just an attempt?

It's just an attempt. I tried to reproduce the error using the Unit test but was not able to.

warehouse/slave.go
```diff
@@ -189,126 +207,170 @@ func (job *Payload) getDiscardsTable() string {
...
+			uploadLocation, err = uploader.Upload(
+				ctx,
+				file,
+				jr.loadObjectFolder,
+				tableName,
+				jr.job.SourceID,
+				getBucketFolder(jr.job.UniqueLoadGenID, tableName),
+			)
+		}
+		return uploadLocation, err
```
fracasula, line 295:
```suggestion
		if slices.Contains(warehouseutils.TimeWindowDestinations, jr.job.DestinationType) {
			return uploader.Upload(
				ctx,
				file,
				warehouseutils.GetTablePathInObjectStorage(jr.job.DestinationNamespace, tableName),
				jr.job.LoadFilePrefix,
			)
		}
		
		return uploader.Upload(
			ctx,
			file,
			jr.loadObjectFolder,
			tableName,
			jr.job.SourceID,
			getBucketFolder(jr.job.UniqueLoadGenID, tableName),
		)
```

warehouse/slave.go
```diff
@@ -189,126 +207,170 @@ func (job *Payload) getDiscardsTable() string {
...
+		g, _ := errgroup.WithContext(ctx)
+		g.SetLimit(jr.numLoadFileUploadWorkers)
+
+		for i := 0; i < jr.numLoadFileUploadWorkers; i++ {
+			g.Go(func() error {
+				for uploadJob := range uploadJobChan {
+					select {
+					case <-ctx.Done():
+						return ctx.Err()
+					default:
```
fracasula, line 310: Why not just do:

```go
for {
	select {
	case <-ctx.Done():
		return ctx.Err()
	case uploadJob := <-uploadJobChan:
		// ...
	}
}
```

warehouse/slave.go
```diff
@@ -189,126 +207,170 @@ func (job *Payload) getDiscardsTable() string {
...
+						}
 
-	// Wait for response
-	for {
-		select {
-		case loadFileOutput := <-loadFileOutputChan:
-			loadFileUploadOutputs = append(loadFileUploadOutputs, loadFileOutput)
-			if len(loadFileUploadOutputs) == len(jobRun.outputFileWritersMap) {
-				return loadFileUploadOutputs, nil
+						mu.Lock()
```
fracasula, line 322: I found this code a bit confusing. First of all the `mu` name does not tell what variable it is supposed to protect. It is a common/good practice to do so. In this case it should be `totalUploadTimeMu`.

The other confusing thing is that we're protecting against writes only, perhaps leave a comment saying you expect to read the variable only once on defer after all routines are terminated.

Alternatively we could remove the double variable and just do `var totalUploadTime atomic.Duration`. Wdyt?

warehouse/slave.go
```diff
@@ -189,126 +207,170 @@ func (job *Payload) getDiscardsTable() string {
...
-		}(ctx)
+		}()
+		return jobStream
 	}
-	// Create upload jobs
-	go func() {
-		for tableName, loadFile := range jobRun.outputFileWritersMap {
-			uploadJobChan <- &loadFileUploadJob{tableName: tableName, outputFile: loadFile}
+
+	uploadLoadFile := func(uploader filemanager.FileManager, uploadFile encoding.LoadFileWriter, tableName string) (filemanager.UploadedFile, error) {
```
fracasula, line 270: Since this is a lambda we can simplify this a bit and not pass `uploader`. Same thing you did on `process` down below.

warehouse/slave.go
```diff
@@ -189,126 +207,170 @@ func (job *Payload) getDiscardsTable() string {
...
+	var totalUploadTime time.Duration
+	var mu sync.Mutex
+
+	defer func() {
+		if totalUploadTime > 0 {
+			jr.timerStat("load_file_upload_time").SendTiming(totalUploadTime)
+		}
+	}()
+
+	generate := func() <-chan *uploadJob {
```
fracasula, line 250: I don't understand why we need `generate`. Why don't you iterate over `jr.outputFileWritersMap` inside `process` itself?
You are already leveraging the `group` limit to ensure the maximum number of workers so getting rid of `generate` should be simple and it would help us decrease the complexity of the `JobRun.uploadLoadFiles` method.

Have a look at the `git diff` below. I suggest we apply all the changes from the diff.

As a sidenote (which you'll find in the diff as well) you are using a group with context but you're ignoring the context. If you don't plan to use it then you could just use a simple group or do the synchronisation natively. Since I think the group context in this case it's going to be useful, I've included it in the diff:

```diff
diff --git a/warehouse/slave.go b/warehouse/slave.go
index 06952f3de..467cb6939 100644
--- a/warehouse/slave.go
+++ b/warehouse/slave.go
@@ -247,63 +247,48 @@ func (jr *JobRun) uploadLoadFiles(ctx context.Context) ([]loadFileUploadOutput,
 		}
 	}()
 
-	generate := func() <-chan *uploadJob {
-		jobStream := make(chan *uploadJob, len(jr.outputFileWritersMap))
-		go func() {
-			defer close(jobStream)
-
-			for tableName, loadFile := range jr.outputFileWritersMap {
-				select {
-				case <-ctx.Done():
-					return
-				default:
-					jobStream <- &uploadJob{
-						tableName:  tableName,
-						outputFile: loadFile,
-					}
-				}
-			}
-		}()
-		return jobStream
-	}
-
-	uploadLoadFile := func(uploader filemanager.FileManager, uploadFile encoding.LoadFileWriter, tableName string) (filemanager.UploadedFile, error) {
+	uploadLoadFile := func(
+		ctx context.Context, uploader filemanager.FileManager, uploadFile encoding.LoadFileWriter, tableName string,
+	) (filemanager.UploadedFile, error) {
 		file, err := os.Open(uploadFile.GetLoadFile().Name())
 		if err != nil {
-			return filemanager.UploadedFile{}, fmt.Errorf("opening file: %w", err)
+			return filemanager.UploadedFile{}, fmt.Errorf("opening file %s: %w", uploadFile.GetLoadFile().Name(), err)
 		}
-		defer file.Close()
+		defer func() { _ = file.Close() }()
 
-		var uploadLocation filemanager.UploadedFile
 		if slices.Contains(warehouseutils.TimeWindowDestinations, jr.job.DestinationType) {
-			uploadLocation, err = uploader.Upload(
+			return uploader.Upload(
 				ctx,
 				file,
 				warehouseutils.GetTablePathInObjectStorage(jr.job.DestinationNamespace, tableName),
 				jr.job.LoadFilePrefix,
 			)
-		} else {
-			uploadLocation, err = uploader.Upload(
-				ctx,
-				file,
-				jr.loadObjectFolder,
-				tableName,
-				jr.job.SourceID,
-				getBucketFolder(jr.job.UniqueLoadGenID, tableName),
-			)
 		}
-		return uploadLocation, err
+		return uploader.Upload(
+			ctx,
+			file,
+			jr.loadObjectFolder,
+			tableName,
+			jr.job.SourceID,
+			getBucketFolder(jr.job.UniqueLoadGenID, tableName),
+		)
 	}
 
-	process := func(uploadJobChan <-chan *uploadJob) <-chan *result {
+	process := func() <-chan *result {
 		processStream := make(chan *result, len(jr.outputFileWritersMap))
 
-		g, _ := errgroup.WithContext(ctx)
+		g, groupCtx := errgroup.WithContext(ctx)
 		g.SetLimit(jr.numLoadFileUploadWorkers)
 
-		for i := 0; i < jr.numLoadFileUploadWorkers; i++ {
-			g.Go(func() error {
-				for uploadJob := range uploadJobChan {
+		go func() {
+			defer close(processStream)
+
+			for tableName, loadFile := range jr.outputFileWritersMap {
+				uploadJob := &uploadJob{
+					tableName:  tableName,
+					outputFile: loadFile,
+				}
+				g.Go(func() error {
 					select {
 					case <-ctx.Done():
 						return ctx.Err()
@@ -311,6 +296,7 @@ func (jr *JobRun) uploadLoadFiles(ctx context.Context) ([]loadFileUploadOutput,
 						loadFileUploadStart := time.Now()
 
 						uploadOutput, err := uploadLoadFile(
+							groupCtx,
 							uploader,
 							uploadJob.outputFile,
 							uploadJob.tableName,
@@ -340,22 +326,19 @@ func (jr *JobRun) uploadLoadFiles(ctx context.Context) ([]loadFileUploadOutput,
 							},
 						}
 					}
-				}
-				return nil
-			})
-		}
+					return nil
+				})
+			}
 
-		go func() {
 			if err := g.Wait(); err != nil {
 				processStream <- &result{err: err}
 			}
-			close(processStream)
 		}()
 
 		return processStream
 	}
 
-	for processedJob := range process(generate()) {
+	for processedJob := range process() {
 		if err := processedJob.err; err != nil {
 			return nil, fmt.Errorf("uploading load file to object storage: %w", err)
 		}
```

As far as I can see the tests pass, please have a look :+1:

warehouse/slave.go
```diff
@@ -189,126 +207,170 @@ func (job *Payload) getDiscardsTable() string {
...
+
+		return processStream
+	}
+
+	for processedJob := range process(generate()) {
+		if err := processedJob.err; err != nil {
+			return nil, fmt.Errorf("uploading load file to object storage: %w", err)
 		}
+
+		output = append(output, processedJob.uploadOutput)
```
fracasula, line 363: We expect as many jobs as `jr.outputFileWritersMap`, right? Perhaps we could preallocate with a `output := make([]loadFileUploadOutput, len(jr.outputFileWritersMap))`.

warehouse/slave.go
```diff
@@ -189,126 +207,170 @@ func (job *Payload) getDiscardsTable() string {
...
+	}
+	if ctx.Err() != nil {
+		return nil, fmt.Errorf("uploading load file to object storage: %w", ctx.Err())
 	}
```
fracasula, line 367: The upload has already been done though, there is no computation left on our side here. I think we can remove this check and just return. The context can be checked upstream as needed, right?

warehouse/slave.go
```diff
@@ -387,31 +449,36 @@ func (event *BatchRouterEvent) GetColumnInfo(columnName string) (columnInfo ware
...
-
-	defer jobRun.counterStat("staging_files_processed", Tag{Name: "worker_id", Value: strconv.Itoa(workerIndex)}).Count(1)
+	jr := JobRun{
+		job:                      job,
+		whIdentifier:             warehouseutils.GetWarehouseIdentifier(job.DestinationType, job.SourceID, job.DestinationID),
+		stats:                    stats.Default,
+		since:                    time.Since,
+		numLoadFileUploadWorkers: config.GetInt("Warehouse.numLoadFileUploadWorkers", 8),
+		slaveUploadTimeout:       config.GetDuration("Warehouse.slaveUploadTimeout", 10, config.GetDuration("Warehouse.slaveUploadTimeoutInMin", 10, time.Minute)),
+		logger:                   pkgLogger,
```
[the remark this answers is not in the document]
achettyiitr: Yeah. I have already a PR in my local for this. I will share this prior this PR.

warehouse/slave.go
achettyiitr: Great point. Using `atomic.Duration`

warehouse/slave.go
achettyiitr: Currently, these tags are attached here. 

```GO
func (jr *jobRun) timerStat(name string, extraTags ...Tag) stats.Measurement {
	tags := stats.Tags{
		"module":      moduleName,
		"destType":    jr.job.DestinationType,
		"warehouseID": jr.warehouseID(),
		"workspaceId": jr.job.WorkspaceID,
		"destID":      jr.job.DestinationID,
		"sourceID":    jr.job.SourceID,
	}
	for _, extraTag := range extraTags {
		tags[extraTag.Name] = extraTag.Value
	}
	return jr.stats.NewTaggedStat(name, stats.TimerType, tags)
}
```

achettyiitr: Not sure whether it can be moved somewhere?

warehouse/slave.go
achettyiitr: Moved it outside.

warehouse/slave.go
achettyiitr: generate has been removed now.

warehouse/slave.go
achettyiitr: I was overcomplicating it using `generate` since it is doing nothing just adding the `uploadJob` to the channel which can be done in `process` as well.

warehouse/slave.go
achettyiitr: You mean something like this:

```GO
	processStream := process()

	for i := 0; i < len(jr.outputFileWritersMap); i++ {
		result := <-processStream
		
		if err := result.err; err != nil {
			return nil, fmt.Errorf("uploading load file to object storage: %w", err)
		}
		
		output[i] = result.result
	}
```

warehouse/slave.go
achettyiitr: Thanks for pointing out. It's not required.

warehouse/slave.go
```diff
@@ -44,54 +47,69 @@ const (
...
+	slaveUploadTimeout       time.Duration
+	logger                   logger.Logger
+	loadObjectFolder         string
 }
 
-func (jobRun *JobRun) setStagingFileReader() (reader *gzip.Reader, endOfFile bool) {
-	job := jobRun.job
-	pkgLogger.Debugf("Starting read from downloaded staging file: %s", job.StagingFileLocation)
-	stagingFile, err := os.Open(jobRun.stagingFilePath)
+type loadFileUploadOutput struct {
```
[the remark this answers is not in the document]
achettyiitr: Rearranges it.

warehouse/slave.go
```diff
@@ -189,126 +207,170 @@ func (job *Payload) getDiscardsTable() string {
...
+
+	var output []loadFileUploadOutput
+	var totalUploadTime time.Duration
+	var mu sync.Mutex
+
+	defer func() {
+		if totalUploadTime > 0 {
+			jr.timerStat("load_file_upload_time").SendTiming(totalUploadTime)
+		}
+	}()
```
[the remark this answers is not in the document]
achettyiitr: 1 staging file is broken down into multiple load files, I thought the error `Metrics with the same name and label values` was caused because we are sending the metrics multiple times from multiple goroutines. So, instead of sending metrics for each file, we can just send the total upload time metrics?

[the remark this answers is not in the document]
achettyiitr:
```GO
	defer func() {
		if err == nil {
			jr.timerStat("load_file_total_upload_time").SendTiming(totalUploadTime.Load())
		}
	}()
```

warehouse/slave.go
```diff
@@ -44,54 +47,69 @@ const (
...
-	whIdentifier         string
-	stats                stats.Stats
+	job                      Payload
+	stagingFilePath          string
+	uuidTS                   time.Time
+	outputFileWritersMap     map[string]encoding.LoadFileWriter
+	tableEventCountMap       map[string]int
+	stagingFileReader        *gzip.Reader
+	whIdentifier             string
+	stats                    stats.Stats
```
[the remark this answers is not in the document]
achettyiitr: Removed duplicates.

warehouse/slave.go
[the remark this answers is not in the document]
fracasula: But these timers aren't going to change once you've created the `JobRun`, right? How about creating them only once when you create the job run? Perhaps with a `JobRun` constructor. Wdyt?

warehouse/slave.go
achettyiitr: Good point.

warehouse/slave.go
fracasula: Yeah or you can still use `append` and preallocate with `make`: https://go.dev/play/p/a3dB2UQ71pl?v=

achettyiitr: Got it, Setting the capacity.

warehouse/slave.go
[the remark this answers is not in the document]
fracasula: Yes I understand why you redesigned this but I don't think that is the culprit. Also, we should be able to send multiple timings for the same timer concurrently. The underlying framework should handle that.

From a metrics perspective it is correct to measure how long it takes to upload each file. I think we should keep that like it was before so that we can have more useful metrics in our dashboards. Having the total is a plus and we could consider it if we think it is useful and we cannot derive it with a promql query.

warehouse/slave.go
achettyiitr: Made the changes.

warehouse/slave_test.go
```diff
@@ -43,3 +61,195 @@ func TestPickupStagingFileBucket(t *testing.T) {
...
+func (m *mockLoadFileWriter) GetLoadFile() *os.File {
+	return m.file
+}
+
+func TestUploadLoadFilesToObjectStorage(t *testing.T) {
+	pool, err := dockertest.NewPool("")
+	require.NoError(t, err)
+
+	ctx := context.Background()
+	ctxCancel, cancel := context.WithCancel(ctx)
```
fracasula, line 94:
```suggestion
	canceledCtx, cancel := context.WithCancel(context.Background())
```

warehouse/slave_test.go
```diff
@@ -43,3 +61,195 @@ func TestPickupStagingFileBucket(t *testing.T) {
...
+				return time.Second
+			}
+			jr.outputFileWritersMap = writerMap
+
+			ctx := ctx
+			if tc.ctx != nil {
+				ctx = tc.ctx
+			}
+
+			loadFIle, err := jr.uploadLoadFiles(ctx)
```
fracasula, line 231:
```suggestion
			ctx := context.Background()
			if tc.ctx != nil {
				ctx = tc.ctx
			}

			loadFile, err := jr.uploadLoadFiles(ctx)
```

warehouse/slave_test.go
```diff
@@ -43,3 +61,195 @@ func TestPickupStagingFileBucket(t *testing.T) {
...
+			if tc.wantError != nil {
+				require.EqualError(t, err, tc.wantError.Error())
+				return
+			}
```
fracasula, line 235: Could be simplified but you need to change `wantError` to `string`:

```suggestion
			if tc.wantError != "" {
				require.EqualError(t, err, tc.wantError)
				return
			}
```

warehouse/stats.go
```diff
@@ -257,12 +206,41 @@ func (job *UploadJob) recordLoadFileGenerationTimeStat(startID, endID int64) (er
...
+		"destType":    jr.job.DestinationType,
+		"warehouseID": warehouseTagName(jr.job.DestinationID, jr.job.SourceName, jr.job.DestinationName, jr.job.SourceID),
+		"workspaceId": jr.job.WorkspaceID,
+		"destID":      jr.job.DestinationID,
+		"sourceID":    jr.job.SourceID,
+	}
+}
+
+func (jr *jobRun) timerStat(name string, extraTags ...warehouseutils.Tag) stats.Measurement {
+	tags := jr.defaultTags()
```
fracasula, line 221: [optional] `defaultTags` could be `buildTags` and get the `extraTags` as well to avoid the repetition of the `for` loop as well.

warehouse/slave.go
```diff
@@ -185,183 +213,271 @@ func PickupStagingConfiguration(job *Payload) bool {
...
+			uploadLocation, err = uploader.Upload(
+				ctx,
+				file,
+				jr.config.loadObjectFolder,
+				tableName,
+				jr.job.SourceID,
+				getBucketFolder(jr.job.UniqueLoadGenID, tableName),
+			)
+		}
+		jr.uploadTimeStat.SendTiming(jr.since(uploadStart))
```
fracasula, line 272: [optional] could be done on `defer` to make the code easier to follow.

fracasula (approved):
LGTM, added a few comments, mostly optional.

lvrach (approved):

warehouse/slave.go
```diff
@@ -35,140 +41,162 @@ const (
 	statsWorkerClaimProcessingTime      = "worker_claim_processing_time"
 	statsWorkerClaimProcessingFailed    = "worker_claim_processing_failed"
 	statsWorkerClaimProcessingSucceeded = "worker_claim_processing_succeeded"
-	tagWorkerid                         = "workerId"
-)
 
-const (
+	tagWorkerid = "workerId"
```
lvrach, line 45: [minor]
```suggestion
	tagWorkerID = "workerId"
```

warehouse/slave.go
```diff
@@ -35,140 +41,162 @@ const (
...
+	TotalRows             int
+	ContentLength         int64
+	StagingFileID         int64
+	DestinationRevisionID string
+	UseRudderStorage      bool
+}
+
+type asyncJobRunResult struct {
+	Result bool
+	Id     string
```
lvrach, line 67: [minor]
```suggestion
	ID     string
```
We can use JSON tag if we marshal this

lvrach (commented):
super minor stuff

warehouse/slave.go
```diff
@@ -35,140 +41,166 @@ const (
...
+		since:        time.Since,
+		logger:       log,
+	}
+
+	jr.config.numLoadFileUploadWorkers = conf.GetInt("Warehouse.numLoadFileUploadWorkers", 8)
+	if conf.IsSet("Warehouse.slaveUploadTimeout") {
+		jr.config.slaveUploadTimeout = conf.GetDuration("Warehouse.slaveUploadTimeout", 10, time.Minute)
+	} else {
+		jr.config.slaveUploadTimeout = conf.GetDuration("Warehouse.slaveUploadTimeoutInMin", 10, time.Minute)
+	}
```
achettyiitr, line 110: Tried chaining but it is not working as expected:
```
jr.config.slaveUploadTimeout = conf.GetDuration("Warehouse.slaveUploadTimeout", 10, conf.GetDuration("Warehouse.slaveUploadTimeoutInMin", 10, time.Minute))
```
```
func TestGetDuration(t *testing.T) {
	{
		c := New()
		t.Log(c.GetDuration("durationInMin", 5, c.GetDuration("duration", 5, time.Minute)))
	}
	{
		c := New()
		c.Set("duration", "10m")
		t.Log(c.GetDuration("durationInMin", 5, c.GetDuration("duration", 5, time.Minute)))
	}
	{
		c := New()
		c.Set("durationInMin", "20m")
		t.Log(c.GetDuration("durationInMin", 5, c.GetDuration("duration", 5, time.Minute)))
	}
	{
		c := New()
		c.Set("duration", "10m")
		c.Set("durationInMin", "20m")
		t.Log(c.GetDuration("durationInMin", 5, c.GetDuration("duration", 5, time.Minute)))
	}
}
    config_test.go:327: 25m0s
    config_test.go:332: 50m0s
    config_test.go:337: 20m0s
    config_test.go:343: 20m0s
```

PR #3554, opened 2023-06-27
chore: gateway_response_time buckets

fracasula:
# Description

I don't think we need as many buckets as we do by default for the `gateway_response_time metric`. This task is to keep only the relevant buckets.

I propose we remove the boundaries for requests that take longer than 5 minutes. Data wise, knowing that it takes even longer isn't helpful in my opinion and we should try to avoid anything that goes above 1 minute and possibly look into anything that goes above 5.

Screenshot from `prousmt`:
Screenshot from 2023-06-27 11-19-04

Tested locally:
Screenshot from 2023-06-27 14-59-57

Let me know if the boundaries make sense to you and if you think we should have different ones.
Also, if you think we should do the same with other metrics we might as well add them here.

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/gw-response-time-buckets-9ff53d3d1a334d359c61b5e914825a5d) >

## Security

### DIFF
--- a/integration_test/kafka_batching/kafka_batching_test.go
+++ b/integration_test/kafka_batching/kafka_batching_test.go
@@ -319,7 +319,7 @@ func TestKafkaBatching(t *testing.T) {
 			{CumulativeCount: ptr(uint64(0)), UpperBound: ptr(2.5)},
 			{CumulativeCount: ptr(uint64(0)), UpperBound: ptr(5.0)},
 			{CumulativeCount: ptr(uint64(1)), UpperBound: ptr(10.0)}, // 10 is the number of messages we sent
-			{CumulativeCount: ptr(uint64(1)), UpperBound: ptr(30.0)}, // 10 is the number of messages we sent
+			{CumulativeCount: ptr(uint64(1)), UpperBound: ptr(30.0)},
 			{CumulativeCount: ptr(uint64(1)), UpperBound: ptr(60.0)},
 			{CumulativeCount: ptr(uint64(1)), UpperBound: ptr(300.0)},
 			{CumulativeCount: ptr(uint64(1)), UpperBound: ptr(600.0)},
--- a/integration_test/kafka_batching/kafka_batching_test.go
+++ b/integration_test/kafka_batching/kafka_batching_test.go
@@ -307,6 +307,7 @@ func TestKafkaBatching(t *testing.T) {
 	requireHistogramEqual(t, metrics["router_kafka_batch_size"], histogram{
 		name: "router_kafka_batch_size", count: 1, sum: 10,
 		buckets: []*promClient.Bucket{
+			{CumulativeCount: ptr(uint64(0)), UpperBound: ptr(0.002)},
 			{CumulativeCount: ptr(uint64(0)), UpperBound: ptr(0.005)},
 			{CumulativeCount: ptr(uint64(0)), UpperBound: ptr(0.01)},
 			{CumulativeCount: ptr(uint64(0)), UpperBound: ptr(0.025)},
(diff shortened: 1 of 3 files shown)


psrikanth88 (approved):

runner/runner.go
```diff
@@ -131,11 +131,15 @@ func (r *Runner) Run(ctx context.Context, args []string) int {
...
 		config.Set("statsExcludedTags", []string{"workspaceId", "sourceID", "destId"})
 	}
-	stats.Default = stats.NewStats(config.Default, logger.Default, svcMetric.Instance,
+	statsOptions := []stats.Option{
 		stats.WithServiceName(r.appType),
 		stats.WithServiceVersion(r.releaseInfo.Version),
 		stats.WithDefaultHistogramBuckets(defaultHistogramBuckets),
-	)
+	}
+	for histogramName, buckets := range customBuckets {
```
mcchran, line 139: should we consider moving to the use of `summary` instead of `histograms`?

runner/buckets.go
```diff
@@ -0,0 +1,8 @@
+package runner
+
+var customBuckets = map[string][]float64{
+	"gateway.response_time": {
+		0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 60,
+		300, /* 5 mins */
```
atzoum, line 6: We have a timeout of 10sec, thus 300 might not be really required

runner/runner.go
```diff
@@ -131,11 +131,15 @@ func (r *Runner) Run(ctx context.Context, args []string) int {
 		(!config.IsSet("WORKSPACE_NAMESPACE") || strings.Contains(config.GetString("WORKSPACE_NAMESPACE", ""), "free")) {
 		config.Set("statsExcludedTags", []string{"workspaceId", "sourceID", "destId"})
 	}
-	stats.Default = stats.NewStats(config.Default, logger.Default, svcMetric.Instance,
+	statsOptions := []stats.Option{
 		stats.WithServiceName(r.appType),
 		stats.WithServiceVersion(r.releaseInfo.Version),
 		stats.WithDefaultHistogramBuckets(defaultHistogramBuckets),
```
atzoum, line 137: For `defaultHistogramBuckets` we might need 2 more, `0.002` for really fast operations & `20` for slower ones but not slower than 1 min.

runner/buckets.go
fracasula: Why do we have the histogram showing requests taking much more than that then? :thinking: 
In the PR description I posted a screenshot depicting that.

runner/runner.go
fracasula: I'm trying to decrese the buckets if possible :sweat_smile: 
What do you think are the use cases for the `0.002` and `20` buckets?

atzoum: https://rudderlabs.slack.com/archives/C01HTT66UMB/p1687271131609569?thread_ts=1687161567.768049&cid=C01HTT66UMB

runner/buckets.go
atzoum: Http server is configured with a [WriteTimeout](https://github.com/rudderlabs/rudder-server/blob/master/gateway/gateway.go#L1387) of 10sec.
Even if we don't honour this at the handler and take more than 10sec to prepare a response for a client's request, my understanding is that the client will never receive this response, but will receive a proxy error instead

fracasula: If that is the case then yeah, anything >10s is actionable from a metrics perspective. I can remove 300 but I'll leave 60 just in case because the `WriteTimeout` is configurable after all, perhaps some customer wants to increase the timeout. In that case at least we will have it covered until one minute.

fracasula:
@achettyiitr can you please review and let me know if the buckets for warehouse make sense or if we can remove some? My guess is that the ones on the microsecond precision aren't going to be useful, right? Possibly others as well or you reckon we might have timers that do measure in that range anyway?

achettyiitr (approved):

achettyiitr:
> @achettyiitr can you please review and let me know if the buckets for warehouse make sense or if we can remove some? My guess is that the ones on the microsecond precision aren't going to be useful, right? Possibly others as well or you reckon we might have timers that do measure in that range anyway?

microsecond precision is not required.

integration_test/kafka_batching/kafka_batching_test.go
```diff
@@ -318,16 +319,11 @@ func TestKafkaBatching(t *testing.T) {
 			{CumulativeCount: ptr(uint64(0)), UpperBound: ptr(2.5)},
 			{CumulativeCount: ptr(uint64(0)), UpperBound: ptr(5.0)},
 			{CumulativeCount: ptr(uint64(1)), UpperBound: ptr(10.0)}, // 10 is the number of messages we sent
+			{CumulativeCount: ptr(uint64(1)), UpperBound: ptr(30.0)}, // 10 is the number of messages we sent
```
atzoum, line 322:
```suggestion
			{CumulativeCount: ptr(uint64(1)), UpperBound: ptr(30.0)},
```

atzoum (approved):

PR #3559, opened 2023-06-27
chore(jobsdb): dsList lock performance improvements

atzoum:
# Description
Acquiring a dsList write lock when not needed and while jobsdb queries are performing slow can have cascading effects on throughput, thus we now acquire one only when necessary and we are acquiring read locks for shorter periods as well:

- During `addNewDSLoop` jobsdb acquires a `dsList` lock only if its needs to add a new dataset.
- During `refreshDSListLoop` jobsdb acquires a `dsList` lock only if it detects a new dataset.
- While reading, jobsdb releases the `dsList` rlock early. The `migration` rlock that has already been acquired will make sure that no existing dataset will be removed from the list. Any new (rightmost) dataset can be added without any adverse effect.
- While updating job statuses, josdb releases the `dsList` rlock early. The `migration` rlock provides the same guarantees as above.
- Increase default values for jobsdb maxWriters (3) & maxReaders (6)
- Increase jobsdb cache ttl to 60 minutes
- Increase `jobMinRowsMigrateThres` to 20%

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/Pipelines-Sprint-68c1c213e9b840b3a9b3826e3631e39a?p=a5e82a768d524195a5aa34e0505fcc99&pm=s)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -669,7 +669,7 @@ func loadConfig() {
 	*/
 	config.RegisterFloat64ConfigVariable(0.8, &jobDoneMigrateThres, true, "JobsDB.jobDoneMigrateThres")
 	config.RegisterFloat64ConfigVariable(3, &jobStatusMigrateThres, true, "JobsDB.jobStatusMigrateThres")
-	config.RegisterFloat64ConfigVariable(0.05, &jobMinRowsMigrateThres, true, "JobsDB.jobMinRowsMigrateThres")
+	config.RegisterFloat64ConfigVariable(0.2, &jobMinRowsMigrateThres, true, "JobsDB.jobMinRowsMigrateThres")
 	config.RegisterIntConfigVariable(100000, &maxDSSize, true, 1, "JobsDB.maxDSSize")
 	config.RegisterIntConfigVariable(10, &maxMigrateOnce, true, 1, "JobsDB.maxMigrateOnce")
 	config.RegisterIntConfigVariable(10, &maxMigrateDSProbe, true, 1, "JobsDB.maxMigrateDSProbe")
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -873,7 +873,7 @@ func (jd *HandleT) init() {
 				jd.runAlwaysChangesets(templateData)
 
 				// finally refresh the dataset list to make sure [datasetList] field is populated
-				_, err := jd.refreshDSList(l)
+				_, err := jd.doRefreshDSList(l)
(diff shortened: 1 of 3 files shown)


jobsdb/jobsdb.go
```diff
@@ -860,7 +860,7 @@ func (jd *HandleT) init() {
 				jd.runAlwaysChangesets(templateData)
 
 				// finally refresh the dataset list to make sure [datasetList] field is populated
-				_, err := jd.refreshDSList(l)
+				err := jd.doRefreshDSRangeList(l)
```
atzoum, line 863: **Note:** this was actually a bug, but not a critical one, since the ds range list in gateway pods is not of much importance

jobsdb/jobsdb.go
```diff
@@ -963,7 +963,7 @@ func (jd *HandleT) readerSetup(ctx context.Context, l lock.LockToken) {
 	// This is a thread-safe operation.
 	// Even if two different services (gateway and processor) perform this operation, there should not be any problem.
 	jd.recoverFromJournal(ReadWrite)
-	jd.assertError(jd.refreshDSRangeList(l))
+	jd.assertError(jd.doRefreshDSRangeList(l))
```
atzoum, line 966: note: this is a bug, readers need to start with consistent ds & dsRange lists

lvrach (approved):

jobsdb/jobsdb.go
```diff
@@ -2556,6 +2577,11 @@ func (jd *HandleT) addNewDSLoop(ctx context.Context) {
 							if err = setReadonlyDsInTx(tx, latestDS); err != nil {
 								return fmt.Errorf("error making dataset read only: %w", err)
 							}
+						} else {
+							// maybe another node added a new DS that we need to make visible to us
+							if err := jd.refreshDSList(ctx); err != nil {
+								return fmt.Errorf("refreshDSList: %w", err)
+							}
```
Sidddddarth, line 2584: Why can't we use the fresh `dsList` we fetch above in line.2549 to set `jd.dsList` here. Would need the `asyncLock` here as well.

Is it because `migration` might change how the tables look..?

Sidddddarth (approved):

jobsdb/jobsdb.go
Sidddddarth: Since the last dataset and the one that could be potentially created is(are) of interest, I think something like what you've done in `doRefreshDSList` can be done(comparing the last datasets in both arrays).

As for the rangeList, it's refreshed once the transaction ends anyway, so why here again?

PR #3571, opened 2023-06-29
chore(router): penalise continuous non-productive pickup loops by sleeping

atzoum:
# Description

Currently, pickup loops with large discard rates which also reach their query limits consistently are running non-stop, potentially causing unproductive stress to the database, by querying it constantly without any significant positive impact.

This change introduces a sleep-based penalty for such pickup loops: If limits have been reached during the pickup loop and the discard ratio is greater than 60%, the loop will sleep for `1sec` before its next attempt.

## Notion Ticket

[Link](https://www.notion.so/rudderstacks/Pipelines-Sprint-68c1c213e9b840b3a9b3826e3631e39a?p=a0ec9a617fdd4dc589f9116aefe06d4b&pm=s)

## Security


cisse21 (approved):

router/handle.go
```diff
@@ -242,6 +249,14 @@ func (rt *Handle) pickup(ctx context.Context, partition string, workers []*worke
...
 	limitsReached = iteratorStats.LimitsReached
+	discardedRatio := float64(iteratorStats.DiscardedJobs) / float64(iteratorStats.TotalJobs)
+
+	// If the discarded ratio is greater than the penalty threshold,
+	// sleep for a while to avoid having a loop running continuously without producing events
+	if limitsReached && discardedRatio > rt.reloadableConfig.failingJobsPenaltyThreshold {
+		limiterEnd()
+		_ = misc.SleepCtx(ctx, rt.reloadableConfig.failingJobsPenaltySleep)
+	}
 	rt.pipelineDelayStats(partition, firstJob, lastJob)
```
lvrach, line 260: Would it make to move the logic in the function that calls pickup, or in a wrapper function, so that `limiterEnd()` remains on defer?

atzoum: since you noticed I will do it properly in rudder-go-kit :)

atzoum: https://github.com/rudderlabs/rudder-go-kit/pull/71

lvrach (approved):

PR #3580, opened 2023-07-04
fix: jobsdb panics during recovery after backup failure(s)

BonapartePC:
# Description

Journal operation`BACKUP_DS` is currently being marked and completed without a transaction. Due to this, we are making multiple entries of journal start for `BACKUP_DS` when we get an error in `backupDs` function. With this change, any error within a transaction should roll back the change.

## Notion Ticket
https://www.notion.so/rudderstacks/backup-failure-causing-jobsdb-panic-adf15a01e88241699be18ca53d9ab38c

## Security


jobsdb/backup.go
```diff
@@ -72,37 +72,42 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
...
-			err = jd.JournalMarkDone(opID)
-			if err != nil {
-				return fmt.Errorf("mark end of backup operation: %w", err)
+			if err := jd.WithTx(func(tx *Tx) error {
+				opID, err := jd.JournalMarkStartInTx(tx, backupDSOperation, opPayload)
+				if err != nil {
+					return fmt.Errorf("mark start of backup operation: %w", err)
+				}
+				err = jd.backupDS(ctx, backupDSRange)
+				if err != nil {
```
atzoum, line 81:
```suggestion
				if err := jd.backupDS(ctx, backupDSRange); err != nil {
```

jobsdb/backup.go
```diff
@@ -72,37 +72,42 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
...
+					return fmt.Errorf("backup dataset: %w", err)
+				}
+				err = jd.journalMarkDoneInTx(tx, opID)
+				if err != nil {
```
atzoum, line 85:
```suggestion
				if err := jd.journalMarkDoneInTx(tx, opID); err != nil {
```

jobsdb/backup.go
```diff
@@ -72,37 +72,42 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
...
-				return fmt.Errorf(" drop dataset: %w", err)
-			}
-			err = jd.JournalMarkDone(opID)
-			if err != nil {
-				return fmt.Errorf("mark end of drop backup operation: %w", err)
-			}
-			return nil
+			return jd.WithTx(func(tx *Tx) error {
+				// drop dataset after successfully uploading both jobs and jobs_status to s3
+				opID, err := jd.JournalMarkStartInTx(tx, backupDSOperation, opPayload)
```
atzoum, line 95:
```suggestion
				opID, err := jd.JournalMarkStartInTx(tx, backupDropDSOperation, opPayload)
```

jobsdb/backup.go
```diff
@@ -72,37 +72,42 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
...
+				if err != nil {
+					return fmt.Errorf("mark start of backup operation: %w", err)
+				}
+				// Currently, we retry uploading a table for some time & if it fails. We only drop that table & not all `pre_drop` tables.
+				// So, in situation when new table creation rate is more than drop. We will still have pipe up issue.
+				// An easy way to fix this is, if at any point of time exponential retry fails then instead of just dropping that particular
+				// table drop all subsequent `pre_drop` table. As, most likely the upload of rest of the table will also fail with the same error.
+				if err = jd.dropDSInTx(tx, backupDS); err != nil {
```
atzoum, line 103: make a habit of this :) 
```suggestion
				if err := jd.dropDSInTx(tx, backupDS); err != nil {
```

jobsdb/backup.go
```diff
@@ -72,37 +72,42 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
...
+					return fmt.Errorf(" drop dataset: %w", err)
+				}
+				if err = jd.journalMarkDoneInTx(tx, opID); err != nil {
```
atzoum, line 106:
```suggestion
				if err := jd.journalMarkDoneInTx(tx, opID); err != nil {
```

jobsdb/backup.go
BonapartePC: Done

jobsdb/backup.go
BonapartePC: Done

jobsdb/backup.go
BonapartePC: Done

jobsdb/backup.go
BonapartePC: Done

jobsdb/backup.go
BonapartePC: Done

jobsdb/backup.go
```diff
@@ -72,37 +72,40 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
...
-			err = jd.JournalMarkDone(opID)
-			if err != nil {
-				return fmt.Errorf("mark end of drop backup operation: %w", err)
-			}
-			return nil
+			return jd.WithTx(func(tx *Tx) error {
+				// drop dataset after successfully uploading both jobs and jobs_status to s3
+				opID, err := jd.JournalMarkStartInTx(tx, backupDropDSOperation, opPayload)
+				if err != nil {
+					return fmt.Errorf("mark start of backup operation: %w", err)
```
atzoum, line 95:
```suggestion
					return fmt.Errorf("mark start of drop backup operation: %w", err)
```

jobsdb/backup.go
```diff
@@ -72,37 +72,40 @@ func (jd *HandleT) backupDSLoop(ctx context.Context) {
...
+				}
+				if err := jd.backupDS(ctx, backupDSRange); err != nil {
+					return fmt.Errorf("backup dataset: %w", err)
+				}
+				if err := jd.journalMarkDoneInTx(tx, opID); err != nil {
+					return fmt.Errorf("mark end of backup operation: %w", err)
+				}
+				return nil
+			}); err != nil {
```
lvrach, line 87: One of the usages of the journal is detecting if a job got interrupted, so it could recover a half-completed job later.

With this approach here, since `backupDS` is not using the transaction we effectively add a journal entry once the backup job is completed. I am not sure how this will impact our recovery logic, and if it is useful to use the journal pattern like this.

lvrach: It looks like we are using journal to detect incomplete backups and run the following code:

```go
func (jd *HandleT) removeTableJSONDumps() {
	backupPathDirName := "/rudder-s3-dumps/"
	jd.assertError(err)
	files, err := filepath.Glob(fmt.Sprintf("%v%v_job*", tmpDirPath+backupPathDirName, jd.tablePrefix))
	jd.assertError(err)
	for _, f := range files {
		err = os.Remove(f)
		jd.assertError(err)
	}
}
``` 

The cleanup is trivial and we can probably run it after every start, regardless of a successful backup or not. The os tmp folder can also take care of this cleanup doing a system restart.

I would propose to completely remove journaling for backups, we will get right of a lot of code/complexity.

lvrach: NOTE: We need to think a bit more about `backupDropDSOperation`,  the next section.

atzoum: Currently journaling is mostly used for logging, that's the reason we haven't removed it yet. Journal table can provide useful information for debugging jobsdb issues.

BonapartePC: @atzoum @cisse21 Please let me know your thoughts on this as well

jobsdb/backup.go
BonapartePC: Done

jobsdb/backup.go
lvrach: Still wrapping with transactions doesn't help. 
* For debugging you will never know if something was started and later interrupted. 
* The recovery logic gets disabled with this change, so by keeping the code there we are just adding more confusion into the mix

atzoum: We can remove recovery logic altogether, we can also do this as part of a another PR.

lvrach: Sounds fair.

Does it make sense to use journaling mart start/done inside the same transaction?

Note that during recovery, entries can be deleted:
```go
	if undoOp {
		sqlStatement = fmt.Sprintf(`DELETE from "%s_journal" WHERE id=$1`, jd.tablePrefix)
```

Do you have any concerns about using logging instead of a journal? Is it a reliability concern?

atzoum (approved):

cisse21 (approved):

PR #3583, opened 2023-07-05
fix: user transformations timeout

achettyiitr:
# Description

- Add support for handling panics in case of user transformations for timeouts.
- Small refactoring around removing globals.

## Notion Ticket

https://www.notion.so/rudderstacks/Hosted-abort-jobs-timed-out-during-UT-c21526789be046b3bec857bd239b8ae3?pvs=4

## Security

### DIFF
--- a/processor/integrations/integrations.go
+++ b/processor/integrations/integrations.go
@@ -15,7 +15,7 @@ import (
 )
 
 var (
-	jsonFast              = jsoniter.ConfigCompatibleWithStandardLibrary
+	json                  = jsoniter.ConfigCompatibleWithStandardLibrary
 	postParametersTFields []string
 )
 
@@ -53,7 +53,7 @@ type TransResponseT struct {
 
 func CollectDestErrorStats(input []byte) {
 	var integrationStat TransStatsT
-	err := jsonFast.Unmarshal(input, &integrationStat)
+	err := json.Unmarshal(input, &integrationStat)
 	if err == nil {
 		if len(integrationStat.StatTags) > 0 {
 			stats.Default.NewTaggedStat("integration.failure_detailed", stats.CountType, integrationStat.StatTags).Increment()
@@ -63,7 +63,7 @@ func CollectDestErrorStats(input []byte) {
 
 func CollectIntgTransformErrorStats(input []byte) {
 	var integrationStats []TransStatsT
-	err := jsonFast.Unmarshal(input, &integrationStats)
+	err := json.Unmarshal(input, &integrationStats)
 	if err == nil {
 		for _, integrationStat := range integrationStats {
 			if len(integrationStat.StatTags) > 0 {
@@ -75,7 +75,7 @@ func CollectIntgTransformErrorStats(input []byte) {
 
 // GetPostInfo parses the transformer response
 func ValidatePostInfo(transformRawParams PostParametersT) error {
-	transformRaw, err := jsonFast.Marshal(transformRawParams)
+	transformRaw, err := json.Marshal(transformRawParams)
 	if err != nil {
 		return err
 	}
--- a/processor/integrations/integrations.go
+++ b/processor/integrations/integrations.go
@@ -7,45 +7,25 @@ import (
 
 	jsoniter "github.com/json-iterator/go"
 	"github.com/tidwall/gjson"
-	"golang.org/x/exp/slices"
 
-	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/stats"
 	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 	"github.com/rudderlabs/rudder-server/utils/types"
-	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
 )
 
 var (
-	jsonfast              = jsoniter.ConfigCompatibleWithStandardLibrary
-	destTransformURL      string
-	userTransformURL      string
+	jsonFast              = jsoniter.ConfigCompatibleWithStandardLibrary
 	postParametersTFields []string
 )
 
-func Init() {
-	loadConfig()
+func init() {
 	// This is called in init and it should be a one time call. Making reflect calls during runtime is not a great idea.
 	// We unmarshal json response from transformer into PostParametersT struct.
 	// Since unmarshal doesn't check if the fields are present in the json or not and instead just initialze to zero value, we have to manually do this check on all fields before unmarshaling
 	// This function gets a list of fields tagged as json from the struct and populates in postParametersTFields
 	postParametersTFields = misc.GetMandatoryJSONFieldNames(PostParametersT{})
 }
-
-func loadConfig() {
-	destTransformURL = config.GetString("DEST_TRANSFORM_URL", "http://localhost:9090")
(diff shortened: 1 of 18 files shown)


processor/integrations/integrations.go
```diff
@@ -160,35 +140,3 @@ func FilterClientIntegrations(clientEvent types.SingularEventT, destNameIDMap ma
...
-
-// GetUserTransformURL returns the port of running user transform
-func GetUserTransformURL() string {
-	return userTransformURL + "/customTransform"
-}
-
-// GetTrackingPlanValidationURL returns the port of running tracking plan validation
-func GetTrackingPlanValidationURL() string {
-	return destTransformURL + "/v0/validate"
-}
```
achettyiitr, line 194: Moved to transformation package.

processor/transformer/transformer.go
```diff
@@ -86,35 +94,48 @@ type HandleT struct {
...
+		maxRetry int
+
+		failOnTimeout bool
+		failOnError   bool
+
+		destTransformationURL string
+		userTransformationURL string
+	}
+
+	warehouseDestinationMap map[string]struct{}
```
achettyiitr, line 122: Preparing `warehouseDestinationMap` from `warehouseutils.WarehouseDestinations` once instead of checking in slice every time for every transformations call.

processor/integrations/integrations.go
achettyiitr, line 194: Moved to the transformations package since the endpoint is prepared there itself.

processor/integrations/integrations.go
```diff
@@ -7,46 +7,26 @@ import (
...
 	"github.com/rudderlabs/rudder-server/utils/misc"
 	"github.com/rudderlabs/rudder-server/utils/types"
-	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
 )
 
 var (
-	jsonfast              = jsoniter.ConfigCompatibleWithStandardLibrary
-	destTransformURL      string
-	userTransformURL      string
+	jsonFast              = jsoniter.ConfigCompatibleWithStandardLibrary
```
atzoum, line 18:
```suggestion
	json              = jsoniter.ConfigCompatibleWithStandardLibrary
```

runner/runner.go
```diff
@@ -358,7 +357,6 @@ func runAllInit() {
 	customdestinationmanager.Init()
 	routertransformer.Init()
 	gateway.Init()
-	integrations.Init()
```
atzoum, line 361: ❤️

processor/transformer/transformer.go
```diff
@@ -86,35 +94,48 @@ type HandleT struct {
 	receivedStat stats.Measurement
```
atzoum, line 94: this is a good chance to rename some structs:
- `HandleT` --> `Handle`
- `TransformerEvent` --> `TransformerEvent`
- `MetadataT` --> `Metadata`
- `ResponseT` --> `Response`

processor/transformer/transformer.go
atzoum: We might as well do this in an `init` function using a non-exported package-scoped variable

processor/transformer/transformer.go
```diff
@@ -86,35 +94,48 @@ type HandleT struct {
...
+	Setup(*config.Config, logger.Logger, stats.Stats)
+	Transform(ctx context.Context, clientEvents []TransformerEventT, batchSize int) ResponseT
+	UserTransform(ctx context.Context, clientEvents []TransformerEventT, batchSize int) ResponseT
+	Validate(ctx context.Context, clientEvents []TransformerEventT, batchSize int) ResponseT
 }
 
 // NewTransformer creates a new transformer
 func NewTransformer() *HandleT {
 	return &HandleT{}
 }
```
atzoum, line 136: What is the value of producing an uninitialised transformer that is unusable unless you call Setup first?

I would much prefer to have all transformer.Handle fields unexported and a single producer method:

```go
type Opt func(*handle)

func NewTransformer(conf *config.Config, log logger.Logger, stat stats.Stats, opts... Opts) Transformer
```

I understand that this requires further refactoring though

processor/transformer/transformer.go
```diff
@@ -86,35 +94,48 @@ type HandleT struct {
...
-	disableKeepAlives                                                    bool
-	retrySleep                                                           time.Duration
-	timeoutDuration                                                      time.Duration
-	pkgLogger                                                            logger.Logger
-)
-
 func Init() {
-	loadConfig()
 	pkgLogger = logger.NewLogger().Child("processor").Child("transformer")
 }
```
atzoum, line 140: let's try to get rid of the `Init` function and the `pkgLogger`.

processor/transformer/transformer.go
```diff
@@ -125,50 +146,59 @@ func isJobTerminated(status int) bool {
...
+	trans.config.maxHTTPIdleConnections = conf.GetInt("Processor.maxHTTPIdleConnections", 5)
+	trans.config.disableKeepAlives = conf.GetBool("Transformer.Client.disableKeepAlives", true)
+	trans.config.timeoutDuration = conf.GetDuration("HttpClient.procTransformer.timeout", 30, time.Second)
+
+	trans.config.destTransformationURL = conf.GetString("DEST_TRANSFORM_URL", "http://localhost:9090")
+	trans.config.userTransformationURL = conf.GetString("USER_TRANSFORM_URL", trans.config.destTransformationURL)
 
-	trans.guardConcurrency = make(chan struct{}, maxConcurrency)
+	conf.RegisterIntConfigVariable(30, &trans.config.maxRetry, true, 1, "Processor.maxRetry")
+	conf.RegisterBoolConfigVariable(false, &trans.config.failOnTimeout, true, "Processor.Transformer.failOnTimeout")
```
atzoum, line 184: Let's make it more explicit and set this configuration override on hosted rudderstack object already
```suggestion
	conf.RegisterBoolConfigVariable(false, &trans.config.failOnUserTransformTimeout, true, "Processor.Transformer.failOnUserTransformTimeout")
```

atzoum (commented):
LGTM, I have provided some comments which suggest to perform some further refactoring on this package

processor/transformer/transformer.go
achettyiitr: I thought of doing it but it was touching too many files so I skipped.

atzoum: The refactoring tools available in IDEs are powerful enough and can help performing such operations in a safe manner, no matter the amount of files being modified.

Sidddddarth (approved):

fracasula (approved):

processor/transformer/transformer.go
```diff
@@ -36,11 +40,21 @@ const (
 const (
 	StatusCPDown              = 809
 	TransformerRequestFailure = 909
+	TransformerRequestTimeout = 919
+)
+
+var (
+	json                    = jsoniter.ConfigCompatibleWithStandardLibrary
+	warehouseDestinationMap map[string]struct{}
```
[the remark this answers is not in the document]
achettyiitr: Good point. Let me make the changes.

processor/processor_test.go
```diff
@@ -3177,239 +3151,6 @@ var _ = Describe("TestJobSplitter", func() {
 	})
 })
 
-var _ = Describe("TestSubJobMerger", func() {
```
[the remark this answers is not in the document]
achettyiitr: `SubJobMerger` function is being used only for tests and not anywhere else.

PR #3596, opened 2023-07-07
fix: cleanup for warehouse integration tests

achettyiitr:
# Description

Since validation tests are also running which creates the schema but not drops it. Added cleanup during validations.

## Notion Ticket

https://www.notion.so/rudderstacks/Cleanup-integration-tests-fa75cc8c7d9e4ce887988a107d1f684f?pvs=4

## Security

### DIFF
--- a/warehouse/integrations/bigquery/bigquery_test.go
+++ b/warehouse/integrations/bigquery/bigquery_test.go
@@ -126,14 +126,13 @@ func TestIntegration(t *testing.T) {
 	serviceHealthEndpoint := fmt.Sprintf("http://localhost:%d/health", httpPort)
 	health.WaitUntilReady(ctx, t, serviceHealthEndpoint, time.Minute, time.Second, "serviceHealthEndpoint")
 
-	t.Run("Event flow", func(t *testing.T) {
-		ctx := context.Background()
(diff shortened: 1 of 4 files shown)


warehouse/integrations/bigquery/bigquery_test.go
```diff
@@ -353,6 +352,12 @@ func TestIntegration(t *testing.T) {
 	})
 
 	t.Run("Validations", func(t *testing.T) {
+		t.Cleanup(func() {
+			require.NoError(t, testhelper.WithConstantRetries(func() error {
```
fracasula, line 356: I don't like much the constant retries helper, I suggest we move to `require.Eventually`. The reason I'm saying this is because with `Eventually` we can set both a total timeout and a retry interval. By doing this it is not just more readable and intuitive, it usuallly ends up retrying faster leading to a test finishing faster.

fracasula (approved):

warehouse/integrations/bigquery/bigquery_test.go
achettyiitr: Added the changes.

cisse21 (approved):

PR #3597, opened 2023-07-07
chore: add max connections to warehouse and pgnotifier

achettyiitr:
# Description

- Setting maxOpenConnections for Warehouse and PgNotifier db handle.

## Notion Ticket

https://www.notion.so/rudderstacks/Max-open-connection-for-warehouse-and-postgres-91e4987ff26b41f6a7aa337fa3cb8f80?pvs=4

## Security

### DIFF
--- a/services/pgnotifier/pgnotifier.go
+++ b/services/pgnotifier/pgnotifier.go
@@ -128,6 +128,7 @@ func New(workspaceIdentifier, fallbackConnectionInfo string) (notifier PGNotifie
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


fracasula (approved):
Looks good, if needed moving forward we can differentiate between master and slave but for now I don't see a reason why. Let me know if you disagree of course.

achettyiitr:
> Looks good, if needed moving forward we can differentiate between master and slave but for now I don't see a reason why. Let me know if you disagree of course.

I am not sure how much benefit it can provide. If we need separate configs for master and slave we can always set them as ENVs since both of these have separate deployments.

lvrach (approved):

PR #3602, opened 2023-07-07
chore: warehouse backend config refactoring

fracasula:
# Description

Small refactoring for the backend config in the warehouse.

**Changes:**
* removed duplicated code
* decreased number of global variables
* simplified backend config management
* increased test coverage
* better memory management
* rewrite of the warehouse GRPC tests
* removed tenant manager from slaves
   * this is initialized only once in the warehouse `Setup()` if `isMaster()` 

## TODOs
* ~write tests for new backend config~
* run warehouse tests with `-race`
* ~follow up on new `TODO` comments added in the code~

## Notion Ticket

< [Notion Link](https://www.notion.so/rudderstacks/wh-backend-config-d65c58fbc0054100b127b37706d04684) >

## Security


warehouse/warehouse.go
```diff
@@ -1678,13 +1443,18 @@ func Start(ctx context.Context, app app.App) error {
...
+	rruntime.GoForWarehouse(func() {
+		bcManager.Start(ctx)
+	})
+
 	runningMode := config.GetString("Warehouse.runningMode", "")
 	if runningMode == DegradedMode {
 		pkgLogger.Infof("WH: Running warehouse service in degraded mode...")
 		if isMaster() {
-			rruntime.GoForWarehouse(func() {
-				minimalConfigSubscriber(ctx)
```
achettyiitr, line 1686: We will still be needing a config subscriber in case of degraded mode.

warehouse/backend_config.go
```diff
@@ -0,0 +1,300 @@
...
+
+	connectionsMap   map[string]map[string]model.Warehouse // destID -> sourceID -> warehouse map
+	connectionsMapMu sync.RWMutex
+
+	sourceIDsByWorkspace   map[string][]string // workspaceID -> []sourceIDs
+	sourceIDsByWorkspaceMu sync.RWMutex
+}
+
+func (s *backendConfigManager) Start(ctx context.Context) {
+	ch := s.backendConfig.Subscribe(ctx, backendconfig.TopicBackendConfig)
```
achettyiitr, line 87: We should we using wh.tenantManager.WatchConfig(ctx)

warehouse/warehouse.go
```diff
@@ -1063,9 +831,7 @@ func onConfigDataEvent(ctx context.Context, config map[string]backendconfig.Conf
 	pkgLogger.Debug("Got config from config-backend", config)
 
 	enabledDestinations := make(map[string]bool)
-	var connectionFlags backendconfig.ConnectionFlags
 	for _, wConfig := range config {
-		connectionFlags = wConfig.ConnectionFlags // the last connection flags should be enough, since they are all the same in multi-workspace environments
```
achettyiitr, line 1068: Since we are already doing this in bcManager, no need do it here.

warehouse/warehouse.go
```diff
@@ -1074,36 +840,25 @@ func onConfigDataEvent(ctx context.Context, config map[string]backendconfig.Conf
...
 					}
 				}
 			}
 		}
 	}
-	if val, ok := connectionFlags.Services["warehouse"]; ok {
-		if UploadAPI.connectionManager != nil {
-			UploadAPI.connectionManager.Apply(connectionFlags.URL, val)
-		}
-	}
```
achettyiitr, line 1097: Since we are already doing this in bcManager, no need do it here.

warehouse/warehouse_test.go
```diff
@@ -193,173 +190,3 @@ func TestUploadJob_ProcessingStats(t *testing.T) {
 		})
 	}
 }
-
-func Test_GetNamespace(t *testing.T) {
```
achettyiitr, line 197: This is moved to bcManager.

achettyiitr (approved):

cisse21 (approved):

warehouse/backend_config.go
```diff
@@ -20,32 +22,32 @@ import (
...
-		logger:               l,
+		logger:               log,
 		initialConfigFetched: make(chan struct{}),
 		connectionsMap:       make(map[string]map[string]model.Warehouse),
 	}
-	if config.GetBool("ENABLE_TUNNELLING", true) {
+	if c.GetBool("ENABLE_TUNNELLING", true) {
 		bcm.internalControlPlaneClient = cpclient.NewInternalClientWithCache(
-			backendconfig.GetConfigBackendURL(),
+			c.GetString("CONFIG_BACKEND_URL", "https://api.rudderstack.com"),
```
fracasula, line 49: Why do we keep duplicating the variable name and its default? Isn't `GetConfigBackendURL` good enough?

warehouse/backend_config.go
```diff
@@ -86,7 +89,10 @@ func (s *backendConfigManager) Start(ctx context.Context) {
 		select {
 		case <-ctx.Done():
 			return
-		case data := <-ch:
+		case data, ok := <-ch:
+			if !ok {
+				return
```
fracasula, line 93: @achettyiitr nice catch :+1:

warehouse/backend_config.go
achettyiitr: I wanted to avoid calling global function which for backendconfig was difficult to test.

warehouse/backend_config.go
```diff
@@ -222,6 +215,49 @@ func (s *backendConfigManager) processData(ctx context.Context, data map[string]
...
+	}
+
+	namespacePrefix := s.conf.GetString(fmt.Sprintf("Warehouse.%s.customDatasetPrefix", whutils.WHDestNameMap[destType]), "")
+	if namespacePrefix != "" {
+		return whutils.ToProviderCase(destType, whutils.ToSafeNamespace(destType, fmt.Sprintf(`%s_%s`, namespacePrefix, source.Name)))
+	}
+
+	namespace, err := s.schema.GetNamespace(ctx, source.ID, destination.ID)
+	if err != nil {
+		s.logger.Errorw("getting namespace",
```
fracasula, line 247: @achettyiitr we don't have enough context here, right? We should at least log the error.

achettyiitr: Good point. Let me make the changes in a separate PR.

warehouse/backend_config_test.go
```diff
@@ -244,4 +256,219 @@ func TestBackendConfigManager(t *testing.T) {
...
+			},
+			expectedNamespace: "config_result_test_source",
+			setConfig:         true,
+		},
+	}
+
+	for _, tc := range testcases {
+		tc := tc
+
+		t.Run("should return namespace", func(t *testing.T) {
```
fracasula, line 434: @achettyiitr How come the test name is always the same? Shall we include a name for the test in the test case itself? Something that would describe the type of test in order to make it self-explanatory would be nice.

warehouse/warehouse.go
```diff
@@ -1447,12 +1405,23 @@ func Start(ctx context.Context, app app.App) error {
 		}
 	}()
 
+	g, ctx := errgroup.WithContext(ctx)
```
fracasula, line 1408: @achettyiitr it feels like we don't need a group here since none of the two routines are returning an error. also, we're overriding the context with the group one which could cause confusion in the code after the group that might be using the context. wdyt?

warehouse/backend_config.go
```diff
@@ -106,6 +105,12 @@ func (s *backendConfigManager) Subscribe(ctx context.Context) <-chan []model.War
 	ch := make(chan []model.Warehouse, 10)
 	s.subscriptions = append(s.subscriptions, ch)
 
+	s.warehousesMu.Lock()
+	if len(s.warehouses) > 0 {
+		ch <- s.warehouses
+	}
+	s.warehousesMu.Unlock()
```
fracasula: @achettyiitr this version was better, now we're holding the lock for the duration of the whole subscribe. I think we should put this code back.

warehouse/backend_config.go
fracasula: Right, because you replaced the bc with the tenant manager. Let's keep it this way for now then :+1:

warehouse/backend_config.go
achettyiitr: The reason, we are adding this is because there is a race condition happening if we are not subscribers and we get backed config.
So, during subscription also, we are sending the last copy of the data.

warehouse/warehouse.go
```diff
@@ -277,171 +252,49 @@ func (*HandleT) handleUploadJob(uploadJob *UploadJob) error {
...
-					// check this commit to https://github.com/rudderlabs/rudder-server/pull/476/commits/fbfddf167aa9fc63485fe006d34e6881f5019667
-					// to avoid creating goroutine for disabled sources/destinations
-					if _, ok := wh.workerChannelMap[workerName]; !ok {
-						workerChan := wh.initWorker()
-						wh.workerChannelMap[workerName] = workerChan
-					}
-					wh.workerChannelMapLock.Unlock()
+		warehouses = lo.Filter(warehouses, func(warehouse model.Warehouse, _ int) bool {
+			return warehouse.Destination.DestinationDefinition.Name == wh.destType
+		})
```
fracasula, line 260: @achettyiitr we could also do this to save ourselves the next `for` loop:

```go
warehouses = lo.Filter(warehouses, func(warehouse model.Warehouse, _ int) bool {
	if warehouse.Destination.DestinationDefinition.Name != wh.destType {
		return false
	}

	if warehouseutils.IDResolutionEnabled() &&
		slices.Contains(warehouseutils.IdentityEnabledWarehouses, wh.destType) {
		wh.setupIdentityTables(ctx, warehouse)
		if shouldPopulateHistoricIdentities && warehouse.Destination.Enabled {
			// non-blocking populate historic identities
			wh.populateHistoricIdentities(ctx, warehouse)
		}
	}

	return true
})
```

PR #3606, opened 2023-07-10
fix: add index on reported_at column of reports table

satishrudderstack:
# Description

- Adding index on `reported_at` on `reports` table which is used by reporting queries

## Notion Ticket

- https://www.notion.so/rudderstacks/Reports-Flush-Issue-from-Rudder-Server-Instance-9-on-prousmtusmt-c9592bd2b2b74d05a9e083ad8baa7988

## Security


sql/migrations/reports/000005_alter_reports_add_reported_at_index.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX CONCURRENTLY reports_reported_at_index ON reports (reported_at);
```
lvrach, line 1: Concurrently since introduced in Postgres 15 will not be supported in all rudder server setups.

We can run it manually with concurrently in Postgres that support it.

We should also add 'IF NOT EXISTS'  so we can run it manually first with no issues

satishrudderstack: Made this change @lvrach . Marking this conversation as resolved

lvrach (approved):

cisse21 (approved):

PR #3610, opened 2023-07-10
chore: add logs for debugging unmarshalling issues

koladilip:
# Description

This is for debugging: 
1. https://app.squadcast.com/incident/64aa22a772ca03089f7b6ebc
2. Error: 2023-07-09T07:22:41.681Z	ERROR	router.transformer	transformer/transformer.go:291	 [TransformerProxy Unmarshaling]::readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...

## Notion Ticket

< Replace with Notion Link >

## Security

### DIFF
--- a/router/worker.go
+++ b/router/worker.go
@@ -426,7 +426,7 @@ func (w *worker) processDestinationJobs() {
 					if err != nil {
 						errorAt = routerutils.ERROR_AT_TF
 						respStatusCode, respBody = types.RouterUnMarshalErrorCode, fmt.Errorf("transformer response unmarshal error: %w", err).Error()
-						w.logger.Errorf("transformer response unmarshal error: %v, destType: %s, destID: %s, jobID: %d", err, w.rt.destType, destinationID, destinationJob.JobMetadataArray[0].JobID)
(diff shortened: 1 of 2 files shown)


router/worker.go
```diff
@@ -426,6 +426,7 @@ func (w *worker) processDestinationJobs() {
 					if err != nil {
 						errorAt = routerutils.ERROR_AT_TF
 						respStatusCode, respBody = types.RouterUnMarshalErrorCode, fmt.Errorf("transformer response unmarshal error: %w", err).Error()
+						w.logger.Errorf("transformer response unmarshal error: %v, destType: %s, destID: %s, jobID: %d", err, w.rt.destType, destinationID, destinationJob.JobMetadataArray[0].JobID)
```
fracasula, line 429: Does `JobMetadataArray[0]` always exist when reaching this line?

fracasula (approved):

router/worker.go
sanpj2292: Some extreme cases, where in the metadata is being sent as an object(mistkenly) then `JobMetadataArray[0]` wouldn't exist. Hence removing it

sanpj2292 (approved):
IMHO, we need to have this logging hence approving this PR

PR #3613, opened 2023-07-10
chore: handle-non-200-codes-transformerproxy

chandumlg:
# Description

Handle non 200 codes transformerproxy to prevent unmarshal errors.

## Notion Ticket

https://www.notion.so/rudderstacks/d5d15ce4be354e3caae11d4ea9f41477?v=2ad23dce1aeb431a82f783e6607d3cb0&p=9483c109858c4fe8856aee9f0c98234a&pm=c

## Security

### DIFF
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -371,16 +371,14 @@ func (trans *handle) doProxyRequest(ctx context.Context, proxyReqParams *ProxyRe
 			statusCode: http.StatusInternalServerError,
 			err:        err,
 		}
-	} else if resp.StatusCode == http.StatusNotFound {
-		// Actually Router wouldn't send any destination to proxy unless it already exists
-		// But if accidentally such a request is sent, we'd probably need to handle for better
-		// understanding of the response
-		notFoundErr := fmt.Errorf(`post "%s" not found`, req.URL)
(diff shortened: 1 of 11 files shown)


router/transformer/transformer.go
```diff
@@ -371,16 +371,14 @@ func (trans *handle) doProxyRequest(ctx context.Context, proxyReqParams *ProxyRe
 			statusCode: http.StatusInternalServerError,
 			err:        err,
 		}
-	} else if resp.StatusCode == http.StatusNotFound {
-		// Actually Router wouldn't send any destination to proxy unless it already exists
-		// But if accidentally such a request is sent, we'd probably need to handle for better
-		// understanding of the response
-		notFoundErr := fmt.Errorf(`post "%s" not found`, req.URL)
-		trans.logger.Errorf(`[TransformerProxy] (Dest-%[1]v) {Job - %[2]v} Client.Do Failure for %[1]v, with %[3]v`, destName, jobID, notFoundErr)
+	} else if resp.StatusCode != http.StatusOK {
```
sanpj2292, line 374: Should we handle this condition as when resp.StatusCode is not 2xx, then you throw an error ? Or should we handle for status-codes that are > 500, throw a non-success error ?

sanpj2292: 🤔

cisse21 (approved):

PR #3644, opened 2023-07-19
chore: fix stat table prefix

cisse21:
# Description

This PR fixes the issue with tablePrefix causing the stats to being populated incorrectly
Also now we stop in the exactly reverse order as we start to ensure that we don't end up corrupting stats and causing alerts

## Notion Ticket

[ Notion Link ](https://www.notion.so/rudderstacks/negative-pending-event-count-bug-1d57bd99cb874b5b96e6a691820f4a1e?pvs=4)

## Security


cisse21:
Let's replace the prefixes with constants in a separate PR

Sidddddarth (approved):

lvrach (approved):

PR #3685, opened 2023-07-31
fix: ignore workspace if no backup config is found

Sidddddarth:
# Description

Ignore error resulting from a workspace's backup config not being found.
This is done so the backup is not stalled in such case.

## Notion Ticket

[linear cfd link](https://linear.app/rudderstack/issue/RUD-412/subteams03sekbfx0dserver-on-call-lot-of-pre-drop-gw-tables-on)

## Security

### DIFF
--- a/jobsdb/backup.go
+++ b/jobsdb/backup.go
@@ -4,19 +4,21 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
-	"github.com/cenkalti/backoff"
 	"golang.org/x/sync/errgroup"
+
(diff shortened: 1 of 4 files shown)


jobsdb/backup.go
```diff
@@ -553,17 +553,16 @@ func (jd *HandleT) createTableDumps(queryFunc func(int64) string, pathFunc func(
 			}
 			preferences, err := jd.fileUploaderProvider.GetStoragePreferences(workspaceID)
```
atzoum, line 554: I had to go through the actual implementation for understanding that the current implementation only throughs an error when there is no workspace present in settings. Using a sentinel error might be preferable here.

atzoum (approved):

cisse21 (approved):

PR #3688, opened 2023-07-31
chore: gateway backend config initialisation improvements

atzoum:
# Description

- Respect the context in `gateway.backendConfigSubscriber`
- Wait until the first configuration has been consumed by the gateway before starting gateway's http listeners
- Add some more information in the **Workspace Config changed** log, such as the number of workspaces and total sources contained in the updated configuration.

## Linear Ticket

[Link](https://linear.app/rudderstack/issue/PIPE-81/gateway-backend-config-initialisation-improvements)

## Security

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -46,7 +46,6 @@ import (
 	"github.com/rudderlabs/rudder-server/gateway/webhook"
 	"github.com/rudderlabs/rudder-server/jobsdb"
 	"github.com/rudderlabs/rudder-server/middleware"
-	"github.com/rudderlabs/rudder-server/rruntime"
 	sourcedebugger "github.com/rudderlabs/rudder-server/services/debugger/source"
 	"github.com/rudderlabs/rudder-server/services/diagnostics"
 	"github.com/rudderlabs/rudder-server/services/rsources"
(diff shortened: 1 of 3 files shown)


lvrach (approved):

gateway/gateway.go
```diff
@@ -201,6 +200,9 @@ type HandleT struct {
 	rsourcesService       rsources.JobService
 	sourcehandle          sourcedebugger.SourceDebugger
 	whProxy               http.Handler
+
+	configInitialised     bool
+	configInitialisedChan chan struct{}
```
achettyiitr, line 205: Is it possible to use `sync.Once` here instead?

gateway/gateway.go
```diff
@@ -1307,7 +1309,7 @@ This function will block.
 */
 func (gateway *HandleT) StartWebHandler(ctx context.Context) error {
 	gateway.logger.Infof("WebHandler waiting for BackendConfig before starting on %d", webPort)
-	gateway.backendConfig.WaitForConfig(ctx)
+	<-gateway.configInitialisedChan
```
achettyiitr, line 1312: Should we honor context as well here?
```GO
	switch {
	case <-gateway.configInitialisedChan
		return
	case <-ctx:
		return
	}
```

Sidddddarth (approved):

PR #3721, opened 2023-08-08
feat: archive jobs to object storage

Sidddddarth:
# Description

Processor writes the events to a new jobsdb - archivalJobsDB
Archiver reads jobs from archive jobsdb and uploads to user configured object storage(if eligible)
Enabled for gateway jobs(archived at a source level, and further in an hourly folder in the source prefix)

```
BenchmarkArchiverIsolation/numWorkspaces:_25_-_numSourcePerWorkspace:_4_-_numJobsPerSource:_1000_-_totalJobs:_100000-10         	       1	33437059167 ns/op	        33.44 overall_duration_seconds	1571687416 B/op	26652032 allocs/op

BenchmarkArchiverIsolation/numWorkspaces:_100_-_numSourcePerWorkspace:_5_-_numJobsPerSource:_1000_-_totalJobs:_500000-10        	       1	134082533166 ns/op	       134.1 overall_duration_seconds	8210319192 B/op	139518565 allocs/op
```

## Linear Ticket

[Notion Link](https://www.notion.so/rudderstacks/Gateway-archival-a9c85931f711407a9cc0415f2c75ed83?pvs=4)
[feed the archival](https://www.notion.so/rudderstacks/Feed-the-Archival-790caefa33424381953a6fe14679a4f7?pvs=4)

## Security


archiver/archiver_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/rudderlabs/rudder-server/testhelper/destination"
+)
+
+func TestJobsArchival(t *testing.T) {
+	var (
+		prefix        = "some-prefix"
+		minioResource []*destination.MINIOResource
+
+		// test data - contains jobs from 3 workspaces(1 - 1 source, 2 & 3 - 2 sources each)
+		goldenFileJobsFileName = "testdata/MultiWorkspaceBackupJobs.json.gz"
```
lvrach, line 41: [minor] Given its usage I would avoid using `goldenFile`:
* We run comparing the result of the operation with this file or the contents of this file, rather we use it as a seed for our jobs
* Change in archival functionality should require the file to change as well, which is not the case here

 ```suggestion
		seedJobsFileName = "testdata/MultiWorkspaceBackupJobs.json.gz"
```

File: archiver/worker.go [hunk 2]
```diff
@@ -0,0 +1,264 @@
...
+	lastJobCreatedAt := jobs[len(jobs)-1].CreatedAt.UTC()
+	workspaceID := jobs[0].WorkspaceId
+	log := w.log.With("workspaceID", workspaceID)
+
+	gzWriter := fileuploader.NewGzMultiFileWriter()
+	path := fmt.Sprintf(
+		"%v%v.json.gz",
+		lo.Must(misc.CreateTMPDIR())+"/rudder-backups/"+w.sourceID+"/",
+		fmt.Sprintf("%v_%v_%v", firstJobCreatedAt.Unix(), lastJobCreatedAt.Unix(), workspaceID),
+	)
```
lvrach, line 152: * `path.Join`: makes it easier to read in safer across operating systems
* I prefer using the expected types rather than `%v` when the type is known.
```suggestion
	path :=  path.Join(
	    lo.Must(misc.CreateTMPDIR()), 
	    rudder-backups, 
	    w.sourceID, 
	    fmt.Sprintf("%d_%d_%s", firstJobCreatedAt.Unix(), lastJobCreatedAt.Unix(), workspaceID),
	)
```

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+
+	for _, job := range jobs {
+		j, err := marshalJob(job)
+		if err != nil {
+			return "", fmt.Errorf("failed to marshal job: %w", err)
+		}
+		if _, err := gzWriter.Write(path, append(j, '\n')); err != nil {
+			return "", fmt.Errorf("failed to write to file: %w", err)
```
lvrach, line 160: [minor] no need to mention `failed to` on an error message
```suggestion
			return "", fmt.Errorf("write to file: %w", err)
```

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+
+	gzWriter := fileuploader.NewGzMultiFileWriter()
+	path := fmt.Sprintf(
+		"%v%v.json.gz",
+		lo.Must(misc.CreateTMPDIR())+"/rudder-backups/"+w.sourceID+"/",
+		fmt.Sprintf("%v_%v_%v", firstJobCreatedAt.Unix(), lastJobCreatedAt.Unix(), workspaceID),
+	)
+
+	for _, job := range jobs {
+		j, err := marshalJob(job)
```
lvrach, line 155: You might want to use json.Encoder instead and write directly to the `gzWriter` https://pkg.go.dev/encoding/json#NewEncoder.

Less lines of code and possibly less memory allocations

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+	}
+	if err := gzWriter.Close(); err != nil {
+		return "", fmt.Errorf("failed to close file: %w", err)
+	}
+	defer func() { _ = os.Remove(path) }()
+
+	fileUploader, err := w.storageProvider.GetFileManager(workspaceID)
+	if err != nil {
+		log.Errorw("Skipping storing errors since no file manager is found", "error", err)
```
lvrach, line 170: Under which conditions this error is possible? will we get error logs if customer has not configured backups ?

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+		return "", fmt.Errorf("failed to get file manager: %w", err)
+	}
+
+	file, err := os.Open(path)
+	if err != nil {
+		return "", fmt.Errorf("failed to open file %s: %w", path, err)
```
lvrach, line 176: [minor]
```suggestion
		return "", fmt.Errorf("open file %q: %w", path, err)
```

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	defer func() { _ = file.Close() }()
```
lvrach, line 178: [blocking] It is not a good idea to ignore the error from `file.Close()`, during close we might need to flush data to disk, and if it fails data will be lost. Given the criticality of those backups, we need to be extra careful

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+		w.sourceID,
+		w.archiveFrom,
+		firstJobCreatedAt.Format("2006-01-02"),
+		fmt.Sprintf("%d", firstJobCreatedAt.Hour()),
+		w.config.instanceID,
+	}
+	uploadOutput, err := fileUploader.Upload(ctx, file, prefixes...)
+	if err != nil {
+		log.Errorw("failed to upload file to object storage", "error", err)
+		return "", fmt.Errorf("failed to upload file to object storage - %w", err)
```
lvrach, line 189: both logging and returning error

also:
```suggestion
		log.Errorw("failed to upload file to object storage", "error", err)
		return "", fmt.Errorf("upload file to object storage: %w", err)
```

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+				nil,
+			)
+		},
+		func(attempt int) {
+			w.log.Warnw("failed to mark jobs' status", "attempt", attempt)
+		},
+	)
+}
+
+func errJson(err error) []byte {
```
lvrach, line 254: [nit-pick] idiomatic way in go
```suggestion
func errJSON(err error) []byte {
```

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+	m := map[string]string{"error": err.Error()}
```
lvrach, line 255: [minor] prefer usage of struct when possible for performance and type checking
```suggestion
        m := struct {
            Error string `json:"error"`
         }{
            Error: err.Error()
         }
```

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+	b, _ := json.Marshal(m)
+	return b
+}
+
+func locationJson(location string) []byte {
```
lvrach, line 260: [nit-pick]
```suggestion
func locationJSON(location string) []byte {
```

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+	if len(jobs) == 0 {
+		return false
+	}
+
+	if !limitReached && time.Since(w.lastUploadTime) < w.config.uploadFrequency {
+		return false // respect the upload frequency
+	}
+
+	workspaceID := jobs[0].WorkspaceId
+	log = log.With("workspaceID", workspaceID)
```
lvrach, line 66: 👍

lvrach (changes requested):
Only blocking request is checking file.Close() error
https://www.joeshaw.org/dont-defer-close-on-writable-files/

app/cluster/dynamic.go
```diff
@@ -191,10 +199,14 @@ func (d *Dynamic) stop() {
 	}
 	d.logger.Info("Stopping the server")
 	start := time.Now()
+	d.Processor.Stop()
+	d.logger.Debug("Processor stopped")
```
atzoum, line 203: Looks like we are stopping processor twice

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+	lastUploadTime time.Time
+	queryParams    jobsdb.GetQueryParamsT
+}
+
+func (w *worker) Work() bool {
+	log := w.log.With("sourceID", w.sourceID)
+start:
+
+	jobs, limitReached, err := w.getJobs()
+	if err != nil {
```
atzoum, line 52: Shouldn't we panic instead if `w.lifecycle.ctx.Err() == nil`?

archiver/worker.go
atzoum: We are only opening the file for reading and sending its contents to the object storage. 
If this operation fails, the jobs will not be marked as succeeded and archiver will retry, i.e. a new file will be created containing the same jobs. What should we do in case closing the file that we opened for reading fails here?

app/apphandlers/embeddedAppHandler.go
```diff
@@ -200,6 +202,20 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
...
 		jobsdb.WithSkipMaintenanceErr(config.GetBool("Processor.jobsDB.skipMaintenanceError", false)),
 	)
+	defer schemaDB.Close()
+
+	archivalDB := jobsdb.NewForReadWrite(
+		"arc",
+		jobsdb.WithClearDB(options.ClearDB),
+		jobsdb.WithDSLimit(&a.config.processorDSLimit),
+		jobsdb.WithSkipMaintenanceErr(config.GetBool("Processor.jobsDB.skipMaintenanceError", false)),
+		jobsdb.WithJobMaxAge(
```
atzoum, line 212: **Note for future improvement:** `jobsdb.HandleT#doCleanup` method picks up jobs in **executing** state as well.

This means that both the following are theoretically possible:
- Duplicate terminal states are recorded for a job, one by the cleanup goroutine and another by the goroutine which marked jobs as executing.
- The cleanup goroutine marks a job as aborted and the other goroutine marks it as failed.

Thus we might need to reconsider in the future having the following in mind:
1. Dealing with jobs in executing state
2. Keeping pending events consistent

archiver/worker.go
Sidddddarth: >Under which conditions this error is possible?

This particular scenario can only happen when the `workspace` is not part of the config this server instance serves

Sidddddarth: If it reached this point, these jobs will be marked aborted in the next iteration.

atzoum: can we use a sentinel error to make this error path more straightforward?

archiver/worker.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	if err := gzWriter.Close(); err != nil {
+		return "", fmt.Errorf("failed to close file: %w", err)
+	}
+	defer func() { _ = os.Remove(path) }()
+
+	fileUploader, err := w.storageProvider.GetFileManager(workspaceID)
+	if err != nil {
+		log.Errorw("Skipping storing errors since no file manager is found", "error", err)
+		return "", fmt.Errorf("failed to get file manager: %w", err)
```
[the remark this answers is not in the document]
Sidddddarth: true
removing the logs here. Returned error is logged anyway.

archiver/worker.go
lvrach: My bad, due to various comments the order of events was unclear. No a blocking comment, but if `masrhalJob` we don't close the `gzWriter`

archiver/worker.go
```diff
@@ -0,0 +1,268 @@
...
+		"rudder-backups",
+		w.sourceID,
+		fmt.Sprintf("%d_%d_%s.json.gz", firstJobCreatedAt.Unix(), lastJobCreatedAt.Unix(), workspaceID),
+	)
+
+	for _, job := range jobs {
+		j, err := marshalJob(job)
+		if err != nil {
+			return "", fmt.Errorf("marshal job: %w", err)
+		}
```
lvrach, line 160:
```suggestion
		if err != nil {
		        _ = gzWriter.Close()
			return "", fmt.Errorf("marshal job: %w", err)
		}
```

archiver/worker.go
```diff
@@ -0,0 +1,268 @@
...
+		if _, err := gzWriter.Write(filePath, append(j, '\n')); err != nil {
+			return "", fmt.Errorf("write to file: %w", err)
+		}
```
lvrach, line 163:
```suggestion
		if _, err := gzWriter.Write(filePath, append(j, '\n')); err != nil {
		         _ = gzWriter.Close()
			return "", fmt.Errorf("write to file: %w", err)
		}
```

lvrach (approved):

archiver/worker.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func (w *worker) Work() bool {
+start:
+	jobs, limitReached, err := w.getJobs()
+	if err != nil {
+		if w.lifecycle.ctx.Err() == nil {
+			panic(err)
+		}
+		w.log.Errorw("failed to fetch jobs for archiving", "error", err)
```
atzoum, line 55: Cancelling the parent context should not be considered to be an error

atzoum (approved):

PR #3726, opened 2023-08-09
chore: archiver cleanup

fracasula:
# Description

Prepared statements for archiver + more DRY and cleanup.

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-121/warehouse-archiver-cleanup) >

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #3731, opened 2023-08-10
chore: logfield with structured logging

fracasula:
# Description

Experimenting with structured logging in the warehouse.

Rationale:
* by allowing only `KeyValue` parameters with unexported attributes, we enforce the usage of the desired pattern
* we eliminate the risk of passing an odd number of key/values
* the type `Generic` can be removed altogether, forcing to add a new `Field` whenever required. alternatively it can be kept and then used later on to search for `Generic` usage to see if new fields can be created (e.g. used more than once)
* by using functions instead of constants we are able to do some conversions (see `logfield.Error`, now we can pass both a `string` or an `error` without having to do `err.Error()` each time)
* by having a `logfield.Logger` we can enforce the pattern and make sure a `component` is always specified

This relates to a discussion we had in [this other PR](https://github.com/rudderlabs/rudder-server/pull/3726#issuecomment-1672724607).

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-129/warehouse-structured-logging) >

## Security

### DIFF
--- a/warehouse/logfield/logfield.go
+++ b/warehouse/logfield/logfield.go
@@ -1,6 +1,8 @@
 package logfield
 
 import (
+	"sync"
+
 	"github.com/rudderlabs/rudder-go-kit/logger"
 )
 
@@ -59,32 +61,46 @@ var (
 	}
 )
 
-func NewLogger(component string, l logger.Logger) *Logger {
-	return &Logger{component: component, logger: l}
+func NewLogger(component string, l logger.Logger) Logger {
+	return Logger{component: component, logger: l}
 }
 
 type Logger struct {
 	component string
 	logger    logger.Logger
+	once      sync.Once
+}
+
+func (l *Logger) init() {
+	l.once.Do(func() {
+		if l.logger == nil {
+			l.logger = logger.NOP
+		}
+	})
 }
 
 func (l *Logger) Infow(msg string, kvs ...KeyValue) {
+	l.init()
 	l.logger.Infow(msg, l.keyAndValues(kvs)...)
 }
 
 func (l *Logger) Debugw(msg string, kvs ...KeyValue) {
+	l.init()
 	l.logger.Debugw(msg, l.keyAndValues(kvs)...)
 }
 
 func (l *Logger) Errorw(msg string, kvs ...KeyValue) {
+	l.init()
 	l.logger.Errorw(msg, l.keyAndValues(kvs)...)
 }
 
 func (l *Logger) Warnw(msg string, kvs ...KeyValue) {
+	l.init()
 	l.logger.Warnw(msg, l.keyAndValues(kvs)...)
 }
 
 func (l *Logger) Fatalw(msg string, kvs ...KeyValue) {
+	l.init()
 	l.logger.Fatalw(msg, l.keyAndValues(kvs)...)
 }
 
--- a/warehouse/logfield/logfield.go
+++ b/warehouse/logfield/logfield.go
@@ -1,20 +1,20 @@
 package logfield
 
+import (
+	"github.com/rudderlabs/rudder-go-kit/logger"
+)
+
 const (
 	UploadJobID                = "uploadJobID"
 	UploadStatus               = "uploadStatus"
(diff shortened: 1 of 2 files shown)


fracasula:
@lvrach @achettyiitr I had a veeery quick stab and the structured logging issue here. It's just an idea, please have a look. If it's overkilling we can stick with the usual approach.

warehouse/logfield/logfield.go
```diff
@@ -23,7 +23,82 @@ const (
...
+			return KeyValue{k, v}
+		}
+	}
+)
+
+func NewLogger(component string, l logger.Logger) *Logger {
+	return &Logger{component: component, logger: l}
+}
+
+type Logger struct {
```
fracasula, line 66: [optional] Given that the `Logger` is a struct here we could always pass it by value and then use a `sync.Once` to make sure that the internal logger is always either not `nil` or `logger.NOP`. This would help with the tests as well.

achettyiitr: Should we enforce this at the `logger.Logger`?

fracasula: > How are we planning to use this? Are we going to replace logger.Logger with this one everywhere?

That would be the idea yeah. Of course it'll be done incrementally and with time.

> Should we enforce this at the `logger.Logger`?

The `logger.Logger` is generic and it belongs to the `kit`. I was thinking of a logger wrapper that is domain/team specific. Such logger could even live in an `internal` folder and used only in its specific domain. Also, the abstraction layer is very small, it's mostly about providing the fields for consistency so I don't see a problem with having a package like this in each domain.

warehouse/logfield/logfield.go
```diff
@@ -23,7 +25,96 @@ const (
...
+
+func (l *Logger) init() {
+	l.once.Do(func() {
+		if l.logger == nil {
+			l.logger = logger.NOP
+		}
+	})
+}
+
+func (l *Logger) Infow(msg string, kvs ...KeyValue) {
```
koladilip, line 82: KeyValue will be good and provides flexibility and provide consistent way of naming things but they don't any meaning associated with them. Do we need a structure like following?
type Metadata struct {
 DestinationID string
 WorkspaceID string
 SourceID string
 OtherFileds map[string]string
}
something like this may be opinionated but now we know what is workspaceID and sourceID etc so we can integrate them with various tools like loki more easily.

fracasula: Don't we already know that with the approach I'm suggesting? Can you try to explain the pros and cons of your approach and how it would differ from this? I don't think I fully understand your idea.

koladilip: If we need validation of this metadata like DestinationID and WorkspaceID required etc then having something like above structure it is easy to do, also Metadata structure can vary in different situation and teams can extend the base logger with their own validation rules. May be underlying we can have generic key value logger with consistent names and on top we can have metadata structure of teams to easily define validation rules etc. Hence my approach requires work from each team to define their own metadata structure

fracasula: I can try to have a stab at this and push it in another branch to compare.

fracasula: @koladilip is [this](https://github.com/rudderlabs/rudder-server/blob/chore.whLogfield-2/warehouse/logfield/logfield.go) close enough to what you mean?

If yes I don't like these things about it:
* slower computation, the struct can become quite big and every time we log we have to check every field manually versus iterating over a slice
* maintenance, we need to manually add both a field in the struct and the related `if` vs 1 function
* we lose the ability to use the `Generic` function which allows for a more relaxed approach. down the line you simply "Find usages" of the `Generic` function and start to organize potential new fields together

What we can't easily do with the functions approach is validating multiple fields but it sounds like we don't need it (unless you have a specific example in mind) and I personally find that overkilling for a logging middleware.

Validation of single fields can still be done in the function where you define the field itself instead of being decoupled and have the definition in one place (struct attribute) and the validation down below (in the method).

Thoughts?

PR #3745, opened 2023-08-14
feat: snowflake appendmode

fracasula:
# Description

Adding `APPEND` mode for Snowflake.

### Changelog
* DRYing `MERGE INTO` query
* Adding new `loadTableStrategy` feature flag
* Adding `APPEND` support to `LoadTable` with integration tests
* Adding `APPEND` support to `LoadUserTables` with integration tests

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-108/introduce-the-option-to-use-append-mode-for-snowflake) >

## Security


warehouse/integrations/snowflake/snowflake.go
```diff
@@ -166,6 +168,18 @@ func New(conf *config.Config, log logger.Logger, stat stats.Stats) *Snowflake {
...
+	loadTableStrategy := conf.GetString("Warehouse.snowflake.loadTableStrategy", loadTableStrategyMergeMode)
+	switch loadTableStrategy {
+	case loadTableStrategyMergeMode, loadTableStrategyAppendMode:
+		sf.config.loadTableStrategy = loadTableStrategy
+	default:
+		loadTableStrategy = loadTableStrategyMergeMode
+		sf.logger.Errorw(
+			"Received invalid loadTableStrategy, defaulting to MERGE mode",
+			lf.LoadTableStrategy, loadTableStrategy,
+		)
```
achettyiitr, line 180: Should we just return from here instead? Since it's set up only.

fracasula: I thought I'd fail gracefully, just log the error, set the default and carry on. If we re return from there then we don't setup `enableDeleteByJobs` and `slowQueryThreshold`. As an alternative I can make the constructor return an error and propagate it upstream.

fracasula: If we propagate the error upstream to the manager then it's going to end up in `uploadsToProcess` which leads to a `panic` so we could do that as well. Thoughts?

achettyiitr: I think we should we should add an error for the constructor.

achettyiitr: Although, we don't have a proper way to initialize these integrations. We are calling New but it does only partial initialization. Some are calling `Setup` while others rely on `Connect`. Or some are just calling functions like `TestConnection`, and `IsEmpty`.

fracasula: https://github.com/rudderlabs/rudder-server/pull/3745/commits/1a0fc96621e51c2ad15d5f9e90485650dad63ff5

achettyiitr: Do you mean the `warehouse/validations` package?

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -756,56 +739,72 @@ func (sf *Snowflake) loadUserTables(ctx context.Context) map[string]error {
...
+			SELECT DISTINCT *
+			FROM (
+				SELECT "USER_ID" as "ID", %[3]s
+				FROM %[1]s.%[4]q
+				WHERE "USER_ID" IS NOT NULL
+			);`,
+			schemaIdentifier, usersTable,
+			strings.Join(userColNames, ","),
+			identifiesTable,
+		)
```
fracasula, line 763: Let me know what you think about this approach and if you have some other ideas in mind. The problem is that when handling the `IDENTIFIES` table right before we're doing just a `COPY INTO` because we're in `APPEND` mode so we don't have any reference as to what it was copied and we can't therefore make a more selective `INSERT` here.

Anyway, since I'm under the impression that the users table is usually not that big beside some exceptions and considering that the `INSERT OVERWRITE INTO` is more efficient than a `COPY INTO` maybe we can give this a try.

Alternatively I was thinking that we could do two copies when in `APPEND` mode for the `IDENTIFIES`, one in `IDENTIFIES` and one in a `TEMPORARY TABLE`.

fracasula: The advantage of the `INSERT OVERWRITE INTO` is also that the `USERS` table will be duplicates free.

fracasula: @achettyiitr @lvrach :point_up:

fracasula: I'll check if I can benchmark this with an `IDENTIFIES` table that has 30 million records.

achettyiitr: For mattermost, one of the users table is having around 50M entries.

achettyiitr: Also, we would need to merge the traits before appending into users table.

warehouse/integrations/snowflake/snowflake.go
fracasula: > Do you mean the `warehouse/validations` package?

Yes.

warehouse/integrations/snowflake/snowflake.go
achettyiitr: I was thinking we should probably do MERGE always for Users. Since if we keep on appending the entries, when we perform the merging of the traits, it's dataset will keep on increasing.
What do you think? @lvrach

fracasula: Yes, the merging of the traits complicates things since it adds a substantial amount of computation. I need to rethink think. Feedback as usual is welcome guys. We either always do the `MERGE` and or do 2 `COPY` for the `IDENTIFIES` so that here we can be more efficient with the `FIRST_VALUE`s.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -295,20 +307,21 @@ func (sf *Snowflake) DeleteBy(ctx context.Context, tableNames []string, params w
...
+	logFields := []interface{}{
+		lf.SourceID, sf.Warehouse.Source.ID,
+		lf.SourceType, sf.Warehouse.Source.SourceDefinition.Name,
+		lf.DestinationID, sf.Warehouse.Destination.ID,
+		lf.DestinationType, sf.Warehouse.Destination.DestinationDefinition.Name,
+		lf.WorkspaceID, sf.Warehouse.WorkspaceID,
+		lf.Namespace, sf.Namespace,
+		lf.TableName, tableName,
+		lf.LoadTableStrategy, sf.config.loadTableStrategy,
+	}
```
achettyiitr, line 323: ❤️ It definitely cleaned up a lot of duplicate code.

warehouse/integrations/snowflake/snowflake_test.go
```diff
@@ -287,43 +295,56 @@ func TestIntegration(t *testing.T) {
...
-
-				urlConfig := snowflakedb.Config{
-					Account:   cred.Account,
-					User:      cred.User,
-					Role:      cred.Role,
-					Password:  cred.Password,
-					Database:  cred.Database,
-					Warehouse: cred.Warehouse,
+				cancel := bootstrap(tc.appendMode)
+				defer cancel()
```
achettyiitr, line 323: Should we wait here for a graceful shutdown of the server by utilizing either the svcDone channel or passing testing.T of the subtest in bootstrap?

achettyiitr (approved):

warehouse/integrations/snowflake/snowflake_test.go
achettyiitr: Currently `t.Cleanup(func() { <-svcDone })` is being registered at the `TestIntegration` level.

fracasula: @achettyiitr good catch, addressed [here](https://github.com/rudderlabs/rudder-server/pull/3745/commits/2be8bfc54e7d4c1f00c863dc3e293ee2f57abf6a).

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -576,132 +436,266 @@ func (sf *Snowflake) loadTable(ctx context.Context, tableName string, tableSchem
...
+	)
 
-	sf.logger.Infof("SF: Dedup records for table:%s using staging table: %s\n", identityMappingsTable, sqlStatement)
+	sf.logger.Infow("Copying identity mappings for table", lf.TableName, identityMappingsTable, lf.Query, sqlStatement)
 	_, err = dbHandle.ExecContext(ctx, sqlStatement)
 	if err != nil {
-		sf.logger.Errorf("SF: Error running MERGE for dedup: %v\n", err)
-		return
+		sf.logger.Errorw("Error running COPY for table",
+			lf.TableName, identityMappingsTable, lf.Query, sqlStatement, lf.Error, err.Error(),
```
lvrach, line 657: [nit-pick]
```suggestion
			lf.TableName, identityMappingsTable, 
			lf.Query, sqlStatement, 
			lf.Error, err.Error(),
```

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -576,132 +436,266 @@ func (sf *Snowflake) loadTable(ctx context.Context, tableName string, tableSchem
...
+		FROM '%v' %s PATTERN = '.*\.csv\.gz'
+		FILE_FORMAT = ( TYPE = csv FIELD_OPTIONALLY_ENCLOSED_BY = '"' ESCAPE_UNENCLOSED_FIELD = NONE )
+		TRUNCATECOLUMNS = TRUE`,
+		schemaIdentifier, stagingTableName,
+		loadLocation,
+		sf.authString(),
+	)
 
-	sf.logger.Infof("SF: Dedup records for table:%s using staging table: %s\n", identityMappingsTable, sqlStatement)
+	sf.logger.Infow("Copying identity mappings for table", lf.TableName, identityMappingsTable, lf.Query, sqlStatement)
```
lvrach, line 653: Maybe we can:
```suggestion
        logger := sf.logger.With(
        lf.TableName, identityMappingsTable, 
        lf.Query, sqlStatement
       )
	logger.Infow("Copying identity mappings for table")
```

reusing it in the section bellow

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -715,14 +709,17 @@ func (sf *Snowflake) loadUserTables(ctx context.Context) map[string]error {
...
+		lf.SourceID, sf.Warehouse.Source.ID,
+		lf.SourceType, sf.Warehouse.Source.SourceDefinition.Name,
+		lf.DestinationID, sf.Warehouse.Destination.ID,
+		lf.DestinationType, sf.Warehouse.Destination.DestinationDefinition.Name,
+		lf.WorkspaceID, sf.Warehouse.WorkspaceID,
+		lf.Namespace, sf.Namespace,
+		lf.TableName, whutils.UsersTable,
+		lf.LoadTableStrategy, sf.config.loadTableStrategy,
+	}
+	sf.logger.Infow("started loading for identifies and users tables", logFields...)
```
lvrach, line 722: Similarly `.With` can be considered here

lvrach (approved):
Approved, with minor comments

warehouse/integrations/snowflake/snowflake.go
fracasula: I can't reuse the `Query` one but I can reuse the table. Making the changes :+1:

PR #3755, opened 2023-08-17
chore: alerting on reports pileup 

itsmihir:
# Description

Rudder-server aggregates metrics and send the metrics to reporting each minute.

The issue is there can be metrics pileup in server if reporting URL is wrong or reporting is down etc. We want to detect these pileups and setup an alert. 

In this PR we have added a goroutine that will query for the min reported_at report present in the server and will emit a gauge metric with difference in current time and min reported_at every 2 minutes

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-33/alerting-on-reports-pileup-in-server

## Security


enterprise/reporting/reporting.go
```diff
@@ -374,6 +384,19 @@ func (r *HandleT) mainLoop(ctx context.Context, clientName string) {
...
 	r.requestLatency = stats.Default.NewTaggedStat(StatReportingHttpReqLatency, stats.TimerType, tags)
+
+	lastReportedAt := time.Now().UTC().Unix() / 60
+	go func() {
+		// for montering reports pileups
+		for {
+			currentMs := time.Now().UTC().Unix() / 60
+			stats.Default.NewTaggedStat(
+				"reporting_metrics_pileup", stats.GaugeType, stats.Tags{"client": clientName},
+			).Gauge(int(currentMs - lastReportedAt))
```
lvrach, line 395: I understand that we are capturing lag rather than pileup, which in my head is the amount of messages rather the time difference.

I would also suffix with the unit we are capturing: 
```suggestion
			stats.Default.NewTaggedStat(
				"reporting_metrics_ lag_minutes", stats.GaugeType, stats.Tags{"client": clientName},
			).Gauge(int(currentMs - lastReportedAt))
```

However, is better/more conventional to use seconds for measuring duration in Prometheus
https://prometheus.io/docs/practices/naming/

enterprise/reporting/reporting.go
```diff
@@ -374,6 +384,19 @@ func (r *HandleT) mainLoop(ctx context.Context, clientName string) {
...
+			time.Sleep(2 * time.Minute)
```
lvrach, line 396: Avoid time.Sleep especially for such durations of time, can cause issues with graceful shutdown and testing.

```suggestion
			select {
	    case <-ctx.Done():
				break;
			case <-time.After(2 * time.Minute):
			}
		}
			```

**Review by lvrach — CHANGES_REQUESTED** [review]

File: enterprise/reporting/reporting.go (hunk 1 above)
Comment (itsmihir) [reply enterprise/reporting/reporting.go]: reporting metrics are stored at minute level, I don't think using seconds will make difference.

File: enterprise/reporting/reporting.go [hunk 3]
```diff
@@ -218,13 +221,16 @@ func (r *HandleT) getReports(currentMs int64, clientName string) (reports []*typ
 	}
 
 	queryStart := time.Now()
-	err = dbHandle.QueryRow(sqlStatement).Scan(&queryMin)
-	if err != nil && err != sql.ErrNoRows {
+	ctx, cancel := context.WithTimeout(context.Background(), r.dbQueryTimeout)
+	defer cancel()
+	err = dbHandle.QueryRowContext(ctx, sqlStatement).Scan(&queryMin)
+
+	if err != nil && err != sql.ErrNoRows && err != context.DeadlineExceeded {
```
lvrach, line 228: In case `err != sql.ErrNoRows` should we return `err = nil` or use a custom error:

```suggestion
    if err != sql.ErrNoRows {
          return nil, 0, ErrNoReports
    }
	if err != nil && err != context.DeadlineExceeded {
```

It is not a good idea to expose `sql.ErrNoRows` to the rest of the system, since it is implementation/SQL specific.

enterprise/reporting/reporting.go
```diff
@@ -397,6 +423,7 @@ func (r *HandleT) mainLoop(ctx context.Context, clientName string) {
 			continue
 		}
 
+		lastReportedAt = reportedAt
```
lvrach, line 426: Is it ok to update this if reports are empty or `ErrNoReports` (sql. ErrNoRows) is returned  ?

itsmihir: Yes we should update `lastReportedAt`. we don't what to trigger the alert if there are no metric to send to reporting.

itsmihir: If the reports are empty this line will no be executed.

enterprise/reporting/reporting.go
itsmihir: returning custom error is query timeouts and returning nil if there are 0 reports

enterprise/reporting/reporting.go
```diff
@@ -374,6 +384,23 @@ func (r *HandleT) mainLoop(ctx context.Context, clientName string) {
...
 	r.getReportsQueryTime = stats.Default.NewTaggedStat(StatReportingGetReportsQueryTime, stats.TimerType, tags)
 	r.requestLatency = stats.Default.NewTaggedStat(StatReportingHttpReqLatency, stats.TimerType, tags)
+
+	lastReportedAt := time.Now().UTC().Unix() / 60
+	go func() {
+		// for montering reports pileups
+		for {
+			currentMs := time.Now().UTC().Unix() / 60
+			stats.Default.NewTaggedStat(
+				"reporting_metrics_ lag_minutes", stats.GaugeType, stats.Tags{"client": clientName},
```
achettyiitr, line 394:
```suggestion
			stats.Default.NewTaggedStat(
				"reporting_metrics_lag_minutes", stats.GaugeType, stats.Tags{"client": clientName},
```
Any reason why we have an extra space in the tag name?

enterprise/reporting/reporting.go
```diff
@@ -374,6 +384,23 @@ func (r *HandleT) mainLoop(ctx context.Context, clientName string) {
 	r.getMinReportedAtQueryTime = stats.Default.NewTaggedStat(StatReportingGetMinReportedAtQueryTime, stats.TimerType, tags)
 	r.getReportsQueryTime = stats.Default.NewTaggedStat(StatReportingGetReportsQueryTime, stats.TimerType, tags)
 	r.requestLatency = stats.Default.NewTaggedStat(StatReportingHttpReqLatency, stats.TimerType, tags)
+
+	lastReportedAt := time.Now().UTC().Unix() / 60
+	go func() {
+		// for montering reports pileups
```
achettyiitr, line 390:
```suggestion
		// for monitering reports pileups
```

enterprise/reporting/reporting.go
```diff
@@ -374,6 +384,23 @@ func (r *HandleT) mainLoop(ctx context.Context, clientName string) {
...
+			stats.Default.NewTaggedStat(
+				"reporting_metrics_ lag_minutes", stats.GaugeType, stats.Tags{"client": clientName},
+			).Gauge(int(currentMs - lastReportedAt))
+			select {
+			case <-ctx.Done():
+				break
+			case <-time.After(2 * time.Minute):
+			}
+		}
+	}()
```
achettyiitr, line 402: [Optional]
```suggestion

	lastReportedAt := time.Now().UTC().Unix() / 60
	go func() {
		// for monitoring reports pileups
		for {
			currentMs := time.Now().UTC().Unix() / 60
			lagMinutes := int(currentMs - lastReportedAt)

			reportingLag.Gauge(lagMinutes)

			select {
			case <-ctx.Done():
				break
			case <-time.After(2 * time.Minute):
			}
		}
	}()
```

enterprise/reporting/reporting.go
```diff
@@ -374,6 +384,23 @@ func (r *HandleT) mainLoop(ctx context.Context, clientName string) {
 	r.getMinReportedAtQueryTime = stats.Default.NewTaggedStat(StatReportingGetMinReportedAtQueryTime, stats.TimerType, tags)
 	r.getReportsQueryTime = stats.Default.NewTaggedStat(StatReportingGetReportsQueryTime, stats.TimerType, tags)
 	r.requestLatency = stats.Default.NewTaggedStat(StatReportingHttpReqLatency, stats.TimerType, tags)
+
+	lastReportedAt := time.Now().UTC().Unix() / 60
```
achettyiitr, line 388: Since this is being accessed from 2 go-routines, there can be a data race happening.
```suggestion
	lastReportedAt := atomic.NewInt64(time.Now().UTC().Unix() / 60)
```

get: lastReportedAt.Load()
set: lastReportedAt.Store(currentMs)

enterprise/reporting/reporting.go
itsmihir: removed extra space.

enterprise/reporting/reporting.go
itsmihir: It looks more cleaner thanks.

achettyiitr (approved):

enterprise/reporting/reporting.go
```diff
@@ -374,6 +385,27 @@ func (r *HandleT) mainLoop(ctx context.Context, clientName string) {
 	r.getMinReportedAtQueryTime = stats.Default.NewTaggedStat(StatReportingGetMinReportedAtQueryTime, stats.TimerType, tags)
 	r.getReportsQueryTime = stats.Default.NewTaggedStat(StatReportingGetReportsQueryTime, stats.TimerType, tags)
 	r.requestLatency = stats.Default.NewTaggedStat(StatReportingHttpReqLatency, stats.TimerType, tags)
+	reportingLag := stats.Default.NewTaggedStat(
+		"reporting_metrics_lag_minutes", stats.GaugeType, stats.Tags{"client": clientName},
+	)
+
+	lastReportedAt := atomic.NewInt64(time.Now().UTC().Unix() / 60)
+	go func() {
```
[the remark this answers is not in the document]
itsmihir: But this function might not finish because of the infinite for loop. And that will block the `mainLoop` to return.

enterprise/reporting/reporting.go
```diff
@@ -70,14 +72,15 @@ type HandleT struct {
...
-	var sleepInterval, mainLoopSleepInterval time.Duration
+	var sleepInterval, mainLoopSleepInterval, dbQueryTimeout time.Duration
 	var maxOpenConnections int
 	reportingServiceURL := config.GetString("REPORTING_URL", "https://reporting.rudderstack.com/")
 	reportingServiceURL = strings.TrimSuffix(reportingServiceURL, "/")
 	sourcesWithEventNameTrackingDisabled := config.GetStringSlice("Reporting.sourcesWithEventNameTrackingDisabled", []string{})
 
 	config.RegisterDurationConfigVariable(5, &mainLoopSleepInterval, true, time.Second, "Reporting.mainLoopSleepInterval")
 	config.RegisterDurationConfigVariable(30, &sleepInterval, true, time.Second, "Reporting.sleepInterval")
+	config.RegisterDurationConfigVariable(60, &dbQueryTimeout, true, time.Second, "Reporting.dbQueryTimeout")
```
atzoum, line 83: Please switch to the new reloadable config api for all configuration options, e.g.

```go
var dbQueryTimeout *config.Reloadable[time.Duration]


```

enterprise/reporting/reporting.go
```diff
@@ -218,13 +222,20 @@ func (r *HandleT) getReports(currentMs int64, clientName string) (reports []*typ
 	}
 
 	queryStart := time.Now()
-	err = dbHandle.QueryRow(sqlStatement).Scan(&queryMin)
-	if err != nil && err != sql.ErrNoRows {
+	ctx, cancel := context.WithTimeout(context.Background(), r.dbQueryTimeout)
+	defer cancel()
+	err = dbHandle.QueryRowContext(ctx, sqlStatement).Scan(&queryMin)
+
+	if err != nil && err != sql.ErrNoRows && err != context.DeadlineExceeded {
```
atzoum, line 229:
```suggestion
	if err != nil && err != sql.ErrNoRows && ctx.Err() == nil {
```

enterprise/reporting/reporting.go
```diff
@@ -218,13 +222,20 @@ func (r *HandleT) getReports(currentMs int64, clientName string) (reports []*typ
...
 		panic(err)
 	}
+	if err == context.DeadlineExceeded {
```
atzoum, line 232:
```suggestion
	if ctx.Err() != nil {
```

enterprise/reporting/reporting.go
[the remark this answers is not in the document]
atzoum: The `mainLoop` shall only return after all goroutines that it has started are properly stopped

lvrach (approved):

PR #3759, opened 2023-08-18
chore: add support for logging duplicate messageIDs

achettyiitr:
# Description

- Added support for logging duplicate message Ids for an upload for Snowflake.

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-68/get-exact-duplicate-records-for-an-upload-id
- https://linear.app/rudderstack/issue/PIPE-67/addition-logging-for-warehouse-duplicates

## Security


achettyiitr:
Tried the queries for the 30 days interval for mattermost and it looks good.

Screenshot 2023-08-17 at 3 49 49 PM
Screenshot 2023-08-17 at 3 51 53 PM

lvrach (approved):

warehouse/utils/context.go
```diff
@@ -0,0 +1,20 @@
+package warehouseutils
+
+import (
+	"context"
+)
+
+type key int
+
+const (
+	uploadID key = iota
```
fracasula, line 10: [optional] The code is correct but I suggest using `struct` types to handle context keys since `uploadID` can too easily be converted by someone to `int` opening this code to unintentional overwrites of the `uploadID`.

```go
type uploadIDContextKey struct {}

func CtxWithUploadID(ctx context.Context, uid int64) context.Context {
	return context.WithValue(ctx, uploadIDContextKey{}, uid)
}

func UploadIDFromCtx(ctx context.Context) (int64, bool) {
	userIP, ok := ctx.Value(uploadIDContextKey{}).(int64)
	return userIP, ok
}
```

Less lines of code too :stuck_out_tongue:

fracasula (approved):

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -143,6 +145,15 @@ type optionalCreds struct {
 	schemaName string
 }
 
+type duplicateMessage struct {
+	id         string
+	receivedAt time.Time
+}
+
+func (m *duplicateMessage) String() string {
+	return fmt.Sprintf("%s@%s", m.id, m.receivedAt.Format("2006-01-02 15:04:05"))
```
fracasula, line 154: [optional] This can be called for lots of messages in a loop and we're talking a simple string interpolation. I suggest not using reflection here.

warehouse/utils/context.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func CtxWithUploadID(ctx context.Context, uid int64) context.Context {
+	return context.WithValue(ctx, uploadID, uid)
+}
+
+func UploadIDFromCtx(ctx context.Context) (int64, bool) {
+	userIP, ok := ctx.Value(uploadID).(int64)
```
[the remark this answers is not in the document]
achettyiitr: Thanks for pointing it out.

warehouse/integrations/snowflake/snowflake.go
achettyiitr: Good point. Simple interpolation would work here. Just for logging purposes.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -576,6 +597,35 @@ func (sf *Snowflake) loadTable(ctx context.Context, tableName string, tableSchem
...
+	if updated > 0 {
+		logFields := []interface{}{
+			logfield.SourceID, sf.Warehouse.Source.ID,
+			logfield.SourceType, sf.Warehouse.Source.SourceDefinition.Name,
+			logfield.DestinationID, sf.Warehouse.Destination.ID,
+			logfield.DestinationType, sf.Warehouse.Destination.DestinationDefinition.Name,
+			logfield.WorkspaceID, sf.Warehouse.WorkspaceID,
+			logfield.Namespace, sf.Namespace,
+			logfield.TableName, tableName,
+		}
```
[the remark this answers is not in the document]
achettyiitr: No need for this now I guess.

PR #3764, opened 2023-08-19
chore: option to pass *config.Config to jobsdb

Sidddddarth:
# Description

Option(`Optfunc`) to have jobsdb-handle-specific config, which falls back to `config.Default` if not passed.
This can help avoid configuration mixups between tests, each jobsdb handle can have specific config to itself, and also to test actual runtime behaviour by changing config values instead of setting `jobsdb.Handle`'s internal fields.

Also moved and grouped jobsdb config fields around to help with readability. eg.: it wasn't entirely clear(to me at first glance) how the number of `maxOpenConnections` were being set in case of a `write`/`read`/`readWrite` jobsdb.

Additionally:
`misc.GetConnectionString` now takes in `*config.Config` as a parameter, using it to fetch db connection details.

## Linear Ticket

[jobsdb handle specific config](https://linear.app/rudderstack/issue/PIPE-166/jobsdb-handle-specific-config)

## Security

### DIFF
--- a/archiver/archiver_test.go
+++ b/archiver/archiver_test.go
@@ -71,7 +71,6 @@ func TestJobsArchival(t *testing.T) {
 		t.Setenv("JOBS_DB_USER", postgresResource.User)
 		t.Setenv("JOBS_DB_PASSWORD", postgresResource.Password)
 	}
-	jobsdb.Init()
 	misc.Init()
 	jd := &jobsdb.Handle{
 		TriggerAddNewDS: func() <-chan time.Time {
--- a/archiver/archiver_test.go
+++ b/archiver/archiver_test.go
@@ -16,7 +16,6 @@ import (
 	"github.com/tidwall/gjson"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
-	c "github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/stats"
 
 	"github.com/rudderlabs/rudder-go-kit/bytesize"
@@ -24,7 +23,6 @@ import (
 	"github.com/rudderlabs/rudder-go-kit/testhelper/docker/resource"
 	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 	"github.com/rudderlabs/rudder-server/jobsdb"
-	"github.com/rudderlabs/rudder-server/jobsdb/prebackup"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 
 	"github.com/rudderlabs/rudder-server/services/fileuploader"
@@ -52,6 +50,20 @@ func TestJobsArchival(t *testing.T) {
 	postgresResource, err := resource.SetupPostgres(pool, cleanup)
 	require.NoError(t, err, "failed to setup postgres resource")
 
+	config.Reset()
+	defer config.Reset()
+	t.Setenv("MINIO_SSL", "false")
+	t.Setenv("JOBS_DB_DB_NAME", postgresResource.Database)
+	t.Setenv("JOBS_DB_NAME", postgresResource.Database)
+	t.Setenv("JOBS_DB_HOST", postgresResource.Host)
+	t.Setenv("JOBS_DB_PORT", postgresResource.Port)
+	t.Setenv("JOBS_DB_USER", postgresResource.User)
+	t.Setenv("JOBS_DB_PASSWORD", postgresResource.Password)
+	misc.Init()
+
+	jd := jobsdb.NewForReadWrite("archiver", jobsdb.WithClearDB(false))
+	require.NoError(t, jd.Start())
+
 	minioResource = make([]*destination.MINIOResource, uniqueWorkspaces)
 	for i := 0; i < uniqueWorkspaces; i++ {
 		minioResource[i], err = destination.SetupMINIO(pool, cleanup)
@@ -60,31 +72,6 @@ func TestJobsArchival(t *testing.T) {
 
 	jobs, err := readGzipJobFile(seedJobsFileName)
 	require.NoError(t, err, "failed to read jobs file")
-
-	config.Reset()
-	{
-		t.Setenv("MINIO_SSL", "false")
-		t.Setenv("JOBS_DB_DB_NAME", postgresResource.Database)
-		t.Setenv("JOBS_DB_NAME", postgresResource.Database)
-		t.Setenv("JOBS_DB_HOST", postgresResource.Host)
-		t.Setenv("JOBS_DB_PORT", postgresResource.Port)
-		t.Setenv("JOBS_DB_USER", postgresResource.User)
-		t.Setenv("JOBS_DB_PASSWORD", postgresResource.Password)
-	}
-	misc.Init()
-	jd := &jobsdb.Handle{
-		TriggerAddNewDS: func() <-chan time.Time {
-			return make(chan time.Time)
-		},
-	}
-	require.NoError(t, jd.Setup(
-		jobsdb.ReadWrite,
-		false,
-		"gw",
-		[]prebackup.Handler{},
-		nil,
-	))
-	require.NoError(t, jd.Start())
 
 	require.NoError(t, jd.Store(ctx, jobs))
 
@@ -150,7 +137,7 @@ func TestJobsArchival(t *testing.T) {
 	archiver := New(
 		jd,
 		fileUploaderProvider,
-		c.New(),
+		config.Default,
 		stats.Default,
 		WithArchiveTrigger(
 			func() <-chan time.Time {
--- a/archiver/archiver_test.go
+++ b/archiver/archiver_test.go
@@ -26,7 +26,6 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/misc"
 
 	"github.com/rudderlabs/rudder-server/services/fileuploader"
-	"github.com/rudderlabs/rudder-server/testhelper"
 	"github.com/rudderlabs/rudder-server/testhelper/destination"
 )
 
@@ -45,9 +44,8 @@ func TestJobsArchival(t *testing.T) {
 
 	pool, err := dockertest.NewPool("")
 	require.NoError(t, err, "Failed to create docker pool")
-	cleanup := &testhelper.Cleanup{}
 
-	postgresResource, err := resource.SetupPostgres(pool, cleanup)
+	postgresResource, err := resource.SetupPostgres(pool, t)
 	require.NoError(t, err, "failed to setup postgres resource")
 	c := config.New()
 	c.Set("MINIO_SSL", "false")
@@ -63,7 +61,7 @@ func TestJobsArchival(t *testing.T) {
 
 	minioResource = make([]*destination.MINIOResource, uniqueWorkspaces)
 	for i := 0; i < uniqueWorkspaces; i++ {
-		minioResource[i], err = destination.SetupMINIO(pool, cleanup)
+		minioResource[i], err = destination.SetupMINIO(pool, t)
 		require.NoError(t, err, "failed to setup minio resource")
 	}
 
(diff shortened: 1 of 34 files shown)


jobsdb/jobsdb.go
```diff
@@ -531,13 +549,13 @@ var dbInvalidJsonErrors = map[string]struct{}{
...
 // pathPrefix = by default is the jobsdb table prefix, is the path appended before instanceID in s3 folder structure
 func (jd *HandleT) registerBackUpSettings() {
-	config.RegisterBoolConfigVariable(true, &masterBackupEnabled, true, "JobsDB.backup.enabled")
-	config.RegisterBoolConfigVariable(false, &jd.BackupSettings.instanceBackupEnabled, true, fmt.Sprintf("JobsDB.backup.%v.enabled", jd.tablePrefix))
-	config.RegisterBoolConfigVariable(false, &jd.BackupSettings.FailedOnly, false, fmt.Sprintf("JobsDB.backup.%v.failedOnly", jd.tablePrefix))
-	config.RegisterStringConfigVariable(jd.tablePrefix, &pathPrefix, false, fmt.Sprintf("JobsDB.backup.%v.pathPrefix", jd.tablePrefix))
-	config.RegisterDurationConfigVariable(10, &jd.maxBackupRetryTime, false, time.Minute, "JobsDB.backup.maxRetry")
-	config.RegisterDurationConfigVariable(10, &jd.refreshDSTimeout, true, time.Minute, "JobsDB.refreshDS.timeout")
-	config.RegisterDurationConfigVariable(10, &jd.migrateDSTimeout, true, time.Minute, "JobsDB.migrateDS.timeout")
+	jd.configGetter.RegisterBoolConfigVariable(true, &masterBackupEnabled, true, "JobsDB.backup.enabled")
```
atzoum, line 552: why is this left as a global variable?

Sidddddarth: missed it🤡
there's so many😅

jobsdb/integration_test.go
```diff
@@ -575,8 +577,11 @@ func TestJobsDB(t *testing.T) {
 			TriggerMigrateDS: func() <-chan time.Time {
 				return triggerMigrateDS
 			},
+			configGetter: c,
 		}
 		prefix := strings.ToLower(rand.String(5))
+		t.Setenv(config.ConfigKeyToEnv(config.DefaultEnvPrefix, "JobsDB.jobDoneMigrateThreshold"), "0.7")
+		t.Setenv(config.ConfigKeyToEnv(config.DefaultEnvPrefix, "JobsDB.jobMinRowsMigrateThreshold"), "0.6")
```
atzoum, line 584: env variables are global, better set the proper configuration parameters in the config instance itself
```suggestion
		c.Set("JobsDB.jobDoneMigrateThreshold", 0.7)
		c.Set("JobsDB.jobMinRowsMigrateThreshold", 0.6)
```

jobsdb/migration_test.go
```diff
@@ -294,9 +298,17 @@ func TestMigration(t *testing.T) {
...
+		t.Setenv(config.ConfigKeyToEnv(config.DefaultEnvPrefix, "JobsDB.vacuumAnalyzeStatusTableThreshold"), "4")
+		t.Setenv(
+			config.ConfigKeyToEnv(
+				config.DefaultEnvPrefix,
+				"JobsDB.vacuumFullStatusTableThreshold",
+			),
+			fmt.Sprint(
+				originalTableSizes[fmt.Sprintf("%s_job_status_2", tablePrefix)]-1,
+			),
+		)
```
atzoum, line 311: better use `c.Set` instead of `t.Setenv`

router/batchrouter/batchrouter_isolation_test.go
```diff
@@ -258,7 +259,7 @@ func BatchrouterIsolationScenario(t testing.TB, spec *BrtIsolationScenarioSpec)
 	config.Set("RUDDER_TMPDIR", os.TempDir())
 
 	t.Logf("Seeding batch_rt jobsdb with jobs")
-	m.seedBrtDB(t, spec)
+	m.seedBrtDB(t, spec, postgresContainer.DB)
```
atzoum, line 262: I don't think this makes much of a difference here, since we are messing with the global config in this test anyway. Changing a very small part of this test wouldn't make it behave better.

jobsdb/jobsdb.go
```diff
@@ -630,65 +648,6 @@ func init() {
...
-	config.RegisterInt64ConfigVariable(64*bytesize.MB, &backupMaxTotalPayloadSize, true, 1, "JobsDB.maxBackupTotalPayloadSize")
-	config.RegisterDurationConfigVariable(30, &migrateDSLoopSleepDuration, true, time.Second, []string{"JobsDB.migrateDSLoopSleepDuration", "JobsDB.migrateDSLoopSleepDurationInS"}...)
-	config.RegisterDurationConfigVariable(5, &addNewDSLoopSleepDuration, true, time.Second, []string{"JobsDB.addNewDSLoopSleepDuration", "JobsDB.addNewDSLoopSleepDurationInS"}...)
-	config.RegisterDurationConfigVariable(10, &refreshDSListLoopSleepDuration, true, time.Second, []string{"JobsDB.refreshDSListLoopSleepDuration", "JobsDB.refreshDSListLoopSleepDurationInS"}...)
-	config.RegisterDurationConfigVariable(5, &backupCheckSleepDuration, true, time.Second, []string{"JobsDB.backupCheckSleepDuration", "JobsDB.backupCheckSleepDurationIns"}...)
-	config.RegisterDurationConfigVariable(120, &cacheExpiration, true, time.Minute, []string{"JobsDB.cacheExpiration"}...)
-	config.RegisterDurationConfigVariable(24, &jobCleanupFrequency, true, time.Hour, []string{"JobsDB.jobCleanupFrequency"}...)
-}
-
-func Init() {
```
atzoum, line 687: 👍

archiver/archiver_test.go
```diff
@@ -70,20 +69,11 @@ func TestJobsArchival(t *testing.T) {
...
-	}
-	require.NoError(t, jd.Setup(
-		jobsdb.ReadWrite,
-		false,
+	jd := jobsdb.NewForReadWrite(
 		"gw",
-		[]prebackup.Handler{},
-		nil,
-	))
+		jobsdb.WithDBHandle(postgresResource.DB),
```
atzoum, line 75: What's the point of `t.Setenv` above? Would it be better to use `WithConfig` instead of `WithDBHandle`? Supporting both seems like an overkill to me

jobsdb/jobsdb.go
```diff
@@ -497,6 +497,24 @@ type HandleT struct {
 		mu      sync.Mutex
 		started bool
 	}
+
+	configGetter *config.Config
+	config       struct {
```
atzoum, line 502: there are various other fields in the Handle struct which would qualify as config fields

archiver/archiver_test.go
Sidddddarth: tried c.Set("DB.host") etc. but the `GetConnectionString` uses default config handle, so `c.Set("DB.host")` is doesn't work.
In all other tests, we either use `t.Setenv("JOBS_DB_HOST", postgresContainer.Host)` or `config.Set("DB.name", postgresContainer.Database)`(default global config).

`GetConnectionString` is used in a lot of places without respective handle specific config.
So going ahead with `t.Setenv` for now.

jobsdb/jobsdb.go
Sidddddarth: have added them all and segregated things(`Handle.migrationConfig` and `Handle.backupConfig`).
I think there can be some more cleanup but right now I don't see it.

utils/misc/dbutils.go
```diff
@@ -11,13 +11,16 @@ import (
...
-	host := config.GetString("DB.host", "localhost")
-	user := config.GetString("DB.user", "ubuntu")
-	dbname := config.GetString("DB.name", "ubuntu")
-	port := config.GetInt("DB.port", 5432)
-	password := config.GetString("DB.password", "ubuntu") // Reading secrets from
-	sslmode := config.GetString("DB.sslMode", "disable")
+func GetConnectionString(c *config.Config) string {
+	if c == nil {
+		c = config.Default
+	}
```
atzoum, line 17: Passing a config shouldn't be optional, in the long run we would like to get rid of `config.Default`, let's not introduce it where we can avoid it :)

Sidddddarth: a lot of dependants of this function don't have specific config, would need more refactoring in various areas of the repo.

processor/manager_test.go
```diff
@@ -152,19 +150,16 @@ func TestProcessorManager(t *testing.T) {
...
-	maxDSSize := 10
+	config.Reset()
+	c := config.New()
+	c.Set("JobsDB.maxDSSize", 10)
 	// tempDB is created to observe/manage the GW DB from the outside without touching the actual GW DB.
-	tempDB := jobsdb.Handle{
-		MaxDSSize: &maxDSSize,
-		TriggerAddNewDS: func() <-chan time.Time {
-			return triggerAddNewDS
-		},
```
atzoum, line 162: how are we overriding jobsdb now so that it doesn't trigger the addNewDS loop?

processor/manager_test.go
```diff
@@ -152,19 +150,16 @@ func TestProcessorManager(t *testing.T) {
...
-	}
-
-	err := tempDB.Setup(jobsdb.Write, true, "gw", []prebackup.Handler{}, fileuploader.NewDefaultProvider())
-	require.NoError(t, err)
+	tempDB := jobsdb.NewForWrite(
+		"gw",
+		jobsdb.WithConfig(c),
+	)
 	defer tempDB.TearDown()
```
atzoum, line 161: this should be defered after starting the db, in the line below

warehouse/router_test.go
```diff
@@ -906,6 +906,12 @@ func TestRouter(t *testing.T) {
 				pgResource, err := resource.SetupPostgres(pool, t)
 				require.NoError(t, err)
 
+				config.Set("DB.host", pgResource.Host)
+				config.Set("DB.port", pgResource.Port)
+				config.Set("DB.user", pgResource.User)
+				config.Set("DB.password", pgResource.Password)
+				config.Set("DB.name", pgResource.Database)
+
```
atzoum, line 914: using the global config and parallel tests seems like a recipe for disaster to me

processor/manager_test.go
Sidddddarth: It wasn't being used anywhere in the test.

warehouse/router_test.go
Sidddddarth: yeah
only option was to remove the `t.Parallel()`.
can't use `t.Setenv` in parallel tests.

atzoum: does this test rely on these config options?

jobsdb/jobsdb.go
```diff
@@ -534,15 +562,15 @@ var dbInvalidJsonErrors = map[string]struct{}{
 // instanceBackupFailedAndAborted = true => the individual jobdb backsup failed and aborted jobs only
 // pathPrefix = by default is the jobsdb table prefix, is the path appended before instanceID in s3 folder structure
 func (jd *Handle) registerBackUpSettings() {
```
atzoum, line 564: since this doesn't seem to contain only backup settings can we remove the function and move its contents to `loadConfig`?

atzoum:
thank you @Sidddddarth! with respect to the description of this pull request, I believe it deserves some more attention to include all introduced changes

warehouse/router_test.go
Sidddddarth: correction: it doesn't, removing these lines

processor/manager_test.go
atzoum: Could it be that the test wanted to simply disable this trigger?

utils/misc/dbutils.go
atzoum: Yes as simple as that... just present the naked truth in plain sight!

atzoum (approved):

processor/manager_test.go
Sidddddarth: true
didn't think of it that way😅
Fortunately this didn't affect this particular test.
But in the future with the config dependency, we could set the `addDSLoop-timer` to a sufficiently large duration.

cisse21 (approved):

PR #3766, opened 2023-08-19
chore: deltalake error message length

achettyiitr:
# Description

- While creating an external table, we are getting an error since there is already some data present specifying some other schema, and in the error messages, it's dumping the entire other schema which is coming to be very huge. It's reaching around 20MB.
- For delta lake, trim error if it exceeds 64KB

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-164/deltalake-error-length

## Security

### DIFF
--- a/warehouse/integrations/deltalake/deltalake.go
+++ b/warehouse/integrations/deltalake/deltalake.go
@@ -463,13 +463,13 @@ func (d *Deltalake) CreateTable(ctx context.Context, tableName string, columns m
 
 	_, err := d.DB.ExecContext(ctx, query)
 	if err != nil {
-		return fmt.Errorf("creating table: %w", d.trimErrorMessage(err))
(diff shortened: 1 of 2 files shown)


warehouse/integrations/deltalake/deltalake.go
```diff
@@ -461,12 +463,21 @@ func (d *Deltalake) CreateTable(ctx context.Context, tableName string, columns m
...
 	return nil
 }
 
+func (d *Deltalake) trimErrorMessage(baseError error) error {
+	errorString := baseError.Error()
+
+	if len(errorString) <= d.config.maxErrorLength {
+		return baseError
+	}
+	return errors.New(errorString[:d.config.maxErrorLength])
```
fracasula, line 478: It looks easy to have a unit test for this one.

fracasula (approved):

PR #3773, opened 2023-08-22
chore: warehouse tests race detection

fracasula:
# Description

Running warehouse tests with `-race`.

### Changelog
* running warehouse tests with `-race` in order to be able to detect bugs due unregulated concurrent access to the same variable(s)
* added a way to clone `model.Schema` because even if we use a mutex to protect the concurrent access, we're still passing around a reference so a change to the map would still count as a data race
* moved `Schema` into its own package that now has both a public and a private API
   * this was done to decouple and to make sure all concurrent accesses are handled properly since the mutexes are now encapsulated in the type
   * moved public API (exported methods) on top
* moved `uploadSchema` into `UploadJob`
* moved `handleSchemaChange` and its errors to `slave_worker` since it's specific to that business logic

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-119/run-warehouse-tests-with-race) >

## Security


warehouse/integrations/snowflake/snowflake_test.go
```diff
@@ -293,8 +293,6 @@ func TestIntegration(t *testing.T) {
 			tc := tc
 
 			t.Run(tc.name, func(t *testing.T) {
-				t.Parallel()
```
fracasula, line 296: This will go away anyway due to [this PR](https://github.com/rudderlabs/rudder-server/pull/3745).

lvrach (approved):

achettyiitr (approved):

PR #3780, opened 2023-08-24
chore: use memoized payloadFunc for dedup

Sidddddarth:
# Description

using the memoized `payloadFunc` for deduplication.

## Linear Ticket

< Replace with Linear Link >

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1446,7 +1446,7 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 				proc.logger.Errorf("Dropping Job since Source not found for sourceId %q: %v", sourceId, sourceError)
 				continue
 			}
-			payload := ro.Memoize(func() json.RawMessage {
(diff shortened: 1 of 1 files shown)


cisse21 (approved):

achettyiitr (approved):

processor/processor.go
```diff
@@ -1446,10 +1446,17 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 				proc.logger.Errorf("Dropping Job since Source not found for sourceId %q: %v", sourceId, sourceError)
 				continue
 			}
+			payload := ro.Memoize(func() json.RawMessage {
```
fracasula, line 1449: So this one here is called `payload` but later on we got a `payloadFunc` (line 1486) that is also returning a payload, all within the same loop :thinking: What the func is going on? :stuck_out_tongue_winking_eye: 

Jokes aside, imo we could use some better names for these variables.

Sidddddarth: >What the func is going on? 😜

😂

PR #3786, opened 2023-08-25
chore: use structured logging in jobsdb backup

Sidddddarth:
# Description

Using structured logging in `jobsdb.Backup`.
Downgraded some logs to `debug` level(when workspaces are configured to skip backups/`"no storage settings found for workspace"`).

## Linear Ticket

[alerts for backup failures](https://linear.app/rudderstack/issue/PIPE-180/add-alerts-in-case-of-backup-failures)

## Security

### DIFF
--- a/jobsdb/backup.go
+++ b/jobsdb/backup.go
@@ -32,7 +32,7 @@ func (jd *Handle) IsMasterBackupEnabled() bool {
 func (jd *Handle) backupDSLoop(ctx context.Context) {
 	sleepMultiplier := time.Duration(1)
 
-	jd.logger.Info("BackupDS loop is running")
+	jd.logger.Info("backupDSLoop is running")
 
 	for {
 		select {
@@ -103,7 +103,7 @@ func (jd *Handle) backupDSLoop(ctx context.Context) {
 			if !jd.conf.skipMaintenanceError {
 				panic(err)
 			}
-			jd.logger.Errorf("[JobsDB] :: Failed to backup dataset. Err: %s", err.Error())
+			jd.logger.Errorw("failed to backup dataset", "error", err.Error())
 		}
 
 	}
@@ -138,7 +138,7 @@ func (jd *Handle) uploadDumps(ctx context.Context, dumps map[string]string) erro
 		operation := func() error {
 			if err := jd.uploadTableDump(ctx, wrkId, path); err != nil {
 				jd.logger.Errorw(
-					"failed backup for workspaceId",
+					"failed backup upload",
 					"workspaceID", wrkId,
 					"error", err.Error(),
 				)
@@ -149,7 +149,11 @@ func (jd *Handle) uploadDumps(ctx context.Context, dumps map[string]string) erro
 		}
 		g.Go(misc.WithBugsnag(func() error {
 			return backoff.RetryNotify(operation, backoff.WithMaxRetries(backoff.NewExponentialBackOff(), uint64(jd.config.GetInt("JobsDB.backup.maxRetries", 100))), func(err error, d time.Duration) {
-				jd.logger.Errorf("[JobsDB] :: Retrying upload workspaceId %v. Error: %s", wrkId, err.Error())
+				jd.logger.Errorw(
+					"retrying backup upload",
+					"workspaceID", wrkId,
+					"error", err.Error(),
+				)
 			})
 		}))
 	}
@@ -176,7 +180,7 @@ func (jd *Handle) failedOnlyBackup(ctx context.Context, backupDSRange *dataSetRa
 		return nil
 	}
 
-	jd.logger.Infof("[JobsDB] :: Backing up table (failed only backup): %v", tableName)
+	jd.logger.Infow("backing up table (failed only backup)", "table", tableName)
 
 	start := time.Now()
 	getFileName := func(workspaceID string) (string, error) {
@@ -200,7 +204,11 @@ func (jd *Handle) failedOnlyBackup(ctx context.Context, backupDSRange *dataSetRa
 	}()
 	err = jd.uploadDumps(ctx, dumps)
 	if err != nil {
-		jd.logger.Errorf("[JobsDB] :: Failed to upload dumps for table: %s. Error: %v", tableName, err)
+		jd.logger.Errorw(
+			"failed to upload failedOnly jobs backup",
+			"table", tableName,
+			"error", err.Error(),
+		)
 		return nil
 	}
 	stats.Default.NewTaggedStat("total_TableDump_TimeStat", stats.TimerType, stats.Tags{"customVal": jd.tablePrefix}).Since(start)
@@ -227,7 +235,7 @@ func (jd *Handle) backupJobsTable(ctx context.Context, backupDSRange *dataSetRan
 		return nil
 	}
 
-	jd.logger.Infof("[JobsDB] :: Backing up table (jobs table): %v", tableName)
+	jd.logger.Infow("backing up table (jobs table)", "table", tableName)
 
 	start := time.Now()
 
@@ -261,7 +269,8 @@ func (jd *Handle) backupJobsTable(ctx context.Context, backupDSRange *dataSetRan
 	err = jd.uploadDumps(ctx, dumps)
 	if err != nil {
 		jd.logger.Errorw(
-			"Failed to upload backup",
+			"failed to upload jobs backup",
+			"table", tableName,
 			"error", err,
 		)
 		return nil
@@ -292,7 +301,7 @@ func (jd *Handle) backupStatusTable(ctx context.Context, backupDSRange *dataSetR
 		return nil
 	}
 
-	jd.logger.Infof("[JobsDB] :: Backing up table (status table): %v", tableName)
+	jd.logger.Infow("backing up table (status table)", "table", tableName)
 
 	start := time.Now()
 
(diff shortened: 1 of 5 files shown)


jobsdb/backup.go
```diff
@@ -137,7 +137,11 @@ func (jd *Handle) uploadDumps(ctx context.Context, dumps map[string]string) erro
 		path := filePath
 		operation := func() error {
 			if err := jd.uploadTableDump(ctx, wrkId, path); err != nil {
-				jd.logger.Errorf("[JobsDB] :: Failed to upload workspaceId %v. Error: %s", wrkId, err.Error())
+				jd.logger.Errorw(
+					"failed backup for workspaceId",
```
atzoum, line 141: I don't see any need to repeat the workspaceId literal in the error's text
```suggestion
					"failed backup",
```

jobsdb/backup.go
```diff
@@ -554,15 +561,21 @@ func (jd *Handle) createTableDumps(ctx context.Context, queryFunc func(int64) st
 			preferences, err := jd.conf.backup.fileUploaderProvider.GetStoragePreferences(workspaceID)
 			if errors.Is(err, fileuploader.NoStorageForWorkspaceError) {
 				offset++
-				jd.logger.Infof("getting storage preferences failed with error: %w for workspaceID: %s", err, workspaceID)
+				jd.logger.Debugw(
+					"getting storage preferences failed",
+					"workspaceID", workspaceID,
```
atzoum, line 566: where is the error?

jobsdb/backup.go
```diff
@@ -554,15 +561,21 @@ func (jd *Handle) createTableDumps(ctx context.Context, queryFunc func(int64) st
...
 				continue
 			}
 			if err != nil {
 				return fmt.Errorf("getting storage preferences failed with error: %w for workspaceID: %s", err, workspaceID)
 			}
 			if !preferences.Backup(jd.tablePrefix) {
 				offset++
-				jd.logger.Infof("Skipping backup for workspace: %s. Preferences: %v and tablePrefix: %s", workspaceID, preferences, jd.tablePrefix)
+				jd.logger.Debugw(
+					fmt.Sprintf("Skipping backup Preferences: %v", preferences),
```
atzoum, line 576: why are we including preferences in the body and not as a field?

jobsdb/backup.go
```diff
@@ -256,7 +260,10 @@ func (jd *Handle) backupJobsTable(ctx context.Context, backupDSRange *dataSetRan
 	}()
 	err = jd.uploadDumps(ctx, dumps)
 	if err != nil {
-		jd.logger.Errorf("[JobsDB] :: Failed to upload dumps for table: %s. Error: %v", tableName, err)
+		jd.logger.Errorw(
+			"Failed to upload backup",
```
atzoum, line 264: Let's be consistent, all logs should either start with a lowecase or an uppercase character

jobsdb/backup.go
Sidddddarth: The only error possible here is workspace not being present in the server's config.
So didn't think it was important here.

atzoum: What will happen if someone changes the underlying implementation and a new error comes up? how this part of the code can be secured?
Previously I proposed to use a sentinel error for that exact purpose.

jobsdb/backup.go
```diff
@@ -137,15 +137,23 @@ func (jd *Handle) uploadDumps(ctx context.Context, dumps map[string]string) erro
 		path := filePath
 		operation := func() error {
 			if err := jd.uploadTableDump(ctx, wrkId, path); err != nil {
-				jd.logger.Errorf("[JobsDB] :: Failed to upload workspaceId %v. Error: %s", wrkId, err.Error())
+				jd.logger.Errorw(
+					"failed backup upload",
+					"workspaceID", wrkId,
+					"error", err.Error(),
```
fracasula, line 143: It warehouse we have a bunch of constants to simplify consistency and avoid possible typos: https://github.com/rudderlabs/rudder-server/blob/master/warehouse/logfield/logfield.go

It can be done here as well.

Sidddddarth: sidenote:
maybe this `warehouse/logfield` should be placed outside(as `package constants` perhaps, because stats for example use the same keys in a lot places too), because a lot of constants defined in there could be used in other places too.

I'm adding `jobsdb/internal/constants/constants.go`, but think it would better serve placed outside.

fracasula: Yes, sharing it [is being discussed](https://github.com/rudderlabs/rudder-server/pull/3731) already, just not sure how yet plus some of the constants might be domain specific (e.g. uploadID is WH specific).

Sidddddarth: > some of the constants might be domain specific

true, wouldn't be hard to group them though. Personally I'd like to have them in one place.
Anyway, for now I'll continue with `jobsdb/internal/constants`

jobsdb/internal/constants/constants.go
```diff
@@ -0,0 +1,9 @@
+package constants
+
+const (
+	TableName   = "tableName"
+	WorkspaceID = "workspaceID"
+	Err         = "error"
+	CustomVal   = "customVal"
+	TablePrefix = "tablePrefix"
+)
```
atzoum, line 9: Don't we need project-wide logfield constants? Are we supposed to be redefining the same constants in other packages too?

jobsdb/jobsdb.go
```diff
@@ -2557,7 +2558,7 @@ func (jd *Handle) refreshDSList(ctx context.Context) error {
 	start := time.Now()
 	var err error
 	defer func() {
-		stats.Default.NewTaggedStat("refresh_ds_loop", stats.TimerType, stats.Tags{"customVal": jd.tablePrefix, "error": strconv.FormatBool(err != nil)}).Since(start)
+		stats.Default.NewTaggedStat("refresh_ds_loop", stats.TimerType, stats.Tags{constants.CustomVal: jd.tablePrefix, "error": strconv.FormatBool(err != nil)}).Since(start)
```
atzoum, line 2561: Are we going to use the same set of constants for covering both log fields & stat tags?

jobsdb/backup.go
```diff
@@ -137,15 +138,30 @@ func (jd *Handle) uploadDumps(ctx context.Context, dumps map[string]string) erro
 		path := filePath
 		operation := func() error {
 			if err := jd.uploadTableDump(ctx, wrkId, path); err != nil {
-				jd.logger.Errorf("[JobsDB] :: Failed to upload workspaceId %v. Error: %s", wrkId, err.Error())
-				stats.Default.NewTaggedStat("backup_ds_failed", stats.CountType, stats.Tags{"customVal": jd.tablePrefix, "workspaceId": wrkId}).Increment()
+				jd.logger.Errorw(
+					"failed backup upload",
```
atzoum, line 142:
```suggestion
					"failed to upload backup",
```

jobsdb/backup.go
```diff
@@ -137,15 +138,30 @@ func (jd *Handle) uploadDumps(ctx context.Context, dumps map[string]string) erro
...
+					constants.WorkspaceID, wrkId,
+					constants.Err, err.Error(),
+				)
+				stats.Default.NewTaggedStat(
+					"backup_ds_failed",
+					stats.CountType,
+					stats.Tags{
+						constants.CustomVal: jd.tablePrefix,
+						"workspaceId":       wrkId,
```
atzoum, line 151: this is an example why the same constants for logs and stats might not fit the bill

jobsdb/backup.go
```diff
@@ -137,15 +138,30 @@ func (jd *Handle) uploadDumps(ctx context.Context, dumps map[string]string) erro
...
+				).Increment()
 				return err
 			}
 			return nil
 		}
 		g.Go(misc.WithBugsnag(func() error {
 			return backoff.RetryNotify(operation, backoff.WithMaxRetries(backoff.NewExponentialBackOff(), uint64(jd.config.GetInt("JobsDB.backup.maxRetries", 100))), func(err error, d time.Duration) {
-				jd.logger.Errorf("[JobsDB] :: Retrying upload workspaceId %v. Error: %s", wrkId, err.Error())
+				jd.logger.Errorw(
+					"retrying backup upload",
```
atzoum, line 161:
```suggestion
					"retrying uploading backup file",
```

jobsdb/internal/constants/constants.go
Sidddddarth: those were my thoughts too.
It's being worked in another PR(albeit not in a project-wide fashion), so this is kind of a placeholder for now.

jobsdb/backup.go
Sidddddarth: using the same everywhere would simplify our lives, though.
I'll change this to use `workspaceID` too(shouldn't cause problems because we don't use this stat as much now), the others can follow soon.

jobsdb/backup.go
Sidddddarth: going with "failed to upload jobs". The same is being used in archival.

jobsdb/jobsdb.go
Sidddddarth: I think it's convenient, might need some changes like you've pointed out in another comment.

jobsdb/backup.go
atzoum: Isn't `workspaceId` an important stat label though?
My personal opinion is that we should:
1. Not mix log fields and stat tags
2. Use 1 common set of log field constants for rudder-server (warehouse can use a separate one)
3. Use 1 common set of stat tag constants

Sidddddarth: why not mix them..? after all they're only tags - either stats or logs.
I don't see the kind of problems it could cause, please elaborate

Sidddddarth: yes
I was expressing the need to have the same in both places

jobsdb/jobsdb.go
Sidddddarth: cleared this one out too.

lvrach (approved):

atzoum (approved):

PR #3794, opened 2023-08-28
feat: onboard destination google cloud function

ujjwal-ab:
# Description
Resolves INT-135

- Onboard destination Google Cloud Function to Custom Destination Manager

## Linear Ticket

< Replace with Linear Link >

## Security


services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,215 @@
...
+		// Process the response (sample response handling here).
+		if response != nil {
+			fmt.Printf("Function call status code: %d\n", response.HTTPStatusCode)
+			// Handle response content as needed.
+		}
+		fmt.Println("Request successful!")
+
+		respStatus = "Success"
+		responseMessage = "[GoogleCloudFunction] :: Message Payload inserted with messageId :: "
+		return 200, respStatus, responseMessage
```
sanpj2292, line 115: Can we put this in a separate function ?

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,215 @@
...
+		return
+	}
+
+	if resp.Status == "OK" {
+		respStatus = "Success"
+		responseMessage = "[GoogleSheets] :: Message Payload inserted with messageId :: " + parsedJSON.Get("id").String()
+	}
+	fmt.Print(resp.Status)
+
+	return 200, respStatus, responseMessage
```
sanpj2292, line 160: Can we also put this in a separate function ?

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,215 @@
...
+	index := strings.Index(url, "cloudfunctions.net/")
+
+	if index != -1 {
+		// Extract the substring after "cloudfunctions.net/"
+		FUNCTION_NAME = url[index+len("cloudfunctions.net/"):]
+	} else {
+		fmt.Println("No match found")
+	}
+
+	functionName = "projects/" + PROJECT_ID + "/locations/" + REGION + "/functions/" + FUNCTION_NAME
```
sanpj2292, line 212: Can't we get each of these separately as config/event(dynamic config) fields ?

sanpj2292: (Or)
Is there any other easy way to get each of these values from `googleCloudFunctionUrl` field ?

services/streammanager/googlecloudfunction/googlecloudfunction.go
koladilip: move gen1 and gen2 logic create 2 function @ujjwal-ab

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,215 @@
...
+	if destConfig.FunctionEnvironment == "gen1" && destConfig.RequireAuthentication {
+		client := producer.client
+
+		if client == nil {
+			respStatus = "Failure"
+			responseMessage = "[GoogleCloudFunction] error  :: Failed to initialize GoogleCloudFunction client"
+			return 400, respStatus, responseMessage
+		}
+
+		functionName := getFunctionName(destConfig.GoogleCloudFunctionUrl)
```
koladilip, line 86: can we move this to initialisation phase as this need not be done for every event, in the init phase you can even call google api to return proper value, instead of custom parsing logic

ujjwal-ab: Actually, I wasn't able to find any API to get the functionName from the URL, that's why I added some custom logic. Also, this will be needed only in case of `gen1` function not for all.

services/streammanager/googlecloudfunction/googlecloudfunction.go
ujjwal-ab: Couldn't find a simpler way to get these from the URL. If we ask these all from config, it will be kind of repetitive.

services/streammanager/googlecloudfunction/googlecloudfunction.go
ujjwal-ab: Done

services/streammanager/googlecloudfunction/googlecloudfunction.go
ujjwal-ab: Done

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func invokeAuthenticatedGen1Functions(client *Client, functionUrl string, parsedJSON gjson.Result) (statusCode int, respStatus, responseMessage string) {
+	if client == nil {
+		respStatus = "Failure"
+		responseMessage = "[GoogleCloudFunction] error  :: Failed to initialize GoogleCloudFunction client"
+		return 400, respStatus, responseMessage
+	}
+
+	functionName := getFunctionName(functionUrl)
```
koladilip, line 94: I have commented earlier to do this in NewProducer to avoid doing this again again during the event processing.

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,226 @@
...
+	destConfig := &Config{config.Credentials, config.FunctionEnvironment, config.RequireAuthentication, config.GoogleCloudFunctionUrl}
+
+	return &GoogleCloudFunctionProducer{client, destConfig}, err
+}
+
+func (producer *GoogleCloudFunctionProducer) Produce(jsonData json.RawMessage, _ interface{}) (statusCode int, respStatus, responseMessage string) {
+	destConfig := producer.config
+	parsedJSON := gjson.ParseBytes(jsonData)
+
+	if destConfig.FunctionEnvironment == "gen1" && destConfig.RequireAuthentication {
```
koladilip, line 79: what happens if env is gen1 and RequireAuth is false?

ujjwal-ab: It will make REST API call similar to gen2 function as to call through SDK we need to create a client which requires credentials.

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,223 @@
...
+	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
+	"github.com/rudderlabs/rudder-server/services/streammanager/common"
+	"github.com/tidwall/gjson"
+)
+
+type Config struct {
+	Credentials            string `json:"credentials"`
+	FunctionEnvironment    string `json:"functionEnvironment"`
+	RequireAuthentication  bool   `json:"requireAuthentication"`
+	GoogleCloudFunctionUrl string `json:"googleCloudFunctionUrl"`
```
koladilip, line 28: Keep directly function name here instead of url

services/streammanager/googlecloudfunction/googlecloudfunction.go
ujjwal-ab: This is only being used when we are calling through in the case of gen1 functions. Should we move it to NewProducer?

services/streammanager/googlecloudfunction/googlecloudfunction.go
ujjwal-ab: Url is needed in the case of gen2 function.

services/streammanager/googlecloudfunction/googlecloudfunction.go
koladilip: That is fine, as it is done only one time compared to doing for every event

services/streammanager/googlecloudfunction/googlecloudfunction.go
koladilip: Keep both then

ujjwal-ab: Do you mean, I should take both from the user in from UI?

koladilip: Extract function using getFunctionName function and keep it here

ujjwal-ab: Done

koladilip (approved):

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,261 @@
...
+
+	var FUNCTION_NAME string
+
+	index := strings.Index(url, "cloudfunctions.net/")
+
+	if index != -1 {
+		// Extract the substring after "cloudfunctions.net/"
+		FUNCTION_NAME = url[index+len("cloudfunctions.net/"):]
+	} else {
+		fmt.Println("No match found")
```
sanpj2292, line 232: Can we use a logger here instead ?

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,261 @@
...
+
+func invokeGen2Functions(functionUrl, credentials string, requireAuthentication bool, parsedJSON gjson.Result) (statusCode int, respStatus, responseMessage string) {
+	ctx := context.Background()
+
+	jsonBytes := []byte(parsedJSON.String())
+
+	// Create a POST request
+	req, err := http.NewRequest("POST", functionUrl, strings.NewReader(string(jsonBytes)))
+	if err != nil {
+		fmt.Println("Error creating request:", err)
```
sanpj2292, line 144: Can we use a logger here instead ?

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,261 @@
...
+	fmt.Println("Request successful!")
+	// -----------
+
+	respStatus = "Success"
+	responseMessage = "[GoogleCloudFunction] :: Message Payload inserted with messageId :: " + parsedJSON.Get("id").String()
+	return 200, respStatus, responseMessage
+}
+
+func invokeGen2Functions(functionUrl, credentials string, requireAuthentication bool, parsedJSON gjson.Result) (statusCode int, respStatus, responseMessage string) {
+	ctx := context.Background()
```
sanpj2292, line 137: Is the context not being passed from `Produce` method ?

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,261 @@
...
+
+	jsonBytes := []byte(parsedJSON.String())
```
sanpj2292, line 139: I think we can get bytes directly from `gjson.Result`. Did you already check it ?

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,261 @@
...
+	if config.RequireAuthentication {
+		opts = []option.ClientOption{
+			option.WithCredentialsJSON([]byte(config.Credentials)),
+		}
+	}
+
+	service, err := generateService(opts...)
+	// If err is not nil then return
+	if err != nil {
+		pkgLogger.Errorf("[GoogleCloudFunction] error  :: %w", err)
```
sanpj2292, line 75: While logging error I don't think you'd need `[GoogleCloudFunction]`. The logger is already created with `GoogleCloudFunction` as child, so this information should be displayed already. Can you confirm it by checking the logs in your local ?

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,261 @@
...
+		pkgLogger.Errorf("[GOOGLE_CLOUD_FUNCTION] error while calling the Gen2 function :: %v", err)
+		return statusCode, respStatus, responseMessage
+	}
+	defer resp.Body.Close()
+
+	if resp.Status == "OK" {
+		respStatus = "Success"
+		responseMessage = "[GoogleCloudFunction] :: Message Payload inserted with messageId :: " + parsedJSON.Get("id").String()
+	}
+	fmt.Print(resp.Status)
```
sanpj2292, line 183: Remove this

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,261 @@
...
+	if requireAuthentication {
+		ts, err := idtoken.NewTokenSource(ctx, functionUrl, option.WithCredentialsJSON([]byte(credentials)))
+		if err != nil {
+			fmt.Print("failed to create NewTokenSource: %w", err)
+		}
+
+		// Get the ID token, to make an authenticated call to the target audience.
+		token, err := ts.Token()
+		if err != nil {
+			fmt.Print("failed to receive token: %w", err)
```
sanpj2292, line 159: I think we should handle this error in a different way. Are we going to fail the event & make sure it re-tries ? or Is there any other approach planned for this ?

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,261 @@
...
+		fmt.Println("Error creating request:", err)
+		return
+	}
+
+	// Set the appropriate headers
+	req.Header.Set("Content-Type", "application/json")
+	if requireAuthentication {
+		ts, err := idtoken.NewTokenSource(ctx, functionUrl, option.WithCredentialsJSON([]byte(credentials)))
+		if err != nil {
+			fmt.Print("failed to create NewTokenSource: %w", err)
```
sanpj2292, line 153: What is the plan around handling this ?

services/streammanager/googlecloudfunction/googlecloudfunction.go
sanpj2292: Alternate way:
1. remove the else block
2. Above `if index != -1 {`, can we add a logger which outputs the index value at debug scope `pkglogger.Debugf("SearchIndex: %d", index)`

services/streammanager/googlecloudfunction/googlecloudfunction.go
sanpj2292: Not getting passed from `Produce`, hence for now closing this

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,312 @@
...
+
+	// ---------- To be Removed
+	// Process the response (sample response handling here).
+	if response != nil {
+		fmt.Printf("Function call status code: %d\n", response.HTTPStatusCode)
+		// Handle response content as needed.
+	}
+	fmt.Println("Request successful!")
+	// -----------
+
```
sanpj2292, line 172: [Minor] Can we remove this part ?

ujjwal-ab: Removed

services/streammanager/googlecloudfunction/googlecloudfunction.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	if resp.StatusCode == http.StatusOK {
+		respStatus = "Success"
+		responseMessage = "[GoogleCloudFunction] :: Message Payload inserted with messageId :: " + parsedJSON.Get("id").String()
+	} else {
+		respStatus = "Failure"
+		responseMessage = "[GOOGLE_CLOUD_FUNCTION] error :: Function call failed " + string(responseBody)
+		pkgLogger.Error(responseMessage)
+	}
```
sanpj2292, line 219: Are we sure about getting errors only when `statusCode != 200` ?

PR #3799, opened 2023-08-28
chore: router_response_counts stat now shows if it's a retry attempt

Sidddddarth:
# Description

`router_response_counts` stat now tell if an attempt is a retry(true/false).
Checks if `attemptNum > 1`.

## Linear Ticket

[adding retry flag to `router_response_counts`](https://linear.app/rudderstack/issue/PIPE-193/adding-retry=bool-labeltag-to-router-delivery-stats)

## Security

### DIFF
--- a/router/worker.go
+++ b/router/worker.go
@@ -820,6 +820,7 @@ func (w *worker) sendRouterResponseCountStat(status *jobsdb.JobStatusT, destinat
 		"alert": strconv.FormatBool(alert),
 		// To specify at which point failure happened
 		"errorAt": errorAt,
+		"retry":   fmt.Sprint(status.AttemptNum > 1),
 	})
 	routerResponseStat.Count(1)
 }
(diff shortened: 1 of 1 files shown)


router/worker.go
```diff
@@ -820,6 +820,7 @@ func (w *worker) sendRouterResponseCountStat(status *jobsdb.JobStatusT, destinat
 		"alert": strconv.FormatBool(alert),
 		// To specify at which point failure happened
 		"errorAt": errorAt,
+		"retry":   fmt.Sprint(status.AttemptNum > 1),
```
lvrach, line 823: For consistency with the line above shall we do:
```suggestion
		"retry":   strconv.FormatBool(status.AttemptNum > 1),
```

I am not sure which is more performant, I bet `strconv.FormatBool` would be slightly.
Readability wise `FormatBool` is more explicit, and thus better in my eyes

lvrach (approved):

cisse21 (approved):

PR #3800, opened 2023-08-28
chore: cloud extract should always merge

achettyiitr:
# Description

- Cloud extract should always go for MERGE irrespective of what the loadTableStrategy is set.

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-195/cloud-sources-always-append

## Security

### DIFF
--- a/warehouse/integrations/snowflake/snowflake.go
+++ b/warehouse/integrations/snowflake/snowflake.go
@@ -59,7 +59,7 @@ var partitionKeyMap = map[string]string{
 	discardsTable:   `"ROW_ID", "COLUMN_NAME", "TABLE_NAME"`,
 }
 
-var appendSourceCategoryMap = map[string]struct{}{
+var mergeSourceCategoryMap = map[string]struct{}{
 	"cloud":           {},
 	"singer-protocol": {},
 }
@@ -820,14 +820,12 @@ func (sf *Snowflake) LoadIdentityMappingsTable(ctx context.Context) error {
 	return nil
 }
 
-// ShouldAppend returns true if the load table strategy is append mode and the source is event stream
-// currently, source category can either be (cloud, singer-protocol, warehouse, webhook)
+// ShouldAppend returns true if the load table strategy is append mode and the source category is not in "mergeSourceCategoryMap"
 func (sf *Snowflake) ShouldAppend() bool {
 	sourceCategory := sf.Warehouse.Source.SourceDefinition.Category
-	if _, isAppendCategory := appendSourceCategoryMap[sourceCategory]; isAppendCategory {
-		return false
-	}
-	return sf.config.loadTableStrategy == loadTableStrategyAppendMode
+	_, isMergeCategory := mergeSourceCategoryMap[sourceCategory]
+
+	return !isMergeCategory && sf.config.loadTableStrategy == loadTableStrategyAppendMode
 }
 
 func (sf *Snowflake) LoadUserTables(ctx context.Context) map[string]error {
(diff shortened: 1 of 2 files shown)


warehouse/integrations/snowflake/snowflake.go
```diff
@@ -815,6 +820,16 @@ func (sf *Snowflake) LoadIdentityMappingsTable(ctx context.Context) error {
 	return nil
 }
 
+// ShouldAppend returns true if the load table strategy is append mode and the source is event stream
+// currently, source category can either be (cloud, singer-protocol, warehouse, webhook)
+func (sf *Snowflake) ShouldAppend() bool {
```
fracasula, line 825: How are we exporting `ShouldAppend` here?

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -815,6 +820,16 @@ func (sf *Snowflake) LoadIdentityMappingsTable(ctx context.Context) error {
 	return nil
 }
 
+// ShouldAppend returns true if the load table strategy is append mode and the source is event stream
+// currently, source category can either be (cloud, singer-protocol, warehouse, webhook)
```
fracasula, line 824: I'm not sure I understand the comment. If source category is `warehouse` or `webhook` should be append then? They are not in the `map` though.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -815,6 +820,16 @@ func (sf *Snowflake) LoadIdentityMappingsTable(ctx context.Context) error {
...
+func (sf *Snowflake) ShouldAppend() bool {
+	sourceCategory := sf.Warehouse.Source.SourceDefinition.Category
+	if _, isAppendCategory := appendSourceCategoryMap[sourceCategory]; isAppendCategory {
```
fracasula, line 827: If it is an append category then `isAppendCategory` is `true`, right? Then why do we return `false`?

I was expecting something like:

```go
func (sf *Snowflake) ShouldAppend() bool {
	sourceCategory := sf.Warehouse.Source.SourceDefinition.Category
	isAppendCategory := appendSourceCategoryMap[sourceCategory]

	return isAppendCategory && sf.config.loadTableStrategy == loadTableStrategyAppendMode
```

This would contradict what we're trying to do though which is to always `MERGE` if the source category is "cloud extract". So maybe the global `map` should be renamed from `appendSourceCategoryMap` to `mergeSourceCategoryMap` and if the warehouse source matches a key in that map we **do not append**. Correct?

warehouse/integrations/snowflake/snowflake.go
achettyiitr: Sorry, I wanted an inverse map. I will update the comment.

warehouse/integrations/snowflake/snowflake.go
achettyiitr: Sorry, I created an inverse map but failed to update the variable.

warehouse/integrations/snowflake/snowflake.go
fracasula: This is a code smell, we'll likely have to address it later.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -815,6 +820,14 @@ func (sf *Snowflake) LoadIdentityMappingsTable(ctx context.Context) error {
 	return nil
 }
 
+// ShouldAppend returns true if the load table strategy is append mode and the source category is not cloud extract
```
fracasula, line 823:
```suggestion
// ShouldAppend returns true if the load table strategy is append mode and the source category is not in "mergeSourceCategoryMap"
```

fracasula (approved):

achettyiitr:
> This is a code smell, we'll likely have to address it later.

I have a task for this. Currently, everything is too coupled with the warehouse.

lvrach (approved):

PR #3784, opened 2023-08-24
chore: cleanup pg notifier

achettyiitr:
# Description

- Introducing **repository** and **model** for Notifier.
- Renamed it to `notifier` instead of `pgNotifier` since we don't need to expose the implementation details like it is implemented for Postgres.
- Removing globals.
- Removed `notifier.Init()`.
- Addition of `claimLag` and `trackBatchLag` for monitoring purposes.
  - `claimLag` to track between the time when the record was created and the time when it was claimed.
  - `trackBatch` to track the number of track batches running.
  - `trackBatchLag` to track the time and how much is being spent on tracking per publish.
  - `orphanJobs` to observe the number of orphan jobs by maintenance.
- Introduced `background errgroup` to **gracefully shut down the goroutines** being created during **track batch** and **subscribe**.
- Added a separate table called `**pg_notifier_queue_metadata**` for **storing common metadata** instead of sending them in each payload as a separate transaction in an **attempt to make the Notifier Generic**. It will help us create **bigger batches**, and **less tracking** or **polling** for batches to complete. 
- Started using the **topic** to make it **version-compatible while claiming and publishing**. The major use case is, that when the release happens, slaves come first and they start processing jobs and after a while when the master is up, it deletes them. So, it's better to have the compatibility.
- **Moving maintenance workers to Master instead of Slave.** Earlier for the Multi-tenant setup, we used to have many rudder server nodes running in embedded mode and share some shared slaves pool. Now warehouse master is kind of Multi-tenant.

# Discussion
- Should we get rid of maintenance workers by getting a database lock?
- Instead of a claim, can we probably have a `trigger` that updates when the status is marked as failed or waiting?

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-169/notifier-cleanup

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #3805, opened 2023-08-29
chore: uploader can append

fracasula:
# Description

We can append if:
* source is not an ETL source
* source it's not a replay
* source category is not in a specific set
* we are in append mode

### Additional changes
* removed duplicated mocked uploader in favour of MockGen

### TODOs
* use `Uploader.CanAppend()` in the other integrations that support `APPEND` mode

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-104/always-do-merge-if-source-type-is-replayetl-sources-%5Bwarehouse%5D) >

## Security


fracasula:
Blocked by https://github.com/rudderlabs/rudder-go-kit/pull/108

warehouse/utils/utils.go
```diff
@@ -230,6 +230,7 @@ type KeyValue struct {
 	Value interface{}
 }
 
+//go:generate mockgen -destination=../internal/mocks/utils/mock_uploader.go -package mock_uploader github.com/rudderlabs/rudder-server/warehouse/utils Uploader
```
achettyiitr, line 233: Can we move this in a separate package?

fracasula: I had created already a dedicated package for it, namely `warehouse/internal/mocks/utils/mock_uploader.go`. I was planning to have other mocks there as well. What's wrong with it?

fracasula: So what's your suggestion? What's the exact package name that you'd create for hosting this mock?

achettyiitr: The mock package name is fine, I was simply asking if we can move this interface to a separate package `warehouse/uploader`?

services/pgnotifier/pgnotifier.go
```diff
@@ -108,10 +108,10 @@ func loadPGNotifierConfig() {
 	pgNotifierDBPort = config.GetInt("PGNOTIFIER_DB_PORT", 5432)
 	pgNotifierDBPassword = config.GetString("PGNOTIFIER_DB_PASSWORD", "ubuntu") // Reading secrets from
 	pgNotifierDBSSLMode = config.GetString("PGNOTIFIER_DB_SSL_MODE", "disable")
-	config.RegisterIntConfigVariable(3, &maxAttempt, false, 1, "PgNotifier.maxAttempt")
+	config.RegisterIntVar(3, &maxAttempt, 1, "PgNotifier.maxAttempt")
 	trackBatchInterval = time.Duration(config.GetInt("PgNotifier.trackBatchIntervalInS", 2)) * time.Second
-	config.RegisterDurationConfigVariable(5000, &maxPollSleep, true, time.Millisecond, "PgNotifier.maxPollSleep")
-	config.RegisterDurationConfigVariable(120, &jobOrphanTimeout, true, time.Second, "PgNotifier.jobOrphanTimeout")
+	config.RegisterAtomicDurationVar(5000, &maxPollSleep, time.Millisecond, "PgNotifier.maxPollSleep")
+	config.RegisterAtomicDurationVar(120, &jobOrphanTimeout, time.Second, "PgNotifier.jobOrphanTimeout")
```
achettyiitr, line 114: Good catch.

achettyiitr: Did we get this while running the test with `-race`?

achettyiitr (approved):

services/pgnotifier/pgnotifier.go
achettyiitr: This has already been take care of with the new changes. These globals have been [removed](https://github.com/rudderlabs/rudder-server/pull/3784/files#diff-81c88462ea98fe7900948848a9b8e201fd33f6efdf9c6770de3cbd64e657bab9R64-R65).

fracasula: Yes the `-race` highlighted it :+1: I'll try to have a look at that PR too then so maybe we can speed this up. With the atomic change I'm consistently getting failing tests now.

warehouse/utils/utils.go
fracasula: Right, let's move it as part of another cleanup PR where we refactor this a bit. There are several things we want to address, splitting the interface in a more SOLID approach, move the derivated interfaces in appropriate packages and remove the circular references. We can do it all together as a separate change, wdyt?

achettyiitr: Sure.

services/pgnotifier/pgnotifier.go
fracasula: > This has already been take care of with the new changes. These globals have been [removed](https://github.com/rudderlabs/rudder-server/pull/3784/files#diff-81c88462ea98fe7900948848a9b8e201fd33f6efdf9c6770de3cbd64e657bab9R64-R65).

@achettyiitr the race can still happen in your branch as well. if the variable is hot-reloaded while we read then you get a race. it's just going to be harder to see this race in tests because the variable is not global anymore.

achettyiitr: Understood. Hot reloadable then need to concurrent safe then

fracasula:
@lvrach this one can potentially be merged if we temporarily remove the `-race` from the `Makefile`. I still need another approval though.

PR #3814, opened 2023-08-31
chore: grpc cleanup

achettyiitr:
# Description

- Removed globals.

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-189/grpc-cleanup

## Security


warehouse/internal/model/upload.go
```diff
@@ -36,7 +36,10 @@ const (
 	Noop                      JobErrorType = "noop"
 )
 
-var ErrUploadNotFound = errors.New("upload not found")
+var (
+	ErrUploadNotFound = errors.New("upload not found")
+	ErrNoUploadsFound = errors.New("no uploads found")
```
lvrach, line 41: [question] Isn't this the same error? Why do we need to have both ?

warehouse/internal/repo/upload.go
```diff
@@ -680,3 +687,362 @@ func (uploads *Uploads) LastCreatedAt(ctx context.Context, sourceID, destination
...
+	return syncUploadInfos, totalUploads, nil
+}
+
+func (uploads *Uploads) syncsInfo(ctx context.Context, limit, offset int, opts model.SyncUploadOptions, countOver bool) ([]model.UploadInfo, int64, error) {
+	filterQuery, filterArgs := syncUploadQueryArgs(&opts)
+
+	countOverClause := "0 AS total_uploads"
+	if countOver {
+		countOverClause = "COUNT(*) OVER() AS total_uploads"
+	}
```
lvrach, line 729: [question] This part of logic I don't understand, why we return `total_uploads: COUNT(*) OVER()` for multitenant and zero for non-mulitenant ?

lvrach (approved):
No blocking comment, just some questions

warehouse/internal/repo/upload.go
achettyiitr: For non-multi-tenant, we are running a count query instead of embedding it in the same query.
https://github.com/rudderlabs/rudder-server/pull/3814/files#diff-dc7a971c99be83650f101872e58fda0aac0b14297cc4d6376485ed2fec63c8e9R715-R718

achettyiitr: Dhawal made this improvement when the queries to hostedmted for count(*) were taking time. So he embedded `COUNT(*) OVER() AS total_uploads` in the same query to get the total count.

warehouse/internal/model/upload.go
achettyiitr: I thought `ErrUploadNotFound` was only for error when we looked for a particular upload. In my case, I was looking for uploads for a particular source and destination.

achettyiitr: Should I use sql.ErrNoRows instead of sentinal error?

warehouse/internal/repo/upload.go
achettyiitr: I can probably benchmark the query after this PR and can have a single query for both. wdyt?

fracasula (approved):

PR #3820, opened 2023-09-01
chore: use single query for getting jobs

atzoum:
# Description

Router iterator, batchrouter & stash can now perform a single SQL query for retrieving jobs instead of multiple.
The above behaviour is toggleable through a feature flag, `JobsDB.useSingleGetJobsQuery`. If the flag is disabled, rudder-server falls back to its previous behaviour.

## Querying

- For unprocessed jobs we are using a left join with the status table
- For all other job states we are using an inner join with the latest status view
- When the query needs to query both types of states we are using a left join with the latest status view

## Caching

- The no jobs cache is checked for all requested job states separately. e.g. if you ask for failed and unprocessed and an entry exists in the cache for failed, the final query will only contain predicates for unprocessed jobs.
- The no jobs cache is updated for separate job states as well, e.g.
  - If no jobs are returned by the query, entries will be added in the cache for all states
  - If only unprocessed jobs are returned for a request issued about both failed and unprocessed, an entry will be added in the no jobs cache for failed state.

## Linear Ticket

[PIPE-219](https://linear.app/rudderstack/issue/PIPE-219/support-performing-a-single-get-jobs-query-toggleable-behind-a-feature)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -1828,7 +1828,10 @@ type moreToken struct {
 	afterJobID *int64
 }
 
-func (jd *Handle) GetToProcess(ctx context.Context, params GetQueryParams, more MoreToken) (*MoreJobsResult, error) { // skipcq: CRT-P0003
+// getToProcessLegacy returns jobs that are in failed, waiting and unprocessed states using three separate queries
+//
+// Deprecated: shall be removed after successful rollout
+func (jd *Handle) getToProcessLegacy(ctx context.Context, params GetQueryParams, more MoreToken) (*MoreJobsResult, error) { // skipcq: CRT-P0003
 
 	mtoken := &moreTokenLegacy{}
 	if more != nil {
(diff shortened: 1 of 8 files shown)


lvrach (approved):

jobsdb/jobsdb.go
```diff
@@ -1890,6 +1893,30 @@ func (jd *Handle) GetToProcess(ctx context.Context, params GetQueryParams, more
...
+
+	if !jd.config.GetBool("JobsDB.useSingleGetJobsQuery", true) { // TODO: remove condition after successful rollout of sinle query
+		return jd.getToProcessLegacy(ctx, params, more)
+	}
+
+	if params.JobsLimit == 0 {
+		return &MoreJobsResult{More: more}, nil
+	}
+	params.stateFilters = []string{Failed.State, Waiting.State, Unprocessed.State}
+	slices.Sort(params.stateFilters)
```
fracasula, line 1906: Aren't we creating the slice just one line above? Can't we create it in a sorted manner already? Also, do we need it to be sorted?

atzoum: Instead of a comment (e.g. this slice needs to be lexicographically sorted), I prefer to use the `slices.Sort` operation explicitly. 
If this slice needs to be changed in the future, relying on a person to make sure the new slice will be sorted is riskier than using `slices.Sort`

Sidddddarth (approved):

fracasula (approved):

cisse21 (approved):

PR #3835, opened 2023-09-05
refactor: apk usage in Dockerfile

PeterDaveHello:
Less commands, faster build, and smaller image!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,9 +31,8 @@ RUN go build -o rudder-cli ./cmd/rudder-cli/
 
 FROM alpine:${ALPINE_VERSION}
 
-RUN apk update && apk add tzdata
-RUN apk -U --no-cache upgrade && \
-    apk add --no-cache ca-certificates postgresql-client curl bash
+RUN apk --no-cache upgrade && \
+    apk --no-cache add tzdata ca-certificates postgresql-client curl bash
 
 COPY --from=builder rudder-server/rudder-server .
 COPY --from=builder rudder-server/build/wait-for-go/wait-for-go .


achettyiitr (approved):

PeterDaveHello:
@achettyiitr anything I can do 😅 ?

Dockerfile
```diff
@@ -31,9 +31,8 @@ RUN go build -o rudder-cli ./cmd/rudder-cli/
 
 FROM alpine:${ALPINE_VERSION}
 
-RUN apk update && apk add tzdata
-RUN apk -U --no-cache upgrade && \
-    apk add --no-cache ca-certificates postgresql-client curl bash
```
achettyiitr, line 36: Any reason, we are skipping update here?

achettyiitr (commented):
Any reason, we are skipping the update here?

Dockerfile
```diff
@@ -31,9 +31,8 @@ RUN go build -o rudder-cli ./cmd/rudder-cli/
...
+RUN apk --no-cache upgrade && \
```
atzoum, line 34: It is a good idea to always do an update right before doing an upgrade or add command. That way the command will install the latest available packages. [link](https://wiki.alpinelinux.org/wiki/Alpine_Package_Keeper#:~:text=Tip%3A%20With%20remote%20repositories%2C%20it%20is%20a%20good%20idea%20to%20always%20do%20an%20update%20right%20before%20doing%20an%20upgrade%20or%20add%20command.%20That%20way%20the%20command%20will%20install%20the%20latest%20available%20packages.)
```suggestion
RUN apk --no-cache update && apk --no-cache upgrade && \
```

PeterDaveHello: Not exactly, `--no-cache` will handle that for you, and leave no temporary files.

PeterDaveHello:
@achettyiitr @atzoum just FYR, it's kind of Dockerfile best practices:

https://github.com/hadolint/hadolint/wiki/DL3019

PeterDaveHello:
Not sure if there's anything I can do here?

Dockerfile
atzoum: the request was to run `apk --no-cache update` before `apk --no-cache upgrade`, not sure how your response is relevant

PeterDaveHello: `--no-cache` actually will make `apk` ignore any local cache and fetch the latest index without local temp files needed to be cleaned up

atzoum (approved):

atzoum:
Thank you @PeterDaveHello, have you performed this step?
https://github.com/rudderlabs/rudder-server/blob/master/CONTRIBUTING.md#rudderstack-contributor-agreement

PeterDaveHello:
It's done now!

lvrach:
Thanks @PeterDaveHello for the contribution

PR #3836, opened 2023-09-06
chore: add stats for archiver

atzoum:
# Description

Adding stats for monitoring archiver

## Linear Ticket

[PIPE-229](https://linear.app/rudderstack/issue/PIPE-229/add-stats-for-archiver)

## Security

### DIFF
--- a/archiver/archiver.go
+++ b/archiver/archiver.go
@@ -21,7 +21,7 @@ type archiver struct {
 	jobsDB          jobsdb.JobsDB
 	storageProvider fileuploader.Provider
 	log             logger.Logger
-	statHandle      stats.Stats
+	stats           stats.Stats
 
 	archiveTrigger           func() <-chan time.Time
 	adaptivePayloadLimitFunc payload.AdaptiveLimiterFunc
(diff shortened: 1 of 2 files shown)


cisse21 (approved):

archiver/archiver.go
```diff
@@ -162,14 +163,17 @@ func (a *archiver) Start() error {
 		// pinger loop
 		for {
 			if a.config.enabled() {
+				start := time.Now()
 				sources, err := a.jobsDB.GetDistinctParameterValues(ctx, "source_id")
+				a.stats.NewStat("arc_active_partitions_time", stats.TimerType).Since(start)
```
achettyiitr, line 168: We can prepare this stat once as part of the setup? The same goes with the below one as well.

achettyiitr (approved):

Sidddddarth (approved):

archiver/archiver.go
atzoum: Caching previously-created stats is already taken care of for us by the stats package, I prefer not having yet-another-field

PR #3850, opened 2023-09-08
fix: error while cleaning up old jobs: sql: no rows in result set

atzoum:
# Description

Ignoring `sql. ErrNoRows` while deleting journal entries

## Linear Ticket

[PIPE-238](https://linear.app/rudderstack/issue/PIPE-238/error-while-cleaning-up-old-jobs-sql-no-rows-in-result-set)

## Security

### DIFF
--- a/jobsdb/admin.go
+++ b/jobsdb/admin.go
@@ -2,6 +2,8 @@ package jobsdb
 
 import (
 	"context"
+	"database/sql"
+	"errors"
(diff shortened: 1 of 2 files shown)


lvrach:
Should we hotfix ?

lvrach (approved):

lvrach (approved):

PR #3851, opened 2023-09-08
chore: introduce load table total rows

achettyiitr:
# Description

- Removed `GetTotalCountInTable` from integrations package.
- Added `LoadTableStats` to calculate the **rows inserted and updated** while loading the table.
- Centralized `dedup_rows` instead of calling from all destinations.
- Avoid logging SQL queries in Info logs. Currently, we are polluting logs so much in the warehouse. If we still need Queries to be logged, we can set a `SlowQueryThreshold` to a minimal value.
- Added UNIT Test for LoadTable.

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-209/investigate-why-total-count-does-not-honour-timeout-for-30-seconds
- https://linear.app/rudderstack/issue/PIPE-202/avoid-using-count-for-getting-total-count-where-ever-needed

## Security


warehouse/integrations/snowflake/snowflake.go
```diff
@@ -604,26 +627,52 @@ func (sf *Snowflake) copyInto(
...
+		count, err := strconv.Atoi(countString)
+		if err != nil {
+			return 0, fmt.Errorf("converting rows loaded: %w", err)
+		}
+		rowsInserted += int64(count)
+	}
+	if err := rows.Err(); err != nil {
+		return 0, fmt.Errorf("iterating over rows: %w", err)
+	}
+	return rowsInserted, nil
```
achettyiitr, line 675: Didn't find a better way of doing this?

warehouse/types/types.go
```diff
@@ -0,0 +1,6 @@
+package types
+
+type LoadTableStats struct {
```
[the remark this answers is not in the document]
achettyiitr: Good point. Moving it to integrations package.

warehouse/integrations/azure-synapse/azure-synapse.go
```diff
@@ -218,232 +222,394 @@ func (*AzureSynapse) IsEmpty(context.Context, model.Warehouse) (empty bool, err
...
-	for _, column := range previousColumnKeys {
-		if !slices.Contains(sortedColumnKeys, column) {
-			extraColumns = append(extraColumns, column)
-		}
-	}
 	fileNames, err := as.LoadFileDownLoader.Download(ctx, tableName)
-	defer misc.RemoveFilePaths(fileNames...)
+	defer func() {
+		misc.RemoveFilePaths(fileNames...)
+	}()
```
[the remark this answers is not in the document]
achettyiitr: Thanks for bringing this up.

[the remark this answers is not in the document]
achettyiitr: > Then looking at the internals it seems like Download is not cleaning up after itself. If there has been an error, all files that have been downloaded for nothing (and are not returned apparently) could be cleaned up by the downloader itself.

That is true, in case of an error it does not delete the files.

warehouse/integrations/azure-synapse/azure-synapse.go
```diff
@@ -218,232 +222,394 @@ func (*AzureSynapse) IsEmpty(context.Context, model.Warehouse) (empty bool, err
...
+		return nil, "", fmt.Errorf("commit transaction: %w", err)
 	}
-	// deduplication process
+
+	log.Infow("completed loading")
+
+	return &types.LoadTableStats{
+		RowsInserted: rowsInserted - rowsDeleted,
+		RowsUpdated:  rowsDeleted,
+	}, stagingTableName, nil
```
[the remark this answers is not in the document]
achettyiitr: Good point. Let me add a test for this.

warehouse/integrations/azure-synapse/azure-synapse.go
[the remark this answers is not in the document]
achettyiitr: I will move the defer block after the error. No point in keeping up before error check.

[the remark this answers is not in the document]
achettyiitr: Have a separate task and PR for this?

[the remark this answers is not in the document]
achettyiitr: Tracking it [here](https://linear.app/rudderstack/issue/PIPE-286/integrations-cleanup-download-task).

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -603,63 +643,52 @@ func (sf *Snowflake) copyInto(
...
-		return fmt.Errorf("copy into table: %w", err)
+	columns, err := rows.Columns()
+	if err != nil {
+		return 0, fmt.Errorf("getting columns: %w", err)
+	}
+	_, index, found := lo.FindIndexOf(columns, func(item string) bool {
+		return item == "rows_loaded"
+	})
+	if !found {
+		return 0, nil
```
fracasula, line 664: `if !found` we should at least get a warning, right? We expect it to be there always otherwise there is some kind of regression with the sdk, right? Also, a test here would be nice too.

achettyiitr: Sure.

achettyiitr: I am also exploring [COPY_HISTORY](https://docs.snowflake.com/en/sql-reference/functions/copy_history) for this as well.

achettyiitr: Added a warn log.

achettyiitr: We can't use [COPY_HISTORY](https://docs.snowflake.com/en/sql-reference/functions/copy_history) since parallel ingestion can also happen at the same time when Rudderstack is loading.

achettyiitr: Going with the current approach.

warehouse/integrations/deltalake/deltalake_test.go
```diff
@@ -379,6 +384,649 @@ func TestIntegration(t *testing.T) {
 			})
 		}
 	})
+
+	t.Run("Load Table", func(t *testing.T) {
```
achettyiitr, line 388: Keeping this as part of Integration tests instead of Unit Tests, since starting for the cluster/SQL warehouse takes around 10-15 mins.

warehouse/integrations/clickhouse/clickhouse.go
```diff
@@ -1032,9 +1034,25 @@ func (ch *Clickhouse) LoadUserTables(ctx context.Context) (errorMap map[string]e
...
+
+	err = ch.loadTable(ctx, tableName, ch.Uploader.GetTableSchemaInUpload(tableName))
+	if err != nil {
+		return nil, fmt.Errorf("loading table: %w", err)
+	}
+
+	postLoadTableCount, err := ch.totalCountIntable(ctx, tableName)
+	if err != nil {
+		return nil, fmt.Errorf("post load table count: %w", err)
+	}
```
achettyiitr, line 1051: We can probably have flags later to disable it in case it's taking too long to execute. 
Currently, checked for existing customers and don't see any timeout error logs corresponding to `total count in table`.

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -793,41 +915,36 @@ func (bq *BigQuery) dropDanglingStagingTables(ctx context.Context) bool {
...
-		}
-		if !exists {
+		if exists, err := bq.tableExists(ctx, tableName); err != nil {
+			return false, fmt.Errorf("checking if table %s exists: %v", tableName, err)
+		} else if !exists {
 			continue
 		}
-		count, err := bq.GetTotalCountInTable(ctx, tableName)
+
+		metadata, err := bq.db.Dataset(bq.namespace).Table(tableName).Metadata(ctx)
```
fracasula, line 941: This line doesn't seem to have coverage (many others don't but I think it would make sense to have it here). wdyt?

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -793,41 +915,36 @@ func (bq *BigQuery) dropDanglingStagingTables(ctx context.Context) bool {
...
 		if err != nil {
-			return empty, err
-		}
-		if count > 0 {
-			empty = false
-			return empty, nil
+			return false, fmt.Errorf("getting metadata for table %s: %v", tableName, err)
 		}
+		return metadata.NumRows == 0, nil
```
fracasula, line 945: This is also not tested. I suggest you have a look at the codecov reports yourself and double check to verify that at least the main paths are tested.

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -793,41 +915,36 @@ func (bq *BigQuery) dropDanglingStagingTables(ctx context.Context) bool {
...
 	}
-	return
+	return true, nil
```
fracasula, line 947: Also not under test. I guess we end up here if none of the `[]tables` exist.

warehouse/integrations/bigquery/testhelper/setup.go
```diff
@@ -34,3 +44,35 @@ func IsBQTestCredentialsAvailable() bool {
...
+	t.Helper()
+
+	it, err := db.Query(query).Read(context.Background())
+	require.NoError(t, err)
+
+	var records [][]string
+	for {
+		var row []bigquery.Value
+		err = it.Next(&row)
+		if err == iterator.Done {
```
fracasula, line 62: I think by default it would make sense to avoid error comparisons in favour of `errors.Is` even when using a client directly. It would save us the trouble of forgetting to refactor this in case we were to use a wrapper or a different client that wraps the error. wdyt?

warehouse/integrations/bigquery/testhelper/setup.go
```diff
@@ -34,3 +44,35 @@ func IsBQTestCredentialsAvailable() bool {
 	_, err := GetBQTestCredentials()
 	return err == nil
 }
+
+func RecordsFromWarehouse(
```
fracasula, line 48: What is this function actually doing? It's hard to tell without reading its code due to its generic name and lack of comment/docs. We should be able to make it more descriptive.

warehouse/integrations/mssql/mssql.go
```diff
@@ -255,227 +253,344 @@ func (ms *MSSQL) DeleteBy(ctx context.Context, tableNames []string, params wareh
...
+			value = value[:mssqlStringLengthLimit]
+		}
+		if !hasDiacritics(value) {
+			return value, nil
+		} else {
+			byteArr := str2ucs2(value)
+			if len(byteArr) > mssqlStringLengthLimit {
+				byteArr = byteArr[:mssqlStringLengthLimit]
+			}
+			return byteArr, nil
```
fracasula, line 497: This bit is also not tested.

warehouse/integrations/postgres/load.go
```diff
@@ -18,172 +21,231 @@ import (
...
+		logfield.LoadTableStrategy, pg.loadTableStrategy(),
 	)
+	log.Infow("started loading")
 
-	return nil
-}
+	log.Infow("setting search path")
+	searchPathStmt := fmt.Sprintf(`
+		SET search_path TO %q;
+`,
```
fracasula, line 88: [optional] Can't you format it without inserting more bytes into the query that we'll send over with the SQL client?

```suggestion
	searchPathStmt := fmt.Sprintf(`SET search_path TO %q;`,
```

warehouse/integrations/postgres/testdata/docker-compose.postgres.yml
```diff
@@ -0,0 +1,15 @@
+version: "3.9"
+
+services:
+  postgres:
+    image: postgres:15-alpine
```
fracasula, line 5: @lvrach shouldn't we mention in [our docs](https://www.rudderstack.com/docs/destinations/warehouse-destinations/postgresql/) a minimum version that we're compatible with? :thinking:

warehouse/integrations/testhelper/setup.go
```diff
@@ -242,3 +250,166 @@ func EnhanceWithDefaultEnvs(t testing.TB) {
...
+	uploadOutput, err := fm.Upload(
+		context.Background(), f, loadObjectFolder,
+		tableName, sourceID, uuid.New().String()+"-"+tableName,
+	)
+	require.NoError(t, err)
+
+	return uploadOutput
+}
+
+func RecordsFromWarehouse(
```
fracasula, line 278: Is this function the same as the other helper?

warehouse/integrations/testhelper/setup.go
```diff
@@ -242,3 +250,166 @@ func EnhanceWithDefaultEnvs(t testing.TB) {
...
+				return item.Format(time.RFC3339)
+			default:
+				return cast.ToString(item)
+			}
+		}))
+	}
+	return records
+}
+
+func LoadRecords() [][]string {
```
fracasula, line 316: Is this loading any records though? It's just returning a matrix of slices.

fracasula: Same comment goes for the other functions.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -604,62 +632,58 @@ func (sf *Snowflake) copyInto(
...
 
-	return nil
-}
+	_, index, found := lo.FindIndexOf(columns, func(item string) bool {
+		return strings.ToLower(item) == "rows_loaded"
+	})
+	if !found {
+		sf.logger.Warnw("rows_loaded column not found in copy command result", "columns", columns)
+		return &types.LoadTableStats{}, nil
+	}
```
fracasula, line 654: Not under test.

warehouse/integrations/bigquery/bigquery.go
achettyiitr: Thanks for pointing this out. I added test coverage considering only Load Table in mind.

achettyiitr: let me re-verify and add coverage.

warehouse/integrations/bigquery/testhelper/setup.go
achettyiitr: Yes.

warehouse/integrations/bigquery/bigquery.go
achettyiitr: Yes.

warehouse/integrations/testhelper/setup.go
achettyiitr: Renamed to RetrieveRecordsFromWarehouse with the comment.

warehouse/integrations/testhelper/setup.go
achettyiitr: These are helpers for the data we loaded in different scenarios. Update the function name and the added comment as well.

achettyiitr:
@fracasula Thanks for pointing out missing codecov scenarios.

fracasula:
@achettyiitr you got some conflicts on `warehouse/schema.go`.

fracasula (approved):

warehouse/integrations/postgres/testdata/docker-compose.postgres.yml
achettyiitr: Good point. Currently, I can't find anything in [docs](https://www.rudderstack.com/docs/resources/rudderstack-architecture/#database) or [Readme](https://github.com/rudderlabs/rudder-server/blob/master/README.md) mentioning it.

warehouse/upload.go
```diff
@@ -1122,34 +1068,25 @@ func (job *UploadJob) loadTable(tName string) (bool, error) {
...
-		if errEventCount != nil {
-			return
-		}
+	tableUpload, errEventCount := job.tableUploadsRepo.GetByUploadIDAndTableName(job.ctx, job.upload.ID, tName)
+	if errEventCount != nil {
+		return alteredSchema, fmt.Errorf("get table upload: %w", errEventCount)
+	}
 
-		// TODO : Perform the comparison here in the codebase
-		job.guageStat(`pre_load_table_rows`, whutils.Tag{Name: "tableName", Value: strings.ToLower(tName)}).Gauge(int(totalBeforeLoad))
```
achettyiitr, line 1149: pre_load_table_rows we are not using in any way right now, in Grafana or Alerts.

warehouse/integrations/azure-synapse/azure-synapse.go
```diff
@@ -195,255 +206,379 @@ func (as *AzureSynapse) connectionCredentials() *credentials {
...
-				} else {
-					recordInterface = append(recordInterface, value)
-				}
+	csvReader := csv.NewReader(gzipReader)
+
+	for {
+		record, err := csvReader.Read()
+		if err != nil {
+			if err == io.EOF {
+				break
```
lvrach, line 387: [nit-pick]
```suggestion
		if err == EOF {
		   break
		} else if err == io.EOF {

```

warehouse/integrations/postgres/testdata/docker-compose.postgres.yml
lvrach: what's the minimum version we support? 

For Postgres as rudder-server dependency, we already have a check https://github.com/rudderlabs/rudder-server/blob/a6f205db9f87fc02f5c5290ec30bfacd216f59bf/services/validators/envValidator.go#L193 it is 10 which make sense since is no longer support by postgres https://endoflife.date/postgresql

I believe we should have 11 as our oldest supported version and also check this during validation

lvrach (approved):

warehouse/integrations/postgres/testdata/docker-compose.postgres.yml
achettyiitr: @lvrach Are you referring to destination validation here? During the validation for `Postgres` destination, we also check the version.

achettyiitr: We are currently not sure what min postgres we support as a destination.

warehouse/integrations/azure-synapse/azure-synapse.go
achettyiitr: Going with this:

```GO
		record, err := csvReader.Read()
		if err != nil {
			if errors.Is(err, io.EOF) {
				break
			}
			return fmt.Errorf("reading record from: %w", err)
		}
```

warehouse/integrations/postgres/testdata/docker-compose.postgres.yml
fracasula: > We are currently not sure what min postgres we support as a destination.

:point_up: 

Also, might make sense to have our Warehouse integration test run twice, one for the minimum supported version and one for the latest.

warehouse/integrations/azure-synapse/azure-synapse.go
```diff
@@ -218,232 +222,394 @@ func (*AzureSynapse) IsEmpty(context.Context, model.Warehouse) (empty bool, err
...
-		}
-		gzipReader.Close()
-		gzipFile.Close()
+	}
+	if _, err = stmt.ExecContext(ctx); err != nil {
+		return nil, "", fmt.Errorf("executing copyIn statement: %w", err)
 	}
 
-	_, err = stmt.ExecContext(ctx)
+	log.Infow("deleting from load table")
```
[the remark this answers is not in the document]
achettyiitr: I feel we should have this since it's a critical stage in our loading process. Currently, making these as debug logs. If we feel that some are still polluting it, we can change those later. wdyt?

```
log.Debugw("creating staging table")
log.Debugw("creating prepared stmt for loading data")
log.Debugw("committing transaction")
log.Debugw("waiting for load/append/merge job to complete", "jobID", job.ID())
log.Debugw("job statistics")
log.Debugw("setting search path")
log.Debugw("generated manifest", "manifestLocation", manifestLocation)
```

warehouse/integrations/postgres/testdata/docker-compose.postgres.yml
achettyiitr: Corresponding task. https://linear.app/rudderstack/issue/PIPE-378/check-minimum-version-during-warehouse-validations

achettyiitr: Also, should we restrict this during the loading process as well? Maybe we can have a preliminary check with the existing customers, If everybody passes, we can enforce it else we need some action there.

PR #3852, opened 2023-09-11
chore: add rsources stats for dropped events at processor

Sidddddarth:
# Description

- All events failed during the processor's pipeline (failed, filtered) are treated as dropped with respect to rudder-sources stats.
- For dropped events we are not capturing failed records

## Linear Ticket

[Ensure drop events are marked as failed in source [job-status endpoint]](https://linear.app/rudderstack/issue/PIPE-177/ensure-drop-events-are-marked-as-failed-in-source-[job-status)

## Security

### DIFF
--- a/processor/events_response.go
+++ b/processor/events_response.go
@@ -1,0 +1,75 @@
+package processor
+
+import (
+	"time"
+
+	"github.com/rudderlabs/rudder-server/jobsdb"
+	"github.com/rudderlabs/rudder-server/processor/transformer"
+	"github.com/rudderlabs/rudder-server/utils/misc"
+	"github.com/samber/lo"
+)
+
+func (proc *Handle) getDroppedJobs(response transformer.Response, eventsToTransform []transformer.TransformerEvent) []*jobsdb.JobT {
+	// each messageID is one event when sending to the transformer
+	inputMessageIDs := lo.Map(eventsToTransform, func(e transformer.TransformerEvent, _ int) string {
+		return e.Metadata.MessageID
+	})
+
+	// in transformer response, multiple messageIDs could be batched together
+	successFullMessageIDs := make([]string, 0)
+	lo.ForEach(response.Events, func(e transformer.TransformerResponse, _ int) {
+		if len(e.Metadata.MessageIDs) > 0 {
+			successFullMessageIDs = append(successFullMessageIDs, e.Metadata.MessageIDs...)
+		} else {
+			successFullMessageIDs = append(successFullMessageIDs, e.Metadata.MessageID)
+		}
+	})
+	// for failed as well
+	failedMessageIDs := make([]string, 0)
+	lo.ForEach(response.FailedEvents, func(e transformer.TransformerResponse, _ int) {
+		if len(e.Metadata.MessageIDs) > 0 {
+			failedMessageIDs = append(failedMessageIDs, e.Metadata.MessageIDs...)
+		} else {
+			failedMessageIDs = append(failedMessageIDs, e.Metadata.MessageID)
+		}
+	})
+	// the remainder of the messageIDs are those that are dropped
+	// we get jobs for those dropped messageIDs - for rsources_stats_collector
+	droppedMessageIDs, _ := lo.Difference(inputMessageIDs, append(successFullMessageIDs, failedMessageIDs...))
+	droppedMessageIDKeys := lo.SliceToMap(droppedMessageIDs, func(m string) (string, struct{}) { return m, struct{}{} })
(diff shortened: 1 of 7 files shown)


processor/processor.go
```diff
@@ -2216,7 +2216,8 @@ func (proc *Handle) transformSrcDest(
 			var successCountMetadataMap map[string]MetricMetadata
 			eventsToTransform, successMetrics, successCountMap, successCountMetadataMap = proc.getDestTransformerEvents(response, commonMetaData, eventsByMessageID, destination, transformer.UserTransformerStage, trackingPlanEnabled, transformationEnabled)
 			failedJobs, failedMetrics, failedCountMap := proc.getFailedEventJobs(response, commonMetaData, eventsByMessageID, transformer.UserTransformerStage, transformationEnabled, trackingPlanEnabled)
-			proc.saveFailedJobs(failedJobs)
+			droppedJobs := proc.getDroppedJobs(response, eventList)
+			proc.saveFailedJobs(append(failedJobs, droppedJobs...))
```
atzoum, line 2220: I believe that for rudder sources' use cases both failed and dropped jobs in processor's pipeline could be considered as dropped (i.e. non-retryable)

1. We could rename 
  - `proc.saveFailedJobs` to `proc.saveDroppedJobs `
  - `rsources.NewFailedJobsCollector` to `rsources.NewDroppedJobsCollector`
  - `rsources.JobsFailed` to `rsources.JobsDropped`
3. Dropped jobs failed keys shouldn't be captured
4. It would be better to report rsources dropped jobs during `Store`, once for all scenarios as the last operation of the method, after storing router, batchrouter & proc_errors

processor/events_response.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/rudderlabs/rudder-server/jobsdb"
+	"github.com/rudderlabs/rudder-server/processor/transformer"
+	"github.com/rudderlabs/rudder-server/utils/misc"
+	"github.com/samber/lo"
+)
+
+func (proc *Handle) getDroppedJobs(response transformer.Response, eventsToTransform []transformer.TransformerEvent) []*jobsdb.JobT {
+	// each messageID is one event when sending to the transformer
+	inputMessageIDs := lo.Map(eventsToTransform, func(e transformer.TransformerEvent, _ int) string {
+		return e.Metadata.MessageID
```
atzoum, line 15: I am confused why processor's pipeline operates on the assumption that messageIDs are unique. Dedup is a feature that can be turned off, right? Am I missing some other important fact with respect to message ids and the processor's pipeline?

services/debugger/transformation/transformationStatusUploader.go
```diff
@@ -390,24 +391,6 @@ func (h *Handle) processRecordTransformationStatus(tStatus *TransformationStatus
 					IsError:          true,
 				})
 			}
-		} else if failedEvent.Metadata.MessageID != "" {
```
atzoum, line 393: how is this change relevant to the scope of the pull request?

processor/processor.go
```diff
@@ -2225,7 +2236,7 @@ func (proc *Handle) transformSrcDest(
 			var successCountMetadataMap map[string]MetricMetadata
 			eventsToTransform, successMetrics, successCountMap, successCountMetadataMap = proc.getDestTransformerEvents(response, commonMetaData, eventsByMessageID, destination, transformer.UserTransformerStage, trackingPlanEnabled, transformationEnabled)
 			failedJobs, failedMetrics, failedCountMap := proc.getFailedEventJobs(response, commonMetaData, eventsByMessageID, transformer.UserTransformerStage, transformationEnabled, trackingPlanEnabled)
-			proc.saveFailedJobs(failedJobs)
```
atzoum, line 2228: Are we intentionally excluding `failedJobs` from rsources stats? Should we treat them as dropped jobs instead?

Sidddddarth: oh yeah
missed it among other changes.

services/debugger/transformation/transformationStatusUploader.go
Sidddddarth: not really, but it's the same pattern that we saw in processor, so removed it.

atzoum (approved):

processor/events_response.go
cisse21: Yeah that is correct but dedup ideally should never be turned off

Also for many of the downstream destinations like Warehouse they do operate under the assumption that the primary key is the messageID which is unique

cisse21 (approved):

PR #3854, opened 2023-09-11
fix: validations changes

achettyiitr:
# Description

- Error copy for Deltalke connection timeout.
  - In case of validations for deltalake, if we were not able to connect to the SQL warehouse/cluster display a user-friendly message to the consumer that makes sure that their warehouse/cluster is up and running.
- Making timeout configurable.

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-240/databricks-connect-validator

## Security

### DIFF
--- a/warehouse/validations/validations.go
+++ b/warehouse/validations/validations.go
@@ -22,10 +22,11 @@ const (
 )
 
 var (
-	connectionTestingFolder        string
-	pkgLogger                      logger.Logger
-	fileManagerFactory             filemanager.Factory
-	objectStorageValidationTimeout time.Duration
+	connectionTestingFolder string
+	pkgLogger               logger.Logger
+	fileManagerFactory      filemanager.Factory
+	objectStorageTimeout    time.Duration
+	queryTimeout            time.Duration
 )
 
 var (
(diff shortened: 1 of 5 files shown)


fracasula (approved):
LGTM, weird you got so many failures in the CI tests though. Are they related to this change?

achettyiitr:
> LGTM, weird you got so many failures in the CI tests though. Are they related to this change?

These are related to the compose-test SDK we are using.
https://rudderlabs.slack.com/archives/C01HTT66UMB/p1694094292552449

warehouse/validations/validations.go
```diff
@@ -50,7 +51,8 @@ func Init() {
 	connectionTestingFolder = config.GetString("RUDDER_CONNECTION_TESTING_BUCKET_FOLDER_NAME", misc.RudderTestPayload)
 	pkgLogger = logger.NewLogger().Child("warehouse").Child("validations")
 	fileManagerFactory = filemanager.New
-	objectStorageValidationTimeout = 15 * time.Second
+	objectStorageTimeout = config.GetDuration("Warehouse.Validations.ObjectStorageTimeout", 15, time.Second)
+	queryTimeout = config.GetDuration("Warehouse.Validations.QueryTimeout", 25, time.Second)
```
achettyiitr, line 55: Increasing the default timeout from 15 seconds to 25 seconds as sometimes it takes more than 15 seconds. Also, we need to keep it below cp router default timeout (30 seconds).

fracasula: The fact that we need to keep it below 30 seconds and why is good info, we could have a comment about it here right before `queryTimeout = config.GetDuration(...)`.

achettyiitr: Sure.

achettyiitr: Making changes in [here](https://github.com/rudderlabs/rudder-server/pull/3883).

PR #3856, opened 2023-09-12
chore: support generic rules to have routers drain events

Sidddddarth:
# Description

The topic of draining events from a sync gave birth to this.
This is a way to pass generic rules by which (batch)router and processor(only `jobRunID` now) can drain events. How this is set to config is another story.

Routers, BatchRouters now have a `drainer` which provides the method:
`Drain(job *jobsdb.JobT) (bool, string)` to check if their jobs can be drained. Support for draining via configurable destinationIDs(`"Router.toAbortDestinationIDs"`) and jobRunID(`"RSources.toAbortJobRunIDs"`) present now. Adding more drain configurations should be straightforward -> just need to add logic in `type drainer` and `(d *drainer) Drain(...)`.

Processor has a simpler approach, a new config field is added to accommodate configurations to drain by, `jobRunId` for now. Also injecting `*config.Config` into processor now. Plan is to proliferate it's use instead of `config.Default` in processor package(story for another PR).

## Linear Ticket

[drain events for a cancelled sync](https://linear.app/rudderstack/issue/PIPE-122/drain-events-for-disabled-syncs)

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -143,6 +143,8 @@ type Handle struct {
 		eventSchemaV2Enabled      bool
 		archivalEnabled           misc.ValueLoader[bool]
 		eventAuditEnabled         map[string]bool
+
+		toAbortJobRunIDs misc.ValueLoader[string]
 	}
 
 	adaptiveLimit func(int64) int64
@@ -2555,10 +2557,26 @@ func ConvertToFilteredTransformerResponse(events []transformer.TransformerEvent,
 	supportedMessageTypesCache := make(map[string]*cacheValue)
 	supportedMessageEventsCache := make(map[string]*cacheValue)
 
+	toAbortJobRunIDs := config.GetStringVar("", "RSources.toAbortJobRunIDs")
+
 	// filter unsupported message types
-	var resp transformer.TransformerResponse
 	for i := range events {
 		event := &events[i]
+
+		if toAbortJobRunIDs != "" {
+			abortIDs := strings.Split(toAbortJobRunIDs, ",")
+			if slices.Contains(abortIDs, event.Metadata.SourceJobRunID) {
+				failedEvents = append(failedEvents,
+					transformer.TransformerResponse{
+						Output:     event.Message,
+						StatusCode: 400,
+						Metadata:   event.Metadata,
+						Error:      "jobRunID configured to abort",
+					},
+				)
+				continue
+			}
+		}
 
 		if filter {
 			// filter unsupported message types
@@ -2589,21 +2607,41 @@ func ConvertToFilteredTransformerResponse(events []transformer.TransformerEvent,
 				messageEvent, typOk := event.Message["event"].(string)
 				if !typOk {
 					// add to FailedEvents
-					resp = transformer.TransformerResponse{Output: event.Message, StatusCode: 400, Metadata: event.Metadata, Error: "Invalid message event. Type assertion failed"}
-					failedEvents = append(failedEvents, resp)
+					failedEvents = append(
+						failedEvents,
+						transformer.TransformerResponse{
+							Output:     event.Message,
+							StatusCode: 400,
+							Metadata:   event.Metadata,
+							Error:      "Invalid message event. Type assertion failed",
+						},
+					)
 					continue
 				}
 				if !slices.Contains(supportedEvents.values, messageEvent) {
-					resp = transformer.TransformerResponse{Output: event.Message, StatusCode: types.FilterEventCode, Metadata: event.Metadata, Error: "Event not supported"}
-					failedEvents = append(failedEvents, resp)
+					failedEvents = append(
+						failedEvents,
+						transformer.TransformerResponse{
+							Output:     event.Message,
+							StatusCode: types.FilterEventCode,
+							Metadata:   event.Metadata,
+							Error:      "Event not supported",
+						},
+					)
 					continue
 				}
 			}
 
 		}
 		// allow event
-		resp = transformer.TransformerResponse{Output: event.Message, StatusCode: 200, Metadata: event.Metadata}
-		responses = append(responses, resp)
+		responses = append(
+			responses,
+			transformer.TransformerResponse{
+				Output:     event.Message,
+				StatusCode: 200,
+				Metadata:   event.Metadata,
+			},
+		)
 	}
 
 	return transformer.Response{Events: responses, FailedEvents: failedEvents}
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1591,10 +1591,11 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 			}
 			// check if jobRunId is cancelled
 			if len(proc.drainConfig.jobRunIDs.Load()) > 0 {
-				if slices.Contains(
-					proc.drainConfig.jobRunIDs.Load(),
-					commonMetadataFromSingularEvent.SourceJobRunID,
-				) {
+				if commonMetadataFromSingularEvent.SourceJobRunID != "" &&
+					slices.Contains(
+						proc.drainConfig.jobRunIDs.Load(),
+						commonMetadataFromSingularEvent.SourceJobRunID,
+					) {
 					proc.logger.Debugf(
 						"cancelled jobRunID: %s",
 						commonMetadataFromSingularEvent.SourceJobRunID,
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1590,18 +1590,16 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 				continue
 			}
 			// check if jobRunId is cancelled
-			if len(proc.drainConfig.jobRunIDs.Load()) > 0 {
-				if commonMetadataFromSingularEvent.SourceJobRunID != "" &&
-					slices.Contains(
-						proc.drainConfig.jobRunIDs.Load(),
-						commonMetadataFromSingularEvent.SourceJobRunID,
-					) {
-					proc.logger.Debugf(
-						"cancelled jobRunID: %s",
-						commonMetadataFromSingularEvent.SourceJobRunID,
-					)
-					continue
-				}
+			if commonMetadataFromSingularEvent.SourceJobRunID != "" &&
+				slices.Contains(
+					proc.drainConfig.jobRunIDs.Load(),
+					commonMetadataFromSingularEvent.SourceJobRunID,
+				) {
+				proc.logger.Debugf(
+					"cancelled jobRunID: %s",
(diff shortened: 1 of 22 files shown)


router/utils/utils.go
```diff
@@ -47,9 +46,40 @@ func getRetentionTimeForDestination(destID string) time.Duration {
...
+	SourceCategory          string      `json:"source_category"`
+	RecordID                interface{} `json:"record_id"`
+	MessageID               string      `json:"message_id"`
+	EventName               string      `json:"event_name"`
+	EventType               string      `json:"event_type"`
+	WorkspaceID             string      `json:"workspaceId"`
+	RudderAccountID         string      `json:"rudderAccountId"`
+}
+
+type AbortConfigs struct {
```
atzoum, line 68: We could get rid of this config struct if we had a `Drainer` interface with a single method

```go
Drain(job *jobsdb.JobT, jobParams JobParameters, destinationsMap map[string]*DestinationWithSources) (bool, string)
```
This component could be initialised using the config object. Drain logic and configuration could then be internal to this component

Sidddddarth: Please take another look(rt, brt, router/utils).
Processor changes will follow soon.

router/batchrouter/handle_lifecycle.go
```diff
@@ -199,6 +199,7 @@ func (brt *Handle) setupReloadableVars() {
 	brt.uploadFreq = config.GetReloadableDurationVar(30, time.Second, "BatchRouter.uploadFreqInS", "BatchRouter.uploadFreq")
 	brt.mainLoopFreq = config.GetReloadableDurationVar(30, time.Second, "BatchRouter.mainLoopFreq")
 	brt.toAbortDestinationIDs = config.GetReloadableStringVar("", "BatchRouter.toAbortDestinationIDs")
```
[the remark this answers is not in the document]
Sidddddarth: Touché
Using one config for both rt, brt now.

router/types.go
```diff
@@ -11,24 +11,6 @@ import (
...
-	SourceTaskRunID         string      `json:"source_task_run_id"`
-	SourceJobID             string      `json:"source_job_id"`
-	SourceJobRunID          string      `json:"source_job_run_id"`
-	SourceDefinitionID      string      `json:"source_definition_id"`
-	DestinationDefinitionID string      `json:"destination_definition_id"`
-	SourceCategory          string      `json:"source_category"`
-	RecordID                interface{} `json:"record_id"`
-	MessageID               string      `json:"message_id"`
-	WorkspaceID             string      `json:"workspaceId"`
-	RudderAccountID         string      `json:"rudderAccountId"`
```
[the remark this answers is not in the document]
Sidddddarth: as for `RudderAccountId`, it is present in multiple structs in the repo.
I will tackle them in another PR. Or perhaps defer to integrations, since they seem to have introduced it.

processor/processor.go
```diff
@@ -2582,11 +2582,27 @@ func ConvertToFilteredTransformerResponse(events []transformer.TransformerEvent,
 	supportedMessageTypesCache := make(map[string]*cacheValue)
 	supportedMessageEventsCache := make(map[string]*cacheValue)
 
+	toAbortJobRunIDs := config.GetStringVar("", "RSources.toAbortJobRunIDs")
```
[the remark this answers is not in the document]
Sidddddarth: Did not know the presence of such a method.(_How can I speak of that to which I have no knowledge_ 😛)
Tangentially, why do we even need a slice, just a `strings.Contains(config.GetString(abortDestID, jobParams.DestinationID)` would suffice, no? Considering `jobParams.DestinationID` can never be empty.
Also this would need some publishing as people are used to comma separated config probably

processor/processor.go
```diff
@@ -1600,6 +1615,20 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
...
+						proc.drainConfig.jobRunIDs.Load(),
+						commonMetadataFromSingularEvent.SourceJobRunID,
+					) {
+					proc.logger.Debugf(
+						"cancelled jobRunID: %s",
+						commonMetadataFromSingularEvent.SourceJobRunID,
+					)
+					continue
+				}
+			}
```
atzoum, line 1631:
```suggestion
			if commonMetadataFromSingularEvent.SourceJobRunID != "" &&
				slices.Contains(
					proc.drainConfig.jobRunIDs.Load(),
					commonMetadataFromSingularEvent.SourceJobRunID,
				) {
				proc.logger.Debugf(
					"cancelled jobRunID: %s",
					commonMetadataFromSingularEvent.SourceJobRunID,
				)
				continue
			}
```

router/utils/utils.go
```diff
@@ -92,3 +95,60 @@ func EnhanceJsonWithTime(t time.Time, key string, resp []byte) []byte {
...
+type drainer struct {
+	destinationIDs misc.ValueLoader[[]string]
+	jobRunIDs      misc.ValueLoader[[]string]
+}
+
+func (d *drainer) Drain(
+	job *jobsdb.JobT,
+	jobParams JobParameters,
+	destinationsMap map[string]*DestinationWithSources,
+) (bool, string) {
```
atzoum, line 129: `Drain` method shall only accept `job *jobsdb.JobT` as a parameter:
- The method's implementation can unmarshal parameters on its own & only the fields that it is interested in
- Whether a destination is enabled or not, this can be provided through an external function while creating a new drainer, e,.g. `enabledDestinationFunc func(destinationID string) (enabled bool)`

router/types.go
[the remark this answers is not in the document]
cisse21: RudderAccountID is used for OAuth destination IIRC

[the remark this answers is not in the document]
atzoum: Nobody is reading it in server's code though, @sanpj2292 ?

router/utils/utils.go
```diff
@@ -92,3 +95,60 @@ func EnhanceJsonWithTime(t time.Time, key string, resp []byte) []byte {
...
+		job *jobsdb.JobT,
+		jobParams JobParameters,
+		destinationsMap map[string]*DestinationWithSources,
+	) (bool, string)
+}
+
+func NewDrainer(conf *config.Config) Drainer {
+	return &drainer{
+		destinationIDs: conf.GetReloadableStringSliceVar(
+			[]string{},
```
atzoum, line 110:
```suggestion
			nil,
```

router/utils/utils.go
```diff
@@ -92,3 +95,60 @@ func EnhanceJsonWithTime(t time.Time, key string, resp []byte) []byte {
...
+			"Router.toAbortDestinationIDs",
+		),
+		jobRunIDs: conf.GetReloadableStringSliceVar(
+			[]string{},
```
atzoum, line 114:
```suggestion
			nil,
```

router/utils/utils.go
Sidddddarth: why even pass `enabledDestinationFunc`?
if the router worker already knows destination is disabled, it can abort a job right away.
Only otherwise can it check `rt.drainer.Drain(...)`

Sidddddarth: agree about the only-job-as-parameter part

router/utils/utils.go
```diff
@@ -92,3 +100,60 @@ func EnhanceJsonWithTime(t time.Time, key string, resp []byte) []byte {
...
+		return true, DestAbort
+	}
+
+	if jobParams.SourceJobRunID != "" &&
+		slices.Contains(d.jobRunIDs.Load(), jobParams.SourceJobRunID) {
+		return true, JobRunIDCancelled
+	}
+
+	return false, ""
+}
```
atzoum, line 159: This makes more sense to me, i.e. 
1. drainer can decide whether a job can be drained or not, as long as it resolve the destination's configuration from the destination id
2. The only argument `Drain` method should accept is `*jobsdb.JobT` 

```suggestion
type drainer struct {
	destinationIDs      misc.ValueLoader[[]string]
	jobRunIDs           misc.ValueLoader[[]string]
	destinationResolver func(string) (*backendconfig.DestinationT, bool)
}

func (d *drainer) Drain(job *jobsdb.JobT) (bool, string) {
	createdAt := job.CreatedAt
	var parameters JobParameters
	_ = json.Unmarshal(job.Parameters, &parameters)
	destID := parameters.DestinationID
	if time.Since(createdAt) > getRetentionTimeForDestination(destID) {
		return true, "job expired"
	}

	if destination, ok := d.destinationResolver(destID); !ok {
		return true, DestNotFoundInConfig
	} else if !destination.Enabled {
		return true, DestDisabled
	}

	if slices.Contains(d.destinationIDs.Load(), destID) {
		return true, DestAbort
	}

	if parameters.SourceJobRunID != "" &&
		slices.Contains(d.jobRunIDs.Load(), parameters.SourceJobRunID) {
		return true, JobRunIDCancelled
	}

	return false, ""
}
```

router/utils/utils.go
```diff
@@ -24,6 +23,11 @@ const (
 	ERROR_AT_DEL = "delivery"
 	// custom destination manager
 	ERROR_AT_CUST = "custom"
+
+	DestNotFoundInConfig = "destination is not available in the config"
+	DestDisabled         = "destination is disabled"
+	DestAbort            = "destination configured to abort"
+	JobRunIDCancelled    = "cancelled jobRunID"
```
atzoum, line 30:
```suggestion
	DrainReasonDestNotFound = "destination is not available in the config"
	DrainReasonDestDisabled         = "destination is disabled"
	DrainReasonDestAbort            = "destination configured to abort"
	DrainReasonJobRunIDCancelled    = "cancelled jobRunID"
```

processor/processor.go
```diff
@@ -1629,6 +1641,18 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
...
+				slices.Contains(
+					proc.drainConfig.jobRunIDs.Load(),
+					commonMetadataFromSingularEvent.SourceJobRunID,
+				) {
+				proc.logger.Debugf(
+					"cancelled jobRunID: %s",
+					commonMetadataFromSingularEvent.SourceJobRunID,
+				)
+				continue
+			}
```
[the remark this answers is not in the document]
Sidddddarth: Yeah, as it stands all the dropped counts would go to diff at destination filter stage.
Would it be better to move this to the event filter stage?

atzoum (approved):

router/utils/utils.go
```diff
@@ -1,6 +1,7 @@
 package utils
```
lvrach, line 1: the package name is not very indicative of its contents, and in general, util/misc names should be avoided.

We can have a `type` package for `JobParameters`.
and introduce a new `drainer` package for `NewDrainer`

lvrach (approved):

PR #3857, opened 2023-09-12
fix: send reportedAt in milliseconds for error reporting

sanpj2292:
# Description

We used to send `reportedAt` in minutes to reporting service which is causing problems.
With this PR, we would be sending `reportedAt` in milliseconds to reporting service

## Linear Ticket

https://linear.app/rudderstack/issue/INT-608/send-reportedat-in-milliseconds

## Security


koladilip (approved):

lvrach (commented):
This change looks ok for now, but I would recommend two things in a separate PR:
* use time.Time all the way until we store to database or encode the payload
* add a test case for this

lvrach (approved):

PR #3859, opened 2023-09-13
feat: added flags for event audit

kanishkkatara:
# Description

Fetching event audit enabled flag from workspace config and using it in processor to send events to esch_jobs table

## Linear Ticket

https://linear.app/rudderstack/issue/DAT-285/add-ability-for-customer-to-self-serve-the-enablement-disablement-of

## Security


processor/processor.go
```diff
@@ -731,13 +731,15 @@ func (proc *Handle) backendConfigSubscriber(ctx context.Context) {
 				}
 			}
 			workspaceLibrariesMap[workspaceID] = wConfig.Libraries
+			eventAuditEnabled = wConfig.Settings.EventAuditEnabled
```
abhimanyubabbar, line 734: You would override the value in this loop as it is running for multiple workspaces right ?

abhimanyubabbar (changes requested):

atzoum (changes requested):
`eventAuditEnabled` needs to be a `map[string]bool` protected by `configSubscriberLock`

processor/processor.go
```diff
@@ -1490,8 +1492,7 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 			},
 			)
 			if proc.config.eventSchemaV2Enabled && // schemas enabled
-				// source has schemas enabled or if we override schemas for all sources
-				(source.EventSchemasEnabled || proc.config.eventSchemaV2AllSources) &&
```
atzoum, line 1494: Please delete `SourceT#EventSchemasEnabled` in `backendconfig/types.go` and adapt `processor_test.go`

processor/processor.go
```diff
@@ -725,7 +724,9 @@ func (proc *Handle) backendConfigSubscriber(ctx context.Context) {
 			sourceIdDestinationMap = make(map[string][]backendconfig.DestinationT)
 			sourceIdSourceMap      = map[string]backendconfig.SourceT{}
 			destinationIDtoTypeMap = make(map[string]string)
+			eventAuditEnabled      = make(map[string]bool)
 		)
+		proc.config.configSubscriberLock.Lock()
```
atzoum, line 729: please revert this change, the lock shall be obtains as before, to minimize contention

processor/processor.go
```diff
@@ -1490,8 +1492,7 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 			},
 			)
 			if proc.config.eventSchemaV2Enabled && // schemas enabled
-				// source has schemas enabled or if we override schemas for all sources
-				(source.EventSchemasEnabled || proc.config.eventSchemaV2AllSources) &&
+				proc.config.eventAuditEnabled[batchEvent.WorkspaceId] &&
```
atzoum, line 1495: You need to acquire a read lock before reading the value from `eventAuditEnabled` here

processor/processor.go
```diff
@@ -1490,8 +1498,8 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 			},
 			)
 			if proc.config.eventSchemaV2Enabled && // schemas enabled
-				// source has schemas enabled or if we override schemas for all sources
-				(source.EventSchemasEnabled || proc.config.eventSchemaV2AllSources) &&
+				proc.eventAuditEnabled(batchEvent.WorkspaceId) &&
+				proc.config.eventAuditEnabled[batchEvent.WorkspaceId] &&
```
atzoum, line 1502: oops
```suggestion
```

atzoum (approved):

backend-config/types.go
```diff
@@ -121,7 +120,8 @@ func (c *ConfigT) DestinationsMap() map[string]*DestinationT {
 }
 
 type Settings struct {
-	DataRetention DataRetention `json:"dataRetention"`
+	DataRetention     DataRetention `json:"dataRetention"`
+	EventAuditEnabled bool          `json:"eventAuditEnabled"`
```
abhimanyubabbar, line 124: Have we verified that data is unmarshalling into this struct automatically ?

backend-config/backend-config.go
```diff
@@ -156,6 +156,8 @@ func filterProcessorEnabledDestinations(config ConfigT) ConfigT {
 		source.Destinations = destinations
 		modifiedConfig.Sources = append(modifiedConfig.Sources, source)
 	}
+	modifiedConfig.Settings.DataRetention.DisableReportingPII = config.Settings.DataRetention.DisableReportingPII
```
atzoum, line 159: can't we just use the whole `Settings` struct instead of selectively picking fields??

kanishkkatara: DataRetention has other properties that will get overridden, thats why didn't do this

atzoum: Not including workspace settings in the `TopicProcessConfig` looks like a bug to me, cc @cisse21, @BonapartePC

backend-config/types.go
kanishkkatara: yes

lvrach (approved):

backend-config/backend-config.go
cisse21: Yeah true.... this looks like a bug which we had.

abhimanyubabbar (approved):

PR #3862, opened 2023-09-13
chore: app for warehouse

achettyiitr:
# Description

- Introduce the warehouse app.
- Remove globals from the warehouse package.
- Added test cases.

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-218/introduce-warehouse-app

## Security


warehouse/warehouse.go
```diff
@@ -217,269 +274,224 @@ func setupTables(dbHandle *sql.DB) error {
...
 
 // Start starts the warehouse service
-func Start(ctx context.Context, app app.App) error {
-	application = app
-
-	mode := config.GetString("Warehouse.mode", config.EmbeddedMode)
-
-	if dbHandle == nil && !isStandAloneSlave(mode) {
-		return errors.New("warehouse service cannot start, database connection is not setup")
-	}
```
achettyiitr, line 316: No point in throwing an error, since later we are already skipping the start for the warehouse service.

warehouse/trigger/trigger.go
fracasula, line 1: One concern about this package is that it doesn't have anything to do with triggers, it's basically just a `sync.Map`, right?

So I was going to suggest to move it to some `util` package but if I benchmark it first.

```go
package trigger

import (
	"sync"
	"testing"
)

func BenchmarkTrigger(b *testing.B) {
	b.Run("trigger", func(b *testing.B) {
		store := NewStore()

		var wg sync.WaitGroup
		wg.Add(b.N * 3)

		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			go func() {
				defer wg.Done()
				store.Trigger("concurrent")
			}()
			go func() {
				defer wg.Done()
				_ = store.IsTriggered("concurrent")
			}()
			go func() {
				defer wg.Done()
				store.ClearTrigger("concurrent")
			}()
		}

		wg.Wait()
	})
	b.Run("sync map", func(b *testing.B) {
		m := sync.Map{}

		var wg sync.WaitGroup
		wg.Add(b.N * 3)

		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			go func() {
				defer wg.Done()
				m.Store("concurrent", struct{}{})
			}()
			go func() {
				defer wg.Done()
				m.Load("concurrent")
			}()
			go func() {
				defer wg.Done()
				m.Delete("concurrent")
			}()
		}

		wg.Wait()
	})
}
```

```
BenchmarkTrigger/trigger-20         	  720672	      1694 ns/op
BenchmarkTrigger/sync_map-20        	 2196574	      534.1 ns/op
```

warehouse/router.go
```diff
@@ -691,12 +713,40 @@ func (r *router) createUploadJobsFromStagingFiles(ctx context.Context, warehouse
...
+	freqInS := r.uploadFreqInS(syncFrequency)
+
+	r.createJobMarkerMapLock.RLock()
+	defer r.createJobMarkerMapLock.RUnlock()
+
+	lastCreatedAt, ok := r.createJobMarkerMap[warehouse.Identifier]
+	if !ok {
+		return true
+	}
+	return r.now().Sub(lastCreatedAt) > time.Duration(freqInS)*time.Second
```
fracasula, line 737: [very minor] it's simple enough in this case to avoid keeping the lock while calculating this:

```go
r.createJobMarkerMapLock.RLock()
lastCreatedAt, ok := r.createJobMarkerMap[warehouse.Identifier]
r.createJobMarkerMapLock.RUnlock()
```

warehouse/slave_worker_job.go
```diff
@@ -389,6 +389,10 @@ func (jr *jobRun) uploadLoadFiles(ctx context.Context) ([]uploadResult, error) {
 	return output, nil
 }
 
+func (jr *jobRun) bucketFolder(batchID, tableName string) string {
+	return fmt.Sprintf(`%v-%v`, batchID, tableName)
```
fracasula, line 393:
```suggestion
	return batchID+"-"+tableName
```

warehouse/warehouse.go
```diff
@@ -217,269 +274,224 @@ func setupTables(dbHandle *sql.DB) error {
...
+func (a *App) Start(ctx context.Context) error {
```
fracasula, line 287: Same comment as above about the naming. If there is no `Stop` method please consider naming this `Run`.

warehouse/warehouse_test.go
```diff
@@ -0,0 +1,400 @@
...
+					pgResource, err := resource.SetupPostgres(pool, t)
+					require.NoError(t, err)
+
+					webPort, err := kithelper.GetFreePort()
+					require.NoError(t, err)
+
+					ctx, stopServer := context.WithCancel(context.Background())
+
+					port, err := strconv.Atoi(pgResource.Port)
+					require.NoError(t, err)
```
fracasula, line 175: Can't we skip this step and pass `pgResource.Port` directly in the `c.Set()` below?

warehouse/trigger/trigger.go
achettyiitr: Yeah, it has nothing to do with triggers, Just a plain vanilla map. Moving it to utils package.

achettyiitr: Thanks for doing a quick Benchmark. 😄

warehouse/warehouse_test.go
```diff
@@ -0,0 +1,400 @@
...
+		c.Set("WAREHOUSE_JOBS_DB_PORT", "5432")
+		c.Set("WAREHOUSE_JOBS_DB_USER", "ubuntu")
+		c.Set("WAREHOUSE_JOBS_DB_PASSWORD", "ubuntu")
+		c.Set("WAREHOUSE_JOBS_DB_DB_NAME", "ubuntu")
+
+		a := NewApp(mockApp, c, logger.NOP, stats.Default, &bcConfig.NOOP{}, filemanager.New)
+		err = a.Setup(context.Background())
+		require.ErrorContains(t, err, "setting up database: could not check compatibility:")
+	})
+	t.Run("without env vars", func(t *testing.T) {
```
fracasula, line 251: Why without env vars? It looks like we're passing them :thinking:

warehouse/warehouse_test.go
```diff
@@ -0,0 +1,400 @@
...
+				{
+					name:        "degraded mode",
+					runningMode: degradedMode,
+				},
+			}
+
+			for _, subTC := range subTestCases {
+				subTC := subTC
+
+				t.Run(tc.name+" with "+subTC.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
fracasula: Same with other things, for example if master then we start the cron archiver. Are we verifying that expectation somewhere?

fracasula (commented):
Great work so far @achettyiitr, it's looking good :+1:

warehouse/warehouse_test.go
achettyiitr: I will update the test context. It was for `without warehouse env vars`. In that case, we are using `misc.GetConnectionString`

achettyiitr:
> It still has nothing to do with triggers though so maybe we can give it a generic name or remove it completely.
> Thoughts?

Yeah, we can remove this and just pass the dependency for `triggerStore: syncs.Map`

warehouse/warehouse.go
achettyiitr: Thanks for pointing this out. 😄

warehouse/warehouse_test.go
achettyiitr: Since we are getting config.GetInt, I was expecting to set config.Set(<>, <int>). But later realized that [viper uses `cast.ToInt`](https://github.com/spf13/viper/blob/336e9f0a93eec4cec432eadda15d8d0803997d09/viper.go#L987) behind the scenes.

achettyiitr: Changing now.

achettyiitr: Thanks for pointing out.

warehouse/mode.go
```diff
@@ -2,6 +2,8 @@ package warehouse
 
 import "github.com/rudderlabs/rudder-go-kit/config"
 
+const degradedMode = "degraded"
```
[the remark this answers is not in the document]
achettyiitr: I am not sure about this. The rudder server also has a degraded mode which is different from the one kept in the warehouse. In the case of the warehouse, it is pointing to `Warehouse.runningMode`

warehouse/jobs/runner.go
```diff
@@ -107,11 +109,11 @@ func (a *AsyncJobWh) addJobsToDB(payload *AsyncJobPayload) (int64, error) {
...
-// InitAsyncJobRunner Async Job runner's main job is to
+// Start Async Job runner's main job is to
 // 1. Scan the database for entries into wh_async_jobs
 // 2. Publish data to pg_notifier queue
 // 3. Move any executing jobs to waiting
-func (a *AsyncJobWh) InitAsyncJobRunner() error {
+func (a *AsyncJobWh) Start() error {
 	// Start the asyncJobRunner
 	a.logger.Info("[WH-Jobs]: Initializing async job runner")
 	g, ctx := errgroup.WithContext(a.context)
```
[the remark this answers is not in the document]
achettyiitr: > Do we need the group in the first place? Since it is only really using a single routine, shouldn't we just remove it and block at the end with a return a.startAsyncJobRunner(ctx)?

> It would be great if we can have it accept a context as well. Then once the context is canceled Run can return after gracefully terminating all the routines that it started.

I will try to take these up in a separate PR. I have updated the [linear ticket](https://linear.app/rudderstack/issue/PIPE-168/jobs-cleanup) regarding this.

warehouse/warehouse_test.go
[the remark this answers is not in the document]
achettyiitr: Currently, I am just verifying whether it is able to serve HTTP service.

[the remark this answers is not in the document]
achettyiitr: But, currently, I am not sure how I can verify other things.

warehouse/upload.go
```diff
@@ -85,13 +85,13 @@ type UploadJobFactory struct {
...
 	ctx                  context.Context
-	dbHandle             *sqlmiddleware.DB
+	db                   *sqlmiddleware.DB
 	destinationValidator validations.DestinationValidator
 	loadfile             *loadfiles.LoadFileGenerator
 	tableUploadsRepo     *repo.TableUploads
 	recovery             *service.Recovery
 	whManager            manager.Manager
-	pgNotifier           *pgnotifier.PGNotifier
+	notifier             *pgnotifier.PGNotifier
```
fracasula, line 94: Is this used at all? :thinking:

achettyiitr: Good catch. It's not being used.

achettyiitr: Curious, do you find it via some static check?

achettyiitr:
> > It still has nothing to do with triggers though so maybe we can give it a generic name or remove it completely.
> > Thoughts?
> 
> Yeah, we can remove this and just pass the dependency for `triggerStore: syncs.Map`. But I don't want to expose **Store, Delete, and Load** to other modules. Currently, it hides the implementation details. Thoughts?

Thanks, @fracasula, for pointing it out regarding `triggerStore`. No need to have it, since it is just proxying to `syncs.Map`

achettyiitr:
Trigger store to use local map.

atzoum:
@achettyiitr this could be a good chance to fix deprecated usages of the config api in warehouse, e.g. `config.RegisterXxxConfigVariable`.

All other packages have been adapted except for `reporting`, `event_schema` (soon to be removed) and `warehouse`

achettyiitr:
> @achettyiitr this could be a good chance to fix deprecated usages of the config api in warehouse, e.g. `config.RegisterXxxConfigVariable`.
> 
> All other packages have been adapted except for `reporting`, `event_schema` (soon to be removed) and `warehouse`

Picking it up just after these changes get merged.

warehouse/jobs/runner.go
```diff
@@ -107,11 +109,11 @@ func (a *AsyncJobWh) addJobsToDB(payload *AsyncJobPayload) (int64, error) {
 	return jobId, nil
 }
 
-// InitAsyncJobRunner Async Job runner's main job is to
+// Start Async Job runner's main job is to
```
fracasula, line 112:
```suggestion
// Run Async Job runner's main job is to
```

fracasula (approved):
Approving with the last batch of minor comments. Thanks @achettyiitr for the effort! This is good work :+1:

warehouse/warehouse_test.go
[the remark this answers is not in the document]
achettyiitr: Currently, I have added HTTP and GRPC. We can add incrementally.

lvrach (approved):

PR #3864, opened 2023-09-13
chore: go-kit v0.15.9

fracasula:
# Description

Pulling kit with the new config Reloadable API.

The idea is to suppress the linter to incrementally migrate towards the new API. While making the changes I took the liberty to migrate a few packages.

Follow up task (it has 2 sub-issues, 1 for server, 1 for warehouse):
* https://linear.app/rudderstack/issue/PIPE-263/sa1019-config-register-reloadable-functions-are-deprecated

## Sidenote

In some tests reloadable variables are changed manually. For now I fixed them by doing:

```go
disableEventUploads = config.GetReloadableBoolVar(true, rand.UniqueString(10))
```

The ideal way of doing it would be to have proper constructors that take a `*config.Config` as a parameter. Unfortunately that requires refactoring the code to make it work that way. As a temporary hack I used the above trick.

If you disagree I can revert, suppress the linter and then we can take the time to properly refactor the affected packages. In those cases I felt it was overkill so I went ahead with that approach.

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-257/update-rudderserver-to-suppressremove-deprecated-config-calls) >

## Security

### DIFF
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -53,8 +53,9 @@ type embeddedApp struct {
 	}
 }
 
+// nolint:staticcheck // SA1019: config Register reloadable functions are deprecated
 func (a *embeddedApp) loadConfiguration() {
-	config.RegisterBoolConfigVariable(types.DefaultReplayEnabled, &a.config.enableReplay, false, "Replay.enabled")
+	a.config.enableReplay = config.GetBoolVar(types.DefaultReplayEnabled, "Replay.enabled")
 	config.RegisterIntConfigVariable(0, &a.config.processorDSLimit, true, 1, "Processor.jobsDB.dsLimit", "JobsDB.dsLimit")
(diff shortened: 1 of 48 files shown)


fracasula:
> Since not all changes are inline (some code has been moved around) it is not easy to review and visually verify correctness of all changes.

@atzoum Yeah I know, adding a comment on top of every single line it was a bit too much so I moved them in a method and commented that method instead.

Then I removed the `[]string{"key1", "key2"}...` pattern and I refactored the non-reloadable variables since those don't need any particular refactoring.

I'd say, pay more attention where you see that there is an actual `*config.Reloadable[T]` variable being used like in:
* services/debugger/destination/eventDeliveryStatusUploader.go
* services/debugger/source/eventUploader.go
* services/pgnotifier/pgnotifier.go
* services/streammanager/kafka/kafkamanager.go
* warehouse/archive/archiver.go
* warehouse/encoding/encoding.go

Sidddddarth (approved):

PR #3866, opened 2023-09-14
feat(processor): added ability for geolocation enrichment during pipeline processing

abhimanyubabbar:
# Description
The change is adding a pipeline enricher ( geo enrichment ) when processing gateway jobs for a particular source. Below are sequence of events:

* The customer will toggle a boolean flag on the Webapp which will augment the information at source level.
* Rudder server will read this information and if pipeline enrichment is enabled, it will start enriching the geo information into the event based on the request IP.
* We will be using unofficial Maxmind GeoIP libraries to read the Maxmind in memory database and the records will be then enriched accordingly.

## Linear Ticket

[Ticket](https://linear.app/rudderstack/project/geolocation-enrichment-at-gateway-eaf72a1a80ae)

## Security


lvrach (commented):
Looks good to me so far, I would take a look once we add tests

backend-config/types.go
```diff
@@ -69,6 +69,10 @@ type SourceT struct {
 	WriteKey                   string
 	DgSourceTrackingPlanConfig DgSourceTrackingPlanConfigT
 	Transient                  bool
+	EventSchemasEnabled        bool
```
achettyiitr, line 72: EventSchemasEnabled?
```suggestion
```

app/apphandlers/embeddedAppHandler.go
```diff
@@ -231,6 +232,23 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 
 	adaptiveLimit := payload.SetupAdaptiveLimiter(ctx, g)
 
+	var geoEnricher enricher.PipelineEnricher
+	if config.GetBool("GeoEnrichment.enabled", false) {
+
+		dbProvider, err := enricher.NewGeoDBProvider(config, a.log)
+		if err != nil {
+			return fmt.Errorf("creating new instance of db provider: %w", err)
```
achettyiitr, line 240: [minor] May be we can we more specific here?
```suggestion
			return fmt.Errorf("creating new instance of geo db provider: %w", err)
```

app/apphandlers/processorAppHandler.go
```diff
@@ -223,6 +222,23 @@ func (a *processorApp) StartRudderCore(ctx context.Context, options *app.Options
 
 	adaptiveLimit := payload.SetupAdaptiveLimiter(ctx, g)
 
+	var geoEnricher enricher.PipelineEnricher
+	if config.GetBool("GeoEnrichment.enabled", false) {
+
+		dbProvider, err := enricher.NewGeoDBProvider(config.Default, a.log)
+		if err != nil {
+			return fmt.Errorf("creating new instance of db provider: %w", err)
```
achettyiitr, line 230: [minor] May be we can we more specific here?
```suggestion
			return fmt.Errorf("creating new instance of geo db provider: %w", err)
```

services/geolocation/maxmind.go
```diff
@@ -0,0 +1,49 @@
...
+type dbReader struct {
+	reader *maxminddb.Reader
+}
+
+func NewMaxmindDBReader(dbLoc string) (*dbReader, error) {
+	reader, err := maxminddb.Open(dbLoc)
+	if err != nil {
+
+		if _, ok := err.(*fs.PathError); ok {
+			return nil, ErrInvalidDatabase
```
achettyiitr, line 26: We can easily add this code path as well in test?

services/geolocation/maxmind_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/rudderlabs/rudder-server/services/geolocation"
+)
+
+func TestGeolocationReader_Failure(t *testing.T) {
+	t.Run("reader errors out when db is corrupted", func(t *testing.T) {
+		_, err := geolocation.NewMaxmindDBReader("./testdata/corrupted_city_test.mmdb")
+		require.True(
+			t,
+			errors.Is(err, geolocation.ErrInvalidDatabase))
```
achettyiitr, line 17:
```suggestion
		require.ErrorIs(t, err, geolocation.ErrInvalidDatabase)
```

services/geolocation/maxmind_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+func TestGeolocationFetcher(t *testing.T) {
+	// below data is a test database provided by maxmind for testing purposes
+	// https://github.com/maxmind/MaxMind-DB/tree/main/test-data
+	f, err := geolocation.NewMaxmindDBReader("./testdata/city_test.mmdb")
+	require.Nil(t, err)
+
+	t.Run("fetcher returns error ErrInvalidIP if ip is empty or invalid", func(t *testing.T) {
+		lookup, err := f.GeoIP(``)
+		require.True(t, errors.Is(err, geolocation.ErrInvalidIP))
```
achettyiitr, line 29:
```suggestion
		require.ErrorIs(t, err, geolocation.ErrInvalidIP)
```

services/geolocation/maxmind_test.go
```diff
@@ -0,0 +1,56 @@
...
+		require.Nil(t, lookup)
+
+		_, err = f.GeoIP(`invalid-ip`)
+		require.True(t, errors.Is(err, geolocation.ErrInvalidIP))
```
achettyiitr, line 33:
```suggestion
		require.ErrorIs(t, err, geolocation.ErrInvalidIP)
```

services/geolocation/maxmind_test.go
```diff
@@ -0,0 +1,56 @@
...
+	})
+
+	t.Run("fetcher returns empty lookup for IP address not available in database", func(t *testing.T) {
+		emptyLookup, err := f.GeoIP(`1.1.1.1`)
+		require.Nil(t, err)
+		require.Equal(t, geolocation.GeoCity{}, *emptyLookup)
+
+		emptyLookup, err = f.GeoIP(`3a02:ff81::0`) // IPv6 lookup
+		require.Nil(t, err)
+		require.Equal(t, geolocation.GeoCity{}, *emptyLookup)
```
achettyiitr, line 54:
```suggestion
		require.Empty(t, *emptyLookup)
```

services/geolocation/maxmind_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+		lookup, err = f.GeoIP(`2a02:ff40::0`) // IPv6
+		require.Nil(t, err)
+		require.Equal(t, `Europe`, lookup.Continent.Names["en"]) // TODO: Do complete matching here.
+	})
+
+	t.Run("fetcher returns empty lookup for IP address not available in database", func(t *testing.T) {
+		emptyLookup, err := f.GeoIP(`1.1.1.1`)
+		require.Nil(t, err)
+		require.Equal(t, geolocation.GeoCity{}, *emptyLookup)
```
achettyiitr, line 50:
```suggestion
		require.Empty(t, *emptyLookup)
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+
+	"github.com/rudderlabs/rudder-go-kit/config"
+	"github.com/rudderlabs/rudder-go-kit/filemanager"
+	"github.com/rudderlabs/rudder-go-kit/logger"
+	"github.com/rudderlabs/rudder-go-kit/stats"
+	"github.com/rudderlabs/rudder-server/services/geolocation"
+	"github.com/rudderlabs/rudder-server/utils/types"
+)
+
+type GeoDBProvider interface {
```
lvrach, line 19:
```suggestion
type GeoDBFetcher interface {
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+	return nil
+}
+
+type NoOpGeoEnricher struct{}
+
+func (e NoOpGeoEnricher) Enrich(sourceId string, request *types.GatewayBatchRequest) error {
+	return nil
+}
+
+type geoDBProviderImpl struct {
```
lvrach, line 115: [minor]
```suggestion
type geoDB struct {
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+	s3Manager filemanager.FileManager
+}
+
+// GetDB simply fetches the database from the upstream located at the key defined
+// in the argument.
+func (p *geoDBProviderImpl) GetDB(key, downloadPath string) error {
```
lvrach, line 121: Should we pass context here? 
```suggestion
func (p *geoDBProviderImpl) GetDB(ctx context.Context, key, downloadPath string) error {
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// Enrich function runs on a request of GatewayBatchRequest which contains
+// multiple singular events from a source. The enrich function augments the
+// geolocation information per event based on IP address.
+func (e *geoEnricher) Enrich(sourceId string, request *types.GatewayBatchRequest) error {
+	e.logger.Debugw("received a call to enrich gateway events for source", "sourceId", sourceId)
+
+	if request.RequestIP == "" {
+		e.stats.NewTaggedStat("proc_geo_enrincher_empty_ip", stats.CountType, stats.Tags{
```
lvrach, line 71: What if we use a single stat with multiple tags: 

stat name:
`proc_geo_enrincher_request`
tags:
* `sourceId=sourceID`
* `error` 
   * `empty_ip`
   * `lookup_failed`
   * `` (empty)

This way we can easily compare the percentage of empty or lookup failed

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+		fetcher: fetcher,
+		stats:   statClient,
+		logger:  log.Child("geolocation"),
+	}, nil
+}
+
+// Enrich function runs on a request of GatewayBatchRequest which contains
+// multiple singular events from a source. The enrich function augments the
+// geolocation information per event based on IP address.
+func (e *geoEnricher) Enrich(sourceId string, request *types.GatewayBatchRequest) error {
```
lvrach, line 67: [minor]
```suggestion
func (e *geoEnricher) Enrich(sourceID string, request *types.GatewayBatchRequest) error {
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+		return fmt.Errorf("downloading db from upstream and storing in file: %w", err)
+	}
+
+	return nil
+}
+
+func NewGeoDBProvider(conf *config.Config, log logger.Logger) (GeoDBProvider, error) {
+	var (
+		bucket = config.GetString("Geolocation.db.bucket", "rudderstack-geolocation")
+		region = config.GetString("Geolocation.db.bucket.region", "us-east-1")
```
achettyiitr, line 140:
```suggestion
		bucket = conf.GetString("Geolocation.db.bucket", "rudderstack-geolocation")
		region = conf.GetString("Geolocation.db.bucket.region", "us-east-1")
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		e.stats.NewTaggedStat(
+			"proc_geo_enricher_geoip_lookup_failed",
+			stats.CountType,
+			stats.Tags{
+				"sourceId": sourceId,
+				"validIP":  strconv.FormatBool(errors.Is(err, geolocation.ErrInvalidIP)),
+			}).Increment()
+
+		return fmt.Errorf("unable to enrich the request with geolocation: %w", err)
```
lvrach, line 95: [minor] We don't have to mention there was a problem, it is implied since it is an error, the idiomatic way is to describe the action:

```suggestion
		return fmt.Errorf("enriching request with geolocation: %w", err)
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	// for every event if we have context object set
+	// only then we add to geo section
+	for _, event := range request.Batch {
+		if context, ok := event["context"].(map[string]interface{}); ok {
```
lvrach, line 101: Should we create the context if it doesn't exist?

processor/processor.go
```diff
@@ -1417,6 +1426,16 @@ func (proc *Handle) eventAuditEnabled(workspaceID string) bool {
 	return proc.config.eventAuditEnabled[workspaceID]
 }
 
+func (proc *Handle) geoEnrichmentEnabledForSource(sourceId string) bool {
```
lvrach, line 1429: [minor]
```suggestion
func (proc *Handle) geoEnrichmentEnabledForSource(sourceID string) bool {
```

lvrach (approved):
Looks good to me with some non-blocking comments.

Most important is to avoid `t.Setenv` in tests as you have already mentioned.

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+
+		return nil
+	}
+
+	defer func(from time.Time) {
+		e.stats.NewTaggedStat(
+			"proc_geo_enricher_request_latency",
+			stats.TimerType,
+			stats.Tags{"sourceId": sourceId}).Since(from)
+	}(time.Now())
```
achettyiitr, line 83: [optional]
```suggestion
		defer e.stats.NewTaggedStat("proc_geo_enricher_request_latency", stats.TimerType, stats.Tags{"sourceId": sourceId}).RecordDuration()()

```

achettyiitr (changes requested):

internal/enricher/geolocation_test.go
```diff
@@ -0,0 +1,250 @@
...
+		input := &geolocation.GeoCity{}
+		actual := extractGeolocationData(input)
+		require.Equal(t, Geolocation{
+			City:     "",
+			Country:  "",
+			Postal:   "",
+			Location: "",
+			Timezone: "",
+			Region:   "",
+		}, *actual)
```
achettyiitr, line 248:
```suggestion
		require.Empty(t, *actual)
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+			context["geo"] = extractGeolocationData(geoip)
+		}
+	}
+
+	return nil
+}
+
+type NoOpGeoEnricher struct{}
+
+func (e NoOpGeoEnricher) Enrich(sourceId string, request *types.GatewayBatchRequest) error {
```
achettyiitr, line 111: [minor]
```suggestion
func (e NoOpGeoEnricher) Enrich(string,*types.GatewayBatchRequest) error {
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("creating a new s3 manager client: %w", err)
+	}
+
+	return &geoDBProviderImpl{
+		s3Manager: manager,
+	}, nil
+}
+
+func extractGeolocationData(geocity *geolocation.GeoCity) *Geolocation {
```
achettyiitr, line 158: [minor]
```suggestion
func extractGeolocationData(geoCity *geolocation.GeoCity) *Geolocation {
```

internal/enricher/geolocation_test.go
```diff
@@ -0,0 +1,250 @@
...
+	})
+
+	t.Run("it returns ErrInvalidIP in case ip address passed is malformed", func(t *testing.T) {
+		err := enricher.Enrich("source-id", &types.GatewayBatchRequest{
+			RequestIP: `invalid-ip`,
+			Batch:     nil,
+		})
+
+		require.NotNil(t, err)
+		require.True(t, errors.Is(err, geolocation.ErrInvalidIP))
```
achettyiitr, line 79:
```suggestion
		require.ErrorIs(t, err, geolocation.ErrInvalidIP)
```

achettyiitr (changes requested):

go.mod
```diff
@@ -111,6 +111,8 @@ require (
 	google.golang.org/protobuf v1.31.0
 )
 
+require github.com/oschwald/maxminddb-golang v1.11.0
```
atzoum, line 114: Please move this to the existing direct dependencies group

internal/enricher/enricher.go
```diff
@@ -0,0 +1,9 @@
+package enricher
+
+import (
+	"github.com/rudderlabs/rudder-server/utils/types"
+)
+
+type PipelineEnricher interface {
+	Enrich(sourceId string, request *types.GatewayBatchRequest) error
```
atzoum, line 8: please add some doc explaining what does enricher does, how it affects the request

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+	s3Manager filemanager.FileManager
+}
+
+// GetDB simply fetches the database from the upstream located at the key defined
+// in the argument.
+func (p *geoDBProviderImpl) GetDB(key, downloadPath string) error {
+	f, err := os.Open(downloadPath)
+	if err != nil {
+		return fmt.Errorf("creating a file to store db contents: %w", err)
+	}
```
atzoum, line 125: Do we have test coverage for this? I guess we don't:

- `os.Open` tries to open a file only for reading... we cannot write to it
- What if the file doesn't exist?

What is the intention of this method? `GetDB` isn't exact in what it does now

services/geolocation/types.go
```diff
@@ -0,0 +1,46 @@
+package geolocation
+
+type GeoFetcher interface {
+	GeoIP(ip string) (*GeoCity, error)
+}
+
+// The City struct corresponds to the data in the GeoIP2/GeoLite2 City
+// databases. Given we are using the native library to decode the information,
+// we have modified some fields in it to contain the pointer values.
+type GeoCity struct {
```
atzoum, line 10:
```suggestion
type GeolocationInfo struct {
```

services/geolocation/maxmind.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+var (
+	ErrInvalidDatabase = errors.New("invalid database file")
+	ErrInvalidIP       = errors.New("ip for lookup cannot be empty or invalid")
+)
+
+type dbReader struct {
+	reader *maxminddb.Reader
+}
```
atzoum, line 19: Just embed it
```suggestion
type maxminddbReader struct {
	*maxminddb.Reader
}
```

services/geolocation/maxmind.go
```diff
@@ -0,0 +1,49 @@
...
+		if errors.As(err, &maxminddb.InvalidDatabaseError{}) {
+			return nil, ErrInvalidDatabase
+		}
+
+		return nil, fmt.Errorf("opening maxmind reader from location: %w", err)
+	}
+	return &dbReader{reader}, nil
+}
+
+func (f *dbReader) GeoIP(ip string) (*GeoCity, error) {
```
atzoum, line 37:
```suggestion
func (f *dbReader) Locate(ip string) (*GeolocationInfo, error) {
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,185 @@
...
+			}).Increment()
+
+		return fmt.Errorf("unable to enrich the request with geolocation: %w", err)
+	}
+
+	// for every event if we have context object set
+	// only then we add to geo section
+	for _, event := range request.Batch {
+		if context, ok := event["context"].(map[string]interface{}); ok {
+			context["geo"] = extractGeolocationData(geoip)
```
atzoum, line 102: we just need to `extractGeolocationData` once, not every time we iterate over the events of the batch

atzoum (changes requested):

internal/enricher/geolocation.go
abhimanyubabbar: Good point, I'll make the change.

internal/enricher/geolocation.go
abhimanyubabbar: I usually try to follow this approach since we last discussed. Probably would have missed it. Thanks for pointing it out.

internal/enricher/geolocation.go
abhimanyubabbar: That's a question I have right now opened up for product. based on the response, I would update it and the tests as well.

backend-config/types.go
abhimanyubabbar: This was a merge conflict resolution issue. Fixed it. 👍🏼

services/geolocation/maxmind.go
abhimanyubabbar: We have that in test.

internal/enricher/geolocation.go
abhimanyubabbar: Thank you for pointing this out. I have fixed it with `os.Create` function.

services/geolocation/types.go
abhimanyubabbar: As per maxmind db, it is represented as GeoCity and hence the same naming convention. Let me know if you feel the name doesn't do justice ?

achettyiitr (approved):

processor/processor.go
```diff
@@ -82,6 +83,7 @@ type Handle struct {
 
 	logger                    logger.Logger
 	eventSchemaHandler        types.EventSchemasI
+	enricher                  enricher.PipelineEnricher
```
atzoum, line 86: Don't we need a list of enrichers instead?

processor/processor.go
```diff
@@ -1482,6 +1501,12 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 		requestIP := gatewayBatchEvent.RequestIP
 		receivedAt := gatewayBatchEvent.ReceivedAt
 
+		// Only if geo enrichment is enabled for deployment and then the source
+		// only then we perform the enrichment on gateway batch event.
+		if proc.config.geoEnrichmentEnabled && proc.geoEnrichmentEnabledForSource(sourceId) {
+			_ = proc.enricher.Enrich(sourceId, &gatewayBatchEvent)
+		}
+
```
atzoum, line 1509: - If geoenrichment is disabled then no need to add the geoenricher in processor.
- The geoenricher itself should know which sources have geoenrichment enabled or not, processor shouldn't know anything about geoenricher, only have a list of enrichers.
```suggestion
		for _, enricher := range proc.enrichers {
			_ = proc.enricher.Enrich(sourceId, &gatewayBatchEvent)
		}
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,191 @@
...
+type geoDB struct {
+	s3Manager filemanager.FileManager
+}
+
+// GetDB simply fetches the database from the upstream located at the key defined
+// in the argument and stores it in the downloadPath provided.
+func (db *geoDB) GetDB(ctx context.Context, key, downloadPath string) error {
+	// If the file already exists, do not go into the loop of downloading
+	// the file again from s3.
+	if _, err := os.Stat(downloadPath); err == nil {
```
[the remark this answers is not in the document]
abhimanyubabbar: Thank you for pointing it out 👍🏼 I'll fix it by fetch temp and rename.

processor/processor.go
```diff
@@ -1482,6 +1501,12 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 		requestIP := gatewayBatchEvent.RequestIP
 		receivedAt := gatewayBatchEvent.ReceivedAt
 
+		// Only if geo enrichment is enabled for deployment and then the source
+		// only then we perform the enrichment on gateway batch event.
+		if proc.config.geoEnrichmentEnabled && proc.geoEnrichmentEnabledForSource(sourceId) {
+			_ = proc.enricher.Enrich(sourceId, &gatewayBatchEvent)
```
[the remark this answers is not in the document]
abhimanyubabbar: I will fix the flow and log the errors

app/apphandlers/embeddedAppHandler.go
```diff
@@ -231,6 +232,23 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
...
 
+	var geoEnricher enricher.PipelineEnricher
+	if config.GetBool("GeoEnrichment.enabled", false) {
+
+		dbProvider, err := enricher.NewGeoDBFetcher(config, a.log)
+		if err != nil {
+			return fmt.Errorf("creating new instance of geo db fetcher: %w", err)
+		}
+
+		geoEnricher, err = enricher.NewGeoEnricher(dbProvider, config, a.log, stats.Default)
```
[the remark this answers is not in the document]
abhimanyubabbar: I'll expose a Close function as well.

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/rudderlabs/rudder-go-kit/filemanager"
+	"github.com/rudderlabs/rudder-go-kit/logger"
+	"github.com/rudderlabs/rudder-go-kit/stats"
+	"github.com/rudderlabs/rudder-server/services/geolocation"
+	"github.com/rudderlabs/rudder-server/utils/types"
+)
+
+type GeoDBFetcher interface {
+	GetDB(ctx context.Context, key, downloadPath string) error
+}
```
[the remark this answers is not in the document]
abhimanyubabbar: We need to have DI on this to allow for testing scenarios which don't hit the S3. Either we have a specific function type or interface works. Let me know if you think it makes sense ?

backend-config/types.go
```diff
@@ -69,6 +69,9 @@ type SourceT struct {
 	WriteKey                   string
 	DgSourceTrackingPlanConfig DgSourceTrackingPlanConfigT
 	Transient                  bool
+	GeoEnrichment              struct {
+		Enabled bool
+	}
```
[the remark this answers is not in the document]
abhimanyubabbar: I can make the change but didn't wanna think too much into the future use-case.

processor/processor.go
abhimanyubabbar: We can always pass an enricher which chains other enrichers or have an array passed here. Given we started with this change and requirements for next enricher aren't much clear, I left it at that only. Wdyt ?

atzoum: What's the point of defining a generic enricher interface if processor knows implementation details, i.e. that this is a geoenricher. Using a list of enrichers simplifies the case where no enricher is enabled too :)

internal/enricher/geolocation.go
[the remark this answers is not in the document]
atzoum: Yes, it is best to inject a function instead, but why do you even need to inject one for tests? If you have made sure during the test execution to create the file where it needs to be, then the original function shouldn't even try to create an S3Manager...

processor/processor.go
abhimanyubabbar: I agree. I have made the updates now.

internal/enricher/geolocation.go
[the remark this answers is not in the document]
abhimanyubabbar: Initially the idea was to download the file everytime nonetheless it exists on the disk or not. I agree the check of looking for the file locally before trying again would solve the issue of having interface for the DB fetcher separately and can be modelled into a simple function instead. 

Let me make the change now !

backend-config/types.go
```diff
@@ -55,7 +55,6 @@ type DestinationT struct {
 	Transformations       []TransformationT
 	IsProcessorEnabled    bool
 	RevisionID            string
-	Deleted               bool
```
abhimanyubabbar: This was added by me when testing something locally !

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (e *geoEnricher) Close() error {
+	if err := e.fetcher.Close(); err != nil {
+		return fmt.Errorf("closing the geo enricher: %w", err)
+	}
+	return nil
+}
+
+func downloadMaxmindDB(ctx context.Context, conf *config.Config, log logger.Logger) (string, error) {
```
abhimanyubabbar, line 145: Here are the changes relating to download maxmind DB

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,235 @@
...
+		e.stats.NewTaggedStat(
+			"proc_geo_enricher_request",
+			stats.CountType,
+			stats.Tags{
+				"sourceID": source.ID,
+				"error":    errType,
+			}).Increment()
+	}()
+
+	if request.RequestIP == "" {
```
atzoum, line 99: According to latest discoveries, the message itself could contain a `context.ip` field which should take precedence over the request's ip.

app/apphandlers/embeddedAppHandler.go
```diff
@@ -359,3 +372,19 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 
 	return g.Wait()
 }
+
+func setupPipelineEnrichers(conf *config.Config, log logger.Logger, stats stats.Stats) ([]enricher.PipelineEnricher, error) {
+	enrichers := make([]enricher.PipelineEnricher, 0)
```
atzoum, line 377:
```suggestion
	var enrichers []enricher.PipelineEnricher
```

processor/processor_geolocation_test.go
```diff
@@ -0,0 +1,305 @@
...
+			WithGeolocationEnabledAtSource(true).
+			WithClientIP(boxfordIP).
+			WithContextIP("").
+			Run(t, func(t *testing.T, event json.RawMessage) {
+				require.NotEmpty(t, gjson.GetBytes(event, "context.geo").Raw, string(event), "geolocation information should be present")
+				require.Equal(t, "Boxford", gjson.GetBytes(event, "context.geo.city").String(), "clientIP should be used by the geolocation service")
+			})
+	})
+
+	t.Run("enabled with invalid context.ip but valid clientIP", func(t *testing.T) {
```
atzoum, line 87: Not sure if this should be the desired behaviour, given that transformer doesn't perform any kind of validation on these fields. So maybe we should just have a precedence order and don't ignore invalid ips

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return "", fmt.Errorf("downloading file with key: %s from bucket: %s and region: %s, err: %w",
+			dbKey,
+			bucket,
+			region,
+			err)
+	}
+
+	// Finally move the downloaded file from previous temp location to new location
+	err = os.Rename(f.Name(), fullpath)
```
atzoum, line 201: Can we rename an open file without closing it? I believe it is about time to add some test coverage for this function. Starting a MINIO server and pushing a file to it is easy.

processor/processor.go
```diff
@@ -1482,14 +1491,33 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
...
+		}
+		statusList = append(statusList, &newStatus)
+
+		source, err := proc.getSourceBySourceID(sourceId)
+		if err != nil {
+			continue
+		}
+
+		for _, enricher := range proc.enrichers {
+			_ = enricher.Enrich(source, &gatewayBatchEvent)
```
atzoum, line 1514: shouldn't we be logging errors?

app/cluster/integration_test.go
```diff
@@ -212,7 +210,9 @@ func TestDynamicClusterManager(t *testing.T) {
 		rsources.NewNoOpService(),
 		destinationdebugger.NewNoOpService(),
 		transformationdebugger.NewNoOpService(),
-		processor.WithFeaturesRetryMaxAttempts(0))
+		[]enricher.PipelineEnricher{},
```
atzoum, line 213:
```suggestion
		nil,
```

internal/enricher/geolocation.go
atzoum, line 145: some doc would be nice, e.g. downloads the file from s3 if not already downloaded

processor/processor.go
```diff
@@ -608,6 +616,7 @@ func (proc *Handle) loadConfig() {
 	// EventSchemas feature. false by default
 	proc.config.enableEventSchemasFeature = config.GetBoolVar(false, "EventSchemas.enableEventSchemasFeature")
 	proc.config.eventSchemaV2Enabled = config.GetBoolVar(false, "EventSchemas2.enabled")
+	proc.config.geoEnrichmentEnabled = config.GetBoolVar(false, "GeoEnrichment.enabled")
```
atzoum, line 619:
```suggestion
```

processor/processor_geolocation_test.go
abhimanyubabbar: Yeah, that's what I am thinking, we shouldn't check if the values are valid or not, just pass the first available value to locate function.

internal/enricher/geolocation.go
abhimanyubabbar: I am confused a bit on this Aris, you mentioning to write tests for `downloadMaxmindDB` or for `os.Rename` separately ?

atzoum: The function

app/cluster/integration_test.go
abhimanyubabbar: I like passing empty array as it tells what argument it is in the calling argument. As per the code the treatment is same. What do you think ?

abhimanyubabbar:
Some open ended questions which I might need some help in answering would be @atzoum :

* Should we attach lookup for empty IP's  / invalid IP's ?
* Should we add empty lookup if the lookup itself failed ?

The underlying question is that should we just add lookup to every event regardless of lookup succeeds or failed as it keeps parity on the event level flowing through the system or we prevent any lookup ?

app/apphandlers/embeddedAppHandler.go
```diff
@@ -359,3 +372,19 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
...
+
+		geoEnricher, err := enricher.NewGeoEnricher(conf, log, stats)
+		if err != nil {
+			return nil, fmt.Errorf("starting geo enrichment process for pipeline: %w", err)
+		}
+		enrichers = append(enrichers, geoEnricher)
+	}
+
+	return enrichers, nil
+}
```
atzoum, line 390: since this function is being used across app handlers, we could move it out of `embeddedAppHandler.go` to `setup.go`

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,259 @@
...
+		dbKey   = conf.GetString("Geolocation.db.key", "geolite2City.mmdb")
+		bucket  = conf.GetString("Geolocation.db.bucket", "rudderstack-geolocation")
+		region  = conf.GetString("Geolocation.db.bucket.region", "us-east-1")
+		timeout = conf.GetInt("Geolocation.db.download.timeout", 100)
+	)
+
+	log.Infof("downloading new geolocation db from key: %s", dbKey)
+	var (
+		tmpDirPath = strings.TrimSuffix(conf.GetString("RUDDER_TMPDIR", "."), "/")
+		baseDIR    = fmt.Sprintf("%s/geolocation", tmpDirPath)
```
atzoum, line 177: I believe this can be simplified
```suggestion
		baseDIR  = path.Join(conf.GetString("RUDDER_TMPDIR", "."), "geolocation")
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,259 @@
...
+		fullpath   = fmt.Sprintf("%s/%s", baseDIR, dbKey)
```
atzoum, line 178: I believe this can be simplified
```suggestion
		baseDIR  = path.Join(conf.GetString("RUDDER_TMPDIR", "."), "geolocation")
		fullpath = path.Join(baseDIR, dbKey)
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,259 @@
...
+	)
+
+	// If the filepath exists return
+	_, err := os.Stat(fullpath)
+	if err == nil {
+		return fullpath, nil
+	}
```
atzoum, line 185: minor, one line less
```suggestion
	if _, err := os.Stat(fullpath); err == nil {
		return fullpath, nil
	}
```

internal/enricher/geolocation.go
```diff
@@ -0,0 +1,264 @@
...
+		// set it with default as map[string]interface{}
+		if _, ok := event["context"]; !ok {
+			event["context"] = map[string]interface{}{}
+		}
+
+		// if the context is other than map[string]interface{}, continue
+		context, ok := event["context"].(map[string]interface{})
+		if !ok {
+			continue
+		}
```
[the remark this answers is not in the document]
abhimanyubabbar: I envisioned this as a missing pre condition to enrichment. The error would be when all the preconditions like `context` and `geo` within the context are present and then the enrichment fails due to some reason.

Wdyt ?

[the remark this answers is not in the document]
abhimanyubabbar: Another point, I am thinking is we can't control the events that are being sent by the customer. If the event treats the context as some other type than map[string]interface{}, treating it as error might generate unnecessary noise in the system.

We can always mention to start with that these are pre-conditions for the event to be enriched with geo information. 

If we start treating it as an error and logging it, I am worried that we might generate logs and noise in system.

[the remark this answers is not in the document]
atzoum: Context is supposed to be a common field of type object, different types shouldn't be common at all
https://www.rudderstack.com/docs/event-spec/standard-events/common-fields/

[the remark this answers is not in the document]
abhimanyubabbar: I agree but we have no checks in place to disallow for events not conforming to a specified schema, so thought of making it as a pre-condition rather than an error. It's a good point to think about and maybe in future we get more clarity to treat it as an error or otherwise.

atzoum (approved):

PR #3867, opened 2023-09-14
chore: cleanup notifier

achettyiitr:
# Description

- Introducing **repository** and **model** for Notifier.
- Renamed it to `notifier` instead of `pgNotifier` since we don't need to expose the implementation details like it is implemented for Postgres.
- Removing globals.
- Removed `notifier.Init()`.
- Introduced `background errgroup` to **gracefully shut down the goroutines** being created during **track batch** and **subscribe**.

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-169/pgnotifier-init-cleanup
- https://linear.app/rudderstack/issue/PIPE-245/notifier-introduce-repository-and-model

## Security


services/notifier/model/notifier.go
lvrach, line 1: I would suggest the following structure:

moving this file to `notifier/notifier.go` containing models part of a the common interface, shared between multiple implementations. Having to access them from model package seems a bit counter intuitive for a library.


move `notifier/notifier.go` -> `notifier/pgnotifer.go` and instead of notifier.New() -> notifier.NewPG() -- since it a specific implementation.

lvrach (approved):

services/notifier/repo/subscriber.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/rudderlabs/rudder-server/services/notifier/model"
+)
+
+// Claim claims a job for a worker.
+func (n *Notifier) Claim(
```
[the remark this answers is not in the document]
achettyiitr: There is not much we are gaining from it. Was trying to break it into separate files depending on roles.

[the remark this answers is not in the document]
achettyiitr: Will move to `notifier.go`

services/notifier/model/notifier.go
achettyiitr: > moving this file to notifier/notifier.go containing models part of a the common interface, shared between multiple implementations. Having to access them from model package seems a bit counter intuitive for a library.

Makes sense. Since it's a package, it should provide a single interface for accessing it. I was initially targeting a repo package and since this model will be shared so moved models to a separate package. Moving everything to notifier package.

achettyiitr: > move notifier/notifier.go -> notifier/pgnotifer.go and instead of notifier.New() -> notifier.NewPG() -- since it a specific implementation
I am not sure whether the consumers need to be aware of the implementation details like whether it is for Postgres, Pulsar or something else.

services/notifier/notifier.go
```diff
@@ -0,0 +1,536 @@
...
+	n.config.user = n.conf.GetString("PGNOTIFIER_DB_USER", "ubuntu")
+	n.config.database = n.conf.GetString("PGNOTIFIER_DB_NAME", "ubuntu")
+	n.config.port = n.conf.GetInt("PGNOTIFIER_DB_PORT", 5432)
+	n.config.password = n.conf.GetString("PGNOTIFIER_DB_PASSWORD", "ubuntu")
+	n.config.sslMode = n.conf.GetString("PGNOTIFIER_DB_SSL_MODE", "disable")
+	n.config.maxAttempt = n.conf.GetInt("PgNotifier.maxAttempt", 3)
+	n.config.maxOpenConnections = n.conf.GetInt("PgNotifier.maxOpenConnections", 20)
+	n.config.shouldForceSetLowerVersion = n.conf.GetBool("SQLMigrator.forceSetLowerVersion", true)
+	n.config.trackBatchInterval = n.conf.GetDuration("PgNotifier.trackBatchIntervalInS", 2, time.Second)
+	n.config.queryTimeout = n.conf.GetDuration("Warehouse.pgNotifierQueryTimeout", 5, time.Minute)
```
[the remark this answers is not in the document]
achettyiitr: This will require creating a standalone struct which I am not sure is required here.

services/notifier/notifier.go
```diff
@@ -0,0 +1,588 @@
...
+	n.logger.Infof("Deleting all jobs for workspace: %s", n.workspaceIdentifier)
+
+	err := n.repo.resetForWorkspace(ctx, n.workspaceIdentifier)
+	if err != nil {
+		return fmt.Errorf("could not reset notifier for workspace: %s: %w", n.workspaceIdentifier, err)
+	}
+	return nil
+}
+
+func (n *Notifier) GetDBHandle() *sql.DB {
```
fracasula, line 318: Considering that you're trying to make this component generic (you even moved it out of the `warehouse` package!), does it make sense to expose such a method? I mean why would someone using a notifier want the internal db handle?

P.S.: no need to make changes here, I'm just not convinced long term this design is optimal.

services/notifier/notifier.go
```diff
@@ -0,0 +1,588 @@
...
+
+	return n.trackBatch(ctx, batchID), nil
+}
+
+// trackBatch tracks the batch and returns a channel of type PublishResponse
+func (n *Notifier) trackBatch(
+	ctx context.Context,
+	batchID string,
+) <-chan *PublishResponse {
+	publishResCh := make(chan *PublishResponse)
```
fracasula, line 352:
```suggestion
	publishResCh := make(chan *PublishResponse, 1)
```

We know beforehand that we'll be sending a single value only, no need to block the sender, we can have it send it to the buffer, close the channel and return right away. Then the reader can consume that value at its own pace.

services/notifier/notifier.go
```diff
@@ -0,0 +1,588 @@
...
+			if err != nil {
+				onUpdate(&PublishResponse{
+					Err: fmt.Errorf("could not get pending count for batch: %s: %w", batchID, err),
+				})
+				return nil
+			} else if count != 0 {
+				continue
+			}
+
+			jobs, err := n.repo.getByBatchID(ctx, batchID)
```
fracasula, line 386: I have a couple questions about this bit. I'm starting from the assumption that this code does the following in this order:
1. get the jobs and keep them in memory
2. delete the jobs, if something goes wrong return
3. send jobs to the consumer

So my questions are:
1. can you describe the workflow of what happens if we delete the jobs and then the service crashes?
2. could we possibly get and delete them in the same query? :thinking: 

No need to make changes here as well, just asking in order to understand.

services/notifier/notifier.go
```diff
@@ -0,0 +1,588 @@
...
+// Subscribe returns a channel of type Job
+func (n *Notifier) Subscribe(
+	ctx context.Context,
+	workerId string,
+	bufferSize int,
+) <-chan *ClaimJob {
+	jobsCh := make(chan *ClaimJob, bufferSize)
+
+	nextPollInterval := func(pollSleep time.Duration) time.Duration {
+		pollSleep = 2*pollSleep + time.Duration(rand.Intn(100))*time.Millisecond
```
fracasula, line 422: I didn't realize that since go 1.20 we didn't have to `Seed` anymore. Just pointing out in case somebody else missed it.

services/notifier/notifier.go
```diff
@@ -0,0 +1,588 @@
...
+
+// Claim claims a job from the notifier queue
+func (n *Notifier) claim(
+	ctx context.Context,
+	workerID string,
+) (*Job, error) {
+	claimStartTime := n.now()
+
+	claimedJob, err := n.repo.claim(ctx, workerID)
+	if err == sql.ErrNoRows {
```
fracasula, line 479: This is going to fail with wrapped errors.

services/notifier/notifier.go
```diff
@@ -0,0 +1,588 @@
...
+	if err != nil {
+		return fmt.Errorf("could not migrate pg_notifier_queue: %w", err)
+	}
+	return nil
+}
+
+// ClearJobs deletes all jobs for the current workspace.
+func (n *Notifier) ClearJobs(ctx context.Context) error {
+	if n.workspaceIdentifier == "" {
+		return nil
```
fracasula, line 306: We don't have a test case for this.

services/notifier/notifier.go
```diff
@@ -0,0 +1,588 @@
...
+	}
+
+	n.logger.Infof("Deleting all jobs for workspace: %s", n.workspaceIdentifier)
+
+	err := n.repo.resetForWorkspace(ctx, n.workspaceIdentifier)
+	if err != nil {
+		return fmt.Errorf("could not reset notifier for workspace: %s: %w", n.workspaceIdentifier, err)
```
fracasula, line 313: We're missing test cases for the repo methods returning errors as well.

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -97,11 +104,14 @@ type WorkerJobRequest struct {
 	StagingUseRudderStorage      bool
 	UniqueLoadGenID              string
 	RudderStoragePrefix          string
-	Output                       []WorkerJobResponse
 	LoadFilePrefix               string // prefix for the load file name
 	LoadFileType                 string
 }
 
+type WorkerJobRequestMetadata struct {
```
fracasula, line 111: Can't we just pass the upload schema instead? What other metadata are we planning to add?

fracasula (commented):
I left a bunch of questions :slightly_smiling_face:

services/notifier/notifier.go
achettyiitr: Good point. Adding CheckHealth to the notifier instead of exposing DB.

services/notifier/notifier.go
achettyiitr: Good point. Using `rand.New(rand.NewSource(time.Now().UnixNano()))`

services/notifier/notifier.go
achettyiitr: Added tests for these.

services/notifier/notifier.go
achettyiitr: Added tests for these.

services/notifier/notifier.go
achettyiitr: > could we possibly get and delete them in the same query? 🤔

Yes.

achettyiitr: > can you describe the workflow of what happens if we delete the jobs and then the service crashes?

Since it's a master which is tracking for a published job, in case it crashes, upon restart it will delete all the jobs and once the UploadJobs are picked up for processing they are again sent to the notifier for processing.

fracasula (approved):
Good stuff :sunglasses: :+1:

PR #3872, opened 2023-09-15
chore: add rudder-cli source and rm binaries

lvrach:
# Description

* Adding [rudderlabs/rudder-cli](https://github.com/rudderlabs/rudder-cli) repository to `rudder-server/cmd`
* Remove binary images
* Change Dockerfile to build rudder-cli instead of coping binaries 

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-256/move-rudder-cli-to-rudder-server

## Security

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -108,11 +108,6 @@ require (
 )
 
 require (
-	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e // indirect
-	github.com/kataras/tablewriter v0.0.0-20180708051242-e063d29b7c23 // indirect
-)
-
-require (
 	cloud.google.com/go v0.110.6 // indirect
 	cloud.google.com/go/compute v1.23.0 // indirect
 	cloud.google.com/go/compute/metadata v0.2.3 // indirect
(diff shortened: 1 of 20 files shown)


achettyiitr (approved):

go.mod
```diff
@@ -103,6 +107,11 @@ require (
 	google.golang.org/protobuf v1.31.0
 )
 
+require (
+	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e // indirect
+	github.com/kataras/tablewriter v0.0.0-20180708051242-e063d29b7c23 // indirect
+)
```
atzoum, line 113: please move these to the indirect group below

go.mod
```diff
@@ -84,6 +87,7 @@ require (
 	github.com/stretchr/testify v1.8.4
 	github.com/tidwall/gjson v1.16.0
 	github.com/tidwall/sjson v1.2.5
+	github.com/urfave/cli v1.22.2
```
atzoum, line 90: can we skip v1 and just use v2?

lvrach: I will create a task to address this.

lvrach: v1 is seems to be supported btw https://github.com/urfave/cli/releases/tag/v1.22.14, so it not an urgent security reason to upgrade.

atzoum (approved):

PR #3871, opened 2023-09-15
chore: use new reloadable config api in router and batchrouter

atzoum:
# Description

- Use new reloadable config api in router and batchrouter
- Drop support for `forceHonorUploadFrequency` since it was never used
- Minor improvements on router tests, when mock calls are missing

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #3874, opened 2023-09-15
chore(batchrouter): honour upload frequency when limitsReached if destination is failing

atzoum:
# Description

When a destination is failing there is no point in keep retrying jobs continuously even if there are a lot of jobs in the queue (limitsReached in jobsdb query).

## Linear Ticket

[PIPE-283](https://linear.app/rudderstack/issue/PIPE-283/honour-upload-frequency-in-batchrouter-when-limitsreached-but)

## Security


router/batchrouter/handle.go
```diff
@@ -245,7 +248,8 @@ func (brt *Handle) getWorkerJobs(partition string) (workerJobs []*DestinationJob
 		if batchDest, ok := destinationsMap[destID]; ok {
 			var processJobs bool
 			brt.lastExecTimesMu.Lock()
-			if limitsReached && !brt.forceHonorUploadFrequency { // if limits are reached, process all jobs regardless of their upload frequency
```
atzoum, line 248: **Note: ** `forceHonorUploadFrequency` is being removed in #3871

lvrach (approved):

router/batchrouter/handle.go
```diff
@@ -116,6 +116,9 @@ type Handle struct {
 	lastExecTimesMu sync.RWMutex
 	lastExecTimes   map[string]time.Time
 
+	failingDestinationsMu sync.RWMutex
+	failingDestinations   map[string]bool
```
achettyiitr, line 120:
```suggestion
	failingDestinations   map[string]struct{}
```
Is it possible to use empty structs here, since currently it is being used as set only?

achettyiitr (approved):

router/batchrouter/handle.go
atzoum: Then we would need to delete from the set when the destination starts succeeding. Both are equally valid solutions in my head. I went with the one that requires the less amount of lines :)

PR #3877, opened 2023-09-17
fix: update error processing of eloqua

ItsSudip:
# Description

We only showed error messages without any details of which fields failed. We are not delivering the error message on the dashboard. So with this task, we are fixing those issues.

## Linear Ticket

> Resolves INT-654

## Security


koladilip (approved):

chandumlg (approved):

chandumlg:
@ItsSudip Please address the format issues.

PR #3879, opened 2023-09-18
feat: new event type - record

chandumlg:
# Description

New event type "record". (To be used to sync rETL records.)

## Linear Ticket

https://linear.app/rudderstack/issue/INT-650/handling-new-event-type-processing-in-pipeline

## Security

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -16,6 +16,7 @@ const (
 	delimiter                 = "<<>>"
 	eventStreamSourceCategory = "eventStream"
 	extractEvent              = "extract"
+	rETLEvent                 = "record"
 	customVal                 = "GW"
 )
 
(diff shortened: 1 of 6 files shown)


gateway/gateway.go
```diff
@@ -16,6 +16,7 @@ const (
 	delimiter                 = "<<>>"
 	eventStreamSourceCategory = "eventStream"
 	extractEvent              = "extract"
+	rETLEvent                 = "record"
```
koladilip, line 19: can record format be used by server side sdks too or do we want to restrict only to rETL?

chandumlg: I would say rETL. Don't see how would it be useful for server side sdks.

achettyiitr (approved):

atzoum (approved):

PR #3882, opened 2023-09-18
feat: filter events support

chandumlg:
# Description

This pr adds the following features:

1. If transformer returns a statusCode: 298 after transforming an event, that event will be treated as a filtered event. Job (corresponding to that event) will be marked as filtered in jobsdb and the same state is forwarded to reporting service.

2. If transformer returns a statusCode: 299 after transforming an event (esp: during router transform), that event will be treated as a successful one and router doesn't need to perform any more actions. Job will be marked as succeeded in jobsdb.

3. Event filter module in processor, instead of silently filtering the events, returns the filtered events with statusCode: 298. Processor will handle them accordingly.

More details: https://www.notion.so/rudderstacks/Introduce-new-code-b-w-transformer-server-to-filter-out-events-8e8faea134474ac08bd85d63d00eceab

## Linear Ticket

https://linear.app/rudderstack/issue/INT-515/implement-server-side-changes-to-accept-new-status-code-from

## Security


processor/processor.go
```diff
@@ -1171,14 +1190,32 @@ func (proc *Handle) updateMetricMaps(
...
 }
 
-func (proc *Handle) getFailedEventJobs(response transformer.Response, commonMetaData *transformer.Metadata, eventsByMessageID map[string]types.SingularEventWithReceivedAt, stage string, transformationEnabled, trackingPlanEnabled bool) ([]*jobsdb.JobT, []*types.PUReportedMetric, map[string]int64) {
-	failedMetrics := make([]*types.PUReportedMetric, 0)
+func (proc *Handle) getNonSuccessfulMetrics(response transformer.Response, commonMetaData *transformer.Metadata, eventsByMessageID map[string]types.SingularEventWithReceivedAt, stage string, transformationEnabled, trackingPlanEnabled bool) *NonSuccessfulTransformationMetrics {
+	m := &NonSuccessfulTransformationMetrics{}
+
+	var filtered, failed []transformer.TransformerResponse
+	for _, event := range response.FailedEvents {
+		if event.StatusCode == types.FilterEventCode {
```
atzoum, line 1198: Shouldn't the filtered event code be returned by the user transformer as well in case the event is dropped?
IMO to be confident we need integration tests, with an actual transformer, reporting enabled (but not sending reports to a remote server) to verify that filtered events are being recorded in reporting in various stages of the processor's pipeline:
- 1 scenario for message filter
- 1 scenario for destination transformation
- 1 scenario for user transformation
- 1 scenario for tracking plan(?)

cc @pChondros

router/worker.go
```diff
@@ -553,6 +554,12 @@ func (w *worker) processDestinationJobs() {
 				}
 				errorAt = routerutils.ERROR_AT_TF
 			}
+		} else if destinationJob.StatusCode == utilTypes.FilterEventCode || destinationJob.StatusCode == utilTypes.SuppressEventCode {
+			respStatusCode = destinationJob.StatusCode
+			respBody = destinationJob.Error
+			if respBody == "" {
+				respBody = utilTypes.GetFallBackErrorMessage(respStatusCode)
+			}
```
atzoum, line 562: instead of introducing yet another branch, we could "reuse" the existing `else` branch, with the hope of eventually doing something better about the `errorAt` flag, which is used only for stats purposes:

```go
respStatusCode = destinationJob.StatusCode
respBody = destinationJob.Error
switch destinationJob.StatusCode {
case utilTypes.FilterEventCode:
	if respBody == "" {
		respBody = "Event filtered"
	}
case utilTypes.SuppressEventCode:
	if respBody == "" {
		respBody = "Event handled by transformer"
	}
default:
	errorAt = routerutils.ERROR_AT_TF
}
```

processor/processor.go
```diff
@@ -291,13 +304,16 @@ func (proc *Handle) newDestinationTransformationStat(sourceID, workspaceID, tran
 	errTags := misc.CopyStringMap(tags)
 	errTags["error"] = "true"
 	numOutputFailedEvents := proc.statsFactory.NewTaggedStat("proc_transform_stage_out_count", stats.CountType, errTags)
+	errTags["error"] = "filtered"
```
Sidddddarth, line 307: won't `filtered` will be propagated to the above stat as well..?

processor/trackingplan.go
```diff
@@ -113,19 +114,21 @@ func (proc *Handle) validateEvents(groupedEventsBySourceId map[SourceIDT][]trans
...
+		nonSuccessMetrics := proc.getNonSuccessfulMetrics(response, commonMetaData, eventsByMessageID, transformer.TrackingPlanValidationStage, false, true)
 
 		validationStat.numValidationSuccessEvents.Count(len(eventsToTransform))
-		validationStat.numValidationFailedEvents.Count(len(failedJobs))
+		validationStat.numValidationFailedEvents.Count(len(nonSuccessMetrics.failedJobs))
+		validationStat.numValidationFilteredEvents.Count(len(nonSuccessMetrics.filteredJobs))
 		proc.logger.Debug("Validation output size", len(eventsToTransform))
 
-		validatedErrorJobs = append(validatedErrorJobs, failedJobs...)
+		validatedErrorJobs = append(validatedErrorJobs, nonSuccessMetrics.failedJobs...)
```
Sidddddarth, line 124: why not add filtered jobs here..?

utils/types/types.go
```diff
@@ -11,6 +11,22 @@ import (
...
+func GetFallBackErrorMessage(code int) string {
+	switch code {
+	case FilterEventCode:
+		return "Event filtered"
+	case SuppressEventCode:
+		return "Event handled by transformer"
+	}
+
+	return ""
+}
```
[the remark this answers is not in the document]
chandumlg: Got rid of this function and used the strings directly as suggested in the other comment below.

router/worker.go
chandumlg: Cleaned up.

processor/processor.go
```diff
@@ -1171,14 +1196,32 @@ func (proc *Handle) updateMetricMaps(
...
+	m := &NonSuccessfulTransformationMetrics{}
+
+	var filtered, failed []transformer.TransformerResponse
+	for _, event := range response.FailedEvents {
+		if event.StatusCode == types.FilterEventCode {
+			filtered = append(filtered, event)
+		} else {
+			failed = append(failed, event)
+		}
+	}
```
atzoum, line 1209: **minor:** doing the same with less lines
```suggestion
	grouped := lo.GroupBy(response.FailedEvents, func(event transformer.TransformerResponse) bool { return event.StatusCode == types.FilterEventCode })
	filtered, failed := grouped[true], grouped[false]
```

processor/processor.go
```diff
@@ -1187,26 +1230,26 @@ func (proc *Handle) getFailedEventJobs(response transformer.Response, commonMeta
 		)
 		payload, err := jsonfast.Marshal(messages)
```
atzoum, line 1231: Any idea why we are storing a single job containing an array of failed payloads to proc_errors? @cisse21?
And we are using this single job to calculate dropped events as well. Wouldn't that cause skewed rsources metrics @Sidddddarth?

router/worker.go
```diff
@@ -806,7 +821,7 @@ func (w *worker) sendRouterResponseCountStat(status *jobsdb.JobStatusT, destinat
 	destinationTag := misc.GetTagName(destination.ID, destination.Name)
 	var alert bool
 	alert = w.allowRouterAbortedAlert(errorAt)
-	if status.JobState == jobsdb.Succeeded.State {
+	if status.JobState == jobsdb.Succeeded.State || status.JobState == jobsdb.Filtered.State {
```
atzoum, line 824: Any adaptations needed in `sendEventDeliveryStat`?

atzoum (approved):

router/worker.go
chandumlg: IMO, No.

router response counts -> responses router received for an event. respStatusCode label will reveal about the status of the delivery.

event delivery stats -> stats for successfully delivered events (count & time spent by the event in the pipeline since it is received.)

processor/transformer/transformer.go
```diff
@@ -444,12 +444,21 @@ func (trans *handle) doPost(ctx context.Context, rawJSON []byte, url, stage stri
 
 	err := backoff.RetryNotify(
 		func() error {
-			var reqErr error
 			requestStartTime := time.Now()
 
-			trace.WithRegion(ctx, "request/post", func() {
```
chandumlg, line 450: Took some liberty and got rid of this trace. 
a. This is causing lint issue
b. We are not using anyways

psrikanth88 (approved):

PR #3885, opened 2023-09-19
chore: don't create custom destination clients for disabled destination

atzoum:
# Description

There is no point in trying to create clients for custom destinations which are disabled. Even if there are jobs for such destinations in the database, router will drain them before reaching to the point of needing to use the client.

## Linear Ticket

[PIPE-295](https://linear.app/rudderstack/issue/PIPE-295/dont-create-custom-destination-clients-for-disabled-destination)

## Security


achettyiitr (approved):

fracasula (approved):
We don't have a test for this one, feel free to add a case if it's a quick one.

PR #3890, opened 2023-09-20
feat: add filtered and added state for user transformer reports

psrikanth88:
# Description

Introduce filtered and added state in reports for user transformation PU. Fixes OBS-125

## Linear Ticket

https://linear.app/rudderstack/issue/DAT-339/surface-user-transformation-drops-in-webapp

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1372,6 +1372,107 @@ func getDiffMetrics(inPU, pu string, inCountMetadataMap map[string]MetricMetadat
 	return diffMetrics
 }
 
+func generateFilteredAndAddedMetrics(inPU, pu string, messageIds []string, messageIdKeyMap map[string]string, inCountMetadataMap map[string]MetricMetadata, outputEvents []transformer.TransformerEvent, failedEvents []transformer.TransformerResponse) ([]*types.PUReportedMetric, []*types.PUReportedMetric) {
+	successMessageIdMap := make(map[string]int)
+	failedMessageIdMap := make(map[string]int)
+	for _, event := range outputEvents {
+		// handle transformBatch user transformation
+		// consider it as success if messageId is present in Metadata.MessageIDs of output event
+		// but do not increment to not count it as newly added event
+		if event.Metadata.MessageID == "" {
+			for _, messageId := range event.Metadata.MessageIDs {
+				if _, ok := successMessageIdMap[messageId]; !ok {
+					successMessageIdMap[messageId] = 0
+				}
+			}
+			continue
+		}
+
+		if _, ok := successMessageIdMap[event.Metadata.MessageID]; !ok {
+			successMessageIdMap[event.Metadata.MessageID] = 0
+		}
+		successMessageIdMap[event.Metadata.MessageID] += 1
+	}
+	for _, event := range failedEvents {
+		// handle transformBatch user transformation
+		// consider it as failed if messageId is present in Metadata.MessageIDs of failed event
+		if event.Metadata.MessageID == "" {
+			for _, messageId := range event.Metadata.MessageIDs {
+				if _, ok := failedMessageIdMap[messageId]; !ok {
+					failedMessageIdMap[messageId] = 0
+				}
+				failedMessageIdMap[messageId] += 1
+			}
+			continue
+		}
+
+		if _, ok := failedMessageIdMap[event.Metadata.MessageID]; !ok {
+			failedMessageIdMap[event.Metadata.MessageID] = 0
+		}
+		failedMessageIdMap[event.Metadata.MessageID] += 1
+	}
+
+	filteredMetricsMap := make(map[string]*types.PUReportedMetric)
+	addedMetricsMap := make(map[string]*types.PUReportedMetric)
+	for _, messageId := range messageIds {
+		outCount, successOk := successMessageIdMap[messageId]
+		_, failedOk := failedMessageIdMap[messageId]
+		noMessageIdInOutput := !successOk && !failedOk
+		multipleMessageIdInOutput := successOk && outCount > 1
+
+		if noMessageIdInOutput || multipleMessageIdInOutput {
+			key := messageIdKeyMap[messageId]
+			metricMetadata := inCountMetadataMap[key]
+			var eventName, eventType string
+			splitKey := strings.Split(key, MetricKeyDelimiter)
+			if len(splitKey) < 5 {
+				eventName = ""
+				eventType = ""
+			} else {
+				eventName = splitKey[3]
+				eventType = splitKey[4]
+			}
+			metricKey := strings.Join([]string{metricMetadata.sourceID, metricMetadata.destinationID, metricMetadata.sourceTaskRunID, metricMetadata.sourceJobID, metricMetadata.sourceJobRunID, metricMetadata.sourceDefinitionID, metricMetadata.destinationDefinitionID, metricMetadata.sourceCategory, metricMetadata.transformationID, metricMetadata.transformationVersionID, metricMetadata.trackingPlanID, fmt.Sprintf("%d", metricMetadata.trackingPlanVersion), eventName, eventType}, MetricKeyDelimiter)
+
+			if noMessageIdInOutput {
+				if _, ok := filteredMetricsMap[metricKey]; !ok {
+					filteredMetricsMap[metricKey] = &types.PUReportedMetric{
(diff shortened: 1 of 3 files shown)


processor/processor.go
```diff
@@ -2257,6 +2362,8 @@ func (proc *Handle) transformSrcDest(
 				reportMetrics = append(reportMetrics, successMetrics...)
 				reportMetrics = append(reportMetrics, failedMetrics...)
 				reportMetrics = append(reportMetrics, diffMetrics...)
+				reportMetrics = append(reportMetrics, filteredMetrics...)
+				reportMetrics = append(reportMetrics, addedMetrics...)
```
atzoum, line 2366: AFAIK diff metrics were supposed to capture these kind of differences. @Sidddddarth do we need diff metrics anymore?

processor/processor.go
```diff
@@ -1372,6 +1372,107 @@ func getDiffMetrics(inPU, pu string, inCountMetadataMap map[string]MetricMetadat
...
+		// consider it as success if messageId is present in Metadata.MessageIDs of output event
+		// but do not increment to not count it as newly added event
+		if event.Metadata.MessageID == "" {
+			for _, messageId := range event.Metadata.MessageIDs {
+				if _, ok := successMessageIdMap[messageId]; !ok {
+					successMessageIdMap[messageId] = 0
+				}
+			}
+			continue
+		}
```
atzoum, line 1389: I still fail to understand the logic behind it, can you please elaborat?

processor/processor.go
```diff
@@ -1372,6 +1372,107 @@ func getDiffMetrics(inPU, pu string, inCountMetadataMap map[string]MetricMetadat
...
+
+		if _, ok := successMessageIdMap[event.Metadata.MessageID]; !ok {
+			successMessageIdMap[event.Metadata.MessageID] = 0
+		}
```
atzoum, line 1393: This is not really needed, the zero value for ints is already 0!
```suggestion
```

processor/processor.go
```diff
@@ -1372,6 +1372,107 @@ func getDiffMetrics(inPU, pu string, inCountMetadataMap map[string]MetricMetadat
...
+		successMessageIdMap[event.Metadata.MessageID] += 1
+	}
+	for _, event := range failedEvents {
+		// handle transformBatch user transformation
+		// consider it as failed if messageId is present in Metadata.MessageIDs of failed event
+		if event.Metadata.MessageID == "" {
+			for _, messageId := range event.Metadata.MessageIDs {
+				if _, ok := failedMessageIdMap[messageId]; !ok {
+					failedMessageIdMap[messageId] = 0
+				}
```
atzoum, line 1403:
```suggestion
```

processor/processor.go
```diff
@@ -1372,6 +1372,107 @@ func getDiffMetrics(inPU, pu string, inCountMetadataMap map[string]MetricMetadat
...
+				failedMessageIdMap[messageId] += 1
+			}
+			continue
+		}
+
+		if _, ok := failedMessageIdMap[event.Metadata.MessageID]; !ok {
+			failedMessageIdMap[event.Metadata.MessageID] = 0
+		}
```
atzoum, line 1411:
```suggestion
```

processor/processor.go
```diff
@@ -1372,6 +1372,107 @@ func getDiffMetrics(inPU, pu string, inCountMetadataMap map[string]MetricMetadat
...
 }
 
+func generateFilteredAndAddedMetrics(inPU, pu string, messageIds []string, messageIdKeyMap map[string]string, inCountMetadataMap map[string]MetricMetadata, outputEvents []transformer.TransformerEvent, failedEvents []transformer.TransformerResponse) ([]*types.PUReportedMetric, []*types.PUReportedMetric) {
+	successMessageIdMap := make(map[string]int)
+	failedMessageIdMap := make(map[string]int)
+	for _, event := range outputEvents {
+		// handle transformBatch user transformation
+		// consider it as success if messageId is present in Metadata.MessageIDs of output event
+		// but do not increment to not count it as newly added event
+		if event.Metadata.MessageID == "" {
```
atzoum, line 1382: @Sidddddarth is introducing a utility method in #3852. Wouldn't this simplify code?

```go
func (m Metadata) GetMessagesIDs() []string {
	if len(m.MessageIDs) > 0 {
		return m.MessageIDs
	}
	return []string{m.MessageID}
}
```

utils/types/reporting_types.go
```diff
@@ -21,7 +21,9 @@ const (
 )
 
 var (
-	DiffStatus = "diff"
+	DiffStatus     = "diff"
+	FilteredStatus = "filtered"
```
atzoum, line 25: @chandumlg is introducing a filtered job state in jobsdb, we could use the jobsdb literal instead here to avoid proliferation of equal constants

processor/processor.go
```diff
@@ -1372,6 +1372,107 @@ func getDiffMetrics(inPU, pu string, inCountMetadataMap map[string]MetricMetadat
 	return diffMetrics
 }
 
+func generateFilteredAndAddedMetrics(inPU, pu string, messageIds []string, messageIdKeyMap map[string]string, inCountMetadataMap map[string]MetricMetadata, outputEvents []transformer.TransformerEvent, failedEvents []transformer.TransformerResponse) ([]*types.PUReportedMetric, []*types.PUReportedMetric) {
```
atzoum, line 1375: A common fallacy in processor right now is the assumption that messageIds in a batch are unique. This doesn't stand true if deduplication is not enabled.

processor/processor.go
```diff
@@ -1372,6 +1372,107 @@ func getDiffMetrics(inPU, pu string, inCountMetadataMap map[string]MetricMetadat
...
+		outCount, successOk := successMessageIdMap[messageId]
+		_, failedOk := failedMessageIdMap[messageId]
+		noMessageIdInOutput := !successOk && !failedOk
+		multipleMessageIdInOutput := successOk && outCount > 1
+
+		if noMessageIdInOutput || multipleMessageIdInOutput {
+			key := messageIdKeyMap[messageId]
+			metricMetadata := inCountMetadataMap[key]
+			var eventName, eventType string
+			splitKey := strings.Split(key, MetricKeyDelimiter)
```
atzoum, line 1427: it is about time to prepare a typed reportingKey, so that we don't need to repeat ourselves and perform complex parsing calculations, e.g.

```go
type metricKey struct {
	sourceID       string
	destinationID  string
	sourceJobRunID string
	eventName      string
	eventType      string
}
```

processor/processor.go
```diff
@@ -1372,6 +1372,107 @@ func getDiffMetrics(inPU, pu string, inCountMetadataMap map[string]MetricMetadat
...
+			if len(splitKey) < 5 {
+				eventName = ""
+				eventType = ""
+			} else {
+				eventName = splitKey[3]
+				eventType = splitKey[4]
+			}
+			metricKey := strings.Join([]string{metricMetadata.sourceID, metricMetadata.destinationID, metricMetadata.sourceTaskRunID, metricMetadata.sourceJobID, metricMetadata.sourceJobRunID, metricMetadata.sourceDefinitionID, metricMetadata.destinationDefinitionID, metricMetadata.sourceCategory, metricMetadata.transformationID, metricMetadata.transformationVersionID, metricMetadata.trackingPlanID, fmt.Sprintf("%d", metricMetadata.trackingPlanVersion), eventName, eventType}, MetricKeyDelimiter)
```
atzoum, line 1435: Same comment about concatenated strings vs typed keys

processor/processor.go
Sidddddarth: considering that `filtered` means the same thing, `diff` can be removed.
Also, request to reconsider the name `filtered`, would `dropped` be better suited?
And events can be dropped in other areas of the processor as well(e.g.: event filter), not just user transforms, so this could be used at all three stages of the transform(UT, event filter, DT). In fact at the destination filter too.

processor/processor.go
psrikanth88: done

processor/processor.go
psrikanth88: done

processor/processor.go
psrikanth88: Using empty messageID as proxy that output was from transformBatch transformation. Can be done similarly by checking length of GetMessagesIDs() but felt former was more explicit as that is the API contract between server - transformer

processor/processor.go
psrikanth88: for the transformer report - we need the input and output numbers. `succeeded` gives output, input can be derived as `succeeded - diff` since we cannot always use `filtered` and `added` (input = succeeded - added +filtered) as they have nuances around transformBatch

psrikanth88: I am open to name change. Did not want to use `dropped` from product perspective as that has negative connotation when showing to user.

processor/processor.go
psrikanth88: @atzoum @Sidddddarth any suggestions around what would be the unique composite key - that is present in both input and output metadata of transformer. should we do `messageId + jobId` and add jobIds similar to messageIds in transformer response for transformBatch case.

PR #3896, opened 2023-09-21
exp: use subject cache for jobsdb

Sidddddarth:
# Description

subject cache for jobsdb.

Motivation: to reduce the amount of adaptations to jobsdb's cache every time the query pattern changes. And also something that has lesser cognitive load. imo this would simplify(at least cache adaptations - simply add the required tokens) if we were to add more query filters, ex.: both source and destination IDs, `jobRunID`, `user_id` etc.

Idea is that each time a consumer of jobsdb queries for jobs -> it's with a subset of all the query parameters possible. We use this subset and wildcards to create subjects. Any write to jobsdb(store/status update) if contains enough information would then be a subset of the subject used during read. The subject is of the format - `dsIndex.WorkspaceID.customVal.SourceID.DestID.State`
For example:
1. read query subject: `1.*.GA.*.*.Unprocessed`
    a write that is a subset of above subject would be: `1.WS-1.GA.SRC-1.DST-1.Unprocessed`(a store job call)

2. read query subject: `3.*.GA.*.dest1.Failed`
    a write that is a subset of above subject would be: `3.WS-1.GA.SRC-99.dest1.Failed` (a job status update call)

Writes when updating the cache can only pass subject literals(no wildcards).

Introduces package `tokens` in `/internal`(credits: nats-server). This is the package that tells us if one subject is a subset of another.

TODO: implement a ttl for the cache.

Other changes:
1. changed `jobsdb.GetQueryParams.CustomVal` from `[]string` to `string`. Because never saw a query by multiple `customVal`s. With more isolation, this is much less likelier.
2. changed `OR` clause to `AND` in parameters -> query string.
3. changed `jobsbd.updatestatus` signature to accept `customVal string` instead of `customValFilters []string` - because again - haven't seen multiple `customVal`s being passed anyway.
4. no cache business at all for the lastDS in a read jobsdb.

We'd have to enforce that any writes that go into jobsdb(store/status update) must have enough information to form a subject literal(no wildcard allowed in writes)
E.g.: populating `jobsdb.JobStatusT.JobParameters` with it's job's `job.Parameters`, set `status.WorkspaceId = job.WorkspaceId`. And making sure update status calls are grouped by one `customVal`.

## Linear Ticket

< Replace with Linear Link >

## Security

## Summary by CodeRabbit

- **New Features**
  - Search functionality now available with a new search bar for improved navigation and user experience.
- **Enhancements**
  - Streamlined job status update process with simplified filtering.
- **Bug Fixes**
  - Corrected syntax errors in job status update parameters.
- **Refactor**
  - Consolidated custom value filters into a single field for more efficient querying.
- **Performance Improvements**
  - Optimized job status update logic for faster processing.
- **Documentation**
  - Updated documentation to reflect changes in job status update methods and query parameters.

### DIFF
--- a/internal/tokens/tokens.go
+++ b/internal/tokens/tokens.go
@@ -1,0 +1,71 @@
+package tokens
+
+import "strings"
+
+const (
+	fwc   = '>'
+	btsep = '.'
+	pwc   = '*'
+)
+
+// SubjectIsSubsetMatch checks if the subject is a subset of the test
+func SubjectIsSubsetMatch(subject, test string) bool {
+	tsa := [32]string{}
+	tts := tokenizeSubjectIntoSlice(tsa[:0], subject)
+	return isSubsetMatch(tts, test)
+}
+
+func isSubsetMatch(tokens []string, test string) bool {
+	tsa := [32]string{}
+	tts := tokenizeSubjectIntoSlice(tsa[:0], test)
+	return isSubsetMatchTokenized(tokens, tts)
+}
+
+func tokenizeSubjectIntoSlice(tts []string, subject string) []string {
+	start := 0
+	for i := 0; i < len(subject); i++ {
+		if subject[i] == btsep {
+			tts = append(tts, subject[start:i])
+			start = i + 1
+		}
+	}
+	tts = append(tts, subject[start:])
+	return tts
+}
+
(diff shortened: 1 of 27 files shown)


internal/tokens/tokens.go
cisse21, line 1: This seems to be a candidate for rudder-go-kit

PR #3900, opened 2023-09-21
chore: warehouse schema race and cleanup

fracasula:
# Description

Sub issue for the Warehouse integration tests with `-race`.

The plan is to incrementally reintroduce the changes in smaller PRs to better test the changes in isolation. This is the last PR that is going to reintroduce the changes we reverted [here](https://github.com/rudderlabs/rudder-server/pull/3825).

In this Pull Request we are moving the `Schema` into its own package and we're making it safe for concurrent use.
The Warehouse Integration tests are going to be covered by `-race` but **not the unit ones**. For those we are going to tackle the issues we found in other PRs.

In order to enable the `-race` in the unit test we have to change the `test-run` recipe in the `Makefile` as such:

```
test-run: ## Run all unit tests
ifeq ($(filter 1,$(debug) $(RUNNER_DEBUG)),)
	$(eval TEST_CMD = SLOW=0 gotestsum --format pkgname-and-test-fails --)
	$(eval TEST_OPTIONS = -p=1 -v -failfast -shuffle=on -coverprofile=profile.out -covermode=atomic -coverpkg=./... -vet=all --timeout=15m)
else
	$(eval TEST_CMD = SLOW=0 go test)
	$(eval TEST_OPTIONS = -p=1 -v -failfast -shuffle=on -coverprofile=profile.out -covermode=atomic -coverpkg=./... -vet=all --timeout=15m)
endif
ifdef package
    ifeq ($(findstring warehouse,$(package)),warehouse)
	    $(TEST_CMD) $(TEST_OPTIONS) -race $(package) && touch $(TESTFILE) || true
    else
	    $(TEST_CMD) $(TEST_OPTIONS) $(package) && touch $(TESTFILE) || true
    endif
else ifdef exclude
	$(eval FILES = `go list ./... | egrep -iv '$(exclude)'`)
	$(TEST_CMD) -count=1 $(TEST_OPTIONS) $(FILES) && touch $(TESTFILE) || true
else
	$(TEST_CMD) -count=1 $(TEST_OPTIONS) ./... && touch $(TESTFILE) || true
endif
```

### It is recommended to review this PR commit by commit to simplify the review.

Panos' [E2E Tests run](https://github.com/rudderlabs/e2e/actions/runs/6276197143).

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-331/warehouseschema-race-and-cleanup) >

## Security

### DIFF
--- a/warehouse/integrations/clickhouse/clickhouse.go
+++ b/warehouse/integrations/clickhouse/clickhouse.go
@@ -892,7 +892,7 @@ func (ch *Clickhouse) AddColumns(ctx context.Context, tableName string, columnsI
 }
 
 func (ch *Clickhouse) CreateSchema(ctx context.Context) error {
-	if len(ch.Uploader.GetSchemaInWarehouse()) > 0 {
+	if !ch.Uploader.IsWarehouseSchemaEmpty() {
 		return nil
 	}
 
--- a/warehouse/integrations/clickhouse/clickhouse.go
+++ b/warehouse/integrations/clickhouse/clickhouse.go
@@ -770,7 +770,7 @@ func (ch *Clickhouse) schemaExists(ctx context.Context, schemaName string) (exis
 	sqlStatement := "SELECT count(*) FROM system.databases WHERE name = ?"
 	err = ch.DB.QueryRowContext(ctx, sqlStatement, schemaName).Scan(&count)
 	// ignore err if no results for query
-	if err == sql.ErrNoRows {
+	if errors.Is(err, sql.ErrNoRows) {
 		err = nil
 	}
 	exists = count > 0
@@ -977,7 +977,8 @@ func (ch *Clickhouse) FetchSchema(ctx context.Context) (model.Schema, model.Sche
 		return schema, unrecognizedSchema, nil
 	}
 	if err != nil {
-		if clickhouseErr, ok := err.(*clickhouse.Exception); ok && clickhouseErr.Code == 81 {
+		var clickhouseErr *clickhouse.Exception
+		if errors.As(err, &clickhouseErr) && clickhouseErr.Code == 81 {
 			return schema, unrecognizedSchema, nil
 		}
 		return nil, nil, fmt.Errorf("fetching schema: %w", err)
--- a/warehouse/schema/schema.go
+++ b/warehouse/schema/schema.go
@@ -1,8 +1,7 @@
-package warehouse
+package schema
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"reflect"
 	"regexp"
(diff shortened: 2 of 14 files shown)


warehouse/schema/schema.go
```diff
@@ -67,44 +66,48 @@ func NewSchema(
...
-	sh.localSchema = updatedSchema
+	sh.localSchemaMu.RLock()
+	consolidatedSchema = consolidateWarehouseSchema(consolidatedSchema, sh.localSchema)
+	consolidatedSchema = overrideUsersWithIdentifiesSchema(consolidatedSchema, sh.warehouse.Type, sh.localSchema)
+	sh.localSchemaMu.RUnlock()
 
-	return nil
-}
+	consolidatedSchema = enhanceDiscardsSchema(consolidatedSchema, sh.warehouse.Type)
+	consolidatedSchema = enhanceSchemaWithIDResolution(consolidatedSchema, sh.isIDResolutionEnabled(), sh.warehouse.Type)
```
achettyiitr, line 95: Don't we need write lock here?

warehouse/schema/schema.go
```diff
@@ -130,12 +133,138 @@ func (sh *Schema) fetchSchemaFromWarehouse(ctx context.Context, repo fetchSchema
...
+	if err != nil {
+		return false, fmt.Errorf("fetching schema from local: %w", err)
+	}
+	if err := sh.FetchSchemaFromWarehouse(ctx, fetchSchemaRepo); err != nil {
+		return false, fmt.Errorf("fetching schema from warehouse: %w", err)
+	}
+
+	sh.schemaInWarehouseMu.RLock()
+	schemaChanged := sh.hasSchemaChanged(localSchema)
+	defer sh.schemaInWarehouseMu.RUnlock()
```
achettyiitr, line 158: [Optional] Maybe we can simplify this. Currently, we are using `RUnlock` in a defer block only if schemaChanged.
```suggestion
	sh.schemaInWarehouseMu.RLock()
	defer sh.schemaInWarehouseMu.RUnlock()
	schemaChanged := sh.hasSchemaChanged(localSchema)
```

warehouse/schema/schema.go
fracasula: The lock is to protect `localSchema` only here. `consolidatedSchema` is a brand new map and we're not copying any reference in it that might be coming from `localSchema` so we should be cool. Thoughts?

warehouse/schema/schema.go
fracasula: Good catch! :+1: I think I know the keyboard shortcut responsible for this :sweat_smile: Thanks :+1:

warehouse/schema/schema.go
achettyiitr: Sorry, I missed this part. `consolidatedSchema` creates a brand new map.

warehouse/integrations/clickhouse/clickhouse.go
```diff
@@ -770,7 +770,7 @@ func (ch *Clickhouse) schemaExists(ctx context.Context, schemaName string) (exis
 	sqlStatement := "SELECT count(*) FROM system.databases WHERE name = ?"
 	err = ch.DB.QueryRowContext(ctx, sqlStatement, schemaName).Scan(&count)
 	// ignore err if no results for query
-	if err == sql.ErrNoRows {
+	if errors.Is(err, sql.ErrNoRows) {
```
cisse21, line 773: Just wondering if we can move to `ExecContext` instead of `QueryRowContext` to avoid the no rows check?

cisse21 (approved):

PR #3904, opened 2023-09-22
feat: add gzip bodyFormat support

Gauravudia:
# Description

Introducing the new `GZIP` body format for gzip compression. One can now send JSON data from other services (like the transformer) which will be compressed and sent to downstream destinations

## Linear Ticket

https://linear.app/rudderstack/issue/INT-646/mixpanel-support-gzip-compression

## Security


router/network_test.go
```diff
@@ -149,6 +149,7 @@ var _ = Describe("Network", func() {
 				"FORM": map[string]interface{}{},
 				"JSON": map[string]interface{}{},
 				"XML":  map[string]interface{}{},
+				"GZIP": map[string]interface{}{},
```
koladilip, line 152: need to add more tests, to test gzip is happening correctly or not

koladilip (approved):

router/network.go
```diff
@@ -157,6 +175,9 @@ func (network *netHandle) SendPost(ctx context.Context, structData integrations.
 		}
 
 		req.Header.Add("User-Agent", "RudderLabs")
+		if bodyFormat == "GZIP" {
+			req.Header.Add("Content-Encoding", "gzip")
+		}
```
ItsSudip, line 180: We are already getting the headers with the transform event. Can't we skip this step?

Gauravudia: Yes, removed the redundant header.

ItsSudip (approved):

router/network.go
```diff
@@ -124,6 +126,22 @@ func (network *netHandle) SendPost(ctx context.Context, structData integrations.
...
 				payload = strings.NewReader(formValues.Encode())
+			case "GZIP":
+				strValue, ok := bodyValue["payload"].(string)
+				if !ok {
+					return &utils.SendPostResponse{
+						StatusCode:   400,
+						ResponseBody: []byte("400 Unable to construct gzip payload. Unexpected transformer response"),
+					}
+				}
+				zippedData, err := base64.StdEncoding.DecodeString(strValue)
```
lvrach, line 137: why do we need to base64 ?

Gauravudia: Transmitting GZIP-compressed binary data from the transformer to the server can introduce compatibility issues. By encoding the payload at the transformer's end and decoding it at the server's end, ensures seamless data transfer without compatibility concerns.
cc- @koladilip

atzoum: Do we really need to transmit binary data from transformer to rudder-server? What is the use case? If we can skip gzip between transformer-server communication it would be cleaner.

Gauravudia: Use case is to send the gzipped data to mixpanel for a specific batch endpoint [/import ](https://developer.mixpanel.com/reference/import-events)

koladilip: @Gauravudia we can do that too.

Gauravudia: @atzoum Removed the base64 and added the gzip compression in server

router/network.go
```diff
@@ -157,6 +187,9 @@ func (network *netHandle) SendPost(ctx context.Context, structData integrations.
 		}
 
 		req.Header.Add("User-Agent", "RudderLabs")
+		if bodyFormat == "GZIP" {
+			req.Header.Add("Content-Encoding", "gzip")
+		}
```
atzoum, line 192: instead of adding another if clause here, we could use a headers map and fill it per case, e.g.

```go
var payload io.Reader
headers := map[string]string{"User-Agent": "RudderLabs"}
...
switch bodyFormat {
	case "GZIP":
		headers["Content-Encoding"] = "gzip"
...
for key, val := range headers {
	req.Header.Add(key, val)
}
...
```

router/network.go
```diff
@@ -124,6 +126,34 @@ func (network *netHandle) SendPost(ctx context.Context, structData integrations.
...
+						StatusCode:   400,
+						ResponseBody: []byte("400 Unable to parse json list. Unexpected transformer response"),
+					}
+				}
+				var buf bytes.Buffer
+				zw := gzip.NewWriter(&buf)
+				defer zw.Close()
+
+				_, err := zw.Write([]byte(strValue))
+				if err != nil {
```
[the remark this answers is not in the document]
Gauravudia: incorporated the review suggestions

router/network.go
```diff
@@ -124,6 +126,34 @@ func (network *netHandle) SendPost(ctx context.Context, structData integrations.
...
+					return &utils.SendPostResponse{
+						StatusCode:   400,
+						ResponseBody: []byte("400 Unable to compress data. Unexpected response"),
+					}
+				}
+				err = zw.Flush()
+				if err != nil {
```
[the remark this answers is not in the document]
Gauravudia: incorporated the review suggestions

router/network.go
Gauravudia: incorporated the review suggestions

atzoum (approved):

PR #3912, opened 2023-09-26
refactor: reporting feature

atzoum:
# Description

Preparatory work for being able to collect additional error metadata fields for the purpose of replaying errors.

- Introduce a new struct for including error message details in reported data (`types.FailedMessage`). This is required for capturing error index metadata through reporting interfaces.
- Honouring context in goroutines started by the reporting feature. 
- Support new reporting implementations easily by using a variable-sized list of delegates inside reporting mediator.
- Remove `WaitForSetup`, since reporting tables are now being migrated synchronously when the program starts.
- Introduced `types.ReportingSyncer` to decouple database initialisation from syncer's main loop.
- Simplify warehouse's reporting setup conditions: 
  - Only standalone slaves don't setup reporting
  - Database syncers are only initialised and run once for each separate database (we are now using `connInfo` as a key instead of the `clientName`).
  - Injecting `Reporting` interface to `UploadJob` instead of `App` to avoid having to expose `ReportingFeature#GetReportingInstance`.
- Addressing races with concurrent map access.

## Linear Ticket

[PIPE-352](https://linear.app/rudderstack/issue/PIPE-352/refactor-reporting-service)

## Security

### DIFF
--- a/enterprise/reporting/error_reporting.go
+++ b/enterprise/reporting/error_reporting.go
@@ -62,7 +62,8 @@ type ErrorDetailReporter struct {
 	onceInit            sync.Once
 	init                chan struct{}
 	reportingServiceURL string
-	clients             map[string]*types.Client
+	syncersMu           sync.RWMutex
+	syncers             map[string]*types.SyncSource
 	log                 logger.Logger
 	namespace           string
 
@@ -73,8 +74,7 @@ type ErrorDetailReporter struct {
 	maxConcurrentRequests misc.ValueLoader[int]
 	maxOpenConnections    int
 
-	httpClient     *http.Client
-	clientsMapLock sync.RWMutex
+	httpClient *http.Client
 
 	backendConfigMu           sync.RWMutex // protects the following
 	workspaceID               string
@@ -123,7 +123,7 @@ func NewErrorDetailReporter(ctx context.Context) *ErrorDetailReporter {
 		init:                      make(chan struct{}),
 		workspaceIDForSourceIDMap: make(map[string]string),
 		destinationIDMap:          make(map[string]destDetail),
-		clients:                   make(map[string]*types.Client),
+		syncers:                   make(map[string]*types.SyncSource),
 		errorDetailExtractor:      extractor,
 		maxOpenConnections:        maxOpenConnections,
 	}
@@ -180,26 +180,26 @@ func (edr *ErrorDetailReporter) DatabaseSyncer(c types.SyncerConfig) types.Repor
 		c.Label = types.CoreReportingLabel
 	}
 
-	edr.clientsMapLock.Lock()
-	defer edr.clientsMapLock.Unlock()
-	if _, ok := edr.clients[c.ConnInfo]; ok {
+	edr.syncersMu.Lock()
+	defer edr.syncersMu.Unlock()
+	if _, ok := edr.syncers[c.ConnInfo]; ok {
 		return func() {} // returning a no-op syncer since another go routine has already started syncing
 	}
 	dbHandle, err := edr.migrate(c)
 	if err != nil {
 		panic(fmt.Errorf("failed during migration: %v", err))
 	}
-	edr.clients[c.ConnInfo] = &types.Client{SyncerConfig: c, DbHandle: dbHandle}
+	edr.syncers[c.ConnInfo] = &types.SyncSource{SyncerConfig: c, DbHandle: dbHandle}
 
 	return func() {
 		edr.mainLoop(edr.ctx, c)
 	}
 }
 
-func (edr *ErrorDetailReporter) GetClient(clientKey string) *types.Client {
-	edr.clientsMapLock.RLock()
-	defer edr.clientsMapLock.RUnlock()
-	return edr.clients[clientKey]
+func (edr *ErrorDetailReporter) GetSyncer(syncerKey string) *types.SyncSource {
+	edr.syncersMu.RLock()
+	defer edr.syncersMu.RUnlock()
+	return edr.syncers[syncerKey]
 }
 
 func (edr *ErrorDetailReporter) Report(metrics []*types.PUReportedMetric, txn *sql.Tx) {
@@ -300,18 +300,18 @@ func (edr *ErrorDetailReporter) extractErrorDetails(sampleResponse string) error
 	}
 }
 
-func (edr *ErrorDetailReporter) getDBHandle(clientKey string) (*sql.DB, error) {
-	client := edr.GetClient(clientKey)
-	if client != nil {
-		return client.DbHandle, nil
+func (edr *ErrorDetailReporter) getDBHandle(syncerKey string) (*sql.DB, error) {
+	syncer := edr.GetSyncer(syncerKey)
+	if syncer != nil {
+		return syncer.DbHandle, nil
 	}
-	return nil, fmt.Errorf("DBHandle not found for client name: %s", clientKey)
+	return nil, fmt.Errorf("DBHandle not found for syncer name: %s", syncerKey)
 }
 
-func (edr *ErrorDetailReporter) getTags(clientName string) stats.Tags {
+func (edr *ErrorDetailReporter) getTags(label string) stats.Tags {
 	return stats.Tags{
 		"workspaceID": edr.workspaceID,
-		"clientName":  clientName,
+		"clientName":  label,
 		"instanceId":  edr.instanceID,
 	}
 }
@@ -340,7 +340,7 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
 	// Delete in a separate go-routine
 	for {
 		if ctx.Err() != nil {
-			edr.log.Infof("stopping mainLoop for client %s : %s", c.Label, ctx.Err())
+			edr.log.Infof("stopping mainLoop for syncer %s : %s", c.Label, ctx.Err())
 			return
 		}
 		requestChan := make(chan struct{}, edr.maxConcurrentRequests.Load())
@@ -355,7 +355,7 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
 		if len(reports) == 0 {
 			select {
 			case <-ctx.Done():
-				edr.log.Infof("stopping mainLoop for client %s : %s", c.Label, ctx.Err())
+				edr.log.Infof("stopping mainLoop for syncer %s : %s", c.Label, ctx.Err())
 				return
 			case <-time.After(edr.sleepInterval.Load()):
 			}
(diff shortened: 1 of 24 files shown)


enterprise/reporting/setup_test.go
```diff
@@ -77,8 +82,8 @@ func TestSetupForNoop(t *testing.T) {
 					EnterpriseToken: "dummy-token",
 				}
 			}
-			med := NewReportingMediator(logger.NOP, f.EnterpriseToken)
-			require.Equal(t, med.createErrorReportInstance(&backendconfig.NOOP{}), &NOOP{})
+			med := NewReportingMediator(context.Background(), logger.NOP, f.EnterpriseToken, &backendconfig.NOOP{})
+			require.Len(t, med.delegates, tc.expectedDelegates)
```
achettyiitr, line 86: As per the existing code, should we not create any reporting instance here? Since we are comparing it with NOOP?

enterprise/reporting/error_reporting.go
```diff
@@ -58,39 +58,41 @@ type destDetail struct {
...
-	srcWspMutex               sync.RWMutex
-	httpClient                *http.Client
-	clientsMapLock            sync.RWMutex
 
 	errorDetailExtractor *ExtractorHandle
 
 	minReportedAtQueryTime      stats.Measurement
 	errorDetailReportsQueryTime stats.Measurement
-	edReportingRequestLatency   stats.Measurement
+	edrortingRequestLatency     stats.Measurement
```
achettyiitr, line 87:
```suggestion
	edReportingRequestLatency     stats.Measurement
```

enterprise/reporting/error_reporting.go
```diff
@@ -293,58 +278,58 @@ func (edRep *ErrorDetailReporter) migrate(c types.Config) (*sql.DB, error) {
...
+func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
+	<-edr.init
+	tags := edr.getTags(c.Label)
 
 	mainLoopTimer := stats.Default.NewTaggedStat("error_detail_reports_main_loop_time", stats.TimerType, tags)
 	getReportsTimer := stats.Default.NewTaggedStat("error_detail_reports_get_reports_time", stats.TimerType, tags)
 	aggregateTimer := stats.Default.NewTaggedStat("error_detail_reports_aggregate_time", stats.TimerType, tags)
 	getReportsSize := stats.Default.NewTaggedStat("error_detail_reports_size", stats.HistogramType, tags)
-	getAggregatedReportsSize := stats.Default.NewTaggedStat("error_detail_reports_aggregated_size", stats.HistogramType, tags)
+	getAggregatedrortsSize := stats.Default.NewTaggedStat("error_detail_reports_aggregated_size", stats.HistogramType, tags)
```
achettyiitr, line 326:
```suggestion
	getAggregatedReportsSize := stats.Default.NewTaggedStat("error_detail_reports_aggregated_size", stats.HistogramType, tags)
```

enterprise/reporting/error_reporting.go
```diff
@@ -293,58 +278,58 @@ func (edRep *ErrorDetailReporter) migrate(c types.Config) (*sql.DB, error) {
...
 
 	errorDetailReportsDeleteQueryTimer := stats.Default.NewTaggedStat("error_detail_reports_delete_query_time", stats.TimerType, tags)
 
-	edRep.minReportedAtQueryTime = stats.Default.NewTaggedStat("error_detail_reports_min_reported_at_query_time", stats.TimerType, tags)
-	edRep.errorDetailReportsQueryTime = stats.Default.NewTaggedStat("error_detail_reports_query_time", stats.TimerType, tags)
-	edRep.edReportingRequestLatency = stats.Default.NewTaggedStat("error_detail_reporting_request_latency", stats.TimerType, tags)
+	edr.minReportedAtQueryTime = stats.Default.NewTaggedStat("error_detail_reports_min_reported_at_query_time", stats.TimerType, tags)
+	edr.errorDetailReportsQueryTime = stats.Default.NewTaggedStat("error_detail_reports_query_time", stats.TimerType, tags)
+	edr.edrortingRequestLatency = stats.Default.NewTaggedStat("error_detail_reporting_request_latency", stats.TimerType, tags)
```
achettyiitr, line 332:
```suggestion
	edr.edReportingRequestLatency = stats.Default.NewTaggedStat("error_detail_reporting_request_latency", stats.TimerType, tags)
```

enterprise/reporting/error_reporting.go
```diff
@@ -353,32 +338,32 @@ func (edRep *ErrorDetailReporter) mainLoop(ctx context.Context, clientName strin
...
 			}
 			continue
 		}
 
 		aggregationStart := time.Now()
-		metrics := edRep.aggregate(reports)
+		metrics := edr.aggregate(reports)
 		aggregateTimer.Since(aggregationStart)
-		getAggregatedReportsSize.Observe(float64(len(metrics)))
+		getAggregatedrortsSize.Observe(float64(len(metrics)))
```
achettyiitr, line 366:
```suggestion
		getAggregatedReportsSize.Observe(float64(len(metrics)))
```

enterprise/reporting/error_reporting.go
```diff
@@ -510,16 +495,16 @@ func (edRep *ErrorDetailReporter) getReports(ctx context.Context, currentMs int6
...
 		}
 		metrics = append(metrics, dbEdMetric)
 	}
 	return metrics, queryMin.Int64
 }
 
-func (edRep *ErrorDetailReporter) aggregate(reports []*types.EDReportsDB) []*types.EDMetric {
-	groupedReports := lo.GroupBy(reports, func(report *types.EDReportsDB) string {
+func (edr *ErrorDetailReporter) aggregate(reports []*types.EDReportsDB) []*types.EDMetric {
+	groupedrorts := lo.GroupBy(reports, func(report *types.EDReportsDB) string {
```
achettyiitr, line 507:
```suggestion
	groupedReports := lo.GroupBy(reports, func(report *types.EDReportsDB) string {
```

achettyiitr: Seems like variable names are mistakenly typed.

enterprise/reporting/mediator.go
```diff
@@ -13,77 +13,66 @@ import (
...
-	return &ReportingMediator{
-		log:             log,
-		enterpriseToken: enterpriseToken,
+	// error reporting implementation
+	if config.GetBool("Reporting.errorReporting.enabled", false) {
+		errorReporter := NewErrorDetailReporter()
+		rruntime.Go(func() {
+			errorReporter.backendConfigSubscriber(ctx, backendConfig)
+		})
+		rm.delegates = append(rm.delegates, errorReporter)
```
achettyiitr, line 44: As per the existing logic, should we not create only 1 reporter in case of `Reporting.errorReporting.enabled` and `Reporting.enabled` is true

enterprise/reporting/setup_test.go
atzoum: please elaborate, we are comparing expected number of delegates, right?

achettyiitr: It's clarified in [here](https://github.com/rudderlabs/rudder-server/pull/3912#discussion_r1338104614). Marking it as resolved.

achettyiitr (approved):

enterprise/reporting/mediator.go
atzoum: Reporting mediator shall always be created, because even if `Reporting.enabled: false` other reporters might need to be enabled, e.g. the error metadata reporter we will be creating, or a rudder-sources reporter which might come next

enterprise/reporting/setup_test.go
lvrach, line 86: [non-blocking]
I prefer a more black-box approach when it comes to testing here. 

* No accessing unexported fields like `instance` and `delegates`
* Not relying on an implementation detail of a delegate slice to tests that delegates are enabled
* Using `reporting_test` package instead of `reporting` 

I understand this is an incremental step, but I just wanted to express my thoughts when looking at this code.

enterprise/reporting/mediator.go
```diff
@@ -13,77 +13,66 @@ import (
...
+
+	delegates []types.Reporting
 }
 
-func (med *ReportingMediator) createReportInstance(backendConfig backendconfig.BackendConfig) types.Reporting {
-	med.log.Debug("Forming reporting instance")
-	reportingEnabled := config.GetBool("Reporting.enabled", types.DefaultReportingEnabled)
-	if !reportingEnabled || med.enterpriseToken == "" {
-		return &NOOP{}
+func NewReportingMediator(ctx context.Context, log logger.Logger, enterpriseToken string, backendConfig backendconfig.BackendConfig) *ReportingMediator {
```
lvrach, line 21: Here we are passing ctx, but the `NewReportingMediator` method is not really blocking. It runs go routines in the background without a way to block until they return. 



Would it make sense to use the `DatabaseSyncer` as a generic `Run` method, that will accept ctx and will be responsible for running all the `defaultReporter.backendConfigSubscriber` go routines? 

And simplify  NewReportingMediator:
```suggestion
func NewReportingMediator(log logger.Logger, enterpriseToken string, backendConfig backendconfig.BackendConfig) *ReportingMediator {
```

to just create the  ReportingMediator struct:
```go
type ReportingMediator struct {
	log logger.Logger

	delegates []types.Reporting
	backconfSubscribers []func(ctx, ...) 
 }
 ```go
 
 Then `DatabaseSyncer` will use its waitgroup to run delegates and backend config subscribers

File: warehouse/router_test.go [hunk 9]
```diff
@@ -48,13 +46,7 @@ func TestRouter(t *testing.T) {
...
 
 	report := &reporting.Factory{}
-	report.Setup(&backendconfig.NOOP{})
-
-	mockCtrl := gomock.NewController(t)
-	mockApp := mocksApp.NewMockApp(mockCtrl)
-	mockApp.EXPECT().Features().Return(&app.Features{
-		Reporting: report,
-	}).AnyTimes()
+	reportingI := report.Setup(context.Background(), &backendconfig.NOOP{})
```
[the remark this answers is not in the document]
atzoum: it would shadow the reporting package. Can rename this to `reportingInstance`

enterprise/reporting/mediator.go
atzoum: Blocking until all go routines have stopped is not implemented yet. There are 2 different kinds of go routines:

1. backend config subscribers
2. syncers' main loops

Both would need to stop during shutdown. The current improvement is that go routines will stop when the context is canceled, even if the program currently doesn't wait for them to stop.

atzoum: will try to take care of it in this pr

atzoum: @lvrach reporting mediator is now using its own `errgroup`, it spawns goroutines using this group and waits for all of them to finish before returning from its `ReportingSyncer` functions.
`gateway`, `processor` and `warehouse` are all running `ReportingSyncer`s in their own errgroups, i.e. waiting for them to finish.

enterprise/reporting/mediator.go
```diff
@@ -3,87 +3,79 @@ package reporting
...
+		return rm
 	}
-	errorReporter := NewEdReporterFromEnvConfig()
-	rruntime.Go(func() {
-		errorReporter.setup(backendConfig)
+
+	// default reporting implementation
+	defaultReporter := NewDefaultReporter(rm.ctx, rm.log)
+	rm.g.Go(func() error {
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
atzoum: oops :(

warehouse/router_test.go
```diff
@@ -124,7 +116,7 @@ func TestRouter(t *testing.T) {
 
 		r, err := newRouter(
 			ctx,
-			mockApp,
+			&reporting.NOOP{},
```
achettyiitr, line 119: [optional]
```suggestion
			reportingInstance,
```

achettyiitr (approved):

lvrach (approved):

PR #3913, opened 2023-09-26
fix: event schema versions without a sample event

cisse21:
# Description

Sampling events at EventSchemaVersion level to populate more samples in the `schema_version` table

## Linear Ticket

[ Linear Link ](https://linear.app/rudderstack/issue/PIPE-343/fix-empty-sample-event-bug-in-schema-version-table)

## Security


proto/event-schema/types.helper.go
```diff
@@ -60,8 +60,7 @@ func (sm *EventSchemaMessage) Merge(other *EventSchemaMessage) {
 	if len(other.Sample) < len(sm.Sample) { // keep the smallest sample
 		sm.Sample = other.Sample
 	}
-	sm.Sample = other.Sample
```
atzoum, line 63: 👍

atzoum (approved):

PR #3115, opened 2023-03-18
feat: update transformer image to rudderstack org

yashasvibajpai:
# Description

We are intending to discontinue pushing rudder-transformer images to the `rudderlabs` organisation, and hence the instances are being migrated to one central org `rudderstack`.

## Notion Ticket
[Ticket](https://www.notion.so/rudderstacks/Update-all-transformer-docker-references-to-rudderstack-repository-066da4270b5243a084ffed35eff1927f?pvs=4)

## Security

- [ ] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #3916, opened 2023-09-27
chore: use transformerdocker image that is publicly available

lvrach:
# Description

While testing rudder-server in codespaces I got an error when try to run docker-compose up. 

The docker image we are using requires to login to our docker hub.

I've used a public image to make development of rudder-server more acesible.

## Notion ticket 
https://linear.app/rudderstack/issue/PIPE-367/use-public-transformer-image-in-docker-compose-file

## Security


achettyiitr (approved):

atzoum (commented):
what about the others changed by #3115?

atzoum (approved):

PR #3937, opened 2023-10-03
chore: split warehouse package

achettyiitr:
# Description

- Splitting warehouse packages into smaller packages. (**admin**, **api**, **bcm**, **constraints**, **internal/mode**, **router**, **slave**, **util/types**)

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-222/splitting-warehouse-packages-into-individual-packages

## TODO
- Introduce a common constructor API for HTTP and GRPC. Expose 2 methods for `startHTTP` and `startGRPC`.

## Security


warehouse/admin/admin.go
fracasula, line 1: @achettyiitr I thought this task would be a nice opportunity to do simple cleanups to these packages without having to do heavy refactorings. With that I mean trying to remove at least the easy dependencies.

For example if we look at this package (i.e. `admin`) we can easily remove a few deps. The point of this exercise is to see if we can start moving as much logic as we can into independent packages that are responsible for as less things as possible.

To mention a couple dependencies, here we don't need to depend on the backend config manager and neither we need to depend on the main `admin` package. Have a look at my `git` patch attached, just unzip and do a `git apply admin.patch` to have a look.

[admin.zip](https://github.com/rudderlabs/rudder-server/files/12820163/admin.zip)

fracasula (commented):
@achettyiitr I left a comment on the `admin` package. Please consider that I'm not asking to do anything heavy nor we need to cover every single package here.

But I think it's good to keep this in mind whenever we're designing a package or moving things around. Then with time we'll end up with code that is less coupled and easier to test.

I hope it makes sense. In the meantime I'll try to review as much as possible in this PR so that we can merge it soon and then do things incrementally as we see fit.

warehouse/admin/admin.go
fracasula: Something I forgot to mention, it is usually preferrable to assemble all the dependencies together from a main component, in our case it is the `warehouse.App` which kinda works like a `main.go` for simple scripts, right?

In there we can pull the various packages and put them together by effectively doing Dependency Injections.

warehouse/admin/admin.go
```diff
@@ -53,7 +56,7 @@ func (a *Admin) Register() {
 
 // TriggerUpload sets uploads to start without delay
 func (*Admin) TriggerUpload(off bool, reply *string) error {
-	startUploadAlways.Store(!off)
+	router.StartUploadAlways.Store(!off)
```
fracasula, line 59: These calls to global functions or variables are exactly what we need to get rid of :sweat_smile: 
The fact that the `TriggerUpload` method doesn't have a method receiver can be a smell.

fracasula: See if [it makes sense](https://github.com/rudderlabs/rudder-server/pull/3945/files).

warehouse/errors/errors.go
fracasula, line 1: Not sure we need a new package here. It is used by the router only. Why not just have an `errors.go` inside `warehouse/router`?

Also you don't need a `manager.Manager` here, this should suffice:

```go
type errorMapper interface {
	ErrorMappings() []model.JobError
}

type ErrorHandler struct {
	Mapper errorMapper
}
```

Check it here: https://github.com/rudderlabs/rudder-server/pull/3945/commits/56304c46b93bee41d2074ef36efe74b0c43fd6ff

warehouse/api/http.go
```diff
@@ -70,7 +73,7 @@ type Api struct {
 	notifier      *notifier.Notifier
 	bcConfig      backendconfig.BackendConfig
 	tenantManager *multitenant.Manager
-	bcManager     *backendConfigManager
+	bcManager     *bcm.BackendConfigManager
```
fracasula, line 76: We could have an interface here as well. Two methods only are needed.

warehouse/bcm/backend_config.go
```diff
@@ -331,3 +334,20 @@ func deepCopy(src, dest interface{}) error {
...
+) {
+	tags := stats.Tags{
+		"workspaceId":   workspaceID,
+		"module":        "warehouse",
+		"destType":      destType,
+		"warehouseID":   misc.GetTagName(destID, sourceName, destName, misc.TailTruncateStr(sourceID, 6)),
+		"destinationID": destID,
+		"errTag":        errTag,
+	}
+	stats.Default.NewTaggedStat("persist_ssl_file_failure", stats.CountType, tags).Count(1)
```
fracasula, line 352: Can't we inject `stats` instead? We would also justify the method receiver here.

fracasula (approved):

warehouse/bcm/backend_config.go
achettyiitr: I was thinking of injecting it in a separate PR.

achettyiitr: Added the changes.

warehouse/admin/admin.go
achettyiitr: Thanks, @fracasula for pointing out in terms of another dimension of removing dependencies. I was also thinking of passing dependencies from the warehouse/app. The [draft](https://github.com/rudderlabs/rudder-server/pull/3945) you shared really helped me a lot in terms of understanding it in a detailed way.

warehouse/admin/admin.go
achettyiitr: Yes.

warehouse/errors/errors.go
achettyiitr: Yes. we don't need Manager dependency here.

warehouse/api/http.go
achettyiitr: I will add this to the Backlog. I also want to have a unified constructor for `api` package which will have `startHTTPServer` and `startGRPCServer` instead of having separate constructors for HTTP and GRPC.

warehouse/mode/mode.go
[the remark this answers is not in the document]
achettyiitr: Yes.

fracasula:
@achettyiitr 

```
=== FAIL: warehouse/constraints TestConstraintsManager/Violates_index_constraint (0.00s)
    constraint_test.go:80: 
        	Error Trace:	/home/runner/work/rudder-server/rudder-server/warehouse/constraints/constraint_test.go:80
        	Error:      	Not equal: 
        	            	expected: true
        	            	actual  : false
        	Test:       	TestConstraintsManager/Violates_index_constraint
    --- FAIL: TestConstraintsManager/Violates_index_constraint (0.00s)

=== FAIL: warehouse/constraints TestConstraintsManager (0.00s)
```

fracasula:
> @achettyiitr
> 
> ```
> === FAIL: warehouse/constraints TestConstraintsManager/Violates_index_constraint (0.00s)
>     constraint_test.go:80: 
>         	Error Trace:	/home/runner/work/rudder-server/rudder-server/warehouse/constraints/constraint_test.go:80
>         	Error:      	Not equal: 
>         	            	expected: true
>         	            	actual  : false
>         	Test:       	TestConstraintsManager/Violates_index_constraint
>     --- FAIL: TestConstraintsManager/Violates_index_constraint (0.00s)
> 
> === FAIL: warehouse/constraints TestConstraintsManager (0.00s)
> ```

Weird, locally it's passing just fine.

fracasula:
> @achettyiitr
> 
> ```
> === FAIL: warehouse/constraints TestConstraintsManager/Violates_index_constraint (0.00s)
>     constraint_test.go:80: 
>         	Error Trace:	/home/runner/work/rudder-server/rudder-server/warehouse/constraints/constraint_test.go:80
>         	Error:      	Not equal: 
>         	            	expected: true
>         	            	actual  : false
>         	Test:       	TestConstraintsManager/Violates_index_constraint
>     --- FAIL: TestConstraintsManager/Violates_index_constraint (0.00s)
> 
> === FAIL: warehouse/constraints TestConstraintsManager (0.00s)
> ```

The test is flaky. Even locally sometimes it passes and sometimes it fails.

fracasula:
> The test is flaky. Even locally sometimes it passes and sometimes it fails.

@achettyiitr This seems to fix the problem:

```diff
diff --git a/warehouse/constraints/constraint_test.go b/warehouse/constraints/constraint_test.go
index ba149af22..26b519e90 100644
--- a/warehouse/constraints/constraint_test.go
+++ b/warehouse/constraints/constraint_test.go
@@ -75,7 +75,8 @@ func TestConstraintsManager(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
-			cm := New(config.Default)
+			c := config.New()
+			cm := New(c)
 			cv := cm.ViolatedConstraints(tc.destinationType, tc.brEvent, tc.columnName)
 			require.Equal(t, tc.expected.IsViolated, cv.IsViolated)
 			require.True(t, strings.HasPrefix(cv.ViolatedIdentifier, tc.expected.ViolatedIdentifier))
```

warehouse/slave/worker.go
```diff
@@ -508,7 +512,7 @@ func (sw *slaveWorker) runAsyncJob(ctx context.Context, asyncjob jobs.AsyncJobPa
 	return result, nil
 }
 
-func (sw *slaveWorker) destinationFromSlaveConnectionMap(destinationId, sourceId string) (model.Warehouse, error) {
+func (sw *worker) destinationFromSlaveConnectionMap(destinationId, sourceId string) (model.Warehouse, error) {
```
cisse21, line 515: change sw to w?

cisse21 (approved):

warehouse/slave/worker.go
achettyiitr: OOps missed updating the receiver. Updating it.

PR #3956, opened 2023-10-09
fix: set local schema when syncing from remote

achettyiitr:
# Description

- When syncing from remote, we need to set the local schema as well.

## Linear Ticket

- https://linear.app/rudderstack/issue/PIPE-392/sync-remote-schema-fix

## Security

### DIFF
--- a/warehouse/schema/schema.go
+++ b/warehouse/schema/schema.go
@@ -73,7 +73,13 @@ func New(
 	}
 }
 
-// ConsolidateStagingFilesUsingLocalSchema consolidates staging files schema with warehouse schema
+// ConsolidateStagingFilesUsingLocalSchema
+// 1. Fetches the schemas for the staging files
+// 2. Consolidates the staging files schemas
+// 3. Consolidates the consolidated schema with the warehouse schema
+// 4. Enhances the consolidated schema with discards schema
+// 5. Enhances the consolidated schema with ID resolution schema
+// 6. Returns the consolidated schema
 func (sh *Schema) ConsolidateStagingFilesUsingLocalSchema(ctx context.Context, stagingFiles []*model.StagingFile) (model.Schema, error) {
 	consolidatedSchema := model.Schema{}
 	batches := lo.Chunk(stagingFiles, sh.stagingFilesSchemaPaginationSize)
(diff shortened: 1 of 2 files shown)


warehouse/schema/schema.go
```diff
@@ -411,20 +229,148 @@ func enhanceSchemaWithIDResolution(consolidatedSchema model.Schema, isIDResoluti
...
+		return false, fmt.Errorf("fetching schema from local: %w", err)
+	}
+
+	if err := s.FetchSchemaFromWarehouse(ctx, fetchSchemaRepo); err != nil {
+		return false, fmt.Errorf("fetching schema from warehouse: %w", err)
+	}
+
+	s.localSchemaMu.Lock()
+	s.localSchema = localSchema
+	s.localSchemaMu.Unlock()
```
achettyiitr, line 287: This was missing from earlier code when we made changes around it during [this](https://github.com/rudderlabs/rudder-server/pull/3900).

fracasula:
@achettyiitr can you redo this on `release/1.15.x`?

fracasula:
It's hard to tell what you changed because you renamed the methods receivers. You did that all in one commit so can you either not rename the method receiver or do it in a separate commit please?

lvrach (approved):
This refactoring and the addition of tests is the right step forward. 

In my review, I checked for the locks we are currently using, but I am not certain about all of them. 

We should invest in simplifying this code further, and understand if those many locks are indeed needed.

achettyiitr:
> It's hard to tell what you changed because you renamed the methods receivers. You did that all in one commit so can you either not rename the method receiver or do it in a separate commit, please?

Sure. Also, I added an existing test in the master to verify. https://github.com/rudderlabs/rudder-server/pull/3957

warehouse/schema/schema.go
fracasula: Aren't we updating this if the schema has changed?

```go
	if schemaChanged {
		err := sh.updateLocalSchema(ctx, uploadID, sh.schemaInWarehouse)
		if err != nil {
			return false, fmt.Errorf("updating local schema: %w", err)
		}
	}
```

achettyiitr: This is only when there is a difference in local and warehouse schema. If not local schema is left empty.

fracasula: I'm not convinced updating it twice is the answer. Let me recheck the `hasSchemaChanged` logic.

warehouse/schema/schema.go
```diff
@@ -411,20 +229,148 @@ func enhanceSchemaWithIDResolution(consolidatedSchema model.Schema, isIDResoluti
...
+		DestinationType: s.warehouse.Type,
+		Schema:          updatedSchema,
+	})
+	if err != nil {
+		return fmt.Errorf("updating local schema: %w", err)
+	}
+
+	s.localSchemaMu.Lock()
+	s.localSchema = updatedSchema
+	s.localSchemaMu.Unlock()
```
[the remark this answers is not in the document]
achettyiitr: Yes, we can simplify the logic by separating the caching into a different layer.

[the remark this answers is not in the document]
achettyiitr: Database queries can cost us because in some cases even the queries to `INFORMATION_SCHEMAS` tables take a long time.

warehouse/schema/schema.go
achettyiitr: Currently, I am initializing it with what I got from `sh.GetLocalSchema(ctx)` otherwise it is empty right if there is no difference?

fracasula: Yes, it doesn't work if empty, you are correct :+1:

fracasula (approved):
Thanks for the changes @achettyiitr :+1: 

We're targeting `master` here though but yesterday I had created a [release](https://github.com/rudderlabs/rudder-server/pull/3954) and a [prerelease](https://github.com/rudderlabs/rudder-server/pull/3955).

If you guys (cc @lvrach) are planning to merge this into `master` please close them. We need to get some closure on this though, it's starting to become confusing.

warehouse/schema/schema.go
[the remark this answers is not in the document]
achettyiitr: Added a task in the [Backlog](https://linear.app/rudderstack/issue/PIPE-393/simplify-schema-changes) so that we can revisit this.

PR #3961, opened 2023-10-10
feat: introduce pagination in failed-keys endpoint

atzoum:
# Description

## feat: introduce pagination in failed-keys endpoint

Adding support for pagination in failed-keys endpoint in a backwards compatible manner:
- If the request contains a `pageSize` query parameter then pagination will be enabled, otherwise the full resultset will be returned in the response.
- If pagination is requested, then the next page's token will be included in the `paging` field of the response's payload. To retrieve the next page, the client needs to include this token in the `pageToken `query parameter of it's next http request.
- Pagination stops when no more results are returned in the response. When this happens, the `paging` field is omitted from the response's payload as well.

_**Performance**_
Added a test and a benchmark to verify the performance characteristics of the new paging API compared to the previous one. The paging API is more memory efficient but slower, e.g. splitting 10Mil records in 10 pages is 3x slower than returning the full resultset at once. 

Thus, an adequately large page size needs to be used in production environments: **It is advised to use a page size between 100K - 1Mil records**.

## chore: use a normalised data model for storing failed keys

- using a normalised data model for greater disk storage efficiency when there is a large number of failed keys
- deduplicating failed keys for the same job, task, source & destination
- using `ksuid` for `id` column and `text` instead of `jsonb` for `record_id` column in the new data model so that future introduction of pagination becomes a straightforward task & highly performant
- removing previous table `rsources_failed_keys`. In case of a rollback, the previous server version will recreate this table. Data migration after rollback needs to happen manually, using the following statement (a hotfix including this statement could be prepared if needed):
```sql
INSERT INTO rsources_failed_keys (
	job_run_id, 
	task_run_id, 
	source_id, 
	destination_id, 
	record_id, 
	ts
)
SELECT 
	k.job_run_id, 
	k.task_run_id, 
	k.source_id, 
	k.destination_id, 
	to_jsonb(r.record_id), 
	r.ts 
FROM rsources_failed_keys_v2 k 
JOIN rsources_failed_keys_v2_records r on r.id = k.id;
```

**_Trivia_**

One could argue that by introducing normalisation and deduplication here, we might also affect performance by introducing lock contention at the database level, since competing threads will be trying to insert data with the same primary key. Thankfully, that's not the case, we only collect failed keys in router & batchrouter and these two components never update job statuses concurrently for a given destination id. But even if concurrent updates was the case, recording failed keys is an exceptional flow for the system, i.e. it shouldn't matter much. 

## Linear Ticket

Resolves PIPE-396
Resolves PIPE-336

## Security

BEGIN_COMMIT_OVERRIDE
feat: introduce pagination in failed-keys endpoint (#3967)
chore: use a normalised data model for storing failed keys (#3961)
END_COMMIT_OVERRIDE

### DIFF
--- a/services/rsources/handler.go
+++ b/services/rsources/handler.go
@@ -34,7 +34,7 @@ type sourcesHandler struct {
 }
 
 func (sh *sourcesHandler) GetStatus(ctx context.Context, jobRunId string, filter JobFilter) (JobStatus, error) {
-	filters, filterParams := sqlFilters(jobRunId, filter, "")
+	filters, filterParams := sqlFilters(jobRunId, filter)
 
 	sqlStatement := fmt.Sprintf(
 		`SELECT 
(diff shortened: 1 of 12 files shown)


services/rsources/handler.go
```diff
@@ -385,27 +536,37 @@ func (sh *sourcesHandler) setupLogicalReplication(ctx context.Context) error {
...
 
-func sqlFilters(jobRunId string, filter JobFilter) (fragment string, params []interface{}) {
+func sqlFilters(jobRunId string, filter JobFilter, alias string) (fragment string, params []interface{}) {
 	var filterParams []interface{}
-	filters := `WHERE job_run_id = $1`
+	var prefix string
+	if alias != "" {
+		prefix = alias + "."
+	}
+	filters := fmt.Sprintf(`WHERE %sjob_run_id = $1`, prefix)
```
cisse21, line 560: [nit-pick]

```suggestion
	if alias != "" {
		prefix = alias 
	}
	filters := fmt.Sprintf(`WHERE %s.job_run_id = $1`, prefix)
```

services/rsources/handler.go
```diff
@@ -385,27 +536,37 @@ func (sh *sourcesHandler) setupLogicalReplication(ctx context.Context) error {
...
 	filterParams = append(filterParams, jobRunId)
 
 	if len(filter.TaskRunID) > 0 {
-		filters += fmt.Sprintf(` AND task_run_id  = ANY ($%d)`, len(filterParams)+1)
+		filters += fmt.Sprintf(` AND %stask_run_id  = ANY ($%d)`, prefix, len(filterParams)+1)
```
cisse21, line 564: [nit-pick]
```suggestion
		filters += fmt.Sprintf(` AND %s.task_run_id  = ANY ($%d)`, prefix, len(filterParams)+1)
```

services/rsources/handler.go
```diff
@@ -385,27 +536,37 @@ func (sh *sourcesHandler) setupLogicalReplication(ctx context.Context) error {
...
 		filterParams = append(filterParams, pq.Array(filter.TaskRunID))
 	}
 
 	if len(filter.SourceID) > 0 {
-		filters += fmt.Sprintf(` AND source_id = ANY ($%d)`, len(filterParams)+1)
+		filters += fmt.Sprintf(` AND %ssource_id = ANY ($%d)`, prefix, len(filterParams)+1)
```
cisse21, line 569: [nit-pick]
```suggestion
		filters += fmt.Sprintf(` AND %s.source_id = ANY ($%d)`, prefix, len(filterParams)+1)
```

cisse21 (approved):

services/rsources/handler.go
atzoum: Removing the prefix in #3967

lvrach (approved):

services/rsources/handler.go
```diff
@@ -103,78 +104,128 @@ func (*sourcesHandler) IncrementStats(ctx context.Context, tx *sql.Tx, jobRunId
...
+	if err != nil {
+		return JobFailedRecords{ID: jobRunId}, fmt.Errorf("failed to get failed record ids: %w", err)
+	}
 
+	filters := "WHERE r.id = ANY($1)"
+	params := []interface{}{pq.Array(ids)}
+	var limit string
+	if paging.Size > 0 {
+		if nextPageToken, err := NextPageTokenFromString(paging.NextPageToken); err == nil {
+			filters = filters + fmt.Sprintf(` AND r.id >= $%[1]d AND r.record_id > $%[2]d`, len(params)+1, len(params)+2)
```
Sidddddarth, line 177: If this unmarshal fails, we'd end up returning all the rows defeating the purpose of pagination.
We can return an error so the user knows to send the proper params.

services/rsources/http/http.go
```diff
@@ -98,12 +99,20 @@ func (h *handler) failedRecords(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "job_run_id not found", http.StatusBadRequest)
 		return
 	}
+	var paging rsources.PagingInfo
+	if pageSize, ok := r.URL.Query()["pageSize"]; ok && len(pageSize) > 0 {
```
Sidddddarth, line 103: maybe we can have a default limit here(10k records)
And also an upper limit, if the user of the API sends a very large limit, it could affect the system adversely.

Sidddddarth:
Posted two comments that I could think of
LGTM otherwise

services/rsources/http/http.go
atzoum: The API is backwards compatible, i.e. no default limit should be applied here.
The advised page limit for clients is be really high, e.g. 500K-1Mil, thus an upper limit isn't critical imo

Sidddddarth: >backwards compatible

Touché
But in case they use the newer API, wouldn't we want some configurable limit?

>page limit for clients is be really high

I didn't fully understand this, please elaborate

atzoum: 1. Performance will suffer if page size is smaller than 500k (see performance test)
2. The payload is not significant since the endpoint practically returns a list of record ids, thus it should be fine to use pretty large page sizes

services/rsources/handler.go
```diff
@@ -103,78 +107,130 @@ func (*sourcesHandler) IncrementStats(ctx context.Context, tx *sql.Tx, jobRunId
...
+	}()
+	if err != nil {
+		return JobFailedRecords{ID: jobRunId}, fmt.Errorf("failed to get failed record ids: %w", err)
+	}
 
+	filters := "WHERE r.id = ANY($1)"
+	params := []interface{}{pq.Array(ids)}
+	var limit string
+	if paging.Size > 0 {
+		if nextPageToken, err := NextPageTokenFromString(paging.NextPageToken); err == nil {
```
Sidddddarth, line 179: Minute:
This unmarshal could've been done in the httphandler and return early on failure(and avoid above `SELECT id FROM "rsources_failed_keys_v2"` query).

Sidddddarth (approved):

PR #3965, opened 2023-10-11
feat: use append vs merge option from backend config

fracasula:
# Description

Adding support for opt-in deduplication.

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-258/use-append-vs-merge-option-from-backend-config-ui) >

## E2E Warehouse run

https://github.com/rudderlabs/e2e/actions/runs/6587296799

## Updated Flow Diagram

append_vs_merge drawio

## Outstanding questions
* do we want to have a `shouldMerge` method in `mssql` as well [here](https://github.com/rudderlabs/rudder-server/blob/v1.15.3/warehouse/integrations/mssql/mssql.go#L355) like we do in [postgres](https://github.com/rudderlabs/rudder-server/blob/v1.15.3/warehouse/integrations/postgres/load.go#L146)? if we add it, is it OK for existing customers that are using `mssql`?
* should we review also other destinations like [azure synapse](https://github.com/rudderlabs/rudder-server/blob/v1.15.3/warehouse/integrations/azure-synapse/azure-synapse.go#L326) and add a `shouldMerge` there as well?

## TODOs
* document `CanAppend` with a generic statement in the public docs saying that even though the user didn't opt-in with  `enableMerge` we might merge anyway due to other conditions e.g. if they connected an extract source, we replay the data, etc...
* check if we have live configurations for our customers that are using `loadTableStrategy` and replace that with `enableMerge`
   * done here: https://github.com/rudderlabs/rudder-devops/pull/9020

## Security


warehouse/integrations/bigquery/bigquery.go
```diff
@@ -859,7 +859,9 @@ func (bq *BigQuery) connect(ctx context.Context, cred BQCredentials) (*bigquery.
 }
 
 func (bq *BigQuery) dedupEnabled() bool {
-	return bq.config.isDedupEnabled || bq.config.isUsersTableDedupEnabled
+	return bq.config.isDedupEnabled ||
+		bq.config.isUsersTableDedupEnabled ||
+		bq.warehouse.GetBoolDestinationConfig("enableDedup")
```
fracasula, line 864: @lvrach does it make sense to keep the other configuration parameters? Should we keep the opt-in `enableDedup` only?

warehouse/integrations/deltalake/deltalake.go
```diff
@@ -1383,5 +1383,7 @@ func (*Deltalake) DeleteBy(context.Context, []string, warehouseutils.DeleteByPar
 // * the load table strategy is "append" mode
 // * the uploader says we can append
 func (d *Deltalake) ShouldAppend() bool {
-	return d.config.loadTableStrategy == appendMode && d.Uploader.CanAppend()
+	return d.config.loadTableStrategy == appendMode &&
+		d.Uploader.CanAppend() &&
+		!d.Warehouse.GetBoolDestinationConfig("enableDedup")
```
fracasula, line 1388: @lvrach Same question here, does the load table strategy makes sense? Should we remove it?

warehouse/integrations/mssql/mssql.go
```diff
@@ -701,15 +701,17 @@ func (ms *MSSQL) loadUserTables(ctx context.Context) (errorMap map[string]error)
...
-	primaryKey := "id"
-	sqlStatement = fmt.Sprintf(`DELETE FROM %[1]s."%[2]s" FROM %[3]s _source where (_source.%[4]s = %[1]s.%[2]s.%[4]s)`, ms.Namespace, warehouseutils.UsersTable, ms.Namespace+"."+stagingTableName, primaryKey)
-	ms.logger.Infof("MSSQL: Dedup records for table:%s using staging table: %s\n", warehouseutils.UsersTable, sqlStatement)
-	_, err = tx.ExecContext(ctx, sqlStatement)
-	if err != nil {
-		ms.logger.Errorf("MSSQL: Error deleting from main table for dedup: %v\n", err)
-		_ = tx.Rollback()
-		errorMap[warehouseutils.UsersTable] = err
-		return
+	if ms.enableDedup() {
```
fracasula, line 704: We weren't checking any flags for dedup here, just pinging you guys to make sure it makes sense to do so before I add the tests. @lvrach @achettyiitr

warehouse/integrations/postgres/load.go
```diff
@@ -532,35 +532,32 @@ func (pg *Postgres) loadUsersTable(
...
-		  (
-			_source.%[4]s = %[1]s.%[2]s.%[4]s
-		  );
-`,
-		pg.Namespace,
-		warehouseutils.UsersTable,
-		usersStagingTableName,
-		primaryKey,
-	)
+	if pg.enableDedup() {
```
fracasula, line 535: We weren't checking any flags for dedup here, just pinging you guys to make sure it makes sense to do so before I add the tests. @lvrach @achettyiitr

warehouse/integrations/redshift/redshift.go
```diff
@@ -862,37 +862,33 @@ func (rs *Redshift) loadUserTables(ctx context.Context) map[string]error {
...
-		  (
-			_source.%[4]s = %[1]s.%[2]s.%[4]s
-		  );
-`,
-		rs.Namespace,
-		warehouseutils.UsersTable,
-		stagingTableName,
-		primaryKey,
-	)
+	if rs.enableDedup() {
```
fracasula, line 865: We weren't checking any flags for dedup here, just pinging you guys to make sure it makes sense to do so before I add the tests. @lvrach @achettyiitr

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -847,7 +847,9 @@ func (sf *Snowflake) LoadIdentityMappingsTable(ctx context.Context) error {
 // * the load table strategy is "append" mode
 // * the uploader says we can append
 func (sf *Snowflake) ShouldAppend() bool {
-	return sf.config.loadTableStrategy == loadTableStrategyAppendMode && sf.Uploader.CanAppend()
+	return sf.config.loadTableStrategy == loadTableStrategyAppendMode &&
+		sf.Uploader.CanAppend() &&
+		!sf.Warehouse.GetBoolDestinationConfig("enableDedup")
```
fracasula, line 852: Same question about the load table strategy @lvrach 

Also what happens if the [upload.CanAppend()](https://github.com/rudderlabs/rudder-server/blob/release/1.15.x/warehouse/router/upload.go#L724) returns `false` but the user enabled the dedup?

fracasula (commented):
@lvrach @achettyiitr hey guys, I left you a bunch of questions in the PR to double check with you a few things before I spend time with test coverage.

warehouse/integrations/deltalake/deltalake.go
fracasula: By default if nothing is set then we are in merge mode. Only if the customer opts-in we can go in append mode.

P.S.: `enableDedup` -> `enableAppend`.

fracasula: We need to document the behaviour of `upload.CanAppend()`. We can also move the `integration.warehouse.GetBoolDestinationConfig("enableAppend")` into the `upload.CanAppend()` method.

warehouse/integrations/postgres/load.go
fracasula: Remove these checks for users tables.

warehouse/integrations/bigquery/bigquery.go
fracasula: Add a way to disable the flag for all destinations.

fracasula: The flag coming from config.yaml will be called `forceMerge` (default `false`). The flag coming from the backend config should be called `enableAppend`.

fracasula:
memo to self: check if load table strategy is used for deltalake and others for our customers and replace accordingly.

fracasula:
@lvrach @achettyiitr hey guys, please review also the flow diagram and the outstanding questions (see PR description).

warehouse/integrations/postgres/load.go
```diff
@@ -143,7 +130,7 @@ func (pg *Postgres) loadTable(
 	}
 
 	var rowsDeleted int64
-	if !slices.Contains(pg.config.skipDedupDestinationIDs, pg.Warehouse.Destination.ID) {
+	if isUsersTable || pg.shouldMerge() {
```
achettyiitr, line 133: Since we already have a tableName in here, we can decide based on that? wdyt?
```suggestion
	if tableName == warehouseutils.UsersTable || pg.shouldMerge() {
```

warehouse/integrations/bigquery/bigquery_test.go
```diff
@@ -1042,3 +1057,14 @@ func TestUnsupportedCredentials(t *testing.T) {
...
+func cloneWarehouse(t *testing.T, wh model.Warehouse) model.Warehouse {
+	t.Helper()
+	buf, err := json.Marshal(wh)
+	require.NoError(t, err)
+
+	var clone model.Warehouse
+	require.NoError(t, json.Unmarshal(buf, &clone))
+
+	return clone
+}
```
achettyiitr, line 1070: Should we create a helper for this, since it's being used at couple of places?

```GO
func Clone[T](t *testing.T, s T) T {
	t.Helper()

	buf, err := json.Marshal(s)
	require.NoError(t, err)

	var clone T
	require.NoError(t, json.Unmarshal(buf, &clone))

	return clone
}
```

achettyiitr (changes requested):

warehouse/integrations/deltalake/deltalake.go
```diff
@@ -1385,9 +1382,10 @@ func (*Deltalake) DeleteBy(context.Context, []string, warehouseutils.DeleteByPar
...
-// ShouldAppend returns true if:
-// * the load table strategy is "append" mode
-// * the uploader says we can append
-func (d *Deltalake) ShouldAppend() bool {
-	return d.config.loadTableStrategy == appendMode && d.Uploader.CanAppend()
+// ShouldMerge returns true if:
+// * the uploader says we cannot append
+// * the user opted in to merging and we allow merging
+func (d *Deltalake) ShouldMerge() bool {
+	return !d.Uploader.CanAppend() ||
```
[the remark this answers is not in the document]
achettyiitr: Also, should we add destinationID-based flags or workspace-based flags here as well?

```GO
func (job *UploadJob) CanAppend() bool {
	if isSourceETL := job.upload.SourceJobRunID != ""; isSourceETL {
		return false
	}
	if job.warehouse.Source.IsReplaySource() {
		return false
	}
	if _, isMergeCategory := mergeSourceCategoryMap[job.warehouse.Source.SourceDefinition.Category]; isMergeCategory {
		return false
	}
	if job.upload.Retried {
		return false
	}

	a := job.conf.GetBool(fmt.Sprintf("Warehouse.%s.allowMerge", whutils.WHDestNameMap[job.upload.DestinationType]), true)
	b := slices.Contains(job.conf.GetStringSlice(fmt.Sprintf("Warehouse.%s.skipMergeDestinationIDs", whutils.WHDestNameMap[job.upload.DestinationType]), nil), job.upload.DestinationID)
	c := slices.Contains(job.conf.GetStringSlice(fmt.Sprintf("Warehouse.%s.skipMergeWorkspaceIDs", whutils.WHDestNameMap[job.upload.DestinationType]), nil), job.upload.WorkspaceID)
	d := job.warehouse.GetBoolDestinationConfig(model.EnableMergeSetting)
	if !a || !d || b || c {
		return false
	}

	return true
}
```

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -847,45 +853,40 @@ func Connect(context context.Context, cred *BQCredentials) (*bigquery.Client, er
...
-	client, err := Connect(ctx, &cred)
-	return client, err
+	c, err := Connect(ctx, &cred)
+	return c, err
 }
 
-func (bq *BigQuery) dedupEnabled() bool {
-	return bq.config.isDedupEnabled || bq.config.isUsersTableDedupEnabled
+// shouldMerge returns true if:
+// * the uploader says we cannot append
```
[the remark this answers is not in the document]
fracasula: Good catch, changed to:

```go
// ShouldMerge returns true if:
// * the server configuration says we can merge
// * the user opted-in
```

warehouse/internal/model/warehouse.go
```diff
@@ -2,6 +2,17 @@ package model
...
+type DestinationConfigSetting interface{ string() string }
+
+type whDestConfSetting string
+
+func (w whDestConfSetting) string() string { return string(w) }
+
+const (
+	EnableMergeSetting      whDestConfSetting = "enableMerge"
+	UseRudderStorageSetting whDestConfSetting = "useRudderStorage"
+)
```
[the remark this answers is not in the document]
fracasula: Yes, let's remove the prefix. The warehouse domain is already pretty clear given the package name.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -843,11 +823,12 @@ func (sf *Snowflake) LoadIdentityMappingsTable(ctx context.Context) error {
 	return nil
 }
 
-// ShouldAppend returns true if:
+// ShouldMerge returns true if:
 // * the load table strategy is "append" mode
 // * the uploader says we can append
```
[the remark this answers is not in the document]
fracasula: Good catch, changed to:

```go
// ShouldMerge returns true if:
// * the server configuration says we can merge
// * the user opted-in
```

I removed the uploader from the comment since it's not being called here.

warehouse/integrations/bigquery/bigquery_test.go
```diff
@@ -83,72 +77,64 @@ func TestIntegration(t *testing.T) {
...
-	defer cancel()
-
-	go func() {
-		r := runner.New(runner.ReleaseInfo{})
-		_ = r.Run(ctx, []string{"bigquery-integration-test"})
-
-		close(svcDone)
-	}()
-	t.Cleanup(func() { <-svcDone })
+	bootstrapSvc := func(t *testing.T, enableMerge bool) *bigquery.Client {
```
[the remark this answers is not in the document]
achettyiitr: My bad, it's being used from only one place and is not exactly a helper function. Plz ignore.

warehouse/integrations/bigquery/bigquery.go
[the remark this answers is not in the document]
fracasula: Changed to:

```go
// shouldMerge returns true if:
// * the server config says we allow merging
// * the user opted in to merging
```

warehouse/integrations/postgres/load.go
fracasula: Yeah I thought about that, but it felt more robust to have the caller tell  the `loadTable` what behaviour is needed. 

That's because the `loadTable` method seems to be unaware of the concept of "users", it doesn't care about the data, it just needs to load the table.

We could even rename `isUsersTable` to `forceMerge` and remove all references of users from `loadTable`. Thoughts?

achettyiitr: `forceMerge` seems to be a good option.

warehouse/integrations/bigquery/bigquery_test.go
fracasula: Done :+1:

achettyiitr (approved):

lvrach (approved):

PR #3966, opened 2023-10-12
feat: add support source transformation using go integrations

koladilip:
# Description

Plugins will be implemented in this [repo](https://github.com/rudderlabs/rudder-integration-plugins)
reference: https://github.com/rudderlabs/rudder-integration-plugins/blob/main/integrations/sources/webhook.go

## Linear Ticket

< Replace with Linear Link >

## Security


gateway/webhook/webhookTransformer.go
```diff
@@ -45,11 +50,53 @@ func (bt *batchWebhookTransformerT) markResponseFail(reason string) transformerR
...
+		}
+		if eventResponse.Output == nil && eventResponse.OutputToSource == nil {
+			batchResponse.responses[idx] = bt.markResponseFail(response.SourceTransformerFailedToReadOutput)
+			continue
+		}
+		bt.stats.receivedStat.Count(1)
+		batchResponse.responses[idx] = eventResponse
+	}
+	bt.stats.transformTimerStat.Since(transformStart)
+	bt.stats.receivedStat.Count(len(events))
```
fracasula, line 85: Does it make sense to both increment the `receivedStat` in the loop and then at the end with `Count(len(events))`? It looks like a mistake so I thought I'd double check with you.

koladilip: I will remove the inner one

gateway/webhook/webhookTransformer.go
```diff
@@ -45,11 +50,53 @@ func (bt *batchWebhookTransformerT) markResponseFail(reason string) transformerR
 	return resp
 }
 
+func (bt *batchWebhookTransformerT) transformUsingPlugin(plugin plugins.Plugin, payload []byte, sourceType string) transformerBatchResponseT {
+	transformStart := time.Now()
+	var events []map[string]interface{}
+	err := json.Unmarshal(payload, &events)
+	if err != nil {
+		bt.stats.failedStat.Count(len(events))
```
fracasula, line 58: Can we really count the `events` if the unmarshalling failed? :thinking: 
I'm afraid this might end up as a noop call like `Count(0)`.

koladilip: fixed this.

fracasula: Maybe a `failedBatch` metric can be useful so that we can `.Increment()` it. Wdyt?
Otherwise we won't have visibility about those failures in our metrics (unless we do in some other way?).

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/rudderlabs/rudder-server
 
-go 1.21
-
-toolchain go1.21.1
```
fracasula, line 5: Out of curiosity, is this intentional?

koladilip: actually no, go mod tidy was causing some issues so I was trying different things, we will add it back

gateway/webhook/webhookTransformer.go
koladilip: you mean whole batch is failed? that may not happening with plugin implementation.

fracasula: What I meant is that earlier you were bumping a metric if you couldn't unmarshal the payload. Now it's not there anymore so we don't have any visibility on that, right?
Given that and given that given that "N events = batch" I proposed something like `failedBatch.Increment()`. If you don't want to call it batch please propose a different name but it's a technicality. What I'm trying to say is that we don't have a metric there anymore :slightly_smiling_face:

fracasula: Maybe we can have `pluginTransformationSuccess` and `pluginTransformationFailure` that we can increment only once per method call. And then keep also the `receivedStat` with `receivedStat.Count(len(events))`. Wdyt?

koladilip: per event level we can add we can add pluginTransformationSuccess or pluginTransformationFailure, but since we are not doing any batch operation for all events so I think we need to skip failed batch. In reality the failedBatch should never trigger as we are generating that code, unless there is not test case for it.

koladilip: Updated can you see if that make sense

koladilip:
@chandumlg @krishna2020 can you share your feedback too

koladilip:
for now, we are closing this and resume back when required

PR #3987, opened 2023-10-18
fix: change null values to empty strings in spite of "null"

ItsSudip:
# Description

We were sending null values as "null". For null values, Eloqua doesn't show any warning but as we were sending "null" they were showing an invalid value warning. With this PR we are fixing that. We are also considering warning with statusCode `ELQ-00040` as a success when none of the unique fields are invalid. 

## Linear Ticket

>Resolves INT-856

## Security


koladilip (approved):

router/batchrouter/asyncdestinationmanager/eloqua/utils.go
```diff
@@ -93,7 +93,11 @@ func createCSVFile(fields []string, file *os.File, uploadJobInfo *JobInfo, jobId
 		}
 		var values []string
 		for _, field := range fields {
-			values = append(values, data.Message.Data[field].(string))
+			if data.Message.Data[field].(string) == "null" {
+				values = append(values, "")
+			} else {
+				values = append(values, data.Message.Data[field].(string))
+			}
```
cisse21, line 100:
```suggestion
             val, ok := data.Message.Data[field].(string)
			if !ok || val == "null" {
				values = append(values, "")
			} else {
				values = append(values, val)
			}
```

cisse21 (approved):

ItsSudip:
Created another pr to the release branch

PR #3990, opened 2023-10-18
chore: advertise gzip support to transformer through X-Feature-Gzip-Support header

Gauravudia:
# Description

Adding the gzip support header in transformer.go for backward compatibility between server <> transformer

## Linear Ticket

https://github.com/rudderlabs/rudder-server/pull/3904

## Security

### DIFF
--- a/processor/transformer/transformer.go
+++ b/processor/transformer/transformer.go
@@ -455,6 +455,7 @@ func (trans *handle) doPost(ctx context.Context, rawJSON []byte, url, stage stri
 				}
 
 				req.Header.Set("Content-Type", "application/json; charset=utf-8")
+				req.Header.Set("X-Feature-GZIP-SUPPORT", "?1")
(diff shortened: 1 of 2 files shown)


processor/transformer/transformer.go
```diff
@@ -455,6 +455,7 @@ func (trans *handle) doPost(ctx context.Context, rawJSON []byte, url, stage stri
 				}
 
 				req.Header.Set("Content-Type", "application/json; charset=utf-8")
+				req.Header.Set("X-Feature-Gzip-Support", "?1")
```
atzoum, line 458: question: why are we using `?1` as a value?

koladilip: any special meaning or something we come up with it. just curious.

koladilip (approved):

atzoum (approved):

PR #3995, opened 2023-10-19
feat: add version deprecation error-code identification logic

sanpj2292:
# Description

The idea is to identify version deprecation error category from error message & store it for reporting purposes

## Linear Ticket

< Replace with Linear Link >

## Security


enterprise/reporting/error_extractor.go
```diff
@@ -31,21 +32,52 @@ var (
...
-	log              logger.Logger
-	ErrorMessageKeys []string // the keys where in we may have error message
+	log                              logger.Logger
+	ErrorMessageKeys                 []string // the keys where in we may have error message
+	versionDeprecationThresholdScore misc.ValueLoader[int]
 }
 
 func NewErrorDetailExtractor(log logger.Logger) *ExtractorHandle {
 	errMsgKeys := config.GetStringSlice("Reporting.ErrorDetail.ErrorMessageKeys", []string{})
+	versionDepThreshold := config.GetReloadableIntVar(1, 1, "Reporting.ErrorDetail.versionDeprecationThresholdScore")
```
koladilip, line 73: should we keep min score to be 2 as version keep might come in different context also

sanpj2292: This is a reloadable configuration variable. We can put the desired value in operator

koladilip (approved):

enterprise/reporting/error_extractor.go
```diff
@@ -255,3 +287,22 @@ func (ext *ExtractorHandle) CleanUpErrorMessage(errMsg string) string {
 
 	return regexdMsg
 }
+
+func (ext *ExtractorHandle) GetErrorCode(errorMessage string) string {
+	// version deprecation logic
+	var score int
+	var errorCode string
+
+	// Check if the value is a string before converting to lowercase
```
chandumlg, line 296: Remove this comment.

enterprise/reporting/error_extractor.go
```diff
@@ -31,21 +32,52 @@ var (
 	whitespacesRegex = regexp.MustCompile("[ \t\n\r]*") // used in checking if string is a valid json to remove extra-spaces
 
 	defaultErrorMessageKeys = []string{"message", "description", "detail", "title", errorKey, "error_message"}
+	deprecationKeywords     = map[string]int{
```
chandumlg, line 35: Can we ensure lowercase of keys somehow (to make sure we don't have casing issues for the keys added in the future)?

chandumlg (approved):
Approving with minor comments.

enterprise/reporting/error_extractor.go
sanpj2292: Are you asking to deliberately lowercase even though this map now contains lowercased keys ?

sanpj2292: Done

PR #3996, opened 2023-10-19
fix: update error parsing of eloqua

ItsSudip:
# Description

We were sending null values as "null". For null values, Eloqua doesn't show any warning but as we were sending "null" they were showing an invalid value warning. With this PR we are fixing that. We are also considering warning with statusCode ELQ-00040 as a success when none of the unique fields are invalid.

## Linear Ticket

>Resolves INT-856

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/eloqua/manager.go
+++ b/router/batchrouter/asyncdestinationmanager/eloqua/manager.go
@@ -44,13 +44,14 @@ func NewManager(destination *backendconfig.DestinationT) (*EloquaBulkUploader, e
 
 func NewEloquaBulkUploader(destinationName, authorization, baseEndpoint string, eloqua EloquaService) *EloquaBulkUploader {
 	return &EloquaBulkUploader{
-		destName:      destinationName,
-		logger:        logger.NewLogger().Child("batchRouter").Child("AsyncDestinationManager").Child("Eloqua").Child("EloquaBulkUploader"),
-		authorization: authorization,
-		baseEndpoint:  baseEndpoint,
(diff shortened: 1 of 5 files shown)


router/batchrouter/asyncdestinationmanager/eloqua/manager.go
```diff
@@ -44,13 +44,14 @@ func NewManager(destination *backendconfig.DestinationT) (*EloquaBulkUploader, e
...
-		eventsLimit:   common.GetBatchRouterConfigInt64("MaxEventsLimit", destinationName, 1000000),
-		service:       eloqua,
-		jobToCSVMap:   map[int64]int64{},
+		destName:          destinationName,
+		logger:            logger.NewLogger().Child("batchRouter").Child("AsyncDestinationManager").Child("Eloqua").Child("EloquaBulkUploader"),
+		authorization:     authorization,
+		baseEndpoint:      baseEndpoint,
+		fileSizeLimit:     common.GetBatchRouterConfigInt64("MaxUploadLimit", destinationName, 32*bytesize.MB),
+		eventsLimit:       common.GetBatchRouterConfigInt64("MaxEventsLimit", destinationName, 1000000),
+		successStatusCode: common.GetBatchRouterConfigStringMap("SuccessStatusCode", destinationName, []string{"ELQ-00040"}),
```
koladilip, line 53: do we really need to get this from config? is it going to be varied for environment or customers?

ItsSudip: We are not able to reproduce all the statusCode so keeping this scope open if later we need to add any new status code we can directly do it from the env.

koladilip (approved):

PR #3999, opened 2023-10-19
fix: corrupted rsources stats captured by processor for dropped jobs

atzoum:
# Description

Using a unique job id instead of a zero (0) value for every dropped job so that stats are captured correctly.

## Problem
By using the same (0) job id, `in_count` was being incremented for all different stat key combinations (job_run_id, task_run_id, source_id & destination_id), but `failed_count` was being erroneously incremented always for the first stat key, due to a corrupted [jobIdsToStatKeyIndex](https://github.com/rudderlabs/rudder-server/blob/49b376fc8b8f4eb1cb20c4fd949f4bfa869845d6/services/rsources/stats_collector.go#L151-L152).

## Linear Ticket

Resolves PIPE-433

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -2566,10 +2566,13 @@ func (proc *Handle) transformSrcDest(
 	}
 }
 
-func (proc *Handle) saveDroppedJobs(failedJobs []*jobsdb.JobT, tx *jobsdb.Tx) error {
-	if len(failedJobs) > 0 {
+func (proc *Handle) saveDroppedJobs(droppedJobs []*jobsdb.JobT, tx *jobsdb.Tx) error {
+	if len(droppedJobs) > 0 {
+		for i := range droppedJobs { // each dropped job should have a unique jobID in the scope of the batch
+			droppedJobs[i].JobID = int64(i)
+		}
 		rsourcesStats := rsources.NewDroppedJobsCollector(proc.rsourcesService)
-		rsourcesStats.JobsDropped(failedJobs)
+		rsourcesStats.JobsDropped(droppedJobs)
 		return rsourcesStats.Publish(context.TODO(), tx.Tx)
 	}
 	return nil


processor/processor.go
```diff
@@ -2566,10 +2566,13 @@ func (proc *Handle) transformSrcDest(
 	}
 }
 
-func (proc *Handle) saveDroppedJobs(failedJobs []*jobsdb.JobT, tx *jobsdb.Tx) error {
-	if len(failedJobs) > 0 {
+func (proc *Handle) saveDroppedJobs(droppedJobs []*jobsdb.JobT, tx *jobsdb.Tx) error {
+	if len(droppedJobs) > 0 {
+		for i := range droppedJobs { // each dropped job should have a unique jobID in the scope of the batch
+			droppedJobs[i].JobID = int64(i)
```
lvrach, line 2572: should we make a copy of the `jobsdb.JobT` before setting the new `i` ?

Not sure how those `droppedJobs` will be used outside of this method

lvrach: Looking further at the code I understand that those jobs have not been inserted to jobsdb and thus don't have a jobID, is this the case ?

lvrach (approved):

processor/processor.go
atzoum: No, the only purpose of these dropped jobs is to help in capturing these stats. These are not actual jobsdb jobs, they get created [here](https://github.com/rudderlabs/rudder-server/blob/master/processor/events_response.go#L55) and never end up in any jobsdb

achettyiitr, line 2572: Curious can't we use [e.Metadata.JobID](https://github.com/rudderlabs/rudder-server/blob/10c0ffe63c7d472112e3316bd05157b85a2dccc5/processor/transformer/transformer.go#L61) in [here](https://github.com/rudderlabs/rudder-server/blob/f5b8e7b4976945207ec0c011ed6d7fe30e1eb323/processor/events_response.go#L55)?

atzoum: Not really, processor operates on messageIDs

achettyiitr (approved):

PR #4007, opened 2023-10-25
feat: new transformer service to fetch and serve transformer features

anantjain45823:
# Description
Resolves INT-543 and INT-768
**What do we want to achieve?**
      We want to send fetch the `sourceConfig` and send it to transformer for sources as well as we do for destinations.

**How we are planning to do that ?**
      We are introducing new endpoint `{transformUrl}/v1/sourceType` which would accept the data in the format 
       `{event: ingested event, source: sourceObject}`.
       
**How We are Maintaining Backward Compatibility ?**
      If some how transformer is old and **does not** support v1 endpoint then it could break. To control this we are 
       globalising the `/feature` endpoint from we already use in `processor.go` to check for router destinations.
       In feature json response we are checking for a boolean value to check if transformer support `v1` endpoint.
       Also for this `/feature` endpoint we made a new transformer service which is fetching the /feature frequently to be updated about is transformer supports source-V1-Endpoint or not.
       
Notion Research : https://www.notion.so/rudderstacks/Add-new-support-features-json-from-transformer-11deb41d5fe44470822e4b3a9873cd32?pvs=4

## Linear Ticket

https://linear.app/rudderstack/issue/INT-768/servertransformer-features-endpoint-extended-support
https://linear.app/rudderstack/issue/INT-543/send-sourceconfig-from-server-to-transformer

## Security

### DIFF
--- a/gateway/webhook/setup.go
+++ b/gateway/webhook/setup.go
@@ -9,14 +9,17 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
-	"golang.org/x/sync/errgroup"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
+
+	"golang.org/x/sync/errgroup"
+
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
 	gwstats "github.com/rudderlabs/rudder-server/gateway/internal/stats"
 	gwtypes "github.com/rudderlabs/rudder-server/gateway/internal/types"
 	"github.com/rudderlabs/rudder-server/gateway/webhook/model"
+	"github.com/rudderlabs/rudder-server/services/transformer"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 )
 
@@ -42,10 +45,8 @@ func newWebhookStats() *webhookStatsT {
 	return &wStats
 }
 
-func Setup(gwHandle Gateway, stat stats.Stats, opts ...batchTransformerOption) *HandleT {
+func Setup(gwHandle Gateway, transformerFeaturesService transformer.FeaturesService, stat stats.Stats, opts ...batchTransformerOption) *HandleT {
 	webhook := &HandleT{gwHandle: gwHandle, stats: stat, logger: logger.NewLogger().Child("gateway").Child("webhook")}
-
-	sourceTransformerURL := strings.TrimSuffix(config.GetString("DEST_TRANSFORM_URL", "http://localhost:9090"), "/") + "/v0/sources"
 	// Number of incoming webhooks that are batched before calling source transformer
 	webhook.config.maxWebhookBatchSize = config.GetReloadableIntVar(32, 1, "Gateway.webhook.maxBatchSize")
 	// Timeout after which batch is formed anyway with whatever webhooks are available
@@ -75,9 +76,16 @@ func Setup(gwHandle Gateway, stat stats.Stats, opts ...batchTransformerOption) *
 	for i := 0; i < maxTransformerProcess; i++ {
 		g.Go(misc.WithBugsnag(func() error {
 			bt := batchWebhookTransformerT{
-				webhook:              webhook,
-				stats:                newWebhookStats(),
-				sourceTransformerURL: sourceTransformerURL,
+				webhook: webhook,
+				stats:   newWebhookStats(),
+				sourceTransformAdapter: func(ctx context.Context) (sourceTransformAdapter, error) {
+					select {
+					case <-ctx.Done():
+						return nil, ctx.Err()
+					case <-transformerFeaturesService.Wait():
+						return newSourceTransformAdapter(transformerFeaturesService.SourceTransformerVersion()), nil
+					}
+				},
 			}
 			for _, opt := range opts {
 				opt(&bt)
--- a/gateway/webhook/webhook.go
+++ b/gateway/webhook/webhook.go
@@ -19,6 +19,7 @@ import (
 	"github.com/hashicorp/go-retryablehttp"
 	"github.com/samber/lo"
 
+	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
 	gwtypes "github.com/rudderlabs/rudder-server/gateway/internal/types"
@@ -85,9 +86,9 @@ type webhookStatsT struct {
 }
 
 type batchWebhookTransformerT struct {
-	webhook              *HandleT
-	stats                *webhookStatsT
-	sourceTransformerURL string
+	webhook                *HandleT
+	stats                  *webhookStatsT
+	sourceTransformAdapter func(ctx context.Context) (sourceTransformAdapter, error)
 }
 
 type batchTransformerOption func(bt *batchWebhookTransformerT)
@@ -262,6 +263,28 @@ func (webhook *HandleT) batchRequests(sourceDef string, requestQ chan *webhookT)
 // TODO : return back immediately for blank request body. its waiting till timeout
 func (bt *batchWebhookTransformerT) batchTransformLoop() {
 	for breq := range bt.webhook.batchRequestQ {
+
+		// If unable to fetch features from transformer, send GatewayTimeout to all requests
+		// TODO: Remove timeout from here after timeout handler is added in gateway
+		ctx, cancel := context.WithTimeout(context.Background(), config.GetDurationVar(10, time.Second, "WriteTimeout", "WriteTimeOutInSec"))
+		sourceTransformAdapter, err := bt.sourceTransformAdapter(ctx)
+		if err != nil {
+			for _, req := range breq.batchRequest {
+				req.done <- transformerResponse{StatusCode: response.GetErrorStatusCode(response.GatewayTimeout), Err: response.GetStatus(response.GatewayTimeout)}
+			}
+			cancel()
+			continue
+		}
+		cancel()
+
+		transformerURL, err := sourceTransformAdapter.getTransformerURL(breq.sourceType)
+		if err != nil {
+			for _, req := range breq.batchRequest {
+				req.done <- transformerResponse{StatusCode: response.GetErrorStatusCode(response.ServiceUnavailable), Err: response.GetStatus(response.ServiceUnavailable)}
+			}
+			continue
+		}
+
 		var payloadArr [][]byte
 		var webRequests []*webhookT
 		for _, req := range breq.batchRequest {
@@ -297,7 +320,14 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
 				continue
 			}
 
-			payloadArr = append(payloadArr, body)
+			payload, err := sourceTransformAdapter.getTransformerEvent(req.authContext, body)
+			if err != nil {
+				req.done <- transformerResponse{Err: response.GetStatus(response.InvalidWebhookSource)}
+				continue
+			}
+
+			payloadArr = append(payloadArr, payload)
+
 			webRequests = append(webRequests, req)
 		}
 
@@ -313,7 +343,7 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
 		bt.stats.sourceStats[breq.sourceType].numEvents.Count(len(payloadArr))
 
 		transformStart := time.Now()
-		batchResponse := bt.transform(payloadArr, breq.sourceType)
+		batchResponse := bt.transform(payloadArr, transformerURL)
 
 		// stats
 		bt.stats.sourceStats[breq.sourceType].sourceTransform.Since(transformStart)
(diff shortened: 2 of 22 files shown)


services/transformer/features.go
```diff
@@ -0,0 +1,45 @@
...
+	Log          logger.Logger
+	PollInterval time.Duration
+}
+
+type TransformerFeaturesService interface {
+	GetSourceTransformerVersion() string
+	Wait() chan struct{}
+}
+
+func NewTransformerFeatureService(ctx context.Context, config TransformerFeatureServiceConfig) (TransformerFeaturesService, error) {
```
anantjain45823, line 21: use of this ?

lvrach:
We should mark this type of change as a feature

anantjain45823:
> We should mark this type of change as a feature

Updated

processor/manager.go
```diff
@@ -115,33 +113,28 @@ func New(
 	transDebugger transformationdebugger.TransformationDebugger,
 	enrichers []enricher.PipelineEnricher,
 	opts ...Opts,
```
atzoum, line 115: please take care of compilation issues

atzoum:
@anantjain45823 please make sure that compilation and all pull request checks are passing before posting for review.

services/transformer/handler.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/rudderlabs/rudder-go-kit/config"
+	"github.com/rudderlabs/rudder-go-kit/logger"
+	"github.com/rudderlabs/rudder-server/utils/httputil"
+	"github.com/rudderlabs/rudder-server/utils/misc"
+)
+
+type transformerHandler struct {
+	logger    logger.Logger
+	asyncInit *misc.AsyncInit
```
atzoum, line 21: we don't need an `AsyncInit` here, just a `chan struct{}` since we don't need to wait for multiple asynchronous tasks to finish, just one

gateway/webhook/webhookTransformer.go
```diff
@@ -10,11 +10,69 @@ import (
...
+	version string
+	bt      *batchWebhookTransformerT
+}
+
+type V1Adapter struct {
+	version string
+	bt      *batchWebhookTransformerT
+}
+
+type V1TransformerEvent struct {
```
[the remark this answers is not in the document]
chandumlg: This is not truly the payload that we are sending to the transformer. Payload is an array of these.

gateway/webhook/webhookTransformer.go
```diff
@@ -10,11 +10,69 @@ import (
...
+	"github.com/rudderlabs/rudder-go-kit/config"
+	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 	"github.com/rudderlabs/rudder-server/gateway/response"
+	"github.com/rudderlabs/rudder-server/services/transformer"
 	"github.com/rudderlabs/rudder-server/utils/httputil"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 )
 
+type SourceTransformAdapter interface {
+	getTransformerEvent(sourceId string, body []byte) ([]byte, error)
```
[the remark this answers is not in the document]
chandumlg: Changing this depends on the response to the above comment.

gateway/webhook/webhook.go
```diff
@@ -297,7 +310,14 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
 				continue
 			}
 
-			payloadArr = append(payloadArr, body)
+			payload, err := sourceTransformAdapter.getTransformerEvent(req.authContext.SourceID, body)
```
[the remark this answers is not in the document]
chandumlg: Source is a huge object. It has source definition, list of destinations attached and each destination has destination config which is also huge. Instead of copying and passing it around for all requests, why not get it only when required?

gateway/webhook/setup.go
```diff
@@ -59,6 +62,7 @@ func Setup(gwHandle Gateway, stat stats.Stats, opts ...batchTransformerOption) *
 		webhook.config.sourceListForParsingParams[i] = strings.ToLower(s)
 	}
 
+	webhook.transformerFeaturesService = transformerFeaturesService
```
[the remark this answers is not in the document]
chandumlg: We need to save the transformerFeaturesService handle because, we need to wait till transformer features are fetched. This check is performed in `batchTransformLoop`.

services/transformer/features.go
```diff
@@ -0,0 +1,66 @@
...
+
+type TransformerFeaturesServiceConfig struct {
+	PollInterval             time.Duration
+	TransformerURL           string
+	FeaturesRetryMaxAttempts int
+}
+
+type TransformerFeaturesService interface {
+	GetSourceTransformerVersion() string
+	GetTransformerFeatures() json.RawMessage
```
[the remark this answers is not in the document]
chandumlg: Changed the function name to Get as per other comments suggestion.

services/transformer/handler.go
chandumlg: Used channel

gateway/webhook/setup.go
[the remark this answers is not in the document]
atzoum: I proposed an alternative approach before that didn't require storing it in a struct field

gateway/webhook/webhook.go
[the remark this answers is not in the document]
atzoum: authContext is a pointer, the source object can be passed by reference too. We just resolve things once and we are done, instead of constantly looking for them and exposing new methods/ dealing with locks

services/transformer/features.go
```diff
@@ -0,0 +1,65 @@
...
+
+type FeaturesServiceConfig struct {
+	PollInterval             time.Duration
+	TransformerURL           string
+	FeaturesRetryMaxAttempts int
+}
+
+type FeaturesService interface {
+	SourceTransformerVersion() string
+	Get() json.RawMessage
```
atzoum, line 24: The service shouldn't expose the raw json message. Part of the abstraction's responsibility is to encapsulate this and provide higher level functions
```suggestion
	RouterTransform(destType string) bool
```

gateway/webhook/webhookTransformer.go
```diff
@@ -10,11 +10,92 @@ import (
...
+	ID               string
+	OriginalID       string
+	Name             string
+	SourceDefinition backendconfig.SourceDefinitionT
+	Config           map[string]interface{}
+	Enabled          bool
+	WorkspaceID      string
+	WriteKey         string
+	Transient        bool
+}
```
atzoum, line 51: We shouldn't need yet-another source struct
```suggestion
```

gateway/webhook/webhookTransformer.go
```diff
@@ -10,11 +10,92 @@ import (
...
 	"github.com/rudderlabs/rudder-server/utils/misc"
 )
 
+type sourceTransformAdapter interface {
+	getTransformerEvent(sourceId string, body []byte) ([]byte, error)
+	getTransformerURL(sourceType string) string
+}
+
+type v0Adapter struct {
+	version string
```
atzoum, line 27: `version` field is redundant, each struct method can use the necessary constant

gateway/webhook/webhookTransformer.go
```diff
@@ -10,11 +10,92 @@ import (
...
+	bt      *batchWebhookTransformerT
```
atzoum, line 28: We shouldn't leak this whole dependency here just for getting a source for a sourceID (1 single resolver method).
Even better, we shouldn't require a resolver method at all

services/transformer/features.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+	"encoding/json"
+	"time"
+
+	"github.com/rudderlabs/rudder-go-kit/logger"
+)
+
+const (
+	V0_SOURCE_TRANSFORMER_VERSION = "v0"
+	V1_SOURCE_TRANSFORMER_VERSION = "v1"
```
atzoum, line 13:
```suggestion
	V0 = "v0"
	V1 = "v1"
```

services/transformer/handler.go
```diff
@@ -0,0 +1,115 @@
...
+	return t.features
+}
+
+func (t *featuresService) Wait() chan struct{} {
+	return t.waitChan
+}
+
+func (t *featuresService) init(ctx context.Context) {
+	go t.syncTransformerFeatureJson(ctx)
+}
```
atzoum, line 47: This seems to be redundant as a method, we may simply call `rruntime.Go(func() {t.syncTransformerFeatureJson(ctx)})` in `NewFeaturesService`

atzoum, line 47: This seems to be redundant as a method, we may simply call `rruntime.Go(func() {t.syncTransformerFeatureJson(ctx)})` in `NewFeaturesService`
```suggestion
```

services/transformer/features_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package transformer
```
atzoum, line 1: can rename this file to `features_impl.go`

services/transformer/handler.go
```diff
@@ -0,0 +1,115 @@
+package transformer
```
atzoum, line 1: can rename this file to `features_impl.go`

gateway/handle_webhook.go
```diff
@@ -52,3 +53,10 @@ func (gw *Handle) SaveWebhookFailures(reqs []*model.FailedWebhookPayload) error
 	defer cancel()
 	return gw.errDB.Store(ctx, jobs)
 }
+
+func (gw *Handle) GetSource(sourceID string) backendconfig.SourceT {
+	gw.configSubscriberLock.RLock()
+	defer gw.configSubscriberLock.RUnlock()
+
+	return gw.sourceIDSourceMap[sourceID]
+}
```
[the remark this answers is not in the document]
cisse21: I see that the source has been added to authCtx but this function isn't removed. Is this intentional?

services/transformer/features.go
```diff
@@ -0,0 +1,67 @@
...
+	RouterTransform(destType string) bool
+	Wait() chan struct{}
+}
+
+var defaultTransformerFeatures = `{
+	"routerTransform": {
+	  "MARKETO": true,
+	  "HS": true
+	}
+  }`
```
cisse21, line 34:
```suggestion
```
Do we still need this considering all the transformers are way beyond the expected version

services/transformer/features.go
```diff
@@ -0,0 +1,67 @@
...
+	return V0
+}
+
+func (*noopService) Wait() chan struct{} {
+	dummyChan := make(chan struct{})
+	close(dummyChan)
+	return dummyChan
+}
+
+func (*noopService) RouterTransform(destType string) bool {
```
cisse21, line 65:
```suggestion
func (*noopService) RouterTransform(_ string) bool {
```

gateway/handle_webhook.go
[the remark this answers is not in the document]
chandumlg: Now we have the source available in authCtx. This function is no longer necessary.

services/transformer/features.go
chandumlg: That is true. Din't want to break anything, so reused the code from processor.go.

services/transformer/features_impl.go
```diff
@@ -0,0 +1,117 @@
...
+	config   FeaturesServiceConfig
+	features json.RawMessage
+}
+
+func (t *featuresService) SourceTransformerVersion() string {
+	features := t.features
+	if t.features == nil {
+		features = json.RawMessage(defaultTransformerFeatures)
+	}
+
```
atzoum, line 30: This is no longer needed since we are initializing the field with the default value in `NewFeaturesService`
```suggestion
```

services/transformer/features_impl.go
```diff
@@ -0,0 +1,117 @@
...
+	return V0
+}
+
+func (t *featuresService) RouterTransform(destType string) bool {
+	features := t.features
+	if t.features == nil {
+		features = json.RawMessage(defaultTransformerFeatures)
+	}
+
```
atzoum, line 43: This is no longer needed since we are initializing the field with the default value in `NewFeaturesService`
```suggestion
```

chandumlg: Oh! I din't see that change.

atzoum (approved):

gateway/handle_webhook.go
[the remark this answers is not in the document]
chandumlg: Oh! I misread your comment before.

This function is removed now.

gateway/webhook/webhookTransformer.go
```diff
@@ -10,11 +10,75 @@ import (
...
+	switch version {
+	case "v1":
+		return &v1Adapter{}
+	}
+
+	return &v0Adapter{}
+}
+
+func getTransformerURL(version, sourceType string) string {
+	return strings.TrimSuffix(config.GetString("DEST_TRANSFORM_URL", "http://localhost:9090"), "/") + "/" + version + "/sources/" + sourceType
```
lvrach, line 79: Prefer [url.JoinPath](https://pkg.go.dev/net/url@go1.19beta1#JoinPath), for readability and safety.
```suggestion
        bashURL := config.GetString("DEST_TRANSFORM_URL", "http://localhost:9090")
	return url.JoinPath(bashURL, version, "sources", sourceType)
```

You might need to use [PathEscape](https://cs.opensource.google/go/go/+/go1.19beta1:src/net/url/url.go;l=281) if your `sourceType` is not safe to be encoded in the url

lvrach, line 79: I would also suggest here to avoid using `config.GetString` which uses the default config provider and instead DI `config.Config` in your adapter struct. 

This will provide isolation in configuring `DEST_TRANSFORM_URL` for your tests

gateway/webhook/webhookTransformer.go
```diff
@@ -45,17 +109,16 @@ func (bt *batchWebhookTransformerT) markResponseFail(reason string) transformerR
...
 
 	payload := misc.MakeJSONArray(events)
-	url := fmt.Sprintf(`%s/%s`, bt.sourceTransformerURL, strings.ToLower(sourceType))
-	resp, err := bt.webhook.netClient.Post(url, "application/json; charset=utf-8", bytes.NewBuffer(payload))
+	resp, err := bt.webhook.netClient.Post(sourceTransformerURL, "application/json; charset=utf-8", bytes.NewBuffer(payload))
 
 	bt.stats.transformTimerStat.Since(transformStart)
 	if err != nil {
-		err := fmt.Errorf("JS HTTP connection error to source transformer: URL: %v Error: %+v", url, err)
+		err := fmt.Errorf("JS HTTP connection error to source transformer: URL: %v Error: %+v", sourceTransformerURL, err)
```
lvrach, line 121: [minor] using err wrap since we change this code base
```suggestion
		err := fmt.Errorf("JS HTTP connection to source transformer (URL: %q): %w", sourceTransformerURL, err)
```

lvrach (approved):

cisse21 (approved):

PR #4009, opened 2023-10-25
feat: add support for sending form array

koladilip:
# Description
Currently, we are sending everything as simple string values converting array values in go format like below.
Example: ["hello", "world"] -> [hello world], now we will be handling arrays properly

## Linear Ticket

https://linear.app/rudderstack/issue/INT-850

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is used.


router/network_test.go
```diff
@@ -35,6 +35,43 @@ func (c *networkContext) Finish() {
...
+		structData.RequestMethod = "POST"
+		structData.Type = "REST"
+		structData.URL = testServer.URL
+		structData.UserID = "anon_id"
+		structData.Body = map[string]interface{}{
+			"FORM": map[string]interface{}{
+				"name":  "test",
+				"alias": alias,
+			},
+		}
```
achettyiitr, line 67: [minor] 
```suggestion
		structData := integrations.PostParametersT{
			RequestMethod: http.MethodPost,
			Type:          "REST",
			URL:           testServer.URL,
			UserID:        "anon_id",
			Body: map[string]interface{}{
				"FORM": map[string]interface{}{
					"name":  "test",
					"alias": alias,
				},
			},
		}
```

achettyiitr (approved):

router/network.go
```diff
@@ -124,8 +124,16 @@ func (network *netHandle) SendPost(ctx context.Context, structData integrations.
...
+					switch v := val.(type) {
+					case []string:
+						for _, listItem := range v {
+							formValues.Add(key, listItem)
+						}
+					default:
+						formValues.Add(key, fmt.Sprint(val))
+					}
 				}
+				headers["Content-Type"] = "application/x-www-form-urlencoded"
```
achettyiitr, line 136: Do we need this? Not mentioned in the PR description?

koladilip: normally transformer sets this but I thought, setting this from server is more apt as it is preparing the final payload

koladilip:
@krishna2020 we might need to do regression testing for this change right?

PR #4033, opened 2023-10-30
fix(destination): oauth stats prefix

sanpj2292:
# Description

Prefix for some of the stats in OAuth module was missing. We have made some defaults instead of relying on the ones that are sent as arguments.

## Linear Ticket

Resolves INT-939

## Security


fracasula (commented):
`GetTokenInfo` is also creating stats like this:

```go
defer func() {
	authStats.statName = fmt.Sprintf("%v_total_req_latency", refTokenParams.EventNamePrefix)
	// ...
}()
```

I know that it's called at the moment by `FetchToken` and `RefreshToken` only but the method is part of the public API. Since it is exported, someone might call this method and pass an empty prefix and break the metrics again.

Also the approach seems a bit fragile in my opinion, why not just do:

```go
defer func() {
	authStats.statName = "total_req_latency"
	if refTokenParams.EventNamePrefix != "" {
		authStats.statName = refTokenParams.EventNamePrefix + "_" + authStats.statName
	}
	// ...
}()
```

This way we would also decrease the risk of other people breaking this code.

sanpj2292:
> `GetTokenInfo` is also creating stats like this:
> 
> ```go
> defer func() {
> 	authStats.statName = fmt.Sprintf("%v_total_req_latency", refTokenParams.EventNamePrefix)
> 	// ...
> }()
> ```
> 
> I know that it's called at the moment by `FetchToken` and `RefreshToken` only but the method is part of the public API. Since it is exported, someone might call this method and pass an empty prefix and break the metrics again.
> 
> Also the approach seems a bit fragile in my opinion, why not just do:
> 
> ```go
> defer func() {
> 	authStats.statName = "total_req_latency"
> 	if refTokenParams.EventNamePrefix != "" {
> 		authStats.statName = refTokenParams.EventNamePrefix + "_" + authStats.statName
> 	}
> 	// ...
> }()
> ```
> 
> This way we would also decrease the risk of other people breaking this code.

There is no public interface to call `GetTokenInfo`. It can only be invoked by either `FetchToken` or `RefreshToken`. Do you still think the approach is fragile ?
Also the reason why we are sending `EventNamePrefix` as argument is to know who is the caller, is it being called by `RefreshToken` or is it being called from `FetchToken`

cisse21:
> > `GetTokenInfo` is also creating stats like this:
> > ```go
> > defer func() {
> > 	authStats.statName = fmt.Sprintf("%v_total_req_latency", refTokenParams.EventNamePrefix)
> > 	// ...
> > }()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I know that it's called at the moment by `FetchToken` and `RefreshToken` only but the method is part of the public API. Since it is exported, someone might call this method and pass an empty prefix and break the metrics again.
> > Also the approach seems a bit fragile in my opinion, why not just do:
> > ```go
> > defer func() {
> > 	authStats.statName = "total_req_latency"
> > 	if refTokenParams.EventNamePrefix != "" {
> > 		authStats.statName = refTokenParams.EventNamePrefix + "_" + authStats.statName
> > 	}
> > 	// ...
> > }()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This way we would also decrease the risk of other people breaking this code.
> 
> There is no public interface to call `GetTokenInfo`. It can only be invoked by either `FetchToken` or `RefreshToken`. Do you still think the approach is fragile ? Also the reason why we are sending `EventNamePrefix` as argument is to know who is the caller, is it being called by `RefreshToken` or is it being called from `FetchToken`
> 
> Let me know your thoughts

If we just want to know who is the caller wouldn't it be better to pass it as a tag rather than being present in stat name?

fracasula:
> There is no public interface to call `GetTokenInfo`. 

`GetTokenInfo` **is** the public API and with API I mean package API because the method is exported so it can be called by whomever imports the package and break things.

What about @cisse21 proposal?

sanpj2292:
> If we just want to know who is the caller wouldn't it be better to pass it as a tag rather than being present in stat name?

Yes. It makes sense to send them as tags. Changing this now, doesn't make a lot of sense but I can work on a fix that would make sure we would not get empty `EventNamePrefix`

There is an item to re-work on OAuth implementation where in this point will be kept in mind & then re-worked accordingly

sanpj2292:
If say there is dis-comfort about having `GetTokenInfo` as public method, I can make it private for now to stop someone from mis-using this. Let me know what you think

fracasula:
@sanpj2292 wdyt of this instead? https://github.com/rudderlabs/rudder-server/pull/4044

> If say there is dis-comfort about having `GetTokenInfo` as public method, I can make it private for now to stop someone from mis-using this. Let me know what you think

It's not about discomfort, it's effectively a private method but it's exported. The package API should be designed by keeping in mind what is accessible from outside and what is not. That method should not be accessible, right?

sanpj2292:
@fracasula @cisse21 I have made changes as requested, can you guys please check and let me know if they look good ?

fracasula (commented):
I can't find any calls to `SendTimerStats` or `SendCountStats` with tags. 
Wouldn't it make sense to just leave the metric names simpler (i.e. instead of `token_request_latency` => `request_latency`) and then pass a tag like `action=fetch_token` or `refresh_token`, etc...?

sanpj2292:
> I can't find any calls to SendTimerStats or SendCountStats with tags.

This is done to add more flexibility. Can remove it as it is not being used

> Wouldn't it make sense to just leave the metric names simpler (i.e. instead of token_request_latency => request_latency) and then pass a tag like action=fetch_token or refresh_token, etc...?

It feels incomplete I've added `oauth_action` as a prefix to all stats.

sanpj2292:
> I can't find any calls to SendTimerStats or SendCountStats with tags.

Removed it

fracasula (approved):

koladilip (approved):

PR #4042, opened 2023-10-30
feat: introducing warehouse repo withTx

achettyiitr:
# Description

- Introducing `WithTx` to use wherever transactions are used.
- Also, this is a follow-up on #4045 where I am doing a cleanup for updating fields for the uploads table. `WithTx` is required during reporting where we update fields and report to reporting in the same transaction.

## Linear Ticket

- Resolves PIPE-462

## Security


warehouse/internal/repo/repo.go
```diff
@@ -18,3 +19,21 @@ func WithNow(now func() time.Time) Opt {
...
+
+func (r *repo) WithTx(f func(tx *sqlmiddleware.Tx) error) error {
+	tx, err := r.db.Begin()
+	if err != nil {
+		return fmt.Errorf("begin transaction: %w", err)
+	}
+
+	if err := f(tx); err != nil {
+		if rollbackErr := tx.Rollback(); rollbackErr != nil {
+			return fmt.Errorf("rollback transaction for %w: %w", err, rollbackErr)
```
fracasula, line 31: This is like Nando's double :chicken: wrap but with errors :sweat_smile:

fracasula (approved):

warehouse/internal/repo/repo.go
achettyiitr: I am a bit skeptical here. Should we ignore rollback error?

fracasula: Nah, it's OK. It works perfectly fine and it is possible to unwrap both errors :+1:

warehouse/internal/repo/load.go
```diff
@@ -60,58 +59,47 @@ func (repo *LoadFiles) DeleteByStagingFiles(ctx context.Context, stagingFileIDs
...
-	}
-	err = txn.Commit()
-	if err != nil {
-		return fmt.Errorf(`inserting load files: commit: %w`, err)
-	}
-	return
+		defer func() { _ = stmt.Close() }()
+
+		for _, loadFile := range loadFiles {
+			metadata := fmt.Sprintf(`{"content_length": %d, "destination_revision_id": %q, "use_rudder_storage": %t}`, loadFile.ContentLength, loadFile.DestinationRevisionID, loadFile.UseRudderStorage)
```
lvrach, line 85: [minor] I would prefer a proper` json.Marshal()`, IMO is cleaner and safer

warehouse/internal/repo/repo.go
lvrach: I wasn't aware that is possible to wrap two errors that the same time, cool

lvrach (approved):

PR #4057, opened 2023-10-31
chore: state for uploads

achettyiitr:
# Description

- Initial efforts to break the uploads file into smaller chunks so that it would be easier to review and understand the code.
- Some feedback around it in [here](https://github.com/rudderlabs/rudder-server/pull/3965#discussion_r1371729057).
- **Note: There is no change from the existing code.**
- There is a follow-up https://github.com/rudderlabs/rudder-server/pull/4058

## Linear Ticket

- Resolves PIPE-466 

## Security


lvrach (approved):

warehouse/router/state.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, uploadState := range stateTransitions {
+		if currentState == uploadState.inProgress || currentState == uploadState.failed {
+			return uploadState
+		}
+		if currentState == uploadState.completed {
+			return uploadState.nextState
+		}
+	}
+	return nil
+}
```
cisse21, line 102:
```suggestion
func nextState(currentState string) *state {
	if _, ok := stateTransitions[currentState]; ok {
		return stateTransitions[currentState].nextState
	}

	for _, uploadState := range stateTransitions {
		if currentState == uploadState.inProgress || currentState == uploadState.failed {
			return uploadState
		}
	}
	return nil
}
```

cisse21 (approved):

PR #4081, opened 2023-11-06
chore: adapt health.WaitUntilReady

achettyiitr:
# Description

- Adapt health.WaitUntilReady

## Linear Ticket

- Resolves PIPE-495

## Security


warehouse/app_test.go
```diff
@@ -4,11 +4,12 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"net/http"
 	"strconv"
 	"testing"
 	"time"
 
+	"github.com/rudderlabs/rudder-server/testhelper/health"
```
lvrach, line 11: We should move it to go-kit

lvrach (approved):

warehouse/app_test.go
achettyiitr: Good point. @atzoum @WFrancois wdyt?

cisse21 (approved):

warehouse/app_test.go
fracasula: I agree. Here a couple of (optional) suggestions:
1. make it so that it receives an `*http.Request` instead then add an alias that does a simple `GET` to the endpoint by calling the main function that takes a request
    * this is to help in cases where headers need to be passed. also in those cases there would be no need to build the request each time, we can build it once and then `Do`
3. make it return the response body so that we can make assertions on that once we know that the status code is the desired one
4. stop the `probe` inside the function once we're done "probing"
5. rename it to `WaitUntilStatusCode`, also alias for `OK` like in bullet point 1 can be useful

It's OK also as it is, those are just some ideas, we can discuss them in the go-kit PR since not all of them might be useful I guess.

PR #4088, opened 2023-11-06
chore: preferAppend defaults to false if not defined

fracasula:
# Description

Follow up to [this PR](https://github.com/rudderlabs/rudder-server/pull/3965) where we changed the default behaviour for Warehouse destinations so that users would have to opt-in if they want to merge (aka remove duplicates).

Given that the UI changes for enabling the merge are not live we're setting `enableMerge` by default to `true` if it's not set. This maintains backwards compatibility and allows us to decouple the data plane deployment from the control plane one.

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-497/default-enablemerge-to-true-if-neither-true-of-false-are-present) >

## Security


fracasula:
@achettyiitr @lvrach I'm proposing this approach. I have not updated the tests just yet, I'm just showcasing the changes here. Once I get a green light from everybody (product and CS as well) then I can update the tests. Let me know if it doesn't make sense somehow.

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -867,7 +867,7 @@ func (bq *BigQuery) connect(ctx context.Context, cred BQCredentials) (*bigquery.
 // * the server config says we allow merging
 // * the user opted in to merging
 func (bq *BigQuery) shouldMerge() bool {
-	return bq.config.allowMerge && bq.warehouse.GetBoolDestinationConfig(model.EnableMergeSetting)
+	return bq.config.allowMerge && bq.warehouse.GetEnableMergeSetting()
```
lvrach, line 870:
```suggestion
	return bq.config.allowMerge && !bq.warehouse.PreferAppend()
```

What if we inverse our logic here. Instead of enableMerge, we change the option to preferAppend false by default, customers can opt-in from UI.

koladilip: Yes if we want to existing customer to use Merge then no migration is required

fracasula: > if it is null in the config, what will happen if you just edit and save without changing anything ?

in this case `enableMerge` would become `false` making it more risky than `preferAppend`. will rename the field to `preferAppend` to make it more robust.

achettyiitr (approved):

fracasula:
Closing in favour of https://github.com/rudderlabs/rudder-server/pull/4098

PR #4098, opened 2023-11-08
fix: preferAppend defaults to false if not defined

fracasula:
# Description

Follow up to [this PR](https://github.com/rudderlabs/rudder-server/pull/3965) where we changed the default behaviour for Warehouse destinations so that users would have to opt-in if they want to merge (aka remove duplicates).

Given that the UI changes for enabling the merge are not live we're setting `enableMerge` by default to `true` if it's not set. This maintains backwards compatibility and allows us to decouple the data plane deployment from the control plane one.

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-497/default-enablemerge-to-true-if-neither-true-of-false-are-present) >

## Security

### DIFF
--- a/warehouse/internal/model/warehouse.go
+++ b/warehouse/internal/model/warehouse.go
@@ -9,7 +9,7 @@ type destConfSetting string
 func (s destConfSetting) string() string { return string(s) }
 
 const (
-	EnableMergeSetting      destConfSetting = "enableMerge"
+	PreferAppendSetting     destConfSetting = "preferAppend"
 	UseRudderStorageSetting destConfSetting = "useRudderStorage"
 )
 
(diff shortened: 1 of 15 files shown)


lvrach (approved):

warehouse/internal/model/warehouse.go
```diff
@@ -31,3 +31,12 @@ func (w *Warehouse) GetBoolDestinationConfig(key DestinationConfigSetting) bool
...
 	return false
 }
+
+func (w *Warehouse) GetPreferAppendSetting() bool {
+	destConfig := w.Destination.Config
+	value, ok := destConfig[PreferAppendSetting.string()].(bool)
+	if !ok {
+		return false // default value for backwards compatibility
+	}
+	return value
```
atzoum, line 41: **trivial:** just sparing 3 lines :)
```suggestion
	value, _ := destConfig[PreferAppendSetting.string()].(bool)
	return value
```

achettyiitr (approved):

PR #4108, opened 2023-11-09
chore: remove panic and continue with old config on incremental update failure

Sidddddarth:
# Description

incremental update failure causes a panic. Removing that panic.

## Linear Ticket

[Slack thread for context](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1699502654415459)

## Security

### DIFF
--- a/backend-config/backend-config.go
+++ b/backend-config/backend-config.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/url"
 	"reflect"
@@ -177,6 +178,9 @@ func (bc *backendConfigImpl) configUpdate(ctx context.Context) {
 	if err != nil {
 		statConfigBackendError.Increment()
 		pkgLogger.Warnf("Error fetching config from backend: %v", err)
+		if errors.Is(err, ErrIncrementalUpdateFailed) {
+			stats.Default.NewStat("config_backend.ErrIncrementalUpdateFailed", stats.CountType).Increment()
+		}
 
 		bc.initializedLock.RLock()
 		if bc.initialized {
--- a/backend-config/namespace_config.go
+++ b/backend-config/namespace_config.go
@@ -20,8 +20,9 @@ import (
 )
 
 var (
-	jsonfast               = jsoniter.ConfigCompatibleWithStandardLibrary
-	updatedAfterTimeFormat = "2006-01-02T15:04:05.000Z"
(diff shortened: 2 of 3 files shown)


backend-config/namespace_config.go
```diff
@@ -139,10 +140,16 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
 		if workspace == nil { // this workspace was not updated, populate it with the previous config
 			previousConfig, ok := nc.workspacesConfig[workspaceID]
 			if !ok {
-				panic(fmt.Errorf(
-					"workspace %q was not updated but was not present in previous config: %+v",
-					workspaceID, req,
-				))
+				nc.logger.Errorw(
+					"workspace was not updated but was not present in previous config",
```
atzoum, line 144: we can just add an alert on this log right?

backend-config/backend-config.go
```diff
@@ -177,6 +178,9 @@ func (bc *backendConfigImpl) configUpdate(ctx context.Context) {
 	if err != nil {
 		statConfigBackendError.Increment()
 		pkgLogger.Warnf("Error fetching config from backend: %v", err)
+		if errors.Is(err, ErrIncrementalUpdateFailed) {
+			stats.Default.NewStat("config_backend.ErrIncrementalUpdateFailed", stats.CountType).Increment()
+		}
```
atzoum, line 183: do we need another stat?

backend-config/namespace_config.go
```diff
@@ -139,10 +140,16 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
...
+					"workspaceID", workspaceID,
+					"req", req,
+				)
+				// reset state here
+				nc.lastUpdatedAt = time.Time{}
```
atzoum, line 149: `namespaceConfig#Get` can react on the `ErrIncrementalUpdateFailed `by resetting the `lastUpdatedAt` and calling `getFromAPI` again. This way we won't have to wait for the next poll to happen

atzoum:
If we want this to be released as a patch release ASAP it needs to target `1.16.x`, otherwise it will be included in the upcoming `1.17.x` release

Sidddddarth:
> If we want this to be released as a patch release ASAP it needs to target `1.16.x`, otherwise it will be included in the upcoming `1.17.x` release


My bad
Opened [here](https://github.com/rudderlabs/rudder-server/pull/4109)

PR #4136, opened 2023-11-14
fix: add autovacuum_vacuum_cost_limit to the reports table [PIPE-512]

BonapartePC:
# Description

add `autovacuum_vacuum_cost_limit` to the `reports` table

## Security


lvrach:
we wanted to do this for report table right? Do we also see value for jobsdb ?

BonapartePC:
> we wanted to do this for report table right? Do we also see value for jobsdb ?

I'm not sure if we can set it only for reports table. Looks like, this can only be set at database level

BonapartePC:
> we wanted to do this for report table right? Do we also see value for jobsdb ?

Made the changes to add for `reports` table

enterprise/reporting/reporting.go
```diff
@@ -135,7 +138,10 @@ func (r *DefaultReporter) DatabaseSyncer(c types.SyncerConfig) types.ReportingSy
 		MigrationsTable:            "reports_migrations",
 		ShouldForceSetLowerVersion: config.GetBool("SQLMigrator.forceSetLowerVersion", true),
 	}
-	err = m.Migrate("reports")
+	templateData := map[string]int{
+		"AutoVacuumCostLimit": r.autoVacuumCostLimit.Load(),
+	}
+	err = m.MigrateFromTemplates("reports", templateData)
```
achettyiitr, line 144: Question: If we set this in migration, we can change the configuration right, since it will be applied only once?

BonapartePC: You are right. There is no point int making it a reloadable int

sql/migrations/reports/000006_alter_reports_auto_vaccum.up.tmpl
```diff
@@ -0,0 +1 @@
+ALTER TABLE reports SET (autovacuum_vacuum_cost_limit = {{ .AutoVacuumCostLimit }});
```
lvrach, line 1: Will this migration run if we change the configuration? It is better to run this query after migration has been completed to ensure it always going to apply when the server starts.

enterprise/reporting/reporting.go
```diff
@@ -84,6 +85,7 @@ func NewDefaultReporter(ctx context.Context, log logger.Logger, configSubscriber
 	maxConcurrentRequests := config.GetReloadableIntVar(32, 1, "Reporting.maxConcurrentRequests")
 	maxOpenConnections := config.GetIntVar(32, 1, "Reporting.maxOpenConnections")
 	dbQueryTimeout = config.GetReloadableDurationVar(60, time.Second, "Reporting.dbQueryTimeout")
+	autoVacuumCostLimit := config.GetIntVar(2000, 1, "Reporting.autoVacuumCostLimit")
```
achettyiitr, line 88: We can keep the default value to 200
```suggestion
	autoVacuumCostLimit := config.GetIntVar(200, 1, "Reporting.autoVacuumCostLimit")
```

enterprise/reporting/reporting.go
```diff
@@ -135,7 +138,10 @@ func (r *DefaultReporter) DatabaseSyncer(c types.SyncerConfig) types.ReportingSy
 		MigrationsTable:            "reports_migrations",
 		ShouldForceSetLowerVersion: config.GetBool("SQLMigrator.forceSetLowerVersion", true),
 	}
-	err = m.Migrate("reports")
+	templateData := map[string]int{
+		"AutoVacuumCostLimit": r.autoVacuumCostLimit,
+	}
+	err = m.MigrateFromTemplates("reports", templateData)
```
achettyiitr, line 144: Any reason not applying everytime when the server starts? Also, we would not be requried to ALTER if the configuration is not set
```suggestion
	err = m.Migrate("reports")
    if err != nil {
		panic(fmt.Errorf("could not run reports migrations: %w", err))
	}
	if config.IsSet("Reporting.autoVacuumCostLimit") {
		_, err = dbHandle.ExecContext(ctx, fmt.Sprintf("ALTER TABLE reports SET (autovacuum_vacuum_cost_limit = %d", config.GetInt("Reporting.autoVacuumCostLimit", 200)))
		if err != nil {
			panic(err)
		}
	}
```

sql/migrations/reports/000006_alter_reports_auto_vaccum.up.tmpl
BonapartePC: Running the query after migration now.

enterprise/reporting/reporting.go
BonapartePC: Done

enterprise/reporting/reporting.go
BonapartePC: Done

achettyiitr (approved):

enterprise/reporting/reporting.go
```diff
@@ -139,6 +139,12 @@ func (r *DefaultReporter) DatabaseSyncer(c types.SyncerConfig) types.ReportingSy
 	if err != nil {
 		panic(fmt.Errorf("could not run reports migrations: %w", err))
 	}
+	if config.IsSet("Reporting.autoVacuumCostLimit") {
+		_, err = dbHandle.ExecContext(r.ctx, fmt.Sprintf("ALTER TABLE reports SET (autovacuum_vacuum_cost_limit = %d);", config.MustGetInt("Reporting.autoVacuumCostLimit")))
+		if err != nil {
+			panic(err)
+		}
+	}
```
atzoum, line 147: Do we need this to be configurable?

atzoum: If we want this to run always we can create a new reports_always folder and run the migrator with `RunAlways: true`

BonapartePC: Any reason why you think the mentioned change is a better way to go? 🤔

atzoum: Next time we need to add another change that needs to run always we'll just need to prepare a new file, not another if statement

BonapartePC: Makes sense. Made the change

enterprise/reporting/reporting.go
```diff
@@ -139,6 +139,20 @@ func (r *DefaultReporter) DatabaseSyncer(c types.SyncerConfig) types.ReportingSy
 	if err != nil {
 		panic(fmt.Errorf("could not run reports migrations: %w", err))
 	}
+	if config.IsSet("Reporting.autoVacuumCostLimit") {
```
atzoum, line 142: - We can execute the migrator regardless for this option.
- We can pass config in the template context
- The template itself can have an if clause checking the existence of the config option (yes go templates can call functions)

BonapartePC: Made the change

lvrach (approved):

services/sql-migrator/migrator_test.go
```diff
@@ -42,8 +43,14 @@ func TestMigrate(t *testing.T) {
 				MigrationsTable: fmt.Sprintf("migrations_%s", dir),
 				Handle:          postgre.DB,
 			}
-
-			err := m.Migrate(dir)
+			var err error
+			if strings.HasPrefix(dir, "reports_always") {
+				err = m.MigrateFromTemplates("reports_always", map[string]interface{}{
```
atzoum, line 48: can we validate the autovacuum sql by setting the `Reporting.autoVacuumCostLimit` option?

BonapartePC: Done

atzoum (approved):

PR #4137, opened 2023-11-14
chore: filtered and error stats separated in processor

Sidddddarth:
# Description

separates filtered and error stats in processor.

## Linear Ticket

[slack thread](https://rudderlabs.slack.com/archives/C03CX31FNAG/p1699905141075659)

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1132,6 +1132,30 @@ func (proc *Handle) getNonSuccessfulMetrics(
 	return m
 }
 
+func procFilteredCountStat(destType, pu, statusCode string) {
+	stats.Default.NewTaggedStat(
+		"proc_filtered_counts",
+		stats.CountType,
+		stats.Tags{
+			"destName":   destType,
+			"statusCode": statusCode,
+			"stage":      pu,
+		},
+	).Increment()
+}
+
+func procErrorCountsStat(destType, pu, statusCode string) {
+	stats.Default.NewTaggedStat(
+		"proc_error_counts",
+		stats.CountType,
+		stats.Tags{
+			"destName":   destType,
+			"statusCode": statusCode,
+			"stage":      pu,
+		},
+	).Increment()
+}
+
 func (proc *Handle) getTransformationMetrics(
 	transformerResponses []transformer.TransformerResponse,
 	state string,
@@ -1144,9 +1168,9 @@ func (proc *Handle) getTransformationMetrics(
 	statusDetailsMap := make(map[string]map[string]*types.StatusDetail)
 	countMap := make(map[string]int64)
 	var jobs []*jobsdb.JobT
-	isDropped := "false"
+	statFunc := procErrorCountsStat
 	if state == jobsdb.Filtered.State {
-		isDropped = "true"
+		statFunc = procFilteredCountStat
 	}
 	for i := range transformerResponses {
 		failedEvent := &transformerResponses[i]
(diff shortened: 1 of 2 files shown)


processor/processor.go
```diff
@@ -1088,24 +1088,59 @@ func (proc *Handle) updateMetricMaps(
 	sd.ViolationCount += int64(veCount)
 }
 
-func (proc *Handle) getNonSuccessfulMetrics(response transformer.Response, commonMetaData *transformer.Metadata, eventsByMessageID map[string]types.SingularEventWithReceivedAt, inPU, pu string) *NonSuccessfulTransformationMetrics {
+func (proc *Handle) getNonSuccessfulMetrics(
+	response transformer.Response,
+	commonMetaData *transformer.Metadata,
+	eventsByMessageID map[string]types.SingularEventWithReceivedAt,
+	inPU, pu string,
+) *NonSuccessfulTransformationMetrics {
```
atzoum, line 1096: **Personal preference:** I don't find that by splitting a single line, that fits perfectly in the screen, into multiple lines makes the codebase more readable

processor/processor.go
```diff
@@ -1176,6 +1219,7 @@ func (proc *Handle) getTransformationMetrics(transformerResponses []transformer.
 			"destName":   commonMetaData.DestinationType,
 			"statusCode": strconv.Itoa(failedEvent.StatusCode),
 			"stage":      pu,
+			"isDropped":  isDropped,
```
atzoum, line 1222: Why don't we use the actual state instead (aborted, filtered)?

Sidddddarth: there's some confusion between different teams using different vocabulary.
Also, filtered might mean dropped at a point and considered a success at some other point.
But `isDropped` - true/false can only mean one thing

processor/processor.go
Sidddddarth: It's not fitting on my screen, I'm having to turn my head and sometimes worse: scroll!
To each their own though, can revert if it's becoming a nuisance for everyone

processor/processor.go
atzoum: Since filtered is not a proc_error, can't we just not include them in this metric? 
If we want to capture a counter for filtered events in the future we could use another measurement name altogether. Adding filtered events in `proc_error_counts` using an additional label seems to be a bit unintuitive to me

processor/processor.go
```diff
@@ -1088,24 +1088,59 @@ func (proc *Handle) updateMetricMaps(
...
+) ([]*jobsdb.JobT, []*types.PUReportedMetric, map[string]int64) {
 	metrics := make([]*types.PUReportedMetric, 0)
 	connectionDetailsMap := make(map[string]*types.ConnectionDetails)
 	statusDetailsMap := make(map[string]map[string]*types.StatusDetail)
 	countMap := make(map[string]int64)
 	var jobs []*jobsdb.JobT
+	isDropped := "false"
+	if state == jobsdb.Filtered.State {
+		isDropped = "true"
+	}
```
cisse21, line 1143:
```suggestion
	isDropped := strconv.FormatBool(state == jobsdb.Filtered.State)
```

processor/processor.go
Sidddddarth: true
pushed some changes
please take another look

processor/processor.go
Sidddddarth: will try to keep this in mind next time

atzoum (approved):

PR #4170, opened 2023-11-22
chore(processor): include partition tag in stats

atzoum:
# Description

Adding the partition tag in processor stats for better observability during incidents

resolves PIPE-567

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -157,43 +157,43 @@ type Handle struct {
 	sourceObservers []sourceObserver
 }
 type processorStats struct {
-	statGatewayDBR                stats.Measurement
-	statGatewayDBW                stats.Measurement
-	statRouterDBW                 stats.Measurement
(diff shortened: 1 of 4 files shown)


lvrach (commented):
One concern I have here is with the cardinality in hosted env. Do you think is going to be an issue ?

atzoum:
> One concern I have here is with the cardinality in hosted env. Do you think is going to be an issue ?

We are using workspace level isolation in hosted & other component metrics (router/batch router/archiver) are already using the partition label in their metrics. Cardinality will increase, but we should manage it.

If it becomes an issue we can quickly remove the partition tag through configuration (`statsExcludedTags`)

Sidddddarth (approved):
Many of these stat `func`s are only used in one place. Inlining them could help avoid the jumps thru the codebase.
LGTM otherwise

cisse21 (approved):

PR #4171, opened 2023-11-22
chore(router): interruption of iterator should cause router to sleep

atzoum:
# Description

In case router iterator is interrupted prematurely, limits should not be considered to be reached, i.e. router shouldn't immediately run the next loop without sleeping first.

Additionally, in case router iterator is interrupted prematurely and iterator's discarded ratio is above the predefined limit, router's loop should pay the failing job penalty and sleep for additional time.

resolves PIPE-566

## Security


mihir20 (approved):
LGTM

lvrach (approved):

PR #4189, opened 2023-12-05
chore: improve fatal log during panic capture

lvrach:
# Description

Ensuring Fatal log during panic capture contains equivalent information to panic. This will allow us to start using alerts based on logs instead of bugsnag for getting notified.

## Linear Ticket

resolves PIPE-549 https://linear.app/rudderstack/issue/PIPE-594/improve-log-during-panic-capture

## Security


mihir20 (approved):
LGTM

achettyiitr (approved):

PR #4193, opened 2023-12-05
feat: introduce random sleep before clickhouse loads

Sidddddarth:
# Description

Customers facing issues due to continuous inserts on clickhouse. This introduces a `misc.SleepCtx` before every load anywhere between (0, maxDelay] with maxDelay defaulting to 5s.

## Linear Ticket

[Fix Clickhouse "too many parts" errors while inserting data](https://linear.app/rudderstack/issue/PIPE-529/fix-clickhouse-too-many-parts-errors-while-inserting-data)

## Security

## Summary by CodeRabbit

- **New Features**
  - Implemented a feature to introduce a random delay in data loading processes to optimize performance under varying system loads.

- **Refactor**
  - Enhanced the data loading mechanism with a new configuration option for managing operation timing.

### DIFF
--- a/warehouse/integrations/clickhouse/clickhouse.go
+++ b/warehouse/integrations/clickhouse/clickhouse.go
@@ -237,11 +237,11 @@ func New(conf *config.Config, log logger.Logger, stat stats.Stats) *Clickhouse {
 	ch.config.s3EngineEnabledWorkspaceIDs = conf.GetStringSlice("Warehouse.clickhouse.s3EngineEnabledWorkspaceIDs", nil)
 	ch.config.slowQueryThreshold = conf.GetDuration("Warehouse.clickhouse.slowQueryThreshold", 5, time.Minute)
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
-	ch.config.randomLoadDelay = func(destinationID string) time.Duration {
+	ch.config.randomLoadDelay = func(workspaceID string) time.Duration {
 		maxDelay := conf.GetDurationVar(
 			0,
 			time.Second,
-			fmt.Sprintf("Warehouse.clickhouse.%s.maxLoadDelay", destinationID),
+			fmt.Sprintf("Warehouse.clickhouse.%s.maxLoadDelay", workspaceID),
 			"Warehouse.clickhouse.maxLoadDelay",
 		)
 		return time.Duration(float64(maxDelay) * (1 - r.Float64()))
@@ -503,7 +503,7 @@ func (ch *Clickhouse) typecastDataFromType(data, dataType string) interface{} {
 
 // loadTable loads table to clickhouse from the load files
 func (ch *Clickhouse) loadTable(ctx context.Context, tableName string, tableSchemaInUpload model.TableSchema) (err error) {
-	if err = misc.SleepCtx(ctx, ch.config.randomLoadDelay(ch.Warehouse.Destination.ID)); err != nil {
+	if err = misc.SleepCtx(ctx, ch.config.randomLoadDelay(ch.Warehouse.WorkspaceID)); err != nil {
 		return
 	}
 	if ch.UseS3CopyEngineForLoading() {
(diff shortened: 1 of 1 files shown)


Sidddddarth:
@coderabbitai review

warehouse/integrations/clickhouse/clickhouse.go
```diff
@@ -491,6 +503,9 @@ func (ch *Clickhouse) typecastDataFromType(data, dataType string) interface{} {
 
 // loadTable loads table to clickhouse from the load files
 func (ch *Clickhouse) loadTable(ctx context.Context, tableName string, tableSchemaInUpload model.TableSchema) (err error) {
+	if misc.SleepCtx(ctx, ch.config.randomLoadDelay(ch.Warehouse.Destination.ID)) != nil {
```
fracasula, line 506: Wouldn't this apply to all customers?

Sidddddarth: should I put a 0 second default..? I left it at 5 because this is a problem that could affect any and everyone.

fracasula: Yes I think by default we shouldn't change the behaviour unless there is an actual problem.
It looks like they are the only one complaining while the other customers are happy with the delivery times, right? If that's the case let's go with a default of 0 and then add a specific setting for them.

fracasula: How many customers are having this issue? It's not clear to me the extent of the issue we're trying to solve here.

Sidddddarth: it's only one right now😅

fracasula: We could have it with a default of `0` to avoid changing the behaviour when not necessary. Then for them we can set it to an appropriate value.

warehouse/integrations/clickhouse/clickhouse.go
```diff
@@ -234,6 +236,16 @@ func New(conf *config.Config, log logger.Logger, stat stats.Stats) *Clickhouse {
 	ch.config.numWorkersDownloadLoadFiles = conf.GetInt("Warehouse.clickhouse.numWorkersDownloadLoadFiles", 8)
 	ch.config.s3EngineEnabledWorkspaceIDs = conf.GetStringSlice("Warehouse.clickhouse.s3EngineEnabledWorkspaceIDs", nil)
 	ch.config.slowQueryThreshold = conf.GetDuration("Warehouse.clickhouse.slowQueryThreshold", 5, time.Minute)
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	ch.config.randomLoadDelay = func(destinationID string) time.Duration {
```
achettyiitr, line 240: Should we have this at a workspace level rather than destination level, so that it can be easy to maintain for GA and to be consistent across all destinations for a customer? wdyt?

warehouse/integrations/clickhouse/clickhouse.go
```diff
@@ -491,6 +503,9 @@ func (ch *Clickhouse) typecastDataFromType(data, dataType string) interface{} {
 
 // loadTable loads table to clickhouse from the load files
 func (ch *Clickhouse) loadTable(ctx context.Context, tableName string, tableSchemaInUpload model.TableSchema) (err error) {
+	if err = misc.SleepCtx(ctx, ch.config.randomLoadDelay(ch.Warehouse.Destination.ID)); err != nil {
+		return
+	}
```
achettyiitr, line 508: [minor]: Should we do delay only in case some delay is present?
```suggestion
	if delay := ch.config.randomLoadDelay(ch.Warehouse.Destination.ID); delay > 0 {
		if err = misc.SleepCtx(ctx, delay); err != nil {
			return
		}
	}
```

Sidddddarth: wouldn't `time.After` take care of it anyway..?
Meaning if the delay is zero, there's no delay😅

warehouse/integrations/clickhouse/clickhouse.go
Sidddddarth: adopting this👍

achettyiitr (approved):

PR #4196, opened 2023-12-05
fix: source no pending jobs

achettyiitr:
# Description

## Linear Ticket

< Replace with Linear Link >

## Security

### DIFF
--- a/warehouse/source/source.go
+++ b/warehouse/source/source.go
@@ -143,6 +143,13 @@ func (m *Manager) process(ctx context.Context) error {
 	m.logger.Infow("starting source jobs processing")
 
 	for {
+		select {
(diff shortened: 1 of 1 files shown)


lvrach (approved):

warehouse/source/source.go
```diff
@@ -143,6 +143,13 @@ func (m *Manager) process(ctx context.Context) error {
 	m.logger.Infow("starting source jobs processing")
 
 	for {
+		select {
```
Sidddddarth, line 146: instead of moving the sleep above, we could have `processPendingJobs` return early(`nil`) if `len(pendingJobs) == 0`.
Just an alternative, either one's fine

PR #4209, opened 2023-12-07
chore: snowflake debug duplicate messages

achettyiitr:
# Description

- Add support for logging sample duplicate rows for Snowflake. Currently, the plan is to deploy only to mattermost for a few hours and get sample messages which will help in debugging for duplicates.
- We are storing the duplicate data from both the main table and the staging table in a temporary file. Since these duplicates are coming continuously, plan is to just get one of the sample files.

## Linear Ticket

- Resolves PIPE-266

## Security


warehouse/integrations/snowflake/snowflake.go
```diff
@@ -521,63 +509,133 @@ func (sf *Snowflake) joinColumnsWithFormatting(columns []string, format string)
...
-		WHERE
-		  	RECEIVED_AT > (SELECT DATEADD(day,-?,current_date())) AND
-		  	ID IN (
-				SELECT ID
-				FROM `+stagingTable+`
-		  	)
-        ORDER BY RECEIVED_AT ASC
-		LIMIT ?;`,
+	rows, err := db.QueryContext(ctx, `
+		WITH mainCTE AS (
```
fracasula, line 528: What does `CTE` stand for?

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -521,63 +509,133 @@ func (sf *Snowflake) joinColumnsWithFormatting(columns []string, format string)
...
-	ctx context.Context,
-	db *sqlmw.DB,
-	mainTableName,
-	stagingTableName string,
-) ([]duplicateMessage, error) {
+func (sf *Snowflake) trackDuplicateMessageInfo(
+	ctx context.Context, db *sqlmw.DB, mainTableName,
+	stagingTableName, sortedColumns string,
+) error {
 	if !lo.Contains(sf.config.debugDuplicateWorkspaceIDs, sf.Warehouse.WorkspaceID) {
```
fracasula, line 516: [question] Is `workspaceID` too broad? Perhaps `destinationID` would be better?

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -521,63 +509,133 @@ func (sf *Snowflake) joinColumnsWithFormatting(columns []string, format string)
...
+			  SELECT
+				ID
+			  FROM
+				`+stagingTable+`
+			)
+		  ORDER BY
+			RECEIVED_AT ASC
+		  LIMIT
+			?
+		), StagingCTE AS (
```
fracasula, line 552: [optional] Do we really need `StagingCTE`?

* In `mainCTE` we're getting all the records from the last X days that are also present in the staging table.
* In `stagingCTE` we're reading from the staging table instead but we select only the records that are already set as duplicate in `mainCTE`.

I understand that in the `UNION` at the end you want the identifier table but what for? If those records are duplicated they are in both tables, right?

Just saying, maybe just save the `mainCTE` and then add the staging file in the dump unless you find this way to be more convenient.

fracasula (approved):

warehouse/integrations/snowflake/snowflake.go
achettyiitr: They will have the same `messageID`, not other columns, that's what we will be figuring out from the sample.

warehouse/integrations/snowflake/snowflake.go
achettyiitr: Common Table Expression (CTE)

warehouse/integrations/snowflake/snowflake.go
achettyiitr: Initially, it was added only for `mattermost` use case for their workspace. We can probably extend it as well to destinationID.

warehouse/integrations/snowflake/snowflake.go
fracasula: Yeah that's why I said we could include the staging file in the dump, it's data we have already. I get it though, looks more convenient this way and the additional scanning is as big as the data from the staging file so it shouldn't be huge :+1:

PR #4213, opened 2023-12-07
chore: tune shared db connection pooling

atzoum:
# Description

Trying to minimise each rudder-server's footprint on the overall number of shared postgres's connections

## Security

### DIFF
--- a/services/rsources/rsources.go
+++ b/services/rsources/rsources.go
@@ -153,6 +153,7 @@ type JobServiceConfig struct {
 	LocalHostname               string
 	LocalConn                   string
 	MaxPoolSize                 int
+	MinPoolSize                 int
 	SharedConn                  string
 	SubscriptionTargetConn      string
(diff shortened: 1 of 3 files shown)


cisse21 (approved):

services/rsources/rsources.go
```diff
@@ -202,6 +203,9 @@ func NewJobService(config JobServiceConfig) (JobService, error) {
 	if config.MaxPoolSize <= 2 {
 		config.MaxPoolSize = 2 // minimum 2 connections in the pool for proper startup
 	}
+	if config.MinPoolSize <= 0 {
```
Sidddddarth, line 206: why not an upper limit for both of them too..?
```
maxConns := lo.Clamp(config.MaxPoolSize, 2, 3)
minConss := lo.Clamp(config.MinPoolSize, 1, 2)
```

Sidddddarth (approved):

PR #4215, opened 2023-12-07
fix: gateway stat tags when source is disabled

Sidddddarth:
# Description

Seeing empty `workspaceID` in gateway stats when source is disabled. This fixes that. Added a couple tests to verify that.

## Linear Ticket

[slack thread](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1701714466251759)

## Security

### DIFF
--- a/gateway/handle_http_auth_test.go
+++ b/gateway/handle_http_auth_test.go
@@ -11,7 +11,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/rudderlabs/rudder-go-kit/logger"
-	"github.com/rudderlabs/rudder-go-kit/stats"
+	"github.com/rudderlabs/rudder-go-kit/stats/memstats"
 	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 	gwtypes "github.com/rudderlabs/rudder-server/gateway/internal/types"
 )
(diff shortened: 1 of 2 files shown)


gateway/handle_http_auth_test.go
```diff
@@ -76,6 +77,21 @@ func TestAuth(t *testing.T) {
 			body, err := io.ReadAll(w.Body)
 			require.NoError(t, err, "reading response body should succeed")
 			require.Equal(t, "Failed to read writeKey from header\n", string(body))
+			require.Equal(t,
+				float64(1),
+				statsStore.Get(
+					"gateway.write_key_requests",
+					map[string]string{
```
achettyiitr, line 84: [optional]: Also at other below places as well.
```suggestion
					stats.Tags{
```

gateway/handle_http_auth_test.go
```diff
@@ -76,6 +77,21 @@ func TestAuth(t *testing.T) {
 			body, err := io.ReadAll(w.Body)
 			require.NoError(t, err, "reading response body should succeed")
 			require.Equal(t, "Failed to read writeKey from header\n", string(body))
+			require.Equal(t,
```
fracasula, line 80: Could probably use `require.EqualValues` if you want to avoid the cast below (i.e. `float64(1)`).

fracasula (approved):

PR #4224, opened 2023-12-11
chore: autofix corrupted job-status stats

atzoum:
# Description

Whenever corrupted job-status stats are encountered (in < out), gateway fixes them before responding and logs a warning.

## Linear Ticket

resolves PIPE-617

## Security

### DIFF
--- a/services/rsources/rsources.go
+++ b/services/rsources/rsources.go
@@ -34,13 +34,59 @@ type Stats struct {
 	Failed uint `json:"failed"`
 }
 
-func (r Stats) completed() bool {
+func (r *Stats) completed() bool {
 	return r.In == r.Out+r.Failed
+}
+
+func (r *Stats) corrupted() bool {
+	return r.In < r.Out+r.Failed
+}
+
+func (r *Stats) fixCorrupted() {
+	if r.corrupted() {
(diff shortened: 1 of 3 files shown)


services/rsources/rsources.go
```diff
@@ -34,15 +34,60 @@ type Stats struct {
...
 	return r.In == r.Out+r.Failed
 }
 
+func (r *Stats) corrupted() bool {
+	return r.In < r.Out+r.Failed
+}
+
+func (r *Stats) fixCorrupted() {
+	if r.corrupted() {
+		r.In = r.Out + r.Failed
```
mihir20, line 47: What is the reason behind this corruption? Why are we programatically fixing this corruption?

atzoum: This doesn't happen regularly, however there are rare occasions where bugs cause this kind of corruption ([example](https://rudderlabs.slack.com/archives/C02B0A38QQG/p1702224101518269)).

The idea here is to "fix" this corruption virtually for the consuming service to resume without being alerted and fire an alert on our side so that we can further investigate. The alert would be setup on the warning log

achettyiitr (approved):

PR #4225, opened 2023-12-11
chore: include the component in application_name of postgres connection

mihir20:
# Description

We want to manage and identify PostgreSQL connections based on the components that establish them. PostgreSQL provides the application_name parameter, which we can use to tag connections with the name of the component. This can be helpful for monitoring and managing connections.

Usecase:
1. To see how many connections are used by a specific component, we can query the `pg_stat_activity`
2. To identify slow queries by a specific component, we can use the `pg_stat_statements` extension along with `application_name`
3. kill all connections from a specific component
4. To monitor resource usage (e.g., CPU, memory) by a specific component, we can use the `pg_stat_activity` along with other system views.

## Linear Ticket

pipe-605

## Security

### DIFF
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -96,7 +96,7 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 
 	reporting := a.app.Features().Reporting.Setup(ctx, backendconfig.DefaultBackendConfig)
 	defer reporting.Stop()
-	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config, "embedded-app-db-syncer")})
+	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config, "reporting")})
 	g.Go(func() error {
 		syncer()
 		return nil
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -96,7 +96,7 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 
 	reporting := a.app.Features().Reporting.Setup(ctx, backendconfig.DefaultBackendConfig)
 	defer reporting.Stop()
-	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config)})
+	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config, "embedded-app-db-syncer")})
 	g.Go(func() error {
 		syncer()
 		return nil
--- a/app/apphandlers/processorAppHandler.go
+++ b/app/apphandlers/processorAppHandler.go
@@ -108,7 +108,7 @@ func (a *processorApp) StartRudderCore(ctx context.Context, options *app.Options
 
 	reporting := a.app.Features().Reporting.Setup(ctx, backendconfig.DefaultBackendConfig)
 	defer reporting.Stop()
-	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config.Default, "processor-app-db-syncer")})
+	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config.Default, "reporting")})
 	g.Go(misc.WithBugsnag(func() error {
 		syncer()
 		return nil
--- a/app/apphandlers/processorAppHandler.go
+++ b/app/apphandlers/processorAppHandler.go
@@ -108,7 +108,7 @@ func (a *processorApp) StartRudderCore(ctx context.Context, options *app.Options
 
 	reporting := a.app.Features().Reporting.Setup(ctx, backendconfig.DefaultBackendConfig)
 	defer reporting.Stop()
-	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config.Default)})
+	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config.Default, "processor-app-db-syncer")})
 	g.Go(misc.WithBugsnag(func() error {
 		syncer()
 		return nil
--- a/app/apphandlers/setup.go
+++ b/app/apphandlers/setup.go
@@ -61,7 +61,7 @@ func NewRsourcesService(deploymentType deployment.Type) (rsources.JobService, er
 	var rsourcesConfig rsources.JobServiceConfig
 	rsourcesConfig.MaxPoolSize = config.GetInt("Rsources.MaxPoolSize", 3)
 	rsourcesConfig.MinPoolSize = config.GetInt("Rsources.MinPoolSize", 1)
-	rsourcesConfig.LocalConn = misc.GetConnectionString(config.Default)
+	rsourcesConfig.LocalConn = misc.GetConnectionString(config.Default, "rsources-service")
 	rsourcesConfig.LocalHostname = config.GetString("DB.host", "localhost")
 	rsourcesConfig.SharedConn = config.GetString("SharedDB.dsn", "")
 	rsourcesConfig.SkipFailedRecordsCollection = !config.GetBool("Router.failedKeysEnabled", true)
--- a/app/apphandlers/setup.go
+++ b/app/apphandlers/setup.go
@@ -61,7 +61,7 @@ func NewRsourcesService(deploymentType deployment.Type) (rsources.JobService, er
 	var rsourcesConfig rsources.JobServiceConfig
 	rsourcesConfig.MaxPoolSize = config.GetInt("Rsources.MaxPoolSize", 3)
 	rsourcesConfig.MinPoolSize = config.GetInt("Rsources.MinPoolSize", 1)
-	rsourcesConfig.LocalConn = misc.GetConnectionString(config.Default, "rsources-service")
+	rsourcesConfig.LocalConn = misc.GetConnectionString(config.Default, "rsources")
 	rsourcesConfig.LocalHostname = config.GetString("DB.host", "localhost")
 	rsourcesConfig.SharedConn = config.GetString("SharedDB.dsn", "")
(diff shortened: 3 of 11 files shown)


services/validators/envValidator.go
```diff
@@ -124,7 +124,7 @@ func getWorkspaceFromDB(dbHandle *sql.DB) (string, error) {
 }
 
 func createDBConnection() (*sql.DB, error) {
-	psqlInfo := misc.GetConnectionString(config.Default)
+	psqlInfo := misc.GetConnectionString(config.Default, "env-validator")
```
atzoum, line 127: `envValidator` is killing dangling connections during startup based on the `APPLICATION_NAME`. Now that we are not using the exact same `APPLICATION_NAME` across rudder-server, we need to adapt the query used in `killDanglingDBConnections` accordingly.

app/apphandlers/embeddedAppHandler.go
```diff
@@ -96,7 +96,7 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 
 	reporting := a.app.Features().Reporting.Setup(ctx, backendconfig.DefaultBackendConfig)
 	defer reporting.Stop()
-	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config)})
+	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config, "embedded-app-db-syncer")})
```
atzoum, line 99:
```suggestion
	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config, "reporting")})
```

app/apphandlers/processorAppHandler.go
```diff
@@ -108,7 +108,7 @@ func (a *processorApp) StartRudderCore(ctx context.Context, options *app.Options
 
 	reporting := a.app.Features().Reporting.Setup(ctx, backendconfig.DefaultBackendConfig)
 	defer reporting.Stop()
-	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config.Default)})
+	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config.Default, "processor-app-db-syncer")})
```
atzoum, line 111:
```suggestion
	syncer := reporting.DatabaseSyncer(types.SyncerConfig{ConnInfo: misc.GetConnectionString(config.Default, "reporting")})
```

app/apphandlers/setup.go
```diff
@@ -61,7 +61,7 @@ func NewRsourcesService(deploymentType deployment.Type) (rsources.JobService, er
 	var rsourcesConfig rsources.JobServiceConfig
 	rsourcesConfig.MaxPoolSize = config.GetInt("Rsources.MaxPoolSize", 3)
 	rsourcesConfig.MinPoolSize = config.GetInt("Rsources.MinPoolSize", 1)
-	rsourcesConfig.LocalConn = misc.GetConnectionString(config.Default)
+	rsourcesConfig.LocalConn = misc.GetConnectionString(config.Default, "rsources-service")
```
atzoum, line 64:
```suggestion
	rsourcesConfig.LocalConn = misc.GetConnectionString(config.Default, "rsources")
```

backend-config/internal/cache/cache.go
```diff
@@ -132,7 +132,7 @@ func (db *cacheStore) Get(ctx context.Context) ([]byte, error) {
 
 // setupDBConn sets up the database connection, creates the config table if it doesn't exist
 func setupDBConn() (*sql.DB, error) {
-	psqlInfo := misc.GetConnectionString(config.Default)
+	psqlInfo := misc.GetConnectionString(config.Default, "backend-config-cache")
```
atzoum, line 135:
```suggestion
	psqlInfo := misc.GetConnectionString(config.Default, "backend-config")
```

services/validators/envValidator.go
atzoum, line 127:
```suggestion
	psqlInfo := misc.GetConnectionString(config.Default, "")
```

Sidddddarth:
Since we're passing context to every query(I think), how necessary is this..?
Maybe invest into passing context where it's not and make queries more conducive towards a graceful shutdown too.

Sidddddarth:
> > Since we're passing context to every query(I think), how necessary is this..? Maybe invest into passing context where it's not and make queries more conducive towards a graceful shutdown too.
> 
> @Sidddddarth there is no way to gracefully handle a panic in one goroutine while a SQL statement is being executed on another goroutine

touché

lvrach:
The motivation behind this change is to allow better observability, debugging and handling of connections. 

Since we use multiple connection pools per server instance, knowing which connection is established by which component is important.

The component will be concatenated and appear in application_name in pg_stat_activity.

Some use cases:
1. how many connections are used by a component
2. this query takes a lot of time, which component is related to
3. kill all connections from a single component

> Since we're passing context to every query(I think), how necessary is this..? 

How will context help in this case?

lvrach:
@mihir20 please provide a better description for the change. Including motivation behind it.

achettyiitr:
Can we add at here as well with a [notifier](https://github.com/rudderlabs/rudder-server/blob/98498eed972a79016c9e531fc29abb9fa654ba6f/services/notifier/notifier.go#L260) identifier?

utils/misc/dbutils.go
```diff
@@ -21,6 +21,9 @@ func GetConnectionString(c *config.Config) string {
 	// Application Name can be any string of less than NAMEDATALEN characters (64 characters in a standard PostgreSQL build).
 	// There is no need to truncate the string on our own though since PostgreSQL auto-truncates this identifier and issues a relevant notice if necessary.
 	appName := DefaultString("rudder-server").OnError(os.Hostname())
+	if len(componentName) > 0 {
+		appName = fmt.Sprintf("%s-%s", componentName, appName)
+	}
```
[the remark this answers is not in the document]
mihir20: Application Name can be any string of less than 64 characters, if we try to set application name with more than 64 chars PostgreSQL will truncate it to 64 chars, In such cases if we try to set app name as `app name + component name` componentName may get truncated.  If we set component name as prefix we may lose app name which imo is okay as we already know app name.

This way, even if the entire string is truncated to 64 characters, the most important information (the component name) is retained.

wdyt @achettyiitr ?

services/validators/envValidator.go
mihir20: updated query in killDanglingDBConnections 👍

mihir20: Didn't understand why should we not send component name here?

atzoum: This comment was mostly done in relation to [this](https://github.com/rudderlabs/rudder-server/pull/4225#discussion_r1422526443) one

utils/misc/dbutils_test.go
```diff
@@ -0,0 +1,59 @@
...
+			if (err != nil) != tt.wantErr {
+				t.Errorf("SetApplicationNameInDBConnectionURL() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if got != tt.want {
+				t.Errorf("SetApplicationNameInDBConnectionURL() got = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
```
achettyiitr: [optional]:  We can probably simplify this and use require for assertions.
```suggestion
func TestSetApplicationNameInDBConnection(t *testing.T) {
	tests := []struct {
		name    string
		dns     string
		appName string
		want    string
		wantErr bool
	}{
		{
			name:    "Invalid DNS URL",
			dns:     "abc@example.com:5432",
			appName: "rsources",
			want:    "",
			wantErr: true,
		},
		{
			name:    "Add App Name in DNS URL",
			dns:     "postgresql://rudder:password@prousmtusmt-rs-shared-postgresql:5432/jobsdb?sslmode=disable",
			appName: "rsources",
			want:    "postgresql://rudder:password@prousmtusmt-rs-shared-postgresql:5432/jobsdb?application_name=rsources&sslmode=disable",
			wantErr: false,
		},
		{
			name:    "Update App Name in DNS URL",
			dns:     "postgresql://rudder:password@prousmtusmt-rs-shared-postgresql:5432/jobsdb?application_name=random&sslmode=disable",
			appName: "rsources",
			want:    "postgresql://rudder:password@prousmtusmt-rs-shared-postgresql:5432/jobsdb?application_name=rsources&sslmode=disable",
			wantErr: false,
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got, err := misc.SetApplicationNameInDBConnectionURL(tt.dns, tt.appName)
			if tt.wantErr {
				require.Error(t, err, "Expected an error but got nil")
				return
			}
			require.Equal(t, tt.want, got)
		})
	}
}
```

mihir20:
> Can we pass the notifier in the fallback DSN as well in [here](https://github.com/rudderlabs/rudder-server/blob/fa5ceb6863e82d6f16c4f4192fc80e19a91d9894/warehouse/app.go#L172) for the notifier component? Currently, `a.connectionString()` will have `warehouse`.

Done 👍

achettyiitr (approved):

utils/misc/dbutils.go
```diff
@@ -21,11 +22,27 @@ func GetConnectionString(c *config.Config) string {
...
+func SetApplicationNameInDBConnectionURL(connectionUrl, appName string) (string, error) {
+	connUrl, err := url.Parse(connectionUrl)
+	if err != nil {
+		return "", err
+	}
+	queryParams := connUrl.Query()
+	queryParams.Set("application_name", appName)
+	connUrl.RawQuery = queryParams.Encode()
+	return connUrl.String(), nil
+}
```
achettyiitr, line 44: We can probably use a shorter function name here.

```suggestion
// SetAppNameInDBConnURL sets application name in db connection url
// if application name is already present in dns it will get override by the appName
func SetAppNameInDBConnURL(connectionURL, appName string) (string, error) {
	connURL, err := url.Parse(connectionURL)
	if err != nil {
		return "", fmt.Errorf("parsing connection url: %w", err)
	}

	queryParams := connURL.Query()
	queryParams.Set("application_name", appName)
	connURL.RawQuery = queryParams.Encode()
	return connURL.String(), nil
}
```

cisse21 (approved):

PR #4234, opened 2023-12-13
feat: add destinationID to rETL endpoint

mihir20:
# Description

Adding Destination ID to rETL endpoint to provide a way for rETL to send an event from a source to a destination. If there are multiple sources rETL will call `/retl` endpoint multiple times to deliver to multiple destinations. 

- Destination ID will be stored in parameters column along with source ID

## Linear Ticket

Pipe-619

## Security

### DIFF
--- a/gateway/handle.go
+++ b/gateway/handle.go
@@ -264,7 +264,7 @@ func (gw *Handle) getJobDataFromRequest(req *webRequestT) (jobData *jobFromReq, 
 	var (
 		arctx         = req.authContext
 		sourceID      = arctx.SourceID
-		destinationId = arctx.DestinationId
+		destinationID = arctx.DestinationID
 		// Should be function of body
 		workspaceId  = arctx.WorkspaceID
 		userIDHeader = req.userIDHeader
@@ -434,8 +434,8 @@ func (gw *Handle) getJobDataFromRequest(req *webRequestT) (jobData *jobFromReq, 
 		"source_task_run_id": sourcesTaskRunID,
 		"traceparent":        traceParent,
 	}
-	if len(destinationId) != 0 {
-		params["destination_id"] = destinationId
+	if len(destinationID) != 0 {
+		params["destination_id"] = destinationID
 	}
 	marshalledParams, err = json.Marshal(params)
 	if err != nil {
--- a/gateway/handle.go
+++ b/gateway/handle.go
@@ -262,8 +262,9 @@ func (gw *Handle) userWebRequestWorkerProcess(userWebRequestWorker *userWebReque
 // - user in the payload is not allowed to send events (suppressed)
 func (gw *Handle) getJobDataFromRequest(req *webRequestT) (jobData *jobFromReq, err error) {
 	var (
-		arctx    = req.authContext
-		sourceID = arctx.SourceID
+		arctx         = req.authContext
+		sourceID      = arctx.SourceID
+		destinationId = arctx.DestinationId
 		// Should be function of body
 		workspaceId  = arctx.WorkspaceID
 		userIDHeader = req.userIDHeader
@@ -432,6 +433,9 @@ func (gw *Handle) getJobDataFromRequest(req *webRequestT) (jobData *jobFromReq, 
 		"source_job_run_id":  sourcesJobRunID,
 		"source_task_run_id": sourcesTaskRunID,
 		"traceparent":        traceParent,
+	}
+	if len(destinationId) != 0 {
+		params["destination_id"] = destinationId
 	}
 	marshalledParams, err = json.Marshal(params)
 	if err != nil {
--- a/gateway/handle_http_auth.go
+++ b/gateway/handle_http_auth.go
@@ -119,6 +119,41 @@ func (gw *Handle) sourceIDAuth(delegate http.HandlerFunc) http.HandlerFunc {
 	}
 }
 
+// authenticateDestinationIdForSource middleware to authenticate destinationId in the X-Rudder-Destination-Id header.
+// If the destinationId is invalid, the request is rejected.
+// destinationID authentication should be performed only after source is authenticated and source is present in context
+// Following validations are performed
+//  1. Destination should be present for source config
+//  2. Destination should be enabled for the source
+func (gw *Handle) authenticateDestinationIdForSource(delegate http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		reqType := r.Context().Value(gwtypes.CtxParamCallType).(string)
+		arctx := r.Context().Value(gwtypes.CtxParamAuthRequestContext).(*gwtypes.AuthRequestContext)
+		var errorMessage string
+		defer func() {
+			gw.handleHttpError(w, r, errorMessage)
+			gw.handleFailureStats(errorMessage, reqType, arctx)
+		}()
+		destinationID := r.Header.Get("X-Rudder-Destination-Id")
+		if destinationID == "" {
+			// TODO: once rETL team migrates to sending destination id in header make destination id as mandatory
+			delegate.ServeHTTP(w, r)
+			return
+		}
+		isValidDestination, isDestinationEnabled := gw.validateDestinationID(destinationID, arctx)
+		if !isValidDestination {
+			errorMessage = response.InvalidDestinationID
+			return
+		}
+		if !isDestinationEnabled {
+			errorMessage = response.DestinationDisabled
+			return
+		}
+		arctx.DestinationId = destinationID
+		delegate.ServeHTTP(w, r.WithContext(context.WithValue(r.Context(), gwtypes.CtxParamAuthRequestContext, arctx)))
+	}
+}
+
 // replaySourceIDAuth middleware to authenticate sourceID in the X-Rudder-Source-Id header.
 // If the sourceID is valid, i.e. it is a replay source and enabled, the source auth info is added to the request context.
 // If the sourceID is invalid, the request is rejected.
@@ -149,6 +184,19 @@ func (gw *Handle) authRequestContextForSourceID(sourceID string) *gwtypes.AuthRe
 		return sourceToRequestContext(s)
 	}
 	return nil
+}
+
+// validateDestinationID gets request context for a given sourceID. If the sourceID is invalid, returns nil.
+func (gw *Handle) validateDestinationID(destinationID string, arctx *gwtypes.AuthRequestContext) (isValid, destinationDisabled bool) {
+	if len(arctx.Source.Destinations) == 0 {
+		return false, false
+	}
+	for _, destination := range arctx.Source.Destinations {
+		if destination.ID == destinationID {
+			return true, destination.Enabled
+		}
+	}
+	return false, false
 }
 
 // authRequestContextForWriteKey gets request context for a given writeKey. If the writeKey is invalid, returns nil.
@@ -226,6 +274,24 @@ func (gw *Handle) handleFailureStats(errorMessage, reqType string, arctx *gwtype
 				Source:   "noSourceIDInHeader",
 			}
 		case response.SourceDisabled:
+			stat = gwstats.SourceStat{
+				SourceID:    arctx.SourceID,
+				WriteKey:    arctx.WriteKey,
+				ReqType:     reqType,
+				Source:      arctx.SourceTag(),
+				WorkspaceID: arctx.WorkspaceID,
+				SourceType:  arctx.SourceCategory,
+			}
+		case response.DestinationDisabled:
+			stat = gwstats.SourceStat{
+				SourceID:    arctx.SourceID,
+				WriteKey:    arctx.WriteKey,
+				ReqType:     reqType,
+				Source:      arctx.SourceTag(),
+				WorkspaceID: arctx.WorkspaceID,
+				SourceType:  arctx.SourceCategory,
+			}
+		case response.InvalidDestinationID:
 			stat = gwstats.SourceStat{
 				SourceID:    arctx.SourceID,
 				WriteKey:    arctx.WriteKey,
--- a/gateway/handle_http_auth.go
+++ b/gateway/handle_http_auth.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"net/http"
 
+	gwCtx "github.com/rudderlabs/rudder-server/gateway/internal/context"
+
 	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 	gwstats "github.com/rudderlabs/rudder-server/gateway/internal/stats"
 	gwtypes "github.com/rudderlabs/rudder-server/gateway/internal/types"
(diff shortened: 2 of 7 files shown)


gateway/handle.go
```diff
@@ -261,8 +261,9 @@ func (gw *Handle) userWebRequestWorkerProcess(userWebRequestWorker *userWebReque
 // - user in the payload is not allowed to send events (suppressed)
 func (gw *Handle) getJobDataFromRequest(req *webRequestT) (jobData *jobFromReq, err error) {
 	var (
-		arctx    = req.authContext
-		sourceID = arctx.SourceID
+		arctx         = req.authContext
+		sourceID      = arctx.SourceID
+		destinationId = arctx.DestinationId
```
achettyiitr, line 266: it should be destinationID to follow Go's idiomatic capitalization
```suggestion
		destinationID = arctx.DestinationID
```

gateway/response/response.go
```diff
@@ -66,6 +66,12 @@ const (
 	InvalidSourceID = "Invalid source id"
 	// InvalidReplaySource - Invalid replay source
 	InvalidReplaySource = "Invalid replay source"
+	// InvalidDestinationID - Invalid destination id
+	InvalidDestinationID = "Invalid destination id"
+	// DestinationDisabled - Destination is disabled
+	DestinationDisabled = "Destination is disabled"
+	// NoDestinationIdInHeader - Failed to read destination id from header
+	NoDestinationIdInHeader = "Failed to read destination id from header"
```
achettyiitr, line 74: Go's idiomatic capitalization
```suggestion
	NoDestinationIDInHeader = "Failed to read destination id from header"
```

gateway/response/response.go
```diff
@@ -87,6 +93,9 @@ var statusMap = map[string]status{
 	NoSourceIdInHeader:      {message: NoSourceIdInHeader, code: http.StatusUnauthorized},
 	InvalidSourceID:         {message: InvalidSourceID, code: http.StatusUnauthorized},
 	InvalidReplaySource:     {message: InvalidReplaySource, code: http.StatusUnauthorized},
+	DestinationDisabled:     {message: DestinationDisabled, code: http.StatusNotFound},
+	InvalidDestinationID:    {message: InvalidDestinationID, code: http.StatusUnauthorized},
+	NoDestinationIdInHeader: {message: NoDestinationIdInHeader, code: http.StatusUnauthorized},
```
achettyiitr, line 98:
```suggestion
	NoDestinationIDInHeader: {message: NoDestinationIDInHeader, code: http.StatusUnauthorized},
```

achettyiitr, line 98: Shouldn't this be `http.StatusBadRequest`?

achettyiitr (changes requested):

gateway/handle_http_retl.go
```diff
@@ -4,5 +4,5 @@ import "net/http"
 
 // webRetlHandler - handler for retl requests
 func (gw *Handle) webRetlHandler() http.HandlerFunc {
-	return gw.callType("retl", gw.sourceIDAuth(gw.webHandler()))
+	return gw.callType("retl", gw.sourceIDAuth(gw.authenticateDestinationIdForSource(gw.webHandler())))
```
achettyiitr, line 7:
```suggestion
	return gw.callType("retl", gw.sourceIDAuth(gw.authenticateDestinationIDForSource(gw.webHandler())))
```

gateway/internal/types/types.go
```diff
@@ -27,6 +27,8 @@ type AuthRequestContext struct {
 	SourceJobRunID  string
 	SourceTaskRunID string
 	Source          backendconfig.SourceT
+	// DestinationId is optional param, destination id will be present for rETL Request
+	DestinationId string
```
achettyiitr, line 31:
```suggestion
	DestinationID string
```

gateway/response/response.go
mihir20: I was also wondering the same, but in case of `NoSourceId` we are returning `Unauthorised` status code tried to follow the same convention. @atzoum can you confirm the reason for using `Unauthorised` in case of no source id in header and what status should we use in this case?

gateway/handle_http_auth.go
```diff
@@ -119,6 +119,44 @@ func (gw *Handle) sourceIDAuth(delegate http.HandlerFunc) http.HandlerFunc {
...
+		}
+		isValidDestination, isDestinationEnabled := gw.validateDestinationID(destinationID, arctx)
+		if !isValidDestination {
+			errorMessage = response.InvalidDestinationID
+			return
+		}
+		if !isDestinationEnabled {
+			errorMessage = response.DestinationDisabled
+			return
+		}
```
achettyiitr, line 154: We can probably simplify this. We probably don't need validateDestinationID.

```suggestion
		destination, found := lo.Find(arctx.Source.Destinations, func(destination backendconfig.DestinationT) bool {
			return destination.ID == destinationID
		})
		if !found {
			errorMessage = response.InvalidDestinationID
			return
		}
		if !destination.Enabled {
			errorMessage = response.DestinationDisabled
			return
		}
```

gateway/handle_http_auth.go
```diff
@@ -119,6 +119,44 @@ func (gw *Handle) sourceIDAuth(delegate http.HandlerFunc) http.HandlerFunc {
...
+			if !gw.config.GetBool("Gateway.isDestinationIdMandatoryInRequestHeader", false) {
+				delegate.ServeHTTP(w, r)
+				return
+			}
+			errorMessage = response.NoDestinationIdInHeader
+			return
+		}
+		isValidDestination, isDestinationEnabled := gw.validateDestinationID(destinationID, arctx)
+		if !isValidDestination {
+			errorMessage = response.InvalidDestinationID
```
achettyiitr, line 148: Shouldn't this be DestinationNotFound instead of InvalidDestinationID?
```suggestion
			errorMessage = response.DestinationNotFound
```

gateway/handle_http_auth.go
```diff
@@ -119,6 +119,44 @@ func (gw *Handle) sourceIDAuth(delegate http.HandlerFunc) http.HandlerFunc {
...
+		reqType := r.Context().Value(gwtypes.CtxParamCallType).(string)
+		arctx := r.Context().Value(gwtypes.CtxParamAuthRequestContext).(*gwtypes.AuthRequestContext)
+		var errorMessage string
+		defer func() {
+			gw.handleHttpError(w, r, errorMessage)
+			gw.handleFailureStats(errorMessage, reqType, arctx)
+		}()
+		destinationID := r.Header.Get("X-Rudder-Destination-Id")
+		if destinationID == "" {
+			if !gw.config.GetBool("Gateway.isDestinationIdMandatoryInRequestHeader", false) {
```
achettyiitr, line 139:
```suggestion
			if !gw.config.GetBool("Gateway.isDestinationIDMandatoryInRequestHeader", false) {
```

gateway/handle_http_auth.go
```diff
@@ -234,6 +285,33 @@ func (gw *Handle) handleFailureStats(errorMessage, reqType string, arctx *gwtype
...
+			}
+		case response.NoDestinationIdInHeader:
+			stat = gwstats.SourceStat{
+				SourceID:    arctx.SourceID,
+				WriteKey:    arctx.WriteKey,
+				ReqType:     reqType,
+				Source:      arctx.SourceTag(),
+				WorkspaceID: arctx.WorkspaceID,
+				SourceType:  arctx.SourceCategory,
+			}
```
achettyiitr, line 314: Seems like duplicate code?
```suggestion
		case response.DestinationDisabled, response.InvalidDestinationID, response.DestinationNotFound
			stat = gwstats.SourceStat{
				SourceID:    arctx.SourceID,
				WriteKey:    arctx.WriteKey,
				ReqType:     reqType,
				Source:      arctx.SourceTag(),
				WorkspaceID: arctx.WorkspaceID,
				SourceType:  arctx.SourceCategory,
			}
```

gateway/handle_http_auth_test.go
```diff
@@ -49,6 +51,19 @@ func TestAuth(t *testing.T) {
 		return r
 	}
 
+	newRequestWithSourceIdAndDestId := func(sourceId, destinationId string, reqCtx *gwtypes.AuthRequestContext) *http.Request {
```
achettyiitr, line 54:
```suggestion
	newRequestWithSourceIDAndDestID := func(sourceID, destinationID string, reqCtx *gwtypes.AuthRequestContext) *http.Request {
```

gateway/handle_http_auth.go
achettyiitr, line 139: Does `allowReqsWithoutDestIDInHeader` seem like a more meaningful configuration key?

gateway/handle_http_auth.go
```diff
@@ -119,6 +119,44 @@ func (gw *Handle) sourceIDAuth(delegate http.HandlerFunc) http.HandlerFunc {
...
+// authenticateDestinationIdForSource middleware to authenticate destinationId in the X-Rudder-Destination-Id header.
+// If the destinationId is invalid, the request is rejected.
+// destinationID authentication should be performed only after source is authenticated and source is present in context
+// Following validations are performed
+//  1. Destination should be present for source config
+//  2. Destination should be enabled for the source
+func (gw *Handle) authenticateDestinationIdForSource(delegate http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		reqType := r.Context().Value(gwtypes.CtxParamCallType).(string)
+		arctx := r.Context().Value(gwtypes.CtxParamAuthRequestContext).(*gwtypes.AuthRequestContext)
```
fracasula, line 131: Can we please avoid the casting repetition we're doing everytime we get/set a context value for these types?

It's risky since it could lead to a panic (in case it is set as a type and read as another). There are [safer ways](https://github.com/rudderlabs/rudder-server/blob/v1.18.2/warehouse/utils/context.go) to do this, I suggest doing the same here.

gateway/handle_http_auth.go
mihir20: IMO its better to separate out core logic in small functions, if in future we want to modify any logic we will only need to modify small functions.

gateway/handle_http_auth.go
mihir20: can destination not found can be considered as a case of Invalid destination, similar to what we are doing for Source ID not found?

gateway/response/response.go
atzoum: It depends on perspective :)
- Either you consider sourceID/destinationID authentication information and you respond with `HTTP 401` in their absence.
- Or you consider sourceID/destinationID required request information and you respond with `HTTP 400` in their absence.

Personally I'm open to both, no strong opinion. `HTTP 400` might be a better approach.

gateway/handle_http_auth.go
```diff
@@ -119,6 +121,45 @@ func (gw *Handle) sourceIDAuth(delegate http.HandlerFunc) http.HandlerFunc {
 	}
 }
 
+// authenticateDestinationIDForSource middleware to authenticate destinationId in the X-Rudder-Destination-Id header.
+// If the destinationId is invalid, the request is rejected.
+// destinationID authentication should be performed only after source is authenticated and source is present in context
+// Following validations are performed
+//  1. Destination should be present for source config
+//  2. Destination should be enabled for the source
+func (gw *Handle) authenticateDestinationIDForSource(delegate http.HandlerFunc) http.HandlerFunc {
```
atzoum, line 130: please consider picking up a shorter name :)
```suggestion
func (gw *Handle) sourceDestIDAuth(delegate http.HandlerFunc) http.HandlerFunc {
```
Additionally, since the `sourceIDAuth` is a prerequisite for this middleware, maybe it should embed it here, see `replaySourceIDAuth` for a similar example

gateway/internal/context/context.go
```diff
@@ -0,0 +1,19 @@
...
+func GetRequestTypeFromCtx(ctx context.Context) (string, bool) {
+	reqType, ok := ctx.Value(gwtypes.CtxParamCallType).(string)
+	return reqType, ok
+}
+
+// GetAuthRequestFromCtx : get auth request from context
+func GetAuthRequestFromCtx(ctx context.Context) (*gwtypes.AuthRequestContext, bool) {
+	authReqCtx, ok := ctx.Value(gwtypes.CtxParamAuthRequestContext).(*gwtypes.AuthRequestContext)
+	return authReqCtx, ok
+}
```
atzoum, line 19: Shouldn't these be used everywhere now that they exist? What's the point of returning the second value (bool) if nobody cares about it in practice? Would it be more sincere to rename these functions to `MustGetXxx` and panic if casting fails?

gateway/internal/types/types.go
```diff
@@ -27,6 +27,8 @@ type AuthRequestContext struct {
 	SourceJobRunID  string
 	SourceTaskRunID string
 	Source          backendconfig.SourceT
+	// DestinationID is optional param, destination id will be present for rETL Request
```
atzoum, line 30: Not sure if we need this comment. `SourceJobRunID` & `SourceTaskRunID` are both optional as well & we don't mention when they are going to be present. One reason why we would choose not to include such comments would be change, i.e. at some point we will decide to change (when these will be present) and almost certainly we'll forget to update the relevant comments! So we end up with invalid comments!

gateway/internal/context/context.go
mihir20: added handling for bool, we will fail request with ISE if we are not able to get these from context, As we don't want our application to panic in such cases.

gateway/internal/types/types.go
mihir20: IMO its better to keep this comment, and if we start using this field for cases we should update the comment, this way we can keep more readable and intuitive. Or we can just add a comment that this field is optional. That way reading through the code might give us a better understanding that these fields may or many not necessarily 
always be available.

achettyiitr (approved):

gateway/handle_http_auth_test.go
```diff
@@ -391,4 +406,118 @@ func TestAuth(t *testing.T) {
 			require.Equal(t, "Invalid replay source\n", string(body))
 		})
 	})
+
+	t.Run("authDestIDForSource", func(t *testing.T) {
```
achettyiitr, line 410: @mihir20 You might need to update these tests since some response codes got changed.

gateway/handle_http_auth.go
```diff
@@ -119,6 +123,57 @@ func (gw *Handle) sourceIDAuth(delegate http.HandlerFunc) http.HandlerFunc {
...
+func (gw *Handle) authDestIDForSource(delegate http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var errorMessage, reqType string
+		var arctx *gwtypes.AuthRequestContext
+		defer func() {
+			gw.handleHttpError(w, r, errorMessage)
+			gw.handleFailureStats(errorMessage, reqType, arctx)
+		}()
+		reqType, ok := gwCtx.GetRequestTypeFromCtx(r.Context())
+		if !ok {
```
atzoum, line 141: Now we need test scenarios for the `!ok` conditions :)

mihir20: Adding tests should be fine this way we will be able to prevent panics in the server code. @fracasula rightly pointed out that this will be a safer way to extract fields from context. 
https://github.com/rudderlabs/rudder-server/pull/4234#discussion_r1430111046

gateway/handle_http_auth.go
```diff
@@ -119,6 +123,57 @@ func (gw *Handle) sourceIDAuth(delegate http.HandlerFunc) http.HandlerFunc {
 	}
 }
 
+// authDestIDForSource middleware to authenticate destinationId in the X-Rudder-Destination-Id header.
+// If the destinationId is invalid, the request is rejected.
```
cisse21, line 127: Do we really need to validate destinationId at the gateway level? The requests will anyways be filtered out in processor and since it is only an internal endpoint don't think sources would send an invalid destinationId. We could just accept whatever we get in the gateway level and move on

lvrach (approved):

gateway/handle_http_auth.go
mihir20: IMO we should not trust what's coming in requests, we should always validate critical inputs.

PR #4241, opened 2023-12-16
chore: support concurrent addition of failed keys for the same job target key

atzoum:
# Description

Due to how `INSERT... ON CONFLICT DO NOTHING` behaves while using a `Read Committed` isolation level ([link](https://www.postgresql.org/docs/11/transaction-iso.html#XACT-READ-COMMITTED:~:text=INSERT%20with%20an%20ON%20CONFLICT%20DO%20NOTHING%20clause%20may%20have%20insertion%20not%20proceed%20for%20a%20row%20due%20to%20the%20outcome%20of%20another%20transaction%20whose%20effects%20are%20not%20visible%20to%20the%20INSERT%20snapshot.%20Again%2C%20this%20is%20only%20the%20case%20in%20Read%20Committed%20mode.)) the previous statement would return NULL whenever two or more goroutines were trying to insert an entry in `rsources_failed_keys_v2` table for the same job target key.

We are now will be using `INSERT... ON CONFLICT DO UPDATE` which makes sure that:
- Changes from other uncommitted transactions will eventually become visible and taken into account.
- The row will be locked, so that it cannot be deleted by another transaction, until this one completes.

Introducing a new `ts` column, so that we don't need to update columns which are part of an index or the primary key.

## Linear Ticket

< Replace with Linear Link >

## Security


fracasula (approved):

services/rsources/handler.go
```diff
@@ -111,22 +111,16 @@ func (sh *sourcesHandler) AddFailedRecords(ctx context.Context, tx *sql.Tx, jobR
...
-
-	row := tx.QueryRow(`WITH new_key AS (
-		INSERT INTO rsources_failed_keys_v2 (id, job_run_id, task_run_id, source_id, destination_id)
+	row := tx.QueryRow(`INSERT INTO rsources_failed_keys_v2 (id, job_run_id, task_run_id, source_id, destination_id)
 		VALUES ($1, $2, $3, $4, $5) 
-		ON CONFLICT (job_run_id, task_run_id, source_id, destination_id, db_name) DO NOTHING
-		RETURNING id
-	) SELECT COALESCE(
-		(SELECT id FROM new_key),
-		(SELECT id FROM rsources_failed_keys_v2 WHERE job_run_id = $2 AND task_run_id = $3 AND source_id = $4 AND destination_id = $5)
```
achettyiitr, line 122: > (SELECT id FROM rsources_failed_keys_v2 WHERE job_run_id = $2 AND task_run_id = $3 AND source_id = $4 AND destination_id = $5)

Curious, about why this got removed?

achettyiitr (approved):

services/rsources/handler.go
atzoum: it is not needed anymore, since this will always return the appropriate id
```sql
INSERT INTO rsources_failed_keys_v2 
  (id, job_run_id, task_run_id, source_id, destination_id)
  VALUES 
  ($1, $2, $3, $4, $5) 
  ON CONFLICT (job_run_id, task_run_id, source_id, destination_id, db_name) 
  DO UPDATE SET ts = NOW()
  RETURNING id
```

PR #4244, opened 2023-12-18
fix: changing eventNames longer than configured max length to ":max-length-exceeded:" before sending to reporting

vamsikrishnakandi:
# Description
To reduce events with junk eventNames, adding a restriction on length of eventName. Added a configuration, if a valid max length is given, it replaces eventName (if it is longer than max length) to :max-length-exceeded:. How to show this in UI will be handled in web-app. 

Default for the config is set to zero, which will not change anything.

More details in linear ticket and attached notion doc in it.

Completes OBS-126

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-126/validate-eventname-in-reports

## Security


enterprise/reporting/reporting.go
```diff
@@ -589,6 +589,10 @@ func (r *DefaultReporter) Report(ctx context.Context, metrics []*types.PUReporte
 			metric = transformMetricForPII(metric, getPIIColumnsToExclude())
 		}
 
+		if len(metric.StatusDetail.EventName) > 100 {
+			metric.StatusDetail.EventName = "INVALID_EVENT_NAME"
```
achettyiitr, line 593: Is truncation a better option instead of sending `INVALID_EVENT_NAME`?

vamsikrishnakandi: we looked at the data, most of them are stringified JSONs, URLs, and decided to go with this. Attached analysis doc in linear as a comment

satishrudderstack: @achettyiitr One reason to not truncate is cardinality. Since these are mostly junk event names, if we truncate, these are going increase cardinality of event names on reporting

enterprise/reporting/reporting.go
```diff
@@ -589,6 +589,10 @@ func (r *DefaultReporter) Report(ctx context.Context, metrics []*types.PUReporte
 			metric = transformMetricForPII(metric, getPIIColumnsToExclude())
 		}
 
+		if len(metric.StatusDetail.EventName) > 100 {
```
achettyiitr, line 592: Should the maxLimit be configurable?

enterprise/reporting/reporting.go
```diff
@@ -589,6 +589,10 @@ func (r *DefaultReporter) Report(ctx context.Context, metrics []*types.PUReporte
 			metric = transformMetricForPII(metric, getPIIColumnsToExclude())
 		}
 
+		if len(metric.StatusDetail.EventName) > config.GetInt("Reporting.eventNameMaxLength", 100) {
```
atzoum, line 592: why don't we trim down event names instead to the first 100 characters?

vamsikrishnakandi: We are not sending first 100 characters, we are sending `INVALID_EVENT_NAME`

vamsikrishnakandi: https://github.com/rudderlabs/rudder-server/pull/4244#discussion_r1429901263

atzoum: Cardinality can be a serious issue even with event names having less than 100 characters though, e.g. a uuid is 36 characters and ksuid is 28 characters.
Why are we coupling it with the length of characters?

vamsikrishnakandi: Initial thought is to trim them but when we looked at the data, 99.55% of eventNames longer than 100 characters are stringified JSONs or URLs. And rest of them can also be avoided with good instrumentation. So, we decided to go with this as this will also help with cardinality.

achettyiitr (approved):

PR #4247, opened 2023-12-19
feat: filter destinations for events from rETL in processor

mihir20:
# Description
For events received through rETL endpoints, destination ID for an event will be stored in params field in JobsDB. For all the events having destination ID in params should be delivered to only those destinations irrespective of how many other destinations are connected to the source. All the other destinations will be filtered at processor.

Base PR: https://github.com/rudderlabs/rudder-server/pull/4234

## Linear Ticket

pipe-620

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1734,7 +1734,7 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 			// event will be dropped if no valid destination is present
 			// if empty string destinationID is all the destinations for the source is validated
 			// else only passed destinationID will be validated
-			if !proc.isDestinationAvailable(singularEvent, sourceId, destinationID) {
+			if !proc.isDestinationAvailable(singularEvent, sourceID, destinationID) {
 				continue
 			}
 
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1544,6 +1544,8 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 
 	outCountMap := make(map[string]int64) // destinations enabled
 	destFilterStatusDetailMap := make(map[string]map[string]*types.StatusDetail)
+	// map of messageID to destinationID: for messages that needs to be delivered to a specific destinations only
+	msgToDestMap := make(map[string]string)
 
 	spans := make([]stats.TraceSpan, 0, len(jobList))
 	defer func() {
@@ -1551,6 +1553,7 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 			span.End()
 		}
 	}()
+
 	for _, batchEvent := range jobList {
 		var eventParams types.EventParams
 		err := jsonfast.Unmarshal(batchEvent.Parameters, &eventParams)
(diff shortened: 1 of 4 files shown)


processor/processor.go
```diff
@@ -2994,7 +3016,7 @@ func (proc *Handle) isDestinationAvailable(event types.SingularEventT, sourceId
 		return false
 	}
 
-	if enabledDestinationsList := proc.getConsentFilteredDestinations(
+	enabledDestinationsList := proc.getConsentFilteredDestinations(
```
atzoum, line 3019: we can use a single if clause and just add another filtering stage
```go
	if enabledDestinationsList := lo.Filter(
		proc.getConsentFilteredDestinations(
			event,
			lo.Flatten(
				lo.Map(
					enabledDestTypes,
					func(destType string, _ int) []backendconfig.DestinationT {
						return proc.getEnabledDestinations(sourceId, destType)
					},
				),
			),
		), func(dest backendconfig.DestinationT, _ int) bool {
			return destinationID == "" || dest.ID == destinationID
		}); len(enabledDestinationsList) == 0 {
		proc.logger.Debug("No destination to route this event to")
		return false
	}
```

processor/processor.go
```diff
@@ -1697,6 +1705,11 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 				&shallowEventCopy,
 				&backendconfig.DestinationT{},
 			)
+			if len(destinationID) != 0 {
+				enhanceWithDestination(&shallowEventCopy, &backendconfig.DestinationT{
+					ID: destinationID,
+				})
+			}
```
atzoum, line 1712: I fear this won't be reliable under all circumstances, at least if there are mixed events for the same source in the same batch, i.e. some events having a destination_id and some others not having anything or having another destination_id: Both `TransformerEvent.Metadata` & `TransformerEvent.Destination` are being overwritten during `Handle#validateEvents` based on the value found in the first event of the batch! (see `proc#getTransformerEvents`)

We could fill `TransformerEvent.Metadata.DestinationID` and then read it from `TransformerResponse.Metadata.DestinationID`, but having to rely on an external service for making sure that this field is being transferred properly is not the safest approach. Furthermore, `proc#getTransformerEvents` is a generic method used by other steps as well & we wouldn't want to break other steps :/

We might need to find an alternative strategy

mihir20: Ack have thought of another strategy will update the PR

lvrach (approved):

processor/processor.go
```diff
@@ -1724,8 +1731,10 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 			// REPORTING - GATEWAY metrics - END
 
 			// Getting all the destinations which are enabled for this
-			// event
-			if !proc.isDestinationAvailable(singularEvent, sourceID) {
+			// event will be dropped if no valid destination is present
+			// if empty string destinationID is all the destinations for the source is validated
```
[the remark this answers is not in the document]
mihir20: updated now

Sidddddarth (approved):

PR #4248, opened 2023-12-19
feat: tracing support for gateway, processor and router

achettyiitr:
# Description

- Introducing tracing in gateway, processor, and router.
- The scope is just to trace a request from start to end to measure the performance of the system. We don't need a lot of granularity at this time.

### Changes

- The span is initiated in the `webRequestHandler` and then propagated to `JobsDB` via the `parameters` JSON field.

## Linear Ticket

- Resolves PIPE-571
- Resolves PIPE-602

## Security


integration_test/tracing/tracing_test.go
fracasula, line 1: Can we try to remove some of the duplicated code that we have here? It looks like an easy thing to do and it would help with reading and maintaining these tests.

integration_test/tracing/tracing_test.go
```diff
@@ -0,0 +1,871 @@
...
+		trServer := transformertest.NewBuilder().Build()
+		defer trServer.Close()
+
+		pool, err := dockertest.NewPool("")
+		require.NoError(t, err)
+
+		postgresContainer, err := resource.SetupPostgres(pool, t)
+		require.NoError(t, err)
+		zipkin, err := resource.SetupZipkin(pool, t)
+		require.NoError(t, err)
```
fracasula, line 78: Out of all the setup operations this is the most likely to fail so it might make sense to move it to the top.

integration_test/tracing/tracing_test.go
```diff
@@ -0,0 +1,871 @@
...
+		resp, err := (&http.Client{}).Do(req)
+		if err != nil {
+			return err
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			b, _ := io.ReadAll(resp.Body)
+			return fmt.Errorf("failed to send event to rudder server, status code: %d: %s", resp.StatusCode, string(b))
+		}
+		func() { kithttputil.CloseResponse(resp) }()
```
fracasula, line 868: Do we need to wrap it this way? :thinking:

integration_test/tracing/tracing_test.go
```diff
@@ -0,0 +1,871 @@
...
+
+	require.Eventually(t, func() bool {
+		var jobsCount int
+		require.NoError(t, db.QueryRow(fmt.Sprintf(`SELECT count(*) FROM unionjobsdbmetadata('%s',1) WHERE job_state = '%s' AND parameters->>'traceparent' is not NULL;`, queue, state)).Scan(&jobsCount))
+		t.Logf("%s %sJobCount: %d", queue, state, jobsCount)
+		return jobsCount == expectedCount
+	},
+		20*time.Minute,
+		1*time.Second,
+		fmt.Sprintf("%d %s events should be in %s state", expectedCount, queue, state),
```
fracasula, line 820: [optional] could use `Eventuallyf` instead.

processor/processor.go
```diff
@@ -1559,7 +1593,7 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 		}
 		statusList = append(statusList, &newStatus)
 
-		source, err := proc.getSourceBySourceID(sourceId)
+		source, err := proc.getSourceBySourceID(sourceID)
 		if err != nil {
 			continue
```
fracasula, line 1598: This looks like another case where we could set the span status to error, right?

processor/processor.go
```diff
@@ -1959,6 +1993,36 @@ func (proc *Handle) transformations(partition string, in *transformationMessage)
...
+				proc.logger.Warnf("Missing traceParent in transformations for jobId %d", event.Metadata.JobID)
+				continue
+			}
+			if _, ok := traces[event.Metadata.TraceParent]; ok {
+				continue
+			}
+			tags := stats.Tags{
+				"sourceId":      event.Metadata.SourceID,
+				"destinationId": event.Metadata.DestinationID,
+				"workspaceId":   event.Metadata.WorkspaceID,
```
fracasula, line 2016: Given that above here in the processor we also included the jobid and the uuid, it might make sense to include them here as well. Wdyt?

processor/processor.go
```diff
@@ -2077,6 +2143,18 @@ func (proc *Handle) sendQueryRetryStats(attempt int) {
...
 func (proc *Handle) Store(partition string, in *storeMessage) {
+	spans := make([]stats.TraceSpan, 0, len(in.traces))
+	defer func() {
+		for _, span := range spans {
+			span.End()
+		}
+	}()
+	for traceParent, tags := range in.traces {
+		ctx := stats.InjectTraceParentIntoContext(context.TODO(), traceParent)
+		_, span := proc.tracer.Start(ctx, "proc.store", stats.SpanKindConsumer, stats.SpanWithTags(tags))
```
fracasula, line 2154: Maybe the `SpanKindInternal` makes more sense here.

```go
	// SpanKindInternal is a SpanKind for a Span that represents an internal
	// operation within an application.
	SpanKindInternal = trace.SpanKindInternal
```

processor/processor.go
fracasula: We read from the DB already so it's definitely not a consumer at this stage. We're going to store the message though so perhaps it could either be internal or producer.

processor/worker_test.go
```diff
@@ -159,6 +159,10 @@ type mockWorkerHandle struct {
 	limitsReached bool
 }
 
+func (m *mockWorkerHandle) tracer() stats.Tracer {
+	return stats.NOP.NewTracer("my-tracer")
```
fracasula, line 163:
```suggestion
	return stats.NOP.NewTracer("")
```

router/worker.go
```diff
@@ -258,6 +259,31 @@ func (w *worker) transform(routerJobs []types.RouterJobT) []types.DestinationJob
...
+			span.End()
+		}
+	}()
+
+	for _, job := range routerJobs {
+		traceParent := job.JobMetadata.TraceParent
+		if traceParent != "" {
+			if _, ok := traces[traceParent]; !ok {
+				ctx := stats.InjectTraceParentIntoContext(context.Background(), traceParent)
+				_, span := w.rt.tracer.Start(ctx, "rt.transform", stats.SpanKindConsumer, stats.SpanWithTags(stats.Tags{
```
fracasula, line 274: This one also doesn't sound like a consumer span :thinking:

router/worker.go
```diff
@@ -308,6 +359,33 @@ func (w *worker) processDestinationJobs() {
...
+		}
+	}()
+
+	for _, job := range w.destinationJobs {
+		for _, jobMetadata := range job.JobMetadataArray {
+			traceParent := jobMetadata.TraceParent
+			if traceParent != "" {
+				if _, ok := traces[traceParent]; !ok {
+					ctx := stats.InjectTraceParentIntoContext(context.Background(), traceParent)
+					_, span := w.rt.tracer.Start(ctx, "rt.process", stats.SpanKindConsumer, stats.SpanWithTags(stats.Tags{
```
fracasula, line 375: Same, not a consumer?

router/worker.go
```diff
@@ -258,6 +259,31 @@ func (w *worker) transform(routerJobs []types.RouterJobT) []types.DestinationJob
...
+	defer func() {
+		for _, span := range traces {
+			span.End()
+		}
+	}()
+
+	for _, job := range routerJobs {
+		traceParent := job.JobMetadata.TraceParent
+		if traceParent != "" {
+			if _, ok := traces[traceParent]; !ok {
```
fracasula, line 272: This deduplication at the time was done to test things out. If we want to keep it then it might make sense to review the tags. For example if the duplication is due to the fact that the same event is due for multiple destinations then it would be confusing to see the span for one destination only and not for the others. Or we can remove the tag and flatten them. Either way I suggest tinkering a bit with the integration test you built to see how is going to look like with 1 source and 3 destinations. Do you see everything as expected?

integration_test/tracing/tracing_test.go
achettyiitr: For linter. `response body must be closed (bodyclose)`

fracasula: Darn linter wants us to wrap it? I'd rather not wrap and add a `nolint` tbh. Up to you :+1:

integration_test/tracing/tracing_test.go
```diff
@@ -0,0 +1,871 @@
...
+) {
+	t.Helper()
+
+	require.Eventually(t, func() bool {
+		var jobsCount int
+		require.NoError(t, db.QueryRow(fmt.Sprintf(`SELECT count(*) FROM unionjobsdbmetadata('%s',1) WHERE job_state = '%s' AND parameters->>'traceparent' is not NULL;`, queue, state)).Scan(&jobsCount))
+		t.Logf("%s %sJobCount: %d", queue, state, jobsCount)
+		return jobsCount == expectedCount
+	},
+		20*time.Minute,
```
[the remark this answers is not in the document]
achettyiitr: This was pushed by mistake. Reverting back.

processor/processor.go
achettyiitr: As discussed, removing `jobid` and the `uuid` from processor.

integration_test/tracing/tracing_test.go
achettyiitr: Added some helpers.

integration_test/tracing/tracing_test.go
```diff
@@ -0,0 +1,871 @@
...
+		}
+
+		for _, trace := range lo.Filter(lo.Flatten(zipkinTraces), func(trace tracemodel.ZipkinTrace, _ int) bool {
+			return trace.Name == "gw.webrequesthandler"
+		}) {
+			require.Equal(t, "batch", trace.Tags["reqType"])
+			require.Equal(t, "/v1/batch", trace.Tags["path"])
+			require.Equal(t, "source-1", trace.Tags["sourceId"])
+			require.Equal(t, "gateway", trace.Tags["otel.library.name"])
+		}
```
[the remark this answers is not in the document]
achettyiitr: Added a helper for this.

router/worker.go
achettyiitr: With 1 source connected to 3 destinations, we are getting `rt.pickup` and `rt.process` for each destination.


Screenshot 2023-12-20 at 4 03 43 PM
Screenshot 2023-12-20 at 4 03 41 PM
Screenshot 2023-12-20 at 4 03 35 PM

fracasula (approved):

router/handle.go
```diff
@@ -230,6 +238,22 @@ func (rt *Handle) pickup(ctx context.Context, partition string, workers []*worke
 		lastJob = job
 		slot, err := rt.findWorkerSlot(ctx, workers, job, blockedOrderKeys)
 		if err == nil {
+			traceParent := gjson.GetBytes(job.Parameters, "traceparent").String()
+			if traceParent != "" {
+				if _, ok := traces[traceParent]; !ok {
```
lvrach, line 243: Do we need special handling if traceParent is empty? It is not something we do in the processor.

lvrach (approved):
Non-blocking question about handling empty parent trace id. 

We should also consider the memory impact of introducing a span struct per event. What will the tracing overhead be when we choose not to sample an event?

router/handle.go
achettyiitr: If traceParent is empty for some reason, we can't probably trace them back.

achettyiitr: > It is not something we do in the processor.

Are u referring to in here? https://github.com/rudderlabs/rudder-server/blob/d4eca46bfe04d6335b34721496469d4836d7ae98/processor/processor.go#L1562

achettyiitr:
> We should also consider the memory impact of introducing a span struct per event. What will the tracing overhead be when we choose not to sample an event?

We are only storing `traceParent` per event.

fracasula:
> We should also consider the memory impact of introducing a span struct per event. What will the tracing overhead be when we choose not to sample an event?

I did some profiling some time ago but I can't get meaningful data unless we profile this branch itself instead of using a dedicated benchmark.

In any case, this is behind a feature flag and meant to be used in the E2E tests only for now. If tracing isn't enabled there is no noticeable impact. Once we have a release with this we can enable it for E2E and then enable it on-demand and monitor the impact for example on hosted by deploying a single zipkin instance.

router/handle.go
achettyiitr: It got missed. Added now.

lvrach (approved):

PR #4262, opened 2023-12-27
chore: enabling gateway to ingest events even when sharedDB is down

mihir20:
# Description
Gateway was not able to start web server, If sharedDB is down for some reason. Following steps in startup are blocking web server to start.
1. RSources service was not able to start as it was not able to connect with sharedDB and setup db tables and replication.
2. DrainConfigManager was not able start and run migrations for the drain config db tables present in shared DB.

Solution: 
When a new gateway instance is coming up, 
1. while setting up RSources service we will not setup sharedDB tables and replication.
2. while setting up DrainConfigManager we will not halt gateway setup on failure of migration for drain config tables.

We will raise an alert for the failure logs. AI for the alert will be to restart gateway instance.

## Linear Ticket
https://linear.app/rudderstack/issue/PIPE-484/gateway-should-accept-data-even-if-it-is-not-able-to-connect-with

## Security

### DIFF
--- a/app/apphandlers/gatewayAppHandler.go
+++ b/app/apphandlers/gatewayAppHandler.go
@@ -121,7 +121,7 @@ func (a *gatewayApp) StartRudderCore(ctx context.Context, options *app.Options) 
 	if err != nil {
 		return fmt.Errorf("failed to create rate limiter: %w", err)
 	}
-	rsourcesService, err := NewRsourcesService(deploymentType)
+	rsourcesService, err := NewRsourcesService(deploymentType, false)
 	if err != nil {
 		return err
 	}
@@ -132,15 +132,18 @@ func (a *gatewayApp) StartRudderCore(ctx context.Context, options *app.Options) 
 	})
 	drainConfigManager, err := drain_config.NewDrainConfigManager(config, a.log.Child("drain-config"))
 	if err != nil {
-		return fmt.Errorf("drain config manager setup: %v", err)
+		a.log.Errorw("drain config manager setup fail", "error", err)
 	}
-	defer drainConfigManager.Stop()
-	g.Go(misc.WithBugsnag(func() (err error) {
-		return drainConfigManager.DrainConfigRoutine(ctx)
+
+	drainConfigHttpHandler := http.Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "unable to start drain config http handler", http.StatusInternalServerError)
 	}))
-	g.Go(misc.WithBugsnag(func() (err error) {
-		return drainConfigManager.CleanupRoutine(ctx)
-	}))
+
+	if drainConfigManager != nil {
+		defer drainConfigManager.Stop()
+		drainConfigHttpHandler = drainConfigManager.DrainConfigHttpHandler()
+	}
+
 	err = gw.Setup(
 		ctx,
 		config, logger.NewLogger().Child("gateway"), stats.Default,
@@ -148,7 +151,7 @@ func (a *gatewayApp) StartRudderCore(ctx context.Context, options *app.Options) 
 		rateLimiter, a.versionHandler, rsourcesService, transformerFeaturesService, sourceHandle,
 		gateway.WithInternalHttpHandlers(
 			map[string]http.Handler{
-				"/drain": drainConfigManager.DrainConfigHttpHandler(),
+				"/drain": drainConfigHttpHandler,
 			},
 		),
 	)
--- a/app/apphandlers/gatewayAppHandler.go
+++ b/app/apphandlers/gatewayAppHandler.go
@@ -132,13 +132,10 @@ func (a *gatewayApp) StartRudderCore(ctx context.Context, options *app.Options) 
 	})
 	drainConfigManager, err := drain_config.NewDrainConfigManager(config, a.log.Child("drain-config"))
 	if err != nil {
-		a.log.Errorw("drain config manager setup fail", "error", err)
+		a.log.Errorw("drain config manager setup failed while starting gateway", "error", err)
 	}
 
-	drainConfigHttpHandler := http.Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		http.Error(w, "unable to start drain config http handler", http.StatusInternalServerError)
-	}))
-
+	drainConfigHttpHandler := drain_config.ErrorResponder("unable to start drain config http handler")
 	if drainConfigManager != nil {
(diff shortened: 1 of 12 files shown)


internal/drain-config/drainConfig.go
```diff
@@ -45,7 +45,10 @@ func NewDrainConfigManager(conf *config.Config, log logger.Logger) (*drainConfig
 		log.Errorw("db setup", "error", err)
 		return nil, fmt.Errorf("db setup: %v", err)
 	}
-	if err := migrate(db); err != nil {
+	if err := db.Ping(); err != nil {
```
atzoum, line 48: We shouldn't ignore errors in case we are in processor, so whether it will fail or not, should be configurable

services/rsources/handler.go
```diff
@@ -488,6 +488,11 @@ func (sh *sourcesHandler) init() error {
 	}
 
 	if sh.sharedDB != nil {
+		if err := sh.sharedDB.Ping(); err != nil {
```
atzoum, line 491: We shouldn't ignore errors in case we are in processor, so whether it will fail or not, should be configurable

app/apphandlers/gatewayAppHandler.go
```diff
@@ -132,23 +132,26 @@ func (a *gatewayApp) StartRudderCore(ctx context.Context, options *app.Options)
 	})
 	drainConfigManager, err := drain_config.NewDrainConfigManager(config, a.log.Child("drain-config"))
 	if err != nil {
-		return fmt.Errorf("drain config manager setup: %v", err)
+		a.log.Errorw("drain config manager setup fail", "error", err)
```
atzoum, line 135: So in case of an error, gateway will be operating in degraded mode, at least with respect to draining jobs. Do we plan to setup a P0 alert for this?

mihir20: yes we will setup an alert for this

internal/drain-config/drainConfig.go
```diff
@@ -45,17 +45,16 @@ func NewDrainConfigManager(conf *config.Config, log logger.Logger) (*drainConfig
...
 		log.Errorw("db migrations", "error", err)
-		return nil, fmt.Errorf("db migrations: %v", err)
 	}
 	return &drainConfigManager{
 		log:  log,
 		conf: conf,
 		db:   db,
 
 		done: &atomic.Bool{},
-	}, nil
```
atzoum, line 58: Can we add a doc on the producer function explaining this special scenario (returning a non-nil manager with a non nil error)?

internal/drain-config/drainConfig.go
```diff
@@ -189,9 +188,6 @@ func setupDBConn(conf *config.Config) (*sql.DB, error) {
 	if err != nil {
 		return nil, fmt.Errorf("db open: %v", err)
 	}
-	if err := db.Ping(); err != nil {
-		return nil, fmt.Errorf("db ping: %v", err)
-	}
```
achettyiitr, line 194: Don't we need this?

app/apphandlers/gatewayAppHandler.go
achettyiitr, line 135:
```suggestion
		a.log.Errorw("drain config manager setup failed", "error", err)
```

app/apphandlers/gatewayAppHandler.go
```diff
@@ -132,23 +132,26 @@ func (a *gatewayApp) StartRudderCore(ctx context.Context, options *app.Options)
...
-	defer drainConfigManager.Stop()
-	g.Go(misc.WithBugsnag(func() (err error) {
-		return drainConfigManager.DrainConfigRoutine(ctx)
-	}))
-	g.Go(misc.WithBugsnag(func() (err error) {
-		return drainConfigManager.CleanupRoutine(ctx)
+
+	drainConfigHttpHandler := http.Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "unable to start drain config http handler", http.StatusInternalServerError)
 	}))
```
achettyiitr, line 140: This can probably be moved to `drain_config` package.

```GO
func ErrorResponder() http.Handler {
	return http.Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		http.Error(w, "unable to start drain config http handler", http.StatusInternalServerError)
	}))
}
```

Consumer can use this as:
```GO
drainConfigHttpHandler := drain_config.ErrorResponder()
```

app/apphandlers/gatewayAppHandler.go
atzoum: It would be nice to have the relevant pr mentioned here :)

app/apphandlers/gatewayAppHandler.go
```diff
@@ -132,23 +132,26 @@ func (a *gatewayApp) StartRudderCore(ctx context.Context, options *app.Options)
...
+
+	if drainConfigManager != nil {
+		defer drainConfigManager.Stop()
```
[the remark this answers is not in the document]
mihir20: No we don't need `DrainConfigRoutine` and `CleanupRoutine` in gateway, these will be run with processor

internal/drain-config/drainConfig.go
mihir20: No we will just setup the connection pool and not ping the db while setting up. Migrations will fail if DB connection fails.

achettyiitr: Don't we need this for Gateway and Processor?

mihir20: For processor also, migrations will fail if DB connection fails. Processor will handle the error and setup will fail, gateway will just log the error and continue to ingest events.

app/apphandlers/gatewayAppHandler.go
mihir20: https://github.com/rudderlabs/rudder-devops/pull/9820

internal/drain-config/drainConfig.go
mihir20: done

achettyiitr (approved):

internal/drain-config/drainConfig.go
lvrach: Have we consider returning an error but having the gateway app ignoring it?

lvrach (approved):
Approved with non blocking comment

cisse21 (approved):

internal/drain-config/drainConfig.go
mihir20: yes, we are returning error and ignoring it in gateway

PR #4263, opened 2023-12-27
chore: change the max limit to take multiplier of the existing limits

BonapartePC:
# Description

chore: change the max limit to take `multiplier` of the existing limits rather than the same in case the `maxLimit` env is missing. Default: 150% of the current limit

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-645/change-maxlimit-to-use-130percent-of-existing-limits

## Security

### DIFF
--- a/router/throttler/adaptive.go
+++ b/router/throttler/adaptive.go
@@ -14,7 +14,7 @@ type adaptiveThrottleConfig struct {
 	window      misc.ValueLoader[time.Duration]
 	minLimit    misc.ValueLoader[int64]
 	maxLimit    misc.ValueLoader[int64]
-	staticLimit misc.ValueLoader[int64]
+	staticLimit func() int64
 }
 
 func (c *adaptiveThrottleConfig) readThrottlingConfig(config *config.Config, destName, destID string) {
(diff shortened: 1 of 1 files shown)


router/throttler/adaptive.go
```diff
@@ -72,6 +74,10 @@ func (t *adaptiveThrottler) getLimit() int64 {
 	if t.limitFactorMeasurement != nil {
 		t.limitFactorMeasurement.Gauge(limitFactor)
 	}
-	limit := int64(float64(t.config.maxLimit.Load()) * limitFactor)
+	maxLimit := t.config.maxLimit.Load()
+	if maxLimit == 0 {
+		maxLimit = t.config.staticLimit.Load() * 130 / 100 // increase static limit by 30% if maxLimit is not set
```
atzoum, line 79: Can we also have this multiplier configurable & hot reloadable? We could also start with `150%`

atzoum (approved):

router/throttler/adaptive.go
BonapartePC: Done

achettyiitr (approved):

router/throttler/adaptive.go
```diff
@@ -25,16 +25,28 @@ func (c *adaptiveThrottleConfig) readThrottlingConfig(config *config.Config, des
...
+		`Router.throttler.adaptive.limitMultiplier`)
+	limit := config.GetReloadableInt64Var(0, 1,
+		fmt.Sprintf(`Router.throttler.%s.%s.limit`, destName, destID),
+		fmt.Sprintf(`Router.throttler.%s.limit`, destName))
+	c.maxLimit = func() int64 {
+		maxLimit := maxLimit.Load()
+		if maxLimit > 0 {
+			return maxLimit
+		}
+		return limit.Load() * limitMultiplier.Load()
```
atzoum, line 44: Either this needs to be divided by 100, or multiplier become an actual multiplier and not a percentage :)

BonapartePC: Fixed it

cisse21 (approved):

PR #4264, opened 2023-12-28
feat: add 'config' field to proxy request

Gauravudia:
# Description

Resolves INT-1233
Update the transformer <> server proxy contract to support including '**config**' at the root level in the proxy request from the server to the transformer. With these changes, the server will now append an additional 'config' field to the proxy request, propagating it to the transformer for utilization.

## Linear Ticket
https://linear.app/rudderstack/issue/INT-1233/transformerproxy-propagate-destination-config-to-transformerproxy-api

## Security


router/transformer/transformer.go
```diff
@@ -65,6 +65,7 @@ type ProxyRequestMetadata struct {
 type ProxyRequestPayload struct {
 	integrations.PostParametersT
 	Metadata []ProxyRequestMetadata `json:"metadata"`
+	Config   map[string]interface{} `json:"config"`
```
koladilip, line 68: DestinationConfig would be better avoid confusion

koladilip (approved):

achettyiitr (approved):

router/transformer/transformer.go
Gauravudia: Incorporated the changes.

PR #4268, opened 2023-12-28
feat: create openAPI for rudder-server public API

pras-stack:
# Description

- Creates RudderStack HTTP API Documentation using go-swagger.
- Serve SwaggerUI within rudder-server.

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-593/create-openapi-for-rudder-server-public-api

## Security

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,3 +106,21 @@ proto: install-tools ## Generate protobuf files
 .PHONY: bench-kafka
 bench-kafka:
 	go test -count 1 -run BenchmarkCompression -bench=. -benchmem ./services/streammanager/kafka/client
+
+.PHONY: check-swagger
+check-swagger:
+	which swagger || (GO111MODULE=off go get -u github.com/go-swagger/go-swagger/cmd/swagger)
+
+.PHONY: swagger
+swagger: check-swagger
+	GO111MODULE=on go mod vendor  && GO111MODULE=off swagger generate spec -o ./swagger.yaml --scan-models
+
+.PHONY: serve-swagger
+serve-swagger: check-swagger
+	go install github.com/go-swagger/go-swagger/cmd/swagger
+	swagger serve ./swagger.yaml
+
+.PHONY: serve-swagger-ui
+serve-swagger-ui: check-swagger
+	go install github.com/go-swagger/go-swagger/cmd/swagger
+	swagger serve -F swagger ./swagger.yaml
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -1,0 +1,325 @@
+swagger: '2.0'
+info:
+  title: RudderStack HTTP API
+  version: 1.0.0
+  description: >-
+    The RudderStack HTTP API lets you track your customer event data and route
+    it to your desired destinations.
+  termsOfService: 'https://www.rudderstack.com/master-service-agreement/'
+  contact:
+    email: support@rudderstack.com
+  license:
+    name: Apache 2.0
+    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
+basePath: /v1
+consumes:
+  - application/json
+produces:
+  - application/json
+schemes:
+  - http
+paths:
+  /api/v1/identify:
+    post:
+      tags:
+        - HTTP API
+      summary: Identify
+      description: >-
+        The identify call lets you associate a visiting user to their actions
+        and record any associated traits.
+      operationId: Identify
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - name: body
+          in: body
+          required: true
+          schema:
+            $ref: '#/definitions/IdentifyPayload'
+      responses:
+        '200':
+          description: OK
+  /v1/track:
+    post:
+      tags:
+        - HTTP API
+      summary: Track
+      description: >-
+        The track call lets you track user actions along with any properties
+        associated with them.
+      operationId: Track
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - name: body
+          in: body
+          required: true
+          schema:
+            $ref: '#/definitions/TrackPayload'
+      responses:
+        '200':
+          description: OK
+  /v1/page:
+    post:
+      tags:
+        - HTTP API
+      summary: Page
+      description: >-
+        The page call lets you record your website’s page views with any
+        additional relevant information about the viewed page.
+      operationId: Page
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - name: body
+          in: body
+          required: true
+          schema:
+            $ref: '#/definitions/PagePayload'
+      responses:
+        '200':
+          description: OK
+  /v1/screen:
+    post:
+      tags:
+        - HTTP API
+      summary: Screen
+      description: >-
+        The screen call is the mobile equivalent of the page call. It lets you
+        record whenever your user views their mobile screen with any additional
+        relevant information about the screen.
+      operationId: Screen
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - name: body
+          in: body
+          required: true
+          schema:
+            $ref: '#/definitions/ScreenPayload'
+      responses:
+        '200':
+          description: OK
+  /v1/group:
+    post:
+      tags:
+        - HTTP API
+      summary: Group
+      description: >-
+        The group call lets you link an identified user with a group such as a
+        company, organization, or an account. It also lets you record any custom
+        traits associated with that group, like the name of the company, the
+        number of employees, etc.
+      operationId: Group
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - name: body
+          in: body
+          required: true
+          schema:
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -106,3 +106,21 @@ proto: install-tools ## Generate protobuf files
...
+	GO111MODULE=on go mod vendor  && GO111MODULE=off swagger generate spec -o ./swagger.yaml --scan-models
+
+.PHONY: serve-swagger
+serve-swagger: check-swagger
+	go install github.com/go-swagger/go-swagger/cmd/swagger
+	swagger serve ./swagger.yaml
+
+.PHONY: serve-swagger-ui
+serve-swagger-ui: check-swagger
+	go install github.com/go-swagger/go-swagger/cmd/swagger
```
achettyiitr, line 125: We can probably move this to install-tools.

achettyiitr: `check-swagger` is not working. This would be sufficient in Makefile.
```GO
.PHONY: serve-swagger-ui
serve-swagger-ui: install-tools
	swagger validate ./swagger.yaml
	swagger serve -F swagger ./swagger.yaml
```

pras-stack: Done.

Makefile
```diff
@@ -106,3 +106,21 @@ proto: install-tools ## Generate protobuf files
...
+
+.PHONY: swagger
+swagger: check-swagger
+	GO111MODULE=on go mod vendor  && GO111MODULE=off swagger generate spec -o ./swagger.yaml --scan-models
+
+.PHONY: serve-swagger
+serve-swagger: check-swagger
+	go install github.com/go-swagger/go-swagger/cmd/swagger
+	swagger serve ./swagger.yaml
+
```
achettyiitr, line 122: We don't need these right?

Makefile
achettyiitr: You would need to add `go install github.com/go-swagger/go-swagger/cmd/swagger` inside install-tools. Also, we can probably remove

Makefile
pras-stack: Don't we need it for installations ?

achettyiitr: This would be sufficient. `go install github.com/go-swagger/go-swagger/cmd/swagger` which we will be added in `install-tools`.

pras-stack: done.

Makefile
```diff
@@ -106,3 +106,8 @@ proto: install-tools ## Generate protobuf files
 .PHONY: bench-kafka
 bench-kafka:
 	go test -count 1 -run BenchmarkCompression -bench=. -benchmem ./services/streammanager/kafka/client
+
+.PHONY: serve-swagger-ui
+serve-swagger-ui: install-tools
```
achettyiitr, line 111: You would need to add `go install github.com/go-swagger/go-swagger/cmd/swagger@latest` inside install-tools.

pras-stack: Added.

Makefile
```diff
@@ -106,3 +106,18 @@ proto: install-tools ## Generate protobuf files
...
+
+install-tools:
+	go install github.com/golang/mock/mockgen@v1.6.0
+	go install mvdan.cc/gofumpt@latest
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28.1
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2.0
+	go install gotest.tools/gotestsum@v1.10.0
+	go install golang.org/x/tools/cmd/goimports@latest
+	go install github.com/go-swagger/go-swagger/cmd/swagger@latest
+	bash ./scripts/install-golangci-lint.sh v1.55.0
```
achettyiitr, line 118: There is already an install-tools, we can replace it with this.

achettyiitr:
ref: https://github.com/rudderlabs/rudder-server/blob/9842f64571f8f5cc08cce90192383e5ddc375fcf/gateway/handle_lifecycle.go#L363
```
/internal/v1/warehouse/fetch-tables
/internal/v1/job-status
/internal/v2/job-status
/v1/job-status
/v1/audiencelist
/v1/merge
/v1/import
/v1/webhook
/v1/warehouse/pending-events
/v1/warehouse/trigger-upload
/v1/warehouse/jobs
/v1/warehouse/fetch-tables
/v1/warehouse/jobs/status
/beacon/v1/batch
/pixel/v1/track
/pixel/v1/page
```

These are missing in the swagger file. Are there plans for adding this later?

Makefile
pras-stack: Done.

swagger.yaml
```diff
@@ -0,0 +1,512 @@
...
+      description: Handles audience list requests.
+      operationId: AudienceList
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters: []
+      responses:
+        '200':
+          description: OK
```
achettyiitr, line 226: We might need to add other possible response codes as well, Few samples:
[429 -> Too many requests.](https://github.com/rudderlabs/rudder-server/blob/9cb9caff24ba945825610ca280609c754ed400f2/gateway/handle.go#L408)
[400 -> Invalid JSON](https://github.com/rudderlabs/rudder-server/blob/9cb9caff24ba945825610ca280609c754ed400f2/gateway/handle.go#L292C2-L292C2)
[400 -> Not a rudder event](https://github.com/rudderlabs/rudder-server/blob/9cb9caff24ba945825610ca280609c754ed400f2/gateway/handle.go#L300)
[400 -> Non identifiable](https://github.com/rudderlabs/rudder-server/blob/9cb9caff24ba945825610ca280609c754ed400f2/gateway/handle.go#L340)
[413 -> Request body too large.](https://github.com/rudderlabs/rudder-server/blob/master/gateway/handle.go#L426)

swagger.yaml
```diff
@@ -0,0 +1,666 @@
...
+        '402':
+          description: Non Identifiable Request
+        '403':
+          description: Request Body Read Failed
+        '413':
+          description: Request size too large
+        '429':
+          description: Too many requests
+        '498':
+          description: Missing Authorization Header
```
achettyiitr, line 204: Don't think we send these response codes: 402, 403, 498. Can u please check once? This is the [status map](https://github.com/rudderlabs/rudder-server/blob/bd365b50d7b537c5144cee495af6f491ed2736c1/gateway/response/response.go#L74) that we use while sending responses to the client.

pras-stack: 403 is present in the status map, will remove 402 and 498.

swagger.yaml
```diff
@@ -0,0 +1,633 @@
+swagger: '2.0'
```
fracasula, line 1: Do we want to use OAS 3 instead (see [here](https://swagger.io/docs/specification/basic-structure/))? cc @lvrach   

```suggestion
openapi: 3.0.3
```

lvrach: yes lets use openAPI

fracasula: @pras-stack I see that you reposted this for review (Slack) but this file hasn't been updated to use the OpenAPI 3 specs, right?

pras-stack: yes done.

Makefile
```diff
@@ -106,3 +96,18 @@ proto: install-tools ## Generate protobuf files
...
+
+.PHONY: serve-swagger-ui
+serve-swagger-ui: install-tools
```
fracasula, line 111: Shouldn't you remove this recipe? It gives the following:

```
swagger validate --stop-on-error ./swagger.yaml
open ./swagger.yaml: no such file or directory
```

It should be replaced with a recipe that is compatible with the OpenAPI 3 specs and it can provide an HTML file and possibly also serve it via the browser.

fracasula: As mentioned in my previous comment this is the [html2 generator](https://openapi-generator.tech/docs/generators/html2), you can use that one.

Once you have the HTML file you can serve it via Go. If you can't find something reasonably small to do that feel free to build a cmd on your own ([it's quite easy](https://stackoverflow.com/a/39324149/828366)).

openapi.yaml
```diff
@@ -0,0 +1,561 @@
+openapi: 3.0.0
```
fracasula, line 1: We should be able to use 3.1.0. It's worth a try I guess.

https://swagger.io/specification/

```suggestion
openapi: 3.1.0
```

fracasula (changes requested):
@pras-stack can we have two recipes in the Makefile? One to generate the HTML file and one to serve it (the latter would start an HTTP server and open a browser to show the docs generated with the former).

Then we should add a `verify` step in our CI to make sure that the recipe to generate the HTML is always called when people make changes to the `openapi.yaml` file. It's easier done than said actually, you can see an example [here](https://github.com/rudderlabs/rudder-observability-kit/blob/main/.github/workflows/verify.yaml) which is using `git diff`.

openapi.yaml
pras-stack: done.

fracasula:
During standup we also talked about having the `index.html` file served by the rudder-server API itself.
In that case we would need to embed the `index.html` file in the Docker image.
cc @lvrach @achettyiitr

PR #4272, opened 2023-12-29
fix: bigquery job statistics for rate limit

achettyiitr:
# Description

- While getting job statistics for BigQuery, it's getting rate limited for gelato (PRO). 
  ```
  load table: append job statistics: getting job: googleapi: Error 429: Quota exceeded for quota metric 'Requests' and limit 'Requests per day' of service 'bigquery.googleapis.com'
  ```
- We can ignore the errors as these statistics are just for tracking alerts and we don't want to retry the syncs due to the following error.

## Linear Ticket

- Resolves PIPE-653

## Security

### DIFF
--- a/warehouse/integrations/bigquery/bigquery.go
+++ b/warehouse/integrations/bigquery/bigquery.go
@@ -466,6 +466,11 @@ func (bq *BigQuery) jobStatistics(
 	)
 	bqJob, err := bqJobGetCall.Context(ctx).Location(job.Location()).Fields("statistics").Do()
 	if err != nil {
+		// In case of rate limit error, return empty statistics
+		var e *googleapi.Error
+		if errors.As(err, &e) && e.Code == 429 {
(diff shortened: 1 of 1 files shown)


fracasula:
> and we don't want to retry the syncs due to the following error.

So we don't want to retry the syncs if we get throttled by BigQuery? Wouldn't the data be lost then? @achettyiitr

achettyiitr:
> > and we don't want to retry the syncs due to the following error.
> 
> So we don't want to retry the syncs if we get throttled by BigQuery? Wouldn't the data be lost then? @achettyiitr

The following error comes once we have successfully loaded the data in BQ. Then used the `jobID` to get the `statistics` of how many rows got loaded. So, it fails at this point when getting the statistics which is fine for us as it is only for tracking our internal alerts.

fracasula (approved):

cisse21 (approved):

PR #4274, opened 2024-01-02
feat: new internal api for workspace status

achettyiitr:
# Description

- New GET endpoint for `internal/v1/workspace/{workspaceID}/status` to check the status for a workspace whether it is serving or not.
- More about it in [here](https://linear.app/rudderstack/issue/PIPE-627/research-zero-downtime-cross-tier-migration) and [here](https://www.notion.so/rudderstacks/Zero-Downtime-Cross-tier-migration-1e8b7a214d2544528c699c684331ca94?pvs=4).

## Linear Ticket

- Resolves PIPE-652

## Security

### DIFF
--- a/gateway/handle_http.go
+++ b/gateway/handle_http.go
@@ -3,6 +3,9 @@ package gateway
 import (
 	"context"
 	"net/http"
+	"strconv"
+
+	"github.com/go-chi/chi/v5"
 
 	gwtypes "github.com/rudderlabs/rudder-server/gateway/internal/types"
 	"github.com/rudderlabs/rudder-server/gateway/response"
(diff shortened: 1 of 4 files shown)


gateway/handle_http.go
```diff
@@ -66,6 +69,24 @@ func (*Handle) robotsHandler(w http.ResponseWriter, _ *http.Request) {
 	_, _ = w.Write([]byte("User-agent: * \nDisallow: / \n"))
 }
 
+// workspaceStatusHandler - handler for workspace status requests
+func (gw *Handle) workspaceStatusHandler() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		workspaceID := chi.URLParam(r, "workspaceID")
```
fracasula, line 75: Shouldn't we at least check if the workspaceID is empty and return a 400? Also, do we have a format for the workspaceID or it can be any string?

PR #4287, opened 2024-01-05
fix: return only the latest namespace entry when fetching tables from warehouse

am6010:
# Description

Modify the query in the `GetTablesForConnection` method to group by source_id and destination_id. This will allow us to get only the latest namespace per source_id and destination_id pair.

Tested 
- extend unit tests.
- manually tested in profiles-qa env
```
curl --location --request GET 'http://localhost:8081/internal/v1/warehouse/fetch-tables' \
--header 'Content-Type: application/json' \
--data '{
    "connections": [
        {
        "source_id": "2ZRRRlvMuP2BLtI0Fjd9O8IvI1k",
        "destination_id": "2ZRV4eRj1ECbSbaBR6XJp3Tod08"
        }
    ]
}'
```

## Linear Ticket

[ticket](https://linear.app/rudderstack/issue/PRO-1605/show-only-latest-schema-for-each-source-in-sources-list)

## Security


achettyiitr (approved):

warehouse/internal/repo/schema_test.go
```diff
@@ -142,5 +142,50 @@ func TestWHSchemasRepo(t *testing.T) {
...
+		t.Log("multiple")
+		latestNamespace := "latest_namespace"
+		sourceID2 := "source_id_2"
+		destinationID2 := "destination_id_2"
+		connection2 := warehouseutils.SourceIDDestinationID{SourceID: sourceID2, DestinationID: destinationID2}
+		schemaLatest := model.WHSchema{
+			UploadID:        2,
+			SourceID:        sourceID,
+			Namespace:       latestNamespace,
+			DestinationID:   destinationID,
```
fracasula, line 155: Shouldn't this be `destinationID2`? :thinking: 
Same with the `sourceID` above. I thought you would insert two records with the same source and destination and then expect only one in `GetTablesForConnection`.

warehouse/internal/repo/schema_test.go
```diff
@@ -142,5 +142,50 @@ func TestWHSchemasRepo(t *testing.T) {
...
+			DestinationID: destinationID,
+			Namespace:     latestNamespace,
+			Tables:        []string{"table_name_1", "table_name_2"},
+		})
+		require.Contains(t, expectedTableNames, warehouseutils.FetchTableInfo{
+			SourceID:      sourceID2,
+			DestinationID: destinationID2,
+			Namespace:     namespace,
+			Tables:        []string{"table_name_1", "table_name_2"},
+		})
```
fracasula, line 189: You don't really need the `connection2` and `schema2`, right? I find it confusing since the assertion really is about something else here.

I think it would make more sense to check the whole slice (with `ElementsMatch` instead of `Contains`) once before insertion and then again after we insert the latest schema to make sure nothing changes because we already have an entry for that sourceID/destID pair.

What do you think? Something like this:

```suggestion
		t.Log("multiple")
		latestNamespace := "latest_namespace"
		expectedTableNames, err = r.GetTablesForConnection(ctx, []warehouseutils.SourceIDDestinationID{connection})
		require.NoError(t, err)
		require.ElementsMatch(t, expectedTableNames, []warehouseutils.FetchTableInfo{{
			SourceID:      sourceID,
			DestinationID: destinationID,
			Namespace:     latestNamespace,
			Tables:        []string{"table_name_1", "table_name_2"},
		}})
		schemaLatest := model.WHSchema{
			UploadID:        2,
			SourceID:        sourceID,
			Namespace:       latestNamespace,
			DestinationID:   destinationID,
			DestinationType: destinationType,
			Schema:          schemaModel,
			CreatedAt:       now,
			UpdatedAt:       now,
		}
		_, err = r.Insert(ctx, &schemaLatest)
		require.NoError(t, err)
		expectedTableNames, err = r.GetTablesForConnection(ctx, []warehouseutils.SourceIDDestinationID{connection})
		require.NoError(t, err)
		require.ElementsMatch(t, expectedTableNames, []warehouseutils.FetchTableInfo{{
			SourceID:      sourceID,
			DestinationID: destinationID,
			Namespace:     latestNamespace,
			Tables:        []string{"table_name_1", "table_name_2"},
		}}, "the new schema should not have changed the returned tables since we already had a schema for the same source id and destination id pair")
```

fracasula (commented):
Changes look good, I left a comment about the test.

warehouse/internal/repo/schema_test.go
am6010: @fracasula  thank you for reviewing. Yeah, I agree it makes the test more complicated. I was trying to test the output of many connections (source id - destination id pairs). 
I have updated the test only to test one connection. I don't think we need the test before inserting, because we do the same test before.

fracasula (approved):

PR #4297, opened 2024-01-10
chore: error index file path collision

achettyiitr:
# Description

- Identified and addressed an issue with over 10k records (around 2M) for a specific **sourceID** and **failedAt** during PWCC debugging.
- Introduced a **uuid** to the error-index file path, maintaining backward compatibility with the existing format.
- Successfully executed e2e tests [here](https://github.com/rudderlabs/e2e/actions/runs/7475473131/job/20343689226).

## Linear Ticket

- Resolves PIPE-682

## Security

### DIFF
--- a/enterprise/reporting/error_index/worker.go
+++ b/enterprise/reporting/error_index/worker.go
@@ -12,6 +12,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/google/uuid"
+
 	"github.com/rudderlabs/rudder-go-kit/filemanager"
 
 	"github.com/samber/lo"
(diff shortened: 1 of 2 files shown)


enterprise/reporting/error_index/worker.go
```diff
@@ -191,6 +193,7 @@ func (w *worker) uploadJobs(ctx context.Context, jobs []*jobsdb.JobT) ([]*jobsdb
 		if err != nil {
 			return nil, fmt.Errorf("uploading aggregated payloads: %w", err)
 		}
+		w.log.Debugw("successfully uploaded aggregated payloads", "location", uploadFile.Location)
```
fracasula, line 196: Can we use `Debugn` instead?

fracasula (approved):

atzoum (approved):
Assuming that rudder-sources release will precede rudder-server release containing this change, since we wouldn't want to end up in a situation where a replay task would fail to parse an error index file name

achettyiitr:
> cede



> Assuming that rudder-sources release will precede rudder-server release containing this change, since we wouldn't want to end up in a situation where a replay task would fail to parse an error index file name

rudder sources changes are backward compatible. It will work with both formats.

PR #4304, opened 2024-01-11
feat: add support for regulation from features.json

koladilip:
# Description
Read regulation destinations from features.json directly, this change has been done with backward compatibility in mind so releases need not be coordinated with transformer

## Linear Ticket

https://linear.app/rudderstack/issue/INT-1336/add-support-for-regulation-in-featuresjson
Resolves INT-1336

## Security


services/transformer/features.go
```diff
@@ -31,7 +34,8 @@ var defaultTransformerFeatures = `{
 	"routerTransform": {
 	  "MARKETO": true,
 	  "HS": true
-	}
+	},
+	"regulation": ["AM"],
```
chandumlg, line 38: Why put this in default?

services/transformer/features.go
```diff
@@ -53,6 +57,11 @@ func NewNoOpService() FeaturesService {
 
 type noopService struct{}
 
+// Regulation implements FeaturesService.
```
chandumlg, line 60: Delete the comment?

services/transformer/features_impl.go
```diff
@@ -42,6 +43,16 @@ func (t *featuresService) RouterTransform(destType string) bool {
 	return gjson.GetBytes(t.features, "routerTransform."+destType).Bool()
 }
 
+func (t *featuresService) Regulation() []string {
```
chandumlg, line 46: A test for this function would be good.

chandumlg (approved):

services/transformer/features.go
koladilip: Just thought of keeping something like previously, what do you suggest?

services/transformer/features.go
koladilip: Done

cisse21 (approved):

PR #4310, opened 2024-01-12
feat: different retry limits for sources jobs in batchrouter

Sidddddarth:
# Description

to fail fast in case of RSources jobs.

## Linear Ticket

[fixes INT-1325](https://linear.app/rudderstack/issue/PIPE-689/fail-fast-for-batch-router)

## Security


router/batchrouter/handle_async.go
```diff
@@ -73,9 +73,8 @@ func getPollInput(job *jobsdb.JobT) common.AsyncPoll {
 	return common.AsyncPoll{ImportId: importId}
 }
 
-func enhanceErrorResponseWithFirstAttemptedAtt(msg stdjson.RawMessage, errorResp []byte) (time.Time, []byte) {
-	firstAttemptedAt := getFirstAttemptAtFromErrorResponse(msg)
-	return firstAttemptedAt, routerutils.EnhanceJsonWithTime(firstAttemptedAt, "firstAttemptedAt", errorResp)
+func enhanceErrorResponseWithFirstAttemptedAtt(msg stdjson.RawMessage, errorResp []byte) []byte {
```
fracasula, line 76: The fact that the response we're enhancing is an error one doesn't seem to be relevant.

```suggestion
func enhanceResponseWithFirstAttemptedAt(msg stdjson.RawMessage, resp []byte) []byte {
```

router/worker.go
```diff
@@ -1070,13 +1070,23 @@ func (w *worker) sendDestinationResponseToConfigBackend(payload json.RawMessage,
 }
 
 func (w *worker) retryLimitReached(status *jobsdb.JobStatusT) bool {
+	respStatusCode, _ := strconv.Atoi(status.ErrorCode)
+	switch respStatusCode {
+	case types.RouterTimedOutStatusCode,
+		types.RouterUnMarshalErrorCode: // 5xx errors
```
fracasula, line 1076: If there is a problem unmarshaling in the router then there is no point in retrying, right? So we can return `true`?

router/batchrouter/handle_async.go
Sidddddarth: True
I think the function was named as such because it's only used to enhance `jobsdb.JobStatus`'s `ErrorResponse` field.
Changed it anyway.

router/batchrouter/batchrouter_test.go
```diff
@@ -299,6 +299,119 @@ var _ = Describe("BatchRouter", func() {
...
+			batchrouter.uploadFreq = misc.SingleValueLoader(time.Microsecond)
+			batchrouter.mainLoopFreq = misc.SingleValueLoader(time.Microsecond)
+			ctx, cancel := context.WithCancel(context.Background())
+			var wg sync.WaitGroup
+			wg.Add(1)
+			go func() {
+				batchrouter.mainLoop(ctx)
+				wg.Done()
+			}()
+			time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
Sidddddarth: Just used the test above this for a layout, without giving much thought.
Good catch.

router/worker.go
Sidddddarth: What you say makes sense, but this is existing behaviour and I think there's merit to it.
I think it's this way because an unmarshal error means a breach of transformer<->server contract.
Aborting those jobs would only force us to replay in these cases.
Whereas letting them fail gives us time to fix the transformer/server issues causing this.
@rudderlabs/server_team wdyt?

cisse21: We also do have an alert if the unmarshal fails IIRC so we should let it be as is I guess

router/worker.go
```diff
@@ -1070,13 +1070,23 @@ func (w *worker) sendDestinationResponseToConfigBackend(payload json.RawMessage,
...
+		return false
+	}
+
+	if respStatusCode < 500 {
+		return false
+	}
```
achettyiitr, line 1082: [optional]: 
```suggestion
	switch {
	case respStatusCode == types.RouterTimedOutStatusCode || respStatusCode == types.RouterUnMarshalErrorCode:
		return false // Handle 5xx errors
	case respStatusCode < 500:
		return false
	}
```

router/batchrouter/batchrouter_test.go
```diff
@@ -299,6 +299,131 @@ var _ = Describe("BatchRouter", func() {
 			cancel()
 			wg.Wait()
 		})
+
+		It("should abort jobs that have retry limits, with lesser(default) limits for rSources jobs", func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
achettyiitr, line 304:
```suggestion
			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
			defer cancel()
```

achettyiitr (approved):

router/worker.go
fracasula: OK so here we're saying that, since there is a problem with the transformer then we get an alert as Rohith said and while the messages are being retried we fix the bug with the transformer, create a release and deploy it live, hopefully all of this before the messages are discarded/expired.

Is this actually likely to happen? There are other reasons for this?

router/batchrouter/batchrouter_test.go
Sidddddarth: It's being cancelled during the last store call made by the router, `DoAndReturn`.
Also can't defer this cancel, because we're waiting on the `done` channel to close

fracasula (approved):
Approved but I left a question. It's not blocking for the PR since we're talking about existing behaviour but I'd like to know more.

router/batchrouter/batchrouter_test.go
fracasula: There is no risk of leaking here. If the mock doesn't call `cancel` the timeout triggers and `done` is closed anyway.

PR #4313, opened 2024-01-15
chore: add env to enable throttlerV2

BonapartePC:
# Description

add env to enable throttlerV2

## Security


router/throttler/factory.go
```diff
@@ -63,20 +63,23 @@ func (f *factory) Get(destName, destID string) Throttler {
...
-			"destType":      destName,
-		})
-	}
-	at := &adaptiveThrottler{
-		limiter:                f.adaptiveLimiter,
-		algorithm:              newAdaptiveAlgorithm(f.config, adaptiveConf.window),
-		config:                 adaptiveConf,
-		limitFactorMeasurement: limitFactorMeasurement,
+	var at Throttler = &noOpThrottler{}
+	if f.config.GetBool("Router.throttlerV2.enabled", true) {
```
achettyiitr, line 67: By default, should this be false?

BonapartePC: I made it true by default because the actual idea is to run both adaptive and static together. We are only going to set it as false if adaptive causes an issue after the release

router/throttler/factory_test.go
```diff
@@ -52,6 +52,25 @@ func TestFactory(t *testing.T) {
 			}, 2*time.Second, 100*time.Millisecond, "limit: %d, expectedLimit: %d", ta.getLimit(), maxLimit*5/10)
 		})
 	})
+
+	t.Run("when throttlerV2 is false", func(t *testing.T) {
+		config := config.New()
```
achettyiitr, line 57: [optional]: Variable 'config' collides with imported package name 
```suggestion
		conf := config.New()
```

achettyiitr (approved):

lvrach (approved):

PR #4324, opened 2024-01-16
chore: integration test for rETL flow

lvrach:
# Description

An integration test for the recently added `/retl` endpoint. This new endpoint lets users also provide destination id to which they want to route their events to. Tests added to verify sources stats' behaviour has remained the same.

## Linear Ticket

[linear](https://linear.app/rudderstack/issue/PIPE-701/retl-integration-test) resolves PIPE-701

## Security

### DIFF
--- a/integration_test/retl_test/sut.go
+++ b/integration_test/retl_test/sut.go
@@ -57,19 +57,43 @@ type SUT struct {
 
 	URL string
 
-	WorkspaceID string
+	workspaceID string
 	Sources     []srcWithDst
 }
 
 type srcWithDst struct {
-	Source       src
-	Destinations []dst
+	source       src
+	destinations []dst
+}
+
+type batch struct {
+	Batch []record `json:"batch"`
+}
+
+type record struct {
+	Context recordContext `json:"context"`
+
+	Type      string    `json:"type"`
+	MessageID string    `json:"messageId"`
+	UserID    string    `json:"userId"`
+	SentAt    time.Time `json:"sentAt"`
+	Timestamp time.Time `json:"timestamp"`
+}
+
+type rudderSource struct {
+	JobID     string `json:"job_id"`
+	JobRunID  string `json:"job_run_id"`
+	TaskRunID string `json:"task_run_id"`
+}
+
+type recordContext struct {
+	Sources rudderSource `json:"sources"`
 }
 
 func Connect(s src, d ...dst) srcWithDst {
 	return srcWithDst{
-		Source:       s,
-		Destinations: d,
+		source:       s,
+		destinations: d,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


integration_test/retl_test/sut_test.go
```diff
@@ -0,0 +1,359 @@
...
+	var (
+		httpClient = &http.Client{}
+		url        = fmt.Sprintf("%s/v1/job-status/%s?%s", s.URL, jobRunID, url.Values{
+			"task_run_id": []string{jobTaskID},
+		}.Encode()) // job_run_id
+	)
+
+	var status rsources.JobStatus
+
+	res, err := httpClient.Get(url)
```
achettyiitr, line 326: Can we use default client here?
```suggestion
	res, err := http.Get(url)
```

integration_test/retl_test/sut_test.go
```diff
@@ -0,0 +1,359 @@
...
+		t.Fail()
+		return
+	}
+
+	t.Logf("Event Sent Successfully: (%s)", body)
+}
+
+// JobStatus hits the job-status endpoint and returns the status for a sourceID, jobRunID, jobTaskID.
+// If the job is not found, the second return value is false. Any other error is logged and the test is failed.
+func (s *SUT) JobStatus(t *testing.T, sourceID, jobRunID, jobTaskID string) (rsources.JobStatus, bool) {
```
achettyiitr, line 316: sourceID not being used?
```suggestion
func (s *SUT) JobStatus(t *testing.T, jobRunID, jobTaskID string) (rsources.JobStatus, bool) {
```

integration_test/retl_test/sut_test.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	defer func() { httputil.CloseResponse(res) }()
+
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		t.Logf("sendEvent error: %v", err)
+		t.Fail()
+		return
+	}
+	if res.Status != "200 OK" {
```
achettyiitr, line 305: Why not check the status code?
```suggestion
	if res.StatusCode != http.StatusOK {
```

achettyiitr (approved):

integration_test/retl_test/sut.go
```diff
@@ -0,0 +1,382 @@
...
+	pool, err := dockertest.NewPool("")
+	require.NoError(t, err)
+
+	containersGroup, _ := errgroup.WithContext(context.TODO())
+
+	var postgresContainer *resource.PostgresResource
+	var transformerContainer *destination.TransformerResource
+
+	containersGroup.Go(func() (err error) {
+		postgresContainer, err = resource.SetupPostgres(pool, t)
```
fracasula, line 126: I think you need to change these, it now is `postgres.Setup(pool, t)`.

fracasula (approved):

integration_test/retl_test/sut.go
fracasula: I pushed the changes @lvrach

PR #4327, opened 2024-01-17
fix: jobs not draining as fast as we'd like because of rate-limiting etc.

Sidddddarth:
# Description

Drain router jobs faster. Done by sending the job to worker in case of backoff or throttling whereas it was discarded earlier.

## Linear Ticket

[slack thread](https://rudderlabs.slack.com/archives/C02026YSHQ9/p1705411872414129?thread_ts=1705297122.598099&cid=C02026YSHQ9)
[Resolves PIPE-713](https://linear.app/rudderstack/issue/PIPE-713/drain-to-avoid-rate-limiting)

## Security

### DIFF
--- a/router/handle.go
+++ b/router/handle.go
@@ -537,23 +537,29 @@ func (rt *Handle) findWorkerSlot(ctx context.Context, workers []*worker, job *jo
 			"destinationId": parameters.DestinationID,
 		}).Increment()
 	}
+	abortedJob := rt.drainOrRetryLimitReached(job) // if job's aborted, then send it to it's worker right away
 	if eventOrderingDisabled {
 		availableWorkers := lo.Filter(workers, func(w *worker, _ int) bool { return w.AvailableSlots() > 0 })
 		if len(availableWorkers) == 0 {
 			return nil, types.ErrWorkerNoSlot
 		}
+		slot := availableWorkers[rand.Intn(len(availableWorkers))].ReserveSlot()
+		if slot == nil {
+			return nil, types.ErrWorkerNoSlot
+		}
+		if abortedJob {
+			return slot, nil
+		}
 		if rt.shouldBackoff(job) {
+			slot.Release()
 			return nil, types.ErrJobBackoff
 		}
 		if rt.shouldThrottle(ctx, job, parameters) {
+			slot.Release()
 			return nil, types.ErrDestinationThrottled
 		}
 
-		if slot := availableWorkers[rand.Intn(len(availableWorkers))].ReserveSlot(); slot != nil { // skipcq: GSC-G404
-			return slot, nil
-		}
-		return nil, types.ErrWorkerNoSlot
-
+		return slot, nil
 	}
 
 	// checking if the orderKey is in blockedOrderKeys. If yes, returning nil.
@@ -564,7 +570,7 @@ func (rt *Handle) findWorkerSlot(ctx context.Context, workers []*worker, job *jo
 	}
 
 	//#JobOrder (see other #JobOrder comment)
-	if rt.shouldBackoff(job) { // backoff
+	if rt.shouldBackoff(job) && !abortedJob { // backoff
 		blockedOrderKeys[orderKey] = struct{}{}
 		return nil, types.ErrJobBackoff
 	}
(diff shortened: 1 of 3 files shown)


router/handle.go
```diff
@@ -579,7 +585,7 @@ func (rt *Handle) findWorkerSlot(ctx context.Context, workers []*worker, job *jo
 		return nil, types.ErrBarrierExists
 	}
 	rt.logger.Debugf("EventOrder: job %d of orderKey %s is allowed to be processed", job.JobID, orderKey)
-	if rt.shouldThrottle(ctx, job, parameters) {
+	if rt.shouldThrottle(ctx, job, parameters) && !abortedJob {
```
fracasula, line 588: Can't we check if we should throttle before entering the barrier like we do for backoff?

Something like:
1. if shouldBackoff AND !abortedJob THEN return ErrJobBackoff
2. if shouldThrottle AND !abortedJob THEN return ErrDestinationThrottled
3. get worker slot
4. enter barrier THEN return (ErrBarrierExists OR slot)

Wdyt?

router/handle.go
```diff
@@ -589,6 +595,45 @@ func (rt *Handle) findWorkerSlot(ctx context.Context, workers []*worker, job *jo
 	//#EndJobOrder
 }
 
+// checks if job is configured to drain or if it's retry limit is reached
+func (rt *Handle) drainOrRetryLimitReached(job *jobsdb.JobT) bool {
+	drain, _ := rt.drainer.Drain(job)
```
fracasula, line 600: [optional]  Would it make sense to get the drain reason and create a metric out of it?

router/router_test.go
```diff
@@ -152,6 +152,12 @@ func initRouter() {
 	misc.Init()
 }
 
+type drainer struct{}
+
+func (d *drainer) Drain(job *jobsdb.JobT) (bool, string) {
+	return false, ""
+}
```
fracasula, line 159: Can we have a test where we have throttling enabled and jobs that should be drained that makes sure we drain right away? Making sure that `throttler.CheckLimitReached` isn't called could be a strategy.

```suggestion
type drainer struct {
	drain bool
}

func (d *drainer) Drain(job *jobsdb.JobT) (bool, string) {
	return d.drain, ""
}
```

router/handle.go
Sidddddarth: this is being done by the `worker`.
Changed things up a little so that this check is only done here and not again in the worker.
Consequently, changed the error message in case a job is drained as well: `routerutils.EnhanceJSON(job.LastJobStatus.ErrorResponse, "reason", abortReason)`

router/handle.go
Sidddddarth: I'm not sure if this order(`barrier.Enter` and then checking if `throttled`) is being followed for a reason(can't think of one right now).
@atzoum wdyt?

router/handle.go
```diff
@@ -537,23 +543,29 @@ func (rt *Handle) findWorkerSlot(ctx context.Context, workers []*worker, job *jo
 			"destinationId": parameters.DestinationID,
 		}).Increment()
 	}
+	abortedJob, abortReason := rt.drainOrRetryLimitReached(job) // if job's aborted, then send it to it's worker right away
```
fracasula, line 546:
```suggestion
	abortedJob, abortReason := rt.drainOrRetryLimitReached(job) // if job's aborted, then send it to its worker right away
```

router/handle.go
```diff
@@ -586,16 +598,59 @@ func (rt *Handle) findWorkerSlot(ctx context.Context, workers []*worker, job *jo
...
+
+func (rt *Handle) retryLimitReached(status *jobsdb.JobStatusT) bool {
+	respStatusCode, _ := strconv.Atoi(status.ErrorCode)
+	switch respStatusCode {
+	case types.RouterTimedOutStatusCode,
+		types.RouterUnMarshalErrorCode: // 5xx errors
+		return false
+	}
+
+	if respStatusCode < 500 {
```
fracasula, line 632:
```suggestion
	if respStatusCode < http.StatusInternalServerError {
```

fracasula (approved):

mihir20 (approved):
LGTM

cisse21 (approved):

PR #4328, opened 2024-01-18
fix: nil load statistics for bigquery

achettyiitr:
# Description

- In case of rate limit errors, we are sending empty statistics and then trying to access a Load field which will be nil.

## Linear Ticket

- Resolves PIPE-704

## Security

### DIFF
--- a/warehouse/integrations/bigquery/bigquery.go
+++ b/warehouse/integrations/bigquery/bigquery.go
@@ -439,8 +439,9 @@ func (bq *BigQuery) loadTableByAppend(
 
 	log.Infow("completed loading")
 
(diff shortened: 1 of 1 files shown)


cisse21 (approved):

mihir20 (approved):
LGTM

fracasula (commented):
@achettyiitr can you create a follow up task to have this client mocked so that we can cover these cases as well?

achettyiitr:
> @achettyiitr can you create a follow up task to have this client mocked so that we can cover these cases as well?

Tracking it [here](https://linear.app/rudderstack/issue/PIPE-709/get-job-statistics-via-a-dependency-for-bq).

PR #4332, opened 2024-01-18
feat: adding sourceId and destinationId in pipeline info metrics

gane5hvarma:
# Description
Rudder server emits a metric called pipeline_processed_events which has labels state and code. This metric is useful to understand the different status codes when rudder server is processing events to destination. Rudder sources team want to use this info to show it to end users. So that they can know how is the sync performing. For this to achieve we want to add sourceId and destinationId as extra labels to pipeline_processed_events metric.

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-299/rudder-server-add-sourceid-and-destinationid-in-pipeline-processed

## Security


router/batchrouter/handle.go
```diff
@@ -576,6 +576,7 @@ func (brt *Handle) updateJobStatus(batchJobs *BatchedJobs, isWarehouse bool, err
 	transformedAtMap := make(map[string]string)
 	statusDetailsMap := make(map[string]*types.StatusDetail)
 	jobStateCounts := make(map[string]int)
+	jobIdConnectionDetailsMap := make(map[int64]*jobsdb.ConnectionDetailsT)
```
lvrach, line 579: What is the reason for using a pointer to a struct? Fro small struct is more efficient to pass them as is.

also (nitpik) `jobID` instead of `jobId` https://google.github.io/styleguide/go/decisions.html#initialisms

jobsdb/jobsdb.go
```diff
@@ -365,6 +365,11 @@ type JobStatusT struct {
 	WorkspaceId   string          `json:"WorkspaceId"`
 }
 
+type ConnectionDetailsT struct {
```
lvrach, line 368: We no longer use `T` suffix for structs 

```suggestion
type ConnectionDetails struct {
```

router/batchrouter/handle.go
lvrach: I understand the reason we use the maps, but is a hack that we need to address some day

gane5hvarma: I will change it for sourceId and destinationId as well

gane5hvarma:
@lvrach updateProcessedEventsMetrics code for both rt and brt is almost same except module and destType differs. Should i change this function to a common function where rt and brt call them with extra parameters instead of a method function?

Sidddddarth:
> @lvrach updateProcessedEventsMetrics code for both rt and brt is almost same except module and destType differs. Should i change this function to a common function where rt and brt call them with extra parameters instead of a method function?

Sounds good. Maybe something like this..?
```

type JobStatusStats struct {
	Module        string
	DestType      string
	SourceID      string
	DestinationID string
	JobState      string
	StatusCode    string
}

func UpdateProcessedEventsMetrics(statHandle stats.Stats, statsMap map[*JobStatusStats]int) {
	if statHandle == nil {
		statHandle = stats.Default
	}
	for stat, count := range statsMap {
		statHandle.NewTaggedStat(`pipeline_processed_events`, stats.CountType, stats.Tags{
			"module":        stat.Module,
			"destType":      stat.DestType,
			"state":         stat.JobState,
			"code":          stat.StatusCode,
			"sourceId":      stat.SourceID,
			"destinationId": stat.DestinationID,
		}).Count(count)
	}
}
```
You can use `router/utils/utils.go`. It contains other utilities used in both router, batch router.

router/handle.go
```diff
@@ -454,7 +464,7 @@ func (rt *Handle) commitStatusList(workerJobStatuses *[]workerJobStatus) {
 		if err != nil {
 			panic(err)
 		}
-		rt.updateProcessedEventsMetrics(statusList)
+		routerutils.UpdateProcessedEventsMetrics(stats.Default, "router", rt.destType, statusList, jobIDConnectionDetailsMap)
```
gane5hvarma, line 467: should change router string to a const

router/batchrouter/handle.go
```diff
@@ -763,7 +768,7 @@ func (brt *Handle) updateJobStatus(batchJobs *BatchedJobs, isWarehouse bool, err
 	if err != nil {
 		panic(err)
 	}
-	brt.updateProcessedEventsMetrics(statusList)
+	routerutils.UpdateProcessedEventsMetrics(stats.Default, "batch_router", brt.destType, statusList, jobIDConnectionDetailsMap)
```
gane5hvarma, line 771: todo: change batch_router string to a const

router/handle.go
```diff
@@ -314,6 +316,11 @@ func (rt *Handle) stopIteration(err error) bool {
 	return rt.isolationStrategy.StopIteration(err)
 }
 
+type connectionDetails struct {
+	sourceId      string
+	destinationId string
+}
+
```
cisse21:
```suggestion
```

cisse21 (approved):

PR #4352, opened 2024-01-25
feat: add 'config' field to proxy request for v0

Gauravudia:
# Description

Resolves INT-1407
Update the transformer <> server proxy contract to support including 'config' at the root level in the proxy request from the server to the transformer. With these changes, the server will now append an additional 'config' field to the proxy request, propagating it to the transformer for utilization. 
In this PR enabling it for V0. For V1 it was already enabled with this [PR](https://github.com/rudderlabs/rudder-server/pull/4264).

## Linear Ticket

https://linear.app/rudderstack/issue/INT-1407/enable-destination-config-to-v0-proxy-endpoint

## Security


router/transformer/transformer_proxy_adapter.go
```diff
@@ -25,7 +25,8 @@ type transformerProxyAdapter interface {
 
 type ProxyRequestPayloadV0 struct {
 	integrations.PostParametersT
-	Metadata ProxyRequestMetadata `json:"metadata"`
+	Metadata          ProxyRequestMetadata   `json:"metadata"`
+	DestinationConfig map[string]interface{} `json:"destinationConfig"`
```
chandumlg, line 29: This doesn't seem like a breaking change. So, no big deal. But ideally, we should be using v1 and deprecate v0 soon.

chandumlg (approved):

koladilip (approved):

PR #4355, opened 2024-01-27
fix: handle consent management configuration fallback for gcm

saikumarrs:
# Description

Instead of using `event.context.consentManagement.provider` as the differentiator b/w legacy and generic consent management, I've updated the logic prefer GCM config first, then fallback to legacy consents config.

This would ensure that the latest v3 JS SDK would be fully compatible with the server and enable us to switch our beta (GCM) users b/w legacy and generic consent management with the help of a feature flag in webapp.

Still, the users will have to set the consents data appropriately in the UI for either of the config fields.

## Linear Ticket

https://linear.app/rudderstack/issue/SDK-998/make-legacy-consent-configuration-fields-source-specific

## Security

### DIFF
--- a/processor/consent.go
+++ b/processor/consent.go
@@ -50,43 +50,42 @@ func (proc *Handle) getConsentFilteredDestinations(event types.SingularEventT, d
 	}
 
 	return lo.Filter(destinations, func(dest backendconfig.DestinationT, _ int) bool {
-		// This field differentiates legacy and generic consent management
-		if consentManagementInfo.Provider != "" {
-			// Generic consent management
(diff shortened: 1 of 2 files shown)


bardisg (approved):
I assume this is to be meged only after GCM becomes available to all customers and destinations

saikumarrs:
> I assume this is to be merged only after GCM becomes available to all customers and destinations

@bardisg , this would be needed for our phase 1 release as well, if our customers want to go back to using legacy consent management.

achettyiitr (approved):

PR #4425, opened 2024-02-27
fix: handling invalid client id error for bingads audience

shrouti1507:
# Description

we are encountering issue with bingAds bulk API returning client ID as empty string. 
We are handling this by checking the status file row name and eventually aborting all the events in importing state

## Linear Ticket

Resolves INT-1682

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/bing-ads/bulk_uploader.go
+++ b/router/batchrouter/asyncdestinationmanager/bing-ads/bulk_uploader.go
@@ -233,6 +233,7 @@ func (b *BingAdsBulkUploader) GetUploadStats(uploadStatsInput common.GetUploadSt
 	// making an array of those jobIds
 	importList := uploadStatsInput.ImportingList
 	var initialEventList []int64
+	var finalFailedEventList []int64
 	for _, job := range importList {
 		initialEventList = append(initialEventList, job.JobID)
 	}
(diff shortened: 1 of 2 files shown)


router/batchrouter/asyncdestinationmanager/bing-ads/util.go
```diff
@@ -111,6 +111,7 @@ func (b *BingAdsBulkUploader) populateZipFile(actionFile *ActionFileInfo, audien
 		actionFile.EventCount += 1
 		for _, uploadData := range data.Message.List {
 			clientIdI := newClientID(data.Metadata.JobID, uploadData.HashedEmail)
+			fmt.Println("######## clientIdI ###########", clientIdI)
```
cisse21, line 114: use logger instead ?

shrouti1507: removed

router/batchrouter/asyncdestinationmanager/bing-ads/bulk_uploader.go
```diff
@@ -262,10 +263,26 @@ func (b *BingAdsBulkUploader) GetUploadStats(uploadStatsInput common.GetUploadSt
 		eventStatsResponse.StatusCode = response.StatusCode
 	}
 
+	// If the failedJobIds array has only one element and that is 0, that means that all the jobs have failed
```
shrouti1507, line 266: this will be refactored

shrouti1507:
We are closing this PR because we are handling this issue in a simpler way by just looking for list item error named rows. This will allow us to get rid of the problem with minimalistic change. cc @koladilip 
Raised PR : https://github.com/rudderlabs/rudder-server/pull/4429

PR #4429, opened 2024-02-28
fix: always look for list item error for bingads audience 

shrouti1507:
# Description
we are encountering issue with bingAds bulk API returning client ID as empty string. This happening because we are just checking the row name with string "error", which will take custom list error as well that is not supposed to have a client id and will break with 500 error code. 
For now we are checking specifically for list item error rows, where the client id presence is expected . Also added test case around that. 
## Linear Ticket

Resolves INT-1682

## Security


router/batchrouter/asyncdestinationmanager/bing-ads/util.go
```diff
@@ -312,14 +312,14 @@ func (b *BingAdsBulkUploader) readPollResults(filePath string) ([][]string, erro
...
 		// remove the file after the response has been written
-		removeErr := os.Remove(filePath)
-		if removeErr != nil {
-			b.logger.Errorf("Error removing the CSV file: %w", removeErr)
-			if err == nil {
-				err = removeErr
-			}
-
-		}
+		// removeErr := os.Remove(filePath)
```
koladilip, line 315: we need to uncomment this code right?

shrouti1507: yes did

router/batchrouter/asyncdestinationmanager/bing-ads/testdata/status-wrong-audience.zip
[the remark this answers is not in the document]
shrouti1507: done

koladilip (approved):

shrouti1507:
new PR is raised based on hotfix release branch https://github.com/rudderlabs/rudder-server/pull/4430

PR #4457, opened 2024-03-07
fix: terminate goroutines if reporting mainLoop panics

itsmihir:
# Description

when `getReports()` panics, the go-routine that is emitting the lag metrics is not terminated because we were doing `wg.Done()`. Using errorGroup to handle this.

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-371/fix-reporting-module-should-panic-if-querying-to-postgres-fails
resolves OBS-371

## Security


achettyiitr (approved):

Sidddddarth:
A few questions:
if the second routine panics, are we sure the first one would exit?
When the second routine panics, the whole system halts so why would do we care about the first routine now?

A thought here:
Maybe we could return the error instead of panic, this would guarantee the errGroup's context is cancelled and accomplish the purpose stated above. And propagate the error up too, that is have the `mainLoop` return the error and then:
```
func (r *DefaultReporter) DatabaseSyncer(c types.SyncerConfig) types.ReportingSyncer {
....
    return func() {
		r.g.Go(func() error {
			return r.mainLoop(r.ctx, c)
		})
	}
}
```

itsmihir:
@Sidddddarth 
> if the second routine panics, are we sure the first one would exit?

Yes

> When the second routine panics, the whole system halts so why would do we care about the first routine now?

Previously we were doing `wg.wait()` and because of this after the second routine panic it was waiting on the first routine to finish (which will never finish as ctx was never cancelled), so the system kept running.


If we return an error then the system will keep on running. Not sure why we had panic in the first place and is it expected to halt the system in case the database is not reachable?

enterprise/reporting/reporting.go
```diff
@@ -361,126 +377,116 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
...
+
+			getReportsStart := time.Now()
+			reports, reportedAt, err := r.getReports(currentMin, c.ConnInfo)
+			getReportsTimer.Since(getReportsStart)
+			getReportsCount.Observe(float64(len(reports)))
+			if len(reports) == 0 {
+				if err == nil {
+					lastReportedAtTime.Store(loopStart)
+				} else {
+					r.log.Errorw("getting reports", "error", err)
```
lvrach, line 408: I am noticing this as an anti-pattern. 

the `err` value should be checked immediately and then the returns values should be checked. 

It is a convention that if err is not nil, other values should not be populated, thus reports will be empty. However, not checking the error hide a bug, where getReports returns a partially filled array with an error. In that case, error will not be reported.

enterprise/reporting/reporting.go
```diff
@@ -361,126 +377,116 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
...
+				}
+				select {
+				case <-ctx.Done():
+					r.log.Infof("stopping mainLoop for syncer %s : %s", c.Label, ctx.Err())
```
lvrach, line 412: I notice an inconsistent behaviour here:
1. in line 394 https://github.com/rudderlabs/rudder-server/pull/4457/files#diff-5639a722fc70b6a04ec6daf63094d6bb127454dd8587e1b3466785305e515c3eR392 , when context is Err, you return an error
2. here `<-ctx.Done()` you return nil. 

ctx.Err() and ctx.Done() are very similar in how they work, except done is blocking until context gets cancels and .Err() returns a not nil error when context gets cancel.

Also to avoid spamming, I would avoid logging an error and returning it, as well. It would be best if you decided how your component should behave and be consistent with it.

enterprise/reporting/reporting.go
```diff
@@ -361,126 +377,116 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
...
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(r.mainLoopSleepInterval.Load()):
+			}
 		}
-	}
+	})
+
+	_ = g.Wait()
```
lvrach, line 489: It is not a good idea to mute the error here

enterprise/reporting/reporting.go
```diff
@@ -361,126 +377,116 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
...
-					logger.NewErrorField(err),
-				)
-			}
-			if sizeEstimate > config.GetInt64("Reporting.vacuumThresholdBytes", 5*bytesize.GB) {
-				if _, err := dbHandle.ExecContext(ctx, `vacuum full analyze reports;`); err != nil {
+			err = errGroup.Wait()
+			if err == nil {
+				dbHandle, err := r.getDBHandle(c.ConnInfo)
+				if err != nil {
+					panic(err)
```
lvrach, line 451: can we prevent the panic here, and instead propage the error?

enterprise/reporting/reporting.go
```diff
@@ -361,126 +377,116 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
...
-			).Scan(&sizeEstimate); err != nil {
-				r.log.Errorn(
-					`[ Reporting ]: Error getting table size estimate`,
-					logger.NewErrorField(err),
-				)
-			}
-			if sizeEstimate > config.GetInt64("Reporting.vacuumThresholdBytes", 5*bytesize.GB) {
-				if _, err := dbHandle.ExecContext(ctx, `vacuum full analyze reports;`); err != nil {
+			err = errGroup.Wait()
+			if err == nil {
```
lvrach, line 448: wouldn't be more readable if you handled the err case first?

```suggestion
			err = errGroup.Wait()
			if err != nil {
```
hard to read the code in github diff, but I understand we do nothing if there is an error. Is this correct? Maybe we should log it ?

lvrach (commented):
I understand the problem we try to solve, but I feel that the existing code would benefit from some overall restructuring and improvement.

We should not rely on `panic`  stopping the whole system, but rather proper error propagation.

I understand the complexity of proper error propagation, but it can be achieved in iterative steps.  You can start by moving the panic when error-group returns.
https://github.com/rudderlabs/rudder-server/pull/4457/files#diff-5639a722fc70b6a04ec6daf63094d6bb127454dd8587e1b3466785305e515c3eR489

Then we can figure out further improvements.
 
Also, there is no test associated with this fix. It is not the easiest test to write, but it is an important flow for the system and it is currently untested.

I don't want to block the PR by requesting changes, but I would like to see if there is an opportunity to improve things here.

enterprise/reporting/reporting.go
itsmihir: Yes, we should log it. 

By the way, you can change the GitHub diff settings to `hide whitespace`, changes are much more clear that way

Sidddddarth (approved):

PR #4467, opened 2024-03-11
feat: add support to build arm64 image

mihir20:
# Description

Adding support to build arm64 image in GA. 

We will create both arm64 as well as amd64 docker image for rudder-server. A manifest will be created that will add both the builds to a common tag so that based on the arch of host machine correct build will be picked-up

Successful Build: https://github.com/rudderlabs/rudder-server/actions/runs/8537621984/job/23388772427

image

image

image

image

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-889/arm-build-for-rudder-server

## Security

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -280,7 +280,16 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Create multi-arch manifest for Docker OSS
         run: |
-          docker buildx imagetools create -t ${{ needs.docker-oss-meta.outputs.tags }} ${{ needs.docker-oss-meta.outputs.arm64_tags }} ${{ needs.docker-oss-meta.outputs.amd64_tags }}
+          while read -r tag; do
+            echo "$tag"
+
+            arm_tag=$(echo "${{ needs.docker-oss-meta.outputs.arm64_tags }}" | grep "$tag")
+
+            amd_tag=$(echo "${{ needs.docker-oss-meta.outputs.amd64_tags }}" | grep "$tag")
+
+            docker buildx imagetools create -t $tag $arm_tag $amd_tag
+
+          done <<< "${{ needs.docker-oss-meta.outputs.tags }}"
   create-manifest-docker-ent:
     runs-on: ubuntu-latest
     needs: [ docker-ent, docker-ent-meta ]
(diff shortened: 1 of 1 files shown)


cisse21 (approved):

.github/workflows/builds.yml
```diff
@@ -134,4 +238,60 @@ jobs:
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3.0.0
+      - name: Login to DockerHub
+        uses: docker/login-action@v2.1.0
+        with:
+          username: rudderlabs
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Create multi-arch manifest for enterprise
+        run: |
+          docker buildx imagetools create -t ${{ needs.docker-ent.outputs.tags }} ${{ needs.docker-ent-arm64.outputs.tags }} ${{ needs.docker-ent.outputs.tags }}-amd64
```
lvrach, line 297: Could it be possible that `needs.docker-ent.outputs.tags` contains multiple tags and `-amd64` only get applied to the last one?

would it be more appropriate to use suffix https://github.com/docker/metadata-action?tab=readme-ov-file#flavor-input ?

```suggestion
          docker buildx imagetools create -t ${{ needs.docker-ent.outputs.tags }} ${{ needs.docker-ent-arm64.outputs.tags }} ${{ needs.docker-ent.outputs.tags }}
```

.github/workflows/builds.yml
```diff
@@ -134,4 +238,60 @@ jobs:
             BUILD_DATE=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.created'] }}
             VERSION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
             COMMIT_HASH=${{ github.sha }}
-            REVISION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.revision'] }}
\ No newline at end of file
+            REVISION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.revision'] }}
+  docker-suppression-backup-service-arm64:
```
lvrach, line 242: would it be possible to use a matrix https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy to avoid the repetition for arm builds ?

lvrach (approved):

PR #4468, opened 2024-03-11
chore: vacuum reports table when size > 5GB

Sidddddarth:
# Description

vacuum reports table when size estimate goes over 5GB(configurable).

## Linear Ticket

[slack thread](https://rudderlabs.slack.com/archives/C02GW9B6K08/p1709952064963919)

## Security

### DIFF
--- a/enterprise/reporting/reporting.go
+++ b/enterprise/reporting/reporting.go
@@ -21,6 +21,7 @@ import (
 	"github.com/lib/pq"
 	"github.com/samber/lo"
 
+	"github.com/rudderlabs/rudder-go-kit/bytesize"
 	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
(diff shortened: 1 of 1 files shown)


achettyiitr (approved):

enterprise/reporting/reporting.go
```diff
@@ -446,6 +451,27 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
...
+			}
+			if sizeEstimate > config.GetInt64("Reporting.vacuumThresholdBytes", 5*bytesize.GB) {
+				if _, err := dbHandle.ExecContext(ctx, `vacuum full analyze reports;`); err != nil {
+					r.log.Errorn(
+						`[ Reporting ]: Error vacuuming reports table`,
+						logger.NewErrorField(err),
+					)
+				}
+			}
+			vacuumDuration.Since(vacuumStart)
```
achettyiitr, line 474: Should we send these only in case we are doing the vacuum? Basically from the above if condition?

cisse21 (approved):

PR #4471, opened 2024-03-12
chore: fix databricks merge query

cisse21:
# Description
For Databricks the way we have been doing merge was running the merge query on the last 2 days data. This PR attempts to run merge on all the old data and not based on the partitioned data

## Linear Ticket
Fixes PIPE-895

## Security

### DIFF
--- a/warehouse/integrations/deltalake/deltalake.go
+++ b/warehouse/integrations/deltalake/deltalake.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
-	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


warehouse/integrations/deltalake/deltalake.go
```diff
@@ -1168,13 +1107,6 @@ func (d *Deltalake) LoadUserTables(ctx context.Context) map[string]error {
 			columnNames(columnKeys),
 		)
 	} else {
-		if partitionQuery, err = d.partitionQuery(ctx, warehouseutils.UsersTable); err != nil {
```
lvrach, line 1171: [non-blocking]
I am thinking instead of deleting this, we can add it behind a rudder-server config. Use case, dedup on 30-days window.

lvrach (approved):

fracasula (approved):

PR #4477, opened 2024-03-14
fix: always look for list item error for bingads audience (#4430)

cisse21:
# Description

Adding the hotfix from 1.20.x to 1.21.x

## Security


lvrach (approved):

mihir20 (approved):
Should we also add this change in master?

cisse21:
> Should we also add this change in master?

Will backmerge this to master

PR #4492, opened 2024-03-18
fix: too many connections in router transform

Sidddddarth:
# Description

On failures in router transform, there's 30 retries - each retry reading the body but not closing it.
1. Closing the body in case of failures before retrying again.
2. Updated some log statements to use `Errorn` instead of `Error`

## Linear Ticket

[resolves PIPE-846](https://linear.app/rudderstack/issue/PIPE-846/hostedmtedmt-v0-rs-1-is-opening-too-many-connections-to-transformer)

## Security

### DIFF
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -152,17 +152,27 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
 		if err != nil {
 			trans.transformRequestTimerStat.SendTiming(time.Since(s))
 			reqFailed = true
-			trans.logger.Errorf("JS HTTP connection error: URL: %v Error: %+v", url, err)
+			trans.logger.Errorn(
+				"JS HTTP connection error",
+				logger.NewErrorField(err),
+				logger.NewStringField("URL", url),
+			)
 			if retryCount > config.GetInt("Processor.maxRetry", 30) {
 				panic(fmt.Errorf("JS HTTP connection error: URL: %v Error: %+v", url, err))
 			}
 			retryCount++
 			time.Sleep(config.GetDurationVar(100, time.Millisecond, "Processor.retrySleep", "Processor.retrySleepInMS"))
 			// Refresh the connection
+			httputil.CloseResponse(resp)
 			continue
 		}
+
 		if reqFailed {
-			trans.logger.Errorf("Failed request succeeded after %v retries, URL: %v", retryCount, url)
+			trans.logger.Errorn(
+				"Failed request succeeded",
+				logger.NewStringField("URL", url),
+				logger.NewIntField("RetryCount", int64(retryCount)),
+			)
 		}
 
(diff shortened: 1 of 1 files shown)


router/transformer/transformer.go
```diff
@@ -152,17 +152,27 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
 		if err != nil {
 			trans.transformRequestTimerStat.SendTiming(time.Since(s))
 			reqFailed = true
-			trans.logger.Errorf("JS HTTP connection error: URL: %v Error: %+v", url, err)
+			trans.logger.Errorn(
+				"JS HTTP connection error",
```
mihir20, line 156:
```suggestion
				"failed to connect with transformer",
```

router/transformer/transformer.go
```diff
@@ -152,17 +152,27 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
...
 			time.Sleep(config.GetDurationVar(100, time.Millisecond, "Processor.retrySleep", "Processor.retrySleepInMS"))
 			// Refresh the connection
+			httputil.CloseResponse(resp)
 			continue
 		}
+
 		if reqFailed {
-			trans.logger.Errorf("Failed request succeeded after %v retries, URL: %v", retryCount, url)
+			trans.logger.Errorn(
+				"Failed request succeeded",
```
mihir20, line 172:
```suggestion
				"failed request got successful after retries",
```

router/transformer/transformer.go
```diff
@@ -152,17 +152,27 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
...
+				logger.NewErrorField(err),
+				logger.NewStringField("URL", url),
+			)
 			if retryCount > config.GetInt("Processor.maxRetry", 30) {
 				panic(fmt.Errorf("JS HTTP connection error: URL: %v Error: %+v", url, err))
 			}
 			retryCount++
 			time.Sleep(config.GetDurationVar(100, time.Millisecond, "Processor.retrySleep", "Processor.retrySleepInMS"))
 			// Refresh the connection
+			httputil.CloseResponse(resp)
```
mihir20, line 166: possible to move this logic out of for loop and create a function, we can then add a defer statement to close response just after opening reader for the response? explicitly closing response might result in miss in one place or other.

Sidddddarth: Yeah I agree.
In this case I went against that thought though so this would be easy(and quick) to review.
Can take up the suggested refactor at a later point.

lvrach (approved):
We should consider closing the error even if body is read without an error.

 https://andrii-kushch.medium.com/is-it-necessary-to-close-the-body-in-the-http-response-object-in-golang-171c44c9394d

BonapartePC (approved):

PR #4496, opened 2024-03-18
chore: version handler cleanup

achettyiitr:
# Description

- Cleanup for version info handler since these fields are not being set.

## Linear Ticket

- Resolves PIPE-834

## Security

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN go mod download
 COPY . .
 
 RUN BUILD_DATE=$(date "+%F,%T") \
-    LDFLAGS="-s -w -X main.version=${VERSION} -X main.commit=${COMMIT_HASH} -X main.buildDate=$BUILD_DATE -X main.builtBy=${REVISION} -X main.builtBy=${REVISION} -X main.enterpriseToken=${ENTERPRISE_TOKEN} " \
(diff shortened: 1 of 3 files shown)


lvrach:
It might be useful to expose the version in metrics or on an HTTP endpoint.

lvrach (approved):

achettyiitr:
> It might be useful to expose the version in metrics or on an HTTP endpoint.

Currently, we are:
1. Setting version in [stat options](https://github.com/rudderlabs/rudder-server/blob/2c469676b70415a598cf7909fcab3d0b05f7977c/runner/runner.go#L113).
2. Gause metrics ([rudder_server_config](https://github.com/rudderlabs/rudder-server/blob/2c469676b70415a598cf7909fcab3d0b05f7977c/runner/runner.go#L164)) which contains all the release info.
3. Version [endpoints](https://github.com/rudderlabs/rudder-server/blob/2c469676b70415a598cf7909fcab3d0b05f7977c/runner/runner.go#L342).

Do you mean we should expose at either metrics or on an HTTP endpoint?

mihir20 (commented):
Isn't there someway these could be populated in place of we removing these?

achettyiitr:
> Isn't there someway these could be populated in place of we removing these?

We already have a version, why do we need major, minor and patch? Also, not sure what the use of exposing `gitURL` in the version endpoint.

mihir20 (approved):
LGTM

PR #4505, opened 2024-03-20
fix: filter success events before sending it to error reporting table

mihir-4116:
# Description
- In this pr we are filtering successful events before sending it to errors table in reporting DB

## Security

- [ ✅] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

### DIFF
--- a/enterprise/reporting/error_reporting.go
+++ b/enterprise/reporting/error_reporting.go
@@ -173,9 +173,9 @@ func (edr *ErrorDetailReporter) Report(ctx context.Context, metrics []*types.PUR
 
 	reportedAt := time.Now().UTC().Unix() / 60
 	for _, metric := range metrics {
-		if (metric.StatusDetail.StatusCode < http.StatusBadRequest && !lo.Contains([]int{types.FilterEventCode, types.SuppressEventCode}, metric.StatusDetail.StatusCode)) {
-            continue
-        }
+		if metric.StatusDetail.StatusCode < http.StatusBadRequest && !lo.Contains([]int{types.FilterEventCode, types.SuppressEventCode}, metric.StatusDetail.StatusCode) {
+			continue
+		}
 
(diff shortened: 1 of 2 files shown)


enterprise/reporting/error_reporting.go
```diff
@@ -173,6 +173,10 @@ func (edr *ErrorDetailReporter) Report(ctx context.Context, metrics []*types.PUR
 
 	reportedAt := time.Now().UTC().Unix() / 60
 	for _, metric := range metrics {
+		if (metric.StatusDetail.StatusCode < http.StatusBadRequest && !lo.Contains([]int{types.FilterEventCode, types.SuppressEventCode}, metric.StatusDetail.StatusCode)) {
+            continue
+        }
```
achettyiitr, line 178: [optonal]: We can probably move it to a separate function as well.

```suggestion
		if !shouldReport(metric) {
			continue
		}
```

```GO
func shouldReport(metric *types.PUReportedMetric) bool {
	switch {
	case metric.StatusDetail.StatusCode >= http.StatusBadRequest, metric.StatusDetail.StatusCode == types.FilterEventCode, metric.StatusDetail.StatusCode == types.SuppressEventCode:
		return true
	default:
		return false
	}
}
```

achettyiitr (approved):

sanpj2292 (approved):

lvrach (commented):
Can we have a test for this?

Sidddddarth (approved):
LGTM
But why not filter while collecting reports..?

koladilip (approved):

sanpj2292:
> LGTM But why not filter while collecting reports..?

When `Report` method is called, the information is forwarded to several reporters like normal Reporting(metrics table), error reporting(errors table) & error_index_reporting from `mediator.go`.
This filteration makes sense only in error reporting(errors table) & may not required in others. Hence we are doing it in `error_reporting.go`

Let me know if it makes sense

PR #4511, opened 2024-03-21
chore: apart from using context to send data using header

ItsSudip:
# Description

We were using http.request context to send data. We are updating that with the header.
## Linear Ticket

< Replace with Linear Link >

## Security


ItsSudip:
we are not considering this approach

PR #4516, opened 2024-03-22
chore: increase max concurrent request to reporting in case of delay

itsmihir:
# Description

Doubling `maxConcurrentRequests` when reports delay exceeds 10 mins 

## Linear Ticket

[Resolves OBS-365](https://linear.app/rudderstack/issue/OBS-365/tune-concurrency-on-server-for-reporting)

## Security

### DIFF
--- a/enterprise/reporting/reporting.go
+++ b/enterprise/reporting/reporting.go
@@ -68,6 +68,7 @@ type DefaultReporter struct {
 	sourcesWithEventNameTrackingDisabled []string
 	maxOpenConnections                   int
 	maxConcurrentRequests                misc.ValueLoader[int]
+	maxConcurrentRequestsWithDelay       misc.ValueLoader[int]
 
 	getMinReportedAtQueryTime stats.Measurement
 	getReportsQueryTime       stats.Measurement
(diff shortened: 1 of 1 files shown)


enterprise/reporting/reporting.go
```diff
@@ -394,7 +397,13 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
...
 				return ctx.Err()
 			}
-			requestChan := make(chan struct{}, r.maxConcurrentRequests.Load())
+
+			concurrentRequests := r.maxConcurrentRequests.Load()
+			if time.Since(lastReportedAtTime.Load()).Minutes() > 10 {
+				concurrentRequests = r.maxConcurrentRequestsWithDelay.Load()
+			}
+
+			requestChan := make(chan struct{}, concurrentRequests)
```
lvrach, line 406: It would be useful, to introduce a metric here. I would suggest a gauge for `concurrentRequests`.

Is there a metric for `time.Since(lastReportedAtTime.Load())` ?

itsmihir: Yes we have a metrics for `time.Since(lastReportedAtTime.Load())`

PR #4517, opened 2024-03-24
chore: fetch pileup counts concurrently

Sidddddarth:
# Description

Fetch pileup counts concurrently.
Configurable limits(default 10 DSs at a time - `jobsdb.pileupcount.parallelism`)
Each jobsdb(`rt`, `batch_rt`) sequentially.

## Linear Ticket

[Resolves PIPE-924](https://linear.app/rudderstack/issue/PIPE-924/concurrent-pileup-count-fetch-at-startup)

## Security


jobsdb/jobsdb.go
```diff
@@ -1863,66 +1864,73 @@ func (jd *Handle) GetToProcess(ctx context.Context, params GetQueryParams, more
...
+	  )
+	  select
+		count(*),
+		customVal,
+		workspace
+	  from
+		joined
+	  group by
+		customVal,
+		workspace;`
```
fracasula, line 1901: I haven't checked but we should have an index in `job_state`, correct? Just trying to understand if we are leveraging indexes properly here. Mostly because before we were running these sequentially whilst now we're issuing X of these queries in parallel so I think it's even more important to make sure that indexing is helping here.

Sidddddarth:
```
QUERY PLAN
 --- 
 HashAggregate  (cost=19679.89..19679.94 rows=5 width=41) 
   Group Key: j.custom_val, j.workspace_id 
   ->  Hash Right Join  (cost=8774.07..19267.59 rows=54974 width=33) 
         Hash Cond: (rt_job_status_104819.job_id = j.job_id) 
         Filter: (((rt_job_status_104819.job_state)::text <> ALL ('{aborted,succeeded,migrated}'::text[])) OR (rt_job_status_104819.job_id IS NULL)) 
         ->  Unique  (cost=0.42..9795.26 rows=55339 width=189) 
               ->  Index Scan using idx_rt_job_status_104819_jid_id on rt_job_status_104819  (cost=0.42..9518.18 rows=110832 width=189) 
         ->  Hash  (cost=8076.07..8076.07 rows=55807 width=41) 
               ->  Seq Scan on rt_jobs_104819 j  (cost=0.00..8076.07 rows=55807 width=41) 

```
The last status view uses one index on the status table.
no index for `job_state` being used though, will dig deeper into this.

jobsdb/jobsdb.go
```diff
@@ -1863,66 +1864,73 @@ func (jd *Handle) GetToProcess(ctx context.Context, params GetQueryParams, more
...
-		if err != nil {
-			return nil, err
-		}
+	queryString := `with joined as (
+		select
+		  j.job_id as jobID,
+		  j.custom_val as customVal,
+		  s.id as statusID,
+		  s.job_state as jobState,
+		  j.workspace_id as workspace
```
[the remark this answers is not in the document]
Sidddddarth: Makes sense. Will update soon

jobsdb/jobsdb.go
Sidddddarth: Have given it some thought over the day. But what's in use looks like the best way to get what we want.
The same pattern is user everywhere(all routers and other consumers interested in jobs of different states).
Filtering by job_state **_after_** the join is by design as we don't want to query extra jobs.

And as for running them in parallel, I think it should be safe because of the same pattern being used by many routers at once.

jobsdb/jobsdb.go
```diff
@@ -1863,66 +1864,73 @@ func (jd *Handle) GetToProcess(ctx context.Context, params GetQueryParams, more
...
 
-		for rows.Next() {
-			var count sql.NullInt64
-			var customVal string
-			var workspace string
-			err := rows.Scan(&count, &customVal, &workspace)
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(jd.config.GetInt("jobsdb.pileupcount.parallelism", 10))
+	for _, ds := range dsList {
+		ds := ds
```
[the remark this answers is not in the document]
Sidddddarth: Noted
Will revisit this when we've upgraded to 1.22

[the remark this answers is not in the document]
lvrach: We have [upgraded](https://github.com/rudderlabs/rudder-server/commit/a705299f3c2c5f26ff2412634b20bff5bab2f2e7) to 1.22

If you pull latest master it should work

jobsdb/jobsdb.go
fracasula: OK cool, no problem. It's configurable anyway so worst case if something happens we'll set the concurrency to 1 and it'll work like before :+1:

jobsdb/jobsdb.go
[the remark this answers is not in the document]
Sidddddarth: so how does it work now?
I just remove this line..?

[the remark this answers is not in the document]
fracasula: Yes :smile:

jobsdb/jobsdb.go
```diff
@@ -1863,66 +1864,87 @@ func (jd *Handle) GetToProcess(ctx context.Context, params GetQueryParams, more
...
+	  from
+		joined
+	  group by
+		customVal,
+		workspace;`
+
+	g, ctx := errgroup.WithContext(ctx)
+	defaultConcurrency := 10
+	conc := jd.config.GetInt("jobsdb.pileupcount.parallelism", 10)
+	if conc <= 1 || conc > defaultConcurrency {
```
fracasula, line 1903: A concurrency of 1 should be accepted. It might be useful in case we want to turn off concurrency completely.

Sidddddarth: oh yes

lvrach (approved):

processor/processor.go
```diff
@@ -3175,20 +3175,19 @@ func (proc *Handle) countPendingEvents(ctx context.Context) error {
...
-		proc.IncreasePendingEvents(tablePrefix, pileUpStatMap)
-	}
-	return nil
+	return misc.RetryWithNotify(ctx,
+		jobdDBQueryRequestTimeout,
+		jobdDBMaxRetries.Load(),
+		func(ctx context.Context) error {
+			metric.Instance.Reset()
+			for tablePrefix, db := range dbs {
+				if err := db.GetPileUpCounts(ctx); err != nil {
```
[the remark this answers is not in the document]
Sidddddarth: Correct. 
Adopted.

BonapartePC (approved):

processor/processor.go
```diff
@@ -3175,20 +3175,23 @@ func (proc *Handle) countPendingEvents(ctx context.Context) error {
...
+					if err := db.GetPileUpCounts(ctx); err != nil {
+						return fmt.Errorf("pileup counts for %s: %w", tablePrefix, err)
+					}
+					return nil
+				})
+			}
+			return g.Wait()
+		}, func(attempt int) {
+			proc.logger.Warnf("Timeout during GetPileUpCounts, attempt %d", attempt)
+			stats.Default.NewTaggedStat("jobsdb_query_timeout", stats.CountType, stats.Tags{"attempt": fmt.Sprint(attempt), "module": "pileup"}).Increment()
```
fracasula, line 3195: strconv is way faster than fmt.Sprint

fracasula (approved):

processor/processor.go
```diff
@@ -3192,6 +3192,6 @@ func (proc *Handle) countPendingEvents(ctx context.Context) error {
 			return g.Wait()
 		}, func(attempt int) {
 			proc.logger.Warnf("Timeout during GetPileUpCounts, attempt %d", attempt)
-			stats.Default.NewTaggedStat("jobsdb_query_timeout", stats.CountType, stats.Tags{"attempt": fmt.Sprint(attempt), "module": "pileup"}).Increment()
+			stats.Default.NewTaggedStat("jobsdb_query_timeout", stats.CountType, stats.Tags{"attempt": strconv.Itoa(attempt), "module": "pileup"}).Increment()
```
fracasula, line 3195: :heart:

PR #4519, opened 2024-03-25
fix: robust dangling table deletion during recovery

lvrach:
# Description

Recently we had three incidents where redshift crash recovery kept on failing. We select tables matching a prefix and then attempt to delete them. Due to race, the table has already been deleted by another process. Although we have concurrency controls to prevent multiple recoveries from taking place for the same destination, but another destination configured in the same namespace could likely cause this. 

Thus, introducing `IF EXIST` to ensure that deletes failures won't cause sync to fail.

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-926/robust-dangling-tables-deletion

## Security

### DIFF
--- a/warehouse/integrations/redshift/redshift_test.go
+++ b/warehouse/integrations/redshift/redshift_test.go
@@ -1001,12 +1001,11 @@ func TestIntegration(t *testing.T) {
 		})
 
 		t.Run("crashRecover", func(t *testing.T) {
-
 			ctx := context.Background()
 			tableName := "crash_recovery_test_table"
 			stgTableName := warehouseutils.StagingTableName(warehouseutils.RS, tableName, 64)
 			mockUploader := newMockUploader(t, nil, tableName, schemaInUpload, schemaInUpload, warehouseutils.LoadFileTypeParquet)
-		
+
 			rs := redshift.New(config.New(), logger.NOP, stats.NOP)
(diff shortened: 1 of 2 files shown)


achettyiitr (approved):

warehouse/integrations/redshift/redshift_test.go
```diff
@@ -999,6 +999,45 @@ func TestIntegration(t *testing.T) {
...
+					namespace,
+					stgTableName,
+				).Scan(&count)
+				require.NoError(t, err)
+
+				return count == 1
+			}
+			require.True(t, tableExists(t))
+
+			err = rs.CrashRecover(ctx)
```
achettyiitr, line 1036: Should we run this? twice to see, if it working as expected, Basically to test for `IF EXISTS`?

cisse21 (approved):

warehouse/integrations/redshift/redshift_test.go
lvrach: the thing is to test `IF EXISTS` you need to run multiple CrashRecover methods in parallel and time it correctly. The issue happens when the table is deleted between the execution of two queries and that is hard to test consistently and reliably.

So here I am just testing CrashRecover, but not the specific issue.

PR #4520, opened 2024-03-25
exp: add docker logs

lvrach:


lvrach:
Screenshot 2024-03-25 at 10 10 34 PM

lvrach:
port conflict between ipv6 and ipv4

PR #4540, opened 2024-04-03
chore: refactor onConfigDataEvent

cisse21:
# Description
Refactor onConfigDataEvent function

## Linear Ticket

Fixes PIPE-935

## Security


warehouse/router/router_test.go
```diff
@@ -1112,6 +1108,7 @@ func TestRouter(t *testing.T) {
 		}))
 
 		ctx, cancel := context.WithCancel(context.Background())
+		g, _ := errgroup.WithContext(ctx)
```
fracasula, line 1111: Why not using `errgroup.Group` if you're ignoring the context? Internally the context is not stored nor used at all, it is just for the end user:

```go
func WithContext(ctx context.Context) (*Group, context.Context) {
	ctx, cancel := withCancelCause(ctx)
	return &Group{cancel: cancel}, ctx
}
```

If one of the routines errors then the context is going to be cancelled but since you're ignoring it then:
```suggestion
		g := errgroup.Group{}
```

warehouse/router/router.go
```diff
@@ -748,17 +694,41 @@ func (r *Router) updateCreateJobMarker(warehouse model.Warehouse, lastProcessedT
...
 }
 
-// Disable disables a router:)
-func (r *Router) Disable() {
-	r.isEnabled.Store(false)
-}
+func (r *Router) loadStats() {
+	r.stats.processingPendingJobsStat = r.statsFactory.NewTaggedStat("wh_processing_pending_jobs", stats.GaugeType, stats.Tags{
+		"destType": r.destType,
+	})
```
fracasula, line 717: Same tags everywhere. It looks like we could reuse the same map.

lvrach (approved):

PR #4556, opened 2024-04-05
feat: adaptations of metadata for internal batch event

achettyiitr:
# Description

- Settings parameters:
  - IP
  - Payload sanitisation (ReceivedAt, userID)
  - Routing key

## Linear Ticket

- Resolves PIPE-809, PIPE-784

## Security


gateway/handle.go
```diff
@@ -716,30 +717,32 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(
 		out              = make([]jobObject, 0, len(eventsBatch))
 	)
 
-	for idx, v := range eventsBatch {
-		toSet, ok := v.Value().(map[string]interface{})
+	for _, v := range eventsBatch {
+		payload := v.Get("payload")
+		parameters := v.Get("parameters")
+		if !payload.Exists() || !parameters.Exists() {
+			panic("payload or parameters not found in batch")
```
fracasula, line 724: Does this mean that a malformed message that somebody else pushed into Pulsar could make RudderServer crash?

achettyiitr: Do you think we should return 404 from here? And in the src router, if we get 404 we drop it and have an alert on top of it?

fracasula: I'm concerned about crashing the whole thing because we're planning to run RS in EMBEDDED mode. If this were to happen we would stop also the processing pipeline.

Ideally we should return something here that we should get alerts upon, but the GW should be able to stay up serving other requests that are maybe correct. Also the processor and the router should keep running.

PR #4569, opened 2024-04-10
chore: add sourceCategory label to event_delivery_time metric

satishrudderstack:
# Description
- As part of [delays project](https://www.notion.so/rudderstacks/End-to-End-Delays-in-Pipeline-362b6fe1e3b148c68aa538670d2a60da), we want to show end-to-end delay for event stream cloud destinations. We are adding `sourceCategory` label on `event_delivery_time` metric so that we can query using `sourceCategory`
- Tested by deploying a dataplane on dev cluster. We are getting `sourceCategory="warehouse"` for rETL, `sourceCategory=""` for eventStream and `sourceCategory="webhook"` for customerio source
  - Attaching part of result from /metrics from processor pod
Screenshot 2024-04-12 at 2 08 49 PM

completes OBS-453

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-453/add-sourcecategory-label-to-event-delivery-time-metric

## Security


achettyiitr:
For event-stream instead of sending an empty value, should we send some appropriate value like we are doing for [reporting](https://github.com/rudderlabs/rudder-server/blob/af82a6f60521847ae98cbad7dd938ee44c7a9b13/enterprise/reporting/event_stats.go#L31-L34)?

router/worker.go
```diff
@@ -166,6 +166,7 @@ func (w *worker) workLoop() {
 				UserID:             userID,
 				JobID:              job.JobID,
 				SourceID:           parameters.SourceID,
+				SourceCategory:     parameters.SourceCategory,
```
satishrudderstack, line 169: We are using [similar logic](https://github.com/rudderlabs/rudder-server/blob/master/router/handle.go#L354) to populate sourceCategory for reporting

satishrudderstack:
> For event-stream instead of sending an empty value, should we send some appropriate value like we are doing for [reporting](https://github.com/rudderlabs/rudder-server/blob/af82a6f60521847ae98cbad7dd938ee44c7a9b13/enterprise/reporting/event_stats.go#L31-L34)?

We are thinking of updating the `category` column in `source_definitions` table in config-backend db instead of adding default values at multiple places. For eventStream sources, the `category` column is empty currently

achettyiitr (approved):

fracasula (approved):

PR #4578, opened 2024-04-12
fix: actions cancelling on push to master

Sidddddarth:
# Description

Our earlier change was cancelling actions in progress.
[That change](https://github.com/rudderlabs/rudder-server/pull/4573) was unwittingly cancelling runs on push to master as well. 
IMO, it's nice to have this suite run for every commit to master without cancelling.
image

On push to master `github.head_ref` would be empty, so one way to group the workflows would be by falling back on commit hash - `github.sha`.

Alternatively, we could separate workflow files for `push`, `pull-request` for simplicity and not have concurrency groups in `push` workflow.

## Security

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -9,7 +9,7 @@ on:
       - main
   pull_request:
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref }}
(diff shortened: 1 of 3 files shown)


mihir20 (approved):
LGTM

cisse21 (approved):

fracasula (approved):

PR #4587, opened 2024-04-17
chore: adapt internal endpoint to ingest msg in new format

mihir20:
# Description
Updating `/internal/v1/batch` endpoint to consume messages in newer format defined [here](https://github.com/rudderlabs/rudder-schemas/blob/main/go/stream/message.go). We will receive an array of messages and each message can have different sourceID and workspaceID. 

## Linear Ticket

Pipe-982

## Security


gateway/handle.go
```diff
@@ -42,6 +45,11 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/types"
 )
 
+var (
+	streamMsgValidator = stream.NewMessageValidator()
```
achettyiitr, line 49: We can probably avoid this global and pass it as a dependency?
```
type Handle struct {
	config           *config.Config
	logger           logger.Logger
	stats            stats.Stats
	tracer           stats.Tracer
	messageValidator func(msg *stream.Message) error
        ...

}
```

And the setup will look something like this:
```
func (gw *Handle) Setup(
	ctx context.Context,
	config *config.Config, logger logger.Logger, stat stats.Stats,
	application app.App, backendConfig backendconfig.BackendConfig, jobsDB, errDB jobsdb.JobsDB,
	rateLimiter throttler.Throttler, versionHandler func(w http.ResponseWriter, r *http.Request),
	rsourcesService rsources.JobService, transformerFeaturesService transformer.FeaturesService, sourcehandle sourcedebugger.SourceDebugger,
	opts ...OptFunc,
) error {
	gw.config = config
	gw.logger = logger
	gw.stats = stat\
	gw.messageValidator = stream.NewMessageValidator()
```

And usage will look something like:
```
	  for _, msg := range messages {
		if err := gw.messageValidator(&msg); err != nil {
			return nil, errors.New("invalid message")
		}
                 ...
           }
```

gateway/handle.go
```diff
@@ -684,72 +696,69 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
...
+
+	jobs = make([]*jobsdb.JobT, 0, len(messages))
 
-		if isUserSuppressed(workspaceID, userIDFromReq, sourceID) {
+	for _, msg := range messages {
+		err := streamMsgValidator(&msg)
+		if err != nil {
+			return nil, fmt.Errorf("%s", response.InvalidJSON)
+		}
+		arctx := gw.authRequestContextForSourceID(msg.Properties.SourceID)
```
achettyiitr, line 756: We can move arCtx within the if condition for isUserSuppressed

achettyiitr: Also, arCtx would be nil since we removed the sourceIDAuth middleware?

gateway/handle.go
```diff
@@ -759,71 +768,56 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(
 			continue
 		}
 
-		userID := buildUserID("", anonIDFromReq, userIDFromReq)
-		receivedAt, _ := toSet["receivedAt"].(string)
+		receivedAt := gjson.GetBytes(msg.Payload, "received_at").String()
 		if receivedAt == "" {
 			receivedAt = time.Now().Format(misc.RFC3339Milli)
 		}
```
achettyiitr, line 774: This is not required now. We will be having this in msg itself.
```GO
      msg.Properties.ReceivedAt.Format(misc.RFC3339Milli),
```

gateway/handle.go
```diff
@@ -759,71 +768,56 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(
...
-			ReceivedAt: userEvent.receivedAt,
+			Batch:      []json.RawMessage{msg.Payload},
+			ReceivedAt: receivedAt,
 		}
-		payload, err = json.Marshal(eventBatch)
+
+		payload, err := json.Marshal(eventBatch)
 		if err != nil {
-			panic(err)
+			return nil, err
```
achettyiitr, line 804: Can we add some context here?
```suggestion
			return nil,  fmt.Errorf("marshalling event batch: %w", err)
```

gateway/handle.go
```diff
@@ -684,72 +696,69 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
...
-	gw.requestSizeStat.Observe(float64(len(body)))
 
-	type jobObject struct {
-		userID     string
-		events     []map[string]interface{}
-		receivedAt string
+	type singularEventBatch struct {
+		Batch      []json.RawMessage `json:"batch"`
+		ReceivedAt string            `json:"receivedAt"`
 	}
```
achettyiitr, line 731: We can take ip as well. It's been added now in rudder schema.
```suggestion
		type SingularEventBatch struct {
			Batch      []map[string]interface{} `json:"batch"`
			RequestIP  string                   `json:"requestIP"`
			ReceivedAt string                   `json:"receivedAt"`
		}
```

gateway/response/response.go
```diff
@@ -75,6 +75,8 @@ const (
 
 	transPixelResponse = "\x47\x49\x46\x38\x39\x61\x01\x00\x01\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00\x21\xF9\x04" +
 		"\x01\x00\x00\x00\x00\x2C\x00\x00\x00\x00\x01\x00\x01\x00\x00\x02\x02\x44\x01\x00\x3B"
+
+	InvalidMessage = "missing required fields in message"
```
achettyiitr, line 79: It seems to be too generic. Should we scope it down to stream message? wdyt?
```suggestion
	InvalidStreamMessage = "missing required fields in stream message"
```

gateway/handle.go
mihir20: @lvrach I forgot, why we decided to declare it globally instead of injecting it as a dependency?

gateway/handle.go
mihir20: > Also, arCtx would be nil since we removed the sourceIDAuth middleware?
that's why we are initialising it here

gateway/handle.go
```diff
@@ -759,71 +769,52 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(
...
+			TraceParent:     msg.Properties.TraceID,
+		}
+
+		marshalledParams, err := json.Marshal(jobsDBParams)
+		if err != nil {
+			gw.logger.Errorn(
+				"[Gateway] Failed to marshal parameters map. Parameters: %+v",
+				logger.NewField("params", jobsDBParams),
+				obskit.Error(err),
+			)
```
achettyiitr, line 787: We don't need `Parameters: %+v`.
```suggestion
			gw.logger.Errorn("[Gateway] Failed to marshal parameters map",
				logger.NewField("params", jobsDBParams),
				obskit.Error(err),
			)
```

gateway/handle.go
```diff
@@ -684,72 +696,70 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
...
+		UserID          string `json:"user_id"`
+		TraceParent     string `json:"traceparent"`
 	}
-	gw.requestSizeStat.Observe(float64(len(body)))
 
-	type jobObject struct {
-		userID     string
-		events     []map[string]interface{}
-		receivedAt string
+	type singularEventBatch struct {
```
achettyiitr, line 728: Not related to the changes. But this is being created twice. https://github.com/rudderlabs/rudder-server/blob/c2acf9f91b307b0e78ab4eed80afbb5e01c84c18/gateway/handle.go#L470

Can we create it outside?

mihir20: Good point, This might require some refactoring, we can pick this up separately.

gateway/gateway_test.go
```diff
@@ -398,7 +400,6 @@ var _ = Describe("Gateway", func() {
 		conf.Set("Gateway.enableEventSchemasFeature", false)
 		c = &testContext{}
 		c.Setup()
-		c.initializeAppFeatures()
```
achettyiitr, line 401: Any reason we removed it from here and initializing in couple of places

achettyiitr (approved):

gateway/gateway_test.go
mihir20: It was  initializeAppFeatures was initialising some mocks which were used in every test but I wanted to use some other mocks in my test so I had to remove this and add at other places.

gateway/handle.go
```diff
@@ -684,73 +695,71 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
...
+		if err != nil {
+			return nil, errors.New(response.InvalidStreamMessage)
+		}
+		if isUserSuppressed(msg.Properties.WorkspaceID, msg.Properties.UserID, msg.Properties.SourceID) {
 			gw.logger.Infon("suppressed event",
-				logger.NewStringField("sourceID", sourceID),
-				logger.NewStringField("workspaceID", workspaceID),
-				logger.NewStringField("userIDFromReq", userIDFromReq),
+				logger.NewStringField("sourceID", msg.Properties.SourceID),
+				logger.NewStringField("workspaceID", msg.Properties.WorkspaceID),
```
lvrach, line 759:
```suggestion
				obskit.SourceID(msg.Properties.SourceID),
				obskit.WorkspaceID(msg.Properties.WorkspaceID),
```

gateway/handle.go
```diff
@@ -684,73 +695,71 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
...
-		if isUserSuppressed(workspaceID, userIDFromReq, sourceID) {
+	if len(messages) == 0 {
+		return nil, errors.New(response.NotRudderEvent)
+	}
+
+	jobs = make([]*jobsdb.JobT, 0, len(messages))
+
+	for _, msg := range messages {
+		err := gw.streamMsgValidator(&msg)
+		if err != nil {
```
lvrach, line 753: Shall we log the error here, we are losing information since we are returning just `errors.New(response.InvalidStreamMessage)` 
```suggestion
		if err != nil {
```

lvrach (approved):

PR #4593, opened 2024-04-19
chore: flip sourceID and originalSourceID in replay events for transformations

Sidddddarth:
# Description

Flip the `SourceID` and `OriginalSourceID(if != "")` before transformations and back after.

## Linear Ticket

< Replace with Linear Link >

## Security

### DIFF
--- a/processor/transformer/transformer.go
+++ b/processor/transformer/transformer.go
@@ -51,6 +51,7 @@ var json = jsoniter.ConfigCompatibleWithStandardLibrary
 type Metadata struct {
 	SourceID            string                            `json:"sourceId"`
 	SourceName          string                            `json:"sourceName"`
+	OriginalSourceID    string                            `json:"originalSourceId"`
 	WorkspaceID         string                            `json:"workspaceId"`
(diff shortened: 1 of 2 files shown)


processor/transformer/transformer.go
```diff
@@ -342,6 +353,12 @@ func (trans *handle) transform(
 		// Transform is one to many mapping so returned
 		// response for each is an array. We flatten it out
 		for _, transformerResponse := range batch {
+			// flip back sourceID and original sourceID
+			if transformerResponse.Metadata.OriginalSourceID != "" {
```
mihir20, line 357: Not sure if we should do this, there's a possibility that this data undergoes updates during the transformation, which could impact its reliability for our purposes.

Sidddddarth: I agree.
The same pattern is being used for `messageID` in several other places in processor too.
Maybe that's something we should take a look at too.

Sidddddarth: Just found out metadata can't be updated in a user transformation, so should be a non-issue.

PR #4594, opened 2024-04-19
fix: ever increasing idle http connections while fetching transformer features

cisse21:
# Description

Use a single http client to make transformer fetch call
Not using the default transport provided by Go 
Renaming few fields in the features implementation 

## Linear Ticket

Fixes PIPE-1019

## Security


services/transformer/features.go
```diff
@@ -38,12 +40,21 @@ var defaultTransformerFeatures = `{
...
+func NewFeaturesService(ctx context.Context, config *config.Config, featConfig FeaturesServiceConfig) FeaturesService {
 	handler := &featuresService{
 		features: json.RawMessage(defaultTransformerFeatures),
 		logger:   logger.NewLogger().Child("transformer-features"),
 		waitChan: make(chan struct{}),
-		config:   config,
+		options:  featConfig,
+		client: &http.Client{
+			Transport: &http.Transport{
+				DisableKeepAlives:   config.GetBool("Transformer.Client.disableKeepAlives", true),
```
lvrach, line 51: shouldn't this be false by default?
```suggestion
				DisableKeepAlives:   config.GetBool("Transformer.Client.disableKeepAlives", false),
```

lvrach (approved):

koladilip (approved):

services/transformer/features_impl.go
```diff
@@ -11,16 +11,16 @@ import (
...
-	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-server/utils/httputil"
 )
 
 type featuresService struct {
 	logger   logger.Logger
 	waitChan chan struct{}
-	config   FeaturesServiceConfig
+	options  FeaturesServiceConfig
```
mihir20, line 21: Should we rename this struct as well?
```suggestion
	options  FeaturesServiceOptions
```

mihir20 (approved):

achettyiitr (approved):

services/transformer/features.go
cisse21: I see it to be true on transformer [here](https://github.com/rudderlabs/rudder-server/blob/master/processor/transformer/transformer.go#L221)

services/transformer/features_impl.go
cisse21: Done

services/transformer/features.go
cisse21: Changed it to false anyways

PR #4598, opened 2024-04-19
feat: introduce tx idle timeout in postgres connections

lvrach:
# Description

## Motivation
In case, rudder-server processor or gateway crashes (panic, OOM killed), on going transactions remain open for a long time. When those transactions acquire a lock it creates issue for new connections to run their quires properly due locks. He have faced more than one [CFD](https://linear.app/rudderstack/issue/PIPE-1003/couple-of-partial-downtime-alerts-today-for-onefootball-and-frontrow).

## Solution 

Introduce [idle_in_transaction_session_timeout](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-IDLE-IN-TRANSACTION-SESSION-TIMEOUT) connection option when establishing tcp connection.

> Terminate any session that has been idle (that is, waiting for a client query) within an open transaction for longer than the specified amount of time. If this value is specified without units, it is taken as milliseconds. A value of zero (the default) disables the timeout.

> This option can be used to ensure that idle sessions do not hold locks for an unreasonable amount of time. Even when no significant locks are held, an open transaction prevents vacuuming away recently-dead tuples that may be visible only to this transaction; so remaining idle for a long time can contribute to table bloat. See [Section 25.1](https://www.postgresql.org/docs/current/routine-vacuuming.html) for more details.

This setting is generally useful, not only for when their is a crash.

 
Default timeout is set to 5 minutes. A safe, high enough value not to affect any queries running production: 

`max(pg_stat_activity_max_tx_duration{container="postgres-exporter", team="pipelines", state="idle in transaction"})`
Screenshot 2024-04-22 at 7 15 52 AM

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1017/configure-idle-in-transaction-session-timeout
## Security


utils/misc/dbutils.go
```diff
@@ -19,15 +20,20 @@ func GetConnectionString(c *config.Config, componentName string) string {
 	port := c.GetInt("DB.port", 5432)
 	password := c.GetString("DB.password", "ubuntu") // Reading secrets from
 	sslmode := c.GetString("DB.sslMode", "disable")
+	idleTxTimeout := c.GetDuration("DB.IdleTxTimeout", 5, time.Minute)
```
mihir20, line 23: Should we keep this timeout limited to 120 Secs as from the graph in the desc max txn time seems to be 60-65 secs only? Keeping high value of timeout may still result in similar issues we are trying to solve(even-though for a smaller time)

mihir20 (approved):
Minor comment, otherwise LGTM

utils/misc/dbutils.go
fracasula: This timeout affects connections waiting for a client query within an open transaction for longer than the specified amount of time.

So if the client opens a transaction, runs a query, keeps the transaction open, it does something with the results and that "something" takes longer than 120 seconds, then the connection is going to be terminated. This would mean that the client will quite likely have to restart from the beginning.

I know the max in the graph shows as 60-65 but there might be edge cases where it could go above 120 I guess. Hard to be 100% sure without checking all metrics of all time and even then, code changes so you might experience that behaviour anyway at some point. This is the reason why 5 minutes seems safer to me and still reasonable enough not to cause incidents due to connections left alive forever.

fracasula (approved):

PR #4606, opened 2024-04-23
fix: correctly quote connection on subscription creation

lvrach:
# Description

### Issue 
The issue triggered by this change https://github.com/rudderlabs/rudder-server/pull/4598, when deployed to free users.

The underlaying cause was that we didn't properly escaping the connection info string passed to the query. The new connection query contained `'`, causing the parsing to break and the following error to happen:

```
{"level":"ERROR","ts":"2024-04-23T12:27:22.674Z","logger":"runner","caller":"runner/runner.go:274","msg":"Terminal error: rudder core: logical replication in \"hostedmtedmt-v0-rs-postgresql-3.hostedmtedmt-v0-rs-postgresql-headless\": failed to create subscription on shared database: pq: syntax error at or near \"-\""}
```

### Fix 
Using [pq.QuoteLiteral](https://pkg.go.dev/github.com/lib/pq#QuoteLiteral) to properly escape the connection string. 
Updated the test to contain the `options='-c idle_in_transaction_session_timeout=300000'`

### TODOs

* Refactor the test to use rudder-go-kit resource
* Use the misc.GetConnectionString instead of hardcoding `options='-c idle_in_transaction_session_timeout=300000` in tests
* Integration test with sharedDB

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1031/fix-correctly-quote-connection-on-subscription-creation

fixes PIPE-1031 

## Security

### DIFF
--- a/services/rsources/handler_test.go
+++ b/services/rsources/handler_test.go
@@ -1585,7 +1585,7 @@ func newDBResource(pool *dockertest.Pool, networkId, hostname string, params ...
 	internalDSN := fmt.Sprintf("postgres://%[1]s:%[2]s@%[3]s:5432/%[4]s?sslmode=disable", username, password, hostname, database)
 	var (
 		db  *sql.DB
-		dsn = fmt.Sprintf("host=localhost port=%[1]s user=%[2]s password=%[3]s dbname=%[4]s sslmode=disable", port, username, password, database)
+		dsn = fmt.Sprintf("host=localhost port=%[1]s user=%[2]s password=%[3]s dbname=%[4]s sslmode=disable options='-c idle_in_transaction_session_timeout=300000' ", port, username, password, database)
 	)
 	// exponential backoff-retry, because the application in the container might not be ready to accept connections yet
 	if err := pool.Retry(func() (err error) {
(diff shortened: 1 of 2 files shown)


fracasula (approved):

services/rsources/handler_test.go
```diff
@@ -1585,7 +1585,7 @@ func newDBResource(pool *dockertest.Pool, networkId, hostname string, params ...
 	internalDSN := fmt.Sprintf("postgres://%[1]s:%[2]s@%[3]s:5432/%[4]s?sslmode=disable", username, password, hostname, database)
 	var (
 		db  *sql.DB
-		dsn = fmt.Sprintf("host=localhost port=%[1]s user=%[2]s password=%[3]s dbname=%[4]s sslmode=disable", port, username, password, database)
+		dsn = fmt.Sprintf("host=localhost port=%[1]s user=%[2]s password=%[3]s dbname=%[4]s sslmode=disable options='-c idle_in_transaction_session_timeout=300000' ", port, username, password, database)
```
mihir20, line 1588: should we call `GetConnectionString` method here to get connection string?

lvrach: That is what I plan to do with using rudder-go-kit for the db resource. It will take some time to refactoring things here, so I would like to do them in separate PR.

mihir20 (approved):

PR #4619, opened 2024-04-25
feat: monitor warehouse databases

achettyiitr:
# Description

- Integrating with `sqlutil.MonitorDatabase` for warehouse and notifier package.

## Linear Ticket

- Resolves PIPE-894

## Security

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -83,7 +83,7 @@ require (
 	github.com/rudderlabs/analytics-go v3.3.3+incompatible
 	github.com/rudderlabs/bing-ads-go-sdk v0.2.1
 	github.com/rudderlabs/compose-test v0.1.3
-	github.com/rudderlabs/rudder-go-kit v0.27.0
+	github.com/rudderlabs/rudder-go-kit v0.29.0
 	github.com/rudderlabs/rudder-observability-kit v0.0.3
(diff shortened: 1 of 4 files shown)


Sidddddarth (approved):
Suggestion: Considering both the `db`s we're monitoring here are `sqlmw.DB`, maybe we could include this directly in `sqlmw`?
A side-effect(or benefit?) is that all the warehouse connections will begin to be monitored(not sure how that'll go - if they even expose stats we observe). We could add an option to the ones already present in `sqlmw` though, to selectively use this only for databases we're interested in monitoring.

LGTM otherwise

lvrach (approved):

achettyiitr:
> Suggestion: Considering both the `db`s we're monitoring here are `sqlmw.DB`, maybe we could include this directly in `sqlmw`? A side-effect(or benefit?) is that all the warehouse connections will begin to be monitored(not sure how that'll go - if they even expose stats we observe). We could add an option to the ones already present in `sqlmw` though, to selectively use this only for databases we're interested in monitoring.
> 
> LGTM otherwise

Good point though. But for simplicity, since it's just being used one in the warehouse package and one in the notifier package, keeping it as it is.

PR #4627, opened 2024-04-29
chore: add support for live events in internal batch endpoint

mihir20:
# Description

- Added support for live events in internal batch endpoint
- Small refactoring, to make stats function more specific to the usecase and use actual source config and not auth context

## Linear Ticket
https://linear.app/rudderstack/issue/PIPE-842/%5Bingestionsvc%5D-source-live-events

## Security


gateway/handle.go
```diff
@@ -753,15 +765,15 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(reqType string, body []byt
 		err := gw.streamMsgValidator(&msg)
 		if err != nil {
 			gw.logger.Errorn("invalid message in request", logger.NewErrorField(err))
-			return nil, errors.New(response.InvalidStreamMessage)
+			return nil, nil, errors.New(response.InvalidStreamMessage)
 		}
+		arctx := gw.authRequestContextForSourceID(msg.Properties.SourceID)
```
lvrach, line 770: Can we introduce a simpler `gw.writeKeyForSourceID` method?

gateway/handle.go
```diff
@@ -670,6 +671,16 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
 				stats.CountType,
 				gw.newReqTypeStatsTagsWithReason(reqType, ""),
 			).Count(len(jobs))
+
+			// Sending events to config backend
+			for _, job := range jobs {
+				writeKey, ok := jobsUUIDToWriteKeyMap[job.UUID]
```
lvrach, line 677: What if we call `gw.writeKeyForSourceID` here instead of passing the `jobsUUIDToWriteKeyMap` ?

gateway/handle.go
mihir20: good idea

gateway/handle.go
mihir20: do you mean to say, we get the source ID from params and call `writeKeyForSourceID` with sourceID?

gateway/handle_observability.go
```diff
@@ -18,14 +20,13 @@ func (gw *Handle) NewSourceStat(arctx *gwtypes.AuthRequestContext, reqType strin
...
 }
 
-func (gw *Handle) newSourceStatTagsWithReason(arctx *gwtypes.AuthRequestContext, reqType, reason string) stats.Tags {
+func (gw *Handle) newSourceStatTagsWithReason(s backendconfig.SourceT, reqType, reason string) stats.Tags {
 	tags := stats.Tags{
-		"source":       arctx.SourceTag(),
-		"source_id":    arctx.SourceID,
-		"write_key":    arctx.WriteKey,
+		"source":       misc.GetTagName(s.WriteKey, s.Name),
+		"source_id":    s.SourceDefinition.ID,
```
Sidddddarth, line 26: Has something changed..?
```suggestion
		"source_id":    s.ID,
```

Sidddddarth (approved):
Didn't understand the motivation behind changing some stat functions.
Otherwise LGTM

gateway/gateway_test.go
```diff
@@ -48,6 +48,7 @@ import (
 	mocksJobsDB "github.com/rudderlabs/rudder-server/mocks/jobsdb"
 	mocksTypes "github.com/rudderlabs/rudder-server/mocks/utils/types"
 	sourcedebugger "github.com/rudderlabs/rudder-server/services/debugger/source"
+	mockSrcDebugger "github.com/rudderlabs/rudder-server/services/debugger/source/mocks"
```
achettyiitr, line 51:
```suggestion
	mocksrcdebugger "github.com/rudderlabs/rudder-server/services/debugger/source/mocks"
```

gateway/handle.go
```diff
@@ -670,6 +670,17 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
 				stats.CountType,
 				gw.newReqTypeStatsTagsWithReason(reqType, ""),
 			).Count(len(jobs))
+
+			// Sending events to config backend
+			for _, job := range jobs {
+				sourceID := gjson.GetBytes(job.Parameters, "source_id").String()
+				sourceConfig := gw.getSourceConfigFromSourceID(sourceID)
+				if sourceConfig.WriteKey == "" {
+					gw.logger.Errorn("unable to get writeKey for job", logger.NewStringField("uuid", job.UUID.String()))
```
achettyiitr, line 679: Shouldn't it be a warning log? Also, Can we add sourceID as well for debugging?
```suggestion
					gw.logger.Warnn("unable to get writeKey for job",
						obskit.SourceID(sourceID),
						logger.NewStringField("jobUUID", job.UUID.String()),
					)
```

gateway/handle.go
```diff
@@ -819,6 +830,15 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(reqType string, body []byt
...
 }
 
+func (gw *Handle) getSourceConfigFromSourceID(sourceID string) backendconfig.SourceT {
+	gw.configSubscriberLock.RLock()
+	defer gw.configSubscriberLock.RUnlock()
+	if s, ok := gw.sourceIDSourceMap[sourceID]; ok {
+		return s
+	}
+	return backendconfig.SourceT{}
+}
```
achettyiitr, line 840: I feel instead of returning too much information, we should just return, what we want (writeKey). wdyt?
```suggestion
func (gw *Handle) getWriteKeyFromSourceID(sourceID string) (string, bool) {
	gw.configSubscriberLock.RLock()
	defer gw.configSubscriberLock.RUnlock()
	if s, ok := gw.sourceIDSourceMap[sourceID]; ok {
		return s.WriteKey, true
	}
	return "", false
}
```

gateway/handle_observability.go
```diff
@@ -18,14 +20,13 @@ func (gw *Handle) NewSourceStat(arctx *gwtypes.AuthRequestContext, reqType strin
 	}
 }
 
-func (gw *Handle) newSourceStatTagsWithReason(arctx *gwtypes.AuthRequestContext, reqType, reason string) stats.Tags {
+func (gw *Handle) newSourceStatTagsWithReason(s backendconfig.SourceT, reqType, reason string) stats.Tags {
```
achettyiitr, line 23: Didn't get why we need to pass too much information when `*gwtypes.AuthRequestContext` was sufficient?

achettyiitr: Why we are not passing `source_type` in tags?

gateway/handle.go
achettyiitr: I understand, we changed the signature for newSourceStatTagsWithReason which `backendconfig.SourceT` instead of `*gwtypes.AuthRequestContext`. But I didn't understand why it was done.

services/debugger/source/eventUploader.go
```diff
@@ -1,5 +1,6 @@
 package sourcedebugger
 
+//go:generate mockgen -destination=./mocks/mock.go -package=mocks github.com/rudderlabs/rudder-server/services/debugger/source SourceDebugger
```
lvrach, line 3: Do we need to generate a mock? The interface signature seems simple enough to mock by hand.

achettyiitr (approved):

achettyiitr (approved):

gateway/handle_observability.go
mihir20: missed it added now

services/debugger/source/eventUploader.go
mihir20: Its simpler to just generate a mock, mock by hand feels too much manual work(in writing a mock as well as to maintain it)

PR #4661, opened 2024-05-07
chore: disable event name tracking for retl connections

itsmihir:
# Description

Disabling event name tracking for all RETL connections, because many customer uses UUIDs for event names, creates a high level of cardinality.

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-470/disable-event-name-tracking-for-retl-connections

## Security


enterprise/reporting/reporting.go
achettyiitr, line 1: Can we add a test scenario for this?

achettyiitr (approved):

satishrudderstack (approved):

PR #4709, opened 2024-05-25
fix(gateway): send on closed chan - (*Handle).webRequestQ

Sidddddarth:
# Description

Using a waitGroup to wait on all `in flight` requests before closing gateway workers' `webRequestQ`.
It is an anti-pattern to close a channel if there's multiple concurrent senders. So we make sure there's no more senders using a simple waitgroup(`inFlightRequests`). A middleware that adds and `Done`s this waitgroup is then needed.

Eventually during `Shutdown` we wait on this waitgroup before closing the workers' `webRequestQ` channels. All the requests would have been handled by this point.

## Linear Ticket

## Security


lvrach:
If you wait for `server.Shutdown` to return normally all the open connections will have been completed. Why this approach is not enough and we have to introduce this counter?

Sidddddarth:
> If you wait for `server.Shutdown` to return normally all the open connections will have been completed. 

True.

> Why this approach is not enough and we have to introduce this counter?

`server.Shutdown` and `close(webRequestQ)` happens in two different goroutines. This counter is a way to synchronise both. Also note that `server.Shutdown` is a graceful exit, calling it will immediately lead to `ListenAndServe` to return `ErrServerClosed`.
From the documentation:
```
When Shutdown is called, [Serve], [ListenAndServe], and [ListenAndServeTLS] immediately return [ErrServerClosed].
Make sure the program doesn't exit and waits instead for Shutdown to return.
```

The synchronisation introduced in this PR does the second part - waiting for shutdown to return and not exiting already.

lvrach:
> The synchronisation introduced in this PR does the second part - waiting for shutdown to return and not exiting already.

That's my point. We can do this by waiting on shutdown to complete, rather than introducing a counter.

fracasula:
@lvrach I had a chat to @Sidddddarth about this. I also agree that using a counter is not ideal.
I proposed to add proper synchronization around the workers and use a common `ctx` for them.
On shutdown we can `cancel` that context and then wait for the workers to finish (i.e. so that they can persist) and then return.

I think this is in line with what you were proposing as well so I'm moving this back to draft to give Siddarth time to implement.

ktgowtham:
I'm assuming this is not a SIGTERM exit, but for some reason `server.ListenAndServe` is being errored out. We probably need to know the reason for it first. logs didn't help me. based on my assumption, `server.Shutdown(context.Background())` isn't being executed for sure. wdyt? 

`func ListenAndServe(ctx context.Context, server *http.Server, shutdownTimeout ...time.Duration) error {
	return gracefulFunc(ctx, server, server.ListenAndServe, shutdownTimeout...)
}`

fracasula (approved):
Let's see how it goes with this approach. Worst case we get killed after the grace period because some in-flight request is being slow, and that would still be better than a `panic` so :+1:

achettyiitr (approved):

PR #4710, opened 2024-05-25
chore: separate dataset index creation

Sidddddarth:
# Description

Separate out index creation for jobsdb datasets. Motivation stemmed from wanting to create indices after jobs are copied during migration([src](https://www.postgresql.org/docs/current/populate.html#POPULATE-RM-INDEXES)). Includes foreign key constraint.

Removed `addDS` journal entry during migration. Which imo is fine considering the `migrateDS` entry contains info about the created dataset.

## Linear Ticket

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -1317,10 +1317,10 @@ func (jd *Handle) createDSInTx(tx *Tx, newDS dataSetT) error {
 
 	// Create the jobs and job_status tables
 	if err = jd.createDSTablesInTx(ctx, tx, newDS); err != nil {
-		return err
+		return fmt.Errorf("creating DS tables %w", err)
 	}
 	if err = jd.createDSIndicesInTx(ctx, tx, newDS); err != nil {
-		return err
+		return fmt.Errorf("creating DS indices %w", err)
 	}
 
 	err = jd.journalMarkDoneInTx(tx, opID)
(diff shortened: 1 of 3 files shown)


jobsdb/jobsdb.go
```diff
@@ -1325,7 +1316,22 @@ func (jd *Handle) createDSInTx(tx *Tx, newDS dataSetT) error {
 	}
 
 	// Create the jobs and job_status tables
-	if _, err = tx.ExecContext(ctx, fmt.Sprintf(`CREATE TABLE %q (
+	if err = jd.createDSTablesInTx(ctx, tx, newDS); err != nil {
+		return err
+	}
+	if err = jd.createDSIndicesInTx(ctx, tx, newDS); err != nil {
+		return err
+	}
```
achettyiitr, line 1324: Should we add some error context here?
```suggestion
	if err = jd.createDSTablesInTx(ctx, tx, newDS); err != nil {
		return fmt.Errorf("creating DS tables %w", err)
	}
	if err = jd.createDSIndicesInTx(ctx, tx, newDS); err != nil {
		return fmt.Errorf("creating DS indices %w", err)
	}
```

jobsdb/jobsdb.go
```diff
@@ -1338,39 +1344,50 @@ func (jd *Handle) createDSInTx(tx *Tx, newDS dataSetT) error {
...
 		error_response JSONB DEFAULT '{}'::JSONB,
 		parameters JSONB DEFAULT '{}'::JSONB,
-		PRIMARY KEY (job_id, job_state, id));`, newDS.JobStatusTable, newDS.JobTable)); err != nil {
+		PRIMARY KEY (job_id, job_state, id));`, newDS.JobStatusTable)); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (jd *Handle) createDSIndicesInTx(ctx context.Context, tx *Tx, newDS dataSetT) error {
```
achettyiitr, line 1363: Should we add more context around the error branches? Same for `createDSTablesInTx` as well.

achettyiitr (approved):

PR #4716, opened 2024-05-28
chore: use GCSUploadIfNotExist setting for file manager

Sidddddarth:
# Description

Update file manager to use new `GCSUploadIfNotExist` setting. Will avoid the gcs storage overwrite related issues.
Users of this filemanager:
1. archiver/workers
2. stash

Both of them are fit to use this setting.

## Linear Ticket

## Security


achettyiitr (approved):

lvrach (changes requested):
Shouldn't you handle `ErrPreConditionFailed` ?

lvrach:
Can we add test case as well ?

PR #4718, opened 2024-05-29
chore: get processor namespace, instanceID during setup

Sidddddarth:
# Description

Instead of `config.Get...`ing the same thing(kube namespace, instanceID) for every event, we may fetch it once and use that information for all.

## Linear Ticket

## Security

### DIFF
--- a/processor/processorBenchmark_test.go
+++ b/processor/processorBenchmark_test.go
@@ -13,8 +13,9 @@ import (
 )
 
 func Benchmark_makeCommonMetadataFromSingularEvent(b *testing.B) {
+	proc := &Handle{}
 	for i := 0; i < b.N; i++ {
-		_ = makeCommonMetadataFromSingularEvent(
+		_ = proc.makeCommonMetadataFromSingularEvent(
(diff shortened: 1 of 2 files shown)


cisse21 (approved):

processor/processorBenchmark_test.go
```diff
@@ -13,8 +13,9 @@ import (
 )
 
 func Benchmark_makeCommonMetadataFromSingularEvent(b *testing.B) {
```
lvrach, line 15: What's the difference in benchmark? Although in isolation isolation rlocks won't have much impact

Sidddddarth: This was added a couple years ago. When we changed some json parsing logic inside. No intention to benchmark anything now.

achettyiitr (approved):

processor/processorBenchmark_test.go
lvrach: just out of curiosity, is there any observable difference if you run the same benchmark before and after the change?

lvrach (approved):

PR #4728, opened 2024-05-30
fix: archival to archive everything whenever it is running

achettyiitr:
# Description

- Since the current archival runs every 6 Hrs and archives 10K (limit) uploads which are older then 5 days.
- This is causing an issue if we are generating more than 10K uploads in 6Hrs which will create a Backlog and it will keep accumulating.
- We face an issue with our MT system where we have accumulated 1.1M uploads. More about it in [here](https://rudderlabs.slack.com/archives/C01JF5MGUAG/p1717045801751619?thread_ts=1717040388.986919&cid=C01JF5MGUAG).

## Linear Ticket

- Resolves [PIPE-1140](https://linear.app/rudderstack/issue/PIPE-1140/warehouse-archiver-fix)

## Security

### DIFF
--- a/warehouse/archive/archiver.go
+++ b/warehouse/archive/archiver.go
@@ -21,6 +21,7 @@ import (
 	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/filemanager"
 	"github.com/rudderlabs/rudder-go-kit/logger"
+
 	"github.com/rudderlabs/rudder-server/services/archiver/tablearchiver"
 	"github.com/rudderlabs/rudder-server/utils/filemanagerutil"
 	"github.com/rudderlabs/rudder-server/utils/misc"
@@ -63,6 +64,7 @@ type Archiver struct {
 		uploadsArchivalTimeInDays   config.ValueLoader[int]
 		archiverTickerTime          config.ValueLoader[time.Duration]
 		backupRowsBatchSize         config.ValueLoader[int]
+		maxLimit                    config.ValueLoader[int]
 	}
 
 	archiveFailedStat stats.Measurement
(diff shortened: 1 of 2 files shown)


warehouse/archive/archiver.go
```diff
@@ -189,6 +192,57 @@ func (*Archiver) usedRudderStorage(metadata []byte) bool {
...
+	uploadsToArchive, err := a.countUploadsToArchive(ctx)
+	if err != nil {
+		return fmt.Errorf("counting uploads to archive: %w", err)
+	}
+
+	for uploadsToArchive > 0 {
+		if err := a.archiveUploads(ctx); err != nil {
+			return fmt.Errorf("archiving uploads: %w", err)
+		}
+		uploadsToArchive -= a.config.maxLimit.Load()
```
mihir20, line 205:
```suggestion
    maxArchiveLimit := a.config.maxLimit.Load()
	for uploadsToArchive > 0 {
		if err := a.archiveUploads(ctx, maxArchiveLimit); err != nil {
			return fmt.Errorf("archiving uploads: %w", err)
		}
		uploadsToArchive -= maxArchiveLimit
```

Would it be better if we load config only once and pass it everywhere? There can be some issues when we update the config and `archiveUploads` method has a value of `maxLimit = 10000` and we update just before we do `uploadsToArchive -= a.config.maxLimit.Load()`

warehouse/archive/archiver.go
```diff
@@ -189,6 +192,57 @@ func (*Archiver) usedRudderStorage(metadata []byte) bool {
...
+		ctx,
+		sqlStatement,
+		fmt.Sprintf("%d DAY", a.config.uploadsArchivalTimeInDays.Load()),
+		model.ExportedData,
+		pq.Array(skipWorkspaceIDs),
+	).Scan(&totalUploads)
+	return totalUploads, err
+}
+
+func (a *Archiver) archiveUploads(ctx context.Context) error {
```
mihir20, line 245:
```suggestion
func (a *Archiver) archiveUploads(ctx context.Context, maxArchiveLimit int) error {
```

warehouse/archive/archiver.go
```diff
@@ -226,6 +280,7 @@ func (a *Archiver) Do(ctx context.Context) error {
 		fmt.Sprintf("%d DAY", a.config.uploadsArchivalTimeInDays.Load()),
 		model.ExportedData,
 		pq.Array(skipWorkspaceIDs),
+		a.config.maxLimit.Load(),
```
mihir20, line 283:
```suggestion
		maxArchiveLimit,
```

ktgowtham (approved):
LGTM

warehouse/archive/archiver.go
achettyiitr: It should be fine since we would rarely change the limit. Also, it should be fine if there are some uploads left because they will get picked up in the next archival.

mihir20: Even if it is fine IMO we should load this only once before the loop for correctness of the code. Its optional though as long as we are able to archive everything we are good.

mihir20 (approved):

warehouse/archive/archiver.go
achettyiitr: Don't think loading the config would have that kind of performance issue. But in any way I have addressed the comment.

fracasula (approved):

PR #4737, opened 2024-05-30
chore: router delivery req-res debug logs

sanpj2292:
# Description

Request - Response debug logs for router delivery would be present in files. Whenever we want to look at request-response logs we can download these files

## Linear Ticket

Resolves INT-2217

## Security


helper/file/file.go
```diff
@@ -0,0 +1,133 @@
...
+	g, _ := errgroup.WithContext(ctx)
+
+	handle := &FileHandle{
+		dir:                 "mydebuglogs/",
+		fileConfig:          fileConfig,
+		logger:              logger.NewLogger().Child("helper.file"),
+		fieldsChan:          fieldsChan,
+		backgroundCtxCancel: cancel,
+		backgroundWait:      g.Wait,
+	}
```
lvrach, line 81: We don't really have a use for context. Closing the channel is enough to stop the process.
```suggestion
	g, _ := errgroup.WithContext(context.Background())

	handle := &FileHandle{
		dir:                 "mydebuglogs/",
		fileConfig:          fileConfig,
		logger:              logger.NewLogger().Child("helper.file"),
		fieldsChan:          fieldsChan,
		backgroundWait:      g.Wait,
	}
```

helper/file/file.go
```diff
@@ -0,0 +1,133 @@
...
+
+func New(parentMod string, opts ...func(*FileHandle)) (*FileHandle, error) {
+	fileConfig := getFileConfig(parentMod)
+	fieldsChan := make(chan fields, fileConfig.MaxBatchSize)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	g, _ := errgroup.WithContext(ctx)
+
+	handle := &FileHandle{
+		dir:                 "mydebuglogs/",
```
lvrach, line 75: A few things:
* `parentMod` is used only for configuration, a bit of an overkill to make required parameter
* `path` seems very important in our use case, so making it a required parameter. If path is not configure correctly we might end up filling the disk
* `config.GetReloadableDurationVar` avoid using `Reloadable` variant if it isn't hot reloadable. It can be confusing.
* Always prefer DI config, instead of using package-level functions. Thus I suggest passing conf.Conf

please note tabs are broken:
```suggestion
func WithOptsFromConfig(prefix string, conf config.Config) {
      return func(fh *FileHandle) {
	       maxBatchSizeConfPath := fmt.Sprintf("%s.DebugHelper.maxBatchSize", parentMod)
	       maxBatchTimeoutConfPath := fmt.Sprintf("%s.DebugHelper.maxBatchTimeout", parentMod)


           fh.MaxBatchTimeout = conf.GetDurationVar(20, time.Second, fh.MaxBatchTimeout)
           fh.MaxBatchSize = conf.GetIntVar(200, 1, fh.MaxBatchSize)
    }
}

func New(path string, opts ...func(*FileHandle)) (*FileHandle, error) {
	fileConfig := getFileConfig(parentMod)
	fieldsChan := make(chan fields, fileConfig.MaxBatchSize)

	ctx, cancel := context.WithCancel(context.Background())
	g, _ := errgroup.WithContext(ctx)

	handle := &FileHandle{
		dir:                path,
```

PR #4738, opened 2024-05-30
fix: populate singular batch with writeKey in internal handler

lvrach:
# Description

Events coming for HA ingestion did not appear in live-events viewer, due to `writeKey` missing from the payload.

Note: this PR doesn't contain test, as it would require significant changes in our tests suite. I wanted to keep scope small since it is hotfix. I have create a [task](https://linear.app/rudderstack/issue/PIPE-1148/improve-gateway-tests-for-internalv1batch) for them.

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1147/fix-ha-ingestion-live-events

## Security

### DIFF
--- a/gateway/handle.go
+++ b/gateway/handle.go
@@ -674,6 +674,10 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
 			// Sending events to config backend
 			for _, job := range jobs {
 				writeKey := gjson.GetBytes(job.EventPayload, "writeKey").String()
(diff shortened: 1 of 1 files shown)


achettyiitr (approved):

fracasula (commented):
Don't we need any test adaptations due to this change?

lvrach:
> Don't we need any test adaptations due to this change?

I've addressed the problem. Need to work on refactoring those tests, they are hard to maintain.

PR #4746, opened 2024-06-03
fix: trino test

achettyiitr:
# Description

- Using previous version (trinodb/trino:448) for trino test. 
- Else it requires some [additional JVM config changes](https://github.com/rudderlabs/rudder-server/actions/runs/9347809213/job/25725576454?pr=4747) as per the recent release for `trinodb/trino:449`.

## Security

### DIFF
--- a/warehouse/integrations/datalake/testdata/etc/jvm.config
+++ b/warehouse/integrations/datalake/testdata/etc/jvm.config
@@ -9,3 +9,5 @@
 -XX:ReservedCodeCacheSize=256M
 -Djdk.attach.allowAttachSelf=true
 -Djdk.nio.maxCachedBufferSize=2000000
+-XX:+UnlockDiagnosticVMOptions
+-XX:G1NumCollectionsKeepPinned=10000000


cisse21 (approved):

achettyiitr:
> @achettyiitr are the additional JVM config changes that you mention complex at all? I tried following your link but couldn't find them.

Running `trinodb/trino:449` was giving the following error. 

```
+ launcher_opts=(--etc-dir /etc/trino)
+ grep -s -q node.id /etc/trino/node.properties
+ launcher_opts+=("-Dnode.id=${HOSTNAME}")
+ exec /usr/lib/trino/bin/launcher run --etc-dir /etc/trino -Dnode.id=7d368e00ce15
ERROR: Trino requires -XX:+UnlockDiagnosticVMOptions -XX:G1NumCollectionsKeepPinned=10000000 on Java versions lower than 22.0.2 due to JDK-8329528
```

lvrach (approved):

fracasula (approved):
[optional] given that it's easy to run the latest we could consider it. not sure how useful it might be to detect if our integration breaks with datalake but even if there is a small chance that it does it is probably worth it given that it's just 2 env vars.

PR #4781, opened 2024-06-09
feat: snowflake key pair

achettyiitr:
# Description

- Add support for [Key-pair authentication](https://docs.snowflake.com/en/user-guide/key-pair-auth) for Snowflake.
- More about it in [here](https://www.notion.so/rudderstacks/Key-Pair-13cadfa4e44e480fa58b5c6865bfe43f?pvs=4).

## Linear Ticket

- Resolves PIPE-1175

## Security

## Summary by CodeRabbit

- **Refactor**
  - Updated integration tests to use environment variables for checking test credentials availability.
  - Refactored Snowflake integration to support key pair authentication.

- **New Features**
  - Added support for key pair authentication in Snowflake integration.
  - Enhanced Snowflake integration configuration with new parameters for key pair authentication.

- **Chores**
  - Updated dependencies to include `github.com/youmark/pkcs8`.

- **Tests**
  - Expanded test cases to cover key pair encrypted and unencrypted scenarios in Snowflake integration.

### DIFF
--- a/warehouse/integrations/snowflake/snowflake.go
+++ b/warehouse/integrations/snowflake/snowflake.go
@@ -3,8 +3,10 @@ package snowflake
 import (
 	"bytes"
 	"context"
+	"crypto/rsa"
 	"database/sql"
 	"encoding/csv"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"regexp"
@@ -16,6 +18,7 @@ import (
 
 	"github.com/samber/lo"
 	snowflake "github.com/snowflakedb/gosnowflake"
+	"github.com/youmark/pkcs8"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/logger"
@@ -37,14 +40,17 @@ const (
 
 // String constants for snowflake destination config
 const (
-	storageIntegration = "storageIntegration"
-	account            = "account"
-	warehouse          = "warehouse"
-	database           = "database"
-	user               = "user"
-	role               = "role"
-	password           = "password"
-	application        = "Rudderstack_Warehouse"
+	storageIntegration   = "storageIntegration"
+	account              = "account"
+	warehouse            = "warehouse"
+	database             = "database"
+	user                 = "user"
+	role                 = "role"
+	password             = "password"
+	useKeyPairAuth       = "useKeyPairAuth"
+	privateKey           = "privateKey"
+	privateKeyPassphrase = "privateKeyPassphrase"
+	application          = "Rudderstack_Warehouse"
 )
 
 var primaryKeyMap = map[string]string{
@@ -127,14 +133,17 @@ var errorsMappings = []model.JobError{
 }
 
 type credentials struct {
-	account    string
-	warehouse  string
-	database   string
-	user       string
-	role       string
-	password   string
-	schemaName string
-	timeout    time.Duration
+	account              string
+	warehouse            string
+	database             string
+	user                 string
+	role                 string
+	password             string
+	schemaName           string
+	useKeyPairAuth       bool
+	privateKey           string
+	privateKeyPassphrase string
+	timeout              time.Duration
 }
 
 type tableLoadResp struct {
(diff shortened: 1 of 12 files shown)


warehouse/integrations/snowflake/snowflake.go
```diff
@@ -1108,6 +1127,24 @@ func (sf *Snowflake) connect(ctx context.Context, opts optionalCreds) (*sqlmw.DB
 	return middleware, nil
 }
 
+func ParsePrivateKey(privateKey, passPhrase string) (*rsa.PrivateKey, error) {
```
fracasula, line 1130: Exporting stuff just because you need it in your test is a small imo. You could simply rename the test package to `snowflake`, I don't think doing the `snowflake_test` thing is needed really.

fracasula (approved):
Looks good to me, just left a suggestion.

warehouse/integrations/snowflake/snowflake.go
achettyiitr: Sure. let me try this. Last time I remember, we used to have a circular dependency.

achettyiitr: Indeed.

```

achettyiitr: It's because of the validation package. I had [this](https://linear.app/rudderstack/issue/PIPE-170/validations-cleanup) ticket, a long time back to refactor the validations package to remove Init and Globals but it didn't got prioritized.

ktgowtham: do we have a utils package where this method can be moved instead of keeping this under snowflake package? there's a possibility of this method to be used in other integrations. wdyt?

ktgowtham (commented):
We also have a flow where we validate warehouse connection in destination creation flow via cp router. Is that covered too in this PR?

warehouse/integrations/snowflake/snowflake.go
achettyiitr: Since the scope is limited as per Snowflake we are accepting only RSA and not other kinds of private keys like `DSA`, `OPENSSH`, `EC`, etc.. so I would probably keep this under Snowflake. If in future the scope extends we can probably move it to utils.

achettyiitr:
> We also have a flow where we validate warehouse connection in destination creation flow via cp router. Is that covered too in this PR?

We do have a simple test for validations in [here](https://github.com/rudderlabs/rudder-server/blob/ed719f3972e5b803805ef24cc319da741746d0ac/warehouse/integrations/snowflake/snowflake_test.go#L564), but I don't think we need to add these scenarios for validations since we are already capturing it in `Event flow` tests.

ktgowtham:
> > We also have a flow where we validate warehouse connection in destination creation flow via cp router. Is that covered too in this PR?
> 
> We do have a simple test for validations in [here](https://github.com/rudderlabs/rudder-server/blob/ed719f3972e5b803805ef24cc319da741746d0ac/warehouse/integrations/snowflake/snowflake_test.go#L564), but I don't think we need to add these scenarios for validations since we are already capturing it in `Event flow` tests.

I was just checking if the code changes handle the validations that are triggered from UI. or do we need any additional changes in the grpc calls via cp router.

ktgowtham (approved):
We need to time UI changes properly such that UI doesn't get released without server release.

achettyiitr:
> > > We also have a flow where we validate warehouse connection in destination creation flow via cp router. Is that covered too in this PR?
> > 
> > 
> > We do have a simple test for validations in [here](https://github.com/rudderlabs/rudder-server/blob/ed719f3972e5b803805ef24cc319da741746d0ac/warehouse/integrations/snowflake/snowflake_test.go#L564), but I don't think we need to add these scenarios for validations since we are already capturing it in `Event flow` tests.
> 
> I was just checking if the code changes handle the validations that are triggered from UI. or do we need any additional changes in the grpc calls via cp router.

The only change is in the integration config. No changes in web-app, config-backend, cp-router.

PR #4792, opened 2024-06-11
chore: detect tracker not running

lvrach:
# Description

Introduce `warehouse_cron_tracker_tick` metric, always incrementing. If cronTrack loop stop running, the metric will remain stable.

We can setup an alert on `increase(warehouse_cron_tracker_tick[5m]) = 0`

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1183/detect-stuck-cron-tracker

## Security

### DIFF
--- a/warehouse/router/tracker.go
+++ b/warehouse/router/tracker.go
@@ -23,7 +23,14 @@ import (
 // CronTracker Track the status of the staging file whether it has reached the terminal state or not for every warehouse
 // we pick the staging file which is oldest within the range NOW() - 2 * syncFrequency and NOW() - 3 * syncFrequency
 func (r *Router) CronTracker(ctx context.Context) error {
+	tick := r.statsFactory.NewTaggedStat("warehouse_cron_tracker_tick", stats.CountType, stats.Tags{
+		"module":   moduleName,
+		"destType": r.destType,
+	})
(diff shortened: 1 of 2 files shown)


warehouse/router/tracker.go
```diff
@@ -23,7 +23,14 @@ import (
 // CronTracker Track the status of the staging file whether it has reached the terminal state or not for every warehouse
 // we pick the staging file which is oldest within the range NOW() - 2 * syncFrequency and NOW() - 3 * syncFrequency
 func (r *Router) CronTracker(ctx context.Context) error {
+	tick := r.statsFactory.NewTaggedStat("warehouse_cron_tracker_tick", stats.GaugeType, stats.Tags{
+		"module":   moduleName,
+		"destType": r.destType,
+	})
 	for {
+
+		tick.Count(1)
```
achettyiitr, line 32: Shouldn't this be guage?
```suggestion
		tick.Gauge(1)
```

lvrach: Just updated the description.

The idea is to be a count, so we can detect when the counter is not increasing.

achettyiitr (approved):

cisse21 (approved):

PR #4804, opened 2024-06-13
chore: create migration for tracked_user_reports table

mihir20:
# Description

create migration for tracked_user_reports table for POC on monthly tracked users.

Column Name | Data Type | Description
-- | -- | --
id | BIGSERIAL PRIMARY KEY | unique identifier
workspace_id | VARCHAR(64) NOT NULL | workspace identifier
instance_id | VARCHAR(64) NOT NULL | rudder server node identifier
source_id | VARCHAR(64) NOT NULL | source identifier
reported_at | TIMESTAMP WITH TIMEZONE NOT NULL |  
userid_hll | TEXT NOT NULL | serialized HLL for user IDs
anonymousid_hll | TEXT NOT NULL | serialized HLL for anonymous IDs
identified_anonymousid_hll | TEXT NOT NULL | serialized HLL for anonymous IDs where userIDs are known

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1189/create-migrations-for-db-table-on-rudder-server-for-poc

## Security

### DIFF
--- a/sql/migrations/tracked_users/000001_init_schema.up.sql
+++ b/sql/migrations/tracked_users/000001_init_schema.up.sql
@@ -1,0 +1,12 @@
+CREATE TABLE tracked_users_reports (
+                           id BIGSERIAL PRIMARY KEY,
+                           workspace_id VARCHAR(64) NOT NULL,
+                           instance_id VARCHAR(64) NOT NULL,
+                           source_id VARCHAR(64),
+                           reported_at TIMESTAMPTZ NOT NULL,
+                           userid_hll TEXT NOT NULL,
+                           anonymousid_hll TEXT NOT NULL,
+                           identified_anonymousid_hll TEXT NOT NULL
+);
+
+CREATE INDEX idx_tracked_users_reports_reported_at ON tracked_users_reports (reported_at);


lvrach (approved):

sql/migrations/tracked_users/000001_init_schema.up.sql
```diff
@@ -0,0 +1,12 @@
...
+                           workspace_id VARCHAR(64) NOT NULL,
+                           instance_id VARCHAR(64) NOT NULL,
+                           source_id VARCHAR(64),
+                           reported_at TIMESTAMPTZ NOT NULL,
+                           userid_hll TEXT NOT NULL,
+                           anonymousid_hll TEXT NOT NULL,
+                           identified_anonymousid_hll TEXT NOT NULL
+);
+
+CREATE INDEX idx_tracked_users_reports_reported_at ON tracked_users_reports (reported_at);
```
lvrach, line 12: Is this the right index for our query pattern? Do we need source_id included as well?

Well we can always change this so not a blocker

mihir20: As per my understanding we would query everything and aggregate in memory. Correct me if I am wrong here @satishrudderstack

satishrudderstack (approved):

PR #4823, opened 2024-06-20
feat: common flusher for reporting with mtu handler

satishrudderstack:
# Description

- In this pr, along with mtu flushing support, we are also adding a common flusher module.

**Motivation for common flusher**
- We have multiple outbox tables on Rudder Server which needs to be flushed to our central Reporting Service. Currently we have multiple flushers with duplicate code.

**Flusher**
- **CronRunner**: Invokes Flusher's Flush() function periodically and if there's lag increase, it calls Flush() aggressively
- **Flusher**: Core flush logic. It uses db and and a aggregator. This is like common platform for flushing reports from different outbox tables
- **Aggregator**: We can have in-app or in-db aggregators for each table. This takes start, end times as input and returns the aggregated reports as jsons

flusher_components

**Flush steps**
1. Get time range (`start`, `end`) to flush. `start` will be from db. `end` will be min(now() - 30s, end-of-current-hour). We don't want to flush reports where inserts are happening. We always try to flush full window durations which is configured via `flushWindow` unless window is crossing multiple hour boundaries
2. Get aggregated reports. Aggregation can be in-app or in-db
3. Send the aggregated reports to Reporting Service using go routines
4. Delete the reports in the range
flusher

**Features**
- Flush/Aggregation window as configuration. We can increase this window to decrease load on Reporting Service
- Aggregation can be done in-app or in db
- Ability to flush aggressively if lag increases. This is done by not sleeping and increasing the concurrency of go-routines used to make calls to Reporting Service
- Batching of payloads to Reporting Service

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-504/reporting-client-for-tracked-users

## Security


snarkychef:
How do we ensure atomicity of steps 3 & 4 in the above diagram? viz. `send aggregated reports to reporting service` and `delete sent reports`? 

As per my understanding, we used to ensure idempotency using timestamp along with grouping labels as report id. Just wondering how it's affected with the introduction of `Flush/Aggregation window`.

satishrudderstack:
> How do we ensure atomicity of steps 3 & 4 in the above diagram? viz. `send aggregated reports to reporting service` and `delete sent reports`?
> 
> As per my understanding, we used to ensure idempotency using timestamp along with grouping labels as report id. Just wondering how it's affected with the introduction of `Flush/Aggregation window`.

On Reporting Service, we are planning to have hash id generated from the payload for each report and have it as a `id` column on Timescale table. If we receive the same report on Reporting Service, we ignore the report and respond with 200. We have similar logic for `metrics` table

From Rudder Server side, we are calculating the flush window first and we are deleting all the reports in the flush window if all the aggregated reports were all sent successfully. In the aggregated report, we have `min(reported_at)` for the group as timestamp which will be sent to reporting. 

Now, for the case where some reports were sent and there is an error for some reports while sending, we error out and don't delete the reports. We will retry the entire window later and reporting will ignore the reports which it consumed already.

FYI, we don't consider most recent data where inserts are happening. We have config `recentExclusionWindowInSeconds`

Nice point @snarkychef . I have added some code so that our flush window doesn't change on retries

enterprise/reporting/flusher/db/db.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/rudderlabs/rudder-server/enterprise/reporting/flusher/report"
+)
+
+type DB interface {
+	InitDB() error
+	GetStart(ctx context.Context, table string) (time.Time, error)
+	FetchBatch(ctx context.Context, table string, start, end time.Time, limit, offset int) ([]report.RawReport, error)
+	Delete(ctx context.Context, table string, start, end time.Time) error
+	CloseDB() error
+}
```
lvrach, line 17: I would move the interface to flusher instead. 

Also we can avoid InitDB, also consider if CloseDB should be part of the interface or not.

```suggestion
type DB interface {
	GetStart(ctx context.Context, table string) (time.Time, error)
	FetchBatch(ctx context.Context, table string, start, end time.Time, limit, offset int) ([]report.RawReport, error)
	Delete(ctx context.Context, table string, start, end time.Time) error
}
```

enterprise/reporting/flusher/db/postgres_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"testing"
+	"time"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetStart(t *testing.T) {
+	db, mock, err := sqlmock.New()
```
lvrach, line 13: Using an actual Postgres instance for our testing here would be preferable.

Using github.com/rudderlabs/rudder-go-kit/testhelper/docker/resource/postgres

enterprise/reporting/flusher/flusher.go
```diff
@@ -0,0 +1,397 @@
...
+	commonTags stats.Tags
+}
+
+func NewFlusher(db db.DB, log logger.Logger, stats stats.Stats, conf *config.Config, table, reportingURL string, aggregator aggregator.Aggregator, module string) *Flusher {
+	flusherMu.Lock()
+	defer flusherMu.Unlock()
+
+	if instance, exists := flusherInstances[table]; exists {
+		return instance
+	}
```
lvrach, line 77:
```suggestion
```

I would expect the factory to be responsible for doing this. It will also simplify your tests.

enterprise/reporting/flusher/flusher_test.go
```diff
@@ -0,0 +1,168 @@
...
+func setupFlusher(f *Flusher) {
+	f.batchSizeFromDB = config.SingleValueLoader(4)
+	f.batchSizeToReporting = config.SingleValueLoader(2)
+	f.minConcurrentRequests = config.SingleValueLoader(2)
+	f.maxConcurrentRequests = config.SingleValueLoader(5)
+	f.lagThresholdForAggresiveFlushInMins = config.SingleValueLoader(10 * time.Minute)
+	f.flushWindow = config.SingleValueLoader(1 * time.Minute)
+	f.maxOpenConnections = 2
+	f.recentExclusionWindow = config.SingleValueLoader(30 * time.Second)
+}
```
lvrach, line 58: I would try to use the NewFlusher instead of override the non exported methods

enterprise/reporting/flusher/flusher.go
```diff
@@ -0,0 +1,397 @@
...
+func (f *Flusher) send(ctx context.Context, aggReports []json.RawMessage) error {
+	if f.batchSizeToReporting.Load() > 1 {
+		if err := f.sendInBatches(ctx, aggReports); err != nil {
+			return err
+		}
+	} else {
+		if err := f.sendIndividually(ctx, aggReports); err != nil {
+			return err
+		}
+	}
```
lvrach, line 283:
```suggestion
		if err := f.sendInBatches(ctx, aggReports); err != nil {
			return err
		}
```

Can we treat `sendIndividually` as sendInBataches with batch size 1?

enterprise/reporting/flusher/flusher.go
```diff
@@ -0,0 +1,397 @@
...
+	return start, end, false, nil
+}
+
+func (f *Flusher) aggregate(ctx context.Context, start, end time.Time) ([]json.RawMessage, error) {
+	jsonReports, total, unique, err := f.aggregator.Aggregate(ctx, start, end)
+	if err != nil {
+		return nil, err
+	}
+	f.reportsCounter.Observe(float64(total))
+	f.aggReportsCounter.Observe(float64(unique))
```
lvrach, line 269: Do we need total and unique only for observability?

Can we move this observability on the aggregator to simplify the interface?

enterprise/reporting/flusher/client/reporting_client.go
```diff
@@ -0,0 +1,99 @@
...
+	log        logger.Logger
+	tags       stats.Tags
+	stats      stats.Stats
+	reqLatency stats.Measurement
+	reqCount   stats.Counter
+}
+
+func NewReportingClient(url string, log logger.Logger, stats stats.Stats, tags stats.Tags) *ReportingClient {
+	tr := &http.Transport{}
+	nc := &http.Client{Transport: tr, Timeout: config.GetDuration("HttpClient.reporting.timeout", 60, time.Second)}
```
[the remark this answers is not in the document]
satishrudderstack: Moved the client to flusher. Resolving this

enterprise/reporting/flusher/client/reporting_client.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/rudderlabs/rudder-go-kit/config"
+	"github.com/rudderlabs/rudder-go-kit/logger"
+	"github.com/rudderlabs/rudder-go-kit/stats"
+	"github.com/rudderlabs/rudder-server/utils/httputil"
+)
+
+const (
+	StatFlusherHttpReqLatency = "flusher_http_request_duration_seconds"
+	StatFlusherHttpReqCount   = "flusher_http_requests_total"
```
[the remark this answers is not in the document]
satishrudderstack: Changed the names

enterprise/reporting/flusher/aggregator/tracked_users_inapp.go
```diff
@@ -0,0 +1,122 @@
...
+	hll, err := hll.FromBytes(data)
+	if err != nil {
+		return nil, err
+	}
+	return &hll, nil
+}
+
+func marshalReports(aggReportsMap map[string]TrackedUsersReport) ([]json.RawMessage, error) {
+	jsonReports := make([]json.RawMessage, 0, len(aggReportsMap))
+	for _, v := range aggReportsMap {
```
itsmihir, line 114: should we combine all the source IDs for a workspace into a single request to reduce network calls?

enterprise/reporting/flusher/aggregator/tracked_users_inapp.go
```diff
@@ -0,0 +1,122 @@
...
+		data, err := json.Marshal(v)
```
itsmihir, line 115: should we combine all the source ID data for a workspace into a single request to reduce network calls?

satishrudderstack: I don't think we will get much benefit out it in terms of size compression. Most of the size contribution in payload is from HLLs. We are having batching support in flusher where we send multiple reports in a single API call to Reporting Service

enterprise/reporting/flusher/db/postgres_test.go
satishrudderstack: @lvrach I have added a [test](https://github.com/rudderlabs/rudder-server/pull/4823/files#diff-fa63fee478295099e0dc2a7ed417aa52f1a50f411d4cdedfa56452b97c14ccf1) which does the test of entire Runner+Flusher flow using postgres docker. I think it should be enough. Postgres coverage is over 80%

enterprise/reporting/flusher/client/reporting_client_test.go
```diff
@@ -0,0 +1,68 @@
+package client
```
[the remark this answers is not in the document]
satishrudderstack: Removed the client package. It is now a method in flusher itself.

enterprise/reporting/flusher/db/postgres_test.go
satishrudderstack: @lvrach I have removed the db package. Earlier I thought may be can can db agnostic implementation of the flusher. Since, we are making use of postgres queries directly inside aggregator for performance reasons, I think we  can remove the db interface. Our flusher now works only with postgres

lvrach (approved):

enterprise/reporting/flusher/aggregator/tracked_users_inapp.go
```diff
@@ -0,0 +1,119 @@
...
+		"module":   module,
+	}
+	t.reportsCounter = t.stats.NewTaggedStat("reporting_flusher_get_reports_count", stats.HistogramType, tags)
+	t.aggReportsCounter = t.stats.NewTaggedStat("reporting_flusher_get_aggregated_reports_count", stats.HistogramType, tags)
+	return &t
+}
+
+func (a *TrackedUsersInAppAggregator) Aggregate(ctx context.Context, start, end time.Time) (jsonReports []json.RawMessage, err error) {
+	selectColumns := "reported_at, workspace_id, source_id, instance_id, userid_hll, anonymousid_hll, identified_anonymousid_hll"
+	query := fmt.Sprintf(`SELECT %s FROM tracked_users_reports WHERE reported_at >= $1 AND reported_at < $2 ORDER BY reported_at`, selectColumns)
```
lvrach, line 39: Either pass tableName and making dynamic or define it as const and use it for both queries and stats.
```suggestion

const tableName = `tracked_users_reports`

func NewTrackedUsersInAppAggregator(db *sql.DB, s stats.Stats, conf *config.Config, module string) *TrackedUsersInAppAggregator {
	t := TrackedUsersInAppAggregator{db: db, stats: s}
	tags := stats.Tags{
		"instance": conf.GetString("INSTANCE_ID", "1"),
		"table":    tableName,
		"module":   module,
	}
	t.reportsCounter = t.stats.NewTaggedStat("reporting_flusher_get_reports_count", stats.HistogramType, tags)
	t.aggReportsCounter = t.stats.NewTaggedStat("reporting_flusher_get_aggregated_reports_count", stats.HistogramType, tags)
	return &t
}

func (a *TrackedUsersInAppAggregator) Aggregate(ctx context.Context, start, end time.Time) (jsonReports []json.RawMessage, err error) {
	query := `SELECT reported_at, workspace_id, source_id, instance_id, userid_hll, anonymousid_hll, identified_anonymousid_hll FROM ` + tableName + `  WHERE reported_at >= $1 AND reported_at < $2 ORDER BY reported_at` 
```

enterprise/reporting/flusher/flusher.go
```diff
@@ -0,0 +1,348 @@
...
+
+	currentUTC := time.Now().UTC()
+	if start.IsZero() {
+		return 0
+	} else {
+		return currentUTC.Sub(start)
+	}
+}
+
+func (f *Flusher) FlushAggressively(ctx context.Context) bool {
```
mihir20, line 162:
```suggestion
func (f *Flusher) ShouldFlushAggressively(ctx context.Context) bool {
```

enterprise/reporting/flusher/flusher.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+	f.deleteReportsTimer.Since(s)
+
+	return nil
+}
+
+// Since we have hourly/daily/monthly aggregates on Reporting Service, we want the window to be within same hour
+// Don't consider most recent data where there are inserts happening
+// Always try to consider full window of flush interval or till current hour
+func (f *Flusher) getRange(ctx context.Context, currentUTC time.Time) (start, end time.Time, valid bool, err error) {
```
mihir20, line 220:
```suggestion
func (f *Flusher) getRange(ctx context.Context, currentUTC time.Time) (start, end time.Time, isFullFlushWindow bool, err error) {
```

enterprise/reporting/flusher/flusher.go
```diff
@@ -0,0 +1,348 @@
...
+
+func (f *Flusher) send(ctx context.Context, aggReports []json.RawMessage) error {
+	batchSize := f.batchSizeToReporting.Load()
+	concurrency := f.getConcurrency(ctx)
+
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(concurrency)
+	f.concurrentRequests.Gauge(float64(concurrency))
+
+	for i := 0; i < len(aggReports); i += batchSize {
```
mihir20, line 272: you can use lo.chunk https://github.com/samber/lo/blob/d93dd9a2d7d22037f0c301f8d91ac79f582477c4/slice.go#L168

mihir20 (approved):
LGTM

PR #4826, opened 2024-06-20
feat: implement an interface to collect data of tracked users

mihir20:
# Description
- POC Implementation of MTU tracking

Implementing an interface to collect data of tracked user's userID and anonymousID. This data will used to estimate number of unique users we are tracking. We will not store the raw userIDs and anonymousIDs directly for tracking count. We will store IDs in a probabilistic data structure called HyperLogLog. which will help us to estimate the cardinality.

For some events will have userIDs where user is known and for some events user is not known so we will have anonymousID. for some unknown users earlier which are also known know we will have both userID and anonymousID. Then count of unique users tracked will be equal to no of unique userID + number of unique anonymousID - number of unique combinations of (userID+anonymousID).

## Linear Ticket

pipe-1200

## Security

### DIFF
--- a/enterprise/trackedusers/data_collector.go
+++ b/enterprise/trackedusers/data_collector.go
@@ -23,9 +23,10 @@ import (
 
 	"github.com/tidwall/gjson"
 
+	"github.com/segmentio/go-hll"
+
 	"github.com/rudderlabs/rudder-server/jobsdb"
 	txn "github.com/rudderlabs/rudder-server/utils/tx"
-	"github.com/segmentio/go-hll"
 )
 
 const (
@@ -171,11 +172,11 @@ func hllToString(hll *hll.Hll) string {
 	return "\\x"
 }
 
-func combineUserIDAnonymousID(userID string, anonymousID string) string {
+func combineUserIDAnonymousID(userID, anonymousID string) string {
 	return userID + ":" + anonymousID
 }
 
-func (u *UniqueUsersCollector) recordIdentifier(idTypeHllMap map[string]*hll.Hll, identifier string, identifierType string) map[string]*hll.Hll {
+func (u *UniqueUsersCollector) recordIdentifier(idTypeHllMap map[string]*hll.Hll, identifier, identifierType string) map[string]*hll.Hll {
 	if idTypeHllMap == nil {
 		idTypeHllMap = make(map[string]*hll.Hll)
 	}
--- a/enterprise/trackedusers/data_collector.go
+++ b/enterprise/trackedusers/data_collector.go
@@ -1,0 +1,191 @@
+package trackedusers
+
+//go:generate mockgen -destination=./mocks/mock_data_collector.go -package=mockdatacollector github.com/rudderlabs/rudder-server/enterprise/trackedusers DataCollector
+
+import (
+	"context"
+	"database/sql"
+	"encoding/hex"
+	"fmt"
+	"time"
+
+	migrator "github.com/rudderlabs/rudder-server/services/sql-migrator"
+
+	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
+
+	"github.com/rudderlabs/rudder-go-kit/config"
+
+	"github.com/rudderlabs/rudder-go-kit/logger"
+
+	"github.com/lib/pq"
+
+	"github.com/spaolacci/murmur3"
+
+	"github.com/tidwall/gjson"
+
+	"github.com/rudderlabs/rudder-server/jobsdb"
+	txn "github.com/rudderlabs/rudder-server/utils/tx"
+	"github.com/segmentio/go-hll"
+)
+
+const (
+	idTypeUserID                       = "userID"
+	idTypeAnonymousID                  = "anonymousID"
+	idTypeUserIDAnonymousIDCombination = "userIDAnonymousIDCombination"
+
+	// changing this will be non backwards compatible
+	murmurSeed = 123
+)
+
+// DataCollector is interface to collect data from jobs
+type DataCollector interface {
+	CollectData(ctx context.Context, jobs []*jobsdb.JobT, tx *txn.Tx) error
+}
+
+type UniqueUsersCollector struct {
+	log         logger.Logger
+	hllSettings *hll.Settings
+	instanceID  string
+}
+
+func NewUniqueUsersCollector(log logger.Logger, dbConn string) (*UniqueUsersCollector, error) {
+	dbHandle, err := sql.Open("postgres", dbConn)
+	if err != nil {
+		panic(err)
+	}
+	dbHandle.SetMaxOpenConns(1)
+
+	m := &migrator.Migrator{
+		Handle:                     dbHandle,
+		MigrationsTable:            "tracked_users_reports_migrations",
+		ShouldForceSetLowerVersion: config.GetBool("SQLMigrator.forceSetLowerVersion", true),
+	}
+	err = m.Migrate("tracked_users")
+	if err != nil {
+		return nil, fmt.Errorf("could not run tracked_users_reports migrations: %w", err)
+	}
+
+	return &UniqueUsersCollector{
+		log: log,
+		hllSettings: &hll.Settings{
+			Log2m:             config.GetInt("TrackedUsers.precision", 14),
+			Regwidth:          config.GetInt("TrackedUsers.registerWidth", 5),
+			ExplicitThreshold: hll.AutoExplicitThreshold,
+			SparseEnabled:     true,
+		},
+		instanceID: config.GetString("INSTANCE_ID", "1"),
+	}, nil
+}
+
+func (u *UniqueUsersCollector) CollectData(ctx context.Context, jobs []*jobsdb.JobT, tx *txn.Tx) error {
+	if len(jobs) == 0 {
+		return nil
+	}
+	workspaceSourceUserIdTypeMap := make(map[string]map[string]map[string]*hll.Hll)
+	reportedAt := time.Now().UTC()
+	for _, job := range jobs {
+		if job.WorkspaceId == "" {
+			u.log.Warn("workspace_id not found in job", logger.NewIntField("jobId", job.JobID))
+			continue
+		}
+
+		sourceID := gjson.GetBytes(job.Parameters, "source_id").String()
+		if sourceID == "" {
+			u.log.Warn("source_id not found in job parameters", obskit.WorkspaceID(job.WorkspaceId),
+				logger.NewIntField("jobId", job.JobID))
+			continue
+		}
+		userID := gjson.GetBytes(job.EventPayload, "batch.0.userId").String()
+		anonymousID := gjson.GetBytes(job.EventPayload, "batch.0.anonymousId").String()
+
+		if userID == "" && anonymousID == "" {
+			u.log.Warn("both userID and anonymousID not found in job event payload", obskit.WorkspaceID(job.WorkspaceId),
+				logger.NewIntField("jobId", job.JobID))
+			continue
+		}
+
+		if workspaceSourceUserIdTypeMap[job.WorkspaceId] == nil {
+			workspaceSourceUserIdTypeMap[job.WorkspaceId] = make(map[string]map[string]*hll.Hll)
+		}
+
+		if userID != "" {
+			workspaceSourceUserIdTypeMap[job.WorkspaceId][sourceID] = u.recordIdentifier(workspaceSourceUserIdTypeMap[job.WorkspaceId][sourceID], userID, idTypeUserID)
+		}
+
+		if anonymousID != "" {
+			workspaceSourceUserIdTypeMap[job.WorkspaceId][sourceID] = u.recordIdentifier(workspaceSourceUserIdTypeMap[job.WorkspaceId][sourceID], anonymousID, idTypeAnonymousID)
+		}
+
+		if userID != "" && anonymousID != "" {
+			combinedUserIDAnonymousID := combineUserIDAnonymousID(userID, anonymousID)
+			workspaceSourceUserIdTypeMap[job.WorkspaceId][sourceID] = u.recordIdentifier(workspaceSourceUserIdTypeMap[job.WorkspaceId][sourceID], combinedUserIDAnonymousID, idTypeUserIDAnonymousIDCombination)
+		}
+	}
+
+	if len(workspaceSourceUserIdTypeMap) == 0 {
+		u.log.Warn("no data to collect", obskit.WorkspaceID(jobs[0].WorkspaceId))
+		return nil
+	}
+
+	stmt, err := tx.PrepareContext(ctx, pq.CopyIn("tracked_users_reports",
+		"workspace_id",
+		"instance_id",
+		"source_id",
+		"reported_at",
+		"userid_hll",
+		"anonymousid_hll",
+		"identified_anonymousid_hll",
+	))
+	if err != nil {
+		return fmt.Errorf("preparing statement: %v", err)
+	}
+	defer func() { _ = stmt.Close() }()
+
+	for workspaceID, sourceUserIdTypeMap := range workspaceSourceUserIdTypeMap {
+		for sourceID, userIdTypeMap := range sourceUserIdTypeMap {
+			_, err := stmt.Exec(workspaceID,
+				u.instanceID,
+				sourceID,
+				reportedAt,
+				hllToString(userIdTypeMap[idTypeUserID]),
+				hllToString(userIdTypeMap[idTypeAnonymousID]),
+				hllToString(userIdTypeMap[idTypeUserIDAnonymousIDCombination]),
+			)
+			if err != nil {
+				return fmt.Errorf("executing statement: %v", err)
+			}
+
+		}
+	}
+	if _, err = stmt.ExecContext(ctx); err != nil {
+		return fmt.Errorf("executing final statement: %v", err)
+	}
+	return nil
+}
+
+// convert hll to hexadecimal encoding compatible with Aggregate Knowledge HLL Storage Spec
+func hllToString(hll *hll.Hll) string {
+	if hll != nil {
+		return "\\x" + hex.EncodeToString(hll.ToBytes())
+	}
+	return "\\x"
+}
+
+func combineUserIDAnonymousID(userID string, anonymousID string) string {
+	return userID + ":" + anonymousID
+}
+
+func (u *UniqueUsersCollector) recordIdentifier(idTypeHllMap map[string]*hll.Hll, identifier string, identifierType string) map[string]*hll.Hll {
+	if idTypeHllMap == nil {
+		idTypeHllMap = make(map[string]*hll.Hll)
+	}
+	if idTypeHllMap[identifierType] == nil {
+		newHll, err := hll.NewHll(*u.hllSettings)
+		if err != nil {
+			panic(err)
+		}
+		idTypeHllMap[identifierType] = &newHll
+	}
+	idTypeHllMap[identifierType].AddRaw(murmur3.Sum64WithSeed([]byte(identifier), murmurSeed))
+	return idTypeHllMap
+}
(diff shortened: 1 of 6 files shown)


enterprise/trackedusers/data_collector.go
```diff
@@ -0,0 +1,192 @@
...
+	log         logger.Logger
+	hllSettings *hll.Settings
+	instanceID  string
+}
+
+func NewUniqueUsersCollector(log logger.Logger, dbConn string) (*UniqueUsersCollector, error) {
+	dbHandle, err := sql.Open("postgres", dbConn)
+	if err != nil {
+		panic(err)
+	}
```
lvrach, line 56: don't panic !
```suggestion
	dbHandle, err := sql.Open("postgres", dbConn)
	if err != nil {
		return err
	}
```

enterprise/trackedusers/data_collector.go
```diff
@@ -0,0 +1,192 @@
...
+
+	m := &migrator.Migrator{
+		Handle:                     dbHandle,
+		MigrationsTable:            "tracked_users_reports_migrations",
+		ShouldForceSetLowerVersion: config.GetBool("SQLMigrator.forceSetLowerVersion", true),
+	}
+	err = m.Migrate("tracked_users")
+	if err != nil {
+		return nil, fmt.Errorf("could not run tracked_users_reports migrations: %w", err)
+	}
```
lvrach, line 67: I would avoid running a migration as part of new.

We could instead:
1. Expose a separate method for running migration. Expecting users of the method to call it. 
2. or use sync.once on `CollectData` to ensure it run the first time CollectData call was made, but makes error handling a bit more complicated.

We should pass a context to migration as well.

enterprise/trackedusers/data_collector.go
```diff
@@ -0,0 +1,192 @@
...
+	return &UniqueUsersCollector{
+		log: log,
+		hllSettings: &hll.Settings{
+			Log2m:             config.GetInt("TrackedUsers.precision", 14),
+			Regwidth:          config.GetInt("TrackedUsers.registerWidth", 5),
+			ExplicitThreshold: hll.AutoExplicitThreshold,
+			SparseEnabled:     true,
+		},
+		instanceID: config.GetString("INSTANCE_ID", "1"),
+	}, nil
```
lvrach, line 78: pass a config.Config, don't use the global config package functions.

This will make it easier/safer to configure during tests. t.SetEnv is not safe on parallel executions.

enterprise/trackedusers/data_collector.go
```diff
@@ -0,0 +1,192 @@
+package trackedusers
```
lvrach, line 1: [subjective] I would go with mtu for package name. As it seems the dominant name used among business and engineering to refer to this feature. It is also short and unique.

```suggestion
package mtu
```

enterprise/trackedusers/data_collector.go
```diff
@@ -0,0 +1,192 @@
...
+
+	"github.com/rudderlabs/rudder-go-kit/logger"
+
+	"github.com/lib/pq"
+
+	"github.com/spaolacci/murmur3"
+
+	"github.com/tidwall/gjson"
+
+	"github.com/segmentio/go-hll"
```
lvrach, line 26: [minor] 3rd party deps can be grouped together
```suggestion
	"github.com/lib/pq"
	"github.com/spaolacci/murmur3"
	"github.com/tidwall/gjson"
	"github.com/segmentio/go-hll"
```

enterprise/trackedusers/data_collector.go
```diff
@@ -0,0 +1,192 @@
...
+	"fmt"
+	"time"
+
+	migrator "github.com/rudderlabs/rudder-server/services/sql-migrator"
+
+	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
+
+	"github.com/rudderlabs/rudder-go-kit/config"
+
+	"github.com/rudderlabs/rudder-go-kit/logger"
```
lvrach, line 18: [minor]
```suggestion
	migrator "github.com/rudderlabs/rudder-server/services/sql-migrator"
	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
	"github.com/rudderlabs/rudder-go-kit/config"
	"github.com/rudderlabs/rudder-go-kit/logger"
```

enterprise/trackedusers/data_collector.go
```diff
@@ -0,0 +1,192 @@
+package trackedusers
+
+//go:generate mockgen -destination=./mocks/mock_data_collector.go -package=mockdatacollector github.com/rudderlabs/rudder-server/enterprise/trackedusers DataCollector
```
lvrach, line 3: I would avoid using mockgen for now. 

It has simple enough interface that makes it easy to create a custom mock. Usually, the generated mock is not very convenient to use, and leading to partial testing in practise.

A good mock for the MTU collector could count the number of unique IDs per workspace per anon/user/identify call. Thus instead of asserting if method was called with the correct arguments, we can test the outcome. 

Also consider exposing a NOOP implementation, it would be useful for disabling the functionality and for tests that don't require testing mtu.

enterprise/trackedusers/data_collector.go
```diff
@@ -0,0 +1,192 @@
...
+	dbHandle.SetMaxOpenConns(1)
+
+	m := &migrator.Migrator{
+		Handle:                     dbHandle,
+		MigrationsTable:            "tracked_users_reports_migrations",
+		ShouldForceSetLowerVersion: config.GetBool("SQLMigrator.forceSetLowerVersion", true),
+	}
+	err = m.Migrate("tracked_users")
+	if err != nil {
+		return nil, fmt.Errorf("could not run tracked_users_reports migrations: %w", err)
```
lvrach, line 66: [minor]
```suggestion
		return nil, fmt.Errorf("migrating `tracked_users_reports` table: %w", err)
```

enterprise/trackedusers/data_collector.go
```diff
@@ -0,0 +1,192 @@
...
+	idTypeUserIDAnonymousIDCombination = "userIDAnonymousIDCombination"
+
+	// changing this will be non backwards compatible
+	murmurSeed = 123
+)
+
+// DataCollector is interface to collect data from jobs
+type DataCollector interface {
+	CollectData(ctx context.Context, jobs []*jobsdb.JobT, tx *txn.Tx) error
+}
```
lvrach, line 44: [subjective]
```suggestion
type UserTracker interface {
	Track(ctx context.Context, tx *txn.Tx,  jobs []*jobsdb.JobT) error
}
```

* `Data` is a bit generic, I do prefer the more applicable `User` term.
* Track sounds better than Collect, as we do keep any of the jobs rather use them to infer some data

enterprise/trackedusers/data_collector.go
```diff
@@ -0,0 +1,192 @@
...
+			continue
+		}
+
+		sourceID := gjson.GetBytes(job.Parameters, "source_id").String()
+		if sourceID == "" {
+			u.log.Warn("source_id not found in job parameters", obskit.WorkspaceID(job.WorkspaceId),
+				logger.NewIntField("jobId", job.JobID))
+			continue
+		}
+		userID := gjson.GetBytes(job.EventPayload, "batch.0.userId").String()
```
ktgowtham, line 99: Just double checking if each job corresponds to one event. and we're fetching userId at event level and not for a batch. There's a chance of a batch API call to have multiple userIds(like a backend service doing batch API calls). I'm assuming even if we store in gw jobs under batch, we store one event only. rt?
Also, we should skip non event stream sources(retl, etl etc)

mihir20: each job can have multiple events but events are batched per user. Adding filter for non event stream sources

enterprise/trackedusers/data_collector.go
mihir20: Track feels like we are tracking user activity
how about ?
```go
type UsersReporter interface {
	ReportUsers(ctx context.Context, reports []*UsersReport, tx *txn.Tx) error
}
```

enterprise/trackedusers/data_collector.go
mihir20: in this particular using mocks makes more sense as we just want to make sure that this methods are correctly called with correct parameters in unit tests, additional logic can be tested in integration tests.

enterprise/trackedusers/data_collector.go
mihir20: migrator currently do not takes context as a param, I can create a new task for adding support for context while running migrations.

ktgowtham (approved):

enterprise/trackedusers/users_reporter.go
```diff
@@ -0,0 +1,223 @@
...
+	idTypeUserIDAnonymousIDCombination = "userIDAnonymousIDCombination"
+
+	// changing this will be non backwards compatible
+	murmurSeed = 123
+)
+
+type UsersReport struct {
+	WorkspaceID              string
+	SourceID                 string
+	UserIDHll                *hll.Hll
```
lvrach, line 37: [minor] for consistancy
```suggestion
	UserIDHll                *hll.HLL
```

enterprise/trackedusers/users_reporter.go
```diff
@@ -0,0 +1,223 @@
...
+			hllToString(report.AnonymousIDHLL),
+			hllToString(report.IdentifiedAnonymousIDHLL),
+		)
+		if err != nil {
+			return fmt.Errorf("executing statement: %v", err)
+		}
+
+	}
+	if _, err = stmt.ExecContext(ctx); err != nil {
+		return fmt.Errorf("executing final statement: %v", err)
```
lvrach, line 193:
```suggestion
		return fmt.Errorf("executing final statement: %w", err)
```

enterprise/trackedusers/users_reporter.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		return fmt.Errorf("preparing statement: %v", err)
+	}
+	defer func() { _ = stmt.Close() }()
+
+	for _, report := range reports {
+		_, err := stmt.Exec(report.WorkspaceID,
+			u.instanceID,
+			report.SourceID,
+			time.Now(),
```
lvrach, line 182: It is better to DI `time.Now` for testing:

```suggestion
			u.Now(),
```

enterprise/trackedusers/users_reporter.go
```diff
@@ -0,0 +1,223 @@
...
+			hllToString(report.UserIDHll),
+			hllToString(report.AnonymousIDHLL),
+			hllToString(report.IdentifiedAnonymousIDHLL),
+		)
+		if err != nil {
+			return fmt.Errorf("executing statement: %v", err)
```
lvrach, line 188:
```suggestion
			return fmt.Errorf("executing statement: %w", err)
```

enterprise/trackedusers/users_reporter.go
```diff
@@ -0,0 +1,223 @@
...
+		"workspace_id",
+		"instance_id",
+		"source_id",
+		"reported_at",
+		"userid_hll",
+		"anonymousid_hll",
+		"identified_anonymousid_hll",
+	))
+	if err != nil {
+		return fmt.Errorf("preparing statement: %v", err)
```
lvrach, line 174:
```suggestion
		return fmt.Errorf("preparing statement: %w", err)
```

enterprise/trackedusers/users_reporter.go
```diff
@@ -0,0 +1,223 @@
...
+		reports = append(reports, sourceIDReports...)
+	}
+	return reports
+}
+
+func (u *UniqueUsersReporter) ReportUsers(ctx context.Context, reports []*UsersReport, tx *txn.Tx) error {
+	if len(reports) == 0 {
+		return nil
+	}
+	stmt, err := tx.PrepareContext(ctx, pq.CopyIn("tracked_users_reports",
```
lvrach, line 164: Should we move this to a constant?
```suggestion
	stmt, err := tx.PrepareContext(ctx, pq.CopyIn(trackUsersTable,
```

enterprise/trackedusers/users_reporter.go
```diff
@@ -0,0 +1,223 @@
...
+				WorkspaceID:              workspaceID,
+				SourceID:                 sourceID,
+				UserIDHll:                userIdTypeMap[idTypeUserID],
+				AnonymousIDHLL:           userIdTypeMap[idTypeAnonymousID],
+				IdentifiedAnonymousIDHLL: userIdTypeMap[idTypeUserIDAnonymousIDCombination],
+			}
+		})...)
+		reports = append(reports, sourceIDReports...)
+	}
+	return reports
```
lvrach, line 157: Isn't this equivalent and simpler? Is there a performance reason?
```suggestion
	reports := make([]*UsersReport, 0)
	for workspaceID, sourceUserMp := range workspaceSourceUserIdTypeMap {
		reports = append(reports, lo.MapToSlice(sourceUserMp, func(sourceID string, userIdTypeMap map[string]*hll.Hll) *UsersReport {
			return &UsersReport{
				WorkspaceID:              workspaceID,
				SourceID:                 sourceID,
				UserIDHll:                userIdTypeMap[idTypeUserID],
				AnonymousIDHLL:           userIdTypeMap[idTypeAnonymousID],
				IdentifiedAnonymousIDHLL: userIdTypeMap[idTypeUserIDAnonymousIDCombination],
			}
		})...)
	}
	return reports
```

enterprise/trackedusers/users_reporter_test.go
```diff
@@ -0,0 +1,304 @@
...
+		EventPayload: []byte(`{"receivedAt":"2021-06-06T20:26:39.598+05:30","writeKey":"writeKey","requestIP":"[::1]",  "batch": [{"anonymousId":"anon_id","channel":"android-sdk","context":{"app":{"build":"1","name":"RudderAndroidClient","namespace":"com.rudderlabs.android.sdk","version":"1.0"},"device":{"id":"49e4bdd1c280bc00","manufacturer":"Google","model":"Android SDK built for x86","name":"generic_x86"},"library":{"name":"com.rudderstack.android.sdk.core"},"locale":"en-US","network":{"carrier":"Android"},"screen":{"density":420,"height":1794,"width":1080},"traits":{"anonymousId":"49e4bdd1c280bc00"},"user_agent":"Dalvik/2.1.0 (Linux; U; Android 9; Android SDK built for x86 Build/PSR1.180720.075)"},"event":"Demo Track","integrations":{"All":true},"messageId":"b96f3d8a-7c26-4329-9671-4e3202f42f15","originalTimestamp":"2019-08-12T05:08:30.909Z","properties":{"category":"Demo Category","floatVal":4.501,"label":"Demo Label","testArray":[{"id":"elem1","value":"e1"},{"id":"elem2","value":"e2"}],"testMap":{"t1":"a","t2":4},"value":5},"rudderId":"a-292e-4e79-9880-f8009e0ae4a3","sentAt":"2019-08-12T05:08:30.909Z","type":"track"}]}`),
```
lvrach, line 36: Can we avoid this repetition here? Can you use `prepareJob` Are all the fields needed for the test?
If it is important to have a full payload, we can move it on a json file in testdata folder.

enterprise/trackedusers/users_reporter_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+				rows, err := postgresContainer.DB.Query("SELECT workspace_id, source_id, userid_hll, anonymousid_hll, identified_anonymousid_hll FROM tracked_users_reports")
+				require.NoError(t, err)
+				require.NoError(t, rows.Err())
+				defer func() { _ = rows.Close() }()
+				var entry UsersReport
+				entries := make([]UsersReport, 0)
+				for rows.Next() {
+					var userIDHllStr, annIDHllStr, combHllStr string
+					err = rows.Scan(&entry.WorkspaceID, &entry.SourceID, &userIDHllStr, &annIDHllStr, &combHllStr)
```
lvrach, line 272: Just for completion let's test reportedAt by mocking/DI `now`

enterprise/trackedusers/users_reporter_test.go
```diff
@@ -0,0 +1,304 @@
...
+						result[e.WorkspaceID] = make(map[string]int)
+					}
+					if e.UserIDHll != nil {
+						result[e.WorkspaceID][e.SourceID] += int(e.UserIDHll.Cardinality())
+					}
+					if e.AnonymousIDHLL != nil {
+						result[e.WorkspaceID][e.SourceID] += int(e.AnonymousIDHLL.Cardinality())
+					}
+					if e.IdentifiedAnonymousIDHLL != nil {
+						result[e.WorkspaceID][e.SourceID] -= int(e.IdentifiedAnonymousIDHLL.Cardinality())
```
lvrach, line 214:
```suggestion
						result[e.WorkspaceID][e.SourceID] += int(e.IdentifiedAnonymousIDHLL.Cardinality())
```

enterprise/trackedusers/users_reporter_test.go
```diff
@@ -0,0 +1,304 @@
...
+				trackedUsers: map[string]map[string]int{},
+			},
+		}
+
+		for _, tc := range testCases {
+			t.Run(tc.name, func(t *testing.T) {
+				collector := &UniqueUsersReporter{log: logger.NOP, hllSettings: &hllSettings}
+				reports := collector.GenerateReportsFromJobs(tc.jobs, tc.sourceIDtoFilter)
+
+				result := make(map[string]map[string]int)
```
lvrach, line 202: It would be preferable to explicitly define your result instead of generating it. It is only a few fields, it would make your test cases more readable and easy to reason about.

Also, it prevents the same bugs in both generation and code from being hidden.

lvrach (approved):
Overall, things look great. I like the approach.

I left some non-blocking comments. I would like to emphasise the ones regarding testing.

enterprise/trackedusers/users_reporter.go
mihir20: yeah, `sourceIDReports := make([]*UsersReport, 0, len(sourceUserMp))` this is reducing mallocs and slice resizing. Not a significant optimisation though as we have to append to reports slice for which size is not known

mihir20:
```
BenchmarkUniqueUsersReporter_GenerateReportsFromJobs/GenerateReportsFromJobs-12         	      87	  11674817 ns/op	  254280 B/op	   30174 allocs/op
BenchmarkUniqueUsersReporter_GenerateReportsFromJobs/GenerateReportsFromJobs-new-12     	     100	  12442578 ns/op	  254040 B/op	   30169 allocs/op
```

enterprise/trackedusers/users_reporter.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+//go:generate mockgen -destination=./mocks/mock_user_reporter.go -package=mockuserreporter github.com/rudderlabs/rudder-server/enterprise/trackedusers UsersReporter
+
+// UsersReporter is interface to report unique users from reports
+type UsersReporter interface {
+	ReportUsers(ctx context.Context, reports []*UsersReport, tx *txn.Tx) error
+	GenerateReportsFromJobs(jobs []*jobsdb.JobT, sourceIdFilter map[string]bool) []*UsersReport
+	MigrateDatabase(dbConn string, conf *config.Config) error
+}
```
lvrach, line 51: Should we define this interface on the consumer instead?

Do we need the generated mock?

```suggestion
// UsersReporter is interface to report unique users from reports
type UsersReporter interface {
	ReportUsers(ctx context.Context, reports []*UsersReport, tx *txn.Tx) error
	GenerateReportsFromJobs(jobs []*jobsdb.JobT, sourceIdFilter map[string]bool) []*UsersReport
}
```
Should MigrateDatabase be part of the interface for processor?

lvrach (approved):

enterprise/trackedusers/users_reporter.go
mihir20: - MigrateDatabase need to be part of this interface as we are running migration as a part of app startup.
- Do we need the generated mock? - removed mock generator
- Should we define this interface on the consumer instead? we need 1 interface here as well as factory can return noop implementation as well, more in the next PR https://github.com/rudderlabs/rudder-server/pull/4827

PR #4827, opened 2024-06-20
feat: push events data from processor to unique users data collector

mihir20:
- POC Implementation of MTU tracking

base pr: https://github.com/rudderlabs/rudder-server/pull/4826

# Description
push events to uniques users data collector, data collector will create hlls and save it in the database. Later reporting service will read the data from the DB, aggregate it and send it to reporting system

## Linear Ticket

pipe-1201

## Security


processor/processor.go
```diff
@@ -2360,6 +2372,13 @@ func (proc *Handle) Store(partition string, in *storeMessage) {
 				}
 			}
 
+			if proc.isTrackedUsersCollectionEnabled() {
```
lvrach, line 2375:
```suggestion
```
No need for a check here, we can use Nop implementation to disable MTU tracking without adding this logic inside processor.

enterprise/trackedusers/factory.go
```diff
@@ -0,0 +1,11 @@
+package trackedusers
+
+import "github.com/rudderlabs/rudder-go-kit/logger"
+
+type Factory struct {
+	Log logger.Logger
+}
+
+func (f *Factory) Setup(dbConn string) (DataCollector, error) {
```
lvrach, line 9: Potential place to decide if user tracking will be enabled

processor/processor.go
```diff
@@ -2154,6 +2165,7 @@ func (proc *Handle) transformations(partition string, in *transformationMessage)
 }
 
 type storeMessage struct {
+	gatewayJobs   []*jobsdb.JobT
```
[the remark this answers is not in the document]
ktgowtham: I agree with @lvrach on this. We can explore computing hll in `processJobsForDest` stage. similar to some reporting metrics that are being computed there. and pass the computed value to the store layer and store this in a transaction.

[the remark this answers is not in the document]
mihir20: went ahead with passing the computed HLLs, tried to think of some other approach as well but couldn't find any better approach

processor/processor.go
```diff
@@ -1656,6 +1665,10 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 			continue
 		}
 
+		if source.SourceDefinition.Category != "" && !strings.EqualFold(source.SourceDefinition.Category, sourceCategoryWebhook) {
```
ktgowtham, line 1668: `webhook` cateogory sources are event stream sources. We shouldn't be filtering `webhook`.
 rETL and ETL has `warehouse` and `singer-protocol`

mihir20: yeah we are only selecting sources where category is empty string or `webhook` rest all we are ignoring

ktgowtham: my bad. read the code wrong

ktgowtham (approved):

processor/processor.go
```diff
@@ -1583,6 +1590,8 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 	// map of jobID to destinationID: for messages that needs to be delivered to a specific destinations only
 	jobIDToSpecificDestMapOnly := make(map[int64]string)
 
+	nonEventStreamSourceIds := make(map[string]bool)
```
cisse21, line 1593: Let's compute this once in `configBackendSubscriber` and use it across the code? Any issues with this approach ?

processor/processor.go
```diff
@@ -2018,6 +2031,7 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 
 		subJobs.hasMore,
 		subJobs.rsourcesStats,
+		proc.trackedUsersReporter.GenerateReportsFromJobs(subJobs.subJobs, nonEventStreamSourceIds),
```
cisse21, line 2034:
```suggestion
		proc.trackedUsersReporter.GenerateReportsFromJobs(jobList, nonEventStreamSourceIds),
```

cisse21 (approved):
Added minor comments

processor/processor.go
mihir20: Its better to keep it in memory instead of calculating again and again. Assuming that data won't be too much to keep in memory.

PR #4836, opened 2024-06-24
chore: webhook integration tests

lvrach:
Introducing integration tests for webhook sources.

## Motivation

They were a couple of regression originated from changes in rudder-server code related to webhooks.

Increase confidence of pipeline and integration team releases, whilst keeping test reliable, maintainable and fast. 

## Implemenation

* Use actual transformer 
* Expand the scope of the testing surface:
   * input: HTTP request to rudder-server
   * output: HTTP response, payload written to gatewayDB and proc_err (if any)
* Use JSON files for test cases (as table-driven test) to promote portability and maintainability 
* Create a new source and workspace per test case to isolate failures and allow parallel test execution.
* Delete [testhelper/destination/transformer.go](https://github.com/rudderlabs/rudder-server/pull/4836/files#diff-553822f1b490cf2ae001fe2bb7dca1f8389f0e31a6ffb647b393fbf5bed8b35f) in favour of rudder-go-kit one.

### necessary evils

* I had to use the full gateway component, ideally only webhook component should have been enough, but there is a high dependency between them. 
* I wanted to avoid spinning Postgres and use jobsDB, but there is a complex dependency with gateway and resources.

### Future notes  

* Currently testcases for two destinations have been added (shopify, slack)
* Plan is to generate all test cases from transformer repository, instead of manually recreating them here. Allowing us to maintain a single source of truth
* There a few more things I would like to do in a followup iteration:
   1. some improvements for config builder, so I don't have to inject workspaceId manually. Ideally workspace builder should generate source and destinations.
   2. refactor gateway Setup, use more optimal params for DI to reduce number of arguments and ease testing.

## Linear Ticket

https://linear.app/rudderstack/issue/PRI-4/webhook-integration-tests

## Security


koladilip (approved):

gateway/webhook/integration_test.go
```diff
@@ -0,0 +1,361 @@
...
+
+				assert.JSONEq(t, string(p), string(batch.Batch[0]))
+			}
+
+			r, err = errDB.GetUnprocessed(ctx, jobsdb.GetQueryParams{
+				WorkspaceID: workspaceID,
+				// ParameterFilters: []jobsdb.ParameterFilterT{{
+				// 	Name:  "source_id",
+				// 	Value: sourceID,
+				// }},
```
fracasula, line 289:
```suggestion
```

gateway/webhook/integration_test.go
```diff
@@ -0,0 +1,361 @@
...
+				JobsLimit: 1,
+			})
+			require.NoError(t, err)
+			assert.Len(t, r.Jobs, len(tc.Output.ErrQueue))
```
fracasula, line 293: Out of curiosity, how come sometimes you use `require` and sometimes `assert`? :thinking:

gateway/webhook/integration_test.go
```diff
@@ -0,0 +1,361 @@
...
+	})
+	t.Setenv("DEST_TRANSFORM_URL", transformerContainer.TransformerURL)
+
+	<-transformerFeaturesService.Wait()
+
+	bcs := make(map[string]backendconfig.ConfigT)
+
+	testSetup := loadTestSetup(t)
+
+	for i, tc := range testSetup.cases {
```
fracasula, line 160: [optional] I found this a bit weird that you're iterating over the test cases twice. Maybe you can create a function for this an call it at the beginning of each iteration in the other loop?

fracasula (approved):

gateway/webhook/integration_test.go
```diff
@@ -0,0 +1,361 @@
...
+			}
+
+			r, err = errDB.GetUnprocessed(ctx, jobsdb.GetQueryParams{
+				WorkspaceID: workspaceID,
+				// ParameterFilters: []jobsdb.ParameterFilterT{{
+				// 	Name:  "source_id",
+				// 	Value: sourceID,
+				// }},
+				JobsLimit: 1,
+			})
```
sanpj2292, line 291: Why aren't we querying with `source_id` in errors DB ?

sanpj2292 (approved):

gateway/webhook/integration_test.go
lvrach: We need to prepare backendconfig before initialising the service

gateway/webhook/integration_test.go
lvrach: debug artefact, uncomment the code

gateway/webhook/integration_test.go
lvrach: debug artifact, uncommented the code

gateway/webhook/integration_test.go
lvrach: I wanted to start introducing assert in my test to make the failed output a bit more useful, by having multiple assert logs instead of a single one. I kept the `require` for more severe errors. I understand the readability issue from inconsistency, need to reconsider this.

PR #4848, opened 2024-06-28
chore: add stage to transformer request time metric

cisse21:
# Description

- Added stage in transformer request time metric
- Remove unused fastHttp package

## Linear Ticket

Fixes PIPE-1243

## Security


processor/transformer/transformer.go
```diff
@@ -266,12 +252,7 @@ func NewTransformer(conf *config.Config, log logger.Logger, stat stats.Stats, op
 
 // Transform function is used to invoke destination transformer API
 func (trans *handle) Transform(ctx context.Context, clientEvents []TransformerEvent, batchSize int) Response {
-	if len(clientEvents) == 0 {
-		return Response{}
-	}
-	destType := clientEvents[0].Destination.DestinationDefinition.Name
-	transformURL := trans.destTransformURL(destType)
-	return trans.transform(ctx, clientEvents, transformURL, batchSize, destTransformerStage)
+	return trans.transform(ctx, clientEvents, trans.destTransformURL(clientEvents[0].Destination.DestinationDefinition.Name), batchSize, destTransformerStage)
```
ktgowtham, line 255: Shouldn't it be safe to check the length? even if the current caller ensures a empty slice is not sent?

ktgowtham (approved):
LGTM. added a minor comment. fasthttp client didn't help and hence we're removing it?

processor/transformer/transformer.go
cisse21: We have a check in the caller [here](https://github.com/rudderlabs/rudder-server/blob/master/processor/processor.go#L2662)

We can be safe but we should be safe even in `UserTransform` and `Validate` too as they are similar. Felt removing is the best way forward

cisse21:
> LGTM. added a minor comment. fasthttp client didn't help and hence we're removing it?

We had a CFD after enabling `fasthttp` with transformer receiving invalid json for a very small percentage of requests

processor/transformer/transformer.go
```diff
@@ -150,16 +146,15 @@ func GetVersion() string {
...
-			return "Unable to read response from transformer."
-		}
-
-		return string(bodyBytes)
+	if resp.StatusCode != http.StatusOK {
+		return defaultBuildVersion
+	}
+	bodyBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		fmt.Println("Unable to read response into bytes with error : ", err.Error())
```
lvrach, line 154: Should we use a logger here?

achettyiitr (approved):

lvrach (approved):

processor/transformer/transformer.go
cisse21: I wanted to get rid of this function too if we were okay with it

processor/transformer/transformer.go
sivashanmukh: Just looking at this function: It is just doing a pass through. Couldn't we just get rid of this and rename trans.transform to trans.Transform?

processor/transformer/transformer.go
sivashanmukh: I didn't see transformer_build_version set in either of user_transformer or dest_transformer pods enviornments in multi-tenant rudder-transformer and user-transformer. Not sure if its being set from elsewhere. 

Can potentially get rid of this altogether 👍

processor/transformer/transformer.go
cisse21: @sivashanmukh will take it up in a new PR. Will merge this PR for now

PR #4855, opened 2024-07-02
chore: add missing transformer sent stat

cisse21:
# Description

Add missing `sentStat` in `transformer.go`

Right now if the response from transformer has mismatch in the contract we expect we end up discarding those jobs to proc errors but ideally there should be no breach in contract and if there is a breach in contract we should stop processing and fix the same. Since the current code structure does not allow propagating errors, adding a panic to stop processing

## Linear Ticket

Fixes PIPE-1248

## Security


processor/transformer/transformer.go
```diff
@@ -395,27 +393,22 @@ func (trans *handle) request(ctx context.Context, url, stage string, data []Tran
...
+		// Panic the processor to avoid replays
 		if err != nil {
 			trans.logger.Errorf("Data sent to transformer : %v", string(rawJSON))
 			trans.logger.Errorf("Transformer returned : %v", string(respData))
-
-			respData = []byte(fmt.Sprintf("Failed to unmarshal transformer response: %s", string(respData)))
-
-			transformerResponses = nil
-			statusCode = http.StatusBadRequest
+			panic(err)
```
fracasula, line 409: PR description and title are quite misleading, right? You're not just adding a missing stat.
How come we are panicking now? Can you add some context in the PR description please?

cisse21: I just added some clean up too .. Will add description for the same

processor/transformer/transformer.go
```diff
@@ -270,6 +270,7 @@ func (trans *handle) transform(
 		trackWg.Done()
 	}()
 
+	trans.sentStat.Count(len(clientEvents))
```
Sidddddarth, line 273: we could do this after actually sending the events

Sidddddarth (approved):

ktgowtham (approved):
LGTM.

PR #4856, opened 2024-07-02
chore: add metrics for unique users reporter

mihir20:
# Description

Add following metrics in rudder server

- Latency of generating HLLs

- Size in each HLL inserted in the outbox table

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1245/add-metrics-for-rudder-server

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -206,6 +206,7 @@ type processorStats struct {
 	processJobThroughput          func(partition string) stats.Measurement
 	transformationsThroughput     func(partition string) stats.Measurement
 	DBWriteThroughput             func(partition string) stats.Measurement
+	trackedUsersReportGeneration  func(partition string) stats.Measurement
 }
 
 type DestStatT struct {
@@ -604,6 +605,11 @@ func (proc *Handle) Setup(
 	}
 	proc.stats.DBWriteThroughput = func(partition string) stats.Measurement {
 		return proc.statsFactory.NewTaggedStat("processor_db_write_throughput", stats.CountType, stats.Tags{
+			"partition": partition,
+		})
+	}
+	proc.stats.trackedUsersReportGeneration = func(partition string) stats.Measurement {
+		return proc.statsFactory.NewTaggedStat("processor_tracked_users_report_gen_time", stats.TimerType, stats.Tags{
 			"partition": partition,
 		})
 	}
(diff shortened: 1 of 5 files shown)


processor/processor.go
```diff
@@ -607,6 +608,11 @@ func (proc *Handle) Setup(
 			"partition": partition,
 		})
 	}
+	proc.stats.trackedUsersReportGeneration = func(partition string) stats.Measurement {
+		return proc.statsFactory.NewTaggedStat("processor_tracked_users_report_gen_time", stats.TimerType, stats.Tags{
```
lvrach, line 612: [minor] from prometheus naming best practises, postfix with the unit
```suggestion
		return proc.statsFactory.NewTaggedStat("processor_tracked_users_report_gen_seconds", stats.TimerType, stats.Tags{
```

runner/buckets.go
```diff
@@ -108,5 +108,13 @@ var (
...
 			0.1, 0.5, 1, 5, 60, 300, 600, 1800, 3600, 12 * 3600, 24 * 3600,
 		},
+		"tracked_users_hll_bytes": {
+			float64(10 * bytesize.B),  // for hll containing single id = 8B
+			float64(100 * bytesize.B), // for hll containing 10 ids = 80B
+			float64(1 * bytesize.KB),  // for hll containing 100 ids = 800B
+			float64(3 * bytesize.KB),  // for hll containing 300 ids = 2400B
+			float64(6 * bytesize.KB),  // for hll containing 700 ids = 5600B
+			float64(10 * bytesize.KB), // max size for hll with log2m=14, regWidth=5 = 10KB
+		},
```
lvrach, line 118: Will need to fine tune the buckets for this one I believe
```suggestion
		},
		"processor_tracked_users_report_gen_seconds": {
		     // 1ms, 10ms, 50ms, 100ms, 0.5s, 1s
		}
```

mihir20: yes this is in micro seconds, I will update this.

mihir20: For now we can keep this metric for some time, we can decide and remove later after checking in prod. May be in next server release?

runner/buckets.go
```diff
@@ -108,5 +108,19 @@ var (
...
 		},
+		"tracked_users_hll_bytes": {
+			float64(10 * bytesize.B),  // for hll containing single id = 8B
+			float64(100 * bytesize.B), // for hll containing 10 ids = 80B
+			float64(400 * bytesize.B), // for hll containing 50 ids = 400B
+			float64(1 * bytesize.KB),  // for hll containing 100 ids = 800B
+			float64(3 * bytesize.KB),  // for hll containing 300 ids = 2400B
+			float64(6 * bytesize.KB),  // for hll containing 700 ids = 5600B
+			float64(10 * bytesize.KB), // max size for hll with log2m=14, regWidth=5 = 10KB
+			float64(41 * bytesize.KB), // max size for hll with log2m=16, regWidth=5 = 40KB
```
satishrudderstack, line 119: @mihir20 Are we able to pick these buckets for the metric. In the load test I remember it was not using these buckets ?

satishrudderstack (approved):
LGTM

runner/buckets.go
mihir20: In load test we were using statsd for metrics which do not support custom buckets, opentelemetry supports custom bucket but we metrics were not flowing when we enabled otel in devcluster.

lvrach (approved):

PR #4859, opened 2024-07-02
feat: integrating pgx for postgres compatible drivers

achettyiitr:
# Description

- integrating [pgx](https://github.com/jackc/pgx) for postgres compatible drivers in these packages:
  - warehouse.
  - notifier.
  - integrations/postgres.
  - integrations/redshift.
 - Skipped tests other then warehouse because we did some changes in reporting which is only compatible with stdlib.
 - Corresponding go-kit pgx changes. https://github.com/rudderlabs/rudder-go-kit/pull/552

## Linear Ticket

-  Resolves PIPE-1236

## Security

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -10,36 +10,6 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
   cancel-in-progress: true
 jobs:
-  integration:
-    name: Integration
-    runs-on: 'ubuntu-20.04'
-    strategy:
-      matrix:
-        FEATURES: [ oss ,enterprise ]
-    steps:
-      - name: Disable IPv6 (temporary fix)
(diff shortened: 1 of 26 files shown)


testhelper/docker/resource/postgres/postgres.go
```diff
@@ -0,0 +1,157 @@
...
+		// 2. create a sql.DB and verify connection
+		if db, err = sql.Open("postgres", dbDSN); err != nil {
+			return fmt.Errorf("opening database: %w", err)
+		}
+		pgxConf, err := pgx.ParseConfig(dbDSN)
+		if err != nil {
+			return fmt.Errorf("parsing pgx config: %w", err)
+		}
+
+		db := stdlib.OpenDB(*pgxConf)
```
achettyiitr, line 127: This is copied from go-kit with this specific change only.

PR #4860, opened 2024-07-02
chore: add integration tests for mtu reporting

mihir20:
# Description

Add integration tests for tracked users reporting

## Linear Ticket

pipe-1202

## Security

### DIFF
--- a/integration_test/trackedusersreporting/tracked_users_reporting_test.go
+++ b/integration_test/trackedusersreporting/tracked_users_reporting_test.go
@@ -116,17 +116,17 @@ func getCardinalityFromDB(t *testing.T, postgresResource *postgres.Resource) map
 		if result[e.WorkspaceID] == nil {
 			result[e.WorkspaceID] = make(map[string]int)
 		}
-		userHllBytes, err := hex.DecodeString(e.userIDHll[2:])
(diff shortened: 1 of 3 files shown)


ktgowtham (approved):

integration_test/trackedusersreporting/tracked_users_reporting_test.go
```diff
@@ -0,0 +1,298 @@
...
+	}, 1*time.Minute, 5*time.Second, "unexpected number of events received, count of events: %d", tc.webhook.RequestsCount())
+
+	// TODO: once flusher is implemented, add a mock reporting server to check is we are getting correct cardinality of users
+	cardinalityMap := getCardinalityFromDB(t, tc.postgresResource)
+	require.Equal(t, 2, cardinalityMap[workspaceID][sourceID])
+	cancel()
+	require.NoError(t, wg.Wait())
+}
+
+func getCardinalityFromDB(t *testing.T, postgresResource *postgres.Resource) map[string]map[string]int {
```
lvrach, line 95: This might stop working when flusher code gets introduced. Given that fluster will get and then delete them from DB. So we will need to adapt the test to mock the reporting server instead of reading the DB.

lvrach (approved):

PR #4887, opened 2024-07-10
chore: flipping SourceID & OriginalSourceID during transformation

BonapartePC:
# Description

In this PR, we are flipping the `OriginalSourceID` and `SourceID` during the transformation and flipping them back once the transformation is done. This is to support replays where transformation can be based on the SourceId of a particular source which we are trying to replay.

## Security


fracasula:
There is no need to merge this one, right? Shall we convert it to draft just in case?

cisse21:
It is better we merge this. Since we have a lot of customer using sourceID in their user transformations

processor/transformer/transformer.go
```diff
@@ -230,7 +231,33 @@ func (trans *handle) Transform(ctx context.Context, clientEvents []TransformerEv
...
+	response := trans.transform(ctx, clientEvents, trans.userTransformURL(), batchSize, userTransformerStage)
+	for _, event := range response.Events {
+		// flip back sourceID and original sourceID
+		if event.Metadata.OriginalSourceID != "" {
+			originalSourceID := event.Metadata.SourceID
+			event.Metadata.SourceID = event.Metadata.OriginalSourceID
+			event.Metadata.OriginalSourceID = originalSourceID
+		}
+	}
+	for _, event := range response.FailedEvents {
```
Sidddddarth, line 252: Both these `flip back`s can be moved to the `(*handle)transform(...) Response` method at the end where we populate the `Response`(`failedEvents` and `outClientEvents`).

Sidddddarth (approved):

Sidddddarth:
Also change the description and PR name since we're going to merge it anyway.

cisse21 (approved):

BonapartePC:
@coderabbitai summary

PR #4890, opened 2024-07-11
chore: update default mtu config

mihir20:
# Description

- Update default value for MTU precision to 16
- Set timezone for reported_at in UTC as flusher is expecting time in UTC timezone. currently we are storing timezone in the DB so it should not an issue. 

## Security

### DIFF
--- a/enterprise/trackedusers/users_reporter.go
+++ b/enterprise/trackedusers/users_reporter.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/rudderlabs/rudder-server/utils/timeutil"
+
 	"github.com/rudderlabs/rudder-go-kit/stats"
 
 	"github.com/rudderlabs/rudder-server/jobsdb"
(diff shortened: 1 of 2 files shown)


ktgowtham (approved):

enterprise/trackedusers/users_reporter.go
```diff
@@ -62,15 +62,15 @@ func NewUniqueUsersReporter(log logger.Logger, conf *config.Config, stats stats.
...
+			Log2m:             conf.GetInt("TrackedUsers.precision", 16),
 			Regwidth:          conf.GetInt("TrackedUsers.registerWidth", 5),
 			ExplicitThreshold: hll.AutoExplicitThreshold,
 			SparseEnabled:     true,
 		},
 		instanceID: config.GetString("INSTANCE_ID", "1"),
 		stats:      stats,
 		now: func() time.Time {
-			return time.Now()
+			return time.Now().UTC()
```
achettyiitr, line 73: We can use timeutil package.
```GO
func NewUniqueUsersReporter(log logger.Logger, conf *config.Config, stats stats.Stats) (*UniqueUsersReporter, error) {
	return &UniqueUsersReporter{
		log: log,
		hllSettings: &hll.Settings{
			Log2m:             conf.GetInt("TrackedUsers.precision", 14),
			Regwidth:          conf.GetInt("TrackedUsers.registerWidth", 5),
			ExplicitThreshold: hll.AutoExplicitThreshold,
			SparseEnabled:     true,
		},
		instanceID: config.GetString("INSTANCE_ID", "1"),
		stats:      stats,
		now:        timeutil.Now,
	}, nil
}
```

achettyiitr (approved):

PR #4891, opened 2024-07-12
chore: include extra columns for errors reporting table

sanpj2292:
# Description

- Adding columns for `errors` table
   - sampleResponse
   - sampleEvent
   - eventType
   - eventName

## Linear Ticket

Resolves INT-1056

## Security


lvrach (approved):

sql/migrations/error_detail_reports/000002_alter_error_detail_reports.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE error_detail_reports
+ADD COLUMN IF NOT EXISTS sample_event JSONB DEFAULT '{}'::JSONB,
```
lvrach, line 2: No a blocker, but an FYI:

We are considering  using TEXT, if we are not planning to query the json in the database, to avoid potential parsing/validation issues with POSTGRES.

```suggestion
ADD COLUMN IF NOT EXISTS sample_event TEXT DEFAULT '{}',
```

lvrach (approved):

cisse21 (approved):

sql/migrations/error_detail_reports/000002_alter_error_detail_reports.up.sql
sanpj2292: > if we are not planning to query the json in the database

We are looking to query json in DB 😅

ktgowtham:
@satishrudderstack , do we need to worry about anything on reporting side wrt size? cc: @sanpj2292

satishrudderstack:
> @satishrudderstack , do we need to worry about anything on reporting side wrt size? cc: @sanpj2292

@ktgowtham @sanpj2292 Right now, the `errors` table on reporting is ~40GB with 31 days retention. With `event_name` cardinality and `sample_event`, I think it would grow big. I think we can enable and keep an eye on the table size on TimescaleDB.

PR #4899, opened 2024-07-16
chore: add handling for alias events in tracked users report

mihir20:
# Description

We will add previousID that we received in the identifiedAnonID bucket so that it will be substracted from the total count. e.g

```js
e.g. we receive events
		{type:track, anonymousID: anon1}
		{type:track, userID: user1}
		{type:track, userID: user2}
		{type:identify, userID: user1, anonymousID: anon1}
		{type:alias, previousId: user2, userID: user1}
		
		userHLL: {user1, user2}, 
		anonHLL: {anon1}, 
		identifiedAnonHLL: {user1-anon1, user2}
		
		cardinality: len(userHLL)+len(anonHLL)-len(identifiedAnonHLL): 2+1-2 = 1
```

https://www.rudderstack.com/docs/event-spec/standard-events/alias/

ref: https://rudderlabs.slack.com/archives/C075KC5G81G/p1720670222299599

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1305/add-handling-for-alias-events-in-mtu

## Security


enterprise/trackedusers/users_reporter.go
```diff
@@ -143,6 +145,23 @@ func (u *UniqueUsersReporter) GenerateReportsFromJobs(jobs []*jobsdb.JobT, sourc
...
+		// {type:track, anonymousID: anon1}
+		// {type:track, userID: user1}
+		// {type:track, userID: user2}
+		// {type:identify, userID: user1, anonymousID: anon1}
+		// {type:alias, previousId: user2, userID: user1}
+		// userHLL: {user1, user2}, anonHLL: {anon1}, identifiedAnonHLL: {user1-anon1, user2}
+		// cardinality: len(userHLL)+len(anonHLL)-len(identifiedAnonHLL): 2+1-2 = 1
+		if eventType == eventTypeAlias {
+			previousID := gjson.GetBytes(job.EventPayload, "batch.0.previousId").String()
+			if previousID != "" {
```
ktgowtham, line 161: Should we add a safe check to not add if prevId is same as userId(may be due to some bug in instrumentation)? 
Just to avoid an edge case, which could decrease the actual user count.

ktgowtham (approved):

ktgowtham (approved):

enterprise/trackedusers/users_reporter.go
```diff
@@ -143,6 +145,23 @@ func (u *UniqueUsersReporter) GenerateReportsFromJobs(jobs []*jobsdb.JobT, sourc
...
+		// e.g. we receive events
+		// {type:track, anonymousID: anon1}
+		// {type:track, userID: user1}
+		// {type:track, userID: user2}
+		// {type:identify, userID: user1, anonymousID: anon1}
+		// {type:alias, previousId: user2, userID: user1}
+		// userHLL: {user1, user2}, anonHLL: {anon1}, identifiedAnonHLL: {user1-anon1, user2}
+		// cardinality: len(userHLL)+len(anonHLL)-len(identifiedAnonHLL): 2+1-2 = 1
+		if eventType == eventTypeAlias {
+			previousID := gjson.GetBytes(job.EventPayload, "batch.0.previousId").String()
```
satishrudderstack, line 160: Why are we only looking at the first entry in `batch` and not all entries ?

satishrudderstack (approved):

enterprise/trackedusers/users_reporter.go
mihir20: Each batch will have only have 1 event. Even if on gateway we receive a batch request we break that down and we create a new entry for each event in a batch. and each entry will be like `{"batch":[{event...}]}`

PR #4903, opened 2024-07-17
chore: send connection config to router transform

Sidddddarth:
# Description

Rudder server now sends `connection config` to router transform with each job.
Due to this being pertinent only in case of sources jobs, we have a check to drop job only if it's a sources job.

## Linear Ticket

[Resolves PIPE-1222](https://linear.app/rudderstack/issue/PIPE-1222/read-connection-config-and-use-it-in-relevant-places)

## Security


ktgowtham:
Adding @koladilip . He wanted to review the PR.

koladilip:
@Sidddddarth you will be raising for process or proxy later?

router/types/types.go
```diff
@@ -18,6 +18,16 @@ type RouterJobT struct {
...
 	JobMetadata JobMetadataT               `json:"metadata"`
 	Destination backendconfig.DestinationT `json:"destination"`
+	Connection  backendconfig.Connection   `json:"connection"`
+}
+
+type SourceDest struct {
+	SourceID, DestinationID string
+}
+
+type Connection struct {
```
koladilip, line 28: where is this used?

Sidddddarth: It contains the connection config - so used to populate wherever needed
I've created this new struct just to have the `connectionID` here as well - could help with stats or maybe other uses in the future.

koladilip (approved):

ktgowtham (approved):
LGTM

koladilip:
we tested this @Sidddddarth

router/types/types.go
cisse21, line 28:
```suggestion
type ConnectionWithID struct {
```

router/types/types.go
```diff
@@ -18,6 +18,16 @@ type RouterJobT struct {
...
+	ConnectionID string
+	ConnectionT  backendconfig.Connection
```
cisse21, line 30:
```suggestion
	Connection  backendconfig.Connection
```

cisse21 (approved):
Approving with a nitpick

PR #4904, opened 2024-07-17
feat: drop trigger for populating warehouse schema versions

lvrach:
# Description

Stop populating `wh_schema_version` table.

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1318/stop-populating-schema-versions

Resolves PIPE-1318

## Security

### DIFF
--- a/sql/migrations/warehouse/000025_drop_trigger_schema_versions.sql.up.sql
+++ b/sql/migrations/warehouse/000025_drop_trigger_schema_versions.sql.up.sql
@@ -1,0 +1,2 @@
+
+DROP TRIGGER IF EXISTS warehouse_schema_history_trigger ON wh_schemas;


achettyiitr (approved):

cisse21 (approved):

sql/migrations/warehouse/000025_drop_trigger_schema_versions.sql.up.sql
```diff
@@ -0,0 +1,2 @@
+
+DROP TRIGGER IF EXISTS warehouse_schema_history_trigger ON wh_schemas;
```
achettyiitr, line 2: Should we also get rid of the `wh_schema_versions` table as well?

lvrach: I am cautious when deleting data using migrations. `wh_schema_versions` should be a safe table to delete, my only concern are hidden dependencies from open source customers.

PR #4906, opened 2024-07-18
feat: redshift iam support

achettyiitr:
# Description

- Add support for redshift IAM.
- Integrated [sqlconnect-go](https://github.com/rudderlabs/sqlconnect-go) for redshift driver integration.

## Linear Ticket

- Resolves PIPE-1321

## Security


warehouse/integrations/testhelper/setup.go
```diff
@@ -308,6 +308,11 @@ func RetrieveRecordsFromWarehouse(
 			switch item := item.(type) {
 			case time.Time:
 				return item.Format(time.RFC3339)
+			case string:
+				if t, err := time.Parse(time.RFC3339Nano, item); err == nil {
+					return t.Format(time.RFC3339)
+				}
+				return item
```
achettyiitr, line 315: This is required because the driver is trying to Parst and because the [redshift driver is formatting it to strin](https://github.com/rudderlabs/sqlconnect-go/blob/main/sqlconnect/internal/redshift/driver/rows.go#L86-L99)g and we are loosing the data type.

warehouse/integrations/redshift/redshift.go
```diff
@@ -969,6 +963,87 @@ func (rs *Redshift) connect(ctx context.Context) (*sqlmiddleware.DB, error) {
...
+		User:   url.UserPassword(user, password),
+		Host:   fmt.Sprintf("%s:%s", host, port),
+		Path:   database,
+	}
+
+	params := url.Values{}
+	params.Add("sslmode", "require")
+
+	if timeout > 0 {
+		params.Add("connect_timeout", fmt.Sprintf("%d", timeout/time.Second))
```
lvrach, line 1030: if timeout < time.Second, we will set `connect_timeout=0` .

Does connect_timeout supports decimals ?
What the behaviour if it is zero? Does it mean no timeout ?

It might be unexpected if someone configures 500ms, and there is actually no connection timeout.

We can consider ceiling the division or throw an error if less than a second is provided

lvrach (approved):

.github/workflows/tests.yaml
```diff
@@ -84,6 +84,12 @@ jobs:
         with:
           username: rudderlabs
           password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Set Redshift Ginkgo AWS Credentials
+        if: matrix.destination == 'redshift'
+        run: |
+          echo "Setting Redshift Ginkgo AWS Credentials..."
+          echo "AWS_ACCESS_KEY_ID=${{ secrets.AWS_REDSHIFT_GINKGO_ACCESS_KEY_ID }}" >> $GITHUB_ENV
+          echo "AWS_SECRET_ACCESS_KEY=${{ secrets.AWS_REDSHIFT_GINKGO_SECRET_ACCESS_KEY }}" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
achettyiitr: Wanted to add these env's just for redshift integration test only, since it's only specific to redshift test.

warehouse/integrations/redshift/redshift.go
achettyiitr: It's better to throw error is `timeout < time.Second`. Making the changes.

fracasula (approved):

PR #4913, opened 2024-07-22
chore: refactor dedup package

cisse21:
# Description

## Linear Ticket

## Security


mihir20 (approved):
LGTM

services/dedup/dedup.go
```diff
@@ -30,33 +23,8 @@ func DefaultPath() string {
 
 // New creates a new deduplication service. The service needs to be closed after use.
 func New(path string) Dedup {
```
mihir20, line 25: [optional] I guess we can remove path from the input as it might be more relevant to badger and might not be relevant to some other DB.

cisse21: I wanted to remove this but I forgot. Will remove this

services/dedup/badger/badger.go
```diff
@@ -66,13 +109,13 @@ func (d *badgerDB) Set(kvs []KeyValue) error {
...
 
-func (d *badgerDB) Close() {
+func (d *BadgerDB) Close() {
 	close(d.close)
 	<-d.gcDone
 	_ = d.badgerDB.Close()
 }
 
-func (d *badgerDB) start() {
+func (d *BadgerDB) Start() {
```
lvrach, line 118: I would try to "hide" Start method, we can implicitly call start once the first time a call to Get or Set is called. 

We should also return an error instead of a panic.  

```go
func (d *BadgerDB) Set(kvs []types.KeyValue) error {
    err := d.init() 
    if err != nil {
        return err
    }
}
```
```suggestion
func (d *BadgerDB) init() error {
    d.once.Do(func () {
      d.badgerDB, d.initErr = badger.Open(d.opts)
      ....
    })
    return d.initErr
```

lvrach: We would also need to change the signature for Get

```
func (d *BadgerDB) Get(key string) (int64, error)
```

This will be useful for scyllaDB client, as any method call could result into an error

services/dedup/badger/badger.go
```diff
@@ -1,20 +1,23 @@
...
 	"github.com/dgraph-io/badger/v4"
+	"github.com/dgraph-io/badger/v4/options"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
-
+	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
 
 	"github.com/rudderlabs/rudder-server/rruntime"
+	"github.com/rudderlabs/rudder-server/services/dedup/types"
```
lvrach, line 16: Would it make more sense if `types` are on `dedup` package instead, do we need a separate package for them?

cisse21: It was creating a cyclic dependency

Sidddddarth (approved):

PR #4922, opened 2024-07-23
feat: introduce scylla dedup

cisse21:
# Description

- Introduces integration with scyllaDB for dedup
- Has multiple modes but defaulted to badger for backwards compatibility
- Has mirror modes which do dual writes but read from either badger or scylla

## Linear Ticket

Fixes PIPE-1354

## Security


cisse21:
@coderabbitai summary

services/dedup/scylla/scylla.go
```diff
@@ -0,0 +1,93 @@
...
+	keysList := lo.PartitionBy(lo.Values(keys), func(kv types.KeyValue) string {
+		return kv.WorkspaceId
+	})
+	for _, keys := range keysList {
+		batches := lo.Chunk(keys, d.batchSize)
+		for _, batch := range batches {
+			scyllaBatch := d.scylla.NewBatch(gocql.LoggedBatch)
+			for _, key := range batch {
+				scyllaBatch.Entries = append(scyllaBatch.Entries, gocql.BatchEntry{
+					Stmt: fmt.Sprintf("INSERT INTO %s (id) VALUES (?,?) USING TTL %d", key.WorkspaceId, d.ttl),
```
lvrach, line 60: Is there a safe way to pass table names in scylla? Something like https://pkg.go.dev/github.com/lib/pq#QuoteIdentifier

services/dedup/mirrorScylla/mirrorScylla.go
```diff
@@ -0,0 +1,40 @@
...
+	}, nil
+}
+
+func (ms *MirrorScylla) Close() {
+	ms.scylla.Close()
+	ms.badger.Close()
+}
+
+func (ms *MirrorScylla) Set(kv types.KeyValue) (bool, int64, error) {
+	return ms.scylla.Set(kv)
```
lvrach, line 34: Shouldn't you run Set for both?

services/dedup/scylla/scylla.go
```diff
@@ -0,0 +1,93 @@
...
+	createTableMap map[string]struct{}
+}
+
+func (d *ScyllaDB) Close() {
+	d.scylla.Close()
+}
+
+func (d *ScyllaDB) Set(kv types.KeyValue) (bool, int64, error) {
+	// Create the table if it doesn't exist
+	if _, ok := d.createTableMap[kv.WorkspaceId]; !ok {
```
lvrach, line 30: We can consider using [sync once values](https://pkg.go.dev/sync#OnceValues) here and possibly some locking for the map. The code may run multiple times if multiple workspaceIds access the same time.

lvrach (commented):
👍 overall, with some comments

services/dedup/scylla/scylla.go
```diff
@@ -0,0 +1,93 @@
...
+	ttl            int
+	batchSize      int
+	createTableMap map[string]struct{}
+}
+
+func (d *ScyllaDB) Close() {
+	d.scylla.Close()
+}
+
+func (d *ScyllaDB) Set(kv types.KeyValue) (bool, int64, error) {
```
mihir20, line 28: This function name doesn't justify what it is doing, its not setting anything, its just checking from the DB. Also Don't we need a in memory cache like we have in badger?

services/dedup/scylla/scylla.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/rudderlabs/rudder-go-kit/stats"
+	"github.com/rudderlabs/rudder-server/services/dedup/types"
+)
+
+type ScyllaDB struct {
+	scylla         *gocql.Session
+	conf           *config.Config
+	stat           stats.Stats
+	ttl            int
+	batchSize      int
```
mihir20, line 20: can you add some comments on what `ttl` and `batchSize` is for?

services/dedup/mirrorScylla/mirrorScylla.go
cisse21: `Set` is used to read where as `Commit` is used to write. Will change the function signature to `Get` instead?

services/dedup/badger/badger.go
```diff
@@ -163,14 +166,62 @@ func (d *BadgerDB) gcLoop() {
 			d.logger.Errorf("Error while getting badgerDB usage: %v", err)
 			continue
 		}
-		statName := "dedup"
+		statName := "Dedup"
```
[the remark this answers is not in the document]
cisse21: find and replace error ...changing it back

services/dedup/dedup.go
```diff
@@ -3,19 +3,46 @@
...
+		}
+		return scylla, nil
+	case MirrorScylla:
+		// Writes happen to both
+		// Read only from Badger
+		return mirrorScylla.NewMirrorScylla(conf, stats)
+	case MirrorBadger:
+		// Writes happen to both
+		// Read only from Badger
+		return mirrorBadger.NewMirrorBadger(conf, stats)
```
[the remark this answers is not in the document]
cisse21: cop paste error .. changing it

services/dedup/scylla/scylla.go
cisse21: https://github.com/rudderlabs/rudder-server/pull/4922#discussion_r1694819117

Didn't want to change the signature from the original contract but can change it to Get since it makes more sense

cisse21: We will need cache... Will make the changes

services/dedup/scylla/scylla.go
cisse21: Not sure how we could use sync once without using a map. Do you have a sample code in mind?

services/dedup/badger/badger.go
```diff
@@ -171,6 +173,53 @@ func (d *BadgerDB) gcLoop() {
...
+	cache    map[string]int64
+}
+
+func (d *Dedup) Get(kv types.KeyValue) (bool, int64, error) {
+	d.cacheMu.Lock()
+	defer d.cacheMu.Unlock()
+	if previous, found := d.cache[kv.Key]; found {
+		return false, previous, nil
+	}
+	previous, found, err := d.badgerDB.Get(kv.Key)
```
fracasula, line 188: @cisse21 I might be missing something but wouldn't holding a lock during DB operations like this affect other workers in the processor pool?
Why not hold the lock just to check the cache, then release it and do a get to allow other workers to use the database?

services/dedup/scylla/scylla.go
fracasula: Something like this should work, right? https://go.dev/play/p/kgVwQclcxIy

services/dedup/badger/badger.go
cisse21:
```
	err := d.badgerDB.init()
	if err != nil {
		return false, 0, err
	}
	d.cacheMu.Lock()

	if previous, found := d.cache[kv.Key]; found {
		return false, previous, nil
	}
	d.cacheMu.Unlock()

	previous, found, err := d.badgerDB.Get(kv.Key)
	if err != nil {
		return false, 0, err
	}
	d.cacheMu.Lock()
	if !found {
		d.cache[kv.Key] = kv.Value
	}
	d.cacheMu.Unlock()
	return !found, previous, nil
```

Do you mean something like this?

services/dedup/scylla/scylla.go
cisse21: https://stackoverflow.com/questions/28447149/key-space-name-issue-with-creating-table-in-cassandra

I don't think there are any ... If the keyspace is capitalised we would have to enclose them in quotes else it would be defaulted to small letters

services/dedup/badger/badger.go
fracasula: @cisse21 yes although in your example you're not unlocking when found. also, maybe we could check the cache again after the `Get` like:

```go
func (d *Dedup) Get(kv types.KeyValue) (bool, int64, error) {
	err := d.badgerDB.init()
	if err != nil {
		return false, 0, err
	}

	d.cacheMu.Lock()
	previous, found := d.cache[kv.Key]
	d.cacheMu.Unlock()
	if found {
		return false, previous, nil
	}

	previous, found, err = d.badgerDB.Get(kv.Key)
	if err != nil {
		return false, 0, err
	}

	d.cacheMu.Lock()
	defer d.cacheMu.Unlock()

	// check the cache again in case the key was set while we were waiting for the lock
	cachePrev, cacheFound := d.cache[kv.Key]
	if cacheFound { // it's in the cache after all, discarding the value we just read
		return false, cachePrev, nil
	}
	if found { // still not in the cache, but it's in the DB so let's refresh the cache
		d.cache[kv.Key] = kv.Value
	}

	return !found, previous, nil
}
```

services/dedup/scylla/scylla.go
```diff
@@ -0,0 +1,135 @@
...
+	// This is essential if we get the same key multiple times in the same batch
+	// Since we are not committing the keys immediately, we need to keep track of the keys in the cache
+	if previous, found := d.cache[kv.Key]; found {
+		return false, previous.Value, nil
+	}
+
+	// Check if the key exists in the DB
+	var value int64
+	err = d.scylla.Query(fmt.Sprintf("SELECT size FROM %s.%s WHERE id = ?", d.keyspace, kv.WorkspaceId), kv.Key).Scan(&value)
+	if err != nil && err != gocql.ErrNotFound {
```
fracasula, line 66: Are you positive this never gets wrapped? Would it make sense to go with `errors.Is` just in case?

cisse21: Updated

services/dedup/badger/badger_test.go
```diff
@@ -0,0 +1,51 @@
...
+	_ = os.RemoveAll(dbPath)
+	conf := config.New()
+	t.Setenv("RUDDER_TMPDIR", dbPath)
+	badger := NewBadgerDB(conf, stats.NOP, DefaultPath())
+	require.NotNil(t, badger)
+	defer badger.Close()
+	t.Run("Same messageID should be deduped from badger", func(t *testing.T) {
+		key1 := types.KeyValue{Key: "a", Value: 1, WorkspaceId: "test"}
+		key2 := types.KeyValue{Key: "a", Value: 1, WorkspaceId: "test"}
+		found, _, err := badger.Get(key1)
```
mihir20, line 32:
```suggestion
		notAvailable, _, err := badger.Get(key1)
```
its not found, we are returning false if key is seen earlier

services/dedup/dedup_test.go
```diff
@@ -4,66 +4,106 @@ import (
...
+			conf.Set("Scylla.Hosts", resource.URL)
+			conf.Set("Scylla.Keyspace", keySpace)
+			conf.Set("Dedup.Mode", tc.name)
+			d, err := dedup.New(conf, stats.Default)
+			require.Nil(t, err)
+			defer d.Close()
+
+			t.Run("if message id is not present in cache and badger db", func(t *testing.T) {
+				found, _, err := d.Get(types.KeyValue{Key: "a", Value: 1, WorkspaceId: "test"})
+				require.Nil(t, err)
```
mihir20, line 73:
```suggestion
				require.NoError(t, err)
```

services/dedup/dedup_test.go
```diff
@@ -4,66 +4,106 @@ import (
...
-	})
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			config.Reset()
+			logger.Reset()
+			misc.Init()
+
+			dbPath := os.TempDir() + "/dedup_test"
+			defer func() { _ = os.RemoveAll(dbPath) }()
+			_ = os.RemoveAll(dbPath)
```
mihir20, line 56:
```suggestion
			dbPath := t.TempDir()
```

services/dedup/mirrorBadger/mirrorBadger.go
```diff
@@ -0,0 +1,40 @@
...
+	}, nil
+}
+
+func (mb *MirrorBadger) Close() {
+	mb.scylla.Close()
+	mb.badger.Close()
+}
+
+func (mb *MirrorBadger) Get(kv types.KeyValue) (bool, int64, error) {
+	return mb.badger.Get(kv)
```
mihir20, line 34: Correct me if I am wrong. our intention here is to populate data in both places(scylla and badger). if we are not doing get on scylla as well, then the in-memory cache in scylla will not be populated so commit will not insert any data in scylla.

we should also add a test to check if data is populated in both places

services/dedup/mirrorScylla/mirrorScylla.go
```diff
@@ -0,0 +1,40 @@
...
+	}, nil
+}
+
+func (ms *MirrorScylla) Close() {
+	ms.scylla.Close()
+	ms.badger.Close()
+}
+
+func (ms *MirrorScylla) Get(kv types.KeyValue) (bool, int64, error) {
+	return ms.scylla.Get(kv)
```
mihir20, line 34: same point here

mihir20 (commented):
not very sure on mirror scylla and mirror badger implementations.

services/dedup/mirrorBadger/mirrorBadger.go
```diff
@@ -0,0 +1,41 @@
...
+	}, nil
+}
+
+func (mb *MirrorBadger) Close() {
+	mb.scylla.Close()
+	mb.badger.Close()
+}
+
+func (mb *MirrorBadger) Get(kv types.KeyValue) (bool, int64, error) {
+	_, _, _ = mb.scylla.Get(kv)
```
mihir20, line 34: What if we handle the error here? In worst case scylla will get populated and badger will not, anyways we are throwing a panic in processor in such cases so everything will be retried.

services/dedup/scylla/scylla_test.go
```diff
@@ -0,0 +1,66 @@
...
+		require.False(t, found)
+	})
+	t.Run("Same messageID should be deduped for same workspace from cache", func(t *testing.T) {
+		key1 := types.KeyValue{Key: "b", Value: 1, WorkspaceId: "test"}
+		key2 := types.KeyValue{Key: "b", Value: 1, WorkspaceId: "test"}
+		found, _, err := scylla.Get(key1)
+		require.Nil(t, err)
+		require.True(t, found)
+		found, _, err = scylla.Get(key2)
+		require.Nil(t, err)
```
lvrach, line 63: [nit-pick] for consistency and improved semantics
```suggestion
		require.NoError(t, err)
		require.False(t, found)
	})
	t.Run("Same messageID should be deduped for same workspace from cache", func(t *testing.T) {
		key1 := types.KeyValue{Key: "b", Value: 1, WorkspaceId: "test"}
		key2 := types.KeyValue{Key: "b", Value: 1, WorkspaceId: "test"}
		found, _, err := scylla.Get(key1)
		require.NoError(t, err)
		require.True(t, found)
		found, _, err = scylla.Get(key2)
		require.NoError(t, err)
```

services/dedup/scylla/scylla.go
```diff
@@ -0,0 +1,139 @@
...
+		for _, batch := range batches {
+			scyllaBatch := d.scylla.NewBatch(gocql.LoggedBatch)
+			for _, key := range batch {
+				scyllaBatch.Entries = append(scyllaBatch.Entries, gocql.BatchEntry{
+					Stmt: fmt.Sprintf("INSERT INTO %s.%s (id,size) VALUES (?,?) USING TTL %d", d.keyspace, key.WorkspaceId, d.ttl),
+					Args: []interface{}{key.Key, key.Value},
+				})
+			}
+			if err := d.scylla.ExecuteBatch(scyllaBatch); err != nil {
+				return fmt.Errorf("error committing keys: %v", err)
```
lvrach, line 103: [minor]
```suggestion
				return fmt.Errorf("committing keys: %v", err)
```

services/dedup/scylla/scylla.go
```diff
@@ -0,0 +1,139 @@
...
+	if !found {
+		d.createTableMap[kv.WorkspaceId] = &sync.Once{}
+		once = d.createTableMap[kv.WorkspaceId]
+	}
+	d.createTableMu.Unlock()
+	once.Do(func() {
+		query := fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s.%s (id text PRIMARY KEY,size bigint) WITH bloom_filter_fp_chance = 0.005;", d.keyspace, kv.WorkspaceId)
+		err = d.scylla.Query(query).Exec()
+	})
+	if err != nil {
```
lvrach, line 51: You can use [sync.OnceValue](https://pkg.go.dev/sync#OnceValue) to store and reuse the error.

In the existing code, in case of an error, only the first goroutine will return the error. 
```suggestion
		d.createTableMap[kv.WorkspaceId] = &sync.OnceValue[error]{}
		once = d.createTableMap[kv.WorkspaceId]
	}
	d.createTableMu.Unlock()

	err := once.Do(func() error {
		query := fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s.%s (id text PRIMARY KEY,size bigint) WITH bloom_filter_fp_chance = 0.005;", d.keyspace, kv.WorkspaceId)
		return d.scylla.Query(query).Exec()
	})
	if err != nil {
```

lvrach: Also, you might want to move this logic to a separate unexported method, and call it even from `Commit`. It depends on how robust you want to make your API, normally Get for a workspace should always happen before a commit.

lvrach (approved):

services/dedup/badger/badger.go
```diff
@@ -40,45 +40,42 @@ func DefaultPath() string {
...
-func NewBadgerDB(path string) *BadgerDB {
-	dedupWindow := config.GetReloadableDurationVar(3600, time.Second, "Dedup.dedupWindow", "Dedup.dedupWindowInS")
-
-	log := logger.NewLogger().Child("dedup")
+func NewBadgerDB(conf *config.Config, stats stats.Stats, path string) *Dedup {
+	dedupWindow := conf.GetReloadableDurationVar(3600, time.Second, "Dedup.dedupWindow", "Dedup.dedupWindowInS")
+	log := logger.NewLogger().Child("Dedup")
 	badgerOpts := badger.
 		DefaultOptions(path).
 		WithCompression(options.None).
```
fracasula, line 48: [question] Out of curiosity... did you test this with compression at all? If yes, what did you observe?

services/dedup/badger/badger.go
```diff
@@ -40,45 +40,42 @@ func DefaultPath() string {
...
 		WithIndexCacheSize(16 << 20). // 16mb
 		WithNumGoroutines(1).
```
fracasula, line 50: [question] Could increasing this number help?

services/dedup/badger/badger.go
```diff
@@ -40,45 +40,42 @@ func DefaultPath() string {
...
-	var err error
-	err = d.init()
-	if err != nil {
-		return 0, false, err
-	}
-	err = d.badgerDB.View(func(txn *badger.Txn) error {
+	err := d.badgerDB.View(func(txn *badger.Txn) error {
 		item, err := txn.Get([]byte(key))
 		if err != nil {
 			return err
```
fracasula, line 81: [question on line 83] Probably not related to this PR but it seems weird so I'll ask.

Why do we do this?

```go
			payloadSize, _ = strconv.ParseInt(string(itemValue), 10, 64)
			found = true
```

Wouldn't it make more sense to do this instead?

```go
			payloadSize, err = strconv.ParseInt(string(itemValue), 10, 64)
			if err == nil {
				found = true
			}
```

I'm saying this, because if we can get the value from the database but it cannot be parsed then we wrote something invalid. We might as well say we couldn't find it? :thinking: WDYT?

services/dedup/mirrorBadger/mirrorBadger_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/rudderlabs/rudder-server/services/dedup/types"
+	"github.com/rudderlabs/rudder-server/utils/misc"
+)
+
+func Test_MirrorBadger(t *testing.T) {
+	config.Reset()
+	logger.Reset()
+	misc.Init()
+
+	dbPath := os.TempDir() + "/dedup_test"
```
fracasula, line 25:
```suggestion
	dbPath := t.TempDir() + "/dedup_test"
```

fracasula: It checks for existence as well so you could remove the `_ = os.RemoveAll(dbPath)` that you have on line 27.
It should also do the cleanup for you if I'm not mistaken:

```go
		c.tempDir, c.tempDirErr = os.MkdirTemp("", pattern)
		if c.tempDirErr == nil {
			c.Cleanup(func() {
				if err := removeAll(c.tempDir); err != nil {
					c.Errorf("TempDir RemoveAll cleanup: %v", err)
				}
			})
		}
```

services/dedup/mirrorScylla/mirrorScylla_test.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+func Test_MirrorBadger(t *testing.T) {
+	config.Reset()
+	logger.Reset()
+	misc.Init()
+
+	dbPath := os.TempDir() + "/dedup_test"
+	defer func() { _ = os.RemoveAll(dbPath) }()
+	_ = os.RemoveAll(dbPath)
```
fracasula, line 27: Same comment as above. Could use `t.TempDir()`.

fracasula (approved):
Approved but I left a few questions and a suggestion for your tests :+1:

services/dedup/badger/badger.go
cisse21: This was an already existing code and I didn't make any changes in the existing functionality

services/dedup/badger/badger.go
cisse21: I think this was done because we care more whether the key was present or not and not the value for the key

PR #4925, opened 2024-07-24
fix: introduce config for get requests in webhook sources

sanpj2292:
# Description

Introduce configuration to allow `GET` requests in webhook sources. Currently we have added `adjust` source by default as one of the sources which sends source event via `GET` calls to rudderstack webhook

## Linear Ticket

Resolves INT-2460

## Security

### DIFF
--- a/gateway/webhook/webhook.go
+++ b/gateway/webhook/webhook.go
@@ -69,7 +69,7 @@ type HandleT struct {
 		webhookBatchTimeout        config.ValueLoader[time.Duration]
 		maxWebhookBatchSize        config.ValueLoader[int]
 		sourceListForParsingParams []string
-		forwardGetRequestForSrcs   []string
+		forwardGetRequestForSrcMap   map[string]struct{}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


shrouti1507 (approved):

koladilip (commented):
tests?

gateway/webhook/webhook.go
```diff
@@ -114,7 +115,7 @@ func (webhook *HandleT) RequestHandler(w http.ResponseWriter, r *http.Request) {
 	var postFrom url.Values
 	var multipartForm *multipart.Form
 
-	if r.Method == "GET" {
+	if r.Method == "GET" && !lo.Contains(webhook.config.forwardGetRequestForSrcs, sourceDefName) {
```
achettyiitr, line 118: We can probably use Map instead of slice here.
```
    // definition
    forwardGetRequestForSrcsMap map[string]struct{}

    // initialization
	webhook.config.forwardGetRequestForSrcsMap = lo.SliceToMap(config.GetStringSliceVar([]string{"adjust"}, "Gateway.webhook.forwardGetRequestForSrcs"), func(item string) (string, struct{}) {
		return item, struct{}{}
	})

    // usage
	if r.Method == "GET" {
		if _, ok := webhook.config.forwardGetRequestForSrcsMap[sourceDefName]; !ok {
			return
		}
	}
```

sanpj2292: Done

sanpj2292:
> tests?

Have added tests. Can you once check and let me know ?

koladilip (approved):

PR #4927, opened 2024-07-24
fix: introduce config for get requests in webhook sources

sanpj2292:
# Description

Introduce configuration to allow GET requests in webhook sources. Currently we have added adjust source by default as one of the sources which sends source event via GET calls to rudderstack webhook

## Linear Ticket

Resolves INT-2460

## Security

### DIFF
--- a/gateway/webhook/setup.go
+++ b/gateway/webhook/setup.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
+	"github.com/samber/lo"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
 
@@ -55,6 +56,14 @@ func Setup(gwHandle Gateway, transformerFeaturesService transformer.FeaturesServ
 	maxTransformerProcess := config.GetIntVar(64, 1, "Gateway.webhook.maxTransformerProcess")
 	// Parse all query params from sources mentioned in this list
 	webhook.config.sourceListForParsingParams = config.GetStringSliceVar([]string{"Shopify", "adjust"}, "Gateway.webhook.sourceListForParsingParams")
+
+	webhook.config.forwardGetRequestForSrcMap = lo.SliceToMap(
+		config.GetStringSliceVar([]string{"adjust"}, "Gateway.webhook.forwardGetRequestForSrcs"),
+		func(item string) (string, struct{}) {
+			return item, struct{}{}
+		},
+	)
+
 	// lowercasing the strings in sourceListForParsingParams
 	for i, s := range webhook.config.sourceListForParsingParams {
 		webhook.config.sourceListForParsingParams[i] = strings.ToLower(s)
--- a/gateway/webhook/webhook_test.go
+++ b/gateway/webhook/webhook_test.go
@@ -518,3 +518,96 @@ func TestPrepareRequestBody(t *testing.T) {
 		})
 	}
 }
+
+func TestAllowGetReqForWebhookSrc(t *testing.T) {
+	initWebhook()
+	ctrl := gomock.NewController(t)
+	mockGW := mockWebhook.NewMockGateway(ctrl)
+	transformerServer := httptest.NewServer(
+		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			defer func() { _ = r.Body.Close() }()
+			body, _ := io.ReadAll(r.Body)
+			var requests []interface{}
+			_ = json.Unmarshal(body, &requests)
+			var responses []transformerResponse
+			for i := 0; i < len(requests); i++ {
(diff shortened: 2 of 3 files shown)


sanpj2292:
hotfix version of https://github.com/rudderlabs/rudder-server/pull/4925

gateway/webhook/webhook_test.go
```diff
@@ -518,3 +518,96 @@ func TestPrepareRequestBody(t *testing.T) {
...
+		description string
+		input       struct {
+			srcMap map[string]struct{}
+			method string
+			srcDef string
+		}
+		output bool
+	}
+
+	cases := []testCase{
```
achettyiitr, line 555: [optional]: We can use an anonymous struct in here.
```suggestion
	[]struct {
		description string
		input       struct {
			srcMap map[string]struct{}
			method string
			srcDef string
		}
		output bool
	}{
```

gateway/webhook/webhook_test.go
```diff
@@ -518,3 +518,96 @@ func TestPrepareRequestBody(t *testing.T) {
 		})
 	}
 }
+
+func TestAllowGetReqForWebhookSrc(t *testing.T) {
```
achettyiitr, line 522: We can probably simplify this test, since we are just testing a unit function and it only depends on config.
```GO
func TestAllowGetReqForWebhookSrc(t *testing.T) {
	cases := []struct {
		name                       string
		forwardGetRequestForSrcMap map[string]struct{}
		method                     string
		srcDef                     string
		expected                   bool
	}{
		{
			name:   "should allow get request for adjust",
			method: http.MethodGet,
			forwardGetRequestForSrcMap: map[string]struct{}{
				"adjust": {},
			},
			srcDef:   "adjust",
			expected: false,
		},
		{
			name:   "should allow post request for adjust",
			method: http.MethodPost,
			forwardGetRequestForSrcMap: map[string]struct{}{
				"adjust": {},
			},
			srcDef:   "adjust",
			expected: false,
		},
		{
			name: "should not allow get request for shopify",
			forwardGetRequestForSrcMap: map[string]struct{}{
				"adjust":     {},
				"customerio": {},
			},
			method:   http.MethodGet,
			srcDef:   "Shopify",
			expected: true,
		},
	}

	for _, tc := range cases {
			wbh := HandleT{}
			wbh.config.forwardGetRequestForSrcMap = tc.forwardGetRequestForSrcMap

			isGetAndNotAllow := wbh.IsGetAndNotAllow(tc.method, tc.srcDef)
			require.Equal(t, tc.expected, isGetAndNotAllow)
		})
	}
}
```

achettyiitr (approved):
Approving with a small comment.

gateway/webhook/setup.go
```diff
@@ -55,6 +56,14 @@ func Setup(gwHandle Gateway, transformerFeaturesService transformer.FeaturesServ
 	maxTransformerProcess := config.GetIntVar(64, 1, "Gateway.webhook.maxTransformerProcess")
 	// Parse all query params from sources mentioned in this list
 	webhook.config.sourceListForParsingParams = config.GetStringSliceVar([]string{"Shopify", "adjust"}, "Gateway.webhook.sourceListForParsingParams")
+
+	webhook.config.forwardGetRequestForSrcMap = lo.SliceToMap(
+		config.GetStringSliceVar([]string{"adjust"}, "Gateway.webhook.forwardGetRequestForSrcs"),
```
lvrach, line 61: I would avoid using a config for this. Similar for `sourceListForParsingParams`.

As configuration skew across environments can cause issues.

gateway/webhook/webhook_test.go
lvrach, line 522: Can we add it as use case on our webhook integration tests ?

gateway/webhook/setup.go
sanpj2292: Can you explain how and why you think using config would cause issues ?

lvrach: This configuration contains source-specific business logic.

* Configuration can diverge in different environments, having a source working in one, but not in the other.
* Configuration circumvents testing and usual release processes.


IMO `Gateway.webhook.forwardGetRequestForSrcs` `Gateway.webhook.sourceListForParsingParams` should be converted to constants any change on them should require a server release and source-specific tests.

We can also consider moving this logic to rudder-transformers. rudder-server or ingestion-svc will pass all the methods (GET, POST, PUT .. ) and `rudder-transfomer` will have logic to reject them or accept them.

lvrach (approved):

gateway/webhook/webhook_test.go
sanpj2292: Yes. I will add it

koladilip (approved):

shrouti1507 (approved):

PR #4930, opened 2024-07-24
fix: new FileManager for every object storage operation

Sidddddarth:
# Description

Reuse the same File manager per workspace for every operation, until there's an update in the data retention settings for said workspace.

Slight change in behaviour: when a workspace opts for self-storage and provider type is empty, it simply means no storage for workspace.

## Linear Ticket

## Security


services/fileuploader/fileuploader.go
```diff
@@ -121,13 +153,28 @@ func (p *provider) updateLoop(ctx context.Context, backendConfig backendconfig.B
...
+				updatedAt:   time.Now(),
 			}
+			filemanagerMap[workspaceId] = lo.Must(filemanager.New(&filemanager.Settings{
+				Provider: bucket.Type,
+				Config:   bucket.Config,
+			}))
 		}
 		p.mu.Lock()
-		p.storageSettings = settings
+		p.storageSettings = settingsMap
```
lvrach, line 176: Do we still need to maintain storageSettings?

services/fileuploader/fileuploader.go
```diff
@@ -121,13 +153,28 @@ func (p *provider) updateLoop(ctx context.Context, backendConfig backendconfig.B
...
+				continue
+			}
+
+			// either newly polled workspace settings or updated storage config - update object storage Filemanager
+			settingsMap[workspaceId] = StorageSettings{
 				Bucket:      bucket,
 				Preferences: preferences,
+				updatedAt:   time.Now(),
 			}
+			filemanagerMap[workspaceId] = lo.Must(filemanager.New(&filemanager.Settings{
```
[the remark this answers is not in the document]
lvrach: should we create a file manager proactively for newly polled configs? We may end up maintaining a lot of managers in memory without receiving events for them.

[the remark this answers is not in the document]
Sidddddarth: > We may end up maintaining a lot of managers in memory without receiving events for them.

True 👍.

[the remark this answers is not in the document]
Sidddddarth: Found something interesting to avoid pre-populating all the File managers: 

`sync.OnceValues[filemanager.FileManager, error]`

services/fileuploader/fileuploader.go
```diff
@@ -67,20 +82,20 @@ func (p *provider) getStorageSettings(workspaceID string) (StorageSettings, erro
 	defer p.mu.RUnlock()
 	settings, ok := p.storageSettings[workspaceID]
 	if !ok {
-		return StorageSettings{}, NoStorageForWorkspaceError
+		return StorageSettings{}, ErrNoStorageForWorkspace
 	}
 	return settings, nil
 }
 
 func (p *provider) GetFileManager(workspaceID string) (filemanager.FileManager, error) {
```
fracasula, line 90: We should consider introducing a context here and do a `select` with `<-p.init`.

services/fileuploader/fileuploader.go
```diff
@@ -97,9 +112,21 @@ func (p *provider) updateLoop(ctx context.Context, backendConfig backendconfig.B
...
+		p.mu.RUnlock()
+		settingsMap := make(map[string]StorageSettings)
+		filemanagerMap := make(map[string]func() (filemanager.FileManager, error))
 		configs := ev.Data.(map[string]backendconfig.ConfigT)
 		for workspaceId, c := range configs {
+			currentWorkspaceConfig, ok := currentSettingsMap[workspaceId]
+			// no change in workspace config, don't process the same config again
+			if ok && !c.UpdatedAt.After(currentWorkspaceConfig.updatedAt) {
+				settingsMap[workspaceId] = currentWorkspaceConfig
+				filemanagerMap[workspaceId] = currentFileManagerMap[workspaceId]
```
fracasula, line 127: Won't we have a race anyway here? You're getting a map which works via address so if the content of the map is changed in line 94, you would still have a read/write race condition here. A test with `-race` should highlight this.

lvrach:
@Sidddddarth which components depend on `services/fileuploader`? Is it only be used for archiving?

Sidddddarth:
> which components depend on services/fileuploader?

Processor/stash as well.

services/fileuploader/fileuploader.go
```diff
@@ -43,6 +48,17 @@ func NewStaticProvider(storageSettings map[string]StorageSettings) Provider {
 		init:            make(chan struct{}),
 		storageSettings: storageSettings,
 	}
+	s.fileManagerMap = make(map[string]func() (filemanager.FileManager, error))
+	for workspaceID, settings := range storageSettings {
+		if settings.Bucket.Type != "" {
```
fracasula, line 53: Would it make sense to return an error in this case?

services/fileuploader/fileuploader.go
```diff
@@ -121,13 +155,30 @@ func (p *provider) updateLoop(ctx context.Context, backendConfig backendconfig.B
...
+				return filemanager.New(&filemanager.Settings{
+					Provider: bucket.Type,
+					Config:   bucket.Config,
+				})
+			})
 		}
 		p.mu.Lock()
-		p.storageSettings = settings
+		p.storageSettings = settingsMap
+		p.fileManagerMap = filemanagerMap
```
fracasula, line 181: When overwriting here, it could be that some file managers will be free for garbage collection, right? Could it be that these file managers have dangling connections that might need cleaning up first? I see no `Close()` method in the interface so I'm talking about what's going on under the hood.

services/fileuploader/fileuploader.go
Sidddddarth: It would, but then we'd have to change function signatures in multiple places imo.
If an error weren't returned, subsequent call to get a file manager would fail with an error in any case.

services/fileuploader/fileuploader.go
Sidddddarth: not using `ctx` just to keep the signature same.
But adopting the select and returning another sentinel error

services/fileuploader/fileuploader.go
```diff
@@ -59,35 +75,38 @@ type provider struct {
...
 	var prefs backendconfig.StoragePreferences
-	settings, err := p.getStorageSettings(workspaceID)
-	if err != nil {
-		return prefs, err
+	select {
+	case <-ctx.Done():
+		return prefs, ctx.Err()
+	case <-p.init:
+	}
+	<-p.init
```
fracasula, line 104: :stuck_out_tongue_winking_eye:

services/fileuploader/fileuploader.go
```diff
@@ -59,35 +75,38 @@ type provider struct {
...
 }
 
-func (p *provider) getStorageSettings(workspaceID string) (StorageSettings, error) {
+func (p *provider) GetFileManager(ctx context.Context, workspaceID string) (filemanager.FileManager, error) {
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-p.init:
+	}
 	<-p.init
```
fracasula, line 87:
```suggestion
```

services/fileuploader/fileuploader.go
Sidddddarth: 🤡

services/fileuploader/fileuploader.go
```diff
@@ -121,13 +157,30 @@ func (p *provider) updateLoop(ctx context.Context, backendConfig backendconfig.B
...
+				})
+			})
 		}
 		p.mu.Lock()
-		p.storageSettings = settings
+		p.storageSettings = settingsMap
+		p.fileManagerMap = filemanagerMap
 		p.mu.Unlock()
 		p.onceInit.Do(func() {
 			close(p.init)
```
fracasula, line 186: Is the second `onceInit` after the for loop needed?

Sidddddarth: I think it was added in the past to cater to some corner case, can't recall the reason though.
No harm in having it though right..?

fracasula: I don't think we should have "just in case" code. I think we should try our best to fully understand the code we are working on.

That said, I think this was done to cover the case where `ch` would get closed right away by `backendConfig`. This means that the `updateLoop` would never get to do an itearation of the `for` loop, thus it would never initialize things.

And if you don't "initialize" things then calls to, say, `GetStoragePreferences` would block forever. Now that we have a context, they would block until the context gets cancelled but in some places we're still passing a background one.

All considered, I think the correct way of doing it though is not to hackily say that we initialized things even if we didn't. I think we that if we are not subscribed to backend config changes, then we should say so with an error.

Perhaps along with the `init` channel, we can have a new one called `notSubscribed` and then do this towards the end of the `updateLoop`:

```go
		p.onceInit.Do(func() {
			close(p.init)
		})
	}

	p.onceInit.Do(func() {
		close(p.notSubscribed)
	})
```

Then we can change the other methods accordingly:

```go
func (p *provider) GetStoragePreferences(ctx context.Context, workspaceID string) (backendconfig.StoragePreferences, error) {
	var prefs backendconfig.StoragePreferences
	select {
	case <-ctx.Done():
		return prefs, ctx.Err()
	case <-p.init:
	case <-p.notSubscribed:
		return ErrNotSubscribed
	}
```

As things are at the moment, we're simply firing the `updateLoop` like this:

```go
go s.updateLoop(ctx, config)
```

If something goes wrong in the backend config then would we even notice?
* if we managed to init at least once we could serve stale settings until the service runs and for some workspaces we could return that there are not settings (if they came up afterwards)
* if we didn't manage to init ever, we wouldn't know and we would keep returning `ErrNoStorageForWorkspace` for example

WDYT?

Sidddddarth: I understood the case in point, and the idea makes sense.

With the mentioned select scenario above though, it's possible that we still return stale configuration when config subscription stops(it could randomly chose between `<-p.init` and `<-p.notSubscribed`). Instead we could use a bool(atomic?) to return early `if !subscribed`.

```
func (p *provider) GetStoragePreferences(ctx context.Context, workspaceID string) (backendconfig.StoragePreferences, error) {
	var prefs backendconfig.StoragePreferences
	if !p.subscribed.Load() {
		return prefs, ErrNotSubscribed
	}
	select {
	case <-ctx.Done():
		return prefs, ctx.Err()
	case <-p.init:
	}
```
And
```
func (p *provider) updateLoop(ctx context.Context, backendConfig backendconfig.BackendConfig) {
	ch := backendConfig.Subscribe(ctx, backendconfig.TopicBackendConfig)
	defer p.subscribed.Load(false)
	for ev := range ch {
		// process config
		p.once.Do(func() {
			close(p.init)
			p.subscribed.Load(true)
		})
	}
}

fracasula: That could work. The alternative would be to have two `select`s as to specify the priority. Your approach seems leaner though considering that there is no way to be always 100% precise and consistent here. So +1 for the atomic value :slightly_smiling_face:

Sidddddarth: Removed the `Once` and the `init` channel too. The bool sufficed.

fracasula: I've added a draft [here](https://github.com/rudderlabs/rudder-server/pull/4989).

cisse21 (approved):

PR #4939, opened 2024-07-26
feat(marketo_bulk_upload): add support to perform marketo bulk upload polling and fetch job status

sandeepdsvs:
# Description

## Linear Ticket

## Security


router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/fetchJobStatus.go
```diff
@@ -0,0 +1,205 @@
...
+	Message string
+}
+
+func (e *PlatformError) Error() string {
+	return e.Message
+}
+
+func getJobsStatus(event MarketoAsyncFailedPayload, jobType string, accessToken string) (string, error) {
+	config := event.Config
+	munchkinId := config["MunchkinId"]
```
koladilip, line 47: Do we need to do a nil check for config?

router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/fetchJobStatus.go
```diff
@@ -0,0 +1,205 @@
...
+	resp, err := handleHttpRequest("GET", url, headers)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	endTime := time.Now()
+	requestTime := endTime.Sub(startTime).Milliseconds()
+
+	fmt.Printf("Request time: %d ms\n", requestTime)
```
koladilip, line 72: use stats or logs

router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/fetchJobStatus.go
```diff
@@ -0,0 +1,205 @@
...
+	jobStatusArr := strings.Split(jobStatus, "\n")
+	headerArr := strings.Split(event.MetaData.CSVHeaders, ",")
+
+	if len(headerArr) == 0 {
+		return nil, &PlatformError{"No csvHeader in metadata"}
+	}
+
+	data := make(map[int64]string)
+
+	fieldSchemaMapping, err := getFieldSchemaMap(accessToken, config["MunchkinId"])
```
koladilip, line 112: can  we use some CSV reader rather than doing manually

router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/fetchJobStatus.go
```diff
@@ -0,0 +1,205 @@
...
+	unsuccessfulJobIdsArr := make([]int64, 0)
+	successfulJobIdsArr := make([]int64, 0)
+
+	for _, element := range jobStatusArr {
+		elemArr := parseCSVLine(element)
+		reasonMessage := elemArr[len(elemArr)-1]
+		elemArr = elemArr[:len(elemArr)-1]
+
+		for key, val := range data {
+			if val == strings.Join(elemArr, ",") {
```
koladilip, line 153: do we need to join for every comparison? can we do this outside the loop?

koladilip, line 153: we should use the lookup field for comparison not compare the whole record
https://developer.adobe.com/marketo-apis/api/mapi/#tag/Bulk-Import-Leads

shrouti1507: There can be one possible change in the functionality. The user needs to ensure that lookup field should be chosen as primary key while configuring the source.
Here is the reason:
Let's say lookup field is Email, and while configuring the source if we choose ID as a primary key, then source can query the rows where id must be present but email can be null / empty string etc. 
In that case,  Marketo throws some warning. 
I was sending the below data in csv:
```
FirstName,LastName,Email,Company
'Able','Baker','', 'marketo'
Able,"Baker","shrouti@gmail.com", "marketo"
```
And while polling I got response :
```
{
    "requestId": "14cf8#1912627bf66",
    "result": [
        {
            "batchId": 7271,
            "importId": "7271",
            "status": "Complete",
            "numOfLeadsProcessed": 2,
            "numOfRowsFailed": 0,
            "numOfRowsWithWarning": 1,
            "message": "Import succeeded, 2 records imported (2 members), 1 warning."
        }
    ],
    "success": true
}
```
For this the code flow will appear in fetch job status part, and if we search with only lookup field, we will not get anything to refer from. Previously we were marking these kind of warnings as failure, but here, it will be automatically marked as success.

The response : 

```
FirstName,LastName,Email,Company,Import Warning Reason
'Able','Baker','', 'marketo',Invalid email address
```
@koladilip let me know if you need some more clarification.

koladilip: this is a rare case and it is fine, we can update a footer note in the form, I don't think this breaks our logic so we are good

PR #4948, opened 2024-07-29
chore: vacuum tracked_users_reports table

mihir20:
# Description
vacuum tracked_users_reports table

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1360/vacuum-handling-for-mtu-tables

## Security

### DIFF
--- a/enterprise/reporting/flusher/flusher.go
+++ b/enterprise/reporting/flusher/flusher.go
@@ -10,6 +10,8 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/rudderlabs/rudder-go-kit/bytesize"
+
 	"github.com/cenkalti/backoff"
 	"golang.org/x/sync/errgroup"
 
@@ -40,6 +42,7 @@ type Flusher struct {
 	batchSizeFromDB                     config.ValueLoader[int]
 	aggressiveFlushEnabled              config.ValueLoader[bool]
 	lagThresholdForAggresiveFlushInMins config.ValueLoader[time.Duration]
+	vacuumThresholdBytes                config.ValueLoader[int64]
 
 	reportingURL          string
 	minConcurrentRequests config.ValueLoader[int]
@@ -51,6 +54,7 @@ type Flusher struct {
 	aggReportsTimer         stats.Measurement
 	sendReportsTimer        stats.Measurement
 	deleteReportsTimer      stats.Measurement
+	vacuumReportsTimer      stats.Measurement
 	concurrentRequests      stats.Measurement
 	reqLatency              stats.Measurement
 	reqCount                stats.Measurement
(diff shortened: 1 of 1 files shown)


enterprise/reporting/flusher/flusher.go
```diff
@@ -303,6 +315,23 @@ func (f *Flusher) delete(ctx context.Context, minReportedAt, maxReportedAt time.
...
+	if err := f.db.QueryRowContext(
+		ctx, fmt.Sprintf(`SELECT pg_table_size(oid) from pg_class where relname='%s'`, f.table),
+	).Scan(&sizeEstimate); err != nil {
+		return fmt.Errorf("error getting table size %w", err)
+	}
+	if sizeEstimate > f.vacuumThresholdBytes.Load() {
+		if _, err := f.db.ExecContext(ctx, fmt.Sprintf("vacuum full analyze %s", f.table)); err != nil {
+			return fmt.Errorf("error vacuuming table %w", err)
+		}
+		f.vacuumReportsTimer.Since(vacuumStart)
```
achettyiitr, line 330: Technically, vacuum time is before running the query.
```suggestion
        vacuumStart := time.Now()
		if _, err := f.db.ExecContext(ctx, fmt.Sprintf("vacuum full analyze %s", f.table)); err != nil {
			return fmt.Errorf("error vacuuming table %w", err)
		}
		f.vacuumReportsTimer.Since(vacuumStart)
```

enterprise/reporting/flusher/flusher.go
```diff
@@ -303,6 +315,23 @@ func (f *Flusher) delete(ctx context.Context, minReportedAt, maxReportedAt time.
...
+func (f *Flusher) vacuum(ctx context.Context) error {
+	vacuumStart := time.Now()
+	var sizeEstimate int64
+	if err := f.db.QueryRowContext(
+		ctx, fmt.Sprintf(`SELECT pg_table_size(oid) from pg_class where relname='%s'`, f.table),
+	).Scan(&sizeEstimate); err != nil {
+		return fmt.Errorf("error getting table size %w", err)
+	}
+	if sizeEstimate > f.vacuumThresholdBytes.Load() {
+		if _, err := f.db.ExecContext(ctx, fmt.Sprintf("vacuum full analyze %s", f.table)); err != nil {
```
achettyiitr, line 327: We can use QuoteIdentifier, to be safely quote the table name.
```suggestion
		if _, err := f.db.ExecContext(ctx, fmt.Sprintf("vacuum full analyze %s", pq.QuoteIdentifier(f.table))); err != nil {
```

enterprise/reporting/flusher/flusher.go
```diff
@@ -303,6 +315,23 @@ func (f *Flusher) delete(ctx context.Context, minReportedAt, maxReportedAt time.
 	return err
 }
 
+func (f *Flusher) vacuum(ctx context.Context) error {
+	vacuumStart := time.Now()
+	var sizeEstimate int64
+	if err := f.db.QueryRowContext(
+		ctx, fmt.Sprintf(`SELECT pg_table_size(oid) from pg_class where relname='%s'`, f.table),
```
achettyiitr, line 322: We can use the parameterized query here.
```suggestion
		ctx, `SELECT pg_table_size(oid) from pg_class where relname = $1`, f.table,
```

achettyiitr (approved):

enterprise/reporting/flusher/flusher.go
mihir20: we don't need to quote table name here

achettyiitr: Yeah we don't essentially need it but it safer approach to passing the table name via QuoteIdentifier.

cisse21 (approved):

PR #4975, opened 2024-08-04
chore: enforce max limit for webhook

achettyiitr:
# Description

- Max limit enforcement on the incoming webhook request. 
- Currently, we are enforcing the source transformation responses we get from the transformer if there is some Output present. 
- We might overwhelm the transformers if we don't enforce the incoming webhook request.

## Linear Ticket

- Resolves PIPE-1382

## Security

### DIFF
--- a/gateway/webhook/webhook.go
+++ b/gateway/webhook/webhook.go
@@ -66,6 +66,7 @@ type HandleT struct {
 	backgroundCancel context.CancelFunc
 
 	config struct {
+		maxReqSize                 config.ValueLoader[int]
 		webhookBatchTimeout        config.ValueLoader[time.Duration]
 		maxWebhookBatchSize        config.ValueLoader[int]
 		sourceListForParsingParams []string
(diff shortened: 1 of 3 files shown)


koladilip (approved):

gateway/webhook/webhook.go
```diff
@@ -334,6 +335,13 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
 				req.done <- transformerResponse{Err: response.GetStatus(response.InvalidJSON)}
 				continue
 			}
+			if len(body) > bt.webhook.config.maxReqSize.Load() {
```
ktgowtham, line 338: How costly is `Load` function? would it be okay to keep it outside the loop? If it's not costly, should be fine.

ktgowtham (approved):

gateway/webhook/webhook.go
achettyiitr: It's just map lookup. https://github.com/rudderlabs/rudder-go-kit/blob/main/config/hotreloadable.go#L784-L790

PR #4976, opened 2024-08-04
chore: adapting obskit for warehouse

achettyiitr:
# Description

- Adapting `obskit` for the warehouse package to standardise log fields and remove those duplicates from `logfield` package.
- I removed the dependency of the `logfield` from `marketobulkupload.go`.
- Also using logger.New* so that we can move to [Non-Sugared Structured format](https://rudderlabs.slack.com/archives/C02LNLLAABC/p1708346761796859) easily in future.

## Linear Ticket

- Resolves PIPE-1383

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/marketobulkupload.go
+++ b/router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/marketobulkupload.go
@@ -99,7 +99,7 @@ func (b *MarketoBulkUploader) Poll(pollInput common.AsyncPoll) common.PollStatus
 		b.logger.Errorw("[Batch Router] Failed to fetch status for",
 			obskit.DestinationType("MARKETO_BULK_UPLOAD"),
 			"body", string(bodyBytes[:512]),
-			"errpr", asyncResponse.Error,
+			"error", asyncResponse.Error,
 		)
 		return common.PollStatusResponse{
 			StatusCode: 500,
--- a/router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/marketobulkupload.go
+++ b/router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/marketobulkupload.go
@@ -12,15 +12,16 @@ import (
 
 	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/logger"
+	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
 
 	"github.com/tidwall/gjson"
 
 	"github.com/rudderlabs/rudder-go-kit/stats"
+
 	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 	"github.com/rudderlabs/rudder-server/jobsdb"
 	"github.com/rudderlabs/rudder-server/router/batchrouter/asyncdestinationmanager/common"
 	"github.com/rudderlabs/rudder-server/utils/misc"
-	lf "github.com/rudderlabs/rudder-server/warehouse/logfield"
 )
 
 type MarketoBulkUploader struct {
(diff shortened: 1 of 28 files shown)


router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/marketobulkupload.go
```diff
@@ -96,9 +97,9 @@ func (b *MarketoBulkUploader) Poll(pollInput common.AsyncPoll) common.PollStatus
 
 	if asyncResponse.Error != "" {
 		b.logger.Errorw("[Batch Router] Failed to fetch status for",
-			lf.DestinationType, "MARKETO_BULK_UPLOAD",
+			obskit.DestinationType("MARKETO_BULK_UPLOAD"),
 			"body", string(bodyBytes[:512]),
-			lf.Error, asyncResponse.Error,
+			"errpr", asyncResponse.Error,
```
fracasula, line 102: There seems to be a typo here.

router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/marketobulkupload.go
```diff
@@ -96,9 +97,9 @@ func (b *MarketoBulkUploader) Poll(pollInput common.AsyncPoll) common.PollStatus
 
 	if asyncResponse.Error != "" {
 		b.logger.Errorw("[Batch Router] Failed to fetch status for",
```
fracasula, line 99: Shouldn't you be using the non-sugared logger? i.e. `Errorn` and not `Errorw`?

warehouse/api/grpc.go
```diff
@@ -539,11 +539,11 @@ func (g *GRPC) RetryWHUploads(ctx context.Context, req *proto.RetryWHUploadsRequ
...
-		lf.WorkspaceID, req.WorkspaceId,
-		lf.SourceID, req.SourceId,
-		lf.DestinationID, req.DestinationId,
-		lf.DestinationType, req.DestinationType,
-		lf.IntervalInHours, req.IntervalInHours,
+		obskit.WorkspaceID(req.WorkspaceId),
+		obskit.SourceID(req.SourceId),
+		obskit.DestinationID(req.DestinationId),
+		obskit.DestinationType(req.DestinationType),
+		logger.NewField(lf.IntervalInHours, req.IntervalInHours),
```
fracasula, line 546: Why `NewField`? Can't you use a "typed" field instead?

router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/marketobulkupload.go
achettyiitr: Yeah, wanted to move to the non-sugared logger as well but the changes will be more. I will create a separate PR for it.

router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/marketobulkupload.go
achettyiitr: OOps.

achettyiitr: Fixing it.

warehouse/api/grpc.go
fracasula: It's an `int64` so `logger.NewIntField` would be the best here. `NewField` is like returning `any` basically.

The whole thing about non-sugared logging is that you always infer the type, this way it does not have to use reflection and that's when you really have a benefit in terms of performance.

So TL;DR, always infer the type and use the `n` version of the methods (e.g. `Withn`, `Infon`, `Errorn`, `Warnn`, etc...).

There is no check around this for performance and complexity reasons so you'll have to be aware and do it. The most we could do is build a linter around this.

achettyiitr: Yeah, we can use NewIntField. Missed it.

router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/marketobulkupload.go
fracasula: Did you see how the logger actually works though?

https://github.com/rudderlabs/rudder-go-kit/blob/main/logger/field.go#L113

If you want to use the field type with the sugared logger then you should use `Expand`. I suggest changing all in this PR so that you don't have to use `Expand` and then remove it when you migrate to `Errorn`.

PR #4977, opened 2024-08-04
feat: users table loading from user load files

achettyiitr:
# Description

- Decoupling or loading of users table from identifies table:
  - Currently, we are querying the `identifies` table for staging data.
  - So, instead of this, we can use `users` load files to populate that data.
- Injecting middleware while creating the client itself instead of using the `getMiddleware` function which is prone to race conditions.

## Linear Ticket

- Resolves PIPE-1347
- Resolves PIPE-1421

## Security

### DIFF
--- a/warehouse/integrations/bigquery/bigquery.go
+++ b/warehouse/integrations/bigquery/bigquery.go
@@ -12,10 +12,12 @@ import (
 
 	"cloud.google.com/go/bigquery"
 	"github.com/samber/lo"
-	bqService "google.golang.org/api/bigquery/v2"
+	bqservice "google.golang.org/api/bigquery/v2"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
+
+	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
 
 	"github.com/rudderlabs/rudder-go-kit/stats"
 
@@ -33,13 +35,13 @@ import (
 )
 
 type BigQuery struct {
-	db         *bigquery.Client
-	middleware *middleware.Client
-	namespace  string
-	warehouse  model.Warehouse
-	projectID  string
-	uploader   warehouseutils.Uploader
-	logger     logger.Logger
+	db        *middleware.Client
+	namespace string
+	warehouse model.Warehouse
+	projectID string
+	uploader  warehouseutils.Uploader
+	conf      *config.Config
+	logger    logger.Logger
 
 	config struct {
 		setUsersLoadPartitionFirstEventFilter bool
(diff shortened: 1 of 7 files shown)


warehouse/integrations/bigquery/bigquery_test.go
```diff
@@ -913,14 +1145,3 @@ func appendEventsMap() whth.EventsCountMap {
...
-func TestUnsupportedCredentials(t *testing.T) {
-	credentials := whbigquery.BQCredentials{
-		ProjectID:   "projectId",
-		Credentials: "{\"installed\":{\"client_id\":\"1234.apps.googleusercontent.com\",\"project_id\":\"project_id\",\"auth_uri\":\"https://accounts.google.com/o/oauth2/auth\",\"token_uri\":\"https://oauth2.googleapis.com/token\",\"auth_provider_x509_cert_url\":\"https://www.googleapis.com/oauth2/v1/certs\",\"client_secret\":\"client_secret\",\"redirect_uris\":[\"urn:ietf:wg:oauth:2.0:oob\",\"http://localhost\"]}}",
-	}
-
-	_, err := whbigquery.Connect(context.Background(), &credentials)
-	assert.NotNil(t, err)
-	assert.Contains(t, err.Error(), "client_credentials.json file is not supported")
-}
```
achettyiitr, line 926: This test is moved to `credentials_test.go`

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -33,13 +35,13 @@ import (
 )
 
 type BigQuery struct {
-	db         *bigquery.Client
-	middleware *middleware.Client
```
achettyiitr, line 37: During connect itself we are setting it, to avoid haveing 2 clients: `db` and `middleware`. Also, we don't need the `getMiddleware() *middleware.Cli`

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -137,54 +139,44 @@ func New(conf *config.Config, log logger.Logger) *BigQuery {
...
+		obskit.Namespace(bq.namespace),
+		logger.NewStringField(logfield.TableName, tableName),
+	)
 	sampleSchema := getTableSchema(columnMap)
 	metaData := &bigquery.TableMetadata{
-		Schema:           sampleSchema,
-		TimePartitioning: &bigquery.TimePartitioning{},
+		Schema: sampleSchema,
+		TimePartitioning: &bigquery.TimePartitioning{
+			Type: bigquery.DayPartitioningType,
```
achettyiitr, line 169: Specifying the partition instead of relying on the default value set by the SDK. It is more understandable now, from the code what kind of partition type we do.

cisse21 (approved):

PR #4992, opened 2024-08-09
chore: update destination id label to be consistent with other modules

itsmihir:
# Description

We use different destination label keys for `event_delivery_time` metric for different modules, for router and warehouse we are using `destID` and for batch_router we are using `destinationId`. 
Changing destination id label for batch_router module to be consistent with other modules.

## Linear Ticket

Resolves OBS-586
https://linear.app/rudderstack/issue/OBS-586/update-labels-in-event-delivery-time-metric

## Security

### DIFF
--- a/router/batchrouter/handle_observability.go
+++ b/router/batchrouter/handle_observability.go
@@ -193,7 +193,7 @@ func (brt *Handle) recordUploadStats(destination Connection, output UploadResult
 			"destination":    destinationTag,
 			"workspaceId":    destination.Source.WorkspaceID,
 			"sourceId":       destination.Source.ID,
-			"destinationId":  destination.Destination.ID,
+			"destID":         destination.Destination.ID,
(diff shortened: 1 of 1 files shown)


mihir20 (commented):
Maybe we can use `destinationId` everywhere. Also we might need to update the grafana dashboard.

itsmihir:
> Maybe we can use `destinationId` everywhere. Also we might need to update the grafana dashboard.

For `batch_router` destination id label was added only 4 months ago in this [PR](https://github.com/rudderlabs/rudder-server/pull/4559/files#diff-667c9fe5d3c4b76a225050414db141a9cd6863ae6ceca580e6a5ace439daed50), so its not being using anywhere. But if we have to change the label key to "destinationId" everywhere else, then it will required more change.

satishrudderstack (approved):

cisse21 (approved):

PR #4999, opened 2024-08-14
chore: sample events based on stage of failure

Sidddddarth:
# Description

Sends input payload received at each router stage to better inform debugging efforts in case of failures.
A transformation failure will report a sample payload that's received by the transformer(jobsdb payload) whereas the delivery failure will report either the transformed payload or simply jobsdb payload if there's no transformation involved.

Also includes the stage at which an event failed to the same end. This will be part of the sample response(field `failedStage`).
To that effect it made sense to do the same in case of successes as well, so the transformed payload is reported if there's a transformation involved.

## Linear Ticket

[Resolves OBS-585](https://linear.app/rudderstack/issue/OBS-585/router-module-destination-delivery-failure-show-the-post-dest)
Resolves PIPE-1277
Resolves PIPE-1120

## Security


cisse21 (approved):

router/worker.go
```diff
@@ -979,6 +985,15 @@ func (w *worker) postStatusOnResponseQ(respStatusCode int, payload json.RawMessa
...
+	switch errorAt {
+	case routerutils.ERROR_AT_TF:
+		inputPayload = destinationJobMetadata.JobT.EventPayload
+		status.ErrorResponse = misc.UpdateJSONWithNewKeyVal(status.ErrorResponse, "failureStage", "RudderStack Transformation Error")
+	default: // includes ERROR_AT_DEL, ERROR_AT_CUST
+		status.ErrorResponse = misc.UpdateJSONWithNewKeyVal(status.ErrorResponse, "failureStage", "Destination Error")
+	}
+
 	status.ErrorResponse = routerutils.EnhanceJSON(status.ErrorResponse, "firstAttemptedAt", firstAttemptedAtTime.Format(misc.RFC3339Milli))
 	status.ErrorResponse = routerutils.EnhanceJSON(status.ErrorResponse, "content-type", respContentType)
```
fracasula, line 998: Would it make sense to always include `errorAt` as part ot the `ErrorResponse` to aid with debugging?

router/worker.go
```diff
@@ -988,47 +1003,59 @@ func (w *worker) postStatusOnResponseQ(respStatusCode int, payload json.RawMessa
...
+	// Saving payload to DB only
+	// 1. if job failed and
+	// 2. if router job undergoes batching or dest transform.
+	if payload != nil && (w.rt.enableBatching || destinationJobMetadata.TransformAt == "router") {
+		if w.rt.reloadableConfig.savePayloadOnError.Load() {
+			status.ErrorResponse = routerutils.EnhanceJSON(status.ErrorResponse, "payload", string(payload))
 		}
+	}
+	// the job failed
+	w.logger.Debugf("Job failed to send, analyzing...")
```
fracasula, line 1024: Please **always** use the `n` (non-sugared) logger even if you don't pass any parameters. We do need a linter for this, but in the meantime try to remember about these please so that we won't have to work too much when we have a linter in place.

```suggestion
	w.logger.Debugn("Job failed to send, analyzing...")
```

fracasula (approved):

router/worker.go
Sidddddarth: In the other cases, the job states and `reason` inform us about why that job failed.

Sidddddarth: I mean to say all cases are already pretty well covered to inform an observer as to why jobs failed at a particular stage.

PR #5011, opened 2024-08-20
fix: terminate loop after long-running transformation is detected

mihir20:
# Description
terminate loop after long-running transformation is detected

## Linear Ticket

-

## Security

### DIFF
--- a/processor/transformer/transformer.go
+++ b/processor/transformer/transformer.go
@@ -552,6 +552,7 @@ func trackLongRunningTransformation(ctx context.Context, stage string, timeout t
 			log.Errorw("Long running transformation detected",
 				"stage", stage,
 				"duration", time.Since(start).String())
+			return // ensure the loop terminates after logging
 		}
 	}
 }


processor/transformer/transformer.go
```diff
@@ -552,6 +552,7 @@ func trackLongRunningTransformation(ctx context.Context, stage string, timeout t
 			log.Errorw("Long running transformation detected",
 				"stage", stage,
 				"duration", time.Since(start).String())
+			return // ensure the loop terminates after logging
```
Sidddddarth, line 555: It's bound to get closed already because the context passed is cancelled.
Am I missing something..?

mihir20: If context is not cancelled and transformation is running for more than the configured timeout this loop will get stuck in infinite loop until context is cancelled.

mihir20: I will not be infinite but still the loop with be stuck and there will be a goroutine that will be running and consuming resources until context is cancelled.

Sidddddarth: we could also remove the `for {...}`

Sidddddarth (approved):

PR #5024, opened 2024-08-25
chore: trino hive cursor error

achettyiitr:
# Description

- Trino HIVE_CURSOR_ERROR when **parquet_use_column_index=true** and **parquet_ignore_statistics=false**.
- Succeeds when either of **parquet_use_column_index=false** or **parquet_ignore_statistics=true**

## Linear Ticket

- Resolves PIPE-1426

## Security


fracasula (commented):
@achettyiitr So we just needed to know a way to get around this error and you're demonstrating that with this PR? No changes needed?

achettyiitr:
> @achettyiitr So we just needed to know a way to get around this error and you're demonstrating that with this PR? No changes needed?

Yeah. Currently, we are not making any changes to the Parquet Library. It would need more time to migrate to a different library.

fracasula (approved):

cisse21 (approved):

PR #5028, opened 2024-08-26
fix: error reporting lag and pii handle

sanpj2292:
# Description

There is a recursive payload that is trying to get stored into `error_detail_reports` table. Since the payload is large and recursive in nature, we are having difficulty in reading it from postgres which is causing delays.

For now, we are planning two things
1. Introduce a lag metric, have an alert configured on it whenever such a delay happens(while trying to fetch data from postgres)
2. For not storing sample response, we have a configuration in rudderstack UI which was not being used for `error_detail_reports` table as it was not storing sample responses. Recently we started storing it. We are planning to respect this configuration for `error_detail_reports` table as well.

## Linear Ticket

Resolves INT-2582

## Security


enterprise/reporting/error_reporting.go
```diff
@@ -317,82 +338,97 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
...
-	for {
-		if ctx.Err() != nil {
-			edr.log.Infof("stopping mainLoop for syncer %s : %s", c.Label, ctx.Err())
-			return
-		}
-		requestChan := make(chan struct{}, edr.maxConcurrentRequests.Load())
-		loopStart := time.Now()
-		currentMs := time.Now().UTC().Unix() / 60
+	var lastReportedAtTime atomic.Time
+	lastReportedAtTime.Store(time.Now())
```
itsmihir, line 342: We need to update `lastReportedAtTime` every time we flush reports to reporting. We can use the output from [this](https://github.com/rudderlabs/rudder-server/pull/5028/files?diff=split&w=1#diff-6067339d2725ff3da3cb6c33d9b28a5becfb9278bd1e5ff14a364bece14584b7R470) query to update the `lastReportedAtTime`

enterprise/reporting/error_reporting.go
```diff
@@ -199,6 +217,10 @@ func (edr *ErrorDetailReporter) Report(ctx context.Context, metrics []*types.PUR
 
 		workspaceID := edr.configSubscriber.WorkspaceIDFromSource(metric.ConnectionDetails.SourceID)
 		metric := *metric
+		if !edr.IsPIIReportingDisabled(workspaceID) {
+			edr.log.Debugf("PII setting is disabled for workspaceId: %s", workspaceID)
```
achettyiitr, line 221: We can use a structured log in here.
```suggestion
			edr.log.Debugn("PII setting disabled", obskit.WorkspaceID(workspaceID))
```

enterprise/reporting/error_reporting.go
```diff
@@ -199,6 +217,10 @@ func (edr *ErrorDetailReporter) Report(ctx context.Context, metrics []*types.PUR
 
 		workspaceID := edr.configSubscriber.WorkspaceIDFromSource(metric.ConnectionDetails.SourceID)
 		metric := *metric
+		if !edr.IsPIIReportingDisabled(workspaceID) {
```
achettyiitr, line 220: Shouldn't it be the other way around? If it is disabled then don't report and return from here.
```suggestion
		if edr.IsPIIReportingDisabled(workspaceID) {
```

enterprise/reporting/error_reporting.go
sanpj2292: This is done. Can you check once ?

mihir20 (commented):
Can we add some tests on when reporting is enabled and when it is disabled?

sanpj2292:
> Can we add some tests on when reporting is enabled and when it is disabled?

Tests have been added. Can you check & let me know ?

enterprise/reporting/error_reporting.go
```diff
@@ -163,6 +166,22 @@ func (edr *ErrorDetailReporter) DatabaseSyncer(c types.SyncerConfig) types.Repor
 	}
 }
 
+func (edr *ErrorDetailReporter) emitLagMetric(ctx context.Context, c types.SyncerConfig, lastReportedAtTime *atomic.Time) error {
```
koladilip, line 169: I didn't get what is happening here, what should this metric capture?

enterprise/reporting/error_reporting.go
```diff
@@ -199,8 +218,12 @@ func (edr *ErrorDetailReporter) Report(ctx context.Context, metrics []*types.PUR
 
 		workspaceID := edr.configSubscriber.WorkspaceIDFromSource(metric.ConnectionDetails.SourceID)
 		metric := *metric
+		if edr.IsPIIReportingDisabled(workspaceID) {
+			edr.log.Debugn("PII setting is disabled for workspaceId:", obskit.WorkspaceID(workspaceID))
+			return nil
+		}
 		destinationDetail := edr.configSubscriber.GetDestDetail(metric.ConnectionDetails.DestinationID)
-		edr.log.Debugf("For DestId: %v -> DestDetail: %v", metric.ConnectionDetails.DestinationID, destinationDetail)
+		edr.log.Debugn("DestinationId & DestDetail details", obskit.DestinationID(metric.ConnectionDetails.DestinationID), logger.NewField("destinationDetail", destinationDetail))
```
koladilip, line 226: can you provide one sample log for this

koladilip (approved):

ktgowtham:
@satishrudderstack , can you look into this PR?

mihir20 (approved):
LGTM

enterprise/reporting/error_reporting.go
```diff
@@ -317,82 +339,109 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
...
+					return ctx.Err()
+				case <-time.After(edr.mainLoopSleepInterval.Load()):
+				}
+				continue
 			}
-			errGroup.Go(func() error {
-				err := edr.sendMetric(errCtx, c.Label, metricToSend)
-				if err != nil {
-					edr.log.Error("Error while sending to Reporting service:", err)
+			lastReportedAtTime.Store(time.Unix(reportedAt*60, 0))
```
itsmihir, line 378: let's move this after checking for `reports` length to be zero (after line 391).

itsmihir (changes requested):

enterprise/reporting/error_reporting.go
```diff
@@ -317,82 +339,109 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
...
+			getReportsTimer.Since(getReportsStart)
+			getReportsSize.Observe(float64(len(reports)))
+
+			if len(reports) == 0 {
+				lastReportedAtTime.Store(loopStart)
+				select {
+				case <-ctx.Done():
+					edr.log.Infof("stopping mainLoop for syncer %s : %s", c.Label, ctx.Err())
+					return nil
```
itsmihir, line 387: lets be consistent and return `ctx.Err()` (from line 437 too)

sanpj2292: updated. can you check now ?

itsmihir (approved):
LGTM

PR #5035, opened 2024-08-28
chore: add trufflehog scan step before image push

lvrach:
# Description

trufflehog scans for leaked secrets before image is pushed to the docker hub.

Here is an example of a [failed](https://github.com/rudderlabs/rudder-server/actions/runs/10612852969/job/29415431999?pr=5035) run, because a secret was detected 

`make sec` is added, to execute the docker scan locally.

### Build without push
First, we run `uses: docker/build-push-action@v6`, using the same attributes as the existing build & push, but:
1. `push: false` we don't want to push the image before scan it first
2. `cache-to: type=local,dest=/tmp/.buildx-cache,mode=max,tag=latest`. This allows us to cache all the image layers, so the next `build-push-action` reuses every layer, so we don't have to rebuild again.
3. `outputs: type=docker,dest=/tmp/local-rudder-server.tar`. Instruct docker to export the docker image as a tar. This is required to call trufflehog, without publishing the image on the registry

### Run TruffleHog
We use the latest docker image trufflehog.
* mounting `-v /tmp:/tmp` so it can scan the tar image file `/tmp/local-rudder-server.tar`
* `--no-verification` uncovers all possible secrets, not just the owns that can be verified. 
* `--fail` if a secret is detected, the step will fail

### Build and Push

Added the following, on the existing `build-push-action` step:
* Added `cache-from: type=local,src=/tmp/.buildx-cache,tag=latest`, to reuse the layer from previous build
* `if: success()` ensuring this step will not run if previous step fails 

## Linear Ticket

https://linear.app/rudderstack/issue/PRI-28/introduce-and-enforce-secret-scanning-checks-as-part-of-our-image

Refs PRI-28

## Security


.github/workflows/builds.yml
```diff
@@ -107,7 +107,32 @@ jobs:
...
+          push: false
+          tags: ${{ matrix.build-config.tags }}
+          labels: ${{ matrix.build-config.labels }}
+          build-args: |
+            BUILD_DATE=${{ needs.docker-oss-meta.outputs.build-date }}
+            VERSION=${{ needs.docker-oss-meta.outputs.version }}
+            COMMIT_HASH=${{ github.sha }}
+            REVISION=${{ needs.docker-oss-meta.outputs.revision }}
+          outputs: type=docker,dest=/tmp/local-rudder-server.tar
+          cache-to: type=local,dest=/tmp/.buildx-cache,mode=max,tag=latest
```
mihir20, line 124: should we use some other tag here instead of latest?

lvrach: We could do that, but it doesn't really matter as long as cache-to and cache-from tags matching. It only affects how things are stored/retrieved from this temporary cache

lvrach: Does it make sense if I rename it to `temp` ?

achettyiitr (approved):

.github/workflows/builds.yml
mihir20: The only doubt i have is if 2 workflows run on same machine, then is there a possibility tags may get overriden by other workflows?

lvrach: Good point. I opted for `${{ runner.temp }}` instead of `/tmp` because it offers more reliable behavior in GitHub Actions workflows. Here's why:

1. Guaranteed uniqueness: `${{ runner.temp }}` provides a unique directory for each job, preventing potential conflicts.
2. Automatic cleanup: GitHub ensures this directory is cleaned after each job, reducing leftover artifacts.
3. Consistent across platforms: It works uniformly across different runner operating systems.

PR #5041, opened 2024-08-29
fix: take tacking plan id from dgsourceTPconfig

akashrpo:
# Description

take tacking plan id from dgsourceTPconfig instead of context.ruddertyper config

## Linear Ticket

[DAT-1551](https://linear.app/rudderstack/issue/DAT-1551/typer-rudder-server-take-tacking-plan-id-from-dgsourcetpconfig-instead)

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1832,19 +1832,12 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
 				&backendconfig.DestinationT{},
 			)
 
-			trackingPlanID := source.DgSourceTrackingPlanConfig.TrackingPlan.Id
 			trackingPlanVersion := source.DgSourceTrackingPlanConfig.TrackingPlan.Version
-			rudderTyperTPID := misc.MapLookup(singularEvent, "context", "ruddertyper", "trackingPlanId")
 			rudderTyperTPVersion := misc.MapLookup(singularEvent, "context", "ruddertyper", "trackingPlanVersion")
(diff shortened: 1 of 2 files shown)


processor/processor.go
```diff
@@ -1831,19 +1831,12 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
...
-			if rudderTyperTPID != nil && rudderTyperTPVersion != nil {
-				if id, ok := rudderTyperTPID.(string); ok && id != "" {
-					trackingPlanID = id
-				}
-				if version, ok := rudderTyperTPVersion.(float64); ok && version > 0 {
-					trackingPlanVersion = int(version)
-				}
+			if version, ok := rudderTyperTPVersion.(float64); ok && version > 0 && rudderTyperTPVersion != nil {
+				trackingPlanVersion = int(version)
 			}
```
achettyiitr, line 1838: We would need to check for nil first? 
```suggestion
			if rudderTyperTPVersion != nil {
				if version, ok := rudderTyperTPVersion.(float64); ok && version > 0 {
					trackingPlanVersion = int(version)
				}
			}
```

processor/processor_test.go
```diff
@@ -984,11 +984,11 @@ var _ = Describe("Tracking Plan Validation", Ordered, func() {
 					},
 				},
 			)
+			Expect(err).To(BeNil())
```
achettyiitr, line 987: We don't need this assertion. We are already checking this above.

achettyiitr (approved):

Jayachand (approved):

PR #5063, opened 2024-09-05
fix: ingore webhook headers support when body is json array

koladilip:
# Description

The webhook headers feature was added in release 1.33 but it is causing issues for sendgrip webhooks as they are sending the request body as a JSON array so we will ignore propagating headers for such cases.

## Linear Ticket

https://linear.app/rudderstack/issue/INT-2504/webhook-source-deckers-has-requested-the-ability-to-forward-the-x-dash

## Security


sanpj2292 (approved):

cisse21 (approved):

koladilip:
tracking here https://github.com/rudderlabs/rudder-server/pull/5064

PR #5064, opened 2024-09-06
fix: revert support for webhook headers

koladilip:
# Description
The webhook headers feature was added in release 1.33 but it is causing issues for SendGrid webhooks as they are sending the request body as a JSON array so we are reverting the change and we will revisit the logic later

## Linear Ticket
https://linear.app/rudderstack/issue/INT-2504/webhook-source-deckers-has-requested-the-ability-to-forward-the-x-dash

## Security

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -80,7 +80,7 @@ require (
 	github.com/rudderlabs/rudder-go-kit v0.38.0
 	github.com/rudderlabs/rudder-observability-kit v0.0.3
 	github.com/rudderlabs/rudder-schemas v0.5.1
-	github.com/rudderlabs/rudder-transformer/go v0.0.0-20240812044419-23196ec42acf
+	github.com/rudderlabs/rudder-transformer/go v0.0.0-20240906042448-f7783d8fb300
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -80,7 +80,7 @@ require (
 	github.com/rudderlabs/rudder-go-kit v0.38.0
 	github.com/rudderlabs/rudder-observability-kit v0.0.3
 	github.com/rudderlabs/rudder-schemas v0.5.1
-	github.com/rudderlabs/rudder-transformer/go v0.0.0-20240812044419-23196ec42acf
+	github.com/rudderlabs/rudder-transformer/go v0.0.0-20240906042448-f7783d8fb300
```
lvrach, line 83: Do we need to revert test cases as well ?

lvrach: My bad, we are moving forward

lvrach (approved):

go.mod
koladilip: these are failing so just upgrade the transformer version

PR #5067, opened 2024-09-09
chore: update router reports payload behind a flag, emit stats to observe sizes

Sidddddarth:
# Description

Emit stats to observe payload sizes:
1. payload in router jobsdb
2. payload being delivered

Puts behaviour from #4999 behind a feature flag(default behaviour is to send jobsdb payload for now)
Adds a `payloadStage` field to error response field in the job status.

## Linear Ticket

[Resolves PIPE-1504](https://linear.app/rudderstack/issue/PIPE-1504/reporting-capture-relevant-payload-and-stage-for-router)

## Security


router/worker.go
```diff
@@ -985,14 +985,18 @@ func (w *worker) postStatusOnResponseQ(respStatusCode int, payload json.RawMessa
...
+	if !w.rt.reportJobsdbPayload.Load() { // TODO: update default/remove this flag after monitoring the payload sizes
+		inputPayload = payload
+	}
+	tags := stats.Tags{
+		"destType":      w.rt.destType,
+		"workspaceID":   destinationJobMetadata.WorkspaceID,
+		"state":         status.JobState,
+		"destinationID": destinationJobMetadata.DestinationID,
 	}
+	stats.Default.NewTaggedStat("router_transformation_payload_size_bytes", stats.HistogramType, tags).Observe(float64(len(destinationJobMetadata.JobT.EventPayload)))
```
ktgowtham, line 998: Why do we need `router_transformation_payload_size_bytes`? Also it's router input payload rt. If we're planning to capture, may be capture at early stage?

ktgowtham: Also, we don't need to capture this multiple times for each router job. We can capture this once per destinationJob.

router/worker.go
```diff
@@ -1012,6 +1016,16 @@ func (w *worker) postStatusOnResponseQ(respStatusCode int, payload json.RawMessa
 		}
 		return
 	}
+	switch errorAt {
```
ktgowtham, line 1019: Can we add additional check of status code and set only if it's failed. just as a safe check. even though this function gets returned if it's success before this block

ktgowtham (approved):

mihir20 (approved):

cisse21 (approved):

PR #5074, opened 2024-09-09
fix: prevent badgerdb hang when closing without prior operations

lvrach:
# Description

## Issue

When transitioning a newly created rudder-server instance that has received no traffic from normal to degraded mode, the process hangs. Causing migration to stuck as well, until instance is restarted, upon restart no transition is required, as rudder-server starts on degrade mode.

Taking a goroutine dump:
```
goroutine 786 [chan receive, 3 minutes]:
github.com/rudderlabs/rudder-server/services/dedup/badger.(*BadgerDB).Close(...)
	/rudder-server/services/dedup/badger/badger.go:118
github.com/rudderlabs/rudder-server/services/dedup/badger.(*Dedup).Close(0x0?)
	/rudder-server/services/dedup/badger/badger.go:230 +0x38
github.com/rudderlabs/rudder-server/processor.(*Handle).Shutdown(0x4001b08008)
	/rudder-server/processor/processor.go:766 +0x50
github.com/rudderlabs/rudder-server/processor.(*LifecycleManager).Stop(0x4000f348c0)
```
You can see that badger.Close hangs, preventing the shutdown to be completed.

I was able to replicate this locally, using the following test:

```go
func TestBadgerClose(t *testing.T) {
	badger := NewBadgerDB(config.New(), stats.NOP, t.TempDir())
	require.NotNil(t, badger)

	t.Log("close badger without any other operation")
	badger.Close()
}
```

### Root-cause

Close was waiting for `d.gcDone` to close, but this will never happen, because `init` was never called.

## Solution

Naive approach is to ensure init is called at the start of close method.  However, this is counter intuitive, and also handling error for init would be needed. Also, in case init error, gc loop go routine won't be call as well.

A safer approach is to introduce a waitgroup, that tracks for any pending background go routines, and won't block if non have been started.

Also, `bgCtx` is introduced as a more conventional wait to pass termination.   
 

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1420/rudder-server-did-not-transition-to-degraded-mode

resolves PIPE-1420

## Security

### DIFF
--- a/services/dedup/badger/badger.go
+++ b/services/dedup/badger/badger.go
@@ -129,8 +129,6 @@ func (d *BadgerDB) init() error {
 	var err error
 
 	d.once.Do(func() {
-		d.wg = sync.WaitGroup{}
-
 		d.badgerDB, err = badger.Open(d.opts)
 		if err != nil {
 			return
--- a/services/dedup/badger/badger.go
+++ b/services/dedup/badger/badger.go
@@ -114,6 +114,7 @@ func (d *BadgerDB) Set(kvs []types.KeyValue) error {
 }
 
 func (d *BadgerDB) Close() {
+	d.init()
 	close(d.close)
 	<-d.gcDone
 	_ = d.badgerDB.Close()
--- a/services/dedup/badger/badger.go
+++ b/services/dedup/badger/badger.go
@@ -1,6 +1,7 @@
 package badger
 
 import (
+	"context"
 	"fmt"
 	"strconv"
 	"sync"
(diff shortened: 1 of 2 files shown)


Sidddddarth (approved):

cisse21 (approved):

services/dedup/badger/badger.go
```diff
@@ -114,22 +118,27 @@ func (d *BadgerDB) Set(kvs []types.KeyValue) error {
...
+		_ = d.badgerDB.Close()
+	}
 }
 
 func (d *BadgerDB) init() error {
 	var err error
 
 	d.once.Do(func() {
+		d.wg = sync.WaitGroup{}
+
```
achettyiitr, line 133: Do we need to overwrite wg? We are already setting in NewBadgerDB?

PR #5089, opened 2024-09-12
fix: gateway internal batch endpoint stats

Sidddddarth:
# Description

Fix some stats in gateway's internal batch endpoint.

## Linear Ticket

[Resolves PIPE-1520](https://linear.app/rudderstack/issue/PIPE-1520/fix-gateway-metrics-for-batch-endpoint-which-is-called-from-ingestion)

## Security


fracasula (approved):
LGTM, although if possible it would be nice to have an assertion in the gateway tests where we check explicitly that the required metrics (e.g. `gateway_write_key_events`) are correctly increased.

cisse21 (approved):

ktgowtham (approved):

PR #5091, opened 2024-09-14
chore: schema and record assertions in warehouse integration test

achettyiitr:
# Description

- Schema and Data validations for warehouse integration tests.
- Instead of running the integration tests with the pre-prepared stale staging files, we are now using events, sending them to the transformer and then. preparing the staging file.
- Same events file (`warehouse/integrations/testdata/*-job.events-*.json`) is being used in all integrations.

## Linear Ticket

- Resolves PIPE-1526

## Security


warehouse/integrations/azure-synapse/azure-synapse.go
```diff
@@ -628,7 +628,7 @@ func (as *AzureSynapse) loadUserTables(ctx context.Context) (errorMap map[string
 						  	select top 1 %[1]q from %[2]s
 						  	where x.id = %[2]s.id
 							  and %[1]q is not null
-							order by X.received_at desc
+							order by received_at desc
```
achettyiitr, line 631: Got this bug after writing integration tests.

cisse21 (approved):

lvrach (approved):

PR #5109, opened 2024-09-18
fix: each router handle stores the entire connections map

Sidddddarth:
# Description

Each router handle stored the connections for all destinations, even those of which were not of the same destType!
It's now filtered so router handles only store connection details of destinations of the same destType.

## Linear Ticket

[Resolves PIPE-1540](https://linear.app/rudderstack/issue/PIPE-1540/shared-backend-config-across-router-instances)

## Security


ktgowtham (approved):

mihir20 (approved):
LGTM

lvrach (approved):

cisse21 (approved):

PR #5120, opened 2024-09-20
chore: no full vacuum for various reports tables

Sidddddarth:
# Description

1. Using `vacuum` instead of `vacuum full` for various reports tables. This only takes lenient locking and doesn't disrupt other reads/writes to the tables. And while it doesn't free up disk space, it makes the space available for newer tuples added to the table.
2. Full vacuum on startup
3. vacuum threshold based on number of deleted rows.

Having done all this, the table size might still grow, but not at the rate when there was no vacuum at all. Each server startup would help free up disk space.

## Linear Ticket

[Resolves PIPE-1517](https://linear.app/rudderstack/issue/PIPE-1549/remove-full-vacuum)

## Security

### DIFF
--- a/enterprise/reporting/error_reporting.go
+++ b/enterprise/reporting/error_reporting.go
@@ -460,7 +460,7 @@ func (edr *ErrorDetailReporter) vacuum(ctx context.Context, dbHandle *sql.DB, c 
 	if sizeEstimate > edr.config.GetInt64("Reporting.errorReporting.vacuumThresholdBytes", 5*bytesize.GB) {
 		vacuumStart := time.Now()
 		vacuumDuration := edr.stats.NewTaggedStat(StatReportingVacuumDuration, stats.TimerType, tags)
-		vaccumStatement := fmt.Sprintf("vacuum full analyze %s", pq.QuoteIdentifier(ErrorDetailReportsTable))
+		vaccumStatement := fmt.Sprintf("vacuum analyze %s", pq.QuoteIdentifier(ErrorDetailReportsTable))
 		if _, err := dbHandle.ExecContext(ctx, vaccumStatement); err != nil {
 			edr.log.Errorn(
 				fmt.Sprintf(`Error vacuuming %s table`, ErrorDetailReportsTable),
--- a/enterprise/reporting/error_reporting.go
+++ b/enterprise/reporting/error_reporting.go
@@ -459,21 +459,27 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
 					deletedRows += len(reports)
 				}
 				// vacuum error_reports_details table
-				var sizeEstimate int64
-				if err := dbHandle.QueryRowContext(
-					ctx,
-					`SELECT pg_table_size(oid) from pg_class where relname = $1`, ErrorDetailReportsTable,
-				).Scan(&sizeEstimate); err != nil {
-					edr.log.Errorn(
-						fmt.Sprintf(`Error getting %s table size estimate`, ErrorDetailReportsTable),
-						logger.NewErrorField(err),
-					)
-				}
-				if deletedRows >= vacuumDeletedRowThreshold.Load() ||
-					(sizeEstimate >= vacuumThresholdBytes.Load() && time.Since(lastVacuum) >= vacuumInterval.Load()) {
+				if deletedRows >= vacuumDeletedRowThreshold.Load() {
 					if err := edr.vacuum(ctx, dbHandle, tags); err == nil {
 						deletedRows = 0
 						lastVacuum = time.Now()
+					}
+				} else {
+					var sizeEstimate int64
+					if err := dbHandle.QueryRowContext(
+						ctx,
+						`SELECT pg_table_size(oid) from pg_class where relname = $1`, ErrorDetailReportsTable,
+					).Scan(&sizeEstimate); err != nil {
+						edr.log.Errorn(
+							fmt.Sprintf(`Error getting %s table size estimate`, ErrorDetailReportsTable),
+							logger.NewErrorField(err),
+						)
+					}
+					if sizeEstimate >= vacuumThresholdBytes.Load() && time.Since(lastVacuum) >= vacuumInterval.Load() {
+						if err := edr.vacuum(ctx, dbHandle, tags); err == nil {
+							deletedRows = 0
+							lastVacuum = time.Now()
+						}
 					}
 				}
 			}
--- a/enterprise/reporting/error_reporting.go
+++ b/enterprise/reporting/error_reporting.go
@@ -346,7 +346,6 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
 	edr.minReportedAtQueryTime = edr.stats.NewTaggedStat("error_detail_reports_min_reported_at_query_time", stats.TimerType, tags)
 	edr.errorDetailReportsQueryTime = edr.stats.NewTaggedStat("error_detail_reports_query_time", stats.TimerType, tags)
 	edr.edReportingRequestLatency = edr.stats.NewTaggedStat("error_detail_reporting_request_latency", stats.TimerType, tags)
-	vacuumDuration := edr.stats.NewTaggedStat(StatReportingVacuumDuration, stats.TimerType, tags)
 
 	var lastReportedAtTime atomic.Time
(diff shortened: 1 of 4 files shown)


Sidddddarth:
went with 100k because the peak rate is around 5k/s
image

Sidddddarth:
error details reports at a higher rate
image

enterprise/reporting/flusher/flusher.go
```diff
@@ -107,6 +106,15 @@ func NewFlusher(db *sql.DB, log logger.Logger, stats stats.Stats, conf *config.C
 
 	f.initCommonTags()
 	f.initStats(f.commonTags)
+	if _, err := db.Exec(
+		fmt.Sprintf("vacuum full analyze %s", pq.QuoteIdentifier(table)),
+	); err != nil {
+		log.Errorn(
+			fmt.Sprintf(`Error full vacuuming %s table`, table),
```
fracasula, line 113: The point of using a non-sugared logger is to avoid reflection (which happens with `fmt.Sprintf`. Also, please use the observability kit helpers whenever possible e.g. `obskit.Error`.

```go
log.Errorn("Cannot do full vacuum analyze", logger.NewStringField("table", table), obskit.Error(err))
```

fracasula (commented):
Could we potentially have this as a feature flag so that we can specify the type of vacuum we want?

enterprise/reporting/error_reporting.go
```diff
@@ -338,6 +346,7 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
 	edr.minReportedAtQueryTime = edr.stats.NewTaggedStat("error_detail_reports_min_reported_at_query_time", stats.TimerType, tags)
 	edr.errorDetailReportsQueryTime = edr.stats.NewTaggedStat("error_detail_reports_query_time", stats.TimerType, tags)
 	edr.edReportingRequestLatency = edr.stats.NewTaggedStat("error_detail_reporting_request_latency", stats.TimerType, tags)
+	vacuumDuration := edr.stats.NewTaggedStat(StatReportingVacuumDuration, stats.TimerType, tags)
```
fracasula, line 349: [optional] I know you're just copying this pattern, but do we really need to expose these constants?

enterprise/reporting/error_reporting.go
```diff
@@ -426,9 +436,33 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
...
+						}
+					} else {
+						if _, err := dbHandle.ExecContext(
+							ctx,
+							fmt.Sprintf("vacuum analyze %s", pq.QuoteIdentifier(ErrorDetailReportsTable)),
+						); err != nil {
+							edr.log.Errorn("error vacuuming", logger.NewStringField("table", ErrorDetailReportsTable), obskit.Error(err))
+						} else {
+							deletedRows = 0
+						}
```
fracasula, line 462: This is all duplicated stuff, right? The only thing that changes is the query.

Why not simplify a bit:

```go
// vacuum error_reports_details table
if deletedRows >= edr.config.GetInt("Reporting.errorReporting.vacuumThresholdDeletedRows", 100000) {
	err := edr.vacuumTable(ctx, dbHandle, edr.vacuumFull.Load(), vacuumDuration)
	if err == nil {
		deletedRows = 0
	}
}

// ...

func (edr *ErrorDetailReporter) vacuumTable(ctx context.Context, dbHandle *sql.DB, full bool, stat stats.Timer) error {
	defer stat.RecordDuration()()
	query := "vacuum analyze %s"
	if full {
		query = "vacuum full analyze %s"
	}
	_, err := dbHandle.ExecContext(ctx, fmt.Sprintf(query, pq.QuoteIdentifier(ErrorDetailReportsTable)))
	if err != nil {
		edr.log.Errorn("error vacuuming", logger.NewStringField("table", ErrorDetailReportsTable), obskit.Error(err))
	}
	return err
}
```

enterprise/reporting/reporting.go
```diff
@@ -449,27 +455,32 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
...
-				if err := dbHandle.QueryRowContext(
-					ctx,
-					fmt.Sprintf(`SELECT pg_table_size(oid) from pg_class where relname='%s';`, ReportsTable),
-				).Scan(&sizeEstimate); err != nil {
-					r.log.Errorn(
-						`[ Reporting ]: Error getting table size estimate`,
-						logger.NewErrorField(err),
-					)
-				}
-				if sizeEstimate > config.GetInt64("Reporting.vacuumThresholdBytes", 5*bytesize.GB) {
```
fracasula, line 465: Why drop the support based on size? It sounds very useful.

Shall we make it configurable instead of dropping it?
Like, if either the size threshold or the number of records was considerable then vacuum. Wdyt?

fracasula (changes requested):

enterprise/reporting/reporting.go
Sidddddarth: only removed the size constraint earlier because once the table gets to 5GB(or whatever configured value), it'll `vacuum analyze` every iteration of this loop.

fracasula: Right, we would need a `minimumInterval` to protect ourselves from this. Not hard to implement but also optional. Let's see what others think about it. If nobody gives an opinion I'm OK to drop it :+1:

enterprise/reporting/error_reporting.go
```diff
@@ -453,8 +459,18 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
 					deletedRows += len(reports)
 				}
 				// vacuum error_reports_details table
+				var sizeEstimate int64
+				if err := dbHandle.QueryRowContext(
+					ctx,
+					`SELECT pg_table_size(oid) from pg_class where relname = $1`, ErrorDetailReportsTable,
```
fracasula, line 465: Shall we run this query only if the deleted rows condition fails? If it succeeds there is no need to run it.

enterprise/reporting/error_reporting.go
```diff
@@ -426,9 +455,33 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
...
+					deletedRows += len(reports)
 				}
 				// vacuum error_reports_details table
-				edr.vacuum(ctx, dbHandle, c)
+				if deletedRows >= vacuumDeletedRowThreshold.Load() {
+					if err := edr.vacuum(ctx, dbHandle, tags); err == nil {
+						deletedRows = 0
+						lastVacuum = time.Now()
+					}
+				} else {
```
fracasula, line 467: Small optimization to avoid running the query for the size estimation unless really necessary.

```suggestion
				} else if time.Since(lastVacuum) >= vacuumInterval.Load() {
```

Of course you'll have to remove the same condition from the nested `if`.

enterprise/reporting/flusher/flusher.go
```diff
@@ -313,24 +329,53 @@ func (f *Flusher) send(ctx context.Context, aggReports []json.RawMessage) error
...
-	if sizeEstimate > f.vacuumThresholdBytes.Load() {
-		vacuumStart := time.Now()
-		if _, err := f.db.ExecContext(ctx, fmt.Sprintf("vacuum full analyze %s", pq.QuoteIdentifier(f.table))); err != nil {
-			return fmt.Errorf("error vacuuming table %w", err)
+	var query string
+	var full bool
+	var performVacuum bool
+	if f.deletedRows >= f.vacuumThresholdDeletedRows.Load() {
+		performVacuum = true
+	} else {
```
fracasula, line 346: Same optimization can be done here.

enterprise/reporting/reporting.go
```diff
@@ -449,27 +462,33 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
...
-					if _, err := dbHandle.ExecContext(ctx, `vacuum full analyze reports;`); err != nil {
+				// vacuum the table
+				if deletedRows >= vacuumDeletedRowsThreshold.Load() {
+					if err := r.vacuum(ctx, dbHandle, tags); err == nil {
+						deletedRows = 0
+						lastVacuum = time.Now()
+					}
+				} else {
```
fracasula, line 475: And here.

fracasula (approved):

BonapartePC (approved):

PR #5122, opened 2024-09-23
chore: set maxIdleConns for jobsdb handles and emit (*sql.DB).Stats

Sidddddarth:
# Description

1. Set `maxIdleConns` for all the jobsdb handles. 
The current default is only 2 max idle conns per pool which causes a lot of turnaround and consistent new connections. Setting it to the same as max open connections.
2. Emit periodic `(*sql.DB).Stats()`. that gauge the total closed(due to `maxIdleConns`) every 15 seconds(configurable).
3. Add tablePrefix to the jobsdb connection pool's `application_name`.

## Security


mihir20 (approved):
LGTM

PR #5131, opened 2024-09-23
fix: wrong order in unionjobsdb, unionjobsdbmetadata functions

Sidddddarth:
# Description

Fixes table ordering in the `unionjobsdb`, `unionjobsdbmetadata` pg functions.

## Security


sql/migrations/node/000008_unionjobsdb_fn.up.sql
achettyiitr, line 1: Since these are part of migrations and we are not using [RunAlways](https://github.com/rudderlabs/rudder-server/blob/336b876f2d5dc334359d99593baa2b9295769541/services/sql-migrator/migrator.go#L37) when running the migration in [here](https://github.com/rudderlabs/rudder-server/blob/master/services/validators/envValidator.go#L129-L145), shouldn't we need to create a new migrations file for both the changes?

PR #5136, opened 2024-09-25
chore: emit event_delivery_time metric at staging file level

itsmihir:
# Description

Currently, we are emitting `event_delivery_time` only based on the first staging file, with this PR we will be emitting it for all the staging files to capture a more accurate event delivery time.

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-680/capture-more-accurate-warehouse-latency

## Security


warehouse/internal/repo/staging.go
```diff
@@ -401,22 +401,41 @@ func (sf *StagingFiles) TotalEventsForUpload(ctx context.Context, upload model.U
 	return total.Int64, nil
 }
 
-func (sf *StagingFiles) FirstEventForUpload(ctx context.Context, upload model.Upload) (time.Time, error) {
-	var firstEvent sql.NullTime
-	err := sf.db.QueryRowContext(ctx, `
+func (sf *StagingFiles) GetEventTimeRangesByUploadID(ctx context.Context, uploadID int64) ([]*model.EventTimeRange, error) {
```
fracasula, line 404: Do you really need to return a slice of pointers? Wouldn't it make more sense to return a slice of values instead? It should be easier on the GC.

warehouse/router/upload.go
```diff
@@ -221,6 +222,15 @@ func (f *UploadJobFactory) NewUploadJob(ctx context.Context, dto *model.UploadJo
 		whutils.Tag{Name: "sourceCategory", Value: uj.warehouse.Source.SourceDefinition.Category},
 	)
 
+	syncFrequency := "1440"
```
fracasula, line 225: Can you add a comment here please? I'm assuming this would be 24h, correct?

warehouse/router/upload.go
```diff
@@ -123,6 +123,7 @@ type UploadJob struct {
 		numStagedEvents                    stats.Measurement
 		uploadSuccess                      stats.Measurement
 		stagingLoadFileEventsCountMismatch stats.Measurement
+		eventDeliveryTime                  stats.Measurement
```
fracasula, line 126: I know you copied from above but just using `Measurement` is wrong in this case. Can you use `stats.Timer` which makes more sense?

warehouse/router/upload_stats.go
```diff
@@ -69,6 +68,17 @@ func (job *UploadJob) generateUploadSuccessMetrics() {
 		return
 	}
 
+	eventTimeRanges, err := job.stagingFileRepo.GetEventTimeRangesByUploadID(job.ctx, job.upload.ID)
+	if err != nil {
+		job.logger.Warnw("event time ranges for upload", logfield.Error, err.Error())
```
fracasula, line 73: Can you please use the non-sugared logger for better performance?

```suggestion
		job.logger.Warnn("event time ranges for upload", obskit.Error(err))
```

Example [here](https://github.com/rudderlabs/rudder-server/blob/v1.34.1/router/batchrouter/handle.go#L421).

fracasula (changes requested):

warehouse/router/upload.go
itsmihir: I used `stats.Measurement` because [job.timerStat](https://github.com/rudderlabs/rudder-server/blob/master/warehouse/router/upload_stats.go#L35) returns `stats.Measurement` type. We can change it.

warehouse/internal/repo/staging_test.go
```diff
@@ -290,6 +290,21 @@ func TestStagingFileRepo_Many(t *testing.T) {
 			require.Nil(t, expectedSchemas)
 		})
 	})
+
+	t.Run("GetEventTimeRangesByUploadID", func(t *testing.T) {
```
fracasula, line 294: There is another scenario you're not testing. If not rows then it should return `nil, nil`.

warehouse/router/upload.go
fracasula: That's because `stats.NewTaggedStat` returns a `Measurement`. It has to because it can return anything. The object you're initializing can be more precise though so it makes sense to use the correct interface as the receiving type. For the returning type we can't do much. Sure we can change `job.timerStat` but not `stats.NewTaggedStat` so the underlying problem would be there anyway.

Thanks for making the change :+1:

fracasula (approved):
LGTM, I added another comment, should be easy to address.

Let's see if @achettyiitr is also happy with the changes.

warehouse/router/upload_stats.go
```diff
@@ -69,6 +69,17 @@ func (job *UploadJob) generateUploadSuccessMetrics() {
...
+	eventTimeRanges, err := job.stagingFileRepo.GetEventTimeRangesByUploadID(job.ctx, job.upload.ID)
+	if err != nil {
+		job.logger.Warnn("event time ranges for upload", obskit.Error(err))
+		return
+	}
+
+	for _, eventTimeRange := range eventTimeRanges {
+		job.stats.eventDeliveryTime.Since(eventTimeRange.FirstEventAt)
+		job.stats.eventDeliveryTime.Since(eventTimeRange.LastEventAt)
+	}
```
satishrudderstack, line 81: We are not capturing the events distribution in the staging files here. I think we should use `total_events` from `wh_staging_files` table here.

warehouse/router/upload_stats.go
```diff
@@ -115,24 +126,6 @@ func (job *UploadJob) recordTableLoad(tableName string, numEvents int64) {
...
-		syncFrequency := "1440"
-		if frequency := job.warehouse.GetStringDestinationConfig(job.conf, model.SyncFrequencySetting); frequency != "" {
-			syncFrequency = frequency
-		}
-		job.timerStat("event_delivery_time",
-			warehouseutils.Tag{Name: "tableName", Value: capturedTableName},
-			warehouseutils.Tag{Name: "syncFrequency", Value: syncFrequency},
-			warehouseutils.Tag{Name: "sourceCategory", Value: job.warehouse.Source.SourceDefinition.Category},
-		).Since(firstEventAt)
-	}
```
satishrudderstack, line 135: Looks like we were not capturing this metric for retried uploads and we don't have that condition now. Why was this @achettyiitr @itsmihir

achettyiitr: It was mostly because when retries it would mess up the metrics because those will be days later.

warehouse/router/upload_stats.go
itsmihir: We currently don't have a way to capture this. There is a hacky way to have a for-loop from 0 to total_events, but I would not prefer it as total_events can be huge.

warehouse/router/upload_stats.go
itsmihir: @achettyiitr can confirm, but I don't think of a reason why we can't capture the retried uploads.

achettyiitr (approved):

PR #5139, opened 2024-09-26
chore: optimise config backend revisionId calls

cisse21:
# Description

PrizePicks CFD

- Reduce the number of config backend API calls being made for a single revisionId
- `VACUUM FULL` notifier table if the size exceeds 5GB

## Linear Ticket

Fixes PIPE-1583

## Security

### DIFF
--- a/services/notifier/repo.go
+++ b/services/notifier/repo.go
@@ -6,6 +6,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/rudderlabs/rudder-go-kit/bytesize"
+	"github.com/rudderlabs/rudder-go-kit/config"
 	"time"
 
 	"github.com/lib/pq"
@@ -302,6 +304,19 @@ func (n *repo) deleteByBatchID(
 	)
 	if err != nil {
 		return fmt.Errorf("deleting by batchID: %w", err)
+	}
+
+	var sizeEstimate int64
+	if err := n.db.QueryRowContext(
+		ctx,
+		fmt.Sprintf(`SELECT pg_table_size(oid) from pg_class where relname='%s';`, notifierTableName),
+	).Scan(&sizeEstimate); err != nil {
+		return fmt.Errorf("size estimate for notifierTable failed with: %w", err)
+	}
+	if sizeEstimate > config.GetInt64("Notifier.vacuumThresholdBytes", 5*bytesize.GB) {
+		if _, err := n.db.ExecContext(ctx, fmt.Sprintf(`vacuum full analyze %q;`, notifierTableName)); err != nil {
+			return fmt.Errorf("deleting by batchID: vacuum: %w", err)
+		}
 	}
 	return nil
 }
--- a/services/notifier/repo.go
+++ b/services/notifier/repo.go
@@ -6,9 +6,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"time"
+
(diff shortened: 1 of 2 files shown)


warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -369,20 +374,20 @@ func (lf *LoadFileGenerator) createFromStaging(ctx context.Context, job *model.U
...
+	revisionIDMap[job.Warehouse.Destination.RevisionID] = job.Warehouse.Destination
+	revisionIds := lo.Uniq(
+		lo.Map(
+			lo.Filter(job.StagingFiles, func(file *model.StagingFile, _ int) bool {
+				return file.DestinationRevisionID != job.Warehouse.Destination.RevisionID
+			}), func(file *model.StagingFile, _ int) string {
+				return file.DestinationRevisionID
+			}))
+	for _, revisionId := range revisionIds {
+		destination, err := lf.ControlPlaneClient.DestinationHistory(ctx, revisionId)
```
Sidddddarth, line 386: Just adding a map check looks simpler than nested `lo.xxx`
```
func (lf *LoadFileGenerator) destinationRevisionIDMap(ctx context.Context, job *model.UploadJob) (revisionIDMap map[string]backendconfig.DestinationT, err error) {
	revisionIDMap = make(map[string]backendconfig.DestinationT)

	// TODO: ensure DestinationRevisionID is populated
	for _, file := range job.StagingFiles {
		revisionID := file.DestinationRevisionID
		// No need to make config backend api call for the current config
		if revisionID == job.Warehouse.Destination.RevisionID {
			revisionIDMap[revisionID] = job.Warehouse.Destination
			continue
		}
		// No need to make config backend api call for the same revision ID
		if _, ok := revisionIDMap[revisionID]; ok {
			continue
		}
		destination, err := lf.ControlPlaneClient.DestinationHistory(ctx, revisionID)
		if err != nil {
			return nil, err
		}
		revisionIDMap[revisionID] = destination
	}
	return
}
```

services/notifier/repo.go
```diff
@@ -303,6 +306,19 @@ func (n *repo) deleteByBatchID(
...
+
+	var sizeEstimate int64
+	if err := n.db.QueryRowContext(
+		ctx,
+		fmt.Sprintf(`SELECT pg_table_size(oid) from pg_class where relname='%s';`, notifierTableName),
+	).Scan(&sizeEstimate); err != nil {
+		return fmt.Errorf("size estimate for notifierTable failed with: %w", err)
+	}
+	if sizeEstimate > config.GetInt64("Notifier.vacuumThresholdBytes", 5*bytesize.GB) {
+		if _, err := n.db.ExecContext(ctx, fmt.Sprintf(`vacuum full analyze %q;`, notifierTableName)); err != nil {
```
lvrach, line 318: Did the full vacuuming helped here?

cisse21: Yes it did. When the size got high the DB queries got slower affecting the throughput, so added the vacuum command to automate it

Sidddddarth (approved):

services/notifier/repo.go
achettyiitr: Should we do it from a master instead of a slave once all the track batch is completed?

cisse21: Let's merge this and take that up if this causes any issues?

achettyiitr (approved):

PR #5146, opened 2024-09-27
chore: add db.sql stat collector

lvrach:
# Description

In this PR, database stats collect is added in all sql.DB instances. This will allow us to monitor open/close connection and connections waiting.

Some additional modifications where due to:
* Passing stats.Stats using DI
* Stats collector, has a name uniqueness limitation. This is to prevent from metrics being accidentally overidden leading to corrupted stats. A lot of tests had to be modified, not use stats.Default. 

## Linear Ticket

https://linear.app/rudderstack/issue/PRI-48/collect-database-sql-stats

resolves PRI-48

## Security


app/apphandlers/processorAppHandler.go
```diff
@@ -149,33 +150,38 @@ func (a *processorApp) StartRudderCore(ctx context.Context, options *app.Options
 		jobsdb.WithClearDB(options.ClearDB),
 		jobsdb.WithDSLimit(a.config.gatewayDSLimit),
 		jobsdb.WithSkipMaintenanceErr(config.GetBool("Gateway.jobsDB.skipMaintenanceError", true)),
+		jobsdb.WithStats(stats.Default),
```
achettyiitr, line 153: We can use statsFactory here since we have already defined it. Same for below occurrence as well.
```suggestion
		jobsdb.WithStats(statsFactory),
```

achettyiitr (commented):
We can also get rid of sqlutil.MonitorDatabase usage as well.

achettyiitr (approved):

BonapartePC (approved):

fracasula (approved):

PR #5170, opened 2024-10-04
chore: one (*sql.DB) pool for all jobsdb

Sidddddarth:
# Description

One connection pool for all jobsdb.
In future, use the same pool across other areas(reporting, backend-config cache, drain-config etc. with appropriate increase in pool configurations)

## Motivation

There's several problems to deal with
By default each jobsdb connection pool could only hold two idle connections, this meant [creating a lot of new connections often](https://rudderlabs.slack.com/archives/C02B0A38QQG/p1726823234425549)
It was also observed that the overall `in_use` connections at any point of time was only around 20-25(this includes reporting, config cache etc.), there was rarely a need for more connections to the database. Also observed was the fact that having too many idle connections lying around caused unnecessary memory usage on postgres pods. A better control over the idle and total connections in use would be great.

So considering these factors, it made sense to have a common connection pool and have different components borrow from it.

To allow for a surge in connection demand, the max open connections config is at a generous distance, which is configurable of course in case 40(default) isn't enough.

 ## Configuration
Using given config(all of these are hot-reloadable):
`maxOpenConnections`: 40
`maxIdleConnections`: 5
`maxIdleConnIdleTime`: 15 Minutes
`maxConnLifetime`: 0

## Linear Ticket

[Resolves PIPE-1602](https://linear.app/rudderstack/issue/PIPE-1602/efficient-db-connection-pool)

## Security


utils/misc/dbutils.go
```diff
@@ -35,6 +40,74 @@ func GetConnectionString(c *config.Config, componentName string) string {
...
+func GetDatabaseConnectionPool(
+	ctx context.Context,
+	conf *config.Config,
+	stat stats.Stats,
+	componentName string,
+) (*sql.DB, error) {
+	connStr := GetConnectionString(conf, componentName)
+	db, err := sql.Open("postgres", connStr)
+	if err != nil {
+		return nil, fmt.Errorf("Error opening connection to database: %w", err)
```
lvrach, line 52:
```suggestion
		return nil, fmt.Errorf("opening connection to database: %w", err)
```

utils/misc/dbutils.go
```diff
@@ -35,6 +40,74 @@ func GetConnectionString(c *config.Config, componentName string) string {
 	)
 }
 
+func GetDatabaseConnectionPool(
```
lvrach, line 43:
```suggestion
func NewDatabaseConnectionPool(
```

utils/misc/dbutils.go
```diff
@@ -35,6 +40,74 @@ func GetConnectionString(c *config.Config, componentName string) string {
...
+	db, err := sql.Open("postgres", connStr)
+	if err != nil {
+		return nil, fmt.Errorf("Error opening connection to database: %w", err)
+	}
+	if err := db.Ping(); err != nil {
+		return nil, fmt.Errorf("Error pinging database: %w", err)
+	}
+	if err := stat.RegisterCollector(
+		collectors.NewDatabaseSQLStats(
+			"jobsdb",
```
lvrach, line 59:
```suggestion
			componentName,
```

utils/misc/dbutils.go
```diff
@@ -35,6 +40,74 @@ func GetConnectionString(c *config.Config, componentName string) string {
...
+			db,
+		),
+	); err != nil {
+		return nil, fmt.Errorf("Error registering database stats collector: %w", err)
+	}
+	/*
+		TODO: find out a reasonably good value for below pool configurations
+	*/
```
lvrach, line 67:
```suggestion
```

utils/misc/dbutils.go
```diff
@@ -35,6 +40,74 @@ func GetConnectionString(c *config.Config, componentName string) string {
...
+
+	maxConnsVar := conf.GetReloadableIntVar(40, 1, "JobsDB.maxOpenConnections")
```
lvrach, line 69: I would avoid using JobsDB, as it is an overloaded term.
```suggestion
	maxConnsVar := conf.GetReloadableIntVar(40, 1, "db.pool.maxOpenConnections")
```

Or even 

```suggestion
	maxConnsVar := conf.GetReloadableIntVar(40, 1, "db.pool.maxOpenConnections", "db."+componentName+".pool.maxOpenConnections")
```

app/apphandlers/embeddedAppHandler.go
```diff
@@ -139,13 +139,21 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 		FeaturesRetryMaxAttempts: 10,
 	})
 
+	dbPool, err := misc.GetDatabaseConnectionPool(ctx, config, statsFactory, "jobsdb")
+	if err != nil {
+		return err
+	}
+	//	Not deferring close here
+	// defer dbPool.Close()
```
lvrach, line 147: What the reason for not closing here? Shall we inject dbPool to embeddedAppHandler. So we always use on connection pool

app/apphandlers/embeddedAppHandler.go
```diff
@@ -139,13 +139,21 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 		FeaturesRetryMaxAttempts: 10,
 	})
 
+	dbPool, err := misc.GetDatabaseConnectionPool(ctx, config, statsFactory, "jobsdb")
```
lvrach, line 142:
```suggestion
	dbPool, err := misc.GetDatabaseConnectionPool(ctx, config, statsFactory, "embedded-app")
```

I would avoid using `jobsdb` for anything else other than jobsdb component

app/apphandlers/embeddedAppHandler.go
Sidddddarth: just to keep the existing behaviour. We don't close gateway db pool, but the others are closed right away. Since we're sharing the pool now, didn't want to close it.
Just left this comment to find out why the gateway pool wasn't being closed earlier because it's closed in gateway app handler anyway.

app/apphandlers/embeddedAppHandler.go
Sidddddarth: makes sense.
Should we suffix with `app`?
`embedded`, `processor`, `gateway` look good enough no?

app/apphandlers/embeddedAppHandler.go
Sidddddarth: With the way things are, injecting a pool into `app`s would be a good idea. This would effectively enforce the same database for all components. But I think it'd be nice to have the option of passing different connections pools to different components.

mihir20 (commented):
@Sidddddarth can you add more details in description on why we are going with single db connection pool?

app/apphandlers/embeddedAppHandler.go
lvrach: processor or gateway can be a bit ambiguous by themselves.

app/apphandlers/embeddedAppHandler.go
lvrach: what is the existing behaviour you are referring to? I see that every component was closing itself, why have we removed that?

Sidddddarth: we close all `*sql.DB` except gateway db. gateway jobsdb is only `Stop`ped.

Sidddddarth: in any case, closing it everywhere
looks like there was no reason after all

app/apphandlers/embeddedAppHandler.go
```diff
@@ -168,14 +181,15 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 		jobsdb.WithDSLimit(a.config.routerDSLimit),
 		jobsdb.WithSkipMaintenanceErr(config.GetBool("Router.jobsDB.skipMaintenanceError", false)),
 		jobsdb.WithStats(statsFactory),
+		jobsdb.WithDBHandle(dbPool),
 	)
-	defer routerDB.Close()
```
BonapartePC, line 172: Any reason we removed this?

Sidddddarth: omitted by mistake.

utils/misc/dbutils.go
```diff
@@ -35,6 +40,71 @@ func GetConnectionString(c *config.Config, componentName string) string {
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+			}
+			updatePoolConfig(db.SetMaxOpenConns, &maxConns, maxConnsVar)
+			updatePoolConfig(db.SetConnMaxIdleTime, &maxIdleTime, maxIdleTimeVar)
+			updatePoolConfig(db.SetMaxIdleConns, &maxIdleConns, maxIdleConnsVar)
+			updatePoolConfig(db.SetConnMaxLifetime, &maxConnLifetime, maxConnLifetimeVar)
```
BonapartePC, line 94: Why not move this below `<-ticker.C`?

utils/misc/dbutils.go
```diff
@@ -35,6 +40,71 @@ func GetConnectionString(c *config.Config, componentName string) string {
...
+		}
+	})
+	return db, nil
+}
+
+func updatePoolConfig[T comparable](f func(T), current *T, conf config.ValueLoader[T]) {
```
BonapartePC, line 100:
```suggestion
func updatePoolConfig[T comparable](setter func(T), current *T, conf config.ValueLoader[T]) {
```
[Optional]

utils/misc/dbutils.go
Sidddddarth: no reason
just used to writing it like this

BonapartePC (approved):

ktgowtham (approved):
LGTM 👍

lvrach (commented):
We need to modify https://github.com/rudderlabs/rudder-server/blob/209a9c1660bea163acd46ce7715a886907c58068/jobsdb/jobsdb.go#L1094

So that db closes the connection only if the connection was initialised from jobsdb.

jobsdb/jobsdb.go
```diff
@@ -769,7 +770,9 @@ func (jd *Handle) init() {
 	jd.loadConfig()
 
 	// Initialize dbHandle if not already set
-	if jd.dbHandle == nil {
+	if jd.dbHandle != nil {
+		jd.sharedConnectionPool = true
```
lvrach, line 774: We can also do:

```suggestion
		jd.onCloseFn = func () {
			 if err := jd.dbHandle.Close(); err != nil {
			     jd.logger.Errorw("error closing db connection", "error", err) 
		          }
		}
```

PR #5181, opened 2024-10-08
fix: set hosted secret as auth instead of workspace token

BonapartePC:
# Description

Set hosted secret as auth instead of workspace token. We can also validate with hosted secret. More info [here](https://github.com/rudderlabs/rudder-config-backend/blob/f83d9f7e7651ee536d7117f373fa6f8b3b956aa5/src/controllers/dataPlane.controller.ts#L40-L53)

## Linear Ticket

Fixes PIPE-1571

## Security

### DIFF
--- a/services/debugger/uploader.go
+++ b/services/debugger/uploader.go
@@ -48,6 +48,7 @@ type uploaderImpl[E any] struct {
 	maxBatchSize, maxRetry, maxESQueueSize config.ValueLoader[int]
 	batchTimeout, retrySleep               config.ValueLoader[time.Duration]
 	region                                 string
+	hostedServiceSecret                    string
 
 	bgWaitGroup sync.WaitGroup
 }
@@ -64,6 +65,7 @@ func (uploader *uploaderImpl[E]) Setup() {
 	uploader.batchTimeout = config.GetReloadableDurationVar(2, time.Second, "Debugger.batchTimeoutInS")
 	uploader.retrySleep = config.GetReloadableDurationVar(100, time.Millisecond, "Debugger.retrySleepInMS")
 	uploader.region = config.GetString("region", "")
+	uploader.hostedServiceSecret = config.GetString("HOSTED_SERVICE_SECRET", "")
 }
 
 func New[E any](url string, transformer Transformer[E]) Uploader[E] {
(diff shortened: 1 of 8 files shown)


albert849 (approved):

albert849 (changes requested):
Improve it

albert849 (approved):

services/debugger/uploader.go
```diff
@@ -128,7 +134,7 @@ func (uploader *uploaderImpl[E]) uploadEvents(eventBuffer []E) {
 			req.URL.RawQuery = q.Encode()
 		}
 		req.Header.Set("Content-Type", "application/json;charset=UTF-8")
-		req.SetBasicAuth(config.GetWorkspaceToken(), "")
+		req.SetBasicAuth(uploader.hostedServiceSecret, "")
```
ktgowtham, line 137: This will break Open source or deployments using workspace token. We need to check the deployment type and then choose the right token.

BonapartePC: Fixed

services/debugger/uploader.go
```diff
@@ -210,3 +221,20 @@ func (uploader *uploaderImpl[E]) flushEvents(ctx context.Context) {
 		}
 	}
 }
+
+func getAuthUsername() (string, error) {
```
lvrach, line 225: Consider moving this auth logic to a different component. deploymentType logic should remain contained in a few components. Can we use Identify method from backend config ? https://github.com/rudderlabs/rudder-server/blob/master/backend-config/namespace_config.go#L228C27-L228C38

Sidddddarth (approved):

services/debugger/uploader.go
BonapartePC: Fixed

ktgowtham (approved):

services/debugger/uploader_test.go
```diff
@@ -298,3 +293,12 @@ var _ = Describe("Uploader", func() {
 		})
 	})
 })
+
+func assertRequest(req *http.Request) {
+	username, password, ok := req.BasicAuth()
+	Expect(ok).To(BeTrue())
+	Expect(username).To(Equal("test"))
+	Expect(password).To(Equal("test"))
```
lvrach, line 301: It's not a good practice to pass the same value to multiple fields.

If the username/password got inverted you won't be able to detect this

lvrach (approved):
Approve, with minor comment

PR #5187, opened 2024-10-09
chore: webhook v0 deprecation while defaulting to v1

vinayteki95:
# Description

Deprecation of webhook transformer service v0 adapter as a preliminary step to move towards v2 spec.
This is stage 0 of 3 for v2 spec upgrade.
Details regarding this upgrade can be found in this [notion doc](https://www.notion.so/rudderstacks/Webhook-Sources-Transformer-Payload-V2-Spec-Migration-109f2b415dd0805db37fc8e76556103f?pvs=4)

## Linear Ticket
Corresponding linear ticket can be found [here.](https://linear.app/rudderstack/issue/INT-2754/stage-0-or-v0-spec-deprecation-in-rudder-server)
Resolves INT-2754
## Note:
This will be a breaking change for older rudder-transformer versions. (older than 1.50.0 rudder-transformer)


gateway/webhook/webhookTransformer_test.go
```diff
@@ -13,24 +13,26 @@ import (
...
+// 	v0Adapter := newSourceTransformAdapter(transformer.V0)
 
-	t.Run("should return the body as is", func(t *testing.T) {
-		testBody := []byte("testBody")
-		retBody, err := v0Adapter.getTransformerEvent(nil, testBody)
-		require.Equal(t, testBody, retBody)
-		require.Nil(t, err)
-	})
-}
+// 	t.Run("should return the right url", func(t *testing.T) {
```
koladilip, line 21: don't keep commented code, we can always get back from git history

koladilip (approved):

ktgowtham (approved):

PR #5188, opened 2024-10-09
chore: cleanup old jobs(beyond maxAge) at startup

Sidddddarth:
# Description

Remove the cleanup loop from jobsdb, because it's not "air-tight"(Doesn't enforce a proper abort for a job).
Also it isn't triggered in many production instances because it triggers once every 24 hours
and more often than not - servers restart within that duration anyway.

So considering these two facts, it's prudent to perform cleanup once at startup and not worry about it later on.
We should also take a look at consumers and have them perform some operation in the off chance that a server were to live for longer durations(days) and jobs are no longer of interest to them.

## Linear Ticket

[Resolves PIPE-1619](https://linear.app/rudderstack/issue/PIPE-1619/clean-up-invalid-jobs)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -863,6 +863,12 @@ func (jd *Handle) init() {
 	if err != nil {
 		panic(fmt.Errorf("failed to run schema migration for %s: %w", jd.tablePrefix, err))
 	}
+	// cleanup old jobs if read/readWrite jobsdb, writer doesn't care about old jobs
+	if jd.ownerType != Write {
+		if err := jd.doCleanup(context.Background(), jd.config.GetInt("jobsdb.cleanupBatchSize", 100)); err != nil {
+			panic(fmt.Errorf("cleaning up old jobs: %w", err))
+		}
+	}
 }
 
 func (jd *Handle) workersAndAuxSetup() {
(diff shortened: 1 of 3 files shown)


mihir20 (approved):
LGTM

jobsdb/jobsdb.go
```diff
@@ -1032,6 +1022,13 @@ func (jd *Handle) readerSetup(ctx context.Context, l lock.LockToken) {
 	// Even if two different services (gateway and processor) perform this operation, there should not be any problem.
 	jd.recoverFromJournal(ReadWrite)
 	jd.assertError(jd.doRefreshDSRangeList(l))
+	jd.assertError(func() error {
+		err := jd.doCleanup(ctx)
```
ktgowtham, line 1026: I would prefer this step to be performed in it's own go routine and not block the startup process for long time.  We did see startup delays if there are lot of pending tables.

Sidddddarth: The way it's written now, it needs a dsListLock(the token passed), so running it async would lead to synchronisation issues.
And this is something that doesn't happen often, and stops when it encounters a DS without any old jobs - so it felt safe to run at startup.

ktgowtham: > And this is something that doesn't happen often, and stops when it encounters a DS without any old jobs
Alright! cool

ktgowtham (approved):

PR #5204, opened 2024-10-16
fix: discard badger transactions

Sidddddarth:
# Description

According to badger documentation:

`func (db *badger.DB) NewTransaction(update bool) *badger.Txn`
```
NewTransaction creates a new transaction.
Badger supports concurrent execution of transactions,
providing serializable snapshot isolation, avoiding write skews.
Badger achieves this by tracking the keys read and at Commit time,
ensuring that these read keys weren't concurrently modified by another transaction.

For read-only transactions, set update to false.
In this mode, we don't track the rows read for any changes.
Thus, any long running iterations done in this mode wouldn't pay this overhead.

Running transactions concurrently is OK.
However, a transaction itself isn't thread safe, and should only be run serially.
It doesn't matter if a transaction is created by one goroutine and passed down to other,
as long as the Txn APIs are called serially.

When you create a new transaction, it is absolutely essential to call Discard().
This should be done irrespective of what the update param is set to.
Commit API internally runs Discard, but running it twice wouldn't cause any issues.

txn := db.NewTransaction(false)
defer txn.Discard()
// Call various APIs.
```

and

`func (txn *badger.Txn) Discard()`
```
Discard discards a created transaction.
This method is very important and must be called.
Commit method calls this internally, however, calling this multiple times doesn't cause any issues.
So, this can safely be called via a defer right when transaction is created.

NOTE: If any operations are run on a discarded transaction, ErrDiscardedTxn is returned.
```

## Linear Ticket

## Security

### DIFF
--- a/services/dedup/badger/badger.go
+++ b/services/dedup/badger/badger.go
@@ -101,28 +101,17 @@ func (d *BadgerDB) Get(key string) (int64, bool, error) {
 
 func (d *BadgerDB) Set(kvs []types.KeyValue) error {
 	defer d.stats.NewTaggedStat("dedup_commit_duration_seconds", stats.TimerType, stats.Tags{"mode": "badger"}).RecordDuration()()
-	return d.set(kvs)
-}
-
-func (d *BadgerDB) set(kvs []types.KeyValue) error {
-	txn := d.badgerDB.NewTransaction(true)
-	defer txn.Discard()
+	wb := d.badgerDB.NewWriteBatch()
+	defer wb.Cancel()
 	for i := range kvs {
 		message := kvs[i]
 		value := strconv.FormatInt(message.Value, 10)
 		e := badger.NewEntry([]byte(message.Key), []byte(value)).WithTTL(d.window.Load())
-		err := txn.SetEntry(e)
-		if errors.Is(err, badger.ErrTxnTooBig) {
-			if err = txn.Commit(); err != nil {
-				return err
-			}
-			txn.Discard()
-			return d.set(kvs[i:])
-		} else if err != nil {
+		if err := wb.SetEntry(e); err != nil {
 			return err
 		}
 	}
-	return txn.Commit()
+	return wb.Flush()
 }
 
 func (d *BadgerDB) Close() {
--- a/services/dedup/badger/badger.go
+++ b/services/dedup/badger/badger.go
@@ -103,18 +103,18 @@ func (d *BadgerDB) Set(kvs []types.KeyValue) error {
 	defer d.stats.NewTaggedStat("dedup_commit_duration_seconds", stats.TimerType, stats.Tags{"mode": "badger"}).RecordDuration()()
 
 	txn := d.badgerDB.NewTransaction(true)
-	for _, message := range kvs {
+	defer txn.Discard()
+	for i := range kvs {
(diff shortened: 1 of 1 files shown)


services/dedup/badger/badger.go
```diff
@@ -101,20 +101,23 @@ func (d *BadgerDB) Get(key string) (int64, bool, error) {
...
+		if errors.Is(err, badger.ErrTxnTooBig) {
 			if err = txn.Commit(); err != nil {
 				return err
 			}
-			txn = d.badgerDB.NewTransaction(true)
-			if err = txn.SetEntry(e); err != nil {
-				return err
-			}
+			txn.Discard()
+			return d.set(kvs[i:])
```
achettyiitr, line 120: We might end up creating a big recursive stack if everything ends up with `badger.ErrTxnTooBig` the following error. We can still re-use the existing logic while making sure that Discard is being called for the transaction. wdyt?

```GO
func (d *BadgerDB) set(kvs []types.KeyValue) error {
	txn := d.badgerDB.NewTransaction(true)
	defer txn.Discard() // Discard will clean up the last transaction, even if there's an early return.

	for i := 0; i < len(kvs); i++ {
		message := kvs[i]
		value := strconv.FormatInt(message.Value, 10)
		e := badger.NewEntry([]byte(message.Key), []byte(value)).WithTTL(d.window.Load())
		err := txn.SetEntry(e)

		if errors.Is(err, badger.ErrTxnTooBig) {
			if err := txn.Commit(); err != nil {
				return err
			}
			
			// Discard the old transaction and create a new one
			txn.Discard()
			txn = d.badgerDB.NewTransaction(true)
			if err = txn.SetEntry(e); err != nil {
				return err
			}
		} else if err != nil {
			return err
		}
	}
	return txn.Commit()
}
```

Sidddddarth: should we go with a `writeBatch` instead?(refer `enterprise/suppress-user/internal/badgerdb/badgerdb.go`)

cisse21 (approved):

services/dedup/badger/badger.go
achettyiitr: How are we going to handle `badger.ErrTxnTooBig `in that case, since writeBatch would do things in a single transaction? Currently, we are creating a new transaction when getting `badger.ErrTxnTooBig`?

Sidddddarth: writeBatch handles it internally
if it gets too big, it commits and starts a new transaction.
We can avoid all this handling on our side

achettyiitr: TIL. Didn't find anything on the API but this function provides the context.

```GO
func (wb *WriteBatch) handleEntry(e *Entry) error {
	if err := wb.txn.SetEntry(e); err != ErrTxnTooBig {
		return err
	}
	// Txn has reached it's zenith. Commit now.
	if cerr := wb.commit(); cerr != nil {
		return cerr
	}
	// This time the error must not be ErrTxnTooBig, otherwise, we make the
	// error permanent.
	if err := wb.txn.SetEntry(e); err != nil {
		wb.err.Store(err)
		return err
	}
	return nil
}
```

achettyiitr (approved):

PR #5205, opened 2024-10-17
feat: warehouse transformer

achettyiitr:
# Description

- Warehouse transformation as a module. 
- It's basically ported from the [warehouse module](https://github.com/rudderlabs/rudder-transformer/tree/develop/src/warehouse) of the [rudder-transformer](https://github.com/rudderlabs/rudder-transformer/tree/develop) from Javascript to GO.
- More about it can be found in [this](https://www.notion.so/rudderstacks/Destination-Transformations-c2ca4536bc4444e9bd3fa7bf64c85f8b?pvs=4) doc.
- Currently, there is a flag to enable warehouse transformations along with Processor using the flag `Processor.enableWarehouseTransformations`.
  - Once enabled, we call the warehouse transformer package API.
  - Compare the response, in case we see any differences, we log it for debugging purposes.

## Linear Ticket

- Resolves WAR-187

## Security


warehouse/transformer/internal/utils/reservedkeywords.go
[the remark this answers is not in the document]
achettyiitr: Yeah, Maintainability wise JSON file is better as compared to keeping a hardcoded lists in the source code.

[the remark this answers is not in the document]
achettyiitr: Will make the changes.

warehouse/transformer/transformer.go
```diff
@@ -0,0 +1,205 @@
...
+func (t *transformer) Transform(_ context.Context, clientEvents []ptrans.TransformerEvent, _ int) (res ptrans.Response) {
+	if len(clientEvents) == 0 {
+		return
+	}
+
+	startTime := t.now()
+	metadata := clientEvents[0].Metadata
+
+	defer func() {
+		tags := stats.Tags{
```
lvrach, line 70: should we add a tag `implementation` to distinguish between `go` or `js-client`

warehouse/transformer/transformer_test.go
```diff
@@ -0,0 +1,753 @@
...
+
+	for _, tc := range testsCases {
+		t.Run(tc.name, func(t *testing.T) {
+			pool, err := dockertest.NewPool("")
+			require.NoError(t, err)
+
+			opts := lo.Map(tc.envOverride, func(item string, index int) transformertest.Option {
+				return transformertest.WithEnv(item)
+			})
+			opts = append(opts, transformertest.WithRepository("rudderstack/develop-rudder-transformer"))
```
lvrach, line 728: how do you access this image? 
```
            	            	failed to pull image: API error (404): pull access denied for rudderstack/develop-rudder-transformer, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

warehouse/transformer/transformer.go
achettyiitr: We are already prefixing `warehouse_` as part of the stat name.

warehouse/transformer/transformer_test.go
achettyiitr: `develop-rudder-transformer` is in rudderlabs and I was testing a [test case](https://github.com/rudderlabs/rudder-transformer/pull/3806) locally. I have commented the test case until the latest transformer image has the changes.

processor/processor.go
```diff
@@ -2765,6 +2776,7 @@ func (proc *Handle) transformSrcDest(
 			proc.logger.Debug("Dest Transform input size", len(eventsToTransform))
 			s := time.Now()
 			response = proc.transformer.Transform(ctx, eventsToTransform, proc.config.transformBatchSize.Load())
+			proc.handleResponseForWarehouseTransformation(ctx, eventsToTransform, response, commonMetaData, eventsByMessageID)
```
ktgowtham, line 2779: won't this run for all destinations? We should run this for only warehouse destinations rt

achettyiitr: Oops. Missed the warehouse destination check. Adding now.

processor/processor.go
```diff
@@ -84,12 +90,18 @@ type trackedUsersReporter interface {
 	GenerateReportsFromJobs(jobs []*jobsdb.JobT, sourceIdFilter map[string]bool) []*trackedusers.UsersReport
 }
 
+type warehouseTransformation interface {
```
shekhar-rudder, line 93: `warehouseTransformation` -> `warehouseTransformer` ?

processor/processor.go
```diff
@@ -159,6 +171,13 @@ type Handle struct {
 		eventAuditEnabled               map[string]bool
 		credentialsMap                  map[string][]transformer.Credential
 		nonEventStreamSources           map[string]bool
+		enableWarehouseTransformations  config.ValueLoader[bool]
+	}
+
+	warehouseTransformerStats struct {
```
shekhar-rudder, line 177: Should `stats` be part of `warehouseTransformation` struct instead of `Handle` struct ?

processor/processor.go
```diff
@@ -3373,6 +3400,68 @@ func (proc *Handle) transformSrcDest(
...
+	differingEvents := proc.warehouseTransDifferEvents(eventsToTransform, pResponse, wResponse, eventsByMessageID)
+	if err := proc.warehouseTransformer.Log(differingEvents, commonMetaData); err != nil {
+		proc.logger.Warnn("Failed to log events for warehouse transformation debugging", obskit.Error(err))
+	}
+	proc.warehouseTransformerStats.logTime.Since(logStartAt)
+}
+
+func (proc *Handle) warehouseTransDifferEvents(
+	eventsToTransform []transformer.TransformerEvent,
+	pResponse, wResponse transformer.Response,
```
shekhar-rudder, line 3431: What does "p" and "w" mean ?

processor/processor.go
```diff
@@ -3373,6 +3400,68 @@ func (proc *Handle) transformSrcDest(
...
+		differedEventsCount  int
+	)
+
+	for i := range pResponse.Events {
+		if reflect.DeepEqual(pResponse.Events[i], wResponse.Events[i]) {
+			continue
+		}
+
+		differedEventsCount++
+		if len(differedSampleEvents) != 0 {
```
shekhar-rudder, line 3454: Won't this check be always false ? `differedSampleEvents` is being initialized with length 0 and the only place its length is getting updated is inside the if block.

processor/processor.go
```diff
@@ -3373,6 +3400,68 @@ func (proc *Handle) transformSrcDest(
...
+		events := lo.Map(eventsToTransform, func(e transformer.TransformerEvent, _ int) types.SingularEventT {
+			return eventsByMessageID[e.Metadata.MessageID].SingularEvent
+		})
+		proc.warehouseTransformerStats.mismatches.Count(len(events))
+		return events
+	}
+
+	var (
+		differedSampleEvents []types.SingularEventT
+		differedEventsCount  int
```
shekhar-rudder, line 3445: Do we need a separate variable for events count ? It can be computed using `len(differedSampleEvents)`

processor/transformer/transformer.go
```diff
@@ -591,7 +603,7 @@ func (trans *handle) destTransformURL(destType string) string {
 	destinationEndPoint := fmt.Sprintf("%s/v0/destinations/%s", trans.config.destTransformationURL, strings.ToLower(destType))
 
 	if _, ok := warehouseutils.WarehouseDestinationMap[destType]; ok {
-		whSchemaVersionQueryParam := fmt.Sprintf("whSchemaVersion=%s&whIDResolve=%v", trans.conf.GetString("Warehouse.schemaVersion", "v1"), warehouseutils.IDResolutionEnabled())
+		whSchemaVersionQueryParam := fmt.Sprintf("whIDResolve=%t", trans.conf.GetBool("Warehouse.enableIDResolution", false))
```
shekhar-rudder, line 606: How does `whSchemaVersion` relate to transformation ?

achettyiitr: Based on this, we used to decide whether we want to have v0/util or v1/util. Currently, we are only supporting v1 now. 
https://github.com/rudderlabs/rudder-transformer/pull/3752/files#diff-bacfe9b0c30628b991e1ec726e1484fa5380fdedab72a9fdda99799906e8de60

processor/processor.go
achettyiitr: No, we are just taking one of the sample events in `differedSampleEvents`. `differedEventsCount` will contain all number of different events.

processor/processor.go
achettyiitr: Yes. Corrected it now.

processor/processor.go
achettyiitr: processor transfromer response and warehouse transformer response.

processor/processor.go
achettyiitr: Let me move this to the warehouse->transformer package.

warehouse/transformer/safe.go
```diff
@@ -0,0 +1,273 @@
...
+		tableName:                    key,
+		skipReservedKeywordsEscaping: tec.intrOpts.skipReservedKeywordsEscaping,
+	}
+	columnNameCached, ok := tec.cache.safeTableNameCache.Load(cacheKey)
+	if !ok {
+		tableName, err = safeTableName(tec.event.Metadata.DestinationType, tec.intrOpts, key)
+		if err != nil {
+			return "", err
+		}
+		tec.cache.safeTableNameCache.Store(cacheKey, tableName)
```
shekhar-rudder, line 91: What is the benefit of creating a cache ? Does the "transform" function involve heavy computation?

warehouse/transformer/safe.go
```diff
@@ -0,0 +1,273 @@
...
+	} else {
+		columnName = columnNameCached.(string)
+	}
+	return columnName
+}
+
+// transformColumnName applies transformation to the input column name based on the destination type and configuration options.
+// If `useBlendoCasing` is enabled, it transforms the column name into Blendo casing.
+// Otherwise, it applies a more general transformation using the `transformName` function.
+func transformColumnName(destType string, intrOpts *intrOptions, destOpts *destOptions, columnName string) string {
```
shekhar-rudder, line 226: `transformColumnName` shares almost all the the code with the `transformTableName` function. Should we extract the common code ?

warehouse/transformer/safe.go
```diff
@@ -0,0 +1,273 @@
...
+// If the name does not start with a letter or underscore, it adds a leading underscore.
+// The name is truncated to 63 characters for Postgres, and the result is converted to lowercase.
+func transformNameToBlendoCase(destType, name string) string {
+	key := reNonAlphanumericOrDollar.ReplaceAllString(name, "_")
+
+	if !reStartsWithLetterOrUnderscore.MatchString(key) {
+		key = "_" + key
+	}
+	if destType == whutils.POSTGRES {
+		key = misc.TruncateStr(name, 63)
```
shekhar-rudder, line 270: 63 is getting repeated at multiple places

warehouse/transformer/transformer_fuzz_test.go
```diff
@@ -0,0 +1,668 @@
...
+	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
+	ptrans "github.com/rudderlabs/rudder-server/processor/transformer"
+	"github.com/rudderlabs/rudder-server/utils/misc"
+	"github.com/rudderlabs/rudder-server/utils/types"
+	"github.com/rudderlabs/rudder-server/warehouse/transformer/testhelper"
+	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+)
+
+var (
+	reDateTime = regexp.MustCompile(
```
shekhar-rudder, line 30: There is already a `reDateTime` defined in utils package. Can that be used instead ?

warehouse/transformer/transformer_fuzz_test.go
```diff
@@ -0,0 +1,668 @@
...
+						Name: destType,
+					},
+				},
+			},
+		}
+		cmpEvents(t, eventContexts, processorTransformer, warehouseTransformer)
+	})
+}
+
+func cmpEvents(t *testing.T, eventContexts []testhelper.EventContext, pTransformer, dTransformer ptrans.DestinationTransformer) {
```
shekhar-rudder, line 602: Can we use `ValidateEvents` method from the `testhelper` package ?

achettyiitr: No, we can't use ValidateEvents. It's basically for comparing the processor and destination transformer response with the expected response.

warehouse/transformer/transformer.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		t.statsFactory.NewTaggedStat("warehouse_dest_transform_request_latency", stats.TimerType, tags).Since(startTime)
+		t.statsFactory.NewTaggedStat("warehouse_dest_transform_requests", stats.CountType, tags).Increment()
+		t.statsFactory.NewTaggedStat("warehouse_dest_transform_input_events", stats.HistogramType, tags).Observe(float64(len(clientEvents)))
+		t.statsFactory.NewTaggedStat("warehouse_dest_transform_output_events", stats.HistogramType, tags).Observe(float64(len(res.Events)))
+		t.statsFactory.NewTaggedStat("warehouse_dest_transform_output_failed_events", stats.HistogramType, tags).Observe(float64(len(res.FailedEvents)))
+	}()
+
+	for i := range clientEvents {
```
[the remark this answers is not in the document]
achettyiitr: There is no need to add `concurrency' here. The processor already makes Transform calls concurrently. Also, these operations will be fast enough since they do not require any API calls or Disk IO.

warehouse/transformer/transformer.go
```diff
@@ -0,0 +1,204 @@
...
+	destType string,
+	data map[string]any, metadata map[string]string,
+) (map[string]any, error) {
+	columns := make(map[string]any, len(data))
+
+	// uuid_ts and loaded_at datatypes are passed from here to create appropriate columns.
+	// Corresponding values are inserted when loading into the warehouse
+	uuidTS := "uuid_ts"
+	if destType == whutils.SNOWFLAKE || destType == whutils.SnowpipeStreaming {
+		uuidTS = "UUID_TS"
```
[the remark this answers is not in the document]
achettyiitr: Using whutils.ToProviderCase.

warehouse/transformer/testhelper/validate.go
```diff
@@ -0,0 +1,118 @@
...
+		require.True(t, ok, "wResponse data must be a map")
+		wRudderEvent, ok := wEventData["rudder_event"].(string)
+		require.True(t, ok, "wResponse rudder_event must be a string")
+		require.JSONEq(t, rudderEvent, wRudderEvent)
+
+		require.JSONEq(t, pRudderEvent, wRudderEvent)
+
+		delete(pEventData, "rudder_event")
+		delete(wEventData, "rudder_event")
+		delete(data, "rudder_event")
```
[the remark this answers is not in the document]
achettyiitr: You can't do a comparison because it's a JSON. Ordering matters.

warehouse/transformer/testhelper/validate.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func validateFailedEventEquality(t *testing.T, pResponse, wResponse ptrans.Response) {
+	t.Helper()
+
+	for i := range pResponse.FailedEvents {
+		require.NotEmpty(t, pResponse.FailedEvents[i].Error)
+		require.NotEmpty(t, wResponse.FailedEvents[i].Error)
+
+		require.NotZero(t, pResponse.FailedEvents[i].StatusCode)
```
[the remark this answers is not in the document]
achettyiitr: No, the error string might be different. We can't actually compare the error string, Javascript handling of error is different from GO handling. So we are just comparing if we received failed events or not.

warehouse/transformer/set.go
```diff
@@ -0,0 +1,199 @@
...
+) error {
+	valJSON, err := json.Marshal(val)
+	if err != nil {
+		return fmt.Errorf("marshalling value: %w", err)
+	}
+	return addDataAndMetadata(tec, pi.prefix+key, string(valJSON), true, data, metadata)
+}
+
+func shouldProcessNestedObject(tec *transformEventContext, val any, pi *prefixInfo) bool {
+	return utils.IsObject(val) && (tec.event.Metadata.SourceCategory != "cloud" || pi.level < 3)
```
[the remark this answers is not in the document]
achettyiitr: This is the same check we have in the [rudder-transformer](https://github.com/rudderlabs/rudder-transformer/blob/95831f1ba4e26e1d2fef59754a18f2b14f4e7be9/src/warehouse/index.js#L258) as well.

warehouse/transformer/set.go
```diff
@@ -0,0 +1,199 @@
...
+	dataType := dataTypeFor(tec.event.Metadata.DestinationType, key, val, isJSONKey)
+
+	data[safeKey] = convertValIfDateTime(val, dataType)
+	metadata[safeKey] = dataType
+	return nil
+}
+
+func isValidJSONPath(tec *transformEventContext, key string, pi *prefixInfo) bool {
+	validLegacyJSONPath := isValidLegacyJSONPathKey(tec.event.Metadata.EventType, pi.prefix+key, pi.level, tec.jsonPathsInfo.legacyKeysMap)
+	validJSONPath := isValidJSONPathKey(pi.completePrefix+key, pi.completeLevel, tec.jsonPathsInfo.keysMap)
```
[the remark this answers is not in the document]
achettyiitr: I think it should be fine, since it's just checking for a key in map.

warehouse/transformer/safe.go
achettyiitr: Added a constant.

warehouse/transformer/datatype.go
```diff
@@ -0,0 +1,89 @@
+package transformer
+
+import (
+	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
+	"github.com/rudderlabs/rudder-server/warehouse/transformer/internal/utils"
+	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+)
+
+func dataTypeFor(destType, key string, val any, isJSONKey bool) string {
```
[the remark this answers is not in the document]
achettyiitr: handleValidJSONPath.

warehouse/transformer/testhelper/validate.go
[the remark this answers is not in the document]
shekhar-rudder: Ok. What about the status code ?

processor/processor.go
achettyiitr: Added test for it as well.

warehouse/transformer/events.go
```diff
@@ -0,0 +1,617 @@
...
+	}
+
+	output := map[string]any{
+		"data": data,
+		"metadata": map[string]any{
+			"table":      table,
+			"columns":    columns,
+			"receivedAt": tec.event.Metadata.ReceivedAt,
+		},
+		"userId": "",
```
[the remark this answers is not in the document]
achettyiitr: We can create a struct with the following information data, metadata and userID. But then later again it will be used to populate the map which is basically `TransformerResponse.Output`. I don't see much value in here with defining the struct.

warehouse/transformer/events.go
```diff
@@ -0,0 +1,617 @@
...
+
+	"github.com/rudderlabs/rudder-server/utils/misc"
+	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
+	"github.com/rudderlabs/rudder-server/warehouse/transformer/internal/response"
+	"github.com/rudderlabs/rudder-server/warehouse/transformer/internal/rules"
+	"github.com/rudderlabs/rudder-server/warehouse/transformer/internal/utils"
+	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+)
+
+func (t *Transformer) trackEvents(tec *transformEventContext) ([]map[string]any, error) {
```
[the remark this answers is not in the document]
achettyiitr: - `transformEventContext` is a data model. I don't think we should have these methods set on a data model. 
- Also, if we do that, then we would need to pass all other information like config and logger as well if we want to get some configuration or log things.

warehouse/transformer/idresolution.go
```diff
@@ -0,0 +1,171 @@
...
+
+	mergeOutput := map[string]any{
+		"data":     data,
+		"metadata": metadata,
+		"userId":   "",
+	}
+	return []map[string]any{mergeOutput}, nil
+}
+
+func mergeProps(message types.SingularEventT, metadata ptrans.Metadata) (*mergeRule, *mergeRule, error) {
```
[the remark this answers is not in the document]
achettyiitr: Yes, sometimes only 1 rule is being set.

warehouse/transformer/internal/rules/rules.go
```diff
@@ -0,0 +1,169 @@
...
+	TrackTableRules = map[string]Rules{
+		"record_id": func(event *ptrans.TransformerEvent) (any, error) {
+			eventType := event.Metadata.EventType
+			canUseRecordID := utils.CanUseRecordID(event.Metadata.SourceCategory)
+			if eventType == "track" && canUseRecordID {
+				cr, err := extractCloudRecordID(event.Message, &event.Metadata, nil)
+				if err != nil {
+					return nil, fmt.Errorf("extracting cloud record id: %w", err)
+				}
+				return utils.ToString(cr), nil
```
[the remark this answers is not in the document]
achettyiitr: No idea, it's what we were doing in [rudder-transformer](https://github.com/rudderlabs/rudder-transformer/blob/6d977feac48905605d5047934d6c95ef603c2470/src/warehouse/config/WHTracksTableConfig.js#L7).

warehouse/transformer/internal/rules/rules.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+			return nil, nil // nolint: nilnil
+		},
+	}
+
+	IdentifyRules = map[string]Rules{
```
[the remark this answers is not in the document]
achettyiitr: It's basically for the user's event; it does not have the default rule override.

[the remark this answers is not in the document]
achettyiitr: Also, let me rename it to `UsersRules` which might be more appropriate.

warehouse/transformer/internal/snakecase/snakecase.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"strings"
+
+	"github.com/dlclark/regexp2"
+	"github.com/samber/lo"
+)
+
+const (
+	// Used to compose unicode character classes.
+	rsAstralRange               = "\\ud800-\\udfff"
```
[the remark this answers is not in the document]
achettyiitr: No idea, just copied these constants from [here](https://github.com/rudderlabs/rudder-transformer/blob/develop/src/warehouse/snakecase/unicodeWords.js#L1).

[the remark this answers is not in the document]
achettyiitr: Removed.

warehouse/transformer/internal/snakecase/snakecase_test.go
```diff
@@ -0,0 +1,232 @@
...
+				actual := extractWords(strings.ToLower(ordinal))
+				require.Equal(t, expected, actual)
+
+				expected = []string{strings.ToUpper(ordinal)}
+				actual = extractWords(strings.ToUpper(ordinal))
+				require.Equal(t, expected, actual)
+			}
+		})
+	})
+	t.Run("extractWords", func(t *testing.T) {
```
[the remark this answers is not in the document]
achettyiitr: Expected values are different, it was more easy to have different test suite for each of them

warehouse/transformer/internal/stringlikeobject/stringlikeobject.go
```diff
@@ -0,0 +1,64 @@
...
+	return (minKey == 0 || minKey == 1) && maxKey-minKey+1 == len(obj)
+}
+
+func isNonNegativeInteger(str string) bool {
+	return lo.EveryBy([]rune(str), func(c rune) bool {
+		return c >= '0' && c <= '9'
+	})
+}
+
+func ToString(obj map[string]any) string {
```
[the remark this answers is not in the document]
achettyiitr: Currently no.

warehouse/transformer/jsonpaths.go
```diff
@@ -0,0 +1,39 @@
...
+	"strings"
+
+	"github.com/rudderlabs/rudder-server/warehouse/transformer/internal/utils"
+)
+
+func extractJSONPathInfo(jsonPaths []string) jsonPathInfo {
+	keysMap, legacyKeysMap := make(map[string]int), make(map[string]int)
+	for _, jsonPath := range jsonPaths {
+		if trimmedJSONPath := strings.TrimSpace(jsonPath); trimmedJSONPath != "" {
+			splitPaths := strings.Split(jsonPath, ".")
```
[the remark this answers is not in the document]
achettyiitr: Good catch. Modified testcases as well.

warehouse/transformer/safe.go
achettyiitr: Yes, it's not very heavy computation, but it still involves a couple of string manipulations, which might be crucial if the string is large. For a particular key since these don't changes, it should be fine to have a cache here.

warehouse/transformer/safe.go
achettyiitr: It's too much of an abstraction to get this common code to a separate function.

warehouse/transformer/types.go
```diff
@@ -0,0 +1,119 @@
...
+		// allowUsersContextTraits when set to false, if context.traits.* is present, it will be added only as context_traits_*
+		// e.g., for context.traits.name, only context_traits_name will be added to the user's table.
+		// For older destinations, it will come as true, and for new destinations this config will not be present, which means we will treat it as false.
+		allowUsersContextTraits bool
+	}
+
+	mergeRule struct {
+		Type, Value any
+	}
+	mergeRulesColumns struct {
```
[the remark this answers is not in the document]
achettyiitr: `Type` and `Value` are of any type.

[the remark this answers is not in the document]
achettyiitr: `Type` and `Value` are of any type.

warehouse/transformer/transformer_fuzz_test.go
achettyiitr: We can use `ValidTimestamp` exported function but here we wanted to capture group values as well from the regex, Since it's just for test, I think it should be fine. 
I don't want to make that regex in the utils package exported.

warehouse/transformer/safe.go
```diff
@@ -0,0 +1,273 @@
...
+	if !shouldSkipSnakeCasing(conf, destType) {
+		namespace = strcase.ToSnake(namespace)
+	}
+	if startsWithDigit(namespace) {
+		namespace = "_" + namespace
+	}
+	if namespace == "" {
+		namespace = "stringempty"
+	}
+	if reservedkeywords.IsNamespace(destType, namespace) {
```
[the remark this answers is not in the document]
shekhar-rudder: Is this change pushed ?

shekhar-rudder (approved):

RanjeetMishra (approved):

warehouse/transformer/safe.go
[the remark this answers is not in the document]
achettyiitr: I think I missed it. Just pushed now.

PR #5211, opened 2024-10-21
chore: limit auth status inactivate requests

sanpj2292:
# Description

As of present day, we are receiving a lot of requests for inactivating auth status. The auth status inactivation happens only when the integration responds with permission denial when accessing a resource kind of error.
In order to acheive limiting, whenever we get an auth status inactivation request, we would use an in-memory cache to store the status of inactivation for the first such request & the next such request will be checking if that destination is present in cache & skip making an API call altogether.

**Note**:
- The meaning of Inactivation is just updating the status to `inactive`

## Linear Ticket

Resolves INT-2744

## Security

### DIFF
--- a/services/oauth/v2/oauth.go
+++ b/services/oauth/v2/oauth.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/tidwall/gjson"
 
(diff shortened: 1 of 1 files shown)


sanpj2292:
This is not required

PR #5212, opened 2024-10-21
fix: dedup - use writeBatch to put keys in badger (#5204)

cisse21:
# Description

## Linear Ticket

## Security


services/dedup/scylla/scylla.go
```diff
@@ -33,6 +33,60 @@ func (d *ScyllaDB) Close() {
...
+
+	// Loop over each workspace and query all messageIDs for that workspace
+	for workspaceID, workspaceJobs := range workspaceJobsMap {
+		var messageIDs []string
+		for _, job := range workspaceJobs {
+			messageIDs = append(messageIDs, job.Key)
+		}
+
+		// Query to get all jobIDs for the given workspaceID and messageIDs
+		query := fmt.Sprintf("SELECT id, size FROM %s.%q WHERE workspaceID = ? AND id IN ?", d.keyspace, d.tableName)
```
mihir20, line 67: instead of querying for every batch can we do something like this? 
```
SELECT id, size
FROM ks.table 
WHERE (workspace_id = x AND message_id IN (list_of_message_ids_x)) 
   OR (workspace_id = y AND message_id IN (list_of_message_ids_y));
```

services/dedup/scylla/scylla.go
```diff
@@ -33,6 +33,60 @@ func (d *ScyllaDB) Close() {
...
+		var size int64
+		for iter.Scan(&dbMessageID, &size) {
+			val, ok := d.cache[dbMessageID]
+			if ok {
+				results[val] = false
+				sizes[val] = size
+			}
+			d.cacheMu.Lock()
+			delete(d.cache, dbMessageID)
+			d.cacheMu.Unlock()
```
mihir20, line 80: didn't understood this, if msgId is not in our local cache but is present in our db, we should store it in our local cache.
Instead here we are checking those msg ids which are not present in our local cache in db if it is present in db we are checking them in cache again but then deleting them from cache.

services/dedup/scylla/scylla.go
mihir20: This will not reduce qps iter will only do a paginated query. Not exactly sure on how iter is implemented for scylla, but I thought it would be similar to that of postgres.

mihir20: discussed offline, this way we would be querying multiple partitions which is not recommended by scylladb

processor/worker.go
```diff
@@ -72,7 +72,11 @@ func (w *worker) start() {
 		defer close(w.channel.transform)
 		defer w.logger.Debugf("preprocessing routine stopped for worker: %s", w.partition)
 		for jobs := range w.channel.preprocess {
-			w.channel.transform <- w.handle.processJobsForDest(w.partition, jobs)
+			val, err := w.handle.processJobsForDest(w.partition, jobs)
```
mihir20, line 75: do we also want to use `processJobsForDestv2` here?

services/dedup/scylla/scylla.go
mihir20: Discussed this offline, we only want to commit when we don't found the key in db so we are deleting the keys that are there in db.

processor/worker.go
cisse21: Fixed

mihir20 (approved):
LGTM

processor/worker_test.go
```diff
@@ -177,6 +177,27 @@ type mockWorkerHandle struct {
 	shouldProcessMultipleSubJobs bool
 }
 
+func (m *mockWorkerHandle) processJobsForDestV2(partition string, subJobs subJob) (*transformationMessage, error) {
```
mihir20, line 180: we should also add a test with `enableParallelScan` set to true

services/dedup/scylla/scylla.go
```diff
@@ -33,6 +33,63 @@ func (d *ScyllaDB) Close() {
...
+	d.cacheMu.Unlock()
+
+	// Loop over each workspace and query all messageIDs for that workspace
+	for workspaceID, workspaceJobs := range workspaceJobsMap {
+		var messageIDs []string
+		for _, job := range workspaceJobs {
+			messageIDs = append(messageIDs, job.Key)
+		}
+
+		messageIDChunks := lo.Chunk(messageIDs, d.batchSize)
```
mihir20, line 66: we should keep the read batch size and write batch size separate

cisse21: Done

ktgowtham (approved):
LGTM

processor/processor.go
```diff
@@ -2046,67 +2158,371 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) *transf
...
 		}
 	}
-
-	if len(statusList) != len(jobList) {
-		panic(fmt.Errorf("len(statusList):%d != len(jobList):%d", len(statusList), len(jobList)))
-	}
 	trackedUsersReportGenStart := time.Now()
-	trackedUsersReports := proc.trackedUsersReporter.GenerateReportsFromJobs(jobList, proc.getNonEventStreamSources())
-	proc.stats.trackedUsersReportGeneration(partition).SendTiming(time.Since(trackedUsersReportGenStart))
+	trackedUsersReports := proc.trackedUsersReporter.GenerateReportsFromJobs(preTrans.jobList, proc.getNonEventStreamSources())
```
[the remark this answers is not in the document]
cisse21: We have it [here](https://github.com/rudderlabs/rudder-server/pull/5212/files#diff-63264a27291123a1413bf5e86a8dcd368b9214681cafd30426fafbdaa2a182faR1953)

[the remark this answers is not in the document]
ktgowtham: I see it in `processJobsForDestV2` but not in `processJobsForDest`? or am I looking something wrong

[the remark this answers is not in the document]
cisse21: We should have it... Let me check once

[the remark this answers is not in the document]
cisse21: Fixed it

PR #5222, opened 2024-10-24
chore: reduce metrics cardinality

mihir20:
# Description

reduce metrics cardinality for rudder-server metrics

## Linear Ticket

pipe-1662

## Security


archiver/worker.go
```diff
@@ -151,7 +151,7 @@ func (w *worker) Stop() {
 }
 
 func (w *worker) uploadJobs(ctx context.Context, jobs []*jobsdb.JobT) (string, error) {
-	defer w.uploadLimiter.Begin(w.sourceID)()
```
cisse21, line 154: Isn't it better to get rid of the argument rather than passing it empty?

mihir20: 1. removing the key will be backwards incompatible change.
2. also I wanted to keep the option available if someone have a low cardinality with respect to the key then they have the option to pass the key. 

As this is a common library I wanted to keep the option open for the developers to choose if they want the key in the metrics or not.

Sidddddarth (approved):

ktgowtham (approved):

PR #5224, opened 2024-10-24
feat: webhook v2 spec upgrade

vinayteki95:
# Description

- V2 adapter is introduced to send v2 spec events
- Source transformer payloads will be built differently for different specs (v1 and v2)
- Features service will be aware of v2 flag from transformer and pick adapter accordingly

## Linear Ticket

Resolves INT-2802

## Security


services/transformer/features.go
```diff
@@ -74,8 +76,8 @@ func (*noopService) Regulations() []string {
 }
 
 func (*noopService) SourceTransformerVersion() string {
-	// v0 is deprecated
-	return V1
+	// v0 is deprecated and upgrading to v2
```
krishna2020, line 79: features_impl.go returns version based on `upgradedToSourceTransformV2` value received by /features response from transformer

when does this come into picture ? and what difference it makes returning v1 vs. v2 here ?

vinayteki95: This is used in some test cases as a placeholder for transformer service. This used to return v0 earlier but as v0 is deprecated v1 is the ideal response.
This doesn't affect the actual logic anywhere.

gateway/webhook/webhookTransformer.go
```diff
@@ -21,22 +25,25 @@ import (
...
+	Source       backendconfig.SourceT `json:"source"`
 }
 
-func (v1 *v1Adapter) getTransformerEvent(authCtx *gwtypes.AuthRequestContext, body []byte) ([]byte, error) {
+func (v1 *v1Adapter) getTransformerEvent(authCtx *gwtypes.AuthRequestContext, eventRequest []byte) ([]byte, error) {
 	source := authCtx.Source
 
 	v1TransformerEvent := V1TransformerEvent{
-		Event: body,
+		EventRequest: eventRequest,
```
koladilip, line 46: this renaming i.e. we need to change a lot of integrations or we need to change in the controller/service layer.

vinayteki95: We won't have to change anything on the transformer as the json that is being build from the structure is till the same
`EventRequest json.RawMessage `json:"event"`` -> here the json field is still **event**

koladilip (approved):

mihir20 (approved):
LGTM

PR #5227, opened 2024-10-24
fix: refresh connections to avoid hotspotting of load distribution

snarkychef:
# Description

## Linear Ticket

## Security

### DIFF
--- a/utils/sysUtils/httpclient.go
+++ b/utils/sysUtils/httpclient.go
@@ -2,10 +2,48 @@
 package sysUtils
 
 import (
+	"context"
 	"net/http"
+	"sync"
+	"time"
 )
 
 // HTTPClient interface
 type HTTPClientI interface {
 	Do(req *http.Request) (*http.Response, error)
+}
+
+type RecycledHTTPClient struct {
+	client *http.Client
+	lock   sync.Mutex
+}
+
+func NewRecycledHTTPClient(ctx context.Context, clientFunc func() *http.Client, ttl time.Duration) *RecycledHTTPClient {
+	recycledHTTPClient := &RecycledHTTPClient{
+		client: clientFunc(),
+	}
+
+	go func() {
+		ticker := time.NewTicker(ttl)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				recycledHTTPClient.lock.Lock()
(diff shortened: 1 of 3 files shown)


utils/sysUtils/httpclient.go
```diff
@@ -2,10 +2,48 @@
...
+		}
+	}()
+
+	return recycledHTTPClient
+}
+
+func (r *RecycledHTTPClient) GetClient() *http.Client {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	return r.client
```
lvrach, line 48: There is simpler way to do this, that doesn't require a background go routine.

During a GetClient request we check the timestamp, of ttl is expired we recreate a new one and return it.

utils/sysUtils/httpclient.go
```diff
@@ -2,10 +2,48 @@
...
+
+	go func() {
+		ticker := time.NewTicker(ttl)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				recycledHTTPClient.lock.Lock()
+				recycledHTTPClient.client = clientFunc()
```
lvrach, line 34: We need to run :

```suggestion
                 recycledHTTPClient.client.CloseIdleConnections()
				recycledHTTPClient.client = clientFunc()
```
https://pkg.go.dev/net/http#Client.CloseIdleConnections

lvrach (changes requested):
Let's add this change behind a flag.

I am bit sceptical if we should use this approach in production or look for a cleaner solution.

However, we can merge it on master behind a flag, and enable in selected cases.

utils/sysUtils/httpclient.go
```diff
@@ -3,9 +3,40 @@ package sysUtils
...
+	}
+}
+
+func (r *RecycledHTTPClient) GetClient() *http.Client {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
+	if r.ttl > 0 && time.Since(r.lastRefreshTime) > r.ttl {
+		r.client.CloseIdleConnections()
+		r.client = r.clientFunc()
```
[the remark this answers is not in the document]
lvrach: Thats true. We should combine this with: `IdleConnTimeout` (can be equal to `r.lastRefreshTime`) Eventually resources would freed up.

PR #5242, opened 2024-10-28
chore: preTransformation is it's own concurrnet step

Sidddddarth:
# Description

separate `preTransform` into its own concurrent step.
Allows us to separate out initial preprocessing(dedup, destination filtering) and following steps(store to arc, esch; validation; multiplexing).

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -127,10 +127,11 @@ type Handle struct {
 	transDebugger              transformationdebugger.TransformationDebugger
 	isolationStrategy          isolation.Strategy
 	limiter                    struct {
-		read       kitsync.Limiter
-		preprocess kitsync.Limiter
-		transform  kitsync.Limiter
-		store      kitsync.Limiter
+		read         kitsync.Limiter
+		preprocess   kitsync.Limiter
+		preTransform kitsync.Limiter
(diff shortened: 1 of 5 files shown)


processor/processor.go
```diff
@@ -1962,36 +1968,38 @@ func (proc *Handle) processJobsForDestV2(partition string, subJobs subJob) (*tra
...
+		jobList:                      jobList,
+		start:                        start,
+		sourceDupStats:               sourceDupStats,
+		dedupKeys:                    dedupKeys,
+	}
+}
+
+func (proc *Handle) generateTransformationMessage(preTrans *preTransformationMessage) *transformationMessage {
+	if proc.limiter.preTransform != nil {
+		defer proc.limiter.preTransform.BeginWithPriority(preTrans.partition, proc.getLimiterPriority(preTrans.partition))()
```
mihir20, line 2001:
```suggestion
		defer proc.limiter.preTransform.BeginWithPriority("", proc.getLimiterPriority(preTrans.partition))()
```
We can skip passing partition as that field is only used for metrics

mihir20 (commented):
LGTM, but can we also add a benchmark test to see if we are actually processing faster.

PR #5252, opened 2024-10-30
fix: parse error due incrorrect csv response from marketo-n

utsabc:
# Description

The issue is due to incorrect csv response from marketo, hence rather than reading all rows at once we are reading each row separately. 

For the erroneous rows we cannot identify which particular job as we cant generate hash to get corresponding job id hence these specific ones will be marked as success (we need to figure out a long term solution for this though)
Screenshot 2024-10-30 at 7 19 48 PM

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/apiService.go
+++ b/router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/apiService.go
@@ -208,9 +208,19 @@ func (m *MarketoAPIService) GetLeadStatus(url string) ([]map[string]string, *API
 
 	reader := csv.NewReader(strings.NewReader(string(body)))
 
-	rows, err := reader.ReadAll()
-	if err != nil {
-		return nil, &APIError{StatusCode: 500, Category: "Retryable", Message: "Error in parsing csv response"}
+	// read each row one by one
+	rows := make([][]string, 0)
+	for {
+		row, err := reader.Read()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			m.logger.Error("Error in parsing csv response", err.Error())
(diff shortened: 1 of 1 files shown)


router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/apiService.go
```diff
@@ -208,9 +208,19 @@ func (m *MarketoAPIService) GetLeadStatus(url string) ([]map[string]string, *API
...
-	if err != nil {
-		return nil, &APIError{StatusCode: 500, Category: "Retryable", Message: "Error in parsing csv response"}
+	// read each row one by one
+	rows := make([][]string, 0)
+	for {
+		row, err := reader.Read()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
```
koladilip, line 218: looks even error comes, record is returned, so can use it, can you try this by sample code

koladilip: we should handle this by removing \n characters in transformer

koladilip (approved):

router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/apiService.go
utsabc: Sure, will take it up

router/batchrouter/asyncdestinationmanager/marketo-bulk-upload/apiService.go
```diff
@@ -208,9 +208,19 @@ func (m *MarketoAPIService) GetLeadStatus(url string) ([]map[string]string, *API
...
+			m.logger.Error("Error in parsing csv response", err.Error())
```
krishna2020, line 219: Is line number captured in err.Error() ?

krishna2020 (approved):

PR #5253, opened 2024-10-30
fix: silent ignore of 4xx errors from reporting

itsmihir:
# Description

We are currently ignoring 4XX errors from reporting service (for example due to payload too large or invalid sourceDefinitionId value). This result in reports being silently dropped without generating any alert. 

By treating 4XX as failure, we ensure that reports are retired, allowing us to detect and receive alerts for high 4XX rates or report backlogs.

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-711/fix-silent-ignore-of-400-errors-in-reporting-service

## Security

### DIFF
--- a/enterprise/reporting/reporting.go
+++ b/enterprise/reporting/reporting.go
@@ -583,11 +583,7 @@ func (r *DefaultReporter) sendMetric(ctx context.Context, netClient *http.Client
 }
 
 func isMetricPosted(status int) bool {
-	if status == 429 {
-		return false
-	}
-
-	return status >= 200 && status < 500
+	return status >= 200 && status < 300
 }
 
 func getPIIColumnsToExclude() []string {


Sidddddarth (approved):

cisse21:
Wouldn't it cause noise if we have a developer doing local testing with app.dev and prod reporting? Wouldn't it be better to classify the errors on reporting side and then add retries over here?

itsmihir:
> Wouldn't it cause noise if we have a developer doing local testing with app.dev and prod reporting? Wouldn't it be better to classify the errors on reporting side and then add retries over here?

We already have error classification in place for the reporting service, and we retry these errors as needed. However, consider a recent issue we encountered: if the source definition ID is incorrect, the reporting service responds with a 400 error. As a result, those reports are completely lost.

cisse21 (approved):

PR #5262, opened 2024-11-04
chore: oauth v2 stats refactor

sanpj2292:
# Description

The way in which stats are written is not easily readable. Through this PR, we are trying to make the stat emission in oauthv2 module a bit more developer friendly and readable

## Linear Ticket

Resolves INT-2744

## Security


services/oauth/v2/oauth.go
```diff
@@ -342,14 +348,15 @@ func (h *OAuthHandler) GetRefreshTokenErrResp(response string, accountSecret *Ac
 // This method hits the Control Plane to get the account information
 // As well update the account information into the destAuthInfoMap(which acts as an in-memory cache)
 func (h *OAuthHandler) fetchAccountInfoFromCp(refTokenParams *RefreshTokenParams, refTokenBody RefreshTokenBodyParams,
-	authStats *OAuthStats, logTypeName string,
+	statsHandler OAuthStatsHandler, logTypeName string,
 ) (int, *AuthResponse, error) {
+	actionType := strings.Join(strings.Fields(strings.ToLower(logTypeName)), "_")
```
sanpj2292, line 353:
```Go
> strings.Join(strings.Fields(strings.ToLower("Refresh Token")), "_")
refresh_token
```

services/oauth/v2/stats.go
```diff
@@ -18,40 +19,57 @@ type OAuthStats struct {
...
+}
+
+func (m *OAuthStatsHandler) mergeTags(tags stats.Tags) stats.Tags {
+	allTags := m.defaultTags
+	for key, value := range tags {
+		allTags[key] = value
 	}
-	s.stats.NewTaggedStat(s.statName, stats.CountType, statsTags).Increment()
+	return allTags
+}
```
Sidddddarth, line 59: This should be the other way around. If we don't want to put values into `m.defaultTags`.
```
func (m *OAuthStatsHandler) mergeTags(tags stats.Tags) stats.Tags {
	for key, value := range m.defaultTags {
		tags[key] = value
	}
	return tags
}
```
Add a map lookup check if you don't want to override with default.

services/oauth/v2/stats.go
```diff
@@ -18,40 +19,57 @@ type OAuthStats struct {
...
+
+func (m *OAuthStatsHandler) getStatName(suffix string) string {
+	return strings.Join([]string{"oauth_action", suffix}, "_")
+}
```
Sidddddarth, line 63: minor
Can inline this function's call.

services/oauth/v2/stats.go
sanpj2292: Used a different way to solve this problem. Please check and let me know if it works

services/oauth/v2/stats.go
sanpj2292: Done

Sidddddarth (approved):

koladilip (approved):

PR #5264, opened 2024-11-06
chore: send multiple reports in a single request to reporting

vamsikrishnakandi:
# Description
Send multiple reports in a single request to reporting
- Removed the keys (from report object) from the identifier to include multiple reports
- For every new identifier we are adding payload for request to the values array
- So when limit for reports count in a request is reached, 
    - we are deleting the identifier, but this will be part of values array
    - new payload object will be created and added to values if new entry is found with same identifier
    - so if there are more reports with same identifier, they are split in to multiple request

## Linear Ticket
Completes OBS-419

## Security


satishrudderstack (approved):

enterprise/reporting/reporting.go
```diff
@@ -336,6 +336,10 @@ func (*DefaultReporter) getAggregatedReports(reports []*types.ReportByStatus) []
 			EventType:      report.StatusDetail.EventType,
 			ErrorType:      report.StatusDetail.ErrorType,
 		})
+
+		if len(metricsByGroup[identifier].StatusDetails) >= r.maxReportsCountInARequest.Load() {
+			delete(metricsByGroup, identifier)
```
mihir20, line 341: instead of deleting, can add this check before adding to the map, delete is a costly operation on a map.

itsmihir: By deleting entry from map we want to create a new batch request, how will adding this check before adding item will help?

itsmihir: Okay got it.
@vamsikrishnakandi Instead of deleting the key, we can assign it a new value if the size exceeds the limit.

vamsikrishnakandi: idea here is, if there are more reports with same identifier that allowed per request, we want to split them into multiple requests. 

so we are deleting the value associated with the identifier when limit is reached. this way next time report comes with same identifier, metric is recreated and pushed to values.

Made the change as you suggested. Now, we create the metric, associate it with the identifier in the group, and push it to values either when the identifier is not found or when the maximum length for StatusDetails is reached (the old value associated with the identifier is replaced when the max length is reached).

enterprise/reporting/reporting.go
```diff
@@ -282,16 +285,13 @@ func (*DefaultReporter) getAggregatedReports(reports []*types.ReportByStatus) []
...
-			strconv.Itoa(report.StatusDetail.StatusCode),
-			report.StatusDetail.EventName, report.StatusDetail.EventType,
 		}
 		return strings.Join(groupingIdentifiers, `::`)
 	}
 
 	for _, report := range reports {
 		identifier := reportIdentifier(report)
-		if _, ok := metricsByGroup[identifier]; !ok {
+		if _, ok := metricsByGroup[identifier]; !ok || len(metricsByGroup[identifier].StatusDetails) >= r.maxReportsCountInARequest.Load() {
```
mihir20, line 294: nit: instead of loading `r.maxReportsCountInARequest` every time we can load this once before the for loop.

enterprise/reporting/reporting_test.go
```diff
@@ -159,81 +159,183 @@ var _ = Describe("Reporting", func() {
...
-				StatusDetails: []*types.StatusDetail{
-					{
-						Status:         "some-status",
-						Count:          3,
-						ViolationCount: 5,
-						StatusCode:     200,
-						SampleResponse: "",
-						SampleEvent:    []byte(`{}`),
-						ErrorType:      "",
+		It("Should provide aggregated reports when batch size is 1", func() {
```
mihir20, line 166: we should add 1 more test for having multiple entries for same identifier in the expected response. same when batch size is more than 1.

vamsikrishnakandi: added this test of batch size > 1. for batch size = 1, this is covered by `Should provide aggregated reports when batch size is 1`

satishrudderstack (approved):

PR #5286, opened 2024-11-12
chore: webhook integration tests commandline

vinayteki95:
# Description

Enhancements in `TestIntegrationWebhook` to accept command-line argument/flag `-webhookversion`.

This flag initialises `rudder-transformer container` with extra env variable `UPGRADED_TO_SOURCE_TRANSFORM_V2=true` enabling webhook V2 spec requests.

## Linear Ticket

Resolves: INT-2880
https://linear.app/rudderstack/issue/INT-2880/stage-1-or-enhancement-of-integration-tests

## Security


koladilip (approved):

rudder-transformer
koladilip, line 1: is this expected?

vinayteki95: No, I think this was a mistake during rebase (recent changes to master seems to have removed rudder-transformer sub-module). Fixed it now.

gateway/webhook/integration_test.go
```diff
@@ -46,6 +47,17 @@ import (
 	"github.com/rudderlabs/rudder-transformer/go/webhook/testcases"
 )
 
+var webhookVersion string
```
achettyiitr, line 50: I didn't get how are we planning to run this test? Do we want to run the test for both v1 and v2? If so, we could have a table driven test for both v1 and v2 instead of this flag?

achettyiitr: I don't see anything modified in `tests.yaml` regarding webhookVersion?

gateway/webhook/integration_test.go
```diff
@@ -67,8 +79,17 @@ func TestIntegrationWebhook(t *testing.T) {
...
 		return nil
 	})
+
 	g.Go(func() (err error) {
-		transformerContainer, err = transformertest.Setup(pool, t)
+		if webhookVersion == "v2" {
+			// To run this test locally with custom repository & image tage, refer the below example.
+			// Example:
+			// 			transformertest.WithRepository("hub.dev-rudder.rudderlabs.com/dockerhub-proxy/rudderstack/develop-rudder-transformer"),
+			// 			transformertest.WithDockerImageTag("feat.sources-v2-spec-controller"),
```
achettyiitr, line 88: Do we need this comment?

gateway/webhook/integration_test.go
vinayteki95: Currently, webhook integration tests are run manually from rudder-transformer by the integrations team [using this](https://github.com/rudderlabs/rudder-transformer/blob/develop/go/README.md). This PR is only to enhance this with a commandline option so that we can test out webhook version with v2 as well.
cc: @sanpj2292 

Changes to run both the tests with both versions will come in a different PR associated with the [v2 server changes](https://linear.app/rudderstack/issue/INT-2802/stage-1-or-v2-spec-or-server-changes-for-v2-spec).

gateway/webhook/integration_test.go
vinayteki95: This helps while running the test cases in local, especially with an ongoing branch.
We can get rid of it as well.

achettyiitr: Yeah, let's get rid of these comments which are self explanatory.

achettyiitr (approved):

PR #5289, opened 2024-11-13
feat: send one sample event / response per label set to reporting

vamsikrishnakandi:
# Description
Send one sample event or response per LabelSet to reporting within the configured duration.
- Generate a hash for each report and store it in memory.
- If the hash is already present in the cache, do not resend the sample event or response. If the hash is not present, send the sample event/response.
- The hash stored in the cache expires after the configured TTL.
- Limit the number of keys stored in the cache. If the limit is exceeded, the hash will not be stored.

## Security

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -249,3 +249,7 @@ PgNotifier:
   retriggerCount: 500
   trackBatchInterval: 2s
   maxAttempt: 3
+Reporting:
+  eventSamplingEnabled: true
+  eventSamplingDurationInMinutes: 60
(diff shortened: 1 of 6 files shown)


cisse21:
Let's store the samples in BadgerDB instead of memory? This can get pretty bloated up in multi-tenant environments like we have seen with live events. WDYT @Sidddddarth @lvrach ?

vamsikrishnakandi:
https://github.com/rudderlabs/rudder-server/pull/5298

PR #5298, opened 2024-11-14
feat: send one sample event, response per label set in the configured duration

vamsikrishnakandi:
# Description
Send one sample event or response per LabelSet to reporting within the configured duration.
- Generate a hash for each report and store it in badger db.
- If the hash is already present in the db, do not resend the sample event or response. If the hash is not present, send the sample event/response.
- The hash stored in the db expires after the configured TTL.

## Linear Ticket
Completes OBS-461

## Security


satishrudderstack:
Since we are going to use ScyllaDB for de-dup, Will we still have badgerdb ?

satishrudderstack:
Since we are migrating to ScyllaDB for de-dup, Will we still have badgerdb  ?

enterprise/reporting/reporting.go
```diff
@@ -615,6 +631,33 @@ func transformMetricForPII(metric types.PUReportedMetric, piiColumns []string) t
...
+
+	if isValidSampleEvent {
+		hash := NewLabelSet(metric).generateHash()
+		found, err := r.eventSampler.Get([]byte(hash))
+		if err != nil {
+			panic(err)
+		}
+
+		if found {
+			metric.StatusDetail.SampleEvent = json.RawMessage(`{}`)
```
itsmihir, line 649: should we set this to NULL? @satishrudderstack

itsmihir (commented):
I’m skeptical about whether the getReports query using ARRAY_AGG will remain compatible with the proposed changes. Consider this scenario: for a given reportedAt value, we might fail to collect the sample event for the first occurrence (perhaps because the TTL hasn’t expired yet) but still manage to collect a sample event for a later occurrence within the same reportedAt. In this case, we won’t flush any sample event to reporting because ARRAY_AGG [1] selects the first record.

enterprise/reporting/event_sampler.go
```diff
@@ -0,0 +1,151 @@
...
+		WithIndexCacheSize(16 << 20). // 16mb
+		WithNumGoroutines(1).
+		WithNumMemtables(conf.GetInt("BadgerDB.numMemtable", 5)).
+		WithValueThreshold(conf.GetInt64("BadgerDB.valueThreshold", 1048576)).
+		WithBlockCacheSize(0).
+		WithNumVersionsToKeep(1).
+		WithNumLevelZeroTables(conf.GetInt("BadgerDB.numLevelZeroTables", 5)).
+		WithNumLevelZeroTablesStall(conf.GetInt("BadgerDB.numLevelZeroTablesStall", 15)).
+		WithSyncWrites(conf.GetBool("BadgerDB.syncWrites", false)).
+		WithDetectConflicts(conf.GetBool("BadgerDB.detectConflicts", false))
```
satishrudderstack, line 53: - I think we should have these BadgerDB configs separate for Reporting. The path can be `Reporting.badgerDB`
```suggestion
		WithLogger(badgerLogger{log}).
		WithCompression(options.None).
		WithIndexCacheSize(16 << 20). // 16mb
		WithNumGoroutines(1).
		WithNumMemtables(conf.GetInt("Reporting.badgerDB.numMemtable", 5)).
		WithValueThreshold(conf.GetInt64("Reporting.badgerDB.valueThreshold", 1048576)).
		WithBlockCacheSize(0).
		WithNumVersionsToKeep(1).
		WithNumLevelZeroTables(conf.GetInt("Reporting.badgerDB.numLevelZeroTables", 5)).
		WithNumLevelZeroTablesStall(conf.GetInt("Reporting.badgerDB.numLevelZeroTablesStall", 15)).
		WithSyncWrites(conf.GetBool("Reporting.badgerDB.syncWrites", false)).
		WithDetectConflicts(conf.GetBool("Reporting.badgerDB.detectConflicts", false))
```

config/config.yaml
```diff
@@ -249,3 +249,6 @@ PgNotifier:
   retriggerCount: 500
   trackBatchInterval: 2s
   maxAttempt: 3
+Reporting:
+  eventSamplingEnabled: false
+  eventSamplingDurationInMinutes: 60
```
satishrudderstack, line 254: Let's use more hierarchical config ?
```suggestion
Reporting:
  eventSampling:
      enabled: false
      durationInMinutes: 60
```

enterprise/reporting/event_sampler.go
satishrudderstack: Also, we should get this badger configs reviewed

config/config.yaml
vamsikrishnakandi: actually, no need to add these in this file, since we are using reloadable configs

enterprise/reporting/event_sampler.go
```diff
@@ -0,0 +1,151 @@
...
+		err := es.db.RunValueLogGC(0.5)
+		if err == nil {
+			goto again
+		}
+	}
+}
+
+func (es *EventSampler) Close() {
+	es.cancel()
+	es.wg.Wait()
```
satishrudderstack, line 135: There might be some issue with `WaitGroup` code here. `es.wg.Add()` is missing

enterprise/reporting/reporting.go
```diff
@@ -634,6 +673,34 @@ func transformMetricForPII(metric types.PUReportedMetric, piiColumns []string) t
...
 
+func (r *DefaultReporter) transformMetricWithEventSampling(metric types.PUReportedMetric, reportedAt int64) (types.PUReportedMetric, error) {
+	if r.eventSampler == nil {
+		return metric, nil
+	}
+
+	isValidSampleEvent := metric.StatusDetail.SampleEvent != nil && string(metric.StatusDetail.SampleEvent) != "{}"
+
+	if isValidSampleEvent {
+		bucket, _ := r.getAggregationBucketMinute(reportedAt, 1)
```
satishrudderstack, line 684:
```suggestion
		bucket, _ := r.getAggregationBucketMinute(reportedAt, int64(r.eventSamplingDuration.Load().Minutes()))
```

enterprise/reporting/event_sampler/badger_event_sampler.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return es.db.Update(func(txn *badger.Txn) error {
+		entry := badger.NewEntry([]byte(key), []byte{1}).WithTTL(es.ttl.Load())
+		return txn.SetEntry(entry)
+	})
+}
+
+func (es *BadgerEventSampler) performGC() {
+	es.wg.Add(1)
+	defer es.wg.Done()
```
satishrudderstack, line 113: We can remove these

enterprise/reporting/event_sampler/badger_event_sampler.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// One call would only result in removal of at max one log file.
+	// As an optimization, we can call it in a loop until it returns an error.
+	for {
+		if err := es.db.RunValueLogGC(0.5); err != nil {
+			break
+		}
+	}
```
satishrudderstack, line 121: If `performGC()` can be a long running function, we should have a way to stop it
```suggestion
    for {
        select {
        case <-es.ctx.Done():
            return
        default:
            if err := es.db.RunValueLogGC(0.5); err != nil {
                return
            }
        }
    }
```

enterprise/reporting/event_sampler/event_sampler_test.go
```diff
@@ -0,0 +1,164 @@
...
+		es, _ := NewEventSampler(ttl, eventSamplerType, eventSamplingCardinality, conf, log)
+		_ = es.Put("key1")
+
+		time.Sleep(600 * time.Millisecond)
+		val1, _ := es.Get("key1")
+
+		assert.False(t, val1, "Expected key1 to be evicted")
+		es.Close()
+	})
+}
```
satishrudderstack, line 49: Both these tests can fail in some cases depending if ttl runs fast/slow. I think we can use `require.Eventually`. `require.Eventually` used in multiple places in our tests

satishrudderstack, line 49:
```suggestion
	t.Run("should not get evicted keys", func(t *testing.T) {
		ttl := conf.GetReloadableDurationVar(100, time.Millisecond, "Reporting.eventSampling.durationInMinutes")
		es, _ := NewEventSampler(ttl, eventSamplerType, eventSamplingCardinality, conf, log)
		defer es.Close()

		_ = es.Put("key1")

		require.Eventually(t, func() bool {
			val1, _ := es.Get("key1")
			return assert.False(t, val1, "Expected key1 to be evicted")
		}, 60*time.Second, 10*time.Millisecond)
	})
```

enterprise/reporting/event_sampler/event_sampler_test.go
```diff
@@ -0,0 +1,164 @@
...
+	t.Run("should not get evicted keys", func(t *testing.T) {
+		es, _ := NewEventSampler(ttl, eventSamplerType, eventSamplingCardinality, conf, log)
+		_ = es.Put("key1")
+
+		time.Sleep(600 * time.Millisecond)
+
+		val1, _ := es.Get("key1")
+
+		assert.False(t, val1, "Expected key1 to be evicted")
+	})
```
satishrudderstack, line 83: Need to fix this test as well using require.Eventually

enterprise/reporting/reporting.go
```diff
@@ -90,11 +95,23 @@ func NewDefaultReporter(ctx context.Context, conf *config.Config, log logger.Log
...
+	eventSamplingCardinality := conf.GetReloadableIntVar(1000000, 1, "Reporting.eventSampling.cardinality")
 	// only send reports for wh actions sources if whActionsOnly is configured
 	whActionsOnly := config.GetBool("REPORTING_WH_ACTIONS_ONLY", false)
 	if whActionsOnly {
 		log.Info("REPORTING_WH_ACTIONS_ONLY enabled.only sending reports relevant to wh actions.")
 	}
+
+	if eventSamplingEnabled.Load() {
+		var err error
+		eventSampler, err = event_sampler.NewEventSampler(eventSamplingDuration, eventSamplerType, eventSamplingCardinality, conf, log)
```
satishrudderstack, line 110: @vamsikrishnakandi Let's make `eventSamplingDuration` a hot reloadable config. Essentially, we read the `eventSamplingDuration` from sampler

enterprise/reporting/reporting.go
```diff
@@ -276,6 +313,7 @@ func (r *DefaultReporter) getReports(currentMs, aggregationIntervalMin int64, sy
 func (r *DefaultReporter) getAggregatedReports(reports []*types.ReportByStatus) []*types.Metric {
 	metricsByGroup := map[string]*types.Metric{}
 	maxReportsCountInARequest := r.maxReportsCountInARequest.Load()
+	bucket, _ := r.getAggregationBucketMinute(reports[0].ReportedAt, 1)
```
satishrudderstack, line 316: `bucket` is only for collecting less sample_events right ? Let's rename it explicitly to `sample_event_bucket` ?

```suggestion
	bucket, _ := r.getAggregationBucketMinute(reports[0].ReportedAt, int64(r.eventSamplingDuration.Load().Minutes()))
```

enterprise/reporting/reporting.go
vamsikrishnakandi: this is reloadable. loading it in eventSampler (`es.ttl.Load()`)

enterprise/reporting/reporting.go
```diff
@@ -329,6 +369,7 @@ func (r *DefaultReporter) getAggregatedReports(reports []*types.ReportByStatus)
 				},
 				ReportMetadata: types.ReportMetadata{
 					ReportedAt: report.ReportedAt * 60 * 1000, // send reportedAt in milliseconds
+					Bucket:     sampleEventBucket * 60 * 1000,
```
satishrudderstack, line 372: [minor]: rename `Bucket` -> `SampleEventBucket` ?

satishrudderstack (approved):

enterprise/reporting/reporting.go
```diff
@@ -90,11 +96,23 @@ func NewDefaultReporter(ctx context.Context, conf *config.Config, log logger.Log
 	maxOpenConnections := config.GetIntVar(32, 1, "Reporting.maxOpenConnections")
 	dbQueryTimeout = config.GetReloadableDurationVar(60, time.Second, "Reporting.dbQueryTimeout")
 	maxReportsCountInARequest := conf.GetReloadableIntVar(10, 1, "Reporting.maxReportsCountInARequest")
+	eventSamplingEnabled := conf.GetReloadableBoolVar(false, "Reporting.eventSampling.enabled")
+	eventSamplingDuration := conf.GetReloadableDurationVar(60, time.Minute, "Reporting.eventSampling.durationInMinutes")
+	eventSamplerType := conf.GetReloadableStringVar("badger", "Reporting.eventSampling.type")
+	eventSamplingCardinality := conf.GetReloadableIntVar(1000000, 1, "Reporting.eventSampling.cardinality")
```
vamsikrishnakandi, line 102: @satishrudderstack should we keep default for `Reporting.eventSampling.type` as `badger` or `in_memory_cache`?

And what should we keep default value for `Reporting.eventSampling.cardinality`?

docker-compose.yml
```diff
@@ -7,27 +7,30 @@ services:
...
-      - db
-      - transformer
-    entrypoint: sh -c '/wait-for db:5432 -- ./rudder-server'
-    ports:
-      - "8080:8080"
-    env_file:
-      - build/docker.env
-    environment:
-      - JOBS_DB_HOST=db
+  #  backend:
```
itsmihir, line 10: let revert this changes.

scripts/generate-event
```diff
@@ -1,16 +1,105 @@
 #!/bin/bash
 
-if [[ $# != 2 ]]; then
-	echo "Usage: ./generate-event <write_key> <end_point>"
-	echo "Example: ./generate-event 1S0ibSaDlSSGkaQuHLi9feJqIUBNAE https://<data_plane_url>/v1/batch"
-	exit
+# Default values
```
itsmihir, line 3: lets revert this file too.

enterprise/reporting/event_sampler/event_sampler.go
```diff
@@ -0,0 +1,46 @@
...
+	var es EventSampler
+	var err error
+
+	switch eventSamplerType.Load() {
+	case BadgerTypeEventSampler:
+		es, err = NewBadgerEventSampler(BadgerEventSamplerPathName, ttl, conf, log)
+	case InMemoryCacheTypeEventSampler:
+		es, err = NewInMemoryCacheEventSampler(ttl, eventSamplingCardinality)
+	default:
+		err = fmt.Errorf("invalid event sampler type: %s", eventSamplerType.Load())
```
cisse21, line 39: probably default to either of the above types and leave a log instead of panicking

enterprise/reporting/event_sampler/event_sampler.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	return es, nil
+}
```
cisse21, line 46:
```suggestion
func NewEventSampler(
	ttl config.ValueLoader[time.Duration],
	eventSamplerType config.ValueLoader[string],
	eventSamplingCardinality config.ValueLoader[int],
	conf *config.Config,
	log logger.Logger,
) (es EventSampler, err error) {
	switch eventSamplerType.Load() {
	case BadgerTypeEventSampler:
		es, err = NewBadgerEventSampler(BadgerEventSamplerPathName, ttl, conf, log)
	case InMemoryCacheTypeEventSampler:
		es, err = NewInMemoryCacheEventSampler(ttl, eventSamplingCardinality)
	default:
		err = fmt.Errorf("invalid event sampler type: %s", eventSamplerType.Load())
	}
	return
}
```

enterprise/reporting/event_sampler/in_memory_cache_event_sampler.go
```diff
@@ -0,0 +1,56 @@
...
+	"sync"
+	"time"
+
+	"github.com/rudderlabs/rudder-go-kit/cachettl"
+	"github.com/rudderlabs/rudder-go-kit/config"
+)
+
+type InMemoryCacheEventSampler struct {
+	cache  *cachettl.Cache[string, bool]
+	mu     sync.Mutex
```
cisse21, line 13: we can get rid of the mu from what I can see.. cachettl already takes care of the safety

enterprise/reporting/event_sampler/badger_event_sampler.go
```diff
@@ -0,0 +1,153 @@
+package event_sampler
```
cisse21, line 1: add tests for both badger samples and in memory samplers

enterprise/reporting/reporting.go
```diff
@@ -90,11 +96,23 @@ func NewDefaultReporter(ctx context.Context, conf *config.Config, log logger.Log
...
 	// only send reports for wh actions sources if whActionsOnly is configured
 	whActionsOnly := config.GetBool("REPORTING_WH_ACTIONS_ONLY", false)
 	if whActionsOnly {
 		log.Info("REPORTING_WH_ACTIONS_ONLY enabled.only sending reports relevant to wh actions.")
 	}
+
+	if eventSamplingEnabled.Load() {
+		var err error
+		eventSampler, err = event_sampler.NewEventSampler(eventSamplingDuration, eventSamplerType, eventSamplingCardinality, conf, log)
```
cisse21, line 111: pass `ctx` into this function and propagate it across

enterprise/reporting/event_sampler/badger_event_sampler.go
```diff
@@ -0,0 +1,153 @@
...
+		ttl:    ttl,
+		ctx:    ctx,
+		cancel: cancel,
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	go es.gcLoop()
```
cisse21, line 69: use `rruntime.Go` instead of `go`

enterprise/reporting/event_sampler/badger_event_sampler.go
vamsikrishnakandi: these are present in `event_sampler_test.go`

enterprise/reporting/reporting.go
satishrudderstack: Like we discussed offline, let's have `badger` as default. We can switch to in-memory implementation if we see this is adding latency into the sytem.

For max cardinality
Assuming a limit of 1-10MB on the size of map and each entry size of ~50 Bytes or so

5 MB * 1024 KB * 1024 B / 50 B = ~100 K

satishrudderstack: @cisse21 Do we have metrics on how much latency badger adds in processing from deduplication use case ? For our case, I think the badger calls might be of the order of 1/5th to 1/10th of dedup calls. Do you see potential latency issues introduced in the processing due to this ?

cisse21 (approved):

PR #5314, opened 2024-11-20
chore: byte index to read singularEvent, propagate payload func to processor pipeline

Sidddddarth:
# Description

Use byte index to get event payload and propagate same to processor pipeline steps - using `sync.OnceValue` instead of `go-kit/memoize` so it can be propagated to other steps. This can be used in setting report payloads or settings event payloads to the error db.

Also updated tests to remove cases where processor picked up jobs from gateway that had multiple events - this hasn't been the case for a long time now.

## Linear Ticket

[Resolves PIPE-1717](https://linear.app/rudderstack/issue/PIPE-1717/avoid-marshalling-in-processor)

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1,6 +1,7 @@
 package processor
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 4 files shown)


mihir20 (approved):
LGTM

PR #5316, opened 2024-11-21
fix: json marshal l errors when parsing poll response from klaviyo

koladilip:
# Description

There is a bug in the Klaviyo implementation where, at times, we receive an empty import ID from the upload. However, when we poll with an empty import ID, we encounter parsing errors because Klaviyo returns a list response when the import ID is empty.

## Linear Ticket

https://linear.app/rudderstack/issue/INT-2820/klaviyo-bulk-upload-at-times-they-are-running-into-errors
Resolves INT-2820

## Security


yashasvibajpai (approved):

router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
```diff
@@ -106,7 +106,9 @@ func (kbu *KlaviyoBulkUploader) Poll(pollInput common.AsyncPoll) common.PollStat
 	importStatuses := make(map[string]string)
 	failedImports := make([]string, 0)
 	for _, importId := range importIds {
-		importStatuses[importId] = "queued"
+		if importId != "" {
```
krishna2020, line 109: does this mean all the events that were part of unsuccessful upload gets re-tried in next attempt as they don't get marked as queued ?

koladilip: yes, they will be retried

PR #5322, opened 2024-11-26
chore: remove unrecognised schema

cisse21:
# Description
remove unrecognised schema from warehouse module

## Linear Ticket

Fixes PIPE-1742

## Security


achettyiitr (approved):

warehouse/router/state_export_data.go
```diff
@@ -398,11 +398,6 @@ func (job *UploadJob) addColumnsToWarehouse(ctx context.Context, tName string, c
 
 	var columnsToAdd []whutils.ColumnInfo
 	for columnName, columnType := range columnsMap {
-		// columns present in unrecognized schema should be skipped
-		if job.schemaHandle.IsColumnInUnrecognizedSchema(tName, columnName) {
```
ktgowtham, line 402: How are we handling this case if we're removing the unrecognized schema. trying to understand the motivation of adding this code in first place.

cisse21: We are getting the columns from warehouse and deciding whether its unrecognised or not

Then we are again checking it the column we get is unrecognised or not in addColumns function but the unrecognised schema is already in warehouse. Did it make sense?

ktgowtham: There's one edge case that could happen. let me know if I'm wrong. 
There's a RECORD type column in a bigquery table. We don't support that [here](https://github.com/rudderlabs/rudder-server/blob/e7599ca07818b4027294f3bd081fd4063f0ab062/warehouse/integrations/bigquery/bigquery.go#L74)
If a field with same name appears in staging file, we might try to create a column(with whatever data type we decide) and fail.
In fetch schema, we're skipping the column in remote schema. [here](https://github.com/rudderlabs/rudder-server/blob/e7599ca07818b4027294f3bd081fd4063f0ab062/warehouse/integrations/bigquery/bigquery.go#L981)

cisse21: Why would we try to create a column which already exists?
We do a add column if not exists query iirc

ktgowtham: yeah, but in fetch schema, we're filtering the column, which we don't support. So the code doesn't know that column exists. I'm assuming the comparison is done with schema returned by fetch schema.

ktgowtham (approved):

PR #5326, opened 2024-11-27
feat: aggregate reports based on the configured interval before sending to reporting

itsmihir:
# Description

Currently, the Rudder Server aggregates reports every minute and then flushes them to the Reporting Service. We want to make the aggregation interval configurable, allowing reports to be aggregated over a specified time period, such as 2 minutes instead of the default 1 minute.

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-415/support-aggregation-time-as-a-configuration-on-rudder-server-ability

## Security


mihir20 (approved):
LGTM

enterprise/reporting/reporting.go
```diff
@@ -345,6 +352,13 @@ func (r *DefaultReporter) getAggregatedReports(reports []*types.ReportByStatus)
 	return values
 }
 
+func (*DefaultReporter) getAggregationBucketMinute(timeMin, aggregationIntervalMin int64) (int64, int64) {
+	bucketStart := timeMin - (timeMin % aggregationIntervalMin)
+	bucketEnd := bucketStart + aggregationIntervalMin
+
+	return bucketStart, bucketEnd
+}
+
```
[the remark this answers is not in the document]
itsmihir: if `aggregationIntervalMin` is less than or equal to 60 mins then [bucketStart, bucketEnd) will belong to same hour. Do you mean we should add a check for aggregationIntervalMin > 60?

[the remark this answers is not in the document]
itsmihir: So should we panic if the value is not a factor of 60?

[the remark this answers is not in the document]
vamsikrishnakandi: we can add a comment next to the flag - // factors of 60 (1, 2, 3, 4, 5, 6, 10, 12, 15, 20, 30, and 60)

enterprise/reporting/reporting.go
```diff
@@ -400,7 +415,12 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
 			currentMin := time.Now().UTC().Unix() / 60
 
 			getReportsStart := time.Now()
-			reports, reportedAt, err := r.getReports(currentMin, c.ConnInfo)
+			aggregationIntervalMin := int64(aggregationInterval.Load().Minutes())
+			if aggregationIntervalMin <= 0 || 60%aggregationIntervalMin != 0 {
+				panic(fmt.Errorf("[ Reporting ]: Error aggregationIntervalMinutes should be a factor of 60, got %d", aggregationIntervalMin))
```
mihir20, line 420: this panic will cause application to halt processing until we fix the config which may take time during code freeze, do you think instead of panic we should add an error log and raise a log based alert for this?

Also please specify why are we adding this check in comments

itsmihir: One potential issue with continuing the processing is that data from the next hour might get aggregated with the current hour’s data. This is undesirable and needs to be avoided.

mihir20: In that case, fire an alert and don't initialize reporter. Processor's main task should be to process the events. Reporting is the secondary task which should not block the primary task.

enterprise/reporting/reporting.go
```diff
@@ -342,6 +350,29 @@ func (r *DefaultReporter) getAggregatedReports(reports []*types.ReportByStatus)
 	return values
 }
 
+func (*DefaultReporter) getAggregationBucketMinute(timeMs, intervalMs int64) (int64, int64) {
+	// If interval is not a factor of 60, then the bucket start will not be aligned to hour start
+	// For example, if intervalMs is 7, and timeMs is 28891085 (6:05) then the bucket start will be 28891079 (5:59)
```
mihir20, line 355:
```suggestion
	// For example, if intervalMs is 7, and timeMs is 28891085 (6:05) then the bucket start will be 28891079 (5:59)
	// and current bucket will contain the data of 2 different hourly buckets, which is should not have happened.
```

enterprise/reporting/reporting.go
```diff
@@ -460,7 +493,9 @@ func (r *DefaultReporter) mainLoop(ctx context.Context, c types.SyncerConfig) {
 				if err != nil {
 					return err
 				}
-				_, err = dbHandle.Exec(`DELETE FROM `+ReportsTable+` WHERE reported_at = $1`, reportedAt)
+				// Use the same aggregationIntervalMin value that was used to query the reports in getReports()
+				bucketStart, bucketEnd := r.getAggregationBucketMinute(reportedAt, aggregationIntervalMin)
+				_, err = dbHandle.Exec(`DELETE FROM `+ReportsTable+` WHERE reported_at >= $1 and reported_at < $2`, bucketStart, bucketEnd)
```
mihir20, line 498: why are we deleting from bucket only we can delete all the entries before last reported time?

satishrudderstack (approved):

enterprise/reporting/reporting.go
itsmihir: we can delete all the reports `< bucketEnd`, I will change the filter in `getReports()` too.

satishrudderstack: [minor]: Very unlikely, but what if there are rows added with older time while we are processing current window

itsmihir: yeah, revert the changes, its better to add the bucketStart

vamsikrishnakandi (approved):

PR #5330, opened 2024-12-01
feat: async framework handling for failures

achettyiitr:
# Description

- With the current setup, we mark the jobs as aborted when the failure happens around Polling. Add support for marking the failed jobs as Failed so they can be retried, aborted jobs will be marked as aborted.

## Linear Ticket

- Resolves PIPE-1755

## Security

- [x] The code changed/added as part of this pull request will not create any security issues with the software's use.


koladilip (approved):

mihir20 (approved):
LGTM

PR #5336, opened 2024-12-04
chore: drop upload id from wh_schema

cisse21:
# Description
Removes unused wh_upload_id from wh_schema table

## Linear Ticket

Fixes WAR-102

## Security

### DIFF
--- a/sql/migrations/warehouse/000026_drop_wh_upload_id_column.up.sql
+++ b/sql/migrations/warehouse/000026_drop_wh_upload_id_column.up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE wh_schemas ALTER COLUMN wh_upload_id DROP NOT NULL;
+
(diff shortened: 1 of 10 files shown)


achettyiitr (approved):

achettyiitr (approved):

sql/migrations/warehouse/000026_drop_wh_upload_id_column.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE wh_schemas DROP COLUMN IF EXISTS wh_upload_id;
```
ktgowtham, line 1: Let's make this nullable in this release. If we revert image, not having this column will create problems. 
In the next release we can drop the column. We can capture a task in linear to not miss this.

ktgowtham (changes requested):

ktgowtham (approved):

PR #5337, opened 2024-12-04
feat: snowpipe http compression for insert request

achettyiitr:
# Description

- Add support for making gzipped Insert requests to the Snowpipe client,

## Linear Ticket

- Resolves PIPE-1765

## Security


koladilip (approved):

router/batchrouter/asyncdestinationmanager/snowpipestreaming/internal/api/api_integration_test.go
```diff
@@ -139,25 +143,104 @@ func TestAPIIntegration(t *testing.T) {
...
+				t.Log("Inserting records")
+				insertRes, err := snowpipeAPI.Insert(ctx, createChannelRes.ChannelID, &model.InsertRequest{
+					Rows:   rows,
+					Offset: "8",
+				})
+				require.NoError(t, err)
+				require.Equal(t, &model.InsertResponse{Success: true, Errors: nil}, insertRes)
+				require.EqualValues(t, tc.payloadSize, statsStore.Get("snowpipe_streaming_request_body_size", stats.Tags{
+					"api": "insert",
+				}).LastValue())
```
mihir20, line 240: we should also check if we are getting correct data after the compression.

achettyiitr: Good point. I was already doing it in "Create channel + Get channel + Insert data + Status" since the config is enabled by default. 
Let me modify the test so that is can do it for both compression enabled and disabled.

mihir20 (approved):
LGTM

PR #5340, opened 2024-12-05
chore: cleanup staging repo to use upload_id for queries

achettyiitr:
# Description

Since `wh_staging_files` has a `upload_id,` we can use that to find associated staging files. We don't need additional filters for `source_id` and `destination_id`.

## Linear Ticket

- Resolves WAR-103

## Security

### DIFF
--- a/warehouse/internal/repo/staging.go
+++ b/warehouse/internal/repo/staging.go
@@ -291,25 +291,16 @@ func (sf *StagingFiles) GetSchemasByIDs(ctx context.Context, ids []int64) ([]mod
 	return schemas, nil
 }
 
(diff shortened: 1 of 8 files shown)


Sidddddarth (approved):
LGTM
Consider removing the indices on these columns(or the columns themselves in the future) too if they're present.

achettyiitr:
> LGTM Consider removing the indices on these columns(or the columns themselves in the future) too if they're present.

We still need `source_id` and `destination_id` filters since they are being used to get the Pending staging files that are not associated with an uploadID.

cisse21 (approved):

PR #5342, opened 2024-12-05
fix: struct tag for connection details reporting type

koladilip:
# Description

The struct tag was wrong earlier, and I renamed the field. Then, it broke the reporting API contract, so I am fixing the struct tag.

## Linear Ticket

https://linear.app/rudderstack/issue/INT-2973/[chore]-propagate-error-category-from-stattags-of-transformer
Resolves INT-2973

## Security

### DIFF
--- a/utils/types/reporting_types.go
+++ b/utils/types/reporting_types.go
@@ -159,7 +159,7 @@ type ConnectionDetails struct {
 	SourceJobID             string `json:"sourceJobId"`
 	SourceJobRunID          string `json:"sourceJobRunId"`
 	SourceDefinitionID      string `json:"sourceDefinitionId"`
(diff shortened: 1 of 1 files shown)


achettyiitr (approved):

itsmihir (approved):

achettyiitr:
@itsmihir @satishrudderstack Not related to this, But can we add some tests around this? Currently, we don't have any test on the payload that is being sent to the reporting.

koladilip:
@

> @itsmihir @satishrudderstack Not related to this, But can we add some tests around this? Currently, we don't have any test on the payload that is being sent to the reporting.

@itsmihir can you create a task for this?

PR #5345, opened 2024-12-05
fix: reporting type status detail

koladilip:
# Description

StatTags was added to the StatusDetail type for category computation but it was getting propagated to reporting API and causing some validation to fail.

## Linear Ticket

https://linear.app/rudderstack/issue/INT-2973/[chore]-propagate-error-category-from-stattags-of-transformer
Resolves INT-2973

## Security


utils/types/reporting_types_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Compare with expected JSON
+	var expected, actual map[string]interface{}
+	if err := json.Unmarshal([]byte(expectedJSON), &expected); err != nil {
+		t.Fatalf("Failed to unmarshal expected JSON: %v", err)
+	}
+	if err := json.Unmarshal(marshaledJSON, &actual); err != nil {
+		t.Fatalf("Failed to unmarshal actual JSON: %v", err)
+	}
+	require.Equal(t, expected, actual)
```
achettyiitr, line 136: We can simply compare the `marshaledJSON` with the expected JSON right?
```suggestion
	marshaledJSON, err := json.Marshal(metric)
	require.NoError(t, err)
	require.JSONEq(t, expectedJSON, string(marshaledJSON))
```

achettyiitr (approved):

cisse21 (approved):

PR #5360, opened 2024-12-10
fix: reports aggregation query to select sample_event and sample_response correctly

itsmihir:
# Description

For warehouse module sample event will always be empty, we need to update the FILTER on ARRAY_AGG in reporting reports aggregation query to select samples response even if sample event is empty. 

## Linear Ticket

Resolves OBS-732
[OBS-732: Fix reports aggregation query on rudder-server to handle empty sample event](https://linear.app/rudderstack/issue/OBS-732/fix-reports-aggregation-query-on-rudder-server-to-handle-empty-sample)

## Security


satishrudderstack (approved):
We should tests for these cases later

vamsikrishnakandi (approved):

PR #5369, opened 2024-12-11
chore: add error msg in the logs when gw req fails

mihir20:
# Description

add error msg in the logs when gw req fails

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1797/better-error-logging-for-ingestion-requests

## Security

### DIFF
--- a/gateway/handle.go
+++ b/gateway/handle.go
@@ -782,9 +782,10 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(reqType string, body []byt
 		stat := gwstats.SourceStat{ReqType: reqType}
 		err := gw.streamMsgValidator(&msg)
 		if err != nil {
+			loggerFields := msg.Properties.LoggerFields()
+			loggerFields = append(loggerFields, obskit.Error(err))
 			gw.logger.Errorn("invalid message in request",
-				logger.NewStringField("properties", msg.Properties.String()),
-				obskit.Error(err))
+				loggerFields...)
 			stat.RequestEventsFailed(1, response.InvalidStreamMessage)
 			stat.Report(gw.stats)
 			return nil, errors.New(response.InvalidStreamMessage)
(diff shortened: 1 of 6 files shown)


gateway/handle.go
```diff
@@ -663,7 +663,7 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
 			stat.Report(gw.stats)
 			goto requestError
```
BonapartePC, line 664: All the errors going to `requestError` should be printing error message. Examples: Line 664 and 680

gateway/handle.go
```diff
@@ -706,14 +706,17 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
 		return
 
 	requestError:
-		errorMessage = err.Error()
+		if errorMessage == "" {
+			errorMessage = err.Error()
+		}
 		status = response.GetErrorStatusCode(errorMessage)
 		responseBody = response.GetStatus(errorMessage)
-		gw.logger.Infon("response",
```
BonapartePC, line 712: Let's also log properties if they are available after unmarshalling

BonapartePC (approved):

gateway/handle.go
```diff
@@ -763,14 +766,14 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(reqType string, body []byt
 	if err != nil {
 		stat.RequestFailed(response.InvalidJSON)
 		stat.Report(gw.stats)
-		return nil, errors.New(response.InvalidJSON)
+		return nil, response.InvalidJSON, fmt.Errorf("unmarshalling request body: %w", err)
```
lvrach, line 769: This is not an elegant way to do this. Using wrapping and using `if errors.As(err, &e)` to extract response error would be a better approach. https://go.dev/blog/go1.13-errors

If this is too complex given the time constrains, I will log the error here.

fracasula (changes requested):

gateway/handle.go
```diff
@@ -763,13 +764,17 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(reqType string, body []byt
...
+			logger.NewStringField("body", string(body)),
+			logger.NewErrorField(err))
 		return nil, errors.New(response.InvalidJSON)
 	}
 	gw.requestSizeStat.Observe(float64(len(body)))
 
 	if len(messages) == 0 {
 		stat.RequestFailed(response.NotRudderEvent)
 		stat.Report(gw.stats)
+		gw.logger.Errorn("no messages in request", logger.NewStringField("body", string(body)))
```
fracasula, line 777: Again, let's not dump the body. Payloads should be encrypted and uploaded somewhere safe, not logged like this.

fracasula: Same applies to all other places where the message payload is being dumped.

fracasula: Uploading should be encrypted imo.

fracasula (approved):

PR #5372, opened 2024-12-13
chore: event_payload column can be JSONB or TEXT

Sidddddarth:
# Description

`event_payload` column in newly created jobs tables will be `TEXT` columns.

Allows migrations from one type to the other, so changing from one to the other shouldn't fail. Changing back to `JSONB` would allow server processes to continue, however utility postgres functions like `unionjobsdb` would fail as the function definition returns table with different column types.

At startup(for `w`, `rw` jobsdbs) - we either update the last DS's event_payload column if empty or create a new dataset if the last has some jobs already - both provided the event type column of the last ds is different from what's configured.

Not a reloadable selection.

Allowed creating `BYTEA` columns but only for benchmark purposes. `TEXT` seems to be doing better and also avoids other character validation problems.

## Linear Ticket

[Resolves PIPE-1649](https://linear.app/rudderstack/issue/PIPE-1649/use-binarytext-to-store-payload-in-jobsdb)

## Security


jobsdb/jobsdb.go
```diff
@@ -1057,8 +1089,59 @@ func (jd *Handle) writerSetup(ctx context.Context, l lock.LockToken) {
...
+		case payloadColumnType(1):
+			payloadType = "bytea"
+		case payloadColumnType(2):
+			payloadType = "text"
+		default:
+			jd.assertError(fmt.Errorf("invalid type: %d", jd.conf.payloadColumnType))
+		}
+		_, err := jd.dbHandle.ExecContext(
+			ctx,
+			fmt.Sprintf(`alter table %q alter column event_payload type %s`, dsList[len(dsList)-1].JobTable, payloadType),
```
ktgowtham, line 1142: What's the need of altering last table? why can't we start with new dataset?

jobsdb/jobsdb.go
```diff
@@ -2215,6 +2307,7 @@ func (jd *Handle) getJobsDS(ctx context.Context, ds dataSetT, lastDS bool, param
 	resultsetStates := map[string]struct{}{}
 	for rows.Next() {
 		var job JobT
+		var payload Payload
```
ktgowtham, line 2310: Why couldn't this be just a []byte?

jobsdb/jobsdb.go
```diff
@@ -770,6 +798,10 @@ func (jd *Handle) init() {
 		jd.config = config.Default
 	}
 
+	if jd.conf.payloadColumnType == 0 {
+		jd.conf.payloadColumnType = payloadColumnType(jd.config.GetIntVar(2, 1, "JobsDB.payloadColumnType"))
```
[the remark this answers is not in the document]
Sidddddarth: because it's faster than using jsonb column

jobsdb/jobsdb.go
Sidddddarth: `rows.Scan` errors if we scan one type into the other.

ktgowtham: What about `jsParameters`? it's a jsonb. but we're reading []byte. I think this should work. I myself wrote a test and it seems to be working fine with jsonb as well as text. Could you double check.

Sidddddarth: If the payload column were to be `TEXT` then we'd see an error when scanning into `[]byte`

Sidddddarth: Correction: `[]byte` satisfies all of them. will revert this bit

jobsdb/migration.go
```diff
@@ -459,18 +459,85 @@ func (jd *Handle) getMigrationList(dsList []dataSetT) (migrateFrom []dsWithPendi
 	return
 }
 
+func getColumnConversion(srcType, destType string) string {
```
lvrach, line 462: Is it possible to find a universal way to do the mapping, that will always work?

jobsdb/jobsdb.go
```diff
@@ -770,6 +798,10 @@ func (jd *Handle) init() {
 		jd.config = config.Default
 	}
 
+	if jd.conf.payloadColumnType == 0 {
+		jd.conf.payloadColumnType = payloadColumnType(jd.config.GetIntVar(int(TEXT), 1, "JobsDB.payloadColumnType"))
+	}
```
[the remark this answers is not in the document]
Sidddddarth: yeah will make the config change, makes sense
as for the default - I've just used the new TEXT column because it's faster..
No reason to not use it right..? We'll have to change all environments to TEXT at some point.

jobsdb/jobsdb.go
Sidddddarth: yeah
removing this entire block
startup continues as is

jobsdb/integration_test.go
```diff
@@ -1111,6 +1112,174 @@ func TestJobsDB_SanitizeJSON(t *testing.T) {
 			string(unprocessedJob.Jobs[0].EventPayload),
 		)
 	}
+	jobDB.TearDown()
+
+	conf.Set("JobsDB.payloadColumnType", 2)
+	textDB := &Handle{config: conf}
+	require.NoError(t, textDB.Setup(ReadWrite, true, strings.ToLower(rand.String(5))))
```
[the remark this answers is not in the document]
Sidddddarth: done - incorporating fuzz tests is pending

jobsdb/jobsdb.go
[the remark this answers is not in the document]
Sidddddarth: changed it to `JSONB`

jobsdb/jobsdb.go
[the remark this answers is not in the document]
Sidddddarth: changed to `JSONB`.

jobsdb/migration.go
```diff
@@ -459,18 +459,77 @@ func (jd *Handle) getMigrationList(dsList []dataSetT) (migrateFrom []dsWithPendi
...
+	return result, nil
+}
+
 func (jd *Handle) migrateJobsInTx(ctx context.Context, tx *Tx, srcDS, destDS dataSetT) (int, error) {
 	defer jd.getTimerStat(
 		"migration_jobs",
 		&statTags{CustomValFilters: []string{jd.tablePrefix}},
 	).RecordDuration()()
 
+	columnTypeMap := map[string]string{srcDS.JobTable: string(JSONB), destDS.JobTable: string(JSONB)}
```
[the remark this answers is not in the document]
cisse21: We would have to change the default at some point and it would effect OSS users too and we will need to have this piece of code then. So don't think this should be under a boolean config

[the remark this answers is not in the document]
ktgowtham: okay, sounds good. one way to go is observe if there's any diff on `migration_jobs` time and take decision. Assuming it won't add much given it's metadata query.

ktgowtham (approved):
Looks good

cisse21 (approved):

PR #5375, opened 2024-12-16
feat: client side load balancing for user transformations

lvrach:
# Description

This PR is attempting to address an imbalance issue observed in production. The rudder-server client tends to open more connections proportionally to older UT or ha-proxy pods. Not allowing the downstream system to scale horizontally.

The PR introduces two alternative HTTP clients to address that:
1. **recycled**: A new client is created every x minutes (configurable). This allows TCP connections to be reestablished in a more balanced way. However: 
    * This approach doesn't guarantee perfect balancing but doesn't allow the system to deteriorate enough.
    * There is an overhead in connection creation and clean-up. 
    * There a possibility of resource leak
2. [**httplb**](https://github.com/bufbuild/httplb) : load-balances TCP connections between available endpoint. Endpoint discovery is done via DNS and requires a headless service in k8s to work. It refreshes newly added or removed pods using `Transformer.Client.ttl`. It uses the same transport as the standard client and the same configuration.
3. **stdlib**: standard HTTP library, what we have today.

Clients can be selected from `Transformer.Client.type` config, by default `stdlib` is used. 

## Testing

* `recycled` was heavily tested in load-test environment under heavy load for a long time. It manage to maintain good enough latencies for a dedicated user transformer.

* httplb is also tested in the same environment, but with mixed results:
  * It maintains perfect balance among all pods.
  * It causes ut pod HPA to scale up and down constantly, fine-tuning HPA might be required
  * Also we want to verify how well it balances between different processes of UT. 
  

Note: due to node.js clustering, UT is spawning multiple processes, all subscribed to the same port.

Here you can see the comparison between `recycled` (left) and `httplb` (right):
Screenshot 2024-11-13 at 7 47 49 PM (1)

Assuming HPA stability is fixed, `httplb` would fair much better than `recycled`.

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1720/client-side-load-balancing-for-transformations

resolves PIPE-1720

## Security


cisse21 (approved):

ktgowtham (approved):
LGTM

PR #5389, opened 2024-12-23
fix: replay tracking plan bug

cisse21:
# Description
This PR is to fix the issue in the replay flow for connections with tracking plan . We were generating the commonMetaData from the input struct and were also altering the metadata before the transformer call while the commonMetaData was generated post the transformer call resulting in incorrect metadata propogated downstream. This PR fixes this by generating the commonMetaData before the transformer call 

## Linear Ticket

Fixes PIPE-1813

## Security

### DIFF
--- a/processor/trackingplan.go
+++ b/processor/trackingplan.go
@@ -89,6 +89,10 @@ func (proc *Handle) validateEvents(groupedEventsBySourceId map[SourceIDT][]trans
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


Sidddddarth (approved):

cisse21:
@mihir20 It isn't that easy to add a test case for this scenario. Talked to Tejo and we would be adding this task to backlog, merging for now

PR #5390, opened 2024-12-23
feat: add support for hash to bingads offline conversions

ItsSudip:
# Description

We are adding support for hashing email and phone for bingads offline conversion destination.

## Linear Ticket

resolves INT-3019

## Security


router/batchrouter/asyncdestinationmanager/bing-ads/offline-conversions/bulk_uploader.go
```diff
@@ -19,15 +19,16 @@ import (
 	"github.com/rudderlabs/rudder-server/router/batchrouter/asyncdestinationmanager/common"
 )
 
-func NewBingAdsBulkUploader(logger logger.Logger, statsFactory stats.Stats, destName string, service bingads.BulkServiceI, client *Client) *BingAdsBulkUploader {
+func NewBingAdsBulkUploader(logger logger.Logger, statsFactory stats.Stats, destName string, service bingads.BulkServiceI, client *Client, isHashRequired bool) *BingAdsBulkUploader {
```
koladilip, line 22: params are increasing so we need do something

koladilip: looks `client *Client` is not used at all

koladilip: instead of isHashRequired, can we pass the destination config and deal with in function

ItsSudip: I have removed `client` as we are not using that. But I am not sending the entire destination config as of now as only `isHashRequired` is required.

yashasvibajpai (approved):

koladilip (approved):

PR #5396, opened 2024-12-24
chore: send sample event to reporting in async destinations

sanpj2292:
# Description

As of today, the destinations operating at batch router do not posess the capability of sending sample event to the reporting service. This is making it very difficult to debug in case of failures.

We are planning add `sampleEvent` data for batch router(async) destinations.

## Linear Ticket

Resolves INT-3077

## Security


router/batchrouter/handle_async.go
```diff
@@ -452,7 +462,14 @@ func (brt *Handle) sendJobsToStorage(batchJobs BatchedJobs) {
 			out.SuccessResponse = fmt.Sprintf(`{"error":"%s"`, rterror.DisabledEgress.Error()) // skipcq: GO-R4002
 		}
 
-		brt.setMultipleJobStatus(out, false, getAttemptNumbers(batchJobs.Jobs), getFirstAttemptAts(batchJobs.Jobs), getOriginalJobParameters(batchJobs.Jobs))
+		brt.setMultipleJobStatus(SetMultipleJobStatusParams{
+			AsyncOutput:           out,
+			Attempted:             false,
```
koladilip, line 467: no need to pass false values as go default false anyway

koladilip (approved):

router/batchrouter/handle_async.go
sanpj2292: Addressed

router/batchrouter/handle_async.go
```diff
@@ -560,17 +590,27 @@ func (brt *Handle) createFakeJob(jobID int64, parameters stdjson.RawMessage) *jo
 	}
 }
 
-func (brt *Handle) getReportMetrics(statusList []*jobsdb.JobStatusT, parametersMap map[int64]stdjson.RawMessage) []*utilTypes.PUReportedMetric {
+type GetReportMetricsParams struct {
+	StatusList    []*jobsdb.JobStatusT
+	ParametersMap map[int64]stdjson.RawMessage
+	JobsList      []*jobsdb.JobT
+}
```
achettyiitr, line 597: Can we move these to `types.go` instead of keeping it here?

achettyiitr, line 597: Does these needs to be exported? Same for `SetMultipleJobStatusParams` as well?
```suggestion
type getReportMetricsParams struct {
	statusList    []*jobsdb.JobStatusT
	parametersMap map[int64]stdjson.RawMessage
	jobsList      []*jobsdb.JobT
}
```

router/batchrouter/handle_async.go
```diff
@@ -560,17 +590,27 @@ func (brt *Handle) createFakeJob(jobID int64, parameters stdjson.RawMessage) *jo
...
 	reportMetrics := make([]*utilTypes.PUReportedMetric, 0)
 	connectionDetailsMap := make(map[string]*utilTypes.ConnectionDetails)
 	transformedAtMap := make(map[string]string)
 	statusDetailsMap := make(map[string]*utilTypes.StatusDetail)
 	routerWorkspaceJobStatusCount := make(map[string]int)
-	for _, status := range statusList {
+	jobsMap := make(map[int64]*jobsdb.JobT)
+	for _, job := range params.JobsList {
+		jobsMap[job.JobID] = job
+	}
```
achettyiitr, line 608: We should pre-allocate the map since we already know what it contains? Or we can just use
[lo](https://github.com/samber/lo) package for this.
```suggestion
	jobsMap := lo.SliceToMap(params.jobsList, func(item *jobsdb.JobT) (int64, *jobsdb.JobT) {
		return item.JobID, item
	})
```

achettyiitr (approved):
Approving with some minor comments.

PR #5407, opened 2025-01-05
fix: warehouse router tracker

achettyiitr:
# Description

- In the previous implementation, we are sending the `warehouse_track_upload_missing` Guage metrics twice. First with value `0` and then if missing upload exists we send value with 1. 
- The way [alerts](https://github.com/rudderlabs/rudderstack-operator/blob/1002f52def9503687cce00057e25e25793c98512/operator-helm/alerts/single-customer-pipelines/warehouse/warehouse-successful-upload.yaml#L1) works is, if the value `1` is continuously fired for `2H` then the alert is being sent. But if the value comes as `0` it will get resolved. For calculating the value it uses whatever it being present at that instant.

## Linear Ticket

- Resolves WAR-114

## Security


warehouse/router/router.go
```diff
@@ -711,14 +712,32 @@ func (r *Router) loadReloadableConfig(whName string) {
 	r.config.enableJitterForSyncs = r.conf.GetReloadableBoolVar(false, "Warehouse.enableJitterForSyncs")
 	r.config.warehouseSyncFreqIgnore = r.conf.GetReloadableBoolVar(false, "Warehouse.warehouseSyncFreqIgnore")
 	r.config.cronTrackerRetries = r.conf.GetReloadableInt64Var(5, 1, "Warehouse.cronTrackerRetries")
+	r.config.uploadBufferTimeInMin = r.conf.GetReloadableDurationVar(180, time.Minute, "Warehouse.uploadBufferTimeInMin")
```
shekhar-rudder, line 715: Is `uploadBufferTimeInMin` solely used for generating metrics and unrelated to the functioning of syncs ? If yes, then the variable name should reflect that

warehouse/router/tracker.go
```diff
@@ -10,196 +10,170 @@ import (
...
+	exists, err := r.checkUploadStatus(ctx, warehouse, createdAt)
+	if err != nil {
+		return err
 	}
 
+	r.recordUploadMissingMetric(warehouse, exists)
+	return nil
+}
+
+func (r *Router) isWithinExcludeWindow(warehouse *model.Warehouse) bool {
```
shekhar-rudder, line 84: Would it be more appropriate to define these new methods, such as `getOldestStagingFile` or `isWithinExcludeWindow`, on a separate cron struct instead of adding them to the `Router`? These methods seem to be unrelated to the core responsibilities of the Router.

warehouse/router/tracker.go
```diff
@@ -10,196 +10,170 @@ import (
...
-			}, b)
-			if err != nil {
-				r.logger.Errorn(
-					"cron tracker failed for",
-					obskit.SourceID(warehouse.Source.ID),
-					obskit.DestinationID(warehouse.Destination.ID),
-					obskit.Error(err),
-				)
-				break
+		r.stats.cronTrackerExecTimestamp.Gauge(execTime.Unix())
```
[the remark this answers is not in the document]
achettyiitr: It should be `r.now()`. Missed updating it.

warehouse/router/router.go
achettyiitr: When you say generating metrics, what do you mean? Currently it's being used for getting latest staging file in `2x`-`3x` time window.

shekhar-rudder: By "generating metrics" I meant that it is being used only in `warehouse_track_upload_missing` computation flow of the cron tracker but not in job syncs.

> Currently it's being used for getting latest staging file in 2x-3x time window.

But only in cron tracker flow, rt ?

warehouse/router/tracker.go
```diff
@@ -10,196 +10,170 @@ import (
...
+
+func (r *Router) checkUploadStatus(ctx context.Context, warehouse *model.Warehouse, createdAt time.Time) (bool, error) {
+	query := `
+		SELECT EXISTS (
+			SELECT 1
+			FROM ` + whutils.WarehouseUploadsTable + `
+			WHERE source_id = $1 AND destination_id = $2 AND
+				  (status = $3 OR status = $4 OR status LIKE $5) AND
+				  updated_at > $6
+		);`
```
RanjeetMishra, line 147: Was just checking indexes for uploads table.

1. Do we keep deleting old entries from UploadsTable else without proper index on time columns this query can keep getting slower with time.
2. Not relevant to this PR. `CREATE INDEX wh_uploads_status_index ON wh_uploads(status text_ops);` index on status column with very low cardinality could have more performance drawbacks.

achettyiitr: > Do we keep deleting old entries from UploadsTable else without proper index on time columns this query can keep getting slower with time.

Yes. https://github.com/rudderlabs/rudder-server/blob/b51605af528cda22c1158f4daeae2675d92ee50a/warehouse/archive/archiver.go#L510

warehouse/router/router.go
achettyiitr: Do you mean something like `cronTrackerUploadBufferTimeInMin`?

warehouse/router/tracker.go
achettyiitr: > Not relevant to this PR. CREATE INDEX wh_uploads_status_index ON wh_uploads(status text_ops); index on status column with very low cardinality could have more performance drawbacks.

True, For columns with low cardinality, PostgreSQL might opt for a sequential scan over using the index, as scanning the table can be more efficient than traversing an index with low selectivity.

achettyiitr: Tracking Query-related things in [this](https://www.notion.so/rudderstacks/Warehouse-Query-Performance-175f2b415dd080d5bca6f78b1dd34c64?pvs=4) doc.

warehouse/router/router.go
shekhar-rudder: yes

achettyiitr: I am not sure that's a good idea. Since it's a generic configuration and not relevant to corn logic. It can be used in other places as well if required. For things like retries, it makes sense to name them as `cronTrackerRetries` but for uploadBufferTimeInMin, I am not sure.

shekhar-rudder (approved):

RanjeetMishra (approved):

PR #5422, opened 2025-01-10
feat: remove unnecessary fields from UT payload

cisse21:
# Description

This pull request aims to enhance the efficiency of the User Transformation (UT) payload by eliminating unnecessary fields, thereby improving performance by reducing payload size.

Changes Introduced:
Field Removal: Unused or redundant fields have been identified and removed from the UT payload structure.

Screenshot 2025-01-13 at 5 21 06 PM

In the image shown above the blue line represent the latency for a User Transformation with master and the yellow line represent the latency with the changes made in this PR. This change leads to 10-15% reduction in latency

## Linear Ticket

Fixes PIPE-1845

## Security

### DIFF
--- a/processor/transformer/transformer.go
+++ b/processor/transformer/transformer.go
@@ -101,10 +101,10 @@ type TransformerEvent struct {
 	Credentials []Credential               `json:"credentials"`
 }
 
-// Dehydrate removes the connection and credentials from the event
+// GetVersionsOnly removes the connection and credentials from the event
 // along with pruning the destination to only include the transformation versionID
 // before sending it to the transformer thereby reducing the payload size
-func (e *TransformerEvent) Dehydrate() *TransformerEvent {
+func (e *TransformerEvent) GetVersionsOnly() *TransformerEvent {
 	tmCopy := *e
 	transformations := make([]backendconfig.TransformationT, 0, len(e.Destination.Transformations))
 	for _, t := range e.Destination.Transformations {
@@ -290,7 +290,7 @@ func (trans *handle) Transform(ctx context.Context, clientEvents []TransformerEv
 func (trans *handle) UserTransform(ctx context.Context, clientEvents []TransformerEvent, batchSize int) Response {
 	var dehydratedClientEvents []TransformerEvent
 	for _, clientEvent := range clientEvents {
-		dehydratedClientEvent := clientEvent.Dehydrate()
+		dehydratedClientEvent := clientEvent.GetVersionsOnly()
 		dehydratedClientEvents = append(dehydratedClientEvents, *dehydratedClientEvent)
 	}
 
--- a/processor/transformer/transformer.go
+++ b/processor/transformer/transformer.go
@@ -101,6 +101,21 @@ type TransformerEvent struct {
 	Credentials []Credential               `json:"credentials"`
 }
 
+func (e *TransformerEvent) Dehydrate() *TransformerEvent {
+	tmCopy := *e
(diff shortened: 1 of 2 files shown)


processor/transformer/transformer.go
```diff
@@ -270,7 +285,13 @@ func (trans *handle) Transform(ctx context.Context, clientEvents []TransformerEv
 
 // UserTransform function is used to invoke user transformer API
 func (trans *handle) UserTransform(ctx context.Context, clientEvents []TransformerEvent, batchSize int) Response {
-	return trans.transform(ctx, clientEvents, trans.userTransformURL(), batchSize, userTransformerStage)
+	var dehydratedClientEvents []TransformerEvent
+	for _, clientEvent := range clientEvents {
+		dehydratedClientEvent := clientEvent.Dehydrate()
```
fracasula, line 290: What is a dehydrated client in this context? :thinking:

lvrach:
@cisse21 can you explain what's the screenshot in the description

processor/transformer/transformer.go
```diff
@@ -270,7 +285,13 @@ func (trans *handle) Transform(ctx context.Context, clientEvents []TransformerEv
 
 // UserTransform function is used to invoke user transformer API
 func (trans *handle) UserTransform(ctx context.Context, clientEvents []TransformerEvent, batchSize int) Response {
```
lvrach, line 287: I was kind hoping for an approach where we introduce a separate client for user transformation. 

As the requirements for dt and ut further diverge, keeping them under the same implementation only adds complexity.

cisse21: @lvrach doing that [here](https://github.com/rudderlabs/rudder-server/pull/5430)

wanted to get this out first in the next release as this would be a quick win. WDYT?

processor/transformer/transformer.go
cisse21: Added a little more in the description. Basically we send a common payload across user transformation , destination transformation and tracking plan which heavily inflates the payload size for each of them. The idea is to reduce the payload size by removing the unnecessary fields

fracasula: Whoever comes across this code will still have troubles understanding what `Dehydrate` means though, right? I don't think they'll go chasing the PR where you created this to get more context.

I suggest adding a comment here and also renaming the `Dehydrate` method to better reflect what it does:
1. PruneFields
2. StripUnnecessaryFields
3. GetVersionsOnly
4. ...

processor/transformer/transformer.go
```diff
@@ -101,6 +101,21 @@ type TransformerEvent struct {
 	Credentials []Credential               `json:"credentials"`
 }
 
+func (e *TransformerEvent) Dehydrate() *TransformerEvent {
```
lvrach, line 104:
```suggestion
func (e *TransformerEvent) Dehydrate() *HydradedTransformerEvent {
```

Instead of using the same struct, with some fields missing, we can introduce new structs with only the fields we want to include.

This will make it easier to understand, what we plan to send and what not. Plus making the object allocation a bit lighter

lvrach (approved):

processor/transformer/transformer.go
cisse21: This would make it a bit difficult downstream as the `transform` function expected `TransformerEvent`. Was planning to do the same in the refactor to reduce the fields further

fracasula (approved):

PR #5426, opened 2025-01-13
refactor: replace schema struct with interface to enable flexible implementation

shekhar-rudder:
# Description

Refactoring

- Replaced the Schema struct with the SchemaHandler interface, ensuring that clients depend on the interface rather than the concrete implementation. This change will enable switching to a different schema implementation via a feature flag.
- `Schema` -> `schema`

## Linear Ticket

resolves WAR-150

## Security

### DIFF
--- a/warehouse/router/upload_test.go
+++ b/warehouse/router/upload_test.go
@@ -22,7 +22,6 @@ import (
 	sqlmiddleware "github.com/rudderlabs/rudder-server/warehouse/integrations/middleware/sqlquerywrapper"
 	"github.com/rudderlabs/rudder-server/warehouse/integrations/redshift"
 	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
-	"github.com/rudderlabs/rudder-server/warehouse/schema"
 	warehouseutils "github.com/rudderlabs/rudder-server/warehouse/utils"
 )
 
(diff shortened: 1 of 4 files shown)


warehouse/schema/schema.go
```diff
@@ -45,7 +45,21 @@ type fetchSchemaRepo interface {
 	FetchSchema(ctx context.Context) (model.Schema, error)
 }
 
-type Schema struct {
+type SchemaHandler interface {
```
achettyiitr, line 48: We can avoid the Schema prefix here since we are already in the schema package.

`The importer of a package will use the name to refer to its contents, so exported names in the package can use that fact to avoid repetition.` https://go.dev/doc/effective_go#names
```suggestion
type Handler interface {
```

warehouse/router/upload_test.go
```diff
@@ -127,17 +126,30 @@ func TestColumnCountStat(t *testing.T) {
...
+			pgResource, err := postgres.Setup(pool, t)
+			require.NoError(t, err)
+
+			uploadJobFactory := &UploadJobFactory{
+				logger:       logger.NOP,
+				statsFactory: statsStore,
+				conf:         conf,
+				db:           sqlmiddleware.New(pgResource.DB),
+			}
+			rs := redshift.New(config.New(), logger.NOP, stats.NOP)
```
achettyiitr, line 145: Can we use a manager in here?
```suggestion
			whManager, err := manager.New(warehouseutils.POSTGRES, conf, logger.NOP, statsStore)
			require.NoError(t, err)
```

achettyiitr (approved):

warehouse/schema/schema.go
shekhar-rudder: Renamed

warehouse/router/upload_test.go
shekhar-rudder: Replaced redshift with manager. I had used redshift because that is how other [tests](https://github.com/rudderlabs/rudder-server/pull/5426/files#diff-3ca4ffe249a74c6545b23670aa07940de31b9f23362020de94830c8f10c7aafeR258) were written

achettyiitr: Yeah, we would need to fix other tests as well.

warehouse/schema/schema.go
```diff
@@ -45,7 +45,21 @@ type fetchSchemaRepo interface {
...
 
-type Schema struct {
+type Handler interface {
+	SyncRemoteSchema(ctx context.Context, fetchSchemaRepo fetchSchemaRepo, uploadID int64) (bool, error)
+	IsWarehouseSchemaEmpty() bool
+	GetTableSchemaInWarehouse(tableName string) model.TableSchema
+	GetLocalSchema(ctx context.Context) (model.Schema, error)
+	UpdateLocalSchema(ctx context.Context, updatedSchema model.Schema) error
+	UpdateWarehouseTableSchema(tableName string, tableSchema model.TableSchema)
+	GetColumnsCountInWarehouseSchema(tableName string) int
```
RanjeetMishra, line 55: @shekhar-rudder 
`GetColumnsCountInWarehouseSchema(tableName string)` Full identifier would be schema.table as same tablename can exist in multiple schemas as well. I did not check this function, can you help share if we are using schema as well when using this func?

RanjeetMishra (approved):

warehouse/schema/schema.go
shekhar-rudder: We don't explicitly use schema in this func. But the warehouse schema which gets fetched at the start is specific to a namespace.
https://github.com/rudderlabs/rudder-server/blob/3c7106ee0e380d49fb46a9a0bda4ecbbcf7054c4/warehouse/integrations/redshift/redshift.go#L1270

https://github.com/rudderlabs/rudder-server/blob/3c7106ee0e380d49fb46a9a0bda4ecbbcf7054c4/warehouse/integrations/redshift/redshift.go#L1223

PR #5437, opened 2025-01-15
chore: transformer 404 error logs

vinayteki95:
# Description

This commit aims to allow logging of 404 errors in the processor while communicating with transformer to trigger log based alerts (loki)

Transformer can respond with 404 errors if the destination is not available in the transformer image but the config-backend has the destination.
This can only happend when some outdated transformer image is used in the deployment. In order to catch such errors and immediately alert (throw loki-log based triggers) the team to make a fix, we aim to log 404 errors and add alerts for those errors.
The below ticket is attached with corresponding alert PR in devops as well.

## Linear Ticket

Resolves INT-3031
https://linear.app/rudderstack/issue/INT-3031/revisit-transformer-and-rudder-server-api-contracts-and-throw-alerts

## Security

### DIFF
--- a/processor/transformer/transformer.go
+++ b/processor/transformer/transformer.go
@@ -480,7 +480,6 @@ func (trans *handle) request(ctx context.Context, url, stage string, data []Tran
 	switch statusCode {
 	case http.StatusOK,
 		http.StatusBadRequest,
-		http.StatusNotFound,
 		http.StatusRequestEntityTooLarge:
 	default:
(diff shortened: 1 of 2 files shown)


utsabc:
Can you add some more context in the description?

vinayteki95:
@utsabc Updated the description with more details. Good catch, this PR is very vague without the context.

utsabc (approved):

lvrach (approved):

PR #5445, opened 2025-01-20
feat: bq load using folder

achettyiitr:
# Description

- BQ load using folder pattern instead of providing all file locations.
- Added loadByFolderPath configuration, by default it's false.

## Linear Ticket

- Resolves WAR-179, WAR-216

## Security

### DIFF
--- a/warehouse/integrations/bigquery/bigquery.go
+++ b/warehouse/integrations/bigquery/bigquery.go
@@ -49,6 +49,7 @@ type BigQuery struct {
 		enableDeleteByJobs                    bool
 		customPartitionsEnabledWorkspaceIDs   []string
 		slowQueryThreshold                    time.Duration
+		loadByFolderPath                      bool
 	}
 }
 
@@ -130,6 +131,7 @@ func New(conf *config.Config, log logger.Logger) *BigQuery {
 	bq.config.enableDeleteByJobs = conf.GetBool("Warehouse.bigquery.enableDeleteByJobs", false)
 	bq.config.customPartitionsEnabledWorkspaceIDs = conf.GetStringSlice("Warehouse.bigquery.customPartitionsEnabledWorkspaceIDs", nil)
 	bq.config.slowQueryThreshold = conf.GetDuration("Warehouse.bigquery.slowQueryThreshold", 5, time.Minute)
+	bq.config.loadByFolderPath = conf.GetBool("Warehouse.bigquery.loadByFolderPath", false)
 
 	return bq
 }
@@ -441,12 +443,12 @@ func (bq *BigQuery) loadTable(ctx context.Context, tableName string) (
 	)
 	log.Infon("started loading")
 
-	loadFileLocation, err := bq.loadFileLocation(ctx, tableName)
+	gcsReferences, err := bq.gcsReferences(ctx, tableName)
 	if err != nil {
-		return nil, nil, fmt.Errorf("getting load file location: %w", err)
+		return nil, nil, fmt.Errorf("getting gcs references: %w", err)
 	}
 
-	gcsRef := bigquery.NewGCSReference(loadFileLocation)
+	gcsRef := bigquery.NewGCSReference(gcsReferences...)
 	gcsRef.SourceFormat = bigquery.JSON
 	gcsRef.MaxBadRecords = 0
 	gcsRef.IgnoreUnknownValues = false
@@ -454,10 +456,10 @@ func (bq *BigQuery) loadTable(ctx context.Context, tableName string) (
 	return bq.loadTableByAppend(ctx, tableName, gcsRef, log)
 }
 
-func (bq *BigQuery) loadFileLocation(
+func (bq *BigQuery) gcsReferences(
 	ctx context.Context,
 	tableName string,
-) (string, error) {
+) ([]string, error) {
 	switch tableName {
 	case warehouseutils.IdentityMappingsTable, warehouseutils.IdentityMergeRulesTable:
 		loadfile, err := bq.uploader.GetSingleLoadFile(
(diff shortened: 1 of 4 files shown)


warehouse/integrations/bigquery/bigquery.go
```diff
@@ -443,44 +441,19 @@ func (bq *BigQuery) loadTable(ctx context.Context, tableName string) (
 	)
 	log.Infon("started loading")
 
-	loadFileLocations, err := bq.loadFileLocations(ctx, tableName)
+	objectLocation, err := bq.uploader.GetSampleLoadFileLocation(ctx, tableName)
```
shekhar-rudder, line 444: noob question - does a folder only contain files or can it have subfolders as well ?

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -443,44 +441,19 @@ func (bq *BigQuery) loadTable(ctx context.Context, tableName string) (
...
 	if err != nil {
-		return nil, nil, fmt.Errorf("getting load file locations: %w", err)
+		return nil, nil, fmt.Errorf("getting sample load file location: %w", err)
 	}
 
-	gcsRef := bigquery.NewGCSReference(warehouseutils.GetGCSLocations(
-		loadFileLocations,
-		warehouseutils.GCSLocationOptions{},
-	)...)
+	gcsRef := bigquery.NewGCSReference(loadFolder(objectLocation))
```
shekhar-rudder, line 449: Will this change also impact performance?

warehouse/integrations/bigquery/bigquery_test.go
```diff
@@ -1095,6 +1097,71 @@ func TestIntegration(t *testing.T) {
...
 		})
+		t.Run("multiple files", func(t *testing.T) {
+			tableName := "multiple_files_test_table"
+			repeat := 10
+			loadObjectFolder := "rudder-warehouse-load-objects"
+			sourceID := "test_source_id"
+
+			prefixes := []string{loadObjectFolder, tableName, sourceID, uuid.New().String() + "-" + tableName}
+
+			loadFiles := lo.RepeatBy(repeat, func(int) whutils.LoadFile {
```
shekhar-rudder, line 1108: Is it worth creating load files concurrently ?

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -692,15 +665,12 @@ func (bq *BigQuery) LoadUserTables(ctx context.Context) (errorMap map[string]err
 }
 
 func (bq *BigQuery) createAndLoadStagingUsersTable(ctx context.Context, stagingTable string) error {
-	loadFileLocations, err := bq.loadFileLocations(ctx, warehouseutils.UsersTable)
+	objectLocation, err := bq.uploader.GetSampleLoadFileLocation(ctx, warehouseutils.UsersTable)
```
RanjeetMishra, line 668: @achettyiitr 
1. Can you help with sample value for objectLocation?
2. Can ignore this, but sharing thought: This would limit us to switch bucket, storage or region like US/EU. Otherwise we would read details for each files and can group by dir path. Anyway this limitation we would have at multiple places in our system.

warehouse/integrations/bigquery/bigquery.go
RanjeetMishra: @achettyiitr Also would dir path contain files only related to this load job? Can you share example of full paths?

achettyiitr: https://www.notion.so/rudderstacks/Slaves-5fed5a5522984924879fef97055f582b?pvs=4#9c3079b24afd4707b1859c2c6b797306

Everytime time load generation happens it will generate a new `<uuid>-<table>` folder, within which all files will present. This folder will contain only the files, it will not contain any folder within it.
`<prefix>/rudder-warehouse-load-objects/<table>/<source_id>/<uuid>-<table>/<timestamp>.<source_id>.<uuid>.<table>.<uuid>.<extension>`

warehouse/integrations/bigquery/bigquery.go
achettyiitr: It shouldn't impact performance.

warehouse/integrations/bigquery/bigquery_test.go
achettyiitr: No. It's just tests. Also, it's just 10 files which should happen very quick.

warehouse/integrations/bigquery/bigquery.go
achettyiitr: > Can you help with sample value for objectLocation?

`<prefix>/rudder-warehouse-load-objects/<table>/<source_id>/<uuid>-<table>/<timestamp>.<source_id>.<uuid>.<table>.<uuid>.<extension>`


> Can ignore this, but sharing thought: This would limit us to switch bucket or region like US/EU. Otherwise we would read details for each files and can group by dir path. Anyway this limitation we would have at multiple places in our system.

No, whenever the load generation happens, all files for all tables will have unique uuid associated with it. Basicallly `<uuid>-<table>`. Staging files can be in different buckets based on when the customer changes the configuration, but during load file generation, Upload Job will have one static config to process, so all the load files will be uploaded with one object storage configuration.

RanjeetMishra (approved):

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -443,41 +441,41 @@ func (bq *BigQuery) loadTable(ctx context.Context, tableName string) (
...
 		loadfile, err := bq.uploader.GetSingleLoadFile(
 			ctx,
 			tableName,
 		)
 		if err != nil {
-			return nil, fmt.Errorf("getting single load file for table %s: %w", tableName, err)
+			return "", fmt.Errorf("getting single load file for table %s: %w", tableName, err)
 		}
-		return []warehouseutils.LoadFile{loadfile}, nil
+		return loadfile.Location, nil
```
shekhar-rudder, line 470: It looks to me that in case of `IdentityMappingsTable` we are returning the file location and the folder location in `default` case. Isn't this inconsistent ?

achettyiitr: Yeah, ID Resolution is some legacy code where load files are not generated on slave instead of master and it does not follow a proper naming convention.

achettyiitr: Also, ID Resolution is only for BQ and Snowflake.

shekhar-rudder (approved):

warehouse/integrations/bigquery/bigquery_test.go
```diff
@@ -1095,6 +1098,85 @@ func TestIntegration(t *testing.T) {
...
+						defer func() { _ = f.Close() }()
+
+						uploadOutput, err := fm.Upload(context.Background(), f, prefixes...)
+						require.NoError(t, err)
+						return whutils.LoadFile{Location: uploadOutput.Location}
+					})
+					mockUploader := newMockUploader(t, loadFiles, tableName, schemaInUpload, schemaInWarehouse)
+
+					c := config.New()
+					c.Set("Warehouse.redshift.loadByFolderPath", tc.loadByFolderPath)
```
shekhar-rudder, line 1141: Warehouse.redshift -> Warehouse.bigquery

Can we add assertions to verify that the correct values or functions are being used in both cases? For instance, the current test checks only the "false" version in both scenarios.

achettyiitr: Yes, added an assertion based on whether it's true or false.

shekhar-rudder (approved):

PR #5449, opened 2025-01-21
chore: generateTransformationMessage is its own concurrent step

Sidddddarth:
# Description

generateTransformationMessage is its own concurrent step.
Includes:
1. store to arc, esch
2. tracking plan validation

## Linear Ticket

[Resolves PIPE-1719](https://linear.app/rudderstack/issue/PIPE-1719/introduce-processor-stages)

## Security


processor/processor.go
```diff
@@ -3283,11 +3282,15 @@ func (proc *Handle) handlePendingGatewayJobs(partition string) bool {
 
 	var transMessage *transformationMessage
 	var err error
-	transMessage, err = proc.processJobsForDest(partition, subJob{
-		subJobs:       unprocessedList.Jobs,
-		hasMore:       false,
-		rsourcesStats: rsourcesStats,
-	})
+	transMessage, err = proc.generateTransformationMessage(proc.processJobsForDest(
```
ktgowtham, line 3285: not related to this PR. but there's a possibility of duplicate persistence of jobs in events schema and archival?

ktgowtham (approved):
Changes look good to me.

processor/processor.go
```diff
@@ -1734,7 +1734,7 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) (*trans
 	for _, batchEvent := range jobList {
 		var eventParams types.EventParams
 		if err := jsonfast.Unmarshal(batchEvent.Parameters, &eventParams); err != nil {
-			return nil, err
+			panic(err)
```
mihir20, line 1737: lets return error from here, we can handle error at 1 place only in `worker.go`

mihir20 (changes requested):

processor/processor.go
Sidddddarth: It's possible yeah

mihir20 (commented):
should we keep this change behind a flag?

Sidddddarth:
@mihir20 
> should we keep this change behind a flag?

It might lead to some complexity in that it'll involve extra handling to choose the channel to put things in based on the flag. And eventually some more at shutdown too. Imo we can go ahead without a flag because the only concern could be memory usage - but that's blocked at the `Store` step in processor anyway.

mihir20 (approved):
LGTM

PR #5451, opened 2025-01-22
chore: make cslb configurable

cisse21:
# Description

- Add configurability options to CSLB Picker
1. 	NewPowerOfTwo
2. 	NewRoundRobin
3. 	NewLeastLoadedRandom
4. 	NewLeastLoadedRoundRobin

https://www.notion.so/rudderstacks/User-Transformer-with-Client-Side-Load-Balancing-190f2b415dd080709a94d3f8ea34ebcd

## Linear Ticket
Fixes PIPE-1860

## Security


ktgowtham (approved):

processor/transformer/transformer.go
```diff
@@ -310,6 +315,39 @@ func (t *HTTPLBTransport) NewRoundTripper(scheme, target string, config httplb.T
...
+
+func (trans *handle) getChecker() health.Checker {
+	checkerType := config.GetString("Transformer.Client.httplb.checkerType", "nop")
+	switch checkerType {
+	case "nop":
+		return health.NopChecker
+	case "polling":
+		return health.NewPollingChecker(
+			health.PollingCheckerConfig{},
+			health.NewSimpleProber(trans.config.userTransformationURL),
```
Sidddddarth, line 344: wouldn't using `trans.config.userTransformationURL` adversely affect destination transformation considering the same client is used..?

Sidddddarth: I guess there's no adverse effects, but we're not probing the right endpoint

processor/transformer/transformer.go
```diff
@@ -310,6 +315,39 @@ func (t *HTTPLBTransport) NewRoundTripper(scheme, target string, config httplb.T
 	return httplb.RoundTripperResult{RoundTripper: t.Transport, Close: t.CloseIdleConnections}
 }
 
+func (trans *handle) getPicker() func(prev picker.Picker, allConns conn.Conns) picker.Picker {
```
lvrach, line 318: we are adding method that contain logic specific to httplb, to a generic component.

It would be preferable to encapsulate this in a separate component. A simple wrapper on httplb.NewClient would be enough.

Something like

```go
HTTPLBFromConfig(c *config.Config) httplb.Client {
pickerType := config.GetString("Transformer.Client.httplb.pickerType", "power_of_two")
...

return httplb.NewClient(....
}
```

cisse21: Done... is this behaviour what you expected?

processor/transformer/transformer.go
cisse21: Will be fixing this in the new separate clients PR

Sidddddarth (approved):

ktgowtham (approved):

Sidddddarth (approved):

PR #5428, opened 2025-01-13
feat: delete staging and load files post successful sync

shekhar-rudder:
# Description

This PR introduces support for cleaning up staging and load files from object storage based on the user preference (cleanupObjectStorageFiles). If file deletion fails, the job will be marked as failed. The cleanup process occurs just before marking the job as complete.

**Implementation Details**
The `UploadJob` struct already includes `stagingFiles`. Using their ids, the corresponding load files are retrieved. Keys are then computed from the locations of both file types, and the files are deleted.

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #5456, opened 2025-01-27
feat: implement object storage delete validations for warehouse destinations

shekhar-rudder:
# Description

With #5428 being merged, this update enhances the validations API for creating/updating destinations by adding a step to verify if RudderStack has object storage delete permissions.

**Implementation Details**
- This new step is included only if the deletion config is set to true.
- It will be the final part of checking the "Object storage" permissions
- Attempt to delete the uploaded file. If the deletion step fails, an error message will be propagated to the web app, informing the user of the issue.

## Security

### DIFF
--- a/warehouse/internal/model/validation.go
+++ b/warehouse/internal/model/validation.go
@@ -4,6 +4,7 @@ import backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 
 const (
 	VerifyingObjectStorage       = "Verifying Object Storage"
+	VerifyingObjectStorageDelete = "Verifying delete permissions in object storage"
 	VerifyingConnections         = "Verifying Connections"
 	VerifyingCreateSchema        = "Verifying Create Schema"
 	VerifyingCreateAndAlterTable = "Verifying Create and Alter Table"
--- a/warehouse/internal/model/validation.go
+++ b/warehouse/internal/model/validation.go
@@ -4,7 +4,6 @@ import backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 
 const (
 	VerifyingObjectStorage       = "Verifying Object Storage"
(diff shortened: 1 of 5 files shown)


warehouse/internal/model/validation.go
```diff
@@ -4,6 +4,7 @@ import backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 
 const (
 	VerifyingObjectStorage       = "Verifying Object Storage"
+	VerifyingObjectStorageDelete = "Verifying delete permissions in object storage"
```
achettyiitr, line 7: Shall it be part of existing step, `Verifying Object Storage`. Current behaviour:
1. Upload file.
2. Download file.

New behaviour: Based on whether the CleanupObjectStorageFilesSetting config is enabled:
1. Upload file.
2. Download file.
3. Delete file.

shekhar-rudder: done

warehouse/validations/validate_test.go
```diff
@@ -50,25 +50,43 @@ func TestValidator(t *testing.T) {
...
-					"secretAccessKey": minioResource.AccessKeySecret,
-					"endPoint":        minioResource.Endpoint,
+				{
+					name:                      "check delete permissions",
+					cleanupObjectStorageFiles: false,
 				},
-			})
-			require.NoError(t, err)
-			require.NoError(t, v.Validate(ctx))
+			}
```
achettyiitr, line 65: Can we have different names for individual tests. Also keep it within `t.Run`.

achettyiitr (approved):

warehouse/validations/validate_test.go
shekhar-rudder: done

RanjeetMishra (approved):

PR #5459, opened 2025-01-29
fix: event sampling order so that error reporter can extract data

koladilip:
# Description
Handle error reporting first and then metrics so that we can extract error details  before it is going to get sampled by metrics reporter

## Linear Ticket

https://linear.app/rudderstack/issue/OBS-763/fix-report-modification-in-reporting-during-event-sampling-rudder
Resolves OBS-763

## Security


koladilip:
closing this for https://github.com/rudderlabs/rudder-server/pull/5454

PR #5462, opened 2025-01-29
feat: add a new implementation of schema with ttl

shekhar-rudder:
# Description

This PR introduces schema invalidation based on a configurable TTL. When a schema entry expires, it is refreshed from the warehouse and updated in storage.

Changes:
- Added an `expires_at` column to the `wh_schemas` table.
- Implemented a new schema version (`schema_v2`), gated by a feature flag.
- In schema_v2:
    1. Entries in the table are marked with an expiration timestamp (now + TTL) upon insert/update.
     2. Reads first validate the schema’s expiration status—if valid, it is returned; otherwise, it is re-fetched from the warehouse.
- Updated database read/write methods to handle the new column.
- Modified the `NewSchema` function to accept `whManager`, allowing schema_v2 to reference the repository.
- Adjusted method signatures in the `Handler` interface to accept a context and return an error, as most schema_v2 methods now interact with the database.
- Extracted `removeDeprecatedColumns` and `tableSchemaDiff` from schemaV1 for reuse in schemaV2.

## Security


warehouse/internal/repo/schema.go
```diff
@@ -130,13 +133,15 @@ func parseWHSchemas(rows *sqlmiddleware.Rows) ([]*model.WHSchema, error) {
 			&schemaPayloadRawRaw,
 			&whSchema.CreatedAt,
 			&whSchema.UpdatedAt,
+			&whSchema.ExpiresAt,
```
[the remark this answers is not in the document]
shekhar-rudder: Updated

warehouse/router/state_export_data.go
```diff
@@ -795,7 +807,11 @@ func (job *UploadJob) columnCountStat(tableName string) {
 	tags := []whutils.Tag{
 		{Name: "tableName", Value: whutils.TableNameForStats(tableName)},
 	}
-	currentColumnsCount := job.schemaHandle.GetColumnsCountInWarehouseSchema(tableName)
+	currentColumnsCount, err := job.schemaHandle.GetColumnsCountInWarehouseSchema(job.ctx, tableName)
+	if err != nil {
+		job.logger.Errorf("error getting column count for table %s: %v", tableName, err)
```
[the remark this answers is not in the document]
shekhar-rudder: Updated

warehouse/schema/schema_v2.go
```diff
@@ -0,0 +1,193 @@
...
+		sh.log.Errorw("error getting schema: %w", err)
+		return true
+	}
+	return len(schema) == 0
+}
+
+func (sh *schemaV2) GetTableSchemaInWarehouse(ctx context.Context, tableName string) model.TableSchema {
+	schema, err := sh.getSchema(ctx)
+	if err != nil {
+		sh.log.Errorw("error getting schema: %w", err)
```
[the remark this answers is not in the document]
shekhar-rudder: Updated

warehouse/schema/schema_v2.go
```diff
@@ -0,0 +1,193 @@
...
+	warehouse                        model.Warehouse
+	log                              logger.Logger
+	ttlInMinutes                     time.Duration
+	schemaRepo                       schemaRepo
+	stagingFilesSchemaPaginationSize int
+	stagingFileRepo                  stagingFileRepo
+	enableIDResolution               bool
+	fetchSchemaRepo                  fetchSchemaRepo
+	now                              func() time.Time
+}
```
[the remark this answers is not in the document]
shekhar-rudder: Added 2 new fields in the struct `cachedSchema` and `cacheExpiry` to avoid DB calls

vyeshwanth (approved):

warehouse/schema/schema_v2.go
```diff
@@ -0,0 +1,216 @@
...
+		schemaRepo:                       v1.schemaRepo,
+		stagingFilesSchemaPaginationSize: v1.stagingFilesSchemaPaginationSize,
+		stagingFileRepo:                  v1.stagingFileRepo,
+		fetchSchemaRepo:                  fetchSchemaRepo,
+		enableIDResolution:               v1.enableIDResolution,
+		now:                              timeutil.Now,
+	}
+	v2.stats.schemaSize = schemaSize
+	return v2
+}
```
achettyiitr, line 52: Since we are already passing v1, can we directly use that instead of passing schemaSize?
```suggestion
func newSchemaV2(v1 *schema, warehouse model.Warehouse, log logger.Logger, ttlInMinutes time.Duration, fetchSchemaRepo fetchSchemaRepo) *schemaV2 {
	v2 := &schemaV2{
		warehouse:                        warehouse,
		log:                              log,
		ttlInMinutes:                     ttlInMinutes,
		schemaRepo:                       v1.schemaRepo,
		stagingFilesSchemaPaginationSize: v1.stagingFilesSchemaPaginationSize,
		stagingFileRepo:                  v1.stagingFileRepo,
		fetchSchemaRepo:                  fetchSchemaRepo,
		enableIDResolution:               v1.enableIDResolution,
		now:                              timeutil.Now,
	}
	v2.stats.schemaSize = v1.stats.schemaSize
	return v2
}
```

warehouse/schema/schema_v2.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (sh *schemaV2) SyncRemoteSchema(ctx context.Context, _ fetchSchemaRepo, uploadID int64) (bool, error) {
+	return false, sh.FetchSchemaFromWarehouse(ctx, sh.fetchSchemaRepo)
+}
+
+func (sh *schemaV2) IsWarehouseSchemaEmpty(ctx context.Context) bool {
+	schema, err := sh.getSchema(ctx)
+	if err != nil {
+		sh.log.Warnf("error getting schema: %v", err)
```
achettyiitr, line 61: Can we use structured logging in here?
```suggestion
		sh.log.Warnn("Error getting schema", obskit.Error(err))
```

achettyiitr: Similar for other places as well.

warehouse/schema/schema_v2.go
achettyiitr, line 1: Don't see any comments in the exported functions. Can we add those?

shekhar-rudder: `schema_v2` is implementing the [Handler](https://github.com/rudderlabs/rudder-server/blob/e0b64f815557061fa009d4badee6cde70b18494c/warehouse/schema/schema.go#L48) interface. And this interface will change once we get rid of the v1 implementation.

warehouse/schema/schema_v2.go
shekhar-rudder: Done

warehouse/schema/schema_v2.go
shekhar-rudder: Done

warehouse/schema/schema_v2.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	if whSchema.Schema == nil {
+		sh.cachedSchema = model.Schema{}
+		sh.cacheExpiry = sh.now().Add(sh.ttlInMinutes)
+		return sh.cachedSchema, nil
+	}
+	if whSchema.ExpiresAt.Before(sh.now()) {
+		sh.log.Infon("Schema expired", obskit.DestinationID(sh.warehouse.Destination.ID), obskit.Namespace(sh.warehouse.Namespace), logger.NewTimeField("expiresAt", whSchema.ExpiresAt))
+		sh.cachedSchemaMu.RUnlock() // Release the lock since "fetchSchemaFromWarehouse" internally acquires a lock
+		schema, err := sh.fetchSchemaFromWarehouse(ctx)
```
vyeshwanth, line 212: For a brief moment, there won't be any lock on the mutex. considering it is a read lock, Even if another reader locked it again, we would release it using `RUnlock`,  so writes in fetch will not be locked as there are no readers. But there is a chance there might be another reader acquiring read lock and there will be two `fetchSchemaFromWarehouse` calls?. Let me know if my understanding is right? is this even possible like two readers attempting to getSchema() concurrently?

shekhar-rudder: Good catch. Yes it is possible for two readers to call getSchema() concurrently. I've updated the code to apply the mutex selectively as needed.

vyeshwanth: Offline discussion: we discussed the case of multiple readers acquiring the read locks and attempting to fetch schema from warehouse. we might have to acquire write lock early once the cache is missed or even at the start of the `getSchema` directly

shekhar-rudder: - Currently there is no scenario where the schemaV2 will be used concurrently. So it is theoretically fine to have multiple concurrent `fetchSchemaFromWarehouse` calls. The only way to prevent this would be to inline the function in `getSchema`

vyeshwanth (approved):

achettyiitr (approved):

PR #5467, opened 2025-02-03
fix: deltalake syncs failing for columns with unhandled data type

shekhar-rudder:
# Description

**Issue**

Deltalake syncs were failing when processing events containing a field of type `DECIMAL(38,0)`, which was not handled by the warehouse code.
- For unhandled data types, the column is skipped in the local table schema. So when a mismatch was detected between the local table schema and event schema, warehouse attempted to add the missing columns in the table. The column addition failed because the column already existed.
- This is a regression introduced from changes in #5322 
- This issue doesn't affect other warehouses since their `AddColumns` method handles already existing columns.

**Fix**

To resolve this, we now retrieve all existing columns of the table before adding new ones. Columns that are already present are skipped, preventing the failure.

## Security

### DIFF
--- a/warehouse/integrations/deltalake/deltalake.go
+++ b/warehouse/integrations/deltalake/deltalake.go
@@ -531,7 +531,7 @@ func (d *Deltalake) AddColumns(ctx context.Context, tableName string, columnsInf
 	if err != nil {
 		return fmt.Errorf("fetch table attributes: %w", err)
 	}
-	columnsInfo = lo.Filter(columnsInfo, func(columnInfo warehouseutils.ColumnInfo, _ int) bool {
+	columnsToAddInfo := lo.Filter(columnsInfo, func(columnInfo warehouseutils.ColumnInfo, _ int) bool {
 		_, ok := tableSchema[columnInfo.Name]
 		return !ok
 	})
@@ -546,7 +546,7 @@ func (d *Deltalake) AddColumns(ctx context.Context, tableName string, columnsInf
 		tableName,
 	))
 
-	for _, columnInfo := range columnsInfo {
+	for _, columnInfo := range columnsToAddInfo {
 		queryBuilder.WriteString(fmt.Sprintf(` %s %s,`, columnInfo.Name, dataTypesMap[columnInfo.Type]))
 	}
 
--- a/warehouse/integrations/deltalake/deltalake.go
+++ b/warehouse/integrations/deltalake/deltalake.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/samber/lo"
+
 	"github.com/rudderlabs/sqlconnect-go/sqlconnect"
 	sqlconnectconfig "github.com/rudderlabs/sqlconnect-go/sqlconnect/config"
 
@@ -35,8 +37,7 @@ const (
 )
 
 const (
-	schemaNotFound       = "[SCHEMA_NOT_FOUND]"
-	columnsAlreadyExists = "already exists in root"
+	schemaNotFound = "[SCHEMA_NOT_FOUND]"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


warehouse/integrations/deltalake/deltalake.go
```diff
@@ -536,17 +548,17 @@ func (d *Deltalake) AddColumns(ctx context.Context, tableName string, columnsInf
...
 
 	query := strings.TrimSuffix(queryBuilder.String(), ",")
 	query += ");"
 
-	_, err := d.DB.ExecContext(ctx, query)
+	_, err = d.DB.ExecContext(ctx, query)
 
 	// Handle error in case of single column
-	if len(columnsInfo) == 1 {
+	if len(updatedColumnsInfo) == 1 {
```
achettyiitr, line 561: We can get rid of this now.

warehouse/integrations/deltalake/deltalake_test.go
```diff
@@ -1319,6 +1319,75 @@ func TestIntegration(t *testing.T) {
...
+		require.NoError(t, err)
+
+		schema, err := d.FetchSchema(ctx)
+		require.NoError(t, err)
+		require.Contains(t, schema, tableName)
+		require.Equal(t, model.TableSchema(map[string]string{
+			"col1": "int",
+			"col2": "string",
+			"col3": "int",
+		}), schema[tableName])
```
achettyiitr, line 1389: Redundant type conversion 
```suggestion
		require.Equal(t, model.TableSchema{
			"col1": "int",
			"col2": "string",
			"col3": "int",
		}, schema[tableName])
```

warehouse/integrations/deltalake/deltalake.go
achettyiitr: Also, columnsAlreadyExists.

warehouse/integrations/deltalake/deltalake_test.go
shekhar-rudder: removed

warehouse/integrations/deltalake/deltalake.go
```diff
@@ -526,6 +526,18 @@ func (d *Deltalake) tableLocationQuery(tableName string) string {
...
+	if err != nil {
+		return fmt.Errorf("fetch table attributes: %w", err)
+	}
+	var updatedColumnsInfo []warehouseutils.ColumnInfo
+	for _, columnInfo := range columnsInfo {
+		_, ok := tableSchema[columnInfo.Name]
+		if !ok {
+			updatedColumnsInfo = append(updatedColumnsInfo, columnInfo)
+		}
+	}
```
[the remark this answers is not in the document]
shekhar-rudder: updated

warehouse/integrations/deltalake/deltalake.go
shekhar-rudder: removed

achettyiitr (approved):

warehouse/integrations/deltalake/deltalake.go
```diff
@@ -526,6 +527,15 @@ func (d *Deltalake) tableLocationQuery(tableName string) string {
 
 // AddColumns adds columns to the table.
 func (d *Deltalake) AddColumns(ctx context.Context, tableName string, columnsInfo []warehouseutils.ColumnInfo) error {
+	tableSchema, err := d.fetchTableAttributes(ctx, tableName)
+	if err != nil {
+		return fmt.Errorf("fetch table attributes: %w", err)
+	}
+	columnsInfo = lo.Filter(columnsInfo, func(columnInfo warehouseutils.ColumnInfo, _ int) bool {
```
RanjeetMishra, line 534: 1. minor: `columnsInfo` essentially are missing columns from tableschema fetched from warehouse? If yes, would `missingColumnsInfo` or `toAddColumnsInfo` be helpful var name?

2. How would idea of fetching table schema at higher cadence solution we are working on impact this change? Or is it in such cases of discrepancy between event schema and local table schema, we will always fetch table schema from warehouse?

RanjeetMishra (approved):

warehouse/integrations/deltalake/deltalake.go
shekhar-rudder: 1. Renamed output of filtering to `columnsToAddInfo`
2. If there is a discrepancy between event schema and local table schema, we currently don't fetch the table schema from the warehouse. The diff is computed and an ALTER statement is executed for missing tables. So I would say schema v2 implementation is independent of this change

RanjeetMishra (approved):

PR #5472, opened 2025-02-04
feat: implement snowpipe destination config validation

shekhar-rudder:
# Description

[Context](https://linear.app/rudderstack/issue/WAR-240/short-term-solution-for-validation-errors-in-sync-errors-for-snowpipe#comment-150d9d94)

Implementation
- Made changes in `StepsToValidate` method to return relevant steps for Snowpipe destination
- Moved `DestinationValidationResponse` struct from `internal/model` package to `validations` package so that it can be accessed in `snowpipestreaming` package
- Added a `validator` field in snowpipe's `Manager` struct which is responsible for validating the destination config
- Made changes in `Upload` function to call the validator whenever there is an authorization error. And if the response is not successful, the error is being returned.

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
@@ -35,6 +35,7 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/timeutil"
 	"github.com/rudderlabs/rudder-server/warehouse/integrations/manager"
 	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+	"github.com/rudderlabs/rudder-server/warehouse/validations"
 )
 
 var json = jsoniter.ConfigCompatibleWithStandardLibrary
@@ -57,6 +58,7 @@ func New(
 		now:                 timeutil.Now,
 		channelCache:        sync.Map{},
 		polledImportInfoMap: make(map[string]*importInfo),
+		validator:           validations.NewDestinationValidator(),
 	}
 
 	m.config.client.url = conf.GetString("SnowpipeStreaming.Client.URL", "http://localhost:9078")
@@ -176,6 +178,10 @@ func (m *Manager) Upload(asyncDest *common.AsyncDestinationStruct) common.AsyncU
 		switch {
 		case errors.Is(err, errAuthz):
 			m.setBackOff(err)
+			response := m.validator.Validate(ctx, asyncDest.Destination)
+			if !response.Success {
+				err = fmt.Errorf("failed to validate snowpipe credentials: %s", response.Error)
+			}
 			return m.failedJobs(asyncDest, err.Error())
 		case errors.Is(err, errBackoff):
 			return m.failedJobs(asyncDest, err.Error())
@@ -216,6 +222,7 @@ func (m *Manager) Upload(asyncDest *common.AsyncDestinationStruct) common.AsyncU
 	)
 	shouldResetBackoff := true // backoff should be reset if authz error is not encountered for any of the tables
 	isBackoffSet := false      // should not be set again if already set
+	var failedReason error
 	for _, info := range uploadInfos {
 		imInfo, discardImInfo, err := m.sendEventsToSnowpipe(ctx, asyncDest.Destination.ID, &destConf, info)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -268,14 +279,18 @@ func (m *Manager) Upload(asyncDest *common.AsyncDestinationStruct) common.AsyncU
...
 		ImportingCount:      len(importingJobIDs),
 		ImportingParameters: importParameters,
 		FailedJobIDs:        failedJobIDs,
 		FailedCount:         len(failedJobIDs),
 		DestinationID:       asyncDest.Destination.ID,
 	}
+	if failedReason != nil {
+		asyncUploadOutput.FailedReason = failedReason.Error()
+	}
+	return asyncUploadOutput
```
shekhar-rudder, line 293: Currently `failedReason` is not being returned in the output. Made changes to include it only for validations error. If required, we can add it for other cases as well.

achettyiitr: Yeah, let's include it for other cases as well. It's should be a very small change.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -176,6 +178,10 @@ func (m *Manager) Upload(asyncDest *common.AsyncDestinationStruct) common.AsyncU
 		switch {
 		case errors.Is(err, errAuthz):
 			m.setBackOff()
+			response := m.validator.Validate(ctx, asyncDest.Destination)
+			if !response.Success {
+				err = fmt.Errorf("failed to validate snowpipe credentials: %s", response.Error)
+			}
```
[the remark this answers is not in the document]
shekhar-rudder: Including the existing `err`, which relates to permissions, could make the error too verbose and potentially confuse the user. I think returning only a specific and granular permission error should make things clearer and easier to understand.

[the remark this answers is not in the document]
achettyiitr: But we also need to include the context correctly; why is RudderStack performing the validation? What is the reason? let's say, while creating the table, we encountered an authorisation error, which is why we are validating.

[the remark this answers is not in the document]
shekhar-rudder: I don’t think additional context is necessary in this case. When sending data to a destination, it is reasonable to expect RudderStack to first attempt a connection to ensure the credentials are valid before proceeding.

warehouse/validations/steps.go
```diff
@@ -19,11 +19,12 @@ func StepsToValidate(dest *backendconfig.DestinationT) *model.StepsResponse {
 		destType = dest.DestinationDefinition.Name
 		steps    []*model.Step
 	)
-
-	steps = []*model.Step{{
-		ID:   len(steps) + 1,
-		Name: model.VerifyingObjectStorage,
-	}}
+	if destType != warehouseutils.SnowpipeStreaming {
```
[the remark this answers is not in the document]
shekhar-rudder: refactored

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
shekhar-rudder: Done. Now sending the very first error in case of multiple `uploadInfos`

warehouse/validations/steps.go
```diff
@@ -15,63 +17,50 @@ func validateStepFunc(_ context.Context, destination *backendconfig.DestinationT
...
 
 func StepsToValidate(dest *backendconfig.DestinationT) *model.StepsResponse {
-	var (
-		destType = dest.DestinationDefinition.Name
-		steps    []*model.Step
-	)
+	destType := dest.DestinationDefinition.Name
+
+	if destType == warehouseutils.SnowpipeStreaming {
+		dest.Config["useKeyPairAuth"] = true // Since we are currently only supporting key pair auth
```
achettyiitr, line 23: Can we move this check to [CreateManager](https://github.com/rudderlabs/rudder-server/blob/06e99fa8cc786347fed0dffa102d107ab8e66d1e/warehouse/validations/validate.go#L497) for Snowpipe Streaming.
 Currently we would need to hardcode this because Snowpipe Streaming only supports key pair Authentication?

achettyiitr (approved):

warehouse/validations/steps.go
shekhar-rudder: Now that you mention it, I believe the validations package shouldn't be aware of the supported authentication types. That configuration should be provided by the caller. I'll move it to `snowpipestreaming` package

vyeshwanth (approved):

PR #5480, opened 2025-02-05
fix: async framework destination live events

achettyiitr:
# Description

- Async destination debugger currently handles only failed and succeeded state, In the case of importing state, it is getting translated to failed status.

## Linear Ticket

- Resolves PIPE-1880

## Security


ktgowtham (approved):

mihir20 (approved):
LGTM

PR #5483, opened 2025-02-06
chore: fix gateway metrics

mihir20:
# Description

We iterate on the jobs and for each job we increase the events by 1 like this:
```
for _, jws := range jobsWithStats {
	jws.stat.RequestEventsSucceeded(1)
```
The problem is that, whenever you call RequestEventsSucceeded it is also increasing the requests:
```
func (ss *SourceStat) RequestEventsSucceeded(num int) {
	ss.events.succeeded += num
	ss.events.total += num
	ss.requests.total++
	ss.requests.succeeded++
}
```
So we should increment events and requests separately. there are some tags like sourceID which are required for events but do not makes sense for internalBatch request. So introducing new methods `EventsSuccess` `EventsFailed` to increment events and requests separately.

## Linear Ticket

-

## Security

### DIFF
--- a/gateway/handle.go
+++ b/gateway/handle.go
@@ -13,9 +13,9 @@ import (
 	"time"
 
 	jsoniter "github.com/json-iterator/go"
+	"github.com/samber/lo"
 
 	"github.com/google/uuid"
-	"github.com/samber/lo"
 	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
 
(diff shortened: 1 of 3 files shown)


gateway/handle.go
```diff
@@ -675,13 +675,15 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
 			})
 			if err = gw.storeJobs(ctx, jobs); err != nil {
 				for _, jws := range jobsWithStats {
-					jws.stat.RequestEventsFailed(1, "storeFailed")
+					jws.stat.EventsFailed(1, "storeFailed")
 					jws.stat.Report(gw.stats)
 				}
+				stat.RequestFailed("storeFailed")
+				stat.Report(gw.stats)
```
fracasula, line 682: This whole approach of jobs with stats embedded smells and it's starting to become a problem.

Calling `Report` each time in a loop doesn't seem like a good idea. I know it was already there but I had missed the first time. As already mentioned [here](https://github.com/rudderlabs/rudder-schemas/pull/40#discussion_r1886524287) we have to be especially careful about what we're doing in loops.

`Report` is calling `NewTaggedStats` which we know uses mutexes under the hood. So imagine doing this for a batch of thousands. We're paying a substantial performance penalty because of bad designed code.

I think this should work the other way around. Instead of having jobs with stats we should have a stats-aware component where you just feed it jobs and their status and then report once at the end.

```go
stats.AddFailedEvent(job, "storeFailed") // gets workspaceID, sourceID from job
stats.AddFailedEvent(job, "storeFailed") // gets workspaceID, sourceID from job
stats.AddFailedEvent(job, "storeFailed") // gets workspaceID, sourceID from job
stats.Report() // stats could have kept maps of tags according to labels like workspaceID etc... and then it reports once per Tags
```

fracasula (approved):
I approved given that you didn't introduce the issue but do keep it in mind please since you also review other people's code, it's important that we design code that is decoupled and efficient. So please do keep an eye out for these issues when you review as well.

Sidddddarth (approved):

fracasula:
@mihir20 when are you planning to release this one? we need to double check codota after the release to see if the metrics make sense.

PR #5493, opened 2025-02-10
chore: sanitise sample event payload and sample response before inserting in reports table

Sidddddarth:
# Description

Sanitise sample event payloads and sample responses before storing to reports table.

## Linear Ticket

[Resolves PIPE-1882](https://linear.app/rudderstack/issue/PIPE-1882/fix-storing-text-json-in-reports-and-add-test-cases)

## Security


fracasula (approved):
LGTM :+1: 

[optional] Maybe you can use an alias for `jsoniter.ConfigCompatibleWithStandardLibrary`.

enterprise/reporting/utils.go
```diff
@@ -104,3 +131,25 @@ func getPIIColumnsToExclude() []string {
 	}
 	return piiColumnsToExclude
 }
+
+func sanitizeJSONForReports(input json.RawMessage) (json.RawMessage, error) {
```
atzoum, line 135: We appear to already have more than enough similar sanitization functions:

1. `jobsdb.sanitizeJSON`
2. `whutils.SanitizeJSON`

I would propose to extract this function to rudder-go-kit and use it wherever necessary

ktgowtham: Will be following up on this [here](https://linear.app/rudderstack/issue/PIPE-1834/gateway-validation-library)

utils/misc/misc.go
```diff
@@ -965,3 +967,35 @@ func GetInstanceID() string {
 	}
 	return ""
 }
+
+var jsonfast = jsoniter.ConfigCompatibleWithStandardLibrary
+
+// SanitizeJSON makes a json payload safe for writing into postgres.
+// 1. Removes any \u0000 string from the payload
+// ~2. Replaces any invalid utf8 characters using github.com/rudderlabs/rudder-go-kit/utf8~
+// 3. unmashals and marshals the payload to remove any extra keys
```
atzoum, line 976:
```suggestion
// 3. unmarshals and marshals the payload to remove any extra keys
```

atzoum, line 976: > remove any extra keys

please elaborate

@lvrach do you remember why this is needed? It was introduced [here](https://github.com/rudderlabs/rudder-server/pull/4762)

atzoum (approved):

PR #5503, opened 2025-02-12
chore: migrate sample event column to text for reporting

itsmihir:
# Description

Recently, we migrated the jobsDB payload column to the text type; however, in the `reports` and `error_detail_reports` tables, we still use the jsonb type to store sample_event. This has led to instances where Postgres failed to cast sample_event from string to jsonb due to special characters, resulting in processor panics.  
This PR migrates the `sample_event` column in these tables from `jsonb` to `text` to ensure consistency and prevent such failures.

## Linear Ticket

Resolves OBS-776

## Security


enterprise/reporting/reporting_test.go
```diff
@@ -504,150 +501,3 @@ func TestGetAggregatedReports(t *testing.T) {
...
-				if !bytes.Equal(gotEvent, tt.wantEvent) {
-					t.Errorf("getSampleWithEventSampling() gotEvent = %s, want %s", gotEvent, tt.wantEvent)
-				}
-				if gotResponse != tt.wantResponse {
-					t.Errorf("getSampleWithEventSampling() gotResponse = %v, want %v", gotResponse, tt.wantResponse)
-				}
-			}
-		})
-	}
-}
```
satishrudderstack, line 653: These sanitization tests were added recently when we migrated to `jobsdb` column to text right ? This sanitization is needed only for json column ?

enterprise/reporting/utils.go
```diff
@@ -46,28 +45,15 @@ func getAggregationBucketMinute(timeMs, intervalMs int64) (int64, int64) {
...
 
 	if !eventSamplingEnabled || eventSampler == nil {
-		// Sanitize both sample event and response before returning
-		if sampleEvent != nil {
-			sampleEvent, err = misc.SanitizeJSON(sampleEvent)
-			if err != nil {
-				return []byte(`{}`), "", err
-			}
-		}
-		sampleResponse = sanitizeStringForReports(sampleResponse)
```
satishrudderstack, line 66: Why are we removing this logic to sanitize sample response ?

processor/processor.go
```diff
@@ -1979,14 +1979,14 @@ func (proc *Handle) processJobsForDest(partition string, subJobs subJob) (*preTr
...
 				reportingtypes.GATEWAY,
-				func() json.RawMessage {
+				func() string {
 					if sourceIsTransient {
-						return []byte(`{}`)
+						return misc.EmptyPayloadString
 					}
 					if payload := event.payloadFunc(); payload != nil {
-						return payload
+						return string(payload)
```
satishrudderstack, line 1987: I think we can optimize a bit here as string conversion could be costly. Instead of converting the `payload` to string here, we can do it in reporting.go after event sampling checks are done and we need to store the sample_event ? We need not modify the `SampleEvent` type in `StatusDetail` struct ?

enterprise/reporting/reporting_test.go
itsmihir: This was added in this [PR](https://github.com/rudderlabs/rudder-server/pull/5493/files#diff-a5d5d7cc37f8c202f47d8f5514c444c7e01cdcd2c64c4c3ea4c5461ee32a325bR594).
Yes, we are doing this sanitization because postgres is not able to convert text to json.

processor/processor.go
itsmihir: we are converting sampleEvent to string [here](https://github.com/rudderlabs/rudder-server/pull/5503/files#diff-1d1dc8727d9e6e324367bb0b660d11f7d06861313f3f5c6c56cda066be221036L70) before sampling. So wouldn't see any improvement.

satishrudderstack (approved):

ktgowtham (approved):

PR #5504, opened 2025-02-12
fix: snowpipe backoff missing for validation error

shekhar-rudder:
# Description

If a Snowpipe destination is created with an invalid private key, the Snowpipe client returns an `ERR_VALIDATION_ERROR` code. Previously, this was handled in the default case, preventing it from being classified as an `Authz` error. As a result, neither backoff nor config validation occurred in the Upload function.

This PR fixes the issue by correctly identifying and returning this error as `Authz`.

- Added `ErrAuthenticationFailed` for invalid user name and `ErrRoleDoesNotExistOrNotAuthorized` for invalid role

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/channel.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/channel.go
@@ -138,7 +138,7 @@ func (m *Manager) createChannel(
 		}
 		m.channelCache.Store(tableName, resp)
 		return resp, nil
-	case internalapi.ErrValidationError:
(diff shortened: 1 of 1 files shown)


router/batchrouter/asyncdestinationmanager/snowpipestreaming/channel.go
```diff
@@ -138,6 +138,8 @@ func (m *Manager) createChannel(
 		}
 		m.channelCache.Store(tableName, resp)
 		return resp, nil
+	case internalapi.ErrValidationError:
+		return nil, fmt.Errorf("%w, %w", errAuthz, err)
```
shekhar-rudder, line 142: Technically speaking, this is an authentication error and not authorization. But I don't think there is a need for that distinction yet.

achettyiitr (approved):

RanjeetMishra (approved):

PR #5508, opened 2025-02-13
feat: gRPC API to expire the warehouse schema of a destination

shekhar-rudder:
# Description

This PR adds a GRPC API which expires all warehouse schemas of a destination in the `wh_schemas` table by setting `expiresAt` column as `now()`. This will ensure that the next time these schemas are fetched from the DB, they will re-fetched from the warehouse.

## Security


proto/warehouse/warehouse.proto
```diff
@@ -20,6 +20,7 @@ service Warehouse {
   rpc RetrieveFailedBatches(RetrieveFailedBatchesRequest) returns (RetrieveFailedBatchesResponse);
   rpc RetryFailedBatches(RetryFailedBatchesRequest) returns (RetryFailedBatchesResponse);
   rpc GetFirstAbortedUploadInContinuousAbortsByDestination(FirstAbortedUploadInContinuousAbortsByDestinationRequest) returns (FirstAbortedUploadInContinuousAbortsByDestinationResponse);
+  rpc SyncWHSchema(SyncWHSchemaRequest) returns (SyncWHSchemaResponse);
```
achettyiitr, line 23:
```suggestion
  rpc SyncWHSchema(SyncWHSchemaRequest) returns (google.protobuf.Empty);
```

warehouse/api/grpc.go
```diff
@@ -973,6 +976,24 @@ func (g *GRPC) RetryFailedBatches(
...
+		lf.DestinationID, req.GetDestinationId(),
+	)
+	log.Infow("Syncing warehouse schema")
+	if req.DestinationId == "" {
+		return &proto.SyncWHSchemaResponse{},
+			status.Errorf(codes.Code(code.Code_INVALID_ARGUMENT), "destinationId cannot be empty")
+	}
+	_, err := g.schemaRepo.SetExpiryForDestination(ctx, req.DestinationId, timeutil.Now())
+	if err != nil {
+		log.Errorw("unable to set expiry for destination", obskit.Error(err))
```
achettyiitr, line 990: Does it need to be a Warning?

warehouse/internal/repo/schema.go
```diff
@@ -252,3 +253,20 @@ func (sh *WHSchema) GetTablesForConnection(ctx context.Context, connections []wa
...
+	result, err := sh.db.ExecContext(ctx, query, expiresAt, sh.now(), destinationID)
+	if err != nil {
+		return 0, fmt.Errorf("updating expiry: %w", err)
+	}
+	rowsAffected, err := result.RowsAffected()
+	if err != nil {
+		return 0, fmt.Errorf("getting rows affected: %w", err)
+	}
+	return int(rowsAffected), nil
+}
```
achettyiitr, line 272: affected rows count is not being used, Can't we do something like this?
```suggestion
func (sh *WHSchema) SetExpiryForDestination(ctx context.Context, destinationID string, expiresAt time.Time) error {
	query := `
		UPDATE ` + whSchemaTableName + `
		SET expires_at = $1, updated_at = $2
		WHERE destination_id = $3
	`
	result, err := sh.db.ExecContext(ctx, query, expiresAt, sh.now(), destinationID)
	if err != nil {
		return fmt.Errorf("updating expiry: %w", err)
	}
	rowsAffected, err := result.RowsAffected()
	if err != nil {
		return fmt.Errorf("getting rows affected: %w", err)
	}
	if rowsAffected == 0 {
		return fmt.Errorf("no rows affected")
	}
	return nil
}
```

warehouse/api/grpc.go
shekhar-rudder: The API not being able to successfully set expiry should be an error , rt ? Why should it be a warning ?

warehouse/internal/repo/schema.go
shekhar-rudder: I don't think we should be returning an error if `rowsAffected == 0`. It is a valid case where there are no entries for a destination in the table and user triggers the API. If you want, I can get rid of `rowsAffected` completely

achettyiitr: Yeah, let's get rid of it if not being used.

proto/warehouse/warehouse.proto
shekhar-rudder: Addressed

warehouse/internal/repo/schema.go
shekhar-rudder: Removed

warehouse/api/grpc.go
```diff
@@ -989,6 +991,24 @@ func (g *GRPC) RetryFailedBatches(
...
+func (g *GRPC) SyncWHSchema(ctx context.Context, req *proto.SyncWHSchemaRequest) (*emptypb.Empty, error) {
+	log := g.logger.With(
+		lf.DestinationID, req.GetDestinationId(),
+	)
+	log.Infow("Syncing warehouse schema")
+	if req.DestinationId == "" {
+		return &emptypb.Empty{},
+			status.Errorf(codes.Code(code.Code_INVALID_ARGUMENT), "destinationId cannot be empty")
+	}
+	err := g.schemaRepo.SetExpiryForDestination(ctx, req.DestinationId, timeutil.Now())
```
achettyiitr, line 1003:
```suggestion
	err := g.schemaRepo.SetExpiryForDestination(ctx, req.DestinationId, g.now())
```

achettyiitr (approved):

vyeshwanth (approved):

PR #5513, opened 2025-02-14
[WIP] feat: configurable json library

atzoum:
# Description

## Linear Ticket

## Security


processor/processor_isolation_test.go
```diff
@@ -267,7 +283,7 @@ func ProcIsolationScenario(t testing.TB, spec *ProcIsolationScenarioSpec) (overa
 		t.Name(),
 	)
 
-	batchSize := 5
+	batchSize := 100
```
atzoum, line 286: revert

router/transformer/transformer_test.go
```diff
@@ -524,16 +525,16 @@ var oauthV2RtTcs = []oauthV2TestCase{
 			},
 		},
 		routerTransformResponses: []types.DestinationJobT{
-			{JobMetadataArray: []types.JobMetadataT{{JobID: 1, WorkspaceID: "wsp"}}, StatusCode: http.StatusOK, Destination: oauthDests[0]},
-			{JobMetadataArray: []types.JobMetadataT{{JobID: 2, WorkspaceID: "wsp"}}, StatusCode: http.StatusUnauthorized, AuthErrorCategory: common.CategoryRefreshToken, Destination: oauthDests[0]},
-			{JobMetadataArray: []types.JobMetadataT{{JobID: 3, WorkspaceID: "wsp"}}, StatusCode: http.StatusOK, Destination: oauthDests[0]},
-			{JobMetadataArray: []types.JobMetadataT{{JobID: 4, WorkspaceID: "wsp"}}, StatusCode: http.StatusUnauthorized, AuthErrorCategory: common.CategoryRefreshToken, Destination: oauthDests[0]},
+			{JobMetadataArray: []types.JobMetadataT{{JobID: 1, WorkspaceID: "wsp", Secret: []byte("null")}}, StatusCode: http.StatusOK, Destination: oauthDests[0], Message: []byte("{}")},
```
atzoum, line 528: **_Note:_** json **standard** library and **sonnet** unmarshall `null` json values to a `json.RawMessage("null")` in contrast to **jsoniter** which unmarshals to `nil` `json.RawMessage`

jsonrs/json_test.go
```diff
@@ -0,0 +1,119 @@
...
+							res, err := j.Marshal(v)
+							require.NoError(t, err)
+							require.Equal(t, string(bytes.Replace([]byte(tpl), []byte("here"), []byte(output), 1)), string(res))
+
+						}
+					})
+				}
+				escapeChar := "�"
+				if name == jsonrs.JsoniterLib {
+					escapeChar = `\ufffd` // unique behaviour: jsoniter uses the \ufffd instead of � that all other libraries use
```
atzoum, line 90: **Note:** unique json-iterator behaviour compared to the other two

jsonrs/sonnet.go
```diff
@@ -0,0 +1,49 @@
...
+		return "", err
+	}
+	return string(m), nil
+}
+
+func (j *sonnetJSON) NewDecoder(r io.Reader) Decoder {
+	// We are always using jsoniter for streaming due to: https://github.com/go-json-experiment/jsonbench?tab=readme-ov-file#streaming
+	return jsoniter.ConfigCompatibleWithStandardLibrary.NewDecoder(r)
+
+	// TODO: if we were to enable sonnet streaming it would unreliable.
```
atzoum, line 37: **Note:** I was unable to root cause the issue, but this shouldn't be a blocker since `json-iterator` appears to provide the only true streaming support anyway.

.github/workflows/tests.yaml
```diff
@@ -177,7 +185,22 @@ jobs:
...
+          TEST_KAFKA_AZURE_EVENT_HUBS_CLOUD_HOST: ${{ secrets.TEST_KAFKA_AZURE_EVENT_HUBS_CLOUD_HOST }}
+          TEST_KAFKA_AZURE_EVENT_HUBS_CLOUD_EVENTHUB_NAME: ${{ secrets.TEST_KAFKA_AZURE_EVENT_HUBS_CLOUD_EVENTHUB_NAME }}
+          TEST_KAFKA_AZURE_EVENT_HUBS_CLOUD_CONNECTION_STRING: ${{ secrets.TEST_KAFKA_AZURE_EVENT_HUBS_CLOUD_CONNECTION_STRING }}
+          TEST_S3_DATALAKE_CREDENTIALS: ${{ secrets.TEST_S3_DATALAKE_CREDENTIALS }}
+          BIGQUERY_INTEGRATION_TEST_CREDENTIALS: ${{ secrets.BIGQUERY_INTEGRATION_TEST_CREDENTIALS }}
+          SNOWPIPE_STREAMING_KEYPAIR_UNENCRYPTED_INTEGRATION_TEST_CREDENTIALS: ${{ secrets.SNOWPIPE_STREAMING_KEYPAIR_UNENCRYPTED_INTEGRATION_TEST_CREDENTIALS }}
+          SNOWPIPE_STREAMING_KEYPAIR_ENCRYPTED_INTEGRATION_TEST_CREDENTIALS: ${{ secrets.SNOWPIPE_STREAMING_KEYPAIR_ENCRYPTED_INTEGRATION_TEST_CREDENTIALS }}
+          RSERVER_JSON_LIBRARY: "jsoniter"
+        run: FORCE_RUN_INTEGRATION_TESTS=true make test exclude="${{ matrix.exclude }}" package=${{ matrix.package }}
+      # TODO: run only one set of tests after sonnet becomes the default implementation
```
atzoum, line 202: **_Note:_** we might need to live with duplicate test runs for sometime

jsonrs/sonnet.go
```diff
@@ -0,0 +1,49 @@
...
+	return string(m), nil
+}
+
+func (j *sonnetJSON) NewDecoder(r io.Reader) Decoder {
+	// We are always using jsoniter for streaming due to: https://github.com/go-json-experiment/jsonbench?tab=readme-ov-file#streaming
+	return jsoniter.ConfigCompatibleWithStandardLibrary.NewDecoder(r)
+
+	// TODO: if we were to enable sonnet streaming it would be unreliable.
+	// Issue is easy to reproduce by running warehouse api tests with sonnet streaming enabled
+	// return sonnet.NewDecoder(r)
```
atzoum, line 39: **_Note:_** I was unable to root cause the issue, but this shouldn't be a blocker since `json-iterator` appears to provide the only true streaming support anyway.

lvrach:
PR title suggestion:
`feat: make json library configurable and introduce sonnet as an option`

lvrach:
I believe the existing test between libraries will benefit with some fuzzing between the 3 libraries, using stdlib as the source of truth.

Basically any input that works on stdlib should work in the other two and given similar results. Giving the existing test cases as corpus a let the fuzzer run for a couple of hours, so we can gain confidence before it reaches prod.

atzoum:
> I believe the existing test between libraries will benefit with some fuzzing between the 3 libraries, using stdlib as the source of truth.
> 
> Basically any input that works on stdlib should work in the other two and given similar results. Giving the existing test cases as corpus a let the fuzzer run for a couple of hours, so we can gain confidence before it reaches prod.

All three libraries have been tested thoroughly [here](https://github.com/go-json-experiment/jsonbench/blob/master/bench_test.go), which includes [this thorough test suite](https://github.com/go-json-experiment/jsonbench/blob/master/testdata/JSONTestSuite/README.md). All three were found to be correct.
Do you consider that some fuzz tests would still be valuable?

fracasula (approved):
jsonrs ftw :+1:

PR #5525, opened 2025-02-18
feat: embedded google pubsub transformer

Sidddddarth:
# Description

Embedding the google pubsub destination transformation in processor transformation.

## Linear Ticket

[Resolves PIPE-1915](https://linear.app/rudderstack/issue/PIPE-1915/port-dt-to-rudder-server-for-kafka-and-pubsub)

## Security


processor/internal/transformer/destination_transformer/transform.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/rudderlabs/rudder-server/processor/types"
+)
+
+type transformerImpl func(ctx context.Context, clientEvents []types.TransformerEvent) types.Response
+
+var embeddedTransformerImpls = map[string]transformerImpl{
+	"GOOGLEPUBSUB": pubsub.Transform,
+}
+
+func (c *Client) Transform(ctx context.Context, clientEvents []types.TransformerEvent) types.Response {
```
lvrach, line 16: Struct methods should live on the same file as struct definition. In general opt for fewer files inside a package even if the file size is becoming larger.

processor/internal/transformer/destination_transformer/transform.go
```diff
@@ -0,0 +1,38 @@
+package destination_transformer
+
+import (
+	"context"
+
+	"github.com/rudderlabs/rudder-server/processor/internal/transformer/destination_transformer/embedded/pubsub"
+	"github.com/rudderlabs/rudder-server/processor/types"
+)
+
+type transformerImpl func(ctx context.Context, clientEvents []types.TransformerEvent) types.Response
```
lvrach, line 10: [minir] 'Impl' suffix is a bit confusing here, as this is not a concrete implementation.

koladilip (approved):
Looks fine, Note: I am not very well versed with pubsub destination

processor/internal/transformer/destination_transformer/embedded/pubsub/pubsub_test.go
```diff
@@ -0,0 +1,676 @@
...
+						StatusCode: http.StatusOK,
+						Metadata:   types.Metadata{},
+					},
+				},
+			},
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
```
lvrach, line 656: It would be useful to add some tests that compare against the node.js implementation similar to how we done it for warehouse transformation.

https://github.com/rudderlabs/rudder-server/blob/master/warehouse/transformer/transformer_test.go#L603

We can even add fuzzing tests, that only compare that given the same input the output can be the same 

https://github.com/rudderlabs/rudder-server/blob/master/warehouse/transformer/transformer_fuzz_test.go#L534

This will give us more confidence, before we even have to test it / compare it in production.

Non-blocking comment we can add this in a follow up PR

lvrach (approved):

processor/internal/transformer/destination_transformer/embedded/pubsub/pubsub.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/rudderlabs/rudder-go-kit/stringify"
+	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
+	utils "github.com/rudderlabs/rudder-server/processor/internal/transformer/destination_transformer/embedded"
+	types "github.com/rudderlabs/rudder-server/processor/types"
+	"github.com/rudderlabs/rudder-server/utils/misc"
+)
+
+var sourceKeys = []string{"properties", "traits", "context.traits"}
+
+func Transform(_ context.Context, events []types.TransformerEvent) types.Response {
```
lvrach, line 18: Deciding to implement `Transform` as function rather than a method makes sense with existing codebase. However, if in the feature we would like to add stats/logs or config access, a struct needs to be introduce to facilitate DI and avoid the use of globals. 

Let's keep simple for now.

PR #5530, opened 2025-02-20
fix: refreshing of datalake expired schemas causing inifinite loop

shekhar-rudder:
# Description

**BUG**: In the `getSchema` method of `schemaV2` if the schema is found to be expired, it calls the `FetchSchema` method of the corresponding warehouse. In case of datalake, it calls `GetLocalSchema` of the uploader which calls the `GetLocalSchema` of schema. This created an inifinte loop as `GetLocalSchema` again calls `getSchema`

**FIX**: Moved the logic of fetching schema from DB from `schema` to `uploader` and removed it from the `Handle` interface

## Security

### DIFF
--- a/warehouse/integrations/datalake/datalake_test.go
+++ b/warehouse/integrations/datalake/datalake_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/xitongsys/parquet-go/reader"
 	"github.com/xitongsys/parquet-go/types"
 
+	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/filemanager"
 
 	"google.golang.org/api/option"
@@ -83,15 +84,15 @@ func TestIntegration(t *testing.T) {
 		jobsDB := whth.JobsDB(t, jobsDBPort)
 
 		testCases := []struct {
-			name           string
-			tables         []string
-			destType       string
-			conf           map[string]interface{}
-			schemaTTL      int
-			prerequisite   func(t testing.TB, ctx context.Context)
-			configOverride map[string]any
-			verifySchema   func(*testing.T, filemanager.FileManager, string)
-			verifyRecords  func(*testing.T, filemanager.FileManager, string, string, string)
+			name               string
+			tables             []string
+			destType           string
+			conf               map[string]interface{}
+			schemaTTLInMinutes int
+			prerequisite       func(t testing.TB, ctx context.Context)
+			configOverride     map[string]any
+			verifySchema       func(*testing.T, filemanager.FileManager, string)
+			verifyRecords      func(*testing.T, filemanager.FileManager, string, string, string)
 		}{
 			{
 				name:     "S3Datalake",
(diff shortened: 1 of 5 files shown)


warehouse/integrations/datalake/datalake_test.go
```diff
@@ -107,6 +109,7 @@ func TestIntegration(t *testing.T) {
 					"prefix":           "some-prefix",
 					"syncFrequency":    "30",
 				},
+				schemaTTL: 0,
```
shekhar-rudder, line 112: regression test

warehouse/schema/schema.go
```diff
@@ -100,7 +99,7 @@ func New(
 		"destinationId": warehouse.Destination.ID,
 	})
 	if conf.GetBoolVar(true, "Warehouse.enableSchemaTTL") {
-		ttlInMinutes := conf.GetDurationVar(720, time.Minute, "Warehouse.schemaTTLInMinutes")
+		ttlInMinutes := conf.GetDurationVar(720, time.Minute, "Warehouse.schemaTtlInMinutes")
```
achettyiitr, line 102: Previous config (`schemaTTLInMinutes`) seems more appropriate. For abbreviations, it's better to keep it as uppercase.

warehouse/integrations/datalake/datalake_test.go
```diff
@@ -86,6 +87,7 @@ func TestIntegration(t *testing.T) {
 			tables         []string
 			destType       string
 			conf           map[string]interface{}
+			schemaTTL      int
```
achettyiitr, line 90:
```suggestion
			schemaTTLInMinutes      int
```

warehouse/router/upload.go
```diff
@@ -942,7 +942,20 @@ func (job *UploadJob) DTO() *model.UploadJob {
 }
 
 func (job *UploadJob) GetLocalSchema(ctx context.Context) (model.Schema, error) {
-	return job.schemaHandle.GetLocalSchema(ctx)
+	schemaRepo := repo.NewWHSchemas(job.db)
```
achettyiitr, line 945: Instead of creating everytime, can we handle it like we handle it for other repos like `uploadsRepo` in `UploadJobFactory`

warehouse/schema/schema.go
shekhar-rudder: With `schemaTTLInMinutes`, the env key will be `SCHEMA_TTLIN_MINUTES` which I found less readable. So I changed it to `schemaTtlInMinutes` which has env key as `SCHEMA_TTL_IN_MINUTES`

achettyiitr: I think that should be fine, we don't use env keys for configuring these configs. Now there is inconsistency because at one place we have `enableSchemaTTL` and at the other place we have `schemaTtlInMinutes`.

We can use something like this for test if we want:
```GO
t.Setenv(config.ConfigKeyToEnv(config.DefaultEnvPrefix, "Warehouse.schemaTtlInMinutes"), strconv.Itoa(tc.schemaTTL))
```

warehouse/router/upload.go
shekhar-rudder: Added a `whSchemaRepo` field in the `UploadJob` struct

warehouse/integrations/datalake/datalake_test.go
shekhar-rudder: Renamed

warehouse/schema/schema.go
shekhar-rudder: Reverted the env casing change

achettyiitr (approved):

vyeshwanth (approved):

PR #5533, opened 2025-02-20
chore: skip previously failed tables

achettyiitr:
# Description

- Skipping previously failed tables when retrying for failed syncs, we can set the failed syncs priority as lower so that it will not affect the real-time syncs. Failed syncs will get processed at a lower priority.
- Using this config, we will not skip the tables which failed for previous syncs. If we don't do it then the current sync will fail till the retry policy and will eventually get marked as aborted.
- In most retry cases, we usually go with the default priority as 100, which means the failed syncs will be processed first and then the real-time ones. But if the number of re-triggered syncs are too many it will take time and affect the real-time syncs.

## Linear Ticket

- Resolves WAR-329.

## Security

### DIFF
--- a/warehouse/router/state_export_data.go
+++ b/warehouse/router/state_export_data.go
@@ -157,7 +157,9 @@ func (job *UploadJob) TablesToSkip() (map[string]model.PendingTableUpload, map[s
 	)
 
 	for _, pendingTableUpload := range job.pendingTableUploads {
-		if pendingTableUpload.UploadID < job.upload.ID && pendingTableUpload.Status == model.TableUploadExportingFailed {
+		if !job.config.skipPreviouslyFailedTables &&
+			pendingTableUpload.UploadID < job.upload.ID &&
+			pendingTableUpload.Status == model.TableUploadExportingFailed {
 			previouslyFailedTableMap[pendingTableUpload.TableName] = pendingTableUpload
 		}
 		if pendingTableUpload.UploadID == job.upload.ID && pendingTableUpload.Status == model.TableUploadExported { // Current upload and table upload succeeded
--- a/warehouse/router/upload.go
+++ b/warehouse/router/upload.go
@@ -102,6 +102,7 @@ type UploadJob struct {
 		maxParallelLoadsWorkspaceIDs        map[string]interface{}
 		columnsBatchSize                    int
 		longRunningUploadStatThresholdInMin time.Duration
+		skipPreviouslyFailedTables          bool
 	}
 
 	errorHandler    ErrorHandler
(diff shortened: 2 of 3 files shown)


warehouse/router/state_export_data.go
```diff
@@ -157,7 +157,9 @@ func (job *UploadJob) TablesToSkip() (map[string]model.PendingTableUpload, map[s
 	)
 
 	for _, pendingTableUpload := range job.pendingTableUploads {
-		if pendingTableUpload.UploadID < job.upload.ID && pendingTableUpload.Status == model.TableUploadExportingFailed {
+		if !job.config.skipPreviouslyFailedTables &&
```
shekhar-rudder, line 160: If this config is `true`, then that means a newer sync can get processed even though there exists an older sync in the backlog.
1. Is this behaviour okay in terms of data correctness ?
2. Do we have this assumption anywhere else in the code that the oldest sync always get picked up first? If yes, we will need changes there as well

warehouse/router/upload_test.go
```diff
@@ -539,8 +533,6 @@ func TestUploadJobT_TablesToSkip(t *testing.T) {
 	})
 
 	t.Run("skip tables", func(t *testing.T) {
-		t.Parallel()
```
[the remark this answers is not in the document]
achettyiitr: Yes.

warehouse/router/state_export_data.go
achettyiitr: > Is this behaviour okay in terms of data correctness ?

This config is basically to be used when we are doing retries, since recently the requests for retries have increased. For retries, we don't want to affect the real-time syncs. With this config set, if we retry the failed syncs with lower priority it would not affect the real-time ones.

> Do we have this assumption anywhere else in the code that the oldest sync always get picked up first? If yes, we will need changes there as well.

No. It depends on various factors like `priority`, `nextRetryTime` whether which sync will be picked up or not. Generally when doing retries we modify these things depending on customer requests whether real-time syncs are more important for them or failed ones. Also, the number of failed sync matter as well.

shekhar-rudder: > This config is basically to be used when we are doing retries, since recently the requests for retries have increased

So post retry, will we turn off the config ? Any downsides to have the config as always on ?

achettyiitr: Yeah, post retry we would need to turn it off. The problem with keeping this config on is the ordering of the delivered events.

warehouse/router/upload.go
```diff
@@ -202,6 +203,7 @@ func (f *UploadJobFactory) NewUploadJob(ctx context.Context, dto *model.UploadJo
 	uj.config.minUploadBackoff = f.conf.GetDurationVar(60, time.Second, "Warehouse.minUploadBackoff", "Warehouse.minUploadBackoffInS")
 	uj.config.maxUploadBackoff = f.conf.GetDurationVar(1800, time.Second, "Warehouse.maxUploadBackoff", "Warehouse.maxUploadBackoffInS")
 	uj.config.retryTimeWindow = f.conf.GetDurationVar(180, time.Minute, "Warehouse.retryTimeWindow", "Warehouse.retryTimeWindowInMins")
+	uj.config.skipPreviouslyFailedTables = f.conf.GetBool("Warehouse.skipPreviouslyFailedTables", false)
```
shekhar-rudder, line 206: Should the config be hot reloadable if we want it to be on only during retries ?

achettyiitr: It's not required, These values are read for every sync, So once we change the config, after that any new syncs are prepared it will have the updated value.

shekhar-rudder (approved):

vyeshwanth (approved):

vyeshwanth:
@achettyiitr checked this already. Approved the change

PR #5542, opened 2025-02-24
fix: error handling for async destinations

koladilip:
# Description

Async destinations are not providing clear errors when destination initialization is failing so customers are getting confused.

## Linear Ticket

https://linear.app/rudderstack/issue/INT-3323/investigate-object-storage-configuration-errors-in-snowflake-to-eloqua
Resolves INT-3323/

## Security

### DIFF
--- a/router/batchrouter/handle_async.go
+++ b/router/batchrouter/handle_async.go
@@ -524,6 +524,17 @@ func (brt *Handle) sendJobsToStorage(batchJobs BatchedJobs) {
 			continue
 		}
 
+		if invalidManager, ok := brt.asyncDestinationStruct[destinationID].Manager.(*asynccommon.InvalidManager); ok {
+			failedAsyncJobs := BatchedJobs{
+				Jobs:       []*jobsdb.JobT{job},
+				Connection: batchJobs.Connection,
+				TimeWindow: batchJobs.TimeWindow,
+				JobState:   jobsdb.Aborted.State,
+			}
+			brt.updateJobStatus(&failedAsyncJobs, false, invalidManager.Error, false)
+			continue
+		}
+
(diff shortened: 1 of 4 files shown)


achettyiitr (approved):

router/batchrouter/handle_async.go
```diff
@@ -476,6 +476,16 @@ func (brt *Handle) sendJobsToStorage(batchJobs BatchedJobs) {
...
+		if invalidManager, ok := brt.asyncDestinationStruct[destinationID].Manager.(*asynccommon.InvalidManager); ok {
+			failedAsyncJobs := BatchedJobs{
+				Jobs:       batchJobs.Jobs,
+				Connection: batchJobs.Connection,
+				TimeWindow: batchJobs.TimeWindow,
+				JobState:   jobsdb.Aborted.State,
+			}
+			brt.updateJobStatus(&failedAsyncJobs, false, invalidManager.Error, false)
+			return
+		}
```
utsabc, line 488:
```suggestion
		if invalidManager, ok := brt.asyncDestinationStruct[destinationID].Manager.(*asynccommon.InvalidManager); !ok {
			failedAsyncJobs := BatchedJobs{
				Jobs:       batchJobs.Jobs,
				Connection: batchJobs.Connection,
				TimeWindow: batchJobs.TimeWindow,
				JobState:   jobsdb.Aborted.State,
			}
			brt.updateJobStatus(&failedAsyncJobs, false, invalidManager.Error, false)
			return
		}
```

Here we are  are handling error case sho shouldnt the condition be `!ok`?

koladilip: we are checking if manager is invalid manager and aborting jobs if so that means ok == true then we need to fail jobs.

utsabc (approved):

PR #5543, opened 2025-02-24
chore: add stats for warehouse process API

achettyiitr:
# Description

- Stats for warehouse process API. Once the stat is added, need to add an alert for non 2xx's response code.

## Linear Ticket

- Resolves WAR-343

## Security

### DIFF
--- a/warehouse/client/warehouse.go
+++ b/warehouse/client/warehouse.go
@@ -93,22 +93,6 @@ func NewWarehouse(baseURL string, statsFactory stats.Stats, opts ...WarehouseOpt
 }
 
 func (w *Warehouse) Process(ctx context.Context, stagingFile StagingFile) error {
-	var (
-		status     = "success"
-		statusCode = http.StatusOK
-	)
-
-	defer func() {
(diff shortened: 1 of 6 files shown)


warehouse/client/warehouse.go
```diff
@@ -87,7 +92,23 @@ func NewWarehouse(baseURL string, opts ...WarehouseOpts) *Warehouse {
 	return warehouse
 }
 
-func (warehouse *Warehouse) Process(ctx context.Context, stagingFile StagingFile) error {
+func (w *Warehouse) Process(ctx context.Context, stagingFile StagingFile) error {
+	var (
```
shekhar-rudder, line 96: I think the default status should be "error" instead of "success". Otherwise if we miss setting `status` and `statusCode` in an error case, it will be treated as success.

achettyiitr: Got rid of the defer statement.

shekhar-rudder (approved):

vyeshwanth (approved):

PR #5557, opened 2025-02-28
chore: varchar handling for mssql and azure synapse

achettyiitr:
# Description

- Prevent data truncation when sending to Microsoft SQL Server and Azure Synapse.
- This involves, querying the `INFORMATION_SCHEMA.COLUMNS` for a particular table, to find the `CHARACTER_MAXIMUM_LENGTH` corresponding varchar columns and truncate based on it.

## Linear Ticket

- Resolves WAR-351

## Security


warehouse/integrations/azure-synapse/azure-synapse.go
```diff
@@ -459,15 +512,19 @@ func (as *AzureSynapse) ProcessColumnValue(
 	case model.BooleanDataType:
 		return strconv.ParseBool(value)
 	case model.StringDataType:
-		if len(value) > stringLengthLimit {
-			value = value[:stringLengthLimit]
+		if varcharLength == varcharMaxLength {
```
RanjeetMishra, line 515: when would we get `varcharMaxLength` as in `-1`? Is it how it returns the value from getVarcharLengths?

RanjeetMishra (approved):

warehouse/integrations/azure-synapse/azure-synapse.go
achettyiitr: We can increase the size to 8000 using VARCHAR(8000) post then we can only set it as VARCHAR(MAX) which has a storage space of 2 GB. Querying the `INFORMATION_SCHEMA.COLUMNS` returns -1 in that case.

https://learn.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-ver16#varchar---n--max--

vyeshwanth (approved):

warehouse/integrations/azure-synapse/azure-synapse.go
```diff
@@ -459,15 +512,19 @@ func (as *AzureSynapse) ProcessColumnValue(
...
+		if len(value) > maxStringLength {
+			value = value[:maxStringLength]
 		}
 		if !hasDiacritics(value) {
 			return value, nil
 		} else {
 			byteArr := str2ucs2(value)
-			if len(byteArr) > stringLengthLimit {
-				byteArr = byteArr[:stringLengthLimit]
+			if len(byteArr) > maxStringLength {
```
shekhar-rudder, line 526: Given that we have already truncated `value` above to `maxStringLength`, is it still possible for `byteArr` length to be more than `maxStringLength` ?

warehouse/integrations/azure-synapse/azure-synapse.go
shekhar-rudder: We can add it as a comment in the code

shekhar-rudder (approved):

warehouse/integrations/azure-synapse/azure-synapse.go
achettyiitr: Since we are already slicing to maxStringLength, it's length will not be more than maxStringLength.

shekhar-rudder: Then can we get rid of `if len(byteArr) > maxStringLength` code since it will always be false ?

achettyiitr: `str2ucs2` function converts a **UTF-8** string to **UCS-2 encoding** using **UTF-16 encoding** where it's length will increase.

PR #5558, opened 2025-03-03
chore: add pattern based matching for version deprecation errors

koladilip:
# Description

The current keys count method is missing cases like `version 2402 is not active`, so now added pattern-based matching for more robust detection of version errors 

## Linear Ticket

[INT-3339: LinkedIn ads version deprecation](https://linear.app/rudderstack/issue/INT-3339/linkedin-ads-version-deprecation)

## Security

### DIFF
--- a/enterprise/reporting/error_extractor.go
+++ b/enterprise/reporting/error_extractor.go
@@ -343,10 +343,47 @@ func getErrorCodeFromStatTags(statTags map[string]string) string {
 func (ext *ExtractorHandle) isVersionDeprecationError(errorMessage string) bool {
 	var score int
 
+	// Convert to lowercase for case-insensitive matching
 	errorMessage = strings.ToLower(errorMessage)
+
+	// 1. Check for direct keyword matches (existing approach)
 	for keyword, s := range lowercasedDeprecationKeywords {
 		if strings.Contains(errorMessage, keyword) {
 			score += s
+		}
+	}
+
+	// 2. Check for version-related patterns with negative context
+	if score <= ext.versionDeprecationThresholdScore.Load() {
+		// Version indicators
+		versionPatterns := []string{"version", "api", "endpoint"}
+
+		// status indicators
+		statusPatterns := []string{
+			"not.*active", "inactive", "no longer.*active",
+			"not.*supported", "unsupported", "no longer.*supported",
+			"not.*valid", "invalid", "no longer.*valid",
+			"not.*available", "unavailable", "no longer.*available",
+			"disabled", "expired", "removed", "discontinued",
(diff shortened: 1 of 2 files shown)


enterprise/reporting/error_extractor.go
```diff
@@ -343,13 +343,50 @@ func getErrorCodeFromStatTags(statTags map[string]string) string {
...
+
+		// status indicators
+		statusPatterns := []string{
+			"not.*active", "inactive", "no longer.*active",
+			"not.*supported", "unsupported", "no longer.*supported",
+			"not.*valid", "invalid", "no longer.*valid",
+			"not.*available", "unavailable", "no longer.*available",
+			"disabled", "expired", "removed", "discontinued",
+			"deprecated", "obsolete",
+		}
```
achettyiitr, line 369: Since these seem static, should we make it global instead of creating it every time?

enterprise/reporting/error_extractor.go
```diff
@@ -343,13 +343,50 @@ func getErrorCodeFromStatTags(statTags map[string]string) string {
...
+				pattern2 := fmt.Sprintf("%s.*%s", sPattern, vPattern)
+
+				if matched, _ := regexp.MatchString(pattern1, errorMessage); matched {
+					score += 3
+					break
+				}
+				if matched, _ := regexp.MatchString(pattern2, errorMessage); matched {
+					score += 3
+					break
+				}
```
achettyiitr, line 385: We can avoid creating pattern2 if pattern1 matches.

```GO
                pattern1 := fmt.Sprintf("%s.*%s", vPattern, sPattern)
				if matched, _ := regexp.MatchString(pattern1, errorMessage); matched {
					score += 3
					break
				}

				pattern2 := fmt.Sprintf("%s.*%s", sPattern, vPattern)
				if matched, _ := regexp.MatchString(pattern2, errorMessage); matched {
					score += 3
					break
				}
```

or 

```
				if matched, _ := regexp.MatchString(fmt.Sprintf("%s.*%s", vPattern, sPattern), errorMessage); matched {
					score += 3
					break
				}
				if matched, _ := regexp.MatchString(fmt.Sprintf("%s.*%s", sPattern, vPattern), errorMessage); matched {
					score += 3
					break
				}
```

achettyiitr (approved):

enterprise/reporting/error_extractor.go
koladilip: done

enterprise/reporting/error_extractor.go
koladilip: done

PR #5570, opened 2025-03-05
fix(jobsdb): clearDB option fails with error

atzoum:
# Description

Lack of tests on the `WithClearDB` option caused this feature to break after we introduced migrations that run always as well as due to other prior code refactoring. This option is required for proper execution of jobsdb benchmarks, thus fixing it and adding a test to handle future regressions.

## Linear Ticket

resolves PIPE-1945

## Security


Sidddddarth (approved):

jobsdb/setup.go
```diff
@@ -12,6 +12,10 @@ func (jd *Handle) SchemaMigrationTable() string {
 	return fmt.Sprintf("%s_schema_migrations", jd.tablePrefix)
 }
 
+func (jd *Handle) RunAlwaysSchemaMigrationTable() string {
```
ktgowtham, line 15: Use this [here](https://github.com/rudderlabs/rudder-server/blob/4ce36da2d4eb3c691f05c7097122c6f0e3dc0c00/jobsdb/setup.go#L37) too?

ktgowtham (approved):
Added minor comment. Approving

PR #5571, opened 2025-03-05
chore(jobsdb): benchmarking scenarios

atzoum:
# Description

Preparing two benchmarking scenarios for jobsdb which can be triggered when server is running in embedded mode.
These scenarios are supporting our benchmarking efforts documented [here](https://www.notion.so/rudderstacks/JobsDB-Benchmarking-Q1-2025-1a7f2b415dd0805f9b84c44758eb565f).

## Linear Ticket

resolves PIPE-1920

## Security


atzoum:
> Nice! Consider performing benchmarks using a cli flag..?(`go run main.go --bench`) as opposed to a config variable. And an end to the benchmark.

Just piggybacking on the rudderstack object for the benchmark's execution in k8s. So far, there hasn't been any need for stopping/starting the benchmark on demand, or running it through cli. For now, for running the benchmark locally, one can simply prepare a local test file.

ktgowtham (approved):

PR #5575, opened 2025-03-05
feat: embedded kafka transformer

itsmihir:
# Description

Embedding the kafka destination transformation in processor.

## Linear Ticket

Resolves PIPE-1941

## Security


koladilip (approved):
kafka has batchTransform also what happens to that now?

itsmihir:
> kafka has batchTransform also what happens to that now?

The batching for Kafka is in the router, so when batching is enabled, we will be making a call to rudder-transformer from router.

processor/internal/transformer/destination_transformer/embedded/kafka/kafka.go
```diff
@@ -0,0 +1,123 @@
...
+			return "", false
+		}
+
+		switch messageType {
+		case "identify", "screen", "page", "group", "alias":
+			{
+				if topic, ok := eventTypeToTopicMap[messageType]; ok {
+					return topic, true
+				}
+				break
```
atzoum, line 106: are these breaks needed?
```suggestion
```

processor/internal/transformer/destination_transformer/embedded/kafka/kafka.go
```diff
@@ -0,0 +1,123 @@
...
+			{
+				eventName, ok := message["event"].(string)
+				if !ok || eventName == "" {
+					return "", false
+				}
+
+				if topic, ok := eventToTopicMap[eventName]; ok {
+					return topic, true
+				}
+				break
```
atzoum, line 118:
```suggestion
```

processor/internal/transformer/destination_transformer/embedded/kafka/kafka.go
```diff
@@ -0,0 +1,123 @@
...
+func filterConfigTopics(message types.SingularEventT, destination backendconfig.DestinationT, eventTypeToTopicMap, eventToTopicMap map[string]string) (string, bool) {
+	if destination.Config["enableMultiTopic"] == true {
+		messageType, ok := message["type"].(string)
+		if !ok {
+			return "", false
+		}
+
+		switch messageType {
+		case "identify", "screen", "page", "group", "alias":
+			{
```
atzoum, line 102: braces are not necessary in case statements in go
```suggestion
```

processor/internal/transformer/destination_transformer/embedded/kafka/kafka.go
lvrach: @itsmihir keep in mind that go has a diffrent behaviour for it switch-case in comparison to other languages:

* cases do NOT fall through by default; you must use the `fallthrough` keyword explicitly to do that. Go opted for this behaviour, since fall through is rarely needed and it can cause bugs and unnecessary clutter. 
* It is not necessary and not idiomatic to use {} in your cases

Only reason to use break inside a switch/case
```go
switch value {
case 1:
    if someCondition {
        break // Exit the case early
    }
    fmt.Println("If you want to break early and not run this statement")
case 2:
    // other code
}
```

lvrach (approved):
Approved, but please address the switch/case comments

processor/internal/transformer/destination_transformer/embedded/kafka/kafka_test.go
```diff
@@ -0,0 +1,909 @@
...
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			got := Transform(context.Background(), c.events)
+			require.Equal(t, c.want, got)
+		})
+	}
+}
+
```
lvrach, line 895: This [comment](https://github.com/rudderlabs/rudder-server/pull/5525#discussion_r2028574533), applies here as well. Can be done for both after merging

PR #5579, opened 2025-03-06
feat: add a common package for validation

mihir20:
# Description

add a package for internal-batch processing, this package can be used at other places as well for the processing of the request payload.

## Linear Ticket

pipe-1834

## Security


gateway/processor/processor.go
```diff
@@ -0,0 +1,23 @@
+package processor
+
+import "github.com/rudderlabs/rudder-schemas/go/stream"
+
+type Processor interface {
```
ktgowtham, line 5: We already have a processor package. this causes confusion. We might need to change this to `PayloadEnricher` and `Process` method to `Enrich`.
Also, we're passing copy of `stream.MessageProperties` from one enricher to other. we can use a pointer. 
We can also use [this](https://github.com/rudderlabs/rudder-server/blob/b52d8a91a56b7e81095638b96a02ba21ab1d9180/enterprise/reporting/mediator.go#L82) design pattern which is used in reporting. Have list of enrichers and loop through them.
let's change the name of package to something else and not processor. and rename the files.

mihir20: I aimed to leave the choice of Processor up to the user, as the order is important in this context. However, in the case of reporting, the order doesn't matter, and the mediator is responsible for selecting all the necessary reporters. Here, the user of the interface will decide which processors are required. 

Additionally, I chose the term "processor" because some implementations are not focused on enrichment but instead perform tasks like validation or sanitization as well.

fracasula: I also found `processor` very confusing and it's a very generic term also. In my head `PayloadEnricher` is still a better name even if the term `processor` wasn't already overloaded here at RudderStack. Plus, it's overloaded so two good reasons to rename the component in my opinion.

All generic names like processor, manager, service, etcetera are poor names and should be avoided. We should come up with something more specific and that doesn't describe a thing (e.g. `processor`) but what it does (its action/side effect). I'll revisit this comment once I'm done with the whole review to see if I can come up with a name myself.

gateway/validator/validator.go
```diff
@@ -0,0 +1,51 @@
...
+	validators []PayloadValidator
+}
+
+// NewValidateMediator creates a new ValidatorMediator with default validators.
+func NewValidateMediator(log logger.Logger, validatorFn func(properties *stream.MessageProperties) error) *ValidatorMediator {
+	return &ValidatorMediator{
+		log: log,
+		validators: []PayloadValidator{
+			&MsgPropertiesValidator{
+				validateFn: validatorFn,
```
ktgowtham, line 27: what's the need of `validatorFn` to be passed along? why can't the implementation be part of `MsgPropertiesValidator` itself?

mihir20: Validate method internally caches some information required for validation, so we want to initialise this method only once and pass that as a dependency to all the users

ktgowtham: my point was to use something like below. Not initialise everytime.

`func NewMsgPropertiesValidator() *MsgPropertiesValidator {
	return &MsgPropertiesValidator{
		validateFn: stream.NewMessagePropertiesValidator(),
	}
}`

and use `NewMsgPropertiesValidator()` here

mihir20: updated the code to use constructor, but `stream.NewMessagePropertiesValidator()` needs to be injected as a dependency.

ktgowtham (approved):

fracasula (changes requested):

gateway/handle.go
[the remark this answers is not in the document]
ktgowtham: https://linear.app/rudderstack/issue/PIPE-1995/clean-up-payload-enrichment-code-from-gateway

gateway/handle_lifecycle.go
```diff
@@ -104,6 +106,7 @@ func (gw *Handle) Setup(
 	gw.conf.maxHeaderBytes = config.GetIntVar(524288, 1, "MaxHeaderBytes")
 	// if set to '0', it means disabled.
 	gw.conf.maxConcurrentRequests = config.GetIntVar(50000, 1, "Gateway.maxConcurrentRequests")
+	gw.conf.enableInternalBatchValidator = config.GetReloadableBoolVar(false, "gateway.enableMsgValidator")
```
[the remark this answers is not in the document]
mihir20: by default if we keep validator enabled we will have to coordinate rudder-server and ingestion service release together. We cannot release ingestion svc without releasing rudder-server.

fracasula (approved):

PR #5581, opened 2025-03-07
refactor: add columns in snowflake using IF NOT EXISTS

shekhar-rudder:
# Description

Currently, when adding columns in Snowflake, we use `ADD COLUMN`, which results in an error if the column already exists. For single-column additions, we catch this error.

This PR updates the query to use `IF NOT EXISTS`, preventing Snowflake from throwing an error for existing columns and removing the need for error handling. However, the query will still fail when adding multiple columns if any column beyond the first already exists in the table.

## Security

### DIFF
--- a/warehouse/integrations/snowflake/snowflake.go
+++ b/warehouse/integrations/snowflake/snowflake.go
@@ -1161,7 +1161,7 @@ func (sf *Snowflake) AddColumns(ctx context.Context, tableName string, columnsIn
 	queryBuilder.WriteString(fmt.Sprintf(`
 		ALTER TABLE
 		  %s.%q
-		ADD COLUMN`,
+		ADD COLUMN IF NOT EXISTS`,
 		schemaIdentifier,
 		tableName,
 	))
@@ -1179,19 +1179,6 @@ func (sf *Snowflake) AddColumns(ctx context.Context, tableName string, columnsIn
 	)
 	log.Infow("Adding columns", lf.Query, query)
 	_, err = sf.DB.ExecContext(ctx, query)
-
-	// Handle error in case of single column
-	if len(columnsInfo) == 1 {
-		if err != nil {
-			if checkAndIgnoreAlreadyExistError(err) {
-				log.Infow("Column already exists",
-					lf.ColumnName, columnsInfo[0].Name,
-					lf.Error, err.Error(),
-				)
-				err = nil
-			}
-		}
-	}
 	return
 }
 
--- a/warehouse/integrations/snowflake/snowflake_test.go
+++ b/warehouse/integrations/snowflake/snowflake_test.go
@@ -1258,6 +1258,99 @@ func TestIntegration(t *testing.T) {
 		require.NoError(t, rows.Err())
 		require.Empty(t, ids2, "no more rows left")
 	})
+
+	t.Run("Add Columns", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


warehouse/integrations/snowflake/snowflake_test.go
```diff
@@ -1258,6 +1258,99 @@ func TestIntegration(t *testing.T) {
...
+				wantErr: false,
+			},
+			{
+				name: "add multiple columns that already exist",
+				columnsInfo: []whutils.ColumnInfo{
+					{Name: "col2", Type: "string"},
+					{Name: "col3", Type: "int"},
+				},
+				wantErr: true,
+			},
```
achettyiitr, line 1331: It's a valid test scenario. If all the columns are already present, it should pass.

warehouse/integrations/snowflake/snowflake_test.go
```diff
@@ -1258,6 +1258,99 @@ func TestIntegration(t *testing.T) {
 		require.NoError(t, rows.Err())
 		require.Empty(t, ids2, "no more rows left")
 	})
+
+	t.Run("Add Columns", func(t *testing.T) {
```
achettyiitr, line 1262: Please add a test scenario where some columns are already present and some are new columns.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -1161,7 +1161,7 @@ func (sf *Snowflake) AddColumns(ctx context.Context, tableName string, columnsIn
 	queryBuilder.WriteString(fmt.Sprintf(`
```
achettyiitr, line 1161: We would need to modify the query to pass `IF NOT EXISTS` for every column.
```GO
	queryBuilder.WriteString(fmt.Sprintf(`
		ALTER TABLE %s.%q ADD COLUMN`,
		schemaIdentifier,
		tableName,
	))

	for _, columnInfo := range columnsInfo {
		queryBuilder.WriteString(fmt.Sprintf(` IF NOT EXISTS %q %s,`, columnInfo.Name, dataTypesMap[columnInfo.Type]))
	}
```

achettyiitr (changes requested):

warehouse/integrations/snowflake/snowflake_test.go
```diff
@@ -1258,6 +1258,92 @@ func TestIntegration(t *testing.T) {
...
+				columnsInfo: []whutils.ColumnInfo{
+					{Name: "col2", Type: "string"},
+				},
+			},
+			{
+				name: "add multiple new + existing columns",
+				columnsInfo: []whutils.ColumnInfo{
+					{Name: "col2", Type: "string"},
+					{Name: "col3", Type: "int"},
+					{Name: "col4", Type: "float"},
```
achettyiitr, line 1325: What is the new column here? All three columns are already present?

achettyiitr: Can we add another test with col3, col4, col5 and col6?

shekhar-rudder: These were not supposed to be independent tests but rather sequential. Now that you have pointed it out, I shouldn't be using `t.Run` here. Refactored the code with a simple for loop

achettyiitr (approved):

PR #5583, opened 2025-03-07
chore: warehouse transformer benchmarking tool

achettyiitr:
# Description

- Benchmarking support for warehouse transformations.

## Linear Ticket

- Resolves WAR-389

## Security

### DIFF
--- a/warehouse/transformer/benchmark/Makefile
+++ b/warehouse/transformer/benchmark/Makefile
@@ -1,0 +1,52 @@
+REQUIRED_VARS_BUILD = DOCKER_USER
+REQUIRED_VARS_DEPLOY = DOCKER_USER K8S_NAMESPACE MODE
+
+define check_vars
+    $(foreach var,$(1),$(if $(value $(var)),,$(error $(var) is not set)))
+endef
+
+ifeq ($(MAKECMDGOALS),build-transformer)
+    $(call check_vars,$(REQUIRED_VARS_BUILD))
+endif
+
+ifeq ($(MAKECMDGOALS),build-push-transformer)
+    $(call check_vars,$(REQUIRED_VARS_BUILD))
+endif
+
+ifeq ($(MAKECMDGOALS),prepare-deployment)
+    $(call check_vars,$(REQUIRED_VARS_DEPLOY))
+endif
+
+ifeq ($(MAKECMDGOALS),deploy-transformer)
+    $(call check_vars,$(REQUIRED_VARS_DEPLOY))
+endif
+
+ifeq ($(MAKECMDGOALS),delete-transformer)
+    $(call check_vars,$(REQUIRED_VARS_DEPLOY))
+endif
+
+IMAGE_NAME = $(DOCKER_USER)/rudder-warehouse-benchmarks:latest
+DOCKERFILE_PATH = ./warehouse/transformer/benchmark/Dockerfile
+DEPLOYMENT_TMP = /tmp/rudder-warehouse-benchmarks-transformer-deployment.yaml
+PLATFORM = linux/arm64
+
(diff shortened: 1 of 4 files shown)


vyeshwanth (approved):

warehouse/transformer/benchmark/Makefile
```diff
@@ -0,0 +1,52 @@
...
+ifeq ($(MAKECMDGOALS),deploy-transformer)
+    $(call check_vars,$(REQUIRED_VARS_DEPLOY))
+endif
+
+ifeq ($(MAKECMDGOALS),delete-transformer)
+    $(call check_vars,$(REQUIRED_VARS_DEPLOY))
+endif
+
+IMAGE_NAME = $(DOCKER_USER)/rudder-warehouse-benchmarks:latest
+DOCKERFILE_PATH = ./warehouse/transformer/benchmark/Dockerfile
```
shekhar-rudder, line 29: Can `DOCKERFILE_PATH` be a relative path - `./Dockerfile` ?

warehouse/transformer/benchmark/main.go
```diff
@@ -0,0 +1,143 @@
...
+	runID := uuid.New().String()
+	conf := config.Default
+
+	logFactory := logger.Default
+	l := logFactory.NewLogger().Child("warehouse-transformer-benchmark").Child(mode)
+
+	clientEvents := make([]types.TransformerEvent, eventsInBatch)
+	for i := 0; i < len(clientEvents); i++ {
+		var transformerEvent types.TransformerEvent
+		err := jsonrs.Unmarshal([]byte(sampleEvent), &transformerEvent)
```
shekhar-rudder, line 63: Since value of `sampleEvent` is fixed, it can be unmarshalled only once

warehouse/transformer/benchmark/main.go
```diff
@@ -0,0 +1,143 @@
...
+		for i := 0; i < iterations; i++ {
+			l.Infof("Running iteration: %d", i+1)
+			t.Transform(gCtx, clientEvents, batchSize)
+		}
+
+		endTime := timeutil.Now().Sub(startTime).Seconds()
+
+		l.Infof("Total events transformer: %d", len(clientEvents)*iterations)
+		l.Infof("Total time for transformations: %v", endTime)
+		l.Infof("Transformation rate (msg/s): %v",
```
shekhar-rudder, line 103: How are we using this benchmark to make decisions? Are we evaluating the `transformation rate` for each input combination?

warehouse/transformer/benchmark/Makefile
achettyiitr: No. We need to build the image from the root directory. That's why we are doing this.
```
build-transformer:
	cd ../../../ && \
	docker buildx build -t $(IMAGE_NAME) -f $(DOCKERFILE_PATH) --platform=$(PLATFORM) .
```

warehouse/transformer/benchmark/main.go
achettyiitr: We can't use a single `transformerEvent`, because it contains `SingularEventT` which is a **map** because **maps** in Go are reference types. One option is to just do Unmarshal everytime. Other option is to do Unmarshal once but then later do deep copy using something like:
```GO
func deepCopy(src, dest interface{}) error {
	buf, err := jsonrs.Marshal(src)
	if err != nil {
		return err
	}
	return jsonrs.Unmarshal(buf, dest)
}
```

In favour of the first approach.

warehouse/transformer/benchmark/main.go
achettyiitr: Yeah, we are evaluating the **Transformation rate** along with the **Kubernetes Compute Resources** (CPU and Memory) and the **Pyroscope Profiles** (CPU and Memory).

PR #5588, opened 2025-03-11
chore: warehouse transformer optimizations for concurrency

achettyiitr:
# Description

- Instead of doing sequential processing for each event, introducing concurrency. We are getting 3.6x times more processing. More about it [this](https://www.notion.so/rudderstacks/Warehouse-Transformation-Optimizations-1b2f2b415dd08039a6a5c70695100411?pvs=4#1b3f2b415dd080a9b971d6f8cfd10f78) doc.

## Linear Ticket

- Resolves WAR-279

## Security

### DIFF
--- a/warehouse/transformer/transformer.go
+++ b/warehouse/transformer/transformer.go
@@ -7,7 +7,7 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/samber/lo"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/rudderlabs/rudder-server/processor/types"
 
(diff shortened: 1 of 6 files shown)


vyeshwanth (approved):

warehouse/transformer/transformer.go
```diff
@@ -71,22 +72,47 @@ func (t *Transformer) Transform(_ context.Context, clientEvents []types.Transfor
...
+	g := errgroup.Group{}
+	g.SetLimit(t.config.concurrentTransformations.Load())
+
+	go func() {
+		defer close(done)
+		for r := range results {
+			if r.Error != "" {
+				res.FailedEvents = append(res.FailedEvents, r)
+			} else {
+				res.Events = append(res.Events, r)
```
shekhar-rudder, line 87: I am assuming that order of events in `res` won't be same as the order in `clientEvents`. Is that okay ?

warehouse/transformer/transformer.go
```diff
@@ -71,22 +72,47 @@ func (t *Transformer) Transform(_ context.Context, clientEvents []types.Transfor
...
+		event := clientEvents[i]
+		event.Metadata.SourceDefinitionType = "" // TODO: Currently, it's getting ignored during JSON marshalling Remove this once we start using it.
+
+		g.Go(func() error {
+			r, err := t.processWarehouseMessage(c, &event)
+			if err != nil {
+				results <- transformerResponseFromErr(&event, err)
+				return nil
+			}
+			for _, item := range r {
```
shekhar-rudder, line 102: Why do we have a 2nd routine for populating `res`? Is it to avoid concurrent access to the slice? If yes, can we use a lock instead?

warehouse/transformer/transformer.go
achettyiitr: Order of events don't matter.

shekhar-rudder (approved):

PR #5594, opened 2025-03-12
chore: refactor webhook pkg to make it reusable in other repos

mihir20:
# Description

- Move webhook authentication logic to new auth package (webhook/auth/auth.go)
- Refactor AuthRequestContext type and move types to gateway/types package
- Update stat reporting interface to use StatReporter instead of concrete SourceStat
- fetch transformer features before starting gateway server
- improved error handling
- Update mock implementations to align with new interface changes
- Consolidate webhook handling code and improve type safety
- Add new source details structure to AuthRequestContext for better data encapsulation

## Linear Ticket

https://linear.app/rudderstack/issue/PIPE-1940/common-webhook-for-ingestion-service-and-gateway

## Security


gateway/webhook/webhook.go
```diff
@@ -461,6 +467,22 @@ func (webhook *HandleT) Register(name string) {
 	}
 }
 
+func (webhook *HandleT) enqueue(name string, req *webhookT) {
+	webhook.requestQMu.Lock()
```
mihir20, line 471: take read lock first then if channel is not there in the map acquire a full lock

gateway/handle_observability.go
```diff
@@ -21,6 +21,17 @@ func (gw *Handle) NewSourceStat(arctx *gwtypes.AuthRequestContext, reqType strin
 	}
 }
 
+func (gw *Handle) NewSourceStatReporter(arctx *gwtypes.AuthRequestContext, reqType string) gwtypes.StatReporter {
```
fracasula, line 24: I wonder why this is a method of `*Handle` in the first place :thinking: 

It is completely decoupled already which is a good thing. It only depends on its arguments and it is not using any attribute from `*Handle`.

I noticed because I saw this interface:

```go
type WebhookHandler interface {
	ProcessWebRequest(writer *http.ResponseWriter, req *http.Request, reqType string, requestPayload []byte, arctx *gwtypes.AuthRequestContext) string
	NewSourceStatReporter(arctx *gwtypes.AuthRequestContext, reqType string) gwtypes.StatReporter
	SaveWebhookFailures([]*model.FailedWebhookPayload) error
}
```

When I saw the interface I thought: why would a "webhook handler" have a function to create stats reporters?

Maybe you can have `NewSourceStatReporter` as a simple function and you could even think of removing it from the `WebhookHandler` interface (which can be renamed into `WebhookRequestHandler` instead:

```go
type WebhookRequestHandler interface {
	ProcessWebRequest(writer *http.ResponseWriter, req *http.Request, reqType string, requestPayload []byte, arctx *gwtypes.AuthRequestContext) string
	SaveWebhookFailures([]*model.FailedWebhookPayload) error
}
```

Then the `NewSourceStatReporter` constructor can be its own dependency as needed.

gateway/webhook/setup.go
```diff
@@ -18,25 +21,35 @@ import (
...
+	NewSourceStatReporter(arctx *gwtypes.AuthRequestContext, reqType string) gwtypes.StatReporter
 	SaveWebhookFailures([]*model.FailedWebhookPayload) error
 }
 
+// RequestMetricsTracker is used to track webhook request metrics on a request basis for OSS customers
+type RequestMetricsTracker interface {
+	TrackRequestMetrics(errorMessage string)
+}
+
 type WebHookI interface {
```
fracasula, line 43: Since I suggested to rename `WebhookHandler` to `WebhookRequestHandler`, and given that in here I see `RequestHandler`, now I wonder what is the actual difference from this interface (i.e. `WebHookI` - terrible name btw, it needs some love from you @mihir20 :D) and the above one `WebhookHandler`.

gateway/webhook/webhook.go
```diff
@@ -209,17 +215,21 @@
 		sourceID:    arctx.SourceID,
 		authContext: arctx,
 	}
-	webhook.requestQMu.RLock()
-	requestQ := webhook.requestQ[sourceDefName]
-	requestQ <- &req
-	webhook.requestQMu.RUnlock()
+	if webhook.config.webhookV2HandlerEnabled {
+		webhook.enqueue(sourceDefName, &req)
```
fracasula, line 219: Why create a new method `enqueue` that duplicates the `Register(name string)` logic plus the code of the `else` below?

Wouldn't it be easier if you just call `Register` and then always execute the `else`? Wdyt?

gateway/webhook/setup.go
```diff
@@ -18,25 +21,35 @@ import (
...
 	RequestHandler(w http.ResponseWriter, r *http.Request)
 	Register(name string)
 }
 
+type TransformerFeaturesService interface {
+	SourceTransformerVersion() string
+	Wait() chan struct{}
+}
```
fracasula, line 51: Can't we simplify this one like we did in the ingestion-svc and just have a function like `type TransformerFeaturesVersionFetcher func () string`?

We either make sure that the webhook doesn't receive traffic is the featuresSvc isn't ready (like we do in ingestion-svc) or we make sure that the function itself blocks until ready first and then returns the version. I think the former would be preferrable. Wdyt?

gateway/webhook/setup.go
mihir20: this interface in not in use we can remove this, But I got your point will do some refactoring around this and above comment 👍

gateway/webhook/webhook.go
mihir20: Register method is not taking read lock, it is always taking a full lock. we can modify Register method to use read lock but Register method is always called at the startup so reading will not be yielding any result and we will have to create a new channel anyways. 
That's why I tried to keep the existing logic as it is and created another method. Also I am planning to remove register method as a part of clean-up.

gateway/handle_lifecycle.go
```diff
@@ -130,7 +137,7 @@ func (gw *Handle) Setup(
 	gw.batchUserWorkerBatchRequestQ = make(chan *batchUserWorkerBatchRequestT, gw.conf.maxDBWriterProcess)
 	gw.irh = &ImportRequestHandler{Handle: gw}
 	gw.rrh = &RegularRequestHandler{Handle: gw}
-	gw.webhook = webhook.Setup(gw, transformerFeaturesService, gw.stats)
+	gw.webhook = webhook.Setup(gw, transformerFeaturesService, gw.stats, gw.config, newSourceStatReporter)
```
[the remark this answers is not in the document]
mihir20: yes, webhook is not dependent directly on gw it depends on an interface and implementation of this interface currently do not depend on webhook but in future it can depend on webhook which may create cyclic dependency.

ktgowtham (approved):

gateway/webhook/auth/auth.go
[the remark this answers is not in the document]
ktgowtham: wdyt about this?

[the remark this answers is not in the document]
mihir20: we need to use same auth functionality in ingestion svc as well, and getting authCtx and handling failure should be done on ingestion svc. We can create an auth function with onFailure and getAuthCtxForWriteKey as function params as well. 
honestly this can be done both ways as logic is not that complex. but creating a pkg here is more extensible if later we want to add more auth related functions we can add to this pkg.

gateway/webhook/webhook.go
```diff
@@ -244,6 +262,19 @@ func (webhook *HandleT) RequestHandler(w http.ResponseWriter, r *http.Request) {
 	ss.Report(webhook.stats)
 }
 
+func getWebhookFailureReason(err string, statusCode int) string {
```
fracasula, line 265: Can you rename the variable to `errMessage`? Because idiomatically speaking when someone sees an `err` variable in Go they know it's going to be an `error` but this isn't the case so for clarity we could simply rename it.

gateway/webhook/webhook.go
```diff
@@ -244,6 +262,19 @@ func (webhook *HandleT) RequestHandler(w http.ResponseWriter, r *http.Request) {
...
+	switch {
+	case err == response.RequestBodyTooLarge:
+		return response.RequestBodyTooLarge
+	case statusCode != 0:
+		return response.SourceTransformerNonSuccessResponse
+	case err != "":
+		return response.SourceTransformerResponseError
```
fracasula, line 272: [optional] Not needed at all of course, but I never got the chance to comment about `fallthrough` so... here we are :joy:

```go
case err != "":
    fallthrough
default:
    return response.SourceTransformerResponseError
```

gateway/webhook/webhookTransformer.go
```diff
@@ -109,18 +112,21 @@ func (v2 *v2Adapter) getAdapterVersion() string {
 
 // ------------------------------
 
-func newSourceTransformAdapter(version string) sourceTransformAdapter {
+func newSourceTransformAdapter(version string, conf *config.Config) sourceTransformAdapter {
 	// V0 Deprecation: this function returns v1 adapter by default, thereby deprecating v0
 	if version == transformer.V2 {
-		return &v2Adapter{}
+		return &v2Adapter{
+			baseTransformerURL: conf.GetString("DEST_TRANSFORM_URL", "http://localhost:9090"),
```
fracasula, line 119: [optional] the conf.GetString is duplicated, perhaps create a var before the ifs?

gateway/handle_lifecycle.go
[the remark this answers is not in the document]
fracasula: ### Circular Dependencies
- The webhook package depends on the gateway package (`gwtypes`), and the gateway package depends on the webhook package, creating a circular dependency.
- Specifically, the circular dependency occurs in the following places:
  - The webhook package depends on the gateway package:
    - In `gateway/webhook/setup.go`, the webhook package imports `gwtypes` from the gateway package and uses it in interface definitions and method signatures.
    - In `gateway/webhook/model/model.go`, the `FailedWebhookPayload` struct uses `*gwtypes.AuthRequestContext` from the gateway package.
    - In `gateway/webhook/webhook.go`, the webhook package imports and uses `gwtypes` for authentication and request handling.
  - The gateway package depends on the webhook package:
    - In `gateway/handle.go`, the gateway package has a field of type `webhook.WebhookRequestHandler`.
    - In `gateway/handle_webhook.go`, the gateway implements the `WebhookRequestProcessor` interface defined in the webhook package.
    - In `gateway/mocks/mockwebhook.go`, the gateway package has mock implementations for webhook interfaces.

[optional] Not required as part of this effort, but I'm starting to think that we could move the shared types to a common package that both gateway and webhook can depend on.

Also for the gw depending on webhook, I think we could solve it by using Dependency Injection and pass a ready to be used webhook handler to the gw.

@mihir20 if you feel like this wouldn't be too complex and want to take a stab at it, I won't complain :sweat_smile:

fracasula (approved):
Approving with a few comments, some are minor some are optional. Please address the minor ones before merging.

vinayteki95 (approved):

gateway/types/types.go
```diff
@@ -26,11 +27,34 @@ type AuthRequestContext struct {
...
+		WriteKey    string
+		Config      map[string]interface{}
+	}
 }
 
 func (arctx *AuthRequestContext) SourceTag() string {
 	return misc.GetTagName(arctx.WriteKey, arctx.SourceName)
 }
+
+type StatReporter interface {
```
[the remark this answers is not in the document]
mihir20: We want to use this from ingestion svc as well, so we had to expose this here.

PR #5596, opened 2025-03-13
chore: optimise drain function

cisse21:
# Description
Optimising Drain function and other relevant code blocks by Unmarshaling Parameters only once in the job lifecycle

## Linear Ticket

Fixes PIPE-1968

## Security


router/types/types.go
```diff
@@ -74,23 +75,24 @@ func (dj *DestinationJobT) JobIDs() map[int64]struct{} {
...
+	FirstAttemptedAt   string                    `json:"firstAttemptedAt"`
+	TransformAt        string                    `json:"transformAt"`
+	WorkspaceID        string                    `json:"workspaceId"`
+	Secret             json.RawMessage           `json:"secret"`
+	JobT               *jobsdb.JobT              `json:"jobsT,omitempty"`
+	WorkerAssignedTime time.Time                 `json:"workerAssignedTime"`
+	DestInfo           json.RawMessage           `json:"destInfo,omitempty"`
+	DontBatch          bool                      `json:"dontBatch"`
+	TraceParent        string                    `json:"traceparent"`
+	Parameters         routerutils.JobParameters `json:"parameters"`
```
atzoum, line 95: Isn't this introducing duplicate information, e.g `SourceID`, `DestinationID`, etc?
Isn't `JobMetadataT` sent/received to/from the transformer as well? Do we need this extra complexity of increasing the payload size and relying on transformer to return it back intact?

cisse21: Yeah that was my only concern of introducing duplicate information but this was needed to skip unmarshalling params at status update for every job and since we unmarshal the transformer response anyways we would get away with another unmarshal and I did not remove the existing fields just because they might be used in Transformer for some reason

I also looked into recreating Parameters from the existing fields of metadata but there were too many missing fields

atzoum: What do we really gain if we anyways pay the cost of marshaling/unmarshaling the parameters in the transformer's request? Can't we avoid that?

cisse21: If we just add more fields into the metadata struct, we can just recreate the parameters from the response but wouldn't it be the same as passing params and unmarshalling it?

atzoum: I would refrain from adding anything to this metadata struct, it is a weak design to use service calls for carrying over information which is irrelevant to the service call itself.
I prefer to keep things it they were, at least until we find a better design that doesn't entail carrying over information through HTTP calls.

atzoum: Can we use the Dehydrate/Hydrate for parameters?

router/types/types.go
```diff
@@ -33,12 +34,13 @@ type ConnectionWithID struct {
...
+func (djs DestinationJobs) Hydrate(jobs map[int64]*jobsdb.JobT, jobParameters map[int64]routerutils.JobParameters) {
 	for i := range djs {
 		for j := range djs[i].JobMetadataArray {
 			if djs[i].JobMetadataArray[j].JobT == nil {
 				djs[i].JobMetadataArray[j].JobT = jobs[djs[i].JobMetadataArray[j].JobID]
 			}
+			djs[i].JobMetadataArray[j].Parameters = jobParameters[djs[i].JobMetadataArray[j].JobID]
 		}
 	}
 }
```
atzoum, line 46:
```suggestion
// Hydrate jobs in the destination jobs' job metadata array
func (djs DestinationJobs) Hydrate(jobs map[int64]lo.Tuple2[*jobsdb.JobT, routerutils.JobParameters]) {
	for i := range djs {
		for j := range djs[i].JobMetadataArray {
			t := jobs[djs[i].JobMetadataArray[j].JobID]
			if djs[i].JobMetadataArray[j].JobT == nil {
				djs[i].JobMetadataArray[j].JobT = t.A
			}
			djs[i].JobMetadataArray[j].Parameters = t.B
		}
	}
}
```

router/types/types.go
```diff
@@ -100,16 +103,19 @@ type TransformMessageT struct {
...
 	for i := range tm.Data {
 		tmCopy.Data = append(tmCopy.Data, tm.Data[i])
 		jobs[tmCopy.Data[i].JobMetadata.JobID] = tmCopy.Data[i].JobMetadata.JobT
+		jobParameters[tmCopy.Data[i].JobMetadata.JobID] = tmCopy.Data[i].JobMetadata.Parameters
 		tmCopy.Data[i].JobMetadata.JobT = nil
+		tmCopy.Data[i].JobMetadata.Parameters = routerutils.JobParameters{}
 	}
-	return &tmCopy, jobs
+	return &tmCopy, jobs, jobParameters
 }
```
atzoum, line 119:
```go
// Dehydrate JobT information from RouterJobT.JobMetadata returning the dehydrated message along with the jobs
func (tm *TransformMessageT) Dehydrate() (*TransformMessageT, map[int64]lo.Tuple2[*jobsdb.JobT, routerutils.JobParameters]) {
	jobs := make(map[int64]lo.Tuple2[*jobsdb.JobT, routerutils.JobParameters])
	tmCopy := *tm
	tmCopy.Data = nil
	for i := range tm.Data {
		tmCopy.Data = append(tmCopy.Data, tm.Data[i])
		jobs[tmCopy.Data[i].JobMetadata.JobID] = lo.T2(tmCopy.Data[i].JobMetadata.JobT, tmCopy.Data[i].JobMetadata.Parameters)
		tmCopy.Data[i].JobMetadata.JobT = nil
		tmCopy.Data[i].JobMetadata.Parameters = routerutils.JobParameters{}
	}
	return &tmCopy, jobs
}
```

router/types.go
```diff
@@ -10,15 +10,17 @@ import (
...
-	status   *jobsdb.JobStatusT
-	payload  json.RawMessage
-	statTags map[string]string
+	userID     string
+	worker     *worker
+	job        *jobsdb.JobT
+	status     *jobsdb.JobStatusT
+	payload    json.RawMessage
+	statTags   map[string]string
+	parameters routerutils.JobParameters
```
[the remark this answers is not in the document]
atzoum: btw, the fact that no test caught this is alarming...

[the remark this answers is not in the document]
cisse21: Aren't the changes over [here](https://github.com/rudderlabs/rudder-server/pull/5596/files#diff-dedd5c61a64be3de818ebab595ad82ba3266c9f8d0857c872dcda86203a0e01bR118) or am I missing something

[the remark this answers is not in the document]
atzoum: forgive me, I was reviewing changes locally and for some reason wasn't getting these changes in my branch :(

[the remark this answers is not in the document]
cisse21: Resolving this comments then :)

atzoum (approved):

ktgowtham (approved):

PR #5598, opened 2025-03-13
fix: group jobs by dest id to transformer from router

koladilip:
# Description
OAuth v2 assumes all jobs in batch are related to a single destination ID, and that only happens if isolation mode is destination, but in hosted we have isolation mode set to none so we are grouping jobs by destination before going to transformer.

Resolves INT-3394

## Linear Ticket

https://linear.app/rudderstack/issue/INT-3394

## Security

### DIFF
--- a/router/worker.go
+++ b/router/worker.go
@@ -277,12 +277,14 @@ func (w *worker) workLoop() {
 func (w *worker) transform(routerJobs []types.RouterJobT) []types.DestinationJobT {
 	// transform limiter with dynamic priority
 	start := time.Now()
-	limiter := w.rt.limiter.transform
-	limiterStats := w.rt.limiter.stats.transform
-	defer limiter.BeginWithPriority("", LimiterPriorityValueFrom(limiterStats.Score(w.partition), 100))()
-	defer func() {
-		limiterStats.Update(w.partition, time.Since(start), len(routerJobs), 0)
-	}()
+	if w.rt.limiter.transform != nil {
+		limiter := w.rt.limiter.transform
+		limiterStats := w.rt.limiter.stats.transform
+		defer limiter.BeginWithPriority("", LimiterPriorityValueFrom(limiterStats.Score(w.partition), 100))()
+		defer func() {
+			limiterStats.Update(w.partition, time.Since(start), len(routerJobs), 0)
+		}()
+	}
 
 	traces := make(map[string]stats.TraceSpan)
 	defer func() {
@@ -308,13 +310,20 @@ func (w *worker) transform(routerJobs []types.RouterJobT) []types.DestinationJob
 			w.rt.logger.Debugn("traceParent is empty during router transform", logger.NewIntField("jobId", job.JobMetadata.JobID))
 		}
 	}
-	w.rt.routerTransformInputCountStat.Count(len(routerJobs))
-	destinationJobs := w.rt.transformer.Transform(
-		transformer.ROUTER_TRANSFORM,
-		&types.TransformMessageT{Data: routerJobs, DestType: strings.ToLower(w.rt.destType)},
-	)
-	w.rt.routerTransformOutputCountStat.Count(len(destinationJobs))
-	w.recordStatsForFailedTransforms("routerTransform", destinationJobs)
+	destinationJobs := make([]types.DestinationJobT, 0)
+	destinationIDRouterJobsMap := lo.GroupBy(routerJobs, func(job types.RouterJobT) string {
+		return job.Destination.ID
+	})
+	for _, destinationIDRouterJobs := range destinationIDRouterJobsMap {
+		w.rt.routerTransformInputCountStat.Count(len(destinationIDRouterJobs))
+		destinationIDJobs := w.rt.transformer.Transform(
+			transformer.ROUTER_TRANSFORM,
+			&types.TransformMessageT{Data: destinationIDRouterJobs, DestType: strings.ToLower(w.rt.destType)},
+		)
+		w.rt.routerTransformOutputCountStat.Count(len(destinationIDJobs))
+		w.recordStatsForFailedTransforms("routerTransform", destinationIDJobs)
+		destinationJobs = append(destinationJobs, destinationIDJobs...)
+	}
 	return destinationJobs
 }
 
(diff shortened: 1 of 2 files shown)


router/worker.go
```diff
@@ -277,12 +277,14 @@ func (w *worker) workLoop() {
...
-		limiterStats.Update(w.partition, time.Since(start), len(routerJobs), 0)
-	}()
+	if w.rt.limiter.transform != nil {
+		limiter := w.rt.limiter.transform
+		limiterStats := w.rt.limiter.stats.transform
+		defer limiter.BeginWithPriority("", LimiterPriorityValueFrom(limiterStats.Score(w.partition), 100))()
+		defer func() {
+			limiterStats.Update(w.partition, time.Since(start), len(routerJobs), 0)
+		}()
+	}
```
atzoum, line 287: please revert and simply add a limiter in your test

router/worker_test.go
```diff
@@ -456,3 +457,150 @@ var _ = Describe("Proxy Request", func() {
...
+	worker := &worker{
+		rt: &Handle{
+			transformer:                    mockTransformer,
+			destType:                       "some_dest_type",
+			logger:                         logger.NOP,
+			routerTransformInputCountStat:  stats.NOP.NewTaggedStat("router_transform_input_count", stats.CountType, stats.Tags{"destType": "some_dest_type"}),
+			routerTransformOutputCountStat: stats.NOP.NewTaggedStat("router_transform_output_count", stats.CountType, stats.Tags{"destType": "some_dest_type"}),
+		},
+	}
+	routerJobs := []types.RouterJobT{
```
atzoum, line 476:
```suggestion
	var limiterWg sync.WaitGroup
	var ctx, cancel = context.WithCancel(context.Background())
	defer limiterWg.Wait()
	defer cancel()
	worker.rt.limiter.transform = kitsync.NewLimiter(ctx, &limiterWg, "transform", math.MaxInt, stats.Default)
	worker.rt.limiter.stats.transform = partition.NewStats()
	routerJobs := []types.RouterJobT{
```

router/worker.go
```diff
@@ -308,13 +310,20 @@ func (w *worker) transform(routerJobs []types.RouterJobT) []types.DestinationJob
...
 		}
 	}
-	w.rt.routerTransformInputCountStat.Count(len(routerJobs))
-	destinationJobs := w.rt.transformer.Transform(
-		transformer.ROUTER_TRANSFORM,
-		&types.TransformMessageT{Data: routerJobs, DestType: strings.ToLower(w.rt.destType)},
-	)
-	w.rt.routerTransformOutputCountStat.Count(len(destinationJobs))
-	w.recordStatsForFailedTransforms("routerTransform", destinationJobs)
+	destinationJobs := make([]types.DestinationJobT, 0)
```
atzoum, line 313:
```suggestion
	destinationJobs := make([]types.DestinationJobT, 0, len(routerJobs))
```

router/worker.go
```diff
@@ -308,13 +310,20 @@ func (w *worker) transform(routerJobs []types.RouterJobT) []types.DestinationJob
...
+	destinationIDRouterJobsMap := lo.GroupBy(routerJobs, func(job types.RouterJobT) string {
+		return job.Destination.ID
+	})
+	for _, destinationIDRouterJobs := range destinationIDRouterJobsMap {
```
atzoum, line 317: Ignoring oauth for a moment, can transformer itself handle transformation of jobs for different destination ids?

router/worker_test.go
koladilip: done

router/worker.go
koladilip: transformer can handle different destination ids events, the problem is the oauth v2 interceptor is written in such a way that it assumes that all events belong to same destination id

router/worker.go
koladilip: output jobs and input jobs need not be the same as there will be batching logic

ItsSudip (approved):

router/worker.go
koladilip: this increases the number of calls to transformer but it doesn't effect the out going requests as we anyway group by destination in transformer more over this only effect free tier as most of the other tiers will run destination isolation.

koladilip: We are also planning to refactor the OAuth v2, but that is a slightly bigger change and requires the transformer service change as well, so we will take it up once this is done.

atzoum: Doing 1 http call with n events in the payload is always going to be better than doing n http calls with 1 event in the payload, that's why we should avoid it if possible.

router/worker.go
```diff
@@ -203,7 +203,7 @@ func (w *worker) workLoop() {
 				logger.NewBoolField("oauthV2Enabled", oauthV2Enabled),
 				obskit.DestinationType(destination.DestinationDefinition.Name),
 			)
-			if authType := oauth.GetAuthType(destination.DestinationDefinition.Config); authType == oauth.OAuth && !oauthV2Enabled {
+			if oauth.IsOAuthDestination(destination.DestinationDefinition.Config) && !oauthV2Enabled {
```
atzoum, line 206:
```suggestion
			if w.rt.isOAuthDestination && !oauthV2Enabled {
```

router/worker.go
```diff
@@ -306,14 +329,11 @@ func (w *worker) transform(routerJobs []types.RouterJobT) []types.DestinationJob
...
-	w.rt.routerTransformInputCountStat.Count(len(routerJobs))
-	destinationJobs := w.rt.transformer.Transform(
-		transformer.ROUTER_TRANSFORM,
-		&types.TransformMessageT{Data: routerJobs, DestType: strings.ToLower(w.rt.destType)},
-	)
-	w.rt.routerTransformOutputCountStat.Count(len(destinationJobs))
-	w.recordStatsForFailedTransforms("routerTransform", destinationJobs)
-	return destinationJobs
+
+	if w.rt.reloadableConfig.oauthV2Enabled.Load() && w.rt.isOAuthDestination {
```
atzoum, line 333: no need to check whether `oauthV2Enabled` if it is not an oauth destination
```suggestion
	if w.rt.isOAuthDestination && w.rt.reloadableConfig.oauthV2Enabled.Load() {
```

router/worker.go
koladilip: done

atzoum (approved):

PR #5605, opened 2025-03-17
feat(processor): count pending events without blocking

atzoum:
# Description

Introducing `rmetrics.PendingEventsRegistry` which captures pending events and publishes them to the global registry, i.e. as prometheus metrics:

- During startup the registry is not published, i.e. no metrics are being pushed to prometheus.
- When processor starts, it runs `josbdb.GetPileupCounts` asynchronously.
- `GetPileupCounts` is now time-aware, i.e. it calculates pending events at a reference point in time so that it doesn't get affected by processor's processing stages which run concurrently and can modify the state of jobs.
- While `GetPileupCounts` is running, migrations are paused so that queries for jobs are not being blocked due to locks.
- After `GetPileupCounts` finishes for both `rt` & `batch_rt` jobsdbs, `PendingEventsRegistry` gets published and migrations resume.
- When processor restarts, e.g. due to scheduler, registry is being reset:
  - a new unpublished registry is created
  - the global registry is being cleared
  - registry is published after pileups are counted again

## Linear Ticket

resolves PIPE-1975

## Security


app/apphandlers/embeddedAppHandler.go
```diff
@@ -308,16 +312,18 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
...
-		AdaptiveLimit:    adaptiveLimit,
+		Reporting:             reporting,
+		BackendConfig:         backendconfig.DefaultBackendConfig,
+		RouterDB:              batchRouterDB,
+		ProcErrorDB:           errDBForWrite,
+		TransientSources:      transientSources,
+		RsourcesService:       rsourcesService,
+		Debugger:              destinationHandle,
+		AdaptiveLimit:         adaptiveLimit,
+		PendingEventsRegistry: pendingEventsRegistry,
```
cisse21, line 326: Shouldn't we be doing this in processorAppHandler too?

services/rmetrics/pending_events.go
```diff
@@ -0,0 +1,152 @@
...
+	})
+	pem.registry = publishedRegistry
+}
+
+// Reset resets the registry to a new, non published one and clears the global published metrics registry
+func (pem *pendingEventsRegistry) Reset() {
+	pem.registryMu.Lock()
+	defer pem.registryMu.Unlock()
+	pem.registry = metric.NewRegistry()
+	metric.Instance.Reset()
```
ktgowtham, line 115: Do we need to set `published` to false here?
Is it safe to reset the global instance? will we lose any metrics?

jobsdb/jobsdb.go
```diff
@@ -1849,7 +1851,10 @@ func (jd *Handle) GetToProcess(ctx context.Context, params GetQueryParams, more
 
 var cacheParameterFilters = []string{"source_id", "destination_id"}
 
-func (jd *Handle) GetPileUpCounts(ctx context.Context) error {
+func (jd *Handle) GetPileUpCounts(ctx context.Context, cutoffTime time.Time, increaseFunc rmetrics.IncreasePendingEventsFunc) error {
+	// pause migration to avoid any read locks being blocked during pileup count
+	jd.migrateDSPaused.Store(true)
+	defer jd.migrateDSPaused.Store(false)
 	if !jd.dsMigrationLock.RTryLockWithCtx(ctx) {
```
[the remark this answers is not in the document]
atzoum: Locks need to be acquired in the same order always, otherwise it is a recipe for deadlocks.

services/rmetrics/pending_events.go
atzoum: thanks for catching that!

atzoum: in the global instance only pending events metrics are stored

atzoum: Added a scenario in tests for covering publishing after reset, thanks!

jobsdb/jobsdb.go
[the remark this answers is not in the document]
ktgowtham: exactly

ktgowtham (approved):

jobsdb/jobsdb.go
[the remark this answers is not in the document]
atzoum: @ktgowtham in `doMigrateDS`:
- first we acquire a `dsListLock ` read lock and then we release it.
- then we acquire a `dsMigrationLock` write lock
- then we acquire a `dsListLock` write lock

PR #5607, opened 2025-03-17
feat: introduce workers per partition in processor

cisse21:
# Description

This PR introduces a new worker-based architecture for the event processor, improving the system's ability to handle concurrent event processing. The architecture implements a partitioning mechanism that allows events from the same user to be processed together while enabling parallel processing of events from different users.

- Divides jobs by user into multiple partitions using a murmur hash function
- Processes each partition independently in parallel
- Maintains event order guarantees within each user's event stream
- Improves overall throughput by better utilizing system resources

## Linear Ticket

Fixes PIPE-1962

## Security


processor/worker.go
```diff
@@ -5,41 +5,163 @@ import (
...
 	"github.com/rudderlabs/rudder-server/jobsdb"
 	"github.com/rudderlabs/rudder-server/rruntime"
 	"github.com/rudderlabs/rudder-server/services/rsources"
 	"github.com/rudderlabs/rudder-server/utils/misc"
 )
 
-// newProcessorWorker creates a new worker
+type worker struct {
+	partition string
+	workers   []*partitionWorker
```
ktgowtham, line 19: let's call this `partitionWorkers` or `subWorkers`?

processor/worker.go
```diff
@@ -5,41 +5,163 @@ import (
...
+	// Initialize rsources stats
+	rsourcesStats := rsources.NewStatsCollector(w.handle.rsourcesService(), rsources.IgnoreDestinationID())
+	rsourcesStats.BeginProcessing(jobs.Jobs)
+
+	// Distribute jobs across partitions based on UserID
+	jobsByPartition := w.partitionJobsByUserID(jobs.Jobs, w.handle.config().numPartitions)
+
+	// Send jobs to their respective partitions for processing
+	var wg sync.WaitGroup
+	for partition, partitionJobs := range jobsByPartition {
```
ktgowtham, line 81: here also, `subPartition` to be explicit?

ktgowtham (approved):
LGTM.
Added minor suggestion for naming(optional to handle).

processor/processor_isolation_test.go
```diff
@@ -47,17 +47,39 @@ func TestProcessorIsolation(t *testing.T) {
...
 	})
 
 	t.Run("source isolation", func(t *testing.T) {
-		spec := NewProcIsolationScenarioSpec(isolation.ModeSource, workspaces, jobsPerWorkspace)
+		spec := NewProcIsolationScenarioSpec(isolation.ModeSource, workspaces, jobsPerWorkspace, 1)
+		ProcIsolationScenario(t, spec)
+	})
+}
+
+func TestProcessorIsolationWithWorkers(t *testing.T) {
```
atzoum, line 65: Let's keep just a single test in this file with all necessary scenarios

```go
func TestProcessorIsolation(t *testing.T) {
	const (
		workspaces          = 10
		jobsPerWorkspace    = 100
		workersPerPartition = 3
	)

	for _, isolationMode := range []isolation.Mode{isolation.ModeNone, isolation.ModeWorkspace, isolation.ModeSource} {
		t.Run(fmt.Sprintf("%s isolation", isolationMode), func(t *testing.T) {
			t.Run("1 worker", func(t *testing.T) {
				ProcIsolationScenario(t, NewProcIsolationScenarioSpec(isolationMode, workspaces, jobsPerWorkspace, 1))
			})
			t.Run(fmt.Sprintf("%d workers", workersPerPartition), func(t *testing.T) {
				ProcIsolationScenario(t, NewProcIsolationScenarioSpec(isolationMode, workspaces, jobsPerWorkspace, workersPerPartition))
			})
		})
	}
}
```

processor/worker.go
```diff
@@ -5,41 +5,163 @@ import (
...
 }
 
-// worker picks jobs from jobsdb for the appropriate partition and performs all processing steps for them
+// partitionWorker picks jobs from jobsdb for the appropriate partition and performs all processing steps:
 //  1. preprocess
 //  2. transform
 //  3. store
-type worker struct {
+type partitionWorker struct {
 	partition string
```
atzoum, line 161: Let's keep the concepts and responsibilities clear:
- the _i**solation strategy**_ decides about _**partitions**_
- for each _**partition**_ we have one _**partitionWorker**_
- each **_partitionWorker_** can setup 1 or more _**pipelines**_
- each pipeline has its own **_pipelineKey_** and _**pipelineWorker**_

Sounds good?

Untitled Diagram drawio

processor/worker.go
```diff
@@ -5,41 +5,163 @@ import (
...
 	return w
 }
 
-// worker picks jobs from jobsdb for the appropriate partition and performs all processing steps for them
+// partitionWorker picks jobs from jobsdb for the appropriate partition and performs all processing steps:
 //  1. preprocess
 //  2. transform
 //  3. store
-type worker struct {
+type partitionWorker struct {
```
atzoum, line 160: let's use different files for `partitionWorker` & `pipelineWorker`

processor/worker.go
```diff
@@ -5,41 +5,163 @@ import (
...
+// Returns true if work was done, false otherwise
+func (w *worker) Work() bool {
+	// If pipelining is disabled, use the legacy job handling path
+	if !w.handle.config().enablePipelining {
+		return w.handle.handlePendingGatewayJobs(w.partition)
+	}
+
+	// Get jobs for this partition
+	jobs := w.handle.getJobs(w.partition)
+	start := time.Now()
```
atzoum, line 56: We should start counting before getting jobs, as it was earlier. Was this change intentional?

```suggestion
	start := time.Now()
	// Get jobs for this partition
	jobs := w.handle.getJobs(w.partition)
```

processor/worker.go
```diff
@@ -5,41 +5,163 @@ import (
...
+
+	// If no jobs were found, return false
+	if len(jobs.Jobs) == 0 {
```
atzoum, line 59: Question: we seem to be removing some stats about `Out of order job_id`, I am not sure I understand what these were about

processor/worker.go
```diff
@@ -5,41 +5,163 @@ import (
...
+	if err := w.handle.markExecuting(w.partition, jobs.Jobs); err != nil {
+		w.logger.Error("Error marking jobs as executing", "error", err)
+		panic(err)
+	}
+
+	// Record throughput metrics
+	w.handle.stats().DBReadThroughput(w.partition).Count(throughputPerSecond(jobs.EventsCount, time.Since(start)))
+
+	// Initialize rsources stats
+	rsourcesStats := rsources.NewStatsCollector(w.handle.rsourcesService(), rsources.IgnoreDestinationID())
```
atzoum, line 73: Don't we need different stats per pipeline? Each pipeline should be independent, right?

processor/worker.go
```diff
@@ -5,41 +5,163 @@ import (
...
+	rsourcesStats.BeginProcessing(jobs.Jobs)
+
+	// Distribute jobs across partitions based on UserID
+	jobsByPartition := w.partitionJobsByUserID(jobs.Jobs, w.handle.config().numPartitions)
```
atzoum, line 77: do we need another method for this simple operation?

```go
	lo.GroupBy(jobs.Jobs, func(job *jobsdb.JobT) int {
		return int(misc.GetMurmurHash(job.UserID) % uint64(w.handle.config().numPartitions))
	})
```

processor/worker.go
```diff
@@ -5,41 +5,163 @@ import (
...
+
+	// Send jobs to their respective partitions for processing
+	var wg sync.WaitGroup
+	for partition, partitionJobs := range jobsByPartition {
+		wg.Add(1)
+		go func(partition int, jobs []*jobsdb.JobT) {
+			defer wg.Done()
+			subJobs := w.handle.jobSplitter(jobs, rsourcesStats)
+			for _, subJob := range subJobs {
+				w.workers[partition].channel.preprocess <- subJob
```
atzoum, line 87: 1. We need to respect the context and terminate gracefully upon context cancellation (stopping)
2. The preprocess channel can be a bottleneck initially, not sure whether we need to start with a greater buffer, e.g. similarly to what we are doing for store buffer

processor/processor.go
```diff
@@ -805,6 +806,7 @@ func (proc *Handle) loadConfig() {
 	proc.config.enablePipelining = config.GetBoolVar(true, "Processor.enablePipelining")
 	proc.config.pipelineBufferedItems = config.GetIntVar(0, 1, "Processor.pipelineBufferedItems")
 	proc.config.subJobSize = config.GetIntVar(defaultSubJobSize, 1, "Processor.subJobSize")
+	proc.config.numPartitions = config.GetIntVar(1, 1, "Processor.numPartitions")
```
atzoum, line 809:
```suggestion
	proc.config.pipelinesPerPartition = config.GetIntVar(1, 1, "Processor.pipelinesPerPartition")
```

atzoum, line 809: We might need to reconsider and potentially remove this piece of code, since it might be too defensive, wdyt?
```go
	// If isolation mode is not none, we need to reduce the values for some of the config variables to more sensible defaults
	if proc.config.isolationMode != isolation.ModeNone {
		defaultSubJobSize = 400
		defaultMaxEventsToProcess = 2000
		defaultPayloadLimit = 20 * bytesize.MB
	}
```

cisse21: We should reconsider everything with respect to payload sizes and adaptive payload limiter since we want to rely on VPA to provide more memory and better throughput than restrict the throughput and keep the system safe. Thats a completely different discussion I guess

processor/worker.go
atzoum: Not true, they are not being processed at the same time & as it stands now:
- statsCollector is not thread safe
- statsCollector's publish will be called multiple times from each pipeline with unpredictable results

Each set of jobs should have its own stats collector

processor/partition_worker.go
```diff
@@ -0,0 +1,135 @@
...
+	workers   []*pipelineWorker
+	logger    logger.Logger
+	stats     *processorStats
+	handle    workerHandle
+	g         *errgroup.Group
+	ctx       context.Context
+}
+
+// newProcessorWorker creates a new worker for the specified partition
+func newPartitionWorker(ctx context.Context, partition string, h workerHandle) *partitionWorker {
```
atzoum, line 29: partition worker should use its own independent context, controlled through `Stop`
```suggestion
func newPartitionWorker(partition string, h workerHandle) *partitionWorker {
```

processor/partition_worker.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// Mark jobs as executing
+	if err := w.handle.markExecuting(w.partition, jobs.Jobs); err != nil {
+		w.logger.Error("Error marking jobs as executing", "error", err)
+		panic(err)
+	}
+
+	// Record throughput metrics
+	w.handle.stats().DBReadThroughput(w.partition).Count(throughputPerSecond(jobs.EventsCount, time.Since(start)))
```
atzoum, line 72: I cannot find us using this metric (`processor_db_read_throughput`) in any of our dashboards or alerts. I propose to delete it. We capture throughput through jobsdb metrics now

processor/partition_worker.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Initialize rsources stats
+	rsourcesStats := rsources.NewStatsCollector(w.handle.rsourcesService(), rsources.IgnoreDestinationID())
```
atzoum, line 75: Please create different stats for each pipeline

processor/pipeline_worker.go
```diff
@@ -3,43 +3,47 @@ package processor
...
 	"github.com/rudderlabs/rudder-server/jobsdb"
 	"github.com/rudderlabs/rudder-server/rruntime"
-	"github.com/rudderlabs/rudder-server/services/rsources"
-	"github.com/rudderlabs/rudder-server/utils/misc"
 )
 
-// newProcessorWorker creates a new worker
-func newProcessorWorker(partition string, h workerHandle) *worker {
-	w := &worker{
+// newWorker creates a new partition worker
```
atzoum, line 12:
```suggestion
// newPipelineWorker creates a new worker which manages a single pipeline of a partition
```

processor/partition_worker.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/rudderlabs/rudder-go-kit/logger"
+	"github.com/rudderlabs/rudder-server/jobsdb"
+	"github.com/rudderlabs/rudder-server/services/rsources"
+	"github.com/rudderlabs/rudder-server/utils/misc"
+)
+
+type partitionWorker struct {
+	partition string
+	workers   []*pipelineWorker
```
atzoum, line 20: minor, just a suggestion
```suggestion
	pipelines   []*pipelineWorker
```

processor/partition_worker.go
```diff
@@ -0,0 +1,145 @@
...
+	logger    logger.Logger
+	stats     *processorStats
+	handle    workerHandle
+	ctx       context.Context
+	cancel    context.CancelFunc
+	g         *errgroup.Group
+}
+
+// newProcessorWorker creates a new worker for the specified partition
```
atzoum, line 26:
```suggestion
// newPartitionWorker creates a new worker for the specified partition
```

processor/partition_worker.go
```diff
@@ -0,0 +1,145 @@
...
+func newPartitionWorker(partition string, h workerHandle) *partitionWorker {
+	w := &partitionWorker{
+		partition: partition,
+		logger:    h.logger().Child(partition),
+		stats:     h.stats(),
+		handle:    h,
+	}
+	w.ctx, w.cancel = context.WithCancel(context.Background())
+	w.g, w.ctx = errgroup.WithContext(w.ctx)
+	// Create workers for each partition
```
atzoum, line 34:
```suggestion
	// Create workers for each pipeline
```

processor/partition_worker.go
```diff
@@ -0,0 +1,144 @@
...
+func (w *partitionWorker) SleepDurations() (min, max time.Duration) {
+	return w.handle.config().readLoopSleep.Load(), w.handle.config().maxLoopSleep.Load()
+}
+
+// Stop stops the worker and waits until all its goroutines have stopped
+func (w *partitionWorker) Stop() {
+	w.cancel()
+	for _, worker := range w.workers {
+		worker.Stop()
+	}
```
atzoum:
```suggestion
	w.cancel()
	_ = w.g.Wait()
	for _, worker := range w.workers {
		worker.Stop()
	}
```

processor/pipeline_worker.go
```diff
@@ -3,43 +3,47 @@ package processor
...
+	w.channel.store = make(chan *storeMessage, storeBufferSize)
+
+	// Start processing goroutines
 	w.start()
 
 	return w
 }
 
-// worker picks jobs from jobsdb for the appropriate partition and performs all processing steps for them
+// pipelineWorker picks jobs from jobsdb for the appropriate partition and performs all processing steps:
```
atzoum, line 38:
```suggestion
// pipelineWorker performs all processing steps of a partition's pipeline:
```

processor/pipeline_worker.go
```diff
@@ -3,43 +3,47 @@ package processor
...
 //  1. preprocess
 //  2. transform
```
atzoum, line 40: where is pretransform?

processor/partition_worker.go
```diff
@@ -0,0 +1,126 @@
...
+
+// SleepDurations returns the min and max sleep durations for the worker
+func (w *partitionWorker) SleepDurations() (min, max time.Duration) {
+	return w.handle.config().readLoopSleep.Load(), w.handle.config().maxLoopSleep.Load()
+}
+
+// Stop stops the worker and waits until all its goroutines have stopped
+func (w *partitionWorker) Stop() {
+	for _, worker := range w.workers {
+		worker.Stop()
```
atzoum, line 124: suggestion: we can stop pipelines concurrently

PR #5625, opened 2025-03-19
refactor: version deprecation detection avoid regex

koladilip:
# Description

Version deprecation extraction uses regex and it consumes a lot of CPU so avoid it

## Linear Ticket

https://linear.app/rudderstack/issue/INT-3339/linkedin-ads-version-deprecation

## Security

### DIFF
--- a/enterprise/reporting/error_extractor.go
+++ b/enterprise/reporting/error_extractor.go
@@ -56,20 +56,19 @@ var (
 
 	// Status-related terms
 	statusPatterns = []string{
-		"not active", "inactive", "no longer active",
+		"not active", "inactive", "no longer active", "invalid",
 		"not supported", "unsupported", "no longer supported",
-		"not valid", "invalid", "no longer valid",
+		"not valid", "no longer valid", "old", "retired",
 		"not available", "unavailable", "no longer available",
 		"disabled", "expired", "removed", "discontinued",
 		"deprecated", "deprecation", "obsolete", "obsolescence",
-		"outdated", "end of life", "end-of-support", "legacy",
-		"retired", "old", "no longer", "upgrade required",
-		"upgrade recommended",
+		"outdated", "end of life", "end of support",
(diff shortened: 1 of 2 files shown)


enterprise/reporting/error_extractor.go
```diff
@@ -353,39 +350,31 @@ func getErrorCodeFromStatTags(statTags map[string]string) string {
 }
 
 func (ext *ExtractorHandle) isVersionDeprecationError(errorMessage string) bool {
```
atzoum, line 352: are there any tests?

koladilip: yes all passing

koladilip: we don't need to maintain original behavior as this only for internal monitoring purpose.

lvrach, line 352: Given the performance-sensitive nature, I would write a benchmark and test other techniques as well:

For instance, create a single compiled regex combining all the terms, and check how it compares against `strings.Contains` 

regexp.Compile("not.*active|inactive|no longer.*active|not.*supported")

depending on the number of elements in the slice, a single pre-compiled regex can be more efficient

vinayteki95 (commented):
Keeping optimisation aside, the logic seems stronger now.

enterprise/reporting/error_reporting_test.go
```diff
@@ -315,6 +315,102 @@ func TestExtractErrorDetails(t *testing.T) {
 				errorCode: "deprecation",
 			},
 		},
```
vinayteki95, line 317: Can we add some negative test cases?
Some of the false positives which are a part of alerts in the last couple of days can be used.

vinayteki95 (approved):

atzoum (approved):

lvrach (approved):

PR #5639, opened 2025-03-24
feat(processor): break down transformations step

cisse21:
# Description
This PR aims at introducing new stages in processor by splitting the current transformations stage into `usertransformer` and `destinationtransformer` stages.

## Additional items
- cleanup unused metrics from processor
- introduce new stage-related metrics in processor for being able to capture stage throughput & latency per job

## Linear Ticket

Fixes PIPE-1973

## Security


atzoum (commented):
I propose some more uniform names for pipeline stages, so that it is easier to distinguish them, e.g.

- `getJobs` -> `getJobsStage`
- `processJobsForDest` -> `preprocessStage`
- `generateTransformationMessage` -> `pretransformStage`
- `usertransformations` -> `userTransformStage`
- `destinationtransformations` ->  `destinationTransformStage`
- `Store` -> `storeStage`

Another thing is that, now, every stage captures the time that it takes through its limiter:
- `proc_read_limiter_working`
- `proc_preprocess_limiter_working`
- etc.

What is missing for us to get a clear picture on which stage is the actual bottleneck is a counter of the number of original jobs that the limiter processes, e.g.
- `proc_read_limiter_jobcount`
- `proc_preprocess_limiter_jobcount`
- etc

processor/processor.go
```diff
@@ -2935,57 +3021,97 @@ func (proc *Handle) transformSrcDest(
 	eventFilterStat.numOutputFilteredEvents.Count(len(nonSuccessMetrics.filteredJobs))
 	eventFilterStat.transformTime.Since(s)
 
-	// Filtering events based on the supported message types - END
```
atzoum, line 2938: this END comment is missing now

processor/processor.go
```diff
@@ -2935,57 +3021,97 @@ func (proc *Handle) transformSrcDest(
...
+	if len(data.eventsToTransform) == 0 {
 		return transformSrcDestOutput{
-			destJobs:        destJobs,
-			batchDestJobs:   batchDestJobs,
-			errorsPerDestID: procErrorJobsByDestID,
-			reportMetrics:   reportMetrics,
-			routerDestIDs:   routerDestIDs,
-			droppedJobs:     droppedJobs,
+			destJobs:        []*jobsdb.JobT{},
+			batchDestJobs:   []*jobsdb.JobT{},
```
atzoum, line 3046:
```suggestion
			destJobs:        nil,
			batchDestJobs:   nil,
```

processor/processor.go
```diff
@@ -3013,20 +3139,20 @@ func (proc *Handle) transformSrcDest(
 				diffMetrics := getDiffMetrics(
 					reportingtypes.EVENT_FILTER,
 					reportingtypes.DEST_TRANSFORMER,
-					inCountMetadataMap,
-					successCountMetadataMap,
-					inCountMap,
+					data.inCountMetadataMap,
+					nil,
```
atzoum, line 3143: was it a bug in the previous implementation that we were using `successCountMetadataMap`?

processor/processor.go
```diff
@@ -3036,6 +3162,7 @@ func (proc *Handle) transformSrcDest(
 		// Save the JSON in DB. This is what the router uses
 		for i := range response.Events {
 			destEventJSON, err := jsonrs.Marshal(response.Events[i].Output)
+			proc.logger.Infof(string(destEventJSON))
```
atzoum, line 3165:
```suggestion
```

processor/processor.go
```diff
@@ -2849,34 +2943,26 @@ func (proc *Handle) transformSrcDest(
...
-	}
-	transformAtFromFeaturesFile := proc.transformerFeaturesService.RouterTransform(destination.DestinationDefinition.Name)
-
 	// Filtering events based on the supported message types - START
 	s := time.Now()
 	eventFilterInCount := len(eventsToTransform)
 	proc.logger.Debug("Supported messages filtering input size", eventFilterInCount)
 	response = ConvertToFilteredTransformerResponse(
 		eventsToTransform,
-		transformAt != "none",
```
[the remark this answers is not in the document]
cisse21: transformAt is only applicable at destination transformation level and not at event filter level

processor/processor.go
cisse21: I felt so.. It doesn't look like the successCountMetadataMap was generated from dest transform call but to maintain the old behaviour we can add it to the struct

processor/processor.go
```diff
@@ -2935,57 +3021,97 @@ func (proc *Handle) transformSrcDest(
...
 		}
 	}
 
+	response := data.response
+	eventsByMessageID := data.eventsByMessageID
+	sourceID, destID := getSourceAndDestIDsFromKey(srcAndDestKey)
+	sourceName := data.eventsToTransform[0].Metadata.SourceName
+	destination := &data.eventsToTransform[0].Destination
+	workspaceID := data.eventsToTransform[0].Metadata.WorkspaceID
+	destType := destination.DestinationDefinition.Name
```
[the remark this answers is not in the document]
atzoum: Isn't it safer? Since after user transform, metadata will be what we get back from transformer, right?

processor/processor.go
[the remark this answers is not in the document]
atzoum: if we are always using `true`, then what's the point of having it as a function argument?

processor/processor.go
```diff
@@ -2955,37 +3084,37 @@ func (proc *Handle) transformSrcDest(
...
 			}
-			proc.handleWarehouseTransformations(ctx, eventsToTransform, response, commonMetaData, eventsByMessageID)
+			proc.handleWarehouseTransformations(ctx, data.eventsToTransform, response, data.commonMetaData, eventsByMessageID)
 
 			destTransformationStat := proc.newDestinationTransformationStat(sourceID, workspaceID, transformAt, destination)
 			destTransformationStat.transformTime.Since(s)
 			transformAt = "processor"
 
-			proc.logger.Debugf("Dest Transform output size %d", len(response.Events))
+			proc.logger.Infof("Dest Transform output size %d", len(response.Events))
```
atzoum, line 3101:
```suggestion
			proc.logger.Debugf("Dest Transform output size %d", len(response.Events))
```

processor/processor.go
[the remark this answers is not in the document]
cisse21: I used them from common metadata rather than from data

processor/processor.go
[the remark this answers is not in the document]
cisse21: I don't want to mess up any behaviour so keeping things as is. We can clean up later if this is no longer relevant

atzoum (approved):

PR #5644, opened 2025-03-26
refactor: version deprecation detection logic

koladilip:
# Description
These combinations help to detect deprecated versions and also reduce the noise.

## Linear Ticket

https://linear.app/rudderstack/issue/INT-1577/[fix]-version-deprecation-error-logic
Resolves INT-1577/

## Security

### DIFF
--- a/enterprise/reporting/error_extractor.go
+++ b/enterprise/reporting/error_extractor.go
@@ -3,6 +3,7 @@ package reporting
 import (
 	"fmt"
 	"regexp"
+	"slices"
 	"strings"
 
 	"github.com/samber/lo"
@@ -14,21 +15,6 @@ import (
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-server/jsonrs"
 )
-
-// Generate strong deprecation terms by combining "deprecated" with version patterns
-func generateStrongDeprecationTerms(keyPatterns, phrasePatterns []string) []string {
-	result := make([]string, 0, len(keyPatterns)+len(phrasePatterns))
-
-	// Add all standalone strong terms
-	result = append(result, phrasePatterns...)
-
-	// Add "deprecated X" for each version pattern
-	for _, v := range keyPatterns {
-		result = append(result, "deprecated "+v)
-	}
-
-	return result
-}
 
 const (
 	responseKey = "response"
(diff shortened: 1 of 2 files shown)


enterprise/reporting/error_extractor.go
```diff
@@ -345,32 +353,18 @@ func getErrorCodeFromStatTags(statTags map[string]string) string {
...
-	}
+func containsDeprecationKey(errorMessage, key string) bool {
+	return strings.HasPrefix(errorMessage, key) || strings.Contains(errorMessage, " "+key)
+}
 
-	return false
+func (ext *ExtractorHandle) isVersionDeprecationError(errorMessage string) bool {
+	// Normalize error message
+	cleanedError := strings.ReplaceAll(strings.ToLower(errorMessage), "-", " ")
+	return slices.ContainsFunc(deprecationKeywordSets, func(keywordSet []string) bool {
```
ktgowtham, line 363: keywords like version, api, endpoint are repeated and if I'm not wrong, we're doing string comparison multiple times. Can this be improved by doing this in two passes? one for repeated keywords and then unique keywords?

enterprise/reporting/error_extractor.go
```diff
@@ -345,32 +353,18 @@ func getErrorCodeFromStatTags(statTags map[string]string) string {
...
-	}
-
-	// Check for strong deprecation terms (using pre-computed list)
-	for _, term := range strongDeprecationTerms {
-		if strings.Contains(cleanedError, term) {
-			return true
-		}
-	}
+func containsDeprecationKey(errorMessage, key string) bool {
+	return strings.HasPrefix(errorMessage, key) || strings.Contains(errorMessage, " "+key)
```
ktgowtham, line 357: Why can't we just search contains? what are the odds that the keyword is a substring of something else?

koladilip: it is quite high, I have come up with only after running against current errors

enterprise/reporting/error_extractor.go
koladilip: updated

ktgowtham (approved):

enterprise/reporting/error_extractor.go
```diff
@@ -357,16 +363,28 @@ func containsDeprecationKey(errorMessage, key string) bool {
...
+func (ext *ExtractorHandle) isVersionDeprecationError(errorMessage string) bool {
+	// Normalize error message
+	cleanedError := strings.ReplaceAll(strings.ToLower(errorMessage), "-", " ")
+	for key, keywordSets := range deprecationKeywordSets {
+		if !containsDeprecationKey(cleanedError, key) {
+			continue
+		}
+		if containsAllKeywords(cleanedError, keywordSets) {
+			return true
+		}
```
satishrudderstack, line 383: [NItpick] How frequently can `containsAllKeyworkds` run ? We are searching for individual patterns in loop. This can be costly. We should have a look at pyroscope cpu profiles after release

FYI, we are considering Aho-corasick for bot pattern matching which can be used here too.

satishrudderstack (approved):

vinayteki95 (approved):

enterprise/reporting/error_extractor.go
koladilip: This is good suggestion but I need to match if word is at the start of message or it have space before, that is hard to do using Aho-corasick matcher

PR #5652, opened 2025-03-26
chore: increase max idle connections per host for kinesis

vinayteki95:
# Description

Kinesis is causing high CPU usage due to high number of TLS handshakes happening. Solution is to increase number of connections that can stay idle leaving room for not killing the connections once done which will reduce TLS connections.
To begin with we are increasing the number of connections from 2 to 64.

## Linear Ticket

Resolves INT-3397
https://linear.app/rudderstack/issue/INT-3397/evaluate-workarounds-for-high-tls-connections-in-expressvpn-due-to

## Security


vinayteki95:
Waiting for go-kit release. Once that is done, will include latest version of it in go.mod to resolve these errors.

services/streammanager/kinesis/kinesismanager.go
```diff
@@ -44,6 +44,9 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (*Kines
 	if err != nil {
 		return nil, err
 	}
+	// Given we use this for only streaming, we can set a high number of idle connections (default is 2)
+	sessionConfig.MaxIdleConnsPerHost = 50
```
atzoum, line 48: expressvpn has more than 50 kinesis destinations, so it is still not enough for them. I see two options:

- Set this value to be equal to the number of destinations
- Make it configurable, so that we can tune it for different customers.

atzoum, line 48: This should be equal to the number of kinesis workers:
```go
config.GetIntVar(64, 1, "Router.KINESIS.noOfWorkers", "Router.noOfWorkers")
```

atzoum: `MaxConnsPerHost` should be adjusted as well based on number of workers

go.mod
```diff
@@ -78,7 +78,7 @@ require (
 	github.com/rudderlabs/analytics-go v3.3.3+incompatible
 	github.com/rudderlabs/bing-ads-go-sdk v0.2.3
 	github.com/rudderlabs/compose-test v0.1.3
-	github.com/rudderlabs/rudder-go-kit v0.48.0
+	github.com/rudderlabs/rudder-go-kit v0.48.1-0.20250327144458-e24b01c6209e
```
atzoum, line 81: let's use the [0.48.1 release](https://github.com/rudderlabs/rudder-go-kit/pull/764)

atzoum (approved):

mihir20 (approved):

PR #5658, opened 2025-03-27
fix: warehouse transformations for mandatory fields

achettyiitr:
# Description

Some mandatory fields like `messageId` and `receivedAt` are sometimes missing or empty, where events are added during the user transformation stage. Since the rudder-server only guarantees these fields at the gateway level, they may be absent when events reach destination transformation.

This update ensures that:
-  `messageId` is assigned a unique value if missing.
-  `receivedAt` is validated and set appropriately, falling back to **event.Metadata.ReceivedAt** or the **current timestamp** if needed.

This follows the same approach as in [rudder-transformer](https://github.com/rudderlabs/rudder-transformer/blob/develop/src/warehouse/index.js#L675-L688).

## Linear Ticket

- Resolves WAR-442

## Security

### DIFF
--- a/warehouse/transformer/events_test.go
+++ b/warehouse/transformer/events_test.go
@@ -3,6 +3,7 @@ package transformer
 import (
 	"context"
 	"net/http"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


warehouse/transformer/transformer.go
```diff
@@ -145,6 +149,21 @@ func (t *Transformer) enhanceContextWithSourceDestInfo(event *types.TransformerE
...
+	}
+	if receivedAt, exists := event.Message["receivedAt"]; !exists || utils.IsBlank(receivedAt) {
+		if strReceivedAt, isString := receivedAt.(string); !isString || !utils.ValidTimestamp(strReceivedAt) {
+			if !utils.ValidTimestamp(event.Metadata.ReceivedAt) {
+				event.Metadata.ReceivedAt = t.now().Format(misc.RFC3339Milli)
+			}
+			event.Message["receivedAt"] = event.Metadata.ReceivedAt
+		}
+	}
+}
```
RanjeetMishra, line 165: 1.  Say if `event.Message["receivedAt"]` does not exist, what would `receivedAt.(string)` do?
2. Do you think we can check for both `event.Message["receivedAt"]` and `event.Metadata.ReceivedAt` for value as well as validTimeStamp and if either is invalid/empty/null, we set both?

achettyiitr: > Say if event.Message["receivedAt"] does not exist, what would receivedAt.(string) do?

- `isString` would be false.

> Do you think we can check for both event.Message["receivedAt"] and event.Metadata.ReceivedAt for value as well as validTimeStamp and if either is invalid/empty/null, we set both?

- Since `event.Metadata.ReceivedAt` is a struct, it will never be nil. It can only be an empty string or an invalid value. For that, we are already checking `utils.ValidTimestamp(event.Metadata.ReceivedAt)` and updating if necessary.
- `event.Message["receivedAt"]` can be invalid/empty/null. In case of nil or invalid type `isString` will be false. In case of an invalid value or empty value, `ValidTimestamp` will be false. We are overriding it with `event.Metadata.ReceivedAt`.


Simplified the logic.
```GO
func (t *Transformer) addMandatoryFields(event *types.TransformerEvent) {
	t.ensureMessageID(event)
	t.ensureReceivedAt(event)
}

func (t *Transformer) ensureMessageID(event *types.TransformerEvent) {
	messageID, exists := event.Message["messageId"]
	if !exists || utils.IsBlank(messageID) {
		event.Metadata.MessageID = "auto-" + t.uuidGenerator()
		event.Message["messageId"] = event.Metadata.MessageID
	}
}

func (t *Transformer) ensureReceivedAt(event *types.TransformerEvent) {
	receivedAt, exists := event.Message["receivedAt"]
	if !exists || utils.IsBlank(receivedAt) {
		t.setDefaultReceivedAt(event)
		return
	}

	strReceivedAt, isString := receivedAt.(string)
	if !isString || !utils.ValidTimestamp(strReceivedAt) {
		t.setDefaultReceivedAt(event)
	}
}

func (t *Transformer) setDefaultReceivedAt(event *types.TransformerEvent) {
	if !utils.ValidTimestamp(event.Metadata.ReceivedAt) {
		event.Metadata.ReceivedAt = t.now().Format(misc.RFC3339Milli)
	}
	event.Message["receivedAt"] = event.Metadata.ReceivedAt
}
```

warehouse/transformer/transformer.go
```diff
@@ -145,6 +149,39 @@ func (t *Transformer) enhanceContextWithSourceDestInfo(event *types.TransformerE
...
 
+func (t *Transformer) addMandatoryFields(event *types.TransformerEvent) {
+	t.ensureMessageID(event)
+	t.ensureReceivedAt(event)
+}
+
+func (t *Transformer) ensureMessageID(event *types.TransformerEvent) {
+	messageID, exists := event.Message["messageId"]
+	if !exists || utils.IsBlank(messageID) {
+		event.Metadata.MessageID = "auto-" + t.uuidGenerator()
```
shekhar-rudder, line 160: Is it possible for `event.Metadata.MessageID` to be present but `event.Message["messageId"]` is empty?

warehouse/transformer/types.go
```diff
@@ -13,7 +13,8 @@ import (
 
 type (
 	Transformer struct {
-		now func() time.Time
+		now           func() time.Time
+		uuidGenerator func() string
```
shekhar-rudder, line 17: uuidGenerator -> generateUuid ?

warehouse/transformer/events_test.go
```diff
@@ -19,6 +25,103 @@ import (
...
+			events := []types.TransformerEvent{{Message: singularEvent, Metadata: tc.metadata, Destination: tc.destination}}
+
+			pResponse := processorTransformer.Transform(ctx, events, batchSize)
+			wResponse := warehouseTransformer.Transform(ctx, events, batchSize)
+
+			require.Equal(t, len(wResponse.Events), len(pResponse.Events))
+			require.Nil(t, wResponse.FailedEvents)
+			require.Nil(t, pResponse.FailedEvents)
+			for i := range wResponse.Events {
+				require.NotEmpty(t, misc.MapLookup(wResponse.Events[i].Output, "data", "id"))
```
shekhar-rudder, line 115: Why are we checking for data.id ? Are we converting message.messageId to data.id ?

warehouse/transformer/events_test.go
```diff
@@ -19,6 +25,103 @@ import (
...
+			eventPayload: `{"type":"track","anonymousId":"anonymousId","userId":"userId","sentAt":"2021-09-01T00:00:00.000Z","timestamp":"2021-09-01T00:00:00.000Z","originalTimestamp":"2021-09-01T00:00:00.000Z","channel":"web","event":"event","request_ip":"5.6.7.8","properties":{"review_id":"86ac1cd43","product_id":"9578257311"},"userProperties":{"rating":3.0,"review_body":"OK for the price. It works but the material feels flimsy."},"context":{"traits":{"name":"Richard Hendricks","email":"rhedricks@example.com","logins":2},"ip":"1.2.3.4"}}`,
+			metadata:     types.Metadata{EventType: "track", DestinationType: "POSTGRES"},
+			destination:  getDestination("POSTGRES", map[string]any{}),
+			verifyResponse: func(t *testing.T, resp types.TransformerResponse) {
+				require.Equal(t, "auto-"+uid, misc.MapLookup(resp.Output, "data", "id"))
+				require.Equal(t, now.Format(misc.RFC3339Milli), misc.MapLookup(resp.Output, "data", "received_at"))
+			},
+		},
+		{
+			name:         "messageId and receivedAt both not empty",
```
shekhar-rudder, line 66: `messageId and receivedAt both empty`

warehouse/transformer/events_test.go
achettyiitr: Yes.

warehouse/transformer/types.go
achettyiitr: `t.generateUuid()` sounds more preferable for a standalone function, not a functional field. Since it's being injected into a struct for customization or testing, I would prefer it to be a noun rather than a verb.

warehouse/transformer/transformer.go
achettyiitr: No. `event.Metadata.MessageID` is basically a derived field from SingularEventT.

RanjeetMishra (approved):

PR #5661, opened 2025-03-27
refactor: remove unused schema v1 code

shekhar-rudder:
# Description

Now that Schema V2 has been fully rolled out to all customers in production, this PR removes the legacy Schema V1 implementation.

**Changes:**
- Removed all code and tests related to the `schema` struct.
- Migrated common methods/interfaces from Schema V1 to the Schema V2 file.
- Copied `ConsolidateStagingFiles` tests from v1 to v2

**Next Steps:**
In the next PR, the client code will be modified to stop fetching the schema directly from the warehouse and instead use the cached schema.

## Security


warehouse/schema/schema_v2.go
achettyiitr, line 1: Shall we rename this file to schema instead of schema_v2. Also for structs as well and other places where v2 is being used?

achettyiitr, line 1: Can we also add the necessary comments for the exported functions which were there in `schema.go`?

shekhar-rudder: Renamed `schema_v2` to `schema`

shekhar-rudder: Added comments for all the methods in the interface

achettyiitr (approved):

vyeshwanth (approved):

PR #5662, opened 2025-03-28
fix: warehouse transformations for tracking plans

achettyiitr:
# Description

The [Tracking Plan properties](https://github.com/rudderlabs/rudder-server/blob/master/processor/trackingplan.go#L28-L51) in the Go code differ in type from the response received from `rudder-transformer`, causing a type mismatch:
- **`context_tracking_plan_version`**:  
  - Unmarshalling response from `rudder-transformer` treats all numbers as `json.Number` (essentially `float64`).  
  - Go code sets it as `int`.
- **`context_violation_errors`**:  
  - Unmarshalling response from `rudder-transformer` returns it as `[]any`.  
  - Go code sets it as `[]types.ValidationError`.
- This results in a type mismatch, though the actual data remains correct.

 ```GO
  - 			"context_tracking_plan_version": int(8),
  + 			"context_tracking_plan_version": float64(8),
  - 			"context_violation_errors": []types.ValidationError{
  - 				{
  - 					Type:     "Required-Missing",
  - 					Message:  "must have required property 'name'",
  - 					Meta:     map[string]string{"instancePath": "/traits", "schemaPath": "#/properties/traits/required"},
  - 					Property: "name",
  - 				},
  - 				{
  - 					Type:    "Datatype-Mismatch",
  - 					Message: "must be string",
  - 					Meta: map[string]string{
  - 						"instancePath": "/traits/email",
  - 						"schemaPath":   "#/properties/traits/properties/e"...,
  - 					},
  - 				},
  - 			},
  + 			"context_violation_errors": []any{
  + 				map[string]any{
  + 					"message": string("must have required property 'name'"),
  + 					"meta": map[string]any{
  + 						"instancePath": string("/traits"),
  + 						"schemaPath":   string("#/properties/traits/required"),
  + 					},
  + 					"property": string("name"),
  + 					"type":     string("Required-Missing"),
  + 				},
  + 				map[string]any{
  + 					"message": string("must be string"),
  + 					"meta": map[string]any{
  + 						"instancePath": string("/traits/email"),
  + 						"schemaPath":   string("#/properties/traits/properties/e"...),
  + 					},
  + 					"property": string(""),
  + 					"type":     string("Datatype-Mismatch"),
  + 				},
  + 			},
  ```

## Linear Ticket

- Resolves WAR-424

## Security

### DIFF
--- a/warehouse/transformer/datatype.go
+++ b/warehouse/transformer/datatype.go
@@ -3,7 +3,10 @@ package transformer
 import (
 	"reflect"
 
+	"github.com/samber/lo"
+
 	"github.com/rudderlabs/rudder-server/jsonrs"
+	"github.com/rudderlabs/rudder-server/processor/types"
 	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
 	"github.com/rudderlabs/rudder-server/warehouse/transformer/internal/utils"
 	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
@@ -90,6 +93,48 @@ func overrideForRedshift(val any, isJSONKey bool) string {
 func convertValIfDateTime(val any, colType string) any {
 	if colType == model.DateTimeDataType {
 		return utils.ToTimestamp(val)
+	}
+	return val
+}
+
+func convertToFloat64IfInteger(val any) any {
+	switch v := val.(type) {
+	case int:
+		return float64(v)
+	case int8:
+		return float64(v)
+	case int16:
+		return float64(v)
+	case int32:
+		return float64(v)
+	case int64:
+		return float64(v)
+	case uint:
+		return float64(v)
+	case uint8:
+		return float64(v)
+	case uint16:
+		return float64(v)
+	case uint32:
+		return float64(v)
+	case uint64:
+		return float64(v)
+	}
+	return val
+}
+
+func convertToSliceIfViolationErrors(val any) any {
+	if validationErrors, ok := val.([]types.ValidationError); ok {
+		result := make([]any, len(validationErrors))
+		for i, e := range validationErrors {
+			result[i] = map[string]any{
+				"type":     e.Type,
+				"message":  e.Message,
+				"meta":     lo.MapValues(e.Meta, func(value, _ string) any { return value }),
+				"property": e.Property,
+			}
+		}
+		return result
 	}
 	return val
 }
--- a/warehouse/transformer/transformer.go
+++ b/warehouse/transformer/transformer.go
@@ -29,6 +29,14 @@ const (
 	redshiftStringLimit = 512
 )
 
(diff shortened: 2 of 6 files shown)


warehouse/transformer/datatype.go
```diff
@@ -93,3 +96,45 @@ func convertValIfDateTime(val any, colType string) any {
...
+}
+
+func convertToSliceIfViolationErrors(val any) any {
+	if validationErrors, ok := val.([]types.ValidationError); ok {
+		result := make([]any, len(validationErrors))
+		for i, e := range validationErrors {
+			result[i] = map[string]any{
+				"type":     e.Type,
+				"message":  e.Message,
+				"meta":     lo.MapValues(e.Meta, func(value, _ string) any { return value }),
```
shekhar-rudder, line 133: Why not assign `e.Meta` directly instead of using `MapValues` ? Is it to create a new map ?

warehouse/transformer/transformer.go
```diff
@@ -27,6 +27,14 @@ const (
 	redshiftStringLimit = 512
 )
 
+// Compile-time check to ensure ValidationError struct remains unchanged.
```
shekhar-rudder, line 30: Are there other structs for which we should add this check ?

warehouse/transformer/datatype.go
achettyiitr: Meta is map[string]string. We need map[string]any.

warehouse/transformer/transformer.go
achettyiitr: No. Only this is sufficient.

shekhar-rudder (approved):

vyeshwanth (approved):

warehouse/transformer/transformer.go
```diff
@@ -29,6 +29,14 @@ const (
...
+var _ = struct {
+	Type     string
+	Message  string
+	Meta     map[string]string
+	Property string
+}(types.ValidationError{})
```
vyeshwanth, line 38: why such compile-time check is necessary for this struct?

achettyiitr: Currently, we are relying on this struct to do the conversion from map[string]string to map[string]any. In case if something changes in future, we need to aware of it like they added new fields or modified some fields. If we don't do that it would break.
```GO
func convertToSliceIfViolationErrors(val any) any {
	if validationErrors, ok := val.([]types.ValidationError); ok {
		result := make([]any, len(validationErrors))
		for i, e := range validationErrors {
			result[i] = map[string]any{
				"type":     e.Type,
				"message":  e.Message,
				"meta":     lo.MapValues(e.Meta, func(value, _ string) any { return value }),
				"property": e.Property,
			}
		}
		return result
	}
	return val
}
```

PR #5669, opened 2025-03-31
feat: introduce common http client for reporting

lvrach:
# Description

## Motivation

To introduce a cost attribution metric for all reporting services. I've noticed coupling and discrepancies between different client implementations, making it hard to test the changes and requiring redundant effort. 

First, I attempted a more [aggressive](https://github.com/rudderlabs/rudder-server/pull/5529) refactor effort. 

However, after discussing this with the PipeOps team, we decided to take a more cynical approach.

## Plan 

* Introduce a common client for all communication with the reporting service
* Support switching to the new common client via a single feature flag
* Intention to remove old client/send method in the next release.

It was also decided
* To break the backward compatibility with metrics exposed by old clients. This might break some dashboards, but no alert should be impacted.

### Metric names
New metrics:

```go
const (
	StatRequestTotalBytes     = "reporting_client_http_request_total_bytes"
	StatTotalDurationsSeconds = "reporting_client_http_total_durations_seconds"
	StatRequestLatency        = "reporting_client_http_request_latency"
	StatHttpRequest           = "reporting_client_http_request"
)
```

Deprecated metrics -- will no longer work after switching to the common client:

```go
	f.reqLatency = f.stats.NewTaggedStat("reporting_flusher_http_request_duration_seconds", stats.TimerType, tags)
	f.reqCount = f.stats.NewTaggedStat("reporting_flusher_http_requests_total", stats.CountType, tags)
	f.sentBytes = f.stats.NewTaggedStat("reporting_flusher_sent_bytes", stats.HistogramType, tags)
	
        edr.stats.NewTaggedStat("error_detail_reporting_http_request", stats.CountType, httpStatTags).Increment()
	edr.edReportingRequestLatency = edr.stats.NewTaggedStat("error_detail_reporting_request_latency", stats.TimerType, tags)
```
### Metric labels 

Changes:

* `workspaceID` -- REMOVED will no longer be captured by the client, as it is not relevant when multiple workspaces are batch to a single call 
* `clientName` -- RENAMED --> `module`
*  `endpoint` --> ADDED: reporting URL 
* `path` --> ADDED: reporting path/service

### `region` query param 

Region query parameter, used only by some reporting clients **has been removed**.

Based on team feedback this query param was used temporary for multi region transition and is no longer needed.

### `REPORTING_URL` env var 

REPORTING_URL now always defaults to `"https://reporting.dev.rudderlabs.com"` 

Some of the existing clients defaulted to `https://reporting.rudderstack.com/`

## To Be Discussed

* Should we use a default value for `REPORTING_URL`, would we rather fail if `REPORTING_URL` is not configured ? It could be a privacy risk for EU customers if reporting us endpoint is configured accidentally.
* Should we introduce `.Ping()` method and corresponding endpoint ? 

## Next steps

1. Rollout `useCommonClient` flag in production
2. Update Grafana dashboards as need
3. Remove all deprecated code (most of it is marked with commends`
4. Start the work for unifying reporting logic

## Linear Ticket

Resolves PRI-168

https://linear.app/rudderstack/issue/PRI-168/introduce-reporting-client-metrics-in-rudder-server-for-cost

## Security


enterprise/reporting/reporting.go
```diff
@@ -80,6 +81,9 @@ type DefaultReporter struct {
 	eventSamplingEnabled  config.ValueLoader[bool]
 	eventSamplingDuration config.ValueLoader[time.Duration]
 	eventSampler          event_sampler.EventSampler
+
+	useCommonClient *config.Reloadable[bool]
```
atzoum, line 85: prefer interfaces
```suggestion
	useCommonClient config.ValueLoader[bool]
```

enterprise/reporting/client/client.go
```diff
@@ -0,0 +1,147 @@
...
+	stats               stats.Stats
+	log                 logger.Logger
+	instanceID          string
+	path                Path
+}
+
+// New creates a new reporting client
+func New(reportingServiceURL string, path Path, conf *config.Config, log logger.Logger, stats stats.Stats) *Client {
+	reportingServiceURL = strings.TrimSuffix(reportingServiceURL, "/")
+	tr := &http.Transport{}
```
atzoum, line 56: why don't we use the default transport, or even provide options for configuration overrides?

lvrach: Good point, I started this client by copying on of the reporting.Send implementations.

Now that I see this again, it doesn't make sense to do it. I'will remove it

lvrach:
@coderabbitai full review

itsmihir (approved):
LGTM, please address the issue where 4xx status codes are being treated as successful responses.

enterprise/reporting/client/client.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+}
+
+func (f *Client) isHTTPRequestSuccessful(status int) bool {
+	if status == 429 {
+		return false
+	}
+
+	return status >= 200 && status < 500
+}
```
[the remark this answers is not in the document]
lvrach: Interestingly, the code started before this PR was introduced https://github.com/rudderlabs/rudder-server/pull/5537. Copying the logic before that.

Apply the change and add a test for it.

enterprise/reporting/client/client.go
```diff
@@ -0,0 +1,161 @@
...
+	return b
+}
+
+// New creates a new reporting client
+func New(path Path, conf *config.Config, log logger.Logger, stats stats.Stats) *Client {
+	reportingServiceURL := conf.GetString("REPORTING_URL", "https://reporting.dev.rudderlabs.com")
+	reportingServiceURL = strings.TrimSuffix(reportingServiceURL, "/")
+
+	return &Client{
+		httpClient: &http.Client{
```
atzoum, line 76: one reason why we would need to override the default transport is for tuning `MaxIdleConnsPerHost` which by default is `2`. This client is going to open connections to a single host for multiple components, correct?

lvrach: Here are the differences between: `http.DefaultTransport` (default) and `&Transport{}` (zero)

| Property| Default  | zero |
| -------- | --------| ------- |
| Proxy     | useEnv | no proxy|
|DialContext.Timeout | 30s | no timeout|
|DialContext.KeepAlive| 30s | 15s |
|ForceAttemptHTTP2| true| false |
|MaxIdleConns| 100| no limit|
| IdleConnTimeout | 90s | no timeout |
| TLSHandshakeTimeout| 10s | no timeout|
|ExpectContinueTimeout| 10s | no timeout|

When it comes to `MaxIdleConnsPerHost` the both have the same behaviour, given the client only connected to a single host. MaxIdleConnsPerHost should effectively overrule MaxIdleConns when it comes to idle connections.

Examining the differences between them. I believe that using zero transport, is almost always a bad idea. 

If we need to change some of the settings is better to do:

```go
transport := http.DefaultTransport.(*http.Transport).Clone()
transport.MaxIdleConnsPerHost = 10
&http.Client{Transport: transport}
```

It remains a question to me, what was the actual intention behind the following :
```go
	tr := &http.Transport{}
	netClient := &http.Client{Transport: tr, Timeout: config.GetDuration("HttpClient.reporting.timeout", 60, time.Second)}
```
As nothing is override on `Transport`

lvrach: > This client is going to open connections to a single host for multiple components, correct?

A single host is going to be used for all reporting-related communication.

Three instances of`reporting/client.Client` going to be used. Depending on the component, a different number of concurrent requests is expected, up to `Reporting.maxConcurrentRequests` (32 by default)

* reporting sends an HTTP request per [reportIdentifier](https://github.com/rudderlabs/rudder-server/blob/63ff9d502ef5e4db7ff32b264ca72635be9bb832/enterprise/reporting/reporting.go#L334) 
* error report an HTTP request per [group](https://github.com/rudderlabs/rudder-server/blob/63ff9d502ef5e4db7ff32b264ca72635be9bb832/enterprise/reporting/error_reporting.go#L676)
* the new flusher is [batching](https://github.com/rudderlabs/rudder-server/blob/63ff9d502ef5e4db7ff32b264ca72635be9bb832/enterprise/reporting/flusher/flusher.go#L313) the grouped keys 

Overall, it looks like many (30) connections are open, due to concurrency and then closed due to MaxIdleConnsPerHost.

lvrach: I find the change of http.DefaultTransport type.

Yes, we cannot exclude the probability of http.DefaultTransport changing. I trust the go dev team would be careful with the changes, most probably the change would be for better rather than worse.

For instance, the might introduce a method and add e sensible default value. You might missed it if you create your own every time.

I would be open to creating our default transport in rudder go kit and have our http client clone and use that.

satishrudderstack (approved):

PR #5675, opened 2025-04-01
chore: datetime detection

achettyiitr:
# Description

- Use some valid timestamp formats for datetime detection instead of `dateparse.ParseAny`.

## Linear Ticket

- Resolves WAR-403

## Security


warehouse/transformer/internal/utils/utils_test.go
```diff
@@ -83,6 +84,30 @@ func TestValidTimestamp(t *testing.T) {
 	}
 }
 
+// BenchmarkValidTimestamp/ValidTimestamp_Valid
+// BenchmarkValidTimestamp/ValidTimestamp_Valid-12         				34838106	 	31.71 ns/op
+// BenchmarkValidTimestamp/ValidTimestamp_Invalid
+// BenchmarkValidTimestamp/ValidTimestamp_Invalid-12       	 			2619600	       	430.6 ns/op
+// BenchmarkValidTimestamp/ValidTimestamp_Invalid_Big_String
+// BenchmarkValidTimestamp/ValidTimestamp_Invalid_Big_String-12        	7234989	       	178.7 ns/op
```
shekhar-rudder, line 92: Why does big string has a lower `ns/op` compared to the invalid case?

achettyiitr: It's because of this check.
```
        if len(input) > validTimestampFormatsMaxLength {
		return false
	}
```

shekhar-rudder (approved):

PR #5692, opened 2025-04-04
fix: snowflake fetch schema privileges

achettyiitr:
# Description

- Use Show Grants to validate the fetch schema step in warehouse destination validation for **SNOWFLAKE**. 
- It's `Warehouse.snowflake.privileges.fetchSchema.enabled` config flag driven. By default it's false. If disabled, use the previous implementation to query **INFORMATION_SCHEMA.COLUMNS**.
- If the configured role in the destination configuration page is empty, use **SHOW GRANTS TO USER** to get the list of roles granted to the user.
- Using **SHOW GRANTS ON SCHEMA** and then filtering for the above roles to check if it has **USAGE** privilege.

## Linear Ticket

- Resolves WAR-485

## Security


warehouse/integrations/snowflake/snowflake.go
```diff
@@ -1446,8 +1458,180 @@ func (sf *Snowflake) LoadTestTable(
...
+	}
+	defer func() { _ = rows.Close() }()
+
+	columns, err := rows.Columns()
+	if err != nil {
+		return nil, fmt.Errorf("getting columns: %w", err)
+	}
+
+	columnIndices := map[string]int{}
+	requiredColumns := [4]string{"name", "privilege", "granted_on", "grantee_name"}
```
RanjeetMishra, line 1575: `granted_on` ... would it be useful for any debugging? Asking if we need to read this, or can ignore.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -1446,8 +1458,180 @@ func (sf *Snowflake) LoadTestTable(
...
+		}
+	}
+	for _, required := range requiredColumns {
+		if _, found := columnIndices[required]; !found {
+			return nil, fmt.Errorf(`required columns %v not found in result: %v`, requiredColumns, columns)
+		}
+	}
+
+	var privileges []privilegeGrant
+	for rows.Next() {
```
RanjeetMishra, line 1592: Row reading code, is there scope of reusing common code for `getUserRoles` and `getShowGrantsPrivileges`?

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -1446,8 +1458,180 @@ func (sf *Snowflake) LoadTestTable(
...
+	if !sf.config.privileges.fetchSchema.enabled {
+		_, err := sf.FetchSchema(ctx)
+		return err
+	}
+
+	var roles []string
+	var err error
+
+	role := sf.Warehouse.GetStringDestinationConfig(sf.conf, model.RoleSetting)
+	if len(role) == 0 {
```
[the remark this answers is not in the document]
achettyiitr: Since the role is not configured, it might be possible that the user is assigned to multiple roles. The default role might not have privileges. We would need to get a list of all the roles assigned to that user.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -1446,8 +1458,180 @@ func (sf *Snowflake) LoadTestTable(
...
+	})
+	if len(missingPrivileges) > 0 {
+		return fmt.Errorf("missing privileges: %v", missingPrivileges)
+	}
+	return nil
+}
+
+func (sf *Snowflake) getUserRoles(ctx context.Context) ([]string, error) {
+	user := sf.Warehouse.GetStringDestinationConfig(sf.conf, model.UserSetting)
+	sqlStatement := fmt.Sprintf("SHOW GRANTS TO USER %q;", user)
```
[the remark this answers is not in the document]
achettyiitr: Roles are case-sensitive. We need double quotes otherwise it will be treated as caps.
https://docs.snowflake.com/en/sql-reference/sql/create-user#required-parameters

warehouse/integrations/snowflake/snowflake.go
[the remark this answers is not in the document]
RanjeetMishra: When we connect with snowflake, we do not pass role explicitly is it? As far as I remember, if no role passed, Snowflake will connect or work with default role. If no default role, then I am not sure what happens, but I am sure Snowflake would have default role set as long as there is at least 1 role associated to user.
Will ping, we can discuss a bit on this.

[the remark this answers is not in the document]
achettyiitr:
```
CREATE USER TEST_DEFAULT_ROLE_1 MUST_CHANGE_PASSWORD = FALSE PASSWORD = "<your_password>";
GRANT ROLE "TEST_ROLE_1" TO USER TEST_DEFAULT_ROLE_1;
DESCRIBE USER TEST_DEFAULT_ROLE_1;
```
If no default role is set and we assign it to that user, it still comes as null.

Screenshot 2025-04-04 at 2 54 07 PM

warehouse/integrations/snowflake/snowflake.go
achettyiitr: Currently, It's being used in tests for some verification.

warehouse/integrations/snowflake/snowflake.go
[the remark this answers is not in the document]
RanjeetMishra: Got it... explicit user and role from our config could have helped us with less lines of code. Lets go with this.

Do you think it makes sense to keep a tech debt task to revisit and make role mandatory in config?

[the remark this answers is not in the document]
achettyiitr: > Do you think it makes sense to keep a tech debt task to revisit and make role mandatory in config?

Yeah, let's add a taste to revisit this. Although things have worked out till now without roles. Don't think this just for this we need to make it mandatory?

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -1446,8 +1460,179 @@ func (sf *Snowflake) LoadTestTable(
...
+	if err != nil {
+		return fmt.Errorf("getting show grants privileges: %w", err)
+	}
+	filteredPrivilegesMap := lo.FilterSliceToMap(privileges, func(item privilegeGrant) (string, bool, bool) {
+		if _, ok := rolesSet[item.granteeName]; !ok {
+			return "", false, false
+		}
+		return item.privilege, true, true
+	})
+	if hasOwnershipPrivileges := filteredPrivilegesMap["OWNERSHIP"]; hasOwnershipPrivileges {
```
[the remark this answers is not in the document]
achettyiitr: Not required. Privileges will always come in CAPS.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -181,6 +193,8 @@ func New(conf *config.Config, log logger.Logger, stat stats.Stats) *Snowflake {
 	sf.config.debugDuplicateTables = lo.Map(conf.GetStringSlice("Warehouse.snowflake.debugDuplicateTables", nil), func(item string, index int) string {
 		return strings.ToUpper(item)
 	})
+	sf.config.privileges.fetchSchema.required = conf.GetStringSlice("Warehouse.snowflake.privileges.fetchSchema.required", []string{"USAGE"})
```
vyeshwanth, line 196: are we looking to override this config? Just want to understand why this is a slice. I am assuming for the current snowflake implementation, `USAGE` is what we need

achettyiitr: Yeah for now USAGE is sufficient. Made it configurable, just in case if it is needed.

warehouse/integrations/snowflake/privileges_test.go
```diff
@@ -0,0 +1,261 @@
...
+			}, privileges)
+			schemaPrivileges, err := sf.getShowGrantsPrivileges(ctx, fmt.Sprintf(`SHOW GRANTS ON SCHEMA %q;`, namespace))
+			require.NoError(t, err)
+			require.ElementsMatch(t, []privilegeGrant{
+				{"WAREHOUSE_INTEGRATION_TESTS.FETCH_SCHEMA_TEST", "USAGE", "SCHEMA", "FETCH_SCHEMA_TEST_ROLE_2"},
+				{"WAREHOUSE_INTEGRATION_TESTS.FETCH_SCHEMA_TEST", "MONITOR", "SCHEMA", "FETCH_SCHEMA_TEST_ROLE_4"},
+				{"WAREHOUSE_INTEGRATION_TESTS.FETCH_SCHEMA_TEST", "MONITOR", "SCHEMA", "FETCH_SCHEMA_TEST_ROLE_5"},
+				{"WAREHOUSE_INTEGRATION_TESTS.FETCH_SCHEMA_TEST", "OWNERSHIP", "SCHEMA", "FETCH_SCHEMA_TEST_ROLE_6"},
+			}, schemaPrivileges)
+			t.Log(sf.TestFetchSchema(ctx))
```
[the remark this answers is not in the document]
achettyiitr: Pushed by mistake. Removing it.

warehouse/integrations/snowflake/snowflake.go
vyeshwanth: nit: we will be forced to change the default value and check for extra grants if our implementation changes. I think we might never use this config. Is this done with any other warehouse previously?

achettyiitr: No. This is the first time we are introducing grants-based privileges for verifying the validations. There is no harm in making this configurable.

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -1446,8 +1460,179 @@ func (sf *Snowflake) LoadTestTable(
...
+	role := sf.Warehouse.GetStringDestinationConfig(sf.conf, model.RoleSetting)
+	if len(role) != 0 {
+		return []string{role}, nil
+	}
+	return sf.getGrantedRoles(ctx)
+}
+
+func (sf *Snowflake) getGrantedRoles(ctx context.Context) ([]string, error) {
+	user := sf.Warehouse.GetStringDestinationConfig(sf.conf, model.UserSetting)
+	sqlStatement := fmt.Sprintf("SHOW GRANTS TO USER %q;", user)
```
vyeshwanth, line 1515: If role is left out on destination config page and snowflake user has a default role set, snowflake uses the default role for that session. Switching to `SHOW GRANTS TO USER` shows all the roles. current logic is checking privileges granted for all the roles. Is this something to check?

achettyiitr: https://github.com/rudderlabs/rudder-server/pull/5692#discussion_r2028297507

achettyiitr: > If role is left out on destination config page and snowflake user has a default role set, snowflake uses the default role for that session.

It's possible that no default role is set as well.

PR #5699, opened 2025-04-07
fix: kafka message batching applies to Confluent Cloud and Azure Event Hub

Sidddddarth:
# Description

Kafka stream manager is used for `Azure Event Hub` and `Confluent Cloud` destinations. The problem is message batching directive(configured for Kafka) is applied across all three as well which can be a problem if batched messages cannot be accepted further down the pipeline. The same compression directive still applies to all three.

## Linear Ticket

Resolves PIPE-1982
[Remove kafka batching on transformer and use client batching with a new config flag](https://linear.app/rudderstack/issue/PIPE-1982/remove-kafka-batching-on-transformer-and-use-client-batching-with-a)

## Security

### DIFF
--- a/services/streammanager/kafka/kafkamanager.go
+++ b/services/streammanager/kafka/kafkamanager.go
@@ -441,7 +441,8 @@ func NewProducerForAzureEventHubs(destination *backendconfig.DestinationT, o com
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


koladilip (approved):
LGTM, will dynamically hot reload.

cisse21 (approved):

PR #5703, opened 2025-04-08
chore: allow limiting query result via payload_size column aggregate

Sidddddarth:
# Description

1. Introduce a `payload_size` column to jobsdb jobs tables.
2. limit get jobs queries via aggregating sum over said column instead of sum over `pg_column_size(event_payload)`.

[Motivation](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1741967424662249)

## Linear Ticket

[Resolves PIPE-2009](https://linear.app/rudderstack/issue/PIPE-2009/fix-payload-limiting-by-adding-actual-column-size-in-jobsdb)

## Security


jobsdb/jobsdb.go
```diff
@@ -1467,6 +1469,7 @@ func (jd *Handle) createDSTablesInTx(ctx context.Context, tx *Tx, newDS dataSetT
 		custom_val VARCHAR(64) NOT NULL,
 		event_payload `+string(columnType)+` NOT NULL,
 		event_count INTEGER NOT NULL DEFAULT 1,
+		payload_size INTEGER NOT NULL DEFAULT 0,
```
Sidddddarth, line 1472: allowing `NULL` would let us query right away by `sum(COALESCE(jobs.payload_size, pg_column_size(jobs.event_payload))) over (order by jobs.job_id) as running_payload_size`?

jobsdb/jobsdb_test.go
```diff
@@ -1536,6 +1537,86 @@ func TestGetDistinctParameterValues(t *testing.T) {
...
+		GetQueryParams{
+			CustomValFilters: []string{customVal},
+			ParameterFilters: []ParameterFilterT{
+				{Name: "destination_id", Value: destinationID},
+			},
+			JobsLimit:        1000,
+			PayloadSizeLimit: 1024 * bytesize.B,
+		})
+	require.NoError(t, err)
+	require.Equal(t, 64, len(unprocessed.Jobs))
```
Sidddddarth, line 1616: I expected lesser jobs. Tried with some different payloads too, but got the same result.

jobsdb/jobsdb.go
atzoum, line 1472: What about existing tables that don't have the column?

atzoum: but we are not doing a COALESCE now, right?

jobsdb/jobsdb_test.go
atzoum: you need to trigger compression threshold in postgresql, i.e. use payloads larger than 2 kB

jobsdb/jobsdb.go
Sidddddarth: I forgot to add the migration template!

Sidddddarth: We aren't yeah.
I just went with the 2 step solution.
Thought of this afterwards.

Sidddddarth:
Since the problem we're trying to address is loading more events into memory than can be safely handled, should we include the size of parameters and the other fields in the number we store to the new column..?

lvrach:
What's the motivation behind this change? Make the query more efficient as we won't have to compute the size every time ?

Sidddddarth:
> What's the motivation behind this change? Make the query more efficient as we won't have to compute the size every time ?

That too, but more importantly `sum(pg_column_size(event_payload)) over ...` aggregates the compressed sizes which means the returned events might occupy more memory on client(beyond the size limit specified) than expected. [ref](https://rudderlabs.slack.com/archives/C01HTT66UMB/p1741967424662249)

jobsdb/jobsdb.go
atzoum, line 1472: - in `JobT` struct we have a `PayloadSize` field, we should delete it
- in `getJobsDS` we select `pg_column_size(jobs.event_payload) as payload_size`, we shouldn't

jobsdb/jobsdb.go
```diff
@@ -2196,14 +2199,18 @@ func (jd *Handle) getJobsDS(ctx context.Context, ds dataSetT, lastDS bool, param
 		// If we are querying only for unprocessed jobs, we should join with the status table instead of the view (performance reasons)
 		joinTable = ds.JobStatusTable
 	}
+	sizeLimitClause := "sum(pg_column_size(jobs.event_payload)) over (order by jobs.job_id) as running_payload_size"
```
atzoum, line 2202:
```suggestion
	runningPayloadSizeSelector := "sum(pg_column_size(jobs.event_payload)) over (order by jobs.job_id) as running_payload_size"
```

jobsdb/jobsdb.go
```diff
@@ -2196,14 +2199,18 @@ func (jd *Handle) getJobsDS(ctx context.Context, ds dataSetT, lastDS bool, param
...
+	if jd.conf.usePayloadSizeColumnToQuery {
+		sizeLimitClause = "sum(jobs.payload_size) over (order by jobs.job_id asc) as running_payload_size"
+	}
```
atzoum, line 2205:
```suggestion
	runningPayloadSizeSelector := "pg_column_size(jobs.event_payload)"
	if jd.conf.usePayloadSizeColumnToQuery {
		runningPayloadSizeSelector = "jobs.payload_size"
	}
```

jobsdb/jobsdb_test.go
```diff
@@ -1536,6 +1537,86 @@ func TestGetDistinctParameterValues(t *testing.T) {
...
...valuevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevaluevalue"}`),
```
atzoum, line 1552: can use rudder-go-kit rand for generating a random string of desired size

jobsdb/jobsdb.go
```diff
@@ -2045,7 +2048,7 @@ func (jd *Handle) doStoreJobsInTx(ctx context.Context, tx *Tx, ds dataSetT, jobL
 		var stmt *sql.Stmt
 		var err error
 
-		stmt, err = tx.PrepareContext(ctx, pq.CopyIn(ds.JobTable, "uuid", "user_id", "custom_val", "parameters", "event_payload", "event_count", "workspace_id"))
+		stmt, err = tx.PrepareContext(ctx, pq.CopyIn(ds.JobTable, "uuid", "user_id", "custom_val", "parameters", "event_payload", "event_count", "payload_size", "workspace_id"))
```
atzoum, line 2051: we need to handle migrateJobsInTx too and there there can be a case where the previous table doesn't have the `payload_size` column

PR #5704, opened 2025-04-08
feat(jobsdb): introduce jobMinRowsLeftMigrateThres

atzoum:
# Description

Introducing `JobsDB.jobMinRowsLeftMigrateThres` configuration property, for handling compaction of jobsdb datasets containing multiple destinations, where one destination is slower than the others. In such a case, jobsdb datasets end up containing some pending jobs (e.g. 40%), but the dataset doesn't get migrated because it is neither small (we count total number of jobs for calculating small tables) nor does it have an adequate percentage of completed jobs (70%).

With this new option, tables with less than 40K pending jobs will become migration candidates and will be migrated (i.e. compacted) alongside some other dataset candidate for migration, as long as one exists.

## Additional items

- Making jobsdb configuration properties tableprefix-aware, i.e. making it possible to have different settings for `gw`, `rt`, `batch_rt`, etc.

## Linear Ticket

resolves PIPE-2017

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -776,7 +776,7 @@ func (jd *Handle) init() {
 	}
 
 	if string(jd.conf.payloadColumnType) == "" {
-		jd.conf.payloadColumnType = payloadColumnType(jd.config.GetStringVar(string(JSONB), "JobsDB.payloadColumnType"))
+		jd.conf.payloadColumnType = payloadColumnType(jd.config.GetStringVar(string(JSONB), jd.configKeys("payloadColumnType")...))
 	}
 
 	if jd.stats == nil {
@@ -895,7 +895,7 @@ func (jd *Handle) workersAndAuxSetup() {
 		cacheParameterFilters,
 		func() time.Duration { return jd.conf.cacheExpiration.Load() },
 		cache.WithWarnOnBranchInvalidation[ParameterFilterT](
-			jd.config.GetReloadableBoolVar(defaultLogCacheBranchInvalidation, "JobsDB."+jd.tablePrefix+".logCacheBranchInvalidation", "JobsDB.logCacheBranchInvalidation"),
+			jd.config.GetReloadableBoolVar(defaultLogCacheBranchInvalidation, jd.configKeys("logCacheBranchInvalidation")...),
 			jd.logger),
 	)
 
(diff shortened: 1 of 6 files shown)


jobsdb/jobsdb.go
```diff
@@ -907,78 +908,56 @@ func (jd *Handle) workersAndAuxSetup() {
 
 func (jd *Handle) loadConfig() {
 	// maxTableSizeInMB: Maximum Table size in MB
-	jd.conf.maxTableSize = jd.config.GetReloadableInt64Var(300, 1000000, "JobsDB.maxTableSizeInMB")
```
Sidddddarth, line 910: maybe consider pulling this out into a method
```
func (jd *Handle) configVarLiteral(option string) []string {
	return []string{"JobsDB."+jd.tablePrefix+"."+option, "JobsDB."+option}
}
```
can help avoid minor mistakes

jobsdb/migration.go
```diff
@@ -657,13 +657,17 @@ func (jd *Handle) checkIfMigrateDS(ds dataSetT) (
...
+	if float64(delCount)/float64(totalCount) > jd.conf.migration.jobDoneMigrateThres.Load() {
+		return true, isSmallDS, recordsLeft, nil
 	}
 
-	if isSmall {
+	if isSmallDS {
+		return true, true, recordsLeft, nil
+	}
+
+	if float64(recordsLeft) < jd.conf.migration.jobMinRowsLeftMigrateThres.Load()*float64(jd.conf.MaxDSSize.Load()) {
```
Sidddddarth, line 670: With this condition, I think we can remove the above `isSmall` check and its config(this new condition becomes `isSmall`).
Too many conditions present might increase cognitive load.

atzoum: Yes, we may remove isSmall and its config in the next release after verifying behaviour of the new config in production 👍

Sidddddarth (approved):

jobsdb/jobsdb.go
```diff
@@ -907,78 +908,56 @@ func (jd *Handle) workersAndAuxSetup() {
...
-	// jobMinRowsMigrateThres: A DS with a low number of rows should be eligible for migration if the number of rows are
-	// less than jobMinRowsMigrateThres percent of maxDSSize (e.g. if jobMinRowsMigrateThres is 5
-	// then DSs that have less than 5% of maxDSSize are eligible for migration)
-	jd.conf.migration.jobMinRowsMigrateThres = func() float64 { return jd.config.GetFloat64("JobsDB.jobMinRowsMigrateThreshold", 0.2) }
+	jd.conf.migration.jobStatusMigrateThres = jd.config.GetReloadableFloat64Var(3, jd.configKeys("jobStatusMigrateThreshold")...)
+	// jobMinRowsMigrateThres: A DS with a low number of total rows should be eligible for migration if the number of total rows are
+	// less than jobMinRowsMigrateThres percent of maxDSSize (e.g. if jobMinRowsMigrateThres is 0.2
+	// then DSs that have less than 20% of maxDSSize total rows are eligible for migration)
+	jd.conf.migration.jobMinRowsMigrateThres = jd.config.GetReloadableFloat64Var(0.2, jd.configKeys("jobMinRowsMigrateThreshold")...)
+	// jobMinRowsMigrateThres: A DS with a low number of pending rows should be eligible for migration if the number of pending rows are
```
ktgowtham, line 941: `jobMinRowsLeftMigrateThres`

ktgowtham (approved):

jobsdb/jobsdb.go
atzoum: this is the old config option, for which the plan is to remove in next release

PR #5705, opened 2025-04-09
feat: add support for table size metadata in warehouse /v1/process API

shekhar-rudder:
# Description

- Updated the `/v1/process` API to accept a new field `BytesPerTable` of type `map[string]int64`.
- Added a new column `bytes_per_table` of type jsonb to the `wh_staging_files table` to persist this data.

## Security


achettyiitr (approved):

RanjeetMishra (approved):
@shekhar-rudder 
For new column say we define json schema as:
```
{
  v: v1,
  d: <data blob>,
}
```

This will easily allow us to decide to save data blob as gzip if we find out storing as compressed will help us. And change can be backward compatible with minimal change at extra cost of few bytes like 5-10 for v, v1, d, and additional json punctuations.

WDYT?

PR #5711, opened 2025-04-09
chore(router): make net client timeout reloadable

koladilip:
# Description
This commit implements the reloadable netClientTimeout parameter in the router package. The implementation allows the HTTP client timeout to be updated at runtime without requiring a server restart, improving operational flexibility.

## Linear Ticket

https://linear.app/rudderstack/issue/INT-3508/make-router-netclienttimeout-hot-reloadable
Resolves INT-3508

## Security

### DIFF
--- a/router/handle_lifecycle.go
+++ b/router/handle_lifecycle.go
@@ -87,12 +87,12 @@ func (rt *Handle) Setup(
 	if rt.netHandle == nil {
 		netHandle := &netHandle{disableEgress: config.GetBool("disableEgress", false)}
 		netHandle.logger = rt.logger.Child("network")
-		netHandle.Setup(destType, rt.netClientTimeout)
+		netHandle.Setup(destType, rt.netClientTimeout.Load())
 		rt.netHandle = netHandle
 	}
 
 	rt.customDestinationManager = customDestinationManager.New(destType, customDestinationManager.Opts{
-		Timeout: rt.netClientTimeout,
+		Timeout: rt.netClientTimeout.Load(),
 	})
 	rt.telemetry = &Diagnostic{}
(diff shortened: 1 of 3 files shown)


vinayteki95 (approved):

router/handle_lifecycle.go
```diff
@@ -87,12 +87,12 @@ func (rt *Handle) Setup(
 	if rt.netHandle == nil {
 		netHandle := &netHandle{disableEgress: config.GetBool("disableEgress", false)}
 		netHandle.logger = rt.logger.Child("network")
-		netHandle.Setup(destType, rt.netClientTimeout)
+		netHandle.Setup(destType, rt.netClientTimeout.Load())
```
atzoum, line 90: The `http.Client`'s timeout isn't reloaded though... same for all other custom destination managers timeouts, they keep their initial value only. 
Furthermore, not all clients will be able to change their timeout value, e.g. aws eventbridge etc.

That being said, I don't believe it is possible, or at least simple, to convert this timeout to a reloadable configuration value.

koladilip: Mainly I am doing this for 
router/worker.go ->
sendCtx, cancel := context.WithTimeout(ctx, w.rt.netClientTimeout.Load())

atzoum: It is risky to change the context's timeout, while the client's timeout remains unchanged. This can lead to behaviours that the person changing the configuration property wouldn't expect.
- We should remove the timeout from the http client itself
- We should use two different configuration properties, one reloadable and another one non-reloadable. We shouldn't be using reloadable configuration properties for components that are not able to adapt to configuration property changes.

koladilip: make sense

PR #5714, opened 2025-04-10
fix: handling isBlank for nested and empty arrays

achettyiitr:
# Description

- **isBlank** to check recursively and return true for an empty string in case of arrays.
- Try out the [isBlank implementation](https://github.com/rudderlabs/rudder-transformer/blob/develop/src/warehouse/config/helpers.js#L9-L16) in rudder-transformer using the [Loash Playground](https://playcode.io/2335384).

```GO
  types.TransformerResponse{
  	Output: map[string]any{
  		"data": map[string]any{
  			... // 5 identical entries
- 			"context_consent_management_denied_consent_ids":  []any{string("")},
  			... // 32 identical entries
  		},
  		"metadata": map[string]any{
  			"columns": map[string]any{
  				... // 5 identical entries
- 				"context_consent_management_denied_consent_ids":  string("string"),
  				... // 34 identical entries
  			},
  		},
  	},
  	... // 3 identical fields
  }
```

## Linear Ticket

- Resolves WAR-498.

## Security


warehouse/transformer/internal/utils/utils.go
```diff
@@ -176,7 +176,13 @@ func IsBlank(value interface{}) bool {
 	case map[string]any:
 		return len(v) == 0
```
RanjeetMishra, line 177: @achettyiitr Any magic checks for map in js transformer? Like empty string as key and len(v) > 0?

RanjeetMishra (approved):

warehouse/transformer/internal/utils/utils.go
achettyiitr: No, it's not required. [_.toString()](https://lodash.com/docs/4.17.15#toString) returns something for an empty map as well.

vyeshwanth (approved):

PR #5718, opened 2025-04-11
fix: handle consent management filtering per connection

cisse21:
# Description

Fixed the consent management based event filtering logic for GCM to prepare in-memory maps based on both source and destination ID. This is because the GCM configuration can be specified per source type unlike the legacy OneTrust and Ketch configuration.

I've also updated the test suites to cover these advanced scenarios.

Ideally, we should store the data per source type and destination ID but making it more specific to the connection level shouldn't hurt. Moreover, when these kind of configurations move to the connection settings in the future, it'll be beneficial.

Without these changes, a destination that's connected to more than one source type and has distinct GCM config for each source types will result in incorrect consent based filtering as the last source type overrides all the data in the map for the destination.

## Linear Ticket

https://linear.app/rudderstack/issue/SDK-3205/fix-consent-management-based-event-filtering-for-cloud-mode

## Security

### DIFF
--- a/processor/consent.go
+++ b/processor/consent.go
@@ -37,8 +37,9 @@ type GenericConsentManagementProviderConfig struct {
 Filters and returns destinations based on the consents configured for the destination and the user consents present in the event.
 
 Supports legacy and generic consent management.
+For GCM based filtering, uses source and destination IDs to fetch the appropriate GCM data from the config.
 */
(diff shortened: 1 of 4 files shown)


ktgowtham (approved):

saikumarrs (approved):

saikumarrs:
For reference, the original PR:
https://github.com/rudderlabs/rudder-server/pull/5716

PR #5723, opened 2025-04-14
fix: jobsdb payload limit used compressed column sizes instead of actual payload length

Sidddddarth:
# Description

Job queries aggregated over `pg_column_size(event_payload)` to limit the result when a size limit is passed. This could lead to loading more jobs into memory than can be handled safely because it's the compressed size of the column(when compressed). Using the `octet_length(event_payload)` instead to find the actual byte size of the event payload column.

## Linear Ticket

[Resolves PIPE-2009](https://linear.app/rudderstack/issue/PIPE-2009/fix-payload-limiting-by-adding-actual-column-size-in-jobsdb)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -406,7 +406,6 @@ type JobT struct {
 	CustomVal     string          `json:"CustomVal"`
 	EventCount    int             `json:"EventCount"`
 	EventPayload  json.RawMessage `json:"EventPayload"`
-	PayloadSize   int64           `json:"PayloadSize"`
 	LastJobStatus JobStatusT      `json:"LastJobStatus"`
 	Parameters    json.RawMessage `json:"Parameters"`
 	WorkspaceId   string          `json:"WorkspaceId"`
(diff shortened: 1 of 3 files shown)


Sidddddarth:
This replaces #5703 .

jobsdb/jobsdb.go
```diff
@@ -2189,9 +2188,9 @@ func (jd *Handle) getJobsDS(ctx context.Context, ds dataSetT, lastDS bool, param
 	sqlStatement := fmt.Sprintf(`SELECT
 									jobs.job_id, jobs.uuid, jobs.user_id, jobs.parameters, jobs.custom_val, jobs.event_payload, jobs.event_count,
 									jobs.created_at, jobs.expire_at, jobs.workspace_id,
-									pg_column_size(jobs.event_payload) as payload_size,
+									octet_length(jobs.event_payload::text) as payload_size,
```
atzoum, line 2191: What is the impact on performance? Does this need to read the whole payload column and count bytes? 
Can we benchmark the two different solutions using 10K dataset with large payloads (e.g. 13K) and measure:
1. query latency
2. postgres cpu

atzoum: Run jobsdb benchmarks for:
- octet_length
- pg_column_size
- separate payload size column

octet_length performance appeared to be better than all other, a backlog of 2Mil pending events with 2.5K payload size and 10K/sec constant event rate was processed in 2m17s vs 4m10s that other solutions.

atzoum (approved):

jobsdb/jobsdb.go
Sidddddarth: Thank you!

cisse21 (approved):

PR #5724, opened 2025-04-14
fix: warehouse transformations concurrent writes

achettyiitr:
# Description

- In processor, [multiplexing happens](https://github.com/rudderlabs/rudder-server/blob/69f1baff51e05c561f0b3122a020e315c62c37bc/processor/processor.go#L2159) when source is connected to multiple destinations. These shallow events point to the same singular event, a map. Modifying and reading this singular event gives `fatal error: concurrent map read and map write` error.
- Also, avoid making modifications to singular events.

## Linear Ticket

- Resolves WAR-488.

## Security

### DIFF
--- a/warehouse/transformer/benchmark/main.go
+++ b/warehouse/transformer/benchmark/main.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/samber/lo"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
@@ -58,15 +59,15 @@ func run(ctx context.Context) error {
 	logFactory := logger.Default
 	l := logFactory.NewLogger().Child("warehouse-transformer-benchmark").Child(mode)
 
-	clientEvents := make([]types.TransformerEvent, eventsInBatch)
-	for i := 0; i < len(clientEvents); i++ {
-		var transformerEvent types.TransformerEvent
-		err := jsonrs.Unmarshal([]byte(sampleEvent), &transformerEvent)
-		if err != nil {
-			return fmt.Errorf("could not unmarshal sample client event: %w", err)
-		}
-		clientEvents[i] = transformerEvent
+	var te types.TransformerEvent
+	err := jsonrs.Unmarshal([]byte(sampleEvent), &te)
+	if err != nil {
+		return fmt.Errorf("could not unmarshal sample client event: %w", err)
 	}
+
+	clientEvents := lo.RepeatBy(eventsInBatch, func(int) types.TransformerEvent {
+		return te
+	})
 
 	t, err := selectTransformer(mode, conf, l)
 	if err != nil {
--- a/warehouse/transformer/internal/rules/rules.go
+++ b/warehouse/transformer/internal/rules/rules.go
@@ -16,12 +16,12 @@ type Rules func(event *types.TransformerEvent) (any, error)
 
 var (
 	DefaultRules = map[string]Rules{
-		"id":                 staticRule("messageId"),
+		"id":                 messageIDFromMetadata(),
 		"anonymous_id":       staticRule("anonymousId"),
 		"user_id":            staticRule("userId"),
 		"sent_at":            staticRule("sentAt"),
 		"timestamp":          staticRule("timestamp"),
-		"received_at":        staticRule("receivedAt"),
+		"received_at":        receivedAtFromMetadata(),
 		"original_timestamp": staticRule("originalTimestamp"),
 		"channel":            staticRule("channel"),
 		"context_ip": func(event *types.TransformerEvent) (any, error) {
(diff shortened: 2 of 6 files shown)


vyeshwanth (approved):

vyeshwanth (approved):

warehouse/transformer/benchmark/main.go
```diff
@@ -58,16 +59,16 @@ func run(ctx context.Context) error {
...
 
-	clientEvents := make([]types.TransformerEvent, eventsInBatch)
-	for i := 0; i < len(clientEvents); i++ {
-		var transformerEvent types.TransformerEvent
-		err := jsonrs.Unmarshal([]byte(sampleEvent), &transformerEvent)
-		if err != nil {
-			return fmt.Errorf("could not unmarshal sample client event: %w", err)
-		}
-		clientEvents[i] = transformerEvent
+	var te types.TransformerEvent
```
shekhar-rudder, line 62: To confirm, the change in this file is specific to `samber/lo` and nothing to do with concurrency, rt ?

warehouse/transformer/transformer.go
```diff
@@ -133,35 +134,43 @@ func (t *Transformer) Transform(_ context.Context, clientEvents []types.Transfor
...
-	messageContextMap["destinationId"] = event.Metadata.DestinationID
-	messageContextMap["destinationType"] = event.Metadata.DestinationType
-
-	event.Message["context"] = messageContextMap
 	return nil
 }
 
+func (t *Transformer) eventContext(tec *transformEventContext) any {
+	contextVal, exists := tec.event.Message["context"]
+	if !t.config.populateSrcDestInfoInContext.Load() {
```
shekhar-rudder, line 160: The if check can be evaluated before `contextVal`

warehouse/transformer/transformer.go
```diff
@@ -171,7 +180,6 @@ func (t *Transformer) ensureMessageID(event *types.TransformerEvent) {
 	messageID, exists := event.Message["messageId"]
 	if !exists || utils.IsBlank(messageID) {
 		event.Metadata.MessageID = "auto-" + t.uuidGenerator()
-		event.Message["messageId"] = event.Metadata.MessageID
```
shekhar-rudder, line 174: Why is `messageId` and `receivedAt` being removed?

warehouse/transformer/transformer_benchmark_test.go
```diff
@@ -10,19 +10,18 @@ import (
...
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
+
 	"github.com/rudderlabs/rudder-server/jsonrs"
 	"github.com/rudderlabs/rudder-server/processor/types"
 )
 
 /*
-Benchmark_Transformer/Identify_(POSTGRES)_-_1000_events
-Benchmark_Transformer/Identify_(POSTGRES)_-_1000_events-12         	       9	 112701398 ns/op
```
shekhar-rudder, line 19: Earlier we had 2 `ns/op`, now only 1. What changed?

warehouse/transformer/internal/rules/rules.go
```diff
@@ -16,12 +16,12 @@ type Rules func(event *types.TransformerEvent) (any, error)
 
 var (
 	DefaultRules = map[string]Rules{
-		"id":                 staticRule("messageId"),
+		"id":                 messageIDFromMetadata(),
```
shekhar-rudder, line 19: Should we add a comment here explaining why we're using the `staticRule` method for some fields, while using a different method for others?

warehouse/transformer/set.go
```diff
@@ -204,6 +205,24 @@ func storeRudderEvent(
 	if err != nil {
 		return fmt.Errorf("marshalling event: %w", err)
 	}
+	if t.config.populateSrcDestInfoInContext.Load() {
+		eventJSON, err = sjson.SetBytes(eventJSON, "context.sourceId", tec.event.Metadata.SourceID)
```
shekhar-rudder, line 209: Looks like this same logic is also present in `eventContext` method? Any reason why we need it at both the places?

achettyiitr: We need the raw events with those additional fields. Since we are not modifying the actual event, we need to set it here.

warehouse/transformer/transformer.go
achettyiitr: We are returning return `contextVal` within the if block right.

warehouse/transformer/transformer.go
achettyiitr: We don't need to modify SingularEvent. Whereever these are being used, we are using Metadata instead of Event.

warehouse/transformer/transformer_benchmark_test.go
achettyiitr: We do have only 1 Benchmark within the test. This one is not required.

PR #5725, opened 2025-04-14
fix: unmarshalling error for staging files without bytes per table

shekhar-rudder:
# Description

**BUG:** Unmarshalling of existing records in the `wh_staging_files` table was failing due to the `bytes_per_table` column having a default value of NULL.

**FIX:**
Updated the insert and read methods to handle the nil case for BytesPerTable

## Security

### DIFF
--- a/warehouse/internal/repo/staging.go
+++ b/warehouse/internal/repo/staging.go
@@ -128,10 +128,15 @@ func (sf *StagingFiles) Insert(ctx context.Context, stagingFile *model.StagingFi
 	if err != nil {
 		return id, fmt.Errorf("marshaling schema: %w", err)
 	}
-
-	bytesPerTableJSON, err := jsonrs.Marshal(stagingFile.BytesPerTable)
-	if err != nil {
-		return id, fmt.Errorf("marshaling bytes per table: %w", err)
+	var bytesPerTablePayload interface{}
+	if stagingFile.BytesPerTable != nil {
+		marshalled, err := jsonrs.Marshal(stagingFile.BytesPerTable)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


warehouse/internal/repo/staging.go
```diff
@@ -128,10 +128,15 @@ func (sf *StagingFiles) Insert(ctx context.Context, stagingFile *model.StagingFi
 	if err != nil {
 		return id, fmt.Errorf("marshaling schema: %w", err)
 	}
-
-	bytesPerTableJSON, err := jsonrs.Marshal(stagingFile.BytesPerTable)
```
shekhar-rudder, line 132: This case wasn't caught by the existing tests since in the DB `bytes_per_table` was being stored as `null` string

achettyiitr (approved):

warehouse/internal/repo/staging.go
shekhar-rudder, line 132: This case was not caught by existing tests since post marshalling value of bytesPerTableJSON was coming as null which was getting saved in the DB. And on unmarshalling this value ("null"), there was no error

RanjeetMishra (approved):

warehouse/internal/repo/staging.go
RanjeetMishra: Default json Marshal behavior, it returns "null" for  `nil`. What we missed here is a test scenario considering an existing entry in warehouse DB?

shekhar-rudder: > What we missed here is a test scenario considering an existing entry in warehouse DB?

Yes

PR #5729, opened 2025-04-15
feat: support for creating upload v2 notifier jobs

shekhar-rudder:
# Description

This PR introduces support for creation of upload_v2 notifier jobs required for batching of staging files

**Key Changes:**
- Files are grouped into v1 and v2 based on the presence of BytesPerTable and a feature flag.
- Extracted common logic for v1 and v2 paths into separate methods
- For now, v2 notifier jobs include only one staging file per job, similar to v1. This is temporary — batching logic will be implemented in the next PR.
- We will no longer track the processing status (success/failure) of staging files in v2.
- v2 jobs will always force creation of load files (see state_generate_load_files.go).

## Linear Ticket

https://linear.app/rudderstack/issue/WAR-461/enable-creation-of-upload-v2-notifier-jobs

## Security

### DIFF
--- a/warehouse/internal/loadfiles/loadfiles.go
+++ b/warehouse/internal/loadfiles/loadfiles.go
@@ -213,7 +213,7 @@ func (lf *LoadFileGenerator) createFromStaging(ctx context.Context, job *model.U
 	if !lf.AllowUploadV2JobCreation(job) {
 		err = lf.createUploadJobs(ctx, job, toProcessStagingFiles, publishBatchSize, uniqueLoadGenID)
 		if err != nil {
-			return 0, 0, err
+			return 0, 0, fmt.Errorf("creating upload jobs: %w", err)
 		}
 		return lf.getLoadFileIDs(ctx, job, stagingFileIDs, uniqueLoadGenID)
 	}
@@ -343,7 +343,7 @@ func (lf *LoadFileGenerator) publishJobs(
 		Priority:     job.Upload.Priority,
 	})
 	if err != nil {
-		return nil, fmt.Errorf("error publishing to notifier: %w", err)
+		return nil, fmt.Errorf("publishing to notifier: %w", err)
 	}
 	return ch, nil
 }
--- a/warehouse/internal/loadfiles/loadfiles.go
+++ b/warehouse/internal/loadfiles/loadfiles.go
@@ -68,8 +68,12 @@ type LoadFileGenerator struct {
 }
 
 type WorkerJobResponse struct {
-	StagingFileID int64            `json:"StagingFileID"`
-	Output        []LoadFileUpload `json:"Output"`
+	StagingFileID int64            `json:"staging_file_id"`
+	Output        []LoadFileUpload `json:"output"`
+}
+
+type WorkerJobResponseV2 struct {
+	Output []LoadFileUpload `json:"output"`
 }
 
 type LoadFileUpload struct {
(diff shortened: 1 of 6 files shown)


warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -140,9 +159,14 @@ func (lf *LoadFileGenerator) ForceCreateLoadFiles(ctx context.Context, job *mode
 	return lf.createFromStaging(ctx, job, job.StagingFiles)
 }
 
+func (lf *LoadFileGenerator) AllowUploadV2JobCreation(job *model.UploadJob) bool {
+	return lf.Conf.GetBool(fmt.Sprintf("Warehouse.%s.enableV2NotifierJob", job.Upload.DestinationType), false) || lf.Conf.GetBool("Warehouse.enableV2NotifierJob", false)
```
vyeshwanth, line 163: shouldn't this be? 
```suggestion
	return lf.Conf.GetBool(fmt.Sprintf("Warehouse.%s.enableV2NotifierJob", job.Upload.DestinationType), false) && lf.Conf.GetBool("Warehouse.enableV2NotifierJob", false)
```
 Toggle has to be enabled and enabled for a destination type. If jobv2 is false for let's say `redshift` and toggle `Warehouse.enableV2NotifierJob` is enabled, we would allow creating v2 jobs.  CMIIW

shekhar-rudder: The plan was to use `Warehouse.enableV2NotifierJob` to switch on for all destinations instead of adding flag for every destination. I never considered the case where I would like to disable it for a specific destination. With the current implementation, I'll need to have all destination flags set to true except for redshift.

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -192,155 +210,52 @@ func (lf *LoadFileGenerator) createFromStaging(ctx context.Context, job *model.U
...
-		uploadSchemaJSON, err := jsonrs.Marshal(struct {
-			UploadSchema model.Schema
-		}{
-			UploadSchema: job.Upload.UploadSchema,
-		})
+	if !lf.AllowUploadV2JobCreation(job) {
+		err = lf.createUploadJobs(ctx, job, toProcessStagingFiles, publishBatchSize, uniqueLoadGenID)
 		if err != nil {
-			return 0, 0, fmt.Errorf("error marshalling upload schema: %w", err)
+			return 0, 0, err
```
vyeshwanth, line 216: minor: better to add context and wrap the error. Makes debugging easier reading the error
```suggestion
			return 0, 0, fmt.Errorf("creating upload jobs: %w", err)
```

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -362,6 +277,228 @@ func (lf *LoadFileGenerator) createFromStaging(ctx context.Context, job *model.U
...
+	}
+	lf.Logger.Infof("[WH]: Publishing %d staging files for %s:%s to notifier", len(messages), destType, destID)
+	ch, err := lf.Notifier.Publish(ctx, &notifier.PublishRequest{
+		Payloads:     messages,
+		JobType:      jobType,
+		UploadSchema: uploadSchemaJSON,
+		Priority:     job.Upload.Priority,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error publishing to notifier: %w", err)
```
vyeshwanth, line 346: nit: we can drop `error` prefix for the error messages

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -192,155 +210,52 @@ func (lf *LoadFileGenerator) createFromStaging(ctx context.Context, job *model.U
...
-			if err = lf.StageRepo.SetStatuses(ctx, successfulStagingFileIDs, warehouseutils.StagingFileSucceededState); err != nil {
-				return fmt.Errorf("setting staging file status to succeeded: %w", err)
-			}
-			return nil
+			return lf.createUploadV2Jobs(gCtx, job, v2Files, publishBatchSize, uniqueLoadGenID)
 		})
 	}
 
-	if err := g.Wait(); err != nil {
+	if err = g.Wait(); err != nil {
```
vyeshwanth, line 246: `g.Wait()` only returns first non-nil error. we don't want to loose the error if both routines return non-nil errors because one is v1 jobs and other is v2 jobs.

shekhar-rudder: `we don't want to loose the error`

Are you talking from a customer perspective or for debugging v2 errors? If latter, then I think the current implementation is fine since for most of the jobs, the type will be either only v1/v2 and not mixed.

vyeshwanth (approved):

achettyiitr (approved):

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -140,9 +159,14 @@ func (lf *LoadFileGenerator) ForceCreateLoadFiles(ctx context.Context, job *mode
 	return lf.createFromStaging(ctx, job, job.StagingFiles)
 }
 
+func (lf *LoadFileGenerator) AllowUploadV2JobCreation(job *model.UploadJob) bool {
```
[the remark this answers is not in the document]
shekhar-rudder: Yes, it is being used in [router](https://github.com/rudderlabs/rudder-server/pull/5729/files#diff-60fb8efe0b73e12027eb87d0f8e7b2ca276ae893cfbbc75e24f6651198a5ae06R25) package

PR #5746, opened 2025-04-17
chore: change jobsdb payload column type default to text

Sidddddarth:
# Description

Change jobsdb column type default to text. This change ensures - when not configured new datasets created have a `text` event_payload column. Left the config and related handling for now so that it can phased out at a later point so it leaves some time for jobsdb to compact and cleanup older jsonb column containing datasets.

## Linear Ticket

[Resolves PIPE-2043](https://linear.app/rudderstack/issue/PIPE-2043/jobsdb-update-default-payload-column-type-to-text)

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -775,7 +775,7 @@ func (jd *Handle) init() {
 	}
 
 	if string(jd.conf.payloadColumnType) == "" {
-		jd.conf.payloadColumnType = payloadColumnType(jd.config.GetStringVar(string(TEXT), jd.configKeys("payloadColumnType")...))
+		jd.conf.payloadColumnType = TEXT
 	}
 
 	if jd.stats == nil {
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -775,7 +775,7 @@ func (jd *Handle) init() {
 	}
 
 	if string(jd.conf.payloadColumnType) == "" {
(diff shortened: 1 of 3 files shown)


cisse21 (approved):

jobsdb/jobsdb.go
```diff
@@ -775,7 +775,7 @@ func (jd *Handle) init() {
 	}
 
 	if string(jd.conf.payloadColumnType) == "" {
-		jd.conf.payloadColumnType = payloadColumnType(jd.config.GetStringVar(string(JSONB), jd.configKeys("payloadColumnType")...))
+		jd.conf.payloadColumnType = payloadColumnType(jd.config.GetStringVar(string(TEXT), jd.configKeys("payloadColumnType")...))
```
atzoum, line 778: we may remove the configuration option altogether, but still support other data types in the rest of the code

jobsdb/jobsdb.go
```diff
@@ -775,7 +775,7 @@ func (jd *Handle) init() {
 	}
 
 	if string(jd.conf.payloadColumnType) == "" {
-		jd.conf.payloadColumnType = payloadColumnType(jd.config.GetStringVar(string(JSONB), jd.configKeys("payloadColumnType")...))
+		jd.conf.payloadColumnType = TEXT
```
atzoum, line 778: we don't need `jd.conf.payloadColumnType` anymore
```suggestion
```

Sidddddarth: It makes our life easier while testing. For example tests where we run a jobsdb with one column type and restart it with another.
Can be removed eventually when we remove the handling for type interchanging.

atzoum (approved):

PR #5747, opened 2025-04-17
fix: warehouse jobs skip processing destinations

achettyiitr:
# Description

- Additional filters will be added to skip destinations during the pickup of sync jobs. Later, this can be exposed in the UI for customers to pause syncs during scenarios like migrations.
- Corresoponding Slack Thread: https://rudderlabs.slack.com/archives/C02AALVFHJN/p1744730044640719 

## Security

### DIFF
--- a/warehouse/internal/repo/upload.go
+++ b/warehouse/internal/repo/upload.go
@@ -78,6 +78,7 @@ type scanFn func(dest ...any) error
 type ProcessOptions struct {
 	SkipIdentifiers                   []string
 	SkipWorkspaces                    []string
+	SkipDestinations                  []string
 	AllowMultipleSourcesForJobsPickup bool
 }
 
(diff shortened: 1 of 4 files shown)


RanjeetMishra (commented):
@achettyiitr 
Can you explain state of system the moment we update config for skipDestination? Any tasks till upload state will keep doing its work, only at last stage of upload we will keep skipping?

And once this skipDestination list is cleared, it will resume?

achettyiitr:
> @achettyiitr Can you explain state of system the moment we update config for skipDestination? Any tasks till upload state will keep doing its work, only at last stage of upload we will keep skipping?
> 
> And once this skipDestination list is cleared, it will resume?

Once the config is set, we will not pick any jobs for those destinations. If there are some jobs already being processed, it will continue processing. 
Upon removal, jobs from those destinations will be picked up.

PR #5753, opened 2025-04-22
feat: update config to add account details with destination in workspaceConfig

ItsSudip:
# Description

Updated config processing in backend-config module to attach account details to the destination

## Linear Ticket

https://linear.app/rudderstack/issue/INT-3447/integrating-workspace-config-to-rudder-server
## Security


backend-config/account_merger_test.go
```diff
@@ -0,0 +1,242 @@
+package backendconfig
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMergeAccounts(t *testing.T) {
```
atzoum, line 9: the name of the test file should follow the name of the actual file that it tests, e.g. `account_association_test.go`

atzoum, line 9:
```suggestion
func TestProcessAccountAssociations(t *testing.T) {
```

backend-config/namespace_config.go
```diff
@@ -168,7 +168,7 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
 		} else {
 			workspace.ApplyReplaySources()
 		}
-
+		workspace.processAccountAssociations()
```
atzoum, line 171: We should only `processAccountAssociations` if `workspace != nil`, i.e. this is a modified workspace. In the other case, the previous config will contain an already processed workspace

backend-config/account_association.go
```diff
@@ -0,0 +1,62 @@
...
+// The process involves three main steps:
+// 1. Creating a map of account definitions for quick lookup
+// 2. Merging account configurations with their definitions
+// 3. Associating the merged accounts with destinations
+func (c *ConfigT) processAccountAssociations() {
+	// Create a lookup map for account definitions using their names as keys
+	accountDefMap := make(map[string]AccountDefinition)
+	for _, accDef := range c.AccountDefinitions {
+		accountDefMap[accDef.Name] = accDef
+	}
```
atzoum, line 16: simpler alternative
```suggestion
	accountDefMap := lo.SliceToMap(c.AccountDefinitions, func(accDef AccountDefinition) (string, AccountDefinition) {
		return accDef.Name, accDef
	})
```

backend-config/account_association.go
```diff
@@ -0,0 +1,62 @@
...
+	accountWithDefMap := make(map[string]AccountWithDefinition)
+	for _, acc := range c.Accounts {
+		accountWithDefMap[acc.Id] = AccountWithDefinition{
+			Id:                    acc.Id,
+			AccountDefinitionName: acc.AccountDefinitionName,
+			Options:               acc.Options,
+			Secret:                acc.Secret,
+			Config:                accountDefMap[acc.AccountDefinitionName].Config,
+		}
+	}
```
atzoum, line 29: simpler alternative
```suggestion
	accountWithDefMap := lo.SliceToMap(c.Accounts, func(acc Account) (string, AccountWithDefinition) {
		return acc.Id, AccountWithDefinition{
			Id:                    acc.Id,
			AccountDefinitionName: acc.AccountDefinitionName,
			Options:               acc.Options,
			Secret:                acc.Secret,
			Config:                accountDefMap[acc.AccountDefinitionName].Config,
		}
	})
```

backend-config/types.go
```diff
@@ -55,6 +55,8 @@ type DestinationT struct {
 	Transformations       []TransformationT
 	IsProcessorEnabled    bool
 	RevisionID            string
+	Account               AccountWithDefinition
+	DeleteAccount         AccountWithDefinition
```
atzoum, line 59: Absence of an `account` or `deleteAccount` should be using nil value semantics, not zero struct value, so that empty fields would be omitted from json altogether.
```suggestion
	Account               *AccountWithDefinition `json:"account,omitempty"`
	DeleteAccount         *AccountWithDefinition `json:"deleteAccount,omitempty"`
```

backend-config/account_association.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// Iterate through all sources and their destinations to set up account associations
+	for i := range c.Sources {
+		for j := range c.Sources[i].Destinations {
+			dest := &c.Sources[i].Destinations[j]
+			c.setDestinationAccounts(dest, accountWithDefMap)
+		}
+	}
```
atzoum, line 37: Assuming that right now, no account information will be propagated through workspace config, we wouldn't want to iterate over all sources and destinations unnecessarily, if there are no accounts to inject.
```suggestion
	if len(accountWithDefMap) > 0 {
		for i := range c.Sources {
			for j := range c.Sources[i].Destinations {
				dest := &c.Sources[i].Destinations[j]
				c.setDestinationAccounts(dest, accountWithDefMap)
			}
		}
	}
```

config/config.yaml
```diff
@@ -106,14 +106,15 @@ Router:
   kafkaDialTimeout: 10s
   minRetryBackoff: 10s
   maxRetryBackoff: 300s
-  noOfWorkers: 64
+  noOfWorkers: 1
```
atzoum, line 109:
```suggestion
  noOfWorkers: 64
```

config/config.yaml
```diff
@@ -106,14 +106,15 @@ Router:
...
   allowAbortedUserJobsCountForProcessing: 1
   maxFailedCountForJob: 3
   retryTimeWindow: 180m
   failedKeysEnabled: true
   saveDestinationResponseOverride: false
   transformerProxy: false
   transformerProxyRetryCount: 15
+  oAuthV2Enabled: true
```
atzoum, line 117: why do we need that?

ItsSudip:
This is some local changes got pushed. Removed those.

config/config.yaml
ItsSudip: This is some local changes got pushed. Removed those.

backend-config/namespace_config.go
```diff
@@ -168,7 +168,9 @@ func (nc *namespaceConfig) getFromAPI(ctx context.Context) (map[string]ConfigT,
 		} else {
 			workspace.ApplyReplaySources()
 		}
-
+		if workspace != nil {
+			workspace.processAccountAssociations()
+		}
```
atzoum, line 173: simply move `workspace.processAccountAssociations()` inside the `else` branch above
```suggestion
```

atzoum (approved):

backend-config/account_association.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// Create a map of accounts merged with their definitions
+	// This combines the account-specific settings with the shared definition settings
+	accountWithDefMap := lo.SliceToMap(c.Accounts, func(acc Account) (string, AccountWithDefinition) {
+		return acc.Id, AccountWithDefinition{
+			Id:                    acc.Id,
+			AccountDefinitionName: acc.AccountDefinitionName,
+			Options:               acc.Options,
+			Secret:                acc.Secret,
+			Config:                accountDefMap[acc.AccountDefinitionName].Config,
```
koladilip, line 27: we shouldn't use name config as it feels like config of Account and may confuse the future developers who will work on this, better to call AccountDefinition, just like destination

backend-config/types.go
```diff
@@ -84,17 +86,47 @@ func (s *SourceT) IsReplaySource() bool {
...
+	UpdatedAt          time.Time                    `json:"updatedAt"`
+	Credentials        map[string]Credential        `json:"credentials"`
+	Connections        map[string]Connection        `json:"connections"`
+	Accounts           []Account                    `json:"accounts"`
+	AccountDefinitions []AccountDefinition          `json:"accountDefinitions"`
+}
+
+type AccountWithDefinition struct {
+	Id                    string                 `json:"id"`
+	AccountDefinitionName string                 `json:"accountDefinitionName"`
```
koladilip, line 126: we can remove AccountDefinitionName and Config and keep one field AccountDefinition

backend-config/types.go
```diff
@@ -55,6 +55,8 @@ type DestinationT struct {
 	Transformations       []TransformationT
 	IsProcessorEnabled    bool
 	RevisionID            string
+	Account               *AccountWithDefinition `json:"account,omitempty"`
```
koladilip, line 58: DeliveryAccount

backend-config/account_association.go
```diff
@@ -0,0 +1,63 @@
...
+	})
+
+	// Create a map of accounts merged with their definitions
+	// This combines the account-specific settings with the shared definition settings
+	accountWithDefMap := lo.SliceToMap(c.Accounts, func(acc Account) (string, AccountWithDefinition) {
+		return acc.Id, AccountWithDefinition{
+			Id:                acc.Id,
+			Options:           acc.Options,
+			Secret:            acc.Secret,
+			AccountDefinition: accountDefMap[acc.AccountDefinitionName].Config,
```
koladilip, line 26: Better to have whole definition right

backend-config/types.go
```diff
@@ -84,17 +86,46 @@ func (s *SourceT) IsReplaySource() bool {
...
+	Connections        map[string]Connection        `json:"connections"`
+	Accounts           []Account                    `json:"accounts"`
+	AccountDefinitions []AccountDefinition          `json:"accountDefinitions"`
+}
+
+type AccountWithDefinition struct {
+	Id                string                 `json:"id"`
+	Options           map[string]interface{} `json:"options"`
+	Secret            map[string]interface{} `json:"secret"`
+	AccountDefinition map[string]interface{} `json:"config"`
```
koladilip, line 128: Better to have whole definition right

koladilip (approved):

PR #5754, opened 2025-04-22
chore: enable oauthv2 by default

ItsSudip:
# Description

We are enabling OAuth v2 by default in router. 

## Linear Ticket

https://linear.app/rudderstack/issue/INT-3551/clean-up-oauth-v1-module-from-rudder-server

## Security


ItsSudip:
We are enabling OAuth V2 from the operator.

PR #5756, opened 2025-04-23
fix: race condition in table schema updates causing incorrect schema to be saved

shekhar-rudder:
# Description

**BUG**
The `UpdateTableSchema` method can be invoked concurrently by multiple goroutines, each fetching its own copy of the schema using `getSchema`. This can result in a race condition where concurrent updates to the schema map may override each other, leading to incorrect data being saved.

**FIX**
- Introduced a lock at the beginning of the UpdateTableSchema method to ensure only one goroutine can execute it at a time, preventing concurrent modifications.
- Refactored getSchema to simplify mutex logic by moving schema fetching to the constructor. This required moving setting of `schemaHandle` in `UploadJob` struct from its constructor to `run` method
- Updated the `New` method of schema to replace `db` with two new arguments `schemaRepo` and `stagingFileRepo` for easier testing

## Security

### DIFF
--- a/sql/migrations/warehouse/000031_add_bytes_per_table_to_staging_files.up.sql
+++ b/sql/migrations/warehouse/000031_add_bytes_per_table_to_staging_files.up.sql
@@ -1,1 +1,1 @@
-ALTER TABLE wh_staging_files ADD COLUMN bytes_per_table JSONB; 
+ALTER TABLE wh_staging_files ADD COLUMN IF NOT EXISTS bytes_per_table JSONB; 


warehouse/schema/schema.go
```diff
@@ -78,56 +77,89 @@ type schema struct {
...
+	if whSchema.Schema == nil {
+		sh.cachedSchema = model.Schema{}
+		return sh, nil
+	}
+	if whSchema.ExpiresAt.After(sh.now()) {
+		sh.cachedSchema = whSchema.Schema
+		return sh, nil
+	}
+	sh.log.Infon("Schema expired", obskit.DestinationID(sh.warehouse.Destination.ID), obskit.Namespace(sh.warehouse.Namespace), logger.NewTimeField("expiresAt", whSchema.ExpiresAt))
+	return sh, sh.fetchSchemaFromWarehouse(ctx)
```
RanjeetMishra, line 133: @shekhar-rudder 
`return sh, sh.fetchSchemaFromWarehouse(ctx)` : Don't this there should be any reason why it should not be fine to let constructor wait for time taking path for fetch schema as well. Just wanted to check once with you.

warehouse/schema/schema.go
```diff
@@ -78,56 +77,89 @@ type schema struct {
...
+		sh.warehouse.Destination.ID,
+		sh.warehouse.Namespace,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("getting schema for namespace: %w", err)
+	}
+	if whSchema.Schema == nil {
+		sh.cachedSchema = model.Schema{}
+		return sh, nil
+	}
```
RanjeetMishra, line 127: `whSchema.Schema == nil` When can sh.schemaRepo.GetForNamespace return nil? And what would nil value mean?

warehouse/schema/schema.go
```diff
@@ -78,56 +77,89 @@ type schema struct {
...
-func (sh *schema) IsSchemaEmpty(ctx context.Context) bool {
-	schema, err := sh.getSchema(ctx)
+func (sh *schema) fetchSchemaFromWarehouse(ctx context.Context) error {
+	start := sh.now()
+	warehouseSchema, err := sh.fetchSchemaRepo.FetchSchema(ctx)
+	if err != nil {
+		return fmt.Errorf("fetching schema: %w", err)
+	}
+	duration := math.Round((sh.now().Sub(start).Minutes() * 1000)) / 1000
+	sh.log.Infon("Fetched schema from warehouse", obskit.DestinationID(sh.warehouse.Destination.ID), obskit.Namespace(sh.warehouse.Type), logger.NewFloatField("timeTakenInMinutes", duration))
```
RanjeetMishra, line 143: minor: Why minutes and not seconds? Would values of few seconds get rounded as 0 if using minutes?

warehouse/schema/schema.go
```diff
@@ -192,23 +221,9 @@ func (sh *schema) isIDResolutionEnabled() bool {
...
-	duration := math.Round((sh.now().Sub(start).Minutes() * 1000)) / 1000
-	sh.log.Infon("Fetched schema from warehouse", obskit.DestinationID(sh.warehouse.Destination.ID), obskit.Namespace(sh.warehouse.Type), logger.NewFloatField("timeTakenInMinutes", duration))
-	removeDeprecatedColumns(warehouseSchema, sh.warehouse, sh.log)
-	return warehouseSchema, sh.saveSchema(ctx, warehouseSchema)
+	sh.cachedSchemaMu.RLock()
+	defer sh.cachedSchemaMu.RUnlock()
+	return tableSchemaDiff(tableName, sh.cachedSchema, tableSchema), nil
 }
 
 func (sh *schema) saveSchema(ctx context.Context, newSchema model.Schema) error {
```
RanjeetMishra, line 229: minor: This we can discuss later while we start testing this fix. Should this func be thread safe instead of usage to be thread safe? Later one, chances of missing to consider this would be high.

warehouse/schema/schema.go
shekhar-rudder: It will be `nil` in case if no entry exists in RudderStack's postgres for that namespace
https://github.com/rudderlabs/rudder-server/blob/e711579eac45033cd8138cbd24172b15aae93bd0/warehouse/internal/repo/schema.go#L114

warehouse/schema/schema.go
shekhar-rudder: > Don't think there should be any reason why it should not be fine to let constructor wait for time taking path for fetch schema as well

Didn't understand this. The constructor IS waiting for the schema fetch call to finish

RanjeetMishra: Ignore this. Resolving.

warehouse/schema/schema.go
shekhar-rudder: > Why minutes and not seconds?

I believe we would be primarily interested in larger values, so I thought it would be better to use minutes instead of seconds.

> Would values of few seconds get rounded as 0 if using minutes?

No. `math.Round((sh.now().Sub(start).Minutes() * 1000)) / 1000` ensures value upto 3 decimals

warehouse/schema/schema.go
RanjeetMishra: In this case, why not call `fetchSchemaFromWarehouse`? As in for expired, we are fetching. Trying to understand difference between expired and nil.

warehouse/schema/schema.go
shekhar-rudder: > Should this func be thread safe

This function only executes a Postgres query and doesn't perform any other operations, so there's no need to use a lock. In the case of concurrent calls from multiple goroutines, the final DB value will be that of the last write.

warehouse/schema/schema.go
RanjeetMishra: I would say can keep it seconds. As in using minutes do not give any advantage as such.

RanjeetMishra (approved):

warehouse/schema/schema.go
shekhar-rudder: Yes no entry would mean that this is the first sync. Post first sync, there will always be an entry. And by entry, I mean the complete row and not just entry in schema column.

warehouse/schema/schema.go
RanjeetMishra: 1. In our usage, we are taking lock to update DB and memory cache both which is right.
 2. `the final DB value will be that of the last write` -- I am still doubtful about this. Say three goroutines are planning to write value A, B, C. Last write can be any of these? Unless we are using it thread safe as in point 1.

shekhar-rudder: By last write, I meant whoever finished last and not in the order which they invoked this function

> Say three goroutines are planning to write value A, B, C. Last write can be any of these?

Yes

PR #5757, opened 2025-04-23
fix: warehouse extracting messageid and receivedat

achettyiitr:
# Description

- **MessageID** and **receivedAt** in the event and metadata are coming differently. This can happen for several reasons, but the source of truth is the event payload rather than the metadata.
  - For one of the destinations, it was happening because during user transformation (2Q9YiwKH1OEKJrEn6zSe8hibYGt). Customer is trying to override **messageID** in event from properties using `updatedEvent.messageId = event.properties.messageId;` or `properties.meta.id;`
- Sample event.
```
{
  "message": {
    "messageId": 1090524,
    "originalTimestamp": "2025-04-22T10:31:31.388Z",
  },
  "metadata": {
    "messageId": "asdasdas-asdas-dasdas-dsadsaas-asdasdsad",
  }
}
```

## Linear Ticket

- Resolves WAR-526.

## Security

### DIFF
--- a/warehouse/transformer/transformer.go
+++ b/warehouse/transformer/transformer.go
@@ -17,7 +17,7 @@ import (
 	"github.com/rudderlabs/rudder-go-kit/stats"
 
 	"github.com/rudderlabs/rudder-server/processor/types"
-	"github.com/rudderlabs/rudder-server/utils/misc"
+	wtypes "github.com/rudderlabs/rudder-server/warehouse/transformer/internal/types"
 
 	"github.com/rudderlabs/rudder-server/utils/timeutil"
 	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
@@ -137,7 +137,6 @@ func (t *Transformer) processWarehouseMessage(cache *cache, event *types.Transfo
 	if err := t.checkValidContext(event); err != nil {
 		return nil, fmt.Errorf("checking valid context: %w", err)
 	}
-	t.addMandatoryFields(event)
 	return t.handleEvent(event, cache)
 }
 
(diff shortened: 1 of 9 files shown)


warehouse/transformer/transformer.go
```diff
@@ -171,46 +170,20 @@ func (t *Transformer) eventContext(tec *transformEventContext) any {
...
 	intrOpts := extractIntrOpts(event.Metadata.DestinationType, event.Message)
 	destOpts := extractDestOpts(event.Metadata.DestinationType, event.Destination.Config)
 	jsonPathsInfo := extractJSONPathInfo(append(intrOpts.jsonPaths, destOpts.jsonPaths...))
 
 	eventType := strings.ToLower(event.Metadata.EventType)
 
+	warehouseEvent := &wtypes.WarehouseTransformerEvent{}
+	warehouseEvent.TransformerEvent = event
+	warehouseEvent.MessageID = utils.ExtractMessageID(event, t.uuidGenerator)
+	warehouseEvent.ReceivedAt = utils.ExtractReceivedAt(event, t.now)
```
shekhar-rudder, line 183: Should we encapsulate all this logic in constructor of `WarehouseTransformerEvent`?

achettyiitr: I am working on a separate PR to not rely on `TransformerEvent` (as we are not depending only a subset of information) and use `WarehouseTransformerEvent` with all the necessary fields required.

vyeshwanth (approved):

shekhar-rudder (approved):

PR #5765, opened 2025-04-24
feat: batch staging files for creating upload_v2 notifier jobs

shekhar-rudder:
# Description

- Renamed `prepareBaseJobRequest` to `generateBaseRequest` and added computing `StagingDestinationConfig` and `LoadFilePrefix` instead of doing it in `publishJobs`
- Renamed `publishJobs` to `publishToNotifier` which now accepts `messages` as argument instead of computing it using `chunk`
- In `createUploadV2Jobs`, added `GroupStagingFiles` to group files based on size and other characteristics

## Security

### DIFF
--- a/warehouse/internal/loadfiles/loadfiles.go
+++ b/warehouse/internal/loadfiles/loadfiles.go
@@ -604,8 +604,7 @@ func (lf *LoadFileGenerator) GroupStagingFiles(files []*model.StagingFile, maxSi
 	}
 	groups := make(map[stagingFileGroupKey][]*fileWithMaxSize)
 
-	filesForSizing := make([]*fileWithMaxSize, len(files))
-	for i, file := range files {
+	for _, file := range files {
 		key := stagingFileGroupKey{
 			UseRudderStorage:      file.UseRudderStorage,
 			DestinationRevisionID: file.DestinationRevisionID,
@@ -617,11 +616,10 @@ func (lf *LoadFileGenerator) GroupStagingFiles(files []*model.StagingFile, maxSi
 				maxSize = size
 			}
 		}
-		filesForSizing[i] = &fileWithMaxSize{
+		groups[key] = append(groups[key], &fileWithMaxSize{
 			file:    file,
 			maxSize: maxSize,
-		}
-		groups[key] = append(groups[key], filesForSizing[i])
+		})
 	}
 
 	result := make([][]*model.StagingFile, 0, len(groups))
--- a/warehouse/internal/loadfiles/loadfiles.go
+++ b/warehouse/internal/loadfiles/loadfiles.go
@@ -277,10 +277,15 @@ func (lf *LoadFileGenerator) getLoadFileIDs(ctx context.Context, job *model.Uplo
 	return loadFiles[0].ID, loadFiles[len(loadFiles)-1].ID, nil
 }
 
-func (lf *LoadFileGenerator) prepareBaseJobRequest(job *model.UploadJob, uniqueLoadGenID string, stagingFile *model.StagingFile) baseWorkerJobRequest {
+func (lf *LoadFileGenerator) generateBaseRequest(
+	job *model.UploadJob,
+	uniqueLoadGenID string,
+	stagingFile *model.StagingFile,
+	destinationRevisionIDMap map[string]backendconfig.DestinationT,
+) baseWorkerJobRequest {
 	destID := job.Upload.DestinationID
 	destType := job.Upload.DestinationType
-	return baseWorkerJobRequest{
+	baseReq := baseWorkerJobRequest{
 		UploadID:                     job.Upload.ID,
 		LoadFileType:                 job.Upload.LoadFileType,
 		SourceID:                     job.Warehouse.Source.ID,
@@ -298,35 +303,21 @@ func (lf *LoadFileGenerator) prepareBaseJobRequest(job *model.UploadJob, uniqueL
 		DestinationRevisionID:        job.Warehouse.Destination.RevisionID,
 		StagingDestinationRevisionID: stagingFile.DestinationRevisionID,
 	}
+	if revisionConfig, ok := destinationRevisionIDMap[stagingFile.DestinationRevisionID]; ok {
+		baseReq.StagingDestinationConfig = revisionConfig.Config
+	}
+	if slices.Contains(warehouseutils.TimeWindowDestinations, job.Warehouse.Type) {
+		baseReq.LoadFilePrefix = lf.GetLoadFilePrefix(stagingFile.TimeWindow, job.Warehouse)
+	}
+	return baseReq
 }
 
-func (lf *LoadFileGenerator) publishJobs(
+func (lf *LoadFileGenerator) publishToNotifier(
 	ctx context.Context,
 	job *model.UploadJob,
-	uniqueLoadGenID string,
-	chunk []*model.StagingFile,
-	destinationRevisionIDMap map[string]backendconfig.DestinationT,
+	messages []stdjson.RawMessage,
 	jobType notifier.JobType,
-	rawPayloadGenerator func(baseReq baseWorkerJobRequest, stagingFile *model.StagingFile) (stdjson.RawMessage, error),
 ) (<-chan *notifier.PublishResponse, error) {
-	destID := job.Upload.DestinationID
-	destType := job.Upload.DestinationType
-	var messages []stdjson.RawMessage
-	for _, stagingFile := range chunk {
-		baseReq := lf.prepareBaseJobRequest(job, uniqueLoadGenID, stagingFile)
-		if revisionConfig, ok := destinationRevisionIDMap[stagingFile.DestinationRevisionID]; ok {
-			baseReq.StagingDestinationConfig = revisionConfig.Config
-		}
-		if slices.Contains(warehouseutils.TimeWindowDestinations, job.Warehouse.Type) {
-			baseReq.LoadFilePrefix = lf.GetLoadFilePrefix(stagingFile.TimeWindow, job.Warehouse)
-		}
-		rawPayload, err := rawPayloadGenerator(baseReq, stagingFile)
-		if err != nil {
-			return nil, fmt.Errorf("error generating raw payload: %w", err)
-		}
-		messages = append(messages, rawPayload)
-	}
-
 	uploadSchemaJSON, err := jsonrs.Marshal(struct {
 		UploadSchema model.Schema `json:"upload_schema"`
 	}{
@@ -335,7 +326,11 @@ func (lf *LoadFileGenerator) publishJobs(
 	if err != nil {
 		return nil, fmt.Errorf("error marshalling upload schema: %w", err)
 	}
+
+	destID := job.Upload.DestinationID
+	destType := job.Upload.DestinationType
 	lf.Logger.Infof("[WH]: Publishing %d staging files for %s:%s to notifier", len(messages), destType, destID)
+
 	ch, err := lf.Notifier.Publish(ctx, &notifier.PublishRequest{
 		Payloads:     messages,
 		JobType:      jobType,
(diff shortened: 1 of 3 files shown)


warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -572,3 +587,107 @@ func toLoadFile(output LoadFileUpload, job *model.UploadJob) model.LoadFile {
...
+func (lf *LoadFileGenerator) GroupStagingFiles(files []*model.StagingFile, maxSizeMB int) [][]*model.StagingFile {
+	// The fields in this struct are determined by the fields being used in the generateBaseRequest
+	type stagingFileGroupKey struct {
+		UseRudderStorage      bool
+		DestinationRevisionID string
+		TimeWindow            time.Time
+	}
+	groups := make(map[stagingFileGroupKey][]*fileWithMaxSize)
+
+	filesForSizing := make([]*fileWithMaxSize, len(files))
```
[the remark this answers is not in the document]
shekhar-rudder: Removed the slice

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -572,3 +587,105 @@ func toLoadFile(output LoadFileUpload, job *model.UploadJob) model.LoadFile {
...
+
+	// For each group, apply size constraints
+	for _, group := range groups {
+		result = append(result, lf.groupBySize(group, maxSizeMB)...)
+	}
+	return result
+}
+
+// splits a group of staging files based on size constraints
+func (lf *LoadFileGenerator) groupBySize(files []*fileWithMaxSize, maxSizeMB int) [][]*model.StagingFile {
```
vyeshwanth, line 635: offline discussion: 
1. `[]*fileWithMaxSize` has a file and largest table byte size in each file. Largest table byte size could be of `tracks` for one file and `identifies` for another. This won't give optimal batching in case if one file has more tracks, other file has less tracks. Ideally, we want large table size belonging to only one table per each file. 

2.  why move last item in slice to the current file index? Last file is going to have lesser table size 
```
                files[i] = files[len(files)-1]
				files = files[:len(files)-1]
```

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -572,3 +587,105 @@ func toLoadFile(output LoadFileUpload, job *model.UploadJob) model.LoadFile {
...
+				i++
+			}
+		}
+
+		// If we couldn't add any files to the batch, add the first file anyway
+		// This ensures we make progress even with files larger than maxSizeBytes
+		if len(currentBatch) == 0 {
+			currentBatch = append(currentBatch, files[0].file)
+			files = files[1:]
+		}
```
RanjeetMishra, line 686: Wondering if this logic can be included in iteration of `for i < len(files) {` itself. Imagine all files maxSize exceeding configured maxSize. Then we are looping for all and then taking out 1 here. And then again we loop through all files. This is worst case scenario of n^2.

warehouse/internal/loadfiles/loadfiles_test.go
```diff
@@ -484,6 +484,283 @@ func TestGetLoadFilePrefix(t *testing.T) {
...
+					{
+						BytesPerTable: map[string]int64{
+							"table1": 40 * 1024 * 1024, // 40MB
+						},
+					},
+				},
+				batchSizes: []int{2},
+			},
+			{
+				name: "multiple files over limit",
```
RanjeetMishra, line 545: There is another at line 512 with same tag `multiple files over limit`?

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -572,3 +587,105 @@ func toLoadFile(output LoadFileUpload, job *model.UploadJob) model.LoadFile {
...
+	file    *model.StagingFile
+	maxSize int64
+}
+
+// GroupStagingFiles groups staging files based on their key characteristics
+// and then applies size constraints within each group. The maxSizeMB parameter controls the maximum size of any table within a group.
+func (lf *LoadFileGenerator) GroupStagingFiles(files []*model.StagingFile, maxSizeMB int) [][]*model.StagingFile {
+	// The fields in this struct are determined by the fields being used in the generateBaseRequest
+	type stagingFileGroupKey struct {
+		UseRudderStorage      bool
```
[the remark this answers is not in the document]
shekhar-rudder: These keys are being used to further group the files. This is required since the final payload has a field `UseRudderStorage` which is applicable for all the files in the payload. So we can't some files with UseRudderStorage=true and others with UseRudderStorage=false.

The other option would be to apart from sending staging file id and location, we also include additional fields in for every staging file - DestinationRevisionID, UseRudderStorage. In majority of the cases, I would expect these values to be same for all files so this would mean unnecessary increase in the payload size

warehouse/internal/loadfiles/loadfiles_test.go
shekhar-rudder: Removed one of the `multiple files over limit` tests

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -572,3 +587,105 @@ func toLoadFile(output LoadFileUpload, job *model.UploadJob) model.LoadFile {
...
+	}
+	return result
+}
+
+// splits a group of staging files based on size constraints
+func (lf *LoadFileGenerator) groupBySize(files []*fileWithMaxSize, maxSizeMB int) [][]*model.StagingFile {
+	maxSizeBytes := int64(maxSizeMB) * 1024 * 1024 // Convert MB to bytes
+	// Sort by the largest table size in each file
+	slices.SortFunc(files, func(a, b *fileWithMaxSize) int {
+		// Not going with b.maxSize - a.maxSize to avoid int overflow
```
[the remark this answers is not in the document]
shekhar-rudder: Changed to `return b.maxSize > a.maxSize`

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -572,3 +587,105 @@ func toLoadFile(output LoadFileUpload, job *model.UploadJob) model.LoadFile {
...
+	result := make([][]*model.StagingFile, 0, len(groups))
+
+	// For each group, apply size constraints
+	for _, group := range groups {
+		result = append(result, lf.groupBySize(group, maxSizeMB)...)
+	}
+	return result
+}
+
+// splits a group of staging files based on size constraints
```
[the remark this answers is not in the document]
shekhar-rudder: done

warehouse/internal/loadfiles/loadfiles.go
shekhar-rudder: 1. Fixed the code, to check for the only the table with max size across all files
2. Updated the element removal logic to avoid any changes in sorting order

warehouse/internal/loadfiles/loadfiles.go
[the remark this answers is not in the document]
RanjeetMishra: I will discuss this offline to understand it better.

warehouse/internal/loadfiles/loadfiles.go
shekhar-rudder: Moved the logic inside the for loop

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -572,3 +587,105 @@ func toLoadFile(output LoadFileUpload, job *model.UploadJob) model.LoadFile {
...
+			}
+
+			if canAdd {
+				// Add file to batch and update table sizes
+				currentBatch = append(currentBatch, files[i])
+				for tableName, size := range files[i].BytesPerTable {
+					batchTableSizes[tableName] += size
+				}
+				// Remove the file while preserving order
+				files = append(files[:i], files[i+1:]...)
```
RanjeetMishra, line 673: I was thinking of something like below or any other variation to avoid files slice update.

```
func (lf *LoadFileGenerator) groupBySize(files []*model.StagingFile, maxSizeMB int) [][]*model.StagingFile {
	var result [][]*model.StagingFile
	maxSizeBytes := int64(maxSizeMB * 1024 * 1024)

	// Find the table with the maximum total size
	tableSizes := lo.Reduce(files, func(acc map[string]int64, file *model.StagingFile, _ int) map[string]int64 {
		for tableName, size := range file.BytesPerTable {
			acc[tableName] += size
		}
		return acc
	}, make(map[string]int64))

	maxTable, maxTableSize := lo.FindKeyBy(tableSizes, func(tableName string, size int64) bool {
		return size == lo.MaxBy(lo.Values(tableSizes), func(a, b int64) bool {
			return a > b
		})
	})

	lf.Logger.Infof("maxTable: %s, maxTableSize: %d", maxTable, maxTableSize)

	// Sort by size contribution to largest table in descending order
	slices.SortFunc(files, func(a, b *model.StagingFile) int {
		return int(b.BytesPerTable[maxTable] - a.BytesPerTable[maxTable])
	})

	start := 0
	for start < len(files) {
		var currentBatch []*model.StagingFile
		batchTableSizes := make(map[string]int64)
		i := start

		for i < len(files) {
			// Check if adding this file would exceed size limit for any table
			canAdd := true
			for tableName, size := range files[i].BytesPerTable {
				newSize := batchTableSizes[tableName] + size
				if newSize > maxSizeBytes {
					canAdd = false
					break
				}
			}

			if canAdd {
				// Add file to batch and update table sizes
				currentBatch = append(currentBatch, files[i])
				for tableName, size := range files[i].BytesPerTable {
					batchTableSizes[tableName] += size
				}
				i++
			} else {
				// If the first file exceeds the size limits, add it as a single batch
				if len(currentBatch) == 0 {
					result = append(result, []*model.StagingFile{files[i]})
					start = i + 1
				} else {
					// Add the current batch and continue from where we left off
					result = append(result, currentBatch)
					start = i
				}
				break
			}
		}

		// If we reached the end of the files slice, add the remaining batch
		if i == len(files) && len(currentBatch) > 0 {
			result = append(result, currentBatch)
			break
		}
	}

	return result
}
```

RanjeetMishra (approved):
LGTM
I actually understood @vyeshwanth's comment better while reviewing this change. 🙏 Good catch!

warehouse/internal/loadfiles/loadfiles.go
RanjeetMishra: Thanks. My bad, it was not n^2 even with this code, missed reading the break part.

warehouse/internal/loadfiles/loadfiles.go
shekhar-rudder: In the above code, there is a bug
```
else {
	// Add the current batch and continue from where we left off
	result = append(result, currentBatch)
	start = i
}
```

If a file is encountered which can't be added to the current batch, that shouldn't lead us to end that batch there and continue with a new batch. We should keep moving forward until we find a file which can be added to the current batch. For example, if there are three files of sizes of 70, 50 and 30 with max size of 100, we shouldn't create the first batch with only 70, after discovering that 50 can't be added to the first batch along with 70.

To avoid slice creation, I think we will have keep track of all the files which have been already processed.

RanjeetMishra: Above I used cursor to generate. :) 
Yes another way is to use another slice with bool processed or not.

shekhar-rudder: Added a map to keep track of processed files. Removed files mutation

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -572,3 +587,109 @@ func toLoadFile(output LoadFileUpload, job *model.UploadJob) model.LoadFile {
...
+	var result [][]*model.StagingFile
+	processed := make(map[int64]bool, len(files))
+
+	for len(processed) < len(files) {
+		// Start a new batch
+		var currentBatch []*model.StagingFile
+		batchTableSizes := make(map[string]int64)
+
+		// Try to add files to the current batch
+		for i := 0; i < len(files); i++ {
```
vyeshwanth, line 656: nit: 
```suggestion
		for i, file := range files {
```

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -572,3 +587,109 @@ func toLoadFile(output LoadFileUpload, job *model.UploadJob) model.LoadFile {
...
+				// add it to its own batch
+				if len(currentBatch) == 0 {
+					result = append(result, []*model.StagingFile{files[i]})
+					processed[files[i].ID] = true
+					break
+				}
+			}
+		}
+
+		if len(currentBatch) > 0 {
```
vyeshwanth, line 689: isn't this always `true` ?

vyeshwanth (approved):

warehouse/internal/loadfiles/loadfiles.go
shekhar-rudder: Done

warehouse/internal/loadfiles/loadfiles.go
```diff
@@ -572,3 +587,109 @@ func toLoadFile(output LoadFileUpload, job *model.UploadJob) model.LoadFile {
...
+			acc[tableName] += size
+		}
+		return acc
+	}, make(map[string]int64))
+
+	maxTable, maxTableSize := lo.FindKeyBy(tableSizes, func(tableName string, size int64) bool {
+		return size == lo.MaxBy(lo.Values(tableSizes), func(a, b int64) bool {
+			return a > b
+		})
+	})
```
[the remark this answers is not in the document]
shekhar-rudder: Done with slight modification (Here output of reduce `tableSizes` is unused)

PR #5805, opened 2025-05-02
fix: warehouse cached schema mismatch

vyeshwanth:
# Description

Issue - In case of updates to a connection (source_1,destination_1, namespace_1), schema belonging to (source_2, destinaton_1, n1) is not valid. 

Fix -  In case of schema change, for one connection identified by three identifiers (sourceId, destinationId, namespace), we want to update the schema belonging to other connections involving same (destinationId, namespace) pair. 

## Linear Ticket
Part of WAR-606

## Security

### DIFF
--- a/warehouse/internal/repo/schema.go
+++ b/warehouse/internal/repo/schema.go
@@ -53,6 +53,31 @@ func (sh *WHSchema) Insert(ctx context.Context, whSchema *model.WHSchema) (int64
 		return id, fmt.Errorf("marshaling schema: %w", err)
 	}
 
+	// update all schemas with the same destination_id and namespace but different source_id
+	// this is to ensure all the connections for a destination have the same schema copy
+	_, err = sh.db.ExecContext(ctx, `
+		UPDATE `+whSchemaTableName+`
+		SET
+			schema = $1,
+			updated_at = $2,
+			expires_at = $3
+		WHERE
+			destination_id = $4 AND
+			namespace = $5 AND
+			source_id != $6;
+`,
+		schemaPayload,
+		now.UTC(),
+		whSchema.ExpiresAt,
+		whSchema.DestinationID,
+		whSchema.Namespace,
+		whSchema.SourceID,
+	)
+	if err != nil {
+		return id, fmt.Errorf("updating related schemas: %w", err)
+	}
+
+	// Then, insert/update the new schema using the unique constraint
 	err = sh.db.QueryRowContext(ctx, `
 		INSERT INTO `+whSchemaTableName+` (
           source_id, namespace, destination_id,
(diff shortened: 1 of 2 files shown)


warehouse/internal/repo/schema.go
```diff
@@ -53,21 +53,48 @@ func (sh *WHSchema) Insert(ctx context.Context, whSchema *model.WHSchema) (int64
...
+	// update all schemas with the same destination_id and namespace but different source_id
+	// this is to ensure all the connections for a destination have the same schema copy
+	_, err = sh.db.ExecContext(ctx, `
+		UPDATE `+whSchemaTableName+`
+		SET
+			schema = $1,
+			updated_at = $2,
+			expires_at = $3
+		WHERE
+			destination_id = $4 AND
```
vyeshwanth, line 65: we have index on `destination_id`, this is performing index scan

warehouse/internal/repo/schema.go
```diff
@@ -53,21 +53,48 @@ func (sh *WHSchema) Insert(ctx context.Context, whSchema *model.WHSchema) (int64
 		return id, fmt.Errorf("marshaling schema: %w", err)
 	}
 
+	// update all schemas with the same destination_id and namespace but different source_id
+	// this is to ensure all the connections for a destination have the same schema copy
+	_, err = sh.db.ExecContext(ctx, `
```
achettyiitr, line 58: Can we do this within a transaction?

vyeshwanth: done.

warehouse/internal/repo/schema.go
```diff
@@ -53,21 +53,54 @@ func (sh *WHSchema) Insert(ctx context.Context, whSchema *model.WHSchema) (int64
 		return id, fmt.Errorf("marshaling schema: %w", err)
 	}
 
-	err = sh.db.QueryRowContext(ctx, `
+	tx, err := sh.db.BeginTx(ctx, nil)
```
achettyiitr, line 56: Can we use repo.WithTx here?

achettyiitr:
```GO
	err = (*repo)(sh).WithTx(ctx, func(tx *sqlmiddleware.Tx) error {
		// update all schemas with the same destination_id and namespace but different source_id
		// this is to ensure all the connections for a destination have the same schema copy
		_, err = tx.ExecContext(ctx, `
		UPDATE `+whSchemaTableName+`
		SET
			schema = $1,
			updated_at = $2,
			expires_at = $3
		WHERE
			destination_id = $4 AND
			namespace = $5 AND
			source_id != $6;
`,
			schemaPayload,
			now.UTC(),
			whSchema.ExpiresAt,
			whSchema.DestinationID,
			whSchema.Namespace,
			whSchema.SourceID,
		)
		if err != nil {
			return fmt.Errorf("updating related schemas: %w", err)
		}

		// Then, insert/update the new schema using the unique constraint
		err = tx.QueryRowContext(ctx, `
		INSERT INTO `+whSchemaTableName+` (
          source_id, namespace, destination_id,
		  destination_type, schema, created_at,
		  updated_at, expires_at
		)
		VALUES
		  ($1, $2, $3, $4, $5, $6, $7, $8)
		ON CONFLICT (
			source_id, destination_id, namespace
		) DO
		UPDATE
		SET
		  schema = $5,
		  updated_at = $7,
		  expires_at = $8
		RETURNING id;
`,
			whSchema.SourceID,
			whSchema.Namespace,
			whSchema.DestinationID,
			whSchema.DestinationType,
			schemaPayload,
			now.UTC(),
			now.UTC(),
			whSchema.ExpiresAt,
		).Scan(&id)
		if err != nil {
			return fmt.Errorf("inserting schema: %w", err)
		}
		return nil
	})
	return id, err
```

vyeshwanth: done. using `WithTx` now

achettyiitr (approved):

RanjeetMishra (approved):

PR #5807, opened 2025-05-02
fix: add check for nil secret on oauthv2

ItsSudip:
# Description

We are adding validation for a nil secret being stored in the cache, for OAuthV2. 

## Linear Ticket

https://linear.app/rudderstack/issue/INT-3599/fix-add-validation-for-nil-secret-being-stored-in-cache-for-oauthv2

## Security

### DIFF
--- a/services/oauth/v2/oauth.go
+++ b/services/oauth/v2/oauth.go
@@ -356,6 +356,10 @@ func (h *OAuthHandler) GetRefreshTokenErrResp(response string, accountSecret *Ac
 	return errorType, message
 }
 
+func isEmptySecret(accountSecret AccountSecret) bool {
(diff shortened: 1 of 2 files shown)


services/oauth/v2/oauth.go
```diff
@@ -429,6 +429,15 @@ func (h *OAuthHandler) fetchAccountInfoFromCp(refTokenParams *RefreshTokenParams
 		}
 		return http.StatusInternalServerError, authResponse, fmt.Errorf("error occurred while fetching/refreshing account info from CP: %s", refErrMsg)
 	}
+
+	if accountSecret.Secret == nil {
```
cisse21, line 433: could it be an empty string too and not just nil?

koladilip (approved):

PR #5810, opened 2025-05-05
feat: use account to decide oauth type of a destination

ItsSudip:
# Description

We are introducing new way to checking oauth type of a destination. We are using account to detect oauth.

## Linear Ticket

## Security


services/oauth/v2/destination_info.go
```diff
@@ -16,9 +17,19 @@ type DestinationInfo struct {
...
 	Config           map[string]interface{}
+	Account          *backendconfig.AccountWithDefinition
 }
 
 func (d *DestinationInfo) IsOAuthDestination(flow common.RudderFlow) (bool, error) {
+	if d.Account != nil {
+		authValue, err := misc.NestedMapLookup(d.Account.AccountDefinition.Config, "refreshOAuthToken")
+		if err != nil {
+			return false, err
+		}
```
utsabc, line 28: Have a test for this case

utsabc: Maybe mock misc.NestedMapLookup to throw error

services/oauth/v2/destination_info_test.go
```diff
@@ -109,6 +112,65 @@ var isOAuthDestTestCases = []destInfoTestCase{
...
+		account: &backendconfig.AccountWithDefinition{
+			AccountDefinition: backendconfig.AccountDefinition{
+				Config: map[string]interface{}{
+					"refreshToken": false,
+				},
+			},
+		},
+		expected: isOAuthResult{
+			isOAuth: false,
+			err: &misc.MapLookupError{
```
krishna2020, line 167: should we throw an error here ? Config can have any other valid values (in future) for non-oauth destinations where `refreshOAuthToken` flag need not be

ItsSudip: yes

krishna2020 (approved):

krishna2020:
@ItsSudip tests are failing

utsabc (approved):

PR #5813, opened 2025-05-05
fix: convert bad request errors to 500 errors in oauth interceptor to prevent panics

koladilip:
## description

this pr addresses the issue described in [INT-2962](https://linear.app/rudderstack/issue/INT-2962/investigate-on-why-bad-request-is-being-sent-in-routertransform) where "bad request" errors in the routertransform component are causing system panics.

## changes

1. modified the oauth interceptor's `postroundtrip` method to handle "bad request" errors by converting them to 500 status codes (which are retryable) instead of propagating errors that could cause panics.

2. changed the return type of `postroundtrip` from `(*http.response, error)` to just `*http.response` since we're no longer returning errors from this function.

3. added tests to verify that "bad request" errors are properly handled and don't cause panics.

4. updated existing tests to reflect the new behavior (returning 500 errors instead of panicking).

## benefits

1. prevents panics: by handling errors gracefully, we prevent panics that could crash the system.

2. makes errors retryable: converting errors to 500 status codes makes them retryable, allowing the system to recover automatically.

3. preserves error information: we still log the original errors for debugging purposes.

## testing

added tests to verify that:

- "bad request" errors are properly handled and don't cause panics

- response body read errors are properly handled

- transport errors are properly handled

- updated existing tests that expected panics to now expect 500 errors

all tests are passing and the code passes linting checks.

Resolves: [INT-2962](https://linear.app/rudderstack/issue/INT-2962/investigate-on-why-bad-request-is-being-sent-in-routertransform)

### DIFF
--- a/services/oauth/v2/http/transport.go
+++ b/services/oauth/v2/http/transport.go
@@ -132,7 +132,7 @@ func (t *OAuthTransport) preRoundTrip(rts *roundTripState) *http.Response {
 	return httpResponseCreator(statusCode, []byte(err.Error()))
 }
 
-func (t *OAuthTransport) postRoundTrip(rts *roundTripState) (*http.Response, error) {
+func (t *OAuthTransport) postRoundTrip(rts *roundTripState) *http.Response {
 	respData, err := io.ReadAll(rts.res.Body)
 	if err != nil {
 		t.log.Errorn("[postRoundTrip] reading response body",
@@ -142,7 +142,8 @@ func (t *OAuthTransport) postRoundTrip(rts *roundTripState) (*http.Response, err
 			logger.NewStringField("flow", string(t.flow)),
 			logger.NewErrorField(err),
 		)
-		return nil, fmt.Errorf("reading response body post roundTrip: %w", err)
+		// Create a new response with a 500 status code
+		return httpResponseCreator(http.StatusInternalServerError, []byte(fmt.Sprintf("[postRoundTrip]Error reading response body: %v", err)))
 	}
 	interceptorResp := oauth.OAuthInterceptorResponse{}
 	// internal function
@@ -164,7 +165,12 @@ func (t *OAuthTransport) postRoundTrip(rts *roundTripState) (*http.Response, err
 			logger.NewStringField("flow", string(t.flow)),
 			logger.NewErrorField(errors.New(string(respData))),
 		)
-		return nil, fmt.Errorf("getting auth error category post roundTrip: %s", string(respData))
+		// Instead of returning an error, set a 500 status code in the interceptor response
+		// This will make the error retryable instead of causing a panic
+		interceptorResp.StatusCode = http.StatusInternalServerError
+		interceptorResp.Response = "[postRoundTrip]Error processing response: " + string(respData)
+		applyInterceptorRespToHttpResp()
+		return rts.res
 	}
 
 	switch authErrorCategory {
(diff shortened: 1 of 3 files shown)


services/oauth/v2/http/transport.go
krishna2020, line 1: As we are handling the failures, do we get an alert whenever transport layer encounters platform issues ?

koladilip: we can update alert we have to capture these new error logs and alert us

services/oauth/v2/http/transport.go
```diff
@@ -164,7 +165,12 @@ func (t *OAuthTransport) postRoundTrip(rts *roundTripState) (*http.Response, err
 			logger.NewStringField("flow", string(t.flow)),
 			logger.NewErrorField(errors.New(string(respData))),
 		)
-		return nil, fmt.Errorf("getting auth error category post roundTrip: %s", string(respData))
+		// Instead of returning an error, set a 500 status code in the interceptor response
+		// This will make the error retryable instead of causing a panic
+		interceptorResp.StatusCode = http.StatusInternalServerError
+		interceptorResp.Response = "Error processing response: " + string(respData)
```
utsabc, line 171: Lets add unique error message to easily understand where this error is happening from.

```suggestion
		interceptorResp.Response = "[In Oauth PostRoundTrip]Error processing response: " + string(respData)
```

services/oauth/v2/http/transport.go
```diff
@@ -178,7 +184,11 @@ func (t *OAuthTransport) postRoundTrip(rts *roundTripState) (*http.Response, err
 				obskit.WorkspaceID(rts.destination.WorkspaceID),
 				obskit.DestinationType(rts.destination.DefinitionName),
 				logger.NewStringField("flow", string(t.flow)))
-			return nil, fmt.Errorf("getting secret from context post roundTrip")
+			// Instead of returning an error, set a 500 status code in the interceptor response
+			interceptorResp.StatusCode = http.StatusInternalServerError
+			interceptorResp.Response = "Error getting secret from context"
```
utsabc, line 189: Same as above
```suggestion
			interceptorResp.Response = "[In Oauth PostRoundTrip]Error getting secret from context"
```

utsabc (approved):

services/oauth/v2/http/transport.go
krishna2020: we have 2 alert definitions that work on logs - yeah, let's make sure alert definition works on these changes too - https://www.notion.so/rudderstacks/oauthv2-alerts-d33f99ad48144e36b6ec188a59222036#153c426b67484f2cbe65bf9e1ca7f8df

koladilip: we will update once this is rolled out

services/oauth/v2/http/transport.go
utsabc: But its easier to read the error messages, and its not a lot of information overload

commit_msg.txt
```diff
@@ -0,0 +1,10 @@
+fix: convert bad request errors to 500 errors in oauth interceptor to prevent panics
+
+this change addresses the issue described in INT-2962 where bad request errors in the routertransform component are causing system panics. by converting these errors to 500 status codes, they become retryable instead of causing panics.
+
+- modified oauth interceptor to handle bad request errors gracefully
+- changed return type of postroundtrip to remove unnecessary error return
+- added tests to verify error handling behavior
+- updated existing tests to reflect new behavior (returning 500 errors instead of panicking)
+
+Resolves: INT-2962
```
cisse21, line 10: Do we need this file?

cisse21 (approved):

commit_msg.txt
koladilip: I will remove this

services/oauth/v2/http/transport.go
koladilip: where will you read this error message from? in events tab this won't be available till it has become terminal state.

PR #5827, opened 2025-05-08
chore: warehouse transformer migration to embedded destination transformer package

achettyiitr:
# Description

- Wanted to integrate the warehouse transformer within the embedded destination transformer package (https://github.com/rudderlabs/rudder-server/pull/5828), but it is creating a cyclic dependency between:
  - github.com/rudderlabs/rudder-server/processor/internal/transformer/
  - github.com/rudderlabs/rudder-server/processor/internal/transformer/destination_transformer/
  - github.com/rudderlabs/rudder-server/warehouse/transformer/
- It also makes sense to keep the warehouse transformer package within the embedded destination transformer package, which contains packages for `pubsub` and `kafka`, as there is no dependency on the warehouse package.

## Linear Ticket

- Resolves WAR-629

## Security


vyeshwanth (approved):

processor/internal/transformer/destination_transformer/embedded/warehouse/internal/model/model.go
```diff
@@ -0,0 +1,15 @@
+package model
+
+type (
+	SchemaType = string
+)
+
+const (
+	StringDataType   SchemaType = "string"
```
shekhar-rudder, line 8: Where are we using these new types?

achettyiitr: Earlier these constants were defined in [here](https://github.com/rudderlabs/rudder-server/blob/master/warehouse/internal/model/schema.go#L11-L21). Now since it is internal to the warehouse package, it can't be used. 
These are being used `datatype.go`.

shekhar-rudder (approved):

PR #5828, opened 2025-05-08
chore: warehouse transformer uploader

achettyiitr:
# Description

- Integrating warehouse transformer within the destination transformer embedded package.
  - `Processor.enableWarehouseTransformations` is by default false. We would need to enable it to start using warehouse transformations.
  - `Processor.verifyWarehouseTransformations` is by default true. If we want to go production, we just need to disable it to stop the comparison logic.
- Instead of logging, Uploading sample events to where we see differences in responses for legacy and embedded.

## Linear Ticket

- Resolves WAR-443, WAR-493

## Security


vyeshwanth (approved):

processor/internal/transformer/destination_transformer/embedded/warehouse/uploader.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/rudderlabs/rudder-go-kit/logger"
+	"github.com/rudderlabs/rudder-go-kit/stringify"
+
+	wtypes "github.com/rudderlabs/rudder-server/processor/internal/transformer/destination_transformer/embedded/warehouse/internal/types"
+	"github.com/rudderlabs/rudder-server/processor/types"
+)
+
+func (t *Transformer) CompareResponsesAndUpload(ctx context.Context, events []types.TransformerEvent, legacyResponse types.Response) {
+	if t.loggedSamplesUploader == nil {
```
shekhar-rudder, line 31: In which case are we not setting `loggedSamplesUploader`?

achettyiitr: In case of error.
```
        var err error
	t.loggedSamplesUploader, err = getSamplingUploader(t.conf, t.logger)
	if err != nil {
		t.logger.Errorn("Failed to create wt sampling file manager", obskit.Error(err))
	}
```

shekhar-rudder (approved):

PR #5835, opened 2025-05-09
fix: update account type to remove id which we are getting as map key

ItsSudip:
# Description

This PR optimizes the `Account` data structure by removing the redundant `Id` field from the `Account` struct. Since we're now using a map structure where the account ID is already the key (`map[string]Account`), storing the same ID inside the struct is redundant and inefficient.

## Changes

- Removed the `Id` field from the `Account` struct in `types.go`
- Updated the `setDestinationAccounts` method to use the map key (account ID) directly when creating `AccountWithDefinition` instances
- Removed redundant `Id` field references from all test cases in `account_association_test.go`

## Benefits

- Eliminates data duplication by not storing the same ID twice (as map key and inside the struct)
- Reduces memory usage by removing redundant fields
- Improves code clarity by making it explicit that the map key is the account ID
- Follows the optimization path started with the previous PR that converted arrays to maps

## Linear Ticket

This PR addresses [INT-3622](https://linear.app/rudderstack/issue/INT-3622/optimize-account-and-accountdefinition-merge-to-destinations), which focuses on optimizing the account and account definition merge process for destinations.

## Security

### DIFF
--- a/backend-config/account_association.go
+++ b/backend-config/account_association.go
@@ -46,7 +46,7 @@ func (c *ConfigT) setDestinationAccounts(dest *DestinationT) {
 					logger.NewStringField("destinationId", dest.ID))
 			}
 			dest.DeliveryAccount = &AccountWithDefinition{
-				Id:                account.Id,
+				Id:                accountID,
 				Options:           account.Options,
 				Secret:            account.Secret,
 				AccountDefinition: accountDefinition,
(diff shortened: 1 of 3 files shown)


backend-config/account_association.go
```diff
@@ -46,7 +46,7 @@ func (c *ConfigT) setDestinationAccounts(dest *DestinationT) {
 					logger.NewStringField("destinationId", dest.ID))
 			}
 			dest.DeliveryAccount = &AccountWithDefinition{
-				Id:                account.Id,
+				Id:                accountID,
```
utsabc, line 49: Can we add test reg this?

ItsSudip: already added can you check once?

utsabc (approved):

cisse21 (approved):

PR #5836, opened 2025-05-09
chore: enrich event with bot details

itsmihir:
# Description

Add bot encicher, which will add bot details present in event parameters to event context. 

## Linear Ticket

[Resolves PIPE-2028] (https://linear.app/rudderstack/issue/PIPE-2028/implement-event-flagging)

## Depedency

https://github.com/rudderlabs/rudder-schemas/pull/66

## Security

### DIFF
--- a/app/apphandlers/setup.go
+++ b/app/apphandlers/setup.go
@@ -150,5 +150,15 @@ func setupPipelineEnrichers(conf *config.Config, log logger.Logger, stats stats.
 		enrichers = append(enrichers, geoEnricher)
 	}
 
+	if conf.GetBool("BotEnrichment.enabled", true) {
+		log.Infof("Setting up the bot pipeline enricher")
+
+		botEnricher, err := enricher.NewBotEnricher()
+		if err != nil {
+			return nil, fmt.Errorf("starting bot enrichment process for pipeline: %w", err)
+		}
+		enrichers = append(enrichers, botEnricher)
+	}
+
 	return enrichers, nil
 }
--- a/internal/enricher/bot.go
+++ b/internal/enricher/bot.go
@@ -1,0 +1,56 @@
+package enricher
+
+import (
+	"errors"
+
+	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
+	"github.com/rudderlabs/rudder-server/processor/types"
+)
+
+type BotDetails struct {
+	Name             string `json:"name,omitempty"`
+	URL              string `json:"url,omitempty"`
+	IsInvalidBrowser bool   `json:"isInvalidBrowser,omitempty"`
+}
+
+type botEnricher struct{}
+
+func NewBotEnricher() (PipelineEnricher, error) {
+	return &botEnricher{}, nil
+}
+
+func (e *botEnricher) Enrich(_ *backendconfig.SourceT, request *types.GatewayBatchRequest, eventParams *types.EventParams) error {
+	var enrichErrs []error
+	for _, event := range request.Batch {
+		// if the event is not a bot, we don't need to enrich it
+		if !eventParams.IsBot {
+			continue
+		}
+
+		// if the context section is missing on the event
+		// set it with default as map[string]interface{}
+		if _, ok := event["context"]; !ok {
+			event["context"] = map[string]interface{}{}
+		}
+
+		// if the context is other than map[string]interface{}, add error and continue
+		context, ok := event["context"].(map[string]interface{})
+		if !ok {
+			enrichErrs = append(enrichErrs, errors.New("event doesn't have a valid context section"))
+			continue
+		}
+
+		context["isBot"] = true
+		context["bot"] = BotDetails{
+			Name:             eventParams.BotName,
+			URL:              eventParams.BotURL,
+			IsInvalidBrowser: eventParams.BotIsInvalidBrowser,
+		}
+	}
+
+	return errors.Join(enrichErrs...)
+}
+
+func (e *botEnricher) Close() error {
+	return nil
+}
--- a/internal/enricher/bot_test.go
+++ b/internal/enricher/bot_test.go
@@ -1,0 +1,202 @@
+package enricher
+
+import (
+	"testing"
+
+	"github.com/rudderlabs/rudder-server/processor/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBotEnricher(t *testing.T) {
+	tests := []struct {
+		name           string
+		request        *types.GatewayBatchRequest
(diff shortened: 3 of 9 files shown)


itsmihir:
@coderabbitai full review

internal/enricher/bot.go
```diff
@@ -0,0 +1,56 @@
...
+
+func NewBotEnricher() (PipelineEnricher, error) {
+	return &botEnricher{}, nil
+}
+
+func (e *botEnricher) Enrich(_ *backendconfig.SourceT, request *types.GatewayBatchRequest, eventParams *types.EventParams) error {
+	var enrichErrs []error
+	for _, event := range request.Batch {
+		// if the event is not a bot, we don't need to enrich it
+		if !eventParams.IsBot {
```
itsmihir, line 26: @ktgowtham If the bot is not detected, we are not adding any fields, are we good with this?

app/apphandlers/setup.go
```diff
@@ -150,5 +150,15 @@
 		enrichers = append(enrichers, geoEnricher)
 	}
 
+	if conf.GetBool("BotEnrichment.enabled", true) {
+		log.Infof("Setting up the bot pipeline enricher")
```
fracasula, line 154: Non-Sugared logger please. Always. Even if you don't pass fields.

fracasula: Actually, can you change the other ones in this file as well?

internal/enricher/bot.go
```diff
@@ -0,0 +1,56 @@
+package enricher
+
+import (
+	"errors"
+
+	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
+	"github.com/rudderlabs/rudder-server/processor/types"
+)
+
+type BotDetails struct {
```
fracasula, line 10: This type doesn't seem to be used outside the package. Do we need to export it?

internal/enricher/bot_test.go
```diff
@@ -0,0 +1,203 @@
...
+				},
+			},
+			expectError: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			enricher, err := NewBotEnricher()
+			assert.NoError(t, err)
```
fracasula, line 191: * `assert.NoError()`: If the assertion fails, the test will **continue execution**. The failure is marked, but the rest of the test will still run.
* `require.NoError()`: If the assertion fails, the test will **immediately terminate**. No further code in that test function will be executed.

How come are you using `assert` here and not `require`?

fracasula (changes requested):

internal/enricher/bot_test.go
itsmihir: Updated to use require

app/apphandlers/setup.go
fracasula: There are more up above, `Info` instead of `Infon`. It'd be nice to have the whole as per standards given how small the file is.

fracasula (approved):
LGTM. It would be nice to have the whole `setup.go` use the non-sugared logger though.

app/apphandlers/setup.go
itsmihir: Updated other places also

processor/processor_bot_enricher_test.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	func() { kithttputil.CloseResponse(resp) }()
+
+	return nil
+}
+
+func (s *botScenario) requireJobsCount(t *testing.T, db *sql.DB, queue, state string, expectedCount int) {
+	require.Eventually(t, func() bool {
+		var jobsCount int
+		require.NoError(t, db.QueryRow(fmt.Sprintf("SELECT count(*) FROM unionjobsdbmetadata('%s',10) WHERE job_state = '%s'", queue, state)).Scan(&jobsCount))
```
fracasula, line 273: `require` makes the test fail right away. Both `require` and `assert` must be avoided inside `Eventually`. The function invoked by `Eventually` is not meant to do any kind of assertion, it is meant to return just `true` or `false`. If it keeps returning `false` for the specified timeout (in this case 20s) then `Eventually` itself will make the test fail. The way you wrote it might make the test fail at the first attempt without the retries every each second.

fracasula: I suggest you just do `return err != nil && jobsCount == expectedCount`.

processor/processor_bot_enricher_test.go
```diff
@@ -0,0 +1,277 @@
...
+		trServer.Close()
+		return err
+	})
+	url := fmt.Sprintf("http://localhost:%d", gwPort)
+	health.WaitUntilReady(ctx, t, url+"/health", 60*time.Second, 10*time.Millisecond, t.Name())
+	return url, postgresContainer.DB, cancel, wg
+}
+
+func (s *botScenario) runRudderServer(ctx context.Context, port int, postgresContainer *postgres.Resource, cbURL, transformerURL, tmpDir string) (err error) {
+	config.Set("enableStats", false)
```
fracasula, line 181: It's a shame that we are still using global variables in our test. I understand that refactoring the runner to make it use a `*config.Config` instance is beyond the scope of this PR, just making a note of this.

You can leave it as is or use `t.Setenv`. Sooner or later we should remove this as it could lead to undesired behaviour cc @ktgowtham

fracasula: This is also preventing us to use `t.Parallel()`.

fracasula (approved):
LGTM but please fix the new test with `Eventually`.

internal/enricher/bot_test.go
```diff
@@ -0,0 +1,225 @@
...
+			name: "non-bot event with name, url and isInvalidBrowser should not be enriched",
+			request: &types.GatewayBatchRequest{
+				Batch: []types.SingularEventT{
+					{
+						"event": "test-event",
+					},
+				},
+			},
+			eventParams: &types.EventParams{
+				IsBot:               false,
```
ktgowtham, line 48: ideally we won't have this test case

processor/processor_bot_enricher_test.go
ktgowtham: https://linear.app/rudderstack/issue/PIPE-2067/get-rid-of-global-variables-in-tests

ktgowtham (approved):

internal/enricher/bot_test.go
itsmihir: Yes, added just in case we introduce a bug with future changes/refactoring.

PR #5842, opened 2025-05-12
fix: reporting common client path with query

lvrach:
# Description

url.JoinPath has a side effect of encoding query parameter. Causing `/metrics?version=v1` to be used as path, rather than `/metrics` with query `?version=v1`.

Added a test to cover this.

## Linear Ticket

Resolves PIPE-2062
https://linear.app/rudderstack/issue/PIPE-2062/fix-reporting-client-path-with-query

## Security

### DIFF
--- a/enterprise/reporting/client/client.go
+++ b/enterprise/reporting/client/client.go
@@ -31,18 +31,18 @@ const (
 )
 
 const (
-	ServiceMetrics      ServiceRoute = "/metrics?version=v1"
-	ServiceRecordErrors ServiceRoute = "/recordErrors"
-	ServiceTrackedUsers ServiceRoute = "/trackedUser"
+	RouteMetrics      Route = "/metrics?version=v1"
+	RouteRecordErrors Route = "/recordErrors"
+	RouteTrackedUsers Route = "/trackedUser"
 )
 
-// ServiceRoute contains the HTTP path and query string for the service.
-type ServiceRoute string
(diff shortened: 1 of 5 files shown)


enterprise/reporting/client/client.go
```diff
@@ -92,10 +92,7 @@ func (c *Client) Send(ctx context.Context, payload any) error {
 		return err
 	}
 
-	u, err := url.JoinPath(c.reportingServiceURL, string(c.path))
-	if err != nil {
-		return err
-	}
+	u := strings.TrimSuffix(c.reportingServiceURL, "/") + string(c.path)
```
atzoum, line 95: can we use `url.JoinPath`?

satishrudderstack (approved):

enterprise/reporting/client/client.go
lvrach: I wasn't happy with regressing to previous implementation either. Your comment nudge me, to come up with a cleaner approach.

I've move the URL construction [here](https://github.com/rudderlabs/rudder-server/pull/5842/files#diff-2c09416a82ca9dbb68294523ea370de02bfcfaab73c98501cc778d20717fe441R45).

* Applying URL parse for both base URL and service route. Properly handling query parameters.
* Rename `Path` to `ServiceRoute`, to semantically encapsulate query params.

satishrudderstack (approved):

atzoum (approved):

PR #5848, opened 2025-05-13
fix: embedded transformations upload

achettyiitr:
# Description

- Use a go-routine when you only need to compare and upload. In case of early return, no need for a go-routine. In most cases, when the limit is reached, we will be unnecessarily spinning a go-routine which is not required.
- Use warehouse transformations for snowpipe streaming as well.
- Resolves WAR-654

## Security

### DIFF
--- a/processor/internal/transformer/destination_transformer/logger.go
+++ b/processor/internal/transformer/destination_transformer/logger.go
@@ -25,13 +25,19 @@ func (c *Client) CompareAndLog(
 		c.log.Warnn("DestinationTransformer sanity check failed")
 		return
 	}
-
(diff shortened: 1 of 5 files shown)


processor/internal/transformer/destination_transformer/logger.go
```diff
@@ -25,13 +25,19 @@ func (c *Client) CompareAndLog(
 		c.log.Warnn("DestinationTransformer sanity check failed")
 		return
 	}
-
 	if c.loggedEvents.Load() >= int64(c.config.maxLoggedEvents.Load()) {
```
itsmihir, line 28: can make `loggedEvents` non-atomic?

itsmihir (approved):

ktgowtham (approved):

PR #5865, opened 2025-05-17
feat: update isOauth function to decide with account

ItsSudip:
# Description

< Replace with adequate description for this PR as per [Pull Request document](https://www.notion.so/rudderstacks/Pull-Requests-40a4c6bd7a5e4387ba9029bab297c9e3) >

## Linear Ticket

< Replace with Linear Link ( [create](https://linear.new?title=Awesome-pr-without-linear-ticket) or [search](https://linear.app/rudderstack/search) linear ticket) or  >

## Security

- [ ] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #5866, opened 2025-05-19
feat: update isOauthFunction to consider account to decide

ItsSudip:
# Description

This changes the `IsOAuthDestination` function in services/oauth/v2/destination_info.go to consider account information when determining if a destination uses OAuth. Specifically, it:

- Adds a new field Account of type *backendconfig.AccountWithDefinition to the DestinationInfo struct
- Updates the IsOAuthDestination function to check the account's configuration for a refreshOAuthToken boolean value when the destination's definition doesn't specify an auth type

## Linear Ticket

## Security

### DIFF
--- a/services/oauth/v2/destination_info.go
+++ b/services/oauth/v2/destination_info.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/samber/lo"
 
+	backendconfig "github.com/rudderlabs/rudder-server/backend-config"
 	"github.com/rudderlabs/rudder-server/services/oauth"
 	"github.com/rudderlabs/rudder-server/services/oauth/v2/common"
 	"github.com/rudderlabs/rudder-server/utils/misc"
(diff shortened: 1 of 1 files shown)


services/oauth/v2/destination_info.go
```diff
@@ -16,12 +17,19 @@ type DestinationInfo struct {
 	DefinitionConfig map[string]interface{}
```
utsabc, line 17: For now add the exact type then we don't need nestedMap lookup

DefinitionConfig = struct {
    refreshOAuthToken bool
}

services/oauth/v2/destination_info.go
```diff
@@ -16,12 +17,19 @@ type DestinationInfo struct {
...
+		if d.Account != nil {
+			authValue, ok := d.Account.AccountDefinition.Config["refreshOAuthToken"].(bool)
+			if !ok {
+				return false, nil
+			}
+			return authValue, nil
+		}
 		return false, nil
 	}
 	authType, ok := authValue.(string)
```
utsabc, line 35: Why only when authValue === nil we are doing this its not either or condition

ItsSudip:
No need for this pr. Will move this changes to [this](https://github.com/rudderlabs/rudder-server/pull/5865) pr.

PR #5869, opened 2025-05-19
chore: use jsonrs from rudder-go-kit

mihir20:
# Description

use jsonrs from rudder-go-kit

## Linear Ticket

pipe-1997

## Security


atzoum (commented):
we no longer need this line
https://github.com/rudderlabs/rudder-server/blob/master/.golangci.yml#L23

atzoum (approved):

fracasula (approved):

PR #5872, opened 2025-05-19
chore: remove redundant structur for account with definition

ItsSudip:
# Description

We are removing redundant structure `accountWithDefinition`. This name is misleading. It is possible AccountDefinition can be undefined in account. So made it optional in account.

## Linear Ticket

## Security


koladilip:
I've made the following changes to the PR:

1. Refactored the `setDestinationAccounts` method in `account_association.go` to use a new utility function `populateAccountToDestination` that handles populating an account to a destination based on the account ID

2. Added a test for JSON serialization/deserialization of the `DestinationT` struct with accounts to ensure that the JSON field names (`deliveryAccount` and `deleteAccount`) are correctly handled

3. Reverted the changes to skip Docker-dependent tests in `backend_config_test.go` as requested

These changes improve code maintainability by reducing duplication and adding test coverage for the JSON serialization aspects of the refactored code.

koladilip (approved):

backend-config/account_association.go
```diff
@@ -21,6 +21,41 @@ func (c *ConfigT) processAccountAssociations() {
...
+//
+// Parameters:
+//   - accountID: The ID of the account to populate
+//   - dest: Pointer to the destination being configured
+//   - accountField: The field name in the destination config (e.g., "rudderAccountId" or "rudderDeleteAccountId")
+//   - flowType: The type of flow (e.g., "delivery" or "regulation")
+//
+// Returns:
+//   - *Account: The populated account object or nil if the account doesn't exist
+func (c *ConfigT) populateAccountToDestination(accountID string, dest *DestinationT, accountField, flowType string) *Account {
```
utsabc, line 35:
```suggestion
func (c *ConfigT) populateAccountToDestination(accountID string, destination *DestinationT, accountField, flowType string) *Account {
```

lets use fully qualified naming convention

backend-config/account_association.go
```diff
@@ -21,6 +21,41 @@ func (c *ConfigT) processAccountAssociations() {
...
+				logger.NewStringField("accountDefinitionName", account.AccountDefinitionName),
+				logger.NewStringField("destinationId", dest.ID))
+		}
+		return &Account{
+			ID:                    accountID,
+			AccountDefinitionName: account.AccountDefinitionName,
+			Options:               account.Options,
+			Secret:                account.Secret,
+			AccountDefinition:     &accountDefinition,
+		}
```
utsabc, line 50: Copilot:
If the account definition does not exist for the given account.AccountDefinitionName, the code still proceeds to return &accountDefinition.

On the line:
accountDefinition, exists := c.AccountDefinitions[account.AccountDefinitionName]

If the definition is not present, accountDefinition will be the zero value of the AccountDefinition type (since map lookups in Go return the zero value when the key does not exist).

The function then returns &accountDefinition, which is a pointer to this zero value.

explanation:
If the account definition does not exist, the variable accountDefinition will hold the zero value for its type, and &accountDefinition will be a pointer to this zero value. This means Account.AccountDefinition will not be nil, but it will point to a default-constructed, empty AccountDefinition struct.

Utsab:
The` &accountDefinition` will not be `nil` if thats the case 
- Should we explicitly set `nil`
- Lets test this scenario and see what are getting as `&accountDefinition`

koladilip: there is problem here, good catch @utsabc

koladilip:
I've addressed the issue raised by @utsabc regarding the handling of non-existent account definitions. The changes include:

1. **Fixed the issue with `&accountDefinition`**: When `AccountDefinitionName` is blank or doesn't exist in the map, we now set `AccountDefinition` to `nil` instead of a pointer to an empty struct.

2. **Added condition for warning logs**: Only log warnings when `AccountDefinitionName` is not empty but doesn't exist in the map. This avoids unnecessary warnings when `AccountDefinitionName` is intentionally blank.

3. **Added a specific test case**: Created a test for blank account definition names that verifies the `AccountDefinition` field is properly set to `nil`.

This ensures that when an account has a blank account definition name, the code doesn't create a misleading pointer to an empty account definition struct, which could cause issues elsewhere in the codebase.

koladilip:
I've further refactored the code to make it more elegant by extracting a dedicated helper function `getAccountDefinition` that handles the account definition lookup logic:

```go
func (c *ConfigT) getAccountDefinition(accountDefinitionName, accountID, accountField, destinationID, flowType string) *AccountDefinition {
    if accountDefinitionName == "" {
        return nil
    }

    if accountDefinition, exists := c.AccountDefinitions[accountDefinitionName]; exists {
        return &accountDefinition
    }

    // Log warning if account definition not found
    pkgLogger.Warnn("Account definition not found in configured accountDefinitions for "+flowType+" flow",
        logger.NewStringField(accountField, accountID),
        logger.NewStringField("accountDefinitionName", accountDefinitionName),
        logger.NewStringField("destinationId", destinationID))
    
    return nil
}
```

This change:

1. **Improves readability**: The function name clearly describes its purpose
2. **Simplifies the main function**: The `populateAccountToDestination` function is now more focused on its core responsibility
3. **Promotes reusability**: The account definition lookup logic can be reused elsewhere if needed
4. **Maintains the same behavior**: All tests continue to pass

The code now handles the case of blank `AccountDefinitionName` more explicitly and elegantly.

koladilip:
I've changed the log level from warning (`Warnn`) to error (`Errorn`) for missing account definitions and accounts. This is more appropriate since these are configuration errors that should be addressed, not just warnings that can be ignored.

backend-config/account_association.go
```diff
@@ -21,6 +21,73 @@ func (c *ConfigT) processAccountAssociations() {
...
+//   - destinationID: The ID of the destination (used for logging)
+//   - flowType: The type of flow (used for logging)
+//
+// Returns:
+//   - *AccountDefinition: A pointer to the account definition or nil if not found
+func (c *ConfigT) getAccountDefinition(accountDefinitionName, accountID, accountField, destinationID, flowType string) *AccountDefinition {
+	if accountDefinitionName == "" {
+		return nil
+	}
+
```
atzoum, line 40: I don't understand why we need this, what it protects us from

backend-config/account_association.go
```diff
@@ -21,6 +21,73 @@ func (c *ConfigT) processAccountAssociations() {
...
+	if account, exists := c.Accounts[accountID]; exists {
+		accountDefinitionPtr := c.getAccountDefinition(
+			account.AccountDefinitionName,
+			accountID,
+			accountField,
+			destination.ID,
+			flowType,
+		)
+
+		return &Account{
```
atzoum, line 75: `account` is already a copy of the value, no need to create another copy and risk forgetting to set some fields in future extensions.

backend-config/account_association.go
```diff
@@ -21,6 +21,73 @@ func (c *ConfigT) processAccountAssociations() {
...
+// Parameters:
+//   - accountDefinitionName: The name of the account definition to look up
+//   - accountID: The ID of the account (used for logging)
+//   - accountField: The field name in the destination config (used for logging)
+//   - destinationID: The ID of the destination (used for logging)
+//   - flowType: The type of flow (used for logging)
+//
+// Returns:
+//   - *AccountDefinition: A pointer to the account definition or nil if not found
+func (c *ConfigT) getAccountDefinition(accountDefinitionName, accountID, accountField, destinationID, flowType string) *AccountDefinition {
```
atzoum, line 36: Passing arguments just for using them in a log statement isn't a good design choice. We can simply pass a properly configured logger instead

backend-config/account_association.go
```diff
@@ -21,6 +21,73 @@ func (c *ConfigT) processAccountAssociations() {
...
+//
+// Parameters:
+//   - accountID: The ID of the account to populate
+//   - destination: Pointer to the destination being configured
+//   - accountField: The field name in the destination config (e.g., "rudderAccountId" or "rudderDeleteAccountId")
+//   - flowType: The type of flow (e.g., "delivery" or "regulation")
+//
+// Returns:
+//   - *Account: The populated account object or nil if the account doesn't exist
+func (c *ConfigT) populateAccountToDestination(accountID string, destination *DestinationT, accountField, flowType string) *Account {
```
atzoum, line 65: Passing arguments just for using them in a log statement isn't a good design choice. We can simply pass a properly configured logger instead

atzoum (commented):
Please consider simplifying to the following two methods
```go
// setDestinationAccounts assigns accounts to a destination based on its configuration.
// It handles two types of account associations:
//
//  1. Regular account [rudderAccountId] - Used for normal event delivery flow
//  2. Delete account [rudderDeleteAccountId] - Used for data deletion/regulation flow
//
// For each account type, it:
//
//  1. Checks if the corresponding account ID exists in the destination config
//  2. Verifies the account exists in the accounts map
//  3. Creates a copy of the account which contains its definition as well
//  4. Assigns it to the appropriate field in the destination
//
// Parameters:
//   - dest: Pointer to the destination being configured
func (c *ConfigT) setDestinationAccounts(dest *DestinationT) {
	// Check and set the regular account if specified in the destination config
	if accountID, ok := dest.Config["rudderAccountId"].(string); ok && accountID != "" {
		dest.DeliveryAccount = c.populateAccountToDestination(accountID, pkgLogger.Withn(
			logger.NewStringField("accountId", accountID),
			logger.NewStringField("destinationId", dest.ID),
			logger.NewStringField("flowType", "delivery"),
		))
	}

	// Check and set the delete account if specified in the destination config
	if deleteAccountID, ok := dest.Config["rudderDeleteAccountId"].(string); ok && deleteAccountID != "" {
		dest.DeleteAccount = c.populateAccountToDestination(deleteAccountID, pkgLogger.Withn(
			logger.NewStringField("accountId", deleteAccountID),
			logger.NewStringField("destinationId", dest.ID),
			logger.NewStringField("flowType", "regulation"),
		))
	}
}

// populateAccountToDestination populates an account to a destination field based on the account ID.
// It creates an copy of the Account object by combining account details with its definition.
//
// Parameters:
//   - accountID: The ID of the account to populate
//   - log: The logger to use for logging errors
//
// Returns:
//   - *Account: The populated account object or nil if the account doesn't exist
func (c *ConfigT) populateAccountToDestination(accountID string, log logger.Logger) *Account {
	if account, ok := c.Accounts[accountID]; ok {
		if accountDefinition, ok := c.AccountDefinitions[account.AccountDefinitionName]; ok {
			account.AccountDefinition = &accountDefinition
		} else {
			log.Errorn("Account definition not found in configured definitions",
				logger.NewStringField("definition", account.AccountDefinitionName),
			)
		}
		return &account
	}
	pkgLogger.Errorn("Account not found in configured accounts")
	return nil
}
```

backend-config/account_association.go
ItsSudip: We are already using accounts for oauth destinations. So for those existing accounts we are not going to get any accountDefinitionName. Returning nil will help us segregating account with accountDefinition and account without accountDefinition.

backend-config/account_association.go
```diff
@@ -21,6 +24,50 @@ func (c *ConfigT) processAccountAssociations() {
...
+//   - *AccountDefinition: A pointer to the account definition or nil if not found
+func (c *ConfigT) getAccountDefinition(accountDefinitionName string, accountAssociationLogger logger.Logger) *AccountDefinition {
+	if accountDefinitionName == "" {
+		return nil
+	}
+	if accountDefinition, exists := c.AccountDefinitions[accountDefinitionName]; exists {
+		return &accountDefinition
+	}
+	// Log error if account definition not found
+	accountAssociationLogger.Errorn("Account definition not found in configured accountDefinitions")
```
utsabc, line 44: Shall we also log the `accountDefinitionName` without which this log doesn't give meaningful insight

backend-config/account_association.go
```diff
@@ -21,6 +24,50 @@ func (c *ConfigT) processAccountAssociations() {
...
+	if account, exists := c.Accounts[accountID]; exists {
+		accountDefinitionPtr := c.getAccountDefinition(
+			account.AccountDefinitionName,
+			accountAssociationLogger,
+		)
+		account.AccountDefinition = accountDefinitionPtr
+		account.ID = accountID
+		return &account
+	}
+	accountAssociationLogger.Errorn("Account not found in configured accounts")
```
utsabc, line 67: Lets also print `accountId` for the same reason as previous

ItsSudip: It is already added to the logger itself. The log will show as below:
`2025-05-21T09:12:25.400+0530	ERROR	backend-config	backend-config/account_association.go:67	Account not found in configured accounts	{"destinationId": "2xO9bhFMegViLzux9VDWLCmsDAJ", "destinationType": "GOOGLE_ADWORDS_REMARKETING_LISTS", "rudderAccountId": "2xO9Z1yXE0Chr51OBFj0WVZbIm1", "flowType": "delivery"}`
Any other information do you think we need to show?

backend-config/account_association_test.go
```diff
@@ -238,7 +240,102 @@ func TestAccountAssociations(t *testing.T) {
 
 		c.processAccountAssociations()
 
-		require.Equal(t, "acc-1", c.Sources[0].Destinations[0].DeliveryAccount.Id)
-		require.Empty(t, c.Sources[0].Destinations[0].DeliveryAccount.AccountDefinition)
+		require.Equal(t, "acc-1", c.Sources[0].Destinations[0].DeliveryAccount.ID)
+		require.Nil(t, c.Sources[0].Destinations[0].DeliveryAccount.AccountDefinition)
+	})
+
+	t.Run("blank account definition name", func(t *testing.T) {
```
utsabc, line 247: Can you add more description in the test case like what is the expectation for this scenario

utsabc: Like we will expect an associated account with destination but it has nil accountDefinition

backend-config/account_association_test.go
```diff
@@ -238,7 +240,102 @@ func TestAccountAssociations(t *testing.T) {
...
+		require.Equal(t, "", c.Sources[0].Destinations[0].DeliveryAccount.AccountDefinitionName)
+		require.Equal(t, map[string]interface{}{"key1": "value1"}, c.Sources[0].Destinations[0].DeliveryAccount.Options)
+		require.Equal(t, map[string]interface{}{"secret1": "secretValue1"}, c.Sources[0].Destinations[0].DeliveryAccount.Secret)
+		require.Nil(t, c.Sources[0].Destinations[0].DeliveryAccount.AccountDefinition)
+
+		// Note: We can't easily test that no warning was logged without mocking the logger,
+		// but the behavior is correct - no warning should be logged for blank AccountDefinitionName
+	})
+
+	t.Run("json serialization of destination with accounts", func(t *testing.T) {
```
utsabc, line 291: Lets add what do we expect as outcome of this test case as well

backend-config/account_association.go
utsabc: I see this is fine

backend-config/account_association.go
```diff
@@ -21,6 +24,50 @@ func (c *ConfigT) processAccountAssociations() {
...
+// populateAccountToDestination populates an account to a destination field based on the account ID.
+// It creates an Account object by combining account details with its definition.
+//
+// Parameters:
+//   - accountID: The ID of the account to populate
+//   - accountAssociationLogger: The logger instance to use for logging
+//
+// Returns:
+//   - *Account: The populated account object or nil if the account doesn't exist
+func (c *ConfigT) populateAccountToDestination(accountID string, accountAssociationLogger logger.Logger) *Account {
```
utsabc, line 57:
```suggestion
func (c *ConfigT) getAccountById(accountID string, accountAssociationLogger logger.Logger) *Account {
```

utsabc, line 57: Better name is `enrichAccountWithDefinition`

backend-config/account_association.go
```diff
@@ -29,55 +76,31 @@ func (c *ConfigT) processAccountAssociations() {
 // For each account type, it:
 // 1. Checks if the corresponding account ID exists in the destination config
 // 2. Verifies the account exists in the accounts map
-// 3. Creates an AccountWithDefinition by combining account details with its definition
+// 3. Creates an account by combining account details with its definition
 // 4. Assigns it to the appropriate field in the destination
 //
 // Parameters:
 //   - dest: Pointer to the destination being configured
 func (c *ConfigT) setDestinationAccounts(dest *DestinationT) {
```
utsabc, line 84: `enrichDestinationWithAccounts` can be better name for this function

backend-config/account_association.go
```diff
@@ -29,55 +76,31 @@ func (c *ConfigT) processAccountAssociations() {
...
-	// Check and set the regular account if specified in the destination config
+	accountAssociationLogger := pkgLogger.Withn(
+		obskit.DestinationID(dest.ID),
+		obskit.DestinationType(dest.DestinationDefinition.Name),
+	)
+	// Check and set the delivery account if specified in the destination config
 	if accountID, ok := dest.Config["rudderAccountId"].(string); ok {
```
atzoum, line 90: is an empty check required as well?
```suggestion
	if accountID, ok := dest.Config["rudderAccountId"].(string); ok && accountID != "" {
```

backend-config/account_association.go
```diff
@@ -29,55 +76,31 @@ func (c *ConfigT) processAccountAssociations() {
...
-		}
+		accountAssociationLogger = accountAssociationLogger.Withn(
+			logger.NewStringField("rudderAccountId", accountID),
+			logger.NewStringField("flowType", "delivery"),
+		)
+		dest.DeliveryAccount = c.populateAccountToDestination(accountID, accountAssociationLogger)
 	}
 
 	// Check and set the delete account if specified in the destination config
 	if deleteAccountID, ok := dest.Config["rudderDeleteAccountId"].(string); ok {
```
atzoum, line 99: is an empty check required as well?
```suggestion
	if deleteAccountID, ok := dest.Config["rudderDeleteAccountId"].(string); ok && deleteAccountID != "" {
```

backend-config/account_association.go
```diff
@@ -21,6 +24,50 @@ func (c *ConfigT) processAccountAssociations() {
...
+// Parameters:
+//   - accountDefinitionName: The name of the account definition to look up
+//   - accountAssociationLogger: The logger instance to use for logging
+//
+// Returns:
+//   - *AccountDefinition: A pointer to the account definition or nil if not found
+func (c *ConfigT) getAccountDefinition(accountDefinitionName string, accountAssociationLogger logger.Logger) *AccountDefinition {
+	if accountDefinitionName == "" {
+		return nil
+	}
```
atzoum, line 39: It would be preferrable to avoid calling this method if accountDefinitionName is empty, since the caller has a better context with respect what an empty accountDefinitionName means. The logic behind the empty check here is not that straightforward
```suggestion
```

backend-config/account_association.go
atzoum: I have comment elsewhere as well, maybe we shouldn't be calling the method if account definition is empty

backend-config/account_association.go
ItsSudip: Moved this logic to the caller.

koladilip (approved):

atzoum (approved):

backend-config/account_association.go
ItsSudip: We have removed the logic of checking empty string of `accountDefinitionName` to the caller.

utsabc (approved):

PR #5874, opened 2025-05-19
chore: refactor config change dtype for config to json.RawMessage

mihir20:
# Description

## Linear Ticket

## Security


services/debugger/source/eventUploader.go
```diff
@@ -8,6 +8,10 @@ import (
 	"sync"
 	"time"
 
+	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
+
+	"github.com/grafana/jsonparser"
```
ktgowtham, line 13: May be as a follow up, we should think of moving it to go-kit given ingestion service and rudder-server uses.

ktgowtham (approved):

services/debugger/source/eventUploader.go
fracasula: @ktgowtham once we have [this lib](https://github.com/rudderlabs/rudder-go-kit/pull/809) we could do a quick refactoring although it might be quite big here in rudder-server.

If we end up proving that the grafana one isn't worth using then we'll have to stick to tidwall everywhere.

PR #5875, opened 2025-05-20
fix: warehouse transformations

achettyiitr:
# Description

- In rudder-transformer, we are using `Number.isInteger` to determine for int datatype, in comparison with Golang, which has a lower range `9223372036854775807`. `654645465456456400000` is being calculated as a float in Golang and an int in JS. Using a `big` package to determine now.
- `utf16` Rune count for being in conjunction with JavaScript.
- GeoLocation enrichment struct getting embedded directly to a singular event. These need to be handled separately, like we are doing for the Tracking Plan.
- Exclude the extract event in case the event name is empty.
- Exclude the identity event in case the userID is empty. 

## Linear Ticket

- Resolves WAR-672.

## Security

### DIFF
--- a/processor/internal/transformer/destination_transformer/embedded/warehouse/datatype.go
+++ b/processor/internal/transformer/destination_transformer/embedded/warehouse/datatype.go
@@ -1,11 +1,12 @@
 package warehouse
 
 import (
-	"unicode/utf8"
+	"fmt"
+	"math/big"
 
 	"github.com/samber/lo"
 
-	"github.com/rudderlabs/rudder-server/jsonrs"
+	"github.com/rudderlabs/rudder-server/internal/enricher"
 	"github.com/rudderlabs/rudder-server/processor/internal/transformer/destination_transformer/embedded/warehouse/internal/model"
 	"github.com/rudderlabs/rudder-server/processor/internal/transformer/destination_transformer/embedded/warehouse/internal/utils"
 	"github.com/rudderlabs/rudder-server/processor/types"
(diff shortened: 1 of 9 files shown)


processor/internal/transformer/destination_transformer/embedded/warehouse/datatype.go
```diff
@@ -142,3 +148,58 @@ func convertToSliceIfViolationErrors(val any) any {
 	}
 	return val
 }
+
+func addDataAndMetadataForContextGeoEnrichment(tec *transformEventContext, data map[string]any, metadata map[string]string, key string, val any) error {
+	if geoLocation, ok := val.(enricher.Geolocation); ok {
+		if len(geoLocation.IP) > 0 {
```
shekhar-rudder, line 154: Instead of having an if block for each field, we can have a for loop

processor/internal/transformer/destination_transformer/embedded/warehouse/events.go
```diff
@@ -236,9 +236,7 @@ func (t *Transformer) extractCommonProps(tec *transformEventContext) (map[string
 	if err = setDataAndMetadataFromRules(tec, commonData, commonMetadata, rules.ExtractRules); err != nil {
 		return nil, nil, fmt.Errorf("extract: setting data and column types from rules: %w", err)
 	}
-
-	eventName, _ = commonData[eventColName].(string)
-	if utils.IsEmptyString(eventName) {
+	if len(strings.TrimSpace(utils.ToString(commonData[eventColName]))) == 0 {
```
shekhar-rudder, line 239: Is the change in this file only about trimming? If yes, can include trimming inside `IsEmptyString` function instead of using `ToString`?

vyeshwanth (approved):

processor/internal/transformer/destination_transformer/embedded/warehouse/datatype.go
achettyiitr: This would require creating an array for key (`_ip`) and data (`geoLocation.IP`) as these are struct fields which is probably not required.

processor/internal/transformer/destination_transformer/embedded/warehouse/events.go
achettyiitr: If we move this to `IsEmptyString`, its implementation for JS would then be different as the empty-spaced string would result in `false` in case of JavaScript. For now keeping it like this.
```
//
//	const isBlank = (value) => {
//	 try {
//	   return _.isEmpty(_.toString(value));
//	 } catch (e) {
//	   console.error(`Error in isBlank: ${e.message}`);
//	   return false;
//	 }
//	};
```

shekhar-rudder (approved):

PR #5876, opened 2025-05-20
chore(jobsdb): migrate incomplete datasets in pairs excluding completed ones

atzoum:
# Description

Avoid migrating a small and incomplete dataset (needing a pair) if all other datasets to be migrated are complete (without pending jobs). 
E.g. if dataset1 is complete and dataset2 needs a pair, only dataset1 will be migrated/deleted. Dataset2 will only be migrated if there is another dataset3 that needs to be migrated and has some incomplete jobs too.

Migrating incomplete datasets takes time and while the migration is happening all other operations (reads, store, update) are being blocked, thus we avoid migrating a single incomplete dataset, even though the resulting dataset would have less jobs to query.

Continuation of #5873

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -1261,12 +1261,12 @@ func (jd *Handle) checkIfFullDSInTx(tx *Tx, ds dataSetT) (bool, error) {
 	}
 
 	if jd.conf.maxDSRetentionPeriod.Load() > 0 {
-		if time.Since(minJobCreatedAt.Time) > jd.conf.maxDSRetentionPeriod.Load() {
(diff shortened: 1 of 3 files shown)


jobsdb/jobsdb.go
```diff
@@ -1276,7 +1276,7 @@ func (jd *Handle) checkIfFullDSInTx(tx *Tx, ds dataSetT) (bool, error) {
 		return true, nil
 	}
 
-	if rowCount > jd.conf.MaxDSSize.Load() {
+	if rowCount >= jd.conf.MaxDSSize.Load() {
```
atzoum, line 1279: **Note:** adding an equality check for simplifying our tests

Sidddddarth (approved):

cisse21 (approved):

PR #5881, opened 2025-05-21
feat: integrate the transformer client with cslb for source transformations

mihir20:
# Description

- introduce httplb for source transformation
- replace retrablehttp lib with exp backoff as retrablehttp has some licensing issues

## Linear Ticket

mihir/pipe-2053

## Security

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.50.4](https://github.com/rudderlabs/rudder-server/compare/v1.50.3...v1.50.4) (2025-06-04)
+
+
+### Bug Fixes
+
+* use jsonrs stdlib for encoding warehouse transformations ([#5945](https://github.com/rudderlabs/rudder-server/issues/5945)) ([055b7de](https://github.com/rudderlabs/rudder-server/commit/055b7debc4c31e7915ed6a0a8fa3391f0dd8422c))
+
 ## [1.50.3](https://github.com/rudderlabs/rudder-server/compare/v1.50.2...v1.50.3) (2025-06-04)
 
 
--- a/processor/internal/transformer/destination_transformer/embedded/warehouse/events_test.go
+++ b/processor/internal/transformer/destination_transformer/embedded/warehouse/events_test.go
@@ -2034,7 +2034,7 @@ func TestEvents(t *testing.T) {
 			},
 			{
 				name:         "track (POSTGRES) jsonPaths (escape characters for &, <, and >)",
-				eventPayload: `{"type":"track","messageId":"messageId","anonymousId":"anonymousId","userId":"userId","sentAt":"2021-09-01T00:00:00.000Z","timestamp":"2021-09-01T00:00:00.000Z","receivedAt":"2021-09-01T00:00:00.000Z","originalTimestamp":"2021-09-01T00:00:00.000Z","channel":"web","event":"event","request_ip":"5.6.7.8","properties":{"location": {"radiation_level": 0.000000006, "city":"Palo Alto <p>Ampersand: &;</p> Palo Alto","state":"California","country":"USA","coordinates":{"latitude":37.4419,"longitude":-122.143,"geo":{"altitude":30.5,"accuracy":5,"details":{"altitudeUnits":"meters","accuracyUnits":"meters"}}}},"review_id":"86ac1cd43","product_id":"9578257311"},"userProperties":{"rating":3.0,"review_body":"OK for the price. It works but the material feels flimsy."},"context":{"traits":{"name":"Richard Hendricks","email":"rhedricks@example.com","logins":2},"ip":"1.2.3.4"}}`,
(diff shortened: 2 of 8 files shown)


internal/transformer-client/client.go
```diff
@@ -96,6 +109,67 @@ func NewClient(config *ClientConfig) Client {
 	}
 }
 
+func NewRetryableClient(config *RetryableClientConfig) RetryableClient {
```
ktgowtham, line 112: lot of duplication in this file. 
why can't we use existing logic in NewClient for http client creation & reuse it here? Have a common method to create http client based on ClientConfig and use it in both `NewClient` and `NewRetryableClient`

atzoum, line 112: Not sure why we need this here. A `RetryableClient` just needs a http.Client & a bunch of configuration options as an input. Can't we create a retryable client in the webhook's setup instead, based on the http.Client that this package will produce?

gateway/webhook/webhookTransformer.go
```diff
@@ -303,3 +312,43 @@ func (bt *batchWebhookTransformerT) transform(events [][]byte, sourceTransformer
...
+
+	_ = backoff.RetryNotify(
+		func() error {
+			var req *http.Request
+			req, err = http.NewRequest("POST", transformerURL, body)
+			if err == nil {
+				req.Header.Set("Content-Type", contentTypeJsonUTF8)
+				resp, err = bt.webhook.httpClient.Do(req) // nolint: bodyclose
+				// retry 5xx errors
+				if err == nil && (resp.StatusCode > http.StatusInternalServerError) {
```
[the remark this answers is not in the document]
mihir20: missed it, thanks for pointing out

gateway/webhook/webhookTransformer.go
```diff
@@ -303,3 +312,43 @@ func (bt *batchWebhookTransformerT) transform(events [][]byte, sourceTransformer
...
+					return fmt.Errorf("non-success status code: %d", resp.StatusCode)
+				}
+			}
+			return err
+		},
+		backoff.WithMaxRetries(
```
atzoum, line 336: We can prepare our own version of retryable http client in `rudder-go-kit` taking advantage of the backoff library and use this instead of  `hashicorp/go-retryablehttp` in many other places.

mihir20: good idea 👍

mihir20: https://github.com/rudderlabs/rudder-go-kit/pull/818

gateway/webhook/webhookTransformer.go
```diff
@@ -303,3 +310,19 @@ func (bt *batchWebhookTransformerT) transform(events [][]byte, sourceTransformer
 	}
 	return batchResponse
 }
+
+func (bt *batchWebhookTransformerT) postWithRetry(transformerURL string, body io.Reader) (*http.Response, error) {
```
atzoum: Is this method now needed? Retries are not being performed by it, but by the client itself

mihir20: we can rename the function to `doPost`, but want to keep this function as I want to keep the logic for creation of request, setting header and doing a request separate instead keeping this in the original function.

gateway/webhook/setup.go
```diff
@@ -87,13 +92,37 @@ func Setup(gwHandle Gateway, transformerFeaturesService TransformerFeaturesServi
...
 
+	transformerClientConfig := &transformerclient.ClientConfig{
+		ClientTimeout: conf.GetDurationVar(30, time.Second, "HttpClient.sourceTransformer.timeout", "HttpClient.webhook.timeout"),
+		ClientTTL:     conf.GetDurationVar(10, time.Second, "Transformer.Client.sourceTransformer.ttl", "Transformer.Client.ttl"),
+		ClientType:    conf.GetStringVar("httplb", "Transformer.Client.sourceTransformer.type", "Transformer.Client.type"),
+		PickerType:    conf.GetStringVar("power_of_two", "Transformer.Client.sourceTransformer.httplb.pickerType", "Transformer.Client.httplb.pickerType"),
+	}
+	transformerClientConfig.TransportConfig.DisableKeepAlives = conf.GetBoolVar(true, "Transformer.Client.sourceTransformer.disableKeepAlives", "Transformer.Client.disableKeepAlives")
+	transformerClientConfig.TransportConfig.MaxConnsPerHost = conf.GetIntVar(100, 1, "Transformer.Client.sourceTransformer.maxHTTPConnections", "Transformer.Client.maxHTTPConnections")
+	transformerClientConfig.TransportConfig.MaxIdleConnsPerHost = conf.GetIntVar(1, 1, "Transformer.Client.sourceTransformer.maxHTTPIdleConnections", "Transformer.Client.maxHTTPIdleConnections")
```
ktgowtham, line 104: Isn't this too conservative? should we increase the default further?

mihir20: for httplb this limit is per resolved ip, there can be many transformer pods and even with 1 idle connections, each transformer pod will have enough connections.

PR #5882, opened 2025-05-21
fix: server panicking with invalid byte sequence for encoding utf8 0x00

atzoum:
# Description

- Removing null bytes while sanitizing. Previously we were only removing a unicode encoded null byte.
- Removing null bytes from UserID while sanitizing

## Security

### DIFF
--- a/jobsdb/integration_test.go
+++ b/jobsdb/integration_test.go
@@ -1056,6 +1056,7 @@ func TestJobsdbSanitizeJSON(t *testing.T) {
 		{
 			string(JSONB),
 			[]testCase{
+				{"\x00", "", nil},
 				{`\u0000`, "", nil},
 				{`\u0000☺\u0000b☺`, "☺b☺", nil},
 				// NOTE: we are not handling the following:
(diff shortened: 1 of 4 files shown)


cisse21 (approved):

achettyiitr (approved):

jobsdb/integration_test.go
```diff
@@ -1093,8 +1094,9 @@ func TestJobsdbSanitizeJSON(t *testing.T) {
 		{
 			string(TEXT),
 			[]testCase{
-				{`\u0000`, `\u0000`, nil},
-				{`\u0000☺\u0000b☺`, `\u0000☺\u0000b☺`, nil},
+				{"\x00", "", nil},
+				{`\u0000`, "", nil},
```
Sidddddarth, line 1098: this change now happens because of the invalid character in user_id column.
It might be prudent to have some of those previous cases too

Sidddddarth (approved):

jobsdb/integration_test.go
atzoum: If you notice, now TEXT column type exhibits the same behaviour as JSONB column

PR #5888, opened 2025-05-22
fix: block internal IP access as destination

cisse21:
# Description

- Apply private/internal ip restrictions is during dns resolution on the http client
- Add dryrun mode to monitor the things which will be blocked
- Add another mode to block ips

## Linear Ticket

Fixes PIPE-2064
Fixes PIPE-2111

## Security


router/network.go
```diff
@@ -28,19 +30,106 @@ import (
...
+			// Log error but continue with other ranges
+			fmt.Printf("Error parsing CIDR block %s: %v\n", cidr, err)
+			continue
+		}
+		ranges = append(ranges, ipRange)
+	}
+
+	return ranges
+}
+
```
atzoum, line 87: I believe that we just need a simlpe `rudder-go-kit` function, named `net.IsPrivateIP`
```go
var privateIPBlocks []*net.IPNet

func init() {
	for _, cidr := range []string{
		"10.0.0.0/8",
		"172.16.0.0/12",
		"192.168.0.0/16",
		"169.254.0.0/16",  // Link-local
		"127.0.0.0/8",     // Loopback
		"::1/128",         // IPv6 loopback
		"fc00::/7",        // IPv6 unique local address
		"fe80::/10",       // IPv6 link-local
	} {
		_, block, _ := net.ParseCIDR(cidr)
		privateIPBlocks = append(privateIPBlocks, block)
	}
}

func IsPrivateIP(ip net.IP) bool {
	for _, block := range privateIPBlocks {
		if block.Contains(ip) {
			return true
		}
	}
	return false
}
```

router/network.go
```diff
@@ -59,6 +148,41 @@ func handleQueryParam(param interface{}) string {
 	}
 }
 
+// validateURLAndHandlePrivateIP checks if the URL resolves to a private IP and handles it based on mode
+func (network *netHandle) validateURLAndHandlePrivateIP(urlStr string) (bool, error) {
```
lvrach, line 152: What's the benefit of doing the additional check here, since we have it on dialContext?

router/network.go
cisse21: This makes more sense.. I was wondering where do I place this block and couldn't find anything appropriate in server, Will move this to go-kit but again the CIDR ranges need to be configurable because
1. All our integration tests run against a webhook and they might end up failing
2. Any of our load tests we run using an internal webhook would fail

thoughts on this?

lvrach: I agree CIDR needs to be configurable for various reasons. We might want to allowlist  some private/internal IPs or even blocklist a public one in production.

router/network.go
cisse21: One reason I can think of is failing fast but yeah it makes sense to remove some duplication. Also the claude solution in the ticket provided both in dial context and also having an additional check

router/network.go
```diff
@@ -287,11 +348,17 @@ func (network *netHandle) Setup(netClientTimeout time.Duration) {
 		defaultTransportCopy.TLSClientConfig = &tlsClientConfig
 		network.logger.Info(network.destType, defaultTransportCopy.TLSClientConfig.NextProtos)
 	}
-	// by default we should have as many idle connections as the number of workers
+
+	network.blockPrivateIPsDryRun = getRouterConfigBool("dryRunMode", network.destType, false)
+	network.blockPrivateIPs = getRouterConfigBool("blockPrivateIPs", network.destType, false)
+	network.logger.Info("blockPrivateIPsDryRun: ", network.blockPrivateIPsDryRun)
+	network.logger.Info("blockPrivateIPs: ", network.blockPrivateIPs)
```
atzoum: let's move these at the top of the method

router/network.go
```diff
@@ -261,19 +281,60 @@ func (network *netHandle) SendPost(ctx context.Context, structData integrations.
...
 	var defaultTransportCopy http.Transport
-	// Not safe to copy DefaultTransport
-	// https://groups.google.com/forum/#!topic/golang-nuts/JmpHoAd76aU
-	// Solved in go1.8 https://github.com/golang/go/issues/26013
 	misc.Copy(&defaultTransportCopy, defaultTransportPointer)
+
+	dialer := &net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}
```
atzoum:
```suggestion
	originalDialContext := defaultTransportCopy.DialContext
```

router/network.go
```diff
@@ -261,19 +281,60 @@ func (network *netHandle) SendPost(ctx context.Context, structData integrations.
...
+						}
+						// In block mode, reject the connection
+						if network.blockPrivateIPs {
+							return nil, ErrDenyPrivateIP
+						}
+					}
+				}
+			}
+		}
+		return dialer.DialContext(ctx, networkType, address)
```
atzoum:
```suggestion
		return originalDialContext(ctx, networkType, address)
```

router/network.go
```diff
@@ -261,19 +281,60 @@ func (network *netHandle) SendPost(ctx context.Context, structData integrations.
...
+				}
+				ips, err := net.LookupIP(host)
+				if err != nil {
+					return nil, err
+				}
+				for _, ip := range ips {
+					if network.blockPrivateIPsCIDRs.Contains(ip) {
+						// In dry run mode, just log and allow the connection
+						if network.blockPrivateIPsDryRun {
+							network.logger.Warnf("Connection to private IP %s detected in dry run mode", ip)
```
atzoum:
```suggestion
							network.logger.Warnn("Connection to private ip detected in dry run mode", logger.NewStringField("ip"), ip)
```

router/network.go
```diff
@@ -261,19 +281,60 @@ func (network *netHandle) SendPost(ctx context.Context, structData integrations.
...
+							return dialer.DialContext(ctx, networkType, address)
```
atzoum:
```suggestion
							return originalDialContext(ctx, networkType, address)
```

atzoum (approved):

PR #5889, opened 2025-05-23
fix: under reporting of load files when using upload id

shekhar-rudder:
# Description

**BUG** - Using partition by on (upload_id, table_name) was picking only one file per table even though there exists multiple files per table

**FIX** - Now we are picking all the files for the given upload_id. Any load files generated in a previous attempt gets deleted so we need not worry about that

## Security

### DIFF
--- a/warehouse/internal/repo/load_test.go
+++ b/warehouse/internal/repo/load_test.go
@@ -159,7 +159,7 @@ func Test_LoadFiles_WithUploadID(t *testing.T) {
 
 		for i := 0; i < 10; i++ {
 			loadFile := model.LoadFile{
-				TableName:             "table_name__" + strconv.Itoa(i),
+				TableName:             "table_name__" + strconv.Itoa(i%3),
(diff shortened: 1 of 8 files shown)


warehouse/internal/repo/load_test.go
```diff
@@ -208,37 +208,6 @@ func Test_LoadFiles_WithUploadID(t *testing.T) {
 		require.Len(t, loadFiles, 0)
 		require.NoError(t, err)
 	})
-
-	t.Run("get latest for stagingID", func(t *testing.T) {
```
shekhar-rudder, line 212: not a valid test scenario anymore

RanjeetMishra (approved):

RanjeetMishra:
@shekhar-rudder Should we raise this PR against release branch for hotfix? Or we can cherry pick this and https://github.com/rudderlabs/rudder-server/pull/5886 both in hotfix release.
CC: @achettyiitr

PR #5906, opened 2025-05-27
fix: under reporting of load files when using upload id (#5889)

shekhar-rudder:
# Description

**BUG** - Using partition by on (upload_id, table_name) was picking only one file per table even though there exists multiple files per table

**FIX** - Now we are picking all the files for the given upload_id. Any load files generated in a previous attempt gets deleted so we need not worry about that

## Security


vyeshwanth:
> **BUG** - Using partition by on (upload_id, table_name) was picking only one file per table even though there exists multiple files per table

Why did we do this? Is this for load file optimisation PR?

PR #5915, opened 2025-05-28
fix: bq load users table failing with unrecognised _PARTITIONTIME

vyeshwanth:
# Description
- Use partition in the warehouse when building deduplication query
- skip writing to partition if bq table is not partitioned or partitioned with other than ingestion time

## Linear Ticket
PART OF WAR-685

## Security

### DIFF
--- a/warehouse/integrations/bigquery/bigquery.go
+++ b/warehouse/integrations/bigquery/bigquery.go
@@ -226,6 +226,7 @@ func (bq *BigQuery) createTableView(ctx context.Context, tableName string, colum
 	}
 
 	// no need to pass partitioningInWarehouse here. dedup query considers the partition column, type from the destination config
+	// should always create a view with partition column and type from destination config
 	deduplicationQuery, err := bq.deduplicationQuery(tableName, columnMap, nil)
 	if err != nil {
 		return fmt.Errorf("deduplication query: %w", err)
--- a/warehouse/integrations/bigquery/bigquery.go
+++ b/warehouse/integrations/bigquery/bigquery.go
@@ -274,7 +274,7 @@ func (bq *BigQuery) deduplicationQuery(tableName string, columnMap model.TableSc
 	// if partitioningInWarehouse is not nil, then we use the partition column and type from the warehouse
 	if partitioningInWarehouse != nil {
 		if partitionColumn != partitioningInWarehouse.Field {
-			bq.logger.Warnn("partition column mismatch in config and warehouse",
+			bq.logger.Debugn("partition column mismatch in config and warehouse",
 				logger.NewStringField("partitionColumnInConfig", partitionColumn),
 				logger.NewStringField("partitionColumnInWarehouse", partitioningInWarehouse.Field),
 			)
@@ -1256,16 +1256,17 @@ func (bq *BigQuery) TestLoadTable(ctx context.Context, location, tableName strin
 	gcsRef.MaxBadRecords = 0
 	gcsRef.IgnoreUnknownValues = false
 
-	partitioningInWarehouse, err := bq.fetchTablePartitionFromWarehouse(ctx, tableName)
-	if err != nil {
-		return fmt.Errorf("fetching table partitioning: %w", err)
+	partitionType, _ := bq.bigqueryPartitionType(bq.partitionType())
+	partitioningInWarehouse := &bigquery.TimePartitioning{
+		Field: bq.partitionColumn(),
+		Type:  partitionType,
 	}
 
 	var outputTable string
 	if bq.avoidPartitionDecorator(partitioningInWarehouse) {
 		outputTable = tableName
 	} else {
-		partitionDate, err := bq.partitionDateByPartitioning(partitioningInWarehouse)
+		partitionDate, err := bq.partitionDate()
 		if err != nil {
 			return fmt.Errorf("partition date: %w", err)
 		}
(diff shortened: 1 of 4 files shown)


warehouse/integrations/bigquery/bigquery.go
```diff
@@ -225,7 +225,7 @@ func (bq *BigQuery) createTableView(ctx context.Context, tableName string, colum
 		return nil
 	}
 
-	deduplicationQuery, err := bq.deduplicationQuery(tableName, columnMap)
+	deduplicationQuery, err := bq.deduplicationQuery(tableName, columnMap, nil)
```
shekhar-rudder, line 228: Why are we passing `nil` here? Lets add a comment

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -688,7 +716,9 @@ func (bq *BigQuery) LoadUserTables(ctx context.Context) (errorMap map[string]err
 
 	log.Infon("Loading data")
 	var partitionedUsersTable string
-	if bq.avoidPartitionDecorator() {
+
+	// avoid writing to partioned table if there is a no or custom partition instead of ingestion time partition i.e _PARTITIONTIME
+	if bq.avoidPartitionDecorator() || usersTablePartitionInWarehouse == nil || usersTablePartitionInWarehouse.Field != "" {
```
shekhar-rudder, line 721: Don't we need this check at other places as well which are using `bq.avoidPartitionDecorator()` ?

shekhar-rudder, line 721: Aren't the conditions `usersTablePartitionInWarehouse == nil` and `usersTablePartitionInWarehouse.Field != ""` effectively opposites? We're trying to check for the absence of a partition field, right?

vyeshwanth: No

`usersTablePartitionInWarehouse == nil` indicates if table has no partition.  If table has no partition we should avoid loading into partitioned tables.

`usersTablePartitionInWarehouse.Field != ""`  indicates there is a custom partition column set other than ingestion time. we should avoid loading into partition in this case

vyeshwanth: I think we should. I was looking at users table but ideally partition could be changed for any table for that matter.

shekhar-rudder: If the partition column is ingestion time, is `usersTablePartitionInWarehouse == nil` true and `usersTablePartitionInWarehouse.Field != ""` false ?

warehouse/integrations/bigquery/bigquery.go
vyeshwanth: create view, we should use whatever is in the destination config.

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -264,6 +271,18 @@ func (bq *BigQuery) deduplicationQuery(tableName string, columnMap model.TableSc
 		partitionColumn, partitionType = bq.partitionColumn(), bq.partitionType()
 	)
 
+	// if partitioningInWarehouse is not nil, then we use the partition column and type from the warehouse
+	if partitioningInWarehouse != nil {
+		if partitionColumn != partitioningInWarehouse.Field {
+			bq.logger.Warnn("partition column mismatch in config and warehouse",
```
achettyiitr, line 277: Do you think we can make this a debug? It will become noise because it will come every time for the customer if they have modified something.

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -264,6 +271,18 @@ func (bq *BigQuery) deduplicationQuery(tableName string, columnMap model.TableSc
...
+				logger.NewStringField("partitionColumnInConfig", partitionColumn),
+				logger.NewStringField("partitionColumnInWarehouse", partitioningInWarehouse.Field),
+			)
+		}
+		partitionColumn = partitioningInWarehouse.Field
```
achettyiitr, line 282: Does Field come as UPPER case?

warehouse/integrations/bigquery/bigquery.go
```diff
@@ -1216,15 +1256,19 @@ func (bq *BigQuery) TestLoadTable(ctx context.Context, location, tableName strin
 	gcsRef.MaxBadRecords = 0
 	gcsRef.IgnoreUnknownValues = false
 
-	partitionDate, err := bq.partitionDate()
+	partitioningInWarehouse, err := bq.fetchTablePartitionFromWarehouse(ctx, tableName)
```
achettyiitr, line 1259: Don't think we need it for Validations. If the tables doesn't exist `fetchTablePartitionFromWarehouse` would error our right? Also, we delete the table post validations so would not be required.

warehouse/integrations/bigquery/bigquery.go
vyeshwanth: Field is nothing but a column name, column names are not case-sensitive in Bigquery.

warehouse/integrations/bigquery/bigquery.go
vyeshwanth: We can make it debug. We are logging sql queries, so we will get to know what partition column is being used.

warehouse/integrations/bigquery/bigquery.go
vyeshwanth: Idea is to keep the validations similar to the final load table. Why would the table doesn't exist? validation creates the table and then loads data?

achettyiitr: Yeah, and then deletes the table as well.

achettyiitr: Post validation.

PR #5917, opened 2025-05-28
chore: proc sample store uploader

achettyiitr:
# Description

- Added store sampler for processor to upload sample events in case storing to Postgres fails.

## Linear Ticket

- Resolves WAR-697

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"path"
 	"runtime/trace"
 	"slices"
 	"strconv"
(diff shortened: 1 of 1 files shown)


ktgowtham (approved):

processor/processor.go
```diff
@@ -433,6 +437,11 @@ func (proc *Handle) Setup(
 		proc.logger.Errorn("failed to create ut sampling file manager", obskit.Error(err))
 		proc.utSamplingFileManager = nil
 	}
+	proc.storeSamplingFileManager, err = getStoreSamplingUploader(proc.conf, proc.logger)
+	if err != nil {
+		proc.logger.Errorn("failed to create store sampling file manager", obskit.Error(err))
+		proc.storeSamplingFileManager = nil
```
cisse21: Why do we need to explicitly mention it as nil? Doesnt the function return nil when an error occurs?

achettyiitr: Removed it. Copied code from the UT sampling file manager.

cisse21 (approved):

PR #5919, opened 2025-05-28
fix(processor): throughput stalling due to mutex deadlock on full connection pool

atzoum:
# Description

When `Processor.enableConcurrentStore` is enabled, the processor performs the following operations concurrently:

- Storing batchrouter jobs
- Storing router jobs
- Storing proc error jobs
- Updating gateway job statuses (transaction gets committed only after all above transactions are successfully committed)

Each of these operations uses a separate database connection.

In the "store router jobs" operation, before opening a connection, a mutex is acquired per destinationID to ensure that only one goroutine inserts jobs for a given destination at a time.

However, this can lead to a deadlock under high load: if all database connections in the pool are occupied by goroutines waiting to acquire a destinationID mutex, and the one goroutine that has acquired the mutex is blocked waiting for a database connection, progress stalls. The result is a circular wait where no goroutine can proceed, causing the processor's throughput to stall..

To fix this deadlock, whenever `Processor.enableConcurrentStore` is enabled, we're acquiring the mutexes early, before any other connection.

Additionally, we're increasing the default connection pool size to 80 and limiting it to 20 for gateway.

## Linear Ticket

resolves PIPE-2101

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -2529,6 +2529,39 @@ func (proc *Handle) sendQueryRetryStats(attempt int) {
 }
 
 func (proc *Handle) storeStage(partition string, in *storeMessage) {
+	lockRouterDestIDs := func() func() {
+		var deferred []func()
+		if len(in.destJobs) == 0 {
+			return func() {}
+		}
+		if len(in.routerDestIDs) > 0 {
+			destIDs := lo.Uniq(in.routerDestIDs)
+			slices.Sort(destIDs)
+			for _, destID := range destIDs {
+				proc.storePlocker.Lock(destID)
+				deferred = append(deferred, func() { proc.storePlocker.Unlock(destID) })
+			}
+		} else {
+			proc.logger.Warnn("empty storeMessage.routerDestIDs",
+				logger.NewStringField("partition", partition),
+				logger.NewStringField("expected",
+					strings.Join(
+						lo.Uniq(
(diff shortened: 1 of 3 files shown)


cisse21 (approved):

processor/processor.go
```diff
@@ -2529,6 +2529,34 @@ func (proc *Handle) sendQueryRetryStats(attempt int) {
 }
 
 func (proc *Handle) storeStage(partition string, in *storeMessage) {
+	lockRouterDestIDs := func() func() {
+		var deferred []func()
+		if len(in.destJobs) > 0 {
```
fracasula, line 2534: [optional] We can avoid a bit of nesting since `deferred` would be empty with no jobs.

```go
if len(in.destJobs) == 0 {
    return func(){}
}
```

processor/processor.go
```diff
@@ -2529,6 +2529,34 @@ func (proc *Handle) sendQueryRetryStats(attempt int) {
...
+			if len(in.routerDestIDs) > 0 {
+				destIDs := lo.Uniq(in.routerDestIDs)
+				slices.Sort(destIDs)
+				for _, destID := range destIDs {
+					proc.storePlocker.Lock(destID)
+					deferred = append(deferred, func() { proc.storePlocker.Unlock(destID) })
+				}
+			} else {
+				proc.logger.Warnw("empty storeMessage.routerDestIDs",
```
fracasula, line 2543: Can you please use `Warnn` for better performance?

fracasula (approved):

PR #5922, opened 2025-05-28
feat: add support for Snowflake Iceberg tables

shekhar-rudder:
# Description

This PR enhances the Snowflake implementation of the Manager interface by adding support for Iceberg tables.

Implementation Details:
- Introduced two new destination settings for Iceberg: `enableIceberg` and `externalVolume`.
- Created a `tableManager` interface to encapsulate table-specific behavior. This interface has two implementations: standard (existing) and Iceberg.
- Moved `dataTypesMap` and the corresponding Iceberg type handling behind this interface.
- `tableManager` is being initialized in the Setup method.
- Refactored `getSortedColumnsFromTableSchema` into a standalone function, making it reusable for `tableManager` instead of being a Snowflake-specific method.
- Removed unused `ColumnsWithDataTypes` method

## Security

### DIFF
--- a/warehouse/integrations/snowflake/table_manager.go
+++ b/warehouse/integrations/snowflake/table_manager.go
@@ -59,8 +59,7 @@ func (m *standardTableManager) addColumnsQuery(schemaIdentifier, tableName strin
 	))
 
 	for _, columnInfo := range columnsInfo {
-		dataType := m.dataTypesMap[columnInfo.Type]
-		queryBuilder.WriteString(fmt.Sprintf(` %q %s,`, columnInfo.Name, dataType))
+		queryBuilder.WriteString(fmt.Sprintf(` IF NOT EXISTS %q %s,`, columnInfo.Name, m.dataTypesMap[columnInfo.Type]))
 	}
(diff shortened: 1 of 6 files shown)


achettyiitr (approved):

warehouse/integrations/snowflake/table_manager.go
```diff
@@ -0,0 +1,127 @@
...
+		dataType, ok := m.dataTypesMap[columnInfo.Type]
+		if !ok {
+			return "", fmt.Errorf("invalid data type: %s", columnInfo.Type)
+		}
+		queryBuilder.WriteString(fmt.Sprintf(` IF NOT EXISTS %q %s,`, columnInfo.Name, dataType))
+	}
+	return strings.TrimSuffix(queryBuilder.String(), ",") + ";", nil
+}
+
+func columnsWithDataTypes(columns model.TableSchema, prefix string, dataTypesMap map[string]string) string {
```
vyeshwanth: what is this prefix for? In both the cases it is empty

shekhar-rudder: It is not required - removed it. The existing code was passing `prefix` as empty string.

PR #5935, opened 2025-06-03
feat: detect and handle memory-fenced responses in rudder-server

cisse21:
# Description

All transformer clients that rudder-server & rudder-ingestion-svc use should have perpetual retries with backoff whenever they encounter a HTTP 503 error with the X-Rudder-Should-Retry: true response header. Client should emit a request failure metric using the X-Rudder-Error-Reason header in the metric's reason label, so that we can visualize errors and setup alerts against them.

## Linear Ticket

Fixes PIPE-2077

## Security


internal/transformer-client/client.go
```diff
@@ -89,11 +94,37 @@ func NewClient(config *ClientConfig) Client {
...
+		cfg,
+		retryablehttp.WithHttpClient(baseClient),
+		retryablehttp.WithCustomRetryStrategy(func(resp *http.Response, err error) (bool, error) {
+			if err != nil {
+				return true, err
+			}
+			if resp.StatusCode == http.StatusServiceUnavailable &&
+				strings.ToLower(resp.Header.Get("X-Rudder-Should-Retry")) == "true" {
+				reason := resp.Header.Get("X-Rudder-Error-Reason")
+				stats.Default.NewTaggedStat("transformer_request_failure", stats.CountType, stats.Tags{"reason": reason}).Count(1)
```
atzoum:
```suggestion
				stats.Default.NewTaggedStat("transformer_client_perpetual_retry_count", stats.CountType, stats.Tags{"reason": reason}).Count(1)
```

internal/transformer-client/client.go
```diff
@@ -89,11 +94,37 @@ func NewClient(config *ClientConfig) Client {
...
+		MaxInterval:     30 * time.Second,
+		MaxElapsedTime:  0, // 0 for infinite
+		Multiplier:      2.0,
 	}
+	return retryablehttp.NewRetryableHTTPClient(
+		cfg,
+		retryablehttp.WithHttpClient(baseClient),
+		retryablehttp.WithCustomRetryStrategy(func(resp *http.Response, err error) (bool, error) {
+			if err != nil {
+				return true, err
```
atzoum: We don't want to retry random errors perpetually, only `HTTP 503` responses with `X-Rudder-Should-Retry": true`.
```suggestion
				return false, backoff.Permanent(err)
```

internal/transformer-client/client.go
```diff
@@ -89,11 +94,37 @@ func NewClient(config *ClientConfig) Client {
...
+		return newRetryableHTTPClient(httplb.NewClient(options...))
 	default:
-		return client
+		return newRetryableHTTPClient(client)
+	}
+}
+
+func newRetryableHTTPClient(baseClient Client) Client {
+	cfg := &retryablehttp.Config{
+		MaxRetry:        0, // 0 for infinite
```
atzoum: Are we confident about that? Maybe we need `-1` instead?
https://github.com/cenkalti/backoff/blob/v4/tries.go#L23-L25

atzoum: let's also have a test verifying the behaviour we want:
- start an httptest server that we control
- start a client with low retry intervals
- verify that the client retries for 5 sec
- switch httptest to return a non retriable response
- verify that the client returns
- as another test, verify that errors are not being retried, e.g. simply perform a request to a port that is not listening.

internal/transformer-client/client.go
```diff
@@ -89,11 +94,37 @@ func NewClient(config *ClientConfig) Client {
...
+		InitialInterval: 1 * time.Second,
+		MaxInterval:     30 * time.Second,
```
atzoum: these needs to be configurable

internal/transformer-client/client.go
```diff
@@ -89,11 +96,37 @@ func NewClient(config *ClientConfig) Client {
...
+		return newRetryableHTTPClient(conf, httplb.NewClient(options...))
 	default:
-		return client
+		return newRetryableHTTPClient(conf, client)
+	}
+}
+
+func newRetryableHTTPClient(conf *config.Config, baseClient Client) Client {
+	cfg := &retryablehttp.Config{
+		MaxRetry:        conf.GetInt("Transformer.Client.maxRetry", 0),
```
atzoum: Are we confident about that? Maybe we need -1 instead?
https://github.com/cenkalti/backoff/blob/v4/tries.go#L23-L25

internal/transformer-client/client.go
```diff
@@ -89,11 +96,37 @@ func NewClient(config *ClientConfig) Client {
...
+		InitialInterval: conf.GetDuration("Transformer.Client.initialInterval", 1, time.Second),
```
atzoum: Let's use the `Transformer.Client.Retryable` prefix for all config options
```suggestion
		InitialInterval: conf.GetDuration("Transformer.Client.Retryable.initialInterval", 1, time.Second),
```

internal/transformer-client/client.go
```diff
@@ -89,11 +96,37 @@ func NewClient(config *ClientConfig) Client {
...
+		retryablehttp.WithCustomRetryStrategy(func(resp *http.Response, err error) (bool, error) {
+			if err != nil {
+				return false, backoff.Permanent(err)
+			}
+			if resp.StatusCode == http.StatusServiceUnavailable &&
+				strings.ToLower(resp.Header.Get("X-Rudder-Should-Retry")) == "true" {
+				reason := resp.Header.Get("X-Rudder-Error-Reason")
+				stats.Default.NewTaggedStat("transformer_client_perpetual_retry_count", stats.CountType, stats.Tags{"reason": reason}).Count(1)
+				resp.Body.Close()
+				return true, fmt.Errorf("memory-fenced: %s", reason)
```
atzoum:
```suggestion
				return true, fmt.Errorf("Got retryable error response from transformer: %s", reason)
```

internal/transformer-client/client_test.go
```diff
@@ -0,0 +1,393 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/rudderlabs/rudder-go-kit/config"
+	kithelper "github.com/rudderlabs/rudder-go-kit/testhelper"
+)
+
+func setupTestConfig() {
+	config.Set("Transformer.Client.maxRetry", 3)
```
atzoum: we shouldn't need to override this
```suggestion
```

internal/transformer-client/client_test.go
```diff
@@ -0,0 +1,393 @@
...
+		require.Equal(t, retryableResponses+1, requestCount, "Should make exactly %d requests", retryableResponses+1)
+
+		require.True(t, elapsed > 100*time.Millisecond, "Should take some time due to retries")
+		require.True(t, elapsed < 5*time.Second, "Should complete before max elapsed time")
+
+		resp.Body.Close()
+	})
+
+	t.Run("stops retrying after max elapsed time", func(t *testing.T) {
+		config.Set("Transformer.Client.maxRetry", 10) // High retry count
```
atzoum:
```suggestion
```

gateway/webhook/setup.go
```diff
@@ -106,7 +106,7 @@ func Setup(gwHandle Gateway, transformerFeaturesService TransformerFeaturesServi
 	transformerClientConfig.Recycle = conf.GetBoolVar(false, "Transformer.Client.sourceTransformer.recycle", "Transformer.Client.recycle")
 	transformerClientConfig.RecycleTTL = conf.GetDurationVar(60, time.Second, "Transformer.Client.sourceTransformer.recycleTTL", "Transformer.Client.recycleTTL")
 
-	transformerClient := transformerclient.NewClient(transformerClientConfig)
+	transformerClient := transformerclient.NewClient(conf, transformerClientConfig)
```
ktgowtham: I think we should fail fast webhook transformation requests given they are executed during a http request and we need to respond to the client asap.
Should we make perpetual retry configurable? 
wdyt @atzoum @mihir20 ?

ktgowtham: with new code, `transformerclient.NewClient` is returning a `retryableHTTPClient`. and after that, we're creating another instance of `retryableHTTPClient`, by using the client returned, with different config. Is this approach okay? can't we just use the retryable client that's first created, using the relevant config?

internal/transformer-client/client.go
```diff
@@ -33,7 +40,7 @@ type Client interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
-func NewClient(config *ClientConfig) Client {
+func NewClient(conf *config.Config, config *ClientConfig) Client {
```
atzoum: we already have a ClientConfig struct, we shouldn't be needing to inject the Config

```go
type ClientConfig struct {
	TransportConfig struct {
		DisableKeepAlives   bool          //	true
		MaxConnsPerHost     int           //	100
		MaxIdleConnsPerHost int           //	10
		IdleConnTimeout     time.Duration //	30*time.Second
	}

	ClientTimeout time.Duration //	600*time.Second
	ClientTTL     time.Duration //	10*time.Second
	ClientType    string        // stdlib(default), httplb
	PickerType    string        // power_of_two(default), round_robin, least_loaded_random, least_loaded_round_robin, random
	Recycle       bool          // false
	RecycleTTL    time.Duration // 60s

	// Configuration for retryable HTTP client in case of [X-Rudder-Should-Retry: true] HTTP 503 responses
	RetryRudderErrors struct {
		Enabled         bool          // true
		MaxRetry        int           // -1 - no limit
		InitialInterval time.Duration // 1s
		MaxInterval     time.Duration // 30s
		MaxElapsedTime  time.Duration // 0s - no limit
		Multiplier      float64       // 2.0
	}
}
```

gateway/webhook/setup.go
atzoum: Yes, we might be better off disabling perpetual retries in case of webhooks. The retryable http client that webhooks are already using should be enough

ktgowtham (approved):

internal/transformer-client/client.go
```diff
@@ -27,73 +49,211 @@ type ClientConfig struct {
...
+
+	// Check if this looks like an explicit configuration (has other values set)
+	hasExplicitConfig := clientConfig.RetryRudderErrors.InitialInterval != 0 ||
+		clientConfig.RetryRudderErrors.MaxInterval != 0 ||
+		clientConfig.RetryRudderErrors.Multiplier != 0
+
+	// Apply defaults only for truly unset values, but respect explicit MaxRetry=0
+	if retryConfig.MaxRetry == 0 && !hasExplicitConfig {
+		retryConfig.MaxRetry = defaultMaxRetry
+	}
```
atzoum: please explain this part

processor/internal/transformer/utils.go
```diff
@@ -39,6 +39,12 @@ func TransformerClientConfig(conf *config.Config, configPrefix string) *transfor
 	transformerClientConfig.TransportConfig.IdleConnTimeout = conf.GetDurationVar(5, time.Second, fmt.Sprintf("Transformer.Client.%s.maxIdleConnDuration", configPrefix), "Transformer.Client.maxIdleConnDuration")
 	transformerClientConfig.Recycle = conf.GetBoolVar(false, fmt.Sprintf("Transformer.Client.%s.recycle", configPrefix), "Transformer.Client.recycle")
 	transformerClientConfig.RecycleTTL = conf.GetDurationVar(60, time.Second, fmt.Sprintf("Transformer.Client.%s.recycleTTL", configPrefix), "Transformer.Client.recycleTTL")
+	transformerClientConfig.RetryRudderErrors.Enabled = conf.GetBoolVar(true, fmt.Sprintf("Transformer.Client.%s.retryRudderErrors.enabled", configPrefix), "Transformer.Client.retryRudderErrors.enabled")
```
atzoum, line 42: we need the same inside router's `transformerClientConfig`

internal/transformer-client/client.go
```diff
@@ -3,14 +3,36 @@
...
+	defaultMaxIdleConnsPerHost = 10
+	defaultIdleConnTimeout     = 30 * time.Second
+	defaultClientTimeout       = 600 * time.Second
+	defaultClientTTL           = 10 * time.Second
+	defaultRecycleTTL          = 60 * time.Second
+	defaultMaxRetry            = -1
+	defaultInitialInterval     = 1 * time.Second
+	defaultMaxInterval         = 30 * time.Second
+	defaultMaxElapsedTime      = 0
+	defaultMultiplier          = 2.0
```
atzoum:
```suggestion

	defaultRetryRudderErrorsMaxRetry        = -1
	defaultRetryRudderErrorsInitialInterval = 1 * time.Second
	defaultRetryRudderErrorsMaxInterval     = 30 * time.Second
	defaultRetryRudderErrorsMaxElapsedTime  = 0
	defaultRetryRudderErrorsMultiplier      = 2.0
```

internal/transformer-client/client.go
```diff
@@ -27,73 +49,211 @@ type ClientConfig struct {
 	PickerType    string        // power_of_two(default), round_robin, least_loaded_random, least_loaded_round_robin, random
 	Recycle       bool          // false
 	RecycleTTL    time.Duration // 60s
+
+	// Configuration for retryable HTTP client in case of [X-Rudder-Should-Retry: true] HTTP 503 responses
+	RetryRudderErrors struct {
+		Enabled         bool          // true
```
atzoum: by default (zero value) we treat it as disabled, right?

internal/transformer-client/client.go
```diff
@@ -27,73 +49,211 @@ type ClientConfig struct {
...
 }
 
 func NewClient(config *ClientConfig) Client {
-	transport := &http.Transport{
-		DisableKeepAlives:   true,
-		MaxConnsPerHost:     100,
-		MaxIdleConnsPerHost: 10,
-		IdleConnTimeout:     30 * time.Second,
+	if config == nil {
+		return newRetryableHTTPClient(buildDefaultClient(), nil)
```
atzoum: why do we need to handle this scenario (nil ClientConfig)? We are not actually using it

internal/transformer-client/client.go
cisse21: I was overthinking a lot. Removed it

router/transformer/transformer_test.go
```diff
@@ -61,6 +61,14 @@ func (a *mockAdapter) getResponse(response []byte, respCode int, metadata []Prox
...
 	initMocks(t)
 
+	config.Set("Transformer.Client.Router.retryRudderErrors.enabled", false)
+	config.Set("HttpClient.backendProxy.timeout", "1s")
+	config.Set("HttpClient.routerTransformer.timeout", "1s")
+
+	defer func() {
+		config.Reset()
+	}()
+
```
atzoum: why do we need this?

internal/transformer-client/client.go
```diff
@@ -27,73 +50,192 @@ type ClientConfig struct {
...
+
+// buildDefaultTransport creates a transport with default settings
+func buildDefaultTransport() *http.Transport {
+	return &http.Transport{
+		DisableKeepAlives:   defaultDisableKeepAlives,
+		MaxConnsPerHost:     defaultMaxConnsPerHost,
+		MaxIdleConnsPerHost: defaultMaxIdleConnsPerHost,
+		IdleConnTimeout:     defaultIdleConnTimeout,
+	}
+}
```
atzoum: we can now inline this in `buildConfiguredTransport`

internal/transformer-client/client.go
```diff
@@ -27,73 +50,192 @@ type ClientConfig struct {
...
-	if config == nil {
-		return client
+		Timeout:   getClientTimeout(config),
 	}
 
-	transport.DisableKeepAlives = config.TransportConfig.DisableKeepAlives
-	if config.TransportConfig.MaxConnsPerHost != 0 {
-		transport.MaxConnsPerHost = config.TransportConfig.MaxConnsPerHost
+	retryableConfig := buildRetryableConfig(config)
+	if retryableConfig != nil {
```
atzoum, line 97: this is never going to be nil

internal/transformer-client/client.go
```diff
@@ -27,73 +50,192 @@ type ClientConfig struct {
...
-	switch config.ClientType {
-	case "httplb":
-		tr := &httplbtransport{
-			MaxConnsPerHost:     transport.MaxConnsPerHost,
-			MaxIdleConnsPerHost: transport.MaxIdleConnsPerHost,
+// buildConfiguredTransport creates a transport with configuration applied
+func buildConfiguredTransport(config *ClientConfig) *http.Transport {
+	transport := buildDefaultTransport()
+
+	if config != nil {
```
atzoum: this is never going to be nil

atzoum (approved):

go.mod
```diff
@@ -116,10 +116,7 @@ require (
 	google.golang.org/protobuf v1.36.6
 )
 
-require (
-	github.com/cenkalti/backoff/v5 v5.0.2 // indirect
-	github.com/containerd/errdefs/pkg v0.3.0 // indirect
-)
+require github.com/cenkalti/backoff/v5 v5.0.2 // indirect
```
atzoum, line 119: we should migrate everything to cenkalti/backoff/v5 at some point

PR #5954, opened 2025-06-06
fix: warehouse transformations responses ordering

achettyiitr:
# Description

- Avoid making modifications, such as ordering a legacy response, as it will be consumed later.
- Avoid concurrent transformations like what's happening in the rudder-transformer JS code. Once we completely migrated to warehouse transformations, we can then later have concurrency set.

## Linear Ticket

- Resolves WAR-742, WAR-697

## Security

### DIFF
--- a/processor/internal/transformer/destination_transformer/embedded/warehouse/uploader.go
+++ b/processor/internal/transformer/destination_transformer/embedded/warehouse/uploader.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"path"
-	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


processor/internal/transformer/destination_transformer/embedded/warehouse/uploader.go
```diff
@@ -101,11 +100,6 @@ func (t *Transformer) sampleDiff(events []types.TransformerEvent, legacyResponse
 		return "" // Don't diff in case there is no response from transformer
 	}
 
-	sortTransformerResponsesByJobID(legacyResponse.Events)
```
shekhar-rudder, line 104: Why were we sorting ?

achettyiitr: So as to compare the event. In rudder-tranformer processing is sequential. But in warehouse transformations, it is concurrent. 
So before comparing, we are sorting to make sure they are of the same order.

RanjeetMishra: @achettyiitr To clarify we are going ahead with safe approach of non concurrency than changing logic which avoids sorting?

achettyiitr: Yeah. If we need to avoid sorting, there is an extra complexity around either deep copying or during comparision.

RanjeetMishra (approved):

PR #5975, opened 2025-06-10
fix: clean up async destination struct after upload is finished

koladilip:
# Description

The PR fixes the logic for cleaning up the async upload state in the batch router.
Previously, the cleanup (asyncStructCleanUp) was being called unconditionally after each upload attempt, regardless of whether the upload iteration was actually completed.
Now, the cleanup is only performed if the upload iteration is truly completed (i.e., when there are no more jobs left to import for that destination).
The UploadInProgress flag is now set to true only when there are jobs still being imported; otherwise, cleanup is performed.
Additionally, the UploadInProgress flag is explicitly reset to false inside the asyncStructCleanUp function to ensure the state is always consistent after cleanup.

**Key Code Changes:**
In pollAsyncStatus, the unconditional setting of UploadInProgress = false was removed, and cleanup is now only called after a successful update.
In asyncUploadWorker, the logic for calling asyncStructCleanUp was moved inside an else block, so it only happens when there are no more jobs to import.
The UploadInProgress flag is now reset to false in asyncStructCleanUp.

**Impact:**
This change ensures that the async upload state is only cleaned up when an upload iteration is truly finished, preventing premature cleanup and potential issues with job processing state.

## Linear Ticket
https://linear.app/rudderstack/issue/INT-3807/fix-cleaning-up-async-destination-struct
Resolves INT-3807
## Security

### DIFF
--- a/router/batchrouter/handle_async.go
+++ b/router/batchrouter/handle_async.go
@@ -323,8 +323,8 @@ func (brt *Handle) pollAsyncStatus(ctx context.Context) {
 				}
 				statusList, err := brt.updatePollStatusToDB(ctx, destinationID, sourceID, importingJob, pollResp)
 				if err == nil {
-					brt.asyncDestinationStruct[destinationID].UploadInProgress = false
 					brt.recordAsyncDestinationDeliveryStatus(sourceID, destinationID, statusList)
+					brt.asyncStructCleanUp(destinationID)
 				}
 			}
 		}
@@ -360,9 +360,7 @@ func (brt *Handle) asyncUploadWorker(ctx context.Context) {
 					brt.asyncDestinationStruct[destinationID].CanUpload = true
 					brt.asyncDestinationStruct[destinationID].PartFileNumber++
 					uploadResponse := brt.asyncDestinationStruct[destinationID].Manager.Upload(brt.asyncDestinationStruct[destinationID])
-					if uploadResponse.ImportingParameters != nil && len(uploadResponse.ImportingJobIDs) > 0 {
-						brt.asyncDestinationStruct[destinationID].UploadInProgress = true
-					}
+
 					brt.setMultipleJobStatus(setMultipleJobStatusParams{
 						AsyncOutput:           uploadResponse,
 						Attempted:             true,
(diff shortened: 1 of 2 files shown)


router/batchrouter/handle_observability.go
```diff
@@ -150,37 +193,34 @@ func (brt *Handle) emitAsyncEventDeliveryTimeMetrics(sourceID, destinationID str
 		}
 
 		// Get original job parameters for this job
-		originalParams, exists := asyncDestStruct.OriginalJobParameters[status.JobID]
```
atzoum: it is not clear what was wrong with the previous implementation and it is also not clear why what we are doing now is better

koladilip: **Problem:** Job parameters were being extracted from asyncDestStruct.OriginalJobParameters, but this field is unreliable because it gets reset immediately after upload completion, before the polling process finishes. This causes the parameters to be unavailable during polling operations.
**Current Solution:** Added fallback logic to check both asyncDestStruct.OriginalJobParameters and the job status for parameters, using whichever source has the data available.
**Root Cause:** The OriginalJobParameters field is being cleared too early in the process - right after upload instead of after the entire polling cycle completes.
**Future Improvement:** The proper fix would be to delay resetting OriginalJobParameters until after polling finishes, but this requires further investigation and will be addressed in a future update.

atzoum: I don't understand why this solution can really fix the issue. Even if it fixes some reporting issues, this early reset that you mentioned, doesn't it cause other issues too?
I guess this is a long-standing issue, so if it has been with us for so long, it shouldn't be that urgent. Why don't we spend the effort to fix it properly?

koladilip: this is fix only for emitting delivery time metrics for async destinations, it wasn't getting emitted as it is getting reset after upload.

koladilip: I have added fix the main issue also you can check.

atzoum: So was the actual issue that we were not properly maintaining the `UploadInProgress` flag causing cleanup to happen where it shouldn't?
Why do we need all these observability-related changes if we are fixing the root cause?

koladilip: make sense, we don't need fallback logic now. I will remove it

koladilip: doing this as regular release.

atzoum (approved):

ItsSudip (approved):

PR #5976, opened 2025-06-11
chore: create new backoff on every reporting request

satishrudderstack:
# Description

With the shared `backoff`, the retries could be waiting for longer time than required. With this PR, we are creating new backoff for every request

## Security


enterprise/reporting/client/client.go
```diff
@@ -157,7 +160,7 @@ func (c *Client) Send(ctx context.Context, payload any) error {
 		return err
 	}
 
-	b := backoff.WithContext(c.backoff, ctx)
```
itsmihir, line 160: we can remove `backoff` from client since we are not using it.

itsmihir (approved):

ktgowtham (approved):

PR #5979, opened 2025-06-11
fix: warehouse transformations deep copying arguments

achettyiitr:
# Description

- Using deepcopy for client events and legacy response, so as to be careful of any modifications within the embedded warehouse transformations module.
- Added test within the processor to cover the warehouse transformations component.

## Linear Ticket

- Resolves WAR-762

## Security

### DIFF
--- a/processor/internal/transformer/destination_transformer/destination_transformer_test.go
+++ b/processor/internal/transformer/destination_transformer/destination_transformer_test.go
@@ -795,10 +795,7 @@ func TestEmbeddedWarehouseTransformer(t *testing.T) {
 				"receivedAt":        "2021-09-01T00:00:00.000Z",
 				"sentAt":            "2021-09-01T00:00:00.000Z",
 				"timestamp":         "2021-09-01T00:00:00.000Z",
-				"traits": map[string]any{
-					"email": float64(12),
-				},
-				"type": "track",
+				"type":              "track",
 			},
 			Metadata: types.Metadata{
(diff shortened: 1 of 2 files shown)


processor/internal/transformer/destination_transformer/destination_transformer_test.go
```diff
@@ -763,3 +768,125 @@ func TestLongRunningTransformation(t *testing.T) {
...
+			Message: map[string]interface{}{
+				"src-key-1":         "test-value-1",
+				"messageId":         "messageId" + strconv.Itoa(index+1),
+				"event":             "event" + strconv.Itoa(index+1),
+				"originalTimestamp": "2021-09-01T00:00:00.000Z",
+				"receivedAt":        "2021-09-01T00:00:00.000Z",
+				"sentAt":            "2021-09-01T00:00:00.000Z",
+				"timestamp":         "2021-09-01T00:00:00.000Z",
+				"traits": map[string]any{
+					"email": float64(12),
```
shekhar-rudder: There is no email column in the output because of `skipTracksTable` being enabled ?

achettyiitr: For the track event, we only process top level message.context, message.properties, message.userProperties, so it won't be present. Removing it to avoid confusion.

shekhar-rudder (approved):

cisse21 (approved):

PR #5987, opened 2025-06-13
fix: signal handling in NotifyContextWithCallback

mihir20:
# Description

test "does not call the callback if context is canceled by us" was flaky as

1. When cancel() is called in the test, it calls both cancel() and signalCancel()
2. The goroutine's select statement might choose either <-signalCtx.Done() or <-ctx.Done()
3. If it happens to select <-signalCtx.Done(), it will incorrectly call the callback function

---

solution: cancel `signalCtx` when ctx is done

## Linear Ticket

-

## Security

### DIFF
--- a/utils/signal/signal.go
+++ b/utils/signal/signal.go
@@ -29,7 +29,6 @@ func NotifyContextWithCallback(fn func(), signals ...os.Signal) (ctx context.Con
 
 	return ctx, func() {
 		cancel()
-		signalCancel()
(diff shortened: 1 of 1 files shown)


ktgowtham (commented):
How about just removing signalCancel() in the return statement? 
We call signalCancel() inside ctx.Done() rt?

ktgowtham (approved):

atzoum (approved):

PR #5996, opened 2025-06-16
chore: gateway warehouse integration test

achettyiitr:
# Description

- Added Warehouse integrations tests with events sent to the Gateway instead of creating a staging file.

## Linear Ticket

- Resolves WAR-773.

## Security


integration_test/warehouse/warehouse_test.go
```diff
@@ -2468,6 +2644,74 @@ func requireReportsCount(
...
+	t *testing.T,
+	db *sql.DB,
+	queue, state string,
+	expectedCount int,
+) {
+	t.Helper()
+
+	require.Eventually(t, func() bool {
+		var jobsCount int
+		require.NoError(t, db.QueryRow(fmt.Sprintf(`SELECT COUNT(*) FROM unionjobsdbmetadata('%s',1) WHERE job_state = '%s';`, queue, state)).Scan(&jobsCount))
```
shekhar-rudder, line 2724: What is `unionjobsdbmetadata` ?

integration_test/warehouse/warehouse_test.go
```diff
@@ -2281,7 +2430,7 @@ func requireStagingFileEventsCount(
 			t.Logf("Staging file events count: %d", eventsCount)
 			return eventsCount == expectedCount
 		},
-		10*time.Second,
+		180*time.Second,
```
shekhar-rudder: Why are we changing 10 -> 180 ?

integration_test/warehouse/warehouse_test.go
```diff
@@ -1590,6 +1597,103 @@ func TestUploads(t *testing.T) {
...
+		return err
+	})
+
+	url := fmt.Sprintf("http://localhost:%d", gwPort)
+	health.WaitUntilReady(ctx, t, url+"/health", 60*time.Second, 10*time.Millisecond, t.Name())
+
+	eventsCount := 10000
+	db := sqlmw.New(postgresContainer.DB)
+
+	err = sendEvents(100, 100, "track", writeKey, url)
```
shekhar-rudder:
```
batchSize = 100
sendEvents(batchSize, eventsCount/batchSize, "track", writeKey, url)
```

integration_test/warehouse/warehouse_test.go
achettyiitr: https://github.com/rudderlabs/rudder-server/blob/d90ea68178d8ad6175d0145568fd7e2d651ed760/sql/migrations/node/000010_unionjobsdb_fn.up.sql#L50

integration_test/warehouse/warehouse_test.go
achettyiitr: Reverted it back. It might not be required since the previous assertion `requireJobsCount(t, postgresContainer.DB, "batch_rt", jobsdb.Succeeded.State, eventsCount)` would be successful.

integration_test/warehouse/warehouse_test.go
```diff
@@ -1590,6 +1597,105 @@ func TestUploads(t *testing.T) {
...
+	requireTableUploadEventsCount(t, ctx, db, eventsCount, []lo.Tuple2[string, any]{
+		{A: "status", B: exportedData},
+		{A: "wh_uploads.source_id", B: sourceID},
+		{A: "wh_uploads.destination_id", B: destinationID},
+		{A: "wh_uploads.namespace", B: namespace},
+	}...)
+	requireDownstreamEventsCount(t, ctx, db, fmt.Sprintf("%s.%s", namespace, "tracks"), eventsCount)
+	requireDistinctDownstreamEventsCount(t, ctx, db, fmt.Sprintf("%s.%s", namespace, "tracks"), "ID", eventsCount)
+
+	cancel()
```
RanjeetMishra, line 1700: @achettyiitr What is this cancel() for? And there is `defer cancel()` before.

achettyiitr: We are calling cancel so that runRudderServer can exit. `defer cancel()` Just in case if we get error earlier even before rudder server starts.

RanjeetMishra (approved):

achettyiitr:
> @achettyiitr As part of this test whenever we enable WH transformation path, would it consider that in code flow?

Yes

PR #5998, opened 2025-06-16
chore: add metrics for duplicate events in staging files

shekhar-rudder:
# Description

Implementation details in https://linear.app/rudderstack/issue/WAR-770

## Security

### DIFF
--- a/warehouse/slave/worker.go
+++ b/warehouse/slave/worker.go
@@ -297,6 +297,10 @@ func (w *worker) processSingleStagingFile(
 	discardsTable := job.discardsTable()
 	sortedTableColumnMap := job.sortedColumnMapForAllTables()
 
+	// Duplicate detection map for messageId
+	messageIdSet := make(map[string]struct{})
+	duplicateCount := 0
+
 	for {
 		ok := bufScanner.Scan()
 		if !ok {
@@ -340,6 +344,19 @@ func (w *worker) processSingleStagingFile(
 		}
 
 		eventLoader := w.encodingFactory.NewEventLoader(writer, job.LoadFileType, job.DestinationType)
+
+		// Duplicate detection by messageId
+		messageIdVal, ok := columnData["messageId"]
+		if ok {
+			messageIdStr, ok := messageIdVal.(string)
+			if ok {
+				if _, exists := messageIdSet[messageIdStr]; exists {
+					duplicateCount++
+				} else {
+					messageIdSet[messageIdStr] = struct{}{}
+				}
+			}
+		}
 
 		for _, columnName := range sortedTableColumnMap[tableName] {
 			if eventLoader.IsLoadTimeColumn(columnName) {
@@ -461,6 +478,15 @@ func (w *worker) processSingleStagingFile(
 		releaseWriter()
 
 		jr.incrementEventCount(tableName)
+	}
+
+	// After processing all lines, increment the metric for duplicates
+	if duplicateCount > 0 {
+		jr.StagingFileDuplicateEvents.Count(duplicateCount)
+		jr.logger.Infon("Found duplicate events in staging file",
+			logger.NewField("stagingFileID", stagingFile.ID),
+			logger.NewIntField("duplicateEvents", int64(duplicateCount)),
+		)
 	}
 
 	jr.logger.Debugn("Process bytes from downloaded staging file",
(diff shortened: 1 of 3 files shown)


warehouse/slave/worker.go
```diff
@@ -297,6 +297,10 @@ func (w *worker) processSingleStagingFile(
 	discardsTable := job.discardsTable()
 	sortedTableColumnMap := job.sortedColumnMapForAllTables()
 
+	// Duplicate detection map for messageId
+	messageIdSet := make(map[string]struct{})
```
achettyiitr:
```suggestion
	messageIDSet := make(map[string]struct{})
```

warehouse/slave/worker.go
```diff
@@ -341,6 +345,19 @@ func (w *worker) processSingleStagingFile(
 
 		eventLoader := w.encodingFactory.NewEventLoader(writer, job.LoadFileType, job.DestinationType)
 
+		// Duplicate detection by messageId
+		messageIdVal, ok := columnData["messageId"]
```
achettyiitr:
```suggestion
		messageIDVal, ok := columnData["messageId"]
```

achettyiitr: It should be the `id` column, not the `messageId`. Also for snowflake it will be in upper case `ID`. Please verify once.

warehouse/slave/worker_job.go
```diff
@@ -145,6 +145,7 @@ type jobRun struct {
 	bytesProcessedStagingFileStat  stats.Measurement
 	bytesDownloadedStagingFileStat stats.Measurement
 	downloadStagingFileFailedStat  stats.Measurement
+	StagingFileDuplicateEvents     stats.Measurement
```
achettyiitr: We can be specific here on what measurement is being used?
```suggestion
	StagingFileDuplicateEvents     stats.Counter
```

achettyiitr: Any reason to keep it exported?
```suggestion
	stagingFileDuplicateEvents     stats.Measurement
```

shekhar-rudder: I went with `stats.Measurement` since that is how all the other stats were typed

shekhar-rudder: Made it unexported

warehouse/slave/worker.go
shekhar-rudder: Addressed

warehouse/slave/worker.go
shekhar-rudder: Addressed

shekhar-rudder: Changed it to id column with case handling

warehouse/slave/worker.go
```diff
@@ -341,6 +345,19 @@ func (w *worker) processSingleStagingFile(
...
+		if ok {
+			iDStr, ok := iDVal.(string)
+			if ok {
+				if _, exists := iDColumnSet[iDStr]; exists {
+					duplicateCount++
+				} else {
+					iDColumnSet[iDStr] = struct{}{}
+				}
+			}
+		}
```
RanjeetMishra: Discussion with @shekhar-rudder to validate:
1. Can same event for tracks and other event table exist in staging file?
2. Irrespective of point 1, does it make sense to do this duplicate check at table level? As in if we were planning to do it at load files stage, thats how we would have checked for duplicates.
CC: @achettyiitr

achettyiitr: 1. Yeah, it's possible.
2. +1

shekhar-rudder: Updated code to check duplicates at table level

warehouse/slave/worker.go
```diff
@@ -297,6 +297,10 @@ func (w *worker) processSingleStagingFile(
 	discardsTable := job.discardsTable()
 	sortedTableColumnMap := job.sortedColumnMapForAllTables()
 
+	// Outer key is table name, inner key is id column value
+	tableIDColumnSet := make(map[string]map[string]struct{})
```
vyeshwanth: can we just do it with one map?

```
type dedupKey struct {
   tableName string
   idValue string
}

var tableIDColSet map[dedupKey]bool
```

shekhar-rudder: Done

vyeshwanth (approved):

PR #5999, opened 2025-06-16
chore: create enterprise images with an additional race enabled build

shekhar-rudder:
# Description

- For every enterprise image/tag that is built and pushed, also build and push a new image with the environment variable RACE_ENABLED=TRUE.
- The new image should have the same tag as the original, but with -race appended (e.g., if the tag is 1.2.3, the new tag should be 1.2.3-race).

## Security

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -190,7 +190,7 @@ jobs:
           context: .
           platforms: ${{ matrix.build-config.platform }}
           push: true
-          tags: ${{ matrix.build-config.tags }}{{ matrix.race == 'true' && '-race' || '' }}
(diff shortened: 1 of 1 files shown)


achettyiitr (approved):

.github/workflows/builds.yml
```diff
@@ -172,6 +175,7 @@ jobs:
             tags: ${{needs.docker-ent-meta.outputs.amd64_tags}}
             labels: ${{needs.docker-ent-meta.outputs.amd64_labels}}
             platform: linux/amd64
+        race: [false, true]
```
RanjeetMishra, line 178: @shekhar-rudder If both we set as false, then `-race` builds will not be generated?

shekhar-rudder: Yes

PR #6002, opened 2025-06-17
feat: add validations for iceberg snowpipe streaming

shekhar-rudder:
# Description

## Security


achettyiitr (approved):

warehouse/integrations/snowflake/snowflake.go
```diff
@@ -254,6 +254,16 @@ func (sf *Snowflake) schemaExists(ctx context.Context) (exists bool, err error)
 	return
 }
 
+func (sf *Snowflake) HasExternalVolumeAccess(ctx context.Context, externalVolume string) error {
+	query := fmt.Sprintf("DESC EXTERNAL VOLUME %s;", externalVolume)
+	// If no error is returned, then the external volume exists and user has access to it
```
RanjeetMishra, line 259: @shekhar-rudder I think this query will only validate metadata about external storage exists. In order to validate if snowflake has access to it, we need to create temp external table and delete?
I see there is step of `VerifyingCreateAndAlterTable` already. Can you check if anything needs to change in that step to use external storage?

warehouse/validations/steps.go
```diff
@@ -21,14 +21,19 @@ func StepsToValidate(dest *backendconfig.DestinationT) *model.StepsResponse {
...
+			{ID: 2, Name: model.VerifyingCreateSchema},
+			{ID: 3, Name: model.VerifyingCreateAndAlterTable},
+			{ID: 4, Name: model.VerifyingFetchSchema},
 		}
+		warehouse := model.Warehouse{
+			Destination: *dest,
+		}
+		if warehouse.GetBoolDestinationConfig(model.EnableIcebergSetting) {
+			steps = append(steps, &model.Step{ID: 5, Name: model.VerifyingExternalVolume})
+		}
```
RanjeetMishra, line 35: So this will not be a new destination, just a flag and other configs? Makes sense if yes.

shekhar-rudder: Yes this won't be a new destination. We will be adding new configs

warehouse/integrations/snowflake/snowflake.go
shekhar-rudder: Yes this query won't validate if the external volume has been configured correctly with right credentials. Let me add it.

PR #6005, opened 2025-06-17
chore: downgrade sync dependency

mihir20:
# Description

downgrade sync dependencies

ref: https://rudderlabs.slack.com/archives/C02HCFVJVAR/p1750151104040389

## Linear Ticket

-

## Security


atzoum (approved):

go.mod
```diff
@@ -17,6 +17,7 @@ replace (
 	golang.org/x/crypto => golang.org/x/crypto v0.32.0
 	golang.org/x/image => golang.org/x/image v0.24.0
 	golang.org/x/net => golang.org/x/net v0.34.0
+	golang.org/x/sync => golang.org/x/sync v0.13.0
```
atzoum, line 20: why a replace?

mihir20: because downgrading is causing other deps to be downgraded. we have fixed a bug in go-kit, I don't want to downgrade go-kit

PR #6012, opened 2025-06-18
fix: incorrect snowpipe channel id being returned if insert recreates the channel

shekhar-rudder:
# Description

**BUG**
A fix was made for preventing duplicates in Snowpipe in PR #5957 . But it missed returning the id of the newly created channel in the `Upload` method. As a result, `Poll` method was still receiving the stale channel id. So the GetStatus API call was failing with 404, causing the event to be marked as failed. This failed event would be retried causing duplicates

**FIX**
- The actual fix was simple to propagate the new channel id from the Insert API method back to the caller. But there was another miss, that post channel recreation the cache was not being updated. So instead this PR, moves the recreation logic from the API layer to the manager layer. This meant reverting the changes done in `api.go` and `apiadapter.go` and the corresponding test files
- In case of 404 status code, the api insert method is now returning an error `ErrChannelNotFound`
- Created a `insert` method in manager which has the channel recreation logic based on the error type. This method is also being called from `discards`
- Added a regression unit test

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
@@ -399,50 +399,60 @@ func (m *Manager) sendEventsToSnowpipe(
 }
 
 func (m *Manager) insert(ctx context.Context, destinationID string, destConf *destConfig, info *uploadInfo, insertReq *model.InsertRequest, channelID string) (string, error) {
-	handleInsertFailure := func(res *model.InsertResponse) error {
-		if deleteErr := m.deleteChannel(ctx, info.tableName, channelID); deleteErr != nil {
-			m.logger.Warnn("Failed to delete channel", obskit.Error(deleteErr))
+	deleteChannel := func(tableName, chID string) {
+		if err := m.deleteChannel(ctx, tableName, chID); err != nil {
+			m.logger.Warnn("Failed to delete channel", obskit.Error(err))
 		}
+	}
+	extractError := func(res *model.InsertResponse, tableName string) error {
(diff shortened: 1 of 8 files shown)


router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -382,38 +382,69 @@ func (m *Manager) sendEventsToSnowpipe(
...
-		ChannelID: channelResponse.ChannelID,
+		ChannelID: channelID,
 		Offset:    offset,
 		Table:     info.tableName,
 		Count:     len(info.events),
 	}
 	return imInfo, discardImInfo, nil
 }
 
+func (m *Manager) insert(ctx context.Context, destinationID string, destConf *destConfig, info *uploadInfo, insertReq *model.InsertRequest, channelID string) (string, error) {
```
RanjeetMishra, line 401: @shekhar-rudder This func implementation can it be modified in below way to make it more clear?
1. Check status of passed channel if it exists? Instead of attempting using Insert func, is there another way to check if channel exists?
2.If channel does not exist: Clear cache and create new/update cache. If creation error return with error and log.
3. Either existing or new channel created successfully. Attempt insert and handle failure. Also insert failure handling would be just 1 way of deleting channel, propagating proper error?

Another thing, lets add log with tablename and channelID for channel to track lifecycle
1. creation
2. channel not found
3. if not logging anywhere can log insert error with tablename and channelid
4. channel deleted (both successfully or with failure)

shekhar-rudder: Refactored the method

RanjeetMishra (approved):

vyeshwanth (approved):

PR #6013, opened 2025-06-18
feat: propagate snowpipe iceberg config to snowpipe service

shekhar-rudder:
# Description

This PR adds propagates the `enableIceberg` field in the destination config to the Snowpipe service so that the service can create Iceberg clients.

Implementation
- Added a new field `EnableIceberg` in the `TableConfig` request body of create channel API

## Security

### DIFF
--- a/warehouse/integrations/snowflake/table_manager.go
+++ b/warehouse/integrations/snowflake/table_manager.go
@@ -78,7 +78,7 @@ func newIcebergTableManager(externalVolume string) tableManager {
 			"bigint":   "number(19,0)",
 			"float":    "double precision",
 			"string":   "varchar",
-			"datetime": "timestamp_ltz(6)",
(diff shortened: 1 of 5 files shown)


warehouse/integrations/snowflake/table_manager.go
```diff
@@ -78,7 +78,7 @@ func newIcebergTableManager(externalVolume string) tableManager {
 			"bigint":   "number(19,0)",
 			"float":    "double precision",
 			"string":   "varchar",
-			"datetime": "timestamp_ltz(6)",
+			"datetime": "timestamp_ntz(6)",
```
shekhar-rudder, line 81: This change is unrelated to rest of the PR. We should be using `timestamp_ntz` instead of `timestamp_ltz`

achettyiitr (approved):

vyeshwanth (approved):

PR #6018, opened 2025-06-18
feat: implement configurable event blocking

itsmihir:
# Description

This PR adds event blocking functionality with a strategic split between gateway and processor components to optimize metrics collection and avoid unnecessary live event recording.

## Architecture & Design:
- Gateway: Contains the blocking logic and marks events for blocking
- Processor: Performs the actual event dropping after deduplication.

Why not have the logic in a single place (in the gateway or the processor)?
- Accurate Metrics: We want to capture metrics after deduplication in the processor to ensure accurate event counts, not before deduplication in the gateway
- Live Events: Source live events are captured at the gateway, and we don't want to send blocked events. If we only had logic in processor, blocked events would still appear in live events.

## Core Features:
- Events can be blocked per workspace based on event type and event name
- Only applies to event stream sources (non-warehouse sources)
- For MVP only track events can be blocked.

## Key Components:
- `isEventBlocked()` method in gateway to determine blocking eligibility
- Integration with `extractJobsFromInternalBatchPayload()` to mark blocked events
- Processor integration to drop events marked as blocked

## Processing Flow
1. Gateway: Evaluates events against blocking rules
2. Gateway: Marks blocked events and skips live event recording
3. Processor: Performs deduplication on all events
4. Processor: Captures metrics post-deduplication (Not part of this PR)
5. Processor: Drops events marked as blocked

## Linear Ticket

[OBS-853
](https://linear.app/rudderstack/issue/OBS-853/add-event-blocking-logic-in-gateway-and-block-events-in-processor)

## Security


itsmihir:
https://github.com/coderabbitai full review

itsmihir:
@coderabbitai full review

gateway/handle.go
```diff
@@ -582,6 +585,52 @@ func (gw *Handle) isUserSuppressed(workspaceID, userID, sourceID string) bool {
...
+		return false
+	}
+
+	if blockedEvents, exists := workspaceSettings.EventBlocking.Events[eventType]; exists {
+		for _, blockedEvent := range blockedEvents {
+			if blockedEvent == eventName {
+				return true
+			}
+		}
+	}
```
cisse21:
```suggestion
	if blockedEvents, exists := workspaceSettings.EventBlocking.Events[eventType]; exists {
		return slices.Contains(blockedEvents, eventName)
	}
```

gateway/handle.go
```diff
@@ -953,6 +1007,11 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(reqType string, body []byt
 			BotAction:           msg.Properties.BotAction,
 		}
 
+		eventName := stringify.Any(gjson.GetBytes(msg.Payload, "event").String())
```
cisse21:
```suggestion
		eventName := gjson.GetBytes(msg.Payload, "event").String()
```

gateway/handle.go
```diff
@@ -777,6 +830,7 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(reqType string, body []byt
 	var (
 		messages         []stream.Message
 		isUserSuppressed = gw.memoizedIsUserSuppressed()
+		isEventBlocked   = gw.memoizedIsEventBlocked()
```
cisse21, line 821: Why can't we add the memoized function to the gw handle and use some the ttl cache in go-kit to evict things over time? Doing caching at a batch level might not be that fruitful

gateway/handle.go
```diff
@@ -744,8 +795,9 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
 }
 
 type jobWithStat struct {
```
cisse21: probably we should be changing the name since its no longer just `jobWithStat`

gateway/handle.go
```diff
@@ -128,6 +130,7 @@ type Handle struct {
 		gwAllowPartialWriteWithErrors        config.ValueLoader[bool]
 		enableInternalBatchValidator         config.ValueLoader[bool]
 		enableInternalBatchEnrichment        config.ValueLoader[bool]
+		enableEventBlocking                  bool
```
cisse21:
```suggestion
		enableEventBlocking                  config.ValueLoader[bool]
```

gateway/handle.go
```diff
@@ -582,6 +585,52 @@ func (gw *Handle) isUserSuppressed(workspaceID, userID, sourceID string) bool {
...
+		}
+		val := gw.isEventBlocked(workspaceID, sourceID, eventType, eventName)
+		cache[key] = val
+		return val
+	}
+}
+
+// isEventBlocked checks if an event should be blocked based on workspace settings
+func (gw *Handle) isEventBlocked(workspaceID, sourceID, eventType, eventName string) bool {
+	if !gw.conf.enableEventBlocking {
```
cisse21:
```suggestion
	if !gw.conf.enableEventBlocking.Load() {
```

gateway/handle_test.go
```diff
@@ -0,0 +1,512 @@
...
+			WriteTimeout                                                                      time.Duration
+			IdleTimeout                                                                       time.Duration
+			allowReqsWithoutUserIDAndAnonymousID                                              config.ValueLoader[bool]
+			gwAllowPartialWriteWithErrors                                                     config.ValueLoader[bool]
+			enableInternalBatchValidator                                                      config.ValueLoader[bool]
+			enableInternalBatchEnrichment                                                     config.ValueLoader[bool]
+			enableEventBlocking                                                               bool
+			webhookV2HandlerEnabled                                                           bool
+		}{
+			enableEventBlocking:           enableEventBlocking,
```
cisse21:
```suggestion
			enableEventBlocking                                                               config.ValueLoader[bool]
			webhookV2HandlerEnabled                                                           bool
		}{
			enableEventBlocking:           config.SingleValueLoader(enableEventBlocking),
```

gateway/handle.go
itsmihir: I chose to keep it a constant boolean to avoid calling Load() for every event.

gateway/handle.go
itsmihir: The main reason for adding the memoized function is to avoid acquiring the configSubscriberLock for every event. Introducing global caching wouldn’t provide enough benefit to justify the added complexity.

gateway/handle.go
cisse21: Calling Load isn't a costly operation as we just read from pointer and this makes it easy to enable and disable without restarts

gateway/handle.go
itsmihir: How about `jobWithMetadata`?

cisse21: Sounds good to me

gateway/handle.go
fracasula: @cisse21 beside a mutex we would need something to expire the keys as well or the map will just grow indefinitely over time. For that purpose I have used [this package](https://github.com/rudderlabs/rudder-go-kit/blob/v0.56.0/cachettl/cachettl.go) in the past, but as Mihir said the complexity would definitely be higher.

gateway/handle.go
```diff
@@ -582,6 +586,48 @@ func (gw *Handle) isUserSuppressed(workspaceID, userID, sourceID string) bool {
...
+		return false
+	}
+
+	workspaceSettings, ok := gw.workspaceIDSettingsMap[workspaceID]
+	if !ok {
+		return false
+	}
+
+	if blockedEvents, exists := workspaceSettings.EventBlocking.Events[eventType]; exists {
+		return slices.Contains(blockedEvents, eventName)
```
fracasula: Out of curiosity, how big could this `blockedEvents` slice become (worst case)?

gateway/handle_test.go
```diff
@@ -0,0 +1,512 @@
...
+
+				// Parse the job parameters to check event blocking properties
+				var eventParams map[string]interface{}
+				err := jsonrs.Unmarshal(job.job.Parameters, &eventParams)
+				require.NoError(t, err, "Should be able to parse job parameters")
+
+				// Check IsEventBlocked parameter
+				isEventBlocked, exists := eventParams["is_event_blocked"]
+				if expectedJob.isEventBlocked {
+					assert.True(t, exists, "Job %d: is_event_blocked parameter should exist for blocked events", i)
```
fracasula: Why `assert`, can't you use `require` and fail the test immediately?

itsmihir: My bad, I missed this. I will add this as a cursor rule, for some reason, it keeps preferring assert over require.

gateway/handle.go
itsmihir: For enterprise workspaces we have a limit of 1000 and for growth we have a limit of 2.

fracasula: Just wondering since doing a sequential loop on big slices just to see if they contain a value is quite inefficient.

itsmihir: yeah, we can store the blocked event name in a map, and just do a lookup here.

cisse21 (approved):

PR #6037, opened 2025-06-23
fix: duplicate load files for datalakes

vyeshwanth:
# Description
- Fix duplicate loadfiles for datalakes using deterministic file name using format `<hash of locations of staging files>.<sourceId>.<file format>`

## Linear Ticket
Part of WAR-666

## Security

### DIFF
--- a/warehouse/internal/loadfiles/mock_loadfile_repo_test.go
+++ b/warehouse/internal/loadfiles/mock_loadfile_repo_test.go
@@ -3,6 +3,7 @@ package loadfiles_test
 import (
 	"context"
 	"slices"
+	"sync"
 
 	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
 )
@@ -10,9 +11,12 @@ import (
 type mockLoadFilesRepo struct {
 	id    int64
 	store []model.LoadFile
+	mu    sync.Mutex
 }
 
 func (m *mockLoadFilesRepo) Insert(_ context.Context, loadFiles []model.LoadFile) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	for _, lf := range loadFiles {
 		lf.ID = m.id + 1
 		m.id += 1
@@ -24,6 +28,8 @@ func (m *mockLoadFilesRepo) Insert(_ context.Context, loadFiles []model.LoadFile
 }
 
 func (m *mockLoadFilesRepo) Delete(_ context.Context, uploadID int64, stagingFileIDs []int64) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	store := make([]model.LoadFile, 0)
 	for _, loadFile := range m.store {
 		if !slices.Contains(stagingFileIDs, loadFile.StagingFileID) {
@@ -36,6 +42,8 @@ func (m *mockLoadFilesRepo) Delete(_ context.Context, uploadID int64, stagingFil
 }
 
 func (m *mockLoadFilesRepo) Get(_ context.Context, uploadID int64, stagingFileIDs []int64) ([]model.LoadFile, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	var loadFiles []model.LoadFile
 	for _, loadFile := range m.store {
 		if slices.Contains(stagingFileIDs, loadFile.StagingFileID) {
--- a/warehouse/slave/worker.go
+++ b/warehouse/slave/worker.go
@@ -12,6 +12,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/samber/lo"
+
 	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
 
 	"github.com/rudderlabs/rudder-go-kit/jsonrs"
@@ -271,6 +273,7 @@ func (w *worker) processSingleStagingFile(
 	jr *jobRun,
 	job *basePayload,
 	stagingFile stagingFileInfo,
+	loadFileNamePrefix string,
 ) error {
 	processingStart := jr.now()
 	var err error
@@ -345,7 +348,7 @@ func (w *worker) processSingleStagingFile(
 		}
 
 		// Create separate load file for each table
-		if writer, releaseWriter, err = jr.writer(tableName, stagingFile); err != nil {
+		if writer, releaseWriter, err = jr.writer(tableName, stagingFile, loadFileNamePrefix); err != nil {
 			return err
 		}
 
@@ -446,7 +449,7 @@ func (w *worker) processSingleStagingFile(
 						eventLoader.AddEmptyColumn(columnName)
 					}
 
-					discardWriter, releaseDiscardWriter, err := jr.writer(discardsTable, stagingFile)
+					discardWriter, releaseDiscardWriter, err := jr.writer(discardsTable, stagingFile, loadFileNamePrefix)
 					if err != nil {
 						return err
 					}
@@ -531,7 +534,7 @@ func (w *worker) processStagingFile(ctx context.Context, job *payload) ([]upload
 	if err := w.processSingleStagingFile(ctx, jr, &job.basePayload, stagingFileInfo{
 		ID:       job.StagingFileID,
 		Location: job.StagingFileLocation,
-	}); err != nil {
+	}, ""); err != nil {
 		return nil, err
 	}
 
@@ -554,6 +557,11 @@ func (w *worker) processMultiStagingFiles(ctx context.Context, job *payloadV2) (
 		jr.cleanup()
 	}()
 
+	// calculate load file name prefix using md5 hash of all staging file locations
+	loadFileNamePrefix := warehouseutils.Md5Hash(lo.Reduce(job.StagingFiles, func(agg string, item stagingFileInfo, index int) string {
+		return agg + item.Location
+	}, ""))
+
 	// Initialize Discards Table
 	discardsTable := job.discardsTable()
 	jr.tableEventCountMap[discardsTable] = 0
@@ -566,7 +574,7 @@ func (w *worker) processMultiStagingFiles(ctx context.Context, job *payloadV2) (
 			w.log.Infon("Processing staging-file for upload_v2 job",
 				logger.NewField("stagingFileID", stagingFile.ID),
 			)
-			if err := w.processSingleStagingFile(gCtx, jr, &job.basePayload, stagingFile); err != nil {
+			if err := w.processSingleStagingFile(gCtx, jr, &job.basePayload, stagingFile, loadFileNamePrefix); err != nil {
 				return fmt.Errorf("processing staging file %d: %w", stagingFile.ID, err)
 			}
 			return nil
--- a/warehouse/slave/worker.go
+++ b/warehouse/slave/worker.go
@@ -13,8 +13,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/samber/lo"
-
 	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
 
 	"github.com/rudderlabs/rudder-go-kit/jsonrs"
--- a/warehouse/slave/worker.go
+++ b/warehouse/slave/worker.go
@@ -9,6 +9,7 @@ import (
 	"math/rand"
 	"reflect"
 	"strconv"
+	"strings"
 	"sync/atomic"
 	"time"
 
@@ -559,9 +560,12 @@ func (w *worker) processMultiStagingFiles(ctx context.Context, job *payloadV2) (
 	}()
 
 	// calculate load file name prefix using md5 hash of all staging file locations
-	loadFileNamePrefix := misc.GetMD5Hash(lo.Reduce(job.StagingFiles, func(agg string, item stagingFileInfo, index int) string {
-		return agg + item.Location
-	}, ""))
+	loadFileNamePrefixBuilder := strings.Builder{}
+	loadFileNamePrefixBuilder.Grow(len(job.StagingFiles) * len(job.StagingFiles[0].Location))
+	for _, stagingFile := range job.StagingFiles {
+		loadFileNamePrefixBuilder.WriteString(stagingFile.Location)
+	}
+	loadFileNamePrefix := misc.GetMD5Hash(loadFileNamePrefixBuilder.String())
 
 	// Initialize Discards Table
 	discardsTable := job.discardsTable()
--- a/warehouse/slave/worker.go
+++ b/warehouse/slave/worker.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/rudderlabs/rudder-go-kit/jsonrs"
 
+	"github.com/rudderlabs/rudder-server/utils/misc"
 	"github.com/rudderlabs/rudder-server/warehouse/bcm"
 	"github.com/rudderlabs/rudder-server/warehouse/constraints"
 	"github.com/rudderlabs/rudder-server/warehouse/utils/types"
@@ -534,7 +535,7 @@ func (w *worker) processStagingFile(ctx context.Context, job *payload) ([]upload
 	if err := w.processSingleStagingFile(ctx, jr, &job.basePayload, stagingFileInfo{
 		ID:       job.StagingFileID,
 		Location: job.StagingFileLocation,
-	}, ""); err != nil {
+	}, misc.GetMD5Hash(job.StagingFileLocation)); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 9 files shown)


warehouse/slave/worker.go
```diff
@@ -531,7 +534,7 @@ func (w *worker) processStagingFile(ctx context.Context, job *payload) ([]upload
 	if err := w.processSingleStagingFile(ctx, jr, &job.basePayload, stagingFileInfo{
 		ID:       job.StagingFileID,
 		Location: job.StagingFileLocation,
-	}); err != nil {
+	}, ""); err != nil {
```
shekhar-rudder: We are passing an empty string here because this code path is not being used anymore with upload_v2 ?

warehouse/utils/utils.go
```diff
@@ -882,3 +883,8 @@ func GetConnectionTimeout(destType, destID string) time.Duration {
 	}
 	return config.GetDuration(warehouseLevelConfig, defaultTimeout, defaultTimeoutUnits)
 }
+
+// Md5Hash returns the MD5 hash of the input string
+func Md5Hash(input string) string {
+	return fmt.Sprintf("%x", md5.Sum([]byte(input)))
```
shekhar-rudder: A function already exists https://github.com/rudderlabs/rudder-server/blob/master/utils/misc/misc.go#L111

warehouse/slave/worker_job.go
```diff
@@ -96,16 +98,30 @@ func (p *basePayload) sortedColumnMapForAllTables() map[string][]string {
...
 	storageProvider := warehouseutils.ObjectStorageType(p.DestinationType, config, useRudderStorage)
+	objectStorageConfig := misc.GetObjectStorageConfig(misc.ObjectStorageOptsT{
+		Provider:                    storageProvider,
+		Config:                      config,
+		UseRudderStorage:            useRudderStorage,
+		RudderStoragePrefixOverride: p.RudderStoragePrefix,
+		WorkspaceID:                 p.WorkspaceID,
+	})
+
+	// create a new config map to avoid modifying the original config map
```
shekhar-rudder: What is the harm in modifying original config map ?

warehouse/slave/worker_job.go
```diff
@@ -96,16 +98,30 @@ func (p *basePayload) sortedColumnMapForAllTables() map[string][]string {
...
+	filemanagerConfig := make(map[string]interface{}, len(objectStorageConfig))
+	for k, v := range objectStorageConfig {
+		filemanagerConfig[k] = v
+	}
+
+	// If the storage provider is GCS, we need to set uploadIfNotExists to true to avoid re-uploading files
+	// GCS requires delete privileges to overwrite files, which won't be granted
+	if storageProvider == warehouseutils.GCS {
```
shekhar-rudder: I think we shouldn't set the `uploadIfNotExist` flag conditionally. The application’s intent is to enforce "uploadIfNotExist" behavior across all file managers, without worrying about how each object storage handles it internally. It should be the responsibility of the underlying library to interpret and apply this behavior as needed.

warehouse/slave/worker_job.go
```diff
@@ -364,12 +395,21 @@ func (jr *jobRun) uploadLoadFiles(ctx context.Context, modifier func(result uplo
...
 		if slices.Contains(warehouseutils.TimeWindowDestinations, jr.job.DestinationType) {
-			return uploader.Upload(
+			uploadOutput, err := uploader.Upload(
 				ctx,
 				file,
 				warehouseutils.GetTablePathInObjectStorage(jr.job.DestinationNamespace, tableName),
 				jr.job.LoadFilePrefix,
 			)
+
+			// upload output will be empty in case of precondition failure
```
shekhar-rudder, line 401: What changed in this PR? Why is `ErrPreConditionFailed` check required ?

warehouse/slave/worker_job.go
```diff
@@ -317,14 +333,29 @@ func (jr *jobRun) path(stagingFileInfo stagingFileInfo) (string, error) {
...
 	if err != nil {
 		return "", fmt.Errorf("getting staging file path: %w", err)
 	}
 
+	stagingFilePathWithoutExt := strings.TrimSuffix(stagingFilePath, ".json.gz")
+
+	if jr.conf.GetBool("Warehouse.useDeterministicLoadFileName", false) && slices.Contains(warehouseutils.TimeWindowDestinations, jr.job.DestinationType) && len(loadFileNamePrefix) > 0 {
+		loadFileName := fmt.Sprintf("%s.%s.%s", loadFileNamePrefix, jr.job.SourceID, warehouseutils.GetLoadFileFormat(jr.job.LoadFileType))
+		// adding uuid to the load file path to ensure that the load file is unique
+		// Even if same batch is processed by same worker, load file path will be unique but file name is constant
```
shekhar-rudder, line 344: At the end of the job, there is a call to cleanup which deletes all the local staging and load files. I think we can rely on that and need not create unique directories.

warehouse/slave/worker_job.go
```diff
@@ -317,14 +333,29 @@ func (jr *jobRun) path(stagingFileInfo stagingFileInfo) (string, error) {
...
+// If Warehouse.useDeterministicLoadFileName is true, the load file path will be unique but the file name will be constant.
+func (jr *jobRun) loadFilePath(stagingFileInfo stagingFileInfo, loadFileNamePrefix string) (string, error) {
 	stagingFilePath, err := jr.path(stagingFileInfo)
 	if err != nil {
 		return "", fmt.Errorf("getting staging file path: %w", err)
 	}
 
+	stagingFilePathWithoutExt := strings.TrimSuffix(stagingFilePath, ".json.gz")
+
+	if jr.conf.GetBool("Warehouse.useDeterministicLoadFileName", false) && slices.Contains(warehouseutils.TimeWindowDestinations, jr.job.DestinationType) && len(loadFileNamePrefix) > 0 {
```
shekhar-rudder, line 341: Why not use deterministic naming for all warehouses ?

warehouse/slave/worker_job.go
vyeshwanth: original one is being passed as arg from the caller, since we are processing multiple files, modifying the original will result in concurrent map access and then panic

warehouse/slave/worker_job.go
vyeshwanth: https://github.com/rudderlabs/rudder-go-kit/commit/be2aff1a9c57721f0a33e60d74f7c280155f3825

filemanager returns ErrorPreConditionFailed if the object exists and `uploadIfNotExists` is set

warehouse/slave/worker_job.go
vyeshwanth: we need unique path, Let's say master restarts. Notifier queue is cleared and all the batches are pushed to queue again. If the load file path within the container is not unique, we will have corrupted files as we use append mode when opening files. End goal is we want unique path but file name should be constant. Hence, creating unique directories

```
	file, err := os.OpenFile(s, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0o660)
```

warehouse/slave/worker_job.go
vyeshwanth: That's problematic because as you can see, when using deterministic file naming, file manager might not upload the file if object already exists and upload result will have empty location. This is fine for datalakes but for other warehouses we need the location later to copy the loadfile into the warehouse table. If we miss the location, we will miss loading data into warehouse

warehouse/slave/worker_job.go
vyeshwanth: Makes sense. we don't have to set this conditionally

warehouse/slave/worker.go
vyeshwanth: I think we can pass the hash.

warehouse/slave/worker_job.go
shekhar-rudder: > file manager might not upload the file if object already exists and upload result will have empty location

Once the file name becomes deterministic, we won't have to rely on the output of the upload call. We can populate the file path using the deterministic name.

warehouse/slave/worker_job.go
shekhar-rudder: Got it. Instead of `O_APPEND` we can use `O_TRUNC` to solve this. But it still will be an issue if the same slave picks up the job again and starts processing it before the 1st attempt has completed.

warehouse/slave/worker_job.go
shekhar-rudder: Instead of copying `filemanagerConfig`, can we copy only `config` ?

warehouse/slave/worker_job.go
vyeshwanth: Exactly. `o_trunc` won't work as long as there is a chance for slave to picking it up again

warehouse/slave/worker_job.go
vyeshwanth: There is bit of implementation detail from the filemanager depending on the object storage config. We need to bring these details into the warehouse, which is unnecessary. 
```
func (m *GcsManager) objectURL(objAttrs *storage.ObjectAttrs) string {
	if m.config.EndPoint != nil && *m.config.EndPoint != "" {
		endpoint := strings.TrimSuffix(*m.config.EndPoint, "/")
		return fmt.Sprintf("%s/%s/%s", endpoint, objAttrs.Bucket, objAttrs.Name)
	}
	return fmt.Sprintf("https://storage.googleapis.com/%s/%s", objAttrs.Bucket, objAttrs.Name)
}
```

Question is why do we want to do this for other warehouses?

warehouse/slave/worker_job.go
vyeshwanth: We can, but it wouldn't be different. Any reason you see for copying config instead of filemanagerConfig?

warehouse/slave/worker_job.go
shekhar-rudder: > Question is why do we want to do this for other warehouses?

To make the code simpler (no if checks).

warehouse/slave/worker_job.go
shekhar-rudder: - minor: Copying `config` would lead to copying of fewer fields.
- major: Copying `config` makes the intent more clearer since it is a method argument. `filemanagerConfig` is being created inside the function, so it is bit harder to understand

vyeshwanth: done. cloning config now

warehouse/internal/loadfiles/mock_loadfile_repo_test.go
```diff
@@ -3,16 +3,20 @@ package loadfiles_test
...
 	"slices"
+	"sync"
 
 	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
 )
 
 type mockLoadFilesRepo struct {
 	id    int64
 	store []model.LoadFile
+	mu    sync.Mutex
```
vyeshwanth, line 14: Added this for the race 

```
Read at 0x00c000720ea0 by goroutine 120:
  github.com/rudderlabs/rudder-server/warehouse/internal/loadfiles_test.(*mockLoadFilesRepo).Insert()
      /home/runner/work/rudder-server/rudder-server/warehouse/internal/loadfiles/mock_loadfile_repo_test.go:17 +0xe8
  github.com/rudderlabs/rudder-server/warehouse/internal/loadfiles.(*LoadFileGenerator).processNotifierResponse()
      /home/runner/work/rudder-server/rudder-server/warehouse/internal/loadfiles/loadfiles.go:406 +0xb30
  github.com/rudderlabs/rudder-server/warehouse/internal/loadfiles.(*LoadFileGenerator).createUploadJobs.func1()
      /home/runner/work/rudder-server/rudder-server/warehouse/internal/loadfiles/loadfiles.go:483 +0xb4
  golang.org/x/sync/errgroup.(*Group).add.func1()
      /home/runner/go/pkg/mod/golang.org/x/sync@v0.15.0/errgroup/errgroup.go:128 +0x141

Previous write at 0x00c000720ea0 by goroutine 119:
  github.com/rudderlabs/rudder-server/warehouse/internal/loadfiles_test.(*mockLoadFilesRepo).Insert()
      /home/runner/work/rudder-server/rudder-server/warehouse/internal/loadfiles/mock_loadfile_repo_test.go:18 +0x119
  github.com/rudderlabs/rudder-server/warehouse/internal/loadfiles.(*LoadFileGenerator).processNotifierResponseV2()
      /home/runner/work/rudder-server/rudder-server/warehouse/internal/loadfiles/loadfiles.go:447 +0x805
  github.com/rudderlabs/rudder-server/warehouse/internal/loadfiles.(*LoadFileGenerator).createUploadV2Jobs.func1()
      /home/runner/work/rudder-server/rudder-server/warehouse/internal/loadfiles/loadfiles.go:528 +0xb4
  golang.org/x/sync/errgroup.(*Group).add.func1()
      /home/runner/go/pkg/mod/golang.org/x/sync@v0.15.0/errgroup/errgroup.go:128 +0x141
```

shekhar-rudder (approved):

warehouse/slave/worker.go
```diff
@@ -554,6 +558,11 @@ func (w *worker) processMultiStagingFiles(ctx context.Context, job *payloadV2) (
 		jr.cleanup()
 	}()
 
+	// calculate load file name prefix using md5 hash of all staging file locations
+	loadFileNamePrefix := misc.GetMD5Hash(lo.Reduce(job.StagingFiles, func(agg string, item stagingFileInfo, index int) string {
```
achettyiitr:
```suggestion
	loadFileNamePrefix := misc.GetMD5Hash(lo.Reduce(job.StagingFiles, func(agg string, item stagingFileInfo, _ int) string {
```

warehouse/slave/worker.go
```diff
@@ -554,6 +558,11 @@ func (w *worker) processMultiStagingFiles(ctx context.Context, job *payloadV2) (
...
+		return agg + item.Location
+	}, ""))
+
```
achettyiitr: Can we use a string builder here with bytes pre-allocated? Concatenating such a large string with locations in an append-only fashion seems slow.

https://www.pixelstech.net/article/1715394789-performance-comparison-of-string-concatenation-in-go-language

achettyiitr:
```GO
        locationsBuilder := strings.Builder{}
	locationsBuilder.Grow(len(job.StagingFiles) * len(job.StagingFiles[0].Location))
	for _, file := range job.StagingFiles {
		locationsBuilder.WriteString(file.Location)
	}
	loadFileNamePrefix := misc.GetMD5Hash(locationsBuilder.String())
```

warehouse/slave/worker_job.go
```diff
@@ -95,12 +97,22 @@ func (p *basePayload) sortedColumnMapForAllTables() map[string][]string {
 }
 
 func (p *basePayload) fileManager(config interface{}, useRudderStorage bool) (filemanager.FileManager, error) {
+	configMap := config.(map[string]interface{})
```
achettyiitr: Do we need to have a safe check here for type cast?

warehouse/slave/worker_job.go
```diff
@@ -95,12 +97,22 @@ func (p *basePayload) sortedColumnMapForAllTables() map[string][]string {
...
+
+	// create a new config map to avoid modifying the original config
+	clonedConfig := make(map[string]interface{})
+	for k, v := range configMap {
+		clonedConfig[k] = v
+	}
+
+	clonedConfig["uploadIfNotExist"] = true
```
achettyiitr: Can we use `maps.Clone` here?
```suggestion
	clonedConfig := maps.Clone(configMap)
```

warehouse/slave/worker_job.go
```diff
@@ -364,12 +391,21 @@ func (jr *jobRun) uploadLoadFiles(ctx context.Context, modifier func(result uplo
...
 				ctx,
 				file,
 				warehouseutils.GetTablePathInObjectStorage(jr.job.DestinationNamespace, tableName),
 				jr.job.LoadFilePrefix,
 			)
+
+			// upload output will be empty in case of precondition failure
+			// but this is acceptable for datalake destinations as file is not downloaded further in the process
+			if errors.Is(err, filemanager.ErrPreConditionFailed) {
+				jr.logger.Warnf("Precondition failed while uploading load file, load file name: %s, table: %s", file.Name(), tableName)
```
achettyiitr: Can we use structured logs in here. `Warnn`

warehouse/slave/worker_job.go
vyeshwanth: It's not happening in the GetObjectStorage, Both the places it's map. Not sure why config is interface{} in the first place. we can have safe cast

warehouse/slave/worker_job.go
vyeshwanth: done. using warnn

warehouse/slave/worker_job.go
vyeshwanth: done using maps.Clone

warehouse/slave/worker_job.go
vyeshwanth: using safe cast

warehouse/slave/worker.go
vyeshwanth: done using builder

achettyiitr (approved):

PR #6044, opened 2025-06-23
chore: add logging to trace double-encoded HLL value issue

itsmihir:
# Description

Occasionally, the reporting service receives a double-encoded invalid HLL value `5c78313139303766`. 

This PR adds strategic debug logging at three critical checkpoints in the HLL data lifecycle to pinpoint exactly where the double encoding occurs:

- Before Database Storage (users_reporter.go): Logs when the problematic value is detected just before storing HLL data to the database
- After Database Retrieval (tracked_users_inapp.go): Logs when the problematic value is found in data retrieved from the database
- After JSON Marshalling (tracked_users_inapp.go): Logs when the problematic value appears in the final marshalled report payload

## Linear Ticket

Resolves OBS-857

## Security

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,53 @@
 # Changelog
 
+## [1.53.1](https://github.com/rudderlabs/rudder-server/compare/v1.53.0...v1.53.1) (2025-06-26)
+
+
+### Bug Fixes
+
+* ensure hasDynamicConfig field always present in DestinationT JSON ([#6057](https://github.com/rudderlabs/rudder-server/issues/6057)) ([e328ffb](https://github.com/rudderlabs/rudder-server/commit/e328ffb37dbd352be5404e21858744bd6b20c0a1))
+* personalise aws v2 implementation fails with unmarshal error ([#6050](https://github.com/rudderlabs/rudder-server/issues/6050)) ([7b1d80d](https://github.com/rudderlabs/rudder-server/commit/7b1d80d545a75b443e8954624209ba80d01a43df))
+* statsExcludedTags causes server to shutdown immediately ([#6060](https://github.com/rudderlabs/rudder-server/issues/6060)) ([6eee5d2](https://github.com/rudderlabs/rudder-server/commit/6eee5d260c71f2e68c37c4e1ebe1957e6d9efdd7))
+
+
+### Miscellaneous
+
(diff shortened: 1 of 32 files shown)


enterprise/reporting/flusher/aggregator/tracked_users_inapp.go
```diff
@@ -55,6 +59,14 @@ func (a *TrackedUsersInAppAggregator) Aggregate(ctx context.Context, start, end
 		if err != nil {
 			return nil, fmt.Errorf("cannot scan row %w", err)
 		}
+
+		if r.UserIDHLLHex == problematicHLLValue || r.AnonymousIDHLLHex == problematicHLLValue || r.IdentifiedAnonymousIDHLLHex == problematicHLLValue {
+			a.logger.Errorn("FOUND PROBLEMATIC HLL VALUE from database",
```
mihir20:
```suggestion
			a.logger.Errorn("found problematic hlll value from database",
```

mihir20: same for others

mihir20 (approved):
FYI, Not only `5c78313139303766` is problematic, but all hll vaues starting with 5c are problematic. but IG this approach will also help in tracing.

itsmihir:
> FYI, Not only `5c78313139303766` is problematic, but all hll vaues starting with 5c are problematic. but IG this approach will also help in tracing.

Yes, also since anonymousIdHLL is always empty, we end up capturing double encoding.

enterprise/reporting/flusher/aggregator/tracked_users_inapp.go
```diff
@@ -12,21 +12,25 @@ import (
 
 	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/jsonrs"
+	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
 )
 
 const tableName = `tracked_users_reports`
 
+const problematicHLLValue = "5c78313139303766"
```
ktgowtham, line 21: it's always the same?

ktgowtham (approved):

enterprise/reporting/flusher/aggregator/tracked_users_inapp.go
itsmihir: Yes, this is double encoding of empty HLL i.e `\x11907f`, and `anonymousIdHLL` is always empty.

PR #6045, opened 2025-06-23
chore: keydb deduplication mirroring

fracasula:
# Description

Testing [keydb](https://github.com/rudderlabs/keydb/blob/alpha1/README.md) capabilities via mirroring.

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-2131/spike-on-deduplication) >

## Security

### DIFF
--- a/services/dedup/dedup.go
+++ b/services/dedup/dedup.go
@@ -3,9 +3,6 @@
 package dedup
 
 import (
-	"fmt"
-	"time"
-
 	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
@@ -24,23 +21,12 @@ func SingleKey(key string) BatchKey {
 
 // New creates a new deduplication service. The service needs to be closed after use.
 func New(conf *config.Config, stats stats.Stats, log logger.Logger) (Dedup, error) {
-	if conf.GetBool("KeyDB.Dedup.Enabled", false) {
-		return keydb.NewKeyDB(conf, stats, log)
-	}
-
 	badgerDedup := badger.NewBadgerDB(conf, stats, badger.DefaultPath())
 
-	if conf.GetBool("KeyDB.Dedup.Mirror", false) {
+	if conf.GetBool("KeyDB.Dedup.Mirror.Enabled", false) {
 		keydbDedup, err := keydb.NewKeyDB(conf, stats, log)
 		if err == nil {
-			m := &mirror{
-				Dedup:  badgerDedup,
-				mirror: keydbDedup,
-				errs:   make(chan error, 1),
-				logger: log,
-			}
-			go m.printErrs(conf)
-			return m, nil
+			return newMirror(badgerDedup, keydbDedup, conf, stats, log), nil
 		}
 		log.Errorn("Failed to create keydb dedup", obskit.Error(err))
 	}
@@ -59,55 +45,4 @@ type Dedup interface {
 
 	// Close closes the deduplication service
 	Close()
-}
-
-type mirror struct {
-	Dedup
-	mirror Dedup
-	errs   chan error
-	logger logger.Logger
-}
-
-func (m *mirror) Allowed(keys ...BatchKey) (map[BatchKey]bool, error) {
-	go func() { // fire & forget
-		if _, err := m.mirror.Allowed(keys...); err != nil {
-			select {
-			case m.errs <- fmt.Errorf("call to Allowed failed: %w", err):
-			default: // leaky bucket to avoid filling the logs if the system fails badly
-			}
-		}
-	}()
-	return m.Dedup.Allowed(keys...)
-}
-
-func (m *mirror) Commit(keys []string) error {
-	go func() { // fire & forget
-		if err := m.mirror.Commit(keys); err != nil {
-			select {
-			case m.errs <- fmt.Errorf("call to Commit failed: %w", err):
-			default: // leaky bucket to avoid filling the logs if the system fails badly
-			}
-		}
-	}()
-	return m.Dedup.Commit(keys)
-}
-
-func (m *mirror) Close() {
-	m.mirror.Close()
-	close(m.errs)
-	m.Dedup.Close()
-}
-
-func (m *mirror) printErrs(conf *config.Config) {
-	ticker := time.NewTicker(conf.GetDuration("KeyDB.Dedup.PrintErrorsInterval", 10, time.Second))
-	for range ticker.C {
-		select {
-		case err, ok := <-m.errs:
-			if !ok {
-				return
-			}
-			m.logger.Errorn("Dedup mirroring error", obskit.Error(err))
-		default:
-		}
-	}
 }
--- a/services/dedup/dedup.go
+++ b/services/dedup/dedup.go
@@ -3,9 +3,15 @@
 package dedup
 
 import (
+	"fmt"
+	"time"
+
 	"github.com/rudderlabs/rudder-go-kit/config"
+	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
+	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
 	"github.com/rudderlabs/rudder-server/services/dedup/badger"
+	"github.com/rudderlabs/rudder-server/services/dedup/keydb"
 	"github.com/rudderlabs/rudder-server/services/dedup/types"
 )
 
@@ -17,8 +23,29 @@ func SingleKey(key string) BatchKey {
 }
 
 // New creates a new deduplication service. The service needs to be closed after use.
-func New(conf *config.Config, stats stats.Stats) (Dedup, error) {
-	return badger.NewBadgerDB(conf, stats, badger.DefaultPath()), nil
+func New(conf *config.Config, stats stats.Stats, log logger.Logger) (Dedup, error) {
+	if conf.GetBool("KeyDB.Dedup.Enabled", false) {
+		return keydb.NewKeyDB(conf, stats, log)
+	}
+
+	badgerDedup := badger.NewBadgerDB(conf, stats, badger.DefaultPath())
+
+	if conf.GetBool("KeyDB.Dedup.Mirror", false) {
+		keydbDedup, err := keydb.NewKeyDB(conf, stats, log)
+		if err == nil {
+			m := &mirror{
+				Dedup:  badgerDedup,
+				mirror: keydbDedup,
+				errs:   make(chan error, 1),
+				logger: log,
+			}
+			go m.printErrs(conf)
+			return m, nil
+		}
+		log.Errorn("Failed to create keydb dedup", obskit.Error(err))
+	}
+
+	return badgerDedup, nil
 }
 
 // Dedup is the interface for deduplication service
@@ -32,4 +59,55 @@ type Dedup interface {
 
 	// Close closes the deduplication service
 	Close()
+}
+
+type mirror struct {
+	Dedup
+	mirror Dedup
+	errs   chan error
+	logger logger.Logger
+}
+
+func (m *mirror) Allowed(keys ...BatchKey) (map[BatchKey]bool, error) {
+	go func() { // fire & forget
+		if _, err := m.mirror.Allowed(keys...); err != nil {
+			select {
+			case m.errs <- fmt.Errorf("call to Allowed failed: %w", err):
+			default: // leaky bucket to avoid filling the logs if the system fails badly
+			}
+		}
+	}()
+	return m.Dedup.Allowed(keys...)
+}
+
+func (m *mirror) Commit(keys []string) error {
+	go func() { // fire & forget
+		if err := m.mirror.Commit(keys); err != nil {
+			select {
+			case m.errs <- fmt.Errorf("call to Commit failed: %w", err):
+			default: // leaky bucket to avoid filling the logs if the system fails badly
+			}
+		}
+	}()
+	return m.Dedup.Commit(keys)
+}
+
+func (m *mirror) Close() {
+	m.mirror.Close()
+	close(m.errs)
+	m.Dedup.Close()
+}
+
+func (m *mirror) printErrs(conf *config.Config) {
+	ticker := time.NewTicker(conf.GetDuration("KeyDB.Dedup.PrintErrorsInterval", 10, time.Second))
+	for range ticker.C {
+		select {
+		case err, ok := <-m.errs:
+			if !ok {
+				return
+			}
+			m.logger.Errorn("Dedup mirroring error", obskit.Error(err))
+		default:
+		}
+	}
 }
--- a/services/dedup/dedup_test.go
+++ b/services/dedup/dedup_test.go
@@ -39,97 +39,159 @@ func Test_Dedup(t *testing.T) {
 	misc.Init()
 
 	dbPath := t.TempDir()
+	conf := config.New()
+	t.Setenv("RUDDER_TMPDIR", dbPath)
 
-	for _, dedupDB := range []string{"badger", "keydb"} {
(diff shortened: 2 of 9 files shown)


services/dedup/dedup.go
```diff
@@ -17,8 +23,29 @@ func SingleKey(key string) BatchKey {
...
+	if conf.GetBool("KeyDB.Dedup.Mirror", false) {
+		keydbDedup, err := keydb.NewKeyDB(conf, stats, log)
+		if err == nil {
+			m := &mirror{
+				Dedup:  badgerDedup,
+				mirror: keydbDedup,
+				errs:   make(chan error, 1),
+				logger: log,
+			}
+			go m.printErrs(conf)
```
atzoum: Can we spawn this goroutine inside the mirror struct and make sure it is stopped during mirror's `Stop`?

services/dedup/dedup.go
```diff
@@ -33,3 +60,54 @@ type Dedup interface {
...
+
+type mirror struct {
+	Dedup
+	mirror Dedup
+	errs   chan error
+	logger logger.Logger
+}
+
+func (m *mirror) Allowed(keys ...BatchKey) (map[BatchKey]bool, error) {
+	go func() { // fire & forget
```
atzoum: It shouldn't be too hard to end up with too many goroutines getting accumulated if keydb is slow enough. 
We could use an error group along with `TryGo` so that we have a cap on the number of goroutines we can spawn & of course emit a metric if this cap is being reached. We should also wait for these goroutines to end during `Stop`.

services/dedup/dedup.go
```diff
@@ -17,8 +23,29 @@ func SingleKey(key string) BatchKey {
 }
 
 // New creates a new deduplication service. The service needs to be closed after use.
-func New(conf *config.Config, stats stats.Stats) (Dedup, error) {
-	return badger.NewBadgerDB(conf, stats, badger.DefaultPath()), nil
+func New(conf *config.Config, stats stats.Stats, log logger.Logger) (Dedup, error) {
+	if conf.GetBool("KeyDB.Dedup.Enabled", false) {
+		return keydb.NewKeyDB(conf, stats, log)
+	}
```
atzoum: keydb is not production ready, we should remove this option for the time being
```suggestion
```

services/dedup/dedup.go
```diff
@@ -33,3 +60,54 @@ type Dedup interface {
...
+			default: // leaky bucket to avoid filling the logs if the system fails badly
+			}
+		}
+	}()
+	return m.Dedup.Allowed(keys...)
+}
+
+func (m *mirror) Commit(keys []string) error {
+	go func() { // fire & forget
+		if err := m.mirror.Commit(keys); err != nil {
```
atzoum: This can easily fail, since:
1. We are committing based on keys that exist in the primary db (local badgerdb) which can be different than the keys that exist in the mirror db
2. There is no guarantee that the mirror's `Commit` will be called after the corresponding `Allowed` call has completed, which can lead to race conditions & unexpected state of the `uncommitted` set

Can we turn the `key has not been previously set` error to a sentinel one and ignore it? We can increment a counter if this sentinel error happens instead of logging an error

services/dedup/dedup.go
```diff
@@ -33,3 +60,54 @@ type Dedup interface {
...
+			case m.errs <- fmt.Errorf("call to Commit failed: %w", err):
+			default: // leaky bucket to avoid filling the logs if the system fails badly
+			}
+		}
+	}()
+	return m.Dedup.Commit(keys)
+}
+
+func (m *mirror) Close() {
+	m.mirror.Close()
```
atzoum: We also need to wait for any running goroutines to stop first

services/dedup/dedup.go
```diff
@@ -33,3 +60,54 @@ type Dedup interface {
...
+type mirror struct {
+	Dedup
+	mirror Dedup
+	errs   chan error
+	logger logger.Logger
+}
+
+func (m *mirror) Allowed(keys ...BatchKey) (map[BatchKey]bool, error) {
+	go func() { // fire & forget
+		if _, err := m.mirror.Allowed(keys...); err != nil {
```
atzoum: Don't we want to increment a metric counter in case of an error? 1 log message every 10 seconds isn't indicative enough

services/dedup/dedup.go
atzoum: Don't we want to increment a metric counter in case of an error? 1 log message every 10 seconds isn't indicative enough

services/dedup/keydb/keydb.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	db := &KeyDB{
+		client: c,
+		window: dedupWindow,
+		logger: log,
+	}
+	db.stats.getTimer = stat.NewTaggedStat("dedup_get_duration_seconds", stats.TimerType, stats.Tags{"mode": "keydb"})
```
atzoum, line 51: For our mirroring use case, we should exclude `key has not been previously set` error from our timings, since this can happen a lot and it doesn't actually perform a remote call.

Maybe our stats should contain an `error` tag as well

services/dedup/dedup.go
```diff
@@ -33,3 +60,54 @@ type Dedup interface {
 	// Close closes the deduplication service
 	Close()
 }
+
+type mirror struct {
```
atzoum: We may move this to a separate `mirror.go` file

services/dedup/keydb/keydb.go
atzoum: On second thought, I believe that for mirroring purposes we should skip the `key has not been previously set` validation in keydb. This would help with accumulating a sizeable amount of keys in keydb as well

services/dedup/dedup.go
atzoum: [This](https://github.com/rudderlabs/rudder-server/pull/6045#discussion_r2164305818) might be preferable

services/dedup/dedup.go
fracasula: Yeah makes sense although a graceful shutdown isn't really necessary for mirroring, I'd have done it for the final thing.

services/dedup/dedup_test.go
```diff
@@ -30,83 +39,97 @@ func Test_Dedup(t *testing.T) {
...
+			startKeydb(t, conf)
+			d, err = dkdb.NewKeyDB(conf, stats.NOP, logger.NOP)
 		}
-		found, err := d.Allowed(kvs...)
 		require.Nil(t, err)
-		for _, kv := range kvs {
-			require.Equal(t, true, found[kv])
+		t.Cleanup(d.Close)
+
+		t.Run(dedupDB+"/key a not present in cache and db", func(t *testing.T) {
```
atzoum: we can wrap tests with another `t.Run(dedupDB)`, which will also take care of cleaning up resources at the proper time

services/dedup/keydb/keydb.go
```diff
@@ -0,0 +1,158 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("getting keys from keydb: %w", err)
+		}
+		d.uncommittedMu.Lock()
+		defer d.uncommittedMu.Unlock()
+		for _, batchKey := range batchKeysToCheck {
+			if !seenInDB[batchKey.Key] {
+				if _, race := d.uncommitted[batchKey.Key]; !race { // if another goroutine managed to set this key, we should skip it
+					result[batchKey] = true
+					d.uncommitted[batchKey.Key] = struct{}{} // mark this key as uncommitted
```
atzoum: There is no guarantee that the key that we're setting in the `uncommitted` set here will be used during a subsequent `Commit` call, since keydb is a mirroring one & we will be committing keys from the primary local db, which will have its own set of keys.
Either we should skip the uncommitted set during mirroring altogether, or at least, we need some periodic routine that is going to clear the `uncommitted` set.

services/dedup/mirror.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+			m.logger.Errorn("Dedup mirroring error", obskit.Error(err))
+		default:
+		}
+	}
+}
+
+func (m *mirror) Close() {
+	m.mirror.Close()
+	close(m.errs)
```
atzoum, line 144: If we close the channel before waiting for goroutines (all other than the `printErrs` one), then it is possible that one of them might panic, while trying to send an error to a closed channel

services/dedup/dedup_test.go
fracasula: I decoupled the keydb tests altogether. Since it doesn't have the uncommitted logic anymore I found it cleaner to have its own. The setup was also different.

services/dedup/keydb/keydb.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	return result, nil
+}
+
+func (d *Dedup) Commit(keys []string) error {
+	kvs := make([]types.BatchKey, len(keys))
+
+	for i, key := range keys {
+		// TODO re-add uncommitted error here if we end up using keydb in production without mirroring
```
atzoum: we would need to rewrite most of it anyway
```suggestion
```

services/dedup/mirror.go
```diff
@@ -0,0 +1,142 @@
...
+		m.metrics.maxRoutinesCount.Increment()
+		m.logLeakyErr(fmt.Errorf("max routines limit reached"))
+	}
+
+	return m.Dedup.Allowed(keys...)
+}
+
+func (m *mirror) Commit(keys []string) error {
+	select {
+	case m.waitGroupSemaphore <- struct{}{}:
```
atzoum: When do we remove it? Shall we use a `kitsync.ErrGroup` instead with `SetLimit`?

services/dedup/mirror.go
```diff
@@ -0,0 +1,146 @@
...
+	dedupMirror := &mirror{
+		Dedup:         d,
+		mirror:        m,
+		group:         group,
+		groupLimit:    groupLimit, // storing limit for enriching error
+		errs:          make(chan error, 1),
+		stopPrintErrs: make(chan struct{}),
+		logger:        log,
+	}
+	dedupMirror.metrics.allowedErrorsCount = s.NewTaggedStat("dedup_err_count", stats.CountType, stats.Tags{
```
atzoum:
```suggestion
	dedupMirror.metrics.allowedErrorsCount = s.NewTaggedStat("dedup_mirroring_err_count", stats.CountType, stats.Tags{
```

services/dedup/mirror.go
```diff
@@ -0,0 +1,146 @@
...
+		"mode": "keydb",
+		"type": "allowed",
+	})
+	dedupMirror.metrics.commitErrorsCount = s.NewTaggedStat("dedup_err_count", stats.CountType, stats.Tags{
```
atzoum:
```suggestion
	dedupMirror.metrics.commitErrorsCount = s.NewTaggedStat("dedup_mirroring_err_count", stats.CountType, stats.Tags{
```

services/dedup/mirror.go
```diff
@@ -0,0 +1,146 @@
...
+		"mode": "keydb",
+		"type": "commit",
+	})
+	dedupMirror.metrics.maxRoutinesCount = s.NewTaggedStat("dedup_max_routines_count", stats.CountType, stats.Tags{
```
atzoum:
```suggestion
	dedupMirror.metrics.maxRoutinesCount = s.NewTaggedStat("dedup_mirroring_max_routines_count", stats.CountType, stats.Tags{
```

services/dedup/mirror.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+}
+
+func (m *mirror) Close() {
+	close(m.stopPrintErrs)
+	m.mirror.Close()
+	m.Dedup.Close()
+	_ = m.group.Wait()
+	close(m.errs)
```
atzoum:
```suggestion
	// first we need to stop all mirroring goroutines
	close(m.stopPrintErrs)
	_ = m.group.Wait()
	close(m.errs)
	// then can close both mirror and primary dedup services
	m.mirror.Close()
	m.Dedup.Close()
	
	
```

services/dedup/keydb/keydb.go
```diff
@@ -0,0 +1,129 @@
...
+	nodeAddresses := conf.GetString("KeyDB.Dedup.Addresses", "")
+	if len(nodeAddresses) == 0 {
+		return nil, fmt.Errorf("keydb dedup: no node addresses provided")
+	}
+	c, err := client.NewClient(client.Config{
+		Addresses:       strings.Split(nodeAddresses, ","),
+		TotalHashRanges: uint32(conf.GetInt("KeyDB.Dedup.TotalHashRanges", 128)),
+		// TODO the client should support exponential backoff and circuit breakers
+		RetryCount: conf.GetInt("KeyDB.Dedup.RetryCount", 60),
+		RetryDelay: conf.GetDuration("KeyDB.Dedup.RetryDelay", 1, time.Second),
```
cisse21, line 40: optional suggestion
```suggestion
	nodeAddresses := conf.GetStringSliceVar([]string{}, "KeyDB.Dedup.Addresses")
	if len(nodeAddresses) == 0 {
		return nil, fmt.Errorf("keydb dedup: no node addresses provided")
	}
	c, err := client.NewClient(client.Config{
		Addresses:       nodeAddresses,
		TotalHashRanges: uint32(conf.GetInt("KeyDB.Dedup.TotalHashRanges", 128)),
		// TODO the client should support exponential backoff and circuit breakers
		RetryCount: conf.GetInt("KeyDB.Dedup.RetryCount", 60),
		RetryDelay: conf.GetDuration("KeyDB.Dedup.RetryDelay", 1, time.Second),
	}, log.Child("keydb"))
```

services/dedup/keydb/keydb.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	return result, nil
+}
+
+func (d *Dedup) Commit(keys []string) error {
+	kvs := make([]types.BatchKey, len(keys))
+	for i, key := range keys {
+		kvs[i] = types.BatchKey{Key: key}
+	}
```
cisse21, line 122:
```suggestion
```
looks unused

services/dedup/keydb/keydb.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// figure out which keys need to be checked against the DB
+	batchKeysToCheck := make([]types.BatchKey, 0, len(batchKeys)) // keys to check in the DB
+	for _, batchKey := range batchKeys {
+		// if the key is already seen in the batch, skip it
+		if _, seen := seenInBatch[batchKey.Key]; seen {
+			continue
+		}
+
+		seenInBatch[batchKey.Key] = struct{}{}
```
cisse21, line 99:
```suggestion
	seenInBatch := make(map[string]bool, len(batchKeys)) // keys already seen in the batch while iterating

	// figure out which keys need to be checked against the DB
	batchKeysToCheck := make([]types.BatchKey, 0, len(batchKeys)) // keys to check in the DB
	for _, batchKey := range batchKeys {
		// if the key is already seen in the batch, skip it
		if seenInBatch[batchKey.Key] {
			continue
		}

		seenInBatch[batchKey.Key] = true
```

cisse21: [nit-pick]

fracasula:
Closing this one to do another against 1.53.x

services/dedup/keydb/keydb.go
fracasula: I had that to begin with and converted it as I had issues with the conversion and couldn't figure out the reason why. Given how simple was to get around the problem I simply did the `strings.Split`, especially after I saw how much code `GetStringSliceVar` has under the hood :sweat_smile:

services/dedup/keydb/keydb.go
fracasula: @cisse21 your approach will make the `t.Run("non-unique keys")` test fail. I'm trying to have it behave as close as possible as the current implementation.

PR #6052, opened 2025-06-25
chore: drop bot events in processor after capturing metrics

itsmihir:
# Description

This PR moves bot event dropping from the ingestion service to the processor to capture accurate reporting metrics before deduplication occurs.

### Background
Previously, bot events were dropped at the ingestion service level. The main issue was that there is currently no way to emit reporting metrics from ingestion, so we were using Prometheus metrics to power the UI which are not accurate. Now we are using reporting metrics in the processor to capture accurate bot event statistics.

### Changes made
- Drop bot events in the processor when botAction is set to "drop" in message parameters
- Block live events recording in gateway for bot events that will be dropped in the processor 
- Capture accurate bot event metrics before the deduplication stage

#### Additional changes
- Move event blocking logic to occur before dedup processing.
- Cleanup `botAction` empty check before enriching them.

## Linear Ticket

Resolves OBS-858

## Security

### DIFF
--- a/gateway/handle_test.go
+++ b/gateway/handle_test.go
@@ -907,44 +907,6 @@ func TestExtractJobsFromInternalBatchPayload_LiveEventRecording(t *testing.T) {
 			description:               "Complex scenario: blocked events and bot drop events skip recording, others don't",
 		},
 		{
-			name: "bot events with invalid browser and drop action should skip live event recording",
-			messages: []stream.Message{
-				{
-					Properties: stream.MessageProperties{
-						RequestType:         "track",
-						RoutingKey:          "routing-key-1",
-						WorkspaceID:         "workspace1",
-						SourceID:            "source-id-1",
-						ReceivedAt:          time.Now(),
-						RequestIP:           "1.1.1.1",
-						IsBot:               true,
-						BotIsInvalidBrowser: true,
-						BotAction:           "drop",
-					},
-					Payload: json.RawMessage(`{"type":"track","event":"PageView","messageId":"msg-1","userId":"user1"}`),
-				},
-				{
-					Properties: stream.MessageProperties{
-						RequestType:         "track",
-						RoutingKey:          "routing-key-2",
-						WorkspaceID:         "workspace1",
-						SourceID:            "source-id-1",
-						ReceivedAt:          time.Now(),
-						RequestIP:           "1.1.1.1",
-						IsBot:               true,
-						BotIsInvalidBrowser: true,
-						BotAction:           "flag",
-					},
-					Payload: json.RawMessage(`{"type":"track","event":"Purchase","messageId":"msg-2","userId":"user1"}`),
-				},
-			},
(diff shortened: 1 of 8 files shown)


processor/processor.go
```diff
@@ -1884,35 +1888,74 @@ func (proc *Handle) preprocessStage(partition string, subJobs subJob) (*preTrans
...
+					transformerEvent,
+					status,
+					reportingtypes.BOT_MANAGEMENT,
+					func() json.RawMessage {
+						return nil
+					},
+					nil,
+				)
+
+				transformerEvent.StatusCode = 0 // reset status code to 0
```
ktgowtham: Need for reset?

itsmihir: Gateway reports currently have statusCode = 0. We’re reusing transformerEvent for them and I didn’t want to modify other metrics. I’ll add a comment for clarity.

gateway/handle_test.go
```diff
@@ -685,3 +685,325 @@ func TestExtractJobsFromInternalBatchPayload_EventBlocking(t *testing.T) {
...
+				},
+			},
+			eventBlockingSettings: backendconfig.EventBlocking{
+				Events: map[string][]string{},
+			},
+			expectedSkipLiveEventRecs: []bool{true, false},
+			description:               "Bot events with invalid browser - only 'drop' action should skip live event recording",
+		},
+		{
+			name: "blocked bot events with drop action should skip live event recording",
```
mihir20: isn't this same as line 770?

processor/processor.go
```diff
@@ -2139,8 +2180,22 @@ func (proc *Handle) pretransformStage(partition string, preTrans *preTransformat
 		return nil, err
 	}
 
-	// REPORTING - GATEWAY metrics - START
 	if proc.isReportingEnabled() {
+
+		// REPORTING - BOT_MANAGEMENT metrics - START
+		reportingtypes.AssertSameKeys(preTrans.botManagementConnectionDetailsMap, preTrans.botManagementStatusDetailsMap)
+		for k, cd := range preTrans.botManagementConnectionDetailsMap {
```
mihir20, line 2187: should we do this in preprocess stage?

mihir20 (commented):
should we add some tests checking bot events are dropped and reporting metrics are being collected as expected?

processor/processor.go
itsmihir: Yeah, we can do this in `preprocess` also, but I want to keep things consistent with how we capture the reporting metrics.

gateway/handle_test.go
itsmihir: Yeah, the test name is same, but one has event blocking setting other doesn't. will update the test name

itsmihir:
> should we add some tests checking bot events are dropped and reporting metrics are being collected as expected?

Yeah I am planning to add some tests for event dropped due to bot and event blocking in a separate PR.

mihir20 (approved):

PR #6059, opened 2025-06-26
feat: implement leaky uploader for invalid payloads

mihir20:
# Description

- Add leaky uploader functionality to handle invalid JSON payloads
- Upload invalid payloads to s3 for debugging and analysis
- Ensure valid payloads are processed normally without being uploaded
- Configure leaky uploader with S3 storage options
- Implement backoff and buffer mechanisms for failed uploads

## Linear Ticket

pipe-2188

## Security

### DIFF
--- a/gateway/gateway_test.go
+++ b/gateway/gateway_test.go
@@ -2756,8 +2756,10 @@ func TestLeakyUploader(t *testing.T) {
 	// Setup Minio test container
 	pool, err := dockertest.NewPool("")
 	require.NoError(t, err)
+	pool.MaxWait = time.Minute
 
 	t.Run("invalid JSON payload should be uploaded to Minio", func(t *testing.T) {
+		t.Parallel()
 		minioContainer, err := minio.Setup(pool, t)
 		require.NoError(t, err)
 
@@ -2779,12 +2781,12 @@ func TestLeakyUploader(t *testing.T) {
 		// verify file was uploaded to Minio
 		require.Eventually(t, func() bool {
 			contents, err := minioContainer.Contents(context.Background(), "gw-failed-events")
-			require.NoError(t, err)
-			return len(contents) != 0
-		}, 2*time.Second, time.Second, "Minio bucket not updated within timeout")
+			return err == nil && len(contents) != 0
+		}, 10*time.Second, time.Second, "Minio bucket not updated within timeout")
 	})
 
 	t.Run("valid JSON payload should be processed normally", func(t *testing.T) {
+		t.Parallel()
 		// Prepare valid payload
 		validPayload := []byte(`[
 			{
@@ -2816,12 +2818,10 @@ func TestLeakyUploader(t *testing.T) {
 		require.Len(t, jobs, 1)
 
 		// Verify no files uploaded to Minio
-		// verify file was uploaded to Minio
 		require.Never(t, func() bool {
 			contents, err := minioContainer.Contents(context.Background(), "gw-failed-events")
-			require.NoError(t, err)
-			return len(contents) != 0
-		}, 2*time.Second, time.Second, "Minio bucket not updated within timeout")
+			return len(contents) != 0 && err == nil
+		}, 5*time.Second, time.Second, "Minio bucket not updated within timeout")
 	})
 }
 
@@ -2851,15 +2851,15 @@ func createTestGatewayWithLeakyUploader(t *testing.T, endpoint, accessKeyID, sec
 	mockApp.EXPECT().Features().Return(enterpriseFeatures).AnyTimes()
 	gw = &Handle{}
 	conf := config.New()
-	conf.Set("gateway.leakyUploader.enabled", true)
-	conf.Set("leakyUploader.Storage.Endpoint", endpoint)
-	conf.Set("leakyUploader.Storage.AccessKeyId", accessKeyID)
-	conf.Set("leakyUploader.Storage.AccessKey", secretKey)
-	conf.Set("leakyUploader.Storage.Bucket", "rudder-saas")
-	conf.Set("leakyUploader.Timeout", "2s")
-	conf.Set("leakyUploader.Storage.DisableSsl", true)
-	conf.Set("leakyUploader.Storage.UseGlue", true)
-	conf.Set("leakyUploader.Storage.S3ForcePathStyle", true)
+	conf.Set("Gateway.leakyUploader.enabled", true)
+	conf.Set("Gateway.leakyUploader.Storage.Endpoint", endpoint)
+	conf.Set("Gateway.leakyUploader.Storage.AccessKeyId", accessKeyID)
+	conf.Set("Gateway.leakyUploader.Storage.AccessKey", secretKey)
+	conf.Set("Gateway.leakyUploader.Storage.Bucket", "rudder-saas")
+	conf.Set("Gateway.leakyUploader.Timeout", "2s")
+	conf.Set("Gateway.leakyUploader.Storage.DisableSsl", true)
+	conf.Set("Gateway.leakyUploader.Storage.UseGlue", true)
+	conf.Set("Gateway.leakyUploader.Storage.S3ForcePathStyle", true)
 	err := gw.Setup(context.Background(), conf, logger.NewLogger().With("component", "test"), stats.NOP, mockApp, mockBackendConfig, mockJobsDB, mockErrJobsDB, mockRateLimiter, mockVersionHandler, rsources.NewNoOpService(), transformer.NewNoOpService(), sourcedebugger.NewNoOpService(), nil)
 	require.NoError(t, err)
 	require.Eventually(t, func() bool {
--- a/gateway/gateway_test.go
+++ b/gateway/gateway_test.go
@@ -16,6 +16,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ory/dockertest/v3"
+
+	"github.com/rudderlabs/rudder-go-kit/testhelper/docker/resource/minio"
+
 	gwtypes "github.com/rudderlabs/rudder-server/gateway/types"
 
 	"github.com/rudderlabs/rudder-go-kit/jsonrs"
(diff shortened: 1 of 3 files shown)


gateway/gateway_test.go
```diff
@@ -2747,3 +2751,127 @@ func waitForBackendConfigInit(gw *Handle) {
...
+		// Call internal batch handler
+		jobs, err := gw.extractJobsFromInternalBatchPayload("batch", validPayload)
+
+		// Verify successful processing
+		require.NoError(t, err)
+		require.NotNil(t, jobs)
+		require.Len(t, jobs, 1)
+
+		// Verify no files uploaded to Minio
+		// verify file was uploaded to Minio
```
fracasula: The comments above are a bit misleading =)

gateway/gateway_test.go
```diff
@@ -2747,3 +2751,127 @@ func waitForBackendConfigInit(gw *Handle) {
 		2*time.Second,
 	).Should(BeTrue())
 }
+
+func TestLeakyUploader(t *testing.T) {
+	// Setup Minio test container
+	pool, err := dockertest.NewPool("")
+	require.NoError(t, err)
```
fracasula, line 2758: If after the require you do `pool.MaxWait = time.Minute` then you could add `t.Parallel()` to the 2 tests below.

gateway/gateway_test.go
```diff
@@ -2747,3 +2751,127 @@ func waitForBackendConfigInit(gw *Handle) {
...
+
+		// Verify error response
+		require.Error(t, err)
+		require.Equal(t, response.InvalidJSON, err.Error())
+		require.Nil(t, jobs)
+
+		// verify file was uploaded to Minio
+		require.Eventually(t, func() bool {
+			contents, err := minioContainer.Contents(context.Background(), "gw-failed-events")
+			require.NoError(t, err)
```
fracasula: No require should be inside `Eventually`. Otherwise it might fail at the first attempt instead of retrying for 2s like you specified below.

gateway/gateway_test.go
```diff
@@ -2747,3 +2751,127 @@ func waitForBackendConfigInit(gw *Handle) {
...
+		jobs, err := gw.extractJobsFromInternalBatchPayload("batch", validPayload)
+
+		// Verify successful processing
+		require.NoError(t, err)
+		require.NotNil(t, jobs)
+		require.Len(t, jobs, 1)
+
+		// Verify no files uploaded to Minio
+		// verify file was uploaded to Minio
+		require.Never(t, func() bool {
```
fracasula, line 2821: This could give a false positive like for example if a delay is introduced, then you'd be happy after 2s but the file gets eventually uploaded.
Can we rely on something else here? For example, do we emit a metric after the upload? If yes we could assert that that metric is emitted and only then check the bucket for files. It would make the test more robust.

gateway/handle_lifecycle.go
```diff
@@ -48,6 +53,11 @@ import (
 	"github.com/rudderlabs/rudder-server/utils/timeutil"
 )
 
+type leakyUpload struct {
```
fracasula: We might be able to come up with a better name. Wdyt of `msgToUpload`?

gateway/handle_lifecycle.go
```diff
@@ -203,9 +213,112 @@ func (gw *Handle) Setup(
...
+		} else {
+			gw.logger.Errorn("failed to create leaky uploader file manager", obskit.Error(err))
+		}
+	}
 	return nil
 }
 
+func getLeakyUploaderFileManager(conf *config.Config, log logger.Logger) (filemanager.FileManager, error) {
+	var (
+		regionHint       = conf.GetStringVar("us-east-1", "leakyUploader.Storage.RegionHint", "AWS_S3_REGION_HINT")
```
fracasula: Can we be more specific and add perhaps `Gateway` to these configuration variables?
Like `Gateway.LeakyUploader.Storage.RegionHint`.

gateway/handle_lifecycle.go
```diff
@@ -203,9 +213,112 @@ func (gw *Handle) Setup(
...
+			g.Go(crash.Wrapper(func() error {
+				leakyUploader(ctx, gw.config, gw.logger, leakyUploaderDone, leakyUploaderBuffer, fm)
+				defer func() {
+					<-leakyUploaderDone
+					close(leakyUploaderBuffer)
+				}()
+				return nil
+			}))
+		} else {
+			gw.logger.Errorn("failed to create leaky uploader file manager", obskit.Error(err))
```
fracasula:
```suggestion
			gw.logger.Errorn("failed to create leaky uploader in gateway", obskit.Error(err))
```

gateway/handle_lifecycle.go
```diff
@@ -203,9 +213,112 @@ func (gw *Handle) Setup(
...
+	backoff := conf.GetDuration("gateway.leakyUploader.backoff", 1, time.Second)
+	instanceName := conf.GetString("INSTANCE_ID", "unknown-instance")
+	log.Infon("starting leaky payload uploader", logger.NewStringField("instanceName", instanceName))
+	defer close(done)
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case upload := <-uploads:
+			file, err := os.CreateTemp("", "payload-dump*.txt")
```
ktgowtham: fm now exposes `UploadReader`. let's use that and avoid using local files.

gateway/handle_lifecycle.go
```diff
@@ -203,9 +213,112 @@ func (gw *Handle) Setup(
...
+			gw.leakyUploader = func(upload leakyUpload) {
+				select {
+				case leakyUploaderBuffer <- leakyUpload{payload: upload.payload, fields: upload.fields}:
+				default:
+					// drop the payload if the channel is full
+				}
+			}
+			g.Go(crash.Wrapper(func() error {
+				leakyUploader(ctx, gw.config, gw.logger, leakyUploaderDone, leakyUploaderBuffer, fm)
+				defer func() {
```
fracasula: Why the `defer`? Aren't we past `leakyUploader` anyway?

gateway/handle_lifecycle.go
```diff
@@ -203,9 +213,112 @@ func (gw *Handle) Setup(
...
+		s3Config,
+		log.Withn(logger.NewStringField("component", "leaky-uploader")),
+		func() time.Duration {
+			return conf.GetDuration("leakyUploader.Timeout", 120, time.Second)
+		},
+	)
+}
+
+func leakyUploader(ctx context.Context, conf *config.Config, log logger.Logger, done chan struct{}, uploads <-chan leakyUpload, fm filemanager.FileManager) {
+	backoff := conf.GetDuration("gateway.leakyUploader.backoff", 1, time.Second)
```
fracasula: Can we have all settings related to this feature properly behind the same hierarchy? Like `Gateway.LeakyUploader...`

gateway/handle_lifecycle.go
```diff
@@ -203,9 +213,112 @@ func (gw *Handle) Setup(
...
+		if err == nil {
+			gw.leakyUploader = func(upload leakyUpload) {
+				select {
+				case leakyUploaderBuffer <- leakyUpload{payload: upload.payload, fields: upload.fields}:
+				default:
+					// drop the payload if the channel is full
+				}
+			}
+			g.Go(crash.Wrapper(func() error {
+				leakyUploader(ctx, gw.config, gw.logger, leakyUploaderDone, leakyUploaderBuffer, fm)
```
fracasula: We could do `go.logger.Child("leakyUploader")`. This way we could easily select all logs coming from this component.

gateway/handle_lifecycle.go
fracasula: We don't need to do this anymore. We have `UploadReader` now :muscle: 
We can skip the file creation, so no write into disks and you can upload data that you have in memory.

gateway/handle_lifecycle.go
```diff
@@ -203,9 +213,112 @@ func (gw *Handle) Setup(
...
+				defer func() {
+					<-leakyUploaderDone
+					close(leakyUploaderBuffer)
```
fracasula: I'm a bit afraid of this setup because you're closing `leakyUploaderBuffer` here and not in the `Shutdown` method of the gateway. This means that there might the possibility that someone invokes the `gw.leakyUploader` function after this channel is closed, leading to a panic.
Can we make sure the GW is in no state to accept requests and that all in-flight are complete before closing this channel?

fracasula (changes requested):

gateway/handle.go
```diff
@@ -779,14 +782,24 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(reqType string, body []byt
 		isUserSuppressed = gw.memoizedIsUserSuppressed()
 		res              []jobWithStat
 		stat             = gwstats.SourceStat{ReqType: reqType}
+		err              error
 	)
 
-	err := jsonrs.Unmarshal(body, &messages)
+	defer func() {
+		// Upload the raw payload via leaky uploader if available
+		if err != nil && gw.leakyUploader != nil {
```
fracasula: Down below at line 807 we do `if len(messages) == 0` and then return an error.
If that were the case we wouldn't upload anything because `err` would be nil, yet the src-router would be stuck, right?

gateway/handle_lifecycle.go
```diff
@@ -203,9 +231,91 @@ func (gw *Handle) Setup(
...
+	if leakyUploaderEnabled {
+		leakyUploaderDone = make(chan struct{})
+		leakyUploaderBuffer = make(chan msgToUpload, config.GetInt("Gateway.leakyUploader.bufferSize", 1))
+		fm, err := getLeakyUploaderFileManager(gw.config, gw.logger)
+		if err == nil {
+			gw.leakyUploader = func(upload msgToUpload) {
+				select {
+				case leakyUploaderBuffer <- msgToUpload{payload: upload.payload, fields: upload.fields}:
+				default:
+					// drop the payload if the channel is full
```
fracasula: Can we print a warning here? Just to be able to detect cases where somehow we can't upload, the src-router is stuck and the logs are empty.

gateway/handle_lifecycle.go
```diff
@@ -203,9 +231,91 @@ func (gw *Handle) Setup(
...
+	instanceName := conf.GetString("INSTANCE_ID", "unknown-instance")
+	log.Infon("starting leaky payload uploader", logger.NewStringField("instanceName", instanceName))
+	defer close(done)
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case upload := <-uploads:
+			fileName := path.Join("gw-failed-events", instanceName, time.Now().Format("2006-01-02"), uuid.New().String())
+			uploadedFile, err := fm.UploadReader(ctx, fileName, strings.NewReader(string(upload.payload)))
```
fracasula: Why the double allocation? Can't you just do `bytes.NewReader(upload.payload)`?

gateway/handle_lifecycle.go
```diff
@@ -203,9 +231,91 @@ func (gw *Handle) Setup(
...
+		func() time.Duration {
+			return conf.GetDuration("Gateway.leakyUploader.Timeout", 120, time.Second)
+		},
+	)
+}
+
+func leakyUploader(ctx context.Context, conf *config.Config, log logger.Logger, done chan struct{}, uploads <-chan msgToUpload, fm filemanager.FileManager) {
+	backoff := conf.GetDuration("Gateway.leakyUploader.backoff", 1, time.Second)
+	instanceName := conf.GetString("INSTANCE_ID", "unknown-instance")
+	log.Infon("starting leaky payload uploader", logger.NewStringField("instanceName", instanceName))
```
fracasula: [optional] You could enrich the logger with `Withn` so that all logs from this function will include the `instanceName`. Although adding the instance name is probably not necessary since we'll have anyway. I suggest you double check and if that's the case then it can be removed altogether.

gateway/handle.go
mihir20: yes, I will add an handling for this where we publish the payload in case of 0 messages also

gateway/handle.go
```diff
@@ -779,14 +782,27 @@ func (gw *Handle) extractJobsFromInternalBatchPayload(reqType string, body []byt
 		isUserSuppressed = gw.memoizedIsUserSuppressed()
 		res              []jobWithStat
 		stat             = gwstats.SourceStat{ReqType: reqType}
+		err              error
 	)
 
-	err := jsonrs.Unmarshal(body, &messages)
+	defer func() {
+		// Upload the raw payload via leaky uploader if available
+		if (err != nil && gw.leakyUploader != nil) || len(messages) == 0 {
```
fracasula: So if there are no messages and `leakyUploader` is `nil` you try to upload, correct?

fracasula (changes requested):

gateway/handle.go
mihir20: fixed in the the latest push

PR #6060, opened 2025-06-26
fix: statsExcludedTags causes server to shutdown immediately

atzoum:
# Description

Ignoring `statsExcludedTags` config change during non-reloadable config restarts, since under some conditions it can be set after server starts, which will cause an unwanted immediate shutdown of the server.

## Security

### DIFF
--- a/main.go
+++ b/main.go
@@ -39,11 +39,16 @@ func main() {
 
 	shutdownOnNonReloadableConfigChange := c.GetReloadableBoolVar(false, "shutdownOnNonReloadableConfigChange")
 	c.OnNonReloadableConfigChange(func(key string) {
-		if shutdownOnNonReloadableConfigChange.Load() {
-			log.Infon("Config change detected, shutting down server...", logger.NewStringField("key", key))
-			cancel()
-		} else {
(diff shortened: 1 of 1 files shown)


cisse21 (approved):

main.go
```diff
@@ -39,11 +39,16 @@ func main() {
...
 	shutdownOnNonReloadableConfigChange := c.GetReloadableBoolVar(false, "shutdownOnNonReloadableConfigChange")
 	c.OnNonReloadableConfigChange(func(key string) {
-		if shutdownOnNonReloadableConfigChange.Load() {
-			log.Infon("Config change detected, shutting down server...", logger.NewStringField("key", key))
-			cancel()
-		} else {
-			log.Infon("Config change detected, but server will not shut down", logger.NewStringField("key", key))
+		switch key {
+		case "statsExcludedTags": // keys to ignore
+			// no-op
```
mihir20, line 44: Should we make this configurable?

atzoum: Nope, we're not using `config.Set` for anything else other than `statsExcludedTags` (for now). I would much rather turning off the feature until we add a missing config key here, rather than having to maintain yet-another config option in our customer objects.

PR #6064, opened 2025-06-27
fix: custom type in payload during enrichment

achettyiitr:
# Description

- We rely on being `SingularEvent` only to contain `map[string]any`, but some context properties are being directly added to the SingularEvent as the struct type like these:
  - [Bot Details](https://github.com/rudderlabs/rudder-server/blob/master/internal/enricher/bot.go#L49-L54).
  - [Geo enrichment](https://github.com/rudderlabs/rudder-server/blob/master/internal/enricher/geolocation.go#L134-L135)
  - [Tracking plan.](https://github.com/rudderlabs/rudder-server/blob/master/processor/trackingplan.go#L32-L50)
- In case of Bot Details, we don't have access to the struct, so we can't infer the type and handle it. Adding support for any custom struct which checks for `reflect.Struct`, `reflect.Ptr`, `reflect.Slice`, `reflect.Array` using reflection.

## Linear Ticket

- Resolves WAR-821

## Security

### DIFF
--- a/processor/internal/transformer/destination_transformer/embedded/warehouse/datatype.go
+++ b/processor/internal/transformer/destination_transformer/embedded/warehouse/datatype.go
@@ -23,7 +23,7 @@ func dataTypeFor(destType, key string, val any, isJSONKey bool) string {
 
 func primitiveType(val any) string {
 	switch v := val.(type) {
-	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64:
+	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64, uintptr:
 		return model.IntDataType
 	case float64:
 		return getFloatType(v)
--- a/processor/internal/transformer/destination_transformer/embedded/warehouse/datatype.go
+++ b/processor/internal/transformer/destination_transformer/embedded/warehouse/datatype.go
@@ -1,15 +1,10 @@
 package warehouse
 
 import (
-	"fmt"
 	"math/big"
 
-	"github.com/samber/lo"
-
-	"github.com/rudderlabs/rudder-server/internal/enricher"
 	"github.com/rudderlabs/rudder-server/processor/internal/transformer/destination_transformer/embedded/warehouse/internal/model"
 	"github.com/rudderlabs/rudder-server/processor/internal/transformer/destination_transformer/embedded/warehouse/internal/utils"
-	"github.com/rudderlabs/rudder-server/processor/types"
 	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
 )
 
@@ -85,7 +80,7 @@ func overrideForRedshift(val any, isJSONKey bool) string {
 
 func shouldUseTextForRedshift(data any) bool {
 	switch v := data.(type) {
-	case []any, []types.ValidationError, map[string]any:
+	case []any, map[string]any:
 		jsonVal, _ := utils.MarshalJSON(v)
 		// Javascript strings are UTF-16 encoded, use utf16 instead of utf8 package for determining the length
 		if utils.UTF16RuneCountInString(string(jsonVal)) > redshiftStringLimit {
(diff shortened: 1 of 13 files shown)


processor/internal/transformer/destination_transformer/embedded/warehouse/internal/utils/utils.go
```diff
@@ -234,8 +270,6 @@ func IsEmptyString(value interface{}) bool {
 			return IsEmptyString(v[0])
 		}
 		return false
-	case []types.ValidationError:
-		return len(v) == 0
```
achettyiitr, line 238: No need for such handling for custom types.

processor/internal/transformer/destination_transformer/embedded/warehouse/datatype.go
```diff
@@ -85,7 +80,7 @@ func overrideForRedshift(val any, isJSONKey bool) string {
 
 func shouldUseTextForRedshift(data any) bool {
 	switch v := data.(type) {
-	case []any, []types.ValidationError, map[string]any:
```
achettyiitr, line 88: No need for such handling for custom datatypes.

processor/internal/transformer/destination_transformer/embedded/warehouse/testhelper/validate.go
```diff
@@ -19,100 +19,98 @@ type EventContext struct {
...
-	ctx := context.Background()
-
-	legacyResponse := pTransformer.Transform(ctx, events)
-	embeddedResponse := dTransformer.Transform(ctx, events)
-
-	validateResponseLengths(t, expectedResponse, legacyResponse, embeddedResponse)
-	validateRudderEventIfExists(t, expectedResponse, legacyResponse, embeddedResponse)
-	validateEventEquality(t, expectedResponse, legacyResponse, embeddedResponse)
-	validateFailedEventEquality(t, legacyResponse, embeddedResponse)
+	expectedResponse, embeddedResponse, legacyResponse = deepCopyResponse(t, expectedResponse), deepCopyResponse(t, embeddedResponse), deepCopyResponse(t, legacyResponse)
```
achettyiitr, line 17: Deep copying here to handle for Marshalling and UnMarshalling scenario as we do in Processsor.

processor/internal/transformer/destination_transformer/embedded/warehouse/transformer.go
```diff
@@ -162,7 +144,7 @@ func (t *Transformer) Transform(_ context.Context, clientEvents []types.Transfor
 		g.Go(func() error {
 			r, err := t.processWarehouseMessage(c, &event)
 			if err != nil {
-				results <- transformerResponseFromErr(&eventMetadata, err)
+				results <- transformerResponseFromErr(&eventMetadata, embeddedutils.GetValidationErrorStatTags(event.Destination), err)
```
achettyiitr, line 147: While deep copying, got this bug. We need to send stats in case of failed events.

processor/internal/transformer/destination_transformer/embedded/warehouse/datatype.go
```diff
@@ -1,15 +1,10 @@
 package warehouse
 
 import (
-	"fmt"
```
shekhar-rudder, line 4: > We rely on being SingularEvent only to contain map[string]any, but some data fields are directly being added, like say for Tracking Plan, Bot Details and Geo enrichment.

Can you add more details about what does "some data fields are directly being added" mean ?

achettyiitr: Added.

processor/internal/transformer/destination_transformer/embedded/warehouse/internal/utils/utils.go
```diff
@@ -95,6 +96,48 @@ func IsObject(val any) bool {
...
+		float32, float64,
+		complex64, complex128,
+		string,
+		[]any,
+		map[string]any:
+		return false
+	}
+	v := reflect.ValueOf(val)
+	switch v.Kind() {
+	case reflect.Struct, reflect.Ptr:
```
shekhar-rudder, line 114: Do we want this method to return true even in case of non-Struct/Slice pointers (like a string pointer) ?

processor/internal/transformer/destination_transformer/embedded/warehouse/testhelper/validate.go
```diff
@@ -1,118 +1,109 @@
...
 
-func validateRudderEventIfExists(t *testing.T, expectedResponse, legacyResponse, embeddedResponse types.Response) {
+func checkForMarshalledFieldsAndRemove(t testing.TB, expected, embedded, legacy []types.TransformerResponse) {
 	t.Helper()
 
-	for i := range legacyResponse.Events {
-		data, ok := expectedResponse.Events[i].Output["data"].(map[string]interface{})
-		if !ok {
-			continue // No data to validate
+	fields := []string{"data.rudder_event"}
```
shekhar-rudder: `fields` has only 1 element. Should we make it a string ?

achettyiitr: Made fields ...string as varadic and passing as function arguements.

vyeshwanth (approved):

processor/internal/transformer/destination_transformer/embedded/warehouse/internal/utils/utils.go
achettyiitr: Yes. Those pointers will need to get converted to actual values. So Marshalling/Unmarshalling is required. Added tests as well for pointers.

shekhar-rudder (approved):

PR #6076, opened 2025-07-02
chore: add server instance id in warehouse staging file metadata

shekhar-rudder:
# Description

https://linear.app/rudderstack/issue/WAR-825/add-instrumentation-for-tracking-rudderstack-node-in-staging-files

## Security


warehouse/internal/model/staging.go
```diff
@@ -31,6 +31,8 @@ type StagingFile struct {
 	SourceJobRunID  string
 	TimeWindow      time.Time
 
+	ServerInstanceId string
```
achettyiitr: nit: Can we make in other places as well.
```suggestion
	ServerInstanceID string
```

PR #6079, opened 2025-07-03
chore: add tests for bot and blocked events dropped in processor

itsmihir:
# Description

This PR adds integration tests to verify the processor correctly drops events based on bot detection and event blocking configurations. The tests ensure that events are properly filtered before reaching the router, preventing unwanted events from being processed downstream.
 
## Linear Ticket

< [PIPE-2199](https://linear.app/rudderstack/issue/PIPE-2199/add-tests-for-bot-and-blocked-events-dropped-in-processor)  >

## Security


processor/processor_event_dropping_test.go
```diff
@@ -0,0 +1,393 @@
...
+				NewBotEventConfig("identify", "", "drop"),
+				NewBotEventConfig("screen", "", "drop"),
+			}).
+			Run(t, func(t *testing.T, scenario *eventDropScenario) {
+				scenario.requireJobsCount(t, "gw", "succeeded", 3)
+				scenario.requireJobsCount(t, "rt", "succeeded", 0)
+				scenario.requireJobsCount(t, "rt", "aborted", 0)
+				scenario.requireJobsCount(t, "rt", "failed", 0)
+
+				scenario.requireTotalJobsCount(t, "rt", 0)
```
mihir20, line 44: for completeness we should also validate reporting metrics

satishrudderstack (approved):

processor/processor_event_dropping_test.go
```diff
@@ -0,0 +1,549 @@
...
+
+				// reporting metrics
+				scenario.requireReportsFromBotManagement(t, botManagementReportExpectations{eventType: "track", eventName: "TestEvent", botDroppedCount: 1})
+				scenario.requireReportsFromBotManagement(t, botManagementReportExpectations{eventType: "identify", eventName: "", botDroppedCount: 1})
+				scenario.requireReportsFromBotManagement(t, botManagementReportExpectations{eventType: "screen", eventName: "", botDroppedCount: 1})
```
satishrudderstack, line 49: One thing I see with these tests is that, it could be possible that reports could be flushed to Reporting Service before these validations

satishrudderstack: Need not fix it as part of this pr, but there is one more issue with these metrics. These reporting metrics are sent to the production reporting instance. I think we should fix it.

processor/processor_event_dropping_test.go
```diff
@@ -0,0 +1,550 @@
...
+	return s
+}
+
+func (s *eventDropScenario) withEventConfigs(configs []eventConfig) *eventDropScenario {
+	s.eventConfigs = configs
+	return s
+}
+
+func (s *eventDropScenario) run(t *testing.T, verification func(t *testing.T, scenario *eventDropScenario)) {
+	config.Reset()
```
fracasula: Is the `config.Reset()` really needed? 🤔 
Aren't we injecting the config properly in the processor?

itsmihir: Yeah, we don't need it, removing it.

fracasula: @itsmihir you're still using a global variable (i.e. `config`) though. Do you think you can get rid of it (if not too complex)?

itsmihir: @fracasula made changes to use `t.Setenv` instead of using global config.

PR #6095, opened 2025-07-10
feat(warehouse): add schema snapshots table and link to staging files (WAR-866)

achettyiitr:
# Description

- Introduces a new table, `wh_staging_file_schema_snapshots`, to store JSON-encoded schema snapshots for staging files, keyed by source, destination, and workspace.
- Adds two columns to **wh_staging_files**:
  - `schema_snapshot_id` (UUID, references the new snapshot table)
  - `schema_snapshot_patch` (TEXT, stores a JSON Patch diff from the snapshot to the actual schema)
- Adds supporting indexes for efficient lookup and retrieval.

# Linear Ticket

[WAR-866: Migration for snapshot table](https://linear.app/rudderstack/issue/WAR-866/migration-for-snapshot-table)

# Security


sql/migrations/warehouse/000032_add_schema_snapshots_to_staging_files.up.sql
```diff
@@ -0,0 +1,27 @@
+-- Create a table to store schema snapshots for staging files.
+-- Each snapshot represents the schema at a specific point in time for a (source, destination, workspace).
+CREATE TABLE IF NOT EXISTS wh_staging_file_schema_snapshots (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(), -- Unique identifier for the snapshot
+    schema TEXT NOT NULL,                          -- JSON-encoded schema at the time of the snapshot
```
achettyiitr, line 5: Should we keep it default as '{}'

sql/migrations/warehouse/000032_add_schema_snapshots_to_staging_files.up.sql
```diff
@@ -0,0 +1,27 @@
...
+    source_id VARCHAR(64) NOT NULL,                -- Source identifier
+    destination_id VARCHAR(64) NOT NULL,           -- Destination identifier
+    workspace_id VARCHAR(64) NOT NULL,             -- Workspace identifier
+    created_at TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT NOW() -- Snapshot creation timestamp
+);
+
+-- Add columns to wh_staging_files to link each file to a schema snapshot and store a JSON Patch diff.
+ALTER TABLE wh_staging_files
+    ADD COLUMN IF NOT EXISTS schema_snapshot_id UUID REFERENCES wh_staging_file_schema_snapshots(id) ON DELETE SET NULL, -- Link to the schema snapshot
+    ADD COLUMN IF NOT EXISTS schema_patch TEXT; -- JSON Patch (RFC 6902) diff from the snapshot to the actual schema
```
shekhar-rudder: Should we call the column `schema_snapshot_patch` instead of `schema_patch` ?

sql/migrations/warehouse/000032_add_schema_snapshots_to_staging_files.up.sql
```diff
@@ -0,0 +1,27 @@
...
+
+-- For fast lookup of staging files by schema_snapshot_id.
```
shekhar-rudder: For what use-case would we need to look up staging files using snapshot id ? Wouldn't it be the other way around ?

sql/migrations/warehouse/000032_add_schema_snapshots_to_staging_files.up.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE INDEX IF NOT EXISTS wh_staging_files_schema_snapshot_id_index
+    ON wh_staging_files(schema_snapshot_id);
+
+-- For efficient retrieval of latest snapshot by (source_id, destination_id, workspace_id, created_at DESC).
+CREATE INDEX IF NOT EXISTS wh_staging_file_schema_snapshots_source_id_destination_id_workspace_id_created_at_index
+    ON wh_staging_file_schema_snapshots(source_id, destination_id, workspace_id, created_at DESC);
+
+-- For filtering snapshots by workspace_id.
+CREATE INDEX IF NOT EXISTS wh_staging_file_schema_snapshots_workspace_id_index
```
shekhar-rudder, line 22: Same here. In which case would we filter snapshots by workspace_id ? I would expect the filtering to be done using (source_id, destination_id)

sql/migrations/warehouse/000032_add_schema_snapshots_to_staging_files.up.sql
shekhar-rudder: `schema` will always be non-empty, so I don't think we should default to `{}`

sql/migrations/warehouse/000032_add_schema_snapshots_to_staging_files.up.sql
achettyiitr: It's for migration.

shekhar-rudder: Lets add it as a comment ?

sql/migrations/warehouse/000032_add_schema_snapshots_to_staging_files.up.sql
achettyiitr: This is for archiving, basically, if the snapshot is not being used, we can archive it. Will add comment.

sql/migrations/warehouse/000032_add_schema_snapshots_to_staging_files.up.sql
```diff
@@ -0,0 +1,27 @@
...
+    ADD COLUMN IF NOT EXISTS schema_snapshot_id UUID REFERENCES wh_staging_file_schema_snapshots(id) ON DELETE SET NULL, -- Link to the schema snapshot
+    ADD COLUMN IF NOT EXISTS schema_patch TEXT; -- JSON Patch (RFC 6902) diff from the snapshot to the actual schema
+
+-- For fast lookup of staging files by schema_snapshot_id.
+CREATE INDEX IF NOT EXISTS wh_staging_files_schema_snapshot_id_index
+    ON wh_staging_files(schema_snapshot_id);
+
+-- For efficient retrieval of latest snapshot by (source_id, destination_id, workspace_id, created_at DESC).
+CREATE INDEX IF NOT EXISTS wh_staging_file_schema_snapshots_source_id_destination_id_workspace_id_created_at_index
+    ON wh_staging_file_schema_snapshots(source_id, destination_id, workspace_id, created_at DESC);
```
shekhar-rudder: Why are we adding (workspace_id, created_at) in the index? Is (source_id, destination_id) not enough ?

achettyiitr: created_at because we are looking for oldest one by sorting. Probably workspaceID we can skip.

shekhar-rudder: Ok lets remove workspace_id

PR #6099, opened 2025-07-14
feat: staging files schema snapshot cached repository with different strategies

achettyiitr:
# Description

- Introduces a new table: `wh_staging_file_schema_snapshots`
  - Stores JSON-encoded schema snapshots for staging files.
  - Keys: source_id, destination_id, workspace_id, and created_at.
- Adds columns to wh_staging_files:
  - `schema_snapshot_id` (UUID, references the new snapshot table)
  - `schema_patch` (TEXT, stores a JSON Patch diff from the snapshot to the actual schema)
- Adds supporting indexes for efficient lookup and retrieval.

# Linear Ticket

[WAR-867: Snapshot cached repository with different strategies](https://linear.app/rudderstack/issue/WAR-867/snapshot-cached-repository-with-different-strategies)

## Security

### DIFF
--- a/warehouse/internal/repo/staging_snapshot.go
+++ b/warehouse/internal/repo/staging_snapshot.go
@@ -1,0 +1,86 @@
+// Package repo provides repository implementations for warehouse entities.
+package repo
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/google/uuid"
+
+	"github.com/rudderlabs/rudder-server/utils/timeutil"
+	sqlmiddleware "github.com/rudderlabs/rudder-server/warehouse/integrations/middleware/sqlquerywrapper"
+	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
+	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+)
+
+const (
+	stagingFileSchemaSnapshotTableName    = whutils.WarehouseStagingFileSchemaSnapshotTable
+	stagingFileSchemaSnapshotTableColumns = `id, schema, source_id, destination_id, workspace_id, created_at`
+)
+
+var ErrNoSchemaSnapshot = errors.New("no schema snapshot found")
+
+type StagingFileSchemaSnapshots repo
+
+// NewStagingFileSchemaSnapshots creates a new StagingFileSchemaSnapshots using the given DB connection.
+func NewStagingFileSchemaSnapshots(db *sqlmiddleware.DB, opts ...Opt) *StagingFileSchemaSnapshots {
+	r := &StagingFileSchemaSnapshots{
+		db:  db,
+		now: timeutil.Now,
+	}
+	for _, opt := range opts {
+		opt((*repo)(r))
+	}
+	return r
+}
+
+// Insert inserts a new schema snapshot into the database and returns its auto-generated ID.
+func (r *StagingFileSchemaSnapshots) Insert(ctx context.Context, sourceID, destinationID, workspaceID string, schemaBytes json.RawMessage) (uuid.UUID, error) {
+	query := `
+		INSERT INTO ` + stagingFileSchemaSnapshotTableName + ` (
+			id, schema, source_id, destination_id, workspace_id, created_at
+		) VALUES ($1, $2, $3, $4, $5, $6)
+		RETURNING id;
+	`
+	id := uuid.New()
+	now := r.now()
+	_, err := r.db.ExecContext(ctx, query,
+		id,
+		schemaBytes,
+		sourceID,
+		destinationID,
+		workspaceID,
+		now,
+	)
+	if err != nil {
+		return uuid.Nil, fmt.Errorf("inserting schema snapshot: %w", err)
+	}
+	return id, nil
+}
+
+// GetLatest returns the most recent schema snapshot for the given source and destination.
+// Returns ErrNoSchemaSnapshot if not found.
+func (r *StagingFileSchemaSnapshots) GetLatest(ctx context.Context, sourceID, destinationID string) (*model.StagingFileSchemaSnapshot, error) {
+	query := `
+		SELECT ` + stagingFileSchemaSnapshotTableColumns + `
+		FROM ` + stagingFileSchemaSnapshotTableName + `
+		WHERE source_id = $1 AND destination_id = $2
+		ORDER BY created_at DESC
+		LIMIT 1;
+	`
+	row := r.db.QueryRowContext(ctx, query, sourceID, destinationID)
+	var snapshot model.StagingFileSchemaSnapshot
+	var schemaRaw []byte
+	if err := row.Scan(&snapshot.ID, &schemaRaw, &snapshot.SourceID, &snapshot.DestinationID, &snapshot.WorkspaceID, &snapshot.CreatedAt); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, ErrNoSchemaSnapshot
+		}
+		return nil, fmt.Errorf("scanning latest schema snapshot by source/dest: %w", err)
+	}
+	snapshot.Schema = schemaRaw
+	snapshot.CreatedAt = snapshot.CreatedAt.UTC()
+	return &snapshot, nil
+}
--- a/warehouse/internal/snapshots/staging_schema.go
+++ b/warehouse/internal/snapshots/staging_schema.go
@@ -1,0 +1,116 @@
+// Package snapshots provides a memory-cached, pluggable-expiry schema snapshot lookup for staging files.
+package snapshots
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math/rand"
+	"time"
+
+	"github.com/google/uuid"
+
+	"github.com/rudderlabs/rudder-go-kit/cachettl"
+	"github.com/rudderlabs/rudder-go-kit/config"
+
+	"github.com/rudderlabs/rudder-server/utils/timeutil"
+	"github.com/rudderlabs/rudder-server/warehouse/internal/model"
+	"github.com/rudderlabs/rudder-server/warehouse/internal/repo"
+)
+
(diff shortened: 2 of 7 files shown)


warehouse/internal/repo/staging_snapshot.go
```diff
@@ -0,0 +1,91 @@
...
+	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+)
+
+const (
+	stagingFileSchemaSnapshotTableName    = whutils.WarehouseStagingFileSchemaSnapshotTable
+	stagingFileSchemaSnapshotTableColumns = `id, schema, created_at, source_id, destination_id, workspace_id`
+)
+
+// ErrNoLatestSchemaSnapshot is returned when no latest schema snapshot is found for the given source, destination, and workspace.
+var ErrNoLatestSchemaSnapshot = errors.New("no latest schema snapshot found for the given source, destination, and workspace")
```
vyeshwanth: nit: can make it `ErrNoSchemaSnapshot` at repo level. Callers can always look for the error and add more context like `schema not found for id`, `schema not found for destination, source` etc.,

warehouse/internal/repo/staging_snapshot.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		return nil, wrapSchemaSnapshotErr("scanning latest schema snapshot by source/dest", err)
+	}
+	snapshot.Schema = schemaRaw
+	return &snapshot, nil
+}
+
+// wrapSchemaSnapshotErr adds context to schema snapshot errors.
+func wrapSchemaSnapshotErr(action string, err error) error {
+	return fmt.Errorf("%s: %w", action, err)
```
vyeshwanth: can inline?

warehouse/internal/cache/staging_snapshot.go
```diff
@@ -0,0 +1,103 @@
+// Package cache provides a memory-cached, pluggable-expiry schema snapshot lookup for staging files.
```
vyeshwanth: can't we use cachettl from go-kit?

achettyiitr: It's acuallt a wrapper over the repo with additional logic around the snapshot strategy. I think the package name was creating confusion. Made it snapshots package.

warehouse/internal/snapshots/staging_schema.go
```diff
@@ -0,0 +1,110 @@
...
+
+		// Expired: insert new and fetch
+		return c.insertAndCache(ctx, sourceID, destinationID, workspaceID, schemaBytes)
+	}
+
+	// Cache miss: fetch from DB
+	snap, err := c.dbRepo.GetLatest(ctx, sourceID, destinationID, workspaceID)
+	if err == nil {
+		if !c.expiryStrategy.IsExpired(snap) {
+			c.cache.Put(key, snap, c.cacheRefreshTTL())
```
vyeshwanth, line 75: can't we just sync the cache expiry duration with expiryStrategy? something like 

`c.cache.Put(key, snap, c.expiryStrategy.Expiry())`

Why do we need to refresh cache or read from DB every 3 hour interval ( default ). Snapshot is immutable right?

achettyiitr: Cache expiry and snapshot expiry are two different things.

achettyiitr: Snapshot expiry can be something like per 10k snapshots.

vyeshwanth: Keeping the cache expiry different but can make it to 12 hours or reduce it.  Right now we are using duration for snapshot expiry but it could be a different from time based strategy. 

@achettyiitr I missed asking this in our offline discussion, why do we want to reduce it to 30 mins if schema sizes are not big for lot of customers?

achettyiitr: So as to avoid keeping it in memory for long time even when it is not needed. Let's say we got a staging file now and then the next one after 5Hrs. No point in keeping it in memory.

warehouse/internal/snapshots/staging_schema.go
```diff
@@ -0,0 +1,111 @@
...
+	c.cache.Put(cacheKey(sourceID, destinationID, workspaceID), snap, c.cacheRefreshTTL())
+	return snap, nil
+}
+
+func (c *StagingFileSchema) cacheRefreshTTL() time.Duration {
+	return c.cacheRefreshInterval.Load() + c.cacheRefreshJitter()
+}
+
+func cacheKey(sourceID, destinationID, workspaceID string) string {
+	return fmt.Sprintf("%s:%s:%s", sourceID, destinationID, workspaceID)
```
shekhar-rudder: Do we need `workspaceID` as part of the cacheKey ?

warehouse/internal/snapshots/staging_schema.go
```diff
@@ -0,0 +1,111 @@
...
+
+// insertAndCache inserts a new snapshot into the DB, fetches it, and caches the result.
+func (c *StagingFileSchema) insertAndCache(ctx context.Context, sourceID, destinationID, workspaceID string, schemaBytes json.RawMessage) (*model.StagingFileSchemaSnapshot, error) {
+	_, err := c.dbRepo.Insert(ctx, sourceID, destinationID, workspaceID, schemaBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	// Fetch the snapshot from the DB
+	snap, err := c.dbRepo.GetLatest(ctx, sourceID, destinationID)
```
shekhar-rudder: We just inserted `schemaBytes` by calling `dbRepo.Insert`. Any harm in assuming it to be latest? It might not be the latest if there are concurrent writes but I don't think we are handling those cases

warehouse/internal/snapshots/staging_schema.go
```diff
@@ -0,0 +1,111 @@
...
+		dbRepo:               dbRepo,
+		cache:                cache,
+		cacheRefreshInterval: cacheRefreshInterval,
+		cacheRefreshJitter:   cacheRefreshJitter,
+		expiryStrategy:       expiryStrategy,
+	}
+}
+
+// Get returns the latest schema snapshot for the given IDs, using cache and DB as needed.
+func (c *StagingFileSchema) Get(ctx context.Context, sourceID, destinationID, workspaceID string, schemaBytes json.RawMessage) (*model.StagingFileSchemaSnapshot, error) {
```
shekhar-rudder: I feel that the Get method is confusing. `Get()` method usually implies a pure lookup. If no value is found in cache/DB, then we can make it the caller's responsibility to call a `Put()` method.

warehouse/internal/snapshots/staging_schema.go
achettyiitr: Yeah, workspaceID can be removed.

warehouse/internal/snapshots/staging_schema.go
achettyiitr: No, we have all details to prepare it. Making the changes.

warehouse/internal/snapshots/staging_schema.go
achettyiitr: Good point. 👍

achettyiitr: Renaming it to `GetOrCreate`.

warehouse/internal/snapshots/staging_schema_test.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+		db := &mockStagingFileSchemaDBRepo{
+			getLatestFunc: func(ctx context.Context, sourceID, destinationID string) (*model.StagingFileSchemaSnapshot, error) {
+				return fresh, nil
+			},
+			insertFunc: func(ctx context.Context, sourceID, destinationID, workspaceID string, schemaBytes json.RawMessage) (uuid.UUID, error) {
+				return fresh.ID, nil
+			},
+		}
+		cache := NewStagingFileSchema(config.New(), db, &fixedExpiryStrategy{expired: true})
```
shekhar-rudder: cache -> schema ?

warehouse/internal/snapshots/staging_schema_test.go
```diff
@@ -0,0 +1,252 @@
...
+		cache.cache.Put(cacheKey("279L3gEKqwruBoKGsXZtSVX7vIy", "27CHciD6leAhurSyFAeN4dp14qZ"), &model.StagingFileSchemaSnapshot{CreatedAt: time.Now().Add(-time.Hour)}, cache.cacheRefreshTTL())
```
shekhar-rudder: `cache.cache.Put` -> Is it possible to use the public `GetOrCreate` method here ?

warehouse/internal/snapshots/staging_schema.go
shekhar-rudder: The name is now more appropriate. However, I still am not sure about the method signature. Currently, we’re requiring the client to pass `schemaBytes` on every call, which forces them to compute it even if it might already be present in the cache. If we instead move the "Create" logic to a separate function, it would allow the client to create the schema lazily. This assumes that schema creation is a non-trivial operation.

warehouse/internal/snapshots/staging_schema_test.go
achettyiitr: Rnamed schemaCache.

warehouse/internal/snapshots/staging_schema.go
achettyiitr: Reason was because there is only one use case around the `/process` API, where we have the modified schema and we want the snapshot shema.

warehouse/internal/snapshots/staging_schema_test.go
achettyiitr: No.

achettyiitr: If we have Get and Put separate, we can have it. But based on the usecase it's required around the /process API, where we have the modified schema and we want the snapshot shema.

shekhar-rudder (approved):

PR #6100, opened 2025-07-14
feat: start populating snapshot id and schema patch

achettyiitr:
# Description

- Updates the warehouse API and internal logic to:
  - Retrieve or create a schema snapshot for each staging file (when enabled).
  - Generate and store a JSON Patch (RFC 6902) diff between the snapshot schema and the current schema.
  - Store both the schema_snapshot_id and the schema_patch in the wh_staging_files table.
- Adds a new utility (warehouse/utils/jsonpatch.go) for generating and applying JSON Patch diffs.
- Updates the staging file model and repository to support the new fields.

## Linear Ticket

[WAR-868: Start populate the snapshotID and the diff schemas](https://linear.app/rudderstack/issue/WAR-868/start-populate-the-snapshotid-and-the-diff-schemas)

## Security


warehouse/internal/api/http_test.go
```diff
@@ -52,6 +56,14 @@ func loadFile(t *testing.T, path string) string {
...
 }
 
+type mockStagingFileSchemaSnapshotGetter struct {
+	getFunc func(ctx context.Context, sourceID, destinationID, workspaceID string, schemaBytes json.RawMessage) (*model.StagingFileSchemaSnapshot, error)
+}
+
+func (m *mockStagingFileSchemaSnapshotGetter) GetOrCreate(ctx context.Context, sourceID, destinationID, workspaceID string, schemaBytes json.RawMessage) (*model.StagingFileSchemaSnapshot, error) {
+	return m.getFunc(ctx, sourceID, destinationID, workspaceID, schemaBytes)
+}
+
```
vyeshwanth, line 66: why not use gomock? Avoid writing boilerplate and rely on generated code.

vyeshwanth (approved):

warehouse/internal/api/http_test.go
achettyiitr: For larger or more complex interfaces, or if you want strict call verification, GoMock is probably preferred. But for a single-method interface, a hand-written mock would be easy to implement and inject.

warehouse/internal/repo/staging.go
```diff
@@ -41,16 +43,18 @@ const stagingTableColumns = `
...
+	SourceTaskRunID               string   `json:"source_task_run_id"`
+	SourceJobID                   string   `json:"source_job_id"`
+	SourceJobRunID                string   `json:"source_job_run_id"`
+	TimeWindowYear                int      `json:"time_window_year"`
+	TimeWindowMonth               int      `json:"time_window_month"`
+	TimeWindowDay                 int      `json:"time_window_day"`
+	TimeWindowHour                int      `json:"time_window_hour"`
+	DestinationRevisionID         string   `json:"destination_revision_id"`
+	ServerInstanceID              string   `json:"server_instance_id"`
+	SnapshotPatchBytes            *int     `json:"snapshot_patch_bytes,omitempty"`
```
shekhar-rudder: SnapshotPatchBytes -> SnapshotPatchSize ?

warehouse/internal/repo/staging.go
```diff
@@ -142,6 +150,17 @@ func (sf *StagingFiles) Insert(ctx context.Context, stagingFile *model.StagingFi
...
 
+	var schemaSnapshotID interface{}
+	if stagingFile.SnapshotSchema != nil {
+		schemaSnapshotID = stagingFile.SnapshotSchema.ID.String()
+	}
+	var schemaPatchPayload interface{}
+	if len(stagingFile.SnapshotPatch) > 0 {
+		schemaPatchPayload = stagingFile.SnapshotPatch
+	} else {
+		schemaPatchPayload = nil
```
shekhar-rudder: Do we need the else case ?

warehouse/internal/repo/staging.go
```diff
@@ -142,6 +150,17 @@ func (sf *StagingFiles) Insert(ctx context.Context, stagingFile *model.StagingFi
 		bytesPerTablePayload = nil
 	}
 
+	var schemaSnapshotID interface{}
+	if stagingFile.SnapshotSchema != nil {
+		schemaSnapshotID = stagingFile.SnapshotSchema.ID.String()
```
shekhar-rudder: If we only need the schema ID, should we avoid populating the entire `SnapshotSchema` struct since it would also include the full schema field unnecessarily?

warehouse/api/http.go
```diff
@@ -103,25 +108,34 @@ func NewApi(
...
+		db:                         db,
+		notifier:                   notifier,
+		bcConfig:                   bcConfig,
+		statsFactory:               statsFactory,
+		tenantManager:              tenantManager,
+		bcManager:                  bcManager,
+		sourceManager:              sourceManager,
+		triggerStore:               triggerStore,
+		stagingRepo:                repo.NewStagingFiles(db),
+		StagingFileSchemaSnapshots: stagingFileSchemaSnapshots,
```
shekhar-rudder: Given that `StagingFileSchemaSnapshots` and `enableStagingFileSchemaSnapshot` are tied to a specific API, should we avoid including them in the API struct where all other fields are generic.

warehouse/internal/api/http.go
```diff
@@ -27,11 +30,18 @@ type stagingFilesRepo interface {
...
+type stagingFileSchemaSnapshotGetter interface {
+	GetOrCreate(ctx context.Context, sourceID, destinationID, workspaceID string, schemaBytes json.RawMessage) (*model.StagingFileSchemaSnapshot, error)
+}
+
 type WarehouseAPI struct {
-	Logger      logger.Logger
-	Stats       stats.Stats
-	Repo        stagingFilesRepo
-	Multitenant *multitenant.Manager
+	Logger                          logger.Logger
```
shekhar-rudder: same here

warehouse/internal/repo/staging.go
achettyiitr: No we don't need it, removing it.

warehouse/internal/repo/staging.go
achettyiitr: Although we are passing `SnapshotSchema` pointer, there will be only 1 that will be circulating around. But since we are not using it, removing that.

warehouse/api/http.go
achettyiitr: Do you mean renaming it to `StagingFileSnapshots` and `enableStagingFileSnapshot`?

achettyiitr: Moved all specific logic to StagingFileSchemaSnapshotHandler.

warehouse/internal/api/http.go
achettyiitr: Moved all specific logic to StagingFileSchemaSnapshotHandler.

PR #6107, opened 2025-07-14
fix: set correct payloadStage in sample response based on reportJobsdbPayload flag

itsmihir:
# Description

## Linear Ticket

## Security

### DIFF
--- a/router/worker.go
+++ b/router/worker.go
@@ -1057,14 +1057,6 @@ func (w *worker) postStatusOnResponseQ(respStatusCode int, destinationJob *types
 		// TODO: update after observing the sizes of the payloads
 	}
 
-	// Saving payload to DB only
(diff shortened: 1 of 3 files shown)


router/worker.go
```diff
@@ -1030,7 +1033,7 @@ func (w *worker) postStatusOnResponseQ(respStatusCode int, destinationJob *types
 			worker:     w,
 			job:        destinationJobMetadata.JobT,
 			status:     status,
-			payload:    inputPayload,
+			payload:    payload,
```
ktgowtham: If it's success, should we keep the `destinationJobMetadata.JobT.EventPayload`. we don't surface it on UI and also, given we might capture huge payloads if we start capturing the router output for successes too?

PR #6114, opened 2025-07-15
chore: reset timer to optimise router batching

cisse21:
# Description

we end up sending smaller batches to router because the timer is not reset after making a call due to batch size. this pr resets the timer every-time a call is made to transformer 
      
## Linear Ticket

fixes PIPE-2204

## Security


mihir20 (approved):
LGTM

PR #6118, opened 2025-07-16
feat: change datatype from jsonb to text for warehouse

achettyiitr:
# Description

- Alters the schema column in `wh_schemas`, `wh_staging_files`, and `wh_uploads` tables from **JSONB** to **TEXT**.
- Alters the payload column in `pg_notifier_queue` from **JSONB** to **TEXT**.
- Both migrations are guarded by feature flags (`Warehouse.enableJsonbToTextMigration` and `PgNotifierQueue.enableJsonbToTextMigration`).
- Ensures idempotency and safe execution by checking column types before altering.

## Linear Ticket

- Resolves WAR-849

## Security


vyeshwanth (approved):

sql/migrations/pg_notifier_queue_always/000001_alter_payload_jsonb_to_text.up.tmpl
```diff
@@ -0,0 +1,18 @@
...
+    DO $$
+    BEGIN
+      IF EXISTS (
+        SELECT 1 FROM information_schema.columns
+        WHERE table_name = 'pg_notifier_queue'
+          AND column_name = 'payload'
+          AND data_type != 'text'
+      ) THEN
+        RAISE NOTICE 'Altering column "payload" in table "pg_notifier_queue" from % to text',
+          (SELECT data_type FROM information_schema.columns WHERE table_name = 'pg_notifier_queue' AND column_name = 'payload');
```
shekhar-rudder: Instead of querying `information_schema` twice, how about we query it once and based on the output we write an if-else block ?

achettyiitr: Done.

PR #6124, opened 2025-07-16
chore: enhance rsources stats collector with component tagging and error handling

mihir20:
# Description

enhance rsources stats collector with component tagging and error handling

## Linear Ticket

-

## Security


jobsdb/jobsdb.go
```diff
@@ -2449,6 +2449,7 @@ func (jd *Handle) addNewDSLoop(ctx context.Context) {
 							if err != nil {
 								return err
 							}
+							jd.logger.Infon("Acquired lock", logger.NewField("ds", latestDS), logger.NewStringField("jobsdb", jd.tablePrefix))
```
fracasula, line 2452: How often would you say we would get these printed out? I'm asking because I'm a bit afraid of noise/pollution.

services/rsources/stats_collector.go
```diff
@@ -67,11 +72,12 @@ func NewStatsCollector(jobservice JobService, opts ...OptFunc) StatsCollector {
 	for _, opt := range opts {
 		opt(sc)
 	}
+	sc.stats.publishTime = statFactory.NewTaggedStat(rsourcesPublishTime, stats.TimerType, stats.Tags{"module": component})
```
fracasula, line 75: Is this a new metric? Do we have defined buckets for it?

jobsdb/jobsdb.go
mihir20: not very often we are printing this everytime we create a new table. we have a log [here](https://github.com/rudderlabs/rudder-server/pull/6124/files#diff-6b1d4978447cbee0288087476d81c9cf578adc1b0cdedfe2fa11327d26499e16L2457) as well

services/rsources/stats_collector.go
mihir20: yes its a new metric, I will add buckets for this

ktgowtham (approved):
Could you also add a log at start and completion of [this](https://github.com/rudderlabs/rudder-server/blob/81dbec0f369b44d3eb5f70120ae5f7db85502de9/services/rsources/handler.go#L426)?

PR #6125, opened 2025-07-17
fix: snowpipe events getting duplicated if status API fails

shekhar-rudder:
# Description

**Problem**

Data duplication can occur in the Snowpipe Streaming destination when the Poll method encounters a 404 error from the Snowpipe service during status polling. This happens if the channelId used for polling is no longer valid (for example, after a Snowpipe service restart). In the previous behavior, such 404s would cause the Poll method to mark events as failed, which would then be retried, leading to duplicate data.

**Solution**

The Polling logic now follows the same error handling pattern as the insert logic: when a 404 (channel not found) is encountered during polling, the system attempts to recover by recreating the channel and continuing the status check, rather than immediately marking the events as failed. This prevents unnecessary retries and data duplication, and ensures more robust handling of transient channel invalidation scenarios.

## Linear Ticket
 
https://linear.app/rudderstack/issue/WAR-920/data-duplication-if-the-getstatus-api-fails-while-polling

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
@@ -544,6 +544,7 @@ func (m *Manager) processPollImportInfos(ctx context.Context, infos []*importInf
 			m.polledImportInfoMap[info.ChannelID] = info
 			continue
 		}
+		info.Failed = false
 		if !inProgress {
 			m.polledImportInfoMap[info.ChannelID] = info
 		}
@@ -597,8 +598,9 @@ func (m *Manager) getImportStatus(ctx context.Context, info *importInfo) (bool, 
 
 		log.Infon("Recreated channel for polling", logger.NewStringField("channelID", recreatedChannel.ChannelID))
 
-		// Update the channel ID in importInfo for future polls. Though I'm not sure if this is needed.
-		info.ChannelID = recreatedChannel.ChannelID
+		// The new channel id is not being associated with the import info. This is because even if we do that,
+		// it will update the in memory map and not the database
+		// So the next time we poll, we will get the old channel id and not the new one
 
 		statusRes2, err2 := m.api.GetStatus(ctx, recreatedChannel.ChannelID)
 		if err2 != nil {


router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -562,19 +562,63 @@ func (m *Manager) getImportStatus(ctx context.Context, info *importInfo) (bool,
...
+		req := buildCreateChannelRequest(m.destination.ID, m.config.instanceID, &destConf, info.Table)
+		recreatedChannel, recreateErr := m.api.CreateChannel(ctx, req)
+		if recreateErr != nil {
+			return false, fmt.Errorf("recreating channel: %w", recreateErr)
+		}
+		m.channelCache.Store(info.Table, recreatedChannel)
+
+		log.Infon("Recreated channel for polling", logger.NewStringField("channelID", recreatedChannel.ChannelID))
+
+		// Update the channel ID in importInfo for future polls. Though I'm not sure if this is needed.
```
achettyiitr: Is this comment needed? Or is it a Question?

achettyiitr: If the polled import channel ID changes, we would need to update it, otherwise next time it will poll for the same previous one.

shekhar-rudder: Updated the comment

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -544,6 +544,7 @@ func (m *Manager) processPollImportInfos(ctx context.Context, infos []*importInf
 			m.polledImportInfoMap[info.ChannelID] = info
 			continue
 		}
+		info.Failed = false
```
shekhar-rudder, line 547: This change is unrelated to rest of the changes in the PR. The bug here is that if in the first poll attempt, if there is an error, `info.Failed` gets marked as `true`. But in the next attempt, if there is no error, we are not clearing up the `Failed` field.

PR #6131, opened 2025-07-17
fix: invalidate warehouse schema cache in case of export error

shekhar-rudder:
# Description

The warehouse schema cache can become outdated, leading to persistent data export failures until the cache expires. To improve reliability, the schema cache should be invalidated immediately upon encountering a sync error during data export. This ensures that the next export attempt fetches the latest schema, reducing the risk of repeated failures due to schema drift. The invalidation should be limited to cases where a sync error occurs during data delivery, and a log entry should be created whenever invalidation happens. This change will help the system recover more quickly from schema mismatches and improve overall data delivery robustness. 

## Linear Ticket

https://linear.app/rudderstack/issue/WAR-937

## Security

### DIFF
--- a/warehouse/router/upload.go
+++ b/warehouse/router/upload.go
@@ -19,6 +19,7 @@ import (
 	"github.com/rudderlabs/rudder-go-kit/filemanager"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
+	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
 
 	"github.com/rudderlabs/rudder-go-kit/jsonrs"
 
@@ -410,7 +411,7 @@ func (job *UploadJob) run() (err error) {
 				// schema is being cleared only in ExportedData case and not in other cases.
 				invErr := job.whSchemaRepo.SetExpiryForDestination(job.ctx, job.warehouse.Destination.ID, job.now())
 				if invErr != nil {
-					job.logger.Errorf("Failed to invalidate schema cache: %v", invErr)
+					job.logger.Errorn("Failed to invalidate schema cache", obskit.Error(err))
 				} else {
 					job.logger.Infon("Invalidated warehouse schema cache due to sync error")
 				}


warehouse/router/upload.go
```diff
@@ -404,6 +404,16 @@ func (job *UploadJob) run() (err error) {
...
+				// Invalidate schema cache on sync error to prevent repeated export failures
+				// This sets the schema expiry to now, so the next attempt will fetch the latest schema
+				// Since we are not checking if the error is due to schema, to prevent unnecessary invalidations
+				// schema is being cleared only in ExportedData case and not in other cases.
+				invErr := job.whSchemaRepo.SetExpiryForDestination(job.ctx, job.warehouse.Destination.ID, job.now())
+				if invErr != nil {
+					job.logger.Errorf("Failed to invalidate schema cache: %v", invErr)
+				} else {
+					job.logger.Infon("Invalidated warehouse schema cache due to sync error")
+				}
```
achettyiitr: Why just logging. It's a db call which should succeeded?

achettyiitr: Or planning to add an alert for it later?

warehouse/router/upload.go
```diff
@@ -404,6 +404,16 @@ func (job *UploadJob) run() (err error) {
 		case model.ExportedData:
 			newStatus = nextUploadState.failed
 			if err = job.exportData(); err != nil {
+				// Invalidate schema cache on sync error to prevent repeated export failures
+				// This sets the schema expiry to now, so the next attempt will fetch the latest schema
+				// Since we are not checking if the error is due to schema, to prevent unnecessary invalidations
+				// schema is being cleared only in ExportedData case and not in other cases.
+				invErr := job.whSchemaRepo.SetExpiryForDestination(job.ctx, job.warehouse.Destination.ID, job.now())
+				if invErr != nil {
+					job.logger.Errorf("Failed to invalidate schema cache: %v", invErr)
```
achettyiitr: Can we use structured logging in here?
```suggestion
					job.logger.Errorn("Failed to invalidate schema cache: %v", invErr)
```

warehouse/router/upload.go
shekhar-rudder: - The sync has already failed. So we will be returning an error message. Are you suggesting to append the invalidation error to the original error ?
- No plans to add alert.

achettyiitr: If there is no plans of adding alert, how will we know if it succeeded?

warehouse/router/upload.go
shekhar-rudder: Done

warehouse/router/upload.go
shekhar-rudder: Alert for this invalidation feels like an overkill. WDYT ?

achettyiitr (approved):

PR #6150, opened 2025-07-22
chore(jobsdb): introduce context with timeout during add new ds loop

atzoum:
# Description

Introducing a context with timeout during addNewDSLoop so that: 
- the loop can stop early during shutdown
- the dsList lock is not being held indefinitely in case of blocking long running queries

## Linear Ticket

resolves PIPE-2254

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -517,10 +517,11 @@ type Handle struct {
 		maxTableSize                   config.ValueLoader[int64]
 		cacheExpiration                config.ValueLoader[time.Duration]
 		addNewDSLoopSleepDuration      config.ValueLoader[time.Duration]
+		addNewDSTimeout                config.ValueLoader[time.Duration]
 		refreshDSListLoopSleepDuration config.ValueLoader[time.Duration]
+		refreshDSTimeout               config.ValueLoader[time.Duration]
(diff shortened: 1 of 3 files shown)


jobsdb/jobsdb.go
```diff
@@ -2446,7 +2449,7 @@ func (jd *Handle) addNewDSLoop(ctx context.Context) {
 							// We will release the list lock after the transaction ends, that's why we need to use an async lock
 							dsListLock, releaseDsListLock, err = jd.dsListLock.AsyncLockWithCtx(ctx)
 							if err != nil {
-								return err
+								return fmt.Errorf("acquiring dsListLock: %w", err)
 							}
 							jd.logger.Infon("Acquired lock", logger.NewField("ds", latestDS), logger.NewStringField("jobsdb", jd.tablePrefix))
 							if _, err = tx.Exec(fmt.Sprintf(`LOCK TABLE %q IN EXCLUSIVE MODE;`, latestDS.JobTable)); err != nil {
```
mihir20:
```suggestion
							if _, err = tx.ExecContext(ctx, fmt.Sprintf(`LOCK TABLE %q IN EXCLUSIVE MODE;`, latestDS.JobTable)); err != nil {
```

mihir20 (approved):

cisse21 (approved):

PR #6170, opened 2025-07-24
chore: cache pg image

cisse21:
# Description

Cache PG image once and reuse them across tests

## Security


.github/workflows/tests.yaml
```diff
@@ -23,12 +23,31 @@ jobs:
...
+        id: cache-postgres-integration
+        uses: actions/cache@v4
+        with:
+          path: ~/image-cache/postgres-17-alpine.tar
+          key: postgres-17-alpine-image-${{ runner.os }}
+
+      - name: Pull and save Postgres 17-alpine image if not cached
+        if: steps.cache-postgres-integration.outputs.cache-hit != 'true'
+        run: |
+          docker pull postgres:17-alpine
```
atzoum: This tag doesn't point to an immutable image, the image will get updated when pg 17.6 is out or alpine 3.23. 
So we can end up with one runner using pg 17.5 and another one pg 17.6

cisse21: What you said makes sense but our tests always pull a `postgres:17-alpine` image and we do that multiple time across tests. Can you think of any better way to do this other than hardcoding the image in the testhelper?

PR #6181, opened 2025-07-28
[WIP] feat(router): support different throttler per event type

atzoum:
# Description

## Linear Ticket

## Security


router/internal/jobiterator/jobiterator.go
```diff
@@ -182,6 +182,12 @@ func (ji *Iterator) Stop() int {
 	return discardedJobs
 }
 
+// StopQueries forces the iterator to stop fetching more jobs from jobsDB by setting the maximum number of queries to the current query count.
+func (ji *Iterator) StopQueries() {
+	ji.maxQueries = ji.state.stats.QueryCount
```
cisse21, line 187: Why do we not set closed to true? but do this instead

router/throttler/internal/static/static_per_event_type.go
```diff
@@ -0,0 +1,44 @@
...
+		window: config.GetReloadableDurationVar(0, time.Second,
+			fmt.Sprintf(`Router.throttler.%s.%s.%s.timeWindow`, destType, destinationID, eventType),
+			fmt.Sprintf(`Router.throttler.%s.%s.timeWindow`, destType, destinationID),
+			fmt.Sprintf(`Router.throttler.%s.%s.timeWindow`, destType, eventType),
+			fmt.Sprintf(`Router.throttler.%s.timeWindow`, destType),
+		),
+		// static cost for per-event-type throttler: cost was originally introduced to address rate limit differences between different event types, so not needed when using per-event-type throttler
+		staticCost: true,
+
+		every: sync.NewEvery(time.Second),
```
cisse21: Should we make this `time.Second` configurable?

cisse21: Similarly at other places

router/throttler/factory.go
```diff
@@ -34,72 +47,54 @@ func NewFactory(config *config.Config, stats stats.Stats) (Factory, error) {
...
-			"Router.throttler.adaptive.enabled"),
-		static:   st,
-		adaptive: at,
+	allEventsAlgorithm, ok := f.allEventTypesAlgos[destinationID]
+	if !ok {
+		allEventsAlgorithm = algorithm.NewAdaptiveAlgorithm(destType, f.config, adaptive.GetAllEventsWindowConfig(f.config, destType, destinationID))
+		f.allEventTypesAlgos[destinationID] = allEventsAlgorithm
 	}
-	return f.throttlers[destID]
+	perEventAlgorithm := algorithm.NewAdaptiveAlgorithm(destType, f.config, adaptive.GetPerEventWindowConfig(f.config, destType, destinationID, eventType))
```
cisse21, line 83: Why is there a cache for `allEventsAlgorithm` and not for `perEventAlgorithm`?

router/throttler/internal/static/static_per_event_type.go
atzoum: Don't think so, just want to avoid updating the gauge every microsecond. Why would anyone want updates at higher/lower rates?

router/throttler/factory.go
atzoum: For 1 destination we are going to create N throttlers, where N the number of event types.
In case of the `allEventsAlgorithm`, all N throttlers need to share the same algorithm implementation, i.e. gather 429 statistics for all event types and adjust throttling factor globally. This doesn't stand true for `perEventAlgorithm` where each throttler needs to have its own algorithm.

_**hint:**_ even though we're creating different throttlers per event type, each throttler will be using the same limiter, but, will be trying to limit based on the same (all events) or different (per event) throttling keys per case.

mihir20: understood:

for destination d1 and eventType throttling disabled
we get event with type t1
we will check f.throttlers["d1:t1"] it is not present in the map
we create a algo for d1 and cache it in f.allEventTypesAlgos["d1"]
next time an event with type t2 and destination d1 comes we will check f.throttlers["d1:t2"]
it will not be present in the map and we will check for algo in f.allEventTypesAlgos["d1"] it will be there we will use it

router/throttler/internal/adaptive/throttler.go
```diff
@@ -0,0 +1,94 @@
...
+	"fmt"
+	"time"
+
+	"github.com/rudderlabs/rudder-go-kit/config"
+	"github.com/rudderlabs/rudder-go-kit/stats"
+	"github.com/rudderlabs/rudder-server/router/throttler/internal/sync"
+	"github.com/rudderlabs/rudder-server/router/throttler/internal/types"
+)
+
+// NewPerEventTypeThrottler constructs a new adaptive throttler for a specific event type of a destination
```
mihir20: added by mistake 😅

router/throttler/internal/types/throttler.go
```diff
@@ -0,0 +1,16 @@
+package types
+
+import "context"
+
+// Throttler interface defines methods for checking and managing request limits.
+// Typically one would have a single throttler per destination id and (optionally) per event type.
+type Throttler interface {
+	// Allow returns true if the limit is not exceeded, false otherwise.
```
mihir20:
```suggestion
	// CheckLimitReached returns true if the limit is not exceeded, false otherwise.
```

router/throttler/internal/adaptive/types.go
```diff
@@ -0,0 +1,12 @@
...
+type Algorithm interface {
+	// ResponseCodeReceived is called when a response is received from the destination
+	ResponseCodeReceived(code int)
+	// Shutdown is called when the throttler is shutting down
+	Shutdown()
+	// limitFactor returns a factor that is used to multiply the limit, a number between 0 and 1
+	LimitFactor() float64
+}
+
+const DefaultMaxLimit = 1000 // 1000 requests per second
```
mihir20:
```suggestion
const DefaultMaxRateLimit = 1000 // 1000 requests per second
```

router/throttler/internal/adaptive/throttler.go
```diff
@@ -0,0 +1,94 @@
...
+
+type throttler struct {
+	destinationID string
+	eventType     string
```
mihir20, line 17: where are we using `eventType`? is this only for tests?

router/throttler/internal/adaptive/throttler.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (t *throttler) validConfiguration() bool {
+	return t.minLimit.Load() > 0 && t.maxLimit() > 0 && t.window.Load() > 0 && t.minLimit.Load() <= t.maxLimit()
+}
+
+func (t *throttler) CheckLimitReached(ctx context.Context, cost int64) (limited bool, retErr error) {
+	t.updateGauges()
+	if !t.validConfiguration() {
+		return false, nil
```
mihir20, line 43: do we need some log or something, to identify that we have setup some incorrect configuration and destination will not be ratelimited?

router/throttler/internal/sync/every.go
```diff
@@ -0,0 +1,32 @@
+package sync
+
+import (
+	"sync"
+	"time"
+)
+
+// Every is a utility that allows executing a function at most once every specified interval.
```
mihir20: should we add this in go-kit?

mihir20 (approved):
LGTM

router/throttler/factory.go
atzoum: Yes. So for this example, we are going to have:
- 2 throttlers for d1, one for t1 & another for t2
- Both throttlers share the same algorithm
- Both throttlers are using the same throttling key, effectively applying the same rate limit for jobs of destination d1 regardless of the event type

router/throttler/internal/sync/every.go
atzoum: yes we should

router/throttler/internal/adaptive/throttler.go
atzoum: yes not really needed

router/throttler/internal/adaptive/throttler.go
atzoum: can log once every 1 minute

router/isolation/isolation_test.go
```diff
@@ -45,16 +52,56 @@ func TestIsolationStrategy(t *testing.T) {
...
+				// No specific config for dest456, should use destination type config
+				c.Set("Router.throttler.WEBHOOK.throttlerPerEventType", true)
+				require.True(t, strategy.StopQueries(types.ErrDestinationThrottled, anotherDestID))
+			})
+
+			t.Run("uses global fallback config", func(t *testing.T) {
+				anotherDestID := "dest789"
+
+				// No specific config for dest789 or destination type, should use global config
+				c.Set("Router.throttler.throttlerPerEventType", true)
```
fracasula, line 102: [nitpick] we're sharing a conf here, not great for test isolation. it's ok for now since there is no need to run these in parallel but I think tests should be designed with complete isolation in mind in the first place.

router/throttler/internal/adaptive/adaptive.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+// NewThrottler constructs a new adaptive throttler that can switch between all event types and per event type throttling.
+func NewThrottler(destType, destinationID, eventType string,
+	perEventAlgorithm, allEventsAlgorithm Algorithm,
+	limiter types.Limiter, config *config.Config, stat stats.Stats, log logger.Logger,
+) types.Throttler {
+	return switcher.NewThrottlerSwitcher(
+		config.GetReloadableBoolVar(false,
+			fmt.Sprintf(`Router.throttler.%s.%s.throttlerPerEventType`, destType, destinationID),
```
fracasula: [optional] I think I saw this snippet of code already somewhere else. if that's the case then it'd be less robust since we'd be fetching the same config twice in 2 different places where a simple typo can become a nightmare to pinpoint.

router/throttling_per_event_type_test.go
```diff
@@ -0,0 +1,374 @@
...
+// The server will respond with a 200 status code and the response body will contain the status code.
+// It will also update the spec with the received counts and last received time for each event type
+func (throttlingPerEventTypeMethods) newWebhook(t *testing.T, spec *throttlingPerEventTypeSpec) *throttlingPerEventTypeWebhook {
+	var wh throttlingPerEventTypeWebhook
+	wh.spec = spec
+
+	wh.Server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		body, err := io.ReadAll(r.Body)
+		require.NoError(t, err, "should be able to read the request body")
+		typeResult := gjson.GetBytes(body, "type")
```
fracasula, line 309: We can start using the [kit jsonparser](https://github.com/rudderlabs/rudder-go-kit/blob/main/jsonparser/jsonparser.go). Once it's used everywhere we can even have rules to enforce it like we did for jsonrs.

atzoum: using jsonparser, but not in test since it is convenient to use gjson here

PR #6183, opened 2025-07-28
chore: modifications and tests related changes for the notifier lz4 and jsonb to text conversion

achettyiitr:
# Description

- Add necessary tests and modifications related to enableJsonbToText and enableLz4Compression configs for PG notifier

## Linear Ticket

- Resolves WAR-854, WAR-849

## Security


services/notifier/notifier_test.go
achettyiitr, line 1: No changes in this file. Only some minor refactorings.

achettyiitr: Also, removing `t.Parallel()` because it is less prone to errors (because of test postgres container not being up if we run too many of them parallely) to run sequential tests where external resources are being used.

vyeshwanth (approved):

PR #6185, opened 2025-07-29
chore: modifications and tests related chagnes for the warehouse lz4 and jsonb to text conversion

achettyiitr:
# Description

- Add necessary tests and modifications related to enableJsonbToText and enableLz4Compression configs for Warehouse

## Linear Ticket

- Resolves WAR-854, WAR-849

## Security

### DIFF
--- a/warehouse/internal/repo/schema.go
+++ b/warehouse/internal/repo/schema.go
@@ -26,7 +26,7 @@ const whSchemaTableColumns = `
 	namespace,
    	destination_id,
 	destination_type,
-	schema,
+	schema::jsonb,
    	created_at,
    	updated_at,
 	expires_at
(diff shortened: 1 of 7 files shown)


warehouse/internal/repo/schema.go
```diff
@@ -26,7 +26,7 @@ const whSchemaTableColumns = `
 	namespace,
    	destination_id,
 	destination_type,
-	schema,
+	schema::jsonb,
```
shekhar-rudder: Can you add some more details? Are the queries failing without these changes?

achettyiitr: Ideally, it was required only at places where we are doing `jsonb_each(s.schema::jsonb)`, removed from other places.

shekhar-rudder (approved):

PR #6200, opened 2025-07-30
feat: warehouse grpc API to get namespace mappings of a destination

shekhar-rudder:
### 📋 **Overview**
Implements the `GetDestinationNamespaces` gRPC API method for the warehouse service . The API allows clients to retrieve the most recent namespace for each source associated with a given destination ID.

### ✅ **Changes Made**
- **Proto**: Added `GetDestinationNamespaces` RPC method and corresponding request/response messages
- **Model**: Added `NamespaceMapping` struct with `SourceID` and `Namespace` fields
- **Repository**: Added `GetDestinationNamespaces` method using `DISTINCT ON` query to get most recent namespace per source
- **gRPC Service**: Implemented API method with input validation, error handling, and structured logging
- **Tests**: Added comprehensive repository and gRPC tests

## Security

### DIFF
--- a/warehouse/api/grpc.go
+++ b/warehouse/api/grpc.go
@@ -1206,7 +1206,7 @@ func (g *GRPC) GetDestinationNamespaces(ctx context.Context, request *proto.GetD
 		}
 	}
 
-	log.Infon(fmt.Sprintf("Successfully retrieved %d destination namespaces", len(protoMappings)))
+	log.Infon("Successfully retrieved destination namespaces", logger.NewIntField("mappingCount", int64(len(protoMappings))))
 	return &proto.GetDestinationNamespacesResponse{
 		NamespaceMappings: protoMappings,
 	}, nil


warehouse/api/grpc.go
```diff
@@ -1176,3 +1176,39 @@ func (g *GRPC) getLatencyAggregationType(
 	}
 	return aggregationType, nil
 }
+
+// GetDestinationNamespaces returns the most recent namespace for each source for a given destination ID.
+func (g *GRPC) GetDestinationNamespaces(ctx context.Context, request *proto.GetDestinationNamespacesRequest) (*proto.GetDestinationNamespacesResponse, error) {
+	log := g.logger.Withn(
+		obskit.DestinationID(request.GetDestinationId()),
+	)
+	log.Infon("Getting destination namespaces")
```
vyeshwanth, line 1184: can be a debug log?

warehouse/api/grpc.go
```diff
@@ -1176,3 +1176,39 @@ func (g *GRPC) getLatencyAggregationType(
...
+	// Convert to proto response
+	protoMappings := make([]*proto.NamespaceMapping, len(mappings))
+	for i, mapping := range mappings {
+		protoMappings[i] = &proto.NamespaceMapping{
+			SourceId:  mapping.SourceID,
+			Namespace: mapping.Namespace,
+		}
+	}
+
+	log.Infon(fmt.Sprintf("Successfully retrieved %d destination namespaces", len(protoMappings)))
```
vyeshwanth: use structured logging instead of fmt.Sprintf ? can be debug log, any reason why we are logging these with info level?

warehouse/api/grpc.go
shekhar-rudder: I don’t think debug logs are useful since we never enable debug level. Given we’re talking about adding more logs for better debuggability, this should be fine.

warehouse/api/grpc.go
shekhar-rudder: Fixed

achettyiitr (approved):

PR #6223, opened 2025-08-11
chore: throttler should emit no warning if not enabled

atzoum:
# Description

Our only means of disabling throttling for a destination type or destination id is by setting the limit or window to zero. Thus it makes no sense to emit a warning log in case some throttler is disabled. Renamed `validConfiguration` to `enabled` to avoid any confusion in the future.

Regression introduced by #6181

## Security

### DIFF
--- a/router/throttler/internal/static/throttler.go
+++ b/router/throttler/internal/static/throttler.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
-	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
 	kitsync "github.com/rudderlabs/rudder-go-kit/sync"
 	"github.com/rudderlabs/rudder-server/router/throttler/internal/types"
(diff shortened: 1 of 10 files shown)


router/throttler/internal/static/throttler.go
```diff
@@ -23,25 +22,23 @@ type throttler struct {
...
-	onceEveryGauge         *kitsync.OnceEvery
-	rateLimitGauge         stats.Gauge
+	onceEveryGauge *kitsync.OnceEvery
+	rateLimitGauge stats.Gauge
 }
 
 func (t *throttler) CheckLimitReached(ctx context.Context, cost int64) (limited bool, retErr error) {
-	t.updateGauges()
-	if !t.validConfiguration() {
-		t.logInvalidConfigWarning()
```
mihir20, line 34: do we not want a warning log for invalid config now?

atzoum: No, since providing invalid config is actually our only means of disabling throttling. We don't want to get warnings for destinations where we've purposefully disabled throttling.
Originally I performed this change assuming that we can explicitly enable/disable throttling, but that's not the case

mihir20 (approved):
LGTM

cisse21 (approved):

PR #6224, opened 2025-08-11
chore: implement configurable deduplication mirror mode

mihir20:
# Description

## Overview
Refactored the dedup package to use a single Dedup instance with pluggable database backends instead of creating multiple instances.

### 1. Core Architecture
- **Before**: Multiple Dedup instances (BadgerDB, KeyDB, Mirror)
- **After**: Single Dedup instance with pluggable DB backends

### 2. Interface Changes
- Added `DB` interface in `types` package for database operations
- Made BadgerDB and KeyDB implement the DB interface
- Created MirrorDB to handle mirroring logic internally

### 3. Main Dedup Structure
```go
// New simplified structure
type Dedup struct {
    db           DB
    uncommittedMu sync.RWMutex
    uncommitted  map[string]struct{}
}
```

#### BadgerDB (`services/dedup/badger/badger.go`)
- `NewBadgerDB` now returns `types.DB` instead of `*Dedup`
- Removed embedded Dedup struct
- Implements DB interface directly

#### KeyDB (`services/dedup/keydb/keydb.go`)
- `NewKeyDB` now returns `types.DB` instead of `*Dedup`
- Removed embedded Dedup struct
- Implements DB interface directly

#### MirrorDB (`services/dedup/mirror.go`)
- New implementation that wraps primary and mirror databases
- Handles asynchronous mirroring operations
- Implements both DB and DedupInterface interfaces

### 5. Configuration Modes
1. `badgerOnlyMode` - BadgerDB only
2. `keyDBOnlyMode` - KeyDB only
3. `mirrorBadgerMode` - BadgerDB primary, KeyDB mirror
4. `mirrorKeyDBMode` - KeyDB primary, BadgerDB mirror

### 6. Test Updates (`services/dedup/dedup_test.go`)
- Updated to use `types.DedupInterface` instead of `dedup.Dedup`
- Modified KeyDB test to use config to select backend
- Preserved all existing test logic and coverage

## Benefits
- Single Dedup instance management
- Cleaner separation of concerns
- Easier testing and maintenance
- Better extensibility for new database backends
- Preserved backward compatibility

## Linear Ticket

pipe-2289

## Security


services/dedup/mirror.go
```diff
@@ -139,12 +166,32 @@ func (m *mirror) printErrs(interval time.Duration) {
...
-	// first we need to stop all mirroring goroutines
-	close(m.stopPrintErrs)
-	_ = m.group.Wait()
-	close(m.errs)
-	// then can close both mirror and primary dedup services
-	m.mirror.Close()
-	m.Dedup.Close()
+// Implement the DedupInterface methods for MirrorDB
+
+func (m *MirrorDB) Allowed(batchKeys ...types.BatchKey) (map[types.BatchKey]bool, error) {
```
ktgowtham: why do we need this method here?

mihir20: not needed removed now, missed to push this earlier

ktgowtham (approved):

services/dedup/types/types.go
```diff
@@ -1,9 +1,39 @@
...
+	// Get returns a map of keys that exist in the database
+	Get(keys []string) (map[string]bool, error)
+	// Set stores keys in the database with a TTL
+	Set(keys []string) error
+	// Close closes the database connection
+	Close()
+}
+
+// DedupInterface is the interface for deduplication service
+type DedupInterface interface {
```
atzoum: Keep interface names clean, use Impl if there is a naming clash with the implementation struct and you need to keep the implementation struct exported
```suggestion
type Dedup interface {
```

processor/processor.go
```diff
@@ -46,7 +48,7 @@ import (
 	"github.com/rudderlabs/rudder-server/rruntime"
 	destinationdebugger "github.com/rudderlabs/rudder-server/services/debugger/destination"
 	transformationdebugger "github.com/rudderlabs/rudder-server/services/debugger/transformation"
-	"github.com/rudderlabs/rudder-server/services/dedup"
+	dedupTypes "github.com/rudderlabs/rudder-server/services/dedup/types"
```
atzoum: long long ago, we agreed that we would be using lowercase for import aliases
```suggestion
	deduptypes "github.com/rudderlabs/rudder-server/services/dedup/types"
```

services/dedup/dedup.go
```diff
@@ -1,17 +1,25 @@
...
 	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
-	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
-	"github.com/rudderlabs/rudder-server/services/dedup/badger"
-	"github.com/rudderlabs/rudder-server/services/dedup/keydb"
 	"github.com/rudderlabs/rudder-server/services/dedup/types"
 )
 
+type Dedup struct {
```
atzoum: doesn't need to be an exported type
```suggestion
type dedup struct {
```

services/dedup/db.go
```diff
@@ -0,0 +1,80 @@
...
+		primary, err := badger.NewBadgerDB(conf, stats, badger.DefaultPath())
+		if err != nil {
+			return nil, err
+		}
+		// Try to create keydb mirror
+		mirror, err := kdb.NewKeyDB(conf, stats, log)
+		if err != nil {
+			// If keydb fails, log and fallback to badger only
+			log.Errorn("failed to create keydb mirror, falling back to badger only", obskit.Error(err))
+			return primary, nil
```
atzoum: I don't think we need this fallback when local badgerdb is the mirror, we should simply return an error after closing the previous db we created

services/dedup/badger/badger.go
```diff
@@ -96,10 +95,11 @@ func NewBadgerDB(conf *config.Config, stat stats.Stats, path string) *Dedup {
 	db.stats.vlogSize = stat.NewTaggedStat("badger_db_size", stats.GaugeType, stats.Tags{"name": "dedup", "type": "vlog"})
 	db.stats.totSize = stat.NewTaggedStat("badger_db_size", stats.GaugeType, stats.Tags{"name": "dedup", "type": "total"})
 
-	return &Dedup{
-		badgerDB:    db,
-		uncommitted: make(map[string]struct{}),
+	err := db.init()
```
atzoum, line 97: since we are now calling `init` in the producer function, we no longer need the `sync.Once`

services/dedup/db.go
```diff
@@ -0,0 +1,80 @@
...
+func getMode(conf *config.Config) Mode {
+	mode := conf.GetString("Dedup.Mirror.Mode", string(badgerOnlyMode))
+	switch Mode(mode) {
+	case mirrorBadgerMode, mirrorKeyDBMode, keyDBOnlyMode:
+		return Mode(mode)
+	default:
+		// Default to mirrorBadgerMode if invalid mode specified
+		return badgerOnlyMode
+	}
+}
```
atzoum: do we really need this? We already have a `default` case in the statement below
```suggestion
```

services/dedup/db.go
```diff
@@ -0,0 +1,80 @@
...
+
+		mirror, err := badger.NewBadgerDB(conf, stats, badger.DefaultPath())
+		if err != nil {
+			// If badger fails, log and fallback to keydb only
+			log.Errorn("failed to create badger mirror, falling back to keydb only", obskit.Error(err))
+			return primary, nil
+		}
+		return NewMirrorDB(primary, mirror, mirrorKeyDBMode, conf, stats, log), nil
+	default:
+		// Default to badger only
```
atzoum, line 73: log a warning

services/dedup/db.go
```diff
@@ -0,0 +1,80 @@
...
+		return badgerOnlyMode
+	}
+}
+
+// NewDB creates a new DB implementation based on configuration
+func NewDB(conf *config.Config, stats stats.Stats, log logger.Logger) (types.DB, error) {
+	mode := getMode(conf)
+
+	switch mode {
+	case badgerOnlyMode:
```
atzoum, line 36: optional: log an info about the mode of dedub db that we are creating

services/dedup/mirror.go
```diff
@@ -13,14 +15,12 @@ import (
...
-const (
-	defaultMaxRoutines = 3000
-)
-
-type mirror struct {
-	Dedup
-	mirror Dedup
+const defaultMaxRoutines = 3000
 
+type MirrorDB struct {
```
atzoum: doesn't need to be exported
```suggestion
type mirrorDB struct {
```

services/dedup/badger/badger.go
```diff
@@ -96,10 +95,11 @@ func NewBadgerDB(conf *config.Config, stat stats.Stats, path string) *Dedup {
...
+	if err != nil {
+		return nil, fmt.Errorf("initializing badger db: %w", err)
 	}
+	return db, nil
 }
 
 func (d *BadgerDB) Get(keys []string) (map[string]bool, error) {
```
atzoum: doesn't need to be exported
```suggestion
func (d *badgerDB) Get(keys []string) (map[string]bool, error) {
```

services/dedup/keydb/keydb.go
```diff
@@ -51,7 +49,7 @@ func NewKeyDB(conf *config.Config, stat stats.Stats, log logger.Logger) (*Dedup,
 	db.stats.getTimer = stat.NewTaggedStat("dedup_get_duration_seconds", stats.TimerType, stats.Tags{"mode": "keydb"})
 	db.stats.setTimer = stat.NewTaggedStat("dedup_set_duration_seconds", stats.TimerType, stats.Tags{"mode": "keydb"})
 
-	return &Dedup{keyDB: db}, nil
+	return db, nil
 }
 
 func (d *KeyDB) Get(keys []string) (map[string]bool, error) {
```
atzoum: doesn't need to be exported
```suggestion
func (d *keyDB) Get(keys []string) (map[string]bool, error) {
```

services/dedup/mirror.go
```diff
@@ -44,85 +44,112 @@ func newMirror(d, m Dedup, conf *config.Config, s stats.Stats, log logger.Logger
...
 	}
-	dedupMirror.metrics.allowedErrorsCount = s.NewTaggedStat("dedup_mirroring_err_count", stats.CountType, stats.Tags{
-		"mode": "keydb",
-		"type": "allowed",
+
+	// Set up metrics
+	mirrorModeStr := "keydb"
+	if mode == mirrorBadgerMode {
+		mirrorModeStr = "badgerdb"
+	}
```
atzoum: We don't need this here since it shouldn't be aware of the various modes. Either use the original `Mode` string, or introduce a `MirrorLabel` method in `Mode` that knows how to return the label.
```suggestion
```

atzoum (commented):
thanks, looks lean and clean!

services/dedup/badger/badger.go
```diff
@@ -55,54 +53,55 @@ func DefaultPath() string {
...
-		WithBaseTableSize(conf.GetInt64Var(1*bytesize.MB, 1, "BadgerDB.Dedup.baseTableSize", "BadgerDB.baseTableSize")).
-		WithBaseLevelSize(conf.GetInt64Var(5*bytesize.MB, 1, "BadgerDB.Dedup.baseLevelSize", "BadgerDB.baseLevelSize")).
-		WithLevelSizeMultiplier(conf.GetIntVar(10, 1, "BadgerDB.Dedup.levelSizeMultiplier", "BadgerDB.levelSizeMultiplier")).
-		WithMaxLevels(conf.GetIntVar(7, 1, "BadgerDB.Dedup.maxLevels", "BadgerDB.maxLevels")).
-		WithNumCompactors(conf.GetIntVar(4, 1, "BadgerDB.Dedup.numCompactors", "BadgerDB.numCompactors")).
-		WithValueThreshold(conf.GetInt64Var(10*bytesize.B, 1, "BadgerDB.Dedup.valueThreshold", "BadgerDB.valueThreshold")).
-		WithSyncWrites(conf.GetBoolVar(false, "BadgerDB.Dedup.syncWrites", "BadgerDB.syncWrites")).
-		WithBlockCacheSize(conf.GetInt64Var(0, 1, "BadgerDB.Dedup.blockCacheSize", "BadgerDB.blockCacheSize")).
-		WithDetectConflicts(conf.GetBoolVar(false, "BadgerDB.Dedup.detectConflicts", "BadgerDB.detectConflicts"))
+		WithIndexCacheSize(conf.GetInt64Var(16*bytesize.MB, 1, "badgerDB.Dedup.indexCacheSize", "badgerDB.indexCacheSize")).
```
atzoum: we shouldn't have converted these to lowercase

services/dedup/db.go
```diff
@@ -0,0 +1,82 @@
...
+	mirrorBadgerMode mode = "mirrorBadger"
+	mirrorKeyDBMode  mode = "mirrorKeyDB"
+)
+
+func (m mode) MirrorLabel() string {
+	mirrorModeStr := "keydb"
+	if m == mirrorBadgerMode {
+		mirrorModeStr = "badgerdb"
+	}
+	return mirrorModeStr
```
atzoum: we need a case statement here, right? Some modes shouldn't have a mirror label

services/dedup/keydb/keydb.go
```diff
@@ -26,35 +24,35 @@ type KeyDB struct {
 	}
 }
 
-func NewKeyDB(conf *config.Config, stat stats.Stats, log logger.Logger) (*Dedup, error) {
-	dedupWindow := conf.GetReloadableDurationVar(3600, time.Second, "KeyDB.Dedup.dedupWindow", "Dedup.dedupWindow", "Dedup.dedupWindowInS")
-	nodeAddresses := conf.GetString("KeyDB.Dedup.Addresses", "")
+func NewKeyDB(conf *config.Config, stat stats.Stats, log logger.Logger) (types.DB, error) {
+	dedupWindow := conf.GetReloadableDurationVar(3600, time.Second, "keyDB.Dedup.dedupWindow", "Dedup.dedupWindow", "Dedup.dedupWindowInS")
+	nodeAddresses := conf.GetString("keyDB.Dedup.Addresses", "")
```
atzoum: this shouldn't be lowercase

services/dedup/db.go
```diff
@@ -0,0 +1,78 @@
...
+		primary, err := badger.NewBadgerDB(conf, stats, badger.DefaultPath())
+		if err != nil {
+			return nil, fmt.Errorf("create badger primary: %w", err)
+		}
+		// try to create keydb mirror
+		mirror, err := kdb.NewKeyDB(conf, stats, log)
+		if err != nil {
+			primary.Close()
+			return nil, fmt.Errorf("create keydb mirror: %w", err)
+		}
```
atzoum: In the scenario where keydsb is the mirror, I believe we should preserve the old behaviour where we were falling back to simple badger and logging a warning, wdyt?

mihir20: yeah makes sense

services/dedup/db.go
mihir20: doing that on line #36

PR #6225, opened 2025-08-12
feat(router): delivery throttler

atzoum:
# Description

Introducing a new layer of throttling in router additional to the existing pickup-level throttling, dubbed **delivery throttling**:
- Delivery throttling can be enabled at a destination endpoint level.
- Delivery throttling will delay the request in order to respect the configured throttling rates.
- Delivery throttling configuration is controlled through the `endpointLabel` field of each transformed job.
- We are not using the actual endpoint to avoid cardinality issues, e.g. different eu vs us endpoints or endpoints having various different query & path params.
- In a typical scenario, we will be using the delivery throttler for endpoints that have low rate limits (e.g. <100 rps) and we will be leveraging a distributed throttling algorithm, so that we respect throttling limits across multiple pods.

**Delivery Throttling Implementation:**

* Added a new delivery throttler mechanism to the router, allowing rate-limited event delivery to destination endpoints based on configurable limits and time windows. This includes support for a new `EndpointLabel` field in `PostParametersT` and relevant configuration keys. [[1]](diffhunk://#diff-704f2b606c8548f91a99db822a0ec0323b4c780a1e9b7b62cada581b79f9ff3dR76) [[2]](diffhunk://#diff-9f951c838d013ed1a4f801b99e1a425f9005fbb96d71c759843b68352e71f6ebL11-R17) [[3]](diffhunk://#diff-737c256269dad4dea6ddd0cc49f6ddd112d7fe76eee6569cf552aa6a7ae7d0a3R121)

**Integration Testing:**

* Introduced an integration test in `router/delivery_throttler_test.go` that verifies the router respects delivery rate limits for webhook events, including burst handling and timing tolerances.
* Added a test configuration template file `router/testdata/deliveryThrottlerTestTemplate.json` to support the integration test setup.

**Throttler Factory Refactor:**

* Refactored the throttler factory interface and usage to distinguish between pickup and delivery throttlers, updating method names and call sites for clarity and correctness. [[1]](diffhunk://#diff-04def44a4cd6a9c8549a994b1738c32da5163b39623f56abea1b20ae1ea14428L180-R187) [[2]](diffhunk://#diff-704f2b606c8548f91a99db822a0ec0323b4c780a1e9b7b62cada581b79f9ff3dL365-R366) [[3]](diffhunk://#diff-704f2b606c8548f91a99db822a0ec0323b4c780a1e9b7b62cada581b79f9ff3dL708-R709)

**Worker Lifecycle Improvements:**

* Improved worker lifecycle management by introducing context cancellation for partition workers, ensuring clean shutdowns and better resource handling. [[1]](diffhunk://#diff-a6dd6410e9533eb3d13a9ec4d47cdf0183abc8b75102b82d701691e8d9c28d9eR28-R34) [[2]](diffhunk://#diff-a6dd6410e9533eb3d13a9ec4d47cdf0183abc8b75102b82d701691e8d9c28d9eR95)

**Dependency and Minor Cleanup:**

* Removed unused `containerd/containerd` dependencies from `go.mod` for cleanup. [[1]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L9) [[2]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L196)
* Added missing import for `time` in `router/throttler/factory.go`.

## Linear Ticket

resolves PIPE-2277

## Security

### DIFF
--- a/router/worker.go
+++ b/router/worker.go
@@ -346,7 +346,7 @@ func (w *worker) process(destinationJobs []types.DestinationJobT) {
 		limiterStats.Update(w.partition, time.Since(start), successCount+errorCount, errorCount)
 	}()
 
-	ctx := context.TODO()
+	ctx := context.TODO() // TODO: use w.ctx and handle graceful shutdown scenario
 
 	transformerProxy := w.rt.reloadableConfig.transformerProxy.Load()
 
@@ -492,7 +492,7 @@ func (w *worker) process(destinationJobs []types.DestinationJobT) {
 							destType := destinationJob.Destination.DestinationDefinition.Name
 							deliveryThrottlerTimeout := w.rt.deliveryThrottlerTimeout.Load()
 							deliveryThrottler := w.rt.throttlerFactory.GetDeliveryThrottler(destType, destinationID, endpointLabel)
-							waitCtx, cancel := context.WithTimeoutCause(ctx, deliveryThrottlerTimeout, errors.New("delivery throttler timeout after "+deliveryThrottlerTimeout.String()))
+							waitCtx, cancel := context.WithTimeoutCause(w.ctx, deliveryThrottlerTimeout, errors.New("delivery throttler timeout after "+deliveryThrottlerTimeout.String()))
 							_, err := deliveryThrottler.Wait(waitCtx)
 							cancel()
 							if err != nil && waitCtx.Err() == nil {


processor/integrations/integrations.go
```diff
@@ -8,14 +8,13 @@ import (
 )
 
 // PostParametersT is a struct for holding all the values from transformerResponse and use them to publish an event to a destination
-// optional is a custom tag introduced by us and is handled by GetMandatoryJSONFieldNames. Its intentionally added
```
atzoum, line 11: _**Note:**_ this kind of validation was removed sometime ago, we no longer need the comment and the tag

router/worker.go
```diff
@@ -42,16 +42,16 @@ type worker struct {
...
+	inputCh    chan workerJob     // the worker's input channel
 
 	inputReservations int                 // number of slots reserved in the input channel
 	barrier           *eventorder.Barrier // barrier to ensure ordering of events
 
 	deliveryTimeStat          stats.Measurement
 	routerDeliveryLatencyStat stats.Measurement
 	routerProxyStat           stats.Measurement
-	latestAssignedTime        time.Time
-	processingStartTime       time.Time
```
atzoum, line 52: _**Note:**_ these were practically unused, leftovers from previous cleanups of unused functionalities

router/worker.go
```diff
@@ -359,379 +359,383 @@ func (w *worker) batchTransform(routerJobs []types.RouterJobT) []types.Destinati
...
+						var respBodyTemp string
+						var respBodyTemps map[int64]string
+						respBodyArr := make([]string, 0)
+						respBodyArrs := make([]map[int64]string, 0)
+						for i, val := range result {
+							w.logger.Debugn(`responseTransform status`,
+								obskit.DestinationType(w.rt.destType),
+								logger.NewBoolField("transformerProxy", w.rt.reloadableConfig.transformerProxy.Load()))
+							errorAt = routerutils.ERROR_AT_DEL
+							if val.EndpointLabel != "" {
```
atzoum: **_TODO:_** use the `default` literal as an `EndpointLabel` if it is empty. This will allow us to use a global delivery throttler for any destination endpoints that don't have an explicit endpoint label

router/throttler/internal/delivery/throttler.go
```diff
@@ -0,0 +1,118 @@
...
+	return time.Since(start), nil
+}
+
+func (t *throttler) enabled() bool {
+	limit := t.limit.Load()
+	window := t.window.Load()
+	return limit > 0 && window > 0
+}
+
+func (t *throttler) GetLimit() int64 {
```
fracasula: Isn't this just used in the test? If yes we can unexport it like the other methods.

router/throttler/internal/delivery/throttler_test.go
```diff
@@ -0,0 +1,910 @@
...
+	"github.com/rudderlabs/rudder-go-kit/config"
+	"github.com/rudderlabs/rudder-go-kit/logger"
+	"github.com/rudderlabs/rudder-go-kit/stats/memstats"
+)
+
+func TestDeliveryThrottler(t *testing.T) {
+	t.Run("NewThrottler", func(t *testing.T) {
+		t.Run("with destination id configuration", func(t *testing.T) {
+			config := config.New()
+			statsStore, err := memstats.New()
```
fracasula: If you're not making any assertions on any metric, you can use `stats.NOP`, it would also make it easier to go through the test.

router/throttler/internal/delivery/throttler_test.go
```diff
@@ -0,0 +1,910 @@
...
+			metrics := statsStore.GetByName("delivery_throttling_rate_limit")
+			require.NotEmpty(t, metrics)
+
+			// Find the metric with correct tags and verify it has value 0 (not updated)
+			found := false
+			for _, metric := range metrics {
+				if metric.Tags["destinationId"] == destinationID &&
+					metric.Tags["destType"] == destType &&
+					metric.Tags["endpointLabel"] == endpointLabel {
+					require.Equal(t, float64(0), metric.Value) // Should remain 0
```
fracasula: [optional] We can have a verbose error message instead than a comment. Might be more useful.

```suggestion
					require.Equalf(t, float64(0), metric.Value, 
						"Gauge value of %s should be 0 when window is 0 (disabled)", metric.Name)
```

router/throttler/internal/delivery/throttler_test.go
```diff
@@ -0,0 +1,910 @@
...
+					// Just verify that some timing was recorded, timing precision may vary
+					require.GreaterOrEqual(t, metric.Value, float64(0))
+					found = true
+					break
+				}
+			}
+			require.True(t, found, "Should find success timer metric")
+		})
+
+		t.Run("does not record timers for disabled throttler", func(t *testing.T) {
```
fracasula: I would suggest changing this test name as I found it misleading because we're still recording timers, it's just that they will have minimal values. How about `records minimal timers for disabled throttler`?

router/throttler/internal/delivery/types.go
```diff
@@ -0,0 +1,16 @@
+package delivery
+
+import (
+	"context"
+	"time"
+
+	"github.com/rudderlabs/rudder-go-kit/logger"
+)
+
+type Logger interface {
```
fracasula: I wonder if we really need this type :thinking:

PR #6232, opened 2025-08-13
chore: stop collecting errors in proc_error jobsdb

atzoum:
# Description

We are no longer be collecting errors in `proc_error` jobsdb as we are planning to remove error collection for debugging purposes in the near future. 

What was happening until now:

- rudder-server was collecting failed jobs from webhook, processor steps, router, and batchrouter, and was storing these failed jobs in the `proc_error` JobsDB.
- periodically, the `stash` processor component was creating backup files from the `proc_error` JobsDB and uploading them to `rudder-proc-err-logs` in object storage.

In the past, we relied on `proc_error` jobs for replays, but that’s no longer the case. At this point, it feels like we’re spending unnecessary resources collecting and storing all this information which no-one actually needs, using as an excuse that it is useful for debugging purposes…

As a first step we are stopping collection and storage of failed jobs to `proc_error`. In a follow up step we will be removing all relevant `proc_error` code altogether.

## Linear Ticket

resolves PIPE-2295

## Security

### DIFF
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -200,7 +200,7 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, options *app.Options)
 	defer batchRouterDB.Close()
 
 	// We need two errorDBs, one in read & one in write mode to support separate gateway to store failures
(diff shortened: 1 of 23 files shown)


mihir20 (approved):
LGTM

PR #6233, opened 2025-08-13
fix: handling empty schema for table level schemas

achettyiitr:
# Description

- Do validation of table-level schemas only in case if some schema is present. 

## Linear Ticket

- Resolves WAR-1017

## Security

### DIFF
--- a/warehouse/internal/repo/schema.go
+++ b/warehouse/internal/repo/schema.go
@@ -231,7 +231,7 @@ func (sh *WHSchema) GetForNamespace(ctx context.Context, destID, namespace strin
 	if err != nil {
 		return model.WHSchema{}, err
 	}
-	if !reflect.DeepEqual(originalSchema.Schema, tableLevelSchemas) {
+	if len(originalSchema.Schema) > 0 && !reflect.DeepEqual(originalSchema.Schema, tableLevelSchemas) {
 		return model.WHSchema{}, errors.New("parent schema does not match parent schema")
 	}
 	return originalSchema, nil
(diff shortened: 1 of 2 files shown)


warehouse/internal/repo/schema.go
```diff
@@ -231,7 +231,7 @@ func (sh *WHSchema) GetForNamespace(ctx context.Context, destID, namespace strin
 	if err != nil {
 		return model.WHSchema{}, err
 	}
-	if !reflect.DeepEqual(originalSchema.Schema, tableLevelSchemas) {
+	if len(originalSchema.Schema) > 0 && !reflect.DeepEqual(originalSchema.Schema, tableLevelSchemas) {
```
shekhar-rudder: If schema is empty, should we do an early return?

achettyiitr: It should be fine, as we are returning in the next line. We can avoid 1 nested block.

shekhar-rudder: I meant that we can return from [here](https://github.com/rudderlabs/rudder-server/pull/6233/files#diff-28d0a99033d94b0ec35c73d633a6ed0f832f2c8dc0b619609881b818933dce72R213). No need to [query](https://github.com/rudderlabs/rudder-server/pull/6233/files#diff-28d0a99033d94b0ec35c73d633a6ed0f832f2c8dc0b619609881b818933dce72R219) the table for table level schemas

achettyiitr: Done.

shekhar-rudder (approved):

PR #6234, opened 2025-08-13
fix: table level schemas

achettyiitr:
# Description

- Table level schemas

## Linear Ticket

- Resolves WAR-1022

## Security


warehouse/internal/repo/schema.go
```diff
@@ -185,8 +185,8 @@ func (sh *WHSchema) Insert(ctx context.Context, whSchema *model.WHSchema) error
 					whSchema.DestinationID,
 					whSchema.DestinationType,
 					tableSchemaPayload,
-					whSchema.CreatedAt.UTC(),
-					whSchema.UpdatedAt.UTC(),
+					now.UTC(),
```
shekhar-rudder, line 188: If the row already exists, we shouldn't be updating `CreatedAt`, rt ?

runner/buckets.go
```diff
@@ -33,6 +33,15 @@ var (
 			float64(1 * bytesize.MB), float64(3 * bytesize.MB), float64(5 * bytesize.MB), float64(10 * bytesize.MB),
 			float64(25 * bytesize.MB), float64(50 * bytesize.MB), float64(100 * bytesize.MB),
 		},
+		"schema_snapshot_patch_size": {
```
shekhar-rudder, line 36: What happens if we don't add the stat here? Does it take some default value?

achettyiitr: Yeah. Defaults are defined [here](https://github.com/rudderlabs/rudder-server/blob/master/runner/buckets.go#L8-L13).

warehouse/internal/repo/schema.go
achettyiitr: It's for the insert case, for update we are only setting these: `schema = $5, updated_at = $7,expires_at = $8;`

shekhar-rudder: Ok. I didn't understand what is the issue with the current code

achettyiitr: For all the values that are inserted/updated, they should have the same timestamp instead of using the previous one.

shekhar-rudder (approved):

RanjeetMishra (approved):

PR #6236, opened 2025-08-14
fix: handle snowpipe service restarts in destination polling logic

shekhar-rudder:
# Description

Implement enhanced batch router polling logic that can detect lost events and mark them for retry:

## Linear Ticket

https://linear.app/rudderstack/issue/WAR-1016/enhance-getimportstatus-logic-to-handle-snowpipe-restarts-and-lost

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
@@ -537,6 +537,8 @@ func (m *Manager) processPollImportInfos(ctx context.Context, infos []*importInf
 		if _, alreadyProcessed := m.polledImportInfoMap[info.ChannelID]; alreadyProcessed {
 			continue
 		}
+		info.FailedJobIds = nil
+		info.Failed = false
 
 		inProgress, err := m.getImportStatus(ctx, info)
 		if err != nil {
@@ -545,8 +547,6 @@ func (m *Manager) processPollImportInfos(ctx context.Context, infos []*importInf
 			m.polledImportInfoMap[info.ChannelID] = info
 			continue
 		}
-		info.FailedJobIds = nil
-		info.Failed = false
 		if !inProgress {
 			m.polledImportInfoMap[info.ChannelID] = info
 		}
@@ -602,14 +602,8 @@ func (m *Manager) getImportStatus(ctx context.Context, info *importInfo) (bool, 
 	// Case 2: Events lost - restart/error scenario
 	if latestInsertedOffset < expectedOffset {
 		log.Infon("Events lost due to Snowpipe restart or error")
-		var start int64
-		if statusRes.Offset == "" {
-			start = 1
-		} else {
-			start = latestCommittedOffset + 1
-		}
 		info.FailedJobIds = &failedJobIds{
-			Start: start,
+			Start: latestCommittedOffset + 1,
 			End:   expectedOffset,
 		}
 		return false, fmt.Errorf("events lost: latestCommittedOffset=%d, latestInsertedOffset=%d, expectedOffset=%d",
@@ -629,6 +623,8 @@ func (m *Manager) getImportStatus(ctx context.Context, info *importInfo) (bool, 
 
 func convertToInt(a string) (int64, error) {
 	if a == "" {
+		// Using math.MinInt64 instead of 0 to avoid skipping events
+		// with negative job ids in case of downscaling of nodes
 		return math.MinInt64, nil
 	}
 	ai, err := strconv.ParseInt(a, 10, 64)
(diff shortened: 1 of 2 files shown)


router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -565,16 +568,74 @@ func (m *Manager) getImportStatus(ctx context.Context, info *importInfo) (bool,
...
+	}
+
+	// Case 1: All events have been flushed - proceed to next batch
+	if latestCommittedOffset == expectedOffset {
+		log.Infon("All events have been flushed successfully")
+		return false, nil
+	}
+
+	// Case 2: Events lost - restart/error scenario
+	if latestInsertedOffset < expectedOffset {
```
vyeshwanth, line 603: `latestInsertedOffset`  - is this from snowflake or the snowpipe service?

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -544,6 +545,8 @@ func (m *Manager) processPollImportInfos(ctx context.Context, infos []*importInf
 			m.polledImportInfoMap[info.ChannelID] = info
 			continue
 		}
+		info.FailedJobIds = nil
+		info.Failed = false
```
RanjeetMishra: @shekhar-rudder Just by reading this func code, I could not understand why these should not be set at start of for loop vs here?

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
shekhar-rudder: This is from snowpipe service (in-memory state). https://www.notion.so/rudderstacks/Snowpipe-service-restart-event-backlog-24cf2b415dd08024b02ee0e9b53029e8?source=copy_link#24df2b415dd080bdb9e0dc3c5aa106e9

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -565,16 +568,74 @@ func (m *Manager) getImportStatus(ctx context.Context, info *importInfo) (bool,
...
+
+	// Case 2: Events lost - restart/error scenario
+	if latestInsertedOffset < expectedOffset {
+		log.Infon("Events lost due to Snowpipe restart or error")
+		var start int64
+		if statusRes.Offset == "" {
+			start = 1
+		} else {
+			start = latestCommittedOffset + 1
+		}
```
RanjeetMishra: In convertToInt, wondering if its fine to use 0 instead of `math.MinInt64`. With this, it could be just start = latestCommittedOffset + 1? Or fine to consider `-1` as well as sentinel return of convertToInt.

I was thinking if we are already processing statsRes in custom convertToInt, do we need another check on raw `statusRes.Offset` vs being able to use the converted offset response.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
shekhar-rudder: That is possible. I can set these values just before calling the `getImportStatus` method

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
RanjeetMishra: Yes
`If no events have been inserted since the service started, set latest_inserted_offset = latest_committed_offset.` this would be important that its being set.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
shekhar-rudder: The reason I didn't go with value 0, instead of `math.MinInt64` to handle negative job ids in the case of downscaling. But I don't think we need the if case. `start = latestCommittedOffset + 1` should work for all the scenarios.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
shekhar-rudder: Done

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
shekhar-rudder: Done

PR #6240, opened 2025-08-18
chore: remove event blocking flag

itsmihir:
# Description

This PR removes the event blocking flag; now the feature is totally controlled by backend config.

## Linear Ticket

Resolves OBS-882

## Security

### DIFF
--- a/gateway/handle.go
+++ b/gateway/handle.go
@@ -130,7 +130,6 @@ type Handle struct {
 		gwAllowPartialWriteWithErrors        config.ValueLoader[bool]
 		enableInternalBatchValidator         config.ValueLoader[bool]
 		enableInternalBatchEnrichment        config.ValueLoader[bool]
-		enableEventBlocking                  config.ValueLoader[bool]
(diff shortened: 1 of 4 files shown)


gateway/handle.go
```diff
@@ -609,10 +608,6 @@ func (gw *Handle) memoizedIsEventBlocked() func(workspaceID, sourceID, eventType
 
 // isEventBlocked checks if an event should be blocked based on workspace settings
 func (gw *Handle) isEventBlocked(workspaceID, sourceID, eventType, eventName string) bool {
-	if !gw.conf.enableEventBlocking.Load() {
-		return false
-	}
-
```
satishrudderstack, line 615: @itsmihir What happens if `eventType` key is not present in the map in line 623 below ?

`return gw.blockedEventsWorkspaceTypeNameMap[workspaceID][eventType][eventName]`

itsmihir: reading from map is safe

PR #6248, opened 2025-08-18
feat: implement error normalizer with similarity detection for reporting

koladilip:
## Description

This PR implements error normalization for enterprise error reporting to prevent error message spam while preserving error occurrence counts. The implementation uses a configurable error normalizer that groups similar errors and applies rate limiting per connection group.

### 🔍 **Error Normalization**
- **Similarity-based grouping** using LCS algorithm to detect and group similar error messages
- **Configurable similarity threshold** (default: 0.75)
- **Canonical error messages**: Returns existing similar errors to preserve count information
- **Rate limiting per connection group**: Limits unique errors per `sourceId:destinationId:eventType:pu` combination

### 🚀 **Rate Limiting**
- **Configurable rate limiting** per connection group key (`sourceId:destinationId:eventType:pu`)
- **Maximum errors per group** per connection (default: 20)
- **Maximum groups** to prevent memory leaks (default: 10,000)
- **Non-blocking behavior**: Returns "RedactedError" when limits are exceeded

### 🧹 **Memory Management**
- **Automatic cleanup** of expired error messages (older than 1 minute)
- **Configurable cleanup intervals** to prevent memory leaks (default: 5 seconds)
- **Smart group eviction**: Removes groups that have reached limits and been inactive for >1 minute
- **Timestamp-based tracking**: Tracks when errors were last seen and when groups were last blocked

### 🔧 **Interface-based Design**
- **ErrorNormalizer interface** for easy testing and extensibility
- **Mock support** for isolated component testing using mockgen
- **Clean separation** of concerns between normalization and reporting

## Configuration

```yaml
Reporting:
  errorReporting:
    normalizer:
      similarityThreshold: 0.75       # Similarity threshold for message grouping (0.0-1.0)
      maxErrorsPerGroup: 20          # Maximum unique errors per connection group
      maxGroups: 10000              # Maximum number of connection groups
      cleanupInterval: "5s"          # How often to run cleanup
```

### Error Normalization Process

1. **Connection Grouping**: Errors are grouped by `sourceId:destinationId:eventType:pu`
2. **Exact Match Check**: First checks for exact error message matches
3. **Similarity Detection**: Uses LCS algorithm to find similar error messages within each group
4. **Rate Limiting**: Limits unique errors per connection group to prevent spam
5. **Cleanup**: Automatically removes expired error messages and inactive groups

### Integration with Error Reporting

* **Seamless integration** with existing error reporting pipeline
* **Early exit conditions** prevent unnecessary processing
* **Preserves error counts** while reducing cardinality
* **Background cleanup** runs independently to manage memory

### Error Group Structure

Each error group maintains:
- **Error messages map**: Maps error messages to their last seen timestamp
- **Last updated timestamp**: When the group's error set last changed
- **Last blocked timestamp**: When the group was last rate limited

## Testing

* **Unit tests** for error normalization logic with table-driven tests
* **Similarity detection tests** with various thresholds and edge cases
* **Rate limiting tests** for connection group limits
* **Cleanup functionality tests** for expired messages and group eviction
* **Mock-based tests** for isolated component testing
* **Integration tests** with error reporting pipeline
* **Concurrent access tests** for thread safety

## Breaking Changes

None. This is a new feature that can be enabled/disabled via configuration.

## Performance Impact

* **Positive**: Reduces database load by limiting error report cardinality
* **Minimal overhead**: Early exit conditions prevent unnecessary processing
* **Memory efficient**: Automatic cleanup prevents memory leaks
* **Thread-safe**: Uses proper synchronization for concurrent access
* **Background cleanup**: Non-blocking cleanup process

## Linear Ticket

INT-4021: Create error rate limiter with configurable truncation

### DIFF
--- a/enterprise/reporting/error_normalizer.go
+++ b/enterprise/reporting/error_normalizer.go
@@ -30,8 +30,88 @@ const (
 	RedactedError = "RedactedError"
 )
 
+type errorEntry struct {
+	message string
+	time    time.Time
+}
+
+type boundedErrorSet struct {
+	entries []*errorEntry          // oldest → newest
+	index   map[string]*errorEntry // message → entry
+}
+
+func newBoundedErrorSet(max int) *boundedErrorSet {
+	return &boundedErrorSet{
+		entries: make([]*errorEntry, 0, max),
+		index:   make(map[string]*errorEntry, max),
+	}
+}
+
+// GetExact looks up exact match
+func (s *boundedErrorSet) GetExact(msg string, now time.Time) (*errorEntry, bool) {
+	if entry, ok := s.index[msg]; ok {
+		entry.time = now
+		s.moveToEnd(entry)
+		return entry, true
+	}
+	return nil, false
+}
+
+// GetSimilar scans for a near match (caller supplies similarity fn)
+func (s *boundedErrorSet) GetSimilar(msg string, now time.Time, similar func(a, b string) bool) (*errorEntry, bool) {
+	for i, entry := range s.entries {
+		if similar(msg, entry.message) {
+			entry.time = now
+			s.moveToEndAt(i)
+			return entry, true
+		}
+	}
+	return nil, false
+}
+
+// Add inserts a new entry, evicting oldest if full
+func (s *boundedErrorSet) Add(msg string, now time.Time) *errorEntry {
+	entry := &errorEntry{message: msg, time: now}
+	s.entries = append(s.entries, entry)
+	s.index[msg] = entry
+	return entry
+}
+
+// DropStale removes messages older than cutoff
+func (s *boundedErrorSet) DropStale(cutoff time.Time) {
+	var drop int
+	for _, entry := range s.entries {
+		if entry.time.Before(cutoff) {
+			delete(s.index, entry.message)
+			drop++
+		} else {
+			break
+		}
+	}
+	if drop > 0 {
+		s.entries = s.entries[drop:]
+	}
+}
+
+// --- internal helpers ---
+
+func (s *boundedErrorSet) moveToEnd(entry *errorEntry) {
+	for i, en := range s.entries {
+		if en == entry {
+			s.moveToEndAt(i)
+			return
+		}
+	}
+}
+
+func (s *boundedErrorSet) moveToEndAt(i int) {
+	entry := s.entries[i]
+	s.entries = append(s.entries[:i], s.entries[i+1:]...)
+	s.entries = append(s.entries, entry)
+}
+
 type errorGroup struct {
-	errors map[string]time.Time
+	errors boundedErrorSet
 	// lastUpdated is the time when the group's error set last changed (new error added or old error removed)
 	lastUpdated time.Time
 	// lastBlocked is the time when the error group was last blocked due to rate limiting
@@ -48,6 +128,7 @@ type errorNormalizer struct {
 	maxErrorsPerGroup   config.ValueLoader[int]
 	maxGroups           config.ValueLoader[int]
 	cleanupInterval     config.ValueLoader[time.Duration]
+	staleTime           config.ValueLoader[time.Duration]
 
 	// Stats manager reference
 	statsManager *ErrorReportingStats
@@ -62,52 +143,55 @@ func NewErrorNormalizer(log logger.Logger, conf *config.Config, statsInstance st
 		maxErrorsPerGroup:   conf.GetReloadableIntVar(20, 1, "Reporting.errorReporting.normalizer.maxErrorsPerGroup"),
 		maxGroups:           conf.GetReloadableIntVar(10000, 1, "Reporting.errorReporting.normalizer.maxGroups"),
 		cleanupInterval:     conf.GetReloadableDurationVar(5, time.Second, "Reporting.errorReporting.normalizer.cleanupInterval"),
+		staleTime:           conf.GetReloadableDurationVar(1, time.Minute, "Reporting.errorReporting.normalizer.staleTime"),
 		statsManager:        statsManager,
 	}
 }
 
 // NormalizeError normalizes the error message for the given connection key
 func (e *errorNormalizer) NormalizeError(ctx context.Context, errorDetailGroupKey types.ErrorDetailGroupKey, msg string) string {
-	// Convert struct key to string for internal use
 	now := time.Now()
 
 	e.mu.Lock()
 	defer e.mu.Unlock()
 
-	c, ok := e.groups[errorDetailGroupKey]
+	maxErrorsPerGroup := e.maxErrorsPerGroup.Load()
+	currentGroup, ok := e.groups[errorDetailGroupKey]
 	if !ok {
 		if len(e.groups) >= e.maxGroups.Load() {
 			return RedactedError
 		}
-		c = &errorGroup{lastUpdated: now, errors: make(map[string]time.Time)}
-		e.groups[errorDetailGroupKey] = c
+		currentGroup = &errorGroup{
+			lastUpdated: now,
+			errors:      *newBoundedErrorSet(maxErrorsPerGroup),
+		}
+		e.groups[errorDetailGroupKey] = currentGroup
 	}
 
 	// Exact match
-	if _, ok := c.errors[msg]; ok {
-		c.errors[msg] = now
+	if _, ok := currentGroup.errors.GetExact(msg, now); ok {
 		return msg
 	}
 
 	// Similarity match
 	similarityThreshold := e.similarityThreshold.Load()
-	for existing := range c.errors {
-		if lcs.Similarity(msg, existing) >= similarityThreshold {
-			c.errors[existing] = now
-			return existing
-		}
+	similarityFunc := func(a, b string) bool {
+		return lcs.Similarity(a, b) >= similarityThreshold
+	}
+
+	if entry, ok := currentGroup.errors.GetSimilar(msg, now, similarityFunc); ok {
+		return entry.message
 	}
 
 	// Rate limit the new error message
-	maxErrorsPerGroup := e.maxErrorsPerGroup.Load()
-	if len(c.errors) >= maxErrorsPerGroup {
-		c.lastBlocked = now
+	if len(currentGroup.errors.entries) >= maxErrorsPerGroup {
+		currentGroup.lastBlocked = now
 		return RedactedError
 	}
 
 	// Insert the new error message
-	c.errors[msg] = now
-	c.lastUpdated = now
+	currentGroup.lastUpdated = now
+	currentGroup.errors.Add(msg, now)
 	return msg
 }
 
@@ -121,39 +205,31 @@ func (e *errorNormalizer) cleanup() {
 	defer e.mu.Unlock()
 
 	now := time.Now()
-	cutoff := now.Add(-time.Minute)
+	cutoff := now.Add(-e.staleTime.Load())
 
 	maxErrorsPerGroup := e.maxErrorsPerGroup.Load()
 
-	for k, c := range e.groups {
-		e.dropStaleMessages(c, cutoff)
+	for k, currentGroup := range e.groups {
+		currentGroup.errors.DropStale(cutoff)
 
-		// Drop if no messages left OR no new unique error for >1m
-		if e.shouldDropCounter(c, now, maxErrorsPerGroup) {
+		// Drop if no messages left OR no new unique error for >staleTime
+		if e.shouldDropCounter(currentGroup, now, maxErrorsPerGroup) {
 			delete(e.groups, k)
 		}
 	}
 }
 
-func (e *errorNormalizer) dropStaleMessages(c *errorGroup, cutoff time.Time) {
-	for msg, t := range c.errors {
-		if t.Before(cutoff) {
-			delete(c.errors, msg)
-			c.lastUpdated = time.Now()
-		}
-	}
-}
-
-func (e *errorNormalizer) shouldDropCounter(c *errorGroup, now time.Time, maxErrorsPerGroup int) bool {
+func (e *errorNormalizer) shouldDropCounter(currentGroup *errorGroup, now time.Time, maxErrorsPerGroup int) bool {
 	// Drop if no messages left
-	if len(c.errors) == 0 {
+	if len(currentGroup.errors.entries) == 0 {
 		return true
 	}
-	// Drop if counter has reached maxErrorsPerGroup and no changes to errors for >1m and
-	// an error has been blocked in the last minute, means this counter is starving other errors
-	if len(c.errors) == maxErrorsPerGroup &&
-		now.Sub(c.lastUpdated) > time.Minute &&
-		now.Sub(c.lastBlocked) < time.Minute {
+	// Drop if counter has reached maxErrorsPerGroup and no changes to errors for >staleTime and
+	// an error has been blocked in the last staleTime, means this counter is starving other errors
+	staleTime := e.staleTime.Load()
+	if len(currentGroup.errors.entries) == maxErrorsPerGroup &&
+		now.Sub(currentGroup.lastUpdated) > staleTime &&
+		now.Sub(currentGroup.lastBlocked) < staleTime {
 		return true
 	}
 	return false
--- a/enterprise/reporting/error_reporting.go
+++ b/enterprise/reporting/error_reporting.go
@@ -244,7 +244,7 @@ func (edr *ErrorDetailReporter) DatabaseSyncer(c types.SyncerConfig) types.Repor
 	}
 }
 
-func (edr *ErrorDetailReporter) emitLagMetric(ctx context.Context, c types.SyncerConfig, lastReportedAtTime *atomic.Time) error {
+func (edr *ErrorDetailReporter) emitLagMetric(ctx context.Context, lastReportedAtTime *atomic.Time) error {
 	// for monitoring reports pileups
 	for {
 		lag := time.Since(lastReportedAtTime.Load())
@@ -479,7 +479,7 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
 	})
 
 	g.Go(func() error {
-		return edr.emitLagMetric(ctx, c, &lastReportedAtTime)
+		return edr.emitLagMetric(ctx, &lastReportedAtTime)
 	})
 
 	g.Go(func() error {
@@ -624,7 +624,7 @@ func (edr *ErrorDetailReporter) mainLoop(ctx context.Context, c types.SyncerConf
 	}
 }
 
-func (edr *ErrorDetailReporter) vacuum(ctx context.Context, dbHandle *sql.DB, tags stats.Tags) error {
+func (edr *ErrorDetailReporter) vacuum(ctx context.Context, dbHandle *sql.DB, _ stats.Tags) error {
 	defer edr.statsManager.VacuumDuration.RecordDuration()()
 	var query string
 	var full bool
@@ -840,7 +840,7 @@ func (edr *ErrorDetailReporter) aggregate(reports []*types.EDReportsDB) []*types
 }
 
 // DEPRECATED: Remove this after migration to commonClient, use edr.commonClient.Send instead.
-func (edr *ErrorDetailReporter) sendMetric(ctx context.Context, label string, metric *types.EDMetric) error {
+func (edr *ErrorDetailReporter) sendMetric(ctx context.Context, _ string, metric *types.EDMetric) error {
 	if edr.useCommonClient.Load() {
 		return edr.commonClient.Send(ctx, metric)
 	}
--- a/enterprise/reporting/error_normalizer_test.go
+++ b/enterprise/reporting/error_normalizer_test.go
@@ -433,3 +433,551 @@ func TestErrorNormalizer_ComprehensiveCleanupCoverage(t *testing.T) {
 	result = ern.NormalizeError(context.Background(), testKey2, "error5")
 	require.Equal(t, RedactedError, result)
 }
+
+func TestBoundedErrorSet_NewBoundedErrorSet(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name     string
+		max      int
+		expected *boundedErrorSet
+	}{
+		{
+			name: "valid max size",
+			max:  10,
+			expected: &boundedErrorSet{
+				entries: make([]*errorEntry, 0, 10),
+				index:   make(map[string]*errorEntry, 10),
+			},
+		},
+		{
+			name: "zero max size",
+			max:  0,
+			expected: &boundedErrorSet{
+				entries: make([]*errorEntry, 0),
+				index:   make(map[string]*errorEntry, 0),
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			result := newBoundedErrorSet(tc.max)
+
+			require.Equal(t, len(tc.expected.entries), len(result.entries))
+			require.Equal(t, cap(tc.expected.entries), cap(result.entries))
+			require.Equal(t, len(tc.expected.index), len(result.index))
+		})
+	}
+}
+
+func TestBoundedErrorSet_GetExact(t *testing.T) {
+	t.Parallel()
+
+	t.Run("exact match found", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		// Add an entry
+		entry := set.Add("test error", now)
+
+		// Test exact match
+		result, found := set.GetExact("test error", now.Add(time.Second))
+
+		require.True(t, found)
+		require.Equal(t, entry, result)
+		require.Equal(t, "test error", result.message)
+		require.Equal(t, now.Add(time.Second), result.time)
+
+		// Verify entry moved to end
+		require.Equal(t, entry, set.entries[len(set.entries)-1])
+	})
+
+	t.Run("exact match not found", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		set.Add("test error", now)
+
+		result, found := set.GetExact("different error", now)
+
+		require.False(t, found)
+		require.Nil(t, result)
+	})
+
+	t.Run("empty set", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		result, found := set.GetExact("test error", now)
+
+		require.False(t, found)
+		require.Nil(t, result)
+	})
+}
+
+func TestBoundedErrorSet_GetSimilar(t *testing.T) {
+	t.Parallel()
+
+	t.Run("similar match found", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		entry := set.Add("test error message", now)
+
+		similarityFunc := func(a, b string) bool {
+			return len(a) == len(b) && len(a) >= 10 // Simple similarity function
+		}
+
+		result, found := set.GetSimilar("test error message", now.Add(time.Second), similarityFunc)
+
+		require.True(t, found)
+		require.Equal(t, entry, result)
+		require.Equal(t, now.Add(time.Second), result.time)
+
+		// Verify entry moved to end
+		require.Equal(t, entry, set.entries[len(set.entries)-1])
+	})
+
+	t.Run("similar match not found", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		set.Add("test error", now)
+
+		similarityFunc := func(a, b string) bool {
+			return false // Never similar
+		}
+
+		result, found := set.GetSimilar("different error", now, similarityFunc)
+
+		require.False(t, found)
+		require.Nil(t, result)
+	})
+
+	t.Run("empty set", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		similarityFunc := func(a, b string) bool {
+			return true
+		}
+
+		result, found := set.GetSimilar("test error", now, similarityFunc)
+
+		require.False(t, found)
+		require.Nil(t, result)
+	})
+
+	t.Run("multiple entries, first match", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		firstEntry := set.Add("first error", now)
+		set.Add("second error", now.Add(time.Second))
+		set.Add("third error", now.Add(time.Second*2))
+
+		similarityFunc := func(a, b string) bool {
+			return a == "first error" && b == "first error"
+		}
+
+		result, found := set.GetSimilar("first error", now.Add(time.Second*3), similarityFunc)
+
+		require.True(t, found)
+		require.Equal(t, firstEntry, result)
+		require.Equal(t, now.Add(time.Second*3), result.time)
+
+		// Verify entry moved to end
+		require.Equal(t, firstEntry, set.entries[len(set.entries)-1])
+	})
+}
+
+func TestBoundedErrorSet_Add(t *testing.T) {
+	t.Parallel()
+
+	t.Run("add to empty set", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		entry := set.Add("test error", now)
+
+		require.Equal(t, "test error", entry.message)
+		require.Equal(t, now, entry.time)
+		require.Equal(t, 1, len(set.entries))
+		require.Equal(t, 1, len(set.index))
+		require.Equal(t, entry, set.index["test error"])
+		require.Equal(t, entry, set.entries[0])
+	})
+
+	t.Run("add multiple entries", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		entry1 := set.Add("error 1", now)
+		entry2 := set.Add("error 2", now.Add(time.Second))
+		entry3 := set.Add("error 3", now.Add(time.Second*2))
+
+		require.Equal(t, 3, len(set.entries))
+		require.Equal(t, 3, len(set.index))
+		require.Equal(t, entry1, set.index["error 1"])
+		require.Equal(t, entry2, set.index["error 2"])
+		require.Equal(t, entry3, set.index["error 3"])
+
+		// Verify order
+		require.Equal(t, entry1, set.entries[0])
+		require.Equal(t, entry2, set.entries[1])
+		require.Equal(t, entry3, set.entries[2])
+	})
+
+	t.Run("add duplicate message", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		entry1 := set.Add("test error", now)
+		entry2 := set.Add("test error", now.Add(time.Second))
+
+		require.Equal(t, 2, len(set.entries))
+		require.Equal(t, 1, len(set.index))               // Only one unique message
+		require.Equal(t, entry2, set.index["test error"]) // Latest entry overwrites
+		require.Equal(t, entry1, set.entries[0])
+		require.Equal(t, entry2, set.entries[1])
+	})
+}
+
+func TestBoundedErrorSet_DropStale(t *testing.T) {
+	t.Parallel()
+
+	t.Run("drop stale entries", func(t *testing.T) {
+		t.Parallel()
+
+		set := newBoundedErrorSet(5)
+		now := time.Now()
+
+		// Add entries with different timestamps
+		set.Add("stale error", now.Add(-time.Hour))
+		freshEntry := set.Add("fresh error", now)
+
+		cutoff := now.Add(-time.Minute)
+
+		set.DropStale(cutoff)
+
+		require.Equal(t, 1, len(set.entries))
+		require.Equal(t, 1, len(set.index))
+		require.Equal(t, freshEntry, set.entries[0])
+		require.Equal(t, freshEntry, set.index["fresh error"])
+		require.Nil(t, set.index["stale error"])
+	})
+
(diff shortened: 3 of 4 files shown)


enterprise/reporting/error_rate_limiter.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	// Check if we've exceeded the rate limit
+	maxErrors := erl.maxErrorsPerMinute.Load()
+	if counter.count >= maxErrors {
+		return false
+	}
+
+	// Check if this is a similar error to existing ones
+	if erl.isSimilarToExisting(errorMessage, counter.messages) {
```
satishrudderstack: Did we consider any other approach for error message similarity ? I think LCS could be slightly expensive in terms of time-complexity.

satishrudderstack:
@koladilip What's the aim of this PR ? Is it only to limit the number of rows we are adding to `error_detail_reports` table or limit the cardinality of error messages we are tracking ? 

With current implementation it looks like, we are skipping reporting if a similar error message is already reported. With this, we'll loose the information about how many times an error has occurred. We won't be able to figure out the top error messages (?)

enterprise/reporting/error_rate_limiter.go
koladilip: we will doing lcs similarity check for first 20 unique messages in particular group only after that we will not extract errors so no more comparisons will happen, I have considered cosine similarity but it didn't work very well. I will share benchmark results.

koladilip:
> @koladilip What's the aim of this PR ? Is it only to limit the number of rows we are adding to `error_detail_reports` table or limit the cardinality of error messages we are tracking ?
> 
> With current implementation it looks like, we are skipping reporting if a similar error message is already reported. With this, we'll loose the information about how many times an error has occurred. We won't be able to figure out the top error messages (?)

This is to reduce the reports we store in rudder-server error_detail_reports, goal of this PR is not solve complete reduction of error cardinality from reporting service perceptive, it might reduce a bit but it is not guaranteed.

we'll loose the information about how many times an error has occurred -> I can fix this by returning the similar error already stored instead of skipping.

vinayteki95:
> > @koladilip What's the aim of this PR ? Is it only to limit the number of rows we are adding to `error_detail_reports` table or limit the cardinality of error messages we are tracking ?
> > With current implementation it looks like, we are skipping reporting if a similar error message is already reported. With this, we'll loose the information about how many times an error has occurred. We won't be able to figure out the top error messages (?)
> 
> This is to reduce the reports we store in rudder-server error_detail_reports, goal of this PR is not solve complete reduction of error cardinality from reporting service perceptive, it might reduce a bit but it is not guaranteed.
> 
> we'll loose the information about how many times an error has occurred -> I can fix this by returning the similar error already stored instead of skipping.

Based on LCS, can we get the "template" of the common string and store that, along with the diverged words replaced with "__", this way the messages can also be as human readable as possible?

Example: "UserId 1234567 is missing, please provide correct id"
Template: "UserId __ is missing, please provide correct id"

go.mod
```diff
@@ -11,6 +11,7 @@ replace (
 	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.10.0
 	github.com/go-jose/go-jose/v3 => github.com/go-jose/go-jose/v3 v3.0.3
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.14
+	github.com/rudderlabs/rudder-go-kit => github.com/rudderlabs/rudder-go-kit v0.61.0
```
atzoum: why do we need this?
```suggestion
```

utils/types/reporting_types.go
```diff
@@ -190,6 +190,62 @@ type PUReportedMetric struct {
...
+		result.StatusDetail = &StatusDetail{
+			Status:         p.StatusDetail.Status,
+			Count:          p.StatusDetail.Count,
+			StatusCode:     p.StatusDetail.StatusCode,
+			SampleResponse: p.StatusDetail.SampleResponse,
+			EventName:      p.StatusDetail.EventName,
+			EventType:      p.StatusDetail.EventType,
+			ErrorType:      p.StatusDetail.ErrorType,
+			ViolationCount: p.StatusDetail.ViolationCount,
+			ErrorDetails:   p.StatusDetail.ErrorDetails,
```
atzoum:
```suggestion
			ErrorDetails:   p.StatusDetail.ErrorDetails,
			StatTags:       maps.Clone(p.StatusDetail.StatTags),
```

utils/types/reporting_types.go
```diff
@@ -190,6 +190,62 @@ type PUReportedMetric struct {
...
+		PUDetails:         p.PUDetails,
+	}
+
+	// Deep copy StatusDetail if it exists
+	if p.StatusDetail != nil {
+		result.StatusDetail = &StatusDetail{
+			Status:         p.StatusDetail.Status,
+			Count:          p.StatusDetail.Count,
+			StatusCode:     p.StatusDetail.StatusCode,
+			SampleResponse: p.StatusDetail.SampleResponse,
```
atzoum:
```suggestion
			SampleResponse: p.StatusDetail.SampleResponse,
			SampleEvent:    bytes.Clone(p.StatusDetail.SampleEvent),
```

utils/types/reporting_types.go
```diff
@@ -190,6 +190,62 @@ type PUReportedMetric struct {
...
+			copy(result.StatusDetail.SampleEvent, p.StatusDetail.SampleEvent)
+		}
+
+		// Deep copy StatTags map
+		if p.StatusDetail.StatTags != nil {
+			result.StatusDetail.StatTags = make(map[string]string, len(p.StatusDetail.StatTags))
+			for k, v := range p.StatusDetail.StatTags {
+				result.StatusDetail.StatTags[k] = v
+			}
+		}
```
atzoum:
```suggestion
```

utils/types/reporting_types.go
```diff
@@ -190,6 +190,62 @@ type PUReportedMetric struct {
...
+			result.StatusDetail.FailedMessages = make([]*FailedMessage, len(p.StatusDetail.FailedMessages))
+			for i, msg := range p.StatusDetail.FailedMessages {
+				if msg != nil {
+					result.StatusDetail.FailedMessages[i] = &FailedMessage{
+						MessageID:  msg.MessageID,
+						ReceivedAt: msg.ReceivedAt,
+					}
+				}
+			}
+		}
```
atzoum:
```suggestion
		if p.StatusDetail.FailedMessages != nil {
			result.StatusDetail.FailedMessages = lo.Map(p.StatusDetail.FailedMessages, func(msg *FailedMessage, _ int) *FailedMessage {
				if msg == nil {
					return nil
				}
				return &FailedMessage{
					MessageID:  msg.MessageID,
					ReceivedAt: msg.ReceivedAt,
				}
			})
		}
```

enterprise/reporting/error_grouper.go
```diff
@@ -0,0 +1,52 @@
...
+// generateMetricGroupKey creates a unique key for connection details
+func (edr *ErrorDetailReporter) generateMetricGroupKey(metric *types.PUReportedMetric) string {
+	keys := []string{
+		metric.SourceID,
+		metric.DestinationID,
+		metric.PU,
+		metric.StatusDetail.EventType,
+	}
+	return strings.Join(keys, "::")
+}
```
atzoum: The grouping key needs not be a `string`. Can simply use a struct, which is comparable and might also be more performant
```go
type errorDetailGroupKey struct {
	SourceID      string
	DestinationID string
	PU            string
	EventType     string
}
```

enterprise/reporting/error_reporting.go
```diff
@@ -238,85 +248,126 @@ func shouldReport(metric types.PUReportedMetric) bool {
...
-		destinationDetail := edr.configSubscriber.GetDestDetail(metric.DestinationID)
-		edr.log.Debugn("DestinationId & DestDetail details",
-			obskit.DestinationID(metric.DestinationID),
-			obskit.DestinationType(destinationDetail.destType),
-			logger.NewStringField("destinationDefinitionID", destinationDetail.destinationDefinitionID))
 
-		// extract error-message & error-code
+		// EXTRACT ERROR DETAILS - This is where error messages are created from SampleResponse
+		destinationDetail := edr.configSubscriber.GetDestDetail(metric.DestinationID)
 		metric.StatusDetail.ErrorDetails = edr.extractErrorDetails(metric.StatusDetail.SampleResponse, metric.StatusDetail.StatTags, destinationDetail.destType)
```
atzoum: Are we intentionally mutating the original `*PUReportedMetric` here?

enterprise/reporting/error_grouper.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"strings"
+
+	"github.com/samber/lo"
+
+	"github.com/rudderlabs/rudder-server/utils/types"
+)
+
+// mergeMetricGroupsByErrorSimilarity merges metrics by LCS similarity on error message
+// Returns a map where key is "connectionKey::errorKey" and value is the grouped metrics
```
atzoum: this doc appears to be outdated

enterprise/reporting/error_reporting.go
```diff
@@ -238,85 +248,126 @@ func shouldReport(metric types.PUReportedMetric) bool {
 }
 
 func (edr *ErrorDetailReporter) Report(ctx context.Context, metrics []*types.PUReportedMetric, txn *Tx) error {
-	edr.log.Debugn("[ErrorDetailReport] Report method called")
-	if len(metrics) == 0 {
+	// Extract error details and filter metrics that should be reported
+	reportableMetrics := edr.extractErrorDetailsAndFilterMetrics(metrics)
```
atzoum, line 282: `reportableMetrics` should be of an internal type other than `PUReportedMetric`. `extractErrorDetailsAndFilterMetrics` shouldn't mutate the existing `metrics`

enterprise/reporting/error_reporting.go
```diff
@@ -238,85 +248,126 @@ func shouldReport(metric types.PUReportedMetric) bool {
...
 
-	stmt, err := txn.PrepareContext(ctx, pq.CopyIn(ErrorDetailReportsTable, ErrorDetailReportsColumns...))
-	if err != nil {
-		edr.log.Errorn("Failed during statement preparation", obskit.Error(err))
-		return fmt.Errorf("preparing statement: %v", err)
+	// Group errors by connection details
+	connectionGroups := edr.groupByConnection(reportableMetrics)
+
+	// Apply rate limiting after grouping (modifies connectionGroups in place)
+	if edr.errorRateLimiter != nil {
```
atzoum: `errorRateLimiter` should never be `nil`. If we are doing this for tests, just prepare a mock errorRateLimiter and use this

enterprise/reporting/error_rate_limiter.go
```diff
@@ -0,0 +1,305 @@
...
+	"github.com/rudderlabs/rudder-go-kit/config"
+	"github.com/rudderlabs/rudder-go-kit/lcs"
+	"github.com/rudderlabs/rudder-go-kit/logger"
+)
+
+//go:generate mockgen -destination=../../../mocks/enterprise/reporting/mock_error_rate_limiter.go -package=mocks github.com/rudderlabs/rudder-server/enterprise/reporting ErrorRateLimiter
+
+// ErrorRateLimiter defines the interface for error rate limiting functionality
+// This interface allows for easier testing by enabling mock implementations
+type ErrorRateLimiter interface {
```
atzoum: Rate limiting is not what this component is actually doing. Should we call it something else instead, e.g. `ErrorNormalizer`?

enterprise/reporting/error_rate_limiter.go
```diff
@@ -0,0 +1,305 @@
...
+type cleanupStrategy interface {
+	// shouldCleanupCounter determines if a counter should be cleaned up
+	shouldCleanupCounter(counter *errorCounter, now time.Time) bool
+
+	// cleanupMessages cleans up individual messages within a counter
+	cleanupMessages(counter *errorCounter, now time.Time, minCleanupInterval time.Duration)
+
+	// handleRateLimit handles what happens when rate limit is reached
+	// Returns true if the error should be allowed (after eviction), false if it should be rejected
+	handleRateLimit(counter *errorCounter, newMessage string, now time.Time) bool
```
atzoum: We shouldn't be needing this method. A single `cleanup` method should be enough
```suggestion
```

enterprise/reporting/error_rate_limiter.go
```diff
@@ -0,0 +1,305 @@
...
+	StartCleanup(ctx context.Context) error
+}
+
+// cleanupStrategy defines the interface for different cleanup strategies
+type cleanupStrategy interface {
+	// shouldCleanupCounter determines if a counter should be cleaned up
+	shouldCleanupCounter(counter *errorCounter, now time.Time) bool
+
+	// cleanupMessages cleans up individual messages within a counter
+	cleanupMessages(counter *errorCounter, now time.Time, minCleanupInterval time.Duration)
```
atzoum: We just need 1 method
```suggestion
	// cleanup determines if a counter should be cleaned up or not. Performs cleanup of messages if the whole counter doesn't need to be cleaned up
	cleanup(counter *errorCounter, now time.Time) bool

```

enterprise/reporting/error_rate_limiter.go
```diff
@@ -0,0 +1,305 @@
...
+	// Returns true if the error should be allowed (after eviction), false if it should be rejected
+	handleRateLimit(counter *errorCounter, newMessage string, now time.Time) bool
+}
+
+// errorCounter tracks errors for a specific combination
+type errorCounter struct {
+	windowEnd          time.Time
+	lastMessageCleanup time.Time // Track when messages were last cleaned from this counter
+	// Track error messages and when each was last used (for LRU strategy)
+	messageTimestamps map[string]time.Time // error message -> last used timestamp
```
atzoum: We may use a list of messages sorted by timestamp ascending so that scheduled cleanup is fast

enterprise/reporting/error_rate_limiter.go
```diff
@@ -0,0 +1,305 @@
...
+			logger.NewIntField("remaining", int64(finalCount)),
+		)
+	}
+
+	e.lastCleanup = now
+}
+
+// StartCleanup starts the periodic cleanup routine
+func (e *errorRateLimiter) StartCleanup(ctx context.Context) error {
+	ticker := time.NewTicker(e.maxCountersCheckInterval.Load())
```
atzoum: cleanup should happen relatively frequently, e.g. every 1-5 seconds and it should be fast

enterprise/reporting/error_rate_limiter.go
```diff
@@ -0,0 +1,305 @@
...
+	// Track error messages and when each was last used (for LRU strategy)
+	messageTimestamps map[string]time.Time // error message -> last used timestamp
+}
+
+// errorRateLimiter provides error rate limiting functionality with configurable strategies
+type errorRateLimiter struct {
+	mu                sync.RWMutex
+	log               logger.Logger
+	lastCleanup       time.Time
+	cleanupInProgress int32
```
atzoum: if cleanup is fast enough it should simply acquire a lock and block other operations
```suggestion
```

go.mod
koladilip: not required.

enterprise/reporting/error_reporting.go
koladilip: created new type

enterprise/reporting/error_reporting.go
koladilip: removed the mutation

utils/types/reporting_types.go
koladilip: Removed this completely, there is already a struct so using that.

utils/types/reporting_types.go
koladilip: we don't mutate SampleEvent

utils/types/reporting_types.go
koladilip: This is removed

utils/types/reporting_types.go
koladilip: This code is removed

utils/types/reporting_types.go
koladilip: Removed

enterprise/reporting/error_rate_limiter.go
koladilip: Done

enterprise/reporting/error_rate_limiter.go
koladilip: Yes done

enterprise/reporting/error_rate_limiter.go
koladilip: Usually counters will be small size with max limit 10k, it is just simple to just loop over without any optimization.

enterprise/reporting/error_rate_limiter.go
koladilip: done

enterprise/reporting/error_rate_limiter.go
koladilip: yes updated

enterprise/reporting/error_grouper.go
koladilip: Removed them

enterprise/reporting/error_reporting.go
koladilip: we are initialising only when it is enabled using config. I will make it always enabled

enterprise/reporting/error_rate_limiter.go
koladilip: Done

enterprise/reporting/error_grouper.go
koladilip: Done

enterprise/reporting/error_reporting.go
```diff
@@ -238,85 +244,124 @@ func shouldReport(metric types.PUReportedMetric) bool {
...
+				metric.Namespace,
+				metric.InstanceID,
+				metric.SourceDefinitionId,
+				metric.SourceID,
+				metric.DestinationDefinitionId,
+				metric.DestinationID,
+				metric.DestType,
+				metric.PU,
+				metric.ReportedAt,
+				totalCount,
```
atzoum: Why are we persisting `totalCount` instead of `metric.Count`?

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,152 @@
...
+	lastBlocked time.Time
+}
+
+type errorNormalizer struct {
+	mu       sync.Mutex
+	log      logger.Logger
+	counters map[types.ErrorDetailGroupKey]*errorCounter
+
+	similarityThreshold config.ValueLoader[float64]
+	maxErrorsPerMinute  config.ValueLoader[int]
```
atzoum: Why `PerMinute`? It is `PerGroup` right?
```suggestion
	maxErrorsPerGroup  config.ValueLoader[int]
```

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,152 @@
...
+	// NormalizeError returns the normalized error message for the given error
+	// Returns "UnknownError" if rate limited, or the normalized error message if should be reported
+	NormalizeError(ctx context.Context, errorDetailGroupKey types.ErrorDetailGroupKey, errorMessage string) string
+
+	// StartCleanup starts the periodic cleanup routine
+	StartCleanup(ctx context.Context) error
+}
+
+type errorCounter struct {
+	messages    map[string]time.Time
```
atzoum: Let's have messages as a list of errors sorted by time ascending. This should help considerably during cleanup, since we could stop iterating after encountering a time beyond the cutoff time
```suggestion
	errors    []lo.Tuple2[string, time.Time] // list of errors sorted by last encountered time ascending
```

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,152 @@
...
+		// Drop if no messages left OR no new unique error for >1m
+		if e.shouldDropCounter(c, now, maxErrorsPerMinute) {
+			delete(e.counters, k)
+		}
+	}
+}
+
+func (e *errorNormalizer) dropStaleMessages(c *errorCounter, cutoff time.Time) {
+	for msg, t := range c.messages {
+		if t.Before(cutoff) {
```
atzoum: iteration should stop when this becomes false, e.g.

```go
func (e *errorNormalizer) dropStaleMessages(c *errorCounter, cutoff time.Time) {
	var idx int
	for i, msg := range c.messages {
		if msg.B.Before(cutoff) {
			idx = i + 1
		} else {
			break
		}
	}
	if idx > 0 {
		c.messages = c.messages[idx:]
	}
}
```

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,152 @@
...
+type errorCounter struct {
+	messages    map[string]time.Time
+	lastUpdated time.Time
+	lastBlocked time.Time
+}
+
+type errorNormalizer struct {
+	mu       sync.Mutex
+	log      logger.Logger
+	counters map[types.ErrorDetailGroupKey]*errorCounter
```
atzoum:
```suggestion
	errorGroups map[types.ErrorDetailGroupKey]*errorGroup
```

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,152 @@
...
+type ErrorNormalizer interface {
+	// NormalizeError returns the normalized error message for the given error
+	// Returns "UnknownError" if rate limited, or the normalized error message if should be reported
+	NormalizeError(ctx context.Context, errorDetailGroupKey types.ErrorDetailGroupKey, errorMessage string) string
+
+	// StartCleanup starts the periodic cleanup routine
+	StartCleanup(ctx context.Context) error
+}
+
+type errorCounter struct {
```
atzoum:
```suggestion
type errorGroup struct {
```

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,152 @@
...
+
+	similarityThreshold := e.similarityThreshold.Load()
+	maxErrorsPerMinute := e.maxErrorsPerMinute.Load()
+	// Similarity
+	for existing := range c.messages {
+		if lcs.Similarity(msg, existing) >= similarityThreshold {
+			c.messages[existing] = now
+			return existing
+		}
+	}
```
atzoum:
```suggestion
	for i, errorTuple := range c.errors {
		if lcs.Similarity(msg, errorTuple.A) >= similarityThreshold {
			c.errors = append(c.errors[:i], c.errors[i+1:]...)
			errorTuple.B = now
			// Append to the end
			c.errors = append(c.errors, errorTuple)
			return errorTuple.A
		}
	}
```

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Rate limit
+	if len(c.messages) >= maxErrorsPerMinute {
+		c.lastBlocked = now
+		return "UnknownError"
+	}
+
+	// Insert
+	c.messages[msg] = now
```
atzoum:
```suggestion
	c.errors = append(c.errors, lo.Tuple2[string, time.Time]{A: msg, B: now})
```

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,152 @@
...
+	// Returns "UnknownError" if rate limited, or the normalized error message if should be reported
+	NormalizeError(ctx context.Context, errorDetailGroupKey types.ErrorDetailGroupKey, errorMessage string) string
+
+	// StartCleanup starts the periodic cleanup routine
+	StartCleanup(ctx context.Context) error
+}
+
+type errorCounter struct {
+	messages    map[string]time.Time
+	lastUpdated time.Time
```
atzoum, line 116: We don't need lastUpdated, just get the last element of the list
```suggestion
```

utils/types/reporting_types.go
```diff
@@ -190,6 +190,74 @@ type PUReportedMetric struct {
...
+	InstanceID              string
+	DestType                string
+	DestinationDefinitionID string
+	ErrorDetails            ErrorDetails
+}
+
+// PUReportedMetricToEDReportsDB converts PUReportedMetric to EDReportsDB
+// This avoids deep copying by extracting only the fields needed for error reporting
+func PUReportedMetricToEDReportsDB(
+	metric *PUReportedMetric,
```
atzoum, line 205: Do we need `ErrorDetails` inside `StatusDetail` anymore?

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,152 @@
...
+	// Convert struct key to string for internal use
+	now := time.Now()
+
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	c, ok := e.counters[errorDetailGroupKey]
+	if !ok {
+		if len(e.counters) >= e.maxCounters.Load() {
+			return "UnknownError"
```
atzoum: Maybe redacted is a better term to use
```suggestion
			return "RedactedError"
```

enterprise/reporting/error_reporting.go
```diff
@@ -238,85 +244,124 @@ func shouldReport(metric types.PUReportedMetric) bool {
...
 
-		edr.stats.NewTaggedStat("error_detail_reporting_failures", stats.CountType, stats.Tags{
-			"errorCode":     metric.StatusDetail.ErrorDetails.Code,
-			"workspaceId":   workspaceID,
-			"destType":      destinationDetail.destType,
-			"sourceId":      metric.SourceID,
-			"destinationId": metric.DestinationID,
-		}).Count(int(metric.StatusDetail.Count))
-
-		sampleEvent, sampleResponse, err := getSampleWithEventSampling(metric, reportedAt, edr.eventSampler, edr.eventSamplingEnabled.Load(), int64(edr.eventSamplingDuration.Load().Minutes()))
```
atzoum, line 282: Don't we need to retain sampling for error details as well? Maybe we should refactor `getSampleWithEventSampling` to not accept a `PUReportedMetric`, e.g.:
```go
func getSampleWithEventSampling(labelSet LabelSet, sampleEvent json.RawMessage, sampleResponse string, reportedAt int64, eventSampler event_sampler.EventSampler, eventSamplingEnabled bool, eventSamplingDuration int64) (sampleEvent json.RawMessage, sampleResponse string, err error) {
```

enterprise/reporting/error_reporting.go
koladilip: this is a mistake, should be metric count.

enterprise/reporting/error_reporting.go
koladilip: We need this. It was mistake. adding back

utils/types/reporting_types.go
koladilip: removed and refactored code.

enterprise/reporting/error_normalizer.go
koladilip: updated

enterprise/reporting/error_normalizer.go
koladilip: didn't implement sorted list of errors as, just keeping map only for tracking errors, sorting for every error seems to costly. I have reduced cleanup interval to 5 secs instead 1 sec, at max there will be 200K messages to check so it shouldn't take much time.

enterprise/reporting/error_normalizer.go
koladilip: didn't implemented this

enterprise/reporting/error_normalizer.go
koladilip: done

enterprise/reporting/error_normalizer.go
koladilip: didn't implement the sorting.

enterprise/reporting/error_normalizer.go
koladilip: didn't implement sorting as we need to sort list for every error processed, it is costly

enterprise/reporting/error_normalizer.go
koladilip: Done

enterprise/reporting/error_reporting.go
koladilip: Done

enterprise/reporting/error_normalizer.go
koladilip: Didn't implement

enterprise/reporting/error_normalizer.go
koladilip: Done

enterprise/reporting/error_reporting.go
```diff
@@ -123,7 +125,9 @@ func NewErrorDetailReporter(
 	eventSamplingCardinality := conf.GetReloadableIntVar(100000, 1, "Reporting.eventSampling.cardinality")
 
 	log := logger.NewLogger().Child("enterprise").Child("error-detail-reporting")
-	extractor := NewErrorDetailExtractor(log)
+	extractor := NewErrorDetailExtractor(log, conf)
+	errorNormalizer := NewErrorNormalizer(log, conf, stats)
+	groupingThreshold := conf.GetReloadableFloat64Var(0.75, "Reporting.errorReporting.grouping.similarityThreshold")
```
atzoum, line 159:
```suggestion
	similarityThreshold := conf.GetReloadableFloat64Var(0.75, "Reporting.errorReporting.grouping.similarityThreshold")
```

enterprise/reporting/error_reporting.go
```diff
@@ -152,6 +156,7 @@ func NewErrorDetailReporter(
 		eventSamplingEnabled:  eventSamplingEnabled,
 		eventSamplingDuration: eventSamplingDuration,
 		eventSampler:          eventSampler,
+		groupingThreshold:     groupingThreshold,
```
atzoum, line 188:
```suggestion
		similarityThreshold:     similarityThreshold,
```

enterprise/reporting/error_reporting.go
```diff
@@ -85,13 +85,15 @@ type ErrorDetailReporter struct {
...
 	errorDetailExtractor *ExtractorHandle
+	errorNormalizer      ErrorNormalizer
 
 	minReportedAtQueryTime      stats.Measurement
 	errorDetailReportsQueryTime stats.Measurement
 	edReportingRequestLatency   stats.Measurement
 	eventSamplingEnabled        config.ValueLoader[bool]
 	eventSamplingDuration       config.ValueLoader[time.Duration]
 	eventSampler                event_sampler.EventSampler
+	groupingThreshold           config.ValueLoader[float64]
```
atzoum, line 124:
```suggestion
	similarityThreshold           config.ValueLoader[float64]
```

enterprise/reporting/error_reporting.go
```diff
@@ -238,85 +244,134 @@ func shouldReport(metric types.PUReportedMetric) bool {
 }
 
 func (edr *ErrorDetailReporter) Report(ctx context.Context, metrics []*types.PUReportedMetric, txn *Tx) error {
-	edr.log.Debugn("[ErrorDetailReport] Report method called")
-	if len(metrics) == 0 {
+	start := time.Now()
+	defer func() {
+		edr.stats.NewStat("error_detail_reporter_report_time", stats.TimerType).Since(start)
```
atzoum: Please keep stats cached, NewStat is usually an expensive operation

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,169 @@
...
+	// Insert the new error message
+	c.errors[msg] = now
+	c.lastUpdated = now
+	return msg
+}
+
+func (e *errorNormalizer) cleanup() {
+	start := time.Now()
+	defer func() {
+		e.stats.NewStat("error_normalizer_cleanup_time", stats.TimerType).Since(start)
```
atzoum:
```suggestion
		e.stats.NewStat("reporting_error_normalizer_cleanup_time", stats.TimerType).Since(start)
```

atzoum (approved):
I am getting better results when using a sorted slice when I have to cleanup 1000 error groups, even taking into account the write delay, please feel free to experiment too
```
-test.shuffle 1756383638044817000
goos: darwin
goarch: arm64
pkg: github.com/rudderlabs/rudder-server/enterprise/reporting
cpu: Apple M1 Pro
BenchmarkErrorNormalizer
BenchmarkErrorNormalizer/map_20
BenchmarkErrorNormalizer/map_20-10         	    1798	    654185 ns/op	       1 B/op	       0 allocs/op
BenchmarkErrorNormalizer/sorted_slice_20
BenchmarkErrorNormalizer/sorted_slice_20-10         	    7554	    156696 ns/op	       0 B/op	       0 allocs/op
BenchmarkErrorNormalizer/map_100
BenchmarkErrorNormalizer/map_100-10                 	     412	   3011142 ns/op	      25 B/op	       0 allocs/op
BenchmarkErrorNormalizer/sorted_slice_100
BenchmarkErrorNormalizer/sorted_slice_100-10        	    1056	   1162006 ns/op	      12 B/op	       0 allocs/op
PASS
ok  	github.com/rudderlabs/rudder-server/enterprise/reporting	5.825s
```
```go
package reporting

import (
	"strconv"
	"testing"
	"time"

	"github.com/samber/lo"
)

const maxErrorsPerGroup = 20

func BenchmarkErrorNormalizer(b *testing.B) {

	const addLoops = 100
	const cleanupLoops = 1000

	runBenchmark := func(b *testing.B, num int, sortedSlice bool) {

		var errors []string
		for i := range num {
			errors = append(errors, "test error message "+strconv.Itoa(i))
		}

		eg := &errGroup{
			errors:    make(map[string]time.Time),
			errorsAlt: make(map[string]struct{}),
			errorsArr: make([]lo.Tuple2[string, time.Time], 0, maxErrorsPerGroup),
		}

		switch sortedSlice {
		case true:
			b.Run("sorted slice "+strconv.Itoa(num), func(b *testing.B) {
				for b.Loop() {
					for range addLoops {
						for j := 0; j < num; j++ {
							eg.addErrorAlt(errors[j])
						}
					}
					for range cleanupLoops {
						eg.cleanupAlt()
					}

				}
			})
		case false:
			b.Run("map "+strconv.Itoa(num), func(b *testing.B) {
				for b.Loop() {
					for range addLoops {
						for j := range num {
							eg.addError(errors[j])
						}
					}
					for range cleanupLoops {
						eg.cleanup()
					}
				}
			})
		}
	}

	runBenchmark(b, 20, false)
	runBenchmark(b, 20, true)

	runBenchmark(b, 100, false)
	runBenchmark(b, 100, true)

}

type errGroup struct {
	errors map[string]time.Time

	errorsAlt map[string]struct{}
	errorsArr []lo.Tuple2[string, time.Time]
}

func (eg *errGroup) addError(err string) {
	eg.errors[err] = time.Now()
}

func (eg *errGroup) addErrorAlt(err string) {
	if _, ok := eg.errorsAlt[err]; ok {
		for i, t := range eg.errorsArr {
			if t.A == err {
				t.B = time.Now()
				eg.errorsArr[i] = t
				// Move element to end
				eg.errorsArr = append(eg.errorsArr[:i], eg.errorsArr[i+1:]...)
				eg.errorsArr = append(eg.errorsArr, t)
				break
			}
		}
	} else {
		eg.errorsAlt[err] = struct{}{}
		eg.errorsArr = append(eg.errorsArr, lo.T2(err, time.Now()))
	}
}

func (eg *errGroup) cleanup() {
	for err, t := range eg.errors {
		if time.Since(t) > time.Minute {
			delete(eg.errors, err)
		}
	}
}

func (eg *errGroup) cleanupAlt() {
	var delIdx int
	for i := 0; i < len(eg.errorsArr); i++ {
		if time.Since(eg.errorsArr[i].B) > time.Minute {
			delIdx = i + 1
			delete(eg.errorsAlt, eg.errorsArr[i].A)
		} else {
			break
		}
	}
	if delIdx > 0 {
		eg.errorsArr = eg.errorsArr[delIdx:]
	}
}

```

enterprise/reporting/error_reporting.go
koladilip: this is not required now, removing

enterprise/reporting/error_reporting.go
koladilip: this is not required now, removing

enterprise/reporting/error_reporting.go
koladilip: this is not required now, removing

koladilip:
I did benmark and sorted is better for cleanup at least 10x times.

enterprise/reporting/error_normalizer.go
koladilip: implemented this idea using slightly different data structures

enterprise/reporting/error_grouper.go
```diff
@@ -0,0 +1,47 @@
...
+		for errorMsg, metrics := range metricsGroupedByError {
+			mergedMetricsGroups[errorMsg] = metrics[0]
+			mergedMetricsGroups[errorMsg].Count = lo.SumBy(metrics, func(metric *types.EDReportsDB) int64 {
+				return metric.Count
+			})
+		}
+		metricGroups[groupKey] = lo.Values(mergedMetricsGroups)
+	}
+	return metricGroups
+}
```
satishrudderstack, line 25: Since the grouping does not include event name, should we remove that label altogether from the `error_details_reports` table ? 

Let's say there are 3 track events, `product_viewed`, `product_added_to_cart`, `product_purchased` and all three have same error, we'll just count the events against one of these event names which might give the impression that all these events are let's say `product_viewed`

enterprise/reporting/error_normalizer.go
```diff
@@ -0,0 +1,249 @@
...
+	return nil, false
+}
+
+// GetSimilar scans for a near match (caller supplies similarity fn)
+func (s *boundedErrorSet) GetSimilar(msg string, now time.Time, similar func(a, b string) bool) (*errorEntry, bool) {
+	for i, entry := range s.entries {
+		if similar(msg, entry.message) {
+			entry.time = now
+			s.moveToEndAt(i)
+			return entry, true
```
satishrudderstack, line 66: Why do we need this `moveToEnd` ? So that it is easier to cleanup ?

If yes, I think it might be better to handle these in the `cleanup` function as the load on `cleanup` function will be less than `GetSimilar` or `GetExact` functions

satishrudderstack (approved):

enterprise/reporting/error_grouper.go
koladilip: we can still keep, it will be still useful sometimes.

enterprise/reporting/error_normalizer.go
koladilip: I did benchmark, sorted version is better overall.

PR #6257, opened 2025-08-20
chore: cleanup for schema snapshoting

achettyiitr:
# Description

- Remove the `enableStagingFileSchemaSnapshot` flag-based changes.
- Stop populating the schema while inserting the staging file.
- Remove unmarshalling and deep equals.

## Linear Ticket

- Resolves WAR-906

## Security

### DIFF
--- a/warehouse/internal/repo/staging.go
+++ b/warehouse/internal/repo/staging.go
@@ -5,7 +5,6 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
-	"reflect"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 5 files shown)


warehouse/internal/repo/staging.go
```diff
@@ -147,10 +145,6 @@ func (sf *StagingFiles) Insert(ctx context.Context, stagingFile *model.StagingFi
 	}
 	now := sf.now()
 
-	schemaPayload, err := jsonrs.Marshal(stagingFile.Schema)
```
shekhar-rudder, line 150: Can we drop `Schema` field from `stagingFile` ?

achettyiitr: How can we drop it? For old data, snapshots are not present. Only the schema is available.

shekhar-rudder: Ok. But instead of saving it as {} in the repositry layer, can we assign it {} in the process API itself?

achettyiitr: IMO process API should not know these details. Even for getting the schemas, it’s the repository layer that knows the details how to compile the schema. Either from snapshot or from the actual one.

shekhar-rudder (approved):

PR #6268, opened 2025-08-22
fix: use source-id for getting cached schemas

achettyiitr:
# Description

- Use sourceID for getting schemas instead of ID to be more deterministic in cases where schemas are available for the same destinationID and namespace.

## Linear Ticket

- Resolves WAR-1064

## Security


warehouse/internal/repo/schema.go
```diff
@@ -329,13 +330,13 @@ func (sh *WHSchema) getTableLevelSchemasForNamespaceWithTx(ctx context.Context,
 				schema,
 				ROW_NUMBER() OVER (
 					PARTITION BY destination_id, namespace, table_name
-					ORDER BY id DESC
+					ORDER BY source_id DESC
```
shekhar-rudder, line 334: How does changing ordering fix the problem? Earlier it was sorting in a particular way and now a different way. There is still a possibility of mismatch, rt?

achettyiitr: Once the table-level schemas are populated, order by sourceID will give you the same results, right? Why would there be mismatches?

achettyiitr: BTW, we are populating table-level schemas first if not populated and then comparing.

vyeshwanth: I didn't understand this as well. why are we sorting by source_id desc.

achettyiitr: Screenshot 2025-08-22 at 2 35 41 PM

When we are getting original schemas, **we pick the latest ID**. Now, when these original schemas get converted to table-level schemas, we don't know whether we will be getting the table-level schemas for the latest ID or some other ID.
If you see the screenshot, there is a mismatch among the original schemas.

vyeshwanth (approved):

PR #6287, opened 2025-09-01
chore: remove proc_error jobsdb

atzoum:
# Description

removing `proc_error` jobsdb and all associated functionality

## Linear Ticket

resolves PIPE-2296

## Security


archiver/archiver_test.go
```diff
@@ -120,10 +114,7 @@ func TestJobsArchival(t *testing.T) {
 				},
 			},
 			Preferences: backendconfig.StoragePreferences{
-				GatewayDumps:     true,
-				BatchRouterDumps: true,
-				RouterDumps:      true,
-				ProcErrorDumps:   true,
+				GatewayDumps: true,
```
fracasula, line 117: I'm not very familiar with this bit. How come you're removing for example `RouterDumps` but not `GatewayDumps`? 🤔

fracasula: This seems to be used in the archiver :+1:

fracasula (approved):

archiver/archiver_test.go
atzoum: Yup, ideally we should remove this as well, but didn't want to extend the scope of this pr

gateway/handle_webhook.go
```diff
@@ -27,45 +17,3 @@ func (gw *Handle) webhookHandler() http.HandlerFunc {
 func (gw *Handle) ProcessTransformedWebhookRequest(w *http.ResponseWriter, r *http.Request, reqType string, payload []byte, arctx *gwtypes.AuthRequestContext) string {
 	return gw.rrh.ProcessRequest(w, r, reqType, payload, arctx)
 }
-
-// SaveWebhookFailures saves errors to the error db
-func (gw *Handle) SaveWebhookFailures(reqs []*model.FailedWebhookPayload) error {
```
mihir20, line 32: @koladilip @vinayteki95 @atzoum should we keep proc-errors for open source users for webhook?

atzoum: This is not a feature we want to have. proc_errors has always been a hack, let's get rid of it

gateway/webhook/setup.go
```diff
@@ -37,8 +36,6 @@ type Gateway interface {
 type WebhookRequestProcessor interface {
 	// ProcessTransformedWebhookRequest processes the transformed webhook request and save it to the gw jobsDB
 	ProcessTransformedWebhookRequest(writer *http.ResponseWriter, req *http.Request, reqType string, requestPayload []byte, arctx *gwtypes.AuthRequestContext) string
-	// SaveWebhookFailures saves the webhook failures to the procErr jobsDB
-	SaveWebhookFailures([]*model.FailedWebhookPayload) error
```
mihir20: we are using this interface in ingestion-svc to publish the failed webhook events in pulsar

satishrudderstack (commented):
I quickly skimmed through and LGTM from reporting as there's no reporting related changes in this PR.

gateway/webhook/setup.go
atzoum: restored with a no-op implementation

mihir20 (approved):

satishrudderstack (approved):
LGTM from reporting side

PR #6296, opened 2025-09-03
fix: remove unwanted table level schemas

achettyiitr:
# Description

- When we fetch the schema from the warehouse, additional tables are also fetched, for which table-level schemas are also generated.
- Now, when these tables get deleted in the customer's warehouse, upon fetching the schema, we will not receive those. But since the table-level schemas got generated, they will be present.
- So, deleting those table-level schemas which are not part of the warehouse schema.
- Adding some logs to the `wh_schemas` repo.

## Linear Ticket

- Resolves WAR-1078.

## Security

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ MOUNT_PATH=/local
 
 # go tools versions
 GOLANGCI=github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.2.1
-gofumpt=mvdan.cc/gofumpt@latest
+gofumpt=mvdan.cc/gofumpt@v0.8.0
 govulncheck=golang.org/x/vuln/cmd/govulncheck@latest
 goimports=golang.org/x/tools/cmd/goimports@latest
 mockgen=go.uber.org/mock/mockgen@v0.6.0
--- a/warehouse/internal/repo/schema.go
+++ b/warehouse/internal/repo/schema.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/lib/pq"
+
 	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/jsonrs"
 	"github.com/rudderlabs/rudder-go-kit/logger"
@@ -17,8 +18,9 @@ import (
 
 	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
 
-	"github.com/rudderlabs/rudder-go-kit/stats"
 	"github.com/samber/lo"
+
+	"github.com/rudderlabs/rudder-go-kit/stats"
 
 	"github.com/rudderlabs/rudder-server/utils/timeutil"
 	sqlmiddleware "github.com/rudderlabs/rudder-server/warehouse/integrations/middleware/sqlquerywrapper"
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -7,7 +7,7 @@ MOUNT_PATH=/local
 
 # go tools versions
 GOLANGCI=github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.2.1
-gofumpt=mvdan.cc/gofumpt@latest
+gofumpt=mvdan.cc/gofumpt@v0.8.0
```
achettyiitr, line 10: https://rudderlabs.slack.com/archives/C01HTT66UMB/p1756879759535659

warehouse/internal/repo/schema.go
```diff
@@ -161,6 +197,11 @@ func (sh *WHSchema) Insert(ctx context.Context, whSchema *model.WHSchema) error
 					tableName,
 				)
 				if err != nil {
+					log.Warnn("Updating table-level related schemas",
+						logger.NewStringField("tableName", tableName),
+						logger.NewStringField("schema", string(schemaPayload)),
+						logger.NewStringField("tableLevelSchem", string(tableSchemaPayload)),
```
shekhar-rudder: tableLevelSchem -> tableLevelSchema/tableSchema

warehouse/internal/repo/schema.go
```diff
@@ -98,6 +112,7 @@ func (sh *WHSchema) Insert(ctx context.Context, whSchema *model.WHSchema) error
 			whSchema.SourceID,
 		)
 		if err != nil {
+			log.Warnn("Updating related schemas", logger.NewStringField("schema", string(schemaPayload)))
```
shekhar-rudder: Why not an error log?

warehouse/internal/repo/schema.go
```diff
@@ -161,6 +197,11 @@ func (sh *WHSchema) Insert(ctx context.Context, whSchema *model.WHSchema) error
 					tableName,
 				)
 				if err != nil {
+					log.Warnn("Updating table-level related schemas",
```
shekhar-rudder: same here

warehouse/internal/repo/schema.go
achettyiitr: Adding err as well for logging.

warehouse/internal/repo/schema.go
```diff
@@ -134,6 +154,27 @@ func (sh *WHSchema) Insert(ctx context.Context, whSchema *model.WHSchema) error
...
+				  AND namespace = $2
+				  AND table_name != ''
+				  AND table_name NOT IN (
+					SELECT unnest($3::text[])
+				  );
+			`,
+					whSchema.DestinationID,
+					whSchema.Namespace,
+					pq.Array(lo.Keys(whSchema.Schema)),
+				)
```
RanjeetMishra, line 172: @achettyiitr 
With SchemaTTL, this query would run less frequently. Query run time should be quick, but `table_name NOT IN` means it will scan all table_name rows from the table.
If I am right, this cleanup we are doing to match table-level schema and wh-schema. No issue with stale tables schema? As in we can think of optimizing cleaning stale tables schema if needed.

WDYT? If yes, I will add a task in backlog and add reference.

RanjeetMishra (approved):

PR #6300, opened 2025-09-03
chore: capture stats for failed transformer client requests

atzoum:
# Description

Using existing metric `processor_transformer_request_time` to capture request errors as well (through a new `success` label) in order to be able to visualize them in dashboards and setup alerts for sources/destinations with constant failures.

## Security


processor/internal/transformer/destination_transformer/destination_transformer.go
```diff
@@ -294,47 +294,56 @@ func (d *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels
 
 	err := backoff.RetryNotify(
 		func() error {
-			var reqErr error
-			requestStartTime := time.Now()
+			start := time.Now()
+			err := func() error {
```
koladilip: can we create common util function for this and reuse every where?

ktgowtham (approved):

PR #6303, opened 2025-09-04
fix: avoid running migrations for notifier from slaves

achettyiitr:
# Description

- Avoid running migrations for notifier from slaves

## Linear Ticket

- Resolves WAR-1095

## Security

### DIFF
--- a/warehouse/app_test.go
+++ b/warehouse/app_test.go
@@ -33,6 +33,7 @@ import (
 	mocksApp "github.com/rudderlabs/rudder-server/mocks/app"
 	mocksBackendConfig "github.com/rudderlabs/rudder-server/mocks/backend-config"
 	proto "github.com/rudderlabs/rudder-server/proto/warehouse"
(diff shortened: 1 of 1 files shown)


shekhar-rudder (approved):

RanjeetMishra (approved):

warehouse/app_test.go
```diff
@@ -107,6 +108,12 @@ func TestApp(t *testing.T) {
 					pgResource, err := postgres.Setup(pool, t)
 					require.NoError(t, err)
 
+					err = (&migrator.Migrator{
```
shekhar-rudder, line 111: Why do we need an explicit migration? Is it because in this test, master mode is not set? If yes, should we update that?

achettyiitr: Yes, this test is for specific modes like slave, where we will not be running the migration.
```
            {
				name:          "slave",
				warehouseMode: config.SlaveMode,
			},
```

PR #6306, opened 2025-09-04
chore: make warehouse enableNotifierHeartbeat flag enabled by default

shekhar-rudder:
# Description

This PR changes the default for `Warehouse.enableNotifierHeartbeat` env from `false` to `true`

## Security

### DIFF
--- a/warehouse/slave/worker.go
+++ b/warehouse/slave/worker.go
@@ -110,7 +110,7 @@ func newWorker(
 	// Increasing maxConcurrentStagingFiles config would also require increasing the memory requests for the slave pods
 	s.config.maxConcurrentStagingFiles = s.conf.GetReloadableIntVar(10, 1, "Warehouse.maxStagingFilesInUploadV2Job")
 	s.config.claimRefreshInterval = s.conf.GetReloadableDurationVar(30, time.Second, "Warehouse.claimRefreshIntervalInS")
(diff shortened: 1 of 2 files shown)


warehouse/slave/worker.go
```diff
@@ -110,7 +110,7 @@ func newWorker(
 	// Increasing maxConcurrentStagingFiles config would also require increasing the memory requests for the slave pods
 	s.config.maxConcurrentStagingFiles = s.conf.GetReloadableIntVar(10, 1, "Warehouse.maxStagingFilesInUploadV2Job")
 	s.config.claimRefreshInterval = s.conf.GetReloadableDurationVar(30, time.Second, "Warehouse.claimRefreshIntervalInS")
-	s.config.enableNotifierHeartbeat = s.conf.GetReloadableBoolVar(false, "Warehouse.enableNotifierHeartbeat")
+	s.config.enableNotifierHeartbeat = s.conf.GetReloadableBoolVar(true, "Warehouse.enableNotifierHeartbeat")
```
achettyiitr: Can we instead get rid of the flag? Since it's stable now, we don't need a feature flag for a mandatory feature this right?

shekhar-rudder: Removed

vyeshwanth (approved):

PR #6325, opened 2025-09-08
chore: renaming mirroring modes

fracasula:
# Description

* Renaming mirroring modes to avoid confusion.
* Adding gauge to show current dedup modes in Grafana dashboards.

## Linear Ticket

< [Linear Link](https://linear.app/rudderstack/issue/PIPE-2373/keydb-follow-ups) >

## Security

### DIFF
--- a/services/dedup/db.go
+++ b/services/dedup/db.go
@@ -33,7 +33,7 @@ func NewDB(conf *config.Config, s stats.Stats, log logger.Logger) (types.DB, err
 	log = log.Withn(logger.NewStringField("mode", string(mirrorMode)))
 	log.Infon("Starting deduplication db")
 
-	gauge := func(primary, mirror string) {
+	modeGauge := func(primary, mirror string) {
 		s.NewTaggedStat("processor_dedup_mode", stats.GaugeType, stats.Tags{
 			"primary": primary,
 			"mirror":  mirror,
@@ -42,14 +42,14 @@ func NewDB(conf *config.Config, s stats.Stats, log logger.Logger) (types.DB, err
 
 	switch mirrorMode {
 	case badgerOnlyMode:
-		gauge("badger", "none")
+		modeGauge("badger", "none")
 		return badger.NewBadgerDB(conf, s, badger.DefaultPath())
 	case keyDBOnlyMode:
 		keydb, err := kdb.NewKeyDB(conf, s, log)
 		if err != nil {
 			return nil, fmt.Errorf("create keydb: %w", err)
 		}
-		gauge("keydb", "none")
+		modeGauge("keydb", "none")
 		return keydb, nil
 	case mirrorToKeyDB:
 		// primary is badger, then we mirror to keydb
@@ -59,11 +59,11 @@ func NewDB(conf *config.Config, s stats.Stats, log logger.Logger) (types.DB, err
 		}
 		mirror, err := kdb.NewKeyDB(conf, s, log)
 		if err != nil {
-			gauge("badger", "none")
+			modeGauge("badger", "none")
 			log.Errorn("Failed to create keydb mirror, falling back to badger only", obskit.Error(err))
 			return primary, nil
 		}
-		gauge("badger", "keydb")
+		modeGauge("badger", "keydb")
 		return NewMirrorDB(primary, mirror, mirrorToKeyDB, conf, s, log), nil
 	case mirrorToBadger:
 		// primary is keydb, then we mirror to badger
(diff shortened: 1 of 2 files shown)


services/dedup/db.go
```diff
@@ -29,48 +28,60 @@ func getMode(conf *config.Config) mode {
...
+			"primary": primary,
+			"mirror":  mirror,
+		}).Gauge(1)
+	}
+
 	switch mirrorMode {
 	case badgerOnlyMode:
-		return badger.NewBadgerDB(conf, stats, badger.DefaultPath())
+		gauge("badger", "none")
+		return badger.NewBadgerDB(conf, s, badger.DefaultPath())
```
ktgowtham, line 46: How about passing a primary true/false here(and for keydb) and use that label to add in relevant metrics(eg: `dedup_get_duration_seconds`)

fracasula: It seems more wasteful though because we'd have to pass `primary=true|false` to both badger and keydb meaning two more labels for gets and puts and other metrics inside there, vs one metric that we spawn only once on startup.

Also, the issue with histogram/counters is that you might not be able to see anything for a selected time range unless you have actual traffic. That's why I went for [this approach](https://github.com/rudderlabs/rudder-server/blob/caad8f078cba5f5a4b3acf93a643a2dd793f5b3b/runner/runner.go#L160).

This is how it would look like: https://rudderstack.grafana.net/goto/ZbnHLsrNR?orgId=1

services/dedup/db.go
```diff
@@ -29,48 +28,60 @@ func getMode(conf *config.Config) mode {
...
 // NewDB creates a new DB implementation based on configuration
-func NewDB(conf *config.Config, stats stats.Stats, log logger.Logger) (types.DB, error) {
+func NewDB(conf *config.Config, s stats.Stats, log logger.Logger) (types.DB, error) {
 	mirrorMode := getMode(conf)
-	log.Infon("starting deduplication db", logger.NewStringField("mode", string(mirrorMode)))
+	log = log.Withn(logger.NewStringField("mode", string(mirrorMode)))
+	log.Infon("Starting deduplication db")
+
+	gauge := func(primary, mirror string) {
+		s.NewTaggedStat("dedup_mode", stats.GaugeType, stats.Tags{
```
mihir20:
```suggestion
		s.NewTaggedStat("processor_dedup_mode", stats.GaugeType, stats.Tags{
```

services/dedup/db.go
```diff
@@ -16,10 +15,10 @@ import (
 type mode string
 
 const (
-	badgerOnlyMode   mode = "badger"
-	keyDBOnlyMode    mode = "keydb"
-	mirrorBadgerMode mode = "mirrorBadger"
```
atzoum, line 21: We are currently using `mirrorBadger` in the configuration ([link](https://github.com/search?q=repo%3Arudderlabs%2Frudder-devops%20mirrorBadger&type=code)). 
After a release containing this change, we would be falling back to `badger` until we fixed the config. Would that be acceptable?

services/dedup/db.go
```diff
@@ -29,48 +28,60 @@ func getMode(conf *config.Config) mode {
...
 
 // NewDB creates a new DB implementation based on configuration
-func NewDB(conf *config.Config, stats stats.Stats, log logger.Logger) (types.DB, error) {
+func NewDB(conf *config.Config, s stats.Stats, log logger.Logger) (types.DB, error) {
 	mirrorMode := getMode(conf)
-	log.Infon("starting deduplication db", logger.NewStringField("mode", string(mirrorMode)))
+	log = log.Withn(logger.NewStringField("mode", string(mirrorMode)))
+	log.Infon("Starting deduplication db")
+
+	gauge := func(primary, mirror string) {
```
atzoum: trivial
```suggestion
	modeGauge := func(primary, mirror string) {
```

services/dedup/db.go
```diff
@@ -29,48 +28,60 @@ func getMode(conf *config.Config) mode {
...
 			primary.Close()
 			return nil, fmt.Errorf("create badger mirror: %w", err)
 		}
-		return NewMirrorDB(primary, mirror, mirrorKeyDBMode, conf, stats, log), nil
+		gauge("keydb", "badger")
+		return NewMirrorDB(primary, mirror, mirrorToBadger, conf, s, log), nil
 	default:
-		log.Warnn("invalid mirror mode, falling back to badger only", logger.NewStringField("mode", string(mirrorMode)))
+		gauge("keydb", "none")
+		log.Warnn("Invalid mirror mode, falling back to badger only")
```
atzoum, line 83: why are we now excluding the requested mirror mode from the log?

services/dedup/db.go
fracasula: Yes, that would be OK :+1:

services/dedup/db.go
fracasula: We are not, because on top of this method I'm doing this:

```go
log = log.Withn(logger.NewStringField("mode", string(mirrorMode)))
```

ktgowtham (approved):

PR #6329, opened 2025-09-10
fix(router): kinesis rate exceeded errors are being aborted with http 400

koladilip:
## Description
added parseKinesisError function: Calls common.ParseAWSError and only changes the status code to 429 if the error is ProvisionedThroughputExceededException
Updated Produce method: Replaces the direct call to common.ParseAWSError with parseKinesisError

## Linear Ticket

Fixes [INT-4117](https://linear.app/rudderstack/issue/INT-4117/abort-throttled-kinesis-events-with-http-400-error)

## Impact
- Improves reliability of Kinesis stream processing when rate limits are encountered
- Prevents events from being aborted with HTTP 400 when they should be retried

## Testing
Test coverage: Verifies that ProvisionedThroughputExceededException returns 429 status code

## Security

### DIFF
--- a/services/streammanager/kinesis/kinesismanager.go
+++ b/services/streammanager/kinesis/kinesismanager.go
@@ -44,6 +44,14 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (common
 	return &KinesisProducer{client: kinesis.NewFromConfig(awsConfig)}, err
 }
 
+func parseKinesisError(err error) (int, string, string) {
+	statusCode, respStatus, responseMessage := common.ParseAWSError(err)
+	if respStatus == "ProvisionedThroughputExceededException" {
+		statusCode = 429
+	}
+	return statusCode, respStatus, responseMessage
+}
+
 // Produce creates a producer and send data to Kinesis.
 func (producer *KinesisProducer) Produce(jsonData json.RawMessage, destConfig interface{}) (int, string, string) {
 	client := producer.client
@@ -91,7 +99,7 @@ func (producer *KinesisProducer) Produce(jsonData json.RawMessage, destConfig in
 	}
 	putOutput, err := client.PutRecord(context.Background(), &putInput)
 	if err != nil {
-		statusCode, respStatus, responseMessage := common.ParseAWSError(err)
+		statusCode, respStatus, responseMessage := parseKinesisError(err)
 		pkgLogger.Errorn("[Kinesis] error",
 			logger.NewIntField("statusCode", int64(statusCode)),
 			logger.NewStringField("respStatus", respStatus),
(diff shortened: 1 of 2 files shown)


services/streammanager/common/common.go
```diff
@@ -27,7 +27,7 @@ type Opts struct {
 }
 
 func mapErrorMessageToStatusCode(errorMessage string, defaultStatusCode int) int {
-	if strings.Contains(errorMessage, "Throttling") {
+	if strings.Contains(errorMessage, "Throttling") || strings.Contains(errorMessage, "Rate exceeded") {
```
atzoum: This is incorrect, please pay attention to the API docs and our log
https://rudderlabs.slack.com/archives/C02H8RKK9AA/p1757484534928659?thread_ts=1757475893.288459&cid=C02H8RKK9AA

koladilip: can you explain the why this is incorrect?

koladilip: {"level":"ERROR","ts":"2025-09-09T20:20:04.999Z","logger":"streammanager.kinesis","caller":"kinesis/kinesismanager.go:95","msg":"[Kinesis] error","statusCode":400,"respStatus":"ProvisionedThroughputExceededException","responseMessage":"Rate exceeded for shard shardId-000000000159 in stream ShapedDatasetStream-3cc578f087ea520aefaaf92c802c1a34 under account 179641707330."}

we are using responseMessage as errorMessage for this function

atzoum: Please pay more attention, where did this come from?
```
"respStatus":"ProvisionedThroughputExceededException"
```

koladilip: Yes we could use this respStatus or responseMessage, since this pattern we are already using so I went with it, if want to something specific only for handling kinesis then we can go with ProvisionedThroughputExceededException

atzoum: I believe you are ignoring the **critical function** which is `ParseAWSError`

koladilip: Yes I rushed a bit here, respStatus coming means it is going to smithy API error block and won't go to the operation error block at all.

services/streammanager/kinesis/kinesismanager.go
```diff
@@ -44,6 +44,14 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (common
 	return &KinesisProducer{client: kinesis.NewFromConfig(awsConfig)}, err
 }
 
+func handleKinesisError(err error) (int, string, string) {
```
atzoum: trivial
```suggestion
func parseKinesisError(err error) (int, string, string) {
```

atzoum (commented):
pr description appears to be outdated

services/streammanager/kinesis/kinesismanager.go
koladilip: Make sense

koladilip: Done

koladilip:
> pr description appears to be outdated

Updated.

atzoum (approved):

PR #6338, opened 2025-09-12
feat(router): jobsdb pickup query batch size adapting to throttling limit

atzoum:
# Description

Routers with destination isolation enabled are now able to adapt their jobsdb pickup query batch size dynamically based on the destination's current throttling limit.

- This feature can be enabled through the hierarchical configuration options (default: **`false`**)
  - `Router.<DEST_TYPE>.pickupQueryThrottlingEnabled` 
  - `Router.pickupQueryThrottlingEnabled`.
- If the destination is using a single throttler for all event types, the pickup query batch size will match the current throttling limit.
- If the destination is using a different throttlers per event type, the pickup query batch size will match the sum of all current throttling limits that have been recently used (within `readSleepSeconds*2`).
- There is an upper limit on the maximum pickup query batch size controlled through (default: **`10000`**)
  - `Router.<DEST_TYPE>.maxJobQueryBatchSize` 
  - `Router.maxJobQueryBatchSize`

By dynamically adjusting the pickup query batch size to match the destination’s current throttling limits, routers no longer over-fetch jobs that cannot be processed immediately. This reduces the number of jobs queried from the database only to be discarded later due to throttling constraints. The result is lower query overhead, more efficient use of system resources, and faster end-to-end job processing since only jobs that can realistically be delivered are retrieved and no sleep penalty is added in the pickup loop due to less discarded jobs (for high job discard ratios (>60%) we impose a sleep penalty of 5 seconds per loop as a back-pressure mechanism).

## Linear Ticket

resolves PIPE-2375

## Security

### DIFF
--- a/router/handle.go
+++ b/router/handle.go
@@ -349,20 +349,26 @@ func (rt *Handle) pickup(ctx context.Context, partition string, workers []*worke
 // getAdaptedJobQueryBatchSize returns the adapted job query batch size based on the throttling limits
 func (*Handle) getAdaptedJobQueryBatchSize(input int, pickupThrottlers func() []throttler.PickupThrottler, readSleep time.Duration, maxLimit int) int {
 	jobQueryBatchSize := input
+	readSleepSeconds := int((readSleep + time.Second - 1) / time.Second) // rounding up to the nearest second
 	// Calculate the total limit of all active throttlers:
 	//
 	//   - if there is a global throttler, use its limit
 	//   - if there are event type specific throttlers, use the sum of their limits
 	totalLimit := lo.Reduce(pickupThrottlers(), func(totalLimit int, throttler throttler.PickupThrottler, index int) int {
-		if index == 0 {
-			return int(throttler.GetLimitPerSecond())
-		}
-		if throttler.GetEventType() == "all" {
+		switch throttler.GetEventType() {
+		case "all":
 			// global throttler, total limit is the limit of the first throttler
+			if index == 0 {
+				return int(throttler.GetLimitPerSecond())
+			}
+			return totalLimit
+		default:
+			// throttler per event type, total limit is the sum of all recently used throttler per event type limits
+			if int(time.Since(throttler.GetLastUsed()).Seconds()) <= readSleepSeconds*2 {
+				return totalLimit + int(throttler.GetLimitPerSecond())
+			}
 			return totalLimit
 		}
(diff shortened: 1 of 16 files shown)


router/handle.go
```diff
@@ -333,6 +346,38 @@ func (rt *Handle) pickup(ctx context.Context, partition string, workers []*worke
 	return pickupCount, limitsReached
 }
 
+// getAdaptedJobQueryBatchSize returns the adapted job query batch size based on the throttling limits
+func (*Handle) getAdaptedJobQueryBatchSize(input int, pickupThrottlers func() []throttler.PickupThrottler, readSleep time.Duration, maxLimit int) int {
+	jobQueryBatchSize := input
+	// Calculate the total limit of all active throttlers:
+	//
+	//   - if there is a global throttler, use its limit
+	//   - if there are event type specific throttlers, use the sum of their limits
```
mihir20, line 356: lets say we currently have a limit of 10 events per second for event types a, b, c, d.
Our logic fetches 40 jobs at a time.

In the database, the first 40 jobs happen to be all of type a, while the next 40 are a mix of b, c, d.
From the first batch, we can only process 10 jobs (due to the per-type limit) and end up discarding the remaining 30.

This results in a 75% discard ratio, which then triggers a sleep penalty.

Do you think if this is a problem? as if we had fetched beyond the first 40 jobs, we could have picked up additional b, c, d jobs and processed more events instead of discarding.

atzoum: Over-fetching or under-fetching will always remain a possibility. Please note that this feature is not enabled by default, and we may not enable it for all destinations using a per-event-type throttler—at least not without careful consideration first.
Also note that we are using the sum of active throttlers, i.e. if only events of type a are encountered by router, our pickup rate will be 10 not 40. If at some point events of type b are encountered, the rate will become 20. In the future we could improve by only considering throttlers that have been used during the past 5 seconds so that we don't take into consideration old, inactive throttlers.

Do you have any suggestions on how we should handle cases where a per-event-type throttler is in place?

mihir20: > we could improve by only considering throttlers that have been used during the past 5 seconds so that we don't take into consideration old, inactive throttlers.

yeah this can work 👍

atzoum: implementing this now 👍

mihir20 (approved):

PR #6344, opened 2025-09-15
fix: populate table level schemas createdAt updatedAt

achettyiitr:
# Description

- Populate correct `createdAt` and `updatedAt` for table-level schemas.

## Linear Ticket

- Resolves WAR-1116

## Security


warehouse/internal/repo/schema_test.go
```diff
@@ -73,7 +73,7 @@ func TestWHSchemasRepo(t *testing.T) {
 				require.NoError(t, err)
 
 				var (
-					now = time.Now().Truncate(time.Second).UTC()
+					now = time.Date(2023, 1, 1, 0, 0, 0, 0, time.UTC)
```
shekhar-rudder, line 76: Any reason this was updated?

shekhar-rudder (approved):

warehouse/internal/repo/schema_test.go
achettyiitr: It's better to use static time for tests.

PR #6346, opened 2025-09-15
chore: limit 1 while getting schemas for namespace

achettyiitr:
# Description

- We are unnecessarily scanning around **608MB** for `grafanainfs` and then taking the first entry and discarding others while scanning schemas. Instead, if we need one single entry, we can put a limit of 1, it’s coming around **32MB**. Notion [link](https://www.notion.so/rudderstacks/Warehouse-indexing-263f2b415dd08049a3f4f781f0bf087d?source=copy_link#26ff2b415dd0805597a0f1c764223139).
Screenshot 2025-09-15 at 1 54 57 PM

## Linear Ticket

- Resolves WAR-1131

## Security

### DIFF
--- a/warehouse/internal/repo/schema.go
+++ b/warehouse/internal/repo/schema.go
@@ -302,8 +302,8 @@ func (sh *WHSchema) getForNamespace(ctx context.Context, destID, namespace strin
 		namespace = $2 AND
 		table_name = ''
 	ORDER BY
-		source_id DESC;
-	`
+		source_id DESC
+	LIMIT 1;`
 
 	rows, err := sh.db.QueryContext(
 		ctx,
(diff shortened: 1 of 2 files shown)


warehouse/internal/repo/schema.go
```diff
@@ -302,8 +302,8 @@ func (sh *WHSchema) getForNamespace(ctx context.Context, destID, namespace strin
 		namespace = $2 AND
 		table_name = ''
 	ORDER BY
-		source_id DESC;
-	`
+		source_id DESC
+	LIMIT 1;`
```
achettyiitr, line 306: Below we are doing something like this:
```GO
    entries, err := parseWHSchemas(rows)
	if err != nil {
		return model.WHSchema{}, fmt.Errorf("parsing rows: %w", err)
	}
	if len(entries) == 0 {
		return model.WHSchema{}, nil
	}
	return *entries[0], err
```

shekhar-rudder (approved):

PR #6376, opened 2025-09-22
feat(processor): configurable delay option for event processing

atzoum:
# Description

Introducing support for configurable delays in the processor’s pipeline for specific partitions (typically sourceIDs). Delay for a specific partition can be configured using the `Processor.preprocessDelay.<partition>` configuration property.

**Example:**
```yaml
Processor.preprocessDelay.sourceId-1: 5m  
```
This ensures that events from `sourceId-1` are processed at least **5 minutes** after their `receivedAt` timestamp.

The processor will still pick up events as usual, but during the `preprocess` stage (after `receivedAt` is read), it will pause processing based on the most recent event in the batch. 
If the processor is stopped while waiting (e.g. shutdown or node migration), the sleep will be interrupted and events will not be processed prematurely. When the processor restarts, processing will resume only after the configured delay has elapsed.

## Linear Ticket

resolves PIPE-2392

## Security

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -59,7 +59,7 @@ require (
 	github.com/gomodule/redigo v1.9.2
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
-	github.com/grafana/jsonparser v0.0.0-20250430123630-2a684464cca1
+	github.com/grafana/jsonparser v0.0.0-20250908162026-5c2524e07b4c
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.27.1
 	github.com/hashicorp/go-retryablehttp v0.7.8
 	github.com/hashicorp/golang-lru/v2 v2.0.7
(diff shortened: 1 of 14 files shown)


ktgowtham (approved):
LGTM. 
but have a basic question: 
why do we have to go the limiter approach and not have a plain [sleeper](https://github.com/rudderlabs/rudder-server/blob/b19ace80276f0af6f56f0f391ef4d94ba553ea32/processor/internal/preprocessdelay/handle.go#L21)?

cisse21 (approved):

atzoum:
> LGTM. but have a basic question: why do we have to go the limiter approach and not have a plain [sleeper](https://github.com/rudderlabs/rudder-server/blob/b19ace80276f0af6f56f0f391ef4d94ba553ea32/processor/internal/preprocessdelay/handle.go#L21)?

If we sleep within the limiter's execution this means that we will be occupying a limiter slot for a longer period of time. limiter's slots are finite and there are occasions where these slots are less than the total number of active sources. In such a case one source's processing sleep would be causing processing delays in another source's processing

PR #6385, opened 2025-09-29
chore: cleanup unused reporting client code after migrating to commonclient

satishrudderstack:
# Description
- We are currently using the common reporting client for flushing the metrics to Reporting Service. As part of this PR, we are cleaning up un-used code.

## Linear Ticket

Fixes OBS-914

## Security


enterprise/reporting/error_reporting.go
itsmihir, line 65: we can remove this as well

PR #6386, opened 2025-09-29
fix: set table type to external for glue

achettyiitr:
# Description

- Set table type as external for Glue tables

## Linear Ticket

- Resolves WAR-1177

## Security


warehouse/integrations/datalake/schema-repository/glue.go
```diff
@@ -93,6 +94,7 @@ func (g *GlueSchemaRepository) CreateTable(ctx context.Context, tableName string
 	tableInput := &types.TableInput{
 		Name:          aws.String(tableName),
 		PartitionKeys: partitionKeys,
+		TableType:     aws.String("EXTERNAL_TABLE"),
```
shekhar-rudder, line 97: Can any string be passed as `TableType` or only some enums ?

warehouse/integrations/datalake/schema-repository/glue.go
```diff
@@ -271,7 +273,8 @@ func (g *GlueSchemaRepository) updateTable(ctx context.Context, tableName string
 	glueInput := &glue.UpdateTableInput{
 		DatabaseName: aws.String(g.Namespace),
 		TableInput: &types.TableInput{
-			Name: aws.String(tableName),
+			Name:      aws.String(tableName),
+			TableType: aws.String("EXTERNAL_TABLE"),
```
shekhar-rudder, line 277: `TableType cannot be null` error for the customer won't be resolved for existing tables unless there is a schema change (which would trigger a call to `updateTable` method). Is that correct?

warehouse/integrations/datalake/schema-repository/glue.go
achettyiitr: Any strings.

warehouse/integrations/datalake/schema-repository/glue.go
achettyiitr: Yes. Unless the update method is called, it won't affect.

shekhar-rudder (approved):

RanjeetMishra (approved):

PR #6387, opened 2025-09-29
refactor: oauth service

atzoum:
# Description

Refactoring the oauth service with main focus on `v2.OAuthHandler` (formerly `v2.Authorizer`) and ultimate goal to introduce circuit breakers in the near future during token retrieval.

### Improvements
- **Method signature cleanup:** 
  - Simplified and specialized method arguments.
  - Methods now return either a value or an error (instead of multiple values).
  - Sentinel error introduced for detecting invalid grant errors.
- **Removed redundancy:** Eliminated extra token expiration check and refresh in BingAds, since `OAuthHandler` already ensures token validity.
- **Type safety:** Narrowed-down cache for better correctness.
- **Renaming for clarity & consistency:**
  - `Authorizer` → `OAuthHandler`
  - `DefinitionName`/`DestDefName`/`DestName` → `DestType`
  - `AccountSecret` → `OAuthToken`
  - `RefreshTokenParams` → `OAuthTokenParams` (also cleaned up some fields)
  - `RefreshTokenBodyParams` → `OAuthTokenRequestBody`
  - `AuthStatusToggleParams` → `StatusRequestParams` (also cleaned up some fields)
- **Structural improvement:** `IsOAuthDestination` should not be a `DestinationInfo` method.

### Fixes
- Refreshing a cached token through control plane during `FetchToken` previously sent the old secret as `null`, preventing refresh.
- `AuthStatusToggle` always returned an error.
- During `preRoundTrip`, if `FetchToken` returned an invalid grant error, the following `AuthStatusToggle` call used an invalid status (`AUTH_STATUS_INACTIVE` instead of `inactive`).
- If the in-memory cache was empty, requests to the control-plane didn’t hint to refresh the resident token before returning it when expired.
- Concurrent modification issue with inflight status requests map in `AuthStatusToggle` (now using a dedicated mutex).
- Regulation worker wasn’t retrying OAuth errors because the way that the error message was injected into the JSON payload, was making it invalid. Test scenarios also missed this due to incorrect verification logic.

## Linear Ticket

resolves PIPE-2445

## Security


regulation-worker/internal/delete/api/api.go
```diff
@@ -156,7 +153,7 @@ func (m *APIManager) deleteWithRetry(ctx context.Context, job model.Job, destina
 		if transportResponse.InterceptorResponse.Response != "" {
 			pkgLogger.Debugn("Actual response received", logger.NewStringField("response", string(respBodyBytes)))
 			// Update the same error response to all as the response received would be []JobRespSchema
-			respBodyBytes, err = sjson.SetRawBytes(respBodyBytes, "#.error", []byte(transportResponse.InterceptorResponse.Response))
+			respBodyBytes, err = sjson.SetBytes(respBodyBytes, "#.error", []byte(transportResponse.InterceptorResponse.Response))
```
atzoum, line 156: **Note:** this is the bug that was corrupting the json response causing retries to never happen

regulation-worker/internal/delete/api/api.go
```diff
@@ -66,9 +66,6 @@ func (m *APIManager) GetSupportedDestinations() []string {
 }
 
 func (m *APIManager) deleteWithRetry(ctx context.Context, job model.Job, destination model.Destination, currentOauthRetryAttempt int) model.JobStatus {
-	if currentOauthRetryAttempt > m.MaxOAuthRefreshRetryAttempts {
-		return job.Status
-	}
```
atzoum, line 71: **Note:** another bug, if we return at this point job status will be `pending`...

router/batchrouter/asyncdestinationmanager/bing-ads/common/token.go
```diff
@@ -22,64 +22,27 @@ type TokenSource struct {
...
-	if currentTime.Before(expirationTime) {
-		return &secret, nil
-	}
-	refreshTokenParams.Secret = authResponse.Account.Secret
-	statusCode, authResponse, err = ts.OauthClientV2.RefreshToken(&refreshTokenParams)
-	// TODO: check if the error is not nil working verify during unit test
-	if err != nil {
-		return nil, fmt.Errorf("error in refreshing access token with this error: %w. StatusCode: %d", err, statusCode)
-	}
-	if err = jsonrs.Unmarshal(authResponse.Account.Secret, &secret); err != nil {
```
atzoum, line 82: **Note:** no need to check and refresh here since `OAuthHandler` is supposed to be doing this already

router/handle_lifecycle.go
```diff
@@ -142,14 +142,8 @@ func (rt *Handle) Setup(
...
 	)
 
-	destinationInfo := oauthv2.DestinationInfo{
-		DefinitionConfig: destinationDefinition.Config,
-		DefinitionName:   destinationDefinition.Name,
-	}
-
 	var err error
-
-	rt.isOAuthDestination, err = destinationInfo.IsOAuthDestination(common.RudderFlowDelivery)
```
atzoum, line 152: **Note:** having `IsOAuthDestination` a method of `DestinationInfo` is a bad design choice. One needs to know the internals of the method to figure out which `DestinationInfo` fields to populate for using the method. This is error-prone. Instead, having a clear function with all required parameters listed is much cleaner.

services/oauth/v2/http/client.go
```diff
@@ -17,7 +17,7 @@ type HttpClientOptionalArgs struct {
 	Transport          http.RoundTripper
 	Augmenter          oauthexts.Augmenter
 	Locker             *sync.PartitionRWLocker
-	OAuthHandler       *oauth.OAuthHandler
+	OAuthHandler       oauth.OAuthHandler
```
atzoum, line 20: **Note:** `OAuthHandler` is now an interface

services/oauth/v2/http/transport.go
```diff
@@ -176,55 +176,49 @@ func (t *OAuthTransport) postRoundTrip(rts *roundTripState) *http.Response {
...
-		if refErr != nil {
-			interceptorResp.Response = refErr.Error()
-		}
-		// refresh token success(retry)
-		// refresh token failed
-		// It can be failed due to the following reasons
-		// 1. invalid grant(abort)
-		// 2. control plan api call failed(retry)
-		// 3. some error happened while returning from RefreshToken function(retry)
-		if authResponse != nil && authResponse.Err == common.RefTokenInvalidGrant {
```
atzoum, line 205: **Note:** this illustrates how returning too many values from a method can make the code more complex when used

koladilip:
@atzoum changes are looking fine but they are many changes so we need to test them carefully right so do you need help on that?

atzoum:
> @atzoum changes are looking fine but they are many changes so we need to test them carefully right so do you need help on that?

Although I feel confident about these changes, it is true that we need to test the actual integrations. I suppose we don't have any automated integration tests for oauth? If so, yes I would need much help, but ideally we should be automating any tests we are performing.

koladilip (approved):

atzoum:
@koladilip 
- As part of this pull request I will work on introducing actual integration tests for the oauth service, i.e. add tests in rudder-server calling the actual oauth endpoints (config-be/rudder-auth).
- @pChondros can run the e2e tests for GARL
- Could you please help with distributing work for any other manual testing that is needed (and potentially come up with a plan for automating these soon)?

ItsSudip (approved):

cisse21 (approved):

PR #6388, opened 2025-09-29
fix: destination namespaces api to contain tablename filter

achettyiitr:
# Description

- Post adding table-level schemas, we would need to add a filter for `table_name` as well for the Destination Namespaces query, as there will be too many entries present. 
- More details in [notion doc](https://www.notion.so/rudderstacks/Warehouse-indexing-263f2b415dd08049a3f4f781f0bf087d?source=copy_link#279f2b415dd080b3bdf7fa5086082bd1).

## Linear Ticket

- Resolves WAR-1175

## Security

### DIFF
--- a/warehouse/internal/repo/schema.go
+++ b/warehouse/internal/repo/schema.go
@@ -597,14 +597,14 @@ func (sh *WHSchema) GetDestinationNamespaces(ctx context.Context, destinationID 
 			namespace
 		FROM ` + whSchemaTableName + `
 		WHERE destination_id = $1
+		AND table_name = ''
 		ORDER BY source_id, updated_at DESC;
 	`
-
 	rows, err := sh.db.QueryContext(ctx, query, destinationID)
 	if err != nil {
 		return nil, fmt.Errorf("querying destination namespaces: %w", err)
 	}
-	defer rows.Close()
+	defer func() { _ = rows.Close() }()
 
 	var mappings []model.NamespaceMapping
 	for rows.Next() {
@@ -615,10 +615,8 @@ func (sh *WHSchema) GetDestinationNamespaces(ctx context.Context, destinationID 
 		}
 		mappings = append(mappings, mapping)
 	}
-
 	if err = rows.Err(); err != nil {
 		return nil, fmt.Errorf("iterating over namespace mappings: %w", err)
 	}
-
 	return mappings, nil
 }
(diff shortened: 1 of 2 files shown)


warehouse/internal/repo/schema.go
```diff
@@ -596,15 +596,14 @@ func (sh *WHSchema) GetDestinationNamespaces(ctx context.Context, destinationID
...
+		WHERE destination_id = $1 AND table_name = ''
 		ORDER BY source_id, updated_at DESC;
 	`
-
 	rows, err := sh.db.QueryContext(ctx, query, destinationID)
 	if err != nil {
 		return nil, fmt.Errorf("querying destination namespaces: %w", err)
 	}
-	defer rows.Close()
+	defer func() { _ = rows.Close() }()
```
shekhar-rudder, line 607: Why this change?

achettyiitr: `rows.Close()` returns an error. It's good practice to replace the return values with `_` instead of ignoring it.

achettyiitr: It's been a standard practice across the rudder-server repo.

warehouse/internal/repo/schema_test.go
```diff
@@ -650,67 +650,53 @@ func TestWHSchemasRepo_GetForNamespace(t *testing.T) {
 
 func TestWHSchemasRepo_GetDestinationNamespaces(t *testing.T) {
 	destinationID := "test_destination_id"
-	conf := config.New()
-	db, ctx := setupDB(t), context.Background()
-	now := timeutil.Now()
-	r := repo.NewWHSchemas(db, conf, logger.NOP)
+	now := time.Date(2023, 1, 1, 0, 0, 0, 0, time.UTC)
```
achettyiitr, line 653: No changes in tests, just refactored it.

warehouse/internal/repo/schema.go
```diff
@@ -596,15 +596,14 @@ func (sh *WHSchema) GetDestinationNamespaces(ctx context.Context, destinationID
 			source_id,
 			namespace
 		FROM ` + whSchemaTableName + `
-		WHERE destination_id = $1
+		WHERE destination_id = $1 AND table_name = ''
```
shekhar-rudder: `fix:` -> If I understand this change correctly, it is a performance fix and not correctness fix, rt?

achettyiitr: It's a bug. When this API got introduced, table level schemas were also happening and this filter got missed.

shekhar-rudder (approved):

PR #6395, opened 2025-09-30
chore: improve populate table level schemas

achettyiitr:
# Description

- Improve table-level schemas.

## Linear Ticket

- Resolves WAR-1184

## Security


sql/migrations/warehouse_always/000001_alter_jsonb_to_text.up.tmpl
```diff
@@ -9,24 +9,24 @@
...
       FROM information_schema.columns
       WHERE table_name = 'wh_uploads' AND column_name = 'schema';
-      
+
       -- Handle wh_schemas table
-      IF wh_schemas_data_type != 'text' THEN
-        RAISE NOTICE 'Altering column "schema" in table "wh_schemas" from % to text', wh_schemas_data_type;
-        ALTER TABLE wh_schemas ALTER COLUMN schema TYPE TEXT USING schema::TEXT;
-        RAISE NOTICE 'Column "schema" in table "wh_schemas" successfully altered to text.';
+      IF wh_schemas_data_type != 'jsonb' THEN
```
shekhar-rudder, line 22: Should we move these "warehouse_always" migrations to the normal migrations list?

warehouse/internal/repo/schema.go
```diff
@@ -354,17 +354,16 @@ func (sh *WHSchema) populateTableLevelSchemasWithTx(ctx context.Context, tx *sql
...
+		LEFT JOIN wh_schemas s2
+		  ON s2.source_id = s.source_id
+		  AND s2.namespace = s.namespace
+		  AND s2.destination_id = s.destination_id
+		  AND s2.table_name = j.key
+		WHERE s.destination_id = $1
+		  AND s.namespace = $2
+		  AND s.table_name = ''
+		  AND s2.id IS NULL
+		ON CONFLICT (source_id, destination_id, namespace, table_name) DO NOTHING;
```
shekhar-rudder, line 366: We didn't have "ON CONFLICT" in the "NOT EXISTS" implementation. Is it required with "LEFT JOIN"?

achettyiitr: Yes. Earlier, we were running an INSERT query only for those which didn't exist. Now, since we are doing JOIN, we need to skip the ones which are already present. We are doing it via `ON CONFLICT`.

sql/migrations/warehouse_always/000001_alter_jsonb_to_text.up.tmpl
achettyiitr: Since all the schema based JSONB <-> TEXT are in one place, I think we should keep it here. Any way it will only run once when the conversion happens.
Also, we would be needing to switch back to TEXT once we completely moved back to table level schemas.

shekhar-rudder (approved):

RanjeetMishra (approved):

PR #6398, opened 2025-09-30
chore: table level schemas

achettyiitr:
# Description

- Table-level schemas

## Linear Ticket

- Resolves WAR-1186

## Security


sql/migrations/warehouse/000037_add_wh_schemas_dest_namespace_table_latest_idx.up.sql
```diff
@@ -0,0 +1,4 @@
+-- Add index on wh_schemas table to speed up queries on destination_id, namespace, and table_name columns (e.g. for schema snapshot query)
+CREATE INDEX CONCURRENTLY IF NOT EXISTS wh_schemas_dest_namespace_table_latest_idx
```
shekhar-rudder: Should we call the index "latest"? If we add another index in the future, it might be confusing

achettyiitr: It was a typo. It should be `wh_schemas_dest_namespace_table_level_idx`

PR #6409, opened 2025-10-01
fix: save webhook failures

mihir20:
# Description

## How It Works

1. When webhook requests are processed, they go through a batch transformation process
2. If any failures occur during transformation (either batch-level or individual request level), the system creates `FailedWebhookPayload` objects
3. These failed payloads are then passed to `SaveWebhookFailures` method
4. This method can be implemented by different services (like ingestion-svc) to handle these failures appropriately
5. In the context of ingestion-svc, these failures are published to proc errors for further processing

## Linear Ticket

pipe-2448

## Security

### DIFF
--- a/gateway/webhook/webhook.go
+++ b/gateway/webhook/webhook.go
@@ -15,6 +15,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/rudderlabs/rudder-server/gateway/webhook/model"
+
 	"github.com/samber/lo"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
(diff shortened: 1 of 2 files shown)


fracasula (dismissed):

fracasula:
@mihir20 if you target master we won't be able to get the new version from go.mod in ingestion-svc though

gateway/webhook/webhook.go
```diff
@@ -433,10 +435,25 @@ func (bt *batchWebhookTransformerT) batchTransformLoop() {
...
+			failedWebhookPayloads := make([]*model.FailedWebhookPayload, len(webRequests))
+			for i, webRequest := range webRequests {
+				failedWebhookPayloads[i] = &model.FailedWebhookPayload{
+					RequestContext: webRequest.authContext,
+					Payload:        payloadArr[i],
+					SourceType:     breq.sourceType,
+					Reason:         batchResponse.batchError.Error(),
+				}
+			}
+			if err := bt.webhook.gwHandle.SaveWebhookFailures(failedWebhookPayloads); err != nil {
```
atzoum, line 447: **question:** iirc we were already doing this in 1 place, are these missed failure scenarios that we're adding?

mihir20: No we removed this code while cleaning up proc errors, but its required for ingestion svc

PR #6411, opened 2025-10-02
[WIP] feat: oauth circuit breakers

atzoum:
# Description

## Linear Ticket

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/yandexmetrica/yandexmetrica_test.go
+++ b/router/batchrouter/asyncdestinationmanager/yandexmetrica/yandexmetrica_test.go
@@ -11,6 +11,7 @@ import (
 	. "github.com/onsi/gomega"
 	"go.uber.org/mock/gomock"
 
+	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
 	kitsync "github.com/rudderlabs/rudder-go-kit/sync"
@@ -50,7 +51,7 @@ var (
 var _ = Describe("Antisymmetric", func() {
 	Describe("NewManager function test", func() {
 		It("should return yandexmetrica manager", func() {
-			yandexmetrica, err := yandexmetrica.NewManager(logger.NOP, stats.NOP, destination, backendconfig.DefaultBackendConfig)
+			yandexmetrica, err := yandexmetrica.NewManager(config.Default, logger.NOP, stats.NOP, destination, backendconfig.DefaultBackendConfig)
 			Expect(err).To(BeNil())
 			Expect(yandexmetrica).NotTo(BeNil())
 		})
@@ -84,7 +85,7 @@ var _ = Describe("Antisymmetric", func() {
 				OAuthHandler: oauthHandler,
 			}
 			httpClient := httpClient.NewOAuthHttpClient(&http.Client{}, oauthv2common.RudderFlowDelivery, &cache, backendconfig.DefaultBackendConfig, augmenter.GetAuthErrorCategoryForYandex, &optionalArgs)
-			yandexmetrica, _ := yandexmetrica.NewManager(logger.NOP, stats.NOP, destination, backendconfig.DefaultBackendConfig)
+			yandexmetrica, _ := yandexmetrica.NewManager(config.Default, logger.NOP, stats.NOP, destination, backendconfig.DefaultBackendConfig)
 			yandexmetrica.Client = httpClient
 			asyncDestination := common.AsyncDestinationStruct{
 				ImportingJobIDs: []int64{1, 2, 3, 4},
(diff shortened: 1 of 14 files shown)


router/batchrouter/asyncdestinationmanager/yandexmetrica/yandexmetrica_test.go
```diff
@@ -50,7 +51,7 @@ var (
 var _ = Describe("Antisymmetric", func() {
 	Describe("NewManager function test", func() {
 		It("should return yandexmetrica manager", func() {
-			yandexmetrica, err := yandexmetrica.NewManager(logger.NOP, stats.NOP, destination, backendconfig.DefaultBackendConfig)
+			yandexmetrica, err := yandexmetrica.NewManager(config.Default, logger.NOP, stats.NOP, destination, backendconfig.DefaultBackendConfig)
```
fracasula: Why not `config.New()`?

router/transformer/transformer_test.go
```diff
@@ -339,7 +339,7 @@ func TestProxyRequest(t *testing.T) {
 				defer srv.Close()
 
 				expTimeDiff := config.SingleValueLoader(1 * time.Minute)
-				tr := NewTransformer(tc.rtTimeout, httpClientTimeout, nil, expTimeDiff, nil)
+				tr := NewTransformer("destType", tc.rtTimeout, httpClientTimeout, nil, expTimeDiff, nil, config.Default)
```
fracasula: Also here we can use `config.New` and avoid sharing the same global object between different tests (even if empty).

router/transformer/transformer_test.go
```diff
@@ -1752,7 +1752,7 @@ func TestTransformNoValidationErrors(t *testing.T) {
 	defer svr.Close()
 	t.Setenv("DEST_TRANSFORM_URL", svr.URL)
 	expTimeDiff := config.SingleValueLoader(1 * time.Minute)
-	tr := NewTransformer(time.Minute, time.Minute, nil, expTimeDiff, nil)
+	tr := NewTransformer("destType", time.Minute, time.Minute, nil, expTimeDiff, nil, config.Default)
```
fracasula: Same comment, this is done in different places.
Also, unrelated to this PR, but in this test file we use `assert.` thrice instead of `require` for some reason.

services/oauth/v2/oauth_breaker.go
```diff
@@ -0,0 +1,247 @@
...
+		// nested success breaker
+		result, err = fn()
+		if err != nil {
+			b.mu.Lock()
+			b.lastError = err
+			b.mu.Unlock()
+		}
+		return nil, err
+	}); ebErr != nil && // need to return the last error if the breaker is open
+		errors.Is(ebErr, gobreaker.ErrOpenState) ||
```
fracasula: Is the `||` wanted here without parenthesis? The 2nd `errors.Is` may run even if `ebErr` is `nil`.

services/oauth/v2/oauth_breaker.go
```diff
@@ -0,0 +1,247 @@
...
+		if err == nil && (previousValue == nil || string(previousValue) != string(result)) {
+			return nil, newTokenErr
+		}
+		return result, nil
+	}); sbErr != nil { // need to return the last value if the success breaker is open
+		if errors.Is(sbErr, gobreaker.ErrOpenState) || errors.Is(sbErr, gobreaker.ErrTooManyRequests) {
+			b.mu.RLock()
+			defer b.mu.RUnlock()
+			b.successBreakerCounter.Increment()
+			return b.lastValue, nil
```
fracasula: Could we get here without managing to update `b.lastValue` on line 211? :thinking:

services/oauth/v2/oauth_breaker.go
```diff
@@ -0,0 +1,247 @@
...
+		breakers: make(map[string]*accountBreaker),
+		opts:     opts,
+	}
+}
+
+// get returns the accountBreaker for the given account ID, creating it if it doesn't exist.
+func (b *accountBreakers) get(params *OAuthTokenParams) *accountBreaker {
+	b.breakersMu.RLock()
+
+	if brk, ok := b.breakers[params.AccountID]; ok {
```
fracasula, line 111: How many accountIDs could we potentially see? Given that there is no TTL or deletions, could this map grow indefinitely and cause memory issues after some time?

atzoum: The total number of accounts cannot be more than the total number of destinations that a server will encounter in its lifetime.

services/oauth/v2/oauth_breaker.go
atzoum: Not really, the success breaker opens whenever multiple different (5) results are returned instead of an error within 1m.

services/oauth/v2/oauth_breaker.go
```diff
@@ -0,0 +1,245 @@
...
+			errorBreakerCounter: b.opts.Stats.NewTaggedStat("oauth_action_error_breaker_tripped", stats.CountType, stats.Tags{
+				"id":          params.AccountID,
+				"workspaceId": params.WorkspaceID,
+				"destType":    params.DestType,
+			}),
+			successBreaker: gobreaker.NewCircuitBreaker(gobreaker.Settings{
+				Name: params.AccountID + "-success",
+				ReadyToTrip: func(counts gobreaker.Counts) bool {
+					// trip the breaker if there are more than successesThreshold successes (new tokens generated) within the interval
+					return int(counts.TotalFailures) >= b.opts.SuccessesThreshold
```
ktgowtham, line 141: We're comparing with `TotalFailures`. Is this correct?

atzoum: Yes, in the context of the breaker it is a failure, but for this specific breaker the successes count as failures 🙃

ktgowtham: ah rt.

vinayteki95 (approved):

PR #6416, opened 2025-10-07
fix(warehouse): int to float datatype handling

RanjeetMishra:
# Description

Enhanced the warehouse schema change handling to properly support integer to float datatype coercion across all warehouse destinations.

For all warehouses, if inferred datatype was int, we used to convert value to int explicitly. [Code Ref](https://github.com/rudderlabs/rudder-server/blob/master/warehouse/slave/worker.go#L404)

In case, inferred Data type from event payload was int and existingDatatype in WH was float, old logic in handleSchema used to explicitly check int type, else it used to consider it as error scenario. This used to work for all warehouses, as we used to convert value to int explicitly. [Code Ref](https://github.com/rudderlabs/rudder-server/blob/master/warehouse/slave/worker.go#L404)
For Snowpipe streaming, such mismatch meant int values will go to rudder_discards.

As all WHs support auto coercion of int to float, this PR considers all relevant value types to auto cast to float.

## Linear Ticket

resolves WAR-1198

## Security

### DIFF
--- a/warehouse/slave/worker.go
+++ b/warehouse/slave/worker.go
@@ -636,8 +636,8 @@ func (w *worker) destinationFromSlaveConnectionMap(destinationId, sourceId strin
 	return conn, nil
 }
 
-// HandleSchemaChange checks if the existing column type is compatible with the new column type
-func HandleSchemaChange(existingDataType, currentDataType model.SchemaType, value any) (any, error) {
+// HandleSchemaChange checks if the existing data type (from warehouse schema) is compatible with the inferred data type (from event data)
+func HandleSchemaChange(existingDataType, inferredDataType model.SchemaType, value any) (any, error) {
 	var (
 		newColumnVal any
 		err          error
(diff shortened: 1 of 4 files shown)


warehouse/slave/worker.go
```diff
@@ -636,45 +636,55 @@ func (w *worker) destinationFromSlaveConnectionMap(destinationId, sourceId strin
...
+			newColumnVal = float64(v)
+		case int32:
+			newColumnVal = float64(v)
+		case int64:
+			newColumnVal = float64(v)
+		case float32:
+			newColumnVal = float64(v)
+		case float64:
+			newColumnVal = v
+		default:
```
shekhar-rudder, line 668: Lets add a log for the default case so that we know that the code is not going there.

warehouse/slave/worker.go
```diff
@@ -636,45 +636,55 @@ func (w *worker) destinationFromSlaveConnectionMap(destinationId, sourceId strin
...
 		} else {
 			newColumnVal = value
 		}
-	} else if (currentDataType == model.IntDataType || currentDataType == model.BigIntDataType) && existingDataType == model.FloatDataType {
-		intVal, ok := value.(int)
-		if !ok {
-			err = fmt.Errorf("incompatible schema conversion from %v to %v", existingDataType, currentDataType)
-		} else {
-			newColumnVal = float64(intVal)
+	} else if (inferredDataType == model.IntDataType || inferredDataType == model.BigIntDataType) && existingDataType == model.FloatDataType {
```
shekhar-rudder, line 654: With the int/float logic inside `HandleSchemaChange`, should we remove it from the other place or do it in the next release?

RanjeetMishra: Added a followup task.https://linear.app/rudderstack/issue/WAR-1203/cleanup-int-float-datatype-handling-from-processsinglestagingfile

Even though confident,  would like to make the change in next release.

warehouse/slave/worker.go
RanjeetMishra: Adding log will be some decent change, need to pass logger. Any other ideas?

shekhar-rudder: No other ideas. Slave caller already has a logger. For snowpipe, the logger is present in caller's caller.

RanjeetMishra: Done
Added 1 additional test scenario for float, int but incompatible value to cover the code path.

shekhar-rudder (approved):

cisse21 (approved):

PR #6688, opened 2026-02-15
fix (gateway/handle.go): limit request body size in getPayloadFromReq…

akshit-git24:


gateway/handle.go
```diff
@@ -648,9 +648,20 @@ func (gw *Handle) getPayloadFromRequest(r *http.Request) ([]byte, error) {
 	start := time.Now()
 	defer gw.bodyReadTimeStat.Since(start)
 
+    //This line limits the reader to maxReqSize + 1 before reading the whole content.
+    r.Body = http.MaxBytesReader(nil,
```
atzoum: Currently, we don't apply request body size limits for `audienceList` events. Therefore, we don't enforce any limit here so that we can first parse the event and [decide later](https://github.com/rudderlabs/rudder-server/blob/db6b404ed8231b8530d57b35bafa78363f6510f1/gateway/handle.go#L468).
We need to figure out first if the requirement for not limiting `audienceList` events is still a valid one.

gateway/handle.go
```diff
@@ -648,9 +648,20 @@ func (gw *Handle) getPayloadFromRequest(r *http.Request) ([]byte, error) {
...
+		                         r.Body,int64(gw.conf.maxReqSize.Load())+1
```
atzoum: Why `maxReqSize+1`?

gateway/handle.go
```diff
@@ -648,9 +648,20 @@ func (gw *Handle) getPayloadFromRequest(r *http.Request) ([]byte, error) {
...
+								)
+
 	payload, err := io.ReadAll(r.Body)
 	_ = r.Body.Close()
 	if err != nil {
+
+        //This snippet of code returns the specific error from MaxBytesReader
+		if err.Error() == "http: request body too large" {
```
atzoum: `errors.AsType[*http.MaxBytesError]` would be more appropriate

akshit-git24:
> This PR is considered to be stale. It has been open 20 days with no further activity thus it is going to be closed in 7 days. To avoid such a case please consider removing the stale label manually or add a comment to the PR.

remove stale

atzoum:
@akshit-git24 due to audienceList we are not able to use the `http.MaxBytesReader`. Closing this pr.

PR #6872, opened 2026-04-11
chore: avoid marshalling proc error job payloads

atzoum:
## Description

Eliminate an expensive `jsonrs.Marshal` on the processor hot path by carrying the original parsed events alongside proc error jobs instead of serializing them into `EventPayload`.

image

### Problem

When building proc error jobs for failed/filtered transformer responses, the processor was:
1. Unmarshalling the original job payload into `[]SingularEventT` (`[]map[string]any`)
2. Re-marshalling the slice back to JSON for `EventPayload` — expensive due to reflection on `map[string]any`
3. The only consumer (`recordEventDeliveryStatus`, the live debugger) would immediately unmarshal it back, then re-marshal each individual event

### Solution

Introduce a `procErrorJob` wrapper that carries the original `[]SingularEventT` alongside the `*jobsdb.JobT`:
- **`getTransformationMetrics`**: no longer marshals events into `EventPayload`
- **`recordEventDeliveryStatus`**: reads `.events` directly, extracts event name/type via map lookup instead of marshal + gjson
- **`droppedJobs`** (`saveDroppedJobs`/rsources): only reads `Parameters`, `JobID`, `UUID` — never touched `EventPayload`, so nil is fine

The only remaining per-event marshal is inside `recordEventDeliveryStatus` for the `DeliveryStatusT.Payload` field, but that only runs when the live debugger is active (rare).

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -276,11 +276,22 @@ type MetricMetadata struct {
 	trackingPlanVersion     int
 }
 
+// procErrorJob wraps a jobsdb.JobT with the original parsed events,
+// avoiding an expensive marshal/unmarshal round-trip through EventPayload.
+type procErrorJob struct {
+	*jobsdb.JobT
+	events []types.SingularEventT
+}
+
+func procErrorJobs(jobs []procErrorJob) []*jobsdb.JobT {
+	return lo.Map(jobs, func(j procErrorJob, _ int) *jobsdb.JobT { return j.JobT })
+}
+
 type NonSuccessfulTransformationMetrics struct {
-	failedJobs       []*jobsdb.JobT
+	failedJobs       []procErrorJob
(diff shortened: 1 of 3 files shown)


fracasula (approved):

processor/processor.go
```diff
@@ -276,11 +276,22 @@ type MetricMetadata struct {
 	trackingPlanVersion     int
 }
 
+// procErrorJob wraps a jobsdb.JobT with the original parsed events,
+// avoiding an expensive marshal/unmarshal round-trip through EventPayload.
+type procErrorJob struct {
+	*jobsdb.JobT
```
mihir20, line 282: Should we also add a comment mentioning jobT will not hold eventPayload

mihir20 (approved):

PR #6873, opened 2026-04-12
feat: add configurable batch size for bulk inserting failed rsources records

mihir20:
# Description

Txn Inserting into `rsources_failed_keys_v2_records` is taking too much time. The core bottleneck is `AddFailedRecords()` for rsources_stats. It uses tx.Prepare + a loop of individual stmt.ExecContext calls. Each ExecContext is a separate network round-trip to Postgres. If a batch
has 1,000 aborted jobs with record_ids, that's 1,000 sequential INSERT round-trips all within one transaction.

- Introduced batch processing for inserting failed records in `rsources_failed_keys_v2_records`
- Added FailedRecordsInsertBatchSize configuration to control batch size
- Implemented batch insert SQL statement with [unnest](https://www.tigerdata.com/blog/boosting-postgres-insert-performance)
- Updated unit tests to cover multi-batch scenarios
- Ensured default batch size is set when configuration is invalid

We could have used the COPY command, but the insert needs `ON CONFLICT (id, record_id) DO UPDATE SET ts = NOW()`, and COPY doesn't support ON CONFLICT

## Linear Ticket

-

## Security

### DIFF
--- a/app/apphandlers/setup.go
+++ b/app/apphandlers/setup.go
@@ -70,6 +70,7 @@ func NewRsourcesService(deploymentType deployment.Type, shouldSetupSharedDB bool
 	}
 	rsourcesConfig.SharedConn = sharedDBConnUrl
 	rsourcesConfig.SkipFailedRecordsCollection = !config.GetBoolVar(true, "Router.failedKeysEnabled")
+	rsourcesConfig.FailedRecordsInsertBatchSize = config.GetIntVar(5000, 1, "Rsources.failedRecordsInsertBatchSize")
 
 	if deploymentType == deployment.MultiTenantType {
 		// For multitenant deployment type we shall require the existence of a SHARED_DB
--- a/app/apphandlers/setup.go
+++ b/app/apphandlers/setup.go
@@ -70,7 +70,7 @@ func NewRsourcesService(deploymentType deployment.Type, shouldSetupSharedDB bool
 	}
 	rsourcesConfig.SharedConn = sharedDBConnUrl
 	rsourcesConfig.SkipFailedRecordsCollection = !config.GetBoolVar(true, "Router.failedKeysEnabled")
-	rsourcesConfig.FailedRecordsInsertBatchSize = config.GetIntVar(5000, 1, "Rsources.failedRecordsInsertBatchSize")
+	rsourcesConfig.FailedRecordsInsertBatchSize = config.GetReloadableIntVar(5000, 1, "Rsources.failedRecordsInsertBatchSize")
 
 	if deploymentType == deployment.MultiTenantType {
 		// For multitenant deployment type we shall require the existence of a SHARED_DB
--- a/services/rsources/handler.go
+++ b/services/rsources/handler.go
@@ -128,10 +128,7 @@ func (sh *sourcesHandler) AddFailedRecords(ctx context.Context, tx *sql.Tx, jobR
 		return fmt.Errorf("scanning rsources_failed_keys_v2 id: %w", err)
 	}
 
-	batchSize := 5000
-	if sh.config.FailedRecordsInsertBatchSize != nil && sh.config.FailedRecordsInsertBatchSize.Load() > 0 {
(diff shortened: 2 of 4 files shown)


services/rsources/handler.go
```diff
@@ -128,19 +129,36 @@ func (sh *sourcesHandler) AddFailedRecords(ctx context.Context, tx *sql.Tx, jobR
 		return fmt.Errorf("scanning rsources_failed_keys_v2 id: %w", err)
 	}
 
-	stmt, err := tx.Prepare(`INSERT INTO rsources_failed_keys_v2_records (id, record_id, code) VALUES ($1, $2, $3) ON CONFLICT (id, record_id) DO UPDATE SET ts = NOW()`)
-	if err != nil {
-		return err
+	batchSize := sh.config.FailedRecordsInsertBatchSize
+	defaultFailedRecordsBatchSize := 5000
+	if batchSize <= 0 {
+		batchSize = defaultFailedRecordsBatchSize
```
atzoum, line 136: need an upper limit based on the max number of parameters allowed

ktgowtham (approved):
LGTM

ktgowtham (approved):

services/rsources/handler.go
```diff
@@ -128,19 +129,37 @@ func (sh *sourcesHandler) AddFailedRecords(ctx context.Context, tx *sql.Tx, jobR
...
-			records[i].Record,
-			records[i].Code,
-		); err != nil {
+	for start := 0; start < len(records); start += batchSize {
+		end := min(start+batchSize, len(records))
+		batch := records[start:end]
+
+		var sb strings.Builder
+		args := make([]any, 0, len(batch)*3)
+		sb.WriteString(`INSERT INTO rsources_failed_keys_v2_records (id, record_id, code) VALUES `)
```
atzoum: Using [unnest](https://www.tigerdata.com/blog/boosting-postgres-insert-performance) might be faster and we would no longer need to have a limit on the number of parameters.
Batching would still be needed and 5K is an acceptable default

mihir20: thanks, updated the pr to use unnest

app/apphandlers/setup.go
```diff
@@ -70,6 +70,7 @@ func NewRsourcesService(deploymentType deployment.Type, shouldSetupSharedDB bool
 	}
 	rsourcesConfig.SharedConn = sharedDBConnUrl
 	rsourcesConfig.SkipFailedRecordsCollection = !config.GetBoolVar(true, "Router.failedKeysEnabled")
+	rsourcesConfig.FailedRecordsInsertBatchSize = config.GetIntVar(5000, 1, "Rsources.failedRecordsInsertBatchSize")
```
atzoum: let's make this reloadable

mihir20: done

services/rsources/handler.go
```diff
@@ -128,18 +128,29 @@ func (sh *sourcesHandler) AddFailedRecords(ctx context.Context, tx *sql.Tx, jobR
...
 
-	stmt, err := tx.Prepare(`INSERT INTO rsources_failed_keys_v2_records (id, record_id, code) VALUES ($1, $2, $3) ON CONFLICT (id, record_id) DO UPDATE SET ts = NOW()`)
-	if err != nil {
-		return err
-	}
-	defer func() { _ = stmt.Close() }()
+	batchSize := 5000
+	if sh.config.FailedRecordsInsertBatchSize != nil && sh.config.FailedRecordsInsertBatchSize.Load() > 0 {
+		batchSize = sh.config.FailedRecordsInsertBatchSize.Load()
+	}
```
atzoum: this should never be nil, we should take care of it in the constructor
```go
	if jsc.FailedRecordsInsertBatchSize == nil {
		jsc.FailedRecordsInsertBatchSize = config.SingleValueLoader(5000)
	}
```
```suggestion
	batchSize := max(1, sh.config.FailedRecordsInsertBatchSize.Load())
```

atzoum (approved):

PR #6896, opened 2026-04-21
feat: add native deletion support for identity removal in gzip files

mihir20:
# Description

`GZIPLocalFileHandler.RemoveIdentity` uses `exec("bash -c sed ...")` to strip suppressed user records from an NDJSON gzip file. Two problems with that:

1. **Shell injection surface**: every suppressed user ID is interpolated into a bash command. We sanitise with `regexp.QuoteMeta`, but those quotes *regex* metacharacters — it does not escape *shell*  
   metacharacters. Inputs like `$(...)` or backticks remain risky.
2. **Does not scale with suppression list size**: sed runs every pattern against every line, so cost grows O(records × users). For 10k records × 100 users, it takes ~1.27s (1 MB/s).

As part of this change, we are using a feature flag and adding native Go support for removing identity. we will be iterating line by line and deleting the lines that have the suppressed userIDs.

[Benchmarks](https://github.com/rudderlabs/rudder-server/pull/6888)

## Linear Ticket

pipe-2904

## Security


ktgowtham (approved):
LGTM. as long as s3 and s3 data lake has the data in the format that this PR expects. assuming we tested that too.

achettyiitr (approved):

fracasula (approved):

aris1009 (approved):

PR #6905, opened 2026-04-23
chore(vuln): zizmor --fix=all findings (SEC-199)

aris1009:
Remediate zizmor findings via `zizmor --fix=all` (SEC-199). Scope limited to `.github/`.


lokey (approved):
Approved: SEC-199 workflow-only changes

PR #6911, opened 2026-04-24
chore: scope workflow permissions to least privilege (SEC-167)

aris1009:
Remediate zizmor excessive-permissions (SEC-167).

actionlint: no regressions before/after.

### DIFF
--- a/.github/workflows/scheduled-release.yml
+++ b/.github/workflows/scheduled-release.yml
@@ -13,7 +13,7 @@ on:
         type: string
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   check-schedule:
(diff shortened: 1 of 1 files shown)


Mrsandeep27 (commented):
Right move — workflow-level least privilege with job-level escalation is the GitHub-recommended pattern. Two thoughts:

**1. Only one job (\`run-release\`) gets \`contents: write\` back.** Worth confirming all *other* jobs in this file (\`check-schedule\`, plus any earlier job from the un-shown portion of the diff) actually only need \`contents: read\`. \`step-security/harden-runner\` warns when a job over-asks; if any of them write tags, push, or create branches, the workflow-level cap will silently break them.

**2. Audit the *other* workflows in this repo** for the same hardening. Token compromise is most useful in workflows that don't touch the release path — \`pr-checks.yml\`, \`scheduled-anything.yml\` etc. — because attackers can stage their access there. Worth a follow-up SEC ticket to sweep all \`.github/workflows/\` files, not just this one.

**3. \`permissions: contents: read\`** at workflow level — since this is the GitHub Actions default for public repos but *not* for some org-level configurations (e.g. write-by-default), explicit is the right move. Bonus: also consider \`pull-requests: read\` if the workflow ever inspects PR data — leaving it unspecified means it inherits whatever the org settings allow.

LGTM on the change itself.

lokey (approved):
Approved: SEC-167 workflow-only changes

PR #6914, opened 2026-04-27
chore: extend transformer client perpetual retries stat

atzoum:
# Description

Extends the `transformer_client_perpetual_retry_count` stat with additional dimensions to give better visibility into perpetual transformer retries:

- **`name` tag**: identifies which transformer client emitted the metric (e.g. `SourceTransformer`, `DestinationTransformer`, `UserTransformer`, `TrackingPlanValidation`, `SourceHydration`, `RouterTransformer`, `TransformerProxy`).
- **`attempt` tag**: the retry attempt number, capped at `5+` to keep cardinality bounded (values `0`, `1`, `2`, `3`, `4`, `5+`).
- **Per-request extra tags via context**: callers can attach extra low-cardinality tags to a request's context using `transformerclient.WithPerpetualRetriesStatsTags(ctx, map[string]string{...})`; the retry strategy merges them into the stat. `UserTransformer` uses this to add a `language` tag (`javascript` / `python`) so JS vs Python user-transformer retries can be differentiated even though they share the same client.

To plumb the `name` through, `transformerclient.NewClient` now takes a `name` parameter, propagated to the custom retry strategy. All call sites have been updated.

Also bumps `rudder-go-kit` to pick up the `attempt int` argument added to the custom retry strategy callback.

## Linear Ticket

N/A

## Security

### DIFF
--- a/gateway/webhook/setup.go
+++ b/gateway/webhook/setup.go
@@ -100,7 +100,7 @@ func Setup(gwHandle Gateway, transformerFeaturesService TransformerFeaturesServi
 	transformerClientConfig.Recycle = conf.GetBoolVar(false, "Transformer.Client.sourceTransformer.recycle", "Transformer.Client.recycle")
 	transformerClientConfig.RecycleTTL = conf.GetDurationVar(60, time.Second, "Transformer.Client.sourceTransformer.recycleTTL", "Transformer.Client.recycleTTL")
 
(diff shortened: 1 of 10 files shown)


fracasula (commented):
Would there be a way to propagate the language for `UserTransformer` at least?

fracasula (approved):

ktgowtham (approved):

PR #6921, opened 2026-04-28
chore: filtering overflow events for snowpipe streaming above 16MB

achettyiitr:

# Description

Snowpipe Streaming insert requests have a [maximum 'insertRows' size](https://github.com/snowflakedb/snowflake-ingest-java/blob/master/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java#L721). Previously, large batches could exceed this limit and fail. This change ensures we stay within the limit and fail only the overflow events deterministically.

- Compute per-event payload size from raw `message.data` JSON and store it as `MessageDataByteSize`.
- While uploading, group events by table and filter out events that would cause a table insert request to exceed the 16MB limit (`maxInsertRequestSizeBytes`).
- Mark any overflow events as failed with reason: `some events were not uploaded because they exceeded the max insert request size: 16777216`.

## Linear Ticket

- Resolves INT-5890

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
@@ -38,8 +38,6 @@ import (
 	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
 )
 
-const maxInsertRequestSizeBytes = 16 * bytesize.MB
-
 func New(
 	conf *config.Config,
 	log logger.Logger,
@@ -73,6 +71,7 @@ func New(
 	m.config.instanceID = conf.GetStringVar("1", "INSTANCE_ID")
 	m.config.maxBufferCapacity = conf.GetReloadableInt64Var(512*bytesize.KB, bytesize.B, "SnowpipeStreaming.maxBufferCapacity")
 	m.config.stuckPipelineThreshold = conf.GetReloadableDurationVar(15, time.Minute, "SnowpipeStreaming.stuckPipelineThresholdInMinutes")
+	m.config.maxInsertRequestSizeBytes = conf.GetReloadableInt64Var(16*bytesize.MB, bytesize.B, "SnowpipeStreaming.maxInsertRequestSizeBytes")
 
 	tags := stats.Tags{
 		"module":        "batch_router",
@@ -206,9 +205,10 @@ func (m *Manager) Upload(_ context.Context, asyncDest *common.AsyncDestinationSt
 		return event.Message.Metadata.Table
 	})
 
-	filteredGroupedEvents, overflowedEvents := m.filterOutEventsThatExceedMaxInsertRequestSize(groupedEvents)
+	maxInsertRequestSizeBytes := m.config.maxInsertRequestSizeBytes.Load()
+	includedGroupedEvents, overflowedEvents := splitEventsExceedingMaxInsertRequestSize(groupedEvents, maxInsertRequestSizeBytes)
 
-	uploadInfos := lo.MapToSlice(filteredGroupedEvents, func(tableName string, tableEvents []*event) *uploadInfo {
+	uploadInfos := lo.MapToSlice(includedGroupedEvents, func(tableName string, tableEvents []*event) *uploadInfo {
 		jobIDs := lo.Map(tableEvents, func(event *event, _ int) int64 {
 			return event.Metadata.JobID
 		})
@@ -359,10 +359,10 @@ func (m *Manager) eventsFromFile(fileName string, eventsCount int) ([]*event, er
 	return events, nil
 }
 
-// filterOutEventsThatExceedMaxInsertRequestSize filters out events that exceed the max insert request size
-// and returns the filtered grouped events and the overflowed events.
-func (m *Manager) filterOutEventsThatExceedMaxInsertRequestSize(groupedEvents map[string][]*event) (map[string][]*event, []*event) {
-	filteredGroupedEvents := make(map[string][]*event, len(groupedEvents))
+// splitEventsExceedingMaxInsertRequestSize splits grouped events into included and overflowed events
+// based on the configured max insert request size.
+func splitEventsExceedingMaxInsertRequestSize(groupedEvents map[string][]*event, maxInsertRequestSizeBytes int64) (map[string][]*event, []*event) {
+	includedGroupedEvents := make(map[string][]*event, len(groupedEvents))
 	overflowedEvents := make([]*event, 0)
 	for tableName, tableEvents := range groupedEvents {
 		// Only account for the insert rows payload size:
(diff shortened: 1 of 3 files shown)


router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -203,7 +205,10 @@ func (m *Manager) Upload(_ context.Context, asyncDest *common.AsyncDestinationSt
 	groupedEvents := lo.GroupBy(events, func(event *event) string {
 		return event.Message.Metadata.Table
 	})
-	uploadInfos := lo.MapToSlice(groupedEvents, func(tableName string, tableEvents []*event) *uploadInfo {
+
+	filteredGroupedEvents, overflowedEvents := m.filterOutEventsThatExceedMaxInsertRequestSize(groupedEvents)
```
shekhar-rudder: As mentioned in the SDK log, why can't we chunk the event into multiple batches in snowpipe service instead of handling it in rudder-server?

achettyiitr: No, we can't chunk in the Snowpipe service. Event `/insert` is associated with an offset. If we are going to chunk on the snowpipe service. What offsets are you going to provide for multiple calls?

It's possible that for insert 1 + offset 1 data got flushed successfully, but for insert 2 and offset 1 it failed. If that happens, we would be marking all events corresponding to insert 2 as succeeded.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -38,6 +38,8 @@ import (
 	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
 )
 
+const maxInsertRequestSizeBytes = 16 * bytesize.MB
```
krishna2020: the current change looks good. Other thought - can this be moved to file preparation phase instead ?

achettyiitr: No. During the file preparation phase, we don't have access to the event Name.

krishna2020 (approved):

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -273,6 +278,15 @@ func (m *Manager) Upload(_ context.Context, asyncDest *common.AsyncDestinationSt
 	if discardImportInfo != nil {
 		importInfos = append(importInfos, discardImportInfo)
 	}
+	if len(overflowedEvents) > 0 {
+		failedJobIDs = append(failedJobIDs, lo.Map(overflowedEvents, func(event *event, _ int) int64 {
```
shekhar-rudder, line 282: Shouldn't it be aborted jobs instead of failed jobs? Are we expecting them to succeed in the next attempt?

achettyiitr: Yeah, overflow jobs came because the async file contained data for that table more than 16MB. We will be marking them as failed, and in the next attempt, they will succeed.

krishna2020: they can succeed, they are getting marked as failed because of buffer being full in the current request.
But there can be other case where event payload size itself can be 16MB (edge scenario) where abort should happen

achettyiitr: > But there can be other case where event payload size itself can be 16MB (edge scenario) where abort should happen

That can only happen via UT right, we don't ingest such a payload in the Gateway. In such a case, upon multiple retries, it will be marked as aborted.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -324,11 +338,18 @@ func (m *Manager) eventsFromFile(fileName string, eventsCount int) ([]*event, er
...
+
 		var e event
-		if err := jsonrs.Unmarshal(scanner.Bytes(), &e); err != nil {
+		if err := jsonrs.Unmarshal(line, &e); err != nil {
 			return nil, fmt.Errorf("failed to unmarshal event: %w", err)
 		}
-		e.setUUIDTimestamp(formattedTS)
+		isUUIDTimestampSet := e.setUUIDTimestamp(formattedTS)
+
+		e.MessageDataByteSize = int64(len(gjson.GetBytes(line, "message.data").Raw))
```
shekhar-rudder, line 349: How about encapsulating `e.MessageDataByteSize` inside `setUUIDTimestamp` method?

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
shekhar-rudder: Should we make it configurable?

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -338,6 +359,35 @@ func (m *Manager) eventsFromFile(fileName string, eventsCount int) ([]*event, er
 	return events, nil
 }
 
+// filterOutEventsThatExceedMaxInsertRequestSize filters out events that exceed the max insert request size
+// and returns the filtered grouped events and the overflowed events.
+func (m *Manager) filterOutEventsThatExceedMaxInsertRequestSize(groupedEvents map[string][]*event) (map[string][]*event, []*event) {
```
shekhar-rudder: Instead of `filter` should we call it `split` since filter usually means to remove

shekhar-rudder: It can be a plain function since it doesn't use any manager fields

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -273,6 +278,15 @@ func (m *Manager) Upload(_ context.Context, asyncDest *common.AsyncDestinationSt
 	if discardImportInfo != nil {
 		importInfos = append(importInfos, discardImportInfo)
 	}
+	if len(overflowedEvents) > 0 {
```
shekhar-rudder, line 281: Lets add a comment saying something like these events are expected to succeed in the next retry and not abort since there is jobid based ordering

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
achettyiitr: It's better to handle it here, because in future if some more modifications come, we can compute it.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
achettyiitr: Not sure if such a comment is helpful, as it is a general behaviour for failure scenarios. `Marking any events as failed and having them picked up in the next processing based on jobid-based ordering is server-specific. It happens for all other failure scenarios.`

shekhar-rudder: `It happens for all other failure scenarios.` -> I felt that it was not obvious why it would succeed in the next attempt and not fail again

shekhar-rudder (approved):

PR #6926, opened 2026-04-29
feat: snowpipe batch status

achettyiitr:

# Description

Adds bulk status polling support for Snowpipe Streaming and integrates it across manager, API adapter, API client, and tests.

- Added bulk status API plumbing:
  - `GetBulkStatus` in `internal/api/api.go`
  - `GetBulkStatus` in `apiadapter.go`
  - `BulkStatusResponse` in `internal/model/model.go`
  - API interface update in `types.go`
- Added config flags:
  - `SnowpipeStreaming.bulkStatusEnabled`
  - `SnowpipeStreaming.bulkStatusChannels`
- Updated polling flow in `snowpipestreaming.go`:
  - routes polling through bulk path when enabled
  - chunks channel IDs for bulk calls
  - handles channel recovery in bulk mode:
    - `notFound` => recreate + re-poll
    - invalid status (`valid=false`) => delete + recreate + re-poll
  - retains existing in-progress/terminal semantics via `isInProgress`
- Added integration test configurability:
  - `runRudderServer(..., configOverrides...)` now accepts runtime config overrides
- Added new integration scenario:
  - `many table (bulk status enabled)` in `integration_test/snowpipestreaming/snowpipestreaming_test.go`

## Linear Ticket

- Resolves INT-5887

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/internal/model/model.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/internal/model/model.go
@@ -76,6 +76,12 @@ type (
 		Valid                bool   `json:"valid"`
 		LatestInsertedOffset string `json:"latestInsertedOffset"`
 	}
+
+	BulkStatusResponse struct {
+		Success  bool                       `json:"success"`
+		Statuses map[string]*StatusResponse `json:"statuses"`
+		NotFound []string                   `json:"notFound"`
+	}
 )
 
 func (c *ChannelResponse) UnmarshalJSON(data []byte) error {
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
@@ -74,7 +74,7 @@ func New(
 	m.config.maxBufferCapacity = conf.GetReloadableInt64Var(512*bytesize.KB, bytesize.B, "SnowpipeStreaming.maxBufferCapacity")
 	m.config.stuckPipelineThreshold = conf.GetReloadableDurationVar(15, time.Minute, "SnowpipeStreaming.stuckPipelineThresholdInMinutes")
 	m.config.bulkStatusEnabled = conf.GetReloadableBoolVar(false, "SnowpipeStreaming.bulkStatusEnabled")
-	m.config.bulkStatusChannels = conf.GetReloadableIntVar(100, 1, "SnowpipeStreaming.bulkStatusChannels")
+	m.config.bulkStatusBatchSize = conf.GetReloadableIntVar(100, 1, "SnowpipeStreaming.bulkStatusBatchSize")
 
 	tags := stats.Tags{
 		"module":        "batch_router",
@@ -847,9 +847,10 @@ func isInProgress(statusRes *model.StatusResponse, info *importInfo, log logger.
 
 func (m *Manager) getBulkStatus(ctx context.Context, channelIDs []string) (map[string]*model.StatusResponse, []string, error) {
 	statuses := make(map[string]*model.StatusResponse)
-	notFound := make([]string, 0)
 
-	channelIDChunks := lo.Chunk(channelIDs, m.config.bulkStatusChannels.Load())
+	var notFound []string
+
+	channelIDChunks := lo.Chunk(channelIDs, m.config.bulkStatusBatchSize.Load())
 	for i := range channelIDChunks {
 		channelIDChunk := channelIDChunks[i]
 		bulkStatusRes, err := m.api.GetBulkStatus(ctx, channelIDChunk)
@@ -872,20 +873,14 @@ func (m *Manager) getNonProcessedImportInfosWithBulkStatus(ctx context.Context, 
 
 	statuses, notFound, err := m.getBulkStatus(ctx, channelIDs)
 	if err != nil {
-		for i := range infos {
-			info := infos[i]
-			info.FailedJobIds = nil
-			info.Failed = true
-			info.Reason = err.Error()
-			m.polledImportInfoMap[info.ChannelID] = info
-		}
-		return nil
+		// In case of failure, we return all the import infos to be polled again
+		m.logger.Warnn("Failed to get bulk status", obskit.Error(err))
+		return infos
 	}
 
-	notFoundMap := make(map[string]struct{}, len(notFound))
-	for _, channelID := range notFound {
-		notFoundMap[channelID] = struct{}{}
-	}
+	notFoundMap := lo.SliceToMap(notFound, func(channelID string) (string, struct{}) {
+		return channelID, struct{}{}
+	})
 
 	var inProgressInfos []*importInfo
 
@@ -895,7 +890,7 @@ func (m *Manager) getNonProcessedImportInfosWithBulkStatus(ctx context.Context, 
 		info.Failed = false
 
 		if _, exists := notFoundMap[info.ChannelID]; exists {
-			recreatedStatusRes, recreateErr := m.handleChannelNotFoundPostBulkStatus(ctx, info)
+			recreatedStatusRes, recreateErr := m.handleChannelRecoveryPostBulkStatus(ctx, info, false)
 			if recreateErr != nil {
 				info.Failed = true
 				info.Reason = recreateErr.Error()
@@ -919,7 +914,7 @@ func (m *Manager) getNonProcessedImportInfosWithBulkStatus(ctx context.Context, 
 
 		statusRes := statuses[info.ChannelID]
 		if !statusRes.Valid {
-			recreatedStatusRes, recreateErr := m.handleInvalidStatusResponsePostBulkStatus(ctx, info)
+			recreatedStatusRes, recreateErr := m.handleChannelRecoveryPostBulkStatus(ctx, info, true)
 			if recreateErr != nil {
 				info.Failed = true
 				info.Reason = recreateErr.Error()
@@ -937,6 +932,12 @@ func (m *Manager) getNonProcessedImportInfosWithBulkStatus(ctx context.Context, 
 				inProgressInfos = append(inProgressInfos, info)
 				continue
 			}
+			m.polledImportInfoMap[info.ChannelID] = info
+			continue
+		}
+		if !statusRes.Success {
+			info.Failed = true
+			info.Reason = fmt.Sprintf("failed to get status with success: %t", statusRes.Success)
 			m.polledImportInfoMap[info.ChannelID] = info
 			continue
 		}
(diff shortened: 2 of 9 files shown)


router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -831,6 +845,226 @@ func isInProgress(statusRes *model.StatusResponse, info *importInfo, log logger.
...
+	}
+
+	return inProgressInfos
+}
+
+func (m *Manager) handleChannelNotFoundPostBulkStatus(ctx context.Context, info *importInfo) (*model.StatusResponse, error) {
+	log := m.logger.Withn(
+		logger.NewStringField("channelId", info.ChannelID),
+		logger.NewStringField("expectedOffset", info.Offset),
+		logger.NewStringField("table", info.Table),
```
shekhar-rudder, line 1002: `handleChannelNotFoundPostBulkStatus` and `handleInvalidStatusResponsePostBulkStatus` are ~60 lines each and nearly identical. The only difference is that the "invalid" path calls `m.deleteChannel()` before recreating. Consider consolidating into a single method with a `deleteFirst bool` parameter, or extracting the shared recreate-and-re-poll tail into a helper — would cut ~50 lines of duplication.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -831,6 +845,226 @@ func isInProgress(statusRes *model.StatusResponse, info *importInfo, log logger.
...
+			m.polledImportInfoMap[info.ChannelID] = info
+			continue
+		}
+
+		statusRes := statuses[info.ChannelID]
+		if !statusRes.Valid {
+			recreatedStatusRes, recreateErr := m.handleInvalidStatusResponsePostBulkStatus(ctx, info)
+			if recreateErr != nil {
+				info.Failed = true
+				info.Reason = recreateErr.Error()
```
shekhar-rudder: Bug: In the non-bulk path (`getImportStatus`), when `Valid == true` but `Success == false`, the channel is explicitly failed:

```go
if !statusRes.Success {
    return false, fmt.Errorf("invalid status response with valid: %t, success: %t", ...)
}
```

The bulk path only checks `!statusRes.Valid` here and falls through to `isInProgress`. A channel returning `Valid=true, Success=false` will silently skip error handling — this is a behavioral divergence from the existing non-bulk path.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -831,6 +845,226 @@ func isInProgress(statusRes *model.StatusResponse, info *importInfo, log logger.
...
+func (m *Manager) getNonProcessedImportInfosWithBulkStatus(ctx context.Context, infos []*importInfo) []*importInfo {
+	channelIDs := lo.Map(infos, func(info *importInfo, _ int) string {
+		return info.ChannelID
+	})
+
+	statuses, notFound, err := m.getBulkStatus(ctx, channelIDs)
+	if err != nil {
+		for i := range infos {
+			info := infos[i]
+			info.FailedJobIds = nil
```
shekhar-rudder: If `getBulkStatus` returns an error (e.g. transient network blip), every info in the batch is permanently marked `Failed = true` and written into `polledImportInfoMap` with no retry opportunity. Consider returning all infos as in-progress on transient errors so they get retried on the next poll cycle, rather than permanently failing the entire batch.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming_test.go
```diff
@@ -1507,6 +1516,361 @@ func TestSnowpipeStreaming(t *testing.T) {
...
+							"test-users-channel":    {Valid: false, Success: false, Offset: "0"},
+						},
+					}, nil
+				},
+			},
+			// getStatusOutputMap: map[string]func() (*model.StatusResponse, error){
+			// 	"test-products-channel": func() (*model.StatusResponse, error) {
+			// 		return &model.StatusResponse{Valid: false, Success: false, Offset: "0"}, nil
+			// 	},
+			// 	"test-users-channel": func() (*model.StatusResponse, error) {
```
shekhar-rudder: Nit: this commented-out `getStatusOutputMap` block should be removed before merge.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -831,6 +845,226 @@ func isInProgress(statusRes *model.StatusResponse, info *importInfo, log logger.
 		latestCommittedOffset, latestInsertedOffset, expectedOffset)
 }
 
+func (m *Manager) getBulkStatus(ctx context.Context, channelIDs []string) (map[string]*model.StatusResponse, []string, error) {
+	statuses := make(map[string]*model.StatusResponse)
+	notFound := make([]string, 0)
+
+	channelIDChunks := lo.Chunk(channelIDs, m.config.bulkStatusChannels.Load())
+	for i := range channelIDChunks {
+		channelIDChunk := channelIDChunks[i]
```
shekhar-rudder, line 905: Nit: this can just be `var notFound []string` — `append` handles nil slices fine.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -646,14 +650,24 @@ func (m *Manager) Poll(_ context.Context, pollInput common.AsyncPoll) common.Pol
```
shekhar-rudder, line 695: Nit: `bulkStatusChannels` reads ambiguously — could mean "channels that have bulk status" rather than a batch/chunk size. Consider `bulkStatusBatchSize` or `bulkStatusChunkSize` to clarify intent.

shekhar-rudder (commented):
A few concerns — one potential bug around a missing `Success` check, a resilience question on bulk error handling, and some cleanup items.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
achettyiitr: Good catch.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -831,6 +845,177 @@ func isInProgress(statusRes *model.StatusResponse, info *importInfo, log logger.
...
+				inProgressInfos = append(inProgressInfos, info)
+				continue
+			}
+			m.polledImportInfoMap[info.ChannelID] = info
+			continue
+		}
+
+		statusRes := statuses[info.ChannelID]
+		if !statusRes.Valid {
+			recreatedStatusRes, recreateErr := m.handleChannelRecoveryPostBulkStatus(ctx, info, true)
```
shekhar-rudder: **Bug: nil pointer dereference**

If the API omits a channel ID from both `Statuses` and `NotFound` (server bug, partial response, etc.), `statusRes` will be `nil` here and `statusRes.Valid` will panic.

```suggestion
		statusRes, ok := statuses[info.ChannelID]
		if !ok {
			info.Failed = true
			info.Reason = fmt.Sprintf("channel %s missing from bulk status response", info.ChannelID)
			m.polledImportInfoMap[info.ChannelID] = info
			continue
		}
		if !statusRes.Valid {
```

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -831,6 +845,177 @@ func isInProgress(statusRes *model.StatusResponse, info *importInfo, log logger.
...
+			info.Reason = statusErr.Error()
+			m.polledImportInfoMap[info.ChannelID] = info
+			continue
+		}
+		if inProgress {
+			inProgressInfos = append(inProgressInfos, info)
+			continue
+		}
+		m.polledImportInfoMap[info.ChannelID] = info
+	}
```
shekhar-rudder: **Nit: code duplication**

The `notFound` block (893–914) and the `!Valid` block (917–938) are copy-paste identical except for the `deleteChannel` bool passed to `handleChannelRecoveryPostBulkStatus`. The `isInProgress` → error → inProgress → store-in-map pattern then repeats a third time in the happy path (946–957).

Consider collapsing the loop body:

```go
for i := range infos {
    info := infos[i]
    info.FailedJobIds = nil
    info.Failed = false

    statusRes, needsRecovery, deleteFirst := classifyBulkResult(info, statuses, notFoundMap)
    if needsRecovery {
        var err error
        statusRes, err = m.handleChannelRecoveryPostBulkStatus(ctx, info, deleteFirst)
        if err != nil {
            info.Failed = true
            info.Reason = err.Error()
            continue
        }
    } else if !statusRes.Success {
        info.Failed = true
        info.Reason = fmt.Sprintf("failed to get status with success: %t", statusRes.Success)
        continue
    }

    inProgress, statusErr := isInProgress(statusRes, info, m.logger)
    if statusErr != nil {
        info.Failed = true
        info.Reason = statusErr.Error()
        continue
    }
    if inProgress {
        inProgressInfos = append(inProgressInfos, info)
        continue
    }
}
```

This removes the three repetitions of the `isInProgress` block, and the `classifyBulkResult` helper also gives a natural place to handle the nil `statusRes` case from the other comment.

achettyiitr: Refactored the code for duplicates.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming_test.go
```diff
@@ -1586,6 +1594,353 @@ func TestSnowpipeStreaming(t *testing.T) {
...
+		require.True(t, output.HasFailed)
+		require.JSONEq(t, `[{"channelId":"test-products-channel","offset":"1003","table":"PRODUCTS","failed":true,"reason":"invalid status response after recreation with valid: false, success: false","count":2},{"channelId":"test-users-channel","offset":"1004","table":"USERS","failed":true,"reason":"invalid status response after recreation with valid: false, success: false","count":2}]`, output.FailedJobParameters)
+		require.EqualValues(t, 4, statsStore.Get("snowpipe_streaming_jobs", stats.Tags{
+			"module":        "batch_router",
+			"workspaceId":   "test-workspace",
+			"destType":      "SNOWPIPE_STREAMING",
+			"destinationId": "test-destination",
+			"status":        "failed",
+		}).LastValue())
+	})
```
shekhar-rudder, line 1782: This test appears to be a duplicate of **"Poll status invalid, recreated channel but still status is invalid with bulk status enabled"** (above). Both tests have identical mock configurations (`getBulkStatusOutputMap`, `deleteChannelOutputMap`, `createChannelOutputMap`) and identical assertions.

If this is meant to cover a different scenario (e.g., the `getBulkStatus` call *itself* returning an error after recreation), the mock should be updated accordingly — for example, making `getBulkStatus` for the recreated channel return an `error` instead of a response with `Valid: false, Success: false`.

Otherwise, one of the two should be removed.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -881,6 +895,163 @@ func isInProgress(statusRes *model.StatusResponse, info *importInfo, log logger.
...
+	var notFound []string
+
+	channelIDChunks := lo.Chunk(channelIDs, m.config.bulkStatusBatchSize.Load())
+	for i := range channelIDChunks {
+		channelIDChunk := channelIDChunks[i]
+		bulkStatusRes, err := m.api.GetBulkStatus(ctx, channelIDChunk)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to get bulk status: %w", err)
+		}
+		if !bulkStatusRes.Success {
```
shekhar-rudder, line 910: Is snowpipe service setting a top level `Success` field? Is this the behaviour across all APIs?

achettyiitr: Yes. Response overrides SuccessResponse. https://github.com/rudderlabs/rudder-snowpipe-clients/blob/main/src/main/java/com/rudderstack/snowpipe/api/dto/BulkChannelStatusResponse.java#L9

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming_test.go
achettyiitr: Fixed it.

shekhar-rudder (approved):

router/batchrouter/asyncdestinationmanager/snowpipestreaming/internal/model/model.go
```diff
@@ -76,6 +76,12 @@ type (
 		Valid                bool   `json:"valid"`
 		LatestInsertedOffset string `json:"latestInsertedOffset"`
 	}
+
+	BulkStatusResponse struct {
+		Success  bool                       `json:"success"`
+		Statuses map[string]*StatusResponse `json:"statuses"`
```
krishna2020, line 82: how does each entry of StatusResponse gets related with channelID ?

achettyiitr: It's a [map](https://github.com/rudderlabs/rudder-snowpipe-clients/blob/32a8258dc7cb2bcb9ca746b0326139a4bbab4477/src/main/java/com/rudderstack/snowpipe/api/ChannelsResource.java#L148-L187) containing channelID vs Status Response.

krishna2020 (approved):

PR #6930, opened 2026-04-30
fix(router): delays in finding worker slots when event ordering is disabled

atzoum:

# Description

## Observed problem

When event ordering is disabled, `Handle.findWorkerSlot` becomes unexpectedly slow at high worker counts. With 512 workers per partition and the experimental buffer-size calculator enabled, each call to find a worker slot took ~47 µs locally — meaning a single pickup batch of 6 000 events spent ~280 ms of CPU just resolving which worker to assign each job to, before any actual processing happened. Latency is even worse on production machines.

## Root causes

The slow path is the event-ordering-disabled branch in `findWorkerSlot`:

```go
availableWorkers := lo.Filter(workers, func(w *worker, _ int) bool { return w.AvailableSlots() > 0 })
if len(availableWorkers) == 0 { return nil, types.ErrWorkerNoSlot }
slot := availableWorkers[rand.Intn(len(availableWorkers))].ReserveSlot()
```

Two compounding problems:

1. **`AvailableSlots()` recomputes the buffer's target capacity on every call.** Inside `workerBuffer.currentCapacity()`, the calculator closure runs (for the experimental calculator: a `SimpleMovingAverage.Load`, multiple `ValueLoader.Load`s, and `math.Max`/`math.Ceil`), plus the stats `OnceEvery.Do` check fires — and this happens **513 times per call** (once per worker for the filter, then once more on the chosen worker's `ReserveSlot`). With stats wired up, this path costs ~47 µs/call.

2. **Filter-then-pick is structurally redundant and racy.** `lo.Filter` allocates a ~5 KB survivor slice, takes 512 RLocks just to ask "do you have a slot?", then we take a write Lock on the chosen worker to actually reserve. The first 512 reads are throwaway work, and there's a (currently benign, single-threaded) race window between the filter and the reservation.

## What we're doing

**1. Cache the buffer capacity, refresh it from a 1 s sampler.**
- `workerBuffer` gains a `cachedCapacity atomic.Int64` field (initialized in the constructor).
- `availableSlots()` reads the cached value with an atomic load instead of invoking the calculator.
- `partitionWorker` runs a single sampler goroutine that ticks once per second and calls `refreshCapacity()` on every worker buffer it owns. The sampler is cancelled cleanly from `Stop()`.
- Since the calculator inputs (config values, throughput SMA) move slowly, a 1 s lag is harmless — production already throttles the corresponding stats observation to every 5 s via `OnceEvery`.

**2. Replace filter+pick with shuffle-and-try.**
- New helper `reserveAnyWorkerSlot(workers)` picks a random starting offset and walks the workers slice with wrap-around, calling `ReserveSlot` until one succeeds.
- `ReserveSlot` is the atomic decision point — no pre-filter step, no race window, no allocation for the survivor slice.
- The random start gives uniform worker selection (in expectation) just like the old `rand.Intn` over the filter result.

**3. New observability stat for slot-resolution latency.**
- `router_pickup_find_worker_slot_time` (TimerType, tags: `destType`, `partition`) — total wall-clock time spent inside `findWorkerSlot` across a single pickup loop. Published once per pickup. Lets us catch regressions in the future, alert on pathological saturation (when buffers run dry and `findWorkerSlot` keeps failing fast across all workers), and confirm the impact of this change in production.

## Benchmark results

Added `BenchmarkFindWorkerSlot` in `router/find_worker_slot_bench_test.go`. It exercises 512 workers across all three production buffer modes (`simple`, `dynamic/standard`, `dynamic/experimental`) and five fill scenarios. Numbers below are from an Apple M1 Pro at 2 s/case for the worst production mode (`dynamic/experimental` — switcher with `enableExperimental=true`, full `workerBufferStats` wired up).

| scenario | before | + cached capacity | + shuffleTry (this PR) |
|---|---|---|---|
| typical (slots widely available) | ~47 µs/op, 4 872 B/op | ~6.7 µs/op, 4 872 B/op | **~45 ns/op, 8 B/op** |
| 50% of workers full | ~47 µs/op | ~6.7 µs/op | **~1.1 µs/op** |
| 99% of workers full | ~47 µs/op | ~6.7 µs/op | **~4.3 µs/op** |
| 99.8% of workers full | ~47 µs/op | ~6.4 µs/op | **~4.4 µs/op** |

Even in the catastrophic 99.8%-full case, the new path wins — and allocations drop from ~5 KB/call to 8 B/call (boxed `int` from `rand.Intn`).

Extrapolated to a 6 000-event pickup batch, in the typical case:
- Before: ~280 ms of CPU just resolving worker slots.
- After: ~0.27 ms — roughly **1 000× lower CPU cost** on the hot path.

## Linear Ticket

resolves PIPE-2949

## Security

### DIFF
--- a/router/handle.go
+++ b/router/handle.go
@@ -238,6 +238,14 @@ func (rt *Handle) pickup(ctx context.Context, partition string, workers []*worke
 	var reservedJobs []reservedJob
 	blockedOrderKeys := make(map[eventorder.BarrierKey]struct{})
 
+	// Total wall-clock time spent in findWorkerSlot across this pickup loop.
+	// Published once at the end so dashboards can spot pathological slot-resolution latency
+	// (e.g. when buffers saturate or the calculator gets expensive).
+	var findWorkerSlotDuration time.Duration
+	defer func() {
+		stats.Default.NewTaggedStat("router_pickup_find_worker_slot_time", stats.TimerType, stats.Tags{"destType": rt.destType, "partition": partition}).SendTiming(findWorkerSlotDuration)
+	}()
+
 	flushTime := time.Now()
 	shouldFlush := func() bool {
 		return len(statusList) > 0 && time.Since(flushTime) > rt.reloadableConfig.pickupFlushInterval.Load()
@@ -287,7 +295,9 @@ func (rt *Handle) pickup(ctx context.Context, partition string, workers []*worke
 			rt.logger.Errorn("Error occurred while unmarshalling job parameters. Panicking", obskit.Error(err))
 			panic(err)
 		}
+		findStart := time.Now()
 		workerJobSlot, err := rt.findWorkerSlot(ctx, workers, job, parameters, blockedOrderKeys)
+		findWorkerSlotDuration += time.Since(findStart)
 		if err == nil {
 			traceParent := jsonparser.GetStringOrEmpty(job.Parameters, "traceparent")
 			if traceParent != "" {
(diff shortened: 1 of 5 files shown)


router/handle.go
```diff
@@ -626,6 +636,24 @@ type workerJobSlot struct {
...
+// no worker has capacity. ReserveSlot is the atomic decision point — there is no
+// pre-filter step (which would be redundant work and would race with the reservation).
+func reserveAnyWorkerSlot(workers []*worker) *reservedSlot {
+	n := len(workers)
+	if n == 0 {
+		return nil
+	}
+	start := rand.Intn(n)
+	for i := range n {
+		if slot := workers[(start+i)%n].ReserveSlot(); slot != nil {
```
fracasula, line 650: Let's say we have a workers slice, say of 10 workers.
0 is free, all others are taken.
Then you get 1 as the index and iterate over all of them, then return `nil`.
Shouldn't you do a second loop in that case and go from `0` to `n`, which in the above case would return the worker at `0`?

fracasula (approved):
LGTM, disregard the comment 👍

router/handle.go
fracasula: I read this wrong, nevermind. I didn't notice the `%n` and I was under the impression that you started from `n` and kept going ahead until the end of the slice 👍

ktgowtham (approved):

PR #6945, opened 2026-05-09
feat: periodic cleanup of rsources shared db tables

atzoum:

# Description

Up until now we were only removing stale rows from each node's **local** rsources db. But orphans can accumulate in the **shared** db when a node leaves the cluster — logical replication only propagates deletes from nodes that are still subscribed, so anything written by a now-departed node is left behind forever.

This PR adds a periodic shared-db cleanup that runs alongside the existing local cleanup loop:

- New `doCleanupSharedTables` reuses the same delete logic (extracted into a `cleanupTablesBefore` helper) but executes against `sharedDB` instead of `localDB`.
- With many nodes pointed at the same shared db, only one should be deleting at a time. The cleanup transaction first calls `pg_try_advisory_xact_lock(sharedCleanupLockID)` and **skips the run** (logs and returns) if another node is already holding the lock. Tx-scoped, so the lock is released automatically on commit/rollback — no leak risk if a node crashes mid-cleanup.
- Retention for the shared cleanup is `Rsources.retention + Rsources.shared.retentionBuffer` (defaults: 72h + 24h = 96h). The buffer guarantees that even with replication lag or a briefly-offline node, the shared row is never GC'd before the corresponding local-side delete has had time to propagate.
- `CleanupLoop` now runs the local and shared loops in parallel via `errgroup.WithContext`. Errors from the local loop still tear down the group (existing behavior); errors from the shared loop are logged and the loop continues — best-effort across nodes.

### New config keys

| key | default | meaning |
|---|---|---|
| `Rsources.shared.cleanup.interval` | `1h` | how often each node attempts a shared-db cleanup |
| `Rsources.shared.retentionBuffer` | `24h` | safety margin on top of `Rsources.retention` before shared rows are eligible for deletion |

## Linear Ticket

resolves PIPE-2963

## Security

### DIFF
--- a/services/rsources/handler.go
+++ b/services/rsources/handler.go
@@ -13,13 +13,18 @@ import (
 	"github.com/lib/pq"
 	"github.com/lib/pq/pqerror"
 	"github.com/segmentio/ksuid"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
 )
 
-const defaultRetentionPeriodInHours = 3 * 24
+const (
+	defaultRetentionPeriodInHours             = 3 * 24
+	defaultSharedRetentionBufferPeriodInHours = 24
+	sharedCleanupLockID                       = 100020002
+)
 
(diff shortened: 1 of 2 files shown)


ktgowtham (approved):

mihir20 (approved):
LGTM

PR #6949, opened 2026-05-11
chore(ci): workflow cleanup (SEC-284)

aris1009:
Disables implicit Go module cache restore on setup-go in this workflow. No behavioral change to tests; cache restore is removed from a privileged job.

Linear: SEC-284


lokey (approved):
Approved: SEC-284 workflow-only changes

PR #6958, opened 2026-05-14
fix(batchrouter): preserve prior LastJobStatus error on drain

achettyiitr:
# Description

When a job is drained in the batch router's `scheduleJobs` (job expired, source not found, destination disabled, etc.), the `ErrorResponse` written to `job_status` was synthesized from an empty `{}` with just `{"reason": "<drain reason>"}`, e.g.

```json
{ "reason": "job expired" }
```

Any upstream error from the previous delivery attempt that lived on `job.LastJobStatus.ErrorResponse` — fields like `error`, `response`, `firstAttemptedAt` — was clobbered, so by the time a job reached status `Aborted` with code `410` we'd lost all the context about *why* it had been failing.

### Change

In `router/batchrouter/worker.go`, base the drained status's `ErrorResponse` on `job.LastJobStatus.ErrorResponse` instead of `[]byte("{}")`. `routerutils.EnhanceJSON` then merges the new `reason` field into the existing object rather than replacing it.

```diff
-                ErrorResponse: routerutils.EnhanceJSON([]byte(`{}`), "reason", finalReason),
+                errResponse := routerutils.EnhanceJSON(job.LastJobStatus.ErrorResponse, "reason", finalReason)
+                ErrorResponse: errResponse,
```

After the fix, a drained job's `ErrorResponse` carries the prior upstream context alongside the drain reason:

```json
{
  "error": "500 internal server error",
  "response": "upstream timed out",
  "firstAttemptedAt": "2024-01-01T00:00:00.000Z",
  "reason": "job expired"
}
```

`ErrorCode` stays `410` and `JobState` stays `Aborted` — no change to drain semantics, only what we surface to debuggers and downstream reporting.

### Tests

New file `router/batchrouter/worker_test.go` with `TestScheduleJobsDrainBranch` covering three cases against a real `scheduleJobs` call (mocked `jobsDB`, fake drainer):

1. **`preserves prior LastJobStatus error fields and stamps the drain reason`** — the bug-fix path. A job with a prior `{"error":"…","response":"…","firstAttemptedAt":"…"}` is drained; asserts all three fields survive and `reason` is stamped on top.
2. **`synthesizes reason when LastJobStatus has no prior error response`** — first-attempt jobs (empty `LastJobStatus.ErrorResponse`) still end up with `{"reason":"job expired"}`. No regression on the existing behavior.
3. **`uses source_not_found reason when the job's source is missing from the destination`** — the second branch of the same drain condition (`!sourceFound`) also benefits from the fix; the prior error is preserved alongside `reason: "source_not_found"`.

## Linear Ticket

- Resolves INT-6358, INT-6356, INT-6349.

## Security

### DIFF
--- a/router/batchrouter/batchrouter_test.go
+++ b/router/batchrouter/batchrouter_test.go
@@ -399,8 +399,8 @@ var _ = Describe("BatchRouter", func() {
 			}).Return(nil)
 			c.mockBatchRouterJobsDB.EXPECT().UpdateJobStatusInTx(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).
 				Do(func(ctx context.Context, _ any, statuses []*jobsdb.JobStatusT) {
-					assertJobStatus(toRetryJobsList[0], statuses[0], jobsdb.Aborted.State, "{\"reason\":\"source_not_found\"}", 130)
-					assertJobStatus(toRetryJobsList[1], statuses[1], jobsdb.Aborted.State, "{\"reason\":\"source_not_found\"}", 4)
+					assertJobStatus(toRetryJobsList[0], statuses[0], jobsdb.Aborted.State, fmt.Sprintf(`{"firstAttemptedAt": %q, "reason":"source_not_found"}`, attempt1.Format(misc.RFC3339Milli)), 130)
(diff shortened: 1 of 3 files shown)


krishna2020 (approved):

ktgowtham (approved):
Tests are failing.

PR #6962, opened 2026-05-15
feat(jobsdb): versioned dataset list and non-blocking drop for completed datasets

atzoum:

# Description

Introducing support for dropping completed dataset tables in a non-blocking fashion without a migration lock:

### 1. Versioned dataset list with reader-drain semantics

A new `versionedDSList` replaces the bare `datasetList` / `datasetRangeList` fields on `Handle`. Each update bumps a version; readers acquire a snapshot tagged with that version and release it when done reading. An additional `wait(through)` API returns a channel that closes once no reader is still using a version `<= through`, so a dropper can block until in-flight queries that still see the old list have finished.

### 2. `nonBlockingCompletedDSDrop` fast path and async drop loop

- A new `dropDSLoop` consumes a `dropDSList` queue. Each entry records the dslist version that must drain before the table can be dropped.
- `addToDropDSList` republishes the dslist without the queued datasets, writes the persistent pre-drop marker (see §3), and pings the loop. The actual `DROP TABLE` happens off the hot path.
- `doMigrateDS` gains a fast path gated by **`jobsdb.<prefix>.nonBlockingCompletedDSDrop`** (default `false`): datasets with zero pending jobs are routed through the async drop loop instead of the in-lock `postMigrateHandleDS` path. When disabled, behavior is unchanged.

### 3. Persistent `pre_drop:v1` table marker

`COMMENT ON TABLE … IS 'rudder:pre_drop:v1'` is written as part of queuing a dataset for drop. This makes the queue crash-safe — the source of truth lives in Postgres, not just in-memory `dropDSList`.

- `getAllTableNames` now joins `pg_class`/`pg_namespace`/`pg_description` and filters out marked tables, masking half-pairs (drops both the jobs and status table when either is marked).
- New SQL migration `000019_jobsdb_functions_pre_drop_filter.up.sql` rewrites `unionjobsdb`, `unionjobsdbmetadata`, and `joborderlog` to apply the same `pg_description` filter, so server-side discovery skips marked tables too.
- `cleanupPreDropTables` runs once at startup of `read` or `readwrite` jobsdbs and after journal recovery, dropping any tables left marked by a previous crashed run.

If there isn't any marker present, behavior remains identical as before.

### 4. `RS001` tolerance on the `UpdateJobStatus` path

The readonly-trigger error code `RS001` was previously only handled on the `Store` path. The `Update` path now wraps each `internalUpdateJobStatusInTx` call in a savepoint: on `ErrStaleDsList` it rolls back to the savepoint, refreshes the dslist under the write lock, updates the `UpdateSafeTx`'s embedded dslist/rangeList snapshot, and retries the whole command. We do **not** replay the caller-owned outer transaction.

To support this, `UpdateSafeTx` now carries its own `dsList`/`dsRangeList` (with a `setDSList` setter) so retries reroute against the refreshed snapshot without re-running `inUpdateSafeCtx`.

This change is harmless until a follow-up PR starts marking status tables readonly during compaction.

## Linear Ticket

resolves PIPE-2996

## Security

---
This is **part 2 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #6979 
- <kbd>&nbsp;3&nbsp;</kbd> #6967 
- <kbd>&nbsp;2&nbsp;</kbd> #6962 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #6963

### DIFF
--- a/jobsdb/admin.go
+++ b/jobsdb/admin.go
@@ -6,6 +6,8 @@ import (
 	"time"
 
 	"github.com/rudderlabs/rudder-go-kit/logger"
+
+	"github.com/rudderlabs/rudder-server/jobsdb/internal/lock"
 )
 
 /*
@@ -43,7 +45,7 @@ func (jd *Handle) deleteJobStatus() {
 				CustomValFilters: []string{jd.tablePrefix},
 			}).RecordDuration()()
 
-		dsList := jd.getDSList()
+		dsList := tx.getDSList()
 
 		for _, ds := range dsList {
 			if err := jd.deleteJobStatusDSInTx(tx.SqlTx(), ds); err != nil {
@@ -102,7 +104,7 @@ func (jd *Handle) failExecuting() {
 			&statTags{CustomValFilters: []string{jd.tablePrefix}},
 		).RecordDuration()()
 
-		dsList := jd.getDSList()
+		dsList := tx.getDSList()
 
 		for _, ds := range dsList {
 			err := jd.failExecutingDSInTx(tx.SqlTx(), ds)
@@ -136,8 +138,13 @@ func (jd *Handle) failExecutingDSInTx(txHandler transactionHandler, ds dataSetT)
 	return err
 }
 
-func (jd *Handle) doCleanup(ctx context.Context) error {
-	if err := jd.abortOldJobs(ctx, jd.getDSList()); err != nil {
+// doCleanup performs cleanup of old jobs and journal entries. It requires a ds list lock token to ensure that the list does not change during the cleanup process.
+func (jd *Handle) doCleanup(ctx context.Context, l lock.LockToken) error {
+	if l == nil {
+		return fmt.Errorf("lock token is required for cleanup")
+	}
+	dsList, _ := jd.dsList.snapshot()
+	if err := jd.abortOldJobs(ctx, dsList); err != nil {
 		return fmt.Errorf("aborting old jobs: %w", err)
 	}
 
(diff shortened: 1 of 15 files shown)


jobsdb/admin.go
```diff
@@ -136,8 +138,13 @@ func (jd *Handle) failExecutingDSInTx(txHandler transactionHandler, ds dataSetT)
 	return err
 }
 
-func (jd *Handle) doCleanup(ctx context.Context) error {
-	if err := jd.abortOldJobs(ctx, jd.getDSList()); err != nil {
+// doCleanup performs cleanup of old jobs and journal entries. It requires a ds list lock token to ensure that the list does not change during the cleanup process.
+func (jd *Handle) doCleanup(ctx context.Context, l lock.LockToken) error {
```
atzoum, line 142: **Note:** modifying this method's signature to highlight that it can only be called with a ds list lock already acquired

jobsdb/versioned_ds_list.go
```diff
@@ -0,0 +1,123 @@
...
+				delete(v.readers, version)
+			} else {
+				v.readers[version]--
+			}
+			v.closeDrainedLocked()
+		})
+	}
+}
+
+func (v *versionedDSList) snapshot() (dataSetTList, dataSetRangeTList) {
```
ktgowtham, line 67: don't we need to increment a reader for this operation like we do in `get()`?

ktgowtham (approved):
LGTM, but added a minor comment. not sure if it has any risk.

jobsdb/versioned_ds_list.go
atzoum: No since we only read snapshots in jobsdb after acquiring a dsList lock

mihir20 (approved):

PR #6964, opened 2026-05-16
chore(jobsdb): disable no results cache state filter optimization

atzoum:

# Description

Makes the per-state `noResultsCache` optimization in `jobsdb` configurable, and **disables it by default**.

The optimization has two parts in `getJobsDS`:
1. **Read-side pruning** — `stateFilters` are narrowed against the cache before querying, skipping states the cache reported as empty for the current `(ds, partitions, workspace, customVals, params)` key.
2. **Write-side commit predicate** — when a queried state is absent from the result set and no limit was reached, a "no jobs of state X" entry is committed to the cache.

Both are now gated behind a single reloadable flag `noResultsCacheStateOptimization`:
- `JobsDB.<tablePrefix>.noResultsCacheStateOptimization`
- `JobsDB.noResultsCacheStateOptimization`

Default is `false` (optimization disabled).

Also fixes a variable-shadowing bug in `getJobsDS` where the outer `payloadSize` was never updated due to an inner shadow, making the post-loop `payloadSize >= PayloadSizeLimit` check dead code. As a result, `LimitsReached` was wrongly reported as `false` when the payload limit was reached at the boundary (exact total) or when a single oversize job exhausted the budget alone. This is one of the concrete mechanisms that can poison the `noResultsCache` — see the reasoning below.

## Reasoning

We suspect the per-state optimization is the root cause of `noResultsCache` being populated with stale entries, which in turn leads to **out-of-order processing**: a state that actually has jobs is wrongly cached as empty, so subsequent reads skip those jobs from one dataset and process later jobs from another dataset first.

The write-side commit predicate `(!ok && !limitsReached)` assumes that when a state is missing from the result set and limits were not reached, we have exhaustively scanned the dataset for that state. This invariant can be broken by silent row-dropping filters that are not part of the cache key, or incorrect calculation of when limits are reached, and the resulting cache entry then poisons later reads.

The `payloadSize` shadowing bug fixed in this PR is a concrete instance of the second class: at the payload-budget boundary `limitsReached` was wrongly returned as `false`, so a state missing from the result set (e.g. because the budget was exhausted before any state-X row appeared) would satisfy `(!ok && !limitsReached)` and get committed to the cache as "no jobs for this state" — a stale entry that subsequent reads would honour, skipping real jobs and triggering out-of-order processing.

Disabling the per-state optimization by default keeps only the coarser `len(jobList) == 0` commit path, which can only record "no rows at all" results and is therefore not vulnerable to per-state misattribution.

The optimization can be re-enabled per-deployment via the reloadable flag while we investigate and harden the invariants.

## Linear Ticket

resolves PIPE-2994

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -2407,7 +2407,7 @@ func (jd *Handle) getJobsDS(ctx context.Context, ds dataSetT, lastDS bool, param
 	var payloadSize int64
 	resultsetStates := map[string]struct{}{}
 	for rows.Next() {
-		var payloadSize int64
+		var rowPayloadSize int64
 		var job JobT
 		var payload []byte
 		var jsState sql.NullString
@@ -2418,7 +2418,7 @@ func (jd *Handle) getJobsDS(ctx context.Context, ds dataSetT, lastDS bool, param
 		var jsErrorResponse []byte
 		var jsParameters []byte
 		err := rows.Scan(&job.JobID, &job.UUID, &job.UserID, &job.Parameters, &job.CustomVal,
-			&payload, &job.EventCount, &job.CreatedAt, &job.ExpireAt, &job.WorkspaceId, &job.PartitionID, &payloadSize, &runningEventCount, &runningPayloadSize,
+			&payload, &job.EventCount, &job.CreatedAt, &job.ExpireAt, &job.WorkspaceId, &job.PartitionID, &rowPayloadSize, &runningEventCount, &runningPayloadSize,
 			&jsState, &jsAttemptNum,
 			&jsExecTime, &jsRetryTime,
 			&jsErrorCode, &jsErrorResponse, &jsParameters)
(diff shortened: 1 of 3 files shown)


ktgowtham (approved):

jobsdb/jobsdb.go
```diff
@@ -2359,7 +2367,7 @@ func (jd *Handle) getJobsDS(ctx context.Context, ds dataSetT, lastDS bool, param
 	var payloadSize int64
 	resultsetStates := map[string]struct{}{}
 	for rows.Next() {
-		var payloadSize int64
+		var rowPayloadSize int64
```
mihir20: rowPayloadSize is not used, we do not need to read this

atzoum: We need to scan as it is part of the resultset. But since we don't use it we can declare it before the `for` loop. 👍

mihir20 (approved):

PR #6966, opened 2026-05-18
fix: snowpipe streaming bulk status offset

achettyiitr:

# Description

- Fixes offset-based duplicate detection for Snowpipe Streaming when `SnowpipeStreaming.bulkStatusEnabled `is on. Duplicate checks now use the bulk status API (with a single pre-fetch per upload) instead of per-channel GetStatus calls.

## Linear Ticket

- Resolves INT-6378

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
@@ -242,6 +242,26 @@ func (m *Manager) Upload(_ context.Context, asyncDest *common.AsyncDestinationSt
 		abortedJobIDs []int64
 		abortReason   string
 	)
+
+	// Get channel IDs from cache to avoid duplicate requests.
+	channelCacheIDs := lo.FilterMap(uploadInfos, func(info *uploadInfo, _ int) (string, bool) {
+		response, ok := m.channelCache.Load(info.tableName)
+		if !ok {
+			return "", false
+		}
+		return response.(*model.ChannelResponse).ChannelID, true
+	})
+
+	// Pre-fetch channel statuses for all channels to avoid duplicate requests.
+	var channelStatuses map[string]*model.StatusResponse
+	if m.config.bulkStatusEnabled.Load() {
+		channelStatuses, _, err = m.getBulkStatus(ctx, channelCacheIDs)
+		if err != nil {
+			m.logger.Warnn("Failed to get bulk status", obskit.Error(err))
+			return m.abortJobs(asyncDest, fmt.Errorf("failed to get bulk status: %w", err).Error())
+		}
+	}
+
 	for _, info := range uploadInfos {
 		fields := []logger.Field{
 			logger.NewStringField("table", info.tableName),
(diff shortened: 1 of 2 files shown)


router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -242,6 +242,26 @@ func (m *Manager) Upload(_ context.Context, asyncDest *common.AsyncDestinationSt
...
+	var channelStatuses map[string]*model.StatusResponse
+	if m.config.bulkStatusEnabled.Load() {
+		channelStatuses, _, err = m.getBulkStatus(ctx, channelCacheIDs)
+		if err != nil {
+			m.logger.Warnn("Failed to get bulk status", obskit.Error(err))
+			return m.abortJobs(asyncDest, fmt.Errorf("failed to get bulk status: %w", err).Error())
+		}
+	}
+
 	for _, info := range uploadInfos {
```
shekhar-rudder, line 265: Are we sure we want to abort all jobs if the bulk status API fails? This could be a transient error. For comparison, `getNonProcessedImportInfosWithBulkStatus` does **not** abort on bulk status failure — it logs the error and returns all import infos to be polled again.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -242,6 +242,26 @@ func (m *Manager) Upload(_ context.Context, asyncDest *common.AsyncDestinationSt
 		abortedJobIDs []int64
 		abortReason   string
 	)
+
+	// Get channel IDs from cache to avoid duplicate requests.
+	channelCacheIDs := lo.FilterMap(uploadInfos, func(info *uploadInfo, _ int) (string, bool) {
+		response, ok := m.channelCache.Load(info.tableName)
+		if !ok {
+			return "", false
+		}
```
shekhar-rudder: `channelCacheIDs` is only used when `bulkStatusEnabled` is true. This can be moved inside the `if m.config.bulkStatusEnabled.Load()` block below to avoid unnecessary computation when the feature is off.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
```diff
@@ -498,8 +519,39 @@ func checkForDuplicateIDsInBatch(events []*event) (duplicateCount int) {
...
+	channelID string,
+	events []*event,
+	channelStatuses map[string]*model.StatusResponse,
+) (duplicateCount int) {
+	var statusRes *model.StatusResponse
+	var err error
+
+	if m.config.bulkStatusEnabled.Load() {
+		var ok bool
+		statusRes, ok = channelStatuses[channelID]
```
shekhar-rudder, line 534: In the `bulkStatusEnabled` branch, all errors are handled internally using `:=` (which shadows this `err`). The `if err != nil` check after the `if/else` block only ever triggers for the `else` branch (`GetStatus`). In the bulk path, `err` will always be `nil` there, making it dead code.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
achettyiitr: Yeah, not needed. Made it by mistake. Marking it failed.

router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
achettyiitr: In case of err for the bulkStatusEnabled branch, we are returning. So the post if `err != nil `is not being used at all. Refactored the code for more clarity.

shekhar-rudder (approved):

krishna2020 (approved):

PR #6978, opened 2026-05-19
feat: per-workspace PyT routing + cold-start handling

mihir20:
# Description

## Summary

Adds support for routing the Python branch of user-transformations to a workspace-scoped PyT service (`pyt-<workspaceID>:9090`) instead of the static `PYTHON_TRANSFORM_URL`, plus the cold-start
observability and retry behavior the HPA scale-from-zero flow needs. JS, mirror, and existing Python paths are identical when the new flag is off.

## Behavior 

- **URL resolution.** New flag `Processor.UserTransformer.perWorkspacePyTEnabled` (default `false`). When on, non-mirror Python transformations route to a per-workspace URL templated from
  `Processor.UserTransformer.perWorkspacePyTURLTemplate` (default `http://pyt-{workspaceID}:9090`). Falls back to the global Python URL when the version-allowlist gates the version out, when mirroring,
  or when `workspaceID` is empty.
- **Cold-start counter.** New `processor_user_transformer_cold_start_errors_total{workspaceID, language}` increments per HTTP attempt classified as cold-start (ECONNREFUSED, DNS not-found, 502, 503)
  only on the per-workspace Python non-mirror path. Tag `language` is hard-coded to `"python"` to keep cardinality bounded. HPA reads `rate(...[1m])` against `activationThreshold: 0.01` to decide 0→1
  scaling.
- **Cold-start retry.** On cold-start detection the inner closure short-circuits with a sentinel `ColdStartError`. The `doPost` tail converts that sentinel to a dedicated status code
  `StatusColdStartWindowFailure = 819`. `sendBatch`'s outer retry treats `819` like `StatusCPDown` — keeps retrying the whole call so the request eventually lands once HPA scales the pod up. Gated by
  `Processor.UserTransformer.perWorkspacePyTEndlessRetries` (default `true`); when disabled, `backoff.Permanent` bails immediately so the batch surfaces a single FailedEvent with status 819 instead of
  panicking the processor.

## Behavior when the flag is off

Identical to current `master`. New config fields default off / endless-retries-on, but the gating predicate means none
of the new code paths are reachable until we explicitly enable the flag.

## Linear Ticket

pipe-3009, pipe-3010

## Security

### DIFF
--- a/processor/internal/transformer/user_transformer/user_transformer.go
+++ b/processor/internal/transformer/user_transformer/user_transformer.go
@@ -48,6 +48,8 @@ func New(conf *config.Config, log logger.Logger, stat stats.Stats, opts ...Opt) 
 	handle.client = transformerclient.NewClient("UserTransformer", transformerutils.TransformerClientConfig(conf, "UserTransformer"))
 	handle.config.userTransformationURL = handle.conf.GetStringVar(handle.conf.GetStringVar("http://localhost:9090", "DEST_TRANSFORM_URL"), "USER_TRANSFORM_URL")
 	handle.config.pythonTransformationURL = handle.conf.GetStringVar("", "PYTHON_TRANSFORM_URL")
+	handle.config.perWorkspacePyTEnabled = handle.conf.GetBoolVar(false, "Processor.UserTransformer.perWorkspacePyTEnabled")
+	handle.config.perWorkspacePyTURLTemplate = handle.conf.GetStringVar("http://pyt-{workspaceID}:9090", "Processor.UserTransformer.perWorkspacePyTURLTemplate")
 	handle.config.pythonTransformConfig = transformerutils.LoadPythonTransformConfig(conf)
 	handle.config.timeoutDuration = conf.GetDurationVar(600, time.Second, "HttpClient.procTransformer.timeout")
 	handle.config.failOnUserTransformTimeout = conf.GetReloadableBoolVar(false, "Processor.UserTransformer.failOnUserTransformTimeout", "Processor.Transformer.failOnUserTransformTimeout")
@@ -84,6 +86,8 @@ type Client struct {
 		timeoutDuration            time.Duration
 		collectInstanceLevelStats  bool
 		batchSize                  config.ValueLoader[int]
+		perWorkspacePyTEnabled     bool
+		perWorkspacePyTURLTemplate string
 	}
 	conf   *config.Config
 	log    logger.Logger
@@ -97,7 +101,11 @@ func (u *Client) Transform(ctx context.Context, clientEvents []types.Transformer
 	}
 	batchSize := u.config.batchSize.Load()
 	transformationLanguage, transformationVersionID, transformationID := transformerutils.GetTransformationInfo(clientEvents)
-	userURL := u.userTransformURL(transformationLanguage, transformationVersionID)
+	workspaceID := ""
+	if len(clientEvents) > 0 {
+		workspaceID = clientEvents[0].Metadata.WorkspaceID
+	}
+	userURL := u.userTransformURL(transformationLanguage, transformationVersionID, workspaceID)
 
 	labels := types.TransformerMetricLabels{
 		Endpoint:         transformerutils.GetEndpointFromURL(userURL),
@@ -409,10 +417,16 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels 
 	return respData, resp.StatusCode, nil
 }
 
-func (u *Client) userTransformURL(language, versionID string) string {
+func (u *Client) userTransformURL(language, versionID, workspaceID string) string {
 	isPython := strings.HasPrefix(language, "python")
-	if isPython && u.config.pythonTransformConfig.IsVersionAllowed(versionID) && u.config.pythonTransformationURL != "" {
-		return u.config.pythonTransformationURL + "/customTransform"
+	if isPython && u.config.pythonTransformConfig.IsVersionAllowed(versionID) {
+		if u.config.perWorkspacePyTEnabled && !u.config.forMirroring && workspaceID != "" {
+			base := strings.ReplaceAll(u.config.perWorkspacePyTURLTemplate, "{workspaceID}", workspaceID)
+			return base + "/customTransform"
+		}
+		if u.config.pythonTransformationURL != "" {
+			return u.config.pythonTransformationURL + "/customTransform"
+		}
 	}
 	return u.config.userTransformationURL + "/customTransform"
 }
--- a/processor/internal/transformer/user_transformer/user_transformer.go
+++ b/processor/internal/transformer/user_transformer/user_transformer.go
@@ -362,11 +362,7 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels 
 			u.stat.NewTaggedStat("transformer_client_total_durations_seconds", stats.CountType, tags).Count(int(duration.Seconds()))
 
 			// This metric is to track cold start errors for PyT, which are expected to be higher than usual due to the nature of PyT scaling.
-			if u.config.perWorkspacePyTEnabled &&
-				!u.config.forMirroring &&
-				strings.HasPrefix(labels.Language, "python") &&
-				labels.WorkspaceID != "" &&
-				isColdStartError(reqErr, resp) {
+			if u.shouldThrowPythonColdStartErr(labels, reqErr, resp) {
 				u.stat.NewTaggedStat(
 					"processor_user_transformer_cold_start_errors_total",
 					stats.CountType,
@@ -425,17 +421,17 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels 
 	if err != nil {
 		u.log.Errorn("User transformation post error",
 			append(labels.ToLoggerFields(), obskit.Error(err))...)
-		if u.config.failOnUserTransformTimeout.Load() && os.IsTimeout(err) {
-			return fmt.Appendf(nil, "transformer request timed out: %s", err), transformerutils.TransformerRequestTimeout, nil
-		} else if u.config.failOnError.Load() {
-			return fmt.Appendf(nil, "transformer request failed: %s", err), transformerutils.TransformerRequestFailure, nil
-		}
 		// Per-workspace PyT: a persistent transport failure on the workspace-
 		// scoped URL is most likely a cold-start window (pod not yet scaled by
 		// HPA, EndpointSlice lag, kube-proxy 5xx). Surface a dedicated status
 		// code so sendBatch retries the whole call until the pod is up instead of treating it as a failed transformation.
 		if errors.Is(err, transformerutils.ErrColdStart) {
 			return fmt.Appendf(nil, "workspace transformer not reachable: %s", err), transformerutils.StatusColdStartWindowFailure, nil
+		}
+		if u.config.failOnUserTransformTimeout.Load() && os.IsTimeout(err) {
+			return fmt.Appendf(nil, "transformer request timed out: %s", err), transformerutils.TransformerRequestTimeout, nil
+		} else if u.config.failOnError.Load() {
+			return fmt.Appendf(nil, "transformer request failed: %s", err), transformerutils.TransformerRequestFailure, nil
 		}
 		return nil, 0, err
 	}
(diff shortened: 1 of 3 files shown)


processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -409,10 +453,34 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels
...
+// kube-proxy returning a no-endpoints status.
+func isColdStartError(err error, resp *http.Response) bool {
+	if err != nil {
+		if errors.Is(err, syscall.ECONNREFUSED) {
+			return true
+		}
+		var dnsErr *net.DNSError
+		return errors.As(err, &dnsErr)
+	}
+	if resp != nil && (resp.StatusCode == http.StatusServiceUnavailable ||
```
itsmihir, line 480: why `http.StatusServiceUnavailable`? 503 status code is used by pyt so that rudder-server retry a request

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -393,6 +430,13 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels
 		} else if u.config.failOnError.Load() {
 			return fmt.Appendf(nil, "transformer request failed: %s", err), transformerutils.TransformerRequestFailure, nil
 		}
+		// Per-workspace PyT: a persistent transport failure on the workspace-
+		// scoped URL is most likely a cold-start window (pod not yet scaled by
+		// HPA, EndpointSlice lag, kube-proxy 5xx). Surface a dedicated status
+		// code so sendBatch retries the whole call until the pod is up instead of treating it as a failed transformation.
+		if errors.Is(err, transformerutils.ErrColdStart) {
+			return fmt.Appendf(nil, "workspace transformer not reachable: %s", err), transformerutils.StatusColdStartWindowFailure, nil
+		}
```
itsmihir, line 430: Should this be above the if condition on line 428? in case failOnError is set to true we wouldn't retry on cold start error.

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -342,6 +361,24 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels
 			u.stat.NewTaggedStat("transformer_client_request_total_bytes", stats.CountType, tags).Count(len(rawJSON))
 			u.stat.NewTaggedStat("transformer_client_total_durations_seconds", stats.CountType, tags).Count(int(duration.Seconds()))
 
+			// This metric is to track cold start errors for PyT, which are expected to be higher than usual due to the nature of PyT scaling.
+			if u.config.perWorkspacePyTEnabled &&
+				!u.config.forMirroring &&
+				strings.HasPrefix(labels.Language, "python") &&
+				labels.WorkspaceID != "" &&
```
itsmihir: can we move all this conditions to `isColdStartError` function?

processor/internal/transformer/user_transformer/user_transformer.go
mihir20: While the pyt pods are not ready, they can send 503 so we would want to retry those requests. Anyway we will just use this metric to scale from 0->1 so that's not an issue. also we are retaining the retry behaviour so IG we should be okay.

processor/internal/transformer/user_transformer/user_transformer.go
mihir20: yeah, makes sense 👍

processor/internal/transformer/user_transformer/user_transformer.go
mihir20: We can keep the isColdStartError function to classify the errors and create a new function for adding other checks.

itsmihir (approved):

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -409,10 +449,42 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels
...
+	}
+	return false
+}
+
+func (u *Client) userTransformURL(language, versionID, workspaceID string) string {
 	isPython := strings.HasPrefix(language, "python")
-	if isPython && u.config.pythonTransformConfig.IsVersionAllowed(versionID) && u.config.pythonTransformationURL != "" {
-		return u.config.pythonTransformationURL + "/customTransform"
+	if isPython && u.config.pythonTransformConfig.IsVersionAllowed(versionID) {
+		if u.config.perWorkspacePyTEnabled && !u.config.forMirroring && workspaceID != "" {
```
fracasula: Shouldn't we complain if `perWorkspacePyTEnabled` is true but the `workspaceID` is empty?

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -409,10 +449,42 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels
 	return respData, resp.StatusCode, nil
 }
 
-func (u *Client) userTransformURL(language, versionID string) string {
+func (u *Client) shouldThrowPythonColdStartErr(labels types.TransformerMetricLabels, err error, resp *http.Response) bool {
+	return u.config.perWorkspacePyTEnabled &&
```
fracasula: We are not checking `IsVersionAllowed` here. We shoud centralize these checks and think about removing `IsVersionAllowed` now that is not needed anymore.

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -248,9 +261,6 @@ func (u *Client) sendBatch(
 			)
 		}),
 	}
-	if !u.config.cpDownEndlessRetries.Load() {
-		retryOptions = append(retryOptions, backoff.WithMaxTries(uint(u.config.maxRetry.Load()+1)))
-	}
 	_ = backoffvoid.Retry(
 		context.Background(),
```
fracasula, line 265: Endless cold-start retry ignores the caller `ctx`. This can outlive Transform / drain. Graceful shutdown / drain semantics break for the new default, we should plumb the outer `ctx` into `backoffvoid.Retry`.

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -72,18 +78,21 @@ func New(conf *config.Config, log logger.Logger, stat stats.Stats, opts ...Opt)
...
+		forMirroring                  bool
+		maxRetry                      config.ValueLoader[int]
+		cpDownEndlessRetries          config.ValueLoader[bool]
+		failOnUserTransformTimeout    config.ValueLoader[bool]
+		failOnError                   config.ValueLoader[bool]
+		maxRetryBackoffInterval       config.ValueLoader[time.Duration]
+		timeoutDuration               time.Duration
+		collectInstanceLevelStats     bool
+		batchSize                     config.ValueLoader[int]
+		perWorkspacePyTEnabled        bool
```
fracasula: Any reason why this isn't reloadable but the endless retries is?

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -48,6 +51,9 @@ func New(conf *config.Config, log logger.Logger, stat stats.Stats, opts ...Opt)
 	handle.client = transformerclient.NewClient("UserTransformer", transformerutils.TransformerClientConfig(conf, "UserTransformer"))
 	handle.config.userTransformationURL = handle.conf.GetStringVar(handle.conf.GetStringVar("http://localhost:9090", "DEST_TRANSFORM_URL"), "USER_TRANSFORM_URL")
 	handle.config.pythonTransformationURL = handle.conf.GetStringVar("", "PYTHON_TRANSFORM_URL")
+	handle.config.perWorkspacePyTEnabled = handle.conf.GetBoolVar(false, "Processor.UserTransformer.perWorkspacePyTEnabled")
+	handle.config.perWorkspacePyTURLTemplate = handle.conf.GetStringVar("http://pyt-{workspaceID}:9090", "Processor.UserTransformer.perWorkspacePyTURLTemplate")
+	handle.config.perWorkspacePyTEndlessRetries = handle.conf.GetReloadableBoolVar(true, "Processor.UserTransformer.perWorkspacePyTEndlessRetries")
```
fracasula, line 56: perWorkspacePyTEndlessRetries=false doesn't seem to be covered by tests

processor/internal/transformer/user_transformer/user_transformer_test.go
```diff
@@ -1407,3 +1413,397 @@ func TestTransformerEvent_ToUserTransformerEvent(t *testing.T) {
...
+		return nil, f.failErr
+	}
+	return &http.Response{
+		StatusCode: http.StatusOK,
+		Header:     hdr,
+		Body:       io.NopCloser(bytes.NewReader(f.successBody)),
+	}, nil
+}
+
+func TestColdStartCounter(t *testing.T) {
```
fracasula, line 1652: We don't verify that the counter stays 0 when each guard is
individually false, even with a cold-start-looking transport error.
Every `TestColdStartCounter` case has `perWorkspacePyTEnabled=true`, `language=pythonfaas`, `workspaceID="ws-A1"`, `forMirroring=false`. None of the four guard negatives is asserted: flag off, mirroring on, JS
language, empty workspaceID.

processor/internal/transformer/user_transformer/user_transformer.go
mihir20: We do not need to check IsVersionAllowed here as this will be for all the pyt. can pick up removal of IsVersionAllowed separately;

fracasula: Yes, please create a task for this 🙏

processor/internal/transformer/user_transformer/user_transformer.go
mihir20: As the processor code currently is, we cannot use caller context; else, we might end up dropping the current jobs. We will have to refactor the processor code for this. will create a separate task for this as well.

processor/internal/transformer/user_transformer/user_transformer.go
mihir20: making reloadable

processor/internal/transformer/user_transformer/user_transformer.go
mihir20: https://linear.app/rudderstack/issue/PIPE-3038/cleanup-transformer-version-filter-in-ut-client

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -388,6 +421,13 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels
 	if err != nil {
 		u.log.Errorn("User transformation post error",
 			append(labels.ToLoggerFields(), obskit.Error(err))...)
+		// Per-workspace PyT: a persistent transport failure on the workspace-
+		// scoped URL is most likely a cold-start window (pod not yet scaled by
+		// HPA, EndpointSlice lag, kube-proxy 5xx). Surface a dedicated status
+		// code so sendBatch retries the whole call until the pod is up instead of treating it as a failed transformation.
+		if errors.Is(err, transformerutils.ErrColdStart) {
+			return fmt.Appendf(nil, "workspace transformer not reachable: %s", err), transformerutils.StatusColdStartWindowFailure, nil
```
fracasula, line 429: Can the error surface to the user in any way? :thinking:
If yes we could remove it from here and just log it, to avoid potentially leaking internal DNSs, IPs and other information.

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -409,9 +449,57 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels
...
+	}
+	if resp != nil && (resp.StatusCode == http.StatusServiceUnavailable ||
+		resp.StatusCode == http.StatusBadGateway) {
+		return true
+	}
+	return false
+}
+
+func (u *Client) userTransformURL(language, versionID, workspaceID string) string {
+	if !strings.HasPrefix(language, "python") {
```
fracasula: There are 3 occurrences already of `strings.HasPrefix(language, "python")`. Sounds like we need a function.

fracasula (approved):
@mihir20 I'm approving this to unblock you so that you can proceed with the release on Monday but I think this is incomplete.

We have an existing feature (mirroring) and a new one introduced here (python transformations for multi-tenant setups). With your implementation the two features simply won't work together given that you're explicitly skipping when `!forMirroring`.

Mirroring is here to stay though and we should retain the ability to mirror also for multi-tenant setups.

A couple considerations about this:
* mirroring will be needed to test different python transformation versions to check sanity diffs (for example [v1.14.0](https://github.com/rudderlabs/openfaas-flask-base/releases/tag/v1.14.0) was never put live but it's handling the metadata the same as the JS transformer. eventually we'll want to do the same here but before enabling we'll need mirroring to evaluate customer impact)
* mirroring should be as close as possible to the production setup. so if we're using one set of pods to transformer per workspace, the same should be done for mirroring.

cc @ktgowtham

processor/internal/transformer/user_transformer/user_transformer.go
mihir20: this will not go to the user; it will be just logged and retried

PR #6788, opened 2026-03-22
refactor(webhook): use gateway request context for transformer adapter timeout

akshit-git24:
Hi @atzoum @mihir20 
# Description

This PR refactors the existing context in the batchTransformLoop in the webhook.go . Currently there is a TODO : Remove timeout from here after timeout handler is added in gateway. The worker was creating an isolated, hardcoded 10-second timer instead of properly inheriting the master HTTP timeout rules from the original incoming webhook request. . After the changes made in [#6751](https://github.com/rudderlabs/rudder-server/pull/6751), we can now safely use the original incoming webhook request as same as of gateway. 

## changes
- Removed `context.WithTimeout` and the associated `cancel()` function calls in `batchTransformLoop`.
- Replaced the manual context with `req.request.Context()`.
- Deleted the resolved `TODO` comment.

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

PR #6988, opened 2026-05-23
fix(webhook): propagate request context in batchTransformLoop

reki7o7:
This PR resolves #6787.

The previous implementation used a hardcoded timeout context when calling the source transformer adapter, which detached execution from the original request lifecycle and ignored gateway-level timeout/cancellation handling.
This change replaces that with the HTTP request context from the batch request so transformer calls respect the existing gateway timeout behavior.
Note: I was unable to fully run the package tests locally because the repository currently has Windows-specific build incompatibilities (`syscall.Stat_t` in `utils/misc`), but the change is minimal and scoped to the requested fix.This is a clean resubmission of the previously stale PR #6788 with the requested minimal change.

### DIFF
--- a/gateway/webhook/webhook.go
+++ b/gateway/webhook/webhook.go
@@ -321,8 +321,7 @@ func (webhook *HandleT) batchRequests(sourceDef string, requestQ chan *webhookT)
 func (bt *batchWebhookTransformerT) batchTransformLoop() {
 	for breq := range bt.webhook.batchRequestQ {
 		// If unable to fetch features from transformer, send GatewayTimeout to all requests
-		// TODO: Remove timeout from here after timeout handler is added in gateway
-		ctx, cancel := context.WithTimeout(context.Background(), config.GetDurationVar(10, time.Second, "WriteTimeout", "WriteTimeOutInSec"))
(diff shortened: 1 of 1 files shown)


gateway/webhook/webhook.go
```diff
@@ -321,8 +321,7 @@ func (webhook *HandleT) batchRequests(sourceDef string, requestQ chan *webhookT)
 func (bt *batchWebhookTransformerT) batchTransformLoop() {
 	for breq := range bt.webhook.batchRequestQ {
 		// If unable to fetch features from transformer, send GatewayTimeout to all requests
-		// TODO: Remove timeout from here after timeout handler is added in gateway
-		ctx, cancel := context.WithTimeout(context.Background(), config.GetDurationVar(10, time.Second, "WriteTimeout", "WriteTimeOutInSec"))
+		ctx := breq.batchRequest[0].request.Context()
```
atzoum, line 324: This doesn’t really improve things. In fact, it introduces a new bug by tying the webhook context to the first request’s context. Since a batch can contain multiple requests, if the client from the first request disconnects, all other clients in the batch will receive an error response.

Maybe we should just remove the TODO to avoid confusing people and agents.

Issue #6349, opened 2025-09-15
Enabling "Enable for cleanup of object storage files (deletion) after successful sync" causes constant "context deadline exceeded" errors


bradtgmurray:
**Describe the bug**
Ever since enabling this flag my logs are full of these context deadline exceeded errors, is there a goroutine that's scoped to the request scope that's being cancelled by the request completing?

<img width="1454" height="1054" alt="Image" src="https://github.com/user-attachments/assets/aeca3540-c168-4485-afe5-5b52c5baaef5" />

**Steps to reproduce the bug**
Check the box, errors start happening

**Expected behavior**
No errors should be seen

rudder-sever: 1.55.2

bradtgmurray:
Not too familiar with the job.ctx lifecycle, but is it still vaild when this defer runs?

https://github.com/rudderlabs/rudder-server/blob/cf9f65324f98dc910498a1cebc89b2618a8c299f/warehouse/router/upload.go#L332

gitcommitshow:
Hi @bradtgmurray , I missed this issue earlier. Did you figure this out?
Do you see any other logs for additional information?
And which warehouse are you facing the issue with?

achettyiitr:
> Not too familiar with the job.ctx lifecycle, but is it still vaild when this defer runs?
> 
> [rudder-server/warehouse/router/upload.go](https://github.com/rudderlabs/rudder-server/blob/cf9f65324f98dc910498a1cebc89b2618a8c299f/warehouse/router/upload.go#L332)
> 
> Line 332 in [cf9f653](/rudderlabs/rudder-server/commit/cf9f65324f98dc910498a1cebc89b2618a8c299f)
> 
>  defer whManager.Cleanup(job.ctx)

We are cleaning up files in [here](https://github.com/rudderlabs/rudder-server/blob/master/warehouse/router/upload.go#L568-L572). By default, the timeout for FileManager is set to 2 minutes. @bradtgmurray Can you try increasing timeout via config: `FileManager.GCS.timeout: 5m`

PR #6996, opened 2026-05-24
fix: use context.Background() for cleanup operations to prevent context deadline exceeded errors

RockyOmvi:
## Description
When "Enable for cleanup of object storage files (deletion) after successful sync" is enabled, cleanup operations were using `job.ctx` which may be cancelled or have a deadline that expires. This caused constant "context deadline exceeded" errors as reported in issue #6349.

## Root Cause
The `job.ctx` context is derived from the router's errgroup context and may be cancelled at any time (e.g., on shutdown, errgroup errors, or parent context cancellation). Cleanup operations like `whManager.Cleanup()` and `cleanupObjectStorageFiles()` were using this context, causing failures when the context was no longer valid.

## Changes
1. **warehouse/router/upload.go**:
   - `defer whManager.Cleanup(job.ctx)` changed to `defer whManager.Cleanup(context.Background())` - ensures warehouse manager cleanup (closing DB connections, dropping dangling staging tables) runs regardless of job context state
   - `job.loadFilesRepo.Get(job.ctx, ...)` changed to `job.loadFilesRepo.Get(context.Background(), ...)` - ensures fetching load files for cleanup isn't tied to job context
   - `errgroup.WithContext(job.ctx)` changed to `errgroup.WithContext(context.Background())` - ensures object storage file deletions aren't cancelled by job context cancellation

2. **warehouse/router/identities.go**:
   - `defer whManager.Cleanup(ctx)` changed to `defer whManager.Cleanup(context.Background())` - same fix for identity table cleanup

## Testing
- Existing tests continue to work since they use `context.Background()` for `job.ctx`
- The filemanager already has its own configurable timeout (default 2 minutes), so using `context.Background()` won't cause hangs

### DIFF
--- a/warehouse/router/identities.go
+++ b/warehouse/router/identities.go
@@ -455,7 +455,7 @@ func (r *Router) populateHistoricIdentities(ctx context.Context, warehouse model
 			_, _ = job.setUploadError(err, model.Aborted)
 			return
 		}
-		defer whManager.Cleanup(ctx)
+		defer whManager.Cleanup(context.Background())
 		_ = job.setUploadStatus(UploadStatusOpts{Status: inProgressState(model.ExportedData)})
 		loadErrors, err := job.loadIdentityTables(true)
 		if err != nil {
--- a/warehouse/router/upload.go
+++ b/warehouse/router/upload.go
@@ -326,7 +326,7 @@ func (job *UploadJob) run() (err error) {
 		_, _ = job.setUploadError(err, InternalProcessingFailed)
 		return err
 	}
-	defer whManager.Cleanup(job.ctx)
+	defer whManager.Cleanup(context.Background())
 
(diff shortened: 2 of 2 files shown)


achettyiitr:
@RockyOmvi As mentioned in the [thread](https://github.com/rudderlabs/rudder-server/issues/6349#issuecomment-4088233198), We are cleaning up files in [here](https://github.com/rudderlabs/rudder-server/blob/master/warehouse/router/upload.go#L568-L572). By default, for [Delete](https://github.com/rudderlabs/rudder-go-kit/blob/main/filemanager/gcsmanager.go#L130-L146) opration the [timeout for FileManager ](https://github.com/rudderlabs/rudder-server/issues/6349#issuecomment-4088233198) is set to 2 minutes.

Updaing ctx, won't fix the issue.

PR #7006, opened 2026-05-25
Fix/#7001

### LINKED ISSUE #7001: bug: unbuffered batchRequestQ in webhook causes global stall across all source types
**Describe the bug**

In `gateway/webhook/setup.go`, the `batchRequestQ` channel is created as **unbuffered**:

```go
webhook.batchRequestQ = make(chan *batchWebhookT)   // setup.go:88
```

This channel is used by the `batchRequests` goroutines (one per source type) to send batches of webhook requests to the `batchTransformLoop` goroutines (up to `maxTransformerProcess`, default 64). Because the channel is unbuffered, every `batchRequestQ <- &breq` call blocks until a `batchTransformLoop` goroutine reads from it.

When all `maxTransformerProcess` goroutines are busy waiting for the transformer HTTP service to respond (which can happen under load or if the transformer is slow), the channel fills up. This cascades:

1. `batchRequests` blocks on `batchRequestQ <- &breq` → it cannot drain `requestQ`
2. `RequestHandler` blocks on `requestQ <- &req` → the HTTP handler never returns
3. **ALL webhook requests for ALL source types are blocked** by a single slow transformer call

There is no isolation between source types in the batching pipeline.

**Code reference**

`gateway/webhook/setup.go:88`:
```go
webhook.batchRequestQ = make(chan *batchWebhookT)
```

`gateway/webhook/webhook.go:306,315,322` — three blocking sends to this channel:
```go
webhook.batchRequestQ <- &breq
```

`gateway/webhook/webhook.go:361` — single unbuffered receiver:
```go
for breq := range bt.webhook.batchRequestQ {
```

**Expected behavior**

The `batchRequestQ` should have a reasonable buffer size (e.g., equal to `maxWebhookBatchSize * number_of_source_types`) so that `batchRequests` goroutines can continue draining their `requestQ` channels even when transformer goroutines are temporarily busy. This provides backpressure isolation between source types.

**Additional context**

The `batchRequests` goroutines loop (lines 288-326) selects between `requestQ` receipts and a batch timeout. When a batch completes, it sends to the unbuffered `batchRequestQ`. If the send blocks, the goroutine cannot service the `requestQ` select case, and all HTTP handlers for that source type pile up — then cascade to other source types as the runtime's goroutine scheduler interleaves them.

bujjibabukatta:
Add buffer to batchRequestQ channel
batchRequestQ was unbuffered, causing batchRequests goroutines to block on every send until a batchTransformLoop worker was free. Under transformer load, this cascaded — blocking requestQ draining and eventually stalling HTTP handlers across all source types.
Adds a configurable buffer (default: maxTransformerProcess, i.e. 64) via Gateway.webhook.batchRequestQSize, allowing batchRequests goroutines to continue draining their per-source requestQ even when all transformer workers are temporarily busy.
Files changed:

gateway/webhook/webhook.go — new batchRequestQSize config field on HandleT
gateway/webhook/setup.go — register config var and pass buffer size to make(chan)

fix #7001

### DIFF
--- a/gateway/webhook/setup.go
+++ b/gateway/webhook/setup.go
@@ -61,6 +61,8 @@ func Setup(gwHandle Gateway, transformerFeaturesService TransformerFeaturesServi
 	webhook := &HandleT{gwHandle: gwHandle, stats: stat, logger: logger.NewLogger().Child("gateway").Child("webhook")}
 	// Number of incoming webhooks that are batched before calling source transformer
 	webhook.config.maxWebhookBatchSize = conf.GetReloadableIntVar(32, 1, "Gateway.webhook.maxBatchSize")
+	// Buffer size for the batchRequestQ channel; isolates batchRequests goroutines from slow transformer workers
+	webhook.config.batchRequestQSize = conf.GetReloadableIntVar(maxTransformerProcess, 1, "Gateway.webhook.batchRequestQSize")
 	// Timeout after which batch is formed anyway with whatever webhooks are available
 	webhook.config.webhookBatchTimeout = conf.GetReloadableDurationVar(20, time.Millisecond, []string{"Gateway.webhook.batchTimeout", "Gateway.webhook.batchTimeoutInMS"}...)
 	// Multiple source transformers are used to generate rudder events from webhooks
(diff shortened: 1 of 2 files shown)


bramhanandlingala (approved):

bujjibabukatta:
Hi @lokey @atzoum @rudder-server-bot  @itsmihir @aris1009 @fracasula  @akshit-git24  Could you please review and approve pull request.

bramhanandlingala (approved):

atzoum:
Not an issue

PR #7007, opened 2026-05-25
chore(jobsdb): lateral join for getting jobs

atzoum:
# Description

Adds a reloadable config option (`getJobsUseLateralJoin`, default `true`) that switches the `getJobsDS` query from joining the `v_last_*` view to a correlated `LATERAL` subquery against the raw job status table.

Driving the join from the jobs side rather than scanning the full `v_last_*` view can yield significant performance improvements, e.g. [plan using v_last](https://explain.dalibo.com/plan/158a74gfc7df8a44) vs [plan using lateral join](https://explain.dalibo.com/plan/f21ghgh581edb851)

## Security


ktgowtham (approved):

mihir20 (approved):
LGTM

PR #7008, opened 2026-05-25
fix(warehouse/slave): protect outputFileWritersMap and tableEventCoun…

### LINKED ISSUE #7002: bug: unprotected concurrent map reads in warehouse slave worker_job.go (data race)
**Describe the bug**

In `warehouse/slave/worker_job.go`, the `uploadLoadFiles` method reads `jr.outputFileWritersMap` and `jr.tableEventCountMap` from both the main goroutine and a spawned upload goroutine **without holding their respective mutexes** (`outputFileWritersMapMu` / `tableEventCountMapMu`).

This is a data race detectable by the Go race detector.

**Code reference**

`warehouse/slave/worker_job.go:415` (main goroutine — no lock):
```go
processStream := make(chan *uploadProcessingResult, len(jr.outputFileWritersMap))
```

`warehouse/slave/worker_job.go:420-450` (upload goroutine — no lock):
```go
go func() {
    for tableName, uploadFile := range jr.outputFileWritersMap {
        // ...
        TotalRows: jr.tableEventCountMap[tableName],
    }
}()
```

`warehouse/slave/worker_job.go:473` (main goroutine — no lock):
```go
output := make([]uploadResult, 0, len(jr.outputFileWritersMap))
```

`warehouse/slave/worker_job.go:483` (main goroutine — no lock):
```go
if len(output) != len(jr.outputFileWritersMap) { ... }
```

**Steps to reproduce**

Run the warehouse tests with the `-race` flag:
```
go test -race ./warehouse/slave/...
```

**Expected behavior**

All reads of `outputFileWritersMap` and `tableEventCountMap` should be protected by `outputFileWritersMapMu.RLock()` / `tableEventCountMapMu.RLock()`:

```go
jr.outputFileWritersMapMu.RLock()
defer jr.outputFileWritersMapMu.RUnlock()
for tableName, uploadFile := range jr.outputFileWritersMap { ... }
```

**Additional context**

The `uploadLoadFiles` function is called from `processStagingFile` (line 499) after a call to `uploadLoadFiles` on the previous line. The maps are populated in `handlePendingStagingFile` which holds the write locks. While in practice `uploadLoadFiles` may be called after all map population is complete, there is no formal synchronization guarantee, and the Go race detector correctly flags these as unsynchronized concurrent reads.

bujjibabukatta:
Fix data race on outputFileWritersMap and tableEventCountMap in uploadLoadFiles

Four unprotected reads of outputFileWritersMap and tableEventCountMap inside uploadLoadFiles were flagged by the Go race detector (go test -race ./warehouse/slave/...). The maps are written under write locks elsewhere but read without any lock in this method.

Fixes by taking snapshots of both maps under RLock before the upload goroutine starts, and using expectedCount (captured once under RLock) for the output slice and length check.
Files changed:

warehouse/slave/worker_job.go — add RLock/RUnlock guards around all four unprotected reads in uploadLoadFiles

fix #7002

### DIFF
--- a/warehouse/slave/worker_job.go
+++ b/warehouse/slave/worker_job.go
@@ -413,6 +413,10 @@ func (jr *jobRun) uploadLoadFiles(ctx context.Context, modifier func(result uplo
 
 	process := func() <-chan *uploadProcessingResult {
 		processStream := make(chan *uploadProcessingResult, len(jr.outputFileWritersMap))
+		jr.outputFileWritersMapMu.RLock()
+		processStreamSize := len(jr.outputFileWritersMap)
+		jr.outputFileWritersMapMu.RUnlock()
+		processStream := make(chan *uploadProcessingResult, processStreamSize)
 
 		g, groupCtx := errgroup.WithContext(ctx)
 		g.SetLimit(jr.config.numLoadFileUploadWorkers)
@@ -420,7 +424,20 @@ func (jr *jobRun) uploadLoadFiles(ctx context.Context, modifier func(result uplo
 		go func() {
 			defer close(processStream)
 
-			for tableName, uploadFile := range jr.outputFileWritersMap {
+			jr.outputFileWritersMapMu.RLock()
++         jr.tableEventCountMapMu.RLock()
++         writersSnapshot := make(map[string]encoding.LoadFileWriter, len(jr.outputFileWritersMap))
++         for k, v := range jr.outputFileWritersMap {
++             writersSnapshot[k] = v
++         }
++         countSnapshot := make(map[string]int, len(jr.tableEventCountMap))
++         for k, v := range jr.tableEventCountMap {
++             countSnapshot[k] = v
++         }
++         jr.tableEventCountMapMu.RUnlock()
++         jr.outputFileWritersMapMu.RUnlock()
+
+			for tableName, uploadFile := range writersSnapshot  {
 				g.Go(func() error {
 					select {
 					case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


bramhanandlingala (approved):

bujjibabukatta:
Hi @atzoum @fracasula  @itsmihir @mihir20 Could you please review and approve pull request ?

bramhanandlingala (approved):

bujjibabukatta:
Hi @ItsSudip  Could you please approve pull request .

warehouse/slave/worker_job.go
achettyiitr, line 1: At the point `uploadLoadFiles` runs, nothing is writing those maps, so no lock is needed — not the `RLock`, not the snapshot, none of it.

The reason is the call sequence in `processMultiStagingFiles`:

```go
for _, f := range job.StagingFiles {
    g.Go(func() error { ... writer()/incrementEventCount() write the maps ... })
}
g.Wait()                  // all writers finished AND synchronized here
jr.closeLoadFiles()
jr.uploadLoadFiles(...)   // runs only after the barrier
```

Every write to `outputFileWritersMap` / `tableEventCountMap` happens inside those errgroup goroutines. `g.Wait()` is a happens-before barrier: it both waits for them to finish and synchronizes their memory effects. So by the time `uploadLoadFiles` is entered, the write phase is over and there's no concurrent writer.

Inside `uploadLoadFiles`, every access to the maps is a read — the `range`, the `len()` calls, `tableEventCountMap[tableName]`. Multiple concurrent reads with no writer is safe in Go and isn't a race. The existing mutexes do their job earlier, guarding the concurrent *write* phase; they're simply not relevant here.

achettyiitr: 1. Can you share the actual `go test -race` output? Which test failed?
2. Since `uploadLoadFiles` runs after `g.Wait()`, there's no writer at that point — what's the concurrent write the race detector saw?
3. Did this build locally? There are two `processStream :=` lines now, which shouldn't compile. Also the `+`.

@bujjibabukatta

PR #7020, opened 2026-05-29
feat(jobsdb): deferred status table lock during compaction

atzoum:
# Description

During non-blocking compaction, the source job status table is locked `EXCLUSIVE` early, before jobs are copied to the destination. Because the job copy includes the (potentially large) `event_payload`, the bulk of the lock window is spent moving data that does not actually need the lock — status updates for jobs in that dataset are blocked for the whole copy.

This PR adds an opt-in mode (`compactionDeferStatusLock`, off by default) that splits the copy into two phases:

1. **Copy pending jobs first, with no status lock held** — the slow, payload-heavy work.
2. **Then lock each source status table one after another** only to copy the latest statuses of the jobs that were moved.

The result is that the status-table lock window shrinks to roughly the (small, payload-free) status copy, minimizing how long status updates are blocked. As a side-effect, jobs that complete between the two phases are still moved to the new table (now carrying their terminal status). These few extra completed rows will get reclaimed on the next compaction.

#7021 runs all tests with this feature enabled

## Benchmark

Deferring the status lock shaves a further ~12% off `nonBlockingCompaction` alone.
Run it with:

```bash
RUN_COMPACTION_BENCH=1 GOMAXPROCS=2 go test ./jobsdb/bench/ -run 'TestBench/compaction' -v -timeout 30m
```

Drain time for 2M jobs (lower is better):

| Scenario                                   | Time      | vs. baseline |
| ------------------------------------------ | --------- | ------------ |
| baseline (blocking compaction)             | 2m56.6s   | —            |
| nonBlockingCompletedDSDrop                 | 2m10.9s   | −25.9%       |
| nonBlockingCompaction                      | 1m54.8s   | −35.0%       |
| nonBlockingCompaction + deferStatusLock    | 1m40.7s   | −43.0%       |

## Linear Ticket

resolves PIPE-3064

## Security

---
This is **part 2 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #7038 
- <kbd>&nbsp;3&nbsp;</kbd> #7021 
- <kbd>&nbsp;2&nbsp;</kbd> #7020 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #7024

### DIFF
--- a/jobsdb/bench/scenario/compaction.go
+++ b/jobsdb/bench/scenario/compaction.go
@@ -58,6 +58,7 @@ type flagSet struct {
 	name                       string
 	nonBlockingCompletedDSDrop bool
 	nonBlockingCompaction      bool
+	compactionDeferStatusLock  bool
 }
 
 func (p *compaction) Run(ctx context.Context) error {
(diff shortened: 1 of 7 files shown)


mihir20 (approved):
LGTM

ktgowtham (approved):

PR #7022, opened 2026-05-29
fix(kafka): support insecure TLS skip verify config

ashkangoleh:
# Description

This PR adds an optional config-driven way to set `InsecureSkipVerify` on the TLS configuration used by regular Kafka destinations.

Supported config keys:

- `Router.KAFKA.tlsInsecureSkipVerify`
- `Router.kafkaTLSInsecureSkipVerify`
- `KAFKA_TLS_INSECURE_SKIP_VERIFY`

Default value: `false`.

## Why

`rudder-go-kit/kafkaclient.TLS` already supports `InsecureSkipVerify`, and the Kafka client already applies it when building the underlying `tls.Config`.

However, Rudder Server was not exposing a config/env-driven way to pass this value when creating the Kafka client for regular Kafka destinations.

This is useful for controlled self-hosted, internal, or test Kafka environments where TLS is enabled but the broker certificate cannot be validated through the configured CA certificate or the system certificate pool.

## Scope

This change is intentionally limited to the regular Kafka destination path in `NewProducer`.

It only applies when `destConfig.SslEnabled` is true.

This PR does not change:

- Kafka destinations with SSL disabled
- SASL behavior
- SSH behavior
- Azure Event Hubs
- Confluent Cloud
- producer batching/compression behavior
- `rudder-go-kit`

## Backward Compatibility

Backward compatible.

The new flag defaults to `false`, so existing Kafka TLS verification behavior remains unchanged unless the operator explicitly enables the new flag.

## Test Plan

- Ran `gofmt` on `services/streammanager/kafka/kafkamanager.go`
- Verified the new flag defaults to `false`
- Verified the flag is only read and applied when Kafka SSL is enabled
- Verified the existing CA certificate path is preserved
- Verified the existing system certificate pool path is preserved
- Verified SASL remains scoped to the SSL-enabled path

## Linear Ticket

No Linear ticket. Small config plumbing fix for Kafka TLS client configuration.

## Security

This PR does not change the default security posture because the new flag defaults to `false`.

When explicitly enabled, `KAFKA_TLS_INSECURE_SKIP_VERIFY=true` disables Kafka broker certificate chain and hostname verification for regular Kafka destinations. This is an operator-controlled escape hatch intended only for controlled internal, self-hosted, or test Kafka deployments where proper certificate validation is not available.

Go's `crypto/tls` documentation warns that `InsecureSkipVerify=true` accepts any certificate and hostname, making TLS susceptible to machine-in-the-middle attacks unless custom verification is used.

### DIFF
--- a/services/streammanager/kafka/kafkamanager.go
+++ b/services/streammanager/kafka/kafkamanager.go
@@ -290,14 +290,24 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (*Produ
 		SSHConfig:   sshConfig,
 	}
 	if destConfig.SslEnabled {
+		tlsInsecureSkipVerify := config.GetBoolVar(
+			false,
+			"Router.KAFKA.tlsInsecureSkipVerify",
+			"Router.kafkaTLSInsecureSkipVerify",
+			"KAFKA_TLS_INSECURE_SKIP_VERIFY",
+		)
 		if destConfig.CACertificate != "" {
 			clientConf.TLS = &client.TLS{
 				CACertificate: []byte(destConfig.CACertificate),
 				Cert:          clientCert,
 				Key:           clientKey,
+				InsecureSkipVerify: tlsInsecureSkipVerify,
 			}
 		} else {
-			clientConf.TLS = &client.TLS{WithSystemCertPool: true}
+			clientConf.TLS = &client.TLS{
+				WithSystemCertPool: true,
+				InsecureSkipVerify: tlsInsecureSkipVerify,
+			}
 		}
 
 		if destConfig.UseSASL { // SASL is enabled only with SSL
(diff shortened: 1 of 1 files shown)


khoramism:
I mean we've been having the same problem and tried to just circle around it. This addition would really simplify our infrastructure.

services/streammanager/kafka/kafkamanager.go
```diff
@@ -290,14 +290,24 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (*Produ
 		SSHConfig:   sshConfig,
 	}
 	if destConfig.SslEnabled {
+		tlsInsecureSkipVerify := config.GetBoolVar(
+			false,
+			"Router.KAFKA.tlsInsecureSkipVerify",
+			"Router.kafkaTLSInsecureSkipVerify",
+			"KAFKA_TLS_INSECURE_SKIP_VERIFY",
+		)
```
atzoum: Just a single config key would be enough. The relevant env variable would be `RSERVER_ROUTER_KAFKA_TLS_INSECURE_SKIP_VERIFY`
```suggestion
		insecureSkipVerify := config.GetBoolVar(false,"Router.KAFKA.tlsInsecureSkipVerify")
```

atzoum (changes requested):

atzoum (approved):

PR #7024, opened 2026-05-30
chore(jobsdb): avoid refreshing ds list from database in case of a stale ds error while updating job statuses

atzoum:
# Description

When updating job statuses, if a background compaction/migration has changed the dataset (DS) layout, the update hits a stale DS list (`ErrStaleDsList`) and retries. Previously the retry re-read the DS list **from the database**, which needs a connection from the main pool — so under high concurrency (in-flight updaters ≥ pool size) every connection is already in use, leaving none to perform the refresh, which can stall the whole update path.

The in-memory DS list cache is already kept up to date by the compaction/migration loop, so the retry doesn't need to touch the database. We now refresh from that cached snapshot under a read lock instead, removing the connection requirement from the retry path entirely.

## Benchmark

A new `compaction` jobsdb bench scenario (`TestBench/compaction`) drains a read-write jobsdb seeded with 2M jobs (20 datasets × 100K, 3KB payload) across 100 destinations, using 100 consumer goroutines against a pool of 80 connections, and measures the time for all consumers to fully drain. It is also used to compare the available compaction strategies under an identical workload.

Run it with:

```bash
RUN_COMPACTION_BENCH=1 GOMAXPROCS=2 go test ./jobsdb/bench/ -run 'TestBench/compaction' -v -timeout 30m
```

Drain time for 2M jobs (lower is better):

| Scenario                          | Time         | vs. baseline |
| --------------------------------- | ------------ | ------------ |
| baseline (blocking compaction)    | 2m56.6s      | —            |
| nonBlockingCompletedDSDrop        | 2m10.9s      | −25.9%       |
| nonBlockingCompaction             | 1m54.8s      | −35.0%       |

## Linear Ticket

resolves PIPE-2997

## Security

---
This is **part 1 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #7038 
- <kbd>&nbsp;3&nbsp;</kbd> #7021 
- <kbd>&nbsp;2&nbsp;</kbd> #7020 
- <kbd>&nbsp;1&nbsp;</kbd> #7024 👈

### DIFF
--- a/jobsdb/bench/bench.go
+++ b/jobsdb/bench/bench.go
@@ -23,6 +23,8 @@ func New(conf *config.Config, stat stats.Stats, log logger.Logger, db *sql.DB) (
 		return scenario.NewSimple(conf, stat, log, db), nil
 	case "two_stage":
 		return scenario.NewTwoStage(conf, stat, log, db), nil
+	case "compaction":
+		return scenario.NewCompaction(conf, stat, log, db), nil
 	default:
 		return nil, fmt.Errorf("unknown jobsdb bench scenario name: %s", conf.GetStringVar("processor", "JobsDB.bench.scenario"))
 	}
--- a/jobsdb/bench/bench_test.go
+++ b/jobsdb/bench/bench_test.go
@@ -3,6 +3,7 @@ package bench_test
 import (
 	"context"
 	"fmt"
+	"os"
 	"testing"
 	"time"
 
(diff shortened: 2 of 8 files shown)


mihir20 (approved):
LGTM

ktgowtham (approved):

PR #6963, opened 2026-05-15
chore(jobsdb): release dsList read lock early during store

atzoum:
🔒 Scanned for secrets using gitleaks 8.30.1

# Description

`jobsdb` currently retains the `dsList` read lock for the full duration of the store callback. As a result, long-running store transactions can block `dsList` writers (`addNewDS`, `migrateDSLoop`). Once a writer begins waiting on the write lock, new `dsList` readers may also become blocked.

Therefore, we are now snapshoting the last dataset under the read lock and release the lock immediately, so the store transaction can take as long as it needs without holding dsList writers back. If this turns out to cause regressions in production, the new reloadable config `jobsDB.holdDSListLockDuringStore` (default `false`) reverts to the previous hold-the-lock-throughout behavior, so no redeployment is required.

## Uncovered latent bugs

Releasing the read lock early exposed two latent bugs in existing tests:

1. **`WithStoreSafeTxFromTx` couldn't retry on a stale ds.** When the caller owned the outer `*Tx` and the store returned `errStaleDsList`, the savepoint inside `doStoreJobsInTx` was never rolled back. The retry inherited an aborted savepoint and the whole transaction failed. Fixed by rolling back to the savepoint before returning `ErrStaleDsList`.

2. **The JSON sanitize path was missing the same stale-ds handling.** Once the rollback was in place, it became obvious that the post-sanitize `store()` retry could itself hit the read-only table and was not handling it either. `doStoreJobsInTx` is now a single loop that handles both conditions uniformly; the sanitize retry still runs at most once.

## Bonus

- `inStoreSafeCtx` no longer logs stale-ds at error level, since it's an expected retryable condition, not a failure.
- When many goroutines hit a stale ds at the same time (which is exactly what happens during a dataset swap), only the first one to grab the dsList write lock actually refreshes. The rest see that the last dataset has already moved past the one they failed on and skip the refresh, going straight back to the retry. This avoids a thundering herd of redundant refreshes.

## Linear Ticket

resolves PIPE-2993

## Security

- [x] The code changed/added as part of this pull request won't create any security issues with how the software is being used.

---
This is **part 1 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #6979 
- <kbd>&nbsp;3&nbsp;</kbd> #6967 
- <kbd>&nbsp;2&nbsp;</kbd> #6962 
- <kbd>&nbsp;1&nbsp;</kbd> #6963 👈

PR #7041, opened 2026-06-02
fix(jobsdb): pq 42P01 error when trying to store to a dataset that has been dropped

atzoum:
# Description

Since #6963, `Store` operates against a snapshot of the dataset list. In the rare case where a compaction manages to drop the previous rightmost dataset between when the snapshot was taken and when the store transaction executes, the `COPY INTO` fails with PostgreSQL error `42P01` (undefined table), treated as an unrecoverable error, causing server to panic.

**Fix:** Treat `42P01` (undefined table) the same as `RS001` (readonly trigger): return `ErrStaleDsList` so `inStoreSafeCtx` refreshes the dataset list and retries on the current last dataset. Set a maximum number of stale ds list retries to avoid infinite loops in test scenarios (default: 3).

## Linear Ticket

resolves PIPE-3080

## Security

### DIFF
--- a/jobsdb/jobsdb.go
+++ b/jobsdb/jobsdb.go
@@ -39,6 +39,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/lib/pq"
+	"github.com/lib/pq/pqerror"
 	"github.com/samber/lo"
 	"github.com/tidwall/gjson"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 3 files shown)


jobsdb/jobsdb.go
```diff
@@ -383,6 +385,9 @@ func (jd *Handle) UpdateJobStatusInTx(ctx context.Context, tx UpdateSafeTx, stat
 		if !errors.Is(err, ErrStaleDsList) {
 			return err
 		}
+		if attempt >= maxRetries {
+			return fmt.Errorf("stale dataset list after %d retries: %w", maxRetries, err)
+		}
```
mihir20, line 390: Should we do this after rollback?

ktgowtham (approved):

jobsdb/jobsdb.go
atzoum: no need to rollback, this would be a terminal failure

mihir20 (approved):

PR #7043, opened 2026-06-03
chore(jobsdb): refactoring and cleanup

atzoum:
## Description

- Drop unnecessary prepared statements
- Remove deprecated `StoreEachBatchRetry` and other dead code
- Rename `ConnectionDetails` → `ConnectionID` and `migration` → `compaction`
- Prefer new config keys over legacy ones
- Split code into smaller, logically grouped files
- Fix stale and missing docs

## Linear Ticket

resolves PIPE-3082

## Security

---
- <kbd>&nbsp;2&nbsp;</kbd> #7048 
- <kbd>&nbsp;1&nbsp;</kbd> #7043 👈


gateway/handle_lifecycle.go
```diff
@@ -487,22 +486,13 @@ func (gw *Handle) dbWriterWorkerProcess() {
 
 		ctx, cancel := context.WithTimeout(context.Background(), gw.conf.WriteTimeout)
 		err := gw.jobsDB.WithStoreSafeTx(ctx, func(tx jobsdb.StoreSafeTx) error {
-			if gw.conf.gwAllowPartialWriteWithErrors.Load() {
-				var err error
-				errorMessagesMap, err = gw.jobsDB.StoreEachBatchRetryInTx(ctx, tx, jobBatches)
```
ktgowtham, line 492: why are we removing this? none of the deployments has this enabled?

ktgowtham, line 492: why are we removing this? none of the deployments has this enabled?

ktgowtham (approved):

gateway/handle_lifecycle.go
atzoum: a. We are not using gateway for ingestion
b. This is an old feature introduced because occasionally some payloads were not able to be stored in jobsdb's jsonb column due to some invalid characters, so this was an attempt to avoid other requests being affected. We've come a long way since then, this shouldn't be an issue anymore even for opensource customers.
c. This code was pretty ugly and in some aspects wrong as well!

fracasula (approved):

PR #7050, opened 2026-06-05
feat(gateway): emit pipelineFirstEventReceived on first source ingestion

akashrpo:
🔒 Scanned for secrets using gitleaks 8.30.0

# Description

Adds the **First Event Received** activation milestone to the Phase 1 PLG funnel (Signup → … → First Connection → **First Event Received** → Converted to Paid). Powers funnel-completion rate and time-to-value (signup → first event).

A new `gateway/internal/firstevent.Tracker` emits `pipelineFirstEventReceived` the first time a source ingests an event, hooked in `getJobDataFromRequest` after jobs are built. Dedup is per-pod via an in-memory `sync.Map` (cheap `ShouldFire` guard runs before the source-config lookup so the hot path stays clean); the warehouse collapses cross-pod duplicates to first-occurrence-per-workspace.

Props: `sourceId`, `sourceType`, `sourceCategory`, `workspaceId`, `isFirstSourceEventInWorkspace`, `receivedAt`. Paired data-gov change relaxes `organization_id` + `time_since_source_created_seconds` to optional on `pipeline_first_event_received_rule` (not in gateway scope; org joined downstream via `workspace_id`).

**Ships disabled** — emits only when `PipelineActivation.enabled=true` and `PipelineActivation.writekey` is set (a product-analytics source in the funnel workspace, separate from the diagnostics writekey). No-op until infra provisions the key, so zero prod impact.

Verified: `go build ./gateway/...` ✓, `go vet` ✓, `firstevent` unit tests 3/3 ✓.

Follow-ups (not in scope): global cross-pod dedup via a DB table; `pipelineFirstDeliveryCompleted`; internal-batch ingest path.

## Linear Ticket

https://linear.app/rudderstack/issue/DAW-3552

## Security


gateway/handle.go
```diff
@@ -549,6 +551,13 @@ func (gw *Handle) getJobDataFromRequest(req *webRequestT) (jobData *jobFromReq,
 	}
 	err = nil
 	jobData.jobs = jobs
+	if len(jobs) > 0 && gw.firstEventTracker.ShouldFire(sourceID) {
+		sourceType := ""
+		if src, ok := gw.getSourceConfigFromSourceID(sourceID); ok {
+			sourceType = src.SourceDefinition.Name
+		}
+		gw.firstEventTracker.Track(sourceID, workspaceId, sourceType, sourceCategory, gw.now().Format(misc.RFC3339Milli))
```
ktgowtham, line 559: This code path doesn't get executed for RS hosted dataplanes. for RS hosted dataplanes, we hit [this](https://github.com/rudderlabs/rudder-server/blob/6e2009a9f942d71082f6419184a293126ae57c5a/gateway/handle_lifecycle.go#L602)

Also, with this kind of instrumentation, we keep on firing this event forever. not just first time. pods come and go, there are multiple pods.
I don't think this is the right way to instrument. 

Can't we run a periodic job for newly created sources(say last month) and check reporting db the first occurrence of gateway event metric and emit that timestamp?

I don't think here is the right place to track this. IMO, this should be done on CP side.

cc: @MiteshSharma

PR #7055, opened 2026-06-05
feat(gateway): throttling support for internal batch endpoint

atzoum:
# Description

Adding support for throttling the internal batch endpoint, using an in-memory throttler. Both event rate and event window is configurable and throttler can be disabled by setting the window to zero.
Throttler expects the presence of `X-Batch-Size` request header. If it is not present it will skip throttling and increment the `gateway_internal_batch_throttle_skip` counter with a `reason: no_header` label.

## Linear Ticket

resolves PIPE-3102

## Security

### DIFF
--- a/gateway/handle.go
+++ b/gateway/handle.go
@@ -52,6 +52,27 @@ type messageValidator interface {
 	Validate(payload []byte, message *stream.MessageProperties) (bool, error)
 }
 
+type handleConfig struct {
+	webPort, maxUserWebRequestWorkerProcess, maxDBWriterProcess                       int
+	maxUserWebRequestBatchSize, maxDBBatchSize, maxHeaderBytes, maxConcurrentRequests int
+	userWebRequestBatchTimeout, dbBatchWriteTimeout                                   config.ValueLoader[time.Duration]
+
+	maxReqSize                           config.ValueLoader[int]
+	enableRateLimit                      config.ValueLoader[bool]
+	internalBatchThrottleEvents          config.ValueLoader[int]
+	internalBatchThrottleWindow          config.ValueLoader[time.Duration]
+	enableSuppressUserFeature            bool
+	diagnosisTickerTime                  time.Duration
+	ReadTimeout                          time.Duration
+	ReadHeaderTimeout                    time.Duration
+	WriteTimeout                         time.Duration
+	IdleTimeout                          time.Duration
+	allowReqsWithoutUserIDAndAnonymousID config.ValueLoader[bool]
+	webhookV2HandlerEnabled              bool
+	internalEndpointsEnabled             bool
+	legacyWarehouseEndpointsEnabled      bool
+}
+
 type Handle struct {
 	// dependencies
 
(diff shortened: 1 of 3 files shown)


gateway/handle.go
```diff
@@ -693,6 +702,25 @@ func (gw *Handle) addToWebRequestQ(_ *http.ResponseWriter, req *http.Request, do
...
 }
 
+// internalBatchAllow checks whether the incoming batch of n events is within the configured
+// global throttle for the internal batch endpoint. It returns true if the request is allowed.
+// Rate and window are passed per-call so the limiter itself is created once and the config
+// can be reloaded without recreating it.
+func (gw *Handle) internalBatchAllow(ctx context.Context, n int) bool {
+	limit := gw.conf.internalBatchThrottleEvents.Load()
+	if limit <= 0 {
+		gw.internalBatchThrottleSkipDisabledCounter.Increment()
```
fracasula, line 716: 3 counters for such a small feature seem a bit overkill. Do we really need to increase a counter everytime we get a request just because we disabled throttling by setting the limit to 0? Wouldn't it make more sense to have a metric for throttle denials?

gateway/handle_test.go
```diff
@@ -992,3 +998,136 @@ func TestExtractJobsFromInternalBatchPayload_LiveEventRecording(t *testing.T) {
...
+	return &Handle{
+		logger:                                   logger.NOP,
+		stats:                                    statsStore,
+		internalBatchLimiter:                     l,
+		bodyReadTimeStat:                         statsStore.NewStat("gateway.http_body_read_time", stats.TimerType),
+		requestSizeStat:                          statsStore.NewStat("gateway.request_size", stats.HistogramType),
+		internalBatchThrottleSkipDisabledCounter: statsStore.NewTaggedStat("gateway.internal_batch_throttle_skip", stats.CountType, stats.Tags{"reason": "disabled"}),
+		internalBatchThrottleSkipNoHeaderCounter: statsStore.NewTaggedStat("gateway.internal_batch_throttle_skip", stats.CountType, stats.Tags{"reason": "no_header"}),
+		internalBatchThrottleSkipInvalidHeaderCounter: statsStore.NewTaggedStat("gateway.internal_batch_throttle_skip", stats.CountType, stats.Tags{"reason": "invalid_header"}),
+		conf: struct {
```
fracasula: This inline struct is erroring for me when I pull the code. This seems like it's been copied from other places as well, might be worth promoting it to its own type, perhaps even in handle.go so that we can reuse it here too. Wdyt?

gateway/handle.go
atzoum: 3 counters still cost less than 1 histogram :)
> Wouldn't it make more sense to have a metric for throttle denials?

Please elaborate on throttle denials

fracasula: For throttle denials I was suggesting a counter though 😅

atzoum: Have a counter to count how many times we return a 429? We have http stats for that purpose

fracasula (approved):

gateway/handle.go
```diff
@@ -709,8 +739,17 @@ func (gw *Handle) internalBatchHandlerFunc() http.HandlerFunc {
...
 
-		// TODO: add tracing
 		gw.logger.LogRequest(r)
+
+		if batchSizeStr := r.Header.Get("X-Batch-Size"); batchSizeStr == "" {
+			gw.internalBatchThrottleSkipNoHeaderCounter.Increment()
+		} else if batchSize, parseErr := strconv.Atoi(batchSizeStr); parseErr != nil || batchSize <= 0 {
+			gw.internalBatchThrottleSkipInvalidHeaderCounter.Increment()
+		} else if !gw.internalBatchAllow(ctx, batchSize) {
+			http.Error(w, response.GetStatus(response.TooManyRequests), http.StatusTooManyRequests)
```
ktgowtham, line 749: > response.TooManyRequests

NIT: should we return back `max events limit reached`

ktgowtham (approved):

gateway/handle.go
atzoum: Since this throttler only applies to the internal batch endpoint and no human or machine will ever going to read the response body, I will defer adding this additional entry into the status map :)

PR #7056, opened 2026-06-05
feat: add BQSTREAM_V2 destination type and bqstreamv2 building blocks

achettyiitr:
# Description
 
PR 1/3 of the BQSTREAM_V2 stack (splits #7019).
 
Adds the `BQSTREAM_V2` destination type, warehouse-side enablement, and all building blocks of the async destination manager — with no runtime behavior yet:
 
**Warehouse enablement**
- Added `BQSTREAM_V2` constant and treated it as a pseudo-warehouse destination (`warehouse/utils`).
- Routed `BQSTREAM_V2` to the BigQuery warehouse manager in `warehouse/integrations/manager` (`FetchSchema`/`CreateSchema`/`CreateTable`/`AddColumns` reuse).
- Embedded warehouse transformer support: `loaded_at` column, JSON path support, full event column type (`string`), users-table skip, BigQuery reserved keywords.
 
**bqstreamv2 building blocks** (`router/batchrouter/asyncdestinationmanager/bqstreamv2`)
- Package types and destination config decoding with namespace normalization (+ unit tests).
- BigQuery data type mapping, in-memory table schema cache with TTL, Storage Write API managed-writer wrapper.
- Event parsing from async files, per-table grouping and size-bounded chunking, event-schema derivation (`events.go`).
- Discard extraction for schema-mismatched values and duplicate ID detection.
- Dynamic protobuf row encoding with numeric/timestamp coercion (`encode.go`), plus protojson equivalence test and benchmarks.
- Schema diffing (`findNewColumns`) and BigQuery error classification (`schemautils.go`, + unit tests).
 
Inert until the Manager orchestration is added and registered (PR 2/3).
 
**Stack:** PR 1 (this) → PR 2 (upload orchestration + registration) → PR 3 (integration tests)
 
## Linear Ticket
 
- Resolves INT-6410
 
## Security
 
## Validation
 
- `go build ./...`
- `go test ./router/batchrouter/asyncdestinationmanager/bqstreamv2/...`
- `go test ./processor/internal/transformer/destination_transformer/embedded/warehouse/...`


warehouse/utils/utils.go
```diff
@@ -136,7 +137,7 @@ var (
 )
 
 func pseudoWarehouseDestinations() map[string]struct{} {
-	all := append(slices.Clone(WarehouseDestinations), SnowpipeStreaming)
+	all := append(slices.Clone(WarehouseDestinations), SnowpipeStreaming, BQStreamV2)
```
saikumarrs: Nit: Can we move the streaming destinations into an list variable of their own instead of an ever growing append call?

processor/internal/transformer/destination_transformer/embedded/warehouse/events.go
```diff
@@ -417,6 +417,7 @@ func (t *Transformer) usersResponse(tec *transformEventContext, commonData map[s
 
 func shouldSkipUsersTable(tec *transformEventContext) bool {
 	return tec.event.Metadata.DestinationType == whutils.SnowpipeStreaming ||
+		tec.event.Metadata.DestinationType == whutils.BQStreamV2 ||
```
saikumarrs: Once https://github.com/rudderlabs/rudder-server/pull/7056/changes#r3371350073 is addressed, this can also be refactored accordingly.

saikumarrs (commented):
Partial review 1

router/batchrouter/asyncdestinationmanager/bqstreamv2/types.go
```diff
@@ -0,0 +1,158 @@
...
+		uuidTS      string
+		rowID       any
+		receivedAt  any
+	}
+
+	Row map[string]any
+)
+
+// Decode decodes the destination configuration from the given map. Also, converts the namespace to the provider case.
+func (d *destConfig) Decode(m map[string]any) error {
```
saikumarrs: Nit: `types.go` has a few functions in it too (`Decode`, the timestamp setters). Can we pull those out and keep this file to type defs only?

router/batchrouter/asyncdestinationmanager/bqstreamv2/events.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/rudderlabs/rudder-server/utils/misc"
+	"github.com/rudderlabs/rudder-server/warehouse/slave"
+	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+)
+
+var (
+	idColumnName         = whutils.ToProviderCase(whutils.BQStreamV2, "id")
+	receivedAtColumnName = whutils.ToProviderCase(whutils.BQStreamV2, "received_at")
+	uuidTSColumnName     = whutils.ToProviderCase(whutils.BQ, "uuid_ts")
+	loadedAtColumnName   = whutils.ToProviderCase(whutils.BQ, "loaded_at")
```
saikumarrs: Why `BQ` for these two when `id`/`received_at` use `BQStreamV2`? It doesn't actually change anything today since `ToProviderCase` is a no-op outside Snowflake, but the mix is confusing — let's just use `BQStreamV2` for all four.

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
```diff
@@ -0,0 +1,73 @@
...
+// Has reports whether a non-expired schema exists for the table, without
+// cloning it.
+func (c *tableSchemaCacheImpl) Has(tableName string, now time.Time) bool {
+	c.mu.RLock()
+	item, ok := c.items[tableName]
+	c.mu.RUnlock()
+	return ok && !now.After(item.expiresAt)
+}
+
+// Peek returns the cached table schema regardless of expiry.
```
saikumarrs, line 36: How is `Peek` helpful?

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
```diff
@@ -0,0 +1,73 @@
...
+func (c *tableSchemaCacheImpl) Peek(tableName string) (whutils.ModelTableSchema, bool) {
+	c.mu.RLock()
+	item, ok := c.items[tableName]
+	c.mu.RUnlock()
```
saikumarrs, line 40: Let's `defer` the unlocks in `Get`/`Has`/`Peek` (and in `Set`/`Invalidate`) the way `Len` already does. Nothing's actually broken today since `Set` clones the schema instead of mutating it in place, but deferring is safer and keeps the whole file consistent.

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Len returns the number of items in the cache.
+func (c *tableSchemaCacheImpl) Len() int {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return len(c.items)
+}
+
+// cloneSchema clones the schema to avoid mutating the original schema.
+func cloneSchema(in whutils.ModelTableSchema) whutils.ModelTableSchema {
```
saikumarrs: `cloneSchema` is basically `maps.Clone(item.schema)` — can we just use that?

router/batchrouter/asyncdestinationmanager/bqstreamv2/schemautils.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	return newColumns
+}
+
+func checkAndIgnoreAlreadyExistError(err error) bool {
+	var e *googleapi.Error
+	if errors.As(err, &e) {
+		// 409 is returned when we try to create a table that already exists
+		// 400 is returned for all kinds of invalid input - so we need to check the error message too
+		if e.Code == 409 || (e.Code == 400 && strings.Contains(e.Message, "already exists in schema")) {
```
saikumarrs: Should this be case-insensitive? BQ's error message casing isn't guaranteed, so the `Contains` check could miss — something like `strings.Contains(strings.ToLower(e.Message), "already exists in schema")`.

router/batchrouter/asyncdestinationmanager/bqstreamv2/events.go
shekhar-rudder: These four column-name globals are split between two providers:

```go
idColumnName         = whutils.ToProviderCase(whutils.BQStreamV2, "id")
receivedAtColumnName = whutils.ToProviderCase(whutils.BQStreamV2, "received_at")
uuidTSColumnName     = whutils.ToProviderCase(whutils.BQ, "uuid_ts")
loadedAtColumnName   = whutils.ToProviderCase(whutils.BQ, "loaded_at")
```

`providerTableName := whutils.ToProviderCase(whutils.BQ, tableName)` in `groupAndChunkEvents` also uses `BQ`. `ToProviderCase` returns the input unchanged for both today, so output matches — but the split makes them silently divergent if BQSTREAM_V2 ever picks up different casing rules. Worth picking one provider type for all four (and the table name).

router/batchrouter/asyncdestinationmanager/bqstreamv2/events.go
```diff
@@ -0,0 +1,229 @@
...
+			currentValue := event.Message.Data[columnName]
+			convertedVal, err := slave.HandleSchemaChange(log, expectedType, actualType, currentValue)
+			if err != nil {
+				event.Message.Data[columnName] = nil // Discard value if conversion fails
+
+				rowID, idExists := event.Message.Data[idColumnName]
+				receivedAt, receivedAtExists := event.Message.Data[receivedAtColumnName]
+
+				if !idExists || !receivedAtExists {
+					continue
```
shekhar-rudder, line 181: When conversion fails but `id` or `received_at` is missing from `event.Message.Data`, the value is still nilled out a few lines above but the discard record is silently dropped. That's data being lost without anyone finding out. At minimum a `log.Warn` here would help operators spot events that are losing values without making it into the discard table.

router/batchrouter/asyncdestinationmanager/bqstreamv2/encode.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+		switch schema[col] {
+		case "datetime":
+			s, ok := v.(string)
+			if !ok {
+				continue
+			}
+			if ts, err := time.Parse(time.RFC3339Nano, s); err == nil {
+				row[col] = ts.UnixMicro()
+			}
```
shekhar-rudder, line 172: If the datetime string fails to parse, the value silently stays as a `string`. It then falls through to `protoValueFor` with field kind `Int64` (TIMESTAMP fields), eventually failing with `"invalid value of type string for INT64 field"` — a confusing message for what is actually a malformed datetime. Consider returning the parse error here so the user-facing error surfaces the real cause.

router/batchrouter/asyncdestinationmanager/bqstreamv2/encode.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, fmt.Errorf("unexpected descriptor type: %T", desc)
+	}
+	return md, nil
+}
+
+// streamWriterKey scopes cached writers to the destination's current project
+// and namespace, so a config revision cannot keep streaming to the old table.
+func streamWriterKey(cfg destConfig, tableName string) string {
+	return cfg.ProjectID + ":" + cfg.Namespace + ":" + tableName
+}
```
shekhar-rudder, line 56: The key is `project:namespace:table` — credentials aren't part of it. If a destination's service-account JSON is rotated/changed, the cached writer keeps using the old credentials until evicted by other means. The doc-comment says the key "scopes cached writers to the destination's current project and namespace, so a config revision cannot keep streaming to the old table" — but credentials are part of the config revision too. Worth confirming this is intentional (and maybe loosening the comment), or including a credential fingerprint in the key.

processor/internal/transformer/destination_transformer/embedded/warehouse/transformer_test.go
```diff
@@ -637,6 +637,8 @@ func setupConfig(resource *transformertest.Resource, configOverride map[string]a
 	c := config.New()
 	c.Set("DEST_TRANSFORM_URL", resource.TransformerURL)
 	c.Set("USER_TRANSFORM_URL", resource.TransformerURL)
+	c.Set("Processor.enableWarehouseTransformations", true)
+	c.Set("Processor.verifyWarehouseTransformations", false)
```
shekhar-rudder:
```go
c.Set("Processor.enableWarehouseTransformations", true)
c.Set("Processor.verifyWarehouseTransformations", false)
```

These keys back `GetReloadableBoolVar` declarations in `destination_transformer.go:96-97` whose defaults are `false` and `true` respectively — i.e., this helper flips both. If the intent is for tests in this file to run with embedded warehouse transformations enabled, flipping the production default in `destination_transformer.go` (one place) would be cleaner than pushing the override into every test setup. Also worth noting: `setupConfig` is shared across all cases here, not just BQSTREAM_V2 — these settings now apply to every pre-existing test using it. Was that intended?

router/batchrouter/asyncdestinationmanager/bqstreamv2/datatype.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"cloud.google.com/go/bigquery"
+	"github.com/samber/lo"
+
+	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+)
+
+// dataTypesMap maps datatype stored in rudder to datatype in bigquery
+var dataTypesMap = map[string]bigquery.FieldType{
+	"boolean":  bigquery.BooleanFieldType,
```
shekhar-rudder: Since BQSTREAM_V2 sits on top of the same BigQuery dialect, a lot of the building blocks here are copies of things that already exist in the BQ stack. Worth thinking about whether we can reuse rather than duplicate — every copy is a future drift risk. A couple of examples I noticed:

- **`dataTypesMap` / `toBigQuerySchema` (this file)** — `warehouse/integrations/bigquery/bigquery.go:65` already has an identical `dataTypesMap`, and `getTableSchema` at line 138 does the same job. The only meaningful difference is the sort (required because `adapt.StorageSchemaToProto2Descriptor` assigns proto field numbers by position). Either export `getTableSchema` and wrap it with a sort here, or move the sort upstream — BQ's `CreateTable` shouldn't care about field order.
- **Reserved-keyword JSON files** — the `BQSTREAM_V2` entries in `processor/.../reservedkeywords/{namespaces,tablescolumns}.json` are byte-for-byte identical to the existing `BQ` entries (95 keywords each, zero difference). An alias at load time (`ReservedKeywords[BQSTREAM_V2] = ReservedKeywords[BQ]`) would avoid two lists that need to be kept in sync by hand.

There are probably more (e.g. routing both `BQ` and `BQSTREAM_V2` through the same warehouse manager already hints at how aligned these dialects are). Worth a once-over before this lands.

shekhar-rudder (commented):
A few thoughts mostly around code reuse with the existing BQ stack and a handful of correctness/operability nits — nothing blocking.

router/batchrouter/asyncdestinationmanager/bqstreamv2/encode.go
```diff
@@ -0,0 +1,153 @@
...
+
+// normalizeRow converts datetime strings into the int64 epoch-micros
+// representation expected by TIMESTAMP fields.
+func normalizeRow(row Row, schema whutils.ModelTableSchema) {
+	for col, v := range row {
+		if v == nil {
+			continue
+		}
+		switch schema[col] {
+		case "datetime":
```
shekhar-rudder: Can we define a constant in datatype.go and use that here for `datetime`?

router/batchrouter/asyncdestinationmanager/bqstreamv2/schemautils.go
```diff
@@ -0,0 +1,47 @@
...
+	"strings"
+
+	"google.golang.org/api/googleapi"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	whutils "github.com/rudderlabs/rudder-server/warehouse/utils"
+)
+
+func findNewColumns(eventSchema, warehouseSchema whutils.ModelTableSchema) []whutils.ColumnInfo {
```
shekhar-rudder, line 14: Should this method also return columns where there is a datatype mismatch between event and warehouse?

router/batchrouter/asyncdestinationmanager/bqstreamv2/streamwriter.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("creating managedwriter client: %w", err)
+	}
+	defer func() {
+		if err != nil {
+			_ = client.Close()
+		}
+	}()
+
+	storageTableSchema, err := adapt.BQSchemaToStorageTableSchema(toBigQuerySchema(tableSchema))
```
shekhar-rudder: Can we use the `descriptorForSchema` method here ?

shekhar-rudder (commented):
Submitting pending review.

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
```diff
@@ -0,0 +1,73 @@
...
+	out := make(whutils.ModelTableSchema, len(in))
+	maps.Copy(out, in)
+	return out
+}
```
shekhar-rudder, line 69: `maps.Clone` (stdlib, Go 1.21+) does the same thing for a single-level map. Since `ModelTableSchema` is `map[string]string`, you can drop this helper entirely and inline `maps.Clone(…)` at the 4 call sites.

```suggestion
func cloneSchema(in whutils.ModelTableSchema) whutils.ModelTableSchema {
	return maps.Clone(in)
}
```

Or just replace every `cloneSchema(x)` call with `maps.Clone(x)` and delete this function.

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
```diff
@@ -0,0 +1,73 @@
...
+	delete(c.items, tableName)
+	c.mu.Unlock()
+}
+
+// Len returns the number of items in the cache.
+func (c *tableSchemaCacheImpl) Len() int {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return len(c.items)
+}
```
shekhar-rudder: Minor: every other method uses explicit `RUnlock()`/`Unlock()` after the map access, but `Len` uses `defer`. For consistency:

```suggestion
func (c *tableSchemaCacheImpl) Len() int {
	c.mu.RLock()
	n := len(c.items)
	c.mu.RUnlock()
	return n
}
```

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Has reports whether a non-expired schema exists for the table, without
+// cloning it.
+func (c *tableSchemaCacheImpl) Has(tableName string, now time.Time) bool {
+	c.mu.RLock()
+	item, ok := c.items[tableName]
+	c.mu.RUnlock()
+	return ok && !now.After(item.expiresAt)
+}
```
shekhar-rudder, line 34: `Has` duplicates the lookup + expiry logic from `Get`. Could simplify to:

```go
func (c *tableSchemaCacheImpl) Has(tableName string, now time.Time) bool {
	_, ok := c.Get(tableName, now)
	return ok
}
```

Trade-off: one unnecessary `maps.Clone` per call. If `Has` is on a hot path where avoiding the clone matters, keep as-is — otherwise the deduplication is worth it.

shekhar-rudder (commented):
Simplification suggestions for `tableschemacache.go`

router/batchrouter/asyncdestinationmanager/bqstreamv2/types.go
```diff
@@ -0,0 +1,158 @@
...
+func (m *Manager) setLoadedAtTimestamp(e *event, formattedTS string) bool {
+	if e.Message.Metadata.Columns == nil {
+		return false
+	}
+	if _, columnExists := e.Message.Metadata.Columns[loadedAtColumnName]; columnExists {
+		e.Message.Data[loadedAtColumnName] = formattedTS
+		return true
+	}
+	return false
+}
```
shekhar-rudder: `setUUIDTimestamp` and `setLoadedAtTimestamp` are identical apart from the column name. Collapse into a single helper:

```go
func setColumnTimestamp(e *event, columnName, formattedTS string) bool {
	if e.Message.Metadata.Columns == nil {
		return false
	}
	if _, ok := e.Message.Metadata.Columns[columnName]; ok {
		e.Message.Data[columnName] = formattedTS
		return true
	}
	return false
}
```

Also, neither method uses the `Manager` receiver — this can be a plain function (as shown above) rather than a method.

shekhar-rudder (commented):
Simplification suggestions for `types.go`

router/batchrouter/asyncdestinationmanager/bqstreamv2/datatype.go
achettyiitr: - Fixed Reserved-keyword JSON files.
- Fixed dataTypesMap.
- toBigQuerySchema requires special handling for sorting, so no need for it to use.

router/batchrouter/asyncdestinationmanager/bqstreamv2/encode.go
achettyiitr: Good point. 
Checked the codebase, we don't need streamWriterKey, as when the config gets updated, we create a new async destination manager. Removing it.

router/batchrouter/asyncdestinationmanager/bqstreamv2/encode.go
achettyiitr: Good point.

router/batchrouter/asyncdestinationmanager/bqstreamv2/events.go
achettyiitr: Fixed.

router/batchrouter/asyncdestinationmanager/bqstreamv2/schemautils.go
achettyiitr: Not needed. Because we are not handling Alters.

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
achettyiitr: Get doesn't return anything if it has expired. Peek is to get the current value so that we can compare these things when fetching the schema.

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
achettyiitr: Not needed. Don't need to use Get and then clone the value, unnecessary, and ignore the value. When using a lock, it is rudimentary to have at least these lines.

```
   c.mu.RLock()
	item, ok := c.items[tableName]
	c.mu.RUnlock()
```

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
achettyiitr: Not needed. Defering it will take the lock for a longer time.

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
achettyiitr: Not needed. Every other function depends on the value, but here we only require the length.

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
saikumarrs: Why would you want to use a expired schema though?

router/batchrouter/asyncdestinationmanager/bqstreamv2/tableschemacache.go
saikumarrs: The point here is to avoid data inconsistencies when rapid cache operations take place.
Is the extra delay significant?

achettyiitr: Yeah, because we are cloning the schema as well and doing some computation.

saikumarrs: Makes sense. Let's unlock right before those time consuming operations then.

saikumarrs (approved):

router/batchrouter/asyncdestinationmanager/bqstreamv2/events.go
```diff
@@ -0,0 +1,243 @@
...
+					rowID:       rowID,
+					receivedAt:  receivedAt,
+				})
+			} else {
+				// Update value if conversion succeeds
+				event.Message.Data[columnName] = convertedVal
+			}
+		}
+		if reflect.TypeOf(event.Message.Data[columnName]) == sliceOfAnyType {
+			marshalledVal, err := jsonrs.Marshal(event.Message.Data[columnName])
```
shekhar-rudder, line 199: This `[]any` marshalling branch is outside the `if expectedType, exists := warehouseSchema[columnName]; exists && actualType != expectedType` block, so it fires for *every* column the event declares — including columns that aren't yet in `warehouseSchema` (new columns) and columns whose types already match. For new columns, slice values get JSON-stringified before the column exists in the target table, which can then surface as a type error from the Storage Write API once the column is created with a non-string type. Intentional, or should this be scoped to the mismatch branch (or to columns whose warehouse type is `string`)?

router/batchrouter/asyncdestinationmanager/bqstreamv2/events.go
```diff
@@ -0,0 +1,243 @@
...
+		providerTableName := whutils.ToProviderCase(whutils.BQStreamV2, tableName)
+
+		var currentChunkBytes int64
+		var currentChunk []*event
+		flush := func() {
+			if len(currentChunk) == 0 {
+				return
+			}
+			groupedAndChunkedEvents[tableName] = append(groupedAndChunkedEvents[tableName], tableEvents{
+				tableName: providerTableName,
```
shekhar-rudder, line 103: Table-name inconsistency: the outer map key here is the raw event table name (`tableName` from the `lo.GroupBy` above), while `tableEvents.tableName` just below is `providerTableName` (`whutils.ToProviderCase(...)`). For BQ these happen to be identical today (ToProviderCase is a no-op for BQ), but the divergence is a footgun — a caller iterating `range groupedAndChunkedEvents` and using the key as a table identifier will get the raw name, while one reading values gets the provider-cased one. Worth picking one and using it for both.

shekhar-rudder (commented):
Two comments — both on `events.go`. One on a likely-unintentional scope of the `[]any` marshalling branch in `getDiscardedRecordsFromEvent`, and one on a raw-vs-provider-cased table name divergence in `groupAndChunkEvents`.

router/batchrouter/asyncdestinationmanager/bqstreamv2/encode.go
```diff
@@ -0,0 +1,177 @@
...
+				return protoreflect.ValueOfFloat64(parsed), nil
+			}
+		case string:
+			if parsed, err := strconv.ParseFloat(v, 64); err == nil {
+				return protoreflect.ValueOfFloat64(parsed), nil
+			}
+		}
+	}
+	return protoreflect.Value{}, fmt.Errorf("invalid value of type %T for %s field", value, fd.Kind())
+}
```
shekhar-rudder, line 154: The error message uses `fd.Kind()` (proto field kind), so for TIMESTAMP/DATETIME columns this reads as `"invalid value of type X for INT64 field"` — confusing from a user perspective, since they configured a `datetime` column, not INT64. Worth threading the warehouse data type (e.g. `schema[col]`) into the error so the surfaced message matches the configured type.

router/batchrouter/asyncdestinationmanager/bqstreamv2/encode.go
```diff
@@ -0,0 +1,177 @@
...
+				continue
+			}
+			fd, ok := fieldsByName[columnName]
+			if !ok {
+				return nil, fmt.Errorf("encoding row: unknown column %q", columnName)
+			}
+			fieldValue, err := protoValueFor(fd, value)
+			if err != nil {
+				return nil, fmt.Errorf("encoding row: column %q: %w", columnName, err)
+			}
```
shekhar-rudder, line 89: These two error returns (unknown column / unencodable value) abort the entire `encodeRows` batch on the first bad row. With Storage Write API batching, that means one malformed event kills the whole AppendRows call (and the whole table chunk). Was per-row skip (with the bad row routed to discards) considered, or is the assumption that `getDiscardedRecordsFromEvent` upstream guarantees nothing unencodable reaches here?

router/batchrouter/asyncdestinationmanager/bqstreamv2/streamwriter.go
```diff
@@ -0,0 +1,78 @@
...
+		return nil, fmt.Errorf("creating managedwriter client: %w", err)
+	}
+	defer func() {
+		if err != nil {
+			_ = client.Close()
+		}
+	}()
+
+	md, err := descriptorForSchema(tableSchema)
+	if err != nil {
```
shekhar-rudder, line 47: `descriptorForSchema` and `adapt.NormalizeDescriptor` are pure CPU work but happen *after* `managedwriter.NewClient` — which is why the deferred close-on-error dance above is needed. Moving the descriptor build above the client creation would let you drop the defer entirely: if descriptor building fails, no client was ever created, so there's nothing to clean up.

shekhar-rudder (commented):
Three more — confusing error message in `protoValueFor`, blast radius of a single bad row in `encodeRows`, and a small reorder in `NewStreamWriter` that removes the need for the deferred close-on-error pattern.

router/batchrouter/asyncdestinationmanager/bqstreamv2/streamwriter.go
achettyiitr: It's still needed for the NewManagedStream creation.

router/batchrouter/asyncdestinationmanager/bqstreamv2/events.go
achettyiitr: This needs to be handled for every piece of data. Currently, during DT, array values are sent as it is, so during post-processing, we convert them to JSON data here. This we do it in warehouse slaves as well for warehouse destinations.

router/batchrouter/asyncdestinationmanager/bqstreamv2/encode.go
achettyiitr: Yes, getDiscardedRecordsFromEvent takes care of it.

shekhar-rudder (approved):

PR #7057, opened 2026-06-05
feat: add and register bq stream v2 async destination manager

achettyiitr:
# Description
 
PR 2/3 of the BQSTREAM_V2 stack (splits #7019). Stacked on #<PR1>.
 
Implements the Manager upload orchestration on top of the building blocks from PR 1, and wires it up:
 
**Upload orchestration**
- Transform using `common.GetMarshalledData(string(job.EventPayload), job.JobID)`.
- Upload flow: reads the async file, groups and chunks events (helpers from PR 1), and processes tables with bounded concurrency.
- Schema lifecycle via the warehouse BigQuery manager (`FetchSchema`, `CreateSchema`, `CreateTable`, `AddColumns`), including dataset-not-found create-and-retry handling and schema caching.
- Discard-path writes to `rudder_discards`; appends encoded rows through cached per-table Storage Write API writers.
- Append error classification into failed vs aborted job IDs; schema cache/writer invalidation on failures.
- Unit tests for the full upload flow (31 scenarios).
 
**Registration**
- Registered `BQSTREAM_V2` in the async destination manager factory via `common.SimpleAsyncDestinationManager`.
- Added `BQSTREAM_V2` to the async destinations whitelist and `misc.BatchDestinations()`.
- Registration unit test and README entry.
 
This PR turns the feature on for configured destinations. ~65% of the diff is unit tests.
 
**Stack:** PR 1 (enablement + building blocks) → PR 2 (this) → PR 3 (integration tests)
 
## Linear Ticket
 
- Resolves INT-6410
 
## Security
 
## Validation
 
- `go test ./router/batchrouter/asyncdestinationmanager/bqstreamv2/...`
- `go test ./router/batchrouter -run Async -count=1`


router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
```diff
@@ -0,0 +1,644 @@
...
+
+			if failedJobError != "" {
+				m.logger.Warnn("Failed to process table",
+					logger.NewStringField("namespace", cfg.Namespace),
+					logger.NewStringField("table", tableName),
+					obskit.Error(errors.New(failedJobError)),
+				)
+
+				if shouldAbort(errors.New(failedJobError)) {
+					aborted = append(aborted, failedJobIDs...)
```
shekhar-rudder, line 248: **Per-table abort classification is dead code.** `failedJobError` is a `string` (set from `result.failedError.Error()`), and `errors.New(failedJobError)` produces a plain error with no gRPC status. `shouldAbort` uses `status.Code(err)`, which returns `codes.Unknown` for a plain error, so this branch always falls through to `failed`.

The effect: a genuinely abortable error in the table path (e.g. `CreateTable`/`AddColumns`/`AppendRows` returning `PermissionDenied`, `FailedPrecondition`, etc.) is classified as retryable and the jobs retry indefinitely instead of aborting — unlike the top-level paths (`refreshSchemaCacheIfNeeded`, discards-table creation) which pass the real wrapped error to `shouldAbort` and classify correctly.

Consider carrying the actual `error` through `tableProcessResult`/`failedJobResult` (instead of stringifying it) and calling `shouldAbort` on it. Also worth noting there's no test covering an abortable error in the events-table path, which is why this slips through.

router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
```diff
@@ -0,0 +1,644 @@
...
+			logger.NewStringField("table", tableName),
+		)
+
+		err := m.createTableSchema(ctx, cfg, tableName, eventsSchema)
+		if err != nil {
+			if !checkAndIgnoreAlreadyExistError(err) {
+				m.logger.Infon("Table schema already exists", logger.NewStringField("table", tableName))
+
+				return fmt.Errorf("table schema already exists: %w", err)
+			}
```
shekhar-rudder: The `"Table schema already exists"` message is incorrect: it's gated on the **negation** of `checkAndIgnoreAlreadyExistError(err)`, so it runs for genuine, non-already-exists creation failures — and that misleading prefix gets wrapped into the error and lands in the persisted `FailedReason`. The check looks inverted; intended is:

```go
if checkAndIgnoreAlreadyExistError(err) {
    m.logger.Infon("Table schema already exists", logger.NewStringField("table", tableName))
    return nil
}
return fmt.Errorf("creating table schema: %w", err)
```

router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
```diff
@@ -0,0 +1,644 @@
...
+		logger: log.Child("bqstreamv2").Withn(
+			obskit.WorkspaceID(destination.WorkspaceID),
+			obskit.DestinationID(destination.ID),
+			obskit.DestinationType(destination.DestinationDefinition.Name),
+			logger.NewStringField("id", uuid.New().String()),
+		),
+		statsFactory:  statsFactory,
+		destination:   destination,
+		streamWriters: make(map[string]*tableStreamWriter),
+		streamWriterFactory: &streamWriterFactoryImpl{
```
shekhar-rudder, line 78: **Cached stream writers leak — `Manager` has no `Close`/`Cleanup`.** The `streamWriters` map holds open Storage Write API streams (and BigQuery clients) that, by design, outlive a single upload (`writerForTable` uses `context.WithoutCancel`). They're only ever released via `invalidateTableCacheAndStreamWriter` (schema change / append error).

But there's no teardown: in `handle_lifecycle.go`, `refreshDestination` → `initAsyncDestinationStruct` builds a **new** `Manager` and overwrites the cached one on every `RevisionID` change, and `Shutdown()` drops it — neither closes the old manager's writers. So every destination config revision (common in multi-tenant) and every shutdown leaks the open streams/connections.

Suggest adding a `Close()` that drains and closes all cached writers, and wiring it into the refresh/teardown path (the interface gap in `common.AsyncDestinationManager` is broader than this PR, but at minimum the leak should be addressed here).

router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
```diff
@@ -0,0 +1,644 @@
...
+					failed = append(failed, failedJobIDs...)
+					failedReason = failedJobError
+				}
+			}
+			return nil
+		})
+	}
+
+	if err := tableErrgroup.Wait(); err != nil {
+		m.logger.Warnn("Failed to process tables", obskit.Error(err))
```
shekhar-rudder: Minor: every `tableErrgroup.Go` callback returns `nil` (errors are handled inside and stored in the per-table result), so `tableErrgroup.Wait()` never returns a non-nil error — this `if err != nil` block is unreachable.

router/batchrouter/asyncdestinationmanager/manager.go
```diff
@@ -47,6 +48,8 @@ func newRegularManager(
 		return snowpipestreaming.New(conf, logger, statsFactory, destination), nil
 	case "SALESFORCE_BULK_UPLOAD":
 		return salesforcebulkupload.NewManager(conf, logger, statsFactory, destination, backendConfig)
+	case "BQSTREAM_V2":
+		return common.SimpleAsyncDestinationManager{UploaderAndTransformer: bqstreamv2.NewManager(conf, logger, statsFactory, destination)}, nil
 	}
```
shekhar-rudder, line 53: SFTP (`sftp/manager.go`) and Lytics (`lytics_bulk_upload/manager.go`) also use `SimpleAsyncDestinationManager`, but they do the wrapping *inside their own package's `NewManager`* and return a fully-formed `common.AsyncDestinationManager`. For consistency with that convention (every other case here returns a ready manager from its own package), consider moving the `SimpleAsyncDestinationManager` wrapping into `bqstreamv2.NewManager`.

router/batchrouter/asyncdestinationmanager/README.md
```diff
@@ -84,6 +84,7 @@ Destinations that require complex polling and status checking:
 - Klaviyo Bulk Upload
 - Lytics Bulk Upload
 - Snowpipe Streaming
+- BigQuery Streaming
```
shekhar-rudder, line 86: This adds "BigQuery Streaming" under *"Destinations that require complex polling and status checking"*, but BQSTREAM_V2 is registered via `SimpleAsyncDestinationManager` with no-op (synchronous) `Poll`/`GetUploadStats`. It conceptually belongs in the *"Simple Async Destinations"* section (line ~94) instead.

utils/misc/misc.go
```diff
@@ -102,7 +102,7 @@ func Init() {
 }
 
 func BatchDestinations() []string {
-	batchDestinations := []string{"S3", "GCS", "MINIO", "RS", "BQ", "AZURE_BLOB", "SNOWFLAKE", "POSTGRES", "CLICKHOUSE", "DIGITAL_OCEAN_SPACES", "MSSQL", "AZURE_SYNAPSE", "S3_DATALAKE", "MARKETO_BULK_UPLOAD", "GCS_DATALAKE", "AZURE_DATALAKE", "DELTALAKE", "BINGADS_AUDIENCE", "ELOQUA", "YANDEX_METRICA_OFFLINE_EVENTS", "SFTP", "BINGADS_OFFLINE_CONVERSIONS", "KLAVIYO_BULK_UPLOAD", "LYTICS_BULK_UPLOAD", "SNOWPIPE_STREAMING", "SALESFORCE_BULK_UPLOAD"}
+	batchDestinations := []string{"S3", "GCS", "MINIO", "RS", "BQ", "AZURE_BLOB", "SNOWFLAKE", "POSTGRES", "CLICKHOUSE", "DIGITAL_OCEAN_SPACES", "MSSQL", "AZURE_SYNAPSE", "S3_DATALAKE", "MARKETO_BULK_UPLOAD", "GCS_DATALAKE", "AZURE_DATALAKE", "DELTALAKE", "BINGADS_AUDIENCE", "ELOQUA", "YANDEX_METRICA_OFFLINE_EVENTS", "SFTP", "BINGADS_OFFLINE_CONVERSIONS", "KLAVIYO_BULK_UPLOAD", "LYTICS_BULK_UPLOAD", "SNOWPIPE_STREAMING", "SALESFORCE_BULK_UPLOAD", "BQSTREAM_V2"}
 	return batchDestinations
```
shekhar-rudder, line 106: Instead of duplicating all the async/sftp destination names here, we could build this from the existing lists — e.g. append `common.asyncDestinations` + `common.sftpDestinations` to the warehouse-specific entries — so a new async destination only needs to be added in one place.

router/batchrouter/asyncdestinationmanager/bqstreamv2/integration.go
```diff
@@ -0,0 +1,122 @@
...
+	return bigQueryManager, nil
+}
+
+// fetchSchemaFromWarehouse fetches the namespace schema and filters it down
+// to the given tables.
+func (m *Manager) fetchSchemaFromWarehouse(ctx context.Context, cfg destConfig, tableNames []string) (whutils.ModelSchema, error) {
+	m.logger.Infon("Fetching schema from warehouse")
+
+	bigQueryManager, err := m.integrationManagerCreator(ctx, cfg)
+	if err != nil {
```
shekhar-rudder, line 21: All four methods in this file create a fresh manager via `integrationManagerCreator` → `manager.New` + `Setup`, and `Setup` dials a brand-new BigQuery client (`bigquery.NewClient`). On top of that, the deferred `Cleanup` runs a `dropDanglingStagingTables` query and `Close()` on every call. So a single `Upload` touching N tables creates ~N+ clients and runs ~N+ dangling-staging-table queries (`fetchSchema`, optional `createSchema`, discards create/add, and one per events table — concurrently across `tableWorkers`).

Consider creating the manager once in `Upload` (before the errgroup), `defer Cleanup` once, and injecting it into `refreshSchemaCacheIfNeeded` / `createTableAndAddColumnsIfNeeded` / the per-table ops. The `cloud.google.com/go/bigquery` client is safe for concurrent use, so a single instance can serve the concurrent table workers.

router/batchrouter/asyncdestinationmanager/bqstreamv2/integration.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil, fmt.Errorf("setting up bigquery manager: %w", err)
+	}
+	return bigQueryManager, nil
+}
+
+// fetchSchemaFromWarehouse fetches the namespace schema and filters it down
+// to the given tables.
+func (m *Manager) fetchSchemaFromWarehouse(ctx context.Context, cfg destConfig, tableNames []string) (whutils.ModelSchema, error) {
+	m.logger.Infon("Fetching schema from warehouse")
```
shekhar-rudder, line 18: Naming: this fetches the full namespace schema and then filters it down to the requested tables, but the name only conveys the fetch. Since the filtering (returning schema only for the tables we asked about) is significant for callers, a name like `fetchSchemaForTables` or `fetchExistingTableSchemas` would reflect both steps.

shekhar-rudder (commented):
Independent code review focused on correctness, concurrency, and the registration wiring, plus a few notes raised during review. The concurrency design in `bqstreamv2.go` is carefully done; the findings below are a mix of one correctness bug (per-table abort classification), a misleading error message, a stream-writer lifecycle leak, and several maintainability/consistency suggestions. Review-only — no code changes.

router/batchrouter/asyncdestinationmanager/bqstreamv2/integration.go
achettyiitr: Renaming to `fetchSchemaFromWarehouseForTables`.

utils/misc/misc.go
achettyiitr: Good point. Same things are getting replicated accross multiple places.

achettyiitr: It's creating a circular dependency. For now keeping it as it is.
```
# github.com/rudderlabs/rudder-server/router/batchrouter/asyncdestinationmanager/bqstreamv2
package github.com/rudderlabs/rudder-server/router/batchrouter/asyncdestinationmanager/bqstreamv2
	imports github.com/rudderlabs/rudder-server/backend-config from bqstreamv2.go
	imports github.com/rudderlabs/rudder-server/backend-config/internal/cache from backend-config.go
	imports github.com/rudderlabs/rudder-server/utils/misc from cache.go
	imports github.com/rudderlabs/rudder-server/router/batchrouter/asyncdestinationmanager/common from misc.go
	imports github.com/rudderlabs/rudder-server/jobsdb from common.go
	imports github.com/rudderlabs/rudder-server/jobsdb/internal/cache from jobsdb.go
	imports github.com/rudderlabs/rudder-server/utils/misc from cache.go: import cycle not allowed
```

router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
achettyiitr: Added as a comment.

router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
```diff
@@ -0,0 +1,626 @@
...
+
+	tableStreamWriter, err := m.streamWriterFactory.NewTableStreamWriter(context.WithoutCancel(ctx), cfg, tableName, schema)
+	if err != nil {
+		return nil, fmt.Errorf("creating table stream writer for table: %w", err)
+	}
+
+	m.logger.Infon("Created table stream writer for table", logger.NewStringField("table", tableName))
+
+	m.streamWriters[tableName] = tableStreamWriter
+
```
shekhar-rudder, line 597: The doc comment above says *"A second lookup after creation keeps the winner when two workers race to build the same writer (the shared discards stream), discarding the loser's duplicate."* — but there's no second lookup here. After the `RLock` miss, both racing workers acquire the write lock in turn and each calls `NewTableStreamWriter`, opening a second gRPC client + managed stream; the second store overwrites the first in the map, so the first writer is never tracked and never `Close()`d — a leaked connection/stream.

Since `processTable` → `streamDiscardedEvents` → `writerForTable(discardsTableName)` runs concurrently across table workers, the shared discards stream is exactly the case the comment describes. Re-check the map after taking the write lock and return the existing writer (closing/discarding the loser) — otherwise this is a real resource leak, not just a stale comment.

router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
```diff
@@ -0,0 +1,626 @@
...
+		m.logger.Infon("No table schema found in cache",
+			logger.NewStringField("namespace", cfg.Namespace),
+			logger.NewStringField("table", tableName),
+		)
+
+		err := m.createTableSchema(ctx, cfg, integrationManagerCreator, tableName, eventsSchema)
+		if err != nil {
+			if !checkAndIgnoreAlreadyExistError(err) {
+				m.logger.Infon("Table schema already exists", logger.NewStringField("table", tableName))
+
```
shekhar-rudder: Misleading log: this fires inside the `!checkAndIgnoreAlreadyExistError(err)` branch — i.e. exactly when the error is *not* an already-exists error and a genuine create failure is about to be returned. The message `"Table schema already exists"` describes the opposite of what happened. Looks like it belongs in the ignored-error path, or should just be removed.

router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
```diff
@@ -0,0 +1,626 @@
...
+	}
+	return response
+}
+
+// streamEventBatches appends each chunk through the table's cached stream
+// writer and reports outcomes per chunk: an acknowledged append is already
+// durable in BigQuery, so only the failed chunks' jobs are retried instead of
+// re-appending (and duplicating) the chunks that landed. Appends are pipelined
+// (fire all, then wait), and the writer and schema cache are evicted on any
+// failure so a retry rebuilds them against the current table schema.
```
shekhar-rudder: This comment claims appends are pipelined so "N batches cost ~one round trip", but `tableStreamWriter.AppendRows` now calls `appendResult.GetResult(ctx)` inline (streamwriter.go), so it blocks per batch. The loop below appends and waits on each chunk sequentially → N round trips, not one. Either the comment is stale from a pre-wrapper version, or the pipelining was lost in the refactor — worth reconciling (the same wording also appears on `streamDiscardedEvents`).

shekhar-rudder (commented):
Review of the BQSTREAM_V2 async manager. One likely resource leak (racing discards stream writer), one misleading log message, and one stale doc comment about pipelining. Test coverage looks thorough.

router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
achettyiitr: Log is not needed, modified the code.

```
		if err := m.createTableSchema(ctx, cfg, integrationManagerCreator, tableName, eventsSchema); err != nil && !checkAndIgnoreAlreadyExistError(err) {
			return fmt.Errorf("creating table schema: %w", err)
		}
```

router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
achettyiitr: Fixed it.

router/batchrouter/asyncdestinationmanager/bqstreamv2/bqstreamv2.go
achettyiitr: Stale comment now. Fixing it.

shekhar-rudder (approved):

saikumarrs (approved):

PR #7066, opened 2026-06-08
fix(jobsdb): jobsdb_lock_total_time for dsMigrationLock missing

atzoum:
## Description

`TryLockWithCtx` was not recording `jobsdb_lock_time` or `jobsdb_lock_total_time` after the lock was released via `Unlock()`. This meant the `dsMigrationLock` (which uses the `TryLockWithCtx` + `Unlock` pattern) was missing held-time metrics.

**Fix:** store the lock acquisition timestamp on the `Locker` struct when `TryLockWithCtx` succeeds, and drain it in `Unlock()` to record both timers. Since only one write lock can be held at a time, a single pair of fields on the struct is sufficient.

## Security

### DIFF
--- a/jobsdb/internal/lock/lock.go
+++ b/jobsdb/internal/lock/lock.go
@@ -26,6 +26,10 @@ type Locker struct {
 	writeLockTimeAsync      stats.Measurement
 	writeLockTotalTime      stats.Measurement
 	writeLockTotalTimeAsync stats.Measurement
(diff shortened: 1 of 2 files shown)


itsmihir (approved):

mihir20 (approved):
LGTM

ktgowtham (approved):

PR #7069, opened 2026-06-08
fix(gateway/webhook): prevent nil-channel send DoS in RequestHandler

### LINKED ISSUE #6999: bug: nil channel send in webhook RequestHandler blocks HTTP handler forever (DoS)
**Describe the bug**

In `gateway/webhook/webhook.go`, the `RequestHandler` looks up the per-source request channel from a map, then sends to it. If the source type has not been registered yet, the map lookup returns a nil channel, and sending to a nil channel blocks **forever** — the HTTP handler goroutine never returns and the response is never written.

This is a denial-of-service vulnerability: a single webhook request for an unregistered source type permanently consumes a goroutine and leaves the client hanging.

**Code reference**

`gateway/webhook/webhook.go:233-238`:
```go
webhook.requestQMu.RLock()
requestQ := webhook.requestQ[sourceDefName]   // nil if sourceDefName not in map
requestQ <- &req                                // BLOCKS FOREVER on nil channel!
webhook.requestQMu.RUnlock()
```

This happens when `webhookV2HandlerEnabled` is `false` (default in `setup.go:80`). In this mode, source types are registered lazily via `processBackendConfig` (`handle_lifecycle.go:350-353`) when backend config is received. If a webhook request arrives:

1. Before backend config is processed (startup race)
2. For a source type not present in the backend config

…the map lookup returns `nil`.

**Steps to reproduce**

1. Configure a webhook source but delay backend config delivery
2. Send a POST request to the webhook endpoint
3. The request never returns — the handler is stuck on `requestQ <- &req` (nil channel)

**Expected behavior**

The handler should verify the channel exists before sending:

```go
requestQ, ok := webhook.requestQ[sourceDefName]
if !ok {
    webhook.failRequest(w, r, "invalid webhook source", http.StatusNotFound)
    webhook.ackCount.Add(1)
    return
}
requestQ <- &req
```

Or alternatively, `Register` should be called unconditionally before the map lookup (removing the `if webhook.config.webhookV2HandlerEnabled` guard at line 233).

kuntal1461:
## Summary

Fixes #6999 — sending on a nil channel in `RequestHandler` blocks the HTTP worker goroutine forever, enabling a denial-of-service condition.

**Root cause:** `webhook.requestQ[sourceDefName]` returns `nil` when the source has not been registered (e.g. `webhookV2HandlerEnabled=false` and no prior `Register` call). Go's nil channel send blocks indefinitely.

**Changes:**

- **Nil-channel guard** — check `requestQ == nil` after the map lookup and return HTTP 503 `service unavailable` immediately instead of blocking.
- **Shutdown-aware enqueue** — wrap the channel send in a `select` with `r.Context().Done()` (client disconnect / deadline) and `webhook.ctx.Done()` (server shutdown). The RLock is held across the select to prevent a concurrent `Shutdown` from closing the channel mid-send.
- **Buffered `done` channel** (size 1) — if `RequestHandler` exits early due to context cancellation, the batcher goroutine can still complete its send without leaking.
- **Context guard on `resp := <-done`** — also wraps the response-wait in a `select` with `r.Context().Done()` so a slow transformer doesn't permanently hold a worker after the client has gone.
- **Store shutdown context** — `HandleT.ctx` added and populated in `Setup` so `RequestHandler` can reference it.

## Test plan

- [x] `TestWebhookRequestHandlerSourceNotRegistered` — V1 mode, no `Register` call → asserts HTTP 503 returned immediately, no block
- [x] `TestWebhookRequestHandlerRequestContextCancellation` — blocking adapter + 100 ms request deadline → asserts handler returns promptly
- [x] All existing webhook unit tests pass (`go test ./gateway/webhook/... -race -count=1`)
- [ ] Manual: run with `webhookV2HandlerEnabled=false` and send a request for an unregistered source — expect immediate 503

### DIFF
--- a/gateway/webhook/setup.go
+++ b/gateway/webhook/setup.go
@@ -119,6 +119,7 @@ func Setup(gwHandle Gateway, transformerFeaturesService TransformerFeaturesServi
 	)
 	ctx, cancel := context.WithCancel(context.Background())
 	webhook.backgroundCancel = cancel
+	webhook.ctx = ctx
 
 	webhook.statReporterCreator = statReporterCreator
 
--- a/gateway/webhook/webhook.go
+++ b/gateway/webhook/webhook.go
@@ -69,6 +69,7 @@ type HandleT struct {
 	batchRequestsWg  sync.WaitGroup
 	backgroundWait   func() error
 	backgroundCancel context.CancelFunc
+	ctx              context.Context // cancelled on Shutdown, before channels are closed
 
 	config struct {
 		maxReqSize                 config.ValueLoader[int]
@@ -212,7 +213,7 @@ func (webhook *HandleT) RequestHandler(w http.ResponseWriter, r *http.Request) {
 		r.Header.Set("Content-Type", "application/json")
 	}
 
-	done := make(chan transformerResponse)
+	done := make(chan transformerResponse, 1)
 	req := webhookT{
 		request:     r,
 		writer:      w,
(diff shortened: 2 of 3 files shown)


kuntal1461:
hi 
any update

PR #7074, opened 2026-06-09
fix(batchrouter): correlate poll results by externalId instead of whole-row hash

ItsSudip:

# Description

## Problem
The Salesforce Bulk Upload connector correlated async poll results back to RudderStack jobs using a SHA-256 hash of the **entire CSV row**. Salesforce coerces field values on store — e.g. datetime microsecond→millisecond truncation (`2025-11-25T03:45:21.14287Z` → `…142Z`) and number-scale rounding — so the rows returned in `successfulResults` / `failedResults` are not byte-identical to what was sent, and the hash no longer matches.

This only surfaces on **partially-failed imports** (`NumberRecordsFailed > 0`); fully-successful imports are marked succeeded without any result matching. In a partially-failed import, the successfully-upserted records couldn't be matched, fell into the reconciliation path, were marked retryable (`FailedKeys`), retried, and — because the mismatch is deterministic — eventually **aborted despite landing in Salesforce**. This drove a 100%-failure alert for a customer (INT-6506); the surfaced error was literally the reconciliation reason string *"Input hash is not found in the data hash to job ID map…"*.

## Changes
- **Correlate on the externalId field value** (the upsert key, which round-trips reliably) instead of the whole-row hash — removes sensitivity to Salesforce coercion of every other column. The externalId field name is now persisted in the import parameters (`SalesforceJobInfo.ExternalIDField`, replacing the now-unused `Headers`) so it's available at poll/stats time.
- **Abort deterministic mismatches instead of retrying.** On a genuine match miss the job is aborted rather than marked failed — retrying just re-uploads the same value, triggers the same coercion, and misses again. **Guard:** if the in-memory correlation map is empty (e.g. a process restart between `Upload` and poll), jobs stay retryable so a restart can't mass-abort an in-flight batch.
- **Validation / observability:** fail the upload fast if the externalId field is empty; warn-log when duplicate externalIds collide within a batch.
- Simplify `calculateHashCode` to take a `string`.
- **Tests** updated + added: externalId-based matching (succeeded/aborted), abort-on-mismatch with a populated map, and the empty-map → retry guard.

## Linear Ticket
https://linear.app/rudderstack/issue/INT-6506

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
+++ b/router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
@@ -70,10 +70,11 @@ func NewUploader(
 		statsFactory:      statsFactory,
 		destName:          destName,
 	}
-	statTags := stats.Tags{"module": "batch_router", "destType": destName}
-	if destination != nil {
-		statTags["destinationId"] = destination.ID
-		statTags["workspaceId"] = destination.WorkspaceID
+	statTags := stats.Tags{
+		"module":        "batch_router",
+		"destType":      destName,
+		"destinationId": destination.ID,
+		"workspaceId":   destination.WorkspaceID,
 	}
 	u.payloadSizeStat = statsFactory.NewTaggedStat("payload_size", stats.HistogramType, statTags)
 	u.eventsPerFileStat = statsFactory.NewTaggedStat("events_per_file", stats.HistogramType, statTags)
@@ -172,26 +173,23 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 	destinationID := asyncDestStruct.Destination.ID
 	filePath := asyncDestStruct.FileName
 
-	input, err := s.readJobsFromFile(filePath)
+	jobs, err := s.readJobsFromFile(filePath)
 	if err != nil {
 		return s.failedJobs(asyncDestStruct, fmt.Sprintf("reading jobs from file: %v", err))
 	}
 
-	objectInfo, err := extractObjectInfo(input)
+	objectInfo, err := extractObjectInfo(jobs)
 	if err != nil {
 		return s.failedJobs(asyncDestStruct, fmt.Sprintf("extracting object info: %v", err))
-	}
-	if objectInfo.ExternalIDField == "" {
-		return s.failedJobs(asyncDestStruct, externalIDFieldEmptyReason)
 	}
 
 	// Drop events without an externalId value up front: with no upsert key they
 	// cannot be sent to Salesforce, so abort them with an error rather than
 	// emitting an uncorrelatable CSV row. Valid events continue to the upload.
-	validJobs := make([]SalesforceAsyncJob, 0, len(input))
+	validJobs := make([]SalesforceAsyncJob, 0, len(jobs))
 	var abortedJobIDs []int64
-	for _, job := range input {
-		if externalIDValue, _ := common.FormatCSVValue(job.Message[objectInfo.ExternalIDField]); externalIDValue == "" {
+	for _, job := range jobs {
+		if len(job.Metadata.ExternalID) == 0 || job.Metadata.ExternalID[0].ID == "" {
 			abortedJobIDs = append(abortedJobIDs, job.Metadata.JobID)
 			continue
 		}
--- a/router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
+++ b/router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
@@ -76,9 +76,9 @@ func NewUploader(
 		"destinationId": destination.ID,
 		"workspaceId":   destination.WorkspaceID,
 	}
-	u.payloadSizeStat = statsFactory.NewTaggedStat("payload_size", stats.HistogramType, statTags)
-	u.eventsPerFileStat = statsFactory.NewTaggedStat("events_per_file", stats.HistogramType, statTags)
-	u.asyncUploadTimeStat = statsFactory.NewTaggedStat("async_upload_time", stats.TimerType, statTags)
+	u.payloadSizeStat = statsFactory.NewTaggedStat("brt_async_dest_payload_size", stats.HistogramType, statTags)
+	u.eventsPerFileStat = statsFactory.NewTaggedStat("brt_async_dest_events_per_file", stats.HistogramType, statTags)
+	u.asyncUploadTimeStat = statsFactory.NewTaggedStat("brt_async_dest_async_upload_time", stats.TimerType, statTags)
 	u.config.maxBufferCapacity = conf.GetReloadableInt64Var(512*bytesize.KB, bytesize.B, "SalesforceBulkUpload.maxBufferCapacity")
 	return u
 }
@@ -121,7 +121,7 @@ func (s *Uploader) Transform(job *jobsdb.JobT) (string, error) {
 		Metadata: SalesforceJobMetadata{
 			JobID:           job.JobID,
 			RudderOperation: "upsert", // we only support upsert
-			ExternalID:      externalIDs,
+			ExternalIDs:     externalIDs,
 		},
 	}
 
@@ -183,40 +183,17 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 		return s.failedJobs(asyncDestStruct, fmt.Sprintf("extracting object info: %v", err))
 	}
 
-	// Drop events without an externalId value up front: with no upsert key they
-	// cannot be sent to Salesforce, so abort them with an error rather than
-	// emitting an uncorrelatable CSV row. Valid events continue to the upload.
-	validJobs := make([]SalesforceAsyncJob, 0, len(jobs))
-	var abortedJobIDs []int64
-	for _, job := range jobs {
-		if len(job.Metadata.ExternalID) == 0 || job.Metadata.ExternalID[0].ID == "" {
-			abortedJobIDs = append(abortedJobIDs, job.Metadata.JobID)
-			continue
-		}
-		validJobs = append(validJobs, job)
-	}
-	if len(abortedJobIDs) > 0 {
-		s.logger.Infon("Aborting events with missing externalId", logger.NewIntField("abortedJobs", int64(len(abortedJobIDs))))
-	}
-
-	// Every event was aborted (no valid externalId); nothing to upload.
-	if len(validJobs) == 0 {
-		return common.AsyncUploadOutput{
-			AbortJobIDs:   abortedJobIDs,
-			AbortReason:   missingExternalIDReason,
-			AbortCount:    len(abortedJobIDs),
-			DestinationID: destinationID,
-		}
-	}
-
-	validJobIDs := lo.Map(validJobs, func(job SalesforceAsyncJob, _ int) int64 {
+	// Every job reaching Upload was already validated at the Transform boundary
+	// (externalId id, type and identifierType are guaranteed present), so we can
+	// upload them all without re-checking the upsert key here.
+	jobIDs := lo.Map(jobs, func(job SalesforceAsyncJob, _ int) int64 {
 		return job.Metadata.JobID
 	})
 
 	csvFilePath, fileSize, err := createCSVFile(
 		destinationID,
 		objectInfo.ExternalIDField,
-		validJobs,
+		jobs,
 	)
 	defer func() {
 		if err := os.Remove(csvFilePath); err != nil {
@@ -228,26 +205,13 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 		return s.failedJobs(asyncDestStruct, fmt.Sprintf("creating CSV: %v", err))
 	}
 
-	// Persist each job's externalId in its importing status params, so the
-	// correlation can be rebuilt from the DB at poll time (restart-safe, no
-	// in-memory state, no payload reads).
-	jobImportingParameters := make(map[int64]stdjson.RawMessage, len(validJobs))
-	for _, job := range validJobs {
-		md, err := jsonrs.Marshal(importingMetadata{ExternalIDHash: hashExternalID(job.Metadata.ExternalID[0].ID)})
-		if err != nil {
-			s.logger.Errorn("marshalling importing metadata", logger.NewIntField("jobID", job.Metadata.JobID), obskit.Error(err))
-			continue
-		}
-		jobImportingParameters[job.Metadata.JobID] = md
-	}
-
 	// Track the size of the CSV we send to Salesforce and how many events it packs.
-	s.eventsPerFileStat.Observe(float64(len(validJobs)))
+	s.eventsPerFileStat.Observe(float64(len(jobs)))
 	s.payloadSizeStat.Observe(float64(fileSize))
 
 	s.logger.Infon("Created CSV file",
 		logger.NewStringField("csvFilePath", csvFilePath),
-		logger.NewIntField("jobs", int64(len(validJobs))),
+		logger.NewIntField("jobs", int64(len(jobs))),
 	)
 	sfJobID, apiError := s.apiService.CreateJob(
 		objectInfo.ObjectType,
@@ -281,29 +245,37 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 
 	s.logger.Infon("Successfully created and closed Salesforce Bulk job", logger.NewStringField("jobID", sfJobID))
 
+	// Persist each job's externalId (hashed) in its importing status params, so the
+	// correlation can be rebuilt from the DB at poll time (restart-safe, no
+	// in-memory state, no payload reads). Built only after the Salesforce job is
+	// created and closed, so we don't do this work for a failed upload.
+	jobImportingParameters := make(map[int64]stdjson.RawMessage, len(jobs))
+	for _, job := range jobs {
+		md, err := jsonrs.Marshal(importingMetadata{ExternalIDHash: hashExternalID(job.Metadata.ExternalIDs[0].ID)})
+		if err != nil {
+			s.logger.Errorn("marshalling importing metadata", logger.NewIntField("jobID", job.Metadata.JobID), obskit.Error(err))
+			continue
+		}
+		jobImportingParameters[job.Metadata.JobID] = md
+	}
+
 	importParameters, err := jsonrs.Marshal(common.ImportParameters{
 		ImportId: SalesforceJobInfo{
 			ID:              sfJobID,
 			ExternalIDField: objectInfo.ExternalIDField,
 		},
-		ImportCount: len(validJobIDs),
+		ImportCount: len(jobIDs),
 	})
 	if err != nil {
 		s.logger.Errorn("marshalling parameters", obskit.Error(err))
 	}
-	output := common.AsyncUploadOutput{
-		ImportingJobIDs:        validJobIDs,
+	return common.AsyncUploadOutput{
+		ImportingJobIDs:        jobIDs,
 		ImportingParameters:    importParameters,
-		ImportingCount:         len(validJobIDs),
+		ImportingCount:         len(jobIDs),
 		JobImportingParameters: jobImportingParameters,
 		DestinationID:          destinationID,
 	}
-	if len(abortedJobIDs) > 0 {
-		output.AbortJobIDs = abortedJobIDs
-		output.AbortReason = missingExternalIDReason
-		output.AbortCount = len(abortedJobIDs)
-	}
-	return output
 }
 
 func (s *Uploader) Poll(_ context.Context, pollInput common.AsyncPoll) common.PollStatusResponse {
@@ -441,6 +413,23 @@ func (s *Uploader) matchRecordsToJobs(
 		externalIDHashToJobID[externalIDHash] = append(externalIDHashToJobID[externalIDHash], job.JobID)
 	}
 
+	// No correlation metadata on any importing job: not a single Salesforce result
+	// can be matched back, so abort the whole batch up front (the records may
+	// already be in Salesforce, so retrying would only re-upsert). This is an
+	// abnormal state (pre-change in-flight jobs or a metadata-persistence bug), so
+	// log loudly for alerting.
+	if len(externalIDHashToJobID) == 0 {
+		s.logger.Errorn(
+			"No externalId correlation metadata found on importing jobs; aborting batch (records may already be in Salesforce)",
+			logger.NewIntField("importing_jobs", int64(len(importingList))),
+		)
+		for _, job := range importingList {
+			metadata.AbortedKeys = append(metadata.AbortedKeys, job.JobID)
+			metadata.AbortedReasons[job.JobID] = noCorrelationMetadataReason
+		}
+		return metadata
+	}
+
 	for _, failedRecord := range failedRecords {
 		key := hashExternalID(failedRecord[externalIDField])
 		if jobIDs, exists := externalIDHashToJobID[key]; exists {
@@ -469,30 +458,18 @@ func (s *Uploader) matchRecordsToJobs(
 		matchedJobIDs := lo.Concat(metadata.SucceededKeys, metadata.AbortedKeys)
 		missingJobIDs, _ := lo.Difference(importingJobIDs, matchedJobIDs)
 
-		// Abort the unmatched jobs. Re-uploading would only re-upsert records
-		// already in Salesforce, so retrying is pointless here.
-		reason := resultCorrelationFailedReason
-		if len(externalIDHashToJobID) == 0 {
-			// We could not correlate ANY result back to a job — abnormal state
-			// (pre-change in-flight jobs or a metadata-persistence bug). Log
-			// loudly so it can be alerted on.
-			s.logger.Errorn(
-				"No externalId correlation metadata found on importing jobs; aborting batch (records may already be in Salesforce)",
-				logger.NewIntField("importing_jobs", int64(len(importingList))),
-				logger.NewIntField("aborted", int64(len(missingJobIDs))),
-			)
-			reason = noCorrelationMetadataReason
-		} else {
-			s.logger.Errorn(
-				"Some Salesforce results could not be correlated back to jobs; aborting them",
-				logger.NewIntField("succeeded_keys", int64(len(metadata.SucceededKeys))),
-				logger.NewIntField("aborted_keys", int64(len(metadata.AbortedKeys))),
-				logger.NewIntField("importing_jobs", int64(len(importingList))),
-			)
-		}
+		// Some results couldn't be correlated back (e.g. Salesforce reformatted the
+		// externalId on store). Abort those jobs — re-uploading would only re-upsert
+		// records already in Salesforce, so retrying is pointless here.
+		s.logger.Errorn(
+			"Some Salesforce results could not be correlated back to jobs; aborting them",
+			logger.NewIntField("succeeded_keys", int64(len(metadata.SucceededKeys))),
+			logger.NewIntField("aborted_keys", int64(len(metadata.AbortedKeys))),
+			logger.NewIntField("importing_jobs", int64(len(importingList))),
+		)
 		metadata.AbortedKeys = append(metadata.AbortedKeys, missingJobIDs...)
 		for _, jobID := range missingJobIDs {
-			metadata.AbortedReasons[jobID] = reason
+			metadata.AbortedReasons[jobID] = resultCorrelationFailedReason
 		}
 	}
 	return metadata
--- a/router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
+++ b/router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
@@ -76,19 +76,14 @@ func (s *Uploader) Transform(job *jobsdb.JobT) (string, error) {
 	traits := gjson.GetBytes(job.EventPayload, "traits")
 	externalID := gjson.GetBytes(job.EventPayload, "context.externalId")
 
-	// Build the metadata object
-	metadata := make(map[string]any)
-	metadata["job_id"] = float64(job.JobID)
-
-	// We are supporting only upsert operation
-	metadata["rudderOperation"] = "upsert"
-
-	// Add externalId to metadata if it exists
-	var externalIdArray []any
-	if externalID.Exists() {
-		if err := jsonrs.Unmarshal([]byte(externalID.Raw), &externalIdArray); err == nil {
-			metadata["externalId"] = externalIdArray
-		}
+	// externalId is mandatory — it is the upsert key. Fail the event up front if
+	// it is absent rather than letting it flow downstream.
+	if !externalID.Exists() {
+		return "", fmt.Errorf("externalId is required but not present in the event")
+	}
+	var externalIDs []SalesforceExternalID
+	if err := jsonrs.Unmarshal([]byte(externalID.Raw), &externalIDs); err != nil {
+		return "", fmt.Errorf("failed to unmarshal externalId: %w", err)
 	}
 
 	// Build the message object
@@ -102,17 +97,20 @@ func (s *Uploader) Transform(job *jobsdb.JobT) (string, error) {
 		})
 	}
 
-	externalIdObject, err := extractFromVDM(externalIdArray)
+	externalIdObject, err := extractFromVDM(externalIDs)
 	if err != nil {
 		return "", fmt.Errorf("failed to extract externalId: %w", err)
 	}
 	// Add externalId field to message
 	message[externalIdObject.ExternalIDField] = externalIdObject.ExternalIDValue
 
-	// Create the final AsyncJob structure
-	asyncJob := common.AsyncJob{
-		Message:  message,
-		Metadata: metadata,
+	asyncJob := SalesforceAsyncJob{
+		Message: message,
+		Metadata: SalesforceJobMetadata{
+			JobID:           job.JobID,
+			RudderOperation: "upsert", // we only support upsert
+			ExternalID:      externalIDs,
+		},
 	}
 
 	// Marshal and return
@@ -124,19 +122,19 @@ func (s *Uploader) Transform(job *jobsdb.JobT) (string, error) {
 	return string(responsePayload), nil
 }
 
-func (s *Uploader) readJobsFromFile(filePath string) ([]common.AsyncJob, error) {
+func (s *Uploader) readJobsFromFile(filePath string) ([]SalesforceAsyncJob, error) {
 	file, err := os.Open(filePath)
 	if err != nil {
 		return nil, fmt.Errorf("opening file: %w", err)
 	}
 	defer file.Close()
 
-	var jobs []common.AsyncJob
+	var jobs []SalesforceAsyncJob
 	scanner := bufio.NewScanner(file)
 	scanner.Buffer(nil, int(s.config.maxBufferCapacity.Load()))
 
 	for scanner.Scan() {
-		var job common.AsyncJob
+		var job SalesforceAsyncJob
 		if err := jsonrs.Unmarshal(scanner.Bytes(), &job); err != nil {
 			return nil, fmt.Errorf("unmarshalling job: %w", err)
 		}
@@ -150,58 +148,61 @@ func (s *Uploader) readJobsFromFile(filePath string) ([]common.AsyncJob, error) 
 	return jobs, nil
 }
 
+func (s *Uploader) failedJobs(asyncDestStruct *common.AsyncDestinationStruct, failedReason string) common.AsyncUploadOutput {
+	return common.AsyncUploadOutput{
+		FailedJobIDs:  asyncDestStruct.ImportingJobIDs,
+		FailedCount:   len(asyncDestStruct.ImportingJobIDs),
+		FailedReason:  failedReason,
+		DestinationID: asyncDestStruct.Destination.ID,
+	}
+}
+
 func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestinationStruct) common.AsyncUploadOutput {
-	destination := asyncDestStruct.Destination
-	destinationID := destination.ID
+	destinationID := asyncDestStruct.Destination.ID
 	filePath := asyncDestStruct.FileName
-	importingJobIDs := asyncDestStruct.ImportingJobIDs
 
 	input, err := s.readJobsFromFile(filePath)
 	if err != nil {
-		return common.AsyncUploadOutput{
-			FailedJobIDs:  importingJobIDs,
-			FailedReason:  fmt.Sprintf("Error reading jobs from file: %v", err),
-			FailedCount:   len(importingJobIDs),
-			DestinationID: destinationID,
-		}
+		return s.failedJobs(asyncDestStruct, fmt.Sprintf("Error reading jobs from file: %v", err))
 	}
 
 	objectInfo, err := extractObjectInfo(input)
 	if err != nil {
-		return common.AsyncUploadOutput{
-			FailedJobIDs:  importingJobIDs,
-			FailedReason:  fmt.Sprintf("Error extracting object info: %v", err),
-			FailedCount:   len(importingJobIDs),
-			DestinationID: destinationID,
-		}
+		return s.failedJobs(asyncDestStruct, fmt.Sprintf("Error extracting object info: %v", err))
 	}
 	if objectInfo.ExternalIDField == "" {
-		return common.AsyncUploadOutput{
-			FailedJobIDs:  importingJobIDs,
-			FailedReason:  "externalId field is empty; cannot correlate poll results back to jobs",
-			FailedCount:   len(importingJobIDs),
-			DestinationID: destinationID,
-		}
+		return s.failedJobs(asyncDestStruct, externalIDFieldEmptyReason)
 	}
 
 	// Drop events without an externalId value up front: with no upsert key they
 	// cannot be sent to Salesforce, so abort them with an error rather than
 	// emitting an uncorrelatable CSV row. Valid events continue to the upload.
-	validJobs := make([]common.AsyncJob, 0, len(input))
-	validJobIDs := make([]int64, 0, len(input))
+	validJobs := make([]SalesforceAsyncJob, 0, len(input))
 	var abortedJobIDs []int64
 	for _, job := range input {
-		jobID := int64(job.Metadata["job_id"].(float64))
 		if externalIDValue, _ := common.FormatCSVValue(job.Message[objectInfo.ExternalIDField]); externalIDValue == "" {
-			abortedJobIDs = append(abortedJobIDs, jobID)
+			abortedJobIDs = append(abortedJobIDs, job.Metadata.JobID)
 			continue
 		}
 		validJobs = append(validJobs, job)
-		validJobIDs = append(validJobIDs, jobID)
 	}
 	if len(abortedJobIDs) > 0 {
 		s.logger.Infon("Aborting events with missing externalId", logger.NewIntField("abortedJobs", int64(len(abortedJobIDs))))
 	}
+
+	// Every event was aborted (no valid externalId); nothing to upload.
+	if len(validJobs) == 0 {
+		return common.AsyncUploadOutput{
+			AbortJobIDs:   abortedJobIDs,
+			AbortReason:   missingExternalIDReason,
+			AbortCount:    len(abortedJobIDs),
+			DestinationID: destinationID,
+		}
+	}
+
+	validJobIDs := lo.Map(validJobs, func(job SalesforceAsyncJob, _ int) int64 {
+		return job.Metadata.JobID
+	})
 
 	csvFilePath, externalIDToJobID, err := createCSVFile(
 		destinationID,
@@ -215,12 +216,7 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 	}()
 	if err != nil {
 		s.logger.Errorn("Error creating CSV", obskit.Error(err))
-		return common.AsyncUploadOutput{
-			FailedJobIDs:  importingJobIDs,
-			FailedReason:  fmt.Sprintf("Error creating CSV: %v", err),
-			FailedCount:   len(importingJobIDs),
-			DestinationID: destinationID,
-		}
+		return s.failedJobs(asyncDestStruct, fmt.Sprintf("Error creating CSV: %v", err))
 	}
 	s.externalIDToJobID = externalIDToJobID
 
@@ -245,12 +241,7 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 	)
 	if apiError != nil {
 		s.logger.Errorn("Error creating Salesforce job for operation upsert.", logger.NewStringField("apiErrorMessage", apiError.Message), logger.NewStringField("category", apiError.Category), logger.NewIntField("statusCode", int64(apiError.StatusCode)))
-		return common.AsyncUploadOutput{
-			FailedJobIDs:  importingJobIDs,
-			FailedReason:  fmt.Sprintf("Error creating Salesforce job: %v", apiError.Message),
-			FailedCount:   len(importingJobIDs),
-			DestinationID: destinationID,
-		}
+		return s.failedJobs(asyncDestStruct, fmt.Sprintf("Error creating Salesforce job: %v", apiError.Message))
 	}
 
 	apiError = s.apiService.UploadData(sfJobID, csvFilePath)
@@ -259,12 +250,7 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 		if err := s.apiService.DeleteJob(sfJobID); err != nil {
 			s.logger.Errorn("Error deleting Salesforce job.", logger.NewStringField("apiErrorMessage", err.Message))
 		}
-		return common.AsyncUploadOutput{
-			FailedJobIDs:  importingJobIDs,
-			FailedReason:  fmt.Sprintf("Error uploading data: %v", apiError.Message),
-			FailedCount:   len(importingJobIDs),
-			DestinationID: destinationID,
-		}
+		return s.failedJobs(asyncDestStruct, fmt.Sprintf("Error uploading data: %v", apiError.Message))
 	}
 
 	apiError = s.apiService.CloseJob(sfJobID)
@@ -273,12 +259,7 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 		if err := s.apiService.DeleteJob(sfJobID); err != nil {
 			s.logger.Errorn("Error deleting Salesforce job.", logger.NewStringField("apiErrorMessage", err.Message))
 		}
-		return common.AsyncUploadOutput{
-			FailedJobIDs:  importingJobIDs,
-			FailedReason:  fmt.Sprintf("Error closing job: %v", apiError.Message),
-			FailedCount:   len(importingJobIDs),
-			DestinationID: destinationID,
-		}
+		return s.failedJobs(asyncDestStruct, fmt.Sprintf("Error closing job: %v", apiError.Message))
 	}
 
 	s.logger.Infon("Successfully created and closed Salesforce Bulk job", logger.NewStringField("jobID", sfJobID))
@@ -293,15 +274,18 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 	if err != nil {
 		s.logger.Errorn("marshalling parameters", obskit.Error(err))
 	}
-	return common.AsyncUploadOutput{
+	output := common.AsyncUploadOutput{
 		ImportingJobIDs:     validJobIDs,
 		ImportingParameters: importParameters,
 		ImportingCount:      len(validJobIDs),
-		AbortJobIDs:         abortedJobIDs,
-		AbortReason:         "externalId is missing for the event; cannot upsert to Salesforce",
-		AbortCount:          len(abortedJobIDs),
 		DestinationID:       destinationID,
 	}
+	if len(abortedJobIDs) > 0 {
+		output.AbortJobIDs = abortedJobIDs
+		output.AbortReason = missingExternalIDReason
+		output.AbortCount = len(abortedJobIDs)
+	}
+	return output
 }
 
 func (s *Uploader) Poll(_ context.Context, pollInput common.AsyncPoll) common.PollStatusResponse {
@@ -464,7 +448,7 @@ func (s *Uploader) matchRecordsToJobs(
 		importingJobIDs := lo.Map(importingList, func(job *jobsdb.JobT, _ int) int64 {
 			return job.JobID
 		})
-		matchedJobIDs := append(append([]int64{}, metadata.SucceededKeys...), metadata.AbortedKeys...)
+		matchedJobIDs := lo.Concat(metadata.SucceededKeys, metadata.AbortedKeys)
 		missingJobIDs, _ := lo.Difference(importingJobIDs, matchedJobIDs)
 
 		if len(s.externalIDToJobID) == 0 {
--- a/router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
+++ b/router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
@@ -3,6 +3,7 @@ package salesforcebulkupload
 import (
 	"bufio"
 	"context"
+	stdjson "encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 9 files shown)


router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -195,6 +204,16 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 	}
 	s.dataHashToJobID = hashToJobID
 
+	// The source is expected to send unique externalIds per batch. Fewer keys
+	// than jobs means duplicates collided onto the same key, which makes
+	// per-job status attribution ambiguous after polling.
+	if len(hashToJobID) < len(input) {
+		s.logger.Warnn("Duplicate externalId values detected in batch; per-job status attribution may be ambiguous",
```
achettyiitr: Are we planning to have an alert on top of it? If you, can you reference the ticket?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/utils.go
```diff
@@ -115,23 +121,17 @@ func createCSVFile(
 		if err := writer.Write(row); err != nil {
 			return "", nil, nil, fmt.Errorf("writing CSV row: %w", err)
 		}
-		hash := calculateHashCode(row)
-		dataHashToJobID[hash] = append(dataHashToJobID[hash], jobID)
+		// Correlate on the externalId value alone, not the whole row: Salesforce
+		// coerces other columns on store (datetime ms-truncation, number scale,
+		// ...) so a whole-row hash would not survive the round-trip.
+		key := calculateHashCode(row[externalIDIndex])
+		dataHashToJobID[key] = append(dataHashToJobID[key], jobID)
```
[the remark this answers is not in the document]
ItsSudip: Addressed by dropping events with an empty/missing externalId in `Upload`, before the CSV is built — they are returned as `AbortJobIDs` with an error since they have no upsert key. Because such events never reach `createCSVFile`, they can no longer collide onto `hash("")`.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -391,8 +410,8 @@ func (s *Uploader) matchRecordsToJobs(
 	}
```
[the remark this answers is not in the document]
ItsSudip: `externalIDField` is always defined here: `Upload` persists it into the import parameters (`SalesforceJobInfo`) and validates it is non-empty before uploading, so it is guaranteed present at `GetUploadStats` time. Missing-externalId is now handled at the input boundary — `Upload` aborts those events up front rather than letting them reach correlation.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -405,8 +424,8 @@ func (s *Uploader) matchRecordsToJobs(
...
```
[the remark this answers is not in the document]
ItsSudip: Empty-externalId events are aborted in `Upload` before they reach the correlation map, so the map never contains `hash("")`. A returned record with an empty externalId therefore matches no job and cannot be mis-attributed, so no skip is needed here.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -405,8 +424,8 @@ func (s *Uploader) matchRecordsToJobs(
...
 
 	for _, successRecord := range successRecords {
-		hash := calculateHashFromRecord(successRecord, headers)
-		if jobIDs, exists := s.dataHashToJobID[hash]; exists {
+		key := calculateHashCode(successRecord[externalIDField])
+		if jobIDs, exists := s.dataHashToJobID[key]; exists {
 			metadata.SucceededKeys = append(metadata.SucceededKeys, jobIDs...)
 		}
 	}
```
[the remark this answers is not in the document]
ItsSudip: Same as the failed-record loop — empty externalIds are dropped in `Upload` before the map is built, so an empty value in a returned record cannot false-match. No guard needed here.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -391,8 +433,8 @@ func (s *Uploader) matchRecordsToJobs(
...
 
 	for _, failedRecord := range failedRecords {
-		hash := calculateHashFromRecord(failedRecord, headers)
-		if jobIDs, exists := s.dataHashToJobID[hash]; exists {
+		key := calculateHashCode(failedRecord[externalIDField])
```
achettyiitr: If it is just single column, do we even need hashing for it?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -175,9 +175,38 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
+			continue
+		}
+		validJobs = append(validJobs, job)
+		validJobIDs = append(validJobIDs, jobID)
+	}
+	if len(abortedJobIDs) > 0 {
+		s.logger.Infon("Aborting events with missing externalId", logger.NewIntField("abortedJobs", int64(len(abortedJobIDs))))
+	}
+
+	csvFilePath, _, hashToJobID, err := createCSVFile(
```
achettyiitr: Since we are not using header, can we omit from createCSVFile return type as it is not being used?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Yes, but it can be pii data, so I thought of hashing it.

achettyiitr: Not needed right now, we are only keeping it in memory, right?

ItsSudip: Yes

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Yes, we can add an alert on it.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Yes removed

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -175,9 +175,38 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
 			DestinationID: destinationID,
 		}
 	}
-	csvFilePath, csvHeaders, hashToJobID, err := createCSVFile(
+	if objectInfo.ExternalIDField == "" {
+		return common.AsyncUploadOutput{
+			FailedJobIDs:  importingJobIDs,
+			FailedReason:  "externalId field is empty; cannot correlate poll results back to jobs",
```
achettyiitr: Can you create a constant for this?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -175,9 +175,38 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
+			FailedCount:   len(importingJobIDs),
+			DestinationID: destinationID,
+		}
```
achettyiitr: Can we create a helper function for this? It's getting repeated everywhere.

```
func (s *Uploader) failedJobs(asyncDest *common.AsyncDestinationStruct, failedReason string) common.AsyncUploadOutput {
	return common.AsyncUploadOutput{
		FailedJobIDs:  asyncDest.ImportingJobIDs,
		FailedCount:   len(asyncDest.ImportingJobIDs),
		FailedReason:  failedReason,
		DestinationID: asyncDest.Destination.ID,
	}
}
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
achettyiitr: Since the externalID is now a mandatory thing, should we fail events here as well if it does not exist?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -175,9 +175,38 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
+	}
+
+	// Drop events without an externalId value up front: with no upsert key they
+	// cannot be sent to Salesforce, so abort them with an error rather than
+	// emitting an uncorrelatable CSV row. Valid events continue to the upload.
+	validJobs := make([]common.AsyncJob, 0, len(input))
+	validJobIDs := make([]int64, 0, len(input))
+	var abortedJobIDs []int64
+	for _, job := range input {
+		jobID := int64(job.Metadata["job_id"].(float64))
```
achettyiitr: While un-marshalling, instead of using the previous type:
```
type AsyncJob struct {
	Message  map[string]any `json:"message"`
	Metadata map[string]any `json:"metadata"`
}
```

We can use something like this to be deterministic, as we can unmarshall the payload and use something like this type:
```
type SalesforceAsyncJob struct {
	Message  map[string]any `json:"message"`
	Metadata struct {
		JobID           int64  `json:"job_id"`
		RudderOperation string `json:"rudderOperation"`
		ExternalID      string `json:"externalId"`
	} `json:"metadata"`
}
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -175,9 +175,38 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
+		if externalIDValue, _ := common.FormatCSVValue(job.Message[objectInfo.ExternalIDField]); externalIDValue == "" {
+			abortedJobIDs = append(abortedJobIDs, jobID)
+			continue
+		}
+		validJobs = append(validJobs, job)
+		validJobIDs = append(validJobIDs, jobID)
+	}
+	if len(abortedJobIDs) > 0 {
```
achettyiitr: What happens if everything is aborted? Do we even need to proceed?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -175,9 +175,38 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
+			FailedCount:   len(importingJobIDs),
+			DestinationID: destinationID,
+		}
+	}
+
+	// Drop events without an externalId value up front: with no upsert key they
+	// cannot be sent to Salesforce, so abort them with an error rather than
+	// emitting an uncorrelatable CSV row. Valid events continue to the upload.
+	validJobs := make([]common.AsyncJob, 0, len(input))
+	validJobIDs := make([]int64, 0, len(input))
```
achettyiitr: If we use something like SalesforceAsyncJob, we don't need validJobIDs as it is embedded within it.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -246,18 +285,21 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
 		s.logger.Errorn("marshalling parameters", obskit.Error(err))
 	}
 	return common.AsyncUploadOutput{
-		ImportingJobIDs:     importingJobIDs,
+		ImportingJobIDs:     validJobIDs,
 		ImportingParameters: importParameters,
-		ImportingCount:      len(importingJobIDs),
+		ImportingCount:      len(validJobIDs),
+		AbortJobIDs:         abortedJobIDs,
+		AbortReason:         "externalId is missing for the event; cannot upsert to Salesforce",
```
achettyiitr: If there are no AbortJobIDs, no need to add AbortReason?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -422,16 +464,33 @@ func (s *Uploader) matchRecordsToJobs(
 		importingJobIDs := lo.Map(importingList, func(job *jobsdb.JobT, _ int) int64 {
 			return job.JobID
 		})
-		missingJobIDs, _ := lo.Difference(importingJobIDs, append(metadata.SucceededKeys, metadata.AbortedKeys...))
-
-		metadata.FailedKeys = missingJobIDs
-		metadata.FailedReasons = lo.SliceToMap(missingJobIDs, func(jobID int64) (int64, string) {
-			return jobID, "Input hash is not found in the data hash to job ID map or the job is not found in the successRespons or failedResponse"
-		})
+		matchedJobIDs := append(append([]int64{}, metadata.SucceededKeys...), metadata.AbortedKeys...)
```
achettyiitr: No need to empty slice at the start.
```suggestion
		matchedJobIDs := lo.Concat(metadata.SucceededKeys, metadata.AbortedKeys)
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -422,16 +464,33 @@ func (s *Uploader) matchRecordsToJobs(
...
+
+		if len(s.externalIDToJobID) == 0 {
+			// No correlation data at all (e.g. the process restarted between
+			// Upload and GetUploadStats and lost the in-memory map). This is a
+			// systemic loss, not a per-record mismatch, so keep these jobs
+			// retryable rather than permanently dropping a batch that may have
+			// succeeded in Salesforce.
+			metadata.FailedKeys = missingJobIDs
+			metadata.FailedReasons = lo.SliceToMap(missingJobIDs, func(jobID int64) (int64, string) {
+				return jobID, "Correlation map is empty (likely a restart between upload and stats); retrying"
```
achettyiitr: Create a const for this?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -422,16 +464,33 @@ func (s *Uploader) matchRecordsToJobs(
...
+			})
+		} else {
+			// The map was populated but these records could not be matched back
+			// to a job, most likely because Salesforce reformatted the
+			// externalId value on store. Retrying would re-upload the same value
+			// and miss again, so abort instead of looping through retries.
+			metadata.AbortedKeys = append(metadata.AbortedKeys, missingJobIDs...)
+			for _, jobID := range missingJobIDs {
+				metadata.AbortedReasons[jobID] = "Could not correlate Salesforce result back to the job: externalId not found in success/failed records (possibly reformatted by Salesforce on store)"
```
achettyiitr: Create a const for this?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — extracted into constants `externalIDFieldEmptyReason` and `missingExternalIDReason` in types.go.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — added the `failedJobs(asyncDestStruct, reason)` helper and routed every failure return (file read, object info, empty field, CSV, CreateJob, UploadData, CloseJob) through it.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — Transform now fails the event up front when externalId is absent, and also surfaces the unmarshal error that was previously swallowed.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — added a typed `SalesforceAsyncJob` (`SalesforceJobMetadata` + `SalesforceExternalID`) used in readJobsFromFile / createCSVFile / Transform, removing the `map[string]any` + `.(float64)` assertion. One change from the suggestion: externalId in metadata is the VDM array, so it is typed as `[]SalesforceExternalID` rather than `string`.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — if every event is aborted (no valid externalId), Upload returns the abort-only output immediately without creating a Salesforce job. Added a test asserting no API calls are made.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — with the typed struct, validJobIDs is now derived via `lo.Map(validJobs, ...)` instead of a separately maintained slice.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — AbortJobIDs/AbortReason/AbortCount are only set when there are aborted jobs.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — using `lo.Concat(metadata.SucceededKeys, metadata.AbortedKeys)`.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/utils.go
```diff
@@ -15,61 +14,49 @@ import (
...
-	firstJob := jobs[0]
-
-	externalIDRaw, ok := firstJob.Metadata["externalId"]
-	if !ok {
+	if len(jobs[0].Metadata.ExternalID) == 0 {
 		return nil, fmt.Errorf("externalId not found in the first job")
 	}
 
-	objectInfo, err := extractFromVDM(externalIDRaw)
+	objectInfo, err := extractFromVDM(jobs[0].Metadata.ExternalID)
```
achettyiitr:
```suggestion
    externalID := jobs[0].Metadata.ExternalID
	if len(externalID) == 0 {
return nil, fmt.Errorf("externalId not found in the first job")
}

	objectInfo, err := extractFromVDM(externalID)
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -76,19 +76,14 @@ func (s *Uploader) Transform(job *jobsdb.JobT) (string, error) {
...
-	// Add externalId to metadata if it exists
-	var externalIdArray []any
-	if externalID.Exists() {
-		if err := jsonrs.Unmarshal([]byte(externalID.Raw), &externalIdArray); err == nil {
-			metadata["externalId"] = externalIdArray
-		}
+	// externalId is mandatory — it is the upsert key. Fail the event up front if
+	// it is absent rather than letting it flow downstream.
+	if !externalID.Exists() {
+		return "", fmt.Errorf("externalId is required but not present in the event")
```
achettyiitr: Can we create a error constant for this?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -150,34 +148,66 @@ func (s *Uploader) readJobsFromFile(filePath string) ([]common.AsyncJob, error)
...
 
 	input, err := s.readJobsFromFile(filePath)
 	if err != nil {
-		return common.AsyncUploadOutput{
-			FailedJobIDs:  importingJobIDs,
-			FailedReason:  fmt.Sprintf("Error reading jobs from file: %v", err),
-			FailedCount:   len(importingJobIDs),
-			DestinationID: destinationID,
-		}
+		return s.failedJobs(asyncDestStruct, fmt.Sprintf("Error reading jobs from file: %v", err))
```
achettyiitr: Error should start with lower case. Also, we don't need to mention error keyword here. We only need to mention the action which is failing. Same for other use case as well.
```suggestion
		return s.failedJobs(asyncDestStruct, fmt.Sprintf("reading jobs from file: %v", err))
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — extracted into the `emptyCorrelationMapReason` constant in types.go.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — extracted into the `resultCorrelationFailedReason` constant in types.go.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/utils.go
ItsSudip: Done — pulled `jobs[0].Metadata.ExternalID` into a local `externalID` and used it for both the length check and `extractFromVDM`.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — added a sentinel error `errExternalIDRequired` and return it instead of an inline `fmt.Errorf`.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — reworded all the `failedJobs` reasons to lowercase, action-only (e.g. `reading jobs from file: %v`, `extracting object info: %v`, `creating CSV: %v`, `creating Salesforce job: %v`, `uploading data: %v`, `closing job: %v`). Also lowercased the two reconciliation reason constants for consistency.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -186,18 +219,35 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
+	}
+
+	// Track the size of the CSV we send to Salesforce and how many events it packs.
+	statTags := stats.Tags{
+		"module":        "batch_router",
+		"destType":      s.destName,
+		"destinationId": asyncDestStruct.Destination.ID,
+		"workspaceId":   asyncDestStruct.Destination.WorkspaceID,
+	}
+	s.statsFactory.NewTaggedStat("events_per_file", stats.HistogramType, statTags).Observe(float64(len(validJobs)))
```
achettyiitr: Can we prepare this stat in the Uploader upfront rather than creating it for every upload?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -186,18 +219,35 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
+	if fileInfo, statErr := os.Stat(csvFilePath); statErr == nil {
+		s.statsFactory.NewTaggedStat("payload_size", stats.HistogramType, statTags).Observe(float64(fileInfo.Size()))
 	}
```
achettyiitr: Instead of doing os.Stat, since the file is being prepared by us, can the createCSVFile function return the byte size as well?

router/batchrouter/handle_lifecycle.go
```diff
@@ -189,7 +189,7 @@ func (brt *Handle) Setup(
 func (brt *Handle) setupReloadableVars() {
 	brt.maxFailedCountForJob = config.GetReloadableIntVar(128, 1, "BatchRouter."+brt.destType+".maxFailedCountForJob", "BatchRouter.maxFailedCountForJob")
 	brt.maxFailedCountForSourcesJob = config.GetReloadableIntVar(3, 1, "BatchRouter.RSources."+brt.destType+".maxFailedCountForJob", "BatchRouter.RSources.maxFailedCountForJob")
-	brt.asyncUploadTimeout = config.GetReloadableDurationVar(30, time.Minute, "BatchRouter."+brt.destType+".asyncUploadTimeout", "BatchRouter.asyncUploadTimeout")
+	brt.asyncUploadTimeout = config.GetReloadableDurationVar(5, time.Second, "BatchRouter."+brt.destType+".asyncUploadTimeout", "BatchRouter.asyncUploadTimeout")
```
achettyiitr: Any reason, we are increase the default timeout for everyone?

ItsSudip: This is unintentional; I am removing it.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — moved the three stats (`payload_size`, `events_per_file`, `async_upload_time`) onto the `Uploader` as fields, created once in `NewUploader`. `Upload` now just calls `.Observe()` / `.Since()` on them instead of `NewTaggedStat` per upload.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — `createCSVFile` now wraps the file in a small `countingWriter` and returns the byte count, so we no longer `os.Stat` afterward. `Upload` passes that size straight into `payload_size`.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -150,34 +159,66 @@ func (s *Uploader) readJobsFromFile(filePath string) ([]common.AsyncJob, error)
...
+	// cannot be sent to Salesforce, so abort them with an error rather than
+	// emitting an uncorrelatable CSV row. Valid events continue to the upload.
+	validJobs := make([]SalesforceAsyncJob, 0, len(input))
+	var abortedJobIDs []int64
+	for _, job := range input {
+		if externalIDValue, _ := common.FormatCSVValue(job.Message[objectInfo.ExternalIDField]); externalIDValue == "" {
+			abortedJobIDs = append(abortedJobIDs, job.Metadata.JobID)
+			continue
+		}
+		validJobs = append(validJobs, job)
```
achettyiitr: Can we handle the error appropriately? A FormatCSVValue error gets reported as "externalId is missing for the event", which will confuse whoever debugs it. Handle the error with its own reason.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/types.go
```diff
@@ -48,8 +51,12 @@ type apiService struct {
 }
 
 type SalesforceJobInfo struct {
-	ID      string   `json:"id"`
-	Headers []string `json:"headers"`
+	ID string `json:"id"`
+	// ExternalIDField is the upsert key field name. Job statuses are correlated
+	// after polling by hashing this field's value (not the whole row), so that
+	// Salesforce's value coercion on other columns (e.g. datetime millisecond
+	// truncation, number scale) does not break the match.
```
achettyiitr, line 57: This comments seems stale, we are not doing hashing anymore?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -60,13 +63,21 @@ func NewUploader(
...
+		externalIDToJobID: make(map[string][]int64),
+		destination:       destination,
+		statsFactory:      statsFactory,
+		destName:          destName,
+	}
+	statTags := stats.Tags{"module": "batch_router", "destType": destName}
+	if destination != nil {
+		statTags["destinationId"] = destination.ID
+		statTags["workspaceId"] = destination.WorkspaceID
 	}
```
achettyiitr: Why would the destination be nil here?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
achettyiitr: Let's capture aborted jobs along with aborted error reason as well.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -60,13 +63,21 @@ func NewUploader(
...
+	u.payloadSizeStat = statsFactory.NewTaggedStat("payload_size", stats.HistogramType, statTags)
+	u.eventsPerFileStat = statsFactory.NewTaggedStat("events_per_file", stats.HistogramType, statTags)
```
achettyiitr: Are [these](https://github.com/rudderlabs/rudder-server/blob/feat.bqstreamv2-2-async-manager/runner/buckets.go#L8-L13) buckets sufficient for them, or we need to create new onces?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/utils.go
```diff
@@ -15,66 +15,68 @@ import (
...
 	}
 
-	externalIDMap, ok := externalIDArray[0].(map[string]any)
-	if !ok {
-		return nil, fmt.Errorf("externalId[0] must be an object")
-	}
-
-	typeStr, _ := externalIDMap["type"].(string)
-	if typeStr == "" {
+	first := externalIDs[0]
```
achettyiitr:
```suggestion
	externalID := externalIDs[0]
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -150,34 +159,66 @@ func (s *Uploader) readJobsFromFile(filePath string) ([]common.AsyncJob, error)
...
+		return s.failedJobs(asyncDestStruct, fmt.Sprintf("reading jobs from file: %v", err))
 	}
 
 	objectInfo, err := extractObjectInfo(input)
 	if err != nil {
+		return s.failedJobs(asyncDestStruct, fmt.Sprintf("extracting object info: %v", err))
+	}
+	if objectInfo.ExternalIDField == "" {
+		return s.failedJobs(asyncDestStruct, externalIDFieldEmptyReason)
+	}
```
achettyiitr: Instead of keeping these checks at the caller, can we move these within extractObjectInfo?

```
func extractFromVDM(externalIDs []SalesforceExternalID) (*ObjectInfo, error) {
	if len(externalIDs) == 0 {
		return nil, fmt.Errorf("externalId must be an array with at least one element")
	}

	first := externalIDs[0] 
	if first.Type == "" {
		return nil, fmt.Errorf("externalId type is required")
	}
    const objectType = strings.Replace(first.Type, "SALESFORCE_BULK_UPLOAD-", "", 1)
    -- Check for IdentifierType, ID and objectType
	return &ObjectInfo{
		ObjectType:      strings.Replace(first.Type, "SALESFORCE_BULK_UPLOAD-", "", 1),
		ExternalIDField: first.IdentifierType,
		ExternalIDValue: first.ID,
	}, nil
}
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -150,34 +159,66 @@ func (s *Uploader) readJobsFromFile(filePath string) ([]common.AsyncJob, error)
...
+
+	// Drop events without an externalId value up front: with no upsert key they
+	// cannot be sent to Salesforce, so abort them with an error rather than
+	// emitting an uncorrelatable CSV row. Valid events continue to the upload.
+	validJobs := make([]SalesforceAsyncJob, 0, len(input))
+	var abortedJobIDs []int64
+	for _, job := range input {
```
achettyiitr: input is very generic, can we use something like events or jobs?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -150,34 +159,66 @@ func (s *Uploader) readJobsFromFile(filePath string) ([]common.AsyncJob, error)
...
+		if externalIDValue, _ := common.FormatCSVValue(job.Message[objectInfo.ExternalIDField]); externalIDValue == "" {
+			abortedJobIDs = append(abortedJobIDs, job.Metadata.JobID)
+			continue
+		}
+		validJobs = append(validJobs, job)
+	}
+	if len(abortedJobIDs) > 0 {
+		s.logger.Infon("Aborting events with missing externalId", logger.NewIntField("abortedJobs", int64(len(abortedJobIDs))))
+	}
```
achettyiitr: This is misleading, as error can be anything while we do common.FormatCSVValue

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -234,32 +276,33 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
 	}
 
 	s.logger.Infon("Successfully created and closed Salesforce Bulk job", logger.NewStringField("jobID", sfJobID))
 
 	importParameters, err := jsonrs.Marshal(common.ImportParameters{
 		ImportId: SalesforceJobInfo{
-			ID:      sfJobID,
-			Headers: csvHeaders,
+			ID:              sfJobID,
+			ExternalIDField: objectInfo.ExternalIDField,
```
achettyiitr, line 264: This might break for existing events in the pipeline; we might need to handle for backward compatability?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/types.go
```diff
@@ -20,13 +20,16 @@ const (
...
-	config          struct {
+	destName            string
+	logger              logger.Logger
+	statsFactory        stats.Stats
+	apiService          APIServiceInterface
+	externalIDToJobID   map[string][]int64
+	destination         *backendconfig.DestinationT
+	payloadSizeStat     stats.Measurement
+	eventsPerFileStat   stats.Measurement
+	asyncUploadTimeStat stats.Measurement
```
achettyiitr: Instead of keeping it generic, can we use stats.TImer & stats.Histogram?

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
achettyiitr: There are no customers using it, we can ignore it.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/types.go
ItsSudip: Done — updated the comment. We correlate on the externalId value directly now (no hashing).

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: It is never nil in production — NewManager always passes a real destination. The guard was only for test callers that passed nil; removed it and updated the tests to pass a real destination.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: `payload_size` already has byte buckets in `customBuckets` (10B–1GB). I added `events_per_file` count buckets (1…100k) in runner/buckets.go. `async_upload_time` is a timer, so the default time buckets fit it.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/utils.go
ItsSudip: Done — renamed `first` to `externalID`.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — moved the field validation into `extractFromVDM` (it now errors when `identifierType` is empty) and removed the separate `ExternalIDField == ""` check in Upload. I did not add an `ID` check there: an empty externalId value is intentionally handled by the Upload partition (the event is aborted), so validating ID in `extractFromVDM` would instead fail it at Transform.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — renamed to `jobs`.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — the partition now checks the typed `job.Metadata.ExternalID[0].ID` instead of `FormatCSVValue(...)`, so there is no swallowed error to mislabel.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/types.go
ItsSudip: Done — typed the fields as `stats.Histogram` (payload_size, events_per_file) and `stats.Timer` (async_upload_time).

achettyiitr (approved):

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -60,13 +63,22 @@ func NewUploader(
...
+		destName:          destName,
+	}
+	statTags := stats.Tags{
+		"module":        "batch_router",
+		"destType":      destName,
+		"destinationId": destination.ID,
+		"workspaceId":   destination.WorkspaceID,
 	}
+	u.payloadSizeStat = statsFactory.NewTaggedStat("payload_size", stats.HistogramType, statTags)
+	u.eventsPerFileStat = statsFactory.NewTaggedStat("events_per_file", stats.HistogramType, statTags)
```
atzoum: use a proper prefix for these metrics, it is too generic now e.g. `brt_async_dest_events_per_file` & `brt_async_dest_payload_size`

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -422,16 +463,33 @@ func (s *Uploader) matchRecordsToJobs(
...
-		missingJobIDs, _ := lo.Difference(importingJobIDs, append(metadata.SucceededKeys, metadata.AbortedKeys...))
-
-		metadata.FailedKeys = missingJobIDs
-		metadata.FailedReasons = lo.SliceToMap(missingJobIDs, func(jobID int64) (int64, string) {
-			return jobID, "Input hash is not found in the data hash to job ID map or the job is not found in the successRespons or failedResponse"
-		})
+		matchedJobIDs := lo.Concat(metadata.SucceededKeys, metadata.AbortedKeys)
+		missingJobIDs, _ := lo.Difference(importingJobIDs, matchedJobIDs)
+
+		if len(s.externalIDToJobID) == 0 {
```
atzoum: Can we store `externalIDs` of each job in the importing job status params?
Use a new `JobImportingParameters map[int64]stdjson.RawMessage`, so that we can simply reconstruct the mappings without relying on in-memory state.
We should avoid reading job payload during poll, since we are trying to skip loading payloads from jobsdb in case of importing jobs

cc @achettyiitr

achettyiitr: For this integration, we are not using reading job payload during poll EventPayload.

achettyiitr: One concern there is, would it be an issue if we store the entire `JobImportingParameters map[int64]stdjson.RawMessage` in the status table. Currently, all the job statuses across all jobs would have the same value, so it would be too much data.

atzoum: @achettyiitr one new field per job, so different job status parameters per job

achettyiitr: Currently, the problem is that in the async framework, we are storing the [same](https://github.com/rudderlabs/rudder-server/blob/fix/INT-6539-klaviyo-abort-non-retryable-errors/router/batchrouter/handle_async.go#L756) importing job parameters for all jobs.

router/batchrouter/handle_async.go
```diff
@@ -745,6 +746,29 @@ func (brt *Handle) setMultipleJobStatus(params setMultipleJobStatusParams) {
...
 				SourceID:      gjson.GetBytes(jobParameters, "source_id").String(),
 			}
+			// Persist per-job metadata (e.g. the externalId) in the stored status
+			// Parameters, so it can be read back at poll time without loading the
+			// job payload or relying on any in-memory state. Example shape:
+			// {
+			//   "importId": {"id": "750Ie000003ZFLSIA4", "externalIdField": "Email"},
+			//   "importCount": 3,
+			//   "metadata": {"externalIdHash": "<sha256 hex of the externalId>"}
+			// }
```
achettyiitr: Do we need to mentions this as it is integration specific?

router/batchrouter/handle_async.go
```diff
@@ -745,6 +746,29 @@ func (brt *Handle) setMultipleJobStatus(params setMultipleJobStatusParams) {
...
+			statusParameters := params.AsyncOutput.ImportingParameters
+			if jobImportingParameters, ok := params.AsyncOutput.JobImportingParameters[jobId]; ok {
+				// copy the shared ImportingParameters before mutating, so each job's
+				// metadata does not leak into the others
+				base := append([]byte(nil), params.AsyncOutput.ImportingParameters...)
```
achettyiitr:
```suggestion
				base := bytes.Clone(params.AsyncOutput.ImportingParameters)
```

router/batchrouter/handle_async.go
```diff
@@ -745,6 +746,29 @@ func (brt *Handle) setMultipleJobStatus(params setMultipleJobStatusParams) {
...
+				if merged, setErr := sjson.SetBytes(base, "metadata", jobImportingParameters); setErr != nil {
+					brt.logger.Errorn("[Batch Router] Failed to persist per-job importing metadata on status params",
+						obskit.DestinationType(brt.destType),
+						logger.NewIntField("jobId", jobId),
+						obskit.Error(setErr),
+					)
+				} else {
+					statusParameters = merged
+				}
```
achettyiitr: nitpick:
```suggestion
				merged, mergedErr := sjson.SetBytes(base, "metadata", jobImportingParameters)
				if mergedErr != nil {
					brt.logger.Errorn("[Batch Router] Failed to persist per-job importing metadata on status params",
						obskit.DestinationType(brt.destType),
						logger.NewIntField("jobId", jobId),
						obskit.Error(mergedErr),
					)
				} else {
					statusParameters = merged
				}
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/types.go
```diff
@@ -79,6 +85,50 @@ type ObjectInfo struct {
...
+	IdentifierType string `json:"identifierType"`
+	Type           string `json:"type"`
+}
+
+// SalesforceJobMetadata is the metadata this connector's Transform writes for
+// each event and reads back during Upload.
+type SalesforceJobMetadata struct {
+	JobID           int64                  `json:"job_id"`
+	RudderOperation string                 `json:"rudderOperation"`
+	ExternalID      []SalesforceExternalID `json:"externalId"`
```
achettyiitr:
```suggestion
	ExternalIDs     []SalesforceExternalID `json:"externalId"`
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -150,34 +160,63 @@ func (s *Uploader) readJobsFromFile(filePath string) ([]common.AsyncJob, error)
...
+	for _, job := range jobs {
+		if len(job.Metadata.ExternalID) == 0 || job.Metadata.ExternalID[0].ID == "" {
+			abortedJobIDs = append(abortedJobIDs, job.Metadata.JobID)
+			continue
+		}
+		validJobs = append(validJobs, job)
+	}
+	if len(abortedJobIDs) > 0 {
+		s.logger.Infon("Aborting events with missing externalId", logger.NewIntField("abortedJobs", int64(len(abortedJobIDs))))
+	}
```
achettyiitr: Thinking back on this, is this even valid because during Transform, we can error during extract VDM information and return error.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -186,18 +225,29 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
+	// in-memory state, no payload reads).
+	jobImportingParameters := make(map[int64]stdjson.RawMessage, len(validJobs))
+	for _, job := range validJobs {
+		md, err := jsonrs.Marshal(importingMetadata{ExternalIDHash: hashExternalID(job.Metadata.ExternalID[0].ID)})
+		if err != nil {
+			s.logger.Errorn("marshalling importing metadata", logger.NewIntField("jobID", job.Metadata.JobID), obskit.Error(err))
+			continue
 		}
+		jobImportingParameters[job.Metadata.JobID] = md
 	}
```
achettyiitr: Move this post job creation, no need to create it if the createJob fails.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -405,33 +456,44 @@ func (s *Uploader) matchRecordsToJobs(
...
-		metadata.FailedReasons = lo.SliceToMap(missingJobIDs, func(jobID int64) (int64, string) {
-			return jobID, "Input hash is not found in the data hash to job ID map or the job is not found in the successRespons or failedResponse"
-		})
+		matchedJobIDs := lo.Concat(metadata.SucceededKeys, metadata.AbortedKeys)
+		missingJobIDs, _ := lo.Difference(importingJobIDs, matchedJobIDs)
+
+		// Abort the unmatched jobs. Re-uploading would only re-upsert records
+		// already in Salesforce, so retrying is pointless here.
+		reason := resultCorrelationFailedReason
+		if len(externalIDHashToJobID) == 0 {
```
achettyiitr: This is not needed right? Since we are storing this, we should be able to get this? Maybe once we calculate externalIDHashToJobID and it's empty, we can return early itself because everything needs to be marked as aborted.

router/batchrouter/handle_async.go
ItsSudip: I have added it as an example to understand the structure.

achettyiitr: Let's not use it because, integrations-wise, the structure would change and becomes invalid. Having the test to understand would be sufficient.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — prefixed all three with `brt_async_dest_`: `brt_async_dest_payload_size`, `brt_async_dest_events_per_file`, `brt_async_dest_async_upload_time`. Also updated the matching bucket keys in `runner/buckets.go` (added a prefixed `payload_size` entry rather than renaming the shared one still used by the other async destinations).

router/batchrouter/handle_async.go
ItsSudip: Good point — it's framework-level, so removed the destination-specific example and left only a generic note that the `metadata` shape is destination-specific.

router/batchrouter/handle_async.go
ItsSudip: Done — using `bytes.Clone(...)`.

router/batchrouter/handle_async.go
ItsSudip: Done — adopted the suggested block; named the error `mergedErr`, with the error log and fallback to the shared params on failure.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/types.go
ItsSudip: Done — renamed to `ExternalIDs` (kept the `externalId` json tag).

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Right — removed the per-event externalId check in `Upload` entirely. Validation now happens at `Transform` via `extractFromVDM`, which errors on empty `id`/`type`/`identifierType`, so events without a valid externalId never reach `Upload`.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — the `jobImportingParameters` map is now built after the Salesforce job is created and closed, so we skip the work when the upload fails.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — when `externalIDHashToJobID` is empty we now return early, marking the whole batch aborted with `noCorrelationMetadataReason` (plus the loud error log for alerting). The reconciliation block below now only handles partial correlation failures.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk_upload_test.go
```diff
@@ -22,6 +25,13 @@ import (
 	salesforcebulkupload "github.com/rudderlabs/rudder-server/router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload"
 )
 
+// extIDHash mirrors the connector's externalId hashing so tests can assert the
+// hashed value persisted in the importing job status params.
+func extIDHash(externalID string) string {
+	sum := sha256.Sum256([]byte(externalID))
+	return hex.EncodeToString(sum[:])
+}
```
achettyiitr: Instead of duplicating it, can we keep the hashExternalID exported?

achettyiitr (approved):

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk_upload_test.go
ItsSudip: Addressed.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -60,13 +64,21 @@ func NewUploader(
...
-		statsFactory:    statsFactory,
-		destName:        destName,
+		logger:       logger,
+		apiService:   apiService,
+		destination:  destination,
+		statsFactory: statsFactory,
+		destName:     destName,
+	}
+	statTags := stats.Tags{
+		"module":        "batch_router",
```
atzoum: `module` is not needed is it?
```suggestion
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -186,18 +202,16 @@ func (s *Uploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestin
...
-			FailedCount:   len(importingJobIDs),
-			DestinationID: destinationID,
-		}
+		return s.failedJobs(asyncDestStruct, fmt.Sprintf("creating CSV: %v", err))
 	}
-	s.dataHashToJobID = hashToJobID
+
+	// Track the size of the CSV we send to Salesforce and how many events it packs.
+	s.eventsPerFileStat.Observe(float64(len(jobs)))
+	s.payloadSizeStat.Observe(float64(fileSize))
```
atzoum, line 209: is casting necessary?
```suggestion
	s.eventsPerFileStat.Observe(len(jobs))
	s.payloadSizeStat.Observe(fileSize)
```

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -76,19 +88,14 @@ func (s *Uploader) Transform(job *jobsdb.JobT) (string, error) {
...
-	if externalID.Exists() {
-		if err := jsonrs.Unmarshal([]byte(externalID.Raw), &externalIdArray); err == nil {
-			metadata["externalId"] = externalIdArray
-		}
+	// externalId is mandatory — it is the upsert key. Fail the event up front if
+	// it is absent rather than letting it flow downstream.
+	if !externalID.Exists() {
+		return "", errExternalIDRequired
+	}
+	var externalIDs []SalesforceExternalID
```
atzoum, line 95: I don't understand why we are using an array, while we always assume this array to have a single value in it

atzoum (commented):
How urgent is this? Can't it be included in next week's release? Do we need a patch release?

ItsSudip:
> How urgent is this? Can't it be included in next week's release? Do we need a patch release?

The customer is waiting for this fix. We already mentioned today as the release date.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: Done — removed the `module` tag. With the new `brt_async_dest_` prefix the metrics are already scoped to batch_router async destinations, so the tag was redundant. Updated the metric test too.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: `stats.Histogram.Observe` is `Observe(value float64)`, and `len(jobs)` is an `int` while `fileSize` is an `int64`, so the explicit `float64(...)` conversion is required for it to compile — Go doesn't convert numeric types implicitly. Kept the casts.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: `context.externalId` is delivered as an array in RudderStack's VDM (mapped-to-destination) format — that's the platform-wide wire shape every VDM destination receives, so `[]SalesforceExternalID` mirrors the incoming JSON as-is. We index `[0]` because there's conventionally a single entry, but the field stays an array so the payload unmarshals directly; a scalar would diverge from the VDM contract and need custom unmarshalling.

atzoum: why does vdm map to an array?

atzoum (approved):

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
ItsSudip: We were supporting externalId for some event stream use case, you can see it [here](https://www.rudderstack.com/docs/sources/event-streams/sdks/mobile-sdk-apis/event-payload-customization-options/#externalid). When we implemented vdm, we used the externalId as an array.

PR #7085, opened 2026-06-12
fix: abort non-retryable Klaviyo upload errors instead of retrying

shekhar-rudder:
## Summary
- Klaviyo 4xx errors (except 429) were incorrectly marked as retryable (`FailedJobIDs`), causing invalid profiles to retry indefinitely, fill batch slots, and push valid events into "prescribed limit" overflow until they were permanently aborted
- `UploadProfiles` now captures the HTTP status code from Klaviyo's response
- `Upload` routes errors by retryability: 4xx (except 429) → `AbortJobIDs` (terminal), 429/5xx/network errors → `FailedJobIDs` (retryable)
- `FailedReason` is now populated in the upload output for better observability

## Test plan
- [x] `TestUpload/Successful_Upload` — all jobs import successfully, no failures
- [x] `TestUpload/Unsuccessful_Upload` — nil response (network error) → jobs retried with `FailedReason`
- [x] `TestUpload/UploadThreeChunksWithMixedFailures` — chunk 1 gets 429 (retried), chunk 2 gets 400 (aborted), chunk 3 succeeds — validates both paths in single test
- [x] Existing Poll and GetUploadStats tests unaffected

## Linear ticket
INT-6539

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
+++ b/router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
@@ -361,7 +361,9 @@ func (kbu *KlaviyoBulkUploader) Upload(_ context.Context, asyncDestStruct *commo
 		FailedCount:         len(failedJobs),
 		AbortJobIDs:         abortedJobs,
 		AbortReason:         abortReason,
+		AbortCount:          len(abortedJobs),
 		ImportingJobIDs:     successJobs,
+		ImportingCount:      len(successJobs),
 		DestinationID:       destination.ID,
 	}
 }
--- a/router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
+++ b/router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"net/http"
 	"os"
 	"strconv"
 	"strings"
@@ -258,6 +259,7 @@ func (kbu *KlaviyoBulkUploader) ExtractProfile(Data Data) Profile {
 func (kbu *KlaviyoBulkUploader) Upload(_ context.Context, asyncDestStruct *common.AsyncDestinationStruct) common.AsyncUploadOutput {
 	destination := asyncDestStruct.Destination
 	var failedJobs []int64
+	var failedReason string
 	var abortedJobs []int64
 	var abortReason string
 	var successJobs []int64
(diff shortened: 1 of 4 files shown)


router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
```diff
@@ -339,15 +351,16 @@ func (kbu *KlaviyoBulkUploader) Upload(_ context.Context, asyncDestStruct *commo
...
 		return kbu.generateKlaviyoErrorOutput("Error while marshaling parameters.", err, importingJobIDs, destinationID)
 	}
-	successJobs, _ = lo.Difference(importingJobIDs, failedJobs)
+	successJobs, _ = lo.Difference(importingJobIDs, append(failedJobs, abortedJobs...))
 	eventsSuccessStat.Count(len(asyncDestStruct.ImportingJobIDs))
 
 	return common.AsyncUploadOutput{
 		ImportingParameters: importParameters,
 		FailedJobIDs:        failedJobs,
+		FailedReason:        failedReason,
```
achettyiitr, line 360: Can we add AbortCount and ImportingCount as well.

router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload_test.go
```diff
@@ -100,12 +100,21 @@ func TestUpload(t *testing.T) {
 		}
 
 		output := uploader.Upload(context.Background(), asyncDestStruct)
-		assert.NotNil(t, output)
-		assert.Equal(t, destination.ID, output.DestinationID)
-		assert.Empty(t, output.FailedJobIDs)
-		assert.Empty(t, output.AbortJobIDs)
-		assert.Empty(t, output.AbortReason)
-		assert.NotEmpty(t, output.ImportingJobIDs)
+		assert.Equal(t, common.AsyncUploadOutput{
```
achettyiitr: Can we use require instead of assert?

router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload_test.go
```diff
@@ -100,12 +100,21 @@ func TestUpload(t *testing.T) {
...
+			ImportingJobIDs:     []int64{1},
+			ImportingParameters: output.ImportingParameters,
```
achettyiitr, line 106: This doesn't feel correct, we are adding the expected output here to match the actual output, can we verify individual fields?

router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
shekhar-rudder: Done. Added `AbortCount: len(abortedJobs)` and `ImportingCount: len(successJobs)` to the return value, and updated test expectations accordingly.

router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload_test.go
shekhar-rudder: Done. Switched the struct comparison assertions from `assert.Equal` to `require.Equal` in the Upload tests.

router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload_test.go
shekhar-rudder: Kept the single struct comparison for a concise assertion, but the `ImportingParameters` is verified separately in a dedicated `require.Equal` assertion right after. The self-reference in the struct is only for `ImportingParameters` since it's a marshaled JSON blob validated independently.

achettyiitr: Maybe we can do something like this, comparing default values does not give much value. Also, if we add new properties to the AsyncUploadOutput, all these tests would break.

```
        require.Equal(t, 1, output.ImportingCount)
		require.Equal(t, []int64{1}, output.ImportingJobIDs)
		require.Equal(t, destination.ID, output.DestinationID)
		require.Equal(t, `{"importId":"importId1","importCount":0}`, string(output.ImportingParameters))
```

shekhar-rudder: > Also, if we add new properties to the AsyncUploadOutput, all these tests would break.

If we add a new field, I think it's fine if the tests fail initially. We should update the existing tests to verify that the new field is being populated correctly.

achettyiitr: IMO, this would require all tests to be updated, which is not needed. The same goes for removing something as well.

shekhar-rudder: > IMO, this would require all tests to be updated, which is not needed.

Wouldn't it be preferable to ensure the new field is being set correctly in the existing tests as well? My guess is that we might have caught missing `FailedReason` if we were asserting the whole response. Also the output here is `Upload` method signature. Since that is part of interface, it is less likely to change.

achettyiitr (approved):

ItsSudip (approved):

PR #7093, opened 2026-06-16
chore(jobsdb): multi-consumer-specific indexes and views

atzoum:
# Description

Adds the `WithMultiConsumer()` constructor option and branches the per-dataset index/view DDL on it.

**Single-consumer JobsDBs** are completely unchanged: they keep the existing `(job_id ASC, id DESC, job_state)` pickup index and the `v_last_<status>` view (`DISTINCT ON (job_id)`).

**Multi-consumer JobsDBs** get a different DDL set per dataset:
- GIN index on `consumers` — serves the `@>` array-containment membership filter used by consumer-scoped pickup queries.
- Composite status index `(job_id ASC, consumer, id DESC, job_state)` — one column wider than the single-consumer index, makes the per-consumer latest-status computation an index-ordered scan.
- `v_last_c_<status>` view — `DISTINCT ON (job_id, consumer)`, returns one latest-status row per *(job, consumer)* pair.

The retention index `(id, job_state) INCLUDE (exec_time)` is consumer-agnostic and created for all handles.

Zero runtime overhead for single-consumer handles: they never create the GIN index, the wider status index, or the `v_last_c_` view.

## Linear Ticket

resolves PIPE-3058

## Security

### DIFF
--- a/jobsdb/jobsdb_dataset_ddl.go
+++ b/jobsdb/jobsdb_dataset_ddl.go
@@ -212,22 +212,41 @@ func (jd *Handle) createDSJobIndicesInTx(ctx context.Context, tx *Tx, newDS data
 			return fmt.Errorf("creating %s index: %w", param, err)
 		}
 	}
+	if jd.conf.multiConsumer {
+		if _, err := tx.ExecContext(ctx, fmt.Sprintf(`CREATE INDEX "idx_%[1]s_consumers" ON %[1]q USING GIN (consumers)`, newDS.JobTable)); err != nil {
+			return fmt.Errorf("creating consumers GIN index: %w", err)
+		}
+	}
 	return nil
 }
 
 // createDSStatusIndicesInTx creates the indices on the job status table plus the
-// v_last view.
+// v_last view. On multi-consumer handles it also creates the v_last_c_ view and
+// the wider (job_id, consumer, id DESC, job_state) pickup index.
+// v_last_ is always created so that a soft downgrade (disabling WithMultiConsumer without
+// the guard) finds the view and can at least start — callers accept single-consumer semantics
+// during a downgrade window.
 func (jd *Handle) createDSStatusIndicesInTx(ctx context.Context, tx *Tx, newDS dataSetT) error {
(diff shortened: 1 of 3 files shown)


ktgowtham (approved):

jobsdb/jobsdb_dataset_ddl.go
```diff
@@ -212,22 +212,41 @@ func (jd *Handle) createDSJobIndicesInTx(ctx context.Context, tx *Tx, newDS data
...
 	if _, err := tx.ExecContext(ctx, fmt.Sprintf(`CREATE VIEW "v_last_%[1]s" AS SELECT DISTINCT ON (job_id) * FROM %[1]q ORDER BY job_id ASC, id DESC`, newDS.JobStatusTable)); err != nil {
-		return fmt.Errorf("create view: %w", err)
+		return fmt.Errorf("creating v_last view: %w", err)
+	}
+
+	if jd.conf.multiConsumer {
+		if _, err := tx.ExecContext(ctx, fmt.Sprintf(`CREATE INDEX "idx_%[1]s_jid_c_id_js" ON %[1]q (job_id ASC, consumer, id DESC, job_state)`, newDS.JobStatusTable)); err != nil {
+			return fmt.Errorf("adding jid_c_id_js index: %w", err)
+		}
+		if _, err := tx.ExecContext(ctx, fmt.Sprintf(`CREATE VIEW "v_last_c_%[1]s" AS SELECT DISTINCT ON (job_id, consumer) * FROM %[1]q ORDER BY job_id ASC, consumer, id DESC`, newDS.JobStatusTable)); err != nil {
```
mihir20, line 243:
```suggestion
		if _, err := tx.ExecContext(ctx, fmt.Sprintf(`CREATE VIEW "v_last_consumer_%[1]s" AS SELECT DISTINCT ON (job_id, consumer) * FROM %[1]q ORDER BY job_id ASC, consumer, id DESC`, newDS.JobStatusTable)); err != nil {
```
can we rename the view? c can be interpreted differently if someone do not have much context?

atzoum: We need to keep names as short as possible, we only have 63 characters available `v_last_c_batch_rt_job_status_` is already 29, which leaves 34 (30 if we take into account compacted table naming) digits for dataset indexes.

mihir20 (approved):

PR #7095, opened 2026-06-16
chore(jobsdb): multi-consumer get and noresults cache handling

atzoum:
# Description

**Consumer-scoped pickup** (`getJobsDS`): 
On a multi-consumer handle, the query adds a GIN membership pre-filter (`jobs.consumers @> ARRAY[$n]`) and scopes the latest-status join to that consumer. Both the view-join strategy (`v_last_c_` ON `job_id AND consumer = $n`) and the lateral-join strategy (`ORDER BY id DESC LIMIT 1` with `AND consumer = $n`) get the consumer-scoped variant. Single-consumer handles ignore the field entirely; their query plans are unchanged.

**`noResultsCache` consumer dimension**: 
The consumer participates in the cache as a `consumer=<value>` parameter-filter key entry, reusing the existing wildcard machinery, no new cache dimension is added. Consumer-scoped lookups key on `consumer=<value>`; all-consumer queries key on `consumer=*`. Invalidation on a status write for consumer X invalidates both `consumer=X` and `consumer=*`. Store writes invalidate only the consumers stored in that batch (plus `*`).

## Linear Ticket

resolves PIPE-3050

## Security

### DIFF
--- a/jobsdb/jobsdb_consumer_test.go
+++ b/jobsdb/jobsdb_consumer_test.go
@@ -96,6 +96,80 @@ func TestMultiConsumerJobsDB_SingleConsumer(t *testing.T) {
 	verifyUnionJobsDBMetadata(t, postgres.DB, prefix, 3)
 }
 
+// TestMultiConsumerJobsDB_ConsumerIsolation verifies that consumer-scoped pickup is isolated:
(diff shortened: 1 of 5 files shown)


ktgowtham (approved):

mihir20 (approved):
LGTM

PR #7096, opened 2026-06-16
chore(jobsdb): multi-consumer compaction

atzoum:
# Description

Updates the dataset compaction paths for supporting the multi-consumer scenarios.

**Consumer trimming**: when carrying a job forward into the destination dataset, already-terminal consumers are stripped from its `consumers` array and only the surviving (still-pending) consumers' status rows are copied.

**Status-table cleanup amplification**: the periodic cleanup (`cleanupStatusTables`) triggers on a statuses-per-job ratio and a table-size threshold that assume one status row per job at steady state. For multi-consumer handles the healthy baseline is one row per *(job, consumer)*, so both thresholds are scaled by the dataset's registry cardinality. A `skipMultiConsumerStatusCompaction` config key (default `true` for multi-consumer JobsDBs) allows for status compaction to be skipped entirely. Dataset-level compaction (move + drop) is unaffected.

## Linear Ticket

resolves PIPE-3052

## Security

---
This is **part 1 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;5&nbsp;</kbd> #7122 
- <kbd>&nbsp;4&nbsp;</kbd> #7121 
- <kbd>&nbsp;3&nbsp;</kbd> #7098 
- <kbd>&nbsp;2&nbsp;</kbd> #7097 
- <kbd>&nbsp;1&nbsp;</kbd> #7096 👈


jobsdb/jobsdb_compaction.go
```diff
@@ -345,16 +345,18 @@ func (jd *Handle) getVacuumFullCandidates(ctx context.Context, dsList []dataSetT
 	}
 
 	toVacuumFull := []string{}
+	threshold := jd.conf.compaction.vacuumFullStatusTableThreshold.Load()
 	for _, ds := range dsList {
-		if tableSizes[ds.JobStatusTable] > jd.conf.compaction.vacuumFullStatusTableThreshold.Load() {
+		scaledThreshold := threshold * max(dsConsumerCounts[ds.Index], 1)
```
mihir20, line 350: Is it worth mentioning that ConsumerCounts is the total unique number of the consumers across all the jobs, as some jobs might have a different number of consumers than others?

jobsdb/jobsdb_compaction.go
```diff
@@ -396,23 +398,38 @@ func (jd *Handle) getCleanUpCandidates(ctx context.Context, dsList []dataSetT) (
 			if jobs == 0 { // using max ds size if we have no stats for the number of jobs
 				jobs = float64(jd.conf.MaxDSSize.Load())
 			}
-			return statuses/jobs > jd.conf.compaction.jobStatusCompactionThres.Load()
+			// For multi-consumer datasets, one status row per (job, consumer) is the healthy baseline,
+			// so scale the denominator by the consumer count to preserve threshold semantics.
```
mihir20, line 402: same here

mihir20 (approved):

jobsdb/jobsdb_compaction.go
atzoum: yes, job status compaction is not efficient now for multi-consumer jobsdbs, but thankfully it is not needed for any of our current use cases

ktgowtham (approved):

PR #7101, opened 2026-06-17
fix(router): shared barrier causing illegal job sequence errors

atzoum:
## Description

A single `*eventorder.Barrier` was shared across all partition workers for a given destination type. Each partition worker calls `barrier.Sync()` at the start of its pickup loop, so one partition worker's `Sync()` could clear `failedJobID` for a key while another partition worker was mid-pickup, after that worker's own `Sync()` had already run. This allowed a higher-JobID job to be admitted into pickup before a lower-JobID job for the same orderKey had completed, producing an illegal sequence error/panic.

Each `partitionWorker` now creates its own `*eventorder.Barrier`. `Sync()` is now only called by the partition worker that owns that barrier, restoring the snapshot guarantee.

### Secondary
Illegal job sequence panics are now configurable, they can be skipped in favour of incrementing a counter instead.

## Linear Ticket

resolves PIPE-2747

## Security

### DIFF
--- a/router/internal/eventorder/eventorder.go
+++ b/router/internal/eventorder/eventorder.go
@@ -64,6 +64,22 @@ func WithOrderingDisabledCheckForBarrierKey(orderingDisabledForKey func(key Barr
 	}
 }
 
+// WithPanicOnIllegalJobSequence controls whether an illegal job sequence causes a panic.
+// When disabled (default), out-of-order jobs are treated as normal and the illegalSequenceCallback is invoked instead.
+func WithPanicOnIllegalJobSequence(panicOnIllegalSequence config.ValueLoader[bool]) OptFn {
+	return func(b *Barrier) {
+		b.panicOnIllegalSequence = panicOnIllegalSequence
+	}
+}
+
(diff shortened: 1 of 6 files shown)


router/internal/eventorder/eventorder.go
```diff
@@ -227,15 +252,20 @@ func (b *Barrier) Wait(key BarrierKey, jobID int64) (wait bool, previousFailedJo
...
+					jobID,
+					!barrier.stateTime.IsZero(),
+					debugInfo))
+			}
+			if b.illegalSequenceCallback != nil {
+				b.illegalSequenceCallback("wait")
+			}
 		}
-		return jobID > failedJob, &failedJob // wait if this is not the failed job
+		return jobID != failedJob, &failedJob // wait if this is not the failed job
```
mihir20, line 268: In what cases can a job smaller than a failed job arrive again?

mihir20 (approved):
I have a doubt; otherwise, LGTM

router/internal/eventorder/eventorder.go
atzoum: There shouldn't be any other case, that's why we are retaining the panic by default

ktgowtham (approved):
great find 👍

PR #7106, opened 2026-06-20
fix(batchrouter): file descriptor leak in upload() — outputFile never closed

### LINKED ISSUE #7071: fix(batchrouter): file descriptor leak in upload() — outputFile never closed
## Bug Description

In `router/batchrouter/handle.go`, the `upload()` function opens a gzip file for uploading to object storage but **never closes the file descriptor**, regardless of whether the upload succeeds or fails.

## Location

`router/batchrouter/handle.go` — function `upload()`, line 384

## Code

```go
outputFile, err := os.Open(gzipFilePath)
if err != nil {
    panic(err)
}
// ❌ no defer outputFile.Close() here

// ... ~70 lines of code ...

uploadOutput, err := uploader.Upload(context.TODO(), outputFile, keyPrefixes...)
if err != nil {
    return UploadResult{Error: err, ...}  // fd leaked
}

return UploadResult{...}  // fd also leaked on success
```

## Impact

`upload()` is called on every batch job cycle for every active destination. Each call leaks one OS file descriptor. Under sustained load this will exhaust the process's fd limit (`ulimit -n`), causing new `os.Open` / socket calls to fail with `too many open files` — which surfaces as dropped events or a crashed process.

## Expected Behaviour

The file descriptor should be closed after the upload completes (success or failure).

## Suggested Fix

Add a `defer` immediately after the `os.Open` call:

```go
outputFile, err := os.Open(gzipFilePath)
if err != nil {
    panic(err)
}
defer func() { _ = outputFile.Close() }()
```

This is already the correct pattern used in the sibling file `archiver/worker.go` (line 187):

```go
file, err := os.Open(filePath)
if err != nil {
    return "", fmt.Errorf("open file %s: %w", filePath, err)
}
defer func() { _ = file.Close() }()
```

## Environment

- File: `router/batchrouter/handle.go`
- Function: `upload()` (lines 240–483)
- Affected: all batch router destinations (warehouse staging + raw data uploads)

kuntal1461:
Closes #7071

## Summary

`outputFile` is opened via `os.Open(gzipFilePath)` in `Handle.upload()` but never closed — leaking one file descriptor on every batch upload call, on both the success and error return paths.

## Fix

Added `defer func() { _ = outputFile.Close() }()` immediately after `os.Open`, matching the existing pattern in `archiver/worker.go:187`.

```go
outputFile, err := os.Open(gzipFilePath)
if err != nil {
    panic(err)
}
defer func() { _ = outputFile.Close() }()  // ← added
```

## Impact without fix

`upload()` is called on every batch job cycle for every active destination. Each call leaks one fd. Under sustained load this exhausts the process fd limit (`ulimit -n`), causing `too many open files` errors — dropped events or a crashed process.

## Test plan

- [ ] `go test ./router/batchrouter/...` passes
- [ ] No new linter warnings

### DIFF
--- a/router/batchrouter/handle.go
+++ b/router/batchrouter/handle.go
@@ -385,6 +385,7 @@ func (brt *Handle) upload(provider string, batchJobs *BatchedJobs, isWarehouse b
 	if err != nil {
 		panic(err)
 	}
+	defer outputFile.Close()
 
 	brt.logger.Debugn("BRT: Starting upload to", logger.NewStringField("provider", provider))
 	var folderName string
--- a/router/batchrouter/handle.go
+++ b/router/batchrouter/handle.go
@@ -385,7 +385,7 @@ func (brt *Handle) upload(provider string, batchJobs *BatchedJobs, isWarehouse b
 	if err != nil {
 		panic(err)
 	}
-	defer outputFile.Close()
+	defer func() { _ = outputFile.Close() }()
 
 	brt.logger.Debugn("BRT: Starting upload to", logger.NewStringField("provider", provider))
 	var folderName string


Rebaka8 (approved):
Verified that the deferred outputFile.Close() is placed immediately after os.Open(), ensuring the file descriptor is released on both successful and error execution paths. The implementation is minimal, addresses the reported resource leak, and follows the existing resource cleanup pattern used elsewhere in the project. Looks good to me.

PR #7107, opened 2026-06-22
fix: klaviyo upload partial failures

achettyiitr:

# Description

Klaviyo's Bulk Profile Import API validates each request **all-or-nothing**: a single profile that fails synchronous validation (e.g. a malformed email) causes Klaviyo to reject the entire batch with HTTP 400, importing **zero** profiles. Today rudder-server mirrors that by aborting the whole chunk, so one bad record takes down every valid profile alongside it.

This PR adds **strip-and-retry on failure**. When Klaviyo returns a 400 whose error pointers identify specific profiles, we abort only those jobs and re-upload the chunk **once** with the remaining valid profiles, so the good records still get delivered.

### Behavior

- **Success** → jobs imported as before.
- **400 that pinpoints specific profiles** → those jobs are aborted (with Klaviyo's per-profile reason), the rest are re-uploaded once; the retry's outcome is final (no looping).
- **400 we can't attribute to a profile** (e.g. invalid `listId`) → whole chunk aborted, unchanged.
- **5xx / 429 / transport errors** → whole chunk marked failed for the batch router to retry, unchanged.

The retry path is bounded to a single extra attempt — there is no loop.

### Implementation

- `uploadChunk` orchestrates the try → strip → retry-once flow.
- `tryUpload` performs one upload attempt and normalizes the status/error.
- `strippableProfiles` decides whether a failure is a per-profile 400 we can safely retry.
- `extractInvalidProfileIndices` parses the offending profile indices from Klaviyo's JSON pointers.
- `recordError` classifies remaining jobs as aborted (4xx) vs failed (5xx/429).
- `chunkUploadResult` (added to `types.go`) carries the per-chunk outcome.

## Linear Ticket

- Resolves INT-6600

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
+++ b/router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -28,6 +29,10 @@ const (
 	MAXPAYLOADSIZE        = 4600000
 	IMPORT_ID_SEPARATOR   = ":"
 )
+
+// profileIndexRegex extracts the profile index from a Klaviyo error pointer such
+// as "/data/attributes/profiles/data/3/attributes/email".
+var profileIndexRegex = regexp.MustCompile(`/profiles/data/(\d+)`)
 
 func createFinalPayload(combinedProfiles []Profile, listId string) Payload {
 	payload := Payload{
@@ -316,33 +321,18 @@ func (kbu *KlaviyoBulkUploader) Upload(_ context.Context, asyncDestStruct *commo
 	var importIds []string // DelimitedImportIds is : separated importIds
 
 	for idx, profileChunk := range profileChunks {
-		combinedPayload := createFinalPayload(profileChunk, listId)
-		uploadResp, err := kbu.KlaviyoAPIService.UploadProfiles(combinedPayload)
-		if err != nil {
-			var uploadError string
-			var statusCode int
-			if uploadResp != nil {
-				statusCode = uploadResp.StatusCode
-				if len(uploadResp.Errors) > 0 {
-					uploadError = uploadResp.Errors.String()
-				}
-			}
-			kbu.Logger.Errorn("Error while uploading profiles",
-				obskit.Error(err),
-				obskit.DestinationID(destinationID),
-				logger.NewStringField("uploadErrors", uploadError))
-
-			if statusCode >= http.StatusBadRequest && statusCode < http.StatusInternalServerError && statusCode != http.StatusTooManyRequests {
-				abortedJobs = append(abortedJobs, jobIDChunks[idx]...)
-				abortReason = fmt.Sprintf("upload rejected by Klaviyo with status %d: %s", statusCode, uploadError)
-			} else {
-				failedJobs = append(failedJobs, jobIDChunks[idx]...)
-				failedReason = fmt.Sprintf("upload failed with status %d: %s", statusCode, uploadError)
-			}
-			continue
+		chunkResult := kbu.uploadChunk(profileChunk, jobIDChunks[idx], listId, destinationID)
+		if len(chunkResult.abortedJobs) > 0 {
+			abortedJobs = append(abortedJobs, chunkResult.abortedJobs...)
+			abortReason = chunkResult.abortReason
 		}
-
-		importIds = append(importIds, uploadResp.Data.Id)
+		if len(chunkResult.failedJobs) > 0 {
+			failedJobs = append(failedJobs, chunkResult.failedJobs...)
+			failedReason = chunkResult.failReason
+		}
+		if chunkResult.importID != "" {
+			importIds = append(importIds, chunkResult.importID)
+		}
 	}
 	importParameters, err := jsonrs.Marshal(common.ImportParameters{
 		ImportId:    strings.Join(importIds, IMPORT_ID_SEPARATOR),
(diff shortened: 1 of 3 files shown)


router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
```diff
@@ -368,6 +358,134 @@ func (kbu *KlaviyoBulkUploader) Upload(_ context.Context, asyncDestStruct *commo
...
+// error could be attributed to a specific profile; if any error is structural
+// (e.g. an invalid list ID, which has no profile pointer) it returns false so the
+// caller can fall back to aborting the whole chunk instead of retrying blindly.
+func extractInvalidProfileIndices(errs ErrorDetailList) (map[int]string, bool) {
+	indices := make(map[int]string)
+	allAttributed := true
+	for _, e := range errs {
+		match := profileIndexRegex.FindStringSubmatch(e.Source.Pointer)
+		if match == nil {
+			allAttributed = false
```
ItsSudip: When we set allAttributed to false, we can return from there, as we are not updating it to true anywhere later.

router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
```diff
@@ -368,6 +358,134 @@ func (kbu *KlaviyoBulkUploader) Upload(_ context.Context, asyncDestStruct *commo
...
+// detail) only when the failure is a synchronous 400 that attributes every error
+// to a specific profile. For any other failure (transport/5xx/429, a 4xx we can't
+// attribute, or no per-profile errors) it returns ok=false, telling the caller to
+// abort/fail the whole chunk rather than retry.
+func strippableProfiles(resp *UploadResp, statusCode int) (map[int]string, bool) {
+	if resp == nil || !isClientError(statusCode) {
+		return nil, false
+	}
+	indices, allAttributed := extractInvalidProfileIndices(resp.Errors)
+	if !allAttributed || len(indices) == 0 {
```
ItsSudip: if allAttributed is true, then len(indices) can't be 0 right?

ItsSudip (approved):

krishna2020:
**`abortReason` collapses multiple per-profile reasons into one**

In `uploadChunk`, the strip loop overwrites `res.abortReason` on each iteration:

```go
res.abortReason = fmt.Sprintf("profile rejected by Klaviyo's synchronous validation: %s", reason)
```

So when a chunk has more than one invalid profile, only the *last* rejected profile's reason survives into `AsyncUploadOutput.AbortReason` (and the cross-chunk assignment in `Upload` then keeps only the last chunk's value). Since this PR is specifically about per-profile attribution, the per-profile detail is mostly discarded in the user-visible abort reason.

Not a regression (the single-reason field predates this PR), but worth considering aggregating the reasons — e.g. joining `idx -> reason` pairs, or at least appending rather than overwriting — so a chunk with several bad profiles doesn't surface an arbitrary single reason.

router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
```diff
@@ -368,6 +358,127 @@ func (kbu *KlaviyoBulkUploader) Upload(_ context.Context, asyncDestStruct *commo
...
+	var res chunkUploadResult
+
+	// First attempt with the full chunk.
+	resp, statusCode, detail, ok := kbu.tryUpload(profiles, listId, destinationID)
+	if ok {
+		res.importID = resp.Data.Id
+		return res
+	}
+
+	badIndices, canStrip := strippableProfiles(resp, statusCode)
```
krishna2020, line 378: Does klaviyo API does the validation synchronously ? In case of errors/issues after poll, they won't get retried ?

ItsSudip:
This looks like a limitation of the framework. Since we return multiple `AbortJobIDs`, we should be able to return `AbortReason` as a map of jobID → reason, so each aborted job can carry its own reason rather than sharing a single one.

router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go
achettyiitr: Yes, some validations happen synchronously and some asynchronously. https://developers.klaviyo.com/en/docs/use_klaviyos_bulk_profile_import_api#validation

achettyiitr: Poll already handles for partial failures.

krishna2020:
**`ImportCount` is always serialized as `0` (pre-existing, but worth fixing here)**

At the marshal site:

```go
importParameters, err := jsonrs.Marshal(common.ImportParameters{
    ImportId:    strings.Join(importIds, IMPORT_ID_SEPARATOR),
    ImportCount: len(successJobs),
})
...
successJobs, _ = lo.Difference(importingJobIDs, append(failedJobs, abortedJobs...))
```

`successJobs` is declared nil and only computed *after* the marshal, so `ImportCount` always serializes as `0`. It's not dead — it's read back during polling (`handle_async.go:106` → `pollInput.ImportCount`) and used in two ways:

- `handle_async.go:361-364` — when `0`, `updatePollStatusToDB` falls back to `maxEventsInABatch` instead of the real count.
- `handle_async.go:369` — the `len(statusList) != ImportCount` mismatch warning is gated on `ImportCount > 0`, so it's effectively disabled for this destination.

Pre-existing (not introduced by this PR), but the fix is a one-liner — move the `successJobs` computation above the marshal so `len(successJobs)` reflects the actual count:

```go
successJobs, _ = lo.Difference(importingJobIDs, append(failedJobs, abortedJobs...))
importParameters, err := jsonrs.Marshal(common.ImportParameters{
    ImportId:    strings.Join(importIds, IMPORT_ID_SEPARATOR),
    ImportCount: len(successJobs),
})
```

achettyiitr:
> **`ImportCount` is always serialized as `0` (pre-existing, but worth fixing here)**
> 
> At the marshal site:
> 
> ```go
> importParameters, err := jsonrs.Marshal(common.ImportParameters{
>     ImportId:    strings.Join(importIds, IMPORT_ID_SEPARATOR),
>     ImportCount: len(successJobs),
> })
> ...
> successJobs, _ = lo.Difference(importingJobIDs, append(failedJobs, abortedJobs...))
> ```
> 
> `successJobs` is declared nil and only computed _after_ the marshal, so `ImportCount` always serializes as `0`. It's not dead — it's read back during polling (`handle_async.go:106` → `pollInput.ImportCount`) and used in two ways:
> 
> * `handle_async.go:361-364` — when `0`, `updatePollStatusToDB` falls back to `maxEventsInABatch` instead of the real count.
> * `handle_async.go:369` — the `len(statusList) != ImportCount` mismatch warning is gated on `ImportCount > 0`, so it's effectively disabled for this destination.
> 
> Pre-existing (not introduced by this PR), but the fix is a one-liner — move the `successJobs` computation above the marshal so `len(successJobs)` reflects the actual count:
> 
> ```go
> successJobs, _ = lo.Difference(importingJobIDs, append(failedJobs, abortedJobs...))
> importParameters, err := jsonrs.Marshal(common.ImportParameters{
>     ImportId:    strings.Join(importIds, IMPORT_ID_SEPARATOR),
>     ImportCount: len(successJobs),
> })
> ```

Seems like an [existing bug](https://github.com/rudderlabs/rudder-server/blob/master/router/batchrouter/asyncdestinationmanager/klaviyobulkupload/klaviyobulkupload.go#L347-L354), let me fix it. Fixed existing test as well.

PR #7108, opened 2026-06-22
chore: recreate stale rsources subscription on shared db

atzoum:
# Description

When rsources logical replication is set up, the shared (subscriber) database holds a subscription per instance, while the local (publisher) database holds the matching replication slot. Today, if that slot disappears — most notably when the local db's persistent volume is deleted and the db is recreated from scratch — the subscription on the shared db survives but is permanently broken: `CREATE SUBSCRIPTION` only ever returns `DuplicateObject` and is skipped, so the slot is never recreated and replication stays silently dead.

This PR makes setup self-healing. Before creating the subscription, we now detect a stale subscription (one whose replication slot no longer exists on the publisher) and drop it so it can be recreated together with a fresh slot. The drop uses `DISABLE` → `SET (slot_name = NONE)` → `DROP` so it doesn't try to contact the publisher for the missing slot.

Also truncates the subscription name to Postgres' 63-byte identifier limit so the catalog lookup matches what the server actually stored (long pod hostnames could otherwise bypass the detection).

## Linear Ticket

resolves PIPE-3134

## Security

### DIFF
--- a/services/rsources/handler.go
+++ b/services/rsources/handler.go
@@ -35,6 +35,18 @@ var ErrInvalidPaginationToken = errors.New("rsources: invalid pagination token")
 // In postgres, the replication slot name can contain lower-case letters, underscore characters, and numbers.
 var replSlotDisallowedChars *regexp.Regexp = regexp.MustCompile(`[^a-z0-9_]`)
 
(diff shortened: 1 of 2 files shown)


ktgowtham (approved):

mihir20 (approved):
LGTM

PR #7119, opened 2026-06-24
feat: monthly active records (MAR) metering pipeline

am6010:
# Description

Adds the **Monthly Active Records (MAR)** metering pipeline to rudder-server, modeled on the existing MTU / tracked-users flow. The feature meters activation records per destination, aggregates them in-app using HyperLogLog, and flushes them through the reporting pipeline.

Key changes:

- **Reporter core** (`enterprise/activationrecords/`): MAR reporter, factory, and noop implementation with gating and HLL settings.
- **Flusher pipeline** (`enterprise/reporting/flusher/`): `ActivationRecordsInAppAggregator` with destination-keyed HLL aggregation, wired into the flusher factory and mediator runner.
- **Processor hook** (`processor/`): meters MAR activation records during pre-transform and strips `context.activation` before storing in the txn store.
- **App wiring** (`app/`): registers the MAR feature and wires both the embedded and processor app handlers.
- **Migration**: adds the `activation_records_reports` table.
- **Tests**: reporter grain/gate/HLL tests, dest-keyed aggregation integration test, end-to-end wire-compat consumer test, and processor strip assertions.

## Review fixes (B1/B2/B3)

Two independent code reviews flagged three blocking issues, now fixed:

- **B1 — `context.activation` leaked to destinations.** The strip mutated the raw `job.EventPayload` bytes, but destination-bound events are separately-parsed `singularEvent` maps. Fixed by stripping `context.activation` per-event in `preprocessStage` (covers the transformer input, `eventsByMessageID`, and the archival `payloadFunc`), unconditionally. The redundant raw-bytes strip is removed. A new processor test captures the destination-bound transformer input and asserts `context.activation` is absent on every event of a multi-event batch.
- **B2 — only `batch[0]` was metered/stripped.** A job batch can hold multiple events; the meter now iterates the full `batch` array (every event metered exactly once), and the per-event strip naturally covers all events.
- **B3 — client-controlled `origin` could abort the store transaction.** The `origin` column is widened from `VARCHAR(64)` to `TEXT`, length-capped defensively, and the MAR write is now best-effort: wrapped in a Postgres `SAVEPOINT` so a failed COPY rolls back only the MAR insert (logging + `activation_records_store_error` stat) instead of failing event delivery.

## Linear Ticket

https://linear.app/rudderstack/issue/PRO-5774

## Security

- [x] Fixes a data-leak where an internal metering namespace (`context.activation`) could reach destination payloads.


enterprise/activationrecords/records_reporter.go
```diff
@@ -0,0 +1,287 @@
...
+			continue
+		}
+
+		sourceID := jsonparser.GetStringOrEmpty(job.Parameters, "source_id")
+		destinationID := jsonparser.GetStringOrEmpty(job.Parameters, "destination_id")
+		if sourceID == "" || destinationID == "" {
+			u.log.Warnn("source_id or destination_id not found in job parameters",
+				obskit.WorkspaceID(job.WorkspaceId),
+				logger.NewIntField("jobId", job.JobID))
+			u.stats.NewTaggedStat("activation_records_skipped", stats.CountType, stats.Tags{
```
mihir20: we can also add this metric when skipping when no workspaceID and make the reason granula for source, destination and workspace missing

enterprise/activationrecords/records_reporter.go
```diff
@@ -0,0 +1,287 @@
...
+		key := recordKey{
+			workspaceID:   job.WorkspaceId,
+			sourceID:      sourceID,
+			destinationID: destinationID,
+		}
+
+		// Extract the raw batch array bytes, then iterate each element.
+		batchRaw := jsonparser.GetValueOrEmpty(job.EventPayload, "batch")
+		if len(batchRaw) == 0 {
+			continue
```
mihir20, line 189: we can also emit `activation_records_skipped` here as these are the condition that should not happen in prod. we can setup alert on these

enterprise/activationrecords/records_reporter.go
```diff
@@ -0,0 +1,287 @@
...
+		if err := jsonrs.Unmarshal(batchRaw, &batchElements); err != nil {
+			// batch field absent or not an array — skip job gracefully.
+			continue
+		}
+		for _, elem := range batchElements {
+			fingerprint := jsonparser.GetStringOrEmpty(elem, "context", "activation", "fingerprint")
+			origin := jsonparser.GetStringOrEmpty(elem, "context", "activation", "origin")
+			if fingerprint == "" || origin == "" {
+				u.stats.NewTaggedStat("activation_records_skipped", stats.CountType, stats.Tags{
+					"reason": "missing_fingerprint_or_origin",
```
mihir20: we can again make this granular to distinguish between missing fingerprint and origin

enterprise/activationrecords/records_reporter.go
```diff
@@ -0,0 +1,287 @@
...
+	// changing this will be non backwards compatible
+	murmurSeed = 123
+
+	activationRecordsTable = "activation_records_reports"
+)
+
+type recordKey struct {
+	workspaceID   string
+	sourceID      string
+	destinationID string
```
mihir20, line 50: do we also need origin in the key?

enterprise/activationrecords/records_reporter.go
```diff
@@ -0,0 +1,287 @@
...
+					// processor pipeline if it ever does happen.
+					u.log.Errorn("creating HLL for activation records", obskit.Error(hllErr))
+					u.stats.NewTaggedStat("activation_records_skipped", stats.CountType, stats.Tags{
+						"reason": "hll_init_failed",
+					}).Increment()
+					continue
+				}
+				truncated, wasTruncated := truncateRunes(origin, 256)
+				if wasTruncated {
+					u.stats.NewTaggedStat("activation_records_origin_truncated", stats.CountType, stats.Tags{}).Increment()
```
mihir20:
```suggestion
					u.stats.NewStat("activation_records_origin_truncated", stats.CountType).Increment()
```

enterprise/activationrecords/records_reporter.go
```diff
@@ -0,0 +1,287 @@
...
+					// Settings are validated at construction, so this is unreachable in
+					// practice; degrade gracefully (skip + stat) rather than crash the
+					// processor pipeline if it ever does happen.
+					u.log.Errorn("creating HLL for activation records", obskit.Error(hllErr))
+					u.stats.NewTaggedStat("activation_records_skipped", stats.CountType, stats.Tags{
+						"reason": "hll_init_failed",
+					}).Increment()
+					continue
+				}
+				truncated, wasTruncated := truncateRunes(origin, 256)
```
mihir20, line 225: can you add a comment on why do we need to truncate the origin?

enterprise/activationrecords/wire_compat_test.go
mihir20, line 1: isn't most of the tests here duplicates of what is already present in activation_records_test.go?

enterprise/reporting/flusher/activation_records_test.go
```diff
@@ -0,0 +1,261 @@
...
+	// validate the payloads sent to Reporting Service
+	reqBodies := testWebhook.Requests()
+	for _, req := range reqBodies {
+		var items []map[string]any
+
+		body, _ := io.ReadAll(req.Body)
+		err := jsonrs.Unmarshal(body, &items)
+		if err != nil {
+			log.Fatalf("Error parsing body: %v", err)
+		}
```
mihir20:
```suggestion
		require.NoError(t, err)
```

processor/processor.go
```diff
@@ -2883,6 +2910,25 @@ func (proc *Handle) storeStage(partition string, pipelineIndex int, in *storeMes
...
+				} else if marErr := proc.activationRecordsReporter.ReportActivationRecords(ctx, in.activationRecordsReports, sqlTx); marErr != nil {
+					proc.logger.Warnn("storing activation records, rolling back to savepoint (best-effort)", obskit.Error(marErr))
+					proc.statsFactory.NewStat("activation_records_store_error", stats.CountType).Increment()
+					if _, rbErr := sqlTx.ExecContext(ctx, "ROLLBACK TO SAVEPOINT activation_records"); rbErr != nil {
+						proc.logger.Warnn("rolling back to savepoint for activation records", obskit.Error(rbErr))
+					}
+				} else if _, relErr := sqlTx.ExecContext(ctx, "RELEASE SAVEPOINT activation_records"); relErr != nil {
+					proc.logger.Warnn("releasing savepoint for activation records", obskit.Error(relErr))
+				}
+			}
```
mihir20: I don't think we need this check; if a transaction fails, the processor panics, and we retry the entire batch again.

mihir20 (commented):
we can also add an end-to-end integration test where events are ingested through gateway and delivered to end destination.

we have a similar test for the trackedUsers: https://github.com/rudderlabs/rudder-server/blob/d69df287c88ddf176faacb89e9d06002eb43e666/integration_test/trackedusersreporting/tracked_users_reporting_test.go

enterprise/activationrecords/records_reporter.go
am6010: Addressed in the current revision — `recordSkip` now emits a granular reason per case: `missing_workspace` (`records_reporter.go:137`), `missing_source` (`:146`), and `missing_destination` (`:154`).

enterprise/activationrecords/records_reporter.go
am6010: Addressed — `recordSkip` (`records_reporter.go:303-307`) increments a tagged `activation_records_skipped` counter with a `reason` label, covering the should-not-happen-in-prod paths (`missing_batch`, `invalid_batch`, `hll_init_failed`). Alerts can key off `reason`.

enterprise/activationrecords/records_reporter.go
am6010: Addressed — these are now distinct skip reasons: `missing_fingerprint` (`records_reporter.go:181`) and `missing_origin` (`:186`).

enterprise/activationrecords/records_reporter.go
am6010: No — `origin` is intentionally kept out of the key (grain is `(workspace, source, destination)`) and carried as a column instead (`records_reporter.go:36-40`). We don't expect a source to change its `origin`, so it's effectively constant per connection; folding it into the key would never change the grouping and would only risk splitting one connection into multiple sketches if the value ever drifted. One HLL per connection is what the reporting backend unions on.

enterprise/activationrecords/records_reporter.go
am6010: Added — see `records_reporter.go:200-205`: `origin` is client-controlled (stamped into the event payload), so we defensively cap its length before writing it to the row / reporting POST, to prevent a malformed or over-long value from bloating either.

enterprise/activationrecords/wire_compat_test.go
am6010: Good catch — the `jsonparser` read-path subtest did overlap with `TestUniqueActivationRecordsReporter`. Removed it in b14c1ccbf and folded its only unique value (proving the reporter resolves the grain from the real gateway `jsonrs` envelope, not hand-written JSON) into grain-field assertions on the remaining test. What's left is the genuinely non-duplicated part: the fingerprint-HLL hex byte round-trip that proves cross-service wire-compat with the reporting backend's postgres `hll` extension.

enterprise/activationrecords/records_reporter.go
mihir20: Then why do we need `origin` ? I thought we would aggregate on this.

processor/processor.go
```diff
@@ -1847,6 +1857,9 @@ func (proc *Handle) preprocessStage(partition string, subJobs subJob, delay time
 		}
 
 		for _, singularEvent := range gatewayBatchEvent.Batch {
+			if ctxMap, ok := singularEvent["context"].(map[string]any); ok {
+				delete(ctxMap, "activation")
+			}
```
itsmihir: Why are we deleting this?

enterprise/activationrecords/records_reporter.go
```diff
@@ -0,0 +1,317 @@
...
+	err = m.Migrate("activation_records")
+	if err != nil {
+		return fmt.Errorf("migrating `activation_records_reports` table: %w", err)
+	}
+	return nil
+}
+
+// GenerateReportsFromJobs aggregates activation records from a batch of jobs.
+// It is FAIL-CLOSED: jobs missing fingerprint or origin are skipped (counted via stats).
+func (u *UniqueActivationRecordsReporter) GenerateReportsFromJobs(jobs []*jobsdb.JobT) []*ActivationRecord {
```
itsmihir, line 133: How are we filtering out event stream sources here?

processor/processor.go
am6010: `context.activation` is internal metering metadata (the stamped fingerprint) — it's only there for MAR counting and must not reach the destination. We read it for metering earlier in `GenerateReportsFromJobs`, then strip it here before the event is delivered. So the record is still counted, just not leaked downstream.

enterprise/activationrecords/records_reporter.go
am6010: We don't filter by source type — we gate on the stamped fingerprint. Only rETL record events carry `context.activation.fingerprint`; event-stream events don't, so they get skipped. So event-stream is excluded implicitly by the gate, not by an explicit filter.

enterprise/activationrecords/records_reporter.go
am6010: Fair point. Since `origin` lives on the connection config, billing can derive it from the connection at query time — so there's no need to carry it through metering. Going to drop it from the pipeline (server + reporting + sources) rather than store/transmit a field that's already derivable.

enterprise/activationrecords/records_reporter.go
itsmihir: Why not rely on sourceId instead to classify if a source is retl?

processor/processor.go
itsmihir: Can customers also set some fields in context? What if they are also setting the activation field for their own use case?

enterprise/activationrecords/records_reporter.go
am6010: Done — now gating on source_category == "warehouse" (gateway-stamped from SourceDefinition.Category), so only reverse-ETL sources are metered and a client can't get counted by stamping a fingerprint on an event-stream source. Read from job params, no config dependency in the hot path.

processor/processor.go
am6010: Good catch — context is customer-extensible, so we now strip only context.activation.fingerprint (the field we stamp) and drop activation only if it's left empty. Customer data under context.activation is preserved.

itsmihir: Can we add a sourceCategory check here, too? do that we only remove it for rETL connection

itsmihir: Also, is there any advantage or harm in keeping fringerpring?

enterprise/activationrecords/records_reporter.go
```diff
@@ -0,0 +1,303 @@
...
+			continue
+		}
+
+		// MAR meters reverse-ETL (warehouse) sources only. Classify by the source
+		// category the gateway stamps from SourceDefinition.Category, rather than
+		// relying solely on the fingerprint gate below: this prevents a client from
+		// being metered by stamping context.activation.fingerprint on a non-rETL
+		// (e.g. event-stream) source. Non-rETL is the expected majority of traffic,
+		// so skip it silently — no per-job skip stat on the hot path.
+		sourceCategory := jsonparser.GetStringOrEmpty(job.Parameters, "source_category")
```
itsmihir, line 154: Let's move this check before the sourceId check so that we don't pay the extra cost of getting sourceId for event stream source

processor/processor.go
am6010: context.activation is internal metering metadata that customers don't consume, and leaving it in pollutes the delivered event schema

processor/processor.go
```diff
@@ -1846,7 +1886,16 @@ func (proc *Handle) preprocessStage(partition string, subJobs subJob, delay time
...
+		// context.activation (fingerprint + origin) is stamped only on reverse-ETL
+		// (warehouse) sources for MAR metering. Strip it before delivery so these internal
+		// metering values never leak downstream — gated on source category so we never
+		// touch a customer-set context.activation on other source types. Metering reads
+		// these from the raw job payloads (see GenerateReportsFromJobs), which are left
+		// intact, so stripping these destination-bound copies is safe.
+		isReverseETLSource := strings.EqualFold(source.SourceDefinition.Category, sourceCategoryWarehouse)
+
 		for _, singularEvent := range gatewayBatchEvent.Batch {
+			stripActivationMetadata(singularEvent, isReverseETLSource)
```
itsmihir: [nit] we can have the check for isReverseETLSource here itself and avoid passing it to stripActivationMetadata

enterprise/activationrecords/records_reporter.go
```diff
@@ -0,0 +1,335 @@
...
+			acc, exists := accumulators[key]
+			if !exists {
+				newHll, hllErr := hll.NewHll(*u.hllSettings)
+				if hllErr != nil {
+					// Settings are validated at construction, so this is unreachable in
+					// practice; degrade gracefully (skip + stat) rather than crash the
+					// processor pipeline if it ever does happen.
+					u.log.Errorn("creating HLL for activation records", obskit.Error(hllErr))
+					u.recordSkip("hll_init_failed")
+					continue
```
itsmihir: should we panic in this case? we panic in users_reporter.

itsmihir (commented):
lets rename flusher/aggregator/types.go -> flusher/aggregator/tracked_users_types.go

itsmihir (approved):

enterprise/activationrecords/records_reporter.go
```diff
@@ -0,0 +1,337 @@
...
+	// Validate the HLL settings once, at startup, so a misconfigured precision or
+	// registerWidth fails fast here (the app handler surfaces this error) instead of
+	// panicking deep in the processor hot path on every metered job.
+	if _, err := hll.NewHll(*hllSettings); err != nil {
+		return nil, fmt.Errorf("invalid activation records HLL settings: %w", err)
+	}
+	return &UniqueActivationRecordsReporter{
+		log:         log,
+		hllSettings: hllSettings,
+		instanceID:  config.GetStringVar("1", "INSTANCE_ID"),
```
mihir20:
```suggestion
		instanceID:  conf.GetStringVar("1", "INSTANCE_ID"),
```

mihir20 (approved):
we can also update pipeline test [here](https://github.com/rudderlabs/rudder-server/blob/60b69b90a78ce92f282a37daf9a79a1b1b776b10/processor/partition_worker_test.go#L392-L392). Otherwise LGTM

PR #7123, opened 2026-06-25
feat: implement Kubernetes deployment scaler for workspace replica management

mihir20:
# Description

Adds `processor/internal/pytscaler` — a small package that reads and sets the replica count of a workspace's per-workspace `rudder-pytransformer` (`pyt-{workspaceID}`) Deployment via the Kubernetes scale subresource.

The pyt Deployments default to 0 replicas (Keda scales them back to 0 when idle), so before the processor forwards a test request
to a workspace's pyt pod, it must scale that Deployment up from zero. This package is the scaling primitive the `Forward` handler (pipe-3144) will call. No call site is wired up yet — this PR only introduces the package and its tests.

## What's in here

- **`Scaler` interface** — two thin primitives over the scale subresource:
    - `GetReplicaCount(ctx, workspaceID) (int, error)`
    - `SetReplicaCount(ctx, workspaceID, count int) error`
- **`k8sScaler`** (unexported) — the real implementation:
    - The pyt Deployment lives in one of several pyt namespaces, not a fixed one. Since the scale subresource is namespace-scoped, it locates the Deployment cluster-wide by `metadata.name=pyt-{lower(workspaceID)}` (a
      universally-supported field selector) and reads the namespace off the match — no namespace config or ExternalName parsing.
    - Operates **only** through the scale subresource (`GetScale`/`UpdateScale`), never a full Deployment update, so the pod spec — including the Kata `RuntimeClass` — is never touched.
    - Transient k8s API errors (429, timeouts, 500, 503) are retried with exponential backoff; everything else fails fast.
- **`noopScaler`** (`NewNoop`) — a do-nothing implementation so the processor starts and serves even when Kubernetes is unavailable (local dev, client-build failure, or the feature disabled). Its `GetReplicaCount` reports `1` so
  a get-before-set caller naturally skips scaling.
- In-cluster client by default, with a kubeconfig fallback for local dev.

## Design notes

- **Idempotency is the caller's job, not this layer's.** `SetReplicaCount` writes the count unconditionally; the "scale to N only when the current count is 0" decision lives in the caller (the `Forward` handler). This keeps the
  k8s layer a pure pass-through to the scale subresource and mirrors the orchestrator's `ReplicaManager` (`GetReplicaCount`/`SetReplicaCount`) terminology.
- **The scale target (`Processor.UserTransformer.pytTestScaleReplicas`) is not read here** — the caller decides the count and passes it in.
- **`Scaler` is the abstraction boundary** — the `Forward` handler depends on the interface, not on Kubernetes. Two implementations (`k8sScaler`, `noopScaler`) plus a fake-client testing need justify the interface.

## Linear Ticket

pipe-3145

## Security

### DIFF
--- a/processor/internal/pytscaler/k8s.go
+++ b/processor/internal/pytscaler/k8s.go
@@ -1,0 +1,136 @@
+package pytscaler
+
+import (
+	"context"
+	"fmt"
+
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/rudderlabs/rudder-go-kit/config"
+	"github.com/rudderlabs/rudder-go-kit/logger"
+)
+
+type k8sScaler struct {
+	client kubernetes.Interface
+	log    logger.Logger
+	retry  retrySettings
+}
+
+func (s *k8sScaler) GetReplicaCount(ctx context.Context, workspaceID string) (int, error) {
+	if workspaceID == "" {
+		return 0, fmt.Errorf("workspaceID is empty")
+	}
+	name := deploymentName(workspaceID)
+	namespace, err := s.resolveNamespace(ctx, name)
+	if err != nil {
+		return 0, err
+	}
+	scale, err := withRetry(ctx, s.retry, func() (*autoscalingv1.Scale, error) {
+		return s.client.AppsV1().Deployments(namespace).GetScale(ctx, name, metav1.GetOptions{})
+	})
+	if err != nil {
+		return 0, fmt.Errorf("getting scale for deployment %s/%s: %w", namespace, name, err)
+	}
+	return int(scale.Spec.Replicas), nil
+}
+
+func (s *k8sScaler) SetReplicaCount(ctx context.Context, workspaceID string, count int) error {
+	if workspaceID == "" {
+		return fmt.Errorf("workspaceID is empty")
+	}
+	name := deploymentName(workspaceID)
+	namespace, err := s.resolveNamespace(ctx, name)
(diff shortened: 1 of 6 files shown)


processor/internal/pytscaler/k8s.go
```diff
@@ -0,0 +1,136 @@
...
+	scale, err := withRetry(ctx, s.retry, func() (*autoscalingv1.Scale, error) {
+		return s.client.AppsV1().Deployments(namespace).GetScale(ctx, name, metav1.GetOptions{})
+	})
+	if err != nil {
+		return 0, fmt.Errorf("getting scale for deployment %s/%s: %w", namespace, name, err)
+	}
+	return int(scale.Spec.Replicas), nil
+}
+
+func (s *k8sScaler) SetReplicaCount(ctx context.Context, workspaceID string, count int) error {
```
fracasula, line 42: Would it make sense to return here not just after the kube-apiserver accepted the write but when the pods are ready?

fracasula (commented):
Should we have something to prevent the keda autoscaler not to be below a certain threshold to avoid issues with this testing functionality? e.g. we scale up and then after 15-30s it is scaled down by keda potentially terminating the pod mid-request.

processor/internal/pytscaler/k8s.go
mihir20: it should be fine I am planning to use an existing transformer client that retries on cold start, which means it will retry until the pod is ready. And also, we will be emitting a cold start metric in case of any cold start failure, which will help in keeping KEDA also happy with the replica count we are manually setting.

mihir20:
> Should we have something to prevent the keda autoscaler not to be below a certain threshold to avoid issues with this testing functionality? e.g. we scale up and then after 15-30s it is scaled down by keda potentially terminating the pod mid-request.

@fracasula, once we scale up and a request is sent to pytransformer, the metrics used in KEDA will not let the pod to go away, and we have a 10-min stabilization window, which will not take the pod away even after 10 min when the traffic is 0.

fracasula (approved):

ktgowtham (approved):

PR #7124, opened 2026-06-25
feat: support compressing payloads in reporting client

atzoum:
# Description

Adds optional gzip compression to the reporting client. When enabled, payloads are gzip-compressed and sent with `Content-Encoding: gzip`.

- Gated behind a reloadable config flag `Reporting.httpClient.compression.enabled` (default `false`), with a configurable level `Reporting.httpClient.compression.level` (default `gzip.DefaultCompression`).
- The compressed body is materialized once before the retry loop so it can be replayed across backoff retries.
- Reports both `reporting_client_http_request_total_bytes` (uncompressed) and the new `reporting_client_http_request_compressed_bytes` (on the wire).

## Linear Ticket

resolves PIPE-3157

## Security

### DIFF
--- a/enterprise/reporting/client/client.go
+++ b/enterprise/reporting/client/client.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


ktgowtham (approved):
This needs a reporting service PR too. reporting service doesn't support gzip request body. We might have to add a middleware. Shouldn't be difficult. 
@itsmihir we can discuss & follow up on this.

itsmihir (approved):

itsmihir:
Reporting already has support for it.

PR #7125, opened 2026-06-25
feat: carry integration version on DestinationT to the transformer

saikumarrs:
## What

Adds an integration major `version` to `DestinationT` and carries it through to the transformer on every path a destination reaches it.

- **backend-config**: new `DestinationT.Version int` tagged `json:"version"` (no `omitempty`, always emitted). rudder-server only carries it — never interprets it.
- Rides along for free on whole-destination embed paths: processor `TransformerEvent`, router `RouterJobT`/`DestinationJobT`, and the compacted variants.
- The delivery **proxy** (slimmed payload) carries it explicitly via `ProxyRequestPayload.DestinationVersion` / `ProxyRequestPayloadV0`, populated in `router/worker.go`.

## Testing

- backend-config: round-trip for single-workspace + namespace unmarshal paths, case-insensitive key, missing→0, and always-present number marshalling.
- Presence in marshalled `TransformerEvent` (golden `goldenUtMirrorClientEvents.json`, now v1) and proxy payload (v0 + v1 adapters).

INT-6484

### DIFF
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -89,7 +89,7 @@ type ProxyRequestPayload struct {
 	integrations.PostParametersT
 	Metadata           []ProxyRequestMetadata `json:"metadata"`
 	DestinationConfig  map[string]any         `json:"destinationConfig"`
-	DestinationVersion int                    `json:"destinationVersion"`
+	DestinationVersion int                    `json:"destinationVersion,omitempty"`
 }
 
 type ProxyRequestParams struct {
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -87,8 +87,9 @@ type ProxyRequestMetadata struct {
 
 type ProxyRequestPayload struct {
 	integrations.PostParametersT
-	Metadata          []ProxyRequestMetadata `json:"metadata"`
-	DestinationConfig map[string]any         `json:"destinationConfig"`
+	Metadata           []ProxyRequestMetadata `json:"metadata"`
+	DestinationConfig  map[string]any         `json:"destinationConfig"`
+	DestinationVersion int                    `json:"destinationVersion"`
 }
 
 type ProxyRequestParams struct {
--- a/router/transformer/transformer_proxy_adapter.go
+++ b/router/transformer/transformer_proxy_adapter.go
@@ -27,8 +27,9 @@ type transformerProxyAdapter interface {
 
 type ProxyRequestPayloadV0 struct {
 	integrations.PostParametersT
-	Metadata          ProxyRequestMetadata `json:"metadata"`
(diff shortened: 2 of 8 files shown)


saikumarrs:
**Transformer counterpart:** https://github.com/rudderlabs/rudder-transformer/pull/5318

Reads `destination.version` (transform/router) and the top-level `destinationVersion` (proxy) and dispatches on the integration major in-file. Pairs with this PR's `DestinationT.Version` plumbing. (INT-6495 / INT-6492)

shekhar-rudder (approved):

router/transformer/transformer_proxy_adapter.go
```diff
@@ -27,8 +27,9 @@ type transformerProxyAdapter interface {
 
 type ProxyRequestPayloadV0 struct {
 	integrations.PostParametersT
-	Metadata          ProxyRequestMetadata `json:"metadata"`
-	DestinationConfig map[string]any       `json:"destinationConfig"`
+	Metadata           ProxyRequestMetadata `json:"metadata"`
+	DestinationConfig  map[string]any       `json:"destinationConfig"`
+	DestinationVersion int                  `json:"destinationVersion"`
```
atzoum: do we need to transfer it over the wire when it is zero?

router/transformer/transformer.go
```diff
@@ -87,8 +87,9 @@ type ProxyRequestMetadata struct {
 
 type ProxyRequestPayload struct {
 	integrations.PostParametersT
-	Metadata          []ProxyRequestMetadata `json:"metadata"`
-	DestinationConfig map[string]any         `json:"destinationConfig"`
+	Metadata           []ProxyRequestMetadata `json:"metadata"`
+	DestinationConfig  map[string]any         `json:"destinationConfig"`
+	DestinationVersion int                    `json:"destinationVersion"`
```
atzoum: do we need to transfer it over the wire when it is zero?

router/transformer/transformer_proxy_adapter.go
saikumarrs: Good call. Added `omitempty` here in c8ad90d61 (and on `DestinationT.Version` for a uniform contract), so a zero/v1-default version is no longer sent over the wire. Absent and `0` both mean v1, so it's lossless — the transformer treats a missing version as v1.

router/transformer/transformer.go
saikumarrs: Good call. Added `omitempty` here in c8ad90d61 (and on `DestinationT.Version` for a uniform contract), so a zero/v1-default version is no longer sent over the wire. Absent and `0` both mean v1, so it's lossless — the transformer treats a missing version as v1.

atzoum (approved):

PR #7126, opened 2026-06-26
feat: add control-plane connection management for processor

mihir20:
# Description

Opens the processor's dataplane-initiated gRPC connection to **cp-router**, reusing the shared `controlplane.ConnectionManager` (the same mechanism rudder-sources and warehouse use). This makes the processor reachable from cp-router so that, test-flow requests can be forwarded to the per-workspace `rudder-pytransformer`.

Only the **node-0** pod opens the connection, so cp-router holds a single inbound connection per dataplane (not one per replica) and node-0 becomes the sole owner of any privileged work served over it.
The whole bootstrap is gated behind a config flag and defaults off — this PR is inert until enabled.

The `Forward` RPC itself is intentionally **left unimplemented in this PR** (a follow-up PR adds op-routing + the HTTP forward to pyt). Until then cp-router receives `codes.Unimplemented`.

## What's in here

- **`proto/processor/processor.proto`** (+ generated `.pb.go` / `_grpc.pb.go`): a generic `ProcessorService` with a single RPC — `Forward(op, workspace_id, payload) → (status_code, body)`. The op is an
  opaque string and payload/body are JSON bytes, so cp-router stays transformation-agnostic and never grows per-op code. Generated with the repo's pinned plugin versions (`protoc-gen-go` v1.33.0,
  `protoc-gen-go-grpc` v1.3.0).
- **`processor/cpservice/cpservice.go`** — `Service`, the `ProcessorService` gRPC handler. Embeds `UnimplementedProcessorServiceServer` so it registers and serves safely while `Forward` is pending.
- **`processor/cpservice/connection.go`** — `Connector`, which wraps `controlplane.ConnectionManager`, registers the handler on the dataplane-side gRPC server, and drives the connection from the workspace
  connection flags published by the control plane (the warehouse pattern). `ShouldConnect` = feature-enabled **&&** node-0.
- **`app/apphandlers/{processor,embedded}AppHandler.go`** — gated bootstrap wired into both modes that run the processor pipeline. GATEWAY mode is intentionally excluded (no processor pipeline; the LLD
  keeps this off the horizontally-scaled gateway).

## Node-0 detection

Reuses `misc.GetInstanceID()` — the same instance-ID ordinal signal `app/cluster/state` uses. It returns the trailing ordinal of `INSTANCE_ID` (`"0"`, `"1"`, …); an unset `INSTANCE_ID` yields `""`, so local runs never connect.

## Config

| Key | Default | Purpose |
  |-----|---------|---------|
| `Processor.cpRouterConnection.enabled` | `false` | Master switch for opening the cp-router connection |
| `Processor.cpRouterConnection.retryInterval` | `0` (→ 1s) | Reconnect backoff for the connection manager |
| `CP_ROUTER_USE_TLS` (reused) | `true` | TLS for the cp-router dial |
| `INSTANCE_ID` (reused) | — | Node-0 detection + `AuthInfo.InstanceID` |

## Linear Ticket

pipe-3143

## Security


processor/cpservice/connection.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+	}
+}
+
+// apply reads the cp-router connection flag for this service from the workspace
+// config and applies it. The flags are identical across workspaces in a
+// multi-workspace deployment, so the first one carrying the flag is enough.
+func (c *Connector) apply(configData map[string]backendconfig.ConfigT) {
+	for _, wConfig := range configData {
+		flags := wConfig.ConnectionFlags
```
ktgowtham, line 115: Not sure if CP is setting these flags today. for MT type deployments(which is the case for all our deployments), connections are made 

for warehouse MT type deployments, this flag is set. 
https://github.com/rudderlabs/rudder-server/blob/f392a5370bf06903bb0afcec122b319048c5bfad/backend-config/namespace_config.go#L181

can we double check?

ktgowtham: we might have to add processor too here

mihir20: The CP router is setting this flag whenever the CP router's HTTP API is called from the config backend.

ktgowtham: that might be needed for open source data planes. but for transformation testing, we can proactively initiate a connection like we do for warehouse if it's a namespace deployment. So the ask is to add rudderstack-processor too, like warehouse is explicitly added

mihir20: I think for the warehouse we are doing this so that connections can be initiated whenever the warehouse master is restarted. CP router also closes the unused connections after some TTL. I don't think our flow is this much critical to initiate connection on every processor restart.

ktgowtham: We'll save lot of time if we have a connection already in place. The connection flags from workspace config approach takes time to be known to the server and it to initiate the connection. 
having a connection available, user doesn't need to wait for long time where the flag is propogated to the server. 
Again, we don't need to follow the connection flag approach for our hosted dataplanes. I think it's for open source dataplanes. For hosted, connections are made using namespace token approach. for both warehouse and rudder-sources.

ktgowtham: cc: @atzoum @achettyiitr as they are well aware of cp router connections mechanisms.

fracasula: > We'll save lot of time if we have a connection already in place

Don't we have metrics for this?

processor/cpservice/connection.go
```diff
@@ -0,0 +1,122 @@
...
+	if region := conf.GetStringVar("", "region"); region != "" {
+		labels["region"] = region
+	}
+
+	log = log.Child("cpservice")
+	cm := &controlplane.ConnectionManager{
+		AuthInfo: controlplane.AuthInfo{
+			Service:         ServiceName,
+			ConnectionToken: connectionToken,
+			InstanceID:      conf.GetStringVar("1", "INSTANCE_ID"),
```
fracasula: Aren't we supposed to connect from node-0 only? I'm asking because I saw that `isNode0` does this:

```go
func isNode0() bool {
	return misc.GetInstanceID() == "0"
}
```

So "0" != "1".

processor/cpservice/connection.go
```diff
@@ -0,0 +1,122 @@
...
+	ch := c.backendConfig.Subscribe(ctx, backendconfig.TopicBackendConfig)
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case data, ok := <-ch:
+			if !ok {
+				return nil
+			}
+			if configData, ok := data.Data.(map[string]backendconfig.ConfigT); ok {
```
fracasula, line 103: Can this be `!ok`? :thinking:

processor/cpservice/connection.go
```diff
@@ -0,0 +1,122 @@
...
+		Logger:          log,
+		RegisterService: func(srv *grpc.Server) { proto.RegisterProcessorServiceServer(srv, handler) },
+	}
+	return &Connector{log: log, backendConfig: backendConfig, cm: cm}, nil
+}
+
+// Run subscribes to backend config and keeps the cp-router connection in sync
+// with the workspace connection flags until ctx is cancelled, at which point it
+// closes the connection. It blocks, so run it in the app errgroup.
+func (c *Connector) Run(ctx context.Context) error {
```
fracasula, line 92: The tests cover `ShouldConnect`, service registration, and an end-to-end `Forward`. But the actual subscribe→decode→`apply`→`stop` loop in `Run`, including the `data.Data.(map[string]backendconfig.ConfigT)` type assertion all the way till close-on-shutdown, has no direct coverage.

processor/cpservice/cpservice_test.go
```diff
@@ -0,0 +1,167 @@
...
+	}{
+		{name: "enabled on node-0", enabled: true, instanceID: "rs-proc-0", want: true},
+		{name: "disabled on node-0", enabled: false, instanceID: "rs-proc-0", want: false},
+		{name: "enabled on node-1", enabled: true, instanceID: "rs-proc-1", want: false},
+		{name: "enabled with unset instance id", enabled: true, instanceID: "", want: false},
+		{name: "enabled on HA node-0", enabled: true, instanceID: "rs-gw-ha-0-abc-def", want: true},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv("INSTANCE_ID", tc.instanceID)
```
fracasula, line 38: Aren't we able to use the config just below this line for this? In connection.go you're doing:

```go
InstanceID:      conf.GetStringVar("1", "INSTANCE_ID"),
```

processor/cpservice/connection.go
ktgowtham: > > We'll save lot of time if we have a connection already in place
> 
> Don't we have metrics for this?

Not sure, probably not. but there are lot of steps involved: 
1. config-be calls cp-router. 
2. cp-router inserts a connection flag row seeing there's no connection. waits for connection to be established.
3. config-be adds this flag in next workspace/namespace config poll from server.
4. server initiates the conenction
5. step 2 now sees the connection and proceeds. 
All this process will take around 5-10s. 
Note that we need to bring up the replica on pyT too. All these can't add up to a really high value or else config-be API from web-app times out(we've a 60s timeout). so better to optimize where-ever we can.

ktgowtham: All I asked is to just add `rudderstack-processor` to the flags, like how `warehouse` is done :D

fracasula: Not questioning anything but since you said "we'll save lot of time" which seemed pretty arbitrary I asked if we have metrics for that so that we can check the actual time. Usually grpc clients come with metrics, if we don't have them then it might be a good opportunity to add some, that's all.

ktgowtham: yeah, shouldn't be hard to add a metric for wait time for a connection

mihir20: @fracasula good point, will add the metrics in the grpc server which we will be implementing in the next pr.

processor/cpservice/connection.go
mihir20: We will only be connecting from node 0. InstanceId is set for all the processor pods via operator; we will not be picking the default value. But I got your point will use misc.InstanceID() here.

processor/cpservice/connection.go
mihir20: shouldn't be will remove the check

processor/cpservice/cpservice_test.go
mihir20: no in `misc.GetInstanceID()` we are using global config

processor/cpservice/connection.go
mihir20: added more tests

processor/cpservice/connection.go
ktgowtham: > All I asked is to just add `rudderstack-processor` to the flags, like how `warehouse` is done :D

Mihir, aren't we going to add this?

mihir20: I missed pushing the changes. I thought I already pushed 😅

mihir20: done now

ktgowtham (approved):

fracasula (approved):

PR #7133, opened 2026-06-29
fix(jobsdb): compaction panics with pq error column js.consumer does not exist at position

atzoum:

# Description

Older `v_last` views were missing the consumer field. Adding it now during database migration.

## Security


itsmihir (approved):

ktgowtham (approved):

sql/migrations/jobsdb/000014_multi_consumer_columns.up.tmpl
```diff
@@ -1,4 +1,5 @@
 {{range .Datasets}}
     ALTER TABLE "{{$.Prefix}}_jobs_{{.}}" ADD COLUMN IF NOT EXISTS consumers TEXT[] NOT NULL DEFAULT '{""}';
     ALTER TABLE "{{$.Prefix}}_job_status_{{.}}" ADD COLUMN IF NOT EXISTS consumer TEXT NOT NULL DEFAULT '';
+    CREATE OR REPLACE VIEW "v_last_{{$.Prefix}}_job_status_{{.}}" AS SELECT DISTINCT ON (job_id) * FROM "{{$.Prefix}}_job_status_{{.}}" ORDER BY job_id ASC, id DESC;
```
atzoum, line 4: **TIL:** In postgres, `SELECT *` in a view is expanded to the explicit column list at `CREATE VIEW` time and stored in `pg_attribute`. It's not re-evaluated on each query.

PR #7137, opened 2026-06-30
chore: enable compression for google pubsub

atzoum:
# Description

Add a configuration option for enabling compression in google pubsub transport

## Security

### DIFF
--- a/services/streammanager/googlepubsub/googlepubsubmanager.go
+++ b/services/streammanager/googlepubsub/googlepubsubmanager.go
@@ -101,6 +101,7 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (*Googl
 	for _, s := range pubsubConfig.EventToTopicMap {
 		topic := client.Publisher(s["to"])
 		topic.PublishSettings.NumGoroutines = config.GetIntVar(25, 1, "Router.GOOGLEPUBSUB.Client.NumGoroutines")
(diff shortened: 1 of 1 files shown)


services/streammanager/googlepubsub/googlepubsubmanager.go
```diff
@@ -101,6 +101,7 @@ func NewProducer(destination *backendconfig.DestinationT, o common.Opts) (*Googl
 	for _, s := range pubsubConfig.EventToTopicMap {
 		topic := client.Publisher(s["to"])
 		topic.PublishSettings.NumGoroutines = config.GetIntVar(25, 1, "Router.GOOGLEPUBSUB.Client.NumGoroutines")
+		topic.PublishSettings.EnableCompression = config.GetBoolVar(true, "Router.GOOGLEPUBSUB.Client.EnableCompression")
```
ktgowtham: fine making it default true? shall we make it false, enable and monitor and make this default true?

saikumarrs (approved):
LGTM except for this concern, https://github.com/rudderlabs/rudder-server/pull/7137#discussion_r3499231403

services/streammanager/googlepubsub/googlepubsubmanager.go
atzoum: sure, I think we should include this in a patch release, wdyt?

ktgowtham (approved):

PR #7141, opened 2026-07-01
feat: support partition migration job fanout per jobsdb

atzoum:
# Description

Partition migrations currently move and flush the `gw`, `rt`, `batch_rt`, and `proc` jobsDBs in lockstep, so the slowest one holds up the rest of the migration. This PR adds support for per-jobsDB migration fan-out (paired with corresponding changes in rudder-schemas and rudder-orchestrator), allowing each jobsDB to move/flush independently.

How it's triggered:
- Opt-in via a reloadable config flag (`PartitionMigration.jobsDBFanoutEnabled`, default off).
- When a source node acknowledges a migration, if fan-out is enabled locally and the orchestrator advertises support for it, the node includes the list of jobsDBs it will move in its acknowledgement instead of moving them all as one unit.
- The orchestrator then fans out the migration jobs per jobsDB accordingly.
- This decision is made once at acknowledgement time and honored for that migration's lifetime, falling back transparently to the existing legacy behavior otherwise.

## Linear Ticket

resolves PIPE-3034

## Security


integration_test/partitionmigration/partitionmigration_embedded_test.go
```diff
@@ -124,7 +138,9 @@ func TestPartitionMigrationEmbeddedMode(t *testing.T) {
 	bc.URL = strings.Replace(bc.URL, "127.0.0.1", localIp, 1) // replace localhost with local IP for docker containers to access
 
 	// start rudder-orchestrator that will run the migration
+	// TODO: revert to the default (latest) tag once jobsdb fan-out (rudder-orchestrator PR #145) is released
 	rudoResource, err := rudo.Setup(pool, t,
+		rudo.WithTag("pr-145"),
```
atzoum: TODO: revert before merging and after orchestrator release

integration_test/partitionmigration/partitionmigration_gwproc_test.go
```diff
@@ -156,7 +158,9 @@ func testPartitionMigrationGatewayProcessorMode(t *testing.T, extraStressWorkspa
 	bc.URL = strings.Replace(bc.URL, "127.0.0.1", localIp, 1) // replace localhost with local IP for docker containers to access
 
 	// start rudder-orchestrator that will run the migrations
+	// TODO: revert to the default (latest) tag once jobsdb fan-out (rudder-orchestrator PR #145) is released
 	rudoResource, err := rudo.Setup(pool, t,
+		rudo.WithTag("pr-145"),
```
atzoum: TODO: revert before merging and after orchestrator release

ktgowtham (approved):

mihir20 (approved):

PR #7142, opened 2026-07-01
chore: support compression in bqstream destinations

atzoum:

# Description

This PR improves the BigQuery streaming (bqstream) destination's transport:

- **Optional payload compression during transport** — it is now possible to gzip-compress the payloads we send to BigQuery over the wire, reducing egress bandwidth. It is opt-in via configuration and disabled by default.
- **Reuse the client as configured by Google** — rather than replacing the HTTP client, we build on top of the client/transport that the Google API library configures, so its authentication and connection defaults are preserved.
- **Worker-based connection pool tuning** — we override the maximum idle connections based on the configured number of workers, consistent with how the other stream managers already tune their pools.

The destination is now exercised through real integration tests against BigQuery.

## Linear Ticket

resolves PIPE-3163

## Security

### DIFF
--- a/services/streammanager/bqstream/bqstreammanager_client.go
+++ b/services/streammanager/bqstream/bqstreammanager_client.go
@@ -1,0 +1,130 @@
+package bqstream
+
+import (
+	"bytes"
+	"compress/gzip"
+	"context"
+	"io"
+	"net"
+	"net/http"
+	"time"
+
+	"golang.org/x/net/http2"
+	"google.golang.org/api/option"
+	htransport "google.golang.org/api/transport/http"
+
+	"github.com/rudderlabs/rudder-go-kit/config"
+)
+
+// newHTTPClient returns the HTTP client used by the BigQuery client. It always
+// applies our tuned base transport (see baseTransport) and, when compress is
+// true, gzip-compresses request bodies. The standard Google authentication
+// middleware is preserved by building the transport via htransport.NewTransport.
+func newHTTPClient(ctx context.Context, compress bool, opts ...option.ClientOption) (*http.Client, error) {
+	base := baseTransport()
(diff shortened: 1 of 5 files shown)


achettyiitr (approved):

services/streammanager/bqstream/bqstreammanager_client.go
```diff
@@ -0,0 +1,130 @@
...
+	// something that is not an *http.Transport, fall back to an equivalent one,
+	// mirroring defaultBaseTransport/fallbackBaseTransport in the Google library.
+	trans := clonedDefaultTransport()
+	if trans == nil {
+		trans = fallbackBaseTransport()
+	}
+	trans.MaxIdleConns = config.GetIntVar(64, 1, "Router.BQSTREAM.httpMaxIdleConns", "Router.BQSTREAM.noOfWorkers", "Router.noOfWorkers")
+	trans.MaxIdleConnsPerHost = config.GetIntVar(64, 1, "Router.BQSTREAM.httpMaxIdleConnsPerHost", "Router.BQSTREAM.noOfWorkers", "Router.noOfWorkers")
+	if http2Trans, err := http2.ConfigureTransports(trans); err == nil {
+		http2Trans.ReadIdleTimeout = 31 * time.Second
```
saikumarrs, line 58: Can we avoid this magic number `31` here?

saikumarrs (approved):

services/streammanager/bqstream/bqstreammanager_client.go
atzoum: we are mirroring the defaults that Google API HTTP client applies
`google.golang.org/api/transport/http/configure_http2_go116.go:24`

PR #7143, opened 2026-07-01
feat: implement Forward RPC for control-plane requests to workspace pyt transformers

mihir20:
# Description

Implements the processor's `Forward` RPC — the control-plane-facing
entrypoint that routes a transformation-**test** request to a workspace's
per-workspace `pyt-{workspaceID}` deployment.

Flow: `config-backend → cp-router → rudderstack-processor (this) → pyt-{workspaceID}`.
On each call the processor maps the op to a pyt endpoint, scales the (Keda-idled,
0-replica) deployment up if needed, forwards the request, and passes the pyt's
HTTP status/body straight back.

> **Python only.** These four endpoints exist for the Python transformation-test
> flow. JavaScript tests stay on rudder-transformer and never reach the processor;
> routing JS here would need separate handling.

## Changes

**`processor/cpservice/` — the `Forward` handler**
- `Forward(ctx, req)` maps `req.Op` (proto `Op` enum) to the matching pyt endpoint
  method; an unspecified/unknown op ⇒ gRPC `InvalidArgument`.
- `ensureScaled` uses get-before-set on the scaler (scale up only when at 0
  replicas). Concurrent Forwards for the same workspace are collapsed via
  `singleflight`, so a burst triggers **one** scale-up (and one cluster-wide
  List), not one per request.
- `NewService` builds a k8s-backed pyt scaler (falling back to a no-op when the
  k8s client can't be built, so the processor still serves) and a
  user_transformer-backed forwarder; `WithScaler`/`WithForwarder` inject fakes in
  tests.
- Error mapping: missing pyt deployment ⇒ `FailedPrecondition`; scaler failure ⇒
  `Internal`; forward failure ⇒ `Unavailable`.

**`processor/internal/transformer/user_transformer/` — the forward client**
- New per-endpoint methods on the existing `Client`: `Test` (`/test`), `TestRun`
  (`/testRun`), `TestLibrary` (`/test-library`), `ExtractLibs` (`/extract-libs`),
  all delegating to a shared `forwardTest` helper.
- `forwardTest` reuses the client's existing HTTP client and cold-start detection
  (`isColdStartError`) but is **time-boxed by the caller's ~60s deadline**: it
  retries only on cold-start signals (connection refused / DNS / 502 / 503) and
  passes any other response through unchanged.
- Reuses the existing `maxRetry` / `maxRetryBackoffInterval` settings; new options
  `WithMaxRetryBackoffInterval` / `WithMaxRetry` let cpservice override them (via
  reloadable `Processor.UserTransformer.pytTest*` config) so the forward path
  retries a cold pyt quickly instead of with the event path's large default.

**`app/apphandlers/` — wiring**
- Processor and embedded handlers construct the service via the updated
  `NewService` when `cpservice.ShouldConnect` is true (node-0 only).

## Config

| Key | Default | Purpose |
  |-----|---------|---------|
| `Processor.UserTransformer.pytTestScaleReplicas` | 2 | replicas to scale a cold pyt up to |
| `Processor.UserTransformer.pytTestMaxRetryBackoffInterval` | 1s | max backoff between forward retries |
| `Processor.UserTransformer.pytTestMaxRetry` | 60 | retry budget for the forward path (ctx deadline is the hard bound) |

## Linear Ticket

pipe-3144

## Security


processor/internal/transformer/user_transformer/user_transformer_test.go
```diff
@@ -1893,3 +1894,92 @@ func TestColdStartCounter(t *testing.T) {
 		})
 	}
 }
+
+func TestForwardTest(t *testing.T) {
```
fracasula, line 1902: @mihir20 can you look into this? 

`forwardTest` builds its target URL via `userTransformBaseURL("python", "", workspaceID)`,
which only returns the per-workspace pyt template when `perWorkspacePyTEnabled` is **true**:

```go
url := u.userTransformBaseURL("python", "", workspaceID) + path   // line 528
// userTransformBaseURL: per-workspace branch is gated on perWorkspacePyTEnabled.Load();
// when false it falls through to userTransformationURL (default http://localhost:9090).
```

`TestForwardTest` sets `perWorkspacePyTURLTemplate` but **never sets
`perWorkspacePyTEnabled=true`**. So every call routes to `http://localhost:9090`, gets
connection-refused, treats that as a cold start, and retries.

mihir20: resolved now

processor/cpservice/forward.go
```diff
@@ -0,0 +1,96 @@
...
+func (s *Service) Forward(ctx context.Context, req *proto.ForwardRequest) (*proto.ForwardResponse, error) {
+	forward, ok := s.forwardForOp(req.Op)
+	if !ok {
+		return nil, status.Errorf(codes.InvalidArgument, "unknown op %v", req.Op)
+	}
+	if req.WorkspaceId == "" {
+		return nil, status.Error(codes.InvalidArgument, "workspaceId is required")
+	}
+
+	if err := s.ensureScaled(ctx, req.WorkspaceId); err != nil {
```
fracasula, line 45: @mihir20 Possible SSRF below, please have a look:

`WorkspaceId` (proto3 `string`, no format constraint) is only lowercased and then `strings.ReplaceAll`'d into `perWorkspacePyTURLTemplate` (default `http://pyt-{workspaceID}:8080`) at `user_transformer.go:632`, used at `:528`, dialed at `:540`.

There is **no validation of `WorkspaceId` anywhere on this path** — `Forward` only checks the op is known and the ID is non-empty (`forward.go`). URL-significant characters (`@`, `#`, `?`, `:`, `/`) survive lowercasing.
- `workspaceID = "a@evil.com#"` → the string `http://pyt-a@evil.com#:8080/test`, which Go's
  `net/url` parses (RFC 3986) as **userinfo=`pyt-a`, host=`evil.com`, port=80**, with `#` starting
  the fragment — so `:8080/test` becomes the *fragment*, `req.URL.Path` is empty, and the outbound
  request is `POST / HTTP/1.1` to `evil.com:80`

processor/cpservice/forward_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/rudderlabs/rudder-go-kit/config"
+	"github.com/rudderlabs/rudder-go-kit/logger"
+
+	"github.com/rudderlabs/rudder-server/processor/internal/pytscaler"
+	proto "github.com/rudderlabs/rudder-server/proto/processor"
+)
+
+// fakeForwarder records which endpoint method was called and returns a canned
+// response, so tests can assert Forward dispatches each op to the right method.
+type fakeForwarder struct {
```
fracasula: Can we move the helpers below and start with the tests?

processor/cpservice/forward_test.go
```diff
@@ -0,0 +1,236 @@
...
+	setCount  int
+}
+
+func (s *fakeScaler) GetReplicaCount(context.Context, string) (int, error) {
+	return s.getReplicas, s.getErr
+}
+
+func (s *fakeScaler) SetReplicaCount(_ context.Context, _ string, count int) error {
+	s.setCalled, s.setCount = true, count
+	return nil
```
fracasula: Are we testing the path where we return an error here?

processor/internal/transformer/user_transformer/user_transformer_test.go
```diff
@@ -1893,3 +1894,95 @@ func TestColdStartCounter(t *testing.T) {
...
+		conf := config.New()
+		conf.Set("Processor.UserTransformer.perWorkspacePyTEnabled", true)
+		conf.Set("Processor.UserTransformer.perWorkspacePyTURLTemplate", srv.URL)
+		client := user_transformer.New(conf, logger.NOP, stats.NOP,
+			user_transformer.WithMaxRetryBackoffInterval(config.NewMockValueLoader(5*time.Millisecond)))
+
+		ctx, cancel := context.WithTimeout(context.Background(), 50*time.Millisecond)
+		defer cancel()
+		_, _, err := client.Test(ctx, "ws-1", []byte(`{}`))
+		require.Error(t, err)
```
fracasula: I think you're expecting a context deadline exceeded here. Could be more specific.

processor/internal/transformer/user_transformer/user_transformer_test.go
```diff
@@ -1893,3 +1894,95 @@ func TestColdStartCounter(t *testing.T) {
...
+		total := 0
+		for _, m := range statsStore.GetByName("processor_user_transformer_cold_start_errors_total") {
+			total += int(m.Value)
+		}
+		require.Equal(t, 2, total, "each cold-start retry should increment the counter")
+	})
+
+	t.Run("stops retrying when the deadline is exceeded", func(t *testing.T) {
+		srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusServiceUnavailable)
```
fracasula, line 1979: Should we add the number of calls here and assert at the end of the test?

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -461,6 +482,99 @@ func (u *Client) doPost(ctx context.Context, rawJSON []byte, url string, labels
...
+// It is the cp-router test path's counterpart to the event-processing path
+// ([Client.doPost]): both target the pyt-{workspaceID} deployment and share the
+// client's maxRetry/maxRetryBackoffInterval settings, but this call is time-boxed
+// by ctx (cp-router's ~60s test deadline). Callers that need a quicker cold-start
+// recovery build the client with [WithMaxRetryBackoffInterval]/[WithMaxRetry] to
+// shrink the backoff and size the retry budget. It retries only on cold-start
+// signals (connection refused, DNS failure, 502/503);
+// any other response is returned as-is for the caller to pass through.
+func (u *Client) forwardTest(ctx context.Context, workspaceID, path string, payload []byte) (int, []byte, error) {
+	url := u.userTransformBaseURL("python", "", workspaceID) + path
```
fracasula, line 540: We should check the configuration before forwarding, right? The configuration is reloadable and we could return a specific error that we can catch on forward.go like:

```go
	if !u.config.perWorkspacePyTEnabled.Load() || u.config.forMirroring {
		return 0, nil, ErrPerWorkspacePyTNotEnabled
	}
```

And then in forward.go upon receiving that error we could return:

```go
status.Error(codes.FailedPrecondition, "per-workspace PyT is not enabled")
```

Wdyt?

processor/cpservice/forward.go
```diff
@@ -0,0 +1,96 @@
...
+// only when it is currently at zero replicas (get-before-set), keeping the call
+// idempotent and a no-op when the deployment is already running.
+//
+// Concurrent Forwards for the same workspace are collapsed by singleflight so a
+// burst of test requests triggers a single scale check — not one per request —
+// avoiding redundant cluster-wide List calls against the k8s API. Requests for
+// different workspaces key on distinct IDs and run independently. (The shared
+// call runs on the first caller's ctx; a fast scale makes that window
+// negligible, and on error the caller retries the whole RPC.)
+func (s *Service) ensureScaled(ctx context.Context, workspaceID string) error {
```
fracasula, line 79: With this new `ensureScaled` you're calling GetReplicaCount and then SetReplicaCount in succession. They are both doing a List() (see `s.resolveNamespace(ctx, name)`). There could be some room for optimization?

ktgowtham (approved):

processor/cpservice/forward.go
mihir20: added a validation layer

processor/cpservice/forward_test.go
mihir20: done

processor/cpservice/forward_test.go
mihir20: added a test for when scaler fails

processor/internal/transformer/user_transformer/user_transformer_test.go
mihir20: done

processor/internal/transformer/user_transformer/user_transformer.go
mihir20: makes sense

processor/cpservice/forward.go
mihir20: updated instead of get and then set, replaced with a single EnsureScaled interface

fracasula (approved):

PR #7147, opened 2026-07-02
chore: add contract tests for test endpoints in pytransformer

mihir20:
# Description
Adds a contract-test suite that pins the wire contract of the four control-plane transformation-test endpoints, comparing the **new architecture** (`rudder-pytransformer`) against the **old architecture**
(`rudder-transformer` in test mode) byte-for-byte.

These tests guard the exact `usertransformer.Client` → `pyt` path that `cpservice.Forward` uses in production, so any behavioral drift between the old and new architectures is caught before it reaches the
control plane.

## Endpoints under contract

| New arch (`rudder-pytransformer`) | Old arch (`rudder-transformer`) |
  |-----------------------------------|----------------------------------|
| `POST /test`         | `POST /transformation/test` |
| `POST /testRun`      | `POST /transformation/testRun` |
| `POST /test-library` | `POST /transformationLibrary/test` |
| `POST /extract-libs` | `POST /extractLibs` |

## How it works

- **Old architecture** — `rudder-transformer` runs with `TRANSFORMER_TEST_MODE=true`: it deploys a per-request OpenFaaS function whose `fprocess` carries the inline code, invokes it, then deletes it. AST
  routes hit a long-lived `fn-ast` function.
- **Dynamic mock OpenFaaS gateway** — unlike the fixed-proxy mock, it honours function deployments: `POST /system/functions` starts a real `openfaas-flask-base` container so the inline code actually
  executes; invocations route to the per-function container and `DELETE` purges it.
- **New architecture** — requests flow through `usertransformer.Client.Test/TestRun/TestLibrary/ExtractLibs` — the same methods `cpservice.Forward` calls in production — covering the exact client → `pyt`
  path.
- Responses are compared **byte-for-byte**, except the documented AST import-violation divergence where `pyt` surfaces the runtime's wording (`"Import of 'os' is not allowed. ..."`) rather than `fn-ast`'s
  (`"Unpermitted import(s). ..."`) — the runtime is treated as the source of truth.

## Coverage

Roughly 35 subtests spanning the four endpoints, including:

- **`/test`** — inline code with logs, library fetching from the config backend, `getCredential` exposure, per-event errors kept inline at HTTP 200, duplicate `messageId`s, metadata without `messageId`,
  compile errors (HTTP 400), verbatim missing-code/missing-events errors, frame-introspection code, library-fetch failures.
- **`/testRun`** — metadata echo per event, dependency/library/credential resolution, per-event errors kept inline with metadata, duplicate `messageId`s, verbatim missing-code/missing-events errors.
- **`/test-library`** — import maps for valid code, rejection of non-whitelisted / `urllib.request` / relative imports (matching the old runtime), module-path keying, syntax errors, verbatim missing-code
  error.
- **`/extract-libs`** — non-whitelisted extraction with validation off, additional libraries under validation, dotted imports, rejection of relative and non-whitelisted imports, verbatim missing-code
  error.

- `startRudderTransformer` now accepts optional extra environment variables (e.g. `TRANSFORMER_TEST_MODE=true`).
- `rudder-pytransformer` container pinned to tag `pr-109` for this suite; to move to `main` before merging this).

## Linear Ticket

pipe-3164

## Security

### DIFF
--- a/integration_test/pytransformer_contract/test_endpoints_contract_test.go
+++ b/integration_test/pytransformer_contract/test_endpoints_contract_test.go
@@ -5,22 +5,23 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httptest"
-	"strconv"
+	"runtime"
 	"strings"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/ory/dockertest/v3"
+	"github.com/ory/dockertest/v3/docker"
 	"github.com/stretchr/testify/require"
 
 	"github.com/rudderlabs/rudder-go-kit/config"
 	"github.com/rudderlabs/rudder-go-kit/jsonrs"
 	"github.com/rudderlabs/rudder-go-kit/logger"
 	"github.com/rudderlabs/rudder-go-kit/stats"
-	kithelper "github.com/rudderlabs/rudder-go-kit/testhelper"
 	"github.com/rudderlabs/rudder-go-kit/testhelper/docker/resource/registry"
 
 	"github.com/rudderlabs/rudder-server/processor/usertransformer"
@@ -113,8 +114,15 @@ func (g *dynamicFaasGateway) handle(w http.ResponseWriter, r *http.Request) {
 		if name == "" {
 			name = req.Name
 		}
-		g.t.Logf("DynamicFaasGateway: deploying %q with fprocess %q", name, req.EnvProcess)
-		url, container, err := startOpenFaasFlaskFprocess(g.t, g.pool, req.EnvProcess)
+		// rudder-transformer builds the fprocess with the config-backend URL as
+		// the host sees it (127.0.0.1 on Linux). The flask container runs in its
+		// own bridge namespace, so rewrite host-local addresses to
+		// host.docker.internal so the function can reach the host. On macOS the
+		// URL is already host.docker.internal, so this is a no-op there.
+		envProcess := strings.ReplaceAll(req.EnvProcess, "127.0.0.1", "host.docker.internal")
+		envProcess = strings.ReplaceAll(envProcess, "localhost", "host.docker.internal")
+		g.t.Logf("DynamicFaasGateway: deploying %q with fprocess %q", name, envProcess)
+		url, container, err := startOpenFaasFlaskFprocess(g.t, g.pool, envProcess)
 		if err != nil {
 			g.t.Errorf("DynamicFaasGateway: starting flask container for %q: %v", name, err)
 			w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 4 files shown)


integration_test/pytransformer_contract/test_endpoints_contract_test.go
```diff
@@ -0,0 +1,1133 @@
...
+}
+
+// register makes the gateway route /function/{name} traffic to url. The
+// container may be nil for externally-managed functions (e.g. fn-ast, whose
+// lifecycle is owned by the test body).
+func (g *dynamicFaasGateway) register(name, url string, container *dockertest.Resource) {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	g.fnURLs[name] = url
+	if container != nil {
```
fracasula: Do we need this check?

integration_test/pytransformer_contract/test_endpoints_contract_test.go
```diff
@@ -0,0 +1,1133 @@
...
+	) (int, []byte) {
+		t.Helper()
+		body, err := jsonrs.Marshal(payload)
+		require.NoError(t, err)
+		statusCode, respBody, err := method(ctx, workspaceID, body)
+		require.NoError(t, err)
+		return statusCode, respBody
+	}
+
+	decodeFlow := func(t *testing.T, body []byte) testFlowResponse {
```
fracasula: Can we cleanup the TestPyTransformerTestEndpointsa bit? These can easily be helpers that live outside the TestPyTransformerTestEndpoints function. Also it'd be beneficial if we can start the test file with the actual tests. We don't care much about the setup so having to open a test file just to scroll until we get to what we need isn't ideal. The goal would be have the test as concise and readable as possible because these are already difficult to review.

integration_test/pytransformer_contract/test_endpoints_contract_test.go
```diff
@@ -0,0 +1,1133 @@
...
+			oldStatus, oldBody := callOld(t, "/transformation/testRun", payload)
+			newStatus, newBody := callNew(t, client.TestRun, payload)
+
+			require.Equal(t, http.StatusBadRequest, newStatus, "body: %s", newBody)
+			require.Equal(t, oldStatus, newStatus)
+			require.Equal(t, "Error: Invalid Request. Missing parameters in transformation code block", decodeError(t, newBody))
+			require.Equal(t, decodeError(t, oldBody), decodeError(t, newBody))
+		})
+
+		t.Run("should return the verbatim missing-events error", func(t *testing.T) {
```
fracasula, line 547: Aren't we missing invalid code that shouldn't compile for "/transformation/testRun"?

integration_test/pytransformer_contract/test_endpoints_contract_test.go
```diff
@@ -0,0 +1,1133 @@
...
+	// Each flask container runs in its own bridge-network namespace — never host
+	// networking — because of-watchdog binds a HARD-CODED Prometheus metrics
+	// port 8081 that is not configurable via env. Under host networking two
+	// flask containers (the long-lived fn-ast and a per-request fn-test) share
+	// the host namespace and collide on 8081: of-watchdog panics with
+	// "bind: address already in use", the container exits, and the invoke port
+	// never comes up ("connection refused"). A per-container namespace keeps
+	// each 8081 private. host.docker.internal (host-gateway) lets the function
+	// reach the host's config backend from inside the namespace.
+	container, err := pool.RunWithOptions(&dockertest.RunOptions{
```
fracasula: `bc_helpers_test.go` already centralizes the Linux-vs-macOS docker quirks:
`newContainerConfig` (host networking on Linux, bridge + `host.docker.internal` +
`PortBindings` on macOS), `containerConfig.url()`, `portStr()`, and `toContainerURL()` /
`dockertesthelper.ToInternalDockerHost`. The existing single-container `startOpenFaasFlask` (`bc_helpers_test.go:409`) uses them, and the pytransformer container solves its metrics-port collision by allocating a unique `METRICS_PORT` env (`bc_helpers_test.go:503-508`).

The new `startOpenFaasFlaskFprocess` instead **hardcodes the macOS-style bridge config for all platforms**, re-implementing `newContainerConfig`'s darwin branch inline.

**Fix:** Give `newContainerConfig` a "force bridge networking regardless of GOOS" option and build `startOpenFaasFlaskFprocess` on top of it (reusing `cfg.ExtraHosts`, `cfg.PortBindings`, `cfg.url()`); route the fprocess host rewrite through the shared `ToInternalDockerHost` helper so
the address-rewriting rules live in one place.

fracasula: I asked specifially the agent to look into this because I remember that we already had helpers that differentiates between Darwin and Linux so we might be duplicating logic unnecessarily here.

fracasula (approved):
Approving with some small asks 👍

ktgowtham (approved):

PR #7151, opened 2026-07-06
fix: update null to #N/A for salesforce bulk upload

ItsSudip:

# Description

Salesforce Bulk API 2.0 treats an empty CSV cell as "leave the field unchanged"; setting a field to null requires the literal `#N/A` sentinel ([Salesforce docs](https://developer.salesforce.com/docs/atlas.en-us.api_bulk_v2.meta/api_bulk_v2/datafiles_prepare_csv.htm)). The `SALESFORCE_BULK_UPLOAD` destination serialized null trait values as empty cells, so null overwrites were silently ignored — fields never got cleared in Salesforce.

## Changes

- `Transform` now encodes JSON `null` trait values as the `#N/A` sentinel (new `salesforceNullSentinel` const), so they reach the upload CSV as the documented field-clear instruction.
- Fields absent from an event are unaffected: they still render as empty cells, which Salesforce ignores. Only explicit nulls clear fields.
- Corrected the doc comment on the shared `common.FormatCSVValue` helper, which wrongly claimed Salesforce Bulk treats empty cells as null (no behavior change there).

## Tests

- Added `TestSalesforceBulk_Transform_NullTraitsBecomeNASentinel`: null traits are encoded as `#N/A`; non-null traits and the externalId pass through untouched.
- Updated `TestSalesforceBulk_createCSVFile_NullValues`: `#N/A` reaches the CSV cell verbatim; absent fields render as empty cells.

## Linear Ticket

Resolves [INT-6566](https://linear.app/rudderstack/issue/INT-6566/null-overwrites-dont-work-on-current-salesforce-sync)

## Security


router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/utils_test.go
```diff
@@ -186,22 +186,23 @@ func TestSalesforceBulk_createCSVFile_NumericNoScientificNotation(t *testing.T)
...
+	// the CSV cell unchanged so Salesforce clears the field. An absent field
+	// renders as an empty cell, which Salesforce ignores (leaves unchanged).
 	jobs := []SalesforceAsyncJob{
 		{
 			Message: map[string]any{
-				"Email":     "middle@example.com",
+				"Email":     "clear@example.com",
 				"FirstName": "Mid",
-				"LastName":  nil,
+				"LastName":  salesforceNullSentinel,
```
saikumarrs: Let's not reuse the constants from the code in the tests.

saikumarrs (approved):

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/utils_test.go
saikumarrs: Inadvertent behavior changes go unnoticed otherwise.

ItsSudip: Addressed

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk_upload_test.go
```diff
@@ -139,6 +139,45 @@ func TestSalesforceBulk_Transform(t *testing.T) {
 	}
 }
 
+func TestSalesforceBulk_Transform_NullTraitsBecomeNASentinel(t *testing.T) {
```
achettyiitr: Can we rather have a check on the output CSV file, because that's what's getting uploaded to Salesforce? `parsed.Message` is still intermediary.

ItsSudip: Done — replaced the Transform-level assertion with an Upload-level test that parses the CSV passed to UploadData: nulls assert as `#N/A` cells, absent fields as empty cells.

router/batchrouter/asyncdestinationmanager/salesforce-bulk-upload/salesforce_bulk.go
```diff
@@ -103,6 +103,13 @@ func (s *Uploader) Transform(job *jobsdb.JobT) (string, error) {
 	// Add all traits to message
 	if traits.Exists() && traits.IsObject() {
 		traits.ForEach(func(key, value gjson.Result) bool {
+			// Salesforce Bulk API ignores empty CSV cells (leaves the field
```
achettyiitr, line 106: Instead of keeping this check here, can we move it to createCSVFile? Because that's the output file that will get uploaded to CSV.

```
for _, job := range input {
		row := make([]string, len(headers))
		for key, value := range job.Message {
			if idx, ok := headerIndex[key]; ok {
				if value == nil {
					row[idx] = salesforceNullSentinel
					continue
				}
				formattedValue, err := common.FormatCSVValue(value)
				if err != nil {
					return "", 0, fmt.Errorf("formatting CSV cell %q: %w", key, err)
				}
				row[idx] = formattedValue
			}
		}
		if err := writer.Write(row); err != nil {
			return "", 0, fmt.Errorf("writing CSV row: %w", err)
		}
	}
```

achettyiitr: Discussed, as we have it panned to handled it in Transform.

achettyiitr (approved):

PR #7159, opened 2026-07-07
fix: classify reverse-ETL sources from config for MAR metering

am6010:
## What & why

MAR (Monthly Active Records) metering for reverse-ETL / Audiences sources produced **no data** in clustered/production deployments: `activation_records_reports` stayed empty and the MAR API returned `[]`.

The reporter gated reverse-ETL classification on the job's `source_category` **param**. That param is populated on the gateway web/auth path, but is **empty on the clustered `/internal/v1/batch` ingestion path** (rudder-sources → ingestion-svc → Pulsar → srcrouter → gateway). So every reverse-ETL record was silently skipped — no rows, no skip stats.

## Change

Classify reverse-ETL (warehouse) sources from the **source config** (`SourceDefinition.Category`) instead of the job param:

- `processor` snapshots a `source_id → category` map once per batch (single lock, category strings only) and passes a lookup `func(sourceID string) string` into the reporter.
- `records_reporter.GenerateReportsFromJobs` gates on that lookup (`== "warehouse"`), still reading the fingerprint from the raw pre-strip job.

Path-independent (works for `/internal/v1/batch`, `/internal/v1/retl`, single-node/mini) and fail-closed: an unknown/missing `source_id` resolves to `""` and is skipped; a spoofed `source_category` param can't force metering.

This supersedes an earlier gateway-side approach (stamping `source_category` in the internal-batch handler). Classifying at the consumer keeps the change contained to the MAR feature and independent of ingestion topology, and mirrors the sibling tracked-users reporter, which already takes a source-config lookup.

## Tests

`records_reporter_test.go`: warehouse source metered with an empty `source_category` param; non-warehouse source skipped; a spoofed `source_category=warehouse` param does not win over config; unknown source skipped fail-closed.

## Validation

Verified end-to-end in dev (`profiles-qa`): reverse-ETL sync → rows in `activation_records_reports` → flushed to the reporting store → MAR API returns per-connection counts (was `[]`).

Linear: [PRO-5810](https://linear.app/rudderstack/issue/PRO-5810)

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -189,6 +189,7 @@ type Handle struct {
 		eventAuditEnabled                         map[string]bool
 		credentialsMap                            map[string][]types.Credential
 		nonEventStreamSources                     map[string]bool
+		sourceIdCategoryMap                       map[string]string
 		enableConcurrentStore                     config.ValueLoader[bool]
 		userTransformationMirroringSanitySampling config.ValueLoader[float64]
 		userTransformationMirroringFireAndForget  config.ValueLoader[bool]
@@ -851,6 +852,7 @@ func (proc *Handle) backendConfigSubscriber(ctx context.Context) {
 			eventAuditEnabled            = make(map[string]bool)
 			credentialsMap               = make(map[string][]types.Credential)
 			nonEventStreamSources        = make(map[string]bool)
+			sourceIdCategoryMap          = make(map[string]string)
 			connectionConfigMap          = make(map[connection]backendconfig.Connection)
 		)
 		for workspaceID, wConfig := range config {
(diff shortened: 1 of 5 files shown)


processor/processor.go
```diff
@@ -2407,7 +2422,12 @@ func (proc *Handle) pretransformStage(partition string, preTrans *preTransformat
 	// GenerateReportsFromJobs reads context.activation from the raw job payloads
 	// (left intact here so metering still works); the destination-bound copies are
 	// already stripped per-event in preprocessStage.
-	activationRecordsReports := proc.activationRecordsReporter.GenerateReportsFromJobs(preTrans.jobList)
+	// Snapshot source_id -> category once per batch (single lock, no per-job SourceT copy,
+	// no error log on unknown sources); the reporter classifies reverse-ETL sources off it.
+	sourceCategoriesBySourceID := proc.getSourceCategoriesBySourceID()
+	activationRecordsReports := proc.activationRecordsReporter.GenerateReportsFromJobs(preTrans.jobList, func(sourceID string) string {
```
itsmihir: why not just pass sourceCategoriesBySourceID directly? what's the purpose of wrapping it with a function?

itsmihir: Another option is to pass `proc.getSourceCategoriesBySourceID()` to GenerateReportsFromJobs and we call it once before the for loop.

mihir20 (approved):

itsmihir (approved):

PR #7161, opened 2026-07-08
fix(gateway/webhook): return 404 for unregistered source instead of blocking on nil channel (#6999)

### LINKED ISSUE #6999: bug: nil channel send in webhook RequestHandler blocks HTTP handler forever (DoS)

kotwal-itpro:
Fixes #6999.

## Problem

`RequestHandler` looked up the per-source request channel from `webhook.requestQ` and immediately sent on it. When the source type had not been `Register()`ed — a condition that can occur when `webhookV2` is disabled and the lazy registration in `processBackendConfig` has not yet run for a given source, or when a request arrives for a source type this instance does not handle — the map lookup returned a nil channel and the send blocked forever. The HTTP handler goroutine never returned, the client hung, and a goroutine leaked per unregistered request. This is the DoS described in #6999.

## Fix

- Comma-ok the map lookup at `gateway/webhook/webhook.go:228`.
- Release the read lock before writing the response (existing lock discipline preserved for the success path).
- Fail fast with the existing `response.InvalidWebhookSource` → HTTP 404 mapping (already defined in `gateway/response`).

## Test

Adds `TestWebhookRequestHandlerReturns404WhenSourceNotRegistered` which:
- Disables `webhookV2HandlerEnabled` (the pre-condition under which the bug reproduces on current HEAD).
- Deliberately omits `Register()`.
- Sends the request in a goroutine guarded by a 2-second `time.After` so the test fails hard (rather than hanging) if the handler ever regresses back to blocking on nil channel.
- Asserts `http.StatusNotFound` and body contains `response.InvalidWebhookSource`.

Passes under `go test -race ./gateway/webhook/...`. Full webhook test suite still passes.

### DIFF
--- a/gateway/webhook/webhook.go
+++ b/gateway/webhook/webhook.go
@@ -225,7 +225,28 @@ func (webhook *HandleT) RequestHandler(w http.ResponseWriter, r *http.Request) {
 		webhook.Register(sourceDefName)
 	}
 	webhook.requestQMu.RLock()
-	requestQ := webhook.requestQ[sourceDefName]
+	requestQ, ok := webhook.requestQ[sourceDefName]
+	if !ok {
+		// The source type was not registered — either because
+		// `webhookV2HandlerEnabled` is false and the lazy registration in
+		// `processBackendConfig` has not yet run for this source, or because
+		// the source type is not one this instance handles. Sending on the
+		// nil channel returned by the map lookup would block this HTTP handler
+		// goroutine forever, leaving the client hanging and leaking a
+		// goroutine per unregistered request. Fail fast with 404 instead.
+		webhook.requestQMu.RUnlock()
+		stat := webhook.statReporterCreator(arctx, reqType)
+		stat.RequestFailed("invalidWebhookSource")
+		stat.Report(webhook.stats)
+		webhook.failRequest(
+			w,
+			r,
+			response.GetStatus(response.InvalidWebhookSource),
+			response.GetErrorStatusCode(response.InvalidWebhookSource),
+		)
+		webhook.ackCount.Add(1)
+		return
+	}
 	requestQ <- &req
 	webhook.requestQMu.RUnlock()
 
--- a/gateway/webhook/webhook_test.go
+++ b/gateway/webhook/webhook_test.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"testing"
 	"testing/iotest"
+	"time"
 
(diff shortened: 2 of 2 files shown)


kotwal-itpro:
Hi team — noticed this got auto-closed by the stale bot after 7 days of inactivity, though it never received a maintainer review. The CLA is signed, the fix addresses the nil-channel DoS in the webhook handler, and it includes a regression test (`TestWebhookRequestHandlerReturns404WhenSourceNotRegistered`) that passes locally, including race detection. Would you be able to reopen this for review? Happy to rebase or update if anything's changed on master since. Thanks!

kotwal-itpro:
Since this couldn't be reopened via the API, I've re-submitted the identical (rebased) fix as #7255. Please review there. Closing loop here.

PR #7166, opened 2026-07-08
fix: detect server-side snowpipe channel failures via per-channel status at stuck threshold

shekhar-rudder:
## Description

The Snowflake Ingest SDK has a bug ([snowflake-ingest-java#1154](https://github.com/snowflakedb/snowflake-ingest-java/issues/1154)) where the **batch** channel-status method silently ignores server-side failure status codes (e.g. `status_code=35`) and reports channels as `valid=true` even when Snowflake has flagged them broken.

Our snowpipe-streaming consumer polls channel status via the **bulk-status** endpoint. When a channel is silently broken, bulk-status keeps returning `valid=true`, so the consumer never enters its normal invalid-channel recovery path. Instead it keeps polling until the **15-minute stuck-pipeline threshold** trips, at which point it logs `"Stuck snowpipe pipeline detected"` (an alert) and fails all in-progress jobs — a 15-minute delay, misreported as a stuck pipeline.

## Change

At the stuck-pipeline threshold in `Manager.Poll`, each still-in-progress channel is re-checked via the **single-channel** status endpoint (`GetStatus`), which correctly surfaces server-side failures (the per-channel method throws, and the snowpipe service returns `success=false`/`valid=false`).

- **Confirmed invalid** channels (`!Valid || !Success`) are failed through the normal failed-import path (channel deleted, jobs retried next batch) **without** the `"Stuck snowpipe pipeline detected"` alert.
- **Genuinely stuck** channels (single-channel check still reports valid) keep the existing alert + `"events still in progress after threshold…"` failure reason.
- **Inconclusive** checks (the `GetStatus` call itself errors) are treated as stuck — fail-safe: the alert is suppressed only on a positive invalid confirmation.

This re-wires `GetStatus` back into the `api` interface and `apiAdapter` (the reverse of #6992, which had removed the per-channel path when bulk-status became the default). The lower-level `internal/api/api.go` `GetStatus` was kept intact by that commit, so no new client code is needed. No change is required in `rudder-snowpipe-clients` — its single-channel `/channels/{id}/status` endpoint already surfaces broken channels. Once the upstream SDK fix ships, this per-channel recheck can be removed.

## Tests

- Re-added `mockAPI.GetStatus` + `getStatusOutputMap`.
- Updated the existing "marks in-progress imports as failed after threshold" test so the single-channel check reports valid (genuinely-stuck path unchanged).
- Added "recovers invalid channel at threshold without stuck-pipeline alert" (invalid → per-channel-invalid reason, no alert).
- Added "treats inconclusive single-channel check at threshold as stuck" (`GetStatus` errors → stuck reason, fail-safe).

Verified with `go build`, full-package `go test`, `gofmt`, and `go vet`.

## Linear Ticket

- Resolves INT-6530

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/apiadapter.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/apiadapter.go
@@ -15,6 +15,7 @@ const (
 	createChannelAPI = "create_channel"
 	deleteChannelAPI = "delete_channel"
 	insertAPI        = "insert"
+	statusAPI        = "status"
 	bulkStatusAPI    = "bulk_status"
 )
 
@@ -110,6 +111,23 @@ func (a *apiAdapter) Insert(ctx context.Context, channelID string, insertRequest
 	}
 	tags["success"] = strconv.FormatBool(resp.Success)
 	tags["code"] = resp.Code
+	return resp, nil
+}
+
+func (a *apiAdapter) GetStatus(ctx context.Context, channelID string) (*model.StatusResponse, error) {
+	a.logger.Debugn("Getting status",
+		logger.NewStringField("channelId", channelID),
+	)
+
+	tags := a.defaultTags(statusAPI)
+	defer a.recordDuration(tags)()
+
+	resp, err := a.api.GetStatus(ctx, channelID)
+	if err != nil {
+		tags["success"] = "false"
+		return nil, err
+	}
+	tags["success"] = strconv.FormatBool(resp.Success)
 	return resp, nil
 }
 
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/snowpipestreaming.go
@@ -701,15 +701,39 @@ func (m *Manager) Poll(_ context.Context, pollInput common.AsyncPoll) common.Pol
 			return common.PollStatusResponse{InProgress: true}
 		}
 
-		m.logger.Warnn("Stuck snowpipe pipeline detected",
-			logger.NewDurationField("duration", duration),
-			logger.NewDurationField("threshold", threshold),
-			logger.NewStringField("importID", pollInput.ImportId),
-		)
+		// The bulk-status endpoint can silently report a server-side broken channel as valid=true
+		// (snowflake-ingest-java#1154), so a channel that looks in-progress here may actually be invalid.
+		// Re-check each in-progress channel via the single-channel status endpoint, which surfaces the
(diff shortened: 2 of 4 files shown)


shekhar-rudder:
## ✅ End-to-end verification (real Snowflake)

Verified this change **end-to-end** — not just the unit tests — by running the PR image through a full local stack and driving all three `Manager.Poll` threshold branches against real infrastructure.

**Setup**
- rudder-server from the CI image `rudderlabs/develop-rudder-server:pr-7166`, full data plane in `rudder-mini`.
- JS source → `SNOWPIPE_STREAMING` destination with real Snowflake key-pair creds.
- **Real `rudder-snowpipe-clients` + real Snowflake**: channels created and rows actually inserted (blobs uploaded + registered; *"All events flushed successfully"*).
- SDK bug ([snowflakedb/snowflake-ingest-java#1154](https://github.com/snowflakedb/snowflake-ingest-java/issues/1154))'s status divergence (which a healthy backend won't produce on demand) reproduced with a small **fault-injection proxy** between rudder-server and the real snowpipe-clients: it forwards `create`/`insert`/`delete` untouched, forces `bulk-status` to report in-progress, and rewrites only the single-channel `/status` per scenario. `stuckPipelineThresholdInMinutes=0` so the recheck fires on the first stuck poll.

**Results** — each branch exercised through the real running server (real `apiAdapter.GetStatus` + HTTP path, which the mocked unit tests don't cover):

| Scenario | single-channel `/status` | Observed |
|---|---|---|
| **Confirmed invalid** | `valid:false` | `WARN "Channel reported invalid at stuck threshold; recovering without stuck-pipeline alert"`; job reason `channel reported invalid by per-channel status check after threshold`; **`Stuck snowpipe pipeline detected` count = 0** ✅ |
| **Genuinely stuck** | `valid:true` | `WARN "Stuck snowpipe pipeline detected"` + reason `events still in progress after threshold: 10s > 0s` ✅ |
| **Inconclusive** | HTTP 500 (errors) | `WARN "Failed to get single-channel status at stuck threshold; treating as stuck"` (after retries) → then stuck alert + stuck reason ✅ |

Fail-safe confirmed: the alert is suppressed **only** on a positive `valid:false` — an errored/unavailable `GetStatus` still raises the stuck alert.

**Notes for anyone reproducing**
- Vault stores the `privateKey` with literal `\n`; convert to real newlines or Snowflake rejects it as invalid PEM.
- The destination's `warehouse` must reach the data plane via the config-backend **API** (a raw DB edit won't recompute the namespace config); `CREATE SCHEMA/TABLE` needs `USAGE` on a warehouse.

The `rudder-snowpipe-clients` service + this verification recipe are being added to `rudder-mini` so this is turnkey next time.

achettyiitr (approved):

maheshkutty (approved):

PR #7167, opened 2026-07-08
fix(warehouse): use pod identity for rudder-storage flow; keep S3-copy user fallback for user S3 destinations

shekhar-rudder:
## What

Stops the `useRudderStorage` flow from authenticating with a broadly-scoped static credential, while **preserving** the shared S3-copy user as the fallback for customer-configured S3 destinations that supply no auth of their own.

- **Rudder-storage flow → AWS SDK default credential chain.** `misc.GetRudderObjectStorageConfig` and the `useRudderStorage` branch of `warehouseutils.CreateAWSSessionConfig` no longer read `RUDDER_AWS_S3_COPY_USER_ACCESS_KEY_ID/KEY`. Access to the intermediate `rudder-warehouse-storage*` buckets now resolves through the AWS SDK default credential chain — i.e. the pod identity role (IRSA) — instead of the static copy-user credential.
- **Non-rudder-storage flow → copy-user fallback kept.** `misc.GetObjectStorageConfig` and the non-rudder-storage branch of `CreateAWSSessionConfig` still fall back to the shared copy-user credentials when a user-configured S3 destination provides neither an IAM role ARN nor access keys. Customers grant this user write access to their bucket precisely so RudderStack can deliver data when they don't supply their own credentials, so this path must be retained.
- **Fix the warehouse COPY temp-credential path (required by the change above).** `GetTemporaryS3Cred` now retrieves already-temporary credentials directly from the default chain when there are no static keys (`sessionConfig.AccessKeyID == ""`), in addition to the role-based case. This is needed because the rudder-storage flow now returns a credential-less session config: `GetSessionToken` rejects the temporary credentials the default chain hands back, so the pod-identity path would otherwise fail. Customer-supplied long-term static keys still go through STS `GetSessionToken` as before.

## Why

Only the `useRudderStorage` flow should stop using the shared copy user — that flow targets RudderStack-owned buckets and should authenticate via pod identity (IRSA). The non-rudder-storage flow is different: our customers explicitly grant the copy user write access to *their* S3 bucket for the case where they configure no IAM role or access keys, so removing the fallback there would break delivery to those destinations.

> Note: an earlier revision of this PR removed the copy user from **both** flows (making missing auth a hard `ErrS3MissingCredentials` failure). That was incorrect for the non-rudder-storage flow and has been reverted — `GetObjectStorageConfig` keeps its original `map[string]any` signature and the sentinel error was dropped.

## Infra pre-conditions

- The pod identity (IRSA) role on the warehouse pods **already has** read/write on `rudder-warehouse-storage*`, so the `useRudderStorage` upload/COPY flow authenticates via the AWS SDK default credential chain with no further infra change.
- `RUDDER_AWS_S3_COPY_USER_ACCESS_KEY_ID/KEY` **must remain configured**, since the non-rudder-storage fallback still consumes them.

## Tests

- Unit tests updated and passing: `TestGetObjectStorageConfig`, `TestGetRudderObjectStorageConfig` (`utils/misc`), `TestCreateAWSSessionConfig` (`warehouse/utils`). They assert that the rudder-storage flow injects no static creds, and that the non-rudder-storage flow falls back to the copy-user credentials when no role/keys are present.
- Restored the `useRudderStorage` warehouse integration cases for **redshift** and **snowflake**. Instead of the app-specific copy-user envs, they now supply generic `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` envs so the AWS SDK default credential chain resolves them automatically — exercising the same credential path the `useRudderStorage` flow uses in production (pod identity / IRSA).
- Restored the **deltalake** `useRudderStorage` case too. Deltalake's COPY runs on Databricks via the S3A `TemporaryAWSCredentialsProvider`, which requires a non-empty session token. In production the `useRudderStorage` path uses IRSA/pod-identity, whose default-chain credentials are already temporary and carry a session token. The test mirrors this: it mints a session token from the static test keys via STS `GetSessionToken` and exports it through the standard `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN` envs — no app-specific credential env reintroduced.
- `go build ./...` and `go vet` clean across affected packages.

### DIFF
--- a/utils/misc/misc.go
+++ b/utils/misc/misc.go
@@ -48,12 +48,6 @@ var (
 	regexGwNonHaOrProcessor = regexp.MustCompile(`^.*-\d+$`)
 )
 
-// ErrS3MissingCredentials is returned when a user-configured (non-rudder-storage)
-// S3 location is configured with neither an IAM role ARN nor access keys. This
-// applies to S3 destinations as well as S3 used for warehouse staging/COPY; such
-// locations must not fall back to any RudderStack-owned credentials.
-var ErrS3MissingCredentials = errors.New("S3 access requires an IAM role ARN or access keys")
-
 const (
 	// RFC3339Milli with milli sec precision
 	RFC3339Milli          = "2006-01-02T15:04:05.000Z07:00"
@@ -650,6 +644,17 @@ func HasAWSRegionInConfig(config any) bool {
 	return true
 }
 
+// GetRudderObjectStorageAccessKeys returns the credentials of the shared
+// RudderStack-owned S3 "copy user". These are used only as a fallback for
+// customer-configured (non-rudder-storage) S3 locations that provide neither an
+// IAM role ARN nor access keys: customers grant this user write access so
+// RudderStack can deliver data to their bucket. The rudder-storage flow no
+// longer uses these keys — it authenticates via the AWS SDK default credential
+// chain (pod identity / IRSA).
+func GetRudderObjectStorageAccessKeys() (accessKeyID, accessKey string) {
+	return config.GetStringVar("", "RUDDER_AWS_S3_COPY_USER_ACCESS_KEY_ID"), config.GetStringVar("", "RUDDER_AWS_S3_COPY_USER_ACCESS_KEY")
+}
+
 func GetRudderObjectStoragePrefix() (prefix string) {
 	return config.GetStringVar(config.GetNamespaceIdentifier(), "RUDDER_WAREHOUSE_BUCKET_PREFIX")
 }
@@ -689,22 +694,26 @@ type ObjectStorageOptsT struct {
 	WorkspaceID                 string
 }
 
-func GetObjectStorageConfig(opts ObjectStorageOptsT) (map[string]any, error) {
+func GetObjectStorageConfig(opts ObjectStorageOptsT) map[string]any {
 	objectStorageConfigMap := opts.Config.(map[string]any)
 	if opts.UseRudderStorage {
-		return GetRudderObjectStorageConfig(opts.RudderStoragePrefixOverride), nil
+		return GetRudderObjectStorageConfig(opts.RudderStoragePrefixOverride)
 	}
 	if opts.Provider == "S3" {
-		if !HasAWSRoleARNInConfig(objectStorageConfigMap) &&
-			!HasAWSKeysInConfig(objectStorageConfigMap) {
-			return nil, ErrS3MissingCredentials
-		}
 		clonedObjectStorageConfig := make(map[string]any)
 		maps.Copy(clonedObjectStorageConfig, objectStorageConfigMap)
 		clonedObjectStorageConfig["externalID"] = opts.WorkspaceID
+		if !HasAWSRoleARNInConfig(objectStorageConfigMap) &&
+			!HasAWSKeysInConfig(objectStorageConfigMap) {
+			// Non-rudder-storage S3 destination without an IAM role ARN or access
+			// keys: fall back to the shared RudderStack S3 copy-user credentials.
+			// Customers grant this user write access to their bucket so RudderStack
+			// can deliver data when they don't supply their own credentials.
+			clonedObjectStorageConfig["accessKeyID"], clonedObjectStorageConfig["accessKey"] = GetRudderObjectStorageAccessKeys()
+		}
 		objectStorageConfigMap = clonedObjectStorageConfig
 	}
-	return objectStorageConfigMap, nil
+	return objectStorageConfigMap
 }
 
 // GetParsedTimestamp returns the parsed timestamp
--- a/utils/misc/misc.go
+++ b/utils/misc/misc.go
@@ -48,6 +48,12 @@ var (
 	regexGwNonHaOrProcessor = regexp.MustCompile(`^.*-\d+$`)
 )
 
+// ErrS3MissingCredentials is returned when a user-configured (non-rudder-storage)
+// S3 location is configured with neither an IAM role ARN nor access keys. This
+// applies to S3 destinations as well as S3 used for warehouse staging/COPY; such
+// locations must not fall back to any RudderStack-owned credentials.
+var ErrS3MissingCredentials = errors.New("S3 access requires an IAM role ARN or access keys")
+
 const (
 	// RFC3339Milli with milli sec precision
 	RFC3339Milli          = "2006-01-02T15:04:05.000Z07:00"
@@ -644,10 +650,6 @@ func HasAWSRegionInConfig(config any) bool {
 	return true
 }
 
-func GetRudderObjectStorageAccessKeys() (accessKeyID, accessKey string) {
-	return config.GetStringVar("", "RUDDER_AWS_S3_COPY_USER_ACCESS_KEY_ID"), config.GetStringVar("", "RUDDER_AWS_S3_COPY_USER_ACCESS_KEY")
-}
-
 func GetRudderObjectStoragePrefix() (prefix string) {
 	return config.GetStringVar(config.GetNamespaceIdentifier(), "RUDDER_WAREHOUSE_BUCKET_PREFIX")
 }
@@ -657,12 +659,10 @@ func GetRegionHint() string {
 }
 
 func GetRudderObjectStorageConfig(prefixOverride string) (storageConfig map[string]any) {
-	// TODO: add error log if s3 keys are not available
 	storageConfig = make(map[string]any)
 	storageConfig["bucketName"] = config.GetStringVar("rudder-warehouse-storage", "RUDDER_WAREHOUSE_BUCKET")
-	storageConfig["accessKeyID"] = config.GetStringVar("", "RUDDER_AWS_S3_COPY_USER_ACCESS_KEY_ID")
-	storageConfig["accessKey"] = config.GetStringVar("", "RUDDER_AWS_S3_COPY_USER_ACCESS_KEY")
 	storageConfig["enableSSE"] = config.GetBoolVar(true, "RUDDER_WAREHOUSE_BUCKET_SSE")
+	// Credentials resolve via the AWS SDK default chain (pod identity / IRSA).
 	// set prefix from override for shared slave type nodes
 	if prefixOverride != "" {
 		storageConfig["prefix"] = prefixOverride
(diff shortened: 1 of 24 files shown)


shekhar-rudder:
## E2E Test Report

Verified against a local `rudder-mini` stack running an image built from this branch (`fix/remove-s3-copy-user` @ `ecd3220fe`; `/version` → `Commit: ecd3220fe`), plus real AWS S3 + Snowflake for the credential-chain paths. Throughout, `RUDDER_AWS_S3_COPY_USER_ACCESS_KEY_ID/KEY` were set to **poison values** to prove they're never read.

### Results

| # | Scenario | Path exercised | Env | Result |
|---|----------|----------------|-----|--------|
| 1 | S3 batch dest **with keys** | `GetObjectStorageConfig` S3+keys — credential gate | local | ✅ Gate passes: flow reaches real S3 `PutObject`, **not** the sentinel (distinguishes cleanly from #3) |
| 2 | Batch **non-S3** provider (MINIO) | provider ≠ S3, unchanged | local | ✅ 25/25 batch-router jobs `succeeded`; object written |
| 3 | S3 batch dest **no keys/role** ⚠️ | `ErrS3MissingCredentials` | local | ✅ 25/25 jobs fail with `"S3 access requires an IAM role ARN or access keys"`; **no object written; no fallback**; copy-user poison never consulted |
| 4 | Batch **useRudderStorage** staging upload | rudder config, no static keys → AWS SDK default chain | real AWS S3 | ✅ Staging file uploaded to `s3://…/rudder-warehouse-staging-logs/…` via default chain (pod-identity emulated by env creds); no static creds in config |
| 5 | Snowflake **useRudderStorage** COPY | `CreateAWSSessionConfig` no-static-creds + `GetTemporaryS3Cred` `AccessKeyID==""` branch (STS `GetSessionToken` **skipped**) | real AWS S3 + Snowflake | ✅ `copying data into main table`; warehouse upload → `exported_data`; **direct Snowflake query: 10 + 10 rows loaded**. COPY succeeded using temporary creds that STS `GetSessionToken` would have *rejected* → proves the STS-skip branch |

⚠️ #3 is the new hard-fail behavior this PR introduces (previously fell back to the shared copy-user).

### Pod-identity emulation (for #4/#5)

No IRSA cluster was available, so the pod-identity path was exercised by feeding AWS creds to rudder-server through the **SDK default credential provider chain** via env vars. The executed code path is identical to IRSA: `GetRudderObjectStorageConfig` injects no credentials → `awsutil.CreateAWSConfig` passes a nil provider → `LoadDefaultConfig` resolves via the default chain (env → shared config → **EKS Pod Identity / IRSA** in prod). #5's COPY succeeding with *temporary* session-token creds confirms `GetTemporaryS3Cred` retrieves them directly and skips `GetSessionToken`.

### Not run

`useRudderStorage=false` **physical** S3 upload (#1 full) and the Snowflake COPY with **user-supplied static keys / role** require long-term IAM keys (`AKIA…`) / an assumable role, which weren't available (only temporary SSO creds). Note a user S3 destination can only be tested with long-term keys anyway: config-backend serves the data plane only the destination-definition's declared config fields, so a `sessionToken` (needed by temporary creds) is stripped from the workspace config. These paths remain covered by the updated unit tests (`TestGetObjectStorageConfig`, `TestGetRudderObjectStorageConfig`, `TestCreateAWSSessionConfig`).

### Verdict

The security-critical behaviors verified end-to-end: (a) missing S3 credentials now **hard-fail** with no fallthrough to any RudderStack-owned credential (#3), and (b) the `useRudderStorage` upload and warehouse COPY both authenticate via the **AWS SDK default chain / pod identity** with no static credentials (#4, #5). ✅

<sub>🤖 E2E run via Claude Code</sub>

warehouse/integrations/snowflake/snowflake_test.go
achettyiitr, line 1: Didn't get why these tests are getting removed? Is the plan to add a different one?

shekhar-rudder: Added back the tests

utils/misc/misc.go
```diff
@@ -689,23 +689,22 @@ type ObjectStorageOptsT struct {
...
+		return GetRudderObjectStorageConfig(opts.RudderStoragePrefixOverride), nil
 	}
 	if opts.Provider == "S3" {
-		clonedObjectStorageConfig := make(map[string]any)
-		maps.Copy(clonedObjectStorageConfig, objectStorageConfigMap)
-		clonedObjectStorageConfig["externalID"] = opts.WorkspaceID
 		if !HasAWSRoleARNInConfig(objectStorageConfigMap) &&
 			!HasAWSKeysInConfig(objectStorageConfigMap) {
-			clonedObjectStorageConfig["accessKeyID"] = config.GetStringVar("", "RUDDER_AWS_S3_COPY_USER_ACCESS_KEY_ID")
-			clonedObjectStorageConfig["accessKey"] = config.GetStringVar("", "RUDDER_AWS_S3_COPY_USER_ACCESS_KEY")
```
achettyiitr, line 704: Do we need to fallback to pod credentials instead of returning error. For redshift, AWS Access Keys are optional?

shekhar-rudder: - We are throwing an error when neither IAM role nor access keys have been configured.
- In this case, useRudderStorage is false, so I don't think we should fallback to pod credential in any scenario

achettyiitr: For redshift, access keys are optional. https://github.com/rudderlabs/rudder-integrations-config/blob/develop/src/configurations/destinations/rs/ui-config.json#L386

shekhar-rudder: But [iamRoleARN](https://github.com/rudderlabs/rudder-integrations-config/blob/77968a80f4ce316d59ee37e252cfe200f420eab0/src/configurations/destinations/rs/ui-config.json#L373) is mandatory. So as long as that is present, server won't throw any error

achettyiitr: Ian role is not mandatory. It’s still option and customer can toggle back.

achettyiitr: It’s required only for the Pre-requisites.

shekhar-rudder: Addressed

warehouse/utils/utils.go
```diff
@@ -674,10 +685,13 @@ func GetTemporaryS3Cred(destination *backendconfig.DestinationT) (string, string
 		return "", "", "", err
 	}
 
-	if sessionConfig.RoleBasedAuth {
+	// Role-based auth and the default credential chain (pod identity) already
+	// yield temporary credentials that include a session token — retrieve them
+	// directly. Only long-term static keys need an explicit GetSessionToken.
+	if sessionConfig.RoleBasedAuth || sessionConfig.AccessKeyID == "" {
```
achettyiitr: Is it still valid? `sessionConfig.AccessKeyID == ""`, it can be in `Non-rudder-storage destination without an IAM role ARN or access keys`.

warehouse/integrations/deltalake/deltalake_test.go
```diff
@@ -234,9 +270,10 @@ func TestIntegration(t *testing.T) {
 				},
 				additionalEnvs: func(destinationID string) map[string]string {
 					return map[string]string{
-						"RUDDER_AWS_S3_COPY_USER_ACCESS_KEY_ID": credentials.AccessKeyID,
-						"RUDDER_AWS_S3_COPY_USER_ACCESS_KEY":    credentials.AccessKey,
-						"RUDDER_WAREHOUSE_BUCKET":               credentials.BucketName,
+						"AWS_ACCESS_KEY_ID":       rudderStorageAccessKeyID,
+						"AWS_SECRET_ACCESS_KEY":   rudderStorageSecretAccessKey,
+						"AWS_SESSION_TOKEN":       rudderStorageSessionToken,
+						"RUDDER_WAREHOUSE_BUCKET": credentials.BucketName,
```
achettyiitr, line 276: Why do we need a token here? Can't we have similar to one in redshift and snowflake?

shekhar-rudder: Redshift/Snowflake tests passed without token but deltalake tests failed. Have added comment explaining why is it needed [here](https://github.com/rudderlabs/rudder-server/pull/7167/changes#diff-8ad3c6da0af87741a94ff8d4986b4fb69b117a490a3c8c705e15c97681e43d99R82)

warehouse/utils/utils.go
achettyiitr: Can we be explicit only for use rudder storage?

warehouse/integrations/deltalake/deltalake_test.go
shekhar-rudder: Deltalake is different from Redshift/Snowflake here because of how the COPY runs.

Redshift and Snowflake execute the `COPY` server-side and accept long-term IAM keys directly, so plain `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` work.

Deltalake performs the S3 read on Databricks via the S3A `TemporaryAWSCredentialsProvider`, which **requires a non-empty session token**. Long-term static keys with no token are rejected.

In production the `useRudderStorage` path resolves session-token-bearing credentials from the AWS SDK default chain (IRSA / pod identity), and `GetTemporaryS3Cred` returns them directly via `Credentials.Retrieve()`. But `Retrieve()` only *resolves* whatever the chain holds — it doesn't mint a token. In the test env the chain resolves the plain static `AWS_*` keys (with an empty `AWS_SESSION_TOKEN`), so the session token comes back empty.

I confirmed this: dropping the token to match Redshift/Snowflake made the deltalake integration test fail at COPY with:

```
[UNAUTHORIZED_ACCESS] Unauthorized access: s3://.../rudder-warehouse-load-objects/...
shaded.databricks.org.apache.hadoop.fs.s3a.CredentialInitializationException:
Provider TemporaryAWSCredentialsProvider has no credentials:
NoAwsCredentialsException: Session credentials in Hadoop configuration: No AWS Credentials
SQLSTATE: 42501
```

So we mint a short-lived session token from the static test keys via STS `GetSessionToken` and export it as `AWS_SESSION_TOKEN`, mirroring the production IRSA/pod-identity path that also yields temporary, session-token credentials — instead of reintroducing an app-specific credential env.

achettyiitr (approved):

ItsSudip (approved):

PR #7173, opened 2026-07-10
fix: handle Snowpipe auth errors during poll phase

shekhar-rudder:
# Description

Snowpipe streaming could enter a false **"stuck pipeline" P0** (`warehouse-snowpipe-stuck`) when a Snowflake auth credential changed underneath a running destination (e.g. a rotated RSA key).

Auth/abort errors were only classified in the **upload** path (`createChannel`). The **poll** recovery path (`handleChannelRecoveryPostBulkStatus`) called `m.api.CreateChannel` directly, bypassing that classification. Because `m.api.CreateChannel` returns `(resp, nil)` even when `resp.Success == false`, an auth failure during poll was:
- stored in `channelCache` as an **invalid** channel (empty channel id), and
- surfaced only as a generic error, leaving the import in-progress until the 15-minute stuck threshold — where it failed with the misleading `"events still in progress after threshold"` message and fired the P0 alert.

## Changes

- **Extract shared classification** — pull the terminal channel-response classification out of `createChannel` into `classifyFailedChannelResponse` (auth / validation / authorization and unsupported-column → `errAbort`; generic otherwise). `createChannel` delegates to it, so the upload path behaves exactly as before.
- **Classify auth errors in the poll recovery path** — `handleChannelRecoveryPostBulkStatus` now classifies a non-success recreate response *before* caching or re-polling. Auth failures **fail fast** with a meaningful reason, the invalid channel is **not cached**, and no stuck-pipeline alert is raised.
- **Propagate the real error** in the upload discards-channel abort — send the underlying `err` instead of the generic `errAbort` sentinel, so the job's abort reason carries the actual cause.

## Tests

- `provideInProgressImportInfos` — recovery auth error fast-fails as abort: asserts the import is not left in-progress, is marked failed with the auth reason (not the stuck message), and the invalid channel is not cached.
- `Upload` — discards-channel auth abort surfaces the meaningful reason (guards the line-201 change).

Full package suite (`SLOW=0 go test ./.../snowpipestreaming/...`) passes; `go vet` and `gofmt` clean.

## Linear Ticket

https://linear.app/rudderstack/issue/INT-6447/handle-auth-errors-during-snowpipe-poll-phase

## Security

### DIFF
--- a/router/batchrouter/asyncdestinationmanager/snowpipestreaming/channel.go
+++ b/router/batchrouter/asyncdestinationmanager/snowpipestreaming/channel.go
@@ -120,13 +120,25 @@ func (m *Manager) createChannel(
 		}
 		m.channelCache.Store(tableName, resp)
 		return resp, nil
+	default:
+		return nil, classifyFailedChannelResponse(resp)
+	}
+}
+
+// classifyFailedChannelResponse maps a non-success CreateChannel response to an error.
+// Validation/authorization failures and unsupported-column failures are wrapped with errAbort
+// (terminal, non-retryable); all other codes yield a generic (retryable) error.
+// It is shared by the upload path (createChannel) and the poll recovery path
+// (handleChannelRecoveryPostBulkStatus) so both classify auth errors identically.
+func classifyFailedChannelResponse(resp *model.ChannelResponse) error {
+	switch resp.Code {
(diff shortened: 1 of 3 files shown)


shekhar-rudder:
Closing — superseded by #7118 (which reworks the same channel abort classification) and the poll-phase auth handling was deprioritized (see INT-6447, now canceled).

PR #7177, opened 2026-07-13
feat: run python transformation tests on ephemeral per-request pyt deployments 

mihir20:
## Description

Python transformation tests (execution ops) run arbitrary, untrusted customer code. Previously, the processor's `cpservice` scaled up the workspace's shared production pyt deployment (`pytscaler`) and
forwarded the test to it — meaning test code ran on the same pods that serve production traffic.

This PR replaces that model with **ephemeral, per-request pyt deployments**: every execution test request gets its own fresh Deployment + Service, created just for that request and torn down after it —
one test per pod.

### New `processor/internal/pytdeployer` package (replaces `pytscaler`)
- `Deployer.RunOnEphemeral(ctx, workspaceID, forward)` owns the full lifecycle: create Deployment + Service → wait for readiness → forward the request → delete both (best-effort).
- Deployment spec mirrors the production per-workspace pytransformer chart: Kata runtime class (`kata-fc`), kata nodepool selector + tolerations, resource requests, image pull secret — all configurable
  via `Processor.pytDeployer.*` config keys (env, labels, pod annotations, and readiness timeout are hot-reloadable).
- **Zone pinning**: pods are pinned to the processor's `AVAILABILITY_ZONE` (node selector + pod label) and the Service selector is scoped to that zone, so cross-zone traffic is avoided.
- All created objects are stamped with contract labels (`app.kubernetes.io/managed-by=rudderstack-processor`, `rudderstack.com/purpose=pyt-test`, `rudderstack.com/workspace-id`) that the reaper CronJob key off.
- Deliberately **no no-op fallback** (unlike `pytscaler`): if a real k8s-backed deployer can't be built, execution ops fail with `FailedPrecondition` — there is no safe fallback target for untrusted code.
  AST ops are unaffected.
- k8s API calls retry with exponential backoff on transient errors (retry helper moved over from `pytscaler`).
- Metrics port exposed on the pod for scraping.

### `cpservice` routing changes
- **Execution ops** (`test`, `testRun`) → fresh ephemeral deployment via `deployer.RunOnEphemeral`.
- **AST ops** (`testLibrary`, `extractLibs`) never execute user code → forwarded straight to a shared static AST deployment (`Processor.pytTestStaticASTURL`), no k8s API involvement.
- **Per-workspace escape hatch**: workspaces flagged via `Processor.pytTestOverrides.<workspaceID>.routeToProduction` (hot-reloadable, with a global fallback key) forward to their production pyt
  deployment instead — for workspaces whose tests need prod-only config (custom DNS, pinned egress IPs, ...) that the generic ephemeral spec can't reproduce. No readiness wait there; the forwarder's
  cold-start retries ride out the scale-from-zero window.
- Removed the singleflight scale-group and scale-replicas config — no longer needed without shared-deployment scaling.
- `Forward` now logs the end-to-end forward duration per op for performance monitoring.

### `user_transformer` forwarder refactor
- Test-forwarding methods (`Test`, `TestRun`, `TestLibrary`, `ExtractLibs`) now take an explicit `baseURL` resolved by the caller instead of deriving the URL from the workspace ID, so the same client
  serves ephemeral, prod-routed, and static-AST targets.
- `ErrPerWorkspacePyTNotEnabled` replaced by `ErrEmptyForwardBaseURL` (mapped to `FailedPrecondition`, since misconfiguration must not be retried away).
- `PerWorkspacePyTBaseURL` extracted as the single place the `{workspaceID}` template substitution lives.

## Linear Ticket

PIPE-3172

## Security


processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -615,6 +641,17 @@ func isColdStartError(err error, resp *http.Response) bool {
...
 		}
+		// Dial timeout ("dial tcp ...: i/o timeout"): the SYN went unanswered —
+		// the pod behind the headless-service DNS record is either still coming
+		// up (IP published before the VM boots) or on its way down (terminating
+		// pod whose record hasn't been withdrawn yet). The two are
+		// indistinguishable from the dialer, and retrying is right for both:
+		// each retry re-resolves DNS, and the pod's replacement (the forward path
+		// scales the deployment up before dialing) answers a later attempt.
+		var opErr *net.OpError
+		if errors.As(err, &opErr) && opErr.Op == "dial" && opErr.Timeout() {
```
fracasula: Syntax update in Go 1.26: replace 'errors.As' with the type-safe 'errors.AsType'

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -559,23 +578,30 @@ func (u *Client) forwardTest(ctx context.Context, workspaceID, path string, payl
...
+				// and the caller's deadline either way. Cold-start signals are
+				// additionally counted; retryability doesn't depend on them.
+				if isColdStartError(err, nil) {
+					u.stat.NewTaggedStat(coldStartErrorsMetric, stats.CountType,
+						stats.Tags{"workspaceID": workspaceID, "language": languagePython}).Increment()
+				}
+				return err
+			}
+			if isColdStartError(nil, resp) {
 				httputil.CloseResponse(resp)
```
fracasula: Why not just move the defer before `if isColdStartError(nil, resp)`?

processor/internal/transformer/user_transformer/user_transformer.go
```diff
@@ -559,23 +578,30 @@ func (u *Client) forwardTest(ctx context.Context, workspaceID, path string, payl
...
-				return backoff.Permanent(err)
-			}
 			defer func() { httputil.CloseResponse(resp) }()
 
 			body, err = io.ReadAll(resp.Body)
 			if err != nil {
-				return backoff.Permanent(err)
+				// Connection broke mid-response — as transient as a failed
+				// request; retry.
+				return err
```
fracasula, line 607: I'm a bit confused here. This used to be a permanent error (i.e. not retryable). Now we're saying it's "as transient as a failed request" and retry indefinitely but we don't have visibility on this, right? No metrics, no logs. We can only wait until max tries is reached and it then propagates.

Are we sure this isn't a regression? Is it really as the comment says "Connection broke mid-response"? Can it be something else as well?

processor/internal/pytdeployer/pytdeployer.go
```diff
@@ -0,0 +1,185 @@
...
+	env              config.ValueLoader[map[string]any]
+	labels           config.ValueLoader[map[string]any]
+	podAnnotations   config.ValueLoader[map[string]any]
+	retry            retrySettings
+}
+
+func loadConfig(conf *config.Config) deployerConfig {
+	return deployerConfig{
+		namespace:       conf.GetStringVar("test-pytransformer", "Processor.pytDeployer.pytTestNamespace"),
+		image:           conf.GetStringVar("", "Processor.pytDeployer.pytTestImage"),
```
fracasula, line 164: What happens if this one stays empty? Is there any validation done?

processor/internal/pytdeployer/pytdeployer.go
```diff
@@ -0,0 +1,185 @@
...
+		imagePullSecret: conf.GetStringVar("regcred", "Processor.pytDeployer.pytTestImagePullSecret"),
```
fracasula: Shouldn't we remove this default as it could be considered a leak and we panic if empty (or error)?

processor/internal/pytdeployer/k8s.go
```diff
@@ -0,0 +1,339 @@
...
+				InitialDelaySeconds: 10, PeriodSeconds: 30, TimeoutSeconds: 10, FailureThreshold: 3,
+			},
+			ReadinessProbe: &corev1.Probe{
+				ProbeHandler: corev1.ProbeHandler{
+					HTTPGet: &corev1.HTTPGetAction{Path: "/health/ready", Port: intstr.FromInt32(pytContainerPort)},
+				},
+				InitialDelaySeconds: 5, PeriodSeconds: 10, TimeoutSeconds: 5, FailureThreshold: 3,
+			},
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
```
fracasula, line 172: Given that it's untrusted code plus it's also testing code (not production grade) I think it would make sense to add a limit as well.

processor/internal/transformer/user_transformer/user_transformer.go
mihir20: Yeah, there can be multiple network errors, so just like processor transform flow, I am retrying for every error. will add some observability here.

processor/internal/pytdeployer/pytdeployer.go
mihir20: there is no validation will add

processor/cpservice/forward.go
```diff
@@ -66,36 +98,37 @@ func (s *Service) Forward(ctx context.Context, req *proto.ForwardRequest) (*prot
...
+// to run on its production pyt deployment instead of an ephemeral one. The
+// flag is a per-workspace reloadable bool —
+// Processor.pytTestOverrides.<workspaceID>.routeToProduction, falling back to
+// the global Processor.pytTestOverrides.routeToProduction — so a workspace
+// (or all of them) can be flagged without a restart. Config keys are matched
+// case-insensitively, mirroring the lowercased deployment-name convention.
+// workspaceID has already passed validWorkspaceID in Forward, so it is safe
+// to interpolate into the key.
+func (s *Service) isProdRouted(workspaceID string) bool {
+	return s.conf.GetReloadableBoolVar(false,
```
fracasula: Given that this is called every time and that nothing would clean up the registered variables (and we could register any number given that we're interpolating the workspaceID), why not register it only once by using `GetReloadableStringMapVar`? I know it's an edge case, and that it's very unlikely that we would register that many, but it seems like an easy change to avoid a potential leak.

processor/internal/pytdeployer/retry.go
```diff
@@ -22,11 +22,11 @@ type retrySettings struct {
 
 func newRetrySettings(conf *config.Config) retrySettings {
 	return retrySettings{
-		initialInterval:     conf.GetDurationVar(200, time.Millisecond, "Processor.UserTransformer.pytScaler.retry.initialInterval"),
-		maxInterval:         conf.GetDurationVar(3, time.Second, "Processor.UserTransformer.pytScaler.retry.maxInterval"),
-		maxElapsedTime:      conf.GetDurationVar(30, time.Second, "Processor.UserTransformer.pytScaler.retry.maxElapsedTime"),
-		multiplier:          conf.GetFloat64Var(1.5, "Processor.UserTransformer.pytScaler.retry.multiplier"),
-		randomizationFactor: conf.GetFloat64Var(0.5, "Processor.UserTransformer.pytScaler.retry.randomizationFactor"),
+		initialInterval:     conf.GetDurationVar(200, time.Millisecond, "Processor.pytDeployer.retry.initialInterval"),
```
fracasula, line 28: I was looking at this file again due to these changes, and I was thinking if perhaps we should extend `isTransient` a bit further and add `k8s.io/apimachinery/pkg/util/net`'s `IsConnectionReset`, `IsConnectionRefused`, `IsProbableEOF` (and optionally a `net.Error.Timeout()` check). It shouldn't cause issues and it could improve the detection a bit more. Wdyt?

processor/internal/pytdeployer/k8s.go
```diff
@@ -0,0 +1,346 @@
...
+	// fail. Skipped entirely when no zone is configured (e.g. single-zone
+	// clusters, local dev): an empty-valued nodeSelector would only match
+	// nodes labeled with an empty zone, i.e. nothing.
+	podLabels := maps.Clone(labels)
+	svcSelector := maps.Clone(labels)
+	if d.config.zone != "" {
+		podSpec.NodeSelector["topology.kubernetes.io/zone"] = d.config.zone
+		podLabels["zone"] = d.config.zone
+		svcSelector["zone"] = d.config.zone
+	}
```
fracasula: Shouldn't we add the zone label first and the clone to avoid potential collisions?

```go
	if d.config.zone != "" {
		podSpec.NodeSelector["topology.kubernetes.io/zone"] = d.config.zone
		labels["zone"] = d.config.zone
	}
	podLabels := maps.Clone(labels)
	svcSelector := maps.Clone(labels)
```

If an operator's `pytTestLabels` includes a key literally named `"zone"` while `AVAILABILITY_ZONE` is also set, then we might have a problem.

A test on this would be nice too.

processor/cpservice/forward.go
```diff
@@ -28,36 +32,64 @@ type endpointForward func(ctx context.Context, workspaceID string, payload []byt
 var validWorkspaceID = regexp.MustCompile(`^[a-zA-Z0-9-]{1,59}$`)
```
fracasula: This regex allows leading/trailing hyphens as well (like `-abc` or `abc-`) while k8s label values must begin and end as alphanumeric or we could get a non-transient `Create` failure. Given that we're already validating it, we might as well make sure this is more precise.

processor/internal/pytdeployer/k8s.go
```diff
@@ -0,0 +1,346 @@
...
+		NodeSelector:       maps.Clone(d.config.nodeSelector),
+		// Tolerations come from config (default: the prod chart's kata
+		// toleration). They only permit scheduling, never force it, so they are
+		// applied regardless of whether the Kata RuntimeClass is enabled.
+		Tolerations:      slices.Clone(d.config.tolerations),
+		ImagePullSecrets: []corev1.LocalObjectReference{{Name: d.config.imagePullSecret}},
+		Containers: []corev1.Container{{
+			Name:  "pytransformer",
+			Image: d.config.image,
+			SecurityContext: &corev1.SecurityContext{
```
fracasula, line 143: I got a comment from a security agent here but haven't verified. Can you please look into it?

```markdown
#### 5. Untrusted-code container lacks `Capabilities.Drop: ["ALL"]` and a `SeccompProfile` (security)

**Why:** The `SecurityContext` sets `RunAsNonRoot`, `RunAsUser/Group`, `ReadOnlyRootFilesystem`, `AllowPrivilegeEscalation: false` — but still omits a capability drop and seccomp profile, so it doesn't meet the PSS "restricted" baseline standard for untrusted workloads (this update added resource Limits but not these). Kata (`kata-fc`) is the primary boundary, so it's defense-in-depth, but a cheap, direct fit for the stated goal.

**Fix:** Add `Capabilities: {Drop: ["ALL"]}` and `SeccompProfile: {Type: RuntimeDefault}`.
```

processor/internal/pytdeployer/k8s.go
```diff
@@ -0,0 +1,346 @@
...
+			// Limits are mandatory here, not an optimization: the container
+			// runs untrusted, non-production-grade code, so it must not be
+			// able to consume unbounded node CPU/memory.
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceCPU:    d.config.cpuRequest,
+					corev1.ResourceMemory: d.config.memoryRequest,
+				},
+				Limits: corev1.ResourceList{
+					corev1.ResourceCPU:    d.config.cpuLimit,
```
fracasula, line 177: We could add an easy check and make sure that limits >= requests

processor/internal/pytdeployer/k8s.go
```diff
@@ -0,0 +1,346 @@
...
+
+// waitReady polls the Deployment until it reports at least one ready
+// replica, bounded by min(ctx's deadline, the configured readiness timeout).
+func (d *k8sDeployer) waitReady(ctx context.Context, name string) error {
+	ctx, cancel := context.WithTimeout(ctx, d.config.readinessTimeout.Load())
+	defer cancel()
+
+	ticker := time.NewTicker(readinessPollInterval)
+	defer ticker.Stop()
+	var lastErr error
```
fracasula, line 272: I see we never reset `lastErr`. Can this be a problem after a successful-but-not-ready `Get`? :thinking:

fracasula (approved):
LGTM, I left a bunch of comments for you to review.

In addition to them I was thinking that we have no "rate limit" whatsoever, so a burst of `test` requests can provision **unbounded Kata microVM pods**. It's good that we'll have a reaper handling cleanups, but what about admissions? cc @ktgowtham 

We could have a simple semaphore in the deployer and optionally a `ResourceQuota` so that a bug/misconfiguration can't exhaust the Kata nodepool.

mihir20:
> LGTM, I left a bunch of comments for you to review.
> 
> In addition to them I was thinking that we have no "rate limit" whatsoever, so a burst of `test` requests can provision **unbounded Kata microVM pods**. It's good that we'll have a reaper handling cleanups, but what about admissions? cc @ktgowtham
> 
> We could have a simple semaphore in the deployer and optionally a `ResourceQuota` so that a bug/misconfiguration can't exhaust the Kata nodepool.

This is a good point; I will pick this as a follow-up task, as this is not a blocker for the release.

processor/internal/pytdeployer/k8s.go
mihir20: If I understood correctly these makes sense:

Capabilities Drop: ["ALL"] — Even non-root containers get a default set of Linux privileges (raw network access, changing file ownership, switching users, etc.). Dropping all of them means the customer's
  Python code has zero privileged operations available — and pytransformer needs none of them anyway (unprivileged port, fixed user, own temp dir).

  Seccomp RuntimeDefault — Filters which kernel syscalls the container can make, blocking the ~40-50 obscure ones that most kernel exploits enter through. Kubernetes leaves this off by default, so setting
  it explicitly guarantees the filter regardless of cluster config.
  
  Why together — Kata's microVM is the primary sandbox; these two are defense-in-depth inside it, and they complete the PSS "restricted" profile (the strictest k8s standard) at zero cost to the app.

processor/internal/pytdeployer/k8s.go
mihir20: fixed now

ktgowtham (approved):

ktgowtham:
> LGTM, I left a bunch of comments for you to review.
> 
> In addition to them I was thinking that we have no "rate limit" whatsoever, so a burst of `test` requests can provision **unbounded Kata microVM pods**. It's good that we'll have a reaper handling cleanups, but what about admissions? cc @ktgowtham
> 
> We could have a simple semaphore in the deployer and optionally a `ResourceQuota` so that a bug/misconfiguration can't exhaust the Kata nodepool.

We should introduce rate limit on control plane side. cc: @mihir20

PR #7184, opened 2026-07-17
feat(router): detect and classify transformer-proxy contract breaches (INT-6839)

saikumarrs:
## What

Makes the transformer-proxy delivery contract observable for [INT-6839](https://linear.app/rudderstack/issue/INT-6839).

A contract breach is when rudder-server sends a batch of jobs to a destination's proxy endpoint and gets back a response it cannot reconcile with what it sent — per-job results end up attached to the **wrong jobs**, or the batch can't be read at all, and no ordinary delivery failure records it. Today `router_transformerproxy_invalid_response` (Go stat `router.transformerproxy.invalid.response`) is emitted for only one of those ways, carries no attribution, and has no alert.

This PR emits the metric at **every** breach site with attribution, classifies infrastructure failures separately so they don't page as breaches, and logs the affected **jobIDs** so on-call can pull the real events out of jobsDB.

## Reasons

| `reason` | When | Contract breach? |
| --- | --- | --- |
| `in out mismatch` | Returned jobIDs differ from those sent, **or** the transformer answered twice for one jobID (one status silently dropped by the map write) | Yes — results attached to the wrong jobs |
| `missing output` | Response carried no `output` field, or an explicit `"output":null` | Yes |
| `unmarshal error` | The response envelope, or its `output` payload, would not parse | Yes |
| `transport error` | The response was **synthesized by the oauth transport** — failed token fetch, augmentation failure, LB round-trip error, unreadable body | No — infrastructure |

All carry **`reason`, `destType`, `destinationId`**, and log the upstream `statusCode`.

### How infra is told apart from a breach

The signal is **provenance, not status**: did the response actually come from the transformer? Only the transformer stamps the `apiVersion` header (`ControllerUtility.deliveryPostProcess`), so a response lacking it did not come from the transformer — an oauth-transport-synthesized error, an ingress/reverse-proxy error page, or an unhandled transformer exception routed through the global error handler rather than `deliveryPostProcess`. `doProxyRequest` records `fromTransformer` from the header's presence, and `infraFailure := !fromTransformer` selects the `transport error` reason. No separate marker header is needed.

The upstream HTTP status is deliberately **not** part of that signal. For v0 the transformer mirrors the destination's delivery status (`deliveryPostProcess`), so any destination 5xx comes from a perfectly healthy transformer — using it would relabel a genuine breach that co-occurs with a destination 5xx as infra, which is exactly the case this exists to catch. The status is logged instead.

Infra only ever selects the *reason* at the points where the response can't be processed; a response that parses is applied exactly as before, so a 5xx still carrying a usable output is delivered normally. A raw transport failure (unreachable transformer, timeout, 404) surfaces as a request error and is emitted as `transport error` too, so the metric is symmetric across oauth and non-oauth destinations.

## The breach logs — jobIDs, not bodies

Each site logs via `Warnn` with `destinationId`, `jobIDs`, and `statusCode`. The three infra branches carry distinct messages so a log line identifies its site.

**The response body is deliberately not logged.** These get debugged by port-forwarding jobsDB and reading the events by jobID, so the jobIDs are what the logs carry. The runbook's lookup uses rudder-server's own `unionjobsdb('rt', N)` to span the rotating `rt_jobs_<n>` datasets and join each job's latest status. Bodies embedded in `ProxyRequestResponseBody` are truncated at 10KB — nothing consumes that field (`router/worker.go` reads only the per-job maps) and a batch response can be multi-MB.

## Behaviour change: `{"output":null}` is no longer a delivery

A response carrying an explicit null output is now treated as `missing output` and returned as an error, so the batch fails and retries.

On **v1** this matches what already happened (the per-job maps came back empty either way, and `hydrateRespStatusCodes` marked every job 500). On **v0** it is a real change: `v0Adapter.getResponse` assigns `respCode` to every job off the request metadata, so a 200 carrying a null output previously marked the whole batch **delivered**. Any destination that legitimately answers `{"output":null}` will now see those jobs fail and retry instead.

That is the intended semantics — a null output is not a delivery — but it is a change in outcome, not just in observability. Covered by tests on both versions.

## Why these labels — and why not `workspaceId`

`reason` drives the runbook branch and the page text; `destType` names the broken integration; `destinationId` pinpoints which destination is breaching (a workspace can run several destinations of one destType).

`workspaceId` was added in an earlier commit and then **deliberately dropped**: nothing consumes it. Alertmanager routes on `namespace`, the runbook never branches on it, and the affected workspace is recoverable from jobsDB — `unionjobsdb` returns `workspace_id`.

## Notable design points from review

- **Detection lives at the caller.** `TransResponse` no longer carries the mismatch flag or the jobID slices; `ProxyRequest` derives the response set from `maps.Keys(routerJobResponseCodes)` and compares against the deduped request jobIDs. All reasons are detected and emitted in one place, through one stats handle.
- **Request-side duplicate jobIDs are legitimate** (`router/worker.go` dedupes only when building final responses), so the set comparison dedupes both sides. **Response-side** duplicates are not — the adapter reports `responseEntries` so `entries > keys` is flagged, since the map write silently keeps the last status.
- **`{"output":null}` counts as missing.** gjson reports an explicit null as existing (`Raw == "null"`), so the type is checked too.

## Tests

Breach classification is asserted through a memstats store: each case pins the **exact set** of emitted reasons, so a healthy response asserting *none* catches a spurious page, not just a mislabel. Covered: all four reasons, `"output":null`, a body with no `output` field, a duplicate-jobID response, a duplicate-jobID **request** (must not page), and a 5xx carrying a usable output (must be applied, not dropped). Each was verified to fail when its guard is removed.

## Downstream / gating

Unblocks the operator alerts in **rudderlabs/rudderstack-operator#1452**, which split by blast radius: `in out mismatch` pages per destination, the response-shape reasons page coarsely per destType, and `transport error` pages from neither.

> **Sequencing:** this PR must be merged **and deployed**, with the new reasons and labels verified emitting in prod, before that alert PR merges.

### DIFF
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -559,13 +560,21 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 	if err != nil {
 		return proxyErrorResponse(respCode, err.Error(), routerJobDontBatchDirectives)
 	}
+	// Compared as sets, so the check does not depend on duplicate jobIDs on either side: a batch can
+	// carry the same JobID more than once (router/worker.go only dedupes when building the final job
+	// responses) and the transformer may collapse or echo those. routerJobResponseCodes is keyed by
+	// the jobIDs the transformer answered for, so its keys are the response set.
+	jobIDsInMetadata := lo.Uniq(proxyJobIDs(proxyReqParams.ResponseData.Metadata))
+	slices.Sort(jobIDsInMetadata)
+	jobIDsInResponse := lo.Keys(transResp.routerJobResponseCodes)
+	slices.Sort(jobIDsInResponse)
 	// Non-fatal: the per-job results were still applied, but they may be attached to the wrong jobs.
 	// Recorded here, alongside the other breach reasons, so all three share one stats handle.
-	if transResp.jobIDMismatch {
+	if !slices.Equal(jobIDsInMetadata, jobIDsInResponse) {
 		emitProxyInvalidResponse(trans.stats, trans.logger, proxyReqParams, "in out mismatch",
 			"[TransformerProxy] JobIDs in out mismatch",
-			logger.NewIntSliceField("jobIDsInMetadata", transResp.jobIDsInMetadata),
-			logger.NewIntSliceField("jobIDsInResponse", transResp.jobIDsInResponse))
+			logger.NewIntSliceField("jobIDsInMetadata", jobIDsInMetadata),
+			logger.NewIntSliceField("jobIDsInResponse", jobIDsInResponse))
 	}
 	integrations.CollectIntegrationFailureDetailedStats(trans.stats, transResp.statTags)
 
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -421,23 +421,28 @@ func emitProxyInvalidResponse(statsHandle stats.Stats, log logger.Logger, reason
 	log.Warnn(msg, append([]logger.Field{obskit.DestinationID(destinationID)}, fields...)...)
 }
 
+// proxyErrorResponse builds the error-path response shared by every early return in ProxyRequest.
+// RespStatusCodes/RespBodys are always empty here by construction: per-job results only exist once
+// the adapter has parsed a response, which has not happened on any path that bails out early.
+func proxyErrorResponse(code int, body string, dontBatch map[int64]bool) ProxyRequestResponse {
+	return ProxyRequestResponse{
+		ProxyRequestStatusCode:   code,
+		ProxyRequestResponseBody: body,
+		RespContentType:          "text/plain; charset=utf-8",
+		RespStatusCodes:          map[int64]int{},
+		RespBodys:                map[int64]string{},
+		DontBatchDirectives:      dontBatch,
+	}
+}
+
 func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequestParams) ProxyRequestResponse {
 	start := time.Now()
-	routerJobResponseCodes := make(map[int64]int)
-	routerJobResponseBodys := make(map[int64]string)
 	routerJobDontBatchDirectives := make(map[int64]bool)
 
 	if len(proxyReqParams.ResponseData.Metadata) == 0 {
 		trans.logger.Warnn("[TransformerProxy] Input metadata is empty",
 			logger.NewStringField("destination", proxyReqParams.DestName))
-		return ProxyRequestResponse{
-			ProxyRequestStatusCode:   http.StatusBadRequest,
-			ProxyRequestResponseBody: "Input metadata is empty",
-			RespContentType:          "text/plain; charset=utf-8",
-			RespStatusCodes:          routerJobResponseCodes,
-			RespBodys:                routerJobResponseBodys,
-			DontBatchDirectives:      routerJobDontBatchDirectives,
-		}
+		return proxyErrorResponse(http.StatusBadRequest, "Input metadata is empty", routerJobDontBatchDirectives)
 	}
 
 	for _, m := range proxyReqParams.ResponseData.Metadata {
@@ -449,26 +454,12 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 
 	payload, err := proxyReqParams.Adapter.getPayload(proxyReqParams)
 	if err != nil {
-		return ProxyRequestResponse{
-			ProxyRequestStatusCode:   http.StatusInternalServerError,
-			ProxyRequestResponseBody: "Payload preparation failed",
-			RespContentType:          "text/plain; charset=utf-8",
-			RespStatusCodes:          routerJobResponseCodes,
-			RespBodys:                routerJobResponseBodys,
-			DontBatchDirectives:      routerJobDontBatchDirectives,
-		}
+		return proxyErrorResponse(http.StatusInternalServerError, "Payload preparation failed", routerJobDontBatchDirectives)
 	}
 
 	proxyURL, err := proxyReqParams.Adapter.getProxyURL(proxyReqParams.DestName)
 	if err != nil {
-		return ProxyRequestResponse{
-			ProxyRequestStatusCode:   http.StatusInternalServerError,
-			ProxyRequestResponseBody: "ProxyURL preparation failed",
-			RespContentType:          "text/plain; charset=utf-8",
-			RespStatusCodes:          routerJobResponseCodes,
-			RespBodys:                routerJobResponseBodys,
-			DontBatchDirectives:      routerJobDontBatchDirectives,
-		}
+		return proxyErrorResponse(http.StatusInternalServerError, "ProxyURL preparation failed", routerJobDontBatchDirectives)
 	}
 
 	// Create metric labels
@@ -498,14 +489,7 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 	trans.stats.NewTaggedStat("transformer_proxy_request_result", stats.CountType, labelsWithSuccess).Increment()
 
 	if requestError != nil {
-		return ProxyRequestResponse{
-			ProxyRequestStatusCode:   respCode,
-			ProxyRequestResponseBody: requestError.Error(),
-			RespContentType:          "text/plain; charset=utf-8",
-			RespStatusCodes:          routerJobResponseCodes,
-			RespBodys:                routerJobResponseBodys,
-			DontBatchDirectives:      routerJobDontBatchDirectives,
-		}
+		return proxyErrorResponse(respCode, requestError.Error(), routerJobDontBatchDirectives)
 	}
 
 	/*
@@ -522,14 +506,7 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 		// The payload is not JSON, so nothing downstream can read it: `output` cannot exist, and
 		// continuing would blank respData and fail inside getResponse on an empty body — losing the
 		// very payload that explains the breach. Stop here and return it.
-		return ProxyRequestResponse{
-			ProxyRequestStatusCode:   respCode,
-			ProxyRequestResponseBody: fmt.Sprintf("[TransformerProxy] response is not valid JSON: %s, err: %v", string(respData), err),
-			RespContentType:          "text/plain; charset=utf-8",
-			RespStatusCodes:          routerJobResponseCodes,
-			RespBodys:                routerJobResponseBodys,
-			DontBatchDirectives:      routerJobDontBatchDirectives,
-		}
+		return proxyErrorResponse(respCode, fmt.Sprintf("[TransformerProxy] response is not valid JSON: %s, err: %v", string(respData), err), routerJobDontBatchDirectives)
 	}
 	if transportResponse.OriginalResponse != "" {
 		respData = []byte(transportResponse.OriginalResponse)
@@ -560,27 +537,13 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 		// There are no per-job results to apply without `output`. Continuing would set respData to
 		// an empty string and fail inside getResponse, reporting an empty body instead of the one
 		// we actually received. Stop here and return it.
-		return ProxyRequestResponse{
-			ProxyRequestStatusCode:   respCode,
-			ProxyRequestResponseBody: fmt.Sprintf("[TransformerProxy] response has no output field: %s", string(respData)),
-			RespContentType:          "text/plain; charset=utf-8",
-			RespStatusCodes:          routerJobResponseCodes,
-			RespBodys:                routerJobResponseBodys,
-			DontBatchDirectives:      routerJobDontBatchDirectives,
-		}
+		return proxyErrorResponse(respCode, fmt.Sprintf("[TransformerProxy] response has no output field: %s", string(respData)), routerJobDontBatchDirectives)
 	}
 	respData = []byte(output.Raw)
 
 	transResp, err := proxyReqParams.Adapter.getResponse(respData, respCode, proxyReqParams.ResponseData.Metadata)
 	if err != nil {
-		return ProxyRequestResponse{
-			ProxyRequestStatusCode:   respCode,
-			ProxyRequestResponseBody: err.Error(),
-			RespContentType:          "text/plain; charset=utf-8",
-			RespStatusCodes:          routerJobResponseCodes,
-			RespBodys:                routerJobResponseBodys,
-			DontBatchDirectives:      routerJobDontBatchDirectives,
-		}
+		return proxyErrorResponse(respCode, err.Error(), routerJobDontBatchDirectives)
 	}
 	// Non-fatal: the per-job results were still applied, but they may be attached to the wrong jobs.
 	// Recorded here, alongside the other breach reasons, so all three share one stats handle.
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -17,6 +17,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/samber/lo"
 	"github.com/tidwall/gjson"
@@ -433,10 +434,15 @@ func (trans *handle) emitProxyInvalidResponse(proxyReqParams *ProxyRequestParams
 // and a batch response can be multi-MB.
 func truncatedBody(b []byte) string {
 	const maxBodyBytes = int(10 * bytesize.KB)
-	if len(b) > maxBodyBytes {
-		return string(b[:maxBodyBytes]) + "...[truncated]"
+	if len(b) <= maxBodyBytes {
+		return string(b)
 	}
-	return string(b)
+	cut := b[:maxBodyBytes]
+	// Trim back off a partial rune so the cut does not emit U+FFFD.
+	if r, size := utf8.DecodeLastRune(cut); r == utf8.RuneError && size <= 1 {
+		cut = cut[:len(cut)-size]
+	}
+	return string(cut) + "...[truncated]"
 }
 
 // proxyErrorResponse builds the error-path response shared by every early return in ProxyRequest.
@@ -510,11 +516,11 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 		return proxyErrorResponse(respCode, requestError.Error(), routerJobDontBatchDirectives)
 	}
 
-	// Responses the oauth transport synthesized itself are infrastructure, not contract breaches, and
-	// get their own reason so the alert can exclude them. The upstream status is deliberately not part
-	// of this: v0 mirrors the destination's delivery status, so a destination 5xx would relabel a
-	// genuine breach as infra.
-	infraFailure := httpPrxResp.transportError
+	// Infrastructure, not a contract breach: either this transport synthesized the response, or it
+	// never came from the transformer at all (an ingress/LB error page sets no apiVersion). Status is
+	// deliberately not part of this - v0 mirrors the destination's delivery status, so a destination
+	// 5xx would relabel a genuine breach as infra.
+	infraFailure := httpPrxResp.transportError || !httpPrxResp.fromTransformer
 
 	/*
 		respData will be in ProxyResponseV0 or ProxyResponseV1
@@ -718,6 +724,10 @@ type httpProxyResponse struct {
 	// received from the transformer. Its body is a plain-text error string, so a non-JSON body on
 	// such a response is an infrastructure failure, not a transformer contract breach.
 	transportError bool
+	// fromTransformer reports that the response carried the apiVersion header, which only the
+	// transformer sets. An ingress/LB error page in front of the transformer carries neither it nor
+	// the transport marker, and is infrastructure rather than a contract breach.
+	fromTransformer bool
 }
 
 func (trans *handle) doProxyRequest(ctx context.Context, proxyUrl string, proxyReqParams *ProxyRequestParams, payload []byte) httpProxyResponse {
@@ -814,9 +824,10 @@ func (trans *handle) doProxyRequest(ctx context.Context, proxyUrl string, proxyR
 	}
 
 	return httpProxyResponse{
-		respData:       respData,
-		statusCode:     resp.StatusCode,
-		transportError: resp.Header.Get(oauthv2httpclient.TransportErrorHeader) != "",
+		respData:        respData,
+		statusCode:      resp.StatusCode,
+		transportError:  resp.Header.Get(oauthv2httpclient.TransportErrorHeader) != "",
+		fromTransformer: resp.Header.Get(apiVersionHeader) != "",
 	}
 }
 
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -412,10 +412,15 @@ func proxyJobIDs(metadata []ProxyRequestMetadata) []int64 {
 // to inspect the affected events. Every breach reason goes through here so the metric keeps a single
 // tag set and a single stats handle. The response body is deliberately not logged: these are
 // debugged by reading the events out of jobsDB by jobID.
-func emitProxyInvalidResponse(statsHandle stats.Stats, log logger.Logger, reason, destType, destinationID, msg string, fields ...logger.Field) {
+//
+// Takes proxyReqParams rather than the destType/destinationId strings so the two cannot be
+// transposed at a call site and silently mis-tag the metric. Callers must be past ProxyRequest's
+// empty-metadata guard, which every breach site is.
+func emitProxyInvalidResponse(statsHandle stats.Stats, log logger.Logger, proxyReqParams *ProxyRequestParams, reason, msg string, fields ...logger.Field) {
+	destinationID := proxyReqParams.ResponseData.Metadata[0].DestinationID
 	statsHandle.NewTaggedStat(`router.transformerproxy.invalid.response`, stats.CountType, stats.Tags{
 		"reason":        reason,
-		"destType":      destType,
+		"destType":      proxyReqParams.DestName,
 		"destinationId": destinationID,
 	}).Increment()
 	log.Warnn(msg, append([]logger.Field{obskit.DestinationID(destinationID)}, fields...)...)
@@ -499,9 +504,15 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 	// unmarshal unsuccessful scenarios
 	// if respData is not a valid json, surface it instead of discarding the error
 	if err := jsonrs.Unmarshal(respData, &transportResponse); err != nil {
-		emitProxyInvalidResponse(trans.stats, trans.logger, "unmarshal error", proxyReqParams.DestName,
-			proxyReqParams.ResponseData.Metadata[0].DestinationID,
-			"[TransformerProxy] proxy response unmarshal failed",
+		// A response synthesized by the oauth transport carries a plain-text error, not the
+		// transformer's payload, so a failed unmarshal there is an infrastructure failure (a short
+		// read, an unreachable transformer) rather than a contract breach. It stays observable under
+		// its own reason so the alert can exclude it without the signal disappearing.
+		reason, msg := "unmarshal error", "[TransformerProxy] proxy response unmarshal failed"
+		if httpPrxResp.transportError {
+			reason, msg = "transport error", "[TransformerProxy] transport returned a non-transformer response"
+		}
+		emitProxyInvalidResponse(trans.stats, trans.logger, proxyReqParams, reason, msg,
 			logger.NewIntSliceField("jobIDs", proxyJobIDs(proxyReqParams.ResponseData.Metadata)))
 		// The payload is not JSON, so nothing downstream can read it: `output` cannot exist, and
 		// continuing would blank respData and fail inside getResponse on an empty body — losing the
@@ -529,9 +540,12 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 		}
 	**/
 	output := gjson.GetBytes(respData, "output")
-	if !output.Exists() {
-		emitProxyInvalidResponse(trans.stats, trans.logger, "missing output", proxyReqParams.DestName,
-			proxyReqParams.ResponseData.Metadata[0].DestinationID,
+	// gjson reports an explicit JSON null as existing (Exists is `Type != Null || len(Raw) != 0`, and
+	// a literal null has Raw == "null"), so the type has to be checked too. Otherwise
+	// `{"output":null}` slips through, unmarshals into an empty response set, and gets reported as an
+	// "in out mismatch" - sending the runbook down the data-corruption branch instead of this one.
+	if !output.Exists() || output.Type == gjson.Null {
+		emitProxyInvalidResponse(trans.stats, trans.logger, proxyReqParams, "missing output",
 			"[TransformerProxy] proxy response missing output",
 			logger.NewIntSliceField("jobIDs", proxyJobIDs(proxyReqParams.ResponseData.Metadata)))
 		// There are no per-job results to apply without `output`. Continuing would set respData to
@@ -548,8 +562,7 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 	// Non-fatal: the per-job results were still applied, but they may be attached to the wrong jobs.
 	// Recorded here, alongside the other breach reasons, so all three share one stats handle.
 	if transResp.jobIDMismatch {
-		emitProxyInvalidResponse(trans.stats, trans.logger, "in out mismatch", proxyReqParams.DestName,
-			proxyReqParams.ResponseData.Metadata[0].DestinationID,
+		emitProxyInvalidResponse(trans.stats, trans.logger, proxyReqParams, "in out mismatch",
 			"[TransformerProxy] JobIDs in out mismatch",
 			logger.NewIntSliceField("jobIDsInMetadata", transResp.jobIDsInMetadata),
 			logger.NewIntSliceField("jobIDsInResponse", transResp.jobIDsInResponse))
@@ -667,6 +680,10 @@ type httpProxyResponse struct {
 	respData   []byte
 	statusCode int
 	err        error
+	// transportError reports that this response was synthesized by the oauth transport rather than
+	// received from the transformer. Its body is a plain-text error string, so a non-JSON body on
+	// such a response is an infrastructure failure, not a transformer contract breach.
+	transportError bool
 }
 
 func (trans *handle) doProxyRequest(ctx context.Context, proxyUrl string, proxyReqParams *ProxyRequestParams, payload []byte) httpProxyResponse {
@@ -763,8 +780,9 @@ func (trans *handle) doProxyRequest(ctx context.Context, proxyUrl string, proxyR
 	}
 
 	return httpProxyResponse{
-		respData:   respData,
-		statusCode: resp.StatusCode,
+		respData:       respData,
+		statusCode:     resp.StatusCode,
+		transportError: resp.Header.Get(oauthv2httpclient.TransportErrorHeader) != "",
 	}
 }
 
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -429,10 +429,8 @@ func (trans *handle) emitProxyInvalidResponse(proxyReqParams *ProxyRequestParams
 	trans.logger.Warnn(msg, append([]logger.Field{obskit.DestinationID(destinationID)}, fields...)...)
 }
 
-// truncatedBody bounds an upstream body embedded in an error string. Nothing consumes
-// ProxyRequestResponseBody (router/worker.go reads only the per-job maps), and a batch response can
-// be multi-MB, so the whole payload is never worth stringifying. Sliced before the conversion so the
-// full body is not allocated first.
+// truncatedBody bounds a body embedded in an error string: nothing reads ProxyRequestResponseBody
+// and a batch response can be multi-MB.
 func truncatedBody(b []byte) string {
 	const maxBodyBytes = int(10 * bytesize.KB)
 	if len(b) > maxBodyBytes {
@@ -512,24 +510,18 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 		return proxyErrorResponse(respCode, requestError.Error(), routerJobDontBatchDirectives)
 	}
 
-	// A response the oauth transport synthesized itself (failed token fetch, augmentation failure, LB
-	// round-trip error, unreadable body) is an infrastructure failure rather than a transformer
-	// contract breach, so it is reported under its own reason and excluded from the paging alert.
-	//
-	// The upstream status is deliberately not part of this signal: for v0 the transformer mirrors the
-	// destination's delivery status (deliveryPostProcess in rudder-transformer), so any destination 5xx
-	// comes from a perfectly healthy transformer and would relabel a genuine breach as infra. The
-	// status is logged instead, and only selects the reason at the points below where the response
-	// cannot be processed - a response that parses is applied exactly as before.
+	// Responses the oauth transport synthesized itself are infrastructure, not contract breaches, and
+	// get their own reason so the alert can exclude them. The upstream status is deliberately not part
+	// of this: v0 mirrors the destination's delivery status, so a destination 5xx would relabel a
+	// genuine breach as infra.
 	infraFailure := httpPrxResp.transportError
 
 	/*
 		respData will be in ProxyResponseV0 or ProxyResponseV1
 	*/
 	var transportResponse oauthv2.TransportResponse // response that we get from oauth-interceptor in postRoundTrip
-	// A non-oauth destination hands back the raw upstream body, so a non-JSON body here is either that
-	// infra failure or a genuinely corrupt transformer response; surface the error instead of
-	// discarding it.
+	// A non-oauth destination hands back the raw upstream body, so a non-JSON body here is either infra
+	// or a corrupt transformer response.
 	if err := jsonrs.Unmarshal(respData, &transportResponse); err != nil {
 		reason, msg := "unmarshal error", "[TransformerProxy] proxy response unmarshal failed"
 		if infraFailure {
@@ -538,8 +530,7 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 		trans.emitProxyInvalidResponse(proxyReqParams, reason, msg,
 			logger.NewIntField("statusCode", int64(httpPrxResp.statusCode)),
 			logger.NewIntSliceField("jobIDs", proxyJobIDs(proxyReqParams.ResponseData.Metadata)))
-		// The payload is not JSON, so nothing downstream can read it: `output` cannot exist, and
-		// continuing would blank respData and fail inside getResponse on an empty body. Stop here.
+		// Not JSON, so `output` cannot exist; continuing would blank respData and fail on an empty body.
 		return proxyErrorResponse(respCode, fmt.Sprintf("[TransformerProxy] response is not valid JSON: %s, err: %v", truncatedBody(respData), err), routerJobDontBatchDirectives)
 	}
 	if transportResponse.OriginalResponse != "" {
@@ -568,8 +559,7 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 	// `{"output":null}` slips through, unmarshals into an empty response set, and gets reported as an
 	// "in out mismatch" - sending the runbook down the data-corruption branch instead of this one.
 	if !output.Exists() || output.Type == gjson.Null {
-		// A transport-synthesized body has no output either; keep it under its own reason so the
-		// breach signal stays clean. Anything else reaching here is a genuine missing-output breach.
+		// A transport-synthesized body has no output either, so it keeps its own reason.
 		reason, msg := "missing output", "[TransformerProxy] proxy response missing output"
 		if infraFailure {
 			reason, msg = "transport error", "[TransformerProxy] transport-synthesized response has no output field"
@@ -577,18 +567,15 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 		trans.emitProxyInvalidResponse(proxyReqParams, reason, msg,
 			logger.NewIntField("statusCode", int64(httpPrxResp.statusCode)),
 			logger.NewIntSliceField("jobIDs", proxyJobIDs(proxyReqParams.ResponseData.Metadata)))
-		// There are no per-job results to apply without `output`. Continuing would set respData to
-		// an empty string and fail inside getResponse, reporting an empty body instead of the one
-		// we actually received. Stop here and return it.
+		// No per-job results to apply without `output`.
 		return proxyErrorResponse(respCode, fmt.Sprintf("[TransformerProxy] response has no output field: %s", truncatedBody(respData)), routerJobDontBatchDirectives)
 	}
 	respData = []byte(output.Raw)
 
 	transResp, err := proxyReqParams.Adapter.getResponse(respData, respCode, proxyReqParams.ResponseData.Metadata)
 	if err != nil {
-		// output was present but its per-job payload will not parse - a genuine contract breach, under
-		// the same "unmarshal error" reason as the outer envelope. A transport-synthesized body that
-		// happened to carry an output field is still infra.
+		// output was present but its payload will not parse - same "unmarshal error" reason as the
+		// outer envelope.
 		reason, msg := "unmarshal error", "[TransformerProxy] proxy response output unmarshal failed"
 		if infraFailure {
 			reason, msg = "transport error", "[TransformerProxy] transport-synthesized response output unmarshal failed"
@@ -598,23 +585,14 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 			logger.NewIntSliceField("jobIDs", proxyJobIDs(proxyReqParams.ResponseData.Metadata)))
 		return proxyErrorResponse(respCode, err.Error(), routerJobDontBatchDirectives)
 	}
-	// Compared as sets, so the check does not depend on duplicate jobIDs on either side: a batch can
-	// carry the same JobID more than once (router/worker.go only dedupes when building the final job
-	// responses) and the transformer may collapse or echo those. routerJobResponseCodes is keyed by
-	// the jobIDs the transformer answered for, so its keys are the response set. This only bites for
-	// v1: v0 keys the map by the request metadata, so the sets are equal by construction and a v0
-	// mismatch is not detectable here (nor was it before).
+	// Compared as sets: a request can legitimately carry the same JobID twice (router/worker.go only
+	// dedupes when building final responses). Only meaningful for v1 - v0 keys the map by the request
+	// metadata, so the sets are equal by construction.
 	jobIDsInMetadata := lo.Uniq(proxyJobIDs(proxyReqParams.ResponseData.Metadata))
 	slices.Sort(jobIDsInMetadata)
 	jobIDsInResponse := slices.Sorted(maps.Keys(transResp.routerJobResponseCodes))
-	// Non-fatal: the per-job results were still applied, but they may be attached to the wrong jobs.
-	// Recorded here, alongside the other breach reasons, so all share one stats handle.
-	//
-	// Two distinct breaches, both "results attached to the wrong jobs":
-	//   sets differ      - the transformer answered for a different set of jobs than we sent
-	//   entries > keys   - it answered twice for the same job, so one status was silently dropped by
-	//                      the map write. Duplicates in the *request* are legitimate (see above) and
-	//                      are why the set comparison dedupes; duplicates in the *response* are not.
+	// Non-fatal: the results were applied, but may be attached to the wrong jobs. Duplicate response
+	// entries collapse to one map key, silently dropping a status - a breach the set check cannot see.
 	duplicateResponses := transResp.responseEntries > len(transResp.routerJobResponseCodes)
 	if !slices.Equal(jobIDsInMetadata, jobIDsInResponse) || duplicateResponses {
 		trans.emitProxyInvalidResponse(proxyReqParams, "in out mismatch",
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -495,9 +495,10 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 	// if respData is not a valid json, surface it instead of discarding the error
 	if err := jsonrs.Unmarshal(respData, &transportResponse); err != nil {
 		trans.stats.NewTaggedStat(`router.transformerproxy.invalid.response`, stats.CountType, stats.Tags{
-			"reason":      "unmarshal error",
-			"destType":    proxyReqParams.DestName,
-			"workspaceId": proxyReqParams.ResponseData.Metadata[0].WorkspaceID,
+			"reason":        "unmarshal error",
+			"destType":      proxyReqParams.DestName,
+			"workspaceId":   proxyReqParams.ResponseData.Metadata[0].WorkspaceID,
+			"destinationId": proxyReqParams.ResponseData.Metadata[0].DestinationID,
 		}).Increment()
 	}
 	if transportResponse.OriginalResponse != "" {
@@ -523,9 +524,10 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 	output := gjson.GetBytes(respData, "output")
 	if !output.Exists() {
 		trans.stats.NewTaggedStat(`router.transformerproxy.invalid.response`, stats.CountType, stats.Tags{
-			"reason":      "missing output",
-			"destType":    proxyReqParams.DestName,
-			"workspaceId": proxyReqParams.ResponseData.Metadata[0].WorkspaceID,
+			"reason":        "missing output",
+			"destType":      proxyReqParams.DestName,
+			"workspaceId":   proxyReqParams.ResponseData.Metadata[0].WorkspaceID,
+			"destinationId": proxyReqParams.ResponseData.Metadata[0].DestinationID,
 		}).Increment()
 	}
 	respData = []byte(output.Raw)
--- a/router/transformer/transformer.go
+++ b/router/transformer/transformer.go
@@ -417,14 +417,14 @@ func proxyJobIDs(metadata []ProxyRequestMetadata) []int64 {
 // Takes proxyReqParams rather than the destType/destinationId strings so the two cannot be
 // transposed at a call site and silently mis-tag the metric. Callers must be past ProxyRequest's
 // empty-metadata guard, which every breach site is.
-func emitProxyInvalidResponse(statsHandle stats.Stats, log logger.Logger, proxyReqParams *ProxyRequestParams, reason, msg string, fields ...logger.Field) {
+func (trans *handle) emitProxyInvalidResponse(proxyReqParams *ProxyRequestParams, reason, msg string, fields ...logger.Field) {
 	destinationID := proxyReqParams.ResponseData.Metadata[0].DestinationID
-	statsHandle.NewTaggedStat(`router.transformerproxy.invalid.response`, stats.CountType, stats.Tags{
+	trans.stats.NewTaggedStat(`router.transformerproxy.invalid.response`, stats.CountType, stats.Tags{
 		"reason":        reason,
 		"destType":      proxyReqParams.DestName,
 		"destinationId": destinationID,
 	}).Increment()
-	log.Warnn(msg, append([]logger.Field{obskit.DestinationID(destinationID)}, fields...)...)
+	trans.logger.Warnn(msg, append([]logger.Field{obskit.DestinationID(destinationID)}, fields...)...)
 }
 
 // proxyErrorResponse builds the error-path response shared by every early return in ProxyRequest.
@@ -498,22 +498,27 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 		return proxyErrorResponse(respCode, requestError.Error(), routerJobDontBatchDirectives)
 	}
 
+	// An upstream 5xx (transformer/LB/gateway) or a transport-synthesized response is an
+	// infrastructure failure, not a transformer contract breach. transportError marks the synthesized
+	// responses (including 4xx ones like a failed token fetch); the status marks a real upstream error
+	// that carried a body. Classifying these as breaches would page the contract owner on our own
+	// infra, so they get a reason the alert can exclude.
+	infraFailure := httpPrxResp.transportError || httpPrxResp.statusCode >= http.StatusInternalServerError
+
 	/*
 		respData will be in ProxyResponseV0 or ProxyResponseV1
 	*/
 	var transportResponse oauthv2.TransportResponse // response that we get from oauth-interceptor in postRoundTrip
-	// unmarshal unsuccessful scenarios
-	// if respData is not a valid json, surface it instead of discarding the error
+	// A non-oauth destination hands back the raw upstream body, so a non-JSON body here is either that
+	// infra failure or a genuinely corrupt transformer response; surface the error instead of
+	// discarding it.
 	if err := jsonrs.Unmarshal(respData, &transportResponse); err != nil {
-		// A response synthesized by the oauth transport carries a plain-text error, not the
-		// transformer's payload, so a failed unmarshal there is an infrastructure failure (a short
-		// read, an unreachable transformer) rather than a contract breach. It stays observable under
-		// its own reason so the alert can exclude it without the signal disappearing.
 		reason, msg := "unmarshal error", "[TransformerProxy] proxy response unmarshal failed"
-		if httpPrxResp.transportError {
+		if infraFailure {
 			reason, msg = "transport error", "[TransformerProxy] transport returned a non-transformer response"
 		}
-		emitProxyInvalidResponse(trans.stats, trans.logger, proxyReqParams, reason, msg,
+		trans.emitProxyInvalidResponse(proxyReqParams, reason, msg,
+			logger.NewIntField("statusCode", int64(httpPrxResp.statusCode)),
 			logger.NewIntSliceField("jobIDs", proxyJobIDs(proxyReqParams.ResponseData.Metadata)))
 		// The payload is not JSON, so nothing downstream can read it: `output` cannot exist, and
 		// continuing would blank respData and fail inside getResponse on an empty body — losing the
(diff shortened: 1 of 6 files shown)


router/transformer/transformer.go
```diff
@@ -491,9 +491,20 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 		respData will be in ProxyResponseV0 or ProxyResponseV1
 	*/
 	var transportResponse oauthv2.TransportResponse // response that we get from oauth-interceptor in postRoundTrip
-	_ = jsonrs.Unmarshal(respData, &transportResponse)
 	// unmarshal unsuccessful scenarios
-	// if respData is not a valid json
+	// if respData is not a valid json, surface it instead of discarding the error
+	if err := jsonrs.Unmarshal(respData, &transportResponse); err != nil {
```
shekhar-rudder, line 540: **🔴 Correctness: invalid-JSON breach double-counts the metric and double-logs.**

This `unmarshal error` branch does not early-return. When `respData` is not valid JSON (HTML/plaintext 502 from a proxy/CDN, truncated body):
1. `Unmarshal` fails → `reason="unmarshal error"` increments + one `Warnn`.
2. Control falls through; `OriginalResponse` stays `""` so `respData` is unchanged.
3. `gjson.GetBytes(respData,"output").Exists()` is then `false` → the `missing output` branch below (line 529) **also** fires with `reason="missing output"` + a second `Warnn`.

So one real breach increments `router.transformerproxy.invalid.response` twice under two reasons and logs twice. Since operator#1452 fires at `>0` on `sum by (…, reason)`, a single non-JSON response trips two reason-series (double page) and makes per-reason counts non-additive.

**Fix:** early-return (or `else if`) after the unmarshal-error emit — a failed unmarshal already implies `output` can't exist, so the second branch only adds noise.

router/transformer/transformer.go
```diff
@@ -514,9 +525,22 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 			}
 		}
 	**/
-	respData = []byte(gjson.GetBytes(respData, "output").Raw)
-
-	transResp, err := proxyReqParams.Adapter.getResponse(respData, respCode, proxyReqParams.ResponseData.Metadata)
+	output := gjson.GetBytes(respData, "output")
```
shekhar-rudder, line 571: **🟡 Cleanup/altitude: three near-identical emit blocks should be one helper.**

This block, the `unmarshal error` block above, and the `in out mismatch` block in `transformer_proxy_adapter.go:164` all build the same `{reason, destType, destinationId}` tag set on the same metric and log jobIDs — with the `lo.Map(...Metadata → JobID)` closure copy-pasted at lines 504 and 537. They've already drifted: the two new sites use `obskit.DestinationID` + a `jobIDs` field; the adapter uses `logger.NewStringField("destinationId")` + `jobIDsInMetadata/jobIDsInResponse`.

Consider collapsing into one `emitInvalidResponse(reason, destType, metadata)` helper so a fourth reason (the pattern invites more) is one line, not another ~10-line copy that can silently diverge in tag set.

router/transformer/transformer_proxy_adapter.go
```diff
@@ -162,9 +162,12 @@ func (v1 *v1Adapter) getResponse(respData []byte, respCode int, metadata []Proxy
 
 	if !reflect.DeepEqual(jobIDsInMetadata, jobIDsInResponse) {
 		stats.Default.NewTaggedStat(`router.transformerproxy.invalid.response`, stats.CountType, stats.Tags{
```
shekhar-rudder: **🟡 Cleanup/testability: metric split across two stats sinks.**

The two new emit sites in `transformer.go` use the injected `trans.stats`; this site uses the global `stats.Default` (pre-existing, but this PR now makes all three the *same* metric). Consequences: the metric is written via two different sinks, and a mock stats handle injected on the handle observes only 2 of the 3 emission sites — the `in out mismatch` reason is invisible to per-handle routing and to tests.

The `emitInvalidResponse` helper suggested above (taking `trans.stats`) would fix this naturally; standalone it needs plumbing a stats handle into `v1Adapter`.

shekhar-rudder (commented):
Automated `/code-review` (high effort). 1 correctness bug + 2 cleanup findings, posted inline. The PR is safe additive instrumentation overall, but finding #1 will make the alert this PR unblocks double-fire.

router/transformer/transformer.go
saikumarrs: Confirmed and fixed in `eda0c00c`. Both breach branches now return a `ProxyRequestResponse` instead of falling through, so the double count is structurally impossible rather than filtered out. Returning also stops the old path from blanking `respData` before the error body is built, which turned out to be the more useful half: three tests were asserting `respData: ` with an empty payload, so the response never carried the thing that explains the failure. They now assert the real body. `ProxyRequestStatusCode` is unchanged, so job outcomes and retry behaviour are unaffected.

router/transformer/transformer.go
saikumarrs: Collapsed into `emitProxyInvalidResponse(statsHandle, log, reason, destType, destinationID, msg, fields...)`, with `proxyJobIDs` replacing the copy-pasted `lo.Map` closure at both sites. A fourth reason is now one call. `eda0c00c`.

One bit of the drift I kept deliberately: the log fields stay variadic rather than normalised to a single `jobIDs`. The mismatch site logs `jobIDsInMetadata` and `jobIDsInResponse` separately because the difference between those two lists is what identifies the mis-attributed jobs, and the runbook step reads both. The `destinationId` field is unified though, via `obskit.DestinationID` in the helper.

router/transformer/transformer_proxy_adapter.go
saikumarrs: Fixed by moving the emit rather than plumbing a second handle into the adapter. Plumbing would have given two injectable sinks instead of one, and the adapter is built in `worker.go` where the transformer handle's stats is not in scope. The adapter now reports the breach on `TransResponse` (`jobIDMismatch` plus the two sorted jobID sets) and `ProxyRequest` records it through `trans.stats`, so all three reasons share one handle and a mock injected on the handle observes every site. `eda0c00c`.

Side effect worth noting: with the emit gone, the `destType` parameter I had threaded into `getResponse` existed only to build that tag, so that signature change is reverted across the interface, both adapters, and their tests. The adapter test that asserted a warning log now asserts the returned mismatch, which is the adapter's actual contract.

router/transformer/transformer.go
```diff
@@ -491,9 +512,25 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
...
+	// if respData is not a valid json, surface it instead of discarding the error
+	if err := jsonrs.Unmarshal(respData, &transportResponse); err != nil {
+		emitProxyInvalidResponse(trans.stats, trans.logger, "unmarshal error", proxyReqParams.DestName,
+			proxyReqParams.ResponseData.Metadata[0].DestinationID,
+			"[TransformerProxy] proxy response unmarshal failed",
+			logger.NewIntSliceField("jobIDs", proxyJobIDs(proxyReqParams.ResponseData.Metadata)))
+		// The payload is not JSON, so nothing downstream can read it: `output` cannot exist, and
+		// continuing would blank respData and fail inside getResponse on an empty body — losing the
+		// very payload that explains the breach. Stop here and return it.
+		return ProxyRequestResponse{
```
shekhar-rudder: **Cleanup (low severity):** this error-path `ProxyRequestResponse` literal — same 6 fields, only the status code and body message vary — is now built at ~7 sites in `ProxyRequest`, and these two new early returns add to it. Not a bug, just maintenance cost: a change to the error-response shape (e.g. adding a breach-reason field) has to be edited across all of them, with a real risk of one site drifting.

Consider a small helper to collapse them:
```go
func (trans *handle) proxyErrorResponse(code int, body string, dontBatch map[int64]bool) ProxyRequestResponse {
    return ProxyRequestResponse{
        ProxyRequestStatusCode:   code,
        ProxyRequestResponseBody: body,
        RespContentType:          "text/plain; charset=utf-8",
        RespStatusCodes:          map[int64]int{},
        RespBodys:                map[int64]string{},
        DontBatchDirectives:      dontBatch,
    }
}
```

_Rest of the review was clean — the swallowed-Unmarshal fix and the missing-`output` guard are behavior-preserving vs. the old fall-through (both previously errored on the same paths via `getResponse`), and moving the mismatch metric onto the injected `trans.stats` is a real improvement. 🤖 via /code-review_

saikumarrs: Collapsed all seven into `proxyErrorResponse(code, body, dontBatch)`. `ef59c04d`.

Kept your empty-literal shape for `RespStatusCodes`/`RespBodys` after confirming the invariant holds: `routerJobResponseCodes` and `routerJobResponseBodys` were declared in `ProxyRequest` and never written on any path - the writes near the end target `transResp`'s maps, not those locals - so every error site was already handing back an empty map. Inlining the literals left both locals dead, so they are removed too. Net -37 lines. I put the reasoning in a comment on the helper so a future change that starts populating per-job results on an error path does not silently drop them.

Two deliberate deviations from the snippet: the success return stays a literal, since it differs in content type, carries `OAuthErrorCategory`, and returns the adapter's populated maps - folding it in would mean parameterising away exactly what makes it different. And it is a plain function rather than a method, matching `proxyJobIDs` and `emitProxyInvalidResponse`, because it does not use the receiver.

shekhar-rudder (commented):
Automated `/code-review` (high effort). Found 3 correctness issues (2 confirmed, 1 plausible) + 1 minor cleanup. All are about **what now trips the metric** — which matters because the downstream `proxy-invalid-response` alert fires at `> 0`. The refactor itself (`proxyErrorResponse`, the `stats.Default`→`trans.stats` move, v0 leaving `jobIDMismatch` zero-valued) is correct. Not flagging the `destinationId` cardinality or missing `workspaceId` — both are deliberate, documented decisions in the PR description.

router/transformer/transformer.go
```diff
@@ -477,23 +489,25 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
...
 
 	/*
 		respData will be in ProxyResponseV0 or ProxyResponseV1
 	*/
 	var transportResponse oauthv2.TransportResponse // response that we get from oauth-interceptor in postRoundTrip
-	_ = jsonrs.Unmarshal(respData, &transportResponse)
 	// unmarshal unsuccessful scenarios
-	// if respData is not a valid json
+	// if respData is not a valid json, surface it instead of discarding the error
+	if err := jsonrs.Unmarshal(respData, &transportResponse); err != nil {
```
shekhar-rudder, line 540: **Transient body-read failures now trip the breach metric (and the `>0` alert).**

When the OAuth interceptor's `postRoundTrip` fails to read the response body it returns HTTP 500 with a **non-JSON** body `[postRoundTrip]Error reading response body: unexpected EOF` and `requestError == nil` (`services/oauth/v2/http/transport.go`). That slips past the earlier `if requestError != nil` guard, `jsonrs.Unmarshal` fails here, and we now emit `reason="unmarshal error"` + a Warn. Before this PR the error was discarded (`_ =`) and no metric was emitted on this path. The `ioReadError` test case (`transformer_test.go:1124`) exercises exactly this.

Net: an ordinary truncated/short HTTP response — a network blip — is now indistinguishable from a genuine contract breach and will page on-call via the `> 0` alert. Consider filtering the transport-level `[postRoundTrip]...` sentinel (or the EOF class) out of the "breach" reason so only genuine non-JSON *destination* responses count.

router/transformer/transformer.go
```diff
@@ -514,18 +528,31 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
 			}
 		}
 	**/
-	respData = []byte(gjson.GetBytes(respData, "output").Raw)
+	output := gjson.GetBytes(respData, "output")
+	if !output.Exists() {
```
shekhar-rudder: **`{"output":null}` is misclassified as `in out mismatch` instead of `missing output`.**

`gjson`'s `.Exists()` returns `true` for an explicit JSON `null`, so this guard is skipped, `respData` becomes the literal `"null"`, and `getResponse` unmarshals it into an empty `ProxyResponseV1` without error. With an empty response set vs. non-empty metadata, `jobIDMismatch` becomes true → the breach is tagged `reason="in out mismatch"` rather than the intended `"missing output"`. Since this PR's whole purpose is per-`reason` attribution for the runbook branch, this mis-routes the on-call response.

Suggest: `if !output.Exists() || output.Type == gjson.Null {`

router/transformer/transformer_proxy_adapter.go
```diff
@@ -160,14 +167,10 @@ func (v1 *v1Adapter) getResponse(respData []byte, respCode int, metadata []Proxy
...
-			"reason": "in out mismatch",
-		}).Increment()
-		v1.logger.Warnn("[TransformerProxy] JobIDs in out mismatch",
-			logger.NewIntSliceField("jobIDsInMetadata", jobIDsInMetadata),
-			logger.NewIntSliceField("jobIDsInResponse", jobIDsInResponse))
-	}
+	// Report the mismatch rather than recording it here. Emitting from the adapter would write this
+	// metric through the package-global stats handle while the other breach reasons use the
+	// caller's, splitting one metric across two sinks and hiding this reason from an injected mock.
+	jobIDMismatch := !reflect.DeepEqual(jobIDsInMetadata, jobIDsInResponse)
```
shekhar-rudder: **Un-deduped metadata JobIDs can produce a false `in out mismatch`.**

`jobIDsInMetadata` is `lo.Map` over all metadata with no dedup, while `jobIDsInResponse` is one entry per response element. Duplicate JobIDs in a batch are a real, known occurrence — `router/worker.go:855` already does `lo.UniqBy(..., JobID)` with a TODO noting the transformer response isn't validated. So metadata `[11,11,21]` vs response `[11,21]` yields a spurious mismatch.

This logic is pre-existing, but this PR promotes it to an alerted metric, so the latent false positive becomes actionable noise. Deduping `jobIDsInMetadata` before the `reflect.DeepEqual` would remove it.

router/transformer/transformer.go
```diff
@@ -400,23 +400,49 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
...
+	return lo.Map(metadata, func(m ProxyRequestMetadata, _ int) int64 {
+		return m.JobID
+	})
+}
+
+// emitProxyInvalidResponse records a transformer-proxy contract breach and logs the jobIDs needed
+// to inspect the affected events. Every breach reason goes through here so the metric keeps a single
+// tag set and a single stats handle. The response body is deliberately not logged: these are
+// debugged by reading the events out of jobsDB by jobID.
+func emitProxyInvalidResponse(statsHandle stats.Stats, log logger.Logger, reason, destType, destinationID, msg string, fields ...logger.Field) {
```
shekhar-rudder: **Minor (cleanup):** `destType` and `destinationID` are always the same two expressions at all three call sites (`proxyReqParams.DestName` and `proxyReqParams.ResponseData.Metadata[0].DestinationID`). Both are strings, so a future fourth call site can transpose them and silently mis-tag the metric. Passing `proxyReqParams` (or the already-built label struct) once would drop the boilerplate and the ordering hazard.

router/transformer/transformer.go
saikumarrs: You are right, and the problem was wider than the one call site. Every response `httpResponseCreator` builds has a plain-text body, so matching on a sentinel string would have covered 1 of the 8 places it is called from and left the rest reporting our own transport faults as transformer contract breaches. Instead the marker goes on the constructor itself: `httpResponseCreator` now sets `X-Rudder-Oauth-Transport-Error`, `doProxyRequest` reads it into `httpProxyResponse.transportError`, and `ProxyRequest` switches the reason to `transport error` when it is set. All 8 sites are covered and a 9th added later gets it for free. `0da26266`, `2041e8f4`.

Worth flagging for the alert side: this is exactly the case that would have paged us for a network fault, so I want it as its own reason rather than folded into `unmarshal error`. Added a regression test on the existing `ioReadError` case that asserts the reason, and it fails with `unmarshal error` if the classification regresses.

router/transformer/transformer.go
saikumarrs: Confirmed - `Exists()` is `t.Type != Null || len(t.Raw) != 0`, and a literal `null` has `Raw == "null"`, so `{"output": null}` was slipping past and failing further down in `getResponse` with a vaguer error instead of being reported as the contract breach it is. Now checks `!output.Exists() || output.Type == gjson.Null`. `2041e8f4`.

router/transformer/transformer_proxy_adapter.go
saikumarrs: Good catch, this would have been a false page. `router/worker.go` dedupes the same way before building responses, so a batch can legitimately carry a JobID twice while the transformer returns one entry per job - metadata `[11,11,21]` against response `[11,21]` was a mismatch purely from the duplicate. `jobIDsInMetadata` now goes through `lo.Uniq` before the sort and compare. `2041e8f4`.

router/transformer/transformer.go
saikumarrs: Switched it to take `proxyReqParams` and pull `DestName` and `Metadata[0].DestinationID` itself, so there is no longer a pair of adjacent same-typed arguments to transpose. That matters more than usual here since `destinationId` is the routing key on the alert - a transposition would have sent the page at the right time with the wrong destination on it. `2041e8f4`.

router/transformer/transformer_proxy_adapter.go
```diff
@@ -151,23 +158,22 @@ func (v1 *v1Adapter) getResponse(respData []byte, respCode int, metadata []Proxy
...
-	jobIDsInMetadata := lo.Map(metadata, func(m ProxyRequestMetadata, _ int) int64 {
+	// Deduped: a batch can legitimately carry the same JobID more than once (router/worker.go dedupes
+	// the same way before building responses), while the transformer returns one entry per job. Without
+	// this, metadata [11,11,21] against response [11,21] is a spurious mismatch - and now an alerted one.
+	jobIDsInMetadata := lo.Uniq(lo.Map(metadata, func(m ProxyRequestMetadata, _ int) int64 {
 		return m.JobID
-	})
+	}))
 	slices.Sort(jobIDsInMetadata)
 	jobIDsInResponse := lo.Map(transformerResponse.Response, func(resp TPDestResponse, _ int) int64 {
```
shekhar-rudder: **Asymmetric dedup can trigger a spurious `in out mismatch` — and this now pages.**

`jobIDsInMetadata` is deduped with `lo.Uniq`, but `jobIDsInResponse` a few lines below is **not**. That makes the `reflect.DeepEqual` comparison sensitive to duplicate JobIDs on the response side.

`router/worker.go` sends **all** metadata entries to the transformer, including duplicate JobIDs — it only dedupes when building the final job responses (`prepareRouterJobResponses`, via `lo.UniqBy`, with the explicit *"TODO: remove this once we enforce the necessary validations in the transformer's response"*). So a batch with a repeated JobID is a real, expected input.

Concretely, for metadata `[11,11,21]`:
- `jobIDsInMetadata` = `Uniq(...)` → `[11,21]`
- if the v1 transformer echoes one response entry per input item → `jobIDsInResponse` → `[11,11,21]`
- `reflect.DeepEqual([11,21], [11,11,21])` → `false` → `jobIDMismatch` fires

The **old** code compared both sides un-deduped (`[11,11,21] == [11,11,21]`) and did *not* flag this. Since this metric now drives the `proxy-invalid-response` paging alert, the one-sided dedup can turn a legitimate duplicate-job batch into a false page.

**Fix:** dedup both sides so the comparison is set-vs-set and robust regardless of whether the transformer collapses or echoes duplicates:

```go
jobIDsInResponse := lo.Uniq(lo.Map(transformerResponse.Response, func(resp TPDestResponse, _ int) int64 {
    return resp.Metadata.JobID
}))
```

saikumarrs: You are right, and this was a regression I introduced - the old un-deduped comparison did not flag this case and my one-sided version does. Confirmed the premise too: `prepareRouterJobResponses` dedupes at `router/worker.go:856` when building the final responses, so duplicates do reach the transformer.

Deduped both sides, which I think is the better framing than fixing the asymmetry: the check is meant to answer "did the transformer return results for the same set of jobs we sent", and as a set comparison it no longer depends on whether the transformer collapses or echoes duplicates - neither of which is specified. A genuine breach is still caught, since response `[11,11]` against metadata `[11,21]` is `{11}` vs `{11,21}`.

Added a table test over both transformer behaviours. The echo case fails without the second `lo.Uniq`, so the regression cannot come back silently. `e3ae4b37`.

router/transformer/transformer_proxy_adapter.go
```diff
@@ -52,6 +51,14 @@ type TransResponse struct {
...
 	authErrorCategory            string
 	statTags                     map[string]string
+	// jobIDMismatch reports that the transformer returned results for a different set of jobIDs
+	// than were sent, meaning per-job results are attributed to the wrong jobs. The adapter only
+	// reports it; the caller records the metric, so every breach reason goes through one stats
+	// handle. jobIDsInMetadata/jobIDsInResponse are the two sorted sets, and their difference is
+	// what makes the breach diagnosable.
+	jobIDMismatch    bool
+	jobIDsInMetadata []int64
+	jobIDsInResponse []int64
```
shekhar-rudder: `jobIDsInMetadata` / `jobIDsInResponse` are derivable state — they don't need to live on `TransResponse`. They're only read inside the `if transResp.jobIDMismatch` log in `ProxyRequest`, and at that point the caller already holds everything needed to reconstruct them:

- `jobIDsInMetadata` → from `proxyReqParams.ResponseData.Metadata` (which the caller passes *into* `getResponse`) via the existing `proxyJobIDs` helper.
- `jobIDsInResponse` → the sorted keys of `transResp.routerJobResponseCodes`, since that map is built as `routerJobResponseCodes[resp.Metadata.JobID] = resp.StatusCode`, so its keys are exactly the unique response jobIDs.

Carrying two slices on a struct returned for *every* proxy response, to serve a *rare* log line, is redundant.

One nuance: dropping only the two slices while keeping `jobIDMismatch` computed in the adapter isn't a clean win — the adapter still has to build both slices to derive the bool, so the caller would end up recomputing them for the log (duplicate work). The cleaner move is to drop **all three** fields and do the comparison at the caller, alongside the other two breach reasons (`unmarshal error`, `missing output`) which are already detected there:

```go
jobIDsInMetadata := lo.Uniq(proxyJobIDs(proxyReqParams.ResponseData.Metadata))
slices.Sort(jobIDsInMetadata)
jobIDsInResponse := lo.Keys(transResp.routerJobResponseCodes)
slices.Sort(jobIDsInResponse)
if !slices.Equal(jobIDsInMetadata, jobIDsInResponse) {
    emitProxyInvalidResponse(trans.stats, trans.logger, proxyReqParams, "in out mismatch",
        "[TransformerProxy] JobIDs in out mismatch",
        logger.NewIntSliceField("jobIDsInMetadata", jobIDsInMetadata),
        logger.NewIntSliceField("jobIDsInResponse", jobIDsInResponse))
}
```

This shrinks the adapter's contract back to "parse the response into the maps" and puts all three breach reasons in one place. Behaviourally identical for both versions: v0 keys `routerJobResponseCodes` by metadata jobIDs, so its keys always equal the metadata set → never a mismatch, same as today.

Trade-off: it makes "map keys == response jobID set" a load-bearing invariant of the caller (worth a comment). If you'd instead prefer the adapter to stay the owner of the version-specific match check, then keeping the current fields is the right call and the slices aren't redundant — but then the middle option (drop slices, keep the bool in the adapter) is the one to avoid.

saikumarrs: Agreed, and I took the full version rather than just deduping the second slice. All three fields are gone from `TransResponse`, and the mismatch check now lives in `ProxyRequest` alongside `unmarshal error` and `missing output`. `jobIDsInResponse` comes from `lo.Keys(transResp.routerJobResponseCodes)` - the map is keyed by the jobIDs the transformer answered for, so its keys are the response set, and map keys are inherently unique. That also retires last round's bug structurally: there is no longer a second slice that can be left un-deduped, so a duplicate-jobID batch can't page by accident regardless of how the transformer handles duplicates.

Added the invariant comment you flagged, on both the map-building site in the adapter and the comparison in the caller. Behaviour is unchanged for both versions - v0 keys the map by request jobIDs so its keys always equal the request set and it never mismatches, same as today.

The adapter no longer imports `lo`/`slices`/`reflect`; its tests now assert the response map is keyed by the deduped set (the invariant), and a new `ProxyRequest` case asserts a transformer answering for a different jobID set pages with `in out mismatch`. `5a725407`.

saikumarrs:
Pushed a follow-up (`601b7887`) from an internal review of this PR's own classification logic - flagging it here since it changes behavior rather than answering a thread.

**Gap:** the `transport error` reason only covered responses the oauth transport synthesizes itself (`httpResponseCreator`). The common real infra failure - the transformer/LB/gateway returning an HTTP 5xx with a body - is a valid HTTP response (`err == nil`) with no marker header, so it reached the content classifier and paged as a *contract breach*:
- oauth destinations: `postRoundTrip` wraps the 5xx body into a valid envelope, the outer unmarshal succeeds, `output` is absent -> `missing output`.
- non-oauth destinations: the raw non-JSON body fails the outer unmarshal -> `unmarshal error`.

Both page the contract owner on a fleet-wide infra event - the exact false positive `transport error` was added to prevent. The existing `lbError` tests actually encoded this (`missing output`).

**Fix:** classify infra up front - `infraFailure := transportError || statusCode >= 500`. The header still pulls its weight (it catches a synthesized 4xx like a failed token fetch that a status check alone would miss); the status check covers upstream 5xx-with-body that the header can't reach. Infra is reported as `transport error` and returned before the content checks, so those only ever see genuine transformer responses. This also drops the "a synthesized body is always non-JSON" assumption - `transportError` is now checked independently of unmarshal success.

Also in the same commit:
- Emit `unmarshal error` on the adapter `getResponse` failure - a present-but-malformed `output` (e.g. `{"output":[1,2,3]}`) was the one breach class incrementing no metric.
- The runner now asserts the *exact set* of breach reasons, so a healthy response asserting none catches a spurious page (not just a mislabel); added a duplicate-jobID batch case (must not page) and asserted the `getResponse` unmarshal reason.
- `emitProxyInvalidResponse` is now a `*handle` method; the upstream `statusCode` is logged on transport errors.

The two `lbError` test cases now expect `transport error`. Happy to walk through any of it.

router/transformer/transformer.go
```diff
@@ -477,23 +496,37 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
...
+		if infraFailure {
+			reason, msg = "transport error", "[TransformerProxy] transport returned a non-transformer response"
+		}
+		trans.emitProxyInvalidResponse(proxyReqParams, reason, msg,
+			logger.NewIntField("statusCode", int64(httpPrxResp.statusCode)),
+			logger.NewIntSliceField("jobIDs", proxyJobIDs(proxyReqParams.ResponseData.Metadata)))
+		// The payload is not JSON, so nothing downstream can read it: `output` cannot exist, and
+		// continuing would blank respData and fail inside getResponse on an empty body — losing the
+		// very payload that explains the breach. Stop here and return it.
+		return proxyErrorResponse(respCode, fmt.Sprintf("[TransformerProxy] response is not valid JSON: %s, err: %v", string(respData), err), routerJobDontBatchDirectives)
```
shekhar-rudder: **[blocking]** The new `transport error` reason makes `router.transformerproxy.invalid.response` fire on routine infrastructure failures, and the gating alert (rudderlabs/rudderstack-operator#1452) fires at `> 0` per `(namespace, destType, destinationId, reason)`.

Every response synthesized by `httpResponseCreator` now carries the header, and each of these reaches this branch with `requestError == nil` (`RoundTrip` returns the synthesized response with a `nil` error, so `doProxyRequest` falls through to the normal return):

- failed token fetch — `services/oauth/v2/http/transport.go:118`
- secret-augmentation failure — `transport.go:129`
- LB / round-trip error — `transport.go:294`
- body-read failure — `transport.go:145`

A single destination with a broken OAuth account emits one count per proxy request for as long as it stays broken, so the alert pages continuously on what is a destination-config problem, not a transformer contract breach. These paths are already covered by `transformer_proxy_request_result{requestSuccess=...}`.

Suggest either not emitting at all when `infraFailure` is true, or excluding `reason="transport error"` in the alert — and if it's the latter, that exclusion should land before or with this PR, since the description gates the alert only on deploy+verify.

router/transformer/transformer.go
```diff
@@ -477,23 +496,37 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
...
+		return proxyErrorResponse(respCode, requestError.Error(), routerJobDontBatchDirectives)
 	}
 
+	// An upstream 5xx (transformer/LB/gateway) or a transport-synthesized response is an
+	// infrastructure failure, not a transformer contract breach. transportError marks the synthesized
+	// responses (including 4xx ones like a failed token fetch); the status marks a real upstream error
+	// that carried a body. It only selects the reason at the points below where the response cannot be
+	// processed - it never short-circuits a response that parses, so a 5xx that still carries a usable
+	// output is applied exactly as before, and only the metric reason changes for the failures.
+	infraFailure := httpPrxResp.transportError || httpPrxResp.statusCode >= http.StatusInternalServerError
```
shekhar-rudder: **[blocking]** `statusCode >= 500` is not a reliable infra signal here, because for **v0** the transformer proxy sets the HTTP status *to the destination's delivery status*:

```ts
// rudder-transformer src/controllers/delivery.ts (deliverToDestination)
ctx.body = { output: deliveryResponse };
ControllerUtility.deliveryPostProcess(ctx, deliveryResponse.status);
```

and `getCompatibleStatusCode` passes non-2xx through unchanged (only 404→400, and 2xx→200). So any destination 5xx yields an HTTP 5xx from a perfectly healthy transformer.

Consequence: a genuine `missing output` / `unmarshal error` breach that happens to co-occur with a destination 5xx gets relabeled `transport error` — precisely the case INT-6839 exists to catch, routed to the wrong runbook branch. The same applies on v1 whenever `authErrorCategory` is set with a 5xx delivery status.

Suggest gating on `httpPrxResp.transportError` alone, and keeping the upstream status as a log field rather than a reason selector.

router/transformer/transformer.go
```diff
@@ -514,18 +547,58 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
...
+	// responses) and the transformer may collapse or echo those. routerJobResponseCodes is keyed by
+	// the jobIDs the transformer answered for, so its keys are the response set. This only bites for
+	// v1: v0 keys the map by the request metadata, so the sets are equal by construction and a v0
+	// mismatch is not detectable here (nor was it before).
+	jobIDsInMetadata := lo.Uniq(proxyJobIDs(proxyReqParams.ResponseData.Metadata))
+	slices.Sort(jobIDsInMetadata)
+	jobIDsInResponse := slices.Sorted(maps.Keys(transResp.routerJobResponseCodes))
+	// Non-fatal: the per-job results were still applied, but they may be attached to the wrong jobs.
+	// Recorded here, alongside the other breach reasons, so all share one stats handle.
+	if !slices.Equal(jobIDsInMetadata, jobIDsInResponse) {
```
shekhar-rudder: **[medium]** Moving to a deduped-set comparison trades a false positive for a false negative.

The old code compared `[1,2]` against `[1,2,2]` and flagged it. The new comparison of `lo.Uniq(metadata)` against `maps.Keys(routerJobResponseCodes)` does not — and the duplicate response entry silently last-write-wins in `v1Adapter.getResponse`:

```go
for _, resp := range transformerResponse.Response {
	routerJobResponseCodes[resp.Metadata.JobID] = resp.StatusCode
	...
}
```

so when the transformer returns two entries for the same jobID with different status codes, one is dropped and nothing is recorded.

Duplicates in the *request* are legitimate (that's the justification in the comment above, and it's correct). Duplicates in the *response* with conflicting statuses are exactly the "attributed to the wrong jobs" breach this PR is meant to surface. Suggest keeping the set comparison and adding a separate check that `len(transformerResponse.Response)` equals its distinct jobID count.

router/transformer/transformer_test.go
```diff
@@ -701,8 +701,13 @@ type oauthv2ProxyTcs struct {
 	destination backendconfig.DestinationT
 
 	ioReadError bool
+	// expected `reason` tag on router.transformerproxy.invalid.response.
+	// Empty means the metric is not asserted for this case.
+	expectedBreachReason string
```
shekhar-rudder, line 706: **[medium]** `missing output` — the flagship new reason — has no test. The five `expectedBreachReason` cases in this file are `transport error` ×3, `unmarshal error` ×1, `in out mismatch` ×1.

The `{"output":null}` case that `transformer.go:551-554` devotes a five-line comment to justifying is untested too, so nothing pins the `output.Type == gjson.Null` half of the condition — dropping it would be a silent regression with a green suite.

Two cases would cover both: a 2xx whose body has no `output` field, and one with `"output":null`.

router/transformer/transformer.go
```diff
@@ -514,18 +547,58 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
...
+		if infraFailure {
+			reason, msg = "transport error", "[TransformerProxy] transport returned a non-transformer response"
+		}
+		trans.emitProxyInvalidResponse(proxyReqParams, reason, msg,
+			logger.NewIntField("statusCode", int64(httpPrxResp.statusCode)),
+			logger.NewIntSliceField("jobIDs", proxyJobIDs(proxyReqParams.ResponseData.Metadata)))
+		// There are no per-job results to apply without `output`. Continuing would set respData to
+		// an empty string and fail inside getResponse, reporting an empty body instead of the one
+		// we actually received. Stop here and return it.
+		return proxyErrorResponse(respCode, fmt.Sprintf("[TransformerProxy] response has no output field: %s", string(respData)), routerJobDontBatchDirectives)
```
shekhar-rudder: **[medium]** This line and `:528` embed the full upstream `respData` into `ProxyRequestResponseBody`, but nothing reads that field. `router/worker.go:531` consumes only `RespStatusCodes` / `RespBodys` / `RespContentType` / `DontBatchDirectives`:

```go
respStatusCodes, respBodyTemps, respContentType = resp.RespStatusCodes, resp.RespBodys, resp.RespContentType
```

and grep finds no other non-test reader of `ProxyRequestResponseBody`.

So the comments here ("reporting an empty body instead of the one we actually received", and "losing the very payload that explains the breach" at `:526-527`) describe a benefit no consumer observes, while the code stringifies a potentially multi-MB batch response on every breach.

Either drop the body from the message, or log it truncated the way `worker.go:493` already does with `misc.TruncateStr(..., 10*bytesize.KB)` — though that would cut against the PR's stated "the response body is deliberately not logged" position, so worth settling which one you want.

router/transformer/transformer_proxy_adapter.go
```diff
@@ -151,24 +146,10 @@ func (v1 *v1Adapter) getResponse(respData []byte, respCode int, metadata []Proxy
...
-	if !reflect.DeepEqual(jobIDsInMetadata, jobIDsInResponse) {
-		stats.Default.NewTaggedStat(`router.transformerproxy.invalid.response`, stats.CountType, stats.Tags{
-			"reason": "in out mismatch",
-		}).Increment()
-		v1.logger.Warnn("[TransformerProxy] JobIDs in out mismatch",
-			logger.NewIntSliceField("jobIDsInMetadata", jobIDsInMetadata),
-			logger.NewIntSliceField("jobIDsInResponse", jobIDsInResponse))
-	}
-
+	// routerJobResponseCodes is keyed by the jobIDs the transformer returned results for, so its keys
```
shekhar-rudder: **[nit]** With the `Warnn` removed, the `logger` field on `v0Adapter` / `v1Adapter` (lines 60 and 63) is now written and never read — `NewTransformerProxyAdapter` at line 181 still takes and stores it, and nothing in this file uses it.

Either drop the field and the parameter, or add a note on why it's being kept.

router/transformer/transformer.go
```diff
@@ -514,18 +547,58 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
...
-			RespStatusCodes:          routerJobResponseCodes,
-			RespBodys:                routerJobResponseBodys,
-			DontBatchDirectives:      routerJobDontBatchDirectives,
+		// output was present but its per-job payload will not parse. On a 2xx this is a genuine
+		// contract breach ("unmarshal error", like the outer envelope); an infra failure that happened
+		// to carry an output-shaped body is still infra.
+		reason, msg := "unmarshal error", "[TransformerProxy] proxy response output unmarshal failed"
+		if infraFailure {
+			reason, msg = "transport error", "[TransformerProxy] transport returned a non-transformer response"
 		}
```
shekhar-rudder, line 597: **[nit]** All three infra branches (here, `:519-520`, and `:559-560`) log the identical message `"[TransformerProxy] transport returned a non-transformer response"`, and the `reason` tag is identical too — so a log line can't tell you which of the three sites produced it, which is the one thing you'd want when triaging.

Distinct messages, or a `site` field on the log, would make these greppable.

shekhar-rudder (commented):
Reviewed against the source at `d638175`; built it and ran `go test ./router/transformer/...` (passes). Also traced `ProxyRequestResponseBody` consumers in `router/worker.go` and the proxy's status-code behaviour in `rudder-transformer`. Seven inline comments — two of them I'd treat as blocking, both about the new `transport error` reason.

router/transformer/transformer.go
saikumarrs: Went with the alert exclusion, and it has already landed: rudderlabs/rudderstack-operator#1452 now selects `reason!="transport error"` (`e6ef26e4`). Merge ordering is the gate - that PR merges before/with the server deploy, so the exclusion is never absent while the reason exists.

Kept the emit rather than dropping it. `transformer_proxy_request_result{requestSuccess}` tells you the request failed but not which of the eight synthesis sites did it or for which destination, and the triage dashboard filters on this metric by reason. Silencing it in the alert while keeping it on the dashboard gets both - a broken OAuth account shows up when you go looking, without paging.

router/transformer/transformer.go
saikumarrs: You are right, and this is the sharper version of a concern I had flagged myself but resolved the wrong way. I had already stopped short-circuiting on the status so a 5xx carrying a usable output is still processed, but that only protected delivery - the reason label was still wrong, and with the exclusion in place a genuine breach during a destination 5xx would have gone unpaged. That is the worst possible failure for this alert.

Dropped it: `infraFailure` is now `httpPrxResp.transportError` alone, and the upstream status is a log field. `a1752601`.

The consequence is the LB-outage cases go back to `missing output`, which is a false positive on a transformer outage. I would rather take that than a false negative on corruption, and it belongs on the alert side anyway - your page-storm comment on #1452 is the right place to fix it, by grouping the systemic reasons coarsely so one outage pages once instead of per destination.

router/transformer/transformer.go
saikumarrs: Agreed - I traded one for the other without noticing the second. Request-side duplicates are legitimate so the set comparison stays deduped, but response-side duplicates are the breach.

`v1Adapter.getResponse` now reports `responseEntries` (the raw entry count, which is not recoverable from the map once the keys collapse) and the caller flags `entries > keys` as `in out mismatch` alongside the set check. v0 reports `len(routerJobResponseCodes)` so it cannot trip. Test covers two entries for one jobID with conflicting statuses - the sets match there, so it fails without the count check. `a1752601`.

router/transformer/transformer_test.go
saikumarrs: Added both: a 2xx whose body has no `output` field, and one with `"output":null`. Needed a small `rawBody` hook in the mock to return a body that bypasses the `{"output": ...}` envelope.

Confirmed each pins its guard - dropping `output.Type == gjson.Null` fails the null case specifically, which was the gap you pointed at. `a1752601`.

router/transformer/transformer.go
saikumarrs: Confirmed - `grep` finds no non-test reader, and `router/worker.go:531` takes only the per-job maps plus content type.

Went with truncation over dropping it: bounded at 10KB, sliced on the byte slice before the string conversion so the multi-MB payload is never allocated. That keeps the field useful to anyone who does reach for it in a debugger without the cost you flagged. The "body deliberately not logged" position is about the log line - the jobIDs are what we log, so you look the events up in jobsDB - and that is unchanged. `a1752601`.

router/transformer/transformer_proxy_adapter.go
saikumarrs: Dropped the field and the parameter - `NewTransformerProxyAdapter(version string)` now, which also cleared the orphaned `mockLogger` in the adapter test. `a1752601`.

router/transformer/transformer.go
saikumarrs: Each site gets its own message now: `failed envelope unmarshal`, `has no output field`, and `output unmarshal failed`, all prefixed `[TransformerProxy] transport-synthesized response ...`. The upstream `statusCode` is on all three too, which is the other thing you would want when triaging. `a1752601`.

saikumarrs: Each site gets its own message now: `failed envelope unmarshal`, `has no output field`, and `output unmarshal failed`, all prefixed `[TransformerProxy] transport-synthesized response ...`. The upstream `statusCode` is on all three too, which is the other thing you would want when triaging. `a1752601`.

router/transformer/transformer.go
shekhar-rudder: Thanks — the exclusion plus the merge-ordering gate closes the four sites I listed. Those all go through `httpResponseCreator`, so they carry the header and `reason!="transport error"` covers them.

It doesn't close the case that opened up in the same round of commits, though. My comment was written against `infraFailure := httpPrxResp.transportError || httpPrxResp.statusCode >= http.StatusInternalServerError` — you can see it in the diff hunk quoted above. `601b7887` added that second half and `a1752601` removed it again, so head is now `infraFailure := httpPrxResp.transportError` (`transformer.go:524`).

That means a genuine upstream 5xx the transport did *not* synthesize — an ingress/LB 502 in front of the transformer, where `RoundTrip` succeeded and handed back a real response — no longer classifies as `transport error` at all:

- **oauth destination**: `getAuthErrorCategory` can't parse the body, so `postRoundTrip` returns the real `rts.res` through `applyInterceptorRespToHttpResp` (`transport.go:158-173`) with no header set. The envelope parses, `respData` is replaced by `OriginalResponse`, `output` is absent → `reason="missing output"`.
- **non-oauth destination**: `RoundTrip` short-circuits at `transport.go:256` and the body is the raw error page → `jsonrs.Unmarshal` fails → `reason="unmarshal error"`.

Neither reason is excluded by the selector, and the alert fires at `> 0`, so these page. The tests in this PR encode it: both `lbError` cases — a plain 502 from the test server — now expect `expectedBreachReason: "missing output"` (`transformer_test.go:1146`, `:1204`). Under the version I commented on those were `transport error` and would have been excluded.

To be clear, I don't think reverting `a1752601` is the answer — your reasoning for dropping the status half is right. For v0 the transformer mirrors the destination's delivery status, so `>= 500` means "the destination returned 5xx", which is routine and would also mask a real breach that happens to co-occur with one.

The signal you actually want is *"did this response come from the transformer at all"*, not *"is it 5xx"*. The `apiVersion` header is already exactly that: `httpResponseCreator` sets it on synthesized responses (`transport.go:74`), the real transformer sets it (we already check it on the transform path at `transformer.go:291-300`), and an ingress or LB error page carries neither. Gating `infraFailure` on "no/!supported `apiVersion` header" instead of on status would put the LB 502 back under `transport error` without relabelling a destination 5xx that carried a usable output.

One wrinkle if you go that way: `httpResponseCreator` builds `http.Header{"apiVersion": ...}` with a non-canonical key, so `Header.Get("apiVersion")` won't find it on a synthesized response (it canonicalises the lookup to `Apiversion`). It does work for real responses, since Go canonicalises headers read off the wire. Worth using `header.Set` there so both paths agree.

shekhar-rudder (commented):
Separate from the `transport error` classification thread (continued [there](https://github.com/rudderlabs/rudder-server/pull/7184#discussion_r3630540617)) — four other things from a pass over the head ref.

1. **[medium]** The infra signal is upstream-controlled. `transformer.go:838-842` reads `resp.Header.Get(oauthv2httpclient.TransportErrorHeader)` off responses that, on the non-oauth passthrough (`transport.go:256`) and the `postRoundTrip` real-response path (`transport.go:172`, `:210`), carry the upstream's headers verbatim. Anything in front of the transformer that sets `X-Rudder-Oauth-Transport-Error` silently suppresses the breach alert. Cheap fix: `res.Header.Del(TransportErrorHeader)` on the genuine-response paths, or carry the flag on a wrapper type / context value rather than a header that can arrive over the wire.

2. **[medium]** Undocumented v0 delivery-semantics change for `{"output":null}`. Before this PR, `respData` became `"null"`, `v0Adapter.getResponse` unmarshalled it without error, and the loop at `transformer_proxy_adapter.go:103-106` assigned `respCode` to **every** job — so a 200 carrying a null output marked the whole batch delivered. With the new guard at `transformer.go:570` it returns an error response with empty maps, and `hydrateRespStatusCodes` (`router/worker.go:815-823`) turns that into 500-for-every-job plus retry.

   That's almost certainly the correct behaviour, but it flips previously-succeeding deliveries into retries for any destination that legitimately answers `{"output":null}`, and it isn't called out in the description. The only `{"output":null}` case is `proxyVersion: "v1"` (`transformer_test.go:1730`) — worth adding the v0 counterpart and a line in the PR body.

3. **[medium]** Test coverage for `transport error` exercises one of the eight `httpResponseCreator` sites. `expectedBreachReason: "transport error"` appears once, on the `ioReadError` case (`transformer_test.go:1065`), which hits `transport.go:145`. The sites that matter most for keeping infra off the page — `FetchToken` failure (`transport.go:118`), augmentation failure (`:129`) and `Transport.RoundTrip` error (`:294`) — have no case pinning their reason. Given that the alert now depends entirely on that label being right, those are the ones worth pinning.

4. **[nit]** `truncatedBody` (`transformer.go:436-442`) slices at a byte offset, so a 10KB cut can land mid-rune and emit `U+FFFD` into the error string. `utf8.DecodeLastRune` trim-back, or a comment saying it's accepted.

router/transformer/transformer.go
saikumarrs: The apiVersion idea is the right one - provenance rather than status is exactly the question, and it lands the LB 502 back under `transport error` without touching the destination-5xx case. Adopted: `infraFailure := httpPrxResp.transportError || !httpPrxResp.fromTransformer`, where `fromTransformer` is the apiVersion header being present on the response. `2fb17b00`.

Two things worth recording from implementing it:

Gated on presence rather than "supported". A version we do not recognise is a real problem, and folding it into `transport error` would put it behind the alert exclusion - the one place it would never be seen. Present-but-unsupported therefore stays on the breach path.

The header is also invisible to `Get` on synthesized responses, which I had not expected: `httpResponseCreator` builds `http.Header{"apiVersion": ...}` as a raw map literal, so the key is never canonicalised and `Get(apiVersionHeader)` returns `""`. That means "no apiVersion" would have caught the synthesized responses too, by accident. I left `transportError` in the condition rather than leaning on that - it is explicit, and it survives someone changing the literal to `Set`.

The `lbError` cases go back to `transport error`. The test mock was setting apiVersion on them, which a real LB error page would not, so it now sets the header only once past that branch.

saikumarrs:
All four addressed in `2fb17b00` (the classification point is answered on [its thread](https://github.com/rudderlabs/rudder-server/pull/7184#discussion_r3630540617)).

**1. Infra signal is upstream-controlled**

Good catch - and it is the same hole I had flagged to myself about the header being a side channel, without following it to "upstream can set it".

`stripTransportErrorHeader` now clears the marker on both paths that hand back a response from the wire (`transport.go:256` passthrough and `:285` before `postRoundTrip`), so only this transport can assert it. Went with stripping over a wrapper type or context value because it keeps the assertion at the single point where a response stops being upstream-controlled.

Test pins it: an upstream that sets `X-Rudder-Oauth-Transport-Error` itself on a body with no output is still reported as `missing output`. It fails without the strip - I checked, because the fix had no coverage otherwise. `2fb17b00`.

**2. Undocumented v0 `{"output":null}` semantics change**

Right, and I had under-reported this. When I traced the delivery impact I checked the v1 path, where the maps are empty either way and `hydrateRespStatusCodes` produces the same 500-and-retry, and concluded the only difference was the recorded content type. v0 is not the same: its loop assigns `respCode` to every job off the request metadata, so a 200 carrying a null output previously marked the whole batch delivered.

Added the v0 counterpart case and called the change out in the PR description under a behaviour-change heading. Agreed it is the correct behaviour - a null output is not a delivery - but it should not have been silent. `2fb17b00`.

**3. `transport error` coverage**

Added the `Transport.RoundTrip` site (`transport.go:294`) - a case that closes the mock transformer before the request, so the transport synthesizes the response and the reason is pinned to `transport error`. `2fb17b00`.

Not covering `FetchToken` (`:118`) and augmentation (`:129`) yet: both need the oauth scaffolding driven into a failure state mid-flight, and the harness currently only feeds control-plane responses for the refresh path. Worth doing, but it is enough scaffolding that I would rather not fold it into this PR - happy to take it as a follow-up unless you want it here.

**4. `truncatedBody` mid-rune cut**

Fixed rather than documented - `truncatedBody` now trims back off a trailing partial rune with `utf8.DecodeLastRune` before converting. `2fb17b00`.

services/oauth/v2/http/transport.go
```diff
@@ -65,11 +65,27 @@ type roundTripState struct {
...
+// silently suppress the breach alert.
+func stripTransportErrorHeader(res *http.Response) {
+	if res != nil && res.Header != nil {
+		res.Header.Del(TransportErrorHeader)
+	}
+}
+
 func httpResponseCreator(statusCode int, body []byte) *http.Response {
+	header := http.Header{"apiVersion": []string{"2"}}
+	header.Set(TransportErrorHeader, "true")
```
shekhar-rudder: **[complexity]** The whole `TransportErrorHeader` + `stripTransportErrorHeader` machinery exists only because synthesized error responses also stamp `apiVersion` here (line 83), so `fromTransformer` (presence of `apiVersion`) can't tell a transport-synthesized 500 apart from a real transformer response. If synthetic responses didn't carry `apiVersion` — or carried a sentinel version — then `!fromTransformer` alone would classify every infra case, and this header, `stripTransportErrorHeader`, its two call sites, the spoof threat model, and one test could all go away.

I assume `apiVersion` is stamped to keep the downstream `strconv.Atoi(resp.Header.Get(apiVersionHeader))` parse (`router/transformer/transformer.go:288`) happy. Can we fix that root cause instead of adding a second marker to work around it? Not blocking — but this is the bulk of the added complexity, and it's accidental rather than intrinsic to the problem.

services/oauth/v2/http/transport.go
```diff
@@ -65,11 +65,27 @@ type roundTripState struct {
...
 
+// TransportErrorHeader marks a response synthesized by this transport rather than received from the
+// upstream service. The body of such a response is a plain-text error string, not the upstream's
+// payload, so callers must not mistake it for a malformed upstream response.
+const TransportErrorHeader = "X-Rudder-Oauth-Transport-Error"
+
+// stripTransportErrorHeader removes the marker from a response that arrived over the wire, so only
+// this transport can assert it. Otherwise anything in front of the transformer could set it and
+// silently suppress the breach alert.
+func stripTransportErrorHeader(res *http.Response) {
```
shekhar-rudder: **[nit]** The strip defends against something in front of the transformer forging `X-Rudder-Oauth-Transport-Error` to silence the breach alert — but who is that adversary in practice? The transformer is trusted infra, an ALB won't emit this header by accident, and destination responses don't pass their headers through as proxy-response headers. It's cheap (one `Del`) and a silenced page is expensive, so it's a defensible hedge, but it reads as belt-and-suspenders guarding an already-trusted boundary. Fine to keep; flagging in case it's guarding a non-threat.

router/transformer/transformer.go
```diff
@@ -400,23 +404,69 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
...
+// truncatedBody bounds a body embedded in an error string: nothing reads ProxyRequestResponseBody
+// and a batch response can be multi-MB.
+func truncatedBody(b []byte) string {
+	const maxBodyBytes = int(10 * bytesize.KB)
+	if len(b) <= maxBodyBytes {
+		return string(b)
+	}
+	cut := b[:maxBodyBytes]
+	// Trim back off a partial rune so the cut does not emit U+FFFD.
+	if r, size := utf8.DecodeLastRune(cut); r == utf8.RuneError && size <= 1 {
```
shekhar-rudder: **[nit]** This partial-rune trim doesn't achieve its stated goal. `utf8.DecodeLastRune` returns `(RuneError, 1)` for an incomplete trailing sequence, so only **one** byte is dropped — a 3–4 byte rune split by the 10KB cut still leaves 1–2 lead/continuation bytes that render as U+FFFD when `string()`-ed. Since this only bounds a truncated error/log string, I'd either loop the trim, trim back to the last rune start, or just drop the rune handling entirely and do a plain byte cut — the current code adds complexity for a cosmetic goal it doesn't meet.

router/transformer/transformer_proxy_adapter.go
```diff
@@ -52,6 +46,9 @@ type TransResponse struct {
 	routerJobDontBatchDirectives map[int64]bool
 	authErrorCategory            string
 	statTags                     map[string]string
+	// responseEntries is the raw per-job entry count. Not derivable from the maps above: two entries
+	// for one jobID collapse to a single key.
+	responseEntries int
```
shekhar-rudder: **[nit]** `responseEntries` is an `int` but the name reads like a slice of entries. A `Count` suffix (`responseEntriesCount`) makes the type and intent obvious — it's the *raw* entry count, deliberately distinct from the deduped map-key count that the duplicate check compares it against. Scope is small: this field, the two assignments below (`len(routerJobResponseCodes)` v0 / `len(transformerResponse.Response)` v1), and `transResp.responseEntries` in `transformer.go`. I'd leave the `logger.NewIntField("responseEntries", ...)` log key string as-is though — that's operator/runbook-facing, so rename only the Go identifier.

services/oauth/v2/http/transport.go
saikumarrs: Adopted, and you are right that it was accidental rather than intrinsic. Two things I found tracing it that make the fix cleaner than a straight swap:

`apiVersion` was never actually keeping any parse happy - the readers all gate on `StatusCode == 200` (`transformer.go:290`, and the processor-side readers do the same) and every synthesized response is non-2xx, so its `apiVersion` was never read. And it was invisible anyway: `httpResponseCreator` set it as a raw map literal (`http.Header{"apiVersion": ...}`), whose key is non-canonical, so `Get("apiVersion")` already returned `""` on synthesized responses. `!fromTransformer` was already doing the whole job; the marker was dead weight sitting on top of it.

So I made the absence intentional instead of accidental: `httpResponseCreator` no longer sets `apiVersion`, and `infraFailure := !httpPrxResp.fromTransformer`. That drops the header const, `stripTransportErrorHeader` and its two sites, the `transportError` field, and the spoof test - about 75 lines net. `c9afa0e1`.

Behaviour is unchanged - the LB-outage and unreachable-transformer cases still resolve to `transport error`, now via apiVersion absence, and both fail if the check is dropped. This is also what I should have done last round instead of keeping `transportError` as a hedge against the literal being canonicalised; removing the literal removes that risk outright.

services/oauth/v2/http/transport.go
saikumarrs: Gone - the strip went with the rest of the marker machinery in `c9afa0e1`, so there is no longer a header to forge.

You had the threat model right, and the new shape settles it in the safe direction rather than by guarding: infra is now "no apiVersion", so a forged header could only *add* apiVersion, which makes a genuine breach page (a false page) rather than silence one. A silenced page was the expensive failure; that direction is now impossible, so there is nothing left to defend.

router/transformer/transformer.go
saikumarrs: Right on both counts - `DecodeLastRune` returns `(RuneError, 1)` for an incomplete tail, so it only ever dropped one byte and left the rest to render as U+FFFD.

Took the "drop it" option: plain byte cut. Nothing reads `ProxyRequestResponseBody`, so a mid-rune cut in a truncated string is not worth the code. `c9afa0e1`.

router/transformer/transformer_proxy_adapter.go
saikumarrs: Renamed to `responseEntriesCount` - the field and its two assignments and the read in `transformer.go`. Left the `logger.NewIntField("responseEntries", ...)` key as-is since that one is runbook-facing. `c9afa0e1`.

shekhar-rudder (approved):

services/oauth/v2/http/transport.go
```diff
@@ -65,11 +65,14 @@ type roundTripState struct {
 	req         *http.Request
 }
 
+// httpResponseCreator builds a response this transport synthesizes on an error path, rather than one
+// received from the upstream service. It carries no apiVersion header - only the transformer sets that
+// - so callers can tell a synthesized response apart from a real one by its absence.
```
atzoum: The pr description mentions something about `X-Rudder-Oauth-Transport-Error` & `httpProxyResponse.transportError`. Why have we diverged? 
We should update the pr description and probably extend the doc here, e.g. to mention that we choose not to include the header so that this covers other non-transformer error responses too, such as intermediate reverse proxies. Btw, an unhandled exception at the transformer will still carry no header right?

router/transformer/transformer.go
```diff
@@ -477,23 +508,32 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
...
 	if requestError != nil {
-		return ProxyRequestResponse{
-			ProxyRequestStatusCode:   respCode,
-			ProxyRequestResponseBody: requestError.Error(),
-			RespContentType:          "text/plain; charset=utf-8",
-			RespStatusCodes:          routerJobResponseCodes,
-			RespBodys:                routerJobResponseBodys,
-			DontBatchDirectives:      routerJobDontBatchDirectives,
-		}
+		return proxyErrorResponse(respCode, requestError.Error(), routerJobDontBatchDirectives)
```
atzoum, line 525: Is the `transport error` reason meant to cover non-OAuth destinations too? As far as I can tell it can't reach them.

`OAuthTransport.RoundTrip` returns `t.Transport.RoundTrip(req)` directly for non-OAuth destinations (transport.go:251-252), so a transformer-unreachable/LB failure comes back as a raw Go error rather than a synthesized response. That lands in `requestError` here and returns before infraFailure is computed, so `emitProxyInvalidResponse` never runs.

router/transformer/transformer.go
```diff
@@ -400,23 +403,65 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
 	return destinationJobs
 }
 
+// proxyJobIDs returns the jobIDs in a proxy request batch — what a responder needs to look the
+// affected events up in jobsDB.
+func proxyJobIDs(metadata []ProxyRequestMetadata) []int64 {
```
atzoum: this could become a method of `ProxyRequestPayload`

```go
func (ppr *ProxyRequestPayload) JobIDs() []int64 {
	return lo.Map(ppr.Metadata, func(m ProxyRequestMetadata, _ int) int64 {
		return m.JobID
	})
}
```

router/transformer/transformer.go
```diff
@@ -400,23 +403,65 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
...
+// to inspect the affected events. Every breach reason goes through here so the metric keeps a single
+// tag set and a single stats handle. The response body is deliberately not logged: these are
+// debugged by reading the events out of jobsDB by jobID.
+//
+// Takes proxyReqParams rather than the destType/destinationId strings so the two cannot be
+// transposed at a call site and silently mis-tag the metric. Callers must be past ProxyRequest's
+// empty-metadata guard, which every breach site is.
+func (trans *handle) emitProxyInvalidResponse(proxyReqParams *ProxyRequestParams, reason, msg string, fields ...logger.Field) {
+	destinationID := proxyReqParams.ResponseData.Metadata[0].DestinationID
+	trans.stats.NewTaggedStat(`router.transformerproxy.invalid.response`, stats.CountType, stats.Tags{
```
atzoum: What's the point of using dots if they are to be translated to underscores?
```suggestion
	trans.stats.NewTaggedStat(`router_transformerproxy_invalid_response`, stats.CountType, stats.Tags{
```

router/transformer/transformer.go
```diff
@@ -400,23 +403,65 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
...
+		"reason":        reason,
+		"destType":      proxyReqParams.DestName,
+		"destinationId": destinationID,
+	}).Increment()
+	trans.logger.Warnn(msg, append([]logger.Field{obskit.DestinationID(destinationID)}, fields...)...)
```
atzoum: Do we need both a metric and a warning log? High frequency logs should be avoided if possible

router/transformer/transformer.go
```diff
@@ -400,23 +403,65 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
...
+}
+
+// truncatedBody bounds a body embedded in an error string: nothing reads ProxyRequestResponseBody
+// and a batch response can be multi-MB.
+func truncatedBody(b []byte) string {
```
atzoum, line 427: We may use `lo.Ellipsis` instead. IMO 10K characters are too many, we shouldn't do more than 1K.

router/transformer/transformer.go
```diff
@@ -477,23 +508,32 @@ func (trans *handle) ProxyRequest(ctx context.Context, proxyReqParams *ProxyRequ
...
 		respData will be in ProxyResponseV0 or ProxyResponseV1
 	*/
 	var transportResponse oauthv2.TransportResponse // response that we get from oauth-interceptor in postRoundTrip
-	_ = jsonrs.Unmarshal(respData, &transportResponse)
-	// unmarshal unsuccessful scenarios
-	// if respData is not a valid json
+	// A non-oauth destination hands back the raw upstream body, so a non-JSON body here is either infra
+	// or a corrupt transformer response.
+	if err := jsonrs.Unmarshal(respData, &transportResponse); err != nil {
+		reason, msg := "unmarshal error", "[TransformerProxy] proxy response unmarshal failed"
```
atzoum: Let's use constants for `unmarshal error`, `transport error`, `missing output`

router/transformer/transformer.go
```diff
@@ -400,23 +403,65 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
...
+// debugged by reading the events out of jobsDB by jobID.
+//
+// Takes proxyReqParams rather than the destType/destinationId strings so the two cannot be
+// transposed at a call site and silently mis-tag the metric. Callers must be past ProxyRequest's
+// empty-metadata guard, which every breach site is.
+func (trans *handle) emitProxyInvalidResponse(proxyReqParams *ProxyRequestParams, reason, msg string, fields ...logger.Field) {
+	destinationID := proxyReqParams.ResponseData.Metadata[0].DestinationID
+	trans.stats.NewTaggedStat(`router.transformerproxy.invalid.response`, stats.CountType, stats.Tags{
+		"reason":        reason,
+		"destType":      proxyReqParams.DestName,
```
atzoum: Can we rename the field `ProxyRequestParams.DestName` to `ProxyRequestParams.DestType`. This is how we call it in rudder-server

router/transformer/transformer.go
```diff
@@ -400,23 +403,65 @@ func (trans *handle) Transform(transformType string, transformMessage *types.Tra
...
+}
+
+// emitProxyInvalidResponse records a transformer-proxy contract breach and logs the jobIDs needed
+// to inspect the affected events. Every breach reason goes through here so the metric keeps a single
+// tag set and a single stats handle. The response body is deliberately not logged: these are
+// debugged by reading the events out of jobsDB by jobID.
+//
+// Takes proxyReqParams rather than the destType/destinationId strings so the two cannot be
+// transposed at a call site and silently mis-tag the metric. Callers must be past ProxyRequest's
+// empty-metadata guard, which every breach site is.
```
atzoum: I would rather we declared a function inside the `handle.ProxyRequest` method, at line 502, after the http request has been performed. This way we can guarantee structurally that this is past ProxyRequest's empty-metadata guard and we no longer need to pass any proxyReqParams parameter at all, so not have to worry about silently mis-tagging metrics...

```go
emitBreach := func(reason, msg string, fields ...logger.Field) {
	destinationID := proxyReqParams.ResponseData.Metadata[0].DestinationID
	trans.stats.NewTaggedStat(`router.transformerproxy.invalid.response`, stats.CountType, stats.Tags{
		"reason":        reason,
		"destType":      proxyReqParams.DestName,
		"destinationId": destinationID,
	}).Increment()
	trans.logger.Warnn(msg, append([]logger.Field{obskit.DestinationID(destinationID)}, fields...)...)
}
```

services/oauth/v2/http/transport.go
saikumarrs: Diverged because the header turned out to be redundant - `httpResponseCreator` set `apiVersion` as a raw map literal, so `Get("apiVersion")` already returned "" on synthesized responses, and `!fromTransformer` was doing the whole job. Dropped the header and the field; classification is `!httpPrxResp.fromTransformer` alone.

To your question: yes, an unhandled transformer exception carries no header. Verified it in rudder-transformer - `ControllerUtility.deliveryPostProcess` is what stamps `apiVersion`, and the delivery controller calls it on every handled outcome (success, destination error, or a caught delivery failure), but an unhandled exception is caught by the global `errorHandlerMiddleware` which sets status and body without going through `deliveryPostProcess`. So it lands as `transport error` - which is right, it is a transformer crash rather than a per-destination contract breach, and it is covered by the transformer-availability alerts instead.

Extended the doc on `httpResponseCreator` to say the absence covers non-transformer errors generally (ingress/reverse proxy, unhandled exceptions), not just this transport's own. Updating the PR description too. `e132612b`.

router/transformer/transformer.go
saikumarrs: Good catch, and fixed. You are right that a raw transport failure lands in `requestError` and returned before emitting - for non-oauth destinations that is the only path (the oauth path synthesizes a response, so those already reached the classifier). Now the `requestError` branch emits `transport error` too, so both show on the metric. It stays excluded from the alert, so this pages nothing - it is metric symmetry, not a new page.

Note the LB-with-a-body case was already covered for non-oauth (a 502 with a body is a real HTTP response, not a Go error, so it reaches `!fromTransformer`); this closes the remaining raw-error gap. Covered by a 404 case that hits the request-error path and fails if the emit is dropped. `e132612b`.

router/transformer/transformer.go
saikumarrs: Done - `func (p *ProxyRequestPayload) JobIDs() []int64`, and the call sites read `proxyReqParams.ResponseData.JobIDs()`. `e132612b`.

router/transformer/transformer.go
saikumarrs: Switched to underscores - `router_transformerproxy_invalid_response`, as a constant. It is what every other stat in this file already uses, and dots vs underscores map to the same Prometheus series, so the operator alert is unaffected. `e132612b`.

router/transformer/transformer.go
saikumarrs: Keeping both, deliberately. These are breach paths, not steady-state - the metric has emitted nothing in prod over 90d, so this is not a high-frequency logger. And the log is load-bearing on its own: it carries the jobIDs, which is the whole triage story - the runbook pulls those and reads the affected events out of jobsDB before they age out. The metric drives the alert; the log is how you act on it. Neither replaces the other.

router/transformer/transformer.go
saikumarrs: Dropped to 1KB. Kept the byte-slice rather than `lo.Ellipsis` on purpose: `lo.Ellipsis` takes a string, so it would allocate the whole body first, and this body can be a multi-MB batch response - the slice truncates before the conversion. `e132612b`.

router/transformer/transformer.go
saikumarrs: Done - `reasonTransportError`, `reasonUnmarshalError`, `reasonMissingOutput`, `reasonInOutMismatch`, alongside the metric-name constant. `e132612b`.

router/transformer/transformer.go
saikumarrs: Renamed `ProxyRequestParams.DestName` to `DestType` and updated the construction in `worker.go` and the tests. `e132612b`.

router/transformer/transformer.go
saikumarrs: Agreed, and did exactly that - `emitBreach` is now a closure declared inside `ProxyRequest` after the http request, so it is structurally past the empty-metadata guard (`Metadata[0]` is always safe) and captures `proxyReqParams`, so there is nothing to thread and nothing to transpose. Also let me reuse it on the request-error path for the symmetry fix above. `e132612b`.

atzoum (approved):

PR #7186, opened 2026-07-19
fix(warehouse): close load file descriptor in identity uploadFile

### LINKED ISSUE #7185: bug(warehouse/identity): file descriptor leak in uploadFile() — outputFile never closed
## Bug Description

In `warehouse/identity/identity.go`, the `uploadFile()` method opens the load file for uploading to object storage but **never closes the file descriptor** — on any path (filemanager creation error, upload error, or success).

This is the same defect class as #7071 (batchrouter `upload()`), but in the identity-resolution flow, which was not covered by that report.

## Location

`warehouse/identity/identity.go` — function `uploadFile()`, line 455

## Code

```go
func (idr *Identity) uploadFile(ctx context.Context, filePath string, txn *sqlmiddleware.Tx, tableName string, totalRecords int) (err error) {
	outputFile, err := os.Open(filePath)
	if err != nil {
		panic(err)
	}
	// ❌ no defer outputFile.Close() anywhere in the function

	storageProvider := warehouseutils.ObjectStorageType(...)
	uploader, err := filemanager.New(&filemanager.Settings{...})
	if err != nil {
		return err                                  // fd leaked
	}
	output, err := uploader.Upload(ctx, outputFile, ...)
	if err != nil {
		return err                                  // fd leaked
	}
	// ...
	return err                                      // fd also leaked on success
}
```

## Impact

`uploadFile()` is called twice per identity-resolution sync — once for merge rules (`processMergeRules`, line 570) and once for identity mappings (line 581). Every warehouse sync with identity resolution enabled leaks 2 OS file descriptors for the lifetime of the process. Under sustained syncing this contributes to fd exhaustion (`too many open files`), the same failure mode described in #7071.

## Secondary issue in the same function

A failed `os.Open` triggers `panic(err)`, crashing the whole server, while every other error path in this function returns an error gracefully. A missing/unreadable temp file should surface as a sync failure, not a process crash.

## Expected Behaviour

The file descriptor should be closed once the upload completes (success or failure), and open failure should return an error instead of panicking.

## Suggested Fix

```go
outputFile, err := os.Open(filePath)
if err != nil {
	return fmt.Errorf("opening load file %s: %w", filePath, err)
}
defer func() { _ = outputFile.Close() }()
```

This matches the pattern already used in the same package at `identity.go:162` (`tableRows.Close()`) and in `warehouse/validations/validate.go:405-411`.

## Environment

- File: `warehouse/identity/identity.go`
- Function: `uploadFile()` (lines 454–494)
- Affected: all warehouse destinations with identity resolution enabled

andyosyndoh:
# Description

`uploadFile()` in `warehouse/identity/identity.go` opened the load file for object-storage upload but never closed the file descriptor — on the filemanager-creation error path, the upload error path, or on success. Since `uploadFile()` runs twice per identity-resolution sync (merge rules + identity mappings), every sync leaked two file descriptors for the lifetime of the process, contributing to fd exhaustion (`too many open files`) under sustained syncing.

This PR:
- adds `defer func() { _ = outputFile.Close() }()` immediately after the `os.Open`, matching the pattern used elsewhere in the package (`identity.go` `tableRows.Close()`) and in `warehouse/validations/validate.go`
- replaces the `panic(err)` on open failure with a wrapped error return, consistent with every other error path in the function

Same defect class as #7071 (batchrouter), but in the identity-resolution flow.

Resolves #7185

## Linear Ticket

N/A — community contribution, tracked in GitHub issue #7185

## Security


Aman-Chaturvedi-x78:
diff --git a/warehouse/identity/identity.go b/warehouse/identity/identity.go
index 83e523d..2867a5d 100644
--- a/warehouse/identity/identity.go
+++ b/warehouse/identity/identity.go
@@ -454,8 +454,10 @@ func (idr *Identity) writeTableToFile(tableName string, txn *sqlmiddleware.Tx, g
 func (idr *Identity) uploadFile(ctx context.Context, filePath string, txn *sqlmiddleware.Tx, tableName string, totalRecords int) (err error) {
 	outputFile, err := os.Open(filePath)
 	if err != nil {
-		panic(err)
+		return fmt.Errorf("opening load file %s: %w", filePath, err)
 	}
+	defer func() { _ = outputFile.Close() }()
+
 	storageProvider := warehouseutils.ObjectStorageType(idr.warehouse.Destination.DestinationDefinition.Name, idr.warehouse.Destination.Config, idr.uploader.UseRudderStorage())
 	uploader, err := filemanager.New(&filemanager.Settings{
 		Provider: storageProvider,

achettyiitr (approved):

warehouse/identity/identity.go
```diff
@@ -454,8 +454,9 @@ func (idr *Identity) writeTableToFile(tableName string, txn *sqlmiddleware.Tx, g
 func (idr *Identity) uploadFile(ctx context.Context, filePath string, txn *sqlmiddleware.Tx, tableName string, totalRecords int) (err error) {
 	outputFile, err := os.Open(filePath)
 	if err != nil {
-		panic(err)
+		return fmt.Errorf("opening load file %s: %w", filePath, err)
```
krishna2020, line 457: As the error is getting gracefully handled now, right metrics need to be added for alerting purpose. Otherwise it introduces a silent failure mode where file open errors may be missed.

krishna2020: cc: @achettyiitr

achettyiitr: In case of error, we mark the syncs as failed. What kind of metrics are we expecting?

krishna2020: regular failure/error metrics to trigger an alert if threshold crossed

achettyiitr: Recently, we added golden alerts, so they should be fine.

krishna2020 (approved):

achettyiitr:
Everything looks good, merging then. https://github.com/rudderlabs/rudder-server/pull/7279

PR #7202, opened 2026-07-21
feat(processor): isolated destination processing pipeline

atzoum:
# Description

Introduces the **proc pool**: a second processor worker pool that consumes a new intermediate **`proc` jobsdb** and runs the post-fan-out stages — `rebuild → user transform → destination transform → store` — isolated **per destination**, so a slow/failing destination no longer blocks the others. It reuses the existing processor stages as-is; only the `rebuild` entry stage is new.

This PR is the **consumer side only**. Nothing writes to the `proc` jobsdb yet — the gw-pool siphon that fans out and persists jobs into it is out of scope and will follow up in a separate pull request. With `Processor.DestinationIsolation.enabled=false` (default) the pool is a no-op.

**Multi-consumer jobsdb.** The `proc` jobsdb is multi-consumer: one job carries multiple consumer IDs (destination IDs), and each `(job, consumer)` pair has its own independent status lifecycle. The proc pool:
- discovers partitions via `GetDistinctConsumers` (one partition = one destination),
- picks up jobs scoped to a consumer (`GetUnprocessed` with `Consumer`), and
- commits per-consumer statuses through the shared `storeStage`, redirected to write against `procDB` instead of `gatewayDB`.

**Rebuild assumptions.** Proc jobs persist only IDs + the event/metadata; at consume time `rebuild` re-hydrates `Destination`/`Connection`/`Libraries`/`Credentials` from the **live** backend config. Therefore:
- Destination filtering and consent are **not** re-applied — they were already decided in the gw pool, and a job is only stored for the destinations that passed them.
- A destination deleted/disabled between fan-out and consume is dropped gracefully to a terminal status. Drop is silent, i.e. it is not being reported, similar to what is happening already during the pretransform stage when we fan out events.

## Linear Ticket

resolves PIPE-3048
resolves PIPE-3054

## Security

---
- <kbd>&nbsp;2&nbsp;</kbd> #7207 
- <kbd>&nbsp;1&nbsp;</kbd> #7202 👈

### DIFF
--- a/app/apphandlers/embeddedAppHandler.go
+++ b/app/apphandlers/embeddedAppHandler.go
@@ -289,6 +289,7 @@ func (a *embeddedApp) StartRudderCore(ctx context.Context, shutdownFn func(), op
 	if config.GetBoolVar(false, "Processor.DestinationIsolation.enabled") {
 		procRWHandle := jobsdb.NewForReadWrite(
 			"proc",
+			jobsdb.WithMultiConsumer(),
 			jobsdb.WithClearDB(options.ClearDB),
 			jobsdb.WithDSLimit(a.config.procDSLimit),
 			jobsdb.WithSkipMaintenanceErr(config.GetBoolVar(false, "Processor.jobsDB.skipMaintenanceError")),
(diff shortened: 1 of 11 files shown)


ktgowtham (approved):

mihir20 (approved):
LGTM

PR #7212, opened 2026-07-24
fix: mid-run pod deaths under concurrent test runs

mihir20:
# Description

Fixes a mid-run pod-death bug in concurrent ephemeral pyt test runs, adds per-operation observability to the processor's CP-facing `Forward` RPC (Python transformation tests), and makes the ephemeral pyt
  deployer's timing knobs configurable.

  ### Bug fix: mid-run pod deaths under concurrent test runs

  Concurrent ephemeral test runs for the same workspace carried identical label sets, so every run's Service selected **every** run's pods. A request forwarded through run A's Service could land on run B's
  pod — which run B deletes the moment it finishes, killing A's request mid-flight ("Subprocess died (exit code -15)").

  Each run's Deployment selector, pod template labels, and Service selector now also carry a per-run name label (`rudderstack.com/run` = the unique `pyt-test-<suffix>` name), making concurrent runs
  label-disjoint: each Service selects only its own run's pods. Covered by a test asserting two same-workspace runs produce Services that select their own pods and not each other's.

  ### New metrics

  | Metric | Type | Tags | What it measures |
  |---|---|---|---|
  | `processor_pyt_forward_rpc_time` | timer | `op`, `route`, `workspaceId` | The whole `Forward` RPC: deployment setup + readiness wait + request |
  | `processor_pyt_request_handle_time` | timer | `op`, `route`, `workspaceId` | Just the HTTP request to the pyt pod |
  | `processor_pyt_readiness_wait_time` | timer | `workspaceId`, `success` | Time spent waiting for the ephemeral Deployment to report ready |
  | `processor_pyt_cleanup_failure_count` | count | `workspaceId` | Best-effort delete failures (work leaked to the reaper) |

  - `route` distinguishes the three forwarding paths: `ephemeral` (fresh pyt deployment), `production` (config-flagged workspaces), `static` (AST ops).
  - On the ephemeral route, `forward_rpc_time − request_handle_time` tells you how much of a slow test was setup vs. actual execution; the readiness-wait timer breaks the setup portion down further.
  `success:false` readiness samples cluster at the configured timeout, making timeouts visible.
  - `processor_pyt_request_handle_time` is only emitted when a request was actually made — a deployer failure before forwarding records no sample. Rejected requests (invalid op/workspaceId) emit nothing.
  - Histogram buckets for all three timers are registered in `runner/buckets.go`.
  - The cold-start / retry counters in `user_transformer.forwardTest` now carry a `path` tag so retries can be attributed per pyt endpoint.

  ### New config keys (`Processor.pytDeployer.*`)

  | Key | Default | Notes |
  |---|---|---|
  | `pytTestReadinessPollInterval` | `250ms` | Reloadable; was a hardcoded constant |
  | `pytTestDeleteTimeout` | `15s` | Reloadable; bounds best-effort cleanup, was a hardcoded constant |
  | `pytTestRunIDLength` | `8` | Random suffix length for ephemeral resource names; clamped to keep `pyt-test-<suffix>` a valid DNS-1035 label |

  ### Implementation notes

  - `pytdeployer.New` now takes `stats.Stats`; `cpservice.Service` retains the stats handle it already received.
  - The three routes in `Forward` share a single `timedForward` closure that emits the request timer itself, so no forward ⇒ no sample, with no extra bookkeeping.

## Linear Ticket

-

## Security

### DIFF
--- a/processor/cpservice/forward.go
+++ b/processor/cpservice/forward.go
@@ -125,6 +125,13 @@ func (s *Service) Forward(ctx context.Context, req *proto.ForwardRequest) (*prot
 	default:
 		statusCode, body, err = timedForward(ctx, s.staticASTURL)
 	}
+	// success here means the RPC delivered a pyt response, whatever its HTTP
+	// status — false marks infrastructure failures (deployer create errors,
+	// readiness timeouts, transport failures), which are the only way to tell
+	// a failed RPC apart from a slow success. On a deployer create failure
+	// this is the ONLY metric emitted: the readiness and request timers never
+	// get a sample because those phases never ran.
+	tags["success"] = strconv.FormatBool(err == nil)
 	s.stat.NewTaggedStat("processor_pyt_forward_rpc_time", stats.TimerType, tags).Since(start)
 	if err != nil {
 		s.log.Warnn("forwarding to pyt",
--- a/processor/cpservice/forward.go
+++ b/processor/cpservice/forward.go
@@ -13,6 +13,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/rudderlabs/rudder-go-kit/logger"
+	"github.com/rudderlabs/rudder-go-kit/stats"
 	obskit "github.com/rudderlabs/rudder-observability-kit/go/labels"
 
 	"github.com/rudderlabs/rudder-server/processor/internal/transformer/user_transformer"
@@ -70,6 +71,7 @@ func (s *Service) Forward(ctx context.Context, req *proto.ForwardRequest) (*prot
 	var statusCode int
 	var body []byte
 	var err error
+	var route string
 	switch {
 	case isExecutionOp && s.isProdRouted(req.WorkspaceId):
 		// Config-flagged workspace: its tests must run with prod-only config
@@ -77,17 +79,25 @@ func (s *Service) Forward(ctx context.Context, req *proto.ForwardRequest) (*prot
 		// reproduce, so forward straight to the workspace's production pyt
 		// deployment. No readiness wait — the forwarder's cold-start retries
 		// ride out the prod deployment's scale-from-zero window.
+		route = "production"
 		statusCode, body, err = forward(ctx,
 			user_transformer.PerWorkspacePyTBaseURL(s.prodURLTemplate, req.WorkspaceId),
 			req.WorkspaceId, req.Payload)
 	case isExecutionOp:
+		route = "ephemeral"
 		statusCode, body, err = s.deployer.RunOnEphemeral(ctx, req.WorkspaceId,
 			func(ctx context.Context, baseURL string) (int, []byte, error) {
 				return forward(ctx, baseURL, req.WorkspaceId, req.Payload)
 			})
 	default:
+		route = "static"
 		statusCode, body, err = forward(ctx, s.staticASTURL, req.WorkspaceId, req.Payload)
 	}
+	s.stat.NewTaggedStat("processor_pyt_forward_time", stats.TimerType, stats.Tags{
+		"op":          req.Op.String(),
+		"route":       route,
+		"workspaceId": req.WorkspaceId,
+	}).Since(start)
 	if err != nil {
 		s.log.Warnn("forwarding to pyt",
 			obskit.WorkspaceID(req.WorkspaceId), logger.NewStringField("op", req.Op.String()), obskit.Error(err))
--- a/processor/cpservice/forward.go
+++ b/processor/cpservice/forward.go
@@ -68,36 +68,49 @@ func (s *Service) Forward(ctx context.Context, req *proto.ForwardRequest) (*prot
 		return nil, status.Error(codes.InvalidArgument, "workspaceId must be 1-59 alphanumeric or hyphen characters, starting and ending with an alphanumeric")
 	}
 
+	route := "static"
+	if isExecutionOp {
+		if s.isProdRouted(req.WorkspaceId) {
+			route = "production"
+		} else {
+			route = "ephemeral"
+		}
+	}
+	tags := stats.Tags{
+		"op":          req.Op.String(),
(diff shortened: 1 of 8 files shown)


processor/cpservice/forward.go
```diff
@@ -67,27 +68,49 @@ func (s *Service) Forward(ctx context.Context, req *proto.ForwardRequest) (*prot
 		return nil, status.Error(codes.InvalidArgument, "workspaceId must be 1-59 alphanumeric or hyphen characters, starting and ending with an alphanumeric")
 	}
 
+	route := "static"
+	if isExecutionOp {
+		if s.isProdRouted(req.WorkspaceId) {
+			route = "production"
```
fracasula: Can we use constants for these?

processor/internal/pytdeployer/k8s.go
```diff
@@ -267,7 +279,7 @@ func (d *k8sDeployer) waitReady(ctx context.Context, name string) error {
 	ctx, cancel := context.WithTimeout(ctx, d.config.readinessTimeout.Load())
 	defer cancel()
 
-	ticker := time.NewTicker(readinessPollInterval)
+	ticker := time.NewTicker(d.config.readinessPollInterval.Load())
```
fracasula: If I'm not mistaken tickers panic with anything <= 0, bringing down the whole processor. Beware of feeding reloadable vars directly here.

processor/cpservice/forward.go
```diff
@@ -67,27 +68,49 @@ func (s *Service) Forward(ctx context.Context, req *proto.ForwardRequest) (*prot
...
+	}
+	// timedForward measures just the HTTP request to the pyt pod and emits the
+	// request timer itself, so on the ephemeral route it excludes the
+	// deployment creation and readiness wait (which RunOnEphemeral performs
+	// before invoking it) that the total timer includes — and a deployer
+	// failure that never invokes it produces no request sample.
+	timedForward := func(ctx context.Context, baseURL string) (int, []byte, error) {
+		execStart := time.Now()
+		statusCode, body, err := forward(ctx, baseURL, req.WorkspaceId, req.Payload)
+		s.stat.NewTaggedStat("processor_pyt_request_handle_time", stats.TimerType, tags).Since(execStart)
```
fracasula: Wouldn't we emit this even on failures like readiness timeouts or deployer failures? We might want to add a success label to tell them apart.

processor/internal/pytdeployer/pytdeployer.go
```diff
@@ -168,19 +183,31 @@ func loadConfig(conf *config.Config) deployerConfig {
...
 	}
 }
 
+// runIDLength loads the random-suffix length for ephemeral resource names,
+// clamped to what keeps "pyt-test-<suffix>" a valid DNS-1035 label (max 63
+// chars, so the suffix can be at most 54) rather than blocking the processor
+// from starting on an out-of-range value.
+func runIDLength(conf *config.Config) int {
+	const prefixLen = len("pyt-test-")
+	length := conf.GetIntVar(8, 1, "Processor.pytDeployer.pytTestRunIDLength")
```
fracasula: I wonder if we need this configurable at all. A configured value of `1` to `3` is accepted silently (~32–~32³ possible names). Under concurrent load two *different* runs can then generate the same name and we would get `AlreadyExists` (which is assumed as "our own retry succeeded").

This means that both runs share one Deployment/Service. Any thoughts?

mihir20: I have kept the default value of 8, Ideally we should not see collisions on 8 but just to be on a safer side I made this configurable.

fracasula: If you still want it to be configurable then let's make sure it is at least 8 and no less? Wdyt? The problem is misconfiguration here. If you pass 1 somehow then you get a nasty silent bug.

processor/internal/pytdeployer/k8s.go
```diff
@@ -70,15 +68,23 @@ func (d *k8sDeployer) RunOnEphemeral(
...
 	}); err != nil {
-		d.delete(ctx, name)
+		d.delete(ctx, name, workspaceID)
 		return 0, nil, fmt.Errorf("creating ephemeral pyt service %s/%s: %w", d.config.namespace, name, err)
 	}
 	// Delete runs regardless of what happens below — on a readiness timeout,
 	// a forward error, or success alike. It never changes the response
 	// returned to the caller; a failed delete is only logged.
-	defer d.delete(ctx, name)
+	defer d.delete(ctx, name, workspaceID)
```
fracasula: Just move this up and you can delete the repetition in the above if.

processor/internal/pytdeployer/k8s.go
```diff
@@ -70,15 +68,23 @@ func (d *k8sDeployer) RunOnEphemeral(
...
 
-	if err := d.waitReady(ctx, name); err != nil {
+	readyStart := time.Now()
+	err := d.waitReady(ctx, name)
+	// Emitted on both outcomes: the success:false series makes readiness
+	// timeouts visible and clusters at the configured readiness timeout.
+	d.stat.NewTaggedStat("processor_pyt_readiness_wait_time", stats.TimerType, stats.Tags{
```
fracasula, line 83: On a Deployment/Service **`Create` failure** (ephemeral route), `RunOnEphemeral` returns *before* `readyStart`, so `processor_pyt_readiness_wait_time` is **never** emitted, and the  forward closure is never invoked so `processor_pyt_request_handle_time` isn't either.

The real problem though it's that basically if that were to happen, the single
  metric for that failed RPC is `processor_pyt_forward_rpc_time`. This means that there would be no way to tell it apart from a slow success.

fracasula (approved):
Looks mostly good, left another comment on the metrics because I think that it could cause confusion and make things hard to understand when looking at the dashboards.

ktgowtham (approved):

PR #7213, opened 2026-07-24
feat(router): report delivery payload for 296 delivered-with-warning

ItsSudip:

# Description

On the router's success path, `postStatusOnResponseQ` reported the **router input payload** (`JobT.EventPayload`) to reporting for every status code — the success branch hardcoded `payload: inputPayload`, so the transformed body actually sent to the destination was never available to reporting. (The `payload` local derived from `reportJobsdbPayload` is only consumed by the failure branch.)

For the **296 "Delivered with Warning"** outcome, the Events tab needs to show what was actually delivered, not the input event. This PR reports `destinationJob.Message` for 296 only.

**Changes**

- `utils/types`: add `DeliveredWithWarningCode = 296` alongside the existing `FilterEventCode` / `SuppressEventCode` constants, so the status code isn't a magic number.
- `router/worker.go`: in the success branch of `postStatusOnResponseQ`, report `destinationJob.Message` instead of the input payload when the status is 296, and set `payloadStage=delivery` on the status so reporting can tell the sample is the delivery body rather than the input event.
- `router/handle.go`, `router/handle_lifecycle.go`: add `store296DeliveryPayload`, a reloadable per-destType bool (default `true`) mirroring the existing `reportJobsdbPayload`. Flipping it off reverts delivery-payload reporting without disabling the warnings feature — the warning status and the destination's response body are unaffected.

**Scope**

The behaviour change is strictly limited to `respStatusCode == 296`. 200, the filter code (298), and the failure path are unchanged — the new value is computed in a separate `successPayload` local rather than reusing the `reportJobsdbPayload`-driven `payload`, so 200-success behaviour is untouched even when that flag is false.

**Testing**

New `TestPostStatusOnResponseQ296DeliveryPayload` covers: 296 → delivery body + `payloadStage=delivery`; 200 → input payload with no marker; 296 with the kill-switch off → input payload; filter code → still `Filtered` with input payload; empty delivery body on 296 → empty, no panic. `go build ./...`, `go vet`, and the full `TestRouter` suite pass.

## Linear Ticket

https://linear.app/rudderstack/issue/INT-6637/rudder-server-add-delivery-payload-for-success-scenario-in-reporting

## Security

### DIFF
--- a/router/handle.go
+++ b/router/handle.go
@@ -76,6 +76,7 @@ type Handle struct {
 	saveDestinationResponse            bool
 	saveDestinationResponseOverride    config.ValueLoader[bool]
 	reportJobsdbPayload                config.ValueLoader[bool]
+	store296DeliveryPayload            config.ValueLoader[bool]
 
 	diagnosisTickerTime time.Duration
 
--- a/router/handle.go
+++ b/router/handle.go
@@ -76,7 +76,7 @@ type Handle struct {
 	saveDestinationResponse            bool
 	saveDestinationResponseOverride    config.ValueLoader[bool]
 	reportJobsdbPayload                config.ValueLoader[bool]
-	store296DeliveryPayload            config.ValueLoader[bool]
+	storeDeliveredWithWarningPayload   config.ValueLoader[bool]
 
 	diagnosisTickerTime time.Duration
 
--- a/router/handle_lifecycle.go
+++ b/router/handle_lifecycle.go
@@ -123,6 +123,7 @@ func (rt *Handle) Setup(
 	rt.eventOrderHalfEnabledStateDuration = config.GetReloadableDurationVar(10, time.Minute, getRouterConfigKeys("eventOrderHalfEnabledStateDuration", destType)...)
 	rt.deliveryThrottlerTimeout = config.GetReloadableDurationVar(5, time.Minute, getRouterConfigKeys("deliveryThrottlerTimeout", destType)...)
 	rt.reportJobsdbPayload = config.GetReloadableBoolVar(true, getRouterConfigKeys("reportJobsdbPayload", destType)...)
+	rt.store296DeliveryPayload = config.GetReloadableBoolVar(true, getRouterConfigKeys("store296DeliveryPayload", destType)...)
 	rt.saveDestinationResponseOverride = config.GetReloadableBoolVar(false, getRouterConfigKeys("saveDestinationResponseOverride", destType)...)
 
 	statTags := stats.Tags{"destType": rt.destType}
--- a/router/handle_lifecycle.go
+++ b/router/handle_lifecycle.go
@@ -123,7 +123,7 @@ func (rt *Handle) Setup(
 	rt.eventOrderHalfEnabledStateDuration = config.GetReloadableDurationVar(10, time.Minute, getRouterConfigKeys("eventOrderHalfEnabledStateDuration", destType)...)
 	rt.deliveryThrottlerTimeout = config.GetReloadableDurationVar(5, time.Minute, getRouterConfigKeys("deliveryThrottlerTimeout", destType)...)
 	rt.reportJobsdbPayload = config.GetReloadableBoolVar(true, getRouterConfigKeys("reportJobsdbPayload", destType)...)
-	rt.store296DeliveryPayload = config.GetReloadableBoolVar(true, getRouterConfigKeys("store296DeliveryPayload", destType)...)
+	rt.storeDeliveredWithWarningPayload = config.GetReloadableBoolVar(false, getRouterConfigKeys("storeDeliveredWithWarningPayload", destType)...)
 	rt.saveDestinationResponseOverride = config.GetReloadableBoolVar(false, getRouterConfigKeys("saveDestinationResponseOverride", destType)...)
 
 	statTags := stats.Tags{"destType": rt.destType}
--- a/router/worker.go
+++ b/router/worker.go
@@ -984,13 +984,21 @@ func (w *worker) postStatusOnResponseQ(respStatusCode int, destinationJob *types
 		if respStatusCode == utilTypes.FilterEventCode {
 			status.JobState = jobsdb.Filtered.State
 		}
+		// For 296 (Delivered with Warning) report the transformed delivery body actually sent to
+		// the destination, so the warnings UX can surface it. Gated by a per-destType kill-switch;
+		// every other success code keeps reporting the router input payload unchanged.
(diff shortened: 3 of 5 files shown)


router/handle.go
```diff
@@ -76,6 +76,7 @@ type Handle struct {
 	saveDestinationResponse            bool
 	saveDestinationResponseOverride    config.ValueLoader[bool]
 	reportJobsdbPayload                config.ValueLoader[bool]
+	store296DeliveryPayload            config.ValueLoader[bool]
```
achettyiitr: Can we keep the variable something like `storeWarningDeliveryPayload`?

router/handle_lifecycle.go
```diff
@@ -123,6 +123,7 @@ func (rt *Handle) Setup(
 	rt.eventOrderHalfEnabledStateDuration = config.GetReloadableDurationVar(10, time.Minute, getRouterConfigKeys("eventOrderHalfEnabledStateDuration", destType)...)
 	rt.deliveryThrottlerTimeout = config.GetReloadableDurationVar(5, time.Minute, getRouterConfigKeys("deliveryThrottlerTimeout", destType)...)
 	rt.reportJobsdbPayload = config.GetReloadableBoolVar(true, getRouterConfigKeys("reportJobsdbPayload", destType)...)
+	rt.store296DeliveryPayload = config.GetReloadableBoolVar(true, getRouterConfigKeys("store296DeliveryPayload", destType)...)
```
achettyiitr: Let's keep it the default false.

achettyiitr:
```suggestion
	rt.storeWarningDeliveryPayload = config.GetReloadableBoolVar(true, getRouterConfigKeys("storeWarningDeliveryPayload", destType)...)
```

router/worker.go
```diff
@@ -984,13 +984,21 @@ func (w *worker) postStatusOnResponseQ(respStatusCode int, destinationJob *types
 		if respStatusCode == utilTypes.FilterEventCode {
 			status.JobState = jobsdb.Filtered.State
 		}
+		// For 296 (Delivered with Warning) report the transformed delivery body actually sent to
+		// the destination, so the warnings UX can surface it. Gated by a per-destType kill-switch;
+		// every other success code keeps reporting the router input payload unchanged.
+		successPayload := inputPayload
```
achettyiitr: Let's rename it to `responsePayload`, `successPayload` is confusing here?
```suggestion
		responsePayload := inputPayload
```

router/worker.go
```diff
@@ -984,13 +984,21 @@ func (w *worker) postStatusOnResponseQ(respStatusCode int, destinationJob *types
...
+		if respStatusCode == utilTypes.DeliveredWithWarningCode && w.rt.store296DeliveryPayload.Load() {
+			successPayload = destinationJob.Message
+			status.ErrorResponse = misc.UpdateJSONWithNewKeyVal(status.ErrorResponse, "payloadStage", "delivery")
+		}
```
achettyiitr: @itsmihir Do we need a "payloadStage" stage as "router_input" in the else case?

router/worker_test.go
achettyiitr, line 1: Let's mention `storeWarningDeliveryPayload` rather than `296`?

router/worker_test.go
```diff
@@ -55,6 +57,69 @@ func createTestWorker(destType string, transformProxy bool, stat stats.Stats) *w
...
+		got := report(newTestWorker(true), utilTypes.FilterEventCode, deliveryBody)
+		require.JSONEq(t, inputPayload, string(got.payload))
+		require.Equal(t, jobsdb.Filtered.State, got.status.JobState)
+	})
+
+	t.Run("empty delivery body on 296 reports empty without panicking", func(t *testing.T) {
+		got := report(newTestWorker(true), utilTypes.DeliveredWithWarningCode, "")
+		require.Empty(t, string(got.payload))
+		require.Contains(t, string(got.status.ErrorResponse), `"payloadStage":"delivery"`)
+	})
```
achettyiitr: Don't we need to check jobState here? Also ErrorResponse?

achettyiitr: Is this even a valid scenario?

router/handle.go
ItsSudip: Renamed to `storeDeliveredWithWarningPayload` (matching the `DeliveredWithWarningCode` constant rather than `storeWarningDeliveryPayload`). Shout if you'd prefer your exact wording. ✅ ed14f30

router/handle_lifecycle.go
ItsSudip: Done — defaulted to `false`, so delivery-payload reporting is opt-in per destType. ✅ ed14f30

ItsSudip: Applied the rename; kept the default `false` per the comment above. ✅ ed14f30

router/worker.go
ItsSudip: Renamed to `responsePayload`. ✅ ed14f30

router/worker_test.go
ItsSudip: Renamed the test to `TestPostStatusOnResponseQStoreDeliveredWithWarningPayload` and reworded the doc comment to reference the flag instead of 296. ✅ ed14f30

router/worker_test.go
ItsSudip: Removed this case (see the thread on the same line), so this is moot now. ✅ ed14f30

ItsSudip: Agreed — a 296 always carries a delivered body, so `destinationJob.Message` won't be empty, and the "no panic" assertion could not fail regardless. Removed the case. ✅ ed14f30

router/worker.go
```diff
@@ -984,13 +984,21 @@ func (w *worker) postStatusOnResponseQ(respStatusCode int, destinationJob *types
 		if respStatusCode == utilTypes.FilterEventCode {
 			status.JobState = jobsdb.Filtered.State
 		}
+		// For 296 (Delivered with Warning) report the transformed delivery body actually sent to
+		// the destination, so the warnings UX can surface it. Gated by a per-destType opt-in flag
+		// (default off); every other success code keeps reporting the router input payload unchanged.
+		responsePayload := inputPayload
```
itsmihir: responsePayload -> payload

router/worker.go
itsmihir: Good to have, don't strictly need it as we don't surface it in UI. We can skip it for now and just add a comment about it.

achettyiitr (approved):

itsmihir (approved):

PR #7217, opened 2026-07-24
feat(router): gate 296 delivered-with-warning behind per-workspace rollout

ItsSudip:

# Description

This branch merged `release/1.81.x` before being squash-merged, so this single commit carries **two** related pieces of the "Delivered with Warning" work:

- **INT-6636** — gate `296 → 200` behind a per-workspace rollout (the primary subject of this PR).
- **INT-6637** — report the transformed delivery payload for 296 (originally #7213 / #7226; its code rode in with this merge).

## INT-6636 — gate 296 → 200

rudder-transformer converts Braze's `200 + warning` response into statusCode **296**. Honouring that unconditionally would let any destination surface 296 — including via handler bugs or destinations not yet onboarded — so rudder-server downgrades `296 → 200` unless the feature is enabled for that job.

**Enablement — the OR of two inputs, both defaulting to off**

| Input | Where | Purpose |
| -- | -- | -- |
| `supportsDeliveredWithWarnings` | `DestinationDefinition.Config`, read at `Setup` and refreshed by the backend-config subscriber | GA switch — enables the feature for every workspace at once |
| `deliveredWithWarningsEnabledWorkspaceIDs` | `Router.<destType>.deliveredWithWarningsEnabledWorkspaceIDs`, reloadable | Controlled-rollout allow-list — enables selected workspaces ahead of GA |

Both empty ⇒ downgrade, so un-onboarded destinations fail closed. `Handle.deliveredWithWarningsEnabled(workspaceID)` is the single place the OR is expressed.

**Changes**

- `utils/types`: add `DeliveredWithWarningCode = 296`.
- `router/handle.go`: `supportsDeliveredWithWarnings` (`atomic.Bool`), the workspace allow-list predicate, the `deliveredWithWarningsEnabled` OR helper, and `statusDowngradedStat` — a `func(from, to int) stats.Counter` constructor for the generic `router_status_downgraded_count{destType,from,to}` metric.
- `router/handle_lifecycle.go`: read the dest-def flag at both dest-def read sites, register the allow-list (mirroring `orderingDisabledWorkspaceIDs`), and wire the downgrade-metric constructor at `Setup`.
- `router/worker.go`: `gateDeliveredWithWarning`, called from `prepareRouterJobResponses` immediately after `hydrateRespStatusCodes` — the single funnel every delivery path (transformer proxy, direct HTTP, custom destination manager) converges on, upstream of `JobStatusT.ErrorCode` and reporting.

**Scope & safety** — only `296` is rewritten; 200/4xx/5xx pass through untouched, response bodies byte-identical. The rewrite-to-200 makes it idempotent across duplicate job IDs, so `router_status_downgraded_count` fires once per job.

## INT-6637 — report delivery payload for 296

On the success path, `postStatusOnResponseQ` reported the **router input payload** for every status code, so the transformed body actually sent to the destination was never available to reporting. For **296** the Events tab needs that delivery body, so this reports `destinationJob.Message` for 296 only and marks the status `payloadStage=delivery`.

- Gated by `storeDeliveredWithWarningPayload` — a reloadable per-destType bool, **default `false`** (opt-in), mirroring `reportJobsdbPayload`.
- Strictly limited to `respStatusCode == 296`; 200, filter (298), and the failure path are unchanged.

**Testing** — `TestDeliveredWithWarningsEnabled` (enablement matrix incl. fail-closed), `TestGateDeliveredWithWarning` (pass-through, downgrade + counter, mixed-workspace batch, mixed-code batch, duplicate-job idempotency), and `TestPostStatusOnResponseQStoreDeliveredWithWarningPayload` (296 → delivery body, 200 → input, flag-off → input, filter → Filtered). `go build ./...`, `go vet`, and the full `TestRouter` suite pass.

## Linear Tickets

- INT-6636: https://linear.app/rudderstack/issue/INT-6636/rudder-server-add-gate-logic-to-update-296-status-code-to-200-when
- INT-6637: https://linear.app/rudderstack/issue/INT-6637/rudder-server-add-delivery-payload-for-success-scenario-in-reporting

## Security

### DIFF
--- a/router/handle.go
+++ b/router/handle.go
@@ -76,6 +76,9 @@ type Handle struct {
 	saveDestinationResponse            bool
 	saveDestinationResponseOverride    config.ValueLoader[bool]
 	reportJobsdbPayload                config.ValueLoader[bool]
+	// supportsDeliveredWithWarnings mirrors the destination definition's capability flag. Written
+	// by the backend-config subscriber and read by workers, hence atomic.
+	supportsDeliveredWithWarnings atomic.Bool
 
 	diagnosisTickerTime time.Duration
 
@@ -107,6 +110,7 @@ type Handle struct {
 	processJobsCountStat           stats.Measurement
 	throttlingErrorStat            stats.Measurement
 	throttledStat                  stats.Measurement
+	warningStatusDowngradedStat    stats.Measurement
 	isolationStrategy              isolation.Strategy
 	backgroundGroup                *errgroup.Group
 	backgroundCtx                  context.Context
@@ -117,6 +121,10 @@ type Handle struct {
 
 	eventOrderingDisabledForWorkspace   func(workspaceID string) bool
 	eventOrderingDisabledForDestination func(destinationID string) bool
+
+	// deliveredWithWarningsEnabledForWorkspace reports whether a workspace is on the
+	// delivered-with-warnings controlled-rollout allow-list.
+	deliveredWithWarningsEnabledForWorkspace func(workspaceID string) bool
 
 	limiter struct {
 		pickup    kitsync.Limiter
@@ -134,6 +142,18 @@ type Handle struct {
 	drainer              routerutils.Drainer
 	drainingPartitionsMu sync.RWMutex
 	drainingPartitions   map[string]struct{} // keeps track of router partitions which are currently draining
+}
+
+// deliveredWithWarningsEnabled reports whether a 296 (Delivered with Warning) status should be
+// honoured for a job in the given workspace — either the destination definition advertises support
+// (general availability), or the workspace is on the controlled-rollout allow-list. Fails closed
+// when the allow-list predicate has not been wired, preserving the default-deny guarantee.
+func (rt *Handle) deliveredWithWarningsEnabled(workspaceID string) bool {
+	if rt.supportsDeliveredWithWarnings.Load() {
+		return true
+	}
+	return rt.deliveredWithWarningsEnabledForWorkspace != nil &&
+		rt.deliveredWithWarningsEnabledForWorkspace(workspaceID)
 }
 
 // activePartitions returns the list of active partitions, depending on the active isolation strategy
--- a/router/handle.go
+++ b/router/handle.go
@@ -110,7 +110,7 @@ type Handle struct {
 	processJobsCountStat           stats.Measurement
 	throttlingErrorStat            stats.Measurement
 	throttledStat                  stats.Measurement
-	statusDowngradedStat           stats.Counter
+	statusDowngradedStat           func(from, to int) stats.Counter
 	isolationStrategy              isolation.Strategy
 	backgroundGroup                *errgroup.Group
 	backgroundCtx                  context.Context
--- a/router/handle.go
+++ b/router/handle.go
@@ -76,6 +76,7 @@ type Handle struct {
 	saveDestinationResponse            bool
 	saveDestinationResponseOverride    config.ValueLoader[bool]
 	reportJobsdbPayload                config.ValueLoader[bool]
+	storeDeliveredWithWarningPayload   config.ValueLoader[bool]
 
 	diagnosisTickerTime time.Duration
 
--- a/router/handle.go
+++ b/router/handle.go
@@ -110,7 +110,7 @@ type Handle struct {
 	processJobsCountStat           stats.Measurement
 	throttlingErrorStat            stats.Measurement
 	throttledStat                  stats.Measurement
-	warningStatusDowngradedStat    stats.Measurement
+	statusDowngradedStat           stats.Counter
 	isolationStrategy              isolation.Strategy
 	backgroundGroup                *errgroup.Group
 	backgroundCtx                  context.Context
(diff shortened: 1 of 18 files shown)


router/handle.go
```diff
@@ -107,6 +110,7 @@ type Handle struct {
 	processJobsCountStat           stats.Measurement
 	throttlingErrorStat            stats.Measurement
 	throttledStat                  stats.Measurement
+	warningStatusDowngradedStat    stats.Measurement
```
achettyiitr:
```suggestion
	warningStatusDowngradedStat    stats.Counter
```

router/handle_lifecycle.go
```diff
@@ -133,6 +136,7 @@ func (rt *Handle) Setup(
 	rt.routerTransformInputCountStat = stats.Default.NewTaggedStat("router_transform_num_input_jobs", stats.CountType, statTags)
 	rt.routerTransformOutputCountStat = stats.Default.NewTaggedStat("router_transform_num_output_jobs", stats.CountType, statTags)
 	rt.batchInputOutputDiffCountStat = stats.Default.NewTaggedStat("router_batch_input_output_diff_jobs", stats.CountType, statTags)
+	rt.warningStatusDowngradedStat = stats.Default.NewTaggedStat("router_warning_status_downgraded_count", stats.CountType, statTags)
```
achettyiitr: Can we keep this generic? statusDowngradedStat, where we can also send from and to as well.

router/handle.go
```diff
@@ -136,6 +144,18 @@ type Handle struct {
...
 
+// deliveredWithWarningsEnabled reports whether a 296 (Delivered with Warning) status should be
+// honoured for a job in the given workspace — either the destination definition advertises support
+// (general availability), or the workspace is on the controlled-rollout allow-list. Fails closed
+// when the allow-list predicate has not been wired, preserving the default-deny guarantee.
+func (rt *Handle) deliveredWithWarningsEnabled(workspaceID string) bool {
+	if rt.supportsDeliveredWithWarnings.Load() {
+		return true
+	}
+	return rt.deliveredWithWarningsEnabledForWorkspace != nil &&
```
achettyiitr: Why will it be null? It will be initialised in the setup?

router/handle.go
```diff
@@ -136,6 +144,18 @@ type Handle struct {
...
+		rt.deliveredWithWarningsEnabledForWorkspace(workspaceID)
+}
```
achettyiitr, line 154: If it is enabled, we would need to check in the workspace; otherwise, return false.

router/handle.go
ItsSudip: Done — `statusDowngradedStat` is now typed `stats.Counter`. ✅ 3e41a5f

router/handle_lifecycle.go
ItsSudip: Made it generic: renamed to `statusDowngradedStat` / `router_status_downgraded_count` with `from` and `to` tags (296 → 200 for this gate), built once in `Setup` — reusable by name + tag schema for any future status-code downgrade.

It's currently built once with fixed `from`/`to`, so recording a *different* downgrade at runtime would need its own handle. If you'd prefer it fully runtime-variable now, I can switch to a `cache.StatsCache` keyed by `{destType, from, to}` (same pattern as `deliveryLatencyStatsCache`) — let me know your preference. ✅ 3e41a5f

router/handle.go
ItsSudip: Dropped the nil guard — `deliveredWithWarningsEnabledForWorkspace` is always wired in `Setup`, as you said. ✅ 3e41a5f

router/handle.go
ItsSudip: Kept it as an OR and simplified to `return rt.supportsDeliveredWithWarnings.Load() || rt.deliveredWithWarningsEnabledForWorkspace(workspaceID)`.

The two inputs are independent: the dest-def `supportsDeliveredWithWarnings` is the GA switch (enables all workspaces at once), and the workspace allow-list is for gradual rollout *before* GA — i.e. while the dest-def flag is still false. So it can't be AND: under AND the allow-list would have no effect until GA flips the dest-def flag on. Shout if you intended something different. ✅ 3e41a5f

router/handle_lifecycle.go
achettyiitr: You can make it generic, by keeping it as a function?

```

	rt.statusDowngradedStat = func(from, to int) stats.Counter {
		return stats.Default.NewTaggedStat("router_status_downgraded_count", stats.CountType, stats.Tags{
			"destType": rt.destType,
			"from":     strconv.Itoa(from),
			"to":       strconv.Itoa(to),
		})
	}
```

ItsSudip: Done — went with your function form: `statusDowngradedStat func(from, to int) stats.Counter`, built in `Setup`, and the gate calls `w.rt.statusDowngradedStat(utilTypes.DeliveredWithWarningCode, http.StatusOK).Count(1)`. ✅ 070838e

achettyiitr (approved):

itsmihir (approved):

PR #7226, opened 2026-07-28
feat(router): report delivery payload for 296 delivered-with-warning

ItsSudip:

# Description

On the router's success path, `postStatusOnResponseQ` reported the **router input payload** (`JobT.EventPayload`) to reporting for every status code — the success branch hardcoded `payload: inputPayload`, so the transformed body actually sent to the destination was never available to reporting.

For the **296 "Delivered with Warning"** outcome, the Events tab needs to show what was actually delivered, not the input event. This PR reports `destinationJob.Message` for 296 only.

**Changes**

- `utils/types`: add `DeliveredWithWarningCode = 296` alongside the existing `FilterEventCode` / `SuppressEventCode` constants, so the status code isn't a magic number.
- `router/worker.go`: in the success branch of `postStatusOnResponseQ`, report `destinationJob.Message` instead of the input payload when the status is 296, and set `payloadStage=delivery` on the status so reporting can tell the sample is the delivery body rather than the input event. Non-296 success responses are unchanged (no `payloadStage` marker — noted in a comment).
- `router/handle.go`, `router/handle_lifecycle.go`: add `storeDeliveredWithWarningPayload`, a reloadable per-destType bool (**default `false`**, opt-in) mirroring the existing `reportJobsdbPayload`. Turning it on enables delivery-payload reporting for 296; turning it off reverts it without affecting the warning status or the destination's response body.

**Scope**

The behaviour change is strictly limited to `respStatusCode == 296`. 200, the filter code (298), and the failure path are unchanged.

**Testing**

`TestPostStatusOnResponseQStoreDeliveredWithWarningPayload` covers: 296 → delivery body + `payloadStage=delivery`; 200 → input payload with no marker; 296 with the flag off → input payload; filter code → still `Filtered` with the input payload. `go build ./...`, `go vet`, and the full `TestRouter` suite pass.

## Linear Ticket

https://linear.app/rudderstack/issue/INT-6637/rudder-server-add-delivery-payload-for-success-scenario-in-reporting

## Security


ItsSudip:
Superseded — the INT-6637 changes already landed in master via #7217's squash-merge (1d772a182), so this PR has no remaining diff to merge. Closing.

PR #7235, opened 2026-07-30
feat(processor): fork event processing for rsources pipelines

atzoum:
# Description

Closes the rsources accounting gap #7207 left open: forking siphons an event to the intermediate `proc` jobsdb before it's transformed, but until now nothing recorded that in rsources until the proc pool eventually drained it. An arbitrarily later, unbounded window in which a retl `jobRunId` could be reported `completed` while its forked events were still sitting unaccounted-for in `proc`. That's why forking was hard-excluded for any event carrying a `SourceJobRunID`.

**The mechanism** 
No new stage is introduced, it piggybacks on the existing source-level `In`/`Out` the same way dropped-job accounting already does (`JobsDropped`). `JobsForked` increments `In` by the number of consumers a job is forked to (e.g. one job forked to 2 destinations bumps `In` by 2), committed in the same transaction as the `proc` write itself so there's no window where the fork is durable but unaccounted-for. `Out` is unaffected: the proc pool's own `rsources.StatsCollector` (`proc_partition_worker.go`) already calls `CollectStats` once per (job, consumer) as each forked job drains to a terminal status. A source now only reports `completed` once every forked consumer has drained, in addition to the existing per-destination router/batch-router accounting, which is untouched.

**Still opt-in** 
Forking retl-tracked events is gated by its own config, disabled by default:

```
Processor.DestinationIsolation.forkRsourcesTrackedJobs
```

### Verification

A new integration test `TestProcessorRsourcesStats` covers both the forked and non-forked paths end to end against the rsources job-status API:
  - `inline`- baseline, isolation off.
  - `retl_excluded` - retl events still stay inline and the numbers match `inline` exactly.
  - `all_forked` - every destination is forked, proving the fork `In` balances against the proc consumer's drained `Out`.
  - `mixed` - a destination that's both forked *and* filtered by a user transformation (a proc-consumer status update with nothing ever reaching the router), and another one is both forked *and* aborted at the router.

Each scenario asserts exact `In`/`Out`/`Failed` at source and destination level, `Completed` via `/internal/v2/job-status`, failed-records resolution back to the sent `recordId` for aborted destinations, jobsdb drain counts, and webhook delivery counts.

## Linear Ticket

resolves PIPE-3055

## Security

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -204,6 +204,7 @@ type Handle struct {
 		pythonTransformConfig                     transformerutils.PythonTransformConfig
 		userTransformationMirroringBlockedIDs     config.ValueLoader[[]string]
 		storeSamplerEnabled                       config.ValueLoader[bool]
+		forkRsourcesTrackedJobs                   bool
 	}
 
 	drainConfig struct {
@@ -827,6 +828,7 @@ func (proc *Handle) loadConfig() {
 	// GWCustomVal is used as a key in the jobsDB customval column
 	proc.config.GWCustomVal = proc.conf.GetStringVar("GW", "Gateway.CustomVal")
 	proc.config.pythonTransformConfig = transformerutils.LoadPythonTransformConfig(proc.conf)
+	proc.config.forkRsourcesTrackedJobs = proc.conf.GetBoolVar(false, "Processor.DestinationIsolation.forkRsourcesTrackedJobs")
 	proc.loadReloadableConfig(defaultPayloadLimit, defaultMaxEventsToProcess)
 }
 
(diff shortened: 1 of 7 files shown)


processor/processor.go
```diff
@@ -3003,6 +3004,9 @@ func (proc *Handle) storeStage(partition string, pipelineIndex int, in *storeMes
 		}
 	}
 	in.rsourcesStats.CollectStats(statusList)
+	// forkedJobs are only ever populated in the gw pool (the proc pool never re-forks),
+	// so this is a no-op when storeStage is invoked via procStoreStage.
+	in.rsourcesStats.JobsForked(in.forkedJobs)
```
mihir20, line 3009: Will this mess up the stats for Rsources?  
For this scenario, an event is ingested from a source connected to 2 destinations. Earlier we would have reported in = 1 and out = 1 for source-level stats. but now we will report in = 3 and out = 3

mihir20 (approved):

processor/processor.go
atzoum: Yup, but in stats are already messed through filtered events. As discussed if we really want to have reliable ingestion stats we would need to tackle it in another way

ktgowtham (approved):

PR #7239, opened 2026-07-31
chore: pyt contract tests for transformations auth

fracasula:
# Description

See [this PR](https://github.com/rudderlabs/rudder-pytransformer/pull/117).

Contract tests for how `rudder-pytransformer` talks to the config backend: hosted-secret
authentication and redirect responses. Tests only — no production code changes in this repo.

Both behaviours matter to rudder-server specifically because they decide whether a config backend misconfiguration is **retryable** or **terminal**. rudder-server aborts any per-event status that is not 200/298, so a terminal answer here means every event for every Python transformation is destroyed with nothing to replay from.

## Config backend authentication — `config_backend_auth_test.go`

Pins both variants of pytransformer's `CONFIG_BACKEND_HOSTED_SECRET`:

- **Unset** (today's deployments, and every self-hosted one) — the request must be unchanged from the build that didn't know about the variable. Asserted by inspecting what the config backend *received*: no `Authorization` header at all. A build sending `Basic Og==` (base64 of `":"`) would pass a "does it still work" check against a permissive backend and 401 against a strict one; only looking at the received header separates the two.
- **Set** — the header is exactly `Basic base64("<secret>:")`, on the transformation fetch *and* the library fetch, and it authenticates against a backend enforcing the real check. The empty password is load-bearing: the config backend rejects any decoded token that doesn't hold
  exactly one colon.
- - **Wrong secret** (stale after a rotation) → 503 + `X-Rudder-Should-Retry` + `X-Rudder-Error-Reason: config_backend_auth_failed`, so events are held rather than aborted. Same for an anonymous fetch against an authenticated backend; a blocked public route (403) reports `config_backend_forbidden` so a bad secret and a missed repoint page differently.
- **Trailing newline** in the secret is trimmed — the shape it actually arrives in when mounted from a k8s secret file.

The mock config backend is a **port of `rudder-config-backend`**, not an invention: `RudderKoaBasicAuth` (including the exactly-one-colon rule and the non-short-circuiting constant-time comparison), `hostedSecretConfig`'s comma-split/trim/drop-blanks parsing with empty passwords, `verifyHostedDataPlaneSecret` collapsing every rejection into one 401, and `blockHostedPublicAccess`'s 403. Provenance is listed file-by-file in the mock's doc comment.

`TestConfigBackendAuthMockMatchesConfigBackend` pins the port itself against cases taken from rudder-config-backend's own suites, and runs without containers — the container tests are only worth their runtime if the thing they authenticate against enforces the real rules.

## Config backend redirects — `redirects_test.go`

pytransformer does not follow redirects, so a 3xx from a misconfigured proxy raises `ConfigBackendRedirectError` → 503 + retry headers.

`TestConfigBackendRedirectIsRetriedNotDropped` is the regression test for the data loss: it holds the config backend broken through a full rudder-server pipeline, asserts nothing is aborted, then repairs it and shows the *original* events come out the far end transformed. Under the previous terminal behaviour those events were gone. It also asserts the redirect target is never contacted.

## Running

These need a `rudder-pytransformer` image; build it from that repo with `make build-ecr-latest`,
then:

```bash
go test -v -count 1 -race ./integration_test/pytransformer_contract -timeout=10m
```

## Linear Ticket

< Fixes [PIPE-3229](https://linear.app/rudderstack/issue/PIPE-3229/pyt-config-backend-authentication) >

## Security

### DIFF
--- a/integration_test/pytransformer_contract/bc_helpers_test.go
+++ b/integration_test/pytransformer_contract/bc_helpers_test.go
@@ -248,6 +248,24 @@ func makeEventWithCredentials(messageID, versionID string, credentials []types.C
 	return ev
 }
 
+// makeEvents creates n TransformerEvents for versionID, optionally carrying library version ids.
+//
+// Message ids are prefixed with versionID so a test sharing one mock config backend across
+// subtests can scope its assertions to its own requests.
+func makeEvents(versionID string, n int, libraryVersionIDs ...string) []types.TransformerEvent {
+	libraries := make([]backendconfig.LibraryT, len(libraryVersionIDs))
+	for i, id := range libraryVersionIDs {
+		libraries[i] = backendconfig.LibraryT{VersionID: id}
+	}
+
+	events := make([]types.TransformerEvent, n)
+	for i := range events {
+		events[i] = makeEvent(fmt.Sprintf("%s-msg-%d", versionID, i+1), versionID)
+		events[i].Libraries = libraries
+	}
+	return events
+}
+
 // configBackendEntry controls what the mock config backend returns for a given versionId.
 //
 // When statusCode is 0 (default), the entry is treated as a normal transformation:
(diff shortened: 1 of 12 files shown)


ktgowtham (approved):
test are failing. pl check

fracasula:
> test are failing. pl check

@ktgowtham As usual they should pass once we merge the pyt PR (the one with the secret support) and then create a release. Then these tests will pull `main` and should pass :crossed_fingers: (they do pass locally in fact).

